PR #4, opened 2015-09-07
Replace mailgun

aryszka:


etcd/etcd_test.go
```diff
@@ -90,7 +90,7 @@ func checkInitial(rd skipper.RawData) bool {
 		return false
 	}
 
-	if r.MatchExp != "PathRegexp(`.*\\.html`)" {
+	if len(r.PathRegexps) != 1 || r.PathRegexps[0] != ".*\\.html" {
```
lmineiro: I know what you did there :)

requestmatch/matcher.go
```diff
@@ -0,0 +1,362 @@
...
+		return ""
+	}
+
+	// clip '/*' and return only the name
+	return param[2:]
+}
+
+// Constructs a Matcher based on the provided definitions. If `ignoreTrailingSlash`
+// is true, the Matcher handles paths with or without a trailing slash equally.
+func Make(ds []Definition, ignoreTrailingSlash bool) (*Matcher, []*DefinitionError) {
```
lmineiro: I've been seeing a lot more such methods being exported as `New...`. Isn't it more idiomatic? Like errors.New(), hash.New(), bytes.NewBuffer()?

aryszka: you're right. I also thought of it. Since now this branch is about the mailgun replacement, I think we can leave it like this. I already have a refactoring branch, and I will add this to the todos for that.

As the idiomatism goes and how I noticed recently, Make* is rarelier used and is usually applied to struct values only, recalling the builtin make, while New* is more often used, and I find it now more appropriate to pointer values, too. This aligns with the fact, that we have an almost never used builtin `new`, where new(Type) is equivalent to &Type{}.

Anyway, let's let this in like this, and then I'll fix this as a refactoring effort.

requestmatch/matcher_test.go
```diff
@@ -0,0 +1,1193 @@
+package requestmatch
```
lmineiro: Should we get rid of all the zalan.do's?

hjacobs: it's definitely not nice to have "real" URLs in the test :-P

requestmatch/matcher_test.go
```diff
@@ -0,0 +1,1193 @@
...
+const (
+	defsCountPhase1 = 1
+	defsCountPhase2 = 100
+	defsCountPhase3 = 10000
+	defsCountPhase4 = 1000000
+)
+
+// defs in eskip format
+const routeDoc = `
+    header: Path("/tessera/header") -> "https://header.mop-taskforce.zalan.do";
```
hjacobs: search & replace "mop-taskforce.zalan.do" with "example.org"!

requestmatch/matcher_test.go
aryszka: yeah... probably. on the long term example.org is better.

requestmatch/matcher_test.go
aryszka: ok, deal.
and thanks for the extra pair of eyes :)

lmineiro:
Apart from that, lgtm

requestmatch/matcher.go
lmineiro: (y)

aryszka:
pushing the domain fix in a second

aryszka:
ok, domain names should be removed

lmineiro:
lgtm

PR #7, opened 2015-09-11
Innkeeper data client

danpersa:


filters/redirect/redirect_test.go
```diff
@@ -24,3 +25,26 @@ func TestRedirect(t *testing.T) {
 		t.Error("invalid location")
 	}
 }
+
+func TestRedirectRelative(t *testing.T) {
+	spec := &Redirect{}
+	f, err := spec.MakeFilter("redirect0", []interface{}{float64(302), "/relative/url"})
```
mo-gr: float64?

lmineiro: http.StatusFound ?

innkeeper/client.go
```diff
@@ -0,0 +1,417 @@
...
+
+	req.Header.Add("Authorization", c.authToken)
+	response, err := c.httpClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	defer response.Body.Close()
+
+	if response.StatusCode == 401 {
```
lmineiro: http.StatusUnauthorized

innkeeper/client.go
```diff
@@ -0,0 +1,417 @@
...
+
+		err = c.authenticate()
+		if err != nil {
+			return nil, err
+		}
+
+		return c.requestData(false, url)
+	}
+
+	if response.StatusCode >= 400 {
```
lmineiro: Mixed feeling about leaving this or actually switch to the first matching constant - http.StatusBadRequest

innkeeper/client_test.go
```diff
@@ -0,0 +1,718 @@
...
+	}
+
+	return "", errors.New(string(authErrorAuthentication))
+}
+
+type innkeeperHandler struct{ data []*routeData }
+
+func (h *innkeeperHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Header.Get("Authorization") != testAuthenticationToken {
+		w.WriteHeader(401)
```
lmineiro: http.StatusUnauthorized

innkeeper/client_test.go
```diff
@@ -0,0 +1,718 @@
...
+
+	c, err := Make(Options{api.URL, false, pollingTimeout, autoAuth(true)})
+	if err != nil {
+		t.Error(err)
+		return
+	}
+
+	defer c.Close()
+
+	// recieve initial
```
lmineiro: typo - receive

innkeeper/client_test.go
```diff
@@ -0,0 +1,718 @@
...
```
lmineiro: Same typo

oauth/oauth.go
```diff
@@ -0,0 +1,123 @@
...
+
+	return ar.AccessToken, nil
+}
+
+func getAuthPostBody(us *userCredentials) string {
+	parameters := url.Values{}
+	parameters.Add("grant_type", grantType)
+	parameters.Add("username", us.Username)
+	parameters.Add("password", us.Password)
+	parameters.Add("scope", "uid fashion_store_route.read")
```
lmineiro: please make this configurable

oauth/oauth_test.go
```diff
@@ -0,0 +1,173 @@
...
+	id, secret, _ := r.BasicAuth()
+	check(id != "theclientid")
+	check(secret != "clientsecret")
+
+	checkForm("grant_type", grantType)
+	checkForm("username", "appusername")
+	checkForm("password", "apppassword")
+	checkForm("scope", "uid fashion_store_route.read")
+
+	w.WriteHeader(200)
```
lmineiro: http.StatusOK ?

oauth/oauth_test.go
```diff
@@ -0,0 +1,173 @@
...
+	enc := json.NewEncoder(w)
+
+	// ignore error
+	enc.Encode(&authResponse{AccessToken: testToken})
+})
+
+var failureHandler http.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	w.WriteHeader(404)
```
lmineiro: http.StatusNotFound

filters/redirect/redirect.go
```diff
@@ -43,7 +44,21 @@ func (f *Redirect) Request(ctx skipper.FilterContext) {}
 
 func (f *Redirect) Response(ctx skipper.FilterContext) {
 	w := ctx.ResponseWriter()
-	w.Header().Set("Location", f.location)
+
+	u, err := url.Parse(f.location)
+
+	if err != nil {
+		w.WriteHeader(501)
```
lmineiro: http.StatusNotImplemented

filters/redirect/redirect_test.go
```diff
@@ -24,3 +25,26 @@ func TestRedirect(t *testing.T) {
...
+	}
+
+	request, _ := http.NewRequest("GET", "https://example.org/some/url", nil)
+
+	ctx := &mock.FilterContext{
+		FResponseWriter: httptest.NewRecorder(),
+		FRequest:        request}
+	f.Response(ctx)
+
+	if ctx.FResponseWriter.(*httptest.ResponseRecorder).Code != 302 {
```
lmineiro: http.StatusFound

filters/redirect/redirect_test.go
aryszka: float64: most forgiving number type. checked earlier, it doesn't cause issues during int conversions. Also, see: http://localhost:8989/pkg/encoding/json/#Unmarshal

aryszka: StatusFound: right, will fix those. (however:

```
grep -rn --exclude="*_test.go" '200\|404\|StatusCode' $GOROOT/src/net/http | grep -v ^Binary | less
```
;))

innkeeper/client.go
```diff
@@ -0,0 +1,417 @@
...
+	ae := apiError{}
+	if err := json.Unmarshal(message, &ae); err != nil {
+		return "", err
+	}
+
+	return ae.ErrorType, nil
+}
+
+func isApiAuthError(error string) bool {
+	switch authErrorType(error) {
```
mo-gr: A switch to either return true or false doesn't feel right... why not somethin like `return error == a || error == b`?

innkeeper/client.go
aryszka: Hey Luis, what's your recommended approach? What we just quickly discussed here, is that comparing `>= http.StatusBadRequest` is fine, we can distinguish between 40x and 50x, but in general, it is fine. I'll can replace the num literal with the constant form, of course. How should this look in your opinion?

innkeeper/client_test.go
aryszka: yup, thanks

innkeeper/client_test.go
aryszka: thanks again

oauth/oauth.go
aryszka: good catch, i have missed this one

innkeeper/client.go
mo-gr: Maybe wrap it in a function isError(response.statusCode). But that function would, be implemented the same way. Checking for 400 <= x < 600

innkeeper/client.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	return exps
+}
+
+func getMatcherExpression(d *routeData) string {
+	m := []string{}
+
+	// there can be only 0 or 1 here, because routes for multiple methods
+	// have been already split
```
mo-gr: does it make sense to assert that and panic otherwise?

innkeeper/client.go
```diff
@@ -0,0 +1,417 @@
...
+	if len(c.doc) == 0 {
+		url := c.createUrl(allRoutesPathRoot)
+		d, err = c.requestData(true, url)
+	} else {
+		url := c.createUrl(updatePathRoot, c.lastModified)
+		d, err = c.requestData(true, url)
+	}
+
+	if err != nil {
+		log.Println("error while receiving innkeeper data;", err)
```
mo-gr: `;`?

oauth/oauth.go
mo-gr: we are in skipper, this should not even be here... imho blocker to merge

filters/redirect/redirect.go
aryszka: uhh, yeah... this is left open and i just forgot to ask it in the end. shall we return 50x errors, or only 404 in these cases? some people don't return 50x to avoid giving tips for attackers. what shall be our approach?

innkeeper/client.go
aryszka: right

innkeeper/client.go
aryszka: it is true that this is fishy, but actually i can create a new field for the singular method, and then it is ensured in compile time. good catch.

innkeeper/client.go
aryszka: right, no need for that

innkeeper/client.go
lmineiro: @aryszka as I said, I have mixed feelings. I like @mo-gr approach. Even if we keep the hardcoded value inside. Becomes really more readable of the intention is

oauth/oauth.go
lmineiro: Agree

innkeeper/client.go
aryszka: wow! how lucky that this was caught: the splitting on methods was just never created :o

filters/redirect/redirect.go
lmineiro: So, this is supposed to be a scenario of the redirect filter which contained an invalid destination URL? If it was, why wasn't it parsed at load time and rejected if route was invalid? Arguably, Inkeeper could've rejected it also...

innkeeper/client_test.go
```diff
@@ -0,0 +1,601 @@
...
+	exps := strings.Split(doc, ";")
+	for i, exp := range exps {
+		exps[i] = strings.Trim(exp, " \n")
+	}
+	sort.Strings(exps)
+	return strings.Join(exps, ";")
+}
+
+func checkDoc(out skipper.RawData, in []*routeData) bool {
+	c := &Client{doc: make(map[int64]map[string]string)}
```
lmineiro: Should we define a type alias for this? I've seen it somewhere else. Maybe this is my C-troll talking...

aryszka: good point

PR #9, opened 2015-10-02
Release cleanup

aryszka:
one (biggest) step of a refactoring. not making a pr directly to master, but only into the review branch of release-cleanup-review. the whole refactoring is not yet done, but it is worth reviewing in separate steps. in this step i removed the common, empty interface package.


filters/filters.go
```diff
@@ -1,38 +1,59 @@
...
+// Typically, there is a single Spec instance of each implementation in a running process, which can create multiple filter
+// instances with different config defined in the configuration on every update.
+type Spec interface {
+
+	// The name of the Spec is used to identify in the configuration which spec a filter is based on.
+	Name() string
+
+	// When the program settings are updated, and they contain filters based on a spec, MakeFilter is
+	// called, and the filter id and the filter specific settings are provided. Returns a filter.
+	CreateFilter(config []interface{}) (Filter, error)
```
lmineiro: nitpicking here, but I really like the "New...()" naming convention. It seems more idiomatic to me

danpersa: In the comment is specified MakeFilter, but the method is called CreateFilter

filters/headerfilter.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (spec *headerFilter) CreateFilter(config []interface{}) (Filter, error) {
+	key, value, err := headerFilterConfig(config)
+	return &headerFilter{typ: spec.typ, key: key, value: value}, err
+}
+
+func (f *headerFilter) Request(ctx FilterContext) {
+	if f.typ == requestHeader {
+		req := ctx.Request()
+		if f.key == "Host" {
```
lmineiro: Should it be case insensitive? http://tools.ietf.org/html/rfc7230#section-3.2 already states they should. I wouldn't expect the filter to force it

filters/redirect.go
```diff
@@ -16,21 +15,21 @@ type Redirect struct {
...
-	if len(c) != 2 {
+	if len(config) != 2 {
 		return invalidArgs()
 	}
 
-	code, ok := c[0].(float64)
+	code, ok := config[0].(float64)
 	if !ok {
 		return invalidArgs()
 	}
```
lmineiro: Should we test the status code range and fail if not ok?

filters/filters.go
aryszka: yup:
New: i agree and it is in the pipe to fix it globally, as part of this same refactor. this pull request is only about the removal of the empty interface package.
Comment: it is also the part of this refactor, to go through the whole documentation, and fix/extend/update it. As reviewers, do you prefer to keep the doc continuously in sync?

filters/headerfilter.go
aryszka: Right! Fixing it.

filters/redirect.go
aryszka: i think in this case we can save that code and leave to the audience whether they want to do something irregular.

filters/filters.go
lmineiro: This being a PR to another branch that will get reviewed I believe it should be fine. The final PR for the destination branch should contain the docs updates

PR #11, opened 2015-10-05
Release cleanup

aryszka:
- moving the eskip package into skipper
- simplifying the route definitions used around the program
- enabling path params for filters
- style changes


routing/matcher.go
```diff
@@ -1,111 +1,335 @@
...
+		return ""
+	}
+
+	// clip '/*' and return only the name
+	return param[2:]
+}
+
+// Constructs a Matcher based on the provided definitions. If `ignoreTrailingSlash`
+// is true, the Matcher handles paths with or without a trailing slash equally.
+func newMatcher(rs []*Route, ignoreTrailingSlash bool) (*matcher, []*definitionError) {
```
mo-gr: boolean flags in APIs are a smell to me. I'd prefer a set of Options (enums), that currently only have one value `ignoreTrailingSlashes`. Either provide a singletonSet with that value or an empty set to change the behaviour.

routing/matcher.go
```diff
@@ -1,111 +1,335 @@
...
+	for _, val := range vals {
+		if check(val) {
+			return true
+		}
+	}
+
+	return false
+}
+
+func matchHeaders(exact map[string]string, hrxs map[string][]*regexp.Regexp, h http.Header) bool {
```
mo-gr: why do we need exact and regexp? couldn't all exacts be represented as regexps? Is this only an optimisation?

routing/matcher_test.go
```diff
@@ -1,79 +1,1161 @@
...
+	testMatch(t, "GET", "/sls/herren/nike/sports", "https://herren-sls.streaming-layout-service.my-department.example.org")
+	testMatch(t, "GET", "/sls/damen/nike/sports", "https://damen-sls.streaming-layout-service.my-department.example.org")
+	testMatch(t, "GET", "/men/nike/sports", "https://herren-en.layout-service.my-department.example.org")
+	testMatch(t, "GET", "/women/nike/sports", "https://damen-en.layout-service.my-department.example.org")
+	testMatch(t, "GET", "/sls-async/men/nike/sports", "https://herren-en.streaming-layout-service.my-department.example.org")
+	testMatch(t, "GET", "/sls-async/women/nike/sports", "https://damen-en.streaming-layout-service.my-department.example.org")
+	testMatch(t, "GET", "/sc/men/nike/sports", "https://herren-en.compositor-layout-service.my-department.example.org")
+	testMatch(t, "GET", "/sc/women/nike/sports", "https://damen-en.compositor-layout-service.my-department.example.org")
+	testMatch(t, "GET", "/sls/men/nike/sports", "https://herren-en.streaming-layout-service.my-department.example.org")
+	testMatch(t, "GET", "/sls/women/nike/sports", "https://damen-en.streaming-layout-service.my-department.example.org")
```
mo-gr: copy pasta? Maybe some generators could be written to generate all those test cases?

routing/routing.go
```diff
@@ -1,37 +1,120 @@
...
+}
+
+func processRoutes(fr filters.Registry, eskipRoutes []*eskip.Route) []*Route {
+	routes := []*Route{}
+	for _, eskipRoute := range eskipRoutes {
+		route, err := processRoute(fr, eskipRoute)
+		if err == nil {
+			routes = append(routes, route)
+		} else {
+			// idividual definition errors are accepted here
```
mo-gr: what means 'accepting errors'?

routing/routing_test.go
```diff
@@ -1,26 +1,18 @@
...
-	return func(req *http.Request) (*Route, map[string]string) {
-		v, p := m.Match(req)
-		r, _ := v.(*Route)
-		return r, p
-	}
-}
-
-func testMatcherWithPath(t *testing.T, m matcher, path string, matchRoute *Route) {
+func testMatcherWithPath(t *testing.T, m matcherFunc, path string, matchRoute *Route) {
 	req, err := http.NewRequest("GET", "http://www.example.com"+path, nil)
```
mo-gr: why does this use example.com but all further examples use example.org </nitpicking>

mo-gr:
What are the go-lang conventions regarding new/make/create... There is a lot of renaming going on here in those three verbs

routing/matcher_test.go
aryszka: it was generated with vim, except for the post and pdp requests :)
actually, we don't need all these regardless of the generating method. it was useful during working on the route lookup itself

routing/routing_test.go
aryszka: all further ones use this one, too, because they use this function
in the test, it proxies requests, that were made to .com, to .org

routing/matcher.go
aryszka: yo, converted into MatchingOptions

mo-gr:
lgtm

routing/matcher.go
aryszka: good question. there can be some performance benefits, though, but doubt that any significant. let me think about it a bit.

aryszka:
regarding go conventions for make/new/create:
there is no clearly defined convention, but there were some conversations in the forums and one can check the std lib's source, too. The thoughts behind these renamings were these:
In the go bultins we have `new` and `make`. `new` allocates memory for a type and returns a pointer to it, while `make` constructs a types as slices, maps and channels, which are not passed around by their pointers, but act more like some sort of 'reference' types. So, I made renaming here for those constructors that return pointers to new structs.
Finally 'create': I used it where neither new or make fit well.

PR #12, opened 2015-10-05
Strip query

mo-gr:


filters/stripquery/stripquery_test.go
```diff
@@ -0,0 +1,79 @@
...
+	c := &mock.FilterContext{nil, req, nil, false, nil}
+	f.Request(c)
+
+	q := c.FRequest.URL.Query()
+	if len(q) > 0 {
+		t.Error("query not removed")
+	}
+}
+
+var headerTests = []struct {
```
aryszka: nice anon struct

filters/stripquery/stripquery.go
```diff
@@ -0,0 +1,95 @@
...
+		return
+	}
+
+	q := url.Query()
+	for k, vv := range q {
+		for _, v := range vv {
+			if r.Header == nil {
+				r.Header = http.Header{}
+			}
+			r.Header.Add(fmt.Sprintf("X-Query-Param-%s", sanitize(k)), v)
```
aryszka: i wonder if we should use http.CanonicalHeaderKey. will this be 'X-Query-Param-foo'?

aryszka:
lgtm

PR #13, opened 2015-10-07
Release cleanup data client

aryszka:


eskip/parser.go
```diff
@@ -306,6 +306,7 @@ func (eskiprcvr *eskipParserImpl) Parse(eskiplex eskipLexer) int {
 	var eskiplval eskipSymType
 	var eskipVAL eskipSymType
 	var eskipDollar []eskipSymType
+	_ = eskipDollar // silence set and not used
```
mo-gr: what?

innkeeper/client.go
```diff
@@ -98,67 +91,200 @@ type apiError struct {
...
+	hs := make(map[string]string)
+	for _, h := range d.Route.MatchHeaders {
+		hs[h.Name] = h.Value
+	}
+
+	return hs
+}
+
+func innkeeperProcotolToScheme(p string) string {
+	return strings.ToLower(p)
```
mo-gr: this looks a bit weak. It should maybe switch/case over the known types and err out if it encounters an unknown string?

innkeeper/client.go
```diff
@@ -98,67 +91,200 @@ type apiError struct {
...
+func splitOnMethods(data []*routeData) []*routeData {
+	var split []*routeData
+	for _, d := range data {
+		if len(d.Route.MatchMethods) == 0 {
+			split = append(split, d)
+			continue
+		}
+
+		for _, m := range d.Route.MatchMethods {
+			copy := d.Route
```
mo-gr: does that actually create a copy? Doesn't look like it

routing/datasource.go
```diff
@@ -0,0 +1,197 @@
...
+
+const (
+	incomingReset incomingType = iota
+	incomingUpdate
+)
+
+type routeDefs map[string]*eskip.Route
+
+type incomingData struct {
+	typ            incomingType
```
mo-gr: `typ` because `type` is a keyword?

routing/datasource.go
```diff
@@ -0,0 +1,197 @@
...
+		return
+	}
+}
+
+func receiveUpdates(c DataClient, pollTimeout time.Duration, out chan<- *incomingData) {
+	for {
+		time.Sleep(pollTimeout)
+		routes, deletedIds, err := c.GetUpdate()
+		if err != nil {
+			log.Println("error while receiving update;", err)
```
mo-gr: this basically ends that client, right? are you sure, you want to be that hard on the first error? maybe a retry?

routing/datasource.go
```diff
@@ -0,0 +1,197 @@
...
+func receiveInitial(c DataClient, pollTimeout time.Duration, out chan<- *incomingData) {
+	for {
+		routes, err := c.GetInitial()
+		if err != nil {
+			log.Println("error while receiveing initial data;", err)
+			time.Sleep(pollTimeout)
+			continue
+		}
+
+		out <- &incomingData{incomingReset, c, routes, nil}
```
mo-gr: always send a reset from a new client? isn't making it hard to use multiple clients in parallel?

routing/datasource.go
```diff
@@ -0,0 +1,197 @@
...
+
+	if d.typ == incomingUpdate {
+		for _, id := range d.deletedIds {
+			delete(defs, id)
+		}
+	}
+
+	if d.typ == incomingReset || d.typ == incomingUpdate {
+		for _, def := range d.upsertedRoutes {
+			defs[def.Id] = def
```
mo-gr: your code suggests, that refs could be nil (line 65), but here, you are assigning to it without checking. Can this break?

routing/datasource.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	return defs
+}
+
+func mergeDefs(defsByClient map[DataClient]routeDefs) []*eskip.Route {
+	mergeById := make(routeDefs)
+	for _, defs := range defsByClient {
+		for id, def := range defs {
+			mergeById[id] = def
```
mo-gr: Is the iteration order of a map specified in go? if multiple clients have a routeDef with the same id, is it determinable, which client will "win" ?

routing/datasource.go
```diff
@@ -0,0 +1,197 @@
...
+	bu, err := url.ParseRequestURI(r.Backend)
+	if err != nil {
+		return "", "", err
+	}
+
+	return bu.Scheme, bu.Host, nil
+}
+
+func createFilter(fr filters.Registry, def *eskip.Filter) (filters.Filter, error) {
+	spec, ok := fr[def.Name]
```
mo-gr: ok? what's wrong with err?

eskip/parser.go
aryszka: probably a new version of goyacc.
(reminder for others: this file is generated, and only stored in the repo for not breaking go get. maybe there is already a better way to deal with this, I just don't know yet.)

innkeeper/client.go
aryszka: right, ok (I wish innkeeper just sent the actual value, since it is already a formal value, no need for a redefined enumeration)

innkeeper/client.go
aryszka: yes. routeData.Route is a value field not a pointer field.

run/run.go
```diff
@@ -62,11 +76,15 @@ func Run(o Options) error {
...
 
 	// create data client
-	dataClient, err := createDataClient(o, auth)
+	dataClients, err := createDataClients(o, auth)
 	if err != nil {
 		return err
 	}
 
+	if len(dataClients) == 0 {
+		log.Println("warning: no route source specified")
```
mo-gr: could be a panic. skipper isn't very useful w/o a dataClient, right?

routing/datasource.go
aryszka: yes. I saw it at many places:

```
grep -rn --include=*.go typ  /Users/aryszka/go/src
```

innkeeper/client.go
mo-gr: scala people love shovelling data in and out of enumerations ;-)

innkeeper/client.go
mo-gr: fair enough

routing/datasource.go
aryszka: well, not really. it is meant to look at it together with the `receiveFromClient` function. I'll update this to make it more obvious.

routing/datasource.go
aryszka: the reset affects only the definitions stored for the given client. so it is fine after all.

mo-gr: udnerstood

routing/datasource.go
aryszka: currently it is 'undefined behavior', but as we discussed, i can make it prioritized.

routing/datasource.go
aryszka: not an err. it is a boolean and it means that the key exists. E.g. to check only if a key exists:

```
if _, exists := myMap["foo"]; exists {
    println("key 'foo' exists")
} else {
    println("key 'foo' does not exist")
}
```

mo-gr: I see. This enables go-maps to store falsey values, right?

routing/datasource.go
mo-gr: discussed to make it defined. But I'd be fine with making that change in a separate, later PR

aryszka: ok, i create an issue for this for now.

routing/datasource.go
aryszka: yeah..

run/run.go
aryszka: I thought of this, but I still prefer the warning only, because even if there is a dataclient, it can be empty, or all routes can get deleted while running, and the result is the same, so it feels more logical to me this way. Also, what if in a weird solution some script should generate the arguments for a process pipe where skipper is only a non-critical process among others.

PR #17, opened 2015-10-12
Release cleanup

aryszka:
cleaning up tests, improving some of the readability, dropping useless tests


proxy/proxy_test.go
```diff
@@ -364,7 +361,7 @@ func TestProcessesRequestWithShuntBackend(t *testing.T) {
 		fr,
 		routing.MatchingOptionsNone,
 		sourcePollTimeout,
-		[]routing.DataClient{testdataclient.New(routes)},
+		[]routing.DataClient{dc},
 		0}), false)
 
 	delay()
```
danpersa: why do we need to delay?

danpersa:
lgtm :+1: for merging

proxy/proxy_test.go
aryszka: the propagation of the route settings and the startup is currently not in sync. In theory, a single 0 delay, or a starting a new go routine would be enough, because the test code is not doing anything else meanwhile, but it turns out, that somehow it is not enough for the scheduler. However, I already considered to make the first routing data request and processing come to sync on startup, and wait for it with a delay. I'll probably create an issue for this.

PR #22, opened 2015-10-15
make header matching case sensitive

aryszka:
#21 fixed


requestmatch/matcher.go
```diff
@@ -114,6 +115,24 @@ func compileRxs(exps []string) ([]*regexp.Regexp, error) {
 	return rxs, nil
 }
 
+func keyValToLower(kv map[string]string) map[string]string {
```
aryszka: this and the one below is actually a painful example how generics could be good sometimes

requestmatch/matcher.go
```diff
@@ -140,8 +159,8 @@ func makeLeaf(d Definition) (*leafMatcher, error) {
 		method:         d.Method(),
 		hostRxs:        hostRxs,
 		pathRxs:        pathRxs,
-		headersExact:   d.Headers(),
-		headersRegexps: allHeaderRxs,
+		headersExact:   keyValToLower(d.Headers()),
+		headersRegexps: headerRegexpsToLower(allHeaderRxs),
```
mo-gr: Why did you decide to go all lowercase on the header instead of the more common canonicalised form?

The godoc says:

        // HTTP defines that header names are case-insensitive.
        // The request parser implements this by canonicalizing the
        // name, making the first character and any characters
        // following a hyphen uppercase and the rest lowercase.

And uses `func CanonicalHeaderKey(s string) string` for that

requestmatch/matcher_test.go
```diff
@@ -1093,6 +1090,44 @@ func TestFreeWildcardParamWithSlash(t *testing.T) {
...
+	rd := &def{routes[0]}
+
+	m, errs := Make([]Definition{rd}, false)
+	if len(errs) != 0 {
+		t.Error("failed to make matcher")
+	}
+
+	r, _ := m.Match(&http.Request{URL: &url.URL{}, Header: http.Header{"Some-Header": []string{"some-value"}}})
+	if r == nil {
+		t.Error("failed to match header, case insensitive, to")
```
mo-gr: to?

requestmatch/matcher_test.go
```diff
@@ -1093,6 +1090,44 @@ func TestFreeWildcardParamWithSlash(t *testing.T) {
...
+	}
+
+	rd := &def{routes[0]}
+
+	m, errs := Make([]Definition{rd}, false)
+	if len(errs) != 0 {
+		t.Error("failed to make matcher")
+	}
+
+	r, _ := m.Match(&http.Request{URL: &url.URL{}, Header: http.Header{"some-header": []string{"some-value"}}})
```
mo-gr: A lot of repetitive boilerplate here.... Couldn't this be pulled into a func that takes an eskip string and a request (and maybe a flag to indicate whether there should be a route or not)?

requestmatch/matcher.go
aryszka: there was some microoptimization on my mind: the initialization step doesn't matter, but the lookup matters, and I just simply considered the ToLower should be faster than CanonicalHeaderKey. but! this is still a good point, and not only because i didn't measure, but because ToLower is utf8 operation while there is a chance that CanonicalHeaderKey is not. Regardless, I make a quick measure, and update this.

requestmatch/matcher_test.go
aryszka: considered this in part in the cleanup branch. There is this reoccurring topic of DRY in unit tests, too dry may hide the little story being tested. In some cases in that branch, I cleaned the tests up to be a bit more DRY, as it is getting more stable slowly, but actually this matcher_test stayed quite similar as what we see here. am I getting it very wrong? should I keep the tests less copy-pasty?

mo-gr: see for DRY in tests in go https://github.com/golang/go/wiki/TableDrivenTests

requestmatch/matcher.go
mo-gr: That is the point. No operation on the request necessary.

requestmatch/matcher.go
mo-gr: I don't think go argues, that they are not good or useful. They only haven't found a satisfying way to add them to the lang, right?

aryszka: yes, right. also, it may be better not to have them then seeing too many abuses

requestmatch/matcher_test.go
aryszka: ok, there are more places in the tests like this. created an issue:
https://github.com/zalando/skipper/issues/25

danpersa:
lgtm

PR #23, opened 2015-10-15
Feature #18 add x flow

### LINKED ISSUE #18: Add a filter to generate a request id
Often there is a flow id that helps gather events and log entries from multiple systems originated for the same customer request.

Such flow id needs to be generated at the earliest and sent down the request path so that the following systems can use it for those purposes.

I suggest we use a common HTTP header `X-Flow-Id` that is created for every request. Generation of this id should favor performance and relax on collision probability since the time window where this is relevant is limited

lmineiro:
The implementation generates a random string with 8 to 64 bytes and sets a HTTP header `X-Flow-Id` that other systems can use to correlate log entries.
The filter accepts some parameters that help control the its behavior, namely, to accept an existing flow id and the length of the generated flow ids
This closes #18


filters/filters.go
```diff
@@ -14,6 +15,10 @@ import (
 	"github.com/zalando/skipper/skipper"
 )
 
+var (
```
aryszka: no need for the braces when there is only a single var, but of course fine like this, too

lmineiro: I believe we can add some more global errors soon :)

filters/flowid/filter.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+type flowId struct {
+	id            string
+	reuseExisting bool
+	flowIdLength  uint8
+}
+
+func New(id string, allowOverride bool, len uint8) skipper.Filter {
+	return &flowId{id, allowOverride, len}
```
mo-gr: reuse existing or allow override? which one is it?

filters/flowid/filter.go
```diff
@@ -0,0 +1,64 @@
...
+
+	if this.reuseExisting {
+		flowId = r.Header.Get(flowIdHeaderName)
+		if isValid(flowId) {
+			return
+		}
+	}
+
+	flowId, err := newFlowId(this.flowIdLength)
+	if err == nil {
```
mo-gr: swallowing an error here. If this goes wrong, it should be logged, right?

filters/flowid/filter.go
```diff
@@ -0,0 +1,64 @@
...
+func (this *flowId) MakeFilter(id string, fc skipper.FilterConfig) (skipper.Filter, error) {
+	var reuseExisting bool
+	if len(fc) > 0 {
+		if r, ok := fc[0].(bool); ok {
+			reuseExisting = r
+		} else {
+			return nil, filters.ErrInvalidFilterParameters
+		}
+	}
+	var flowIdLength uint8 = defaultLen
```
mo-gr: where does the defaultLen come from?

filters/flowid/readme.md
```diff
@@ -0,0 +1,67 @@
...
+
+The filter can be used with many different combinations of parameters. It can also be used without any parameter, since 
+both are options.
+
+### Default parameters
+	FlowId()
+Without any parameters, the filter doesn't reuse existing `X-Flow-Id` headers and generates new ones with 16 bytes.
+
+### Reuse existing flow id
+	FlowId(true)
```
aryszka: in these types of filter arguments, we could use enum strings. FlowId(true) would result in a parsing error.
I recommend: FlowId("reuse")

filters/flowid/readme.md
```diff
@@ -0,0 +1,67 @@
...
+The second parameter is a number that defines the length of the generated flow ids. Valid options are any even number
+between 8 and 254.
+
+## Usage
+
+The filter can be used with many different combinations of parameters. It can also be used without any parameter, since 
+both are options.
+
+### Default parameters
+	FlowId()
```
aryszka: can we use camel casing for the filters, and Pascal casing for the matcher conditions? it would be nice to have a convention here

filters/flowid/readme.md
```diff
@@ -0,0 +1,67 @@
...
+
+	BenchmarkFlowIdLen8-4 	 1000000	      1157 ns/op
+	BenchmarkFlowIdLen10-4	 1000000	      1162 ns/op
+	BenchmarkFlowIdLen12-4	 1000000	      1163 ns/op
+	BenchmarkFlowIdLen14-4	 1000000	      1171 ns/op
+	BenchmarkFlowIdLen16-4	 1000000	      1180 ns/op
+	BenchmarkFlowIdLen32-4	 1000000	      1957 ns/op
+	BenchmarkFlowIdLen64-4	  300000	      3520 ns/op
+
+As you can see, starting at len = 32 (16 random bytes) the performance starts dropping dramatically.
```
aryszka: did you figure the reason? :)

filters/flowid/hashing.go
```diff
@@ -0,0 +1,37 @@
...
+	minLength  = 8
+)
+
+var (
+	ErrInvalidLen  = errors.New(fmt.Sprintf("Invalid length. len must be >= %d and < %d", minLength, maxLength))
+	flowIdRegex, _ = regexp.Compile(`^[\w+/=\-]+$`)
+)
+
+func newFlowId(len uint8) (string, error) {
+	if len < minLength || len > maxLength || len%2 != 0 {
```
aryszka: if it works, it works. just a warning that i managed once to trigger very strange behavior from the compiler by overdefining 'make' and it resulted in a relatively hard-to-figure compilation error.

mo-gr: if `len%2` is a requirement, the ErrInvalidLen should point that out

filters/flowid/hashing.go
```diff
@@ -0,0 +1,37 @@
...
+
+const (
+	defaultLen = 16
+	maxLength  = 254
+	minLength  = 8
+)
+
+var (
+	ErrInvalidLen  = errors.New(fmt.Sprintf("Invalid length. len must be >= %d and < %d", minLength, maxLength))
+	flowIdRegex, _ = regexp.Compile(`^[\w+/=\-]+$`)
```
aryszka: in these cases, you can use MustCompile. it doesn't return an error, but panics immediately when running the tests. The idea is the same, the expression is not coming from outside, but you're one step safer by a panic during the tests.

filters/flowid/hashing.go
```diff
@@ -0,0 +1,37 @@
+package flowid
+
+import (
+	"crypto/rand"
```
mo-gr: Do we actually need cryptographic randomness here? It seems like regular, pseudo randomness should be enough

filters/flowid/filter.go
```diff
@@ -0,0 +1,64 @@
...
+		fc.Request().Header.Set(flowIdHeaderName, flowId)
+	}
+}
+
+func (this *flowId) Response(skipper.FilterContext) {}
+
+func (this *flowId) MakeFilter(id string, fc skipper.FilterConfig) (skipper.Filter, error) {
+	var reuseExisting bool
+	if len(fc) > 0 {
+		if r, ok := fc[0].(bool); ok {
```
aryszka: we don't have bool :(
the doc is not yet in the master, but you can have a look at here:
https://github.com/zalando/skipper/blob/release-cleanup-doc/eskip/doc.go

filters/flowid/hashing.go
```diff
@@ -0,0 +1,37 @@
...
+func newFlowId(len uint8) (string, error) {
+	if len < minLength || len > maxLength || len%2 != 0 {
+		return "", ErrInvalidLen
+	}
+
+	u := make([]byte, hex.DecodedLen(int(len)))
+	buf := make([]byte, len)
+
+	rand.Read(u)
+	hex.Encode(buf, u)
```
mo-gr: Hex is cool and all... I get that, but couldn't you just draw `len` times a random char and concatenate them?

filters/flowid/filter.go
lmineiro: We haven't exactly established this logging strategy yet...

filters/flowid/hashing.go
```diff
@@ -0,0 +1,37 @@
...
+	return string(buf), nil
+}
+
+func isValid(flowId string) bool {
+	return len(flowId) >= minLength && len(flowId) <= maxLength && flowIdRegex.MatchString(flowId)
```
mo-gr: Reggie could also check for the length. not sure if this makes sense though...

filters/flowid/readme.md
```diff
@@ -0,0 +1,67 @@
+# Flow ID Filter
```
mo-gr: `Flow Id` or `Flow ID` - please be consistent

filters/flowid/readme.md
lmineiro: Your call. I haven't noticed, until now, which is out standard

filters/flowid/readme.md
```diff
@@ -0,0 +1,67 @@
...
+The filter takes 2 optional parameters:
+
+1. Accept existing `X-Flow-Id` header
+2. Flow Id length
+	 
+The first parameter is a boolean parameter that, when set to true, will make the filter skip the generation of
+a new flow id. If the existing header value is not a valid flow id it is ignored and a new flow id is also generated.
+
+The second parameter is a number that defines the length of the generated flow ids. Valid options are any even number
+between 8 and 254.
```
mo-gr: Forcing an even number seems strangely arbitrary without knowing the implementation

filters/flowid/readme.md
lmineiro: Went all the way to the getrandom(2) syscall. Without wasting too much time there, I'll blame this on the drainage of entropy pool

filters/flowid/readme.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Usage
+
+The filter can be used with many different combinations of parameters. It can also be used without any parameter, since 
+both are options.
```
mo-gr: s/options./optional/

filters/flowid/hashing.go
lmineiro: Good hint

filters/flowid/filter.go
```diff
@@ -0,0 +1,64 @@
...
+	flowIdHeaderName = "X-Flow-Id"
+)
+
+type flowId struct {
+	id            string
+	reuseExisting bool
+	flowIdLength  uint8
+}
+
+func New(id string, allowOverride bool, len uint8) skipper.Filter {
```
aryszka: i don't see where the filter spec is exported

filters/flowid/filter.go
lmineiro: Right, it's used only here but it is in the hashing file. Will move the const here

filters/flowid/filter.go
aryszka: true, but all errors that couldn't be handled were printed to stderr as a basic approach. i think we should do it here, too.

filters/flowid/readme.md
aryszka: there wasn't :) you are the first person with whom i can agree on it

mo-gr:
Have you considered just having an array of valid chars and then `len` times pulling a random index out of that one. Seems to me, that this might be faster than crypto-random and hex operations

lmineiro:
Pre-defining an alphabet seems like a good idea. I'm sure it will be faster and, you're right, for this we don't need crypto-randomness.

filters/flowid/hashing.go
lmineiro: We don't

filters/flowid/filter.go
lmineiro: Okay

filters/flowid/hashing.go
lmineiro: Would be more expensive to delegate that to the regex. I expect the short-circuit of the boolean evaluation to beat that

filters/flowid/filter.go
lmineiro: Right. I believe I got it now how this should work out...

lmineiro:
$ curl -v http://localhost:9090/flowid
* Hostname was NOT found in DNS cache
*   Trying ::1...
> GET /flowid HTTP/1.1
> User-Agent: curl/7.38.0
> Host: localhost:9090
> Accept: */*
> 
---
$ nc -l 8080
GET /flowid HTTP/1.1
Host: localhost:8080
User-Agent: curl/7.38.0
Accept: */*
X-Flow-Id: 4vI2AarIh5hqqsuD
Accept-Encoding: gzip

aryszka:
lgtm, very nice

filters/filters.go
```diff
@@ -28,6 +29,7 @@ func Register(registry skipper.FilterRegistry) {
 		static.Make(),
 		stripquery.Make(),
 		&redirect.Redirect{},
+		flowid.New(),
```
danpersa: new instead of make? shouldn't we be consistent with naming?

danpersa:
lgtm

filters/filters.go
lmineiro: Yes, we should. It will be New*(). Refactoring the old filters will be next

aryszka: it's already in progress to switch to new, see the release-cleanup branch

PR #35, opened 2015-10-30
Eskip tool

aryszka:
low prio, but can be a useful tool
there some smaller changes to the etcd client, otherwise doesn't affect the main code


cmd/eskip/doc.go
```diff
@@ -0,0 +1,140 @@
...
+
+Delete routes from etcd:
+
+    eskip delete -ids route1,route2,route3
+
+Delete all routes from etcd:
+
+    eskip print | eskip delete
+
+Copy all routes in etcd to under another prefix:
```
mo-gr: the under is misplaced

cmd/eskip/write.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	id, err := flowid.NewFlowId(randomIdLength)
+	if err != nil {
+		return err
+	}
+
+	// replace characters that are not allowed
+	// for eskip route ids.
+	id = routeIdRx.ReplaceAllString(id, "x")
```
mo-gr: why not just use a uuid?

etcd/etcd.go
```diff
@@ -34,14 +37,27 @@ tests.
...
+// RouteInfo contains a route id, plus the loaded and parsed route or
+// the parse error in case of failure.
+type RouteInfo struct {
+
+	// The route id plus the route data or if parsing was successful.
+	eskip.Route
+
+	// The parsing error if the parsing failed.
+	ParseError error
+}
```
mo-gr: I guess this thing shouldn't contain both at the same time, right? Is there an `Either` type in go?

mo-gr:
lgtm

mo-gr:
And huge 👍🏻 for that awesome eskip tool!

cmd/eskip/write.go
aryszka: I didn't want to add an additional dependency, and considered the randomness of this solution good enough for the purpose.

etcd/etcd.go
aryszka: It shouldn't be used other than receiving what was stored here. It means that there is a route here, and it is either valid, or this is the problem with it. There is no such thing as `Either`, but true, this is used in a similar way.

cmd/eskip/doc.go
aryszka: fixed, thanks

PR #39, opened 2015-11-08
fix quickstart instruction in the readme

aryszka:


readme.md
```diff
@@ -25,7 +25,7 @@ Skipper is 'go get' compatible. If needed, create a go workspace first:
 
 Get the skipper packages:
 
-    go get github.com/zalando/skipper
+    go get github.com/zalando/skipper/...
```
danpersa: 3 dots?

aryszka: that's right. three dot means recursively in the subdirectories for the go tool. this change is needed because recently we changed the root package from a main executable into a simple lilbrary package, and the executable was moved under the cmd/skipper and cmd/eskip packages, and go get doesn't call install on these without the three dots.

danpersa:
LGTM :+1:

Issue #40, opened 2015-11-09
make the filter instances identifiable for monitoring purpose


aryszka:

LappleApple:
If "Help Wanted" can also apply, thumbs-up this comment.

PR #42, opened 2015-11-09
add monitoring info to the constructed filter instances

aryszka:
#40


aryszka:
(not meant to master, yet)

PR #43, opened 2015-11-09
add monitoring info to the constructed filter instances

aryszka:
#40 

in one way, i find it laser. in another way, it's ugly :)
what do you think?

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -188,14 +188,14 @@ func callSafe(p func()) {
 }
 
 // applies all filters to a request
-func applyFiltersToRequest(f []filters.Filter, ctx filters.FilterContext) {
+func applyFiltersToRequest(f []*routing.Filter, ctx filters.FilterContext) {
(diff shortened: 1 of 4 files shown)


lmineiro:
I'd prefer to name this new type to something different to make it clear. Something like RouteFilter, FilterWrapper or something else. This is only used by the proxy, isn't it?
Whoever is writing a filter will get one or the other and, eventually, use the wrong one. For those using auto-completing IDEs it can even auto-import the wrong one.

aryszka:
i thought of this, and the argument for this name was that one always references it together with the package, like routing.Filter. but still see the point that it is not the same thing, so ok. RouteFilter is the name i hate the least.
(btw, this one is a structure, the other one is an interface, and this one even implements the other by anonymously including an instance of it, just to be precise :) )

aryszka:
RouteFilter, it became

lmineiro:
(y)

PR #48, opened 2015-11-10
Application and Access log - initial version

aryszka:


aryszka:
created this to the wip branch, only to avoid double work

log/access.go
```diff
@@ -0,0 +1,86 @@
...
+	ff := r.Header.Get("X-Forwarded-For")
+	if ff != "" {
+		return ff
+	}
+
+	return r.RemoteAddr
+}
+
+// strip port from addresses with hostname, ipv4 or ipv6
+func stripPort(address string) string {
```
lmineiro: https://golang.org/pkg/net/#SplitHostPort
Recently revived my ExtractDomain function using that one also

aryszka: updated

log/access.go
```diff
@@ -0,0 +1,78 @@
...
+	h, _, err := net.SplitHostPort(address)
+	if err != nil {
+		h = ""
```
lmineiro: Are you sure this is what you want to do? If address doesn't have a port it may still have a valid address...

PR #51, opened 2015-11-12
Update readme.md

LappleApple:


readme.md
```diff
@@ -3,15 +3,15 @@
 # Skipper
 
 Skipper is an HTTP router that acts as a reverse proxy (with support for flexible route definitions), alters
-requests, and responses with filters. It:
+requests, and responds with filters. It:
```
danpersa: "request -> response" is the way the http people refer to requests responds :) you shouldn't change this...

https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Response_message

aryszka: what I meant here is that both the request and the response can be altered 'by' the filters. this sounds more as the filters were the response.

aryszka: here is the intro paragraph from the documentation (https://godoc.org/github.com/zalando/skipper), slightly modified:

'''
Skipper is an HTTP router that acts as a reverse proxy, mapping incoming requests to multiple HTTP backend services, based on routes selected by the request attributes. In between, both the requests and responses can be augmented by a filter chain defined individually for each route.
'''

Maybe we could use this in the readme, too. Does it look ok?

LappleApple: Sure. The confusion from my edit stems from the original use of a comma that changes the meaning of your intent (I changed "responses" to "responds," to be grammatically consistent). If you use this new intro, here's how it should go (minus the edit, which is a question):

Skipper is an HTTP router that acts as a reverse proxy, mapping incoming requests to multiple HTTP backend services based on routes selected by the request attributes. In between [EDIT: in between what and what?], both the requests and responses can be augmented by a filter chain defined individually for each route.

aryszka: I meant by 'in between' as between the request processing starts and the response is returned, but actually, it isn't technically precise either way. So, I think the simplest would be to simply leave the 'in between' out:

Skipper is an HTTP router that acts as a reverse proxy, mapping incoming requests to multiple HTTP backend services based on routes selected by the request attributes. Both the requests and responses can be augmented by a filter chain defined individually for each route.

PR #52, opened 2015-11-13
Access log

aryszka:
- access log created, not yet wired in into the proxy
- moved the code from log to the logging package


lmineiro:
lgtm

PR #54, opened 2015-11-17
Logging options flags

aryszka:


cmd/skipper/main.go
```diff
@@ -103,6 +109,9 @@ func init() {
 	flag.StringVar(&metricsPrefix, "metrics-prefix", "", metricsPrefixUsage)
 	flag.BoolVar(&debugGcMetrics, "debug-gc-metrics", false, debugGcMetricsUsage)
 	flag.BoolVar(&runtimeMetrics, "runtime-metrics", true, runtimeMetricsUsage)
+	flag.StringVar(&applicationLog, "application-log", "", applicationLogUsage)
+	flag.StringVar(&applicationLogPrefix, "application-log-prefix", "", applicationLogPrefixUsage)
```
lmineiro: I thought we wanted some other default here. `APP`? `APPLICATION`? `APPLICATION_LOG`?

logging/log.go
```diff
@@ -62,5 +56,5 @@ func Init(o Options) {
 	}
 }
 
-func ApplicationLog() *logrus.Logger { return appLog }
+func ApplicationLog() *logrus.Logger { return logrus.StandardLogger() }
```
lmineiro: This one is, basically, available through `logrus.XXX` or, with the alias, `log.XXX`. Do you think it's worth it to export such function?

cmd/skipper/main.go
aryszka: yup, we'll have it in our own appliance. notice that the default for the access log here is that it is turned off. The reason is that when someone just starts the app in dev mode, may not want to be flooded with access log. And then the prefix is not required. Is this a false logic?

logging/log.go
aryszka: true. actually, neither the applog or the accesslog is worth exporting.

PR #59, opened 2015-11-19
Added backend url to the filter context

### LINKED ISSUE #57: Expose the backend url from the route in the FilterContext
Some filters need access to the backend url from the route to whitelist/blacklist certain features

Currently there is no way to do it.

The FilterContext should include a copy of that element from the route

lmineiro:
This closes #57


aryszka:
looks good to me

PR #60, opened 2015-11-20
Logging and monitoring

lmineiro:
This adds both logging and metrics that can be used by monitoring system


.gitignore
```diff
@@ -6,3 +6,4 @@ default.etcd
 oauth/client.json
 oauth/user.json
 .idea
+example.eskip
```
mo-gr: why gitignore example.eskip? Seems like it could be useful to have some examples in the repo

cmd/skipper/main.go
```diff
@@ -30,9 +30,9 @@ package main
 
 import (
 	"flag"
+	log "github.com/Sirupsen/logrus"
```
mo-gr: Logrus itself claims "The core API is unlikely to change much but please version control your Logrus to make sure you aren't fetching latest master on every build." I don't think, we are doing this... Have you considered to do that?

cmd/skipper/main.go
```diff
@@ -90,6 +108,14 @@ func init() {
 	flag.StringVar(&innkeeperPreRouteFilters, "innkeeper-pre-route-filters", "", innkeeperPreRouteFiltersUsage)
 	flag.StringVar(&innkeeperPostRouteFilters, "innkeeper-post-route-filters", "", innkeeperPostRouteFiltersUsage)
 	flag.BoolVar(&devMode, "dev-mode", false, devModeUsage)
+	flag.StringVar(&metricsListener, "metrics-listener", defaultMetricsListener, metricsListenerUsage)
+	flag.StringVar(&metricsPrefix, "metrics-prefix", defaultMetricsPrefix, metricsPrefixUsage)
+	flag.BoolVar(&debugGcMetrics, "debug-gc-metrics", false, debugGcMetricsUsage)
+	flag.BoolVar(&runtimeMetrics, "runtime-metrics", true, runtimeMetricsUsage)
+	flag.StringVar(&applicationLog, "application-log", "", applicationLogUsage)
+	flag.StringVar(&applicationLogPrefix, "application-log-prefix", "", applicationLogPrefixUsage)
+	flag.StringVar(&accessLog, "access-log", "", accessLogUsage)
```
mo-gr: why are some defaults extracted as named constants and some defaults are just used inline? Seems inconsistent

cmd/skipper/main.go
lmineiro: We did, to some extent. We're aware of that warning and this was an informed decision

logging/access.go
```diff
@@ -0,0 +1,126 @@
...
+	for i, key := range keys {
+		values[i] = e.Data[key]
+	}
+
+	return []byte(fmt.Sprintf(f.format, values...)), nil
+}
+
+// Logs an access event in Apache combined log format.
+func Access(entry *AccessEntry) {
+	if accessLog == nil || entry == nil {
```
mo-gr: What is this `accessLog` variable. Where is it coming from? Just referencing a var from another file smells a bit to me. Can't you at least make it explicit, where it is defined?

logging/access.go
```diff
@@ -0,0 +1,126 @@
...
+import (
+	"fmt"
+	"github.com/Sirupsen/logrus"
+	"net"
+	"net/http"
+	"time"
+)
+
+const (
+	dateFormat      = "02/Jan/2006:15:04:05 -0700"
```
mo-gr: Is that how go date format works? You just throw it a date and it figures it out? Weird magic. Anyway, is the separation between date and time really a `:`?

logging/access_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"bytes"
+	"net/http"
+	"testing"
+	"time"
+)
+
+const logOutput = `127.0.0.1 - - [10/Oct/2000:13:55:36 +0000] "GET /apache_pb.gif HTTP/1.1" 418 2326 "" "" 42`
+
+func testRequest() *http.Request {
+	r, _ := http.NewRequest("GET", "http://frank@127.0.0.1", nil)
```
mo-gr: Who is frank?

logging/access_test.go
```diff
@@ -0,0 +1,78 @@
...
+	testAccessLog(t, nil, "")
+}
+
+func TestNoPanicOnMissingRequest(t *testing.T) {
+	entry := testAccessEntry()
+	entry.Request = nil
+	testAccessLog(t, entry, `- - - [10/Oct/2000:13:55:36 +0000] "  " 418 2326 "" "" 42`)
+}
+
+func TestUseXForwarded(t *testing.T) {
```
mo-gr: Can you add a test case for the case, that `X-Forwarded-For` contains multiple IP addresses?
Something like:
```
X-Forwarded-For: 1.2.3.4, 5.4.3.2, 8.8.8.8
```

mo-gr: And maybe also create a test for an IPv6 Address:

```
X-Forwarded-For: 2001:DB8::21f:5bff:febf:ce22:8a2e
```

logging/doc.go
```diff
@@ -0,0 +1,44 @@
...
+To send messages to the application log, import this package and use its
+methods. Example:
+
+    import log "github.com/Sirupsen/logrus"
+
+    func doSomething() {
+        log.Errorf("nothing to do")
+    }
+
+During startup initialization, it possible to redirect the log output
```
mo-gr: ...it is possible to...

logging/handler_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	innerHandler := http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusTeapot)
+	})
+	h := NewHandler(innerHandler)
+
+	h.ServeHTTP(httptest.NewRecorder(), &http.Request{})
+
+	output := accessLog.String()
+	if !strings.Contains(output, strconv.Itoa(http.StatusTeapot)) {
```
mo-gr: :+1: for the Teapot!

logging/handler_test.go
```diff
@@ -0,0 +1,65 @@
...
+		t.Error("failed to log access")
+	}
+}
+
+func TestSets200OnMissingStatus(t *testing.T) {
```
mo-gr: In what case can the status be missing and why is it ok to substitute it with 200 in that case?

logging/log_test.go
```diff
@@ -0,0 +1,52 @@
...
+	var buf bytes.Buffer
+	Init(Options{ApplicationLogOutput: &buf})
+	msg := "Hello, world!"
+	log.Infof(msg)
+	if !strings.Contains(buf.String(), msg) {
+		t.Error("failed to use custom output")
+	}
+}
+
+func TestCustomPrefixForApplicationLog(t *testing.T) {
```
mo-gr: This test validates, that the output contains the prefix. But it does not validate, that the output actually contains the full message. Seems like a possible hideout for a nice bug.

logging/loggingwriter.go
```diff
@@ -0,0 +1,32 @@
+package logging
+
+import "net/http"
+
+// this name reflects it more what it is
```
mo-gr: What is this comment trying to tell me? That we don't suck at naming?

metrics/doc.go
```diff
@@ -0,0 +1,38 @@
...
+https://github.com/dropwizard/metrics
+https://github.com/rcrowley/go-metrics
+
+The collected metrics include detailed information about Skipper's relevant processes while serving requests -
+looking up routes, filters (aggregate and individual), backend communication and forwarding the response to
+the client.
+
+Options
+
+To enable metrics, it needs to be initialized with a Listener address.In this case, Skipper will start an additional
```
mo-gr: Space missing after `address.`

metrics/doc.go
```diff
@@ -0,0 +1,38 @@
...
+
+You can define a custom Prefix to every reported metrics key. This allows you to avoid conflicts between Skipper's
+metrics and other systems if you aggregate them later in some monitoring system. The default prefix is "skipper."
+
+You can also enable some Go garbage collector and runtime metrics using EnableDebugGcMetrics and EnableRuntimeMetrics,
+respectively.
+
+REST API
+
+This listener accepts GET requests on the /metrics endpoint like any other REST api. A request to "/metrics" should
```
mo-gr: should?

lmineiro: Yes, this is related to the next statement. If you don't register debug gc or runtime stats and the instance didn't serve a single request it will return a 404 because it still has no metrics.

Maybe I should make this part more clear - no known metrics (yet) resulting in a 404.

What would you suggest?

lmineiro: "This listener accepts GET requests on the /metrics endpoint like any other REST api. A request to "/metrics" should
return a JSON response including all the collected metrics. Please note that a lot of metrics are created lazily
whenever a request triggers them. This means that your response object content will depend on your own routes and
the filters used. In the case there are no metrics due to inactivity, if can receive a 404 response instead."
Looks reasonable?

metrics/handler.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+// This listener is only used to expose the metrics
+func (mh *metricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	p := r.URL.Path
+	if r.Method == "GET" && (p == "/metrics" || strings.HasPrefix(p, "/metrics/")) {
+		mh.sendMetrics(w, strings.TrimPrefix(p, "/metrics"))
+	} else {
+		http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
```
mo-gr: Why not 404?

cmd/skipper/main.go
lmineiro: The plan was to only declare constant for those defaults which value was not the default empty value for the type. I see that we've might have missed some, like the runtime-metrics.

metrics/handler.go
```diff
@@ -0,0 +1,56 @@
...
+				metrics[prefix+name] = i
+			}
+		})
+	}
+
+	return metrics
+}
+
+func (mh *metricsHandler) sendMetrics(w http.ResponseWriter, p string) {
+	_, k := path.Split(p)
```
mo-gr: If I read the code correctly, it will also accept `/metrics/crap/more/crap/metric-key`. Is that a problem for you?

.gitignore
lmineiro: The readme already suggests how to create this particular file. If we decide to commit one such file we'll change the readme

metrics/metrics.go
```diff
@@ -0,0 +1,165 @@
...
+	"fmt"
+	log "github.com/Sirupsen/logrus"
+	"github.com/rcrowley/go-metrics"
+	"net/http"
+	"time"
+)
+
+type skipperMetrics map[string]interface{}
+
+// Options for initializing metrics collection.
```
mo-gr: I would have never guessed that myself. Thank you for that insightful comment!

metrics/metrics.go
```diff
@@ -0,0 +1,165 @@
...
+	// If set, Go runtime metrics are collected in
+	// addition to the http traffic metrics.
+	EnableRuntimeMetrics bool
+}
+
+const (
+	KeyRouteLookup     = "routelookup"
+	KeyFilterRequest   = "filter.%s.request"
+	KeyFiltersRequest  = "filters.request.%s"
+	KeyProxyBackend    = "backend.%s"
```
mo-gr: what is the difference between filter and filters?

metrics/handler.go
lmineiro: Because it is a BadRequest ;-)

I didn't want to branch the specific 405 (MethodNotAllowed) and 404 (NotFound). Since is the metrics endpoint I don't think it is that important to follow the rfc so strictly. Do you agree?

metrics/metrics.go
```diff
@@ -0,0 +1,165 @@
...
+func Init(o Options) {
+	if o.Listener == "" {
+		log.Infoln("Metrics are disabled")
+		return
+	}
+
+	r := metrics.NewRegistry()
+	if o.EnableDebugGcMetrics {
+		metrics.RegisterDebugGCStats(r)
+		go metrics.CaptureDebugGCStats(r, 5e9)
```
mo-gr: Any way to make that `5e9` more readable? Like `(5 * Time.second)` or something like that?

metrics/handler.go
lmineiro: I guess anyone can be that creative with the metrics prefix. I don't think this would be a problem

metrics/metrics.go
lmineiro: filters (plural) is the aggregate for all of them where the suffix (%s) will be the route
filter (singular) is the measurement for a specific filter (%s)
Both are measured in the request and response

metrics/metrics.go
lmineiro: This was a copy and paste from go-metrics example. Sounds reasonable to make it more readable, indeed.

proxy/proxy.go
```diff
@@ -268,6 +269,15 @@ func shunt(r *http.Request) *http.Response {
 		Request:    r}
 }
 
+// applies all filters to a request
```
mo-gr: thank you again for that insightful comment :+1:

proxy/proxy.go
```diff
@@ -278,84 +288,93 @@ func (p *proxy) roundtrip(r *http.Request, rt *routing.Route) (*http.Response, e
 	return p.roundTripper.RoundTrip(rr)
 }
 
-// http.Handler implementation
-func (p *proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	hterr := func(err error) {
-		// todo: just a bet that we shouldn't send here 50x
-		http.Error(w, http.StatusText(http.StatusNotFound), http.StatusNotFound)
-		log.Println(err)
+// applies all filters to a response
```
mo-gr: `// allies all filters to a response in reverse order`?

proxy/proxy.go
```diff
@@ -278,84 +288,93 @@ func (p *proxy) roundtrip(r *http.Request, rt *routing.Route) (*http.Response, e
...
+		})
 	}
+}
 
-	var (
-		rt     *routing.Route
-		params map[string]string
-	)
+func addBranding(rs *http.Response) {
+	rs.Header.Set("X-Powered-By", "Skipper")
```
mo-gr: a version could be useful

metrics/metrics.go
```diff
@@ -0,0 +1,165 @@
...
+
+func measure(key string, f func()) {
+	if t := getTimer(key); t != nil {
+		t.Time(f)
+	} else {
+		f()
+	}
+}
+
+func MeasureRouteLookup(start time.Time) {
```
mo-gr: Why is the interface to the MeasureXXX functions so inconsistent? some take a func, others take a start time... Choose one and stick with it. (I'd say, always take a func).

skipper.go
```diff
@@ -92,6 +97,47 @@ type Options struct {
...
+	// When /dev/stderr or /dev/stdout is passed in, it will be resolved
+	// to os.Stderr or os.Stdout.
+	//
+	// Warning: passing an arbitrary file will try to open it append
+	// on start and use it, or fail on start, but the current
+	// implementation doesn't support any more proper handling
+	// of temporary failures or log-rolling.
+	ApplicationLogOutput string
+
+	// Application log prefix. Default value: "[APPLICATION_LOG]".
```
mo-gr: This is a lie. The default value is an empty string.

lmineiro: It shouldn't :)

metrics/metrics.go
lmineiro: Early returns that should abort the ServeHTTP function

.gitignore
aryszka: do you mean this file is generated somewhere during the tests? the json files above are.

cmd/skipper/main.go
aryszka: i see actually only that one true. I think that's fine.

logging/access.go
aryszka: that's how go date format works. i agree that it is strange. this particular format is from the apache access log. we didn't test whether this format works with the go date fmt magic, but we saw it used in another library doing the same. (no, the code of that library was not to be reused.)

logging/access.go
aryszka: the dichotomy is caused because of this var is initialized in the log.go, together with the application log, but the bulk of the access log code is in access.go. i check how could it be better

logging/access_test.go
aryszka: frank is frank, i am just not sure about his name. suggestion?

logging/access_test.go
aryszka: legit

aryszka: luis looked up the related go stdlib code. the solution for the logging: it is enough to use the RemoteAddr field, assuming the thast go stdlib keeps doing its job.

on another page: we also should set the X-Forwarded-For and similar header.

logging/log_test.go
aryszka: right, fixing

metrics/metrics.go
aryszka: you're welcome

metrics/metrics.go
aryszka: maybe allfilters?

proxy/proxy.go
aryszka: thanks, i'll create an issue

metrics/metrics.go
lmineiro: :+1:

metrics/handler.go
mo-gr: fair enough

metrics/doc.go
mo-gr: it's just wording. but thanks for the explanation

metrics/handler.go
mo-gr: just wanted to make sure

metrics/metrics.go
mo-gr: Gnarling my teeth, but I accept that...

PR #66, opened 2015-12-01
introduce atomic for the matcher (routing table) pointer

aryszka:


lmineiro:
lgtm

routing/routing.go
```diff
@@ -130,10 +131,22 @@ func feedMatchers(updateBuffer int, current *matcher) (chan<- *matcher, <-chan *
 // Initializes a new routing instance, and starts listening for route
 // definition updates.
 func New(o Options) *Routing {
+	r := &Routing{}
 	initialMatcher, _ := newMatcher(nil, MatchingOptionsNone)
-	matchersIn, matchersOut := feedMatchers(o.UpdateBuffer, initialMatcher)
```
mo-gr: `feedMatchers` is now dead code and can be removed

mo-gr:
lgtm

routing/routing.go
aryszka: right, updating in a sec

PR #67, opened 2015-12-01
Refactored metrics to improve performance

lmineiro:
We found during load testing that the go-metrics mutex caused a relevant delay for super fast operations, like the route lookup.

I tested different strategies and the best approach was to update metrics as a go routine. We'll add results for this with further load testing.

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -77,45 +77,40 @@ func getTimer(key string) metrics.Timer {
 }
 
 func measureSince(key string, start time.Time) {
-	if t := getTimer(key); t != nil {
-		t.UpdateSince(start)
-	}
-}
-
-func measure(key string, f func()) {
-	if t := getTimer(key); t != nil {
-		t.Time(f)
-	} else {
-		f()
-	}
+	d := time.Since(start)
+	go func() {
+		if t := getTimer(key); t != nil {
+			t.Update(d)
+		}
+	}()
 }
 
 func MeasureRouteLookup(start time.Time) {
 	measureSince(KeyRouteLookup, start)
 }
 
-func MeasureFilterRequest(filterName string, f func()) {
-	measure(fmt.Sprintf(KeyFilterRequest, filterName), f)
+func MeasureFilterRequest(filterName string, start time.Time) {
+	measureSince(fmt.Sprintf(KeyFilterRequest, filterName), start)
 }
 
-func MeasureAllFiltersRequest(routeId string, f func()) {
-	measure(fmt.Sprintf(KeyFiltersRequest, routeId), f)
+func MeasureAllFiltersRequest(routeId string, start time.Time) {
+	measureSince(fmt.Sprintf(KeyFiltersRequest, routeId), start)
 }
 
 func MeasureBackend(routeId string, start time.Time) {
 	measureSince(fmt.Sprintf(KeyProxyBackend, routeId), start)
 }
 
-func MeasureFilterResponse(filterName string, f func()) {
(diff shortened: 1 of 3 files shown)


metrics/metrics_test.go
```diff
@@ -82,6 +73,7 @@ func TestMeasurement(t *testing.T) {
 	time.Sleep(5)
 	measureSince("TestMeasurement1", now)
 
+	time.Sleep(20 * time.Millisecond)
```
lmineiro: I'd like to go on record saying that this is ugly as hell and I hate myself for doing it

metrics/metrics_test.go
```diff
@@ -91,9 +83,8 @@ func TestMeasurement(t *testing.T) {
 		t.Error("'TestMeasurement2' metric should only have zeroes")
 	}
 
-	measure("TestMeasurement2", func() {
-		time.Sleep(5)
-	})
+	measureSince("TestMeasurement2", now)
+	time.Sleep(20 * time.Millisecond)
```
lmineiro: ditto

metrics/metrics_test.go
danpersa: at least this doesn't throw an exception as in java...

metrics/metrics_test.go
danpersa: as you're using this two times, you could extract it as a function :)

lmineiro: I refuse to cover my shame with such shenanigans :)

danpersa: lgtm

metrics/metrics.go
```diff
@@ -77,45 +77,40 @@ func getTimer(key string) metrics.Timer {
...
-	measure(fmt.Sprintf(KeyFiltersResponse, routeId), f)
+func MeasureAllFiltersResponse(routeId string, start time.Time) {
+	measureSince(fmt.Sprintf(KeyFiltersResponse, routeId), start)
 }
 
-func MeasureResponse(code int, method string, routeId string, f func()) {
-	measure(fmt.Sprintf(KeyResponse, code, method, routeId), f)
+func MeasureResponse(code int, method string, routeId string, start time.Time) {
+	measureSince(fmt.Sprintf(KeyResponse, code, method, routeId), start)
 }
```
mo-gr: at least the interface is consistent now :)

proxy/proxy.go
```diff
@@ -358,23 +360,24 @@ func (p *proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-			if err != nil {
-				log.Error(err)
-			}
-		})
+		start = time.Now()
+		copyHeader(w.Header(), rs.Header)
+		w.WriteHeader(rs.StatusCode)
+		err := copyStream(w.(flusherWriter), rs.Body)
+		if err != nil {
+			log.Error(err)
```
mo-gr: previously, you were measuring errors as well. Now, you don't. Is that on purpose?

mo-gr:
lgtm

proxy/proxy.go
lmineiro: Yes, this `else` was intentional. I was a bit lazy before because we were wrapping the all thing as a function.
Since now we only do this single interface I decided to only measure success to prevent immediate failures to skew data for the metric.

We should add later specific error metrics, only as counters, not timers. I'll create an issue.

lmineiro: https://github.com/zalando/skipper/issues/69

aryszka: jubilee-boom!

PR #72, opened 2015-12-04
fix lexer to trim comments in last position

aryszka:
#71 + carry returns (@mo-gr)


mo-gr:
Fix looks good to me... but this is pushing RegExps really hard
http://imgs.xkcd.com/comics/regular_expressions.png

aryszka:
agree. it's a dept. need to generate the lexer. thanks for the review!

PR #73, opened 2015-12-04
Added Godeps

Raffo:
Godeps added and Godeps/_workspace added to `.gitignore`. In this way you will NOT commit the sources of the imported projects. The dependencies can then be restored using `godep restore`. 
Let me know if this PR suits your needs.


aryszka:
Nice! Thanks a lot for the contribution!

PR #75, opened 2015-12-06
Route settings log

aryszka:


danpersa:
lgtm :)

routing/routing.go
```diff
@@ -125,6 +126,7 @@ func (r *Routing) startReceivingUpdates(o Options) {
 		for {
 			m := <-c
 			r.matcher.Store(m)
+			log.Println("route settings applied")
```
lmineiro: Does it have the same level as `Info()`?

aryszka: yes it has, per doc and test.

PR #76, opened 2015-12-07
Add innkeeper cli

danpersa:


cmd/eskip/load.go
```diff
@@ -107,6 +107,7 @@ func loadRoutes(in *medium) (loadResult, error) {
 		return loadReader(os.Stdin)
 	case file:
 		return loadFile(in.path)
+		// TODO: case innkeeper
```
danpersa: the logic will come in another pull request

cmd/eskip/media_test.go
```diff
@@ -157,26 +147,34 @@ func TestValidateSelectMedia(t *testing.T) {
 		false,
 		nil,
 		&medium{typ: inlineIds},
-		&medium{
```
danpersa: default cases handled in another method

cmd/eskip/media.go
```diff
@@ -93,25 +103,38 @@ func validateSelectWrite(cmd command, media []*medium) (input, output *medium, e
...
+		if m.typ == etcd || m.typ == innkeeper {
+			out = m
+		} else {
+			in = m
 		}
 	}
 
+	if in == nil {
+		return nil, nil, missingInput
+	} else if in.typ != inlineIds {
```
aryszka: doesn't this make it impossible to do this:

```
eskip print | eskip delete
```

?

cmd/eskip/media_defaults.go
```diff
@@ -0,0 +1,45 @@
+package main
```
aryszka: pls, call it 'mediadefaults.go'. the _ is used to separate tests, platform, architecture and similar things.

cmd/eskip/media_test.go
```diff
@@ -64,16 +64,6 @@ func TestValidateSelectMedia(t *testing.T) {
 		in      *medium
 		out     *medium
 	}{{
-
```
aryszka: why removing this test?

cmd/eskip/args.go
```diff
@@ -124,6 +135,30 @@ func processEtcdArgs(etcdUrls, etcdPrefix string) (*medium, error) {
...
+
+	if oauthToken == "" {
+		return nil, missingOAuthToken
+	}
+
+	if innkeeperUrl == "" {
+		innkeeperUrl = defaultInnkeeperUrl
+	}
+
+	urls, err := stringsToUrls([]string{innkeeperUrl})
```
aryszka: we can change the stringsToUrls signature to stringsToUrls(strs ...string) and then call it here simply with stringsToUrls(innkeeperUrl) while in the etcd with stringsToUrls(etcdUrls...)

cmd/eskip/media_defaults.go
```diff
@@ -0,0 +1,45 @@
...
+	input = in
+	if input == nil {
+		// take innkeeper as default
+		input, err = processEtcdArgs(defaultEtcdUrls, defaultEtcdPrefix)
+
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+	return input, nil, nil
```
aryszka: when using named return args, it is nicer to explicitly set the value for each of them, and just simply call 'return' without args

cmd/eskip/media_defaults_test.go
```diff
@@ -0,0 +1,109 @@
+// Copyright 2015 Zalando SE
```
aryszka: pls rename this file to mediadefaults_test.go, or move the code to media_test.go, it fits there perfectly.

cmd/eskip/args.go
danpersa: will do :)

cmd/eskip/media_defaults.go
danpersa: I will rename it... I don't think we should be afraid of creating more files... it keeps the code more structured... also important is that it keeps your tests structured...

cmd/eskip/media_defaults.go
danpersa: yeah... I was remembering this from the tutorials... but I forgot that you still need a return... I was trying not to put anything at all... :P 

will fix :+1:

cmd/eskip/media_defaults_test.go
danpersa: will do

cmd/eskip/media_test.go
danpersa: as this case will never happen... this function takes care of this :)

    src/github.com/zalando/skipper/cmd/eskip/eskip.go:74

cmd/eskip/media.go
danpersa: that's correct... I got that condition wrong... I will write a test to cover for this and then fix it :)

cmd/eskip/media_test.go
aryszka: even if the command is 'foo'?

cmd/eskip/media_defaults.go
aryszka: :+1: :)

cmd/eskip/media_test.go
danpersa: getCommand will send an invalidCommand err

danpersa: I will add tests to check for this :)

danpersa: added tests

cmd/eskip/eskip.go
```diff
@@ -22,7 +22,7 @@ import (
 
 type (
 	command     string
-	commandFunc func(in, out *medium) error
+	commandFunc func(in, out *medium, writeClient *WriteClient) error
```
aryszka: *WriteClient: why pointer of the interface?

cmd/eskip/readclient.go
```diff
@@ -0,0 +1,127 @@
...
+type inlineReader struct {
+	routes string
+}
+
+type idsReader struct {
+	ids []string
+}
+
+// TODO eliminate duplicate code in write and read clients
+func createReadClient(m *medium) (ReadClient, error) {
```
aryszka: I am really not a fan of applying this type of polymorphism in this case. Meaning the ReadClient and its implementations. The reason is that this doesn't make the code simpler but rather the opposite. The reasons why I don't see any benefit of it:
- these are stateless objects for a simple logic, within a single function that is called and the object is thrown away
- the switch may be hidden here, but it is still here
- with all the 'implementing' types, and their methods, the code just gets much more verbose

cmd/eskip/readclient.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/eskipfile"
+	etcdclient "github.com/zalando/skipper/etcd"
+	innkeeperclient "github.com/zalando/skipper/innkeeper"
+	"io"
+	"io/ioutil"
+	"os"
+)
+
+type ReadClient interface {
```
aryszka: why is this interface public?

cmd/eskip/readclient.go
```diff
@@ -0,0 +1,127 @@
...
+	LoadAndParseAll() ([]*eskip.RouteInfo, error)
+
+	LoadUpdate() (eskip.RouteList, []string, error)
```
aryszka: the LoadUpdate method doesn't seem to be used anywhere

cmd/eskip/eskip.go
```diff
@@ -113,6 +113,18 @@ func main() {
 		exitHint(err)
 	}
 
+	in, out, err = addDefaultMedia(cmd, in, out)
+
+	if err != nil {
+		exitHint(err)
+	}
+
+	writeClient, err := createWriteClient(out)
```
aryszka: maybe this createWriteClient could go inside the command, only when it is needed, similarly to read client?

cmd/eskip/writeclient.go
```diff
@@ -0,0 +1,44 @@
+package main
+
+import "github.com/zalando/skipper/eskip"
+import (
+	"errors"
+	etcdclient "github.com/zalando/skipper/etcd"
+	innkeeperclient "github.com/zalando/skipper/innkeeper"
+)
+
+type WriteClient interface {
```
aryszka: why is this public?

cmd/eskip/writeclient.go
```diff
@@ -0,0 +1,44 @@
...
+	UpsertAll(routes eskip.RouteList) error
+	// delete all items in 'routes' that fulfil 'cond'.
+	DeleteAllIf(routes eskip.RouteList, cond eskip.RoutePredicate) error
+}
+
+var (
+	invalidOutput = errors.New("invalid output")
+)
+
+func createWriteClient(out *medium) (WriteClient, error) {
```
aryszka: see me comments for the ReadClient, this seems to be a similar case.

eskip/eskip.go
```diff
@@ -96,6 +96,21 @@ type Route struct {
 	Backend string
 }
 
+type RoutePredicate func(*Route) bool
+
+type RouteList []*Route
```
aryszka: when reading other go code, I tend to prefer not having empty type aliases like this. It makes sense when it has its own methods appended, or just a number type gets a name for some constants or binary flags, but for such cases recently started preferring the non-aliased format. A slice of routes is just that: a slice of routes. What do you think, would it make sense to remove this alias? As I checked, we don't have these now in the public interfaces of the packages, and I remember removing a lot of my own ones during an earlier refactoring.

aryszka: ooops, just realized, there is a type routeList []*eskip.Route in master, but it is not public. Does it really have to be public?

routeid/routeid.go
```diff
@@ -0,0 +1,29 @@
+package routeid
```
aryszka: this is not useful as a standalone package for anybody. Too trivial logic.

routing/testdataclient/dataclient.go
```diff
@@ -103,6 +103,16 @@ func (c *Client) LoadUpdate() ([]*eskip.Route, []string, error) {
 	return u, d, nil
 }
 
+func (c *Client) UpsertAll(routes eskip.RouteList) error {
```
aryszka: why is this method required?

routing/testdataclient/dataclient.go
```diff
@@ -103,6 +103,16 @@ func (c *Client) LoadUpdate() ([]*eskip.Route, []string, error) {
...
+	// TODO implement this
+	return nil
+}
+
+func (c *Client) DeleteAllIf(routes eskip.RouteList, cond eskip.RoutePredicate) error {
```
aryszka: why is this method required?

cmd/eskip/readclient.go
```diff
@@ -0,0 +1,127 @@
...
+	routes string
+}
+
+type idsReader struct {
+	ids []string
+}
+
+// TODO eliminate duplicate code in write and read clients
+func createReadClient(m *medium) (ReadClient, error) {
+	// no out put, no client
```
aryszka: output, single word?

cmd/eskip/writeclient.go
```diff
@@ -0,0 +1,44 @@
...
+	// no out put, no client
```
aryszka: output single word?

cmd/eskip/writeclient.go
```diff
@@ -0,0 +1,44 @@
+package main
+
+import "github.com/zalando/skipper/eskip"
```
aryszka: please make it a single import block

cmd/eskip/writeclient.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"errors"
+	etcdclient "github.com/zalando/skipper/etcd"
```
aryszka: i think these aliases are very useful here. They are longer, they don't solve any conflict, and just make searching in the code harder.

cmd/eskip/writeclient.go
```diff
@@ -0,0 +1,44 @@
...
+		return nil, nil
+	}
+
+	switch out.typ {
+	case innkeeper:
+		auth := innkeeperclient.CreateInnkeeperAuthentication(innkeeperclient.AuthOptions{InnkeeperAuthToken: out.oauthToken})
+
+		ic, err := innkeeperclient.New(innkeeperclient.Options{
+			Address:        out.urls[0].String(),
+			Insecure:       true,
```
aryszka: i guess this hardcoded true stayed here by accident

cmd/eskip/doc.go
```diff
@@ -70,6 +70,8 @@ const (
 	// flag usage strings:
```
aryszka: please remove "On the roadmap: Innkeeper support." from this doc.

eskipfile/file.go
```diff
@@ -45,9 +45,16 @@ func Open(path string) (*Client, error) {
 	return &Client{routes}, nil
 }
 
+func (c Client) LoadAndParseAll() (routeInfos []*eskip.RouteInfo, err error) {
```
aryszka: 'routeInfos' here is visible in the docs, due to LoadAndParseAll is public, info doesn't have plural form. Please, give it a better name.

cmd/eskip/readclient.go
aryszka: ...and it seems that most implementations of the LoadUpdate function noop. Please remove it.

cmd/eskip/readclient.go
```diff
@@ -0,0 +1,127 @@
...
+			Authentication: auth})
+
+		if err != nil {
+			return nil, err
+		}
+		return ic, nil
+
+	case etcd:
+		client := etcdclient.New(urlsToStrings(m.urls), m.path)
+		return client, nil
```
aryszka: why not just return the result of etcdclient.New?

etcd/etcd.go
```diff
@@ -246,3 +236,29 @@ func (c *Client) Delete(id string) error {
...
+		r.Id = routeid.GenerateIfNeeded(r.Id)
+		err := c.Upsert(r)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *Client) DeleteAllIf(routes eskip.RouteList, cond eskip.RoutePredicate) error {
```
aryszka: this function seems like a convenience function, and it is used only in the eskip tool. Why not put there where it is used?

etcd/etcd.go
```diff
@@ -246,3 +236,29 @@ func (c *Client) Delete(id string) error {
 
 	return err
 }
+
+func (c *Client) UpsertAll(routes eskip.RouteList) error {
```
aryszka: the same as DeleteAllIf

cmd/eskip/writeclient.go
```diff
@@ -0,0 +1,44 @@
...
+	innkeeperclient "github.com/zalando/skipper/innkeeper"
+)
+
+type WriteClient interface {
+	UpsertAll(routes eskip.RouteList) error
+	// delete all items in 'routes' that fulfil 'cond'.
+	DeleteAllIf(routes eskip.RouteList, cond eskip.RoutePredicate) error
+}
+
+var (
```
aryszka: single var, no need for parens

cmd/eskip/eskip_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+func TestGetCommandFail(t *testing.T) {
+	_, err := getCommand([]string{"some", "hello"})
+	if err != invalidCommand {
+		t.Error("hello is an invalid command")
+	}
+}
+
+func TestGetCommandEmpty(t *testing.T) {
+	_, err := getCommand([]string{})
```
aryszka: this is used in the tested code with os.Args. os.Args will never be empty, because the command for the os will be always the first arg. Better to test here something like `[]string{"some"}`.

cmd/eskip/mediadefaults.go
```diff
@@ -0,0 +1,35 @@
+package main
```
aryszka: this code feels a bit like an overkill. Wouldn't it be simpler as a single function? E.g:

```
func defaultMedia(cmd command, in, out *medium) (*medium, *medium, error) {
    var err error
    switch cmd {
    case check, print:
        if in == nil {
            in, err = processEtcdArgs(defaultEtcdUrls, defaultEtcdPrefix)
        }
    case upsert, reset, delete:
        if out == nil {
            out, err = processEtcdArgs(defaultEtcdUrls, defaultEtcdPrefix)
        }
    }

    return in, out, err
}
```

cmd/eskip/media.go
```diff
@@ -93,25 +103,36 @@ func validateSelectWrite(cmd command, media []*medium) (input, output *medium, e
...
-	switch cmd {
-	case check, print:
-		return validateSelectRead(media)
-	case upsert, reset, delete:
-		return validateSelectWrite(cmd, media)
-	default:
-		return nil, nil, invalidCommand
-	}
+	// cmd should be present and valid
+	return commandToValidations[cmd](media)
```
aryszka: isn't commandToValidations only used here? maybe a common `switch` is the right tool here?

eskip/eskip.go
```diff
@@ -96,6 +96,21 @@ type Route struct {
...
+
+// RouteInfo contains a route id, plus the loaded and parsed route or
+// the parse error in case of failure.
+type RouteInfo struct {
```
aryszka: this type definition feels a bit orphan in this package, not sure if this is the right place. I remember we discussed it, but. The reason why it is currently in the etcd package is that in etcd, being a generic storage with an open API, one can store invalid routes or just any data, and depending on the client, it can be useful to report parse errors individually. Currently, this is only used by the eskip tool, that verifies the whether one or more route definitions are valid. The runtime configuration rejects the whole set of routes, in case of any one of them is invalid, because that indicates more serious problems than just a single parse error. But this makes sense only in case of etcd, because as I checked, the recent Innkeeper responses, once we have a valid json response, they will be valid eskip routes. See the `LoadAndParseAll` in innkeeper/client.go of this pull-request, it never sets a parse error. So, I think we should use this type only in the etcd client.

eskip/eskip.go
```diff
@@ -96,6 +96,21 @@ type Route struct {
 	Backend string
 }
 
+type RoutePredicate func(*Route) bool
```
aryszka: this is a trivial type and not used by this package. Also, since I am not sure yet, if the DeleteAllIf methods in the other packages are really need to be public and in those packages, I think this is not required here.

eskipfile/file.go
```diff
@@ -45,9 +45,16 @@ func Open(path string) (*Client, error) {
...
+	for _, route := range c.routes {
+		routeInfos = append(routeInfos, &eskip.RouteInfo{*route, nil})
```
aryszka: similarly to the other places regarding the WriteClient and ReadClient interfaces, this feels a bit of a type-fu.

innkeeper/clientauth.go
```diff
@@ -0,0 +1,32 @@
+package innkeeper
+
+import (
+	"github.com/zalando/skipper/oauth"
+)
+
+// An Authentication object provides authentication to Innkeeper.
+type Authentication interface {
```
aryszka: if the innkeeper package will import the oauth package directly, then this public interface doesn't seem to make much sense. Am I wrong? The original goal was to have less dependencies between the packages, but it's fine, but then we don't really need the extra interface.

innkeeper/clientauth_test.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+func TestCreateInnkeeperAuthenticationClient(t *testing.T) {
+	options := AuthOptions{InnkeeperAuthToken: "",
+		OAuthCredentialsDir: "dir",
+		OAuthUrl:            "url",
+		OAuthScope:          "scope"}
+	auth := CreateInnkeeperAuthentication(options)
+
+	if reflect.TypeOf(auth).String() != "*oauth.OAuthClient" {
```
aryszka: unsee :)

```
if _, ok := auth.(*oauth.OAuthClient); !ok {
    t.Error("wrong authentication client type")
}
```

innkeeper/clientauth_test.go
```diff
@@ -0,0 +1,27 @@
...
+		t.Error("wrong fixed token")
```
aryszka: wrong error message

innkeeper/client.go
```diff
@@ -235,7 +233,7 @@ func convertRoute(id string, d *routeData, preRouteFilters, postRouteFilters []*
 }
 
 // Converts a set of Innkeeper route definitions to their eskip representation.
-func convertData(data []*routeData, preRouteFilters, postRouteFilters []*eskip.Filter) ([]*eskip.Route, []string, string) {
+func convertJsonToEskip(data []*routeData, preRouteFilters, postRouteFilters []*eskip.Filter) ([]*eskip.Route, []string, string) {
```
aryszka: question: do we still need the pre/post-route filters? Now innkeeper, with the new interface, could full control this matter.

innkeeper/client.go
```diff
@@ -320,6 +318,43 @@ func getHttpError(r *http.Response) (error, bool) {
...
+	req.Header.Add(authHeaderName, authToken)
+	req.Header.Set("Content-Type", "application/json")
+	response, err := c.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+
+	defer response.Body.Close()
+
+	if response.StatusCode == http.StatusUnauthorized || response.StatusCode == http.StatusBadRequest {
```
aryszka: wouldn't it be better to check for all non-success status codes?

innkeeper/client.go
```diff
@@ -320,6 +318,43 @@ func getHttpError(r *http.Response) (error, bool) {
...
+
+	res, err := json.Marshal(route)
+
+	req, err := http.NewRequest("POST", url, bytes.NewReader(res))
+
+	if err != nil {
+		return err
+	}
+
+	authToken, err := c.opts.Authentication.GetToken()
```
aryszka: wouldn't it be better to cache the token? like in case of `requestData`? because this may mean to call the oauth service on every single route being uploaded.

innkeeper/client.go
```diff
@@ -374,14 +409,31 @@ func (c *Client) requestData(authRetry bool, url string) ([]*routeData, error) {
...
+	routeList, err := c.LoadAll()
+	if err != nil {
+		return nil, err
+	}
+
+	routeInfos := []*eskip.RouteInfo{}
+
+	for _, route := range routeList {
+		routeInfo := &eskip.RouteInfo{
+			*route, nil}
```
aryszka: see my comments for the `RouteInfo` type, in the eskip package per this pull request. I think this `LoadAndParseAll` makes only sense in case of the etcd client. (the parser error is `nil` in this line.)

innkeeper/client.go
```diff
@@ -390,16 +442,34 @@ func (c *Client) LoadAll() ([]*eskip.Route, error) {
...
 
 	return routes, deleted, nil
 }
+
+func (c *Client) UpsertAll(routes eskip.RouteList) error {
+	// convert the routes to the innkeeper json structs
+	data := convertEskipToInnkeeper(routes)
+
+	for _, route := range data {
+		err := c.writeRoute(c.opts.Address+allRoutesPath, route)
```
aryszka: has the Innkeeper API changed? If it still only does insert and delete, then this function shouldn't be called upsert. Upsert means update if it exists, or insert if it doesn't.

innkeeper/client.go
```diff
@@ -390,16 +442,34 @@ func (c *Client) LoadAll() ([]*eskip.Route, error) {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (c *Client) DeleteAllIf(routes eskip.RouteList, cond eskip.RoutePredicate) error {
+	return nil
```
aryszka: i guess this one here needs to be implemented yet for the eskip tool Innkeeper integration ;)

innkeeper/eskiptojson.go
```diff
@@ -0,0 +1,125 @@
...
+	var (
+		pathMatch     string
+		pathMatchType matchType
+	)
+
+	if r.Path != "" {
+		pathMatch = r.Path
+		pathMatchType = matchStrict
+	} else if len(r.PathRegexps) > 0 {
+		// TODO we should only have one path regexp
```
aryszka: agree, this could be changed generally. Preferably in a separate pull request. We can shoot a `log.Warning` until then, in the cases of multiple PathRegexps.

innkeeper/eskiptojson.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func convertMethod(r *eskip.Route) string {
+	return r.Method
+}
+
+func convertHost(r *eskip.Route) (host string) {
+	if len(r.HostRegexps) > 0 {
+		// we take the first one
+		// TODO HostRegexps should not be an array
```
aryszka: see my comment for pathregexps.

innkeeper/eskiptojson_test.go
```diff
@@ -0,0 +1,188 @@
+package innkeeper
+
+import (
+	"github.com/coreos/etcd/Godeps/_workspace/src/github.com/stretchr/testify/assert"
```
aryszka: please don't import packages from the Godeps workspace, neither ours, especially not from the imported packages' Godeps. IDE hiccup?

On the other hand: there was a reason why we didn't include testing frameworks. Please see, https://golang.org/doc/faq#testing_framework. In this particular case, I recommend to use table testing, that I already saw applied in other tests in this pull request.

cmd/eskip/eskip.go
danpersa: what's the problem with it? :)

cmd/eskip/readclient.go
danpersa: I decoupled the place where we create the stateless object from the place where we use them in order to follow the [Dependency Inversion Principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle). Separation so concerns is the second principle I'd like to follow. It is really important when building complex systems to decouple concerns. Looking at our example, after separating the client creation concern and the loading concern: the readability is improved, as the load.go file is not much smaller. Also, as soon as you want to add a new reader, you don't have to understand what everything happens in the load file. You just create a new reader, in the separate file. Testability is improved, as you are now able to mock the ReadClient.

The [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single_responsibility_principle) is the second princile I'd like to respect. And abstraction which needs to load routes, should only do one thing load the routes, as it's name suggests. Creating a client is another concern, and it's the responsibility of another abstraction.

Regarding your comments:
- yes, the switch is still there, but you don't have to dig inside of the load.go file to find it. As soon as you need to deal with a readclient, you open its file and add, or modify a client.
- yes, the code has a few more lines. But now it's consistent. Before the change, we were many kinds of abstractions in order to read the lines. For innkeeper, etcd and file we had the DataClient interface, and for the rest we had inline implementations. Now, when we want to read  routes, we have one abstraction: ReadClient. And having less lines of code doesn't make the code more understandable (see bad written Scala :P)

cmd/eskip/readclient.go
danpersa: I'll make it private

cmd/eskip/eskip.go
aryszka: well, i am asking for the reason why is it there. If there is no reason for it, then it shouldn't be there.

cmd/eskip/readclient.go
danpersa: ups... I forgot it here... my initial intention was to completely replace the DataClient with a ReadClient and a WriteClient. That's also the reason for making them public :)

cmd/eskip/eskip.go
danpersa: I have the same comment

cmd/eskip/writeclient.go
danpersa: fixed

cmd/eskip/writeclient.go
danpersa: see the answer of your comment

eskip/eskip.go
danpersa: it has to be public, as it's used in other packages. If you want to remove it, we can open an issue.

routeid/routeid.go
danpersa: it's used in both the innkeeper and etcd packages. I created a new package in order not to create circular dependencies between packages.

routing/testdataclient/dataclient.go
danpersa: removed

routing/testdataclient/dataclient.go
danpersa: removed

cmd/eskip/readclient.go
danpersa: removed

cmd/eskip/writeclient.go
danpersa: fixed

cmd/eskip/writeclient.go
danpersa: fixed

cmd/eskip/writeclient.go
danpersa: I will read you comment as "they are **not** very useful"

actually, somebody else defined this aliases, I only moved them. There is an innkeeper variable in the `main` package in conflict with the package name.

cmd/eskip/writeclient.go
danpersa: I will create a ticket for this, as we need a new flag in eskip

danpersa: https://github.com/zalando/skipper/issues/80

eskipfile/file.go
danpersa: from: https://en.wiktionary.org/wiki/information

    information ‎(usually uncountable, plural informations)

cmd/eskip/readclient.go
danpersa: fixed

cmd/eskip/readclient.go
aryszka: i deeply disagree with most of this in the context of this particular code.
- the dependency inversion principle shouldn't be about some simple functions, but more complex structures. It looks like you are introducing here unnecessary dependencies, just to have an excuse to apply a principle that is unnecessary in this case.
- you write "separation of concerns" is "important when building complex systems". That's right, but for one, this tool is not a complex system. For second, the explanation of the example sounds wrong to me because the client creation concern is introduced for many of the cases where this concern was not there before (for more cases than there was a client), also because the readability is not improved (just read this code), as while the size of the load file got smaller, the total size of the code got much bigger, also because it is not true that you don't have to understand what everything happens in the load file to add a reader, one shouldn't add a reader without understanding something as simple what happens in the load file, and also because the the testability hasn't improved by being able to mock the read client, because the read client wasn't there to start with and there was no need to mock it.
- single responsibility principle: creating a client is not another concern when there is no need to create a client
- i am looking at load.go in master and i don't get the last two comments from the reply. Which abstractions do you mean? Why the files? The load.go in master has 181 lines including the license notice, and the longest function has 21 lines. Aren't we overkilling here something with the ReadClient?

etcd/etcd.go
danpersa: I deal with this method as soon as I start dealing with the delete for innkeeper https://github.com/zalando/skipper/issues/81

etcd/etcd.go
danpersa: I deal with this method as soon as I start dealing with the delete for innkeeper #81

cmd/eskip/eskip_test.go
danpersa: done

cmd/eskip/readclient.go
aryszka: please, just don't replace it yet. Go interfaces are quite different than java interfaces, and we should discuss it first.

In general, i think it would be nicer adding small features and doing full refactorings in separate pull requests. The design considerations for the latter could be discussed before making so big changes, too.

cmd/eskip/eskip.go
aryszka: what do you mean?

eskip/eskip.go
aryszka: but it wasn't used in other packages as it is only being introduced here. What I mean is that it was only used in the cmd/eskip package, and wasn't part of the public interface.

routeid/routeid.go
aryszka: options:
- for something as trivial as this, even code duplication is fine
- it can be put into a package that already exists and used by both depending packages.

Please don't create a package for a single function, either above solution is better.

cmd/eskip/writeclient.go
aryszka: yep, sorry i mean as 'not very useful'. I see now, it is because of the constants `etcd` and `innkeeper`. (y)

cmd/eskip/writeclient.go
aryszka: why can't we create it within this pull request? then we wouldn't publish insecure solution.

eskipfile/file.go
aryszka: well, it says usually uncountable, and I don't know if this is an unusual case. And any other more detailed explanation says that the plural form is wrong. And just asked a native, and she says it's wrong. Are you sure that it's good here?

cmd/eskip/mediadefaults.go
danpersa: I prefer the the version I implemented as it's really close to the StrategyPattern

cmd/eskip/media.go
danpersa: we already know where each command (print, upsert) goes, so we don't need to decide this every time we call it. The advantage of having separate functions is also that we give name to concepts. So the readability is improvoed, as we split the commands into Read and Write commands.

eskip/eskip.go
danpersa: I think we should add parsing errors to innkeeper also. I know it's not there now, but I plan to do it...

eskip/eskip.go
danpersa: I created this ticket https://github.com/zalando/skipper/issues/81 for deleting

cmd/eskip/mediadefaults.go
aryszka: as the suggested code above should show, too, i think the strategy pattern is a big overkill for this case. we're just setting default output for write commands and default input for read commands, and nothing else.

cmd/eskip/media.go
aryszka: i am not talking about the separate functions, those are already there anyway. i am talking about the commandToValidations map. it introduces a separate package level variable, instead of the switch.

eskip/eskip.go
aryszka: can there be parsing errors in the innkeeper client? in which situations? if the innkeeper response is a single json document, how do identify which individual route's parsing failed? how can unparseable routes get into Innkeeper's storage?

innkeeper/clientauth.go
danpersa: I guess we can think about it, and fix it in another pull request.

My feeling is that removing abstractions, makes code more monolithic and starting to break the SRP and DIP.

innkeeper/clientauth_test.go
danpersa: yes, that's what I was looking for! thanks :)

innkeeper/clientauth.go
aryszka: well, that's ok. About srp and dip, to me it feels that the current one is more both srp and dip, since currently the innkeeper client doesn't directly depend on the oauth package (only in the documentation :))

innkeeper/clientauth_test.go
danpersa: fixed

innkeeper/client.go
danpersa: Created https://github.com/zalando/skipper/issues/82

innkeeper/client.go
danpersa: we use a FixedToken here... the token is provided as a command line arg

innkeeper/client.go
danpersa: I want to have error parsing for innkeeper also...

danpersa: https://github.com/zalando/skipper/issues/83

innkeeper/client.go
danpersa: the upsert method is part of the interface... and it does upsert for etcd for example...

innkeeper/client.go
danpersa: https://github.com/zalando/skipper/issues/81

innkeeper/eskiptojson.go
danpersa: definitely a new pull request... this one got but enough https://github.com/zalando/skipper/issues/84

innkeeper/eskiptojson.go
danpersa: https://github.com/zalando/skipper/issues/84

innkeeper/eskiptojson_test.go
danpersa: fixed the import.

for this particular case, the asserts make the code more readable.

putting there a dozen of ifs would really make the code ugly.

table testing doesn't fit, as we check for different stuff in each of the tests...

eskip/eskip.go
danpersa: Created this one to investigate: https://github.com/zalando/skipper/issues/83

eskipfile/file.go
danpersa: I included your concern here: https://github.com/zalando/skipper/issues/83

cmd/eskip/eskip.go
danpersa: removed

cmd/eskip/readclient.go
danpersa: probably we'll not agree on this one :P

cmd/eskip/readclient.go
danpersa: I will not modify it in this pull request ;)

cmd/eskip/eskip.go
danpersa: What's I'd like to achieve here in the future is the total opposite of what you're saying. I'd like to have the commands only get a read, and write client as parameters. This completely decouples the command by the `medium`. A command doesn't need to know about the `medium`, it only needs a way to read and a way to write routes. And this is represented by the ReadClient and WriteClient.

eskip/eskip.go
danpersa: removed the RouteList

cmd/eskip/writeclient.go
danpersa: fixed

cmd/eskip/doc.go
danpersa: done

cmd/eskip/media.go
danpersa: I guess each solution has it's downsides and advantages :)

innkeeper/client.go
danpersa: this is right now in sync with innkeeper, as it only return those status codes...

routeid/routeid.go
danpersa: * are you suggesting of copy pasting the same code, together with the tests in two different places? I really can't do that.
* there not such common package

I don't think we should be afraid of creating new packages.

cmd/eskip/eskip.go
aryszka: cool, but still why not to handle both the createWriteClient and the createReadClient the same way? either both could be here or else.

routeid/routeid.go
aryszka: which test? what I mean in case option one, is that one `if` statement and random string generation with limited set of characters can be copypasted, yes, and doesn't even need to be separately tested. what needs to be tested is the calling code, that whatever it does, the route gets an id.

for the second option, the eskip package is fine, because that's where the route type and the id field is defined.

No, I am not afraid of creating new packages, as you can see by the packages I already created. See what I said above: please don't create a package with a single function. To be more precise: please don't create a new package with a single function that is as trivial as this.

cmd/eskip/media.go
aryszka: what are the downsides of the switch in this case?

innkeeper/client.go
aryszka: The problem with doing this is that in case of a 500, the calling code will believe that the request was successful, and can't even make a log entry or anything.

innkeeper/client.go
aryszka: i think this is not right. This client accepts currently anything that implements the Authentication interface. It is true that it can be the FixedToken but it also can be anything else. For the integrity of the package, the package should make sure that it properly deals with what it offers.

innkeeper/client.go
aryszka: in what cases can innkeeper routes have individual parsing errors?

innkeeper/client.go
aryszka: which interface? also etcd does do an actual upsert.
Am I correct that this Innkeeper API call does not do an upsert? If not, please don't call this function an upsert.

innkeeper/eskiptojson.go
aryszka: ok, can you please add log.Warning until then, in case there is any data loss?

routeid/routeid.go
danpersa: Moved routeid to eskip package

cmd/eskip/media.go
danpersa: A switch means that we want to make a decision. In our case, we don't need to make any decisions, as we know at compile time that for upsert we want to call the `validateSelectWrite` procedure. I think the map expresses this better.

The downside of the map, as you mentioned, is that we introduces another variable. This should of course, be a constant, but go doesn't allow it.

If you really insist I can do a switch, but I find the map solution nicer :)

eskipfile/file.go
danpersa: I don't get the problem. In order for a client to load and parse the routes, it need to implement the correct interface, isn't it?

innkeeper/client.go
danpersa: To be fixed in another pull request. I created https://github.com/zalando/skipper/issues/85 so we won't forget.

innkeeper/client.go
danpersa: As the client accepts anything that implements the Authentication interface, this is the contract between the two components. Of course, the token can be valid or not and we take this into account by returning a proper error.

innkeeper/client.go
danpersa: I guess it's up to the one who works at #83 to figure it out. In case everything is always perfect, we will remove the RouteInfo as you suggested. The goal of this #83 ticket and other tickets I created is to extract the non critical stuff out of this already monolithic pull request. So we can deal with each smaller problem individually after we merge this one.

innkeeper/client.go
danpersa: I was talking about the `writeClient` interface. Our write client should be consistent. We should decide if we want to do the upsert for all of the clients or not.

innkeeper/eskiptojson.go
danpersa: I added

    log.Println("Warn: We should only have one path regexp")

as .Warning does not exist in the `log` package

innkeeper/eskiptojson.go
danpersa: I added the warn

cmd/eskip/eskip.go
danpersa: I extracted the creation of the clients in the main function

cmd/eskip/media.go
mo-gr: To me, those two maps (`commandToValidations` and `commands` in eskip.go) smell a lot like some weird inside-out data structures (call it OO if you insist). I think, `command` should not be a type alias for a string anymore but instead be a proper struct containing a command-name (the current command), a `validation` function pointer and a `commandFunc` function pointer. This way, the weird map lookups would go away and we could do proper validations and command execution dispatched on the actual command.

cmd/eskip/mediadefaults.go
mo-gr: I kind of agree with dan here... Having a defaultWrite/defaultRead and assigning them to the commands is more readable to me. But I would mix in my comment https://github.com/zalando/skipper/pull/76/files#r48719213 here as well to get rid of the map. All those maps from `string->someFunc` are a smell in my nose.

cmd/eskip/readclient.go
```diff
@@ -0,0 +1,112 @@
...
+
+type inlineReader struct {
+	routes string
+}
+
+type idsReader struct {
+	ids []string
+}
+
+// TODO eliminate duplicate code in write and read clients
```
mo-gr: Are you planing to do this before or after this PR is merged?

cmd/eskip/readclient.go
```diff
@@ -0,0 +1,112 @@
...
+
+		ic, err := innkeeperclient.New(innkeeperclient.Options{
+			Address:        m.urls[0].String(),
+			Insecure:       true,
+			Authentication: auth})
+
+		if err != nil {
+			return nil, err
+		}
+		return ic, nil
```
mo-gr: I'd prefer a named function here to keep all the case-branches one-liners

eskip/eskip.go
mo-gr: As long as we use this type somewhere, I think a type alias like this one makes code more readable (even, or especially, for simple types like this)

cmd/eskip/readclient.go
danpersa: done

cmd/eskip/readclient.go
danpersa: extracting that function fixed this also ;)

cmd/eskip/media.go
danpersa: @mo-gr, could you do a pull request to this branch with this small change? :)

mo-gr: "small change"... -_-

innkeeper/client.go
mo-gr: I agree with dan. I would not add caching of tokens here. If the thing, that provides the token would have to make an expensive call, it is the responsibility of that thing to cache.

innkeeper/client.go
mo-gr: I agree with arpad here. So +1 for #85

mo-gr:
I agree, that we should try to merge this epic PR and handle the remaining things in separate PRs. From my point of view, there is no blocker open anymore. I will do a few tests locally to make sure, this branch won't break anything and then just merge it so we can all move on to smaller PRs.

cmd/eskip/media.go
mo-gr: Do you agree in general? Then we should just create an issue for that

danpersa: Yes, I created #86

cmd/eskip/eskip.go
mo-gr: aligned

cmd/eskip/media.go
mo-gr: And here it is #86

mo-gr:
From my point of view, all issues are either fixed or moved to separate issues. This PR doesn't seem to break anything. So I will merge it now. Let's try to keep our PRs a bit smaller from now on ;-)

PR #77, opened 2015-12-08
fix header setting in the header filter

aryszka:


lmineiro:
We can fix it later in the way we discussed. I believe the best fix now is #78

aryszka:
this is not done yet, only created to check the part in the proxy, if it is the same as what we talked about.

PR #78, opened 2015-12-08
Always set the Host header from the original request by default

lmineiro:
This single line is required to reuse the Host from the original request. This was our original expectation.
As we found out, the new request is built with the Host from the URL.

Go's SDK doesn't actually have the Host header in the Headers.

After we refactor the header filter we can still provide the option to change the host in the upstream request.


mo-gr:
I think, go's behaviour is common, since the Host is not optional in HTTP/1.1. Therefore lots of libraries see it as part of the request instead of a header. Anyways... fix looks good to me

PR #90, opened 2016-01-14
Custom predicates

aryszka:


doc.go
```diff
@@ -177,7 +179,54 @@ cmd/eskip command package, and/or enter in the command line:
...
-library and extended with custom filters and/or custom data sources.
+library and extended with custom filters, predicates and data sources.
+
+
+Custom Predicates
+
+To create a custom predicate, the PredicateSpec of the routing package
+needs to be implemented. The PredicateSpec is used to create the actual
+predicate with concrete arguments, during the route definitions are
+processed by the routing package.
```
lmineiro: Confusing english here, what exactly is the idea?

eskip/eskip.go
```diff
@@ -111,76 +139,100 @@ type RouteInfo struct {
...
+		if err != nil {
+			return err
 		}
-	}
 
-	return ss, nil
-}
+		switch m.name {
+		case "Path":
+			if pathSet {
```
lmineiro: When does it get true?

eskip/eskip.go
```diff
@@ -111,76 +139,100 @@ type RouteInfo struct {
...
+		case "Host":
+			if args, err = getStringArgs(1, m.args); err == nil {
+				route.HostRegexps = append(route.HostRegexps, args[0])
+			}
+		case "PathRegexp":
+			if args, err = getStringArgs(1, m.args); err == nil {
+				route.PathRegexps = append(route.PathRegexps, args[0])
+			}
+		case "Method":
+			if methodSet {
```
lmineiro: When does it get true?

eskip/eskip.go
```diff
@@ -44,13 +53,28 @@ type parsedRoute struct {
 	backend  string
 }
 
+// A CustomPredicate object represents a parsed, in-memory, route matching predicate
+// that is defined by extensions.
+type CustomPredicate struct {
```
mo-gr: Why call it `CustomPredicate`? What would be the drawback of only having `Predicate` and implementing the default ones exactly like those (only that they are bundled)?

eskip/eskip.go
```diff
@@ -85,6 +109,10 @@ type Route struct {
 	// E.g. HeaderRegexp("Accept", /\Wapplication\/json\W/)
 	HeaderRegexps map[string][]string
 
+	// Custom predicates to match.
+	// E.g. Traffic(.3)
+	CustomPredicates []*CustomPredicate
```
mo-gr: To expand on my previous suggestion, this would only be `Predicates []*Predicate` Seems cleaner to me

eskip/eskip.go
aryszka: makes sense, in general. i see two things with that atm:
- it would probably make it a quite big pr. Rather a refactor than a feature.
- the path predicate has a distinguished role

eskip/eskip.go
aryszka: thx, true, and the tests are missing

eskip/eskip.go
aryszka: thx, true, and the tests are missing

doc.go
aryszka: yeah... (shyface)

one needs to provide the 'factory' object, the spec, and its create method should return the actual predicate object, and it is called when the routing tree is constructed. can we hammer it out together?

eskip/eskip.go
lmineiro: I also vote for a simplified Predicate. I thought this to be new code, are there already any dependencies on `CustomPredicate`?

aryszka: it's new code. It can be renamed to Predicate, not an issue. What would be too big, is unifying the already existing predicates into the same general Predicate type. A good idea, but a separate pr.

eskip/eskip.go
aryszka: fixed

eskip/eskip.go
aryszka: fixed

doc.go
aryszka: gave it a try to be a bit more clear.

eskip/eskip.go
aryszka: so.. now i reduced a bit the usage of the term 'Custom', because what is custom. On the other hand, while I it is true that it was a bit short-sighted back then to hardcode the types of the prepacked predicates, and it would be much nicer to just provide a built-in set of commonly used predicate implementations, this change can easily become an epic refactor, also hard to do it in a backwards compatible way, so I think we should do it in a separate PR. Can't wait for it though :)

aryszka:
the .String() was forgotten, now added it

eskip/eskip.go
```diff
@@ -44,13 +53,28 @@ type parsedRoute struct {
...
+
+	// The name of the custom predicate as referenced
+	// in the route definition. E.g. 'Foo'.
+	Name string
+
+	// The arguments of the predicate as defined in the
+	// route definition. The arguments can be of type
+	// float64 or string (string for both strings and
+	// regular expressions).
+	Args []interface{}
```
danpersa: this is the `Either` in scala... or a `float64 | string` in [Ceylon](http://ceylon-lang.org/documentation/1.2/reference/structure/type/) :)

aryszka: that's right

lmineiro: ... and a Variant in Visual Basic! Yay!

mo-gr: Also inwoon as a *sum type* in theory

aryszka: everything's just an n-bit register at the end of the day. can i merge?

lmineiro:
lgtm

PR #91, opened 2016-01-15
Added possibility to break the filter chain

lmineiro:


filters/filters.go
```diff
@@ -38,10 +38,13 @@ type FilterContext interface {
 	// route.
 	Served() bool
```
aryszka: sure that we need this? maybe we can mark it deprecated, too, and until removed, use it only in the context of the old MarkServed() way? just wondering.

proxy/proxy.go
```diff
@@ -323,61 +327,62 @@ func (p *proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
 	metrics.MeasureRouteLookup(start)
 
 	start = time.Now()
-	f := rt.Filters
+	routeFilters := rt.Filters
 	c := newFilterContext(w, r, params, p.preserveOriginal, rt)
-	p.applyFiltersToRequest(f, c)
+	processedFilters := p.applyFiltersToRequest(routeFilters, c)
```
aryszka: i see you found a way to do it nice without an actual stack :+1:

aryszka:
i like it!

filters/filters.go
lmineiro: We don't need it, indeed. It should be only the internal flag. I'll add the deprecation documentation

lmineiro: The proxy needs to check if the request has been served already. Should we switch to another strategy like, checking if there is already a Response object?

mo-gr: If there is an impliciit way to determine the state of this, we should use it. Otherwise it is unnecessary data-denormalisation and will only lead to bugs in the long run

lmineiro: Well, this is just the interface. The proxy package provides its own filterContext implementation and it can share the state in any other way, available only to the package.

I'll deprecate this method and figure out a way for the proxy's internal filterContext

proxy/proxy.go
```diff
@@ -323,61 +327,62 @@ func (p *proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		start = time.Now()
+		if rt.Shunt {
+			rs = shunt(r)
+		} else {
+			rs, err = p.roundtrip(r, rt)
 			if err != nil {
+				http.Error(w,
+					http.StatusText(http.StatusInternalServerError),
+					http.StatusInternalServerError)
 				log.Error(err)
```
mo-gr: Errors don't add branding. Intentionally?

proxy/proxy_test.go
```diff
@@ -390,6 +390,79 @@ func TestAppliesFilters(t *testing.T) {
...
+
+	if _, has := r.Header["X-Unexpected"]; has {
+		t.Error("Request has an unexpected header from a filter after the breaker in the chain")
+	}
+
+	if _, has := w.Header()["X-Unexpected"]; has {
+		t.Error("Response has an unexpected header from a filter after the breaker in the chain")
+	}
+
+	if w.Code != http.StatusUnauthorized && w.Body.String() != "Impossible body" {
```
mo-gr: I would make that two separate ifs. One for statusCode and one for the body

mo-gr:
:+1:

PR #93, opened 2016-01-15
fine tune deprected filter shunting

aryszka:


proxy/proxy.go
```diff
@@ -375,21 +376,23 @@ func (p *proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	start = time.Now()
-	addBranding(response.Header)
-	copyHeader(w.Header(), response.Header)
-	w.WriteHeader(response.StatusCode)
-	err := copyStream(w.(flusherWriter), response.Body)
-	if err != nil {
-		log.Error(err)
-	} else {
-		metrics.MeasureResponse(response.StatusCode, r.Method, rt.Id, start)
+	if !c.served {
```
aryszka: this is to fully support the deprecated functionality without breaking old stuff

proxy/proxy.go
```diff
@@ -375,21 +376,23 @@ func (p *proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	start = time.Now()
-	response := c.Response()
-	if p.preserveOriginal {
-		c.originalResponse = cloneResponseMetadata(response)
+	if !c.served && p.preserveOriginal {
```
aryszka: this takes into account, that filters on the response chain can also call Serve() and override the response from the backend or from the previous 'serving' filter

PR #94, opened 2016-01-17
Etcd client

aryszka:


aryszka:
replacing the deprecated etcd client lib with direct http requests + fixing the update bug that missed updates

etcd/etcd.go
```diff
@@ -183,43 +379,56 @@ func (c *Client) LoadAll() ([]*eskip.Route, error) {
 // or update.
 //
 // It uses etcd's watch functionality that results in blocking this call
-// until the next change is detected in etcd.
+// until the next change is detected in etcd or reaches the configured hard
+// timeout.
 func (c *Client) LoadUpdate() ([]*eskip.Route, []string, error) {
```
aryszka: The reasoning behind the fix for the update bug looks more or less like this:

Want to continue supporting the unified interface that hides if the implementation is short-poll or long-poll.  Would have been happy to go short-polling etcd, but turns out, it doesn't really support it, only if loading everything each time, which can be too much. It supports long-polling, but when long-polling a directory, it returns every updated node in a separate round, which would mean, when returned directly to the routing package, that a new tree would be constructed on every updated route, which, again, would be too much.

Considered three solutions:

1. Do continuous long-poll in the background concurrently, collect the updates, and whenever the routing package requests for updates, return the ones that are collected to this point.

2. Try to guess that a batch of updates belong together based on the X-Etcd-Index header and the individual modifiedIndexes of the nodes.

3. Use a timeout for the long-poll, that relatively large for the etcd connection, but is relatively small for detecting the updates in the routing. When the timeout is reached during the long-polling, consider it as the end of a batch of changes, and return to the calling code.

#1 depends on the calling code to regularly fetch the collected updates, which is a bit dangerous in a way, that it may lead to memleaks. Found that #2 only works when the etcd cluster is not used to store anything else, only these routes, due to how the etcd indexes work. #3 introduces this unnatural timeout value, but the gap between the expected etcd communication timings and the user perception is large enough, and this felt the most robust solution, so I went with #3.

PR #101, opened 2016-01-30
fix etcd timeout when etcd host in proxy mode

aryszka:


routing/datasource.go
```diff
@@ -39,6 +39,24 @@ type incomingData struct {
...
 
+func (d *incomingData) Log() {
+	typeString := "unknown"
+	switch d.typ {
+	case incomingReset:
+		typeString = "reset"
+	case incomingUpdate:
+		typeString = "update"
+	}
+
```
lmineiro: Extract these to
```
func (typ incomingType) String() string {
	switch typ {
	case incomingReset:
		return "reset"
	case incomingUpdate:
		return "update"
	default:
		return "unknown"
		
	}
}
```
https://play.golang.org/p/rOPW1n0ec0

etcd/etcd.go
```diff
@@ -167,6 +170,15 @@ func (c *Client) tryEndpoints(mreq func(string) (*http.Request, error)) (*http.R
...
 
+		if isTimeout(err) {
+			break
+		} else if uerr, ok := err.(*url.Error); ok {
+			if isTimeout(uerr.Err) {
+				err = uerr.Err
+				break
+			}
+		}
+
```
lmineiro: I am probably missing something here. It seems to me that a timeout on the first endpoint prevents you from trying any remaining endpoint.

aryszka: yes, on purpose. the multiple endpoints are tried for the cases of outage or other connection problems. the timeout is not part of these retry causing errors, but it is used to identify the case when no new changes occured for a certain time. Or we can say so, too, that the timeout is not a remote problem, it is enforced on the client side, so it is independent of the health state of the remote endpoint. On a one step higher level, we use the timeouts in this client, to identify batches of updates, because otherwise etcd would notify skipper for each route change one by one.

routing/datasource.go
aryszka: updated, thx

etcd/etcd.go
lmineiro: Understood. It's ugly, but I get it ;-)

aryszka: ugly, i see :)

PR #102, opened 2016-01-30
add insecure flag for eskip connections

aryszka:


aryszka:
@danpersa what do you think, shall we enable the insecure flag for the innkeeper connections, too?

lmineiro:
I think it makes sense for it to affect all the data clients

danpersa:
+1 to enable thelm

danpersa:
lgtm

PR #103, opened 2016-02-01
Use the standard header for innkeeper authorization

danpersa:


innkeeper/client.go
```diff
@@ -334,7 +334,7 @@ func (c *Client) writeRoute(url string, route *routeData) error {
 		return err
 	}
 
-	req.Header.Add(authHeaderName, authToken)
+	req.Header.Add(authHeaderName, "Bearer " + authToken)
```
lmineiro: How does innkeeper handle multiple `Authorization` headers? Should be a `.Set()` instead?

aryszka: true

aryszka: updated

lmineiro:
lgtm

danpersa:
lgtm auch

PR #104, opened 2016-02-01
Update readme.md

LappleApple:
I moved the item about being customizable up to the first paragraph to highlight it better and deleted the item from the list


lmineiro:
lgtm

PR #108, opened 2016-02-10
Source predicate

mo-gr:
This is a first draft of a source predicate, that enables routing
decisions based on the source of a request.

Right now it is build (and registered) as a custom predicate. We could
discuss adding it to the built in predicates or leaving it like that as
a nice and simple example of how to add custom predicates to skipper.

Also, the test cases might not be complete. I'm open to suggestions for
more test cases/corner cases, that are not yet tested.

### DIFF
--- a/predicates/source/source.go
+++ b/predicates/source/source.go
@@ -22,20 +22,20 @@ The source predicate supports one or more IP addresses with or without a netmask
 Examples:
 
     // only match requests from 1.2.3.4
-    example1: Source("1.2.3.4") -> "http://example.org"
+    example1: Source("1.2.3.4") -> "http://example.org";
 
     // only match requests from 1.2.3.0 - 1.2.3.255
-    example2: Source("1.2.3.0/24") -> "http://example.org"
+    example2: Source("1.2.3.0/24") -> "http://example.org";
 
     // only match requests from 1.2.3.4 and the 2.2.2.0/24 network
-    example3: Source("1.2.3.4", "2.2.2.0/24") -> "http://example.org"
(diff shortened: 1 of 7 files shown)


util/net.go
```diff
@@ -0,0 +1,35 @@
+package netutils
```
aryszka: for one we should call the directories containing the packages the same name as the package. this can be useful regarding tooling and human expectations, too. also a custom in go.

on the other hand, i kinda dislike package names like 'util', 'common', etc. Can we call this simply 'net'?

on the third hand, this functionality feels kind of a triviality, or at least a bordercase. do we expect it to be globally useful? do we expect this package to grow with time? is anyone already aware of stuff that could go here during a refactoring?

predicates/source/source.go
```diff
@@ -0,0 +1,94 @@
...
+is available. If the X-Forwarded-For header is not present or does not contain
+a valid source address, the source IP of the incoming request is used for
+matching.
+
+The source predicate supports one or more IP addresses with or without a netmask.
+
+Examples:
+
+    // only match requests from 1.2.3.4
+    example1: Source("1.2.3.4") -> "http://example.org"
```
aryszka: it would be nice to have a semicolon separating the definitions, even if these examples not necessarily are meant as a single document.

util/net.go
mo-gr: I agree with your dislike regarding `util`, `common` etc. It might be nice to have the same logic for extracting the source of a request in multiple places. But I'm fine with inlining this for now inside the source predicate

aryszka: ok. let's call it net, both the directory and package, and see if it grows with time.

aryszka:
lgtm

PR #114, opened 2016-03-02
extend innkeeper marshaling with custom predicates

aryszka:


innkeeper/client_test.go
```diff
@@ -296,6 +296,15 @@ func TestParsingInnkeeperComplexRoute(t *testing.T) {
...
 					}]
 				},
+                "predicates": [{
+                    "name": "customPredicate1"
+                }, {
+                    "name": "customPredicate2",
+                    "args": []
+                }, {
+                    "name": "customPredicate3",
+                    "args": ["string1", 3.14]
```
danpersa: do we really need float numbers? for filters we only integers (at least in Innkeeper)

aryszka: why only integers?

aryszka: does innkeeper support 64 bit integers only?

aryszka: udpated to check for 64 bit integers. please, let us know if that's not ok.

aryszka: after consulting dan: limited the innkeeper numbers to 32 bit, signed integers

innkeeper/client_test.go
```diff
@@ -479,6 +501,33 @@ func TestConvertRouteHeaders(t *testing.T) {
...
 
+func TestConvertRoutePredicates(t *testing.T) {
+	d := &routeData{Route: routeDef{Predicates: []customPredicate{{
+		Name: "customPredicate1",
+	}, {
+		Name: "customPredicate2",
+		Args: []interface{}{},
+	}, {
+		Name: "customPredicate3",
+		Args: []interface{}{"string1", 3.14},
```
danpersa: shouldn't this be an integer?

innkeeper/client_test.go
danpersa: shouldn't line 306 have an integer there?

danpersa:
lgtm

PR #115, opened 2016-03-03
re-enable real numbers for innkeeper

aryszka:


danpersa:
except the failing check, lgtm

Issue #49, opened 2015-11-11
fix the time sensitivity of the tests


aryszka:
tests are randomly failing due to the non-sync initial propagation of the route settings. however, this is not an issue in normal operation, it causes annoyances in dev time and ci/cd.

LappleApple:
Is #189 close to being merged? Then this can be closed.

aryszka:
yes, merging it will automatically close this

PR #117, opened 2016-03-03
add test proxy

aryszka:
created an initial proxy testing
some go fmt sneaked in, too

### DIFF
--- a/proxy/proxytest/proxytest.go
+++ b/proxy/proxytest/proxytest.go
@@ -1,0 +1,20 @@
+package proxytest
+
+import (
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/proxy"
+	"github.com/zalando/skipper/routing"
+	"github.com/zalando/skipper/routing/testdataclient"
+	"net/http/httptest"
+	"time"
+)
+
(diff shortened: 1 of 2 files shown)


proxy/proxytest/proxytest.go
```diff
@@ -0,0 +1,20 @@
...
+	"net/http/httptest"
+	"time"
+)
+
+func New(fr filters.Registry, routes ...*eskip.Route) *httptest.Server {
+	dc := testdataclient.New(routes)
+	rt := routing.New(routing.Options{FilterRegistry: fr, DataClients: []routing.DataClient{dc}})
+	pr := proxy.New(rt, proxy.OptionsNone)
+	tsp := httptest.NewServer(pr)
+	time.Sleep(12 * time.Millisecond) // propagate data client routes
```
mo-gr: Isn't there a nicer way? Some channel to listen to for routes to be propagated maybe? This is always just a workaround. Either you don't wait long enough and introduce race conditions in tests or you waste time :-/

mo-gr:
But in general: :+1:

proxy/proxytest/proxytest.go
aryszka: true... we already discussed. so far, i've been hesitating to introduce a synchronization point that is only used in tests. #49

aryszka:
ok, i guess i just merge it for now.

PR #121, opened 2016-03-06
Upgrade static

aryszka:
This pull request aims to fix the static filter to comply with the new way of shunting. The piping is a delicate issue. The reason is to fulfil 3 requirements:

1. benefit from the net/http.ServeFile implementation
2. still allow downstream filters to modify the response metadata
3. avoid buffering whole files


lmineiro:
I realize that we may have messed this one up. The fact that this became this complicated is a good sign.

lmineiro:
lgtm

aryszka:
i think the original version of this filter was messed up, i made it in a hurry with little care just push out somehow the humans.txt.

PR #122, opened 2016-03-10
simple endpointErrors to string convertation

ewgRa:
Small cosmetic refactoring.

Just review code and mention that this method can be done in more easy way.


aryszka:
thanks!

PR #123, opened 2016-03-10
client try endpoints optimization

ewgRa:
Optimization and more understandable client.tryEndpoints

Seems rotation not tested well, so, I'm not sure that this is correct optimization, but test passed and have same rotation instructions as original code.

This version will work faster, since less array copy operations and so on.

Since I'm not 100% sure that it is not broke something, merge it at your own risk or after more test will be cover more cases.

### DIFF
--- a/etcd/etcd.go
+++ b/etcd/etcd.go
@@ -204,11 +204,11 @@ func (c *Client) tryEndpoints(mreq func(string) (*http.Request, error)) (*http.R
 		endpointErrs = append(endpointErrs, err)
 	}
 
-	if lastEndpoint > 0 {
-		if lastEndpoint > 1 {
-			c.endpoints = append(c.endpoints[lastEndpoint:], c.endpoints[:lastEndpoint]...)
-		}
+	if lastEndpoint > 1 {
+		c.endpoints = append(c.endpoints[lastEndpoint:], c.endpoints[:lastEndpoint]...)
+	}
 
+	if lastEndpoint == len(c.endpoints) {
 		err = &endpointErrors{endpointErrs}
 	}
 
--- a/etcd/etcd.go
+++ b/etcd/etcd.go
@@ -208,7 +208,7 @@ func (c *Client) tryEndpoints(mreq func(string) (*http.Request, error)) (*http.R
 				index++
 			}
 
-			if index != 0 && index != len(c.endcheckpoints) {
+			if index != 0 && index != len(c.endpoints) {
 				c.endpoints = append(c.endpoints[index:], c.endpoints[:index]...)
 			}
 
--- a/etcd/etcd.go
+++ b/etcd/etcd.go
@@ -174,16 +174,15 @@ func isTimeout(err error) bool {
 // are tried. It returns the response to the first successful request.
 func (c *Client) tryEndpoints(mreq func(string) (*http.Request, error)) (*http.Response, error) {
 	var (
-		endpoints    []string
 		req          *http.Request
 		rsp          *http.Response
 		err          error
 		endpointErrs []error
+		lastEndpoint int
 	)
 
-	endpoints = c.endpoints
-	for len(endpoints) > 0 {
-		req, err = mreq(endpoints[0] + "/v2/keys")
+	for lastEndpoint = 0; lastEndpoint < len(c.endpoints); lastEndpoint++ {
+		req, err = mreq(c.endpoints[lastEndpoint] + "/v2/keys")
 		if err != nil {
 			return nil, err
 		}
@@ -203,13 +202,13 @@ func (c *Client) tryEndpoints(mreq func(string) (*http.Request, error)) (*http.R
 		}
 
 		endpointErrs = append(endpointErrs, err)
-		endpoints = endpoints[1:]
 	}
 
-	rotate := len(c.endpoints) - len(endpoints)
-	c.endpoints = append(c.endpoints[rotate:], c.endpoints[:rotate]...)
+	if lastEndpoint > 0 {
+		if lastEndpoint > 1 {
+			c.endpoints = append(c.endpoints[lastEndpoint:], c.endpoints[:lastEndpoint]...)
+		}
 
-	if len(endpoints) == 0 {
 		err = &endpointErrors{endpointErrs}
 	}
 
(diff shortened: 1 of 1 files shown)


ewgRa:
also "for" probably can be replaced with "range" and code will be looks even better.

etcd/etcd.go
```diff
@@ -204,13 +203,13 @@ func (c *Client) tryEndpoints(mreq func(string) (*http.Request, error)) (*http.R
 		}
 
 		endpointErrs = append(endpointErrs, err)
-		endpoints = endpoints[1:]
 	}
 
-	rotate := len(c.endpoints) - len(endpoints)
-	c.endpoints = append(c.endpoints[rotate:], c.endpoints[:rotate]...)
+	if lastEndpoint > 0 {
```
aryszka: this would consider it an error if a single endpoint fails, while the original considered the outcome an error only if all the endpoints failed.

etcd/etcd.go
```diff
@@ -204,13 +203,13 @@ func (c *Client) tryEndpoints(mreq func(string) (*http.Request, error)) (*http.R
...
+		if lastEndpoint > 1 {
+			c.endpoints = append(c.endpoints[lastEndpoint:], c.endpoints[:lastEndpoint]...)
```
aryszka: not completely sure about the check `> 1` for the rotation

aryszka:
Hi @ewgRa , thanks for the pull request. see my comment. And yes, you are completely right about the missing tests, it is my debt to create them. For now, I believe we can leave the rotation as is.

etcd/etcd.go
ewgRa: agree, fixed

etcd/etcd.go
ewgRa: if lastEndpoint <= 1 this means that there is no checkpoints or only one checkpoint. Rotation haven't sense.

ewgRa:
fix commited

ewgRa:
Also another one fix, for rotation.

ewgRa:
arghhh, need to be checked again. Seems there is not need to rotate when all checkpoints failed.

aryszka:
how i see, at the rotation step, lastEndpoint either points to the successful endpoint or is equal to the length of the endpoints. so i think the `-1` in the slicing expression is not needed. the check for `> 1` probably should be `> 0 && < len(c.endpoints)`. what do you think?

ewgRa:
@aryszka can you review please my last commit?

For me it is a best implementation.
If all failed - return all errors, nothing to rotate. If some failed but one success - rotate to success and return response. Simple, fast.


If it is worse than previous implementation, feel free to say it and we return to previous implemetation and "how I see" part.

Also if you need squash commits - feel free to request it.

etcd/etcd.go
```diff
@@ -175,23 +175,25 @@ func isTimeout(err error) bool {
...
 		}
 
 		rsp, err = c.client.Do(req)
 		if err == nil {
-			break
+			if index != 0 {
+				c.endpoints = append(c.endpoints[index:], c.endpoints[:index]...)
+			}
+
+			return rsp, nil
```
aryszka: i like this. however, the same should be done to the timeout cases: return nil, err instead of break. then it would be cool.

aryszka:
reviewed. see my comment regarding the timeout result.

i agree with your approach, only the timeout thingy needs to be fixed yet. The reason for this weird handling of the timeout is that we had some problems with the long polling in production with local etcd proxies.

> if it is worse...
no, it is not. your approach makes it definitely more readable.

> ...squash...
no, not yet. we still need to create contribution guidelines, until that happens, we don't go nitpicking on commit squash :)

thanks a lot for the contribution!

ewgRa:
done, if I understand right, if timeout happens - than stop looking to next checkpoints and return last error related to timeout.

ewgRa:
also I think if there is timeoutError - we need totate index+1?

ewgRa:
I think yes, so, when timeout - rotate to next "live" checkpoint.

ewgRa:
Sorry, was typo :) Recently was contests with checkpoints :)

ewgRa:
From "master" it seems like when timeout happens - next time it will again ask this timed out endpoint, if I'm right, is this a bug?

aryszka:
hi, sorry for not getting back to this earlier. So.. no the timeout is considered an error when initially all routes are fetched, but it is not an error in case of waiting for an update, because in that case it means that simply just there were no updates.

aryszka:
normally we wouldn't need the timeout at all for the updates and just do normal long-polling, but apparently, there were some problems when etcd is accessed through a local etcd proxy instance. this may be quite ugly now, not defending it, but this was the best we could come up with.

aryszka:
if we start from the state after commit 8c365f3c5fef9199557aa630cfa605c3973ec50b, and replace `break` statements on the `isTimeout`, to do a `return nil, err` instead of `break`, then it would work as intended.

ewgRa:
@aryszka Are you sure about replace break to return? If just replace break to return - there will be no rotation if timeout happens. And it will be not like in original code.

For now I drop this skip endpoint if it raise timeout, now it looks like original code. I will tomorrow read your explanations about this timeout, because for example "when initially all routes are fetched" I don't see in original code.

As for me original code always consider timeout as "response" and make break and do not go to next endpoints. It is like answer and we decide that all another endpoints also give timeouts. Probably it is true for network timeout, but for url timeout (I think) next endpoint can be live.

Probably there is tests needed :)

aryszka:
i see, you are right, the endpoints that failed before the one timeouting, should be rotated.

So, the point is that the timeout case is considered by the caller whether it is an error or not, see LoadAndParseAll and LoadUpdate. I'll try to find my etcd startup scripts, with which the weird behavior, that lead to this timeout handling, can be reproduced.

You are completely right about the tests. I'll see if i can ship some tests until tomorrow.

ewgRa:
It is absolutely not urgent for me, so, feel free to write tests when you have time.

When I will have time, I will try to write them myself and check mentioned LoadAndParseAll and LoadUpdate.

ewgRa:
there is a tests that cover tryEndpoints very well #129

In case when they will be merged, I will merge them to this branch. For now I merge them locally and they passed, so, refactoring fit original logic.

ewgRa:
Rebased with upstream master - tests passed.

Anything else? Another test cases that new code will fail? Squash?

ewgRa:
+ dropped check that always true

aryszka:
thanks :+1:

ewgRa:
yw! 

sorry for the mass of messages and commits )

PR #128, opened 2016-03-13
Path params

aryszka:


aryszka:
closing for now, reconsidering the template syntax

danielbayerlein:
Is there currently an alternative for this feature?

aryszka:
@danielbayerlein yes and no, unfortunately rather no. When you implement a custom filter, you can access the path parameters, https://godoc.org/github.com/zalando/skipper/filters#FilterContext with the `PathParam(string) string` function. What is missing is a nice way to reference these parameters in the eskip route config format.

PR #129, opened 2016-03-13
#124: add enpoints rotation tests

ewgRa:
PR for #124 - tests for endpoints rotation

### DIFF
--- a/etcd/etcd_test.go
+++ b/etcd/etcd_test.go
@@ -21,6 +21,9 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
+	"strings"
+	"net"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


etcd/etcd_test.go
```diff
@@ -116,6 +119,98 @@ func TestReceivesError(t *testing.T) {
...
+	c, err := New(Options{Endpoints: []string{s.URL, s2.URL, "neverreached"}, Prefix: "/skippertest-invalid"})
+
+	if err != nil {
+		t.Error(err)
+		return
+	}
+
+	_, err = c.LoadAll()
+
+	if err == nil {
```
aryszka: i guess this error check is irrelevant here because it doesn't fail this time due to transer error, but because of the empty paylod. fine, though :+1:

PR #133, opened 2016-04-02
Fix matcher

### LINKED ISSUE #116: Surprising route behaviour with more specific sub-path routes
If I have the following three routes:

`a: Path("/foo/*_") -> "https://foo.org"`
`b: Path("/foo/bar") && Traffic(0.5, "bar", "foobar") -> "https://bar.org"`
`z: * -> "https://catch.all"`

I would expect a request with the path `/foo/bar` to hit `foo.org` in half the time and `bar.org` the other half. However, all the requests, that don't hit `bar.org` end up being route to `catch.all`.

This is probably due to the way, the path-match tree is constructed, but it is very surprising and dangerous behaviour.

ewgRa:
fix #116 with use changes in https://github.com/zalando/pathmux/pull/2

### DIFF
--- a/routing/matcher.go
+++ b/routing/matcher.go
@@ -28,11 +28,11 @@ type leafRequestMatcher struct {
 	path string
 }
 
-func (m *leafRequestMatcher) Match(value interface{}) bool {
+func (m *leafRequestMatcher) Match(value interface{}) (bool, interface{}) {
 	v := value.(*pathMatcher)
 	l := matchLeaves(v.leaves, m.r, m.path)
 
-	return l != nil
+	return l != nil, l
 }
 
 type leafMatcher struct {
@@ -247,10 +247,10 @@ func newMatcher(rs []*Route, o MatchingOptions) (*matcher, []*definitionError) {
 }
 
 // matches a path in the path trie structure.
-func matchPathTree(tree *pathmux.Tree, path string, lrm *leafRequestMatcher) (leafMatchers, map[string]string) {
-	v, params := tree.LookupMatcher(path, lrm)
+func matchPathTree(tree *pathmux.Tree, path string, lrm *leafRequestMatcher) (leafMatchers, map[string]string, *leafMatcher) {
+	v, params, value := tree.LookupMatcher(path, lrm)
 	if v == nil {
-		return nil, nil
+		return nil, nil, nil
 	}
 
 	// prepend slash in case of free form wildcards path segments (`/*name`),
@@ -261,7 +261,7 @@ func matchPathTree(tree *pathmux.Tree, path string, lrm *leafRequestMatcher) (le
 		params[pm.freeWildcardParam] = freeParam
 	}
 
-	return pm.leaves, params
+	return pm.leaves, params, value.(*leafMatcher)
 }
 
 // matches the path regexp conditions in a leaf matcher.
@@ -373,8 +373,8 @@ func (m *matcher) match(r *http.Request) (*Route, map[string]string) {
 	lrm := &leafRequestMatcher{r, path}
 
 	// first match fixed and wildcard paths
-	leaves, params := matchPathTree(m.paths, path, lrm)
-	l := matchLeaves(leaves, r, path)
+	_, params, l := matchPathTree(m.paths, path, lrm)
+
 	if l != nil {
 		return l.route, params
 	}
--- a/routing/matcher.go
+++ b/routing/matcher.go
@@ -28,7 +28,7 @@ type leafRequestMatcher struct {
 	path string
 }
 
-func (m *leafRequestMatcher) IsMatch(value interface{}) bool {
+func (m *leafRequestMatcher) Match(value interface{}) bool {
 	v := value.(*pathMatcher)
 	l := matchLeaves(v.leaves, m.r, m.path)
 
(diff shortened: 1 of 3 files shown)


routing/matcher.go
```diff
@@ -235,8 +247,14 @@ func newMatcher(rs []*Route, o MatchingOptions) (*matcher, []*definitionError) {
 }
 
 // matches a path in the path trie structure.
-func matchPathTree(tree *pathmux.Tree, path string) (leafMatchers, map[string]string) {
-	v, params := tree.Lookup(path)
+func matchPathTree(tree *pathmux.Tree, path string, lrm *leafRequestMatcher) (leafMatchers, map[string]string) {
+	var matcher pathmux.Matcher = nil
```
ewgRa: this strange thing needed, because in Tree somehow m != nil check true for nil. Perhaps bug in Go?

routing/matcher.go
```diff
@@ -358,8 +376,10 @@ func (m *matcher) match(r *http.Request) (*Route, map[string]string) {
 		path = path[:len(path)-1]
 	}
 
+	lrm := &leafRequestMatcher{r, path}
+
 	// first match fixed and wildcard paths
-	leaves, params := matchPathTree(m.paths, path)
+	leaves, params := matchPathTree(m.paths, path, lrm)
 	l := matchLeaves(leaves, r, path)
```
ewgRa: perhaps this double work, since in lrm already this work done. Just need somehow get leaf from lrm and this can save some performance.

routing/matcher.go
ewgRa: More here https://github.com/zalando/pathmux/pull/2/files#r58353118

routing/matcher.go
aryszka: maybe we can change the pathmux matcher a bit, so that it would return not only a bool but also the matched value.

ewgRa: Yes, it is possible, changes in pathmux will be needed (Matcher.Match, node.search, maybe other palces).

Also I see another one option. Just save latest matched leave in leafRequestMatcher (kind of cache) and take it in matcher.match.
Can we rely on such logic? I just fear that it will be lovely place for future bugs.
What do you think?

If you have same feelings, than I will commit changes later to pathmux and skipper.

aryszka: i agree that saving the latest is a bit shaky solution, especially for future maintenance, and i would prefer the way where match returns the matched value.

ewgRa: Ok, than I will commit changes to pathmux soon.

ewgRa:
1. Changes done. Can you benchmark skipper/routing?
For BenchmarkGeneric I have +11% with this final changes.

2. My laptop can done only BenchmarkGeneric, but on BenchmarkPathTree1 get kill signal and "out of memory" in syslog. Is this expected and it is only GC problem or there memory leaks?

3. Also see pathmux issue comments.

ewgRa:
So, I just reduce random routes counts in tests and run some benchmarks.

```
./benchcmp master-bench.txt fix-matcher-matchleaves-twice-call-bench.txt
benchmark                          old ns/op     new ns/op     delta
BenchmarkGeneric-4                 160205        193990        +21.09%
BenchmarkPathTree1-4               1085          1590          +46.54%
BenchmarkPathTree2-4               1367          1866          +36.50%
BenchmarkPathTree3-4               2198          3102          +41.13%
BenchmarkPathTree4-4               2754          3906          +41.83%
BenchmarkConstructionGeneric-4     208088        288422        +38.61%
BenchmarkConstructionMass-4        133039264     181532283     +36.45%
```

```
./benchcmp master-bench.txt fix-matcher-bench.txt 
benchmark                          old ns/op     new ns/op     delta
BenchmarkGeneric-4                 160205        192500        +20.16%
BenchmarkPathTree1-4               1085          1544          +42.30%
BenchmarkPathTree2-4               1367          1811          +32.48%
BenchmarkPathTree3-4               2198          3155          +43.54%
BenchmarkPathTree4-4               2754          3438          +24.84%
BenchmarkConstructionGeneric-4     208088        292211        +40.43%
BenchmarkConstructionMass-4        133039264     184731451     +38.85%
```

I interpret it as there is no big difference if call matchLeaves twice, or try to save this call by introduce return match value in Matcher.

But this "save call" way have twice degradation in pathmux performance if there is good benchmarks there.

Also as total skipper performance degradation about ~30%, that seems huge for me.

So, my opinion, revert last commit that save this matchLeaves call by return value from Matcher in skipper and pathmux, and just add missed call Matcher for catchAll case in pathmux.

routing/matcher.go
```diff
@@ -235,10 +247,10 @@ func newMatcher(rs []*Route, o MatchingOptions) (*matcher, []*definitionError) {
 }
 
 // matches a path in the path trie structure.
-func matchPathTree(tree *pathmux.Tree, path string) (leafMatchers, map[string]string) {
-	v, params := tree.Lookup(path)
+func matchPathTree(tree *pathmux.Tree, path string, lrm *leafRequestMatcher) (leafMatchers, map[string]string, *leafMatcher) {
```
aryszka: i think it would be enough if this returned only the route and the params

aryszka:
lgtm

routing/matcher.go
ewgRa: done here https://github.com/zalando/skipper/pull/142

PR #135, opened 2016-04-06
add served set for new shunting

aryszka:


lmineiro:
lgtm

mo-gr:
The `ServeHTTP` in proxy.go only looks for `c.served` in line 405 and 411. Is that correct? There are also some lines, that could be replaced with `c.Served()` now in that file, like 304

aryszka:
in kine 405: this is cool. The reason is that in the deprecated way of shunting, the filters wrote directly onto the response writer. In the new way of shunting, they can provide a body reader in the response object that they define. So, in the old way you cannot map the response onto the response writer, while in the new way, you have to.

for using Served(): well it is actually only line 304, and i considered this case as an exception from under the rule of 'use the getter when available'. it feels to me more explicit this way. Note that this duality will be gone with version 2, where we remove support for the deprecated shunting.

mo-gr:
Ok. No issues then. But please fix the broken tests

PR #139, opened 2016-04-21
provide an option to set custom data clients

aryszka:
a demo can be found here:
[https://github.com/aryszka/skipper-dataclient-demo](https://github.com/aryszka/skipper-dataclient-demo)


mo-gr:
lgtm

PR #142, opened 2016-04-21
less return arguments in matchPathTree

ewgRa:
Come from https://github.com/zalando/skipper/pull/133#discussion_r60588967


aryszka:
...and thanks for the contribution! :+1:

PR #146, opened 2016-04-22
Predicates that allow temporary routes. Between predicate makes the r…

c00ler:
…oute available only certain time interval. Before predicate makes the route available only before specified date. After predicate makes the route available only after specified date.

### DIFF
--- a/predicates/interval/interval.go
+++ b/predicates/interval/interval.go
@@ -1,17 +1,3 @@
-// Copyright 2015 Zalando SE
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-// http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
 /*
 Package interval implements custom predicates to match routes
 only during some period of time. Package includes three predicates:
--- a/predicates/interval/interval.go
+++ b/predicates/interval/interval.go
@@ -1,6 +1,8 @@
 /*
 Package interval implements custom predicates to match routes
-only during some period of time. Package includes three predicates:
+only during some period of time.
+
+Package includes three predicates:
 Between, Before and After. All predicates can be created using the date
 represented as a string in RFC3339 format (see https://golang.org/pkg/time/#pkg-constants),
 int64 or float64 number. float64 number will be converted into int64
@@ -11,13 +13,11 @@ range of dates. Range is a closed range, so boundaries are included in
 the range. Between predicate requires two dates to be constructed.
 Upper boundary must be after lower boundary.
 
-Before predicate matches only if current date is before the specified
-date. Only one date is required to construct the predicate. Boundary
-is not included in the range.
+Before predicate matches only if current date is before or equal to
+the specified date. Only one date is required to construct the predicate.
 
-After predicate matches only if current date is after the specified
-date. Only one date is required to construct the predicate. Boundary
-is not included in the range.
+After predicate matches only if current date is after or equal to
+the specified date. Only one date is required to construct the predicate.
 
 Examples:
 
@@ -34,10 +34,10 @@ Examples:
 package interval
 
 import (
-	"errors"
 	"net/http"
 	"time"
 
+	"github.com/zalando/skipper/predicates"
 	"github.com/zalando/skipper/routing"
 )
 
@@ -49,27 +49,25 @@ const (
 	after
 )
 
-var InvalidArgsError = errors.New("invalid arguments")
-
 type spec struct {
 	typ intervalType
 }
 
 type predicate struct {
-	typ   intervalType
-	begin time.Time
-	end   time.Time
-	time  func() time.Time
+	typ     intervalType
+	begin   time.Time
+	end     time.Time
+	getTime func() time.Time
 }
 
 // Creates Between predicate.
-func NewBetweenPredicate() routing.PredicateSpec { return &spec{between} }
+func NewBetween() routing.PredicateSpec { return &spec{between} }
 
 // Creates Before predicate.
-func NewBeforePredicate() routing.PredicateSpec { return &spec{before} }
+func NewBefore() routing.PredicateSpec { return &spec{before} }
 
 // Creates After predicate.
-func NewAfterPredicate() routing.PredicateSpec { return &spec{after} }
+func NewAfter() routing.PredicateSpec { return &spec{after} }
 
 func (s *spec) Name() string {
 	switch s.typ {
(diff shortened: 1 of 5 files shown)


predicates/interval/interval.go
```diff
@@ -0,0 +1,182 @@
...
+only during some period of time. Package includes three predicates:
+Between, Before and After. All predicates can be created using the date
+represented as a string in RFC3339 format (see https://golang.org/pkg/time/#pkg-constants),
+int64 or float64 number. float64 number will be converted into int64
+number.
+
+Between predicate matches only if current date is inside the specified
+range of dates. Range is a closed range, so boundaries are included in
+the range. Between predicate requires two dates to be constructed.
+Upper boundary mast be after lower boundary.
```
lmineiro: typo, "must"

predicates/interval/interval.go
```diff
@@ -0,0 +1,182 @@
...
+
+type intervalType int
+
+const (
+	between intervalType = iota
+	before
+	after
+)
+
+var InvalidArgsError = errors.New("invalid arguments")
```
lmineiro: Maybe this should be refactored to a predicate general error, just like we did for the filters

predicates/interval/interval.go
```diff
@@ -0,0 +1,182 @@
+// Copyright 2015 Zalando SE
```
lmineiro: I don't think we were repeating the header on every file...

predicates/interval/interval.go
c00ler: Makes sense, but I would create a separate task for it because it will have an impact on zalando specific filters. What do you think?

lmineiro: I was just suggesting to move it to predicate package, not filter or higher.
Besides, it's more idiomatic to name these vars `ErrSomething`

predicates/interval/interval.go
c00ler: Looks like that since last time I was looking into our guidelines something has changed. Now it is not mandatory. I'll remove it.

predicates/interval/interval.go
aryszka: ErrInvalidPredicateArgs would be nice. I agree with having a common one. We already talked about having a single predicate package, or, we could make it completely consistent with the filters, we could move the common definitions in the /predicate package, and put the simple predicates like this into a builtin package, and only the complex ones would have a separate package.

predicates/interval/interval.go
```diff
@@ -0,0 +1,168 @@
...
+
+type predicate struct {
+	typ   intervalType
+	begin time.Time
+	end   time.Time
+	time  func() time.Time
+}
+
+// Creates Between predicate.
+func NewBetweenPredicate() routing.PredicateSpec { return &spec{between} }
```
aryszka: It would be enough to call these functions `NewBetween`, `NewBefore` and `NewAfter`

predicates/interval/interval.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	return time.Time{}, false
+}
+
+func (p *predicate) Match(r *http.Request) bool {
+	now := p.time()
+
+	switch p.typ {
+	case between: // Between is inclusive and Before and After are exclusive
```
aryszka: why not the standard [...) way of intervals? wouldn't people expect rather that way?

predicates/interval/interval.go
```diff
@@ -0,0 +1,168 @@
...
+		if (p.begin.Before(now) || p.begin.Equal(now)) && (p.end.After(now) || p.end.Equal(now)) {
+			return true
+		}
+	case before:
+		if p.end.After(now) {
```
aryszka: `return p.end.After(now)`? ;)

predicates/interval/interval.go
```diff
@@ -0,0 +1,168 @@
...
+		if len(args) != 2 {
+			return nil, InvalidArgsError
+		}
+	default:
+		if len(args) != 1 {
+			return nil, InvalidArgsError
+		}
+	}
+
+	time := func() time.Time {
```
aryszka: `time := time.Now`

predicates/interval/interval.go
```diff
@@ -0,0 +1,168 @@
...
+
+type spec struct {
+	typ intervalType
+}
+
+type predicate struct {
+	typ   intervalType
+	begin time.Time
+	end   time.Time
+	time  func() time.Time
```
aryszka: this is a nice and simple way of making it testable, but looking at the tests, it could be a package level variable, too, to make it even simpler.

predicates/interval/interval.go
```diff
@@ -0,0 +1,168 @@
+/*
+Package interval implements custom predicates to match routes
+only during some period of time. Package includes three predicates:
```
aryszka: I would cut this paragraph after the first sentence. See how godoc renders package headlines, when listing subdirectories. https://godoc.org/github.com/zalando/skipper

predicates/interval/interval.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	return time.Time{}, time.Time{}, false
+}
+
+func parseArg(arg interface{}) (time.Time, bool) {
+	switch a := arg.(type) {
+	case string:
+		t, err := time.Parse(time.RFC3339, a)
+		if err == nil {
```
aryszka: we kind of lose this information here. I see that for many cases we just return the same InvalidArgsError, though.. meh

skipper.go
```diff
@@ -307,7 +308,7 @@ func Run(o Options) error {
 	}
 
 	// include bundeled custom predicates
-	o.CustomPredicates = append(o.CustomPredicates, source.New())
+	o.CustomPredicates = append(o.CustomPredicates, source.New(), interval.NewBetweenPredicate(), interval.NewBeforePredicate(), interval.NewAfterPredicate())
```
aryszka: we didn't define strict rules on the line lengths, but this feels a bit too long. Can we cut it somehow?

aryszka:
Nice!
See my minor comments. If you prefer to keep it as it is now, then ready to merge.

aryszka:
Actually, I am still wondering about the inclusive/exclusive decision, because that's something hard to fix in a backwards compatible way, once it's out.

lmineiro:
I'm also thinking about the same as @aryszka 

Considering a practical scenario where I'd like to schedule a route for the near future, I'd use the inclusive lower bound and exclusive upper bound.

`someRoute: Between('yyyy-mm-dd 00:00:00', 'yyyy-mm-dd+1 00:00:00') -> ...` would give no surprise as a full 24h period instead of some shenanigan like `someRoute: Between('yyyy-mm-dd 00:00:00', 'yyyy-mm-dd 23:59:59.999') -> ...`

The same actually applies for the other types:

`otherRoute: Before('yyyy-mm-dd 00:00:00') -> ...` would be easy to reason as "Do that until that day (dd)", which would not apply even for a single second.

`lastRoute: After('yyyy-mm-dd 00:00:00') -> ...` would, on the other hand, be expected to be something like, starting on that day (dd), do that...

Obviously there's no big practical difference. I'm inclined to agree with @aryszka that going with [a,b) would cause least surprise, with the above reasoning added

predicates/interval/interval.go
c00ler: 👍

predicates/interval/interval.go
c00ler: Yes, I will change the implementation, so all intervals are inclusive.

lmineiro: Erm, I don't think this was the suggestion. Lower interval inclusive and upper interval exclusive if I got it right. This was what my examples suggested

c00ler: You got it right, I made it my way :) I think it is a bit confusing for non tech people. If upper boundary is excluded then Before and After will be have a bit different. In Before the date will be excluded, because it is the right boundary. In After the date will be included, because it is the left boundary. I made it simpler, all explicitly defined dates are included.

c00ler:
@lmineiro I changed the implementation to always exclude right boundary from the range.

PR #147, opened 2016-05-02
Added write flag to log file

alexanderbh:
By default OpenFile will use the O_RDONLY but to write to the access log file it must be opened with write access.


lmineiro:
lgtm
thanks

Issue #148, opened 2016-05-13
Skipper should support SSL/TLS


Raffo:
- It should be possible to turn on and off. 
- It should be possible to specify Key, CA, Certificate.

lmineiro:
Sounds good. What do you think of HTTP/2? We could combine this :)

Raffo:
This [should come out of the box with HTTPS ](https://golang.org/pkg/net/http/)

````
The http package has transparent support for the HTTP/2 protocol when using HTTPS.
````

lmineiro:
That was actually what I meant.

It's probably more test code than anything else

hmatinho:
could you support getting certificates from letsencrypt automatically and managing them ?

Raffo:
@hmatinho I believe this is out of scope for this request. You can maybe open another issue and discuss is there what you would love to see implemented.

aryszka:
@hmatinho  what I would prefer, would be that skipper could monitor the provided path to the certificate and key, and automatically fetch the updates if any from there. And then letsencrypt's certbot standalone, or any other update method, could run separately from skipper, and update the certs in the configured directory. This way we wouldn't create a built-in dependency on a particular service (even if letsencrypt bears all our love :)). Would this approach fit your requirements/workflow?

Raffo:
Closed by #150.

PR #150, opened 2016-05-18
Implemented TLS support

Raffo:
Implements #148 . Please provide feedback.


cmd/skipper/main.go
```diff
@@ -30,11 +30,12 @@ package main
...
 import (
 	"flag"
+	"strings"
+	"time"
+
 	log "github.com/Sirupsen/logrus"
 	"github.com/zalando/skipper"
 	"github.com/zalando/skipper/proxy"
-	"strings"
-	"time"
```
linki: why that change in import order? isn't the canonical way to first import stdlib packages and then third party? even [goimports](https://github.com/golang/tools/blob/master/cmd/goimports/goimports.go#L7-L21) does it this way.

cmd/skipper/main.go
```diff
@@ -70,6 +71,8 @@ const (
 	accessLogUsage                 = "output file for the access log, When not set, /dev/stderr is used"
 	accessLogDisabledUsage         = "when this flag is set, no access log is printed"
 	debugEndpointUsage             = "when this address is set, skipper starts an additional listener returning the original and transformed requests"
+	certPathTLSUsage               = "path of the certificate file"
+	keyPathTLSUsage                = "path of the key"
```
linki: "path of" should be "path to"

cmd/skipper/main.go
Raffo: It was not done in the canonical way before, it's correct now.

skipper.go
```diff
@@ -27,11 +33,6 @@ import (
 	"github.com/zalando/skipper/predicates/source"
 	"github.com/zalando/skipper/proxy"
 	"github.com/zalando/skipper/routing"
-	"io"
-	"net/http"
-	"os"
-	"path"
-	"time"
```
linki: see import related question above

skipper.go
```diff
@@ -159,6 +160,11 @@ type Options struct {
 	AccessLogDisabled bool
 
 	DebugListener string
+
+	//Path of certificate when using TLS
+	CertPathTLS string
+	//Path of
```
linki: unfinished comment

Raffo: Thanks, will fix that.

skipper_test.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func TestHTTPServer(t *testing.T) {
+	o := Options{
+		Address: ":9090",
+	}
+	proxy := initializeProxy()
+	go listenAndServe(proxy, &o)
+	r, err := http.Get("http://localhost:9090")
+	if err == nil {
```
linki: this test will be green if `err != nil` which we probably don't want

skipper_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	proxy := initializeProxy()
+	go listenAndServe(proxy, &o)
+
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+	}
+	client := &http.Client{Transport: tr}
+	r, err := client.Get("https://localhost:9091")
+	if err == nil {
```
linki: this test will be green if `err != nil` which we probably don't want

cmd/skipper/main.go
Raffo: I'm not sure if it should be "path of" or "path to", my English is failing me here, but I see that in other places in skipper that "path to" is used so I'll change it. Moreover, I would like to understand if this description is enough (according to you) given that is fairly shorter than other usage descriptions in skipper.

skipper.go
Raffo: See answer above.

skipper_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+func initializeProxy() *http.Handler {
+	filterRegistry := builtin.MakeRegistry()
+	proxy := proxy.New(routing.New(routing.Options{
+		FilterRegistry: filterRegistry,
+		DataClients:    []routing.DataClient{}}), proxy.OptionsNone)
+	return &proxy
+}
+
+func TestMissingOptionHTTP(t *testing.T) {
```
linki: how about: TestOptionsDefaultToHTTP

skipper_test.go
```diff
@@ -0,0 +1,98 @@
...
+	o := Options{}
+	if o.isHTTPS() {
+		t.FailNow()
+	}
+}
+
+func TestWithOptionHTTPS(t *testing.T) {
```
linki: how about: TestOptionsWithCertificateUsesHTTPS

skipper_test.go
Raffo: Right, I can probably just ignore the error then.

skipper_test.go
```diff
@@ -0,0 +1,98 @@
...
+	o := Options{CertPathTLS: "foo", KeyPathTLS: "bar"}
+	if !o.isHTTPS() {
+		t.FailNow()
+	}
+}
+
+func TestWrongCert(t *testing.T) {
```
linki: how about: `TestOptionsWithWrongCertPathFails`

skipper_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	proxy := initializeProxy()
+
+	err := listenAndServe(proxy, &o)
+	if err == nil {
+		t.Fatal(err)
+	}
+}
+
+func TestWrongKey(t *testing.T) {
```
linki: how about: `TestOptionsWithWrongKeyPathFails`

cmd/skipper/main.go
linki: oh, right :flushed: my bad. must be color blind or sth.

cmd/skipper/main.go
linki: i wouldn't mind if it were more detailed. like:

* when using https, the path on the local filesystem to the certificate file (including any intermediates)
* when using https, the path on the local filesystem to the certificate's private key file

skipper.go
linki: nevermind

skipper_test.go
linki: depends: since the test would be green that basically means "if the server doesn't come up at all, tls works". imo, the test should either fail in this case (so one can investigate what is actually causing it) or you have some other test that ensures the server comes up fine so that you can ignore this error here.

cmd/skipper/main.go
Raffo: I like them, but without "when using HTTPS". Actually you use HTTPS when you specify them.

skipper_test.go
Raffo: TestOptionsDefaultsToHTTP is better.

skipper_test.go
Raffo: I would go with ```TestWithWrongCertPathFails```

cmd/skipper/main.go
aryszka: good point anyway, on my list to apply linting

aryszka:
nice, thanks!

PR #152, opened 2016-05-19
fix race condition in listener tests

aryszka:
fixing listener tests:
- wait for listener started on the happy path
- find random available port for test listeners

### DIFF
--- a/skipper_test.go
+++ b/skipper_test.go
@@ -97,7 +97,7 @@ func TestWithWrongCertPathFails(t *testing.T) {
 func TestWithWrongKeyPathFails(t *testing.T) {
 	a, err := findAddress()
 	if err != nil {
-
+		t.Fatal(err)
 	}
 
 	o := Options{Address: a,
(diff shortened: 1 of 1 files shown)


aryszka:
this caused problems for me when ran the tests, can because of different platforms and timings etc

skipper_test.go
```diff
@@ -34,74 +77,91 @@ func TestOptionsWithCertUsesHTTPS(t *testing.T) {
...
+	err = listenAndServe(proxy, &o)
 	if err == nil {
 		t.Fatal(err)
 	}
 }
 
 func TestWithWrongKeyPathFails(t *testing.T) {
-	o := Options{Address: ":9091",
+	a, err := findAddress()
+	if err != nil {
```
Raffo: This err check is empty, it must have the same behavior as above.

Raffo:
You are totally right, it was pretty naive before and much better now. If you address my only comment it's OK to merge for me.

aryszka:
thanks, fixed

PR #158, opened 2016-05-29
Extract delayed

aryszka:
Introducing a new package, filters/serve.

- It implements a function ServeHTTP, that can be used to shunt requests with the stdlib standard http middleware interface, http.Handler:

```
var handler = http.StripPrefix(webRoot, http.FileServer(http.Dir(root)))

func (f *myFilter) Request(ctx filters.FilterContext) {
	serve.ServeHTTP(ctx, handler)
}
```

- On a one step lower level, to support more flexible scenarios, it implements a generic PipedBody, that can be used to serve custom or modified content, without full buffering or blocking the filter chain:

```
func (f *myFilter) Response(ctx filters.FilterContext) {
	in := ctx.Response().Body
	out := serve.NewPipedBody()
	ctx.Response().Body = out

	ctx.Response().Header.Del("Content-Lenght")
	ctx.Response().Header.Set("Content-Encoding", "gzip")
	ctx.Response().Header.Add("Vary", "Accept-Encoding")

	go func() {
		defer in.Close()

		gz := gzip.NewWriter(out)
		defer gz.Close()

		_, err := io.Copy(gz, in) // timeout handled through the original body
		if err == nil {
			err = io.EOF
		}

		out.CloseWithError(err)
	}()
}
```

### DIFF
--- a/filters/builtin/redirect.go
+++ b/filters/builtin/redirect.go
@@ -37,9 +37,9 @@ func NewRedirect() filters.Spec { return &redirect{deprecated: true} }
 
 // Returns a new filter Spec, whose instances create an HTTP redirect
 // response. It shunts the request flow, meaning that the filter chain on
-// the request path is not continued, and the backend is not the request
-// is not forwarded to the backend. Instances expect two parameters: the
-// redirect status code and the redirect location.
+// the request path is not continued. The request is not forwarded to the
+// backend. Instances expect two parameters: the redirect status code and
+// the redirect location.
 // Name: "redirectTo".
 func NewRedirectTo() filters.Spec { return &redirect{deprecated: false} }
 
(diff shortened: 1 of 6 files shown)


filters/builtin/redirect.go
```diff
@@ -31,11 +31,19 @@ type redirect struct {
 // response. Marks the request as served. Instances expect two
 // parameters: the redirect status code and the redirect location.
 // Name: "redirect".
+//
+// This filter is deprecated, use RedirectTo instead.
 func NewRedirect() filters.Spec { return &redirect{deprecated: true} }
 
+// Returns a new filter Spec, whose instances create an HTTP redirect
+// response. It shunts the request flow, meaning that the filter chain on
+// the request path is not continued, and the backend is not the request
```
mo-gr: sentence is broken. better: 'It shunts the request flow, meaning that the filter chain on the request path is not continued. The request is not forwarded to the backend.'

mo-gr:
lgtm.

aryszka:
thanks

PR #161, opened 2016-06-01
Compression

aryszka:


filters/builtin/compress.go
```diff
@@ -0,0 +1,323 @@
...
+	"application/octet-stream",
+}
+
+var (
+	gzipPool    = &sync.Pool{}
+	deflatePool = &sync.Pool{}
+)
+
+func init() {
+	for i := 0; i < runtime.NumCPU()*4; i++ {
```
mo-gr: Why four times the number of cpus in the pool?

aryszka: tested with different values by binary steps, this turned out to be the best. Not sure what are the actual chances for cache misses.

filters/builtin/compress.go
```diff
@@ -0,0 +1,323 @@
...
+
+	if !stringsContain(mime, ct) {
+		return false
+	}
+
+	return true
+}
+
+func acceptedEncoding(r *http.Request) string {
+	var encs encodings
```
mo-gr: I'm kind of surprised that there is nothing like that in the go std lib... But the code is clear and understandable... so no offense

filters/builtin/compress.go
```diff
@@ -0,0 +1,323 @@
...
+	if !stringsContain(r.Header["Vary"], "Accept-Encoding", http.CanonicalHeaderKey) {
+		r.Header.Add("Vary", "Accept-Encoding")
+	}
+}
+
+// Not handled encoding is considered as an implementation error, since
+// these functions are only called from inside the package, and the
+// encoding should be selected from a predefined set.
+func unsupported() {
+	panic(unsupportedEncoding)
```
mo-gr: I understand the reasoning, but I'm not sure I like the idea of a filter panicking...

mo-gr: Ok... thinking about it, it is probably just fine to panic here

filters/builtin/compress.go
```diff
@@ -0,0 +1,323 @@
...
+		gw, err := gzip.NewWriterLevel(nil, gzip.BestSpeed)
+		if err != nil {
+			// This is considered as an implementation error, since the compress/gzip doc
+			// states that it returns an error only if the compression level is invalid.
+			panic(err)
+		}
+
+		return gw
+	case "deflate":
+		fw, err := flate.NewWriter(nil, flate.BestSpeed)
```
mo-gr: Have you considered making the gzip/deflate implementation an option to the filter spec?

filters/builtin/compress.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+func encode(out *io.PipeWriter, in io.ReadCloser, enc string) {
+	pool := encoderPool(enc)
+	pe := pool.Get()
+	var e encoder
+	if pe == nil {
+		e = newEncoder(enc)
+	} else {
+		e = pe.(encoder)
```
mo-gr: no one needs generics ;-)

filters/builtin/compress.go
aryszka: for a moment yes, but then i decided to simplify it, because the used encoding typicall so much more depends on what the client puts into the accept-encoding header. We can think about it. How would it work?

filters/builtin/compress.go
aryszka: oh yeah, we should collect the examples :)

mo-gr:
lgtm

filters/builtin/compress.go
mo-gr: No, I mean just provide an option for gzip.BestCompression (or whatever) when you configure skippers filters

aryszka: ok max compression would look in eskip like:

```
compress(9, "...", "x-custom-mime/zalando")
```

mo-gr: even better. perfect!

PR #164, opened 2016-06-06
Connection pooling

aryszka:
The main change in this pull request is in proxy/proxy.go. It enables setting higher connection pool size. It also implements periodic closing of idle connections, in order to support scenarios where the backend services are replaced (e.g. redeployed).

The changes try to be backwards compatible.

There is a lot of noise in the PR, that was needed to make testing less resource consuming, because seemed like, at least on some machines, they were incosistently failing on refused connections.


cmd/skipper/main.go
```diff
@@ -61,6 +61,8 @@ const (
 	sourcePollTimeoutUsage         = "polling timeout of the routing data sources, in milliseconds"
 	insecureUsage                  = "flag indicating to ignore the verification of the TLS certificates of the backend services"
 	proxyPreserveHostUsage         = "flag indicating to preserve the incoming request 'Host' header in the outgoing requests"
+	idleConnsPerHostUsage          = "maximum idle connections per backend host"
+	closeIdleConnsPeriodUsage      = "period of closing all idle connections in seconds. Not closing when less than 0"
```
lmineiro: Seeing below that this is an actual Duration I was wondering if it wouldn't cause less surprise and doc hunting if it's actually named "Duration".
We could also accept a string as the parameter and user ParseDuration to give the end user great flexibility (We could still default to seconds for plain integers)

proxy/proxy.go
```diff
@@ -31,41 +31,93 @@ const (
...
-	// previous benchmarks.
-	//
-	// Also: it should be a parameter.
-	idleConnsPerHost = 64
+	// The default value set for http.Transport.MaxIdleConnsPerHost.
+	DefaultIdleConnsPerHost = 64
+
+	// The default period at which the idle connections are forcibly
+	// closed.
+	DefaultCloseIdleConnsPeriod = 12 * time.Second
```
lmineiro: Is this default also copied from somewhere else or are we introducing it? If we're introducing it, I'd increase this to 20secs/30secs to align with the default DNS TTL (20sec) or the JVMs 30sec

proxy/proxy.go
```diff
@@ -169,23 +224,56 @@ func mapRequest(r *http.Request, rt *routing.Route, host string) (*http.Request,
...
+func New(r *routing.Routing, options Options, pr ...PriorityRoute) *Proxy {
+	return NewProxy(ProxyOptions{
+		Routing:              r,
+		Flags:                ProxyFlags(options),
+		PriorityRoutes:       pr,
+		CloseIdleConnsPeriod: -time.Second})
+}
+
+// Creates a proxy.
+func NewProxy(o ProxyOptions) *Proxy {
```
lmineiro: I see the dilemma... to avoid breaking the API, we now get `proxy.NewProxy(...)` ... Would it be too shocking to export only this one under skipper pkg? I understand this means having the factory method further away from the implementation but it actually only uses exported stuff. `skipper.NewProxy(...)` would seem more idiomatic...

proxy/proxy.go
```diff
@@ -31,41 +31,93 @@ const (
...
+	DefaultIdleConnsPerHost = 64
+
+	// The default period at which the idle connections are forcibly
+	// closed.
+	DefaultCloseIdleConnsPeriod = 12 * time.Second
 )
 
-type Options uint
+// ProxyFlags control the behavior of the proxy.
+type ProxyFlags uint
```
lmineiro: Erm... not very idiomatic, is it? this is already in pkg proxy... It should be named `Flags` only

proxy/proxy.go
```diff
@@ -31,41 +31,93 @@ const (
...
+	OptionsDebug             = Options(ProxyDebug)
+)
 
-	// Flag indicating whether the outgoing request to the backend
-	// should use by default the 'Host' header of the incoming request,
-	// or the host part of the backend address, in case filters don't
-	// change it.
-	OptionsProxyPreserveHost
+// Proxy initialization options.
+type ProxyOptions struct {
```
lmineiro: Same as above, although I understand the will to keep the old one and hence, naming this one differently.

skipper.go
```diff
@@ -96,9 +96,21 @@ type Options struct {
 	// Polling timeout of the routing data sources.
 	SourcePollTimeout time.Duration
 
-	// Flags controlling the proxy behavior.
+	// Deprecated. See ProxyFlags. When used together with ProxyFlags,
+	// the values will be combined with |.
 	ProxyOptions proxy.Options
 
+	// Flags controlling the proxy behavior.
+	ProxyFlags proxy.ProxyFlags
```
lmineiro: Yeah... definitely... `proxy.ProxyFlags` looks funky

cmd/skipper/main.go
aryszka: i like the parse duration. I prefer the Period instead of Duration, because it is a periodic action.

proxy/proxy.go
aryszka: we are introducing it. OK to do 20s, thx.

proxy/proxy.go
aryszka: I don't really like to put it into the root package. This is a temporary situation. Once releasing a large version, we will drop the deprecated call. Would that be cool?

proxy/proxy.go
aryszka: Right, thx.

proxy/proxy.go
aryszka: I would handle this one together with the NewProxy function.

skipper.go
aryszka: ok

Issue #159, opened 2016-05-30
Support to upgrade protocol http code 101


aryszka:

szuecs:
As far as I understand we can implement a feature which can do upgrade connections of http in a generic way, such that you can use WS, Spdy or what other protocol the client and backend agrees to handshake.
Because of that I changed the title to make the feature request more clear.

aryszka:
agree

PR #167, opened 2016-06-17
Feature handle upgrade requests

szuecs:
This PR adds feature from issue #159


proxy/upgrade.go
```diff
@@ -0,0 +1,195 @@
...
+	"os"
+	"strings"
+	"sync"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// isUpgradeRequest returns true if and only if there is a "Connection"
+// key with the value "Upgrade" in Headers of the given request.
+func isUpgradeRequest(req *http.Request) bool {
```
aryszka: this one concerns me a bit, even if we consider the current implementation an experimental feature. The issue is that it always upgrades on any request that has Connection: upgrade. This may break other use cases. One of the reasons is that the proxy would not execute the response filters on upgrade requests. If it is enough now to support websockets, then it is cool, because then in this function, we could simply check the Upgrade header, in addition to the Connection, if it is set to "websocket", because Skipper hasn't supported websockets sofar, so nothing would get broken. But, e.g. if there is a browser out there that tries to upgrade to e.g. spdy (or h2c, who knows), then the requests coming from these browsers would be upgraded, and the response filters would be omitted for those users. If restricting the upgrade to websockets is not enough, then one possible way to come around this issue for now would be to create a cmd line flag where the experimental feature is explicitly enabled. Then later, when the feature is finalized in a more organic way, it can be enabled by default. Let's discuss it.

cmd/skipper/main.go
```diff
@@ -39,13 +39,14 @@ import (
...
-	defaultRuntimeMetrics       = true
-	defaultApplicationLogPrefix = "[APP]"
+	defaultAddress               = ":9090"
+	defaultEtcdPrefix            = "/skipper"
+	defaultSourcePollTimeout     = int64(3000)
+	defaultMetricsListener       = ":9911"
+	defaultMetricsPrefix         = "skipper."
+	defaultRuntimeMetrics        = true
+	defaultApplicationLogPrefix  = "[APP]"
+	defaultBackendFlushIntervall = 20 * time.Millisecond
```
aryszka: spelling: interval with single 'l'. see other places, too.

proxy/proxy.go
```diff
@@ -213,6 +221,12 @@ func mapRequest(r *http.Request, rt *routing.Route, host string) (*http.Request,
 	u.Scheme = rt.Scheme
 	u.Host = rt.Host
 
+	backendURL, err := url.Parse(rt.Backend)
```
aryszka: a topic of ongoing discussion...
but since I see that now the filters are executed on the request flow, we could use the basicAuth filter instead of putting the user and the password in the backend address. Note that the backend address is limited by design to scheme and host (https://godoc.org/github.com/zalando/skipper/eskip#hdr-Backend)

proxy/upgrade.go
```diff
@@ -0,0 +1,195 @@
...
+
+// TODO: add user here
+type auditLog struct {
+	Method   string `json:"method"`
+	Path     string `json:"path"`
+	Query    string `json:"query"`
+	Fragment string `json:"fragment"`
+}
+
+// serveHTTP inspects the request and either proxies an upgraded connection directly,
```
aryszka: does it really inspect the request, or that happens before? maybe changed from a previous version?

proxy/upgrade_test.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	gotProt := getUpgradeRequest(req)
+	if gotProt != prot {
+		t.Errorf("%s != %s for %+v", gotProt, prot, req)
+	}
+
+}
+
+func TestServeHTTP(t *testing.T) {
+	// TODO
```
aryszka: hmmm... :)

aryszka:
looks almost good. Let's discuss the isUpgradeRequest checking and the basic auth in the mapRequest.

cmd/skipper/main.go
szuecs: done

proxy/upgrade.go
szuecs: I would like to use it with SPDY/3.1. 
I am ok to have an --experimental-upgrade flag

1)
If you want to apply the filters on each request and response, you have to dissect all upgradeable protocols. That means writing protocol dissectors for each protocol. I think it should not be done within skipper.

2)
The problem with hijacking is that we do not own the sockets anymore. 

3)
Maybe there is also no http in between anymore so filters does not make sense to apply.

proxy/proxy.go
szuecs: It will not work for upgraded connections that have to get a net.Conn from the backend request.

proxy/upgrade.go
szuecs: fixed

proxy/upgrade.go
aryszka: 1) for web sockets, it makes sense to apply the filters for the initial round of establishing the connection. For http2, I don't know yet how to do it, but it would be nice, if we could implement it some time soon. For spdy, as you prefer, here just wanted to make sure that other functionality gets not broken.

2) for web sockets, it is ok, we only care about the initiation phase. For, http2, the same: i don't know yet. SPDY: yes, true, makes not much sense to invest too much efforts to achieve filtering.

3) this depends on the protocol that we upgrade to. based on this: https://github.bus.zalan.do/pathfinder/land-of-mosaic/pull/4, we'll need to care only about 'WebSocket' in the midterm future. At least I hope.

so after all, with the --experimental-upgrade flag, ship it!

proxy/proxy.go
aryszka: why? applyFiltersToRequest is called, and it sets the Authorization header, then here in the maprequest, a new request object is created, and the cloneHeader copies the Authorization header among others. The end result is the same as if it was taken from the backend url.

proxy/upgrade.go
szuecs: 3) 
This is an internal PR, that seems to have no pointer to Upgrade protocol, http2, SPDY nor websocket.

aryszka: haha, didn't check what i am pasting in the rush :)
what i meant was this: http://www.iana.org/assignments/http-upgrade-tokens/http-upgrade-tokens.xhtml

whatever. with the flag, it'll be cool for now

szuecs: I added the flag -experimental-upgrade to switch it on.

proxy/proxy.go
szuecs: I checked it again and you are right. We have not to copy the User from URL. I can drop the backendURL from mapRequest.

aryszka:
:+1: 

thanks!

PR #171, opened 2016-06-22
Innkeeper

aryszka:
applies latest innkeeper interface.

WARNING: not backwards compatible!
WARNING: the upsert/reset/delete functionality for innkeeper in the eskip tool was not tested!

### DIFF
--- a/innkeeper/eskiptojson.go
+++ b/innkeeper/eskiptojson.go
@@ -5,6 +5,13 @@ import (
 	"log"
 )
 
+type matchType string
+
+const (
+	matchStrict = matchType("STRICT")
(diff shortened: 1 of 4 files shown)


innkeeper/eskiptojson.go
```diff
@@ -5,6 +5,59 @@ import (
 	"log"
 )
 
+type matchType string
```
danpersa: we don't need the code in this file anymore, isn't it?

aryszka: true.. i just didn't want to change the write direction this time. i'll do that in a next round. this one is only needed for the other, unused, code to compile.

danpersa:
👍

PR #174, opened 2016-07-06
Diagnostics tools - Network throttling

aryszka:
This pull request implements network throttling: latency, max bandwidth or custom chunking with delays.

### DIFF
--- a/filters/diag/diag.go
+++ b/filters/diag/diag.go
@@ -1,0 +1,286 @@
+package diag
+
+import (
+	"io"
+	"math/rand"
+	"net/http"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/filters/serve"
+)
+
+const minChunkSize = 512
+
+const (
+	RandomName           = "randomContent"
+	LatencyName          = "latency"
+	ChunksName           = "chunks"
+	BandwidthName        = "bandwidth"
+	BackendLatencyName   = "backendLatency"
+	BackendBandwidthName = "backendBandwidth"
+	BackendChunksName    = "backendChunks"
+)
+
+type throttleType int
+
+const (
+	latency throttleType = iota
+	bandwidth
+	chunks
+	backendLatency
+	backendBandwidth
+	backendChunks
+)
+
+type random struct {
+	len int
+}
+
+type throttle struct {
+	typ       throttleType
+	chunkSize int
+	delay     time.Duration
+}
+
+var randomChars = []byte("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789")
+
+func kbps2bpms(kbps float64) float64 {
+	return kbps * 1024 / 1000
+}
+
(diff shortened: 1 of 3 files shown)


mo-gr:
👍

szuecs:
:+1:

filters/diag/diag.go
```diff
@@ -0,0 +1,344 @@
...
+func NewRandom() filters.Spec { return &random{} }
+
+// NewLatency creates a filter specification whose filter instances can be used
+// to add additional latency to responses. It expects the latency in milliseconds
+// as an argument. It always adds this value in addition to the natural latency,
+// and does not do any adjustments. Eskip example:
+//
+// 	* -> latency(120) -> "https://www.example.org";
+//
+func NewLatency() filters.Spec { return &throttle{typ: latency} }
```
lmineiro: I would've preferred that all the latency arguments to be strings and accept valid ParseDuration values. We could still accept simple numbers and default that to ms. It's a matter of taste but I'd like to see arguments such as "1s", "500µs", etc.
One thing to filter out, just like right now, would be negative latencies.

filters/diag/diag.go
```diff
@@ -0,0 +1,344 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &throttle{t.typ, chunkSize, delay}, nil
+}
+
+func (t *throttle) goThrottle(in io.ReadCloser, close bool) io.ReadCloser {
+	if t.chunkSize <= 0 {
+		if t.delay > 0 {
```
lmineiro: this is not needed: "A negative or zero duration causes Sleep to return immediately."

filters/diag/diag.go
```diff
@@ -0,0 +1,344 @@
...
+		defer func() {
+			w.CloseWithError(err)
+			if close {
+				in.Close()
+			}
+		}()
+
+		b := make([]byte, defaultChunkSize)
+		for err == nil {
+			n := 0
```
lmineiro: Should we be specific about `uint64` or at least some other kind of `uint`?

filters/diag/diag.go
```diff
@@ -0,0 +1,344 @@
...
+
+			if err == nil {
+				delay := t.delay
+
+				switch t.typ {
+				case bandwidth, backendBandwidth:
+					delay -= time.Now().Sub(start)
+				}
+
+				if delay >= 0 {
```
lmineiro: It's not needed

filters/diag/diag_test.go
```diff
@@ -0,0 +1,610 @@
...
+func checkMessage(expect messageExp, start time.Time, body io.Reader) error {
+	if expect.header > 0 && !checkWithTolerance(start, expect.header) {
+		return fmt.Errorf("expected time to header, %v, %v, %v", start, time.Now(), expect.header)
+	}
+
+	totalRead := 0
+
+	cstart := start
+	for _, c := range expect.chunks {
+		n := 0
```
lmineiro: Same remark about the data type as before

lmineiro:
This is cool stuff.
I added some minor remarks, none of them blockers for me.

Issue #97, opened 2016-01-28
eskip pretty printing option


aryszka:
consider if this should be done by eskip, or just provide an awk example

pc-alves:
Working on this issue I came across a question: Should indentation be

1. a single space (`' '`)
2. a single tab (`\t`)
3. one of the above as default, and an option for the number of spaces (`-s4` to have 4 spaces)

Any thoughts? Comments?

aryszka:
in our practice, two spaces became a de-facto standard. It could be the default for the tool, too. For overriding, I would prefer a single option, where one can specify just anything as the atomic indentation:

```
-indent "    "
```

the value of this parameter could be validated with "\s*", because we accept all \s when parsing eskip.

lmineiro:
Wondering if we could consider the "existing" OFS, RS, ... environment variables that already control the output of some well known tools, line AWK

LappleApple:
Hey @aryszka @pc-alves @lmineiro This issue started in Jan 2016. Anything we can do here--add a Help Wanted label, add something to the README, etc.? Would like to close.

pc-alves:
PR #180 implemented pretty-print with a default option of two spaces for indentation.
@aryszka suggested the `-indent "    "` command line option, which was not implemented. I left the issue open for this reason. But I'm fine with closing it, and having the `indent` option as a separate issue.

LappleApple:
Thanks, @pc-alves.

PR #180, opened 2016-07-21
Added pretty print functionality

pc-alves:
fixes issue #97

### DIFF
--- a/cmd/eskip/doc.go
+++ b/cmd/eskip/doc.go
@@ -77,10 +77,11 @@ const (
 	prependFileUsage    = "prepend filters from a file to each patched route"
 	appendFiltersUsage  = "append filters to each patched route"
 	appendFileUsage     = "append filters from a file to each patched route"
+	prettyUsage         = "prints routes in a more readable format"
 
 	// command line help (1):
 	help1 = `Usage: eskip <command> [media flags] [--] [file]
-Commands: check|print|upsert|reset|delete
+Commands: check|print|upsert|reset|delete|patch
 Verify, print, update or delete Skipper routes.
 See more: https://github.com/zalando/skipper
 
--- a/eskip/string.go
+++ b/eskip/string.go
@@ -90,7 +90,7 @@ func (r *Route) filterString(pretty bool) string {
 		sfilters = appendFmt(sfilters, "%s(%s)", f.Name, argsString(f.Args))
 	}
   if pretty {
-		return strings.Join(sfilters, "\n\t-> ")
+		return strings.Join(sfilters, "\n  -> ")
 	}
 	return strings.Join(sfilters, " -> ")
 }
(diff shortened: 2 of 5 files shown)


cmd/eskip/doc.go
```diff
@@ -77,10 +77,11 @@ const (
 	prependFileUsage    = "prepend filters from a file to each patched route"
 	appendFiltersUsage  = "append filters to each patched route"
 	appendFileUsage     = "append filters from a file to each patched route"
+	prettyUsage         = "prints routes in a more readable format"
 
 	// command line help (1):
 	help1 = `Usage: eskip <command> [media flags] [--] [file]
-Commands: check|print|upsert|reset|delete
+Commands: check|print|upsert|reset|delete|patch
```
szuecs: This seems not to be relevant for this PR, but maybe correct to fix it.

eskip/string.go
```diff
@@ -84,12 +84,14 @@ func (r *Route) predicateString() string {
...
 }
 
-func (r *Route) filterString() string {
+func (r *Route) filterString(pretty bool) string {
 	var sfilters []string
 	for _, f := range r.Filters {
 		sfilters = appendFmt(sfilters, "%s(%s)", f.Name, argsString(f.Args))
 	}
-
+  if pretty {
```
szuecs: Please use gofmt to format your code.

cmd/eskip/doc.go
aryszka: yes, thanks for the fix

eskip/string.go
```diff
@@ -103,15 +105,23 @@ func (r *Route) backendString() string {
 
 // Serializes a route expression. Omits the route id if any.
 func (r *Route) String() string {
+	return r.Print(false)
+}
+
+func (r *Route) Print(pretty bool) string {
```
aryszka: i like it. Can we please do the same for `func String(routes ...*Route) string` to stay a bit consistent?

eskip/string_test.go
```diff
@@ -94,6 +94,56 @@ func TestDocString(t *testing.T) {
 		`route2: Path("/some/path") -> "https://www.example.org"`)
 }
 
+func TestPrintNonPretty(t *testing.T) {
+	for i, item := range []struct {
+		route    string
+		expected string
+	} {
+		{
+			"route1: Method(\"GET\") -> filter(\"expression\") -> <shunt>",
```
aryszka: you can also use backtick strings for this, so you don't need to escape doubleqoutes

eskip/string.go
pc-alves: Not sure what you mean.
a) Having a `Print` function for multiple routes, with a pretty option?
b) Using the `Print` function in `func String(routes ...*Route) string`?

aryszka: sorry for not being completely clear. I meant a). See `func (r *Route) String()` string and `func String(routes ...*Route) string`. There could be a `func Print(routes ...*Route) string`, too.

pc-alves: Check the latest commit (7aa655c). I think I got it right.

aryszka:
:+1:

aryszka:
thanks, @pc-alves !

Issue #155, opened 2016-05-24
Add a CONTRIBUTING.md


LappleApple:
Would also identify 2-3 issues you really want help with that are good candidates for community to-do's, and add them in the README. I could do the add, if you share the issue #'s here.

LappleApple:
@aryszka @lmineiro @danpersa Went ahead and made one. This project has a fair number of open issues; to encourage contributions I would highlight a few in the README that you really need.

aryszka:
solved by #183

PR #183, opened 2016-07-29
doc: Update Contributing.md

melTob:
#155


melTob:
@LappleApple your feedback is highly appreciated

LappleApple:
@melTob Actually never mind, these are fine!

PR #192, opened 2016-08-07
Close current body

### LINKED ISSUE #191: close the actual body in the proxy
if a filter replaces the response body from the backend, the proxy still closes the one from the backend and doesn't close the one from the filter. This way, if the body set by the filter needs to be closing, the filter has no signal when to do it. Also it limits composability in cases when multiple filters would replace the response body. Change the proxy so, that it always closes the actual body after the response filters were executed, and the filters have the responsibility to close the received body in case they replace it.

aryszka:
Closes #191


grassator:
@aryszka any reason not to merge this?

PR #196, opened 2016-08-11
Query

csenol:
This commit adds QueryParam predicate to match routes with some query params
This is my first Go code. Please review @aryszka @lmineiro

### DIFF
--- a/predicates/query/query.go
+++ b/predicates/query/query.go
@@ -1,0 +1,110 @@
+/*
+Package source implements a custom predicate to match routes
+based on the Query Params in URL
+
+It supports checking existence of query params and also checking whether
+query params value match to a given regular exp
+
+Examples:
+
+    // Checking existence of a query param
+    // matches http://example.org?bb=a&query=withvalue
+    example1: QueryParam("query") -> "http://example.org";
+
+    //Even a query param without a value
+    // matches http://example.org?bb=a&query=
+    example1: QueryParam("query") -> "http://example.org";
+
+    //matches with regexp
+    // matches http://example.org?bb=a&query=example
+    example1: QueryParam("query", "^example$") -> "http://example.org";
+
+    //matches with regexp and multiple values of query param
+    // matches http://example.org?bb=a&query=testing&query=example
+    example1: QueryParam("query", "^example$") -> "http://example.org";
+
+*/
+
+package query
+
+import (
+	"github.com/zalando/skipper/predicates"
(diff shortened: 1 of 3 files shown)


lmineiro:
Looks like you missed gofmt...

csenol:
right, formatted and updated

predicates/query/query.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/zalando/skipper/routing"
+	"net/http"
+	"regexp"
+)
+
+type MatchType int
+
+const (
+	Exists  MatchType = 1
+	Matches MatchType = 2
```
aryszka: while correct, this can be written somewhat nicer:

```
type MatchType int

const (
    Exists  MatchType = iota
    Matches
)
```

iota is a go idiom for these kind of constants.

predicates/query/query.go
```diff
@@ -0,0 +1,110 @@
...
+const name = "QueryParam"
+
+func New() routing.PredicateSpec { return &spec{} }
+
+func (s *spec) Name() string {
+	return name
+}
+
+func (s *spec) Create(args []interface{}) (routing.Predicate, error) {
+	lengs := len(args)
```
aryszka: interesting var name `lengs`, if it's a signature habit, i like it. But typically len(slice) doesn't need to be put in a var, because it is a builtin syntax. nicer to just use len(args).

predicates/query/query.go
```diff
@@ -0,0 +1,110 @@
...
+package query
+
+import (
+	"github.com/zalando/skipper/predicates"
+	"github.com/zalando/skipper/routing"
+	"net/http"
+	"regexp"
+)
+
+type MatchType int
```
aryszka: I think MatchType, Exists, and Matches should be private to the package. This means the should start with lowercase. In go the visibility is package based, lower case means package only, while upper case means visible outside of the package. (like nothing and `public` in some other language)

aryszka:
very nice first-time go! thanks for the contribution. Please, change the visibility of the matchtype type and its enum values. I also recommend to use the iota there.

I like how the behavior in the documentation is covered by the tests.

predicates/query/query.go
Raffo: You can also use iota +1 in case you don't want to have it starting from 0... which can make sense cause 0 is the "default" value for ints.

csenol:
@aryszka @Raffo Thanks for comments. Updated

PR #197, opened 2016-08-19
Add Weights to predicates

csenol:
This is not for a PR but validating an approach. How about adding weights to predicates ? This might be backward incompatible  . Is there a way to add default value to interface methods ?


csenol:
@lmineiro @aryszka Looking for your input

csenol:
Got the input

PR #201, opened 2016-08-24
Add replicateTo filter. Which can be used for shadow backends

csenol:
Addin `replicateTo` filter for shadow backends.
I used DefaultHttp client but not sure with its performance as we discussed with @lmineiro 
Looking for your feedbacks @aryszka @lmineiro

### DIFF
--- a/filters/builtin/builtin.go
+++ b/filters/builtin/builtin.go
@@ -7,6 +7,7 @@ import (
 	"github.com/zalando/skipper/filters"
 	"github.com/zalando/skipper/filters/diag"
 	"github.com/zalando/skipper/filters/flowid"
+	"github.com/zalando/skipper/filters/tee"
 )
 
 const (
@@ -67,6 +68,7 @@ func MakeRegistry() filters.Registry {
 		diag.NewBackendLatency(),
 		diag.NewBackendBandwidth(),
 		diag.NewBackendChunks(),
+		tee.NewTee(),
 	} {
 		r.Register(s)
 	}
--- a/filters/tee/tee.go
+++ b/filters/tee/tee.go
@@ -1,0 +1,140 @@
+package tee
+
+import (
+	"fmt"
+	log "github.com/Sirupsen/logrus"
+	"github.com/zalando/skipper/filters"
+	"net/http"
+	"net/url"
+	"regexp"
+)
+
+const (
+	Name = "tee"
+)
+
+type teeSpec struct{}
+
+type teeType int
+
+const (
+	asBackend teeType = iota + 1
+	pathModified
+)
+
+type tee struct {
+	client      *http.Client
+	typ         teeType
+	host        string
+	scheme      string
+	rx          *regexp.Regexp
+	replacement string
+}
+
+// Returns a new tee filter Spec, whose instances execute
+// the exact same Request against a shadow backend.
+// parameters: shadow backend url, optional - the path(as a regexp) to match and the replacement string.
+// Name: "tee".
+// Example
+// Path("/api/v3") -> tee("https://api.example.com") -> "http://example.org/"
+// This route wil send incoming  request to http://example.org/api/v3 but will also send
+// a copy of the query to https://api.example.com/api/v3.
+// Example
+// Path("/api/v3") -> tee("https://api.example.com", ".*", "/v1/" ) -> "http://example.org/"
+// This route wil send incoming request to http://example.org/api/v3 but will also send
+// a copy of the request to https://api.example.com/v1/ . Note that scheme and path are changed
+
+func NewTee() *teeSpec {
+	return &teeSpec{}
+}
+
+//We do not touch response at all
+func (r *tee) Response(filters.FilterContext) {}
+
+// Request is copied and then modified to adopt changes in new backend
+func (r *tee) Request(fc filters.FilterContext) {
+	copyOfRequest := cloneRequest(r, fc.Request())
+	go func() {
+		rsp, err := r.client.Do(&copyOfRequest)
+		if err != nil {
+			log.Warn("error while tee request", err)
+		}
+		defer rsp.Body.Close()
+	}()
+}
+
+// copies requests changes URL and Host in request.
+// If 2nd and 3rd params are given path is also modified by applying regexp
+func cloneRequest(rep *tee, req *http.Request) http.Request {
+	copyOfRequest := new(http.Request)
+	*copyOfRequest = *req
+	copyUrl := new(url.URL)
+	*copyUrl = *req.URL
+	copyOfRequest.URL = copyUrl
+	copyOfRequest.URL.Host = rep.host
+	copyOfRequest.URL.Scheme = rep.scheme
+	copyOfRequest.Host = rep.host
+	//Need toc copy body
+	// copyOfRequest.Body = ???
+	//request URI should be empty
+	copyOfRequest.RequestURI = ""
+	if rep.typ == pathModified {
+		copyOfRequest.URL.Path = string(rep.rx.ReplaceAllString(copyOfRequest.URL.Path, rep.replacement))
+	}
+	return *copyOfRequest
+}
+
+// Creates out tee Filter
+// If only one parameter is given shadow backend is used as it is specified
+// If second and third parameters are also set, then path is modified
+func (spec *teeSpec) CreateFilter(config []interface{}) (filters.Filter, error) {
+	var teeInit tee = tee{client: &http.Client{}}
+
+	if len(config) == 0 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+	backend, ok := config[0].(string)
+	if !ok {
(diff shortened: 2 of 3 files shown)


filters/replicateto/replicateto.go
```diff
@@ -0,0 +1,146 @@
+// Copyright 2015 Zalando SE
```
aryszka: please don't use the license text in the code files. The reason for many files having it, is that initially we thought that it is required. Turned out no, and since then when changing a file, we remove them, and when new ones are added, we don't add the license excerpt.

filters/replicateto/replicateto.go
```diff
@@ -0,0 +1,146 @@
...
+type replicateTo struct {
+	typ         replicateType
+	host        string
+	scheme      string
+	rx          *regexp.Regexp
+	replacement []byte
+}
+
+// Returns a new replication filter Spec, whose instances execute
+// exactly same Request against a shadow backend.
```
aryszka: exactly same -> the exact same

filters/replicateto/replicateto.go
```diff
@@ -0,0 +1,146 @@
...
+// parameters: shadow backend url, optional - the expression to match and the replacement string.
+// Name: "replicateTo".
+
+// Example
+// Path("http://example.com/api/v3") -> replicateTo("https://api.example.com") -> "http://example.org/"
```
aryszka: not a valid path, it doesn't need the scheme nor the host

filters/replicateto/replicateto.go
```diff
@@ -0,0 +1,146 @@
...
+//This route wil send incoming query to http://example.org/api/v3 but will also send
+// a copy of the query to https://api.example.com/api/v3. Note that scheme is changed
+
+// Example
+// Path("http://example.com/api/v3") -> replicateTo("https://api.example.com", ".*", "/v1/" ) -> "http://example.org/"
+//This route wil send incoming query to http://example.org/api/v3 but will also send
```
aryszka: send incoming query -> send incoming queries -> (why 'query'?) -> send incoming requests

filters/replicateto/replicateto.go
```diff
@@ -0,0 +1,146 @@
...
+	return &replicateToSpec{}
+}
+
+//We do not touch response at all
+func (r *replicateTo) Response(filters.FilterContext) {}
+
+// Request is copied and then modified to adopt changes in new backend
+func (r *replicateTo) Request(fc filters.FilterContext) {
+	copyOfRequest := modifyRequest(r, fc.Request())
+	go http.DefaultClient.Do(&copyOfRequest)
```
aryszka: I think initially the default settings for the http client are ok. I don't really see now a good way to pass individual TLS and connection pooling sttings for them, even if they might be important. As a last resort, one can always cascade any router instances, or processes.

While the default settings are OK, please use a new instance for the replicateToSpec. The reason is that http.DefaultClient is globally accessible, and any code can change its settings. I would add a field for the `replicateToSpec` of `client *http.Client` and would initialize it in `NewReplicateTo`, with default settings.

filters/replicateto/replicateto.go
```diff
@@ -0,0 +1,146 @@
...
+// Example
+// Path("http://example.com/api/v3") -> replicateTo("https://api.example.com") -> "http://example.org/"
+//This route wil send incoming query to http://example.org/api/v3 but will also send
+// a copy of the query to https://api.example.com/api/v3. Note that scheme is changed
+
+// Example
+// Path("http://example.com/api/v3") -> replicateTo("https://api.example.com", ".*", "/v1/" ) -> "http://example.org/"
+//This route wil send incoming query to http://example.org/api/v3 but will also send
+// a copy of the query to https://api.example.com/v1/ . Note that scheme and path are changed
+
```
aryszka: This empty line renders the doc invisible in the generated doc, pls remove the empty line here. You can test the docs with `godoc -http :8080`.

filters/replicateto/replicateto.go
aryszka: the http.Client.Do function returns a response and an error. Two things are missing:

- the response body needs to be closed
- it can be important for diagnostics to log unhandled errors

```
rsp, err := r.client.Do(&copyOfRequest)
if err != nil {
    log.Error("error while replicating request", err)
}

defer rsp.Body.Close()
```

since this happens now in a goroutine, we need to wrap it in a closure:

```
func() {
    rsp, err := r.client.Do(&copyOfRequest)
    if err != nil {
        log.Error("error while replicating request", err)
    }

    defer rsp.Body.Close()
}()
```

filters/replicateto/replicateto.go
```diff
@@ -0,0 +1,146 @@
...
+// Request is copied and then modified to adopt changes in new backend
+func (r *replicateTo) Request(fc filters.FilterContext) {
+	copyOfRequest := modifyRequest(r, fc.Request())
+	go http.DefaultClient.Do(&copyOfRequest)
+}
+
+// copies requests changes URL and Host in request.
+// If 2nd and 3rd params are given path is also modified by applying regexp
+func modifyRequest(rep *replicateTo, req *http.Request) http.Request {
```
aryszka: it would be more explicit to call it `cloneRequest`

aryszka: besides: this is not enough, because this way the two outgoing requests share the body. This may not be completely straightfoward to do without blocking and excessive buffering. Please have a look at using io.Pipe and io.TeeReader (see some good naming just there :) ). We can try to do it together some time.

aryszka: this issue needs to be tested, too, so that the outgoing body is read full on the two test backends.

aryszka:
generally, I really like it.

Somehow the replicateTo name is not of my favour, I would love however if it was called 'tee', like the unix tool. I know I am not a champ of naming, but 'tee' for this one may be good.

The question about the client: see my inline comments. It is really a matter of the settings, we can start with the defaults, but it is safer to create a new instance for the spec at least.

aryszka:
See my comment for the cloning of the incoming body. I think it cannot be merged until that's sorted.

filters/replicateto/replicateto.go
```diff
@@ -0,0 +1,140 @@
...
+	typ         replicateType
+	host        string
+	scheme      string
+	rx          *regexp.Regexp
+	replacement []byte
+}
+
+// Returns a new replication filter Spec, whose instances execute
+// the exact same Request against a shadow backend.
+// parameters: shadow backend url, optional - the expression to match and the replacement string.
```
lmineiro: please mention "path". The expression is used against the path, correct?

lmineiro:
I'm also in favor of naming this filter Tee

filters/replicateto/replicateto.go
```diff
@@ -0,0 +1,140 @@
...
+			log.Warn("error while replicating request", err)
+		}
+		defer rsp.Body.Close()
+	}()
+}
+
+// copies requests changes URL and Host in request.
+// If 2nd and 3rd params are given path is also modified by applying regexp
+func cloneRequest(rep *replicateTo, req *http.Request) http.Request {
+	tmp := *req
```
lmineiro: I'd follow the strategy of https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy.go#L150 to clone the request

There seems to be a lot of (de)referencing going on here

filters/replicateto/replicateto.go
```diff
@@ -0,0 +1,140 @@
...
+	pathModified
+)
+
+type replicateTo struct {
+	client      *http.Client
+	typ         replicateType
+	host        string
+	scheme      string
+	rx          *regexp.Regexp
+	replacement []byte
```
lmineiro: See note below about `ReplaceAllString`

filters/replicateto/replicateto.go
```diff
@@ -0,0 +1,140 @@
...
+	copyOfRequest.URL = copyUrl
+	copyOfRequest.URL.Host = rep.host
+	copyOfRequest.URL.Scheme = rep.scheme
+	copyOfRequest.Host = rep.host
+	//Need toc copy body
+	// copyOfRequest.Body = ???
+	//request URI should be empty
+	copyOfRequest.RequestURI = ""
+	if rep.typ == pathModified {
+		copyOfRequest.URL.Path = string(rep.rx.ReplaceAll([]byte(copyOfRequest.URL.Path), rep.replacement))
```
lmineiro: Overkill? Maybe `ReplaceAllString` serves the same purpose here without so much byte/string casting going around

filters/replicateto/replicateto.go
csenol: @lmineiro Great for pointing out to that method. Do you think if we could use same approach for copying body as well ?

aryszka: the body is not 'there' at the moment of copy. it is streaming, and the data needs to be duplicated. the trick is to figure how to do it without blocking any of the outgoing requests but not buffering too much either.

PR #204, opened 2016-09-01
#203 Share TLSClientConfig with upgrade protocol proxy

### LINKED ISSUE #203: Share tls.Config between proxy and upgradeProxy
It would make sense to share the `tls.Config` used by the proxy with the updateProxy instead of using two separate, but identical tls.Configs.

Our specific use case is that we are using a slightly modified version of skipper with the changes discussed in #179 and for this we need the `tls.Config` to be shared.

mikkeloscar:
Normal proxy and upgrade proxy should configure tls the same way, so it
makes sense to share the `*tls.Config` already initilized with the
normal proxy.

Note this exposes the `*http.Transport` instead of "hiding" it behind the `RoundTripper` interface.

Close #203

### DIFF
--- a/proxy/upgrade.go
+++ b/proxy/upgrade.go
@@ -147,9 +147,7 @@ func (p *upgradeProxy) dialBackend(req *http.Request) (net.Conn, error) {
 			}
 			err = tlsConn.VerifyHostname(hostToVerify)
 			if err != nil {
-				if tlsConn != nil {
-					_ = tlsConn.Close()
-				}
+				tlsConn.Close()
 				return nil, err
 			}
 		}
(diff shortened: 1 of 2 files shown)


proxy/upgrade.go
```diff
@@ -134,30 +135,25 @@ func (p *upgradeProxy) dialBackend(req *http.Request) (net.Conn, error) {
...
-				_ = tlsConn.Close()
+
+		if !p.insecure {
+			hostToVerify, _, err := net.SplitHostPort(dialAddr)
+			if err != nil {
+				return nil, err
+			}
+			err = tlsConn.VerifyHostname(hostToVerify)
+			if err != nil {
+				if tlsConn != nil {
```
aryszka: i think this could be just:

```
tlsConn.Close()
```

aryszka: or:

```
return nil, tlsConn.Close()
```

aryszka:
:+1:

proxy/upgrade.go
mikkeloscar: Thanks, fixed

PR #206, opened 2016-09-07
Tee filter

csenol:


filters/tee/tee.go
```diff
@@ -0,0 +1,172 @@
...
+	clone.URL.Host = t.host
+	clone.URL.Scheme = t.scheme
+	clone.Host = t.host
+
+	pr, pw := io.Pipe()
+	tr := &teeTie{req.Body, pw}
+	clone.Body = pr
+	req.Body = tr
+	//Setting to empty string otherwise go-http doesn't allow having it in client request
+	clone.RequestURI = ""
```
lmineiro: "" is the zero value for string. What am I missing?

csenol: While copying request from incoming request, we clone `RequestURI`, too. And If I use incoming request with http.client with that value set I got this error `level=warning msg="error while tee requestGet http://127.0.0.1:51406/: http: Request.RequestURI can't be set in client requests." 
`  Therefore I am setting it to empty string.

aryszka: true. I think the low level copy doesn't work out well here. Consider maybe: https://github.com/zalando/skipper/blob/master/proxy/proxy.go#L223

lmineiro: Sorry, I missed line  nr 99, it's clear for me now

lmineiro:
👍

filters/tee/tee.go
```diff
@@ -0,0 +1,172 @@
+package tee
+
+import (
+	"fmt"
+	log "github.com/Sirupsen/logrus"
+	"github.com/zalando/skipper/filters"
```
aryszka: in new code, we tend to use the "CodeReviewComments": https://github.com/golang/go/wiki/CodeReviewComments#imports
for imports. Here, the logrus and filters imports could be below the stdlib ones, separated by an empty line. This is, of course, not obvious when just generally looking at the whole code base.

filters/tee/tee.go
```diff
@@ -0,0 +1,172 @@
...
+// Name: "tee".
+// Example
+// Path("/api/v3") -> tee("https://api.example.com") -> "http://example.org/"
+// This route wil send incoming  request to http://example.org/api/v3 but will also send
+// a copy of the query to https://api.example.com/api/v3.
+// Example
+// Path("/api/v3") -> tee("https://api.example.com", ".*", "/v1/" ) -> "http://example.org/"
+// This route wil send incoming request to http://example.org/api/v3 but will also send
+// a copy of the request to https://api.example.com/v1/ . Note that scheme and path are changed
+
```
aryszka: this empty line makes the above documentation not visible in godoc. Please remove it and then it will be generated automatically like e.g. https://godoc.org/github.com/zalando/skipper/filters/diag#NewBackendBandwidth

filters/tee/tee.go
```diff
@@ -0,0 +1,172 @@
+package tee
```
aryszka: the package should have a documentation, at least a headline. See: https://godoc.org/github.com/zalando/skipper#pkg-subdirectories

filters/tee/tee.go
```diff
@@ -0,0 +1,172 @@
...
+	clone.Host = t.host
+
+	pr, pw := io.Pipe()
+	tr := &teeTie{req.Body, pw}
+	clone.Body = pr
+	req.Body = tr
+	//Setting to empty string otherwise go-http doesn't allow having it in client request
+	clone.RequestURI = ""
+	if t.typ == pathModified {
+		clone.URL.Path = string(t.rx.ReplaceAllString(clone.URL.Path, t.replacement))
```
aryszka: ReplaceAllString returns a string, unnecessary to convert: https://godoc.org/regexp#Regexp.ReplaceAllString

filters/tee/tee.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	return *clone
+}
+
+// Creates out tee Filter
+// If only one parameter is given shadow backend is used as it is specified
+// If second and third parameters are also set, then path is modified
+func (spec *teeSpec) CreateFilter(config []interface{}) (filters.Filter, error) {
+	var teeInit tee = tee{client: &http.Client{}}
```
aryszka: this is correct, but typically the below, shorthand form would be used instead in most go programs:

```
tee := tee{client: &http.Client{}}
```

aryszka: also, typically, when we finally return a pointer type of a structure from a function, we initialize it as a pointer, and just return that. But, of course, this is also correct.

filters/tee/tee.go
```diff
@@ -0,0 +1,172 @@
...
+	if len(config) == 1 {
+		teeInit.typ = asBackend
+		return &teeInit, nil
+	}
+
+	//modpath
+	if len(config) == 3 {
+		expr, ok := config[1].(string)
+		if !ok {
+			return nil, fmt.Errorf("Invalid filter config in %s, expecting regexp and string, got: %v", Name, config)
```
aryszka: one good practice in go is not capitalize error messages for better composition: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

filters/tee/tee_test.go
```diff
@@ -0,0 +1,129 @@
...
+		t.Errorf("Tee Request Path not modified, %v", modifiedRequest.URL.Path)
+	}
+
+	originalRequest := fc.Request()
+	if originalRequest.URL.Path != "/api/v3" {
+		t.Errorf("Incoming Request Scheme modified, %v", originalRequest.URL.Path)
+	}
+}
+
+type MyHandler struct {
```
aryszka: would be somewhat nicer to have the type definition at the top of the file

filters/tee/tee_test.go
```diff
@@ -0,0 +1,129 @@
...
+	shadowServer := httptest.NewServer(shadowHandler)
+	shadowUrl := shadowServer.URL
+	defer shadowServer.Close()
+
+	originalHandler := &MyHandler{name: "original"}
+	originalServer := httptest.NewServer(originalHandler)
+	originalUrl := originalServer.URL
+	defer originalServer.Close()
+
+	routeStr := "route1: * -> Tee(\"" + shadowUrl + "\")" + " -> " + "\"" + originalUrl + "\";"
```
aryszka: consider using the backtick "`" strings for the case of the escaped doublequotes '"'. also, this can be the right use case for fmt.Sprintf() https://godoc.org/fmt#Sprintf

filters/tee/tee_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	route, _ := eskip.Parse(routeStr)
+	registery := make(filters.Registry)
```
aryszka: typo, "registry", see the one just next to it

filters/tee/tee_test.go
```diff
@@ -0,0 +1,129 @@
...
+	p := proxytest.New(registery, route...)
+	defer p.Close()
+
+	req, _ := http.NewRequest("GET", p.URL, strings.NewReader("TESTEST"))
+	req.Host = "www.example.org"
+	req.Header.Set("X-Test", "true")
+	req.Close = true
+	rsp, _ := (&http.Client{}).Do(req)
+	rsp.Body.Close()
+	if shadowHandler.body != originalHandler.body {
```
aryszka: the problem with this test is that, in case of some lower level failure, both the shadowHandler and the originalHandler can have empty body, and the test may give false positive. Either make the test fail in case of error at line 119, or check if the two bodies are actually equal to "TESTTEST", or both (best ;) )

filters/tee/tee.go
```diff
@@ -0,0 +1,172 @@
...
+	if t.typ == pathModified {
+		clone.URL.Path = string(t.rx.ReplaceAllString(clone.URL.Path, t.replacement))
+	}
+	return *clone
+}
+
+// Creates out tee Filter
+// If only one parameter is given shadow backend is used as it is specified
+// If second and third parameters are also set, then path is modified
+func (spec *teeSpec) CreateFilter(config []interface{}) (filters.Filter, error) {
```
aryszka: i think we should have the non-happy path covered, too. The main reason is that the filters are typically created from their textual, eskip representation, during runtime long after the initial startup, and we want to be sure that they fail when they have to.

aryszka:
Nice! Most of my comments are rather just stylistic remarks, feel free to cherry-pick from them, but please fix these ones:

- the assertion in the test TestTeeEndToEndBody
- the package comment on line 1
- the fix for godoc on line 53, NewTee

filters/tee/tee.go
csenol: @aryszka Do you mean having tests for non-happy part in test or some implementation for some non-happy path which I can't see ? If it is latter case, could you give me some examples ? I'll add more tests for former case

filters/tee/tee.go
csenol: Added doc.go with some exmaples

filters/tee/tee.go
aryszka: e.g. len(config) == 0

csenol: Okay, by looking at your reference I understand you meant tests

PR #209, opened 2016-10-10
Pathsubtree

aryszka:
create PathSubtree() predicate
cover existing Path() predicate cases with additional tests


lmineiro:
:+1:

PR #213, opened 2016-10-19
rename tee to follow conventions

aryszka:
also fixes some docs (the examples from the constructor can be already found in doc.go)

### DIFF
--- a/eskip/doc.go
+++ b/eskip/doc.go
@@ -160,11 +160,20 @@ filters:
 
     status(418)
 
+	tee("https://audit-logging.example.org")
+
 For details about the built-in filters, please, refer to the
 documentation of the skipper/filters package. Skipper is designed to be
 extendable primarily by implementing custom filters, for details about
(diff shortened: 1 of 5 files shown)


lmineiro:
Does this break backwards compatibility?

aryszka:
yes, it does. It's been out for 20 days. I would make here an exception from the rule, but we cal also support the capitalized variant as deprecated for a while.

lmineiro:
Yeah, 20 days is not that long. I'd be fine with whatever approach. Your call.

csenol:
sorry for this. I thought Path/Host/PathRegexp have this convention but now I realize the pattern. Predicates are capitalized whereas filters are all small letters. Maybe we can write this convention to somewhere if it is already not around .

aryszka:
no worries. and yes good idea to mention the convention

aryszka:
updated with keeping the deprecated version and documenting the naming conventions in eskip docs

Issue #182, opened 2016-07-26
RFC: Extend *.eskip format with basic abstraction facilities


grassator:
The following proposal is to implement several abstraction facilities to *.eskip format. The focus is to have minimal changes to the syntax of the files and introduce bare minimum extra grammar.

## Example

```
@include "

commonFilterChain():
    filterOne()
    -> filterTwo()
    -> localeResolver();

AllHosts():
    Host(/my[.]host[.]name|my[.]other[.]host[.]name/)

$commonPathPrefix: "/prefix";
replacePath($path):
    modPath(".*", $path);

foo:
    AllHosts() && Path('/foo/(\w)+')
    -> commonFilterChain()
    -> pathWithPrefix("/bar")
    -> "http://mosaic9.org";
```

### User-defined constants

A constant MUST be able to represent the following base types:

* Number
* String
* Regular Expression

Each constant identifier MUST be prefixed with `$`. A constant definition MUST consist of an identifier and it's value separated by `:` and terminated by `;` located at the top level in an eskip file:

```
$someNum: 42;
$someStr: "foobar";
$someRegex: /foo|bar/; 
```

Constants are global and MUST have a globally unique name. A conflict in names MUST be reported by the `eskip check` command.

### User-defined filter chains

A user-defined filter chain has the same syntax as the route, but MUST have a parameter list enclosed in round braces following the identifier. Parameter list MUST contain zero or more parameters. Each parameter MUST be a locally-bound constant.

The body of the filter chain MUST consist of one or more filter "calls", separated by `->`:

```
commonFilterChain():
    filterOne()
    -> filterTwo()
    -> localeResolver();
```

Any violations of on of the following conditions MUST be reported by the `eskip check` command:

* User-defined filter chains are global and MUST have a globally unique name.
* User-defined filter chains MUST be defined before their usage.
* Self or mutual recursion is not allowed. 
* An identifier for a user-defined filter chain MUST start with a lowercase latin character.

### User-defined predicates

User-defined predicates have the same syntax as the route, but MUST have a parameter list enclosed in round braces following the identifier. Parameter list MUST contain zero or more parameters. Each parameter MUST be a locally-bound constant.

The body of the predicate  MUST consist of one or more predicate "calls", separated by `&&`:

```
PathOnDefaultHost($path):
    Host("mosaic9.org") && Path($path)
```

Any violations of on of the following conditions MUST be reported by the `eskip check` command:

* User-defined predicates are global and MUST have a globally unique name.
[...]

PR #215, opened 2016-10-24
Added support for transformation with path params

evangelion1204:
Hi,

I introduced a new path `transformPath` filter, but maybe this support could be added to `setPath` after all. It support to do the following and transform the path according to the specified parameters. An example is a legacy framework like ZF1.

```
route: Path("/path/:greeting/:name") => transformPath("/default/greeting/${greeting}/name/${name}") => "https://legacy.example.com";
```

In the end I would like to support for query params too.

```
route: Path("/path/:greeting/:name") => transformPath("/default?greeting=${greeting}&name=${name}") => "https://legacy.example.com";
```


filters/builtin/path.go
```diff
@@ -99,6 +132,12 @@ func (f *modPath) Request(ctx filters.FilterContext) {
 		req.URL.Path = f.rx.ReplaceAllString(req.URL.Path, f.replacement)
 	case fullReplace:
 		req.URL.Path = f.replacement
+	case transformReplace:
+		req.URL.Path = f.replacement
+		for _, placeholder := range f.placeholders {
+			req.URL.Path = strings.Replace(req.URL.Path, "${"+placeholder+"}", ctx.PathParam(placeholder), -1)
+		}
+		fmt.Println(req.URL.Path)
```
aryszka: tracing was left in

evangelion1204:
@aryszka @lmineiro 
I dropped the `transformPath` and added some sort of draft for `TemplateString` to `setPath` that could be used now in every filter to support `${name}` for headers, queries ...

@aryszka yeah, I just used the placeholder/template format I am used to in newest ES6 and from the bash.

filters/builtin/path.go
```diff
@@ -79,13 +89,36 @@ func createSetPath(config []interface{}) (filters.Filter, error) {
...
+	if len(config) != 1 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	replacement, ok := config[0].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	rx := regexp.MustCompile("\\$\\{(\\w+)\\}")
```
evangelion1204: See below refactoring and comments.

evangelion1204:
@aryszka 
Merged your PR and added the query filters, that my use-case is actually really supported 😄 you can also see my comment in #182

grassator:
@evangelion1204 How would one escape this replacement? I know it's quite unlikely, but I believe that every interpolation should have escaping possible.

filters/builtin/query.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+type modQuery struct {
+	behavior modQueryBehavior
+	name     *template.Template
+	value    *template.Template
+}
+
+// Returns a new dropQuery filter Spec, whose instances execute
+// drops a corresponding query param on the request path.
```
aryszka: "...whose instances drop a corresponding..."

filters/builtin/query.go
```diff
@@ -0,0 +1,132 @@
...
+// The templating feature will stay in Skipper, but the syntax of the
+// templating may change.
+//
+// See also: https://github.com/zalando/skipper/issues/182
+//
+// Name: "dropQuery".
+func NewDropQuery() filters.Spec { return &modQuery{behavior: drop} }
+
+// Returns a new setQuery filter Spec, whose instances replace
+// the request path.
```
aryszka: "the request path" -> "the request query" or "the query parameters"

aryszka:
@grassator you're right, a syntax needs escaping. However, I think this is 'okish' for now, because the feature is marked experimental, and the syntax may need to be changed together with #182 . It can be still useful until that.

lmineiro:
Wondering if we could come up with a different name for the template package. I spent a good part of the review thinking this was about the builtin `template.New()`. We should try to make it less ambiguous.

aryszka:
@lmineiro fair point. Unfortunately, I don't have a good name on my mind now.

grassator:
@aryszka @lmineiro How about "interpolate" for the name?

aryszka:
@grassator then rather "template". Very honestly, "interpolate" sounds a bit elitist to me for this purpose. @lmineiro , note that the stdlib has already two packages called "template", text/template and html/template. Soon, when dealing with #182, i can easily imagine an eskip/template, or similar. Whatever, the point here is that when one sees template during reviewing, already has to check if it is text/template or html/template ;)

If no veto, then I merge this tomorrow.

lmineiro:
I know about those two from the stdlib. I find horrible all that repetition of Template & Tree on both. My opinion is that two wrongs don't make one right.

This `template.New()` is public. We won't get rid of it.

I'm also struggling to understand what was wrong with the original notation in the first place. Every other library and framework handling routes that I checked recently - httprouter, gin, Play framework, also adopted `:varname`. Only gorilla followed a different path due to the included regex matching. Is this a NIH issue?

aryszka:
@lmineiro :varname is good for capturing, and it stays as it is now. This here is about inserting the already captured value. Does Gin do something like that?

yes, the package is public, but the whole stuff is marked experimental. We can mark the template package, too, true. TBH i am not sure at all if it deserves its own package. Since it's part of the syntax, it could reside in the eskip package. Then it could be eskip.Template.

aryszka:
actually! how would it be having it in the eskip package? eskip.Template?

lmineiro:
Would be the least bad. If the method was something like `eskip.NewTemplate()` there would be absolutely no ambiguity. Even if the `eskip.Template` type itself is exported it already sounds like a better version to me. Your thoughts?

aryszka:
perfectly agree

evangelion1204:
Ok, should I then
1) move the template type to eskip package
2) Change method from `New` to `NewTemplate`
3) fix the marked comments 😄 

@aryszka Don't wonder, I am currently ill, maybe I feel tomorrow evening good enough to apply the changes

filters/builtin/path.go
```diff
@@ -79,13 +89,36 @@ func createSetPath(config []interface{}) (filters.Filter, error) {
...
+	matches := rx.FindAllStringSubmatch(replacement, -1)
+	placeholders := make([]string, len(matches))
+
+	for index, placeholder := range matches {
+		placeholders[index] = placeholder[1]
```
evangelion1204: I am not sure if `FindStringSubmatch` is enough, it will only find the leftmost one, I want to find all `${...}` maybe there are more than one.

evangelion1204:
@aryszka ok, done. Also left a comment for @lmineiro above.

lmineiro:
Which one?

aryszka:
lgtm

evangelion1204:
@lmineiro 
https://github.com/zalando/skipper/pull/215#discussion_r84638742 about the `FindStringSubmatch`.

PR #216, opened 2016-10-25
Added support for basic auth

evangelion1204:
Hi,

we currently need basic auth in Zalando-Lounge. Therefor I added a PR to support a filter with basic auth.
```
* -> basicAuth("/path/to/htpasswd") -> "https://example.com";
```


filters/auth/basic.go
```diff
@@ -0,0 +1,62 @@
...
+const (
+	Name                      = "basicAuth"
+	ForceBasicAuthHeaderName  = "WWW-Authenticate"
+	ForceBasicAuthHeaderValue = "Basic realm=\"User Visible Realm\""
+)
+
+type basicSpec struct{}
+
+type basic struct {
+	configFile    string
```
aryszka: this field seems not to be used

filters/auth/basic_test.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	req, err := http.NewRequest("GET", "https://www.example.org/", nil)
+	if err != nil {
+		t.Error(err)
+	}
+
+	ctx := &filtertest.Context{FRequest: req}
+	f.Request(ctx)
+	if ctx.Response().Header.Get(ForceBasicAuthHeaderName) != ForceBasicAuthHeaderValue && ctx.Served() {
```
aryszka: maybe worth testing the status code, too

filters/auth/basic_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+func TestWithSuccessfulAuth(t *testing.T) {
+	spec := NewBasicAuth()
+	f, err := spec.CreateFilter([]interface{}{"htpasswd.test"})
+	if err != nil {
+		t.Error(err)
+	}
+
+	req, err := http.NewRequest("GET", "https://www.example.org/", nil)
+	req.SetBasicAuth("myName", "myPassword")
```
aryszka: i would test the case of wong username and password combination

filters/auth/basic.go
evangelion1204: Indeed, was the code of my first implementation.

filters/auth/basic_test.go
evangelion1204: Indeed, going to add it.

evangelion1204:
@aryszka @lmineiro feel free to take a look, should I rebase?

aryszka:
lgtm

filters/auth/basic.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	configFile, ok := config[0].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	htpasswd := auth.HtpasswdFileProvider(configFile)
+	authenticator := auth.NewBasicAuthenticator("Basic Realm", htpasswd)
```
lmineiro: Since the Realm ends up being something visible to users, would you consider accepting another (optional) parameter to customize the Realm name?

evangelion1204:
@lmineiro Added a second param to define the name which is at least visible in Firefox 😄

filters/auth/htpasswd.test
```diff
@@ -0,0 +1 @@
+myName:$apr1$8G2wkTu0$I4Mjw4DYGOfB71lwNRX531
```
lmineiro: can you move this file to a testdata subfolder?

filters/auth/basic.go
```diff
@@ -0,0 +1,72 @@
...
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	realmName := DefaultRealmName
+
+	if len(config) == 2 {
+		definedName, ok := config[1].(string)
+		if ok {
+			realmName = definedName
+		}
```
lmineiro: Would be more idiomatic to have:
```
if definedName, ok := config[1].(string); ok {
    realmName = definedName
}
```
Also, I'd prefer to have a one time build of the full header value like `Basic Realm="` + definedName + `"` allowing the code above to be a simple assignment:
```
header.Set(ForceBasicAuthHeaderName, a.realmName)
```

filters/auth/basic_test.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		t.Error(err)
+	}
+
+	req, err := http.NewRequest("GET", "https://www.example.org/", nil)
+	if err != nil {
+		t.Error(err)
+	}
+
+	expectedBasicAuthHeaderValue := ForceBasicAuthHeaderValue + "\"Basic Realm\""
```
lmineiro: Small note, you can use the back tick to include the double quote as part of the string, like
```
`"Basic Realm"`
```

lmineiro: Also, you could add a test case to assert the new optional parameter ;-)

filters/auth/htpasswd.test
evangelion1204: Yeah, then I can also rename the file.

filters/auth/basic.go
evangelion1204: > if definedName, ok := config[1].(string); ok {
>     realmName = definedName
> }

wasn't aware of that, honestly I am not go expert 😄  thx for sharing.

filters/auth/basic_test.go
evangelion1204: Good points about the ticks, regarding the assert I need to check how to do that ;)

evangelion1204: I think with more combinations it would make sense, in that case I would leave it as it is.

PR #219, opened 2016-11-08
Added a check for repeating route ids to `eskip check`

grassator:
This is a quite easy mistake to make, and hard to debug, so I've added a check for it.

### DIFF
--- a/cmd/eskip/load.go
+++ b/cmd/eskip/load.go
@@ -77,6 +77,17 @@ func loadRoutesChecked(in *medium) ([]*eskip.Route, error) {
 	return lr.routes, checkParseErrors(lr)
 }
 
+func checkRepeatedRouteIds(routes []*eskip.Route) error {
+	ids := map[string]struct{}{}
+	for _, route := range routes {
+		if _, ok := ids[route.Id]; ok {
+			return errors.New("Repeating route with id " + route.Id)
+		}
+		ids[route.Id] = struct{}{}
+	}
+	return nil
+}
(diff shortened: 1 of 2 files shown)


aryszka:
```
eskip print routes.eskip | cut -f 1 -d : | sort | uniq -c | grep -v -E '^'\\'s*1' && echo bad || echo ok
```

? :)

let me think a bit. Not necessarily bad, just departs a bit from the 'unixyness' of the eskip tool.

cmd/eskip/load.go
```diff
@@ -77,6 +77,17 @@ func loadRoutesChecked(in *medium) ([]*eskip.Route, error) {
...
+	ids := map[string]struct{}{}
+	for _, route := range routes {
+		if _, ok := ids[route.Id]; ok {
+			return errors.New("Repeating route with id " + route.Id)
+		}
+		ids[route.Id] = struct{}{}
+	}
+	return nil
+}
+
```
aryszka: if you use `map[string]bool`, then the if check could simply look like: `if ids[route.Id] {`

grassator:
@aryszka Switched to the map of bool as you proposed.

I think the composability of eskip didn't change with this commit. My first instinct was to just put something like what you typed in into our makefile, however I always want to the application to do the right thing by default.

This is just about UX—new users won't know about the behavior with duplicate route id, so they won't know to check for it and will be frustrated.

aryszka:
:+1:

PR #221, opened 2016-11-11
Decimal printing

aryszka:


lmineiro:
What is this for btw?

aryszka:
this is to avoid printing exponential form of larger numbers and .0000 decimals at the same time to be able to read back the output. for the real thing, i'll need to fix the parser. admittedly a hack. the current way is also :(

grassator:
👍

PR #223, opened 2016-11-22
Eskip json

grassator:
This PR also continues some preparation work on refactoring eskip to not treat some predicates as built-in (Path in this case).

### DIFF
--- a/cmd/eskip/load.go
+++ b/cmd/eskip/load.go
@@ -15,11 +15,11 @@
 package main
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"github.com/zalando/skipper/eskip"
 	"io/ioutil"
-	"encoding/json"
 )
 
 type loadResult struct {
@@ -116,7 +116,7 @@ func printCmd(a cmdArgs) error {
 		e := json.NewEncoder(stdout)
 		e.SetEscapeHTML(false)
 		if err := e.Encode(lr.routes); err != nil {
-			println(err.Error())
+			return err
 		}
 	} else {
 		for _, r := range lr.routes {
--- a/cmd/eskip/load.go
+++ b/cmd/eskip/load.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"github.com/zalando/skipper/eskip"
 	"io/ioutil"
+	"encoding/json"
 )
 
 type loadResult struct {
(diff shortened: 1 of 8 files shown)


cmd/eskip/load.go
```diff
@@ -111,14 +112,22 @@ func printCmd(a cmdArgs) error {
...
-		if perr, hasError := lr.parseErrors[r.Id]; hasError {
-			printStderr(r.Id, perr)
-		} else {
-			if r.Id == "" {
-				fmt.Fprintln(stdout, r.String())
+	if printJson {
+		e := json.NewEncoder(stdout)
+		e.SetEscapeHTML(false)
+		if err := e.Encode(lr.routes); err != nil {
+			println(err.Error())
```
aryszka: instead of `println(err.Error()` pls `return err`

cmd/eskip/load.go
```diff
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"github.com/zalando/skipper/eskip"
 	"io/ioutil"
+	"encoding/json"
```
aryszka: `go fmt ./...`

eskip/eskip.go
```diff
@@ -320,3 +324,119 @@ func GenerateIfNeeded(existingId string) string {
...
+	return json.Marshal(&struct {
+		Name string        `json:"name"`
+		Args []interface{} `json:"args"`
+	}{
+		Name: f.Name,
+		Args: args,
+	})
+}
+
+func (f *Predicate) MarshalJSON() ([]byte, error) {
```
aryszka: to avoid duplication, there could be a s.g. like:

```
func marhsalNameArgs(name string, args []interface{}) ([]byte, error) {
    	if args == nil {
		args = []interface{}{}
	}

	return json.Marshal(&struct {
		Name string        `json:"name"`
		Args []interface{} `json:"args"`
	}{
		Name: name,
		Args: args,
	})
}
```

and called from both the predicate and the filter.

+1 for the anonymous struct

eskip/eskip.go
```diff
@@ -320,3 +324,119 @@ func GenerateIfNeeded(existingId string) string {
...
+			rjf = append(rjf, &Predicate{
+				Name: "HeaderRegexp",
+				Args: []interface{}{k, v},
+			})
+		}
+	}
+
+	for _, p := range r.Predicates {
+		rjf = append(rjf, p)
+	}
```
aryszka: better:

```
rjf = append(rjf, r.Predicates...)
```

eskip/eskip.go
```diff
@@ -320,3 +324,119 @@ func GenerateIfNeeded(existingId string) string {
 	id = routeIdRx.ReplaceAllString(id, "x")
 	return "route" + id
 }
+
+func marshalJsonPredicates(r *Route) []*Predicate {
+	rjf := make([]*Predicate, 0)
```
aryszka: some nitpicking would be to set the capacity: `rjf := make([]*Predicate, 0, len(r.Predicates))`. Not significant in this case.

grassator:
@aryszka all should be fixed

PR #224, opened 2016-11-22
fix query param predicate documentation

aryszka:


lmineiro:
?

aryszka:
the documentation for this predicate was not visible in godoc

Issue #227, opened 2016-11-29
kube-skipper: data client for Kubernetes Ingress resources


hjacobs:
Create a new "data source" for Skipper to configure routes automatically from declared `Ingress` resources:
* Poll on Kubernetes `Ingress` resource http://kubernetes.io/docs/user-guide/ingress/
* Configure routes accroding to defined host and path matching

Idea is to use skipper as a HTTP router in combination with [kube-ingress-aws-controller](https://github.com/zalando-incubator/kube-ingress-aws-controller) to support `Ingress` on [Kubernetes on AWS](https://github.com/zalando-incubator/kubernetes-on-aws):

* kube-ingress-aws-controller
  * Deployed as deployment with replicas=1
  * Watch on Ingress resources
  * Configure one Application Load Balancer (ALB/ELBv2) per used SSL certificate
  * Attach Auto Scaling Group to ALB (using skipper NodePort)
* skipper with Kubernetes data client
  * Deployed as DaemonSet (thus running on every node)
  * Exposes NodePort on every EC2 instance
  * Watch on Ingress resources
  * Configure routes by defined host/path matching

hjacobs:
@aryszka started working on this: https://github.com/zalando/skipper/tree/kube-client

melTob:
@hjacobs @aryszka can we close this? I assume it`s already done.

hjacobs:
Yes, you can close this :smile:

PR #229, opened 2016-12-07
Kube client

aryszka:
partial implementation of #227 and https://github.com/zalando-incubator/kubernetes-on-aws/issues/169

(currently verbose debug logs left in the code to support testing of deployments)

### DIFF
--- a/dataclients/kube/definitions.go
+++ b/dataclients/kube/definitions.go
@@ -1,5 +1,10 @@
 package kube
 
+type metadata struct {
+	Namespace string `json:"namespace"`
+	Name      string `json:"name"`
+}
+
 type backend struct {
 	ServiceName string `json:"serviceName"`
 	ServicePort string `json:"servicePort"`
@@ -25,7 +30,8 @@ type ingressSpec struct {
 }
 
 type ingressItem struct {
-	Spec *ingressSpec `json:"spec"`
+	Metadata *metadata    `json:"metadata"`
+	Spec     *ingressSpec `json:"spec"`
 }
 
 type ingressList struct {
--- a/dataclients/kube/definitions.go
+++ b/dataclients/kube/definitions.go
@@ -1,0 +1,45 @@
+package kube
+
+type backend struct {
+	ServiceName string `json:"serviceName"`
+	ServicePort string `json:"servicePort"`
+}
+
+type pathRule struct {
+	Path    string   `json:"path"`
+	Backend *backend `json:"backend"`
+}
+
+type httpRule struct {
+	Paths []*pathRule `json:"paths"`
+}
+
+type rule struct {
+	Host string    `json:"host"`
+	Http *httpRule `json:"http"`
+}
+
+type ingressSpec struct {
+	DefaultBackend *backend `json:"backend"`
+	Rules          []*rule  `json:"rules"`
+}
+
+type ingressItem struct {
+	Spec *ingressSpec `json:"spec"`
+}
+
+type ingressList struct {
+	Items []*ingressItem `json:"items"`
+}
+
+type serviceSpec struct {
+	ClusterIP string `json:"clusterIP"`
+}
+
+type service struct {
+	Spec *serviceSpec `json:"spec"`
+}
+
+type serviceList struct {
+	Items []*service `json:"items"`
+}
--- a/dataclients/kube/definitions.go
+++ b/dataclients/kube/definitions.go
@@ -7,7 +7,7 @@ type metadata struct {
 
 type backend struct {
 	ServiceName string `json:"serviceName"`
-	ServicePort string `json:"servicePort"`
+	ServicePort int    `json:"servicePort"`
 }
 
 type pathRule struct {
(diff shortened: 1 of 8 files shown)


dataclients/kube/kube.go
```diff
@@ -0,0 +1,303 @@
...
+// logs if invalid, but proceeds with the valid ones
+// should report failures in Ingress status
+//
+// TODO:
+// - check how to set failures in ingress status
+func (c *Client) ingressToRoutes(items []*ingressItem) []*eskip.Route {
+	routes := make([]*eskip.Route, 0, len(items))
+	for _, i := range items {
+		if i.Metadata == nil || i.Metadata.Namespace == "" || i.Metadata.Name == "" {
+			log.Errorf("invalid ingress item: missing metadata")
```
mikkeloscar: This item should be skipped if it's invalid right?

dataclients/kube/kube.go
```diff
@@ -0,0 +1,303 @@
...
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/zalando/skipper/eskip"
+)
+
+// FEATURE:
+// - provide option to limit the used namespaces?
+
+const (
+	defaultAPIAddress = "http://localhost:8001"
```
mikkeloscar: Maybe it's a premature optimization, but I would make it possible to configure API access based on the environment when running in a cluster, so you don't have to have a 80+ MiB sidecar with `kubectl` just for this.

Example of how it's done in the kubernetes go client: https://github.com/kubernetes/client-go/blob/master/rest/config.go#L270-L298

I aslo made a PR for this in mate: https://github.com/zalando-incubator/mate/pull/21 :)

dataclients/kube/kube.go
aryszka: right

aryszka: updated, thx

mikkeloscar:
:+1:

skipper.go
```diff
@@ -66,6 +53,10 @@ type Options struct {
 	// Skip TLS certificate check for etcd connections.
 	EtcdInsecure bool
 
+	// Kubernetes API URL for ingress. By default, Kubernetes ingress is
+	// disabled.
+	KubeUrl string
```
hjacobs: I would propose using the long form "kubernetes" everywhere (we don't need to save any ASCII chars, do we?)

dataclients/kube/definitions.go
```diff
@@ -0,0 +1,115 @@
+package kube
```
hjacobs: same here: I would propose using the long form "kubernetes" everywhere (we don't need to save any ASCII chars, do we?)

dataclients/kube/kube.go
hjacobs: not sure what your conventions say: but I would log this as warning and not error (as we will happily continue with the next item)

hjacobs:
LGTM, see my comments regarding naming (I would prefer to use "kubernetes" in code, we are also not abbreviating "inkeeper" etc) and log level (stuff we can continue on is warning for me).

dataclients/kube/kube.go
```diff
@@ -0,0 +1,304 @@
...
+	rsp, err := http.Get(url)
+	if err != nil {
+		log.Debugf("request to %s failed: %v", url, err)
+		return err
+	}
+
+	log.Debugf("request to %s succeeded", url)
+	defer rsp.Body.Close()
+
+	if rsp.StatusCode != 200 {
```
Raffo: 200 -> http.StatusOK

skipper.go
```diff
@@ -233,6 +224,10 @@ func createDataClients(o Options, auth innkeeper.Authentication) ([]routing.Data
 		clients = append(clients, etcdClient)
 	}
 
+	if o.KubeUrl != "" {
```
Raffo: Url -> URL

dataclients/kube/kube.go
aryszka: yes, true, changing it

dataclients/kube/definitions.go
aryszka: for the other case, with the address option, i agree and changing right away. For the package name, i am a bit hesitant, but without a clear rational reasoning. Anyone with additional opinion on this.

dataclients/kube/kube.go
aryszka: right

skipper.go
aryszka: yes.. should start to use this convention

skipper.go
```diff
@@ -66,6 +53,10 @@ type Options struct {
 	// Skip TLS certificate check for etcd connections.
 	EtcdInsecure bool
 
+	// Kubernetes API base address for ingress. If not specififed,
+	// Kubernetes ingress is disabled.
+	KubernetesAPIAddress string
```
hjacobs: Why is it "Address" here and "Url" below? It should be Url here too I guess...

aryszka: i follow this naming convention to use address when it is only an address of a server, like http://localhost:8080 but no path or else. The innkeeper one just may be a full url, i don't remember. Is this dumb? Or are there better conventions for this distinction?

aryszka: maybe baseURL?

PR #231, opened 2016-12-08
Non-existence predicate for query parameters

haldun:


mo-gr:
Imho this is needless duplication. If we really want such a predicate to exist, I'd prefer to fold it into the query.go source file and use the exact same logic, just inverted.

However, I'm not yet convinced we really do need this predicate. What is your use case?

PR #233, opened 2016-12-09
change dockerfile: support eskip and skipper, change to alpine base image

aryszka:
(see PR title)

### DIFF
--- a/packaging/Dockerfile
+++ b/packaging/Dockerfile
@@ -1,6 +1,5 @@
 FROM alpine
 MAINTAINER Skipper Maintainers <team-pathfinder@zalando.de>
 RUN mkdir -p /usr/bin
-ADD skipper /usr/bin/skipper
-ADD eskip /usr/bin/eskip
+ADD skipper eskip /usr/bin/
 ENV PATH $PATH:/usr/bin
--- a/packaging/Dockerfile
+++ b/packaging/Dockerfile
@@ -1,8 +1,6 @@
-FROM registry.opensource.zalan.do/pathfinder/archlinux:latest-SNAPSHOT
+FROM alpine
 MAINTAINER Skipper Maintainers <team-pathfinder@zalando.de>
-RUN pacman --noconfirm -Syyu
-RUN pacman --noconfirm -S git go
-RUN mkdir -p /opt/go-workspace
-ENV GOPATH /opt/go-workspace
-ENV PATH $PATH:$GOPATH/bin
-RUN go get github.com/zalando/skipper/...
+RUN mkdir -p /usr/bin
+ADD skipper /usr/bin/skipper
+ADD eskip /usr/bin/eskip
+ENV PATH $PATH:/usr/bin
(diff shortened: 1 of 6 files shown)


packaging/Makefile
```diff
@@ -0,0 +1,19 @@
...
+SKIPPER_IMAGE    ?= $(SKIPPER_REGISTRY)/skipper:$(SKIPPER_VERSION)
+
+default: docker-build
+
+skipper:
+	CGO_ENABLED=0 go build github.com/zalando/skipper/cmd/skipper
+
+eskip:
+	CGO_ENABLED=0 go build github.com/zalando/skipper/cmd/eskip
+
```
lmineiro: We should forcibly build the binaries for Linux since that is the container OS

packaging/Makefile
```diff
@@ -0,0 +1,19 @@
...
+docker-build: skipper eskip
+	docker build -t $(SKIPPER_IMAGE) .
```
lmineiro: Also create a .dockerignore file to avoid pushing unnecessary files to the docker build context (At least the packaging/Makefile and the packaging/readme.md come to mind)

packaging/Makefile
aryszka: true

packaging/Makefile
aryszka: ok

packaging/Dockerfile
```diff
@@ -1,8 +1,6 @@
...
+FROM alpine
 MAINTAINER Skipper Maintainers <team-pathfinder@zalando.de>
-RUN pacman --noconfirm -Syyu
-RUN pacman --noconfirm -S git go
-RUN mkdir -p /opt/go-workspace
-ENV GOPATH /opt/go-workspace
-ENV PATH $PATH:$GOPATH/bin
-RUN go get github.com/zalando/skipper/...
+RUN mkdir -p /usr/bin
+ADD skipper /usr/bin/skipper
```
mo-gr: I think this could be one line `ADD skipper eskip /usr/bin`. This would make the resulting container have one layer less

packaging/Makefile
```diff
@@ -0,0 +1,19 @@
+.PHONY: skipper eskip docker-push
```
mo-gr: technically, skipper and eskip are not PHONY. They both generate a binary and you could use make here to only build it when actually necessary

packaging/Dockerfile
aryszka: right

packaging/Makefile
aryszka: yup, just then if you change something in the code, make may not detect that it needs to rebuild the binaries. Is there way to tell make that some source has changed? something like:

```
skipper: ../**/*.go
```

for me it didn't work by default. Any suggestions?

packaging/Makefile
```diff
@@ -0,0 +1,19 @@
...
+
+SKIPPER_VERSION  ?= $(shell git rev-parse HEAD)
```
aryszka: yes, i know. Semantic versioning is in the pipeline, but that comes with some more labor, regarding reviewing the deprecations. Planning to fix the default image version together with that. If this is now not completely unacceptable, then this ugly versions are my shot here. Suggestions?

PR #236, opened 2016-12-14
Kube client

aryszka:
as part of #227 and https://github.com/zalando-incubator/kubernetes-on-aws/issues/169

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -32,6 +32,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/filters/builtin"
+	"github.com/zalando/skipper/predicates/source"
 )
 
 // FEATURE:
@@ -42,7 +43,15 @@ const (
 	ingressesURI         = "/apis/extensions/v1beta1/ingresses"
 	serviceURIFmt        = "/api/v1/namespaces/%s/services/%s"
 	healthcheckRouteID   = "kube__healthz"
+	healthcheckPath      = "/kube-system/healthz"
 )
+
+var internalIPs = []interface{}{
+	"10.0.0.0/8",
+	"192.168.0.0/16",
+	"172.16.0.0/12",
+	"127.0.0.1/32",
+}
 
 // Options is used to initialize the Kubernetes DataClient.
 type Options struct {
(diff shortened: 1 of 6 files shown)


dataclients/kubernetes/kube.go
```diff
@@ -73,8 +97,13 @@ func New(o Options) *Client {
 		o.KubernetesURL = defaultKubernetesURL
 	}
 
-	log.Debugf("kube client initialized with api address: %s", o.KubernetesURL)
-	return &Client{apiURL: o.KubernetesURL}
+	log.Debugf("kube client initialized with api address: %s; with healtcheck: %t",
```
lmineiro: typo `healtcheck`

dataclients/kubernetes/kube.go
```diff
@@ -279,14 +317,49 @@ func (c *Client) loadAndConvert() ([]*eskip.Route, error) {
 	return r, nil
 }
 
+func healthcheckRoute(healthy bool) *eskip.Route {
+	status := http.StatusOK
+	if !healthy {
+		status = http.StatusNotFound
```
lmineiro: I'm a bit unsure if it shouldn't be a 5XX family error code. Like `503 - Service Unavailable`.
4XX are client errors and this is not the case. In practical terms it serves the same purpose.
What do you think?

dataclients/kubernetes/kube_test.go
```diff
@@ -207,6 +288,12 @@ func (api *testAPI) getTestService(uri string) (*service, bool) {
 }
 
 func (api *testAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if api.failNext {
+		api.failNext = false
+		w.WriteHeader(http.StatusServiceUnavailable)
```
lmineiro: http.StatusServiceUnavailable - nice! See my comment above.

dataclients/kubernetes/kube.go
hjacobs: Yes, 5xx is more appropriate! (does not make a practical difference, but we should follow HTTP standards..)

dataclients/kubernetes/kube.go
aryszka: right

dataclients/kubernetes/kube.go
aryszka: ok, thx

hjacobs:
:+1:

hjacobs:
Tests failing..

PR #238, opened 2016-12-16
Setting Content-Type header to application/json in debug

haldun:
`Content-Type` header of responses from debug listener have `plain/text`, however the body contains JSON. This change replaces it with `application/json`.


lmineiro:
Great stuff. Thanks for finding this.

lmineiro:
lgtm

PR #245, opened 2017-01-11
Basic per backend host metrics

mo-gr:
The background of this PR is the following:

We would like to create a dashboard, that shows us a distribution of where
skipper routes the requests to. Skipper currently supports metrics on a
per-backend base, but only keyed via the route-id. For our use-case, we would
need a mapping from route-id to backend-host. This is very brittle.

This is a simple solution, that would also provide a metric based on the
hostname of the backend. This would allow us to aggregate in ZMON. We could for
example create three buckets:
1. all requests, that go to *.example.org
2. all requests, that go to *.example.com
3. all others.
Plotting the request count over those three buckets would be exactly what
we want.

I tried to keep this as general as possible to be useful to others as well and
made it opt-in, because it might not be interesting for all users of skipper.

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -39,19 +39,24 @@ type Options struct {
 	// If set, detailed total response time metrics will be collected
 	// for each host, additionally grouped by status and method.
 	EnableServeHostMetrics bool
+
+	// If set, detailed response time metrics will be collected
+	// for each backend host
+	EnableBackendHostMetrics bool
 }
 
 const (
-	KeyRouteLookup     = "routelookup"
-	KeyRouteFailure    = "routefailure"
-	KeyFilterRequest   = "filter.%s.request"
-	KeyFiltersRequest  = "allfilters.request.%s"
-	KeyProxyBackend    = "backend.%s"
-	KeyFilterResponse  = "filter.%s.response"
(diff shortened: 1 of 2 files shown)


lmineiro:
I was wondering if this shouldn't be merged or done together with the existing servehost.* metrics.

I understand that those use the edge request Host and your change seems to require the Host of the backend.

I'm also not sure if the port, sometimes present in the backend URL, is relevant here.

aryszka:
@lmineiro how would the merge with the servehost metrics look?

the port technically makes it a different address, where there can be something else, i vote for having the port.

metrics/metrics.go
```diff
@@ -155,6 +160,12 @@ func (m *Metrics) MeasureBackend(routeId string, start time.Time) {
 	m.measureSince(fmt.Sprintf(KeyProxyBackend, routeId), start)
 }
 
+func (m *Metrics) MeasureBackendHost(routeBackendHost string, start time.Time) {
+	if m.options.EnableBackendHostMetrics {
+		m.measureSince(fmt.Sprintf(KeyProxyBackendHost, hostForKey(routeBackendHost)), start)
```
lmineiro: Does hostForKey() bring the port, if present?

aryszka: yes

PR #247, opened 2017-01-16
Service unavailable error

### LINKED ISSUE #241: Kubernetes: Skipper throws Internal Server Error for service without endpoints
I created a Kubernetes service and Ingress and the service has no endpoints (wrong label selector):

```
$ http https://myapp-ingress.example.org/
HTTP/1.1 500 Internal Server Error
Connection: keep-alive
Content-Length: 22
Date: Fri, 16 Dec 2016 22:24:39 GMT

Internal Server Error
```

I think it's because the service has no endpoints:
```
$ zkubectl describe svc myapp
Name:			myapp
Namespace:		default
Labels:			application=myapp
Selector:		<none>
Type:			ClusterIP
IP:			10.3.0.179
Port:			<unset>	80/TCP
Endpoints:		<none>
Session Affinity:	None
```

I would expect Skipper to return HTTP error 503 in that case:

> 503 Service Unavailable
> The server is currently unavailable (because it is overloaded or down for maintenance). Generally, this is a temporary state

aryszka:
return 503 Service Unavailable on errors when connecting to a backend service

Closes #241

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -20,6 +20,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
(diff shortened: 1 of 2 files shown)


proxy/proxy.go
```diff
@@ -568,10 +569,13 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 			if err != nil {
 				p.metrics.IncErrorsBackend(rt.Id)
-				sendError(w,
-					http.StatusText(http.StatusInternalServerError),
-					http.StatusInternalServerError)
-				p.metrics.MeasureServe(rt.Id, r.Host, r.Method, http.StatusInternalServerError, startServe)
+				code := http.StatusInternalServerError
+				if _, ok := err.(net.Error); ok {
```
hjacobs: not sure what this does and what values `err` can have

hjacobs:
LGTM, but somebody with more insights should probably have a look?

/cc @lmineiro @Raffo

lmineiro:
lgtm

Issue #251, opened 2017-01-18
consider using ULID for flowid


lmineiro:
Recently learned about https://github.com/oklog/ulid which, without entropy, seems very performant and could offer lexicographically sortable flow Ids.

I also like the alphabet better.

Existing filter can be refactored to support a a new parameter defining the generator.

head to head benchmarking should be enough to add to the docs.

I suggest the fastest to become the default

lmineiro:
Results on Mac OS X, 2 Ghz Core i7
current flowid
```
BenchmarkFlowIdLen8-8    	20000000	       109 ns/op
BenchmarkFlowIdLen10-8   	10000000	       127 ns/op
BenchmarkFlowIdLen12-8   	10000000	       163 ns/op
BenchmarkFlowIdLen14-8   	10000000	       165 ns/op
BenchmarkFlowIdLen16-8   	10000000	       175 ns/op
BenchmarkFlowIdLen32-8   	 5000000	       277 ns/op
BenchmarkFlowIdLen64-8   	 3000000	       477 ns/op
```
oklog/ulid
```
BenchmarkNew/WithCryptoEntropy-8         	 2000000	       937 ns/op	  17.06 MB/s
BenchmarkNew/WithEntropy-8               	20000000	        73.5 ns/op	 217.81 MB/s
BenchmarkNew/WithoutEntropy-8            	50000000	        35.5 ns/op	 451.29 MB/s
BenchmarkMustNew/WithCryptoEntropy-8     	 2000000	       954 ns/op	  16.77 MB/s
BenchmarkMustNew/WithEntropy-8           	20000000	        79.5 ns/op	 201.26 MB/s
BenchmarkMustNew/WithoutEntropy-8        	50000000	        40.0 ns/op	 399.71 MB/s
BenchmarkParse-8                         	50000000	        33.8 ns/op	 769.97 MB/s
BenchmarkMustParse-8                     	50000000		        37.6 ns/op	 692.25 MB/s
BenchmarkString-8                        	20000000	        73.3 ns/op	 218.29 MB/s
BenchmarkMarshal/Text-8                  	20000000	        60.6 ns/op	 264.19 MB/s
BenchmarkMarshal/TextTo-8                	50000000	        26.1 ns/op	 613.48 MB/s
BenchmarkMarshal/Binary-8                	300000000	         5.11 ns/op	3132.90 MB/s
BenchmarkMarshal/BinaryTo-8              	2000000000	         1.81 ns/op	8849.95 MB/s
[...]

PR #253, opened 2017-01-19
refactored flowid to support having custom generators

lmineiro:
Added a Universally Unique Lexicographically Sortable Identifier (ULID) flow ID generator

Relevant results:
```
BenchmarkFlowIdULIDGenerator/Std-8  	10000000	       207 ns/op
BenchmarkFlowIdBuiltInGenerator/16-8         	10000000	       188 ns/op
BenchmarkFlowIdBuiltInGenerator/26-8         	 5000000	       254 ns/op
```

Previous default uses a 16 char flowid (188ns) and the new ULID generator generates a 26 char flowid (207ns)

I'd say that the extra cost is negligible and this new generator could become the new flowid default generator. I'd be particularly interested in having the sortable flowids.
To be fair, if we wanted 26 char long flowid using the current algorithm it would be actually slower (254ns) than the ULID.


filters/flowid/generator-ulid.go
```diff
@@ -0,0 +1,40 @@
...
+func newULIDGenerator() flowIDGenerator {
+	return newULIDGeneratorWithEntropyProvider(rand.New(rand.NewSource(time.Now().UTC().UnixNano())))
+}
+
+func newULIDGeneratorWithEntropyProvider(r io.Reader) flowIDGenerator {
+	return &ulidGenerator{r: r}
+}
+
+func (g *ulidGenerator) Generate() (string, error) {
+	g.Lock()
```
aryszka: this is somewhat fishy to me. I see https://github.com/oklog/ulid#usage, but I wonder what is the cost of calling rand.New() + reading those few bytes from it on every request. I see the chance that it is better than both doing the synchronization here or using a sync.Pool. Let's do some parallel benchmark on it some time.

lmineiro:
Most changes are compromises from poorly written code and bad decisions I made when writing the filter long time ago. NewFlowID() should, probably, never have been exported in the first place.

I struggled myself to decide on how to go forward with the exporting of the new generator interface. I had in mind to make it a first class citizen and went against it at the last minute, giving preference to review feedback.

I think we're both thinking in the same way, overall, how to go forward with this.

filters/flowid/generator-ulid.go
lmineiro: Creating new ones on every call will hurt the garbage collector. My first attempt resulted in:

```
BenchmarkFlowIdULIDGenerator/Std-8  	  100000	     11762 ns/op
```
For doing it like this instead:
```
	//g.Lock()
	r := rand.New(rand.NewSource(time.Now().UTC().UnixNano()))
	id, err := ulid.New(ulid.Now(), r)
	//g.Unlock()
```

I added details for all those attempts as comments for #251. A completely, non thread safe, usage of the same entropy provider wasn't faster than 170ns.

lmineiro: More detailed, benchmem results: (not parallel benchmarks)

New entropy provider on each request:
```
BenchmarkFlowIdULIDGenerator/Std-8         	  100000	     11807 ns/op	    5456 B/op	       4 allocs/op
```

Using the mutex:
```
BenchmarkFlowIdULIDGenerator/Std-8         	10000000	       203 ns/op	      48 B/op	       2 allocs/op
```

lmineiro: Parallel benchmarks:

New entropy provider on each request:
```
BenchmarkFlowIdULIDGenerator-8   	 1000000	      2543 ns/op	    5456 B/op	       4 allocs/op
```

Using the mutex:
```
BenchmarkFlowIdULIDGenerator-8   	 5000000	       384 ns/op	      48 B/op	       2 allocs/op
```

filters/flowid/hashing.go
```diff
@@ -10,33 +10,53 @@ import (
...
 
 var (
 	ErrInvalidLen = errors.New(fmt.Sprintf("Invalid length. Must be between %d and %d", MinLength, MaxLength))
 	flowIdRegex   = regexp.MustCompile(`^[0-9a-zA-Z+-]+$`)
 )
 
-// NewFlowId returns a random flowId using the flowIdAlphabet with length l.
-//
+// NewFlowId creates a new built-in generator with the defined length and returns a flowid
+// This exported function is deprecated and the new flowIdGenerator interface should be used
```
lmineiro: Was trying to avoid breaking backwards compatibility. I saw that we ended up using it also to generate random route ids.

lmineiro:
We still need one last decision - do we change the default to the new ULID? It's already faster for parallel use, than the default with 16 chars.

filters/flowid/filter.go
```diff
@@ -54,16 +72,13 @@ func (spec *flowIdSpec) CreateFilter(fc []interface{}) (filters.Filter, error) {
...
+				"please check updated docs")
 		}
 	}
-	return &flowId{reuseExisting, flowIdLength}, nil
+	return &flowId{reuseExisting: reuseExisting, generator: spec.generator}, nil
 }
 
-func (spec *flowIdSpec) Name() string { return Name }
+// Name returns the canonical filter name
+func (_ *flowIdSpec) Name() string { return Name }
```
aryszka: i see that in the same cloud of services one would want just a single type of flowId, but for the record: i would have allowed the custom implementations to have different names. Since they can still do it with completely independent filters, it's cool maybe even better so.

filters/flowid/generator.go
```diff
@@ -0,0 +1,11 @@
+package flowid
```
aryszka: Question, have you considered that this interface is something that the code in filter.go expects? Maybe, it could moved to there. I still think that it doesn't deserve its own file.

filters/flowid/standard.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	return &standardGenerator{length: l}, nil
+}
+
+// Generate returns a new Flow ID from the built-in generator with the configured length
+// The alphabet is limited to 64 elements and requires a random 6 bit value to index any of them.
+// The cost to rnd.IntXX is not very relevant but the bit shifting operations are faster.
+// For this reason a single call to rnd.Int63 is used and its bits are mapped up to 10 chunks of 6 bits each.
+// The byte data type carries 2 additional bits for the next chunk which are cleared with the alphabet bit mask.
```
aryszka: your're aware that comments of the unpublished members, like standardGenerator won't be visible in godoc? This doc contains useful information. Please move the gist of it to the NewStandardGenerator, as somthing like NewStandardGenerator creates a new FlowID generator that ... <this content>

aryszka:
"We still need one last decision - do we change the default to the new ULID? It's already faster for parallel use, than the default with 16 chars."

had thought about it: we focus on parallel use, OK. Only very special or very dumb code would get broken by this change because of the byte size, true. On the other hand, well, it still could. I think we should stick to the status quo for now, but in the next breaking release, i want to have the ulid as the default.

Does this change the preferred naming?

filters/flowid/filter.go
lmineiro: I understand and I also spent some time thinking about it. Since this is a relatively easy functionality to deliver, I propose we wait for someone to specifically to ask for it. Until then, we keep the previous functionality of having one single flow ID format per skipper instance.

filters/flowid/generator.go
lmineiro: It doesn't bother me to move this to the filter.go file. I also thought how the entire set of Generator interface and implementations could be moved anywhere else. This was, basically, the reason why I already left it in this shape. Since we don't have any plans on moving them now, and they'll be exported afterwards, the interface might as well be moved to the filter.go file.

filters/flowid/standard.go
lmineiro: I didn't want to lose the comments and couldn't come up with a good location :)

lmineiro: Isn't this a comment on the exported `Generate()`?

lmineiro: Moved to `NewStandardGenerator()`

filters/flowid/filter.go
aryszka: cool

aryszka:
shipit

PR #256, opened 2017-01-20
Flowid generators interface

aryszka:
this sub PR contains some suggestions to the package's exported interface. It provides the possibility to define new filters with custom generator implementations. It contains two open questions:

- shall we split `flowId("reuse", 8)` and `flowId("reuse", "ulid")` to `flowId("reuse", 8)` and `ulidFlowId("reuse")`?
- whether it is fine that the custom filters created by `WithGenerator(name, g)` won't have route specific arguments this way?

### DIFF
--- a/filters/flowid/filter.go
+++ b/filters/flowid/filter.go
@@ -12,15 +12,35 @@ const (
 	HeaderName          = "X-Flow-Id"
 )
 
-type flowIdSpec struct{}
+// The Generator interface can be used to define custom request tracing Flow ID implementations.
+type Generator interface {
+
+	// Generate returns a new Flow ID using the implementation specific format or an error in case of failure.
+	Generate() (string, error)
+
+	// MustGenerate behaves like Generate but panics on failure instead of returning an error.
+	MustGenerate() string
+}
+
+type flowIdSpec struct{
+	name string
+	generator Generator
+}
 
 type flowId struct {
 	reuseExisting bool
-	generator     flowIDGenerator
+	generator     Generator
 }
 
-func New() *flowIdSpec {
+// and here i would still consider if we should have a flowId() and a ulidFlowId() filter with different names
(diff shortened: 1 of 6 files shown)


lmineiro:
I think we should go ahead with a single flowid generator design.

We export the creation of any flowid.Generator which shall be used as the argument to create the flowId filter spec, as you also suggested.

We can also keep the backwards compatibility `flowid.New()` that creates a new spec using the StandardGenerator.

Should we export directly the functions that create new, specific, Generators, like `NewStandardGenerator()` and `NewULIDGenerator()`  since they'll have different "constructors" or we only export a more generic `NewGenerator(generatorID string, args ... interface{})` ?

I'd would also prefer to have only a `flowId()` or `flowId("reuse")` filter approach, probably still accepting the `flowId("reuse", n)` for backwards compatibility but logging an error/warning of this deprecation.

Backwards compatibility is something to put a lot of focus or are you willing to break it?

aryszka:
> I think we should go ahead with a single flowid generator design.

ok

> Should we export directly the functions that create new, specific, Generators, like NewStandardGenerator() and NewULIDGenerator() since they'll have different "constructors" or we only export a more generic NewGenerator(generatorID string, args ... interface{}) ?

I vote for exporting them directly, but don't mind the other way too much either.

> I'd would also prefer to have only a flowId() or flowId("reuse") filter approach, probably still accepting the flowId("reuse", n) for backwards compatibility but logging an error/warning of this deprecation.

yes, this would be good

> Backwards compatibility is something to put a lot of focus or are you willing to break it?

I would like to avoid breaking it in this round, but we can mark it deprecated in the docs. I am working on proper versioning, and once we have it, we can define a major release where all the deprecated stuff is gone.

lmineiro:
Agreed on the interface, merged with #253

Issue #248, opened 2017-01-18
Watch for changes in `routes-file`


mo-gr:
If skipper is started with a `-routes-file` argument, it would be cool, if skipper would watch for changes to that file and update the routing tables accordingly.

lmineiro:
I did some initial research with https://github.com/fsnotify/fsnotify, which seems to be the de facto standard for this.

So far, under Mac OS X this seems to be a luck game. The project mentions issues with Spotlight. I tried to apply the mentioned workarounds but they don't seem to really solve the problem.

The events I got:
```
[APP]INFO[0024] event: "example.eskip": CHMOD                
[APP]INFO[0024] event: "example.eskip": REMOVE               
[APP]INFO[0024] event: "example.eskip": CREATE               
[APP]INFO[0025] event: "example.eskip": CHMOD                
[APP]INFO[0041] event: "example.eskip": CHMOD                
[APP]INFO[0041] event: "example.eskip": REMOVE               
[APP]INFO[0041] event: "example.eskip": CREATE               
[APP]INFO[0041] event: "example.eskip": CHMOD                
[APP]INFO[0041] event: "example.eskip": REMOVE               
[APP]INFO[0041] event: "example.eskip": CREATE               
[APP]INFO[0044] event: "example.eskip": CHMOD                
[APP]INFO[0189] event: "example.eskip": CHMOD                
[APP]INFO[0191] event: "example.eskip": CHMOD                
[APP]INFO[0200] event: "example.eskip": RENAME 
```

And then it stopped working. The fsnotify seems to be the more elegant way to do this but there seems to be some shortcomings.

I'll test this on Linux (native and Docker) to see if it behaves any different.

lmineiro:
https://github.com/andreaskoch/go-fswatch could be a reasonable alternative

lmineiro:
Tests using https://github.com/rjeczalik/notify are more promising and seem to work fine under Mac OS X:

```
[APP]INFO[0024] detected changes (notify.Write) in routes file "example.eskip" 
[APP]INFO[0024] route settings, update, route: all: * -> "https://example.org" 
[APP]INFO[0024] route settings, update, route: foo: Path("/foo") -> setPath("/") -> "https://example.org" 
[APP]INFO[0024] route settings, update, route: bar: Path("/bar") -> setPath("/") -> "https://example.org" 
[APP]INFO[0024] route settings received                      
[APP]INFO[0024] route settings applied 
```

kode4food:
[...]

PR #257, opened 2017-01-21
first draft of fswatch on the routes file

lmineiro:
DON'T MERGE
This is a first draft for #248 and I'd like to get early feedback.

It seems to work fine on Mac OS X. It would be great if someone else out there could test with other OSes.
Under Docker, both with alpine and ubuntu I couldn't make it work. I tried both by having the file inside the container and also by mapping a volume.

```
docker run --name skipper --rm -v $(pwd)/example.eskip:/example.eskip -t skipper -routes-file example.eskip
```

inotifywatch did get the events
```
root@40c37740610e:/# inotifywatch -v -t 10 /example.eskip 
Establishing watches...
Total of 1 watches.
Finished establishing watches, now collecting statistics.
Will listen for events for 10 seconds.
total  modify  attrib  filename
12     6       6       /example.eskip
```

### DIFF
--- a/eskipfile/file.go
+++ b/eskipfile/file.go
@@ -16,36 +16,73 @@
 Package eskipfile implements a DataClient for reading the skipper route
 definitions from an eskip formatted file when opened.
 
+It supports runtime updates of routes by watching changes to the file.
+
 (See the DataClient interface in the skipper/routing package and the eskip
 format in the skipper/eskip package.)
 */
 package eskipfile
 
 import (
-	"github.com/zalando/skipper/eskip"
 	"io/ioutil"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/rjeczalik/notify"
+	"github.com/zalando/skipper/eskip"
 )
 
 // A Client contains the route definitions from an eskip file.
-type Client struct{ routes []*eskip.Route }
+type Client struct {
+	path     string
+	routes   []*eskip.Route
+	fsevents chan notify.EventInfo
+}
 
-// Opens an eskip file and parses it, returning a DataClient implementation.
+// Open reads an eskip file and parses it, returning a DataClient implementation.
 // If reading or parsing the file fails, returns an error.
-func Open(path string) (*Client, error) {
+func Open(p string) (*Client, error) {
+	r, err := readRoutesFromFile(p)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Client{routes: r, path: p}, nil
+}
+
+// Watch behaves like Open but it starts watching the eskip file for changes.
+// This enabled runtime updates of routes when using this DataClient.
+// The filesystem watcher created by this call should be properly Closed using this Client's Close function.
+func Watch(p string) (*Client, error) {
+	c, err := Open(p)
+	if err != nil {
+		return nil, err
+	}
+	c.fsevents = make(chan notify.EventInfo, 1)
+	// We only care about Write events and Rename for "atomic" saves from some editors
+	if err := notify.Watch(c.path, c.fsevents, notify.Write, notify.Rename); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+// Close does a clean shutdown of the Client. It is not needed unless the Client was created using the
+// Watch function, since it creates a filesystem watcher
+func (c *Client) Close() {
+	if c.fsevents != nil {
+		notify.Stop(c.fsevents)
+	}
+}
+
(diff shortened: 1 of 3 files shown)


aryszka:
on linux (fairly bare and recent arch linux):

```
[APP]ERRO[0013] error while receiving update;open routes.eskip: no such file or directory 
```

```
:) skipper -routes-file routes.eskip
[APP]INFO[0000] metrics listener on :9911/metrics            
[APP]INFO[0000] route settings, update, route: example: * -> setRequestHeader("Accept", "flate") -> "https://www.example.org" 
[APP]INFO[0013] detected changes in routes file "routes.eskip", reloading ... 
[APP]ERRO[0013] error while receiving update;open routes.eskip: no such file or directory 
[APP]INFO[0013] route settings, update, route: example: * -> setRequestHeader("Accept", "flate") -> "https://www.example.org" 
[APP]INFO[0013] route settings received                      
[APP]INFO[0013] route settings applied 
```

what i did, started skipper with -routes-file routes.eskip, then edited the file and saved, and from [0013] is what happened. Hope, it helps.

aryszka:
another possible issue: removing the file was not detected. Question: would users expect to have the following scenario working:

1. start skipper with an eskip file: `skipper -routes-file example.eskip`
2. move the file into backup: `mv example.eskip{,.backup}`
3. create a new file with different content, called example.eskip

I can imagine that the expected behavior would be that all routes get deleted after step 2 and new ones are created in step 3

aryszka:
how about using watch, but as a backup, also regularly check the file in a polling fashion? It could work so that if the poll detects that the watch didn't trigger when it should have, the data client reinitiates itself with resubscription to the watch events and sending a routes reset to the routing?

lmineiro:
Deliberately, only Write and Rename events were added to the watch. I actually expected Delete not to be detected, although I don't have a clear indication which event would be triggered by a delete.

It's very weird that the reload didn't work out. It uses exactly the same function to read the routes file as Open/Watch, which seems to have succeeded on:
```
[APP]INFO[0000] route settings, update, route: example: * -> setRequestHeader("Accept", "flate") -> "https://www.example.org"
```

Also, the fact that on [0013] there are also logs about the routing being reloaded looks weird. Does the routing manager reload, even on error from a dataclient?

The above: 1,2,3 strategy is basically what I left as a comment about the watching Rename. Most editors provide "atomic" save by:
1. Save current content on new, temp file
2. Rename temp file to the effective file
3. Send Rename event

aryszka:
"Also, the fact that on [0013] there are also logs about the routing being reloaded looks weird. " -> i think this is a consequence of the update reporting an error. In this case the routing asks for a full load.

aryszka:
"Most editors provide "atomic" save by" -> checked it was the case. When simply appending to the file, the update worked :)

lmineiro:
Interesting. This would mean that it fired a Write event when the file was deleted, then causing the failure to open it.

lmineiro:
@mo-gr can you test this with your favourite editor "du jour"?

aryszka:
makes sense if vim did exactly the thing you described

lmineiro:
Seems that vim moves file.eskip to file.eskip~ (backup) and then saves the buffer to a new, file.eskip. file.eskip~ keeps the old inode, which can probably mess with the watches.

eskipfile/file.go
```diff
@@ -16,45 +16,105 @@
...
 }
 
 // Returns the parsed route definitions found in the file.
-func (c Client) LoadAll() ([]*eskip.Route, error) { return c.routes, nil }
+func (c *Client) LoadAll() ([]*eskip.Route, error) { return c.routes, nil }
+
+// LoadUpdate is No-Op if the Client was created using Open.
+// When created using Watch, it monitors the filesystem for changes on the eskip file and enables runtime updates
+// of the routes
+func (c *Client) LoadUpdate() ([]*eskip.Route, []string, error) {
```
mo-gr: What is the purpose of the string array here?

mo-gr:
Tested successfully on macOS 10.12.2 with emacs and vim.

eskipfile/file.go
lmineiro: As far as I understood, this part of the DataClient interface is used to report the IDs of deleted routes. Maybe @aryszka can confirm.

aryszka:
@lmineiro @mo-gr "IDs of deleted routes": correct.

aryszka:
@lmineiro Closing this PR for now in favor of https://github.com/zalando/skipper/pull/588

If we want to make the file watch smarter then the version in #588 , let's discuss it.

PR #258, opened 2017-01-26
Introducing sem ver

aryszka:
DON'T MERGE JUST YET

Before merging, we have to tag an initial version. I'd like to start with v0.0.9, is it ok?

How it works:

1. on pushing to branches, it runs the tests as currently
2. on creating a pull request, it runs the tests, checks if formatting was applied, and runs go vet
3. merge:
a) when merging to master, it bumps the patch version
b) when merging to master, and the commit message contains "minor-release" (the branch name is automatically picked when merging), bumps the minor version
c) like 3b but for the major version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,10 @@
-SOURCES = $(shell find . -name '*.go')
+SOURCES =         $(shell find . -name '*.go')
+CURRENT_VERSION = $(shell git tag | sort -V | tail -n1)
+VERSION =         $(CURRENT_VERSION)
+NEXT_MAJOR =      $(shell go run packaging/version/version.go major $(currentVersion))
+NEXT_MINOR =      $(shell go run packaging/version/version.go minor $(currentVersion))
+NEXT_PATCH =      $(shell go run packaging/version/version.go patch $(currentVersion))
+COMMIT_HASH =     $(shell git rev-parse --short HEAD)
 
 default: build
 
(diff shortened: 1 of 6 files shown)


lmineiro:
NICE!

Makefile
```diff
@@ -17,15 +35,58 @@ shortcheck: build
...
 	go vet ./...
 
-fmt:
+fmt: $(SOURCES)
 	gofmt -w $(SOURCES)
 
-checkfmt:
-	if [[ $$(gofmt -d $(SOURCES)) ]]; then false; else true; fi
+check-fmt: $(SOURCES)
+	if [ "$$(gofmt -d $(SOURCES))" != "" ]; then false; else true; fi
```
mo-gr: couldn't this be achieved with the `test` command? An if/else that does nothing but return true/false always looks weird to me

packaging/version/version.go
```diff
@@ -0,0 +1,67 @@
...
+	case cmdPatch:
+		*patch++
+	}
+}
+
+func main() {
+	cmd, current, err := command()
+	if err != nil {
+		log.Println(usage())
+		return
```
mo-gr: return with non-zero return code maybe?

mo-gr:
Awesome!

packaging/version/version.go
aryszka: right, os even imported already

aryszka: log.Fatalln even better

PR #261, opened 2017-01-27
Changed handling of transient kube apiserver errors

### LINKED ISSUE #260: Kubernetes: do not switch back the health check to false when failing to load Ingress resources
We should never toggle back the health check to `false` in case of transient Kubernetes errors, as this would remove the node from the load balancer. The current behavior will disable all apps if the API server is temporary unavailable (e.g. because of etcd problems).

Only this `if` needs to be removed: https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/kube.go#L347

Raffo:
Closes #260.

### DIFF
--- a/dataclients/kubernetes/kube_test.go
+++ b/dataclients/kubernetes/kube_test.go
@@ -855,10 +855,10 @@ func TestHealthcheckInitial(t *testing.T) {
 
 		r, err := dc.LoadAll()
 		if err != nil {
-			t.Error(err)
+			t.Log("Got an error, routes will not be changed. This is fine.")
 		}
 
-		checkHealthcheck(t, r, true, false)
+		checkHealthcheck(t, r, false, false)
 	})
 }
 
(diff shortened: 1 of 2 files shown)


lmineiro:
@aryszka should confirm that returning err to the routing handler will keep old routing table

lmineiro:
lgtm

Raffo:
Tests failing, nice! On it.

aryszka:
@lmineiro @Raffo "err to the routing handler will keep old routing table": correct

dataclients/kubernetes/kube_test.go
```diff
@@ -855,10 +855,10 @@ func TestHealthcheckInitial(t *testing.T) {
 
 		r, err := dc.LoadAll()
 		if err != nil {
```
aryszka: i think here we want to test if we actually have an error, so i recommend something like this:

```
if err == nil {
    t.Error("error expected")
}
```

and the call to `checkHealthCheck` is unnecessary from that point

dataclients/kubernetes/kube_test.go
```diff
@@ -987,15 +987,15 @@ func TestHealthcheckReload(t *testing.T) {
 			ProvideHealthcheck: true,
 		})
 
-		dc.LoadAll()
+		prevr, _ := dc.LoadAll()
 		api.failNext = true
 
 		r, err := dc.LoadAll()
 		if err != nil {
-			t.Error("failed to fail")
```
aryszka: similarly to my other comment.

Actually, i think the tests "use healthcheck, fail" and "use healthcheck, reload fail" are not necessary anymore because they were testing a behavior that was removed. At least we can fix something by purely removing code :)

dataclients/kubernetes/kube_test.go
aryszka: better delete "use healthcheck, fail" and "use healthcheck, reload fail" tests. See the explanation in the other comment.

dataclients/kubernetes/kube_test.go
Raffo: Ok, then I can remove them.

hjacobs:
:+1:

PR #262, opened 2017-02-02
Adding zappr config file

### LINKED ISSUE #235: Configure Zappr
We should enable Zappr, right?

Joneser:
Adding the zappr config file to resolve #235

### DIFF
--- a/.zappr.yaml
+++ b/.zappr.yaml
@@ -1,6 +1,8 @@
 approvals:
-  minimum: 2
+  minimum: 1
   ignore: pr_opener
-  pattern: "^(:\\+1:|👍)$"
+  pattern: "^(:\\+1:|👍|lgtm|shipit)$"
   veto:
     pattern: "^(:\\-1:|👎)$"
+  from:
+    collaborators: true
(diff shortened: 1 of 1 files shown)


.zappr.yaml
```diff
@@ -0,0 +1,6 @@
+approvals:
+  minimum: 2
+  ignore: pr_opener
+  pattern: "^(:\\+1:|👍)$"
```
aryszka: can we also use shipit and lgtm?

.zappr.yaml
```diff
@@ -0,0 +1,6 @@
+approvals:
+  minimum: 2
```
aryszka: i would bet 1 would be enough for an official approval

aryszka: would add this:

```
  from:
    collaborators: true
```

or is this the default?

(https://zappr.readthedocs.io/en/latest/setup/)

Joneser: It isn't there by default, i'll add along with your other suggestions

Joneser:
@aryszka Config updated :)

aryszka:
lgtm

PR #266, opened 2017-02-02
add prefix to error messages where it was missing

### LINKED ISSUE #263: improve error logging
there are some error log calls that don't pass any additional information or code reference with the error. E.g: `log.Error(err)`. Prepend some explanation or reference to these cases.

aryszka:
add a prefix to error messages that log only an error object. Closes #263


lmineiro:
👍

PR #275, opened 2017-02-17
Prototype for watching files & directories for routes

sarslanhan:
**DON'T MERGE**

Prototype for #248.

You can load directories using the same command for files.

I could test a little but create, update, deletes and renames for files are working.


kode4food:
Hi all, can someone explain to me the use case for this feature?

lmineiro:
This is exactly for what #248 mentions - hot reload of routes without restarting the process.

We saw the benefit this can have during development but also for all other users that want to keep skipper running and still be able to update the routes at runtime without any big infrastructure pieces like etcd or innkeeper.

aryszka:
@sarslanhan closing this branch now in favor of https://github.com/zalando/skipper/pull/588 .

If we want to make the file watch smarter then the version in #588 , let's discuss it.

PR #276, opened 2017-02-21
go vet fix: check error before using response

aryszka:
go vet ./... spots now the issue being fixed here


Raffo:
That's a classic go mistake I always check in PRs, if go vet keeps getting better I'll be useless :D

Raffo:
👍

Issue #269, opened 2017-02-06
Skipper produces 500 "Internal Server Error" HTTP status code on EOF


hjacobs:
We observe sporadic issues with Skipper connecting to Kubernetes `ClusterIP` service. Relevant Skipper log lines:

500:
```
02:26:40.814 log='[APP]time="2017-02-05T01:26:39Z" level=error msg=EOF'
02:26:40.814 log='52.59.216.130 - - [05/Feb/2017:01:26:39 +0000] "GET / HTTP/1.1" 500 22 "" "zmon-worker/v130" 2 hjacobs-test.example.org'
```

500:
```
20:10:14.338 log='[APP]time="2017-02-05T19:10:10Z" level=error msg=EOF'
20:10:14.338 log='52.59.216.130 - - [05/Feb/2017:19:10:10 +0000] "GET / HTTP/1.1" 500 22 "" "zmon-worker/v130" 14 hjacobs-test.example.org'
```

Expected behavior: Skipper should never produce "Internal Server Error", but instead handle the problem and report an appropriate status code (maybe 503?).

aryszka:
the msg=EOF is weird. At first sight not sure what's the source of it, spotted in other deployments, too. Deploying this could help: https://github.com/zalando/skipper/pull/266/files , can I get +1 there if the PR is ok?

aryszka:
https://github.com/zalando/skipper/releases/tag/v0.9.5

aryszka:
code to reproduce:

https://gist.github.com/aryszka/6da6e379750994ae348646a88ecd84db

(TLS is not needed for reproducing it, but here's a makefile that generates certs for TLS tests and runs the above code:
https://gist.github.com/aryszka/0d1a657ae6f7ac660a17cab79424173c)

aryszka:
considering the behavior of the code linked above, the issue can be that the connection close on the server doesn't happen immediately, even if it is reported to be closed. In this case the 'write: broken pipe' error makes sense, and the only weird thing is when it reports EOF.

So reporting '503 Service Unavailable' seems legit. Or maybe '502 Bad Gateway'?

szuecs:
I would suggest 502 or 504 https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_Server_Error

503 would mean that skipper itself is unavailable.

aryszka:
[...]

PR #277, opened 2017-02-21
set the body to nil for outgoing requests, when the content length is zero

aryszka:
this suppose to fix the trivially reproducible part of this issue: https://github.com/zalando/skipper/issues/269

EDIT: to emphasize, it enables http.Transport to retry the connection

See: https://golang.org/src/net/http/request.go#L1300

(naturally, there was a facepalm moment :) )

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -227,7 +227,12 @@ func mapRequest(r *http.Request, rt *routing.Route, host string) (*http.Request,
 	u.Scheme = rt.Scheme
 	u.Host = rt.Host
 
-	rr, err := http.NewRequest(r.Method, u.String(), r.Body)
+	body := r.Body
+	if r.ContentLength == 0 {
+		body = nil
+	}
+
+	rr, err := http.NewRequest(r.Method, u.String(), body)
 	if err != nil {
 		return nil, err
 	}
@@ -576,7 +581,7 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 				sendError(w, http.StatusText(code), code)
 				p.metrics.MeasureServe(rt.Id, r.Host, r.Method, code, startServe)
-				log.Error("error during backend roundtrip", err)
+				log.Error("error during backend roundtrip: ", err)
 				return
 			}
 		}
--- a/proxy/proxy_test.go
+++ b/proxy/proxy_test.go
@@ -936,7 +936,6 @@ func TestRoundtripperRetry(t *testing.T) {
 			return
 		}
 
-		println("closing now")
 		if err := l.lastConn.Close(); err != nil {
 			t.Error(err)
 			return
--- a/proxy/proxy_test.go
+++ b/proxy/proxy_test.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"math/rand"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
@@ -51,6 +52,11 @@ type testProxy struct {
 	log     *loggingtest.Logger
 	routing *routing.Routing
 	proxy   *Proxy
+}
+
+type listener struct {
+	inner    net.Listener
+	lastConn net.Conn
 }
 
 func (cors *preserveOriginalSpec) Name() string { return "preserveOriginal" }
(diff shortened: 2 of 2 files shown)


hjacobs:
@aryszka do you have "more data" for me to show that this would fix our 5xx problem (or is it just one part of it)?

aryszka:
@hjacobs yes, the linked line in the stdlib is used as one of the conditions by http.Transport to decide whether a request can be retried. In our code, this mapRequest() call happens right before we call http.Transport.RoundTrip().

See also: https://golang.org/src/net/http/transport.go#L411

There is no explicit guarantee that it fixes the whole issue, because this is working only for those cases where the connection is taken from the pool and not for new connections. On the other hand, the provided test consistently can reproduce the EOF errors, if the fix is removed.

I don't know yet if all of the cases observed with kubernetes fall into this case of broken connections from the pool or are there such cases with new connections, too, for that I am still running a test.

As for the new connections, i'm currently preparing a fix in another branch, that just wraps a roundtripper with custom retry.

aryszka:
@hjacobs tests reproduced the last 24 hours 4 times. Each of them fall into the case which gets fixed by this PR:

```
{"log":"2017/02/21 23:44:21 it is from pool\n","stream":"stderr","time":"2017-02-21T23:44:21.023001147Z"}  
{"log":"2017/02/21 23:44:21 http/1.1\n","stream":"stderr","time":"2017-02-21T23:44:21.023008484Z"}  
{"log":"2017/02/21 23:44:21 from peek true\n","stream":"stderr","time":"2017-02-21T23:44:21.023012873Z"}  
{"log":"2017/02/21 23:44:21 checking whether to set request canceler\n","stream":"stderr","time":"2017-02-21T23:44:21.023168254Z"}  
{"log":"2017/02/21 23:44:21 checking retry\n","stream":"stderr","time":"2017-02-21T23:44:21.023180108Z"}  
{"log":"2017/02/21 23:44:21 it is not replayable\n","stream":"stderr","time":"2017-02-21T23:44:21.023184712Z"}  
{"log":"2017/02/21 23:44:21 setting request canceler\n","stream":"stderr","time":"2017-02-21T23:44:21.023188614Z"}  
{"log":"2017/02/21 23:44:21 sending error from read loop\n","stream":"stderr","time":"2017-02-21T23:44:21.023192675Z"}  
{"log":"2017/02/21 23:44:21 from resc, mapping from read loop true\n","stream":"stderr","time":"2017-02-21T23:44:21.023894047Z"}  
{"log":"2017/02/21 23:44:21 setting canceler to nil\n","stream":"stderr","time":"2017-02-21T23:44:21.023908607Z"}  
{"log":"2017/02/21 23:44:21 pconn roundtrip return true\n","stream":"stderr","time":"2017-02-21T23:44:21.023914256Z"}  
{"log":"2017/02/21 23:44:21 checking retry\n","stream":"stderr","time":"2017-02-21T23:44:21.023918574Z"}  
{"log":"2017/02/21 23:44:21 it is not replayable\n","stream":"stderr","time":"2017-02-21T23:44:21.023922478Z"}  
{"log":"2017/02/21 23:44:21 not retrying\n","stream":"stderr","time":"2017-02-21T23:44:21.023926653Z"}  
{"log":"2017/02/21 23:44:21 unpacking error\n","stream":"stderr","time":"2017-02-21T23:44:21.023930617Z"}  
{"log":"2017/02/21 23:44:21 error from roundtrip\n","stream":"stderr","time":"2017-02-21T23:44:21.023934724Z"}  
```

mo-gr:
👍

aryszka:
as for reference:

```
    // Body is the request's body.
    //
    // For client requests a nil body means the request has no
    // body, such as a GET request. The HTTP Client's Transport
    // is responsible for calling the Close method.
    //
    // For server requests the Request Body is always non-nil
    // but will return EOF immediately when no body is present.
    // The Server will close the request body. The ServeHTTP
    // Handler does not need to.
    Body io.ReadCloser
```

and

```
    // ContentLength records the length of the associated content.
    // The value -1 indicates that the length is unknown.
    // Values >= 0 indicate that the given number of bytes may
    // be read from Body.
    // For client requests, a value of 0 with a non-nil Body is
    // also treated as unknown.
    ContentLength int64
```

Raffo:
Great! 👍

hjacobs:
:+1:

PR #280, opened 2017-02-23
automatically create docker images on releases

aryszka:
start creating docker images on every release automatically


szuecs:
:+1:

PR #285, opened 2017-02-28
try to fix ci build

aryszka:
clean fails on ci when an already clean build starts


aryszka:
not enough (pierone login failing)

PR #288, opened 2017-03-03
Support HTTPS requests in the Docker image

shuhei:
Added certificates to `Dockerfile` to support HTTPS requests. Without it, I got errors like `x509: failed to load system roots and no roots provided`.

Also:

- Updated env variables in the packaging README
- Fixed error on the first docker build

### DIFF
--- a/packaging/Dockerfile
+++ b/packaging/Dockerfile
@@ -1,5 +1,6 @@
 FROM alpine
 MAINTAINER Skipper Maintainers <team-pathfinder@zalando.de>
+RUN apk --no-cache add ca-certificates && update-ca-certificates
 RUN mkdir -p /usr/bin
 ADD skipper eskip /usr/bin/
 ENV PATH $PATH:/usr/bin
--- a/packaging/Makefile
+++ b/packaging/Makefile
@@ -13,7 +13,7 @@ eskip:
 	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build github.com/zalando/skipper/cmd/eskip
 
 clean:
-	rm skipper eskip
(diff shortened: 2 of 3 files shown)


hjacobs:
Thanks for the PR!

lmineiro:
We've been mounting the host cert dir and this is how I, personally, would like to keep it.

aryszka:
question: does having defaults prevent mounting custom ones?

shuhei:
@lmineiro Ah, it's a good idea. But because <q>the primary use case for this docker image is to have a quick'n'dirty skipper available quick</q> according to [the packaging README](https://github.com/zalando/skipper/blob/master/packaging/readme.md), it would be great if it has default ones or documentation at least.

@Imineiro @aryszka What do you think?

lmineiro:
@shuhei good point on the quick'n'dirty promise.
@aryszka it won't stop us from mounting but we also shouldn't really encourage using such Docker image, out of the box, for production

ideahitme:
are we using zalando alpine image with preinstalled ca-certificates for skipper releases ? then you might just change Docker base image to registry.opensource.zalan.do/alpine and keep the same docker image for internal and external users ?

mikkeloscar:
I think the certs should be included in the image.

One use case is running it on Kubernetes were the pod should be self-contained. You don't want to rely on where the node stores certificates, and you don't want to get `hostPath` privileges if it's not needed.

aryszka:
sounds like a legit argument to me

aryszka:
:+1:

mikkeloscar:
:+1:

PR #290, opened 2017-03-06
Get rid of kubectl proxy in skipper

ideahitme:
1. Add kubernetes-in-cluster flag to specify if the skipper is deployed as a pod vs running locally
2. If in-cluster use service account to authenticate with API server
3. Otherwise fall back to the current implementation 

JUST started, and I wanted to add a do-not-merge label, but apparently I don't have access to that. Will ping you once done

### DIFF
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -42,7 +42,9 @@ const (
 	addressUsage                   = "network address that skipper should listen on"
 	etcdUrlsUsage                  = "urls of nodes in an etcd cluster, storing route definitions"
 	etcdPrefixUsage                = "path prefix for skipper related data in etcd"
-	kubernetesURLUsage             = "kubernetes API base url for the ingress data client; when set, it enables ingress"
+	kubernetesUsage                = "enables skipper to generate routes for ingress resources in kubernetes cluster"
+	kubernetesInClusterUsage       = "specify if skipper is running inside kubernetes cluster"
+	kubernetesURLUsage             = "kubernetes API base url for the ingress data client; requires kubectl proxy enabled; omit if kubernetes-in-cluster is set to true"
 	kubernetesHealthcheckUsage     = "automatic healthcheck route for internal IPs with path /kube-system/healthz; valid only with kubernetes-url"
 	innkeeperUrlUsage              = "API endpoint of the Innkeeper service, storing route definitions"
 	innkeeperAuthTokenUsage        = "fixed token for innkeeper authentication"
@@ -90,6 +92,8 @@ var (
 	proxyPreserveHost         bool
 	idleConnsPerHost          int
 	closeIdleConnsPeriod      string
+	kubernetes                bool
+	kubernetesInCluster       bool
 	kubernetesURL             string
 	kubernetesHealthcheck     bool
 	innkeeperUrl              string
@@ -129,6 +133,8 @@ func init() {
 	flag.IntVar(&idleConnsPerHost, "idle-conns-num", proxy.DefaultIdleConnsPerHost, idleConnsPerHostUsage)
 	flag.StringVar(&closeIdleConnsPeriod, "close-idle-conns-period", strconv.Itoa(int(proxy.DefaultCloseIdleConnsPeriod/time.Second)), closeIdleConnsPeriodUsage)
 	flag.StringVar(&etcdPrefix, "etcd-prefix", defaultEtcdPrefix, etcdPrefixUsage)
+	flag.BoolVar(&kubernetes, "kubernetes", false, kubernetesUsage)
+	flag.BoolVar(&kubernetesHealthcheckUsage, "kubernetes-in-cluster", false, kubernetesInCluster)
 	flag.StringVar(&kubernetesURL, "kubernetes-url", "", kubernetesURLUsage)
 	flag.BoolVar(&kubernetesHealthcheck, "kubernetes-healthcheck", true, kubernetesHealthcheckUsage)
 	flag.StringVar(&innkeeperUrl, "innkeeper-url", "", innkeeperUrlUsage)
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -44,7 +44,7 @@ const (
 	etcdPrefixUsage                = "path prefix for skipper related data in etcd"
 	kubernetesUsage                = "enables skipper to generate routes for ingress resources in kubernetes cluster"
 	kubernetesInClusterUsage       = "specify if skipper is running inside kubernetes cluster"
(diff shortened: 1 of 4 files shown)


cmd/skipper/main.go
```diff
@@ -129,6 +133,8 @@ func init() {
 	flag.IntVar(&idleConnsPerHost, "idle-conns-num", proxy.DefaultIdleConnsPerHost, idleConnsPerHostUsage)
 	flag.StringVar(&closeIdleConnsPeriod, "close-idle-conns-period", strconv.Itoa(int(proxy.DefaultCloseIdleConnsPeriod/time.Second)), closeIdleConnsPeriodUsage)
 	flag.StringVar(&etcdPrefix, "etcd-prefix", defaultEtcdPrefix, etcdPrefixUsage)
+	flag.BoolVar(&kubernetes, "kubernetes", false, kubernetesUsage)
+	flag.BoolVar(&kubernetesHealthcheckUsage, "kubernetes-in-cluster", false, kubernetesInCluster)
```
hjacobs: references wrong const..

ideahitme:
seems to work fine in the kubernetes cluster, just need to fix the tests and ready for review

ideahitme:
will add unit tests, but can be reviewed now

cmd/skipper/main.go
```diff
@@ -42,7 +42,9 @@ const (
 	addressUsage                   = "network address that skipper should listen on"
 	etcdUrlsUsage                  = "urls of nodes in an etcd cluster, storing route definitions"
 	etcdPrefixUsage                = "path prefix for skipper related data in etcd"
-	kubernetesURLUsage             = "kubernetes API base url for the ingress data client; when set, it enables ingress"
+	kubernetesUsage                = "enables skipper to generate routes for ingress resources in kubernetes cluster"
+	kubernetesInClusterUsage       = "specify if skipper is running inside kubernetes cluster"
+	kubernetesURLUsage             = "kubernetes API base url for the ingress data client; requires kubectl proxy enabled; omit if kubernetes-in-cluster is set to true"
```
hjacobs: nitpick: "url" -> "URL"

dataclients/kubernetes/kube.go
```diff
@@ -79,37 +92,130 @@ type Options struct {
...
 
 var nonWord = regexp.MustCompile("\\W")
 
 var (
-	errServiceNotFound     = errors.New("service not found")
-	errServicePortNotFound = errors.New("service port not found")
+	errServiceNotFound      = errors.New("service not found")
+	errServicePortNotFound  = errors.New("service port not found")
+	errAPIServerURLNotFound = errors.New("kubernetes API server URL could not be constructed from env vars")
+	errInvalidCertificate   = errors.New("discovered certificate is invalid")
```
szuecs: The failure message is missleading. Why it is "discovered"? I would suggest something like "invalid CA".

dataclients/kubernetes/kube.go
```diff
@@ -79,37 +92,130 @@ type Options struct {
...
+func buildAPIURL(o Options) (string, error) {
+	if !o.KubernetesInCluster {
+		if o.KubernetesURL == "" {
+			return defaultKubernetesURL, nil
+		}
+		return o.KubernetesURL, nil
+	}
+
+	host, port := os.Getenv(serviceHostEnvVar), os.Getenv(servicePortEnvVar)
+	if len(host) == 0 || len(port) == 0 {
```
hjacobs: why is `len` used here instead of comparing with empty string?

dataclients/kubernetes/kube.go
```diff
@@ -79,37 +92,130 @@ type Options struct {
...
+}
+
+func (c *Client) createRequest(method, url string, body io.Reader) (*http.Request, error) {
+	req, err := http.NewRequest(method, url, body)
+	if err != nil {
+		return nil, err
+	}
+
+	if c.token != "" {
+		req.Header.Set("Authorization", "Bearer "+c.token)
```
hjacobs: go fmt missing? no spaces around `+`..

skipper.go
```diff
@@ -53,8 +53,14 @@ type Options struct {
...
-	// Kubernetes API base URL for ingress. If not specififed,
-	// Kubernetes ingress is disabled.
+	// If set enables skipper to generate based on ingress resources in kubernetes cluster
+	Kubernetes bool
+
+	// If set makes skipper to authenticate with the kubernetes API server with service account assigned to the skipper POD
+	// If omitted skipper will rely on kubectl proxy to authenticate with API server
+	KubernetesInCluster bool
+
+	// Kubernetes API base URL. Only makes sense if KubernetesInCluster is set to false. If omitted and skipper not deployed to the cluster, the default API URL will be used
```
hjacobs: nitpick: "If omitted and skipper is not deployed to the cluster, .." (missing "is")

hjacobs:
Overall the logic looks fine to me.

szuecs:
Looks also good to me

dataclients/kubernetes/kube.go
ideahitme: my editor auto-collapses spaces in this case, so it seems valid with gofmt :D

hjacobs:
:+1:

dataclients/kubernetes/kube.go
aryszka: looks, ok. You can run `make fmt` to apply the same formatting as the ci expects, or run `make check-fmt` to just check if it is the same as what the ci expects. (The task runs `gofmt -w`, which has probably the same settings as most editor defaults, but worth checking before push.)

dataclients/kubernetes/kube.go
```diff
@@ -79,37 +92,130 @@ type Options struct {
...
+
+func buildAPIURL(o Options) (string, error) {
+	if !o.KubernetesInCluster {
+		if o.KubernetesURL == "" {
+			return defaultKubernetesURL, nil
+		}
+		return o.KubernetesURL, nil
+	}
+
+	host, port := os.Getenv(serviceHostEnvVar), os.Getenv(servicePortEnvVar)
```
aryszka: woudln't it be simpler to have this as a single env var? the host and the port in one?

aryszka: maybe calling it KUBERNETES_SERVICE or KUBERNETES_SERVICE_ADDRESS

aryszka:
nice! Minor thing i would consider if it is not better to put the host and the port into a single env var.

aryszka:
:+1:

dataclients/kubernetes/kube.go
ideahitme: https://github.com/kubernetes/kubernetes/blob/1a46a167f3c1baecb7e0e6a84137330982861ec2/staging/src/k8s.io/client-go/rest/config.go#L278 
the env vars are injected by kubernetes, i just followed the way it is done in the client-go :)

PR #292, opened 2017-03-06
patch for stdlib profile path handling

aryszka:


mo-gr:
:+1:

mo-gr:
but why are you not just using /debug/?

mo-gr:
:+1:

Issue #287, opened 2017-03-03
tee() filter - context closed


predragstefanovic:
Hi, we would like to use tee() filter, but at the moment we are encountering an error on mirroring requests. 
Here is the route example we use on ```https://main_host``` gateway: 
```
Path("/mirror") && Method("GET")
  -> tee("https://host_to_receive_mirrored_traffic", "^/.*", "/mirrorEndpoint" )
  -> setPath("/reset-path")
  -> "https://main_host";

catch_all: *
  -> status(401)
  -> <shunt>;
```
After requesting: ```GET /mirror``` we get 401 as expected but there is no traffic on ```https://host_to_receive_mirrored_traffic```. Related skipper logs:
```
[03/Mar/2017:13:33:52 +0000]  "GET /mirror HTTP/1.1" 401 0 "" "" 15 main_host
[APP]time="2017-03-03T13:33:52Z" level=warning msg="error while tee requestGet https://host_to_receive_mirrored_traffic/mirrorEndpoint: context canceled" 
```

PR #294, opened 2017-03-08
avoid copying unknown fields for the tee() request

aryszka:
This PR changes how the requests for the tee() filter are cloned. It replaces value based shallow copy with explicitly setting each known field in the clone.

Closes: #287

### DIFF
--- a/filters/tee/tee.go
+++ b/filters/tee/tee.go
@@ -70,7 +70,7 @@ func (tt *teeTie) Read(b []byte) (int, error) {
 
 	if n > 0 {
 		if _, werr := tt.w.Write(b[:n]); werr != nil {
-			log.Error("error while tee request", werr)
+			log.Error("tee: error while tee request", werr)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


filters/tee/tee.go
```diff
@@ -88,40 +88,55 @@ func (r *tee) Response(filters.FilterContext) {}
...
 		if err != nil {
-			log.Warn("error while tee request", err)
-		}
-		if err == nil {
-			defer rsp.Body.Close()
+			log.Warn("tee: error while tee request", err)
+			return
 		}
+
+		defer rsp.Body.Close()
```
aryszka: defer: true, fixing

ioutil.Discard: how to do it?
as i understand we're on the client side in this case, and we'll receive an *http.Response

aryszka: defer: true, fixing

ioutil.Discard: how to do it?
as i understand we're on the client side in this case, and we'll receive an *http.Response

PR #299, opened 2017-03-13
packaging in a clean directory does not work

szuecs:
We use the packaging Makefile in our Jenkins task to build and we have a clean directory, so this will break.

    rm skipper eskip
    rm: cannot remove 'skipper': No such file or directory
    rm: cannot remove 'eskip': No such file or directory
    Makefile:16: recipe for target 'clean' failed
    make: *** [clean] Error 1


aryszka:
:+1:

PR #300, opened 2017-03-15
tee filter fixes

aryszka:
- fix forwarding headers
- fix teeing requests with no body
- sync tests

### DIFF
--- a/filters/tee/tee_test.go
+++ b/filters/tee/tee_test.go
@@ -21,9 +21,12 @@ var (
 	teeArgsWithModPath = []interface{}{"https://api.example.com", ".*", "/v1/"}
 )
 
-type MyHandler struct {
-	name string
-	body string
+type myHandler struct {
+	t      *testing.T
+	name   string
+	header http.Header
+	body   string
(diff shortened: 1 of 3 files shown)


filters/tee/tee_test.go
```diff
@@ -136,17 +152,76 @@ func TestTeeEndToEndBody(t *testing.T) {
...
 
 	testingStr := "TESTEST"
-	req, _ := http.NewRequest("GET", p.URL, strings.NewReader(testingStr))
+	req, err := http.NewRequest("GET", p.URL, strings.NewReader(testingStr))
+	if err != nil {
+		t.Error(err)
+	}
+
 	req.Host = "www.example.org"
-	req.Header.Set("X-Test", "true")
```
szuecs: Why do you drop this header from the test?

szuecs:
:+1:

filters/tee/tee_test.go
aryszka: it didn't hurt there but there was no assertion for it. I wrote a new test case for checking the headers.

PR #302, opened 2017-03-21
Wire backend host metrics

aryszka:


mo-gr:
👍

hjacobs:
What does it do? We really need to document all the different metrics options, I have no clue what to use for what..

aryszka:
I agree with improving on the doc.

This particular type of metrics does the same as the per-route or per-host metrics but for backend hosts. It collects the time to serve a request separately for each backend, method and status.

aryszka:
@hjacobs the PR that added the metrics, but didn't wire it into the global options:

https://github.com/zalando/skipper/pull/245

PR #305, opened 2017-04-05
Add '^' to start of path when getting them from kubernetes

jchanam:
We've found a bug regarding the paths when reading from kubernetes ingresses.

Before, the following rule was create:

- Host: test.com Path: /test

With this, any of the following URIs matched:

- test.com/test
- test.com/foo/test
- test.com/foo/bar/test

Now, with this fix, only paths starting with the path defined on the ingress match.

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -341,7 +341,7 @@ func (c *Client) convertPathRule(ns, name, host string, prule *pathRule) (*eskip
 
 	var pathExpressions []string
 	if prule.Path != "" {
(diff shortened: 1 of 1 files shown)


aryszka:
:+1:

lmineiro:
shipit

lmineiro:
👍

aryszka:
@jchanam makes sense, thanks a lot

hjacobs:
@aryszka @lmineiro PR for https://github.com/zalando-incubator/kubernetes-on-aws please :smirk:

Issue #298, opened 2017-03-09
Kubernetes: handle sigterm to support graceful shutdown in daemonsets


aryszka:
The scenario would look like this:

1. handle SIGTERM
2. when received, disable the healthcheck but continue serving
3. get killed by the kubelet

@hjacobs please add the required details

hjacobs:
That sounds about right.

szuecs:
closed in favour of #307

aryszka:
done in #307

PR #306, opened 2017-04-05
add sigterm handler to change the return value for kubernetes

szuecs:
add sigterm handler to change the return value for /kube-system/healthz from 200 to 503 to signal unhealthy to ALB, if got a SIGTERM

#298

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -30,8 +30,10 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"os/signal"
 	"regexp"
 	"strings"
+	"syscall"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -109,6 +111,7 @@ type Client struct {
 	provideHTTPSRedirect bool
 	token                string
 	current              map[string]*eskip.Route
+	healthy              bool
 }
 
 var nonWord = regexp.MustCompile("\\W")
@@ -146,6 +149,7 @@ func New(o Options) (*Client, error) {
 		provideHTTPSRedirect: o.ProvideHTTPSRedirect,
 		current:              make(map[string]*eskip.Route),
 		token:                token,
+		healthy:              true,
 	}, nil
 }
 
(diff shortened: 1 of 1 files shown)


hjacobs:
:+1:

dataclients/kubernetes/kube.go
```diff
@@ -498,9 +502,21 @@ func (c *Client) LoadAll() ([]*eskip.Route, error) {
...
 }
 
+func (c *Client) registerSigtermHandler() {
+	log.Info("register sigterm handler")
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGTERM)
+	// Block until a signal is received.
+	s := <-sigs
+	log.Infof("shutdown, caused by %s, set healthCheck to be unhealty", s)
+	c.healthy = false
```
aryszka: writing here causes a race condition because healthy is accessed from multiple goroutines. Since neither is not a performance critical path, a simple Mutex would be enough here.

aryszka: data race to be precise

dataclients/kubernetes/kube.go
```diff
@@ -498,9 +502,21 @@ func (c *Client) LoadAll() ([]*eskip.Route, error) {
 	c.current = mapRoutes(r)
 	log.Debugf("all routes loaded and mapped")
 
+	go c.registerSigtermHandler()
```
aryszka: this will spawn a go routine for every test that never return. We should try to make the data client cleanup its resources (this is the first such resource). I have sg on my mind how, making a suggestion in a sub-branch.

aryszka: of course, this is true not only for the tests :)

when someone uses skipper as an embedded library, may need to have control over the cleanup

aryszka: even better.. with a select, we don't even need the goroutine here. Then no need for the cleanup either. Pushing it in a bit.

dataclients/kubernetes/kube.go
aryszka: see my suggested changes to this branch for how to come around the data race

dataclients/kubernetes/kube.go
aryszka: see my suggested changes to this branch for how to come around the data race

Issue #237, opened 2016-12-15
travis build fails sometimes due to test calling out start an etcd process


aryszka:

ideahitme:
https://travis-ci.org/zalando/skipper/builds/208397133 - failing 
retry build - 
https://travis-ci.org/zalando/skipper/builds/208437789 - success

aryszka:
thanks! the solution i have on mind is to mock away etcd

szuecs:
We stopped testing etcd on Travis so this can be closed, not?

szuecs:
etcd is not tested in travis, so close

PR #311, opened 2017-04-18
Implement caching for kubernetes service endpoints

jchanam:
When an ingress has a multiple routes, a request to the K8S API is made for each route and endpoint.

If the endpoint is the same, we can cache the response and there's no need to ask again. On the next iteration, this cache will be invalidated, so if the service changes, we'll be able to notice it.

This is very helpful in ingresses with many routes (we use 50+ in one real-world example) as the repeated calls to K8S API are slow and congest traffic on the whole. Also, each request generates a log entry, which can result in huge log files.

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -349,14 +349,25 @@ func (c *Client) convertDefaultBackend(i *ingressItem) (*eskip.Route, bool, erro
 	return r, true, nil
 }
 
-func (c *Client) convertPathRule(ns, name, host string, prule *pathRule) (*eskip.Route, error) {
+func (c *Client) convertPathRule(ns, name, host string, prule *pathRule, servicesURLs map[string]string) (*eskip.Route, map[string]string, error) {
 	if prule.Backend == nil {
-		return nil, fmt.Errorf("invalid path rule, missing backend in: %s/%s/%s", ns, name, host)
+		return nil, nil, fmt.Errorf("invalid path rule, missing backend in: %s/%s/%s", ns, name, host)
 	}
-
-	address, err := c.getServiceURL(ns, prule.Backend.ServiceName, prule.Backend.ServicePort)
-	if err != nil {
-		return nil, err
+	serviceKey := ns + prule.Backend.ServiceName + prule.Backend.ServicePort.String()
+	var (
+		address string
+		err     error
+	)
+	if val, ok := servicesURLs[serviceKey]; !ok {
+		address, err = c.getServiceURL(ns, prule.Backend.ServiceName, prule.Backend.ServicePort)
+		if err != nil {
+			return nil, nil, err
+		}
+		servicesURLs[serviceKey] = address
+		log.Debugf("New route for %s/%s/%s", ns, prule.Backend.ServiceName, prule.Backend.ServicePort)
+	} else {
+		address = val
+		log.Debugf("Route for %s/%s/%s already known", ns, prule.Backend.ServiceName, prule.Backend.ServicePort)
 	}
 
 	var pathExpressions []string
@@ -370,7 +381,7 @@ func (c *Client) convertPathRule(ns, name, host string, prule *pathRule) (*eskip
 		Backend:     address,
 	}
 
-	return r, nil
+	return r, servicesURLs, nil
 }
 
 // logs if invalid, but proceeds with the valid ones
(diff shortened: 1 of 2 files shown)


hjacobs:
Thanks for the PR!

hjacobs:
There seems to be some problem (https://travis-ci.org/zalando/skipper/builds/223222447):

```
panic: assignment to entry in nil map [recovered]
	panic: assignment to entry in nil map
```

dataclients/kubernetes/kube.go
```diff
@@ -405,7 +422,7 @@ func (c *Client) ingressToRoutes(items []*ingressItem) []*eskip.Route {
 			host := []string{"^" + strings.Replace(rule.Host, ".", "[.]", -1) + "$"}
 
 			for _, prule := range rule.Http.Paths {
-				r, err := c.convertPathRule(i.Metadata.Namespace, i.Metadata.Name, rule.Host, prule)
+				r, servicesURLs, err = c.convertPathRule(i.Metadata.Namespace, i.Metadata.Name, rule.Host, prule, servicesURLs)
```
lmineiro: It is enough to pass the serviceURLs map it in the argument since map types are reference types. The additions inside the method will be carried over between calls.

lmineiro: A nice way to confirm and, actually desired, is to write a test for the new behavior of the convertPathRule() function.

jchanam:
Hi!

I've done the changed you requested. Thanks to @lmineiro and @szuecs for your help and suggestions.

hjacobs:
@lmineiro @aryszka can we please fix the Travis CI build to not always fail because of missing `etcd` binary?

hjacobs:
:+1:

hjacobs:
@lmineiro btw, you can expect anybody to write tests if your Travis CI build always fails :smirk:

lmineiro:
@hjacobs we still couldn't find the time to fix https://github.com/zalando/skipper/issues/237
A PR would be highly appreciated

lmineiro:
@jchanam wan't to give it a shot writing a test for this caching behavior?

szuecs:
@jchanam Please create a new test function "TestConvertPathRule" in https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/kube_test.go

jchanam:
I've tried to add a test. I'm not sure if it's correct or not, but I'm having a PASS :dancer: 

Could you tell me if it's ok?

dataclients/kubernetes/kube.go
```diff
@@ -349,14 +349,25 @@ func (c *Client) convertDefaultBackend(i *ingressItem) (*eskip.Route, bool, erro
...
-func (c *Client) convertPathRule(ns, name, host string, prule *pathRule) (*eskip.Route, error) {
+func (c *Client) convertPathRule(ns, name, host string, prule *pathRule, servicesURLs map[string]string) (*eskip.Route, error) {
 	if prule.Backend == nil {
 		return nil, fmt.Errorf("invalid path rule, missing backend in: %s/%s/%s", ns, name, host)
 	}
-
-	address, err := c.getServiceURL(ns, prule.Backend.ServiceName, prule.Backend.ServicePort)
-	if err != nil {
-		return nil, err
+	serviceKey := ns + prule.Backend.ServiceName + prule.Backend.ServicePort.String()
```
aryszka: it can reduce the chance for naming conflicts if we apply some formatting for the key, e.g:

`fmt.Sprintf("__srv__%s__%s__%v", ns, prule.Backend.ServiceName, prule.Backend.ServicePort)`

dataclients/kubernetes/kube_test.go
```diff
@@ -828,6 +828,61 @@ func TestIngress(t *testing.T) {
 	})
 }
 
+func TestConvertPathRule(t *testing.T) {
```
aryszka: this tests for regression. Wondering if there was a nice way to test the improvement in the change set. E.g extending the test api handler with counter for each path, and check that the counter values are as expected.

dataclients/kubernetes/kube.go
jchanam: Why that can reduce conflicts? With the same values, it will generate the same key. I mean, if this key:
`namespaceservice80`
have a conflict, the key:
`__srv__namespace__service__80`
will hace the exact same probability of collisions.

aryszka: i was thinking about cases where we have a) namespace=foo,service=barbaz and b) namespace=foobar,service=baz . I know it doesn't eliminate the conflicts completely, only reduces the chance. For that we'd need to escape somehow each occurence of __ in the components.

I just checked what we do with the routeIDs, and see now that those are also not completely right, because there the '_' occurences are simply deleted. This has the potential of some rare but annoying bugs.

We can fix this for both the service url keys and route IDs as a separate issue, so I am not insisting on doing it here.

aryszka:
:+1:

lmineiro:
👍

aryszka:
merging now and fixing the build separately as it is broken due to this PR

PR #313, opened 2017-04-19
Install etcd in Travis

ese:


hack/build_etcd.sh
```diff
@@ -0,0 +1,20 @@
...
+echo "Using ETCD version $ETCD_VERSION"
+
+git clone https://github.com/coreos/etcd.git hack/etcd
+cd hack/etcd
+git checkout $ETCD_VERSION
+./build
+
+${TRAVIS:?"This is not a Travis build. All Done"}
+#Temporal solution to travis issue #155
+sudo rm -rf /dev/shm && sudo ln -s /run/shm /dev/shm
```
aryszka: is this to get more shared memory?

aryszka:
@ese thanks for the PR

aryszka:
:+1:

hack/build_etcd.sh
```diff
@@ -0,0 +1,16 @@
...
+if [  $# -gt 0 ]
+    then
+    ETCD_VERSION="$1";
+    else
+    ETCD_VERSION="master";
+fi
+
+echo "Using ETCD version $ETCD_VERSION"
+
+mkdir /hack
```
aryszka: for me this fails. Shouldn't the `hack` directory be created inside the workdir instead of the root?

ese: I don't know where place etcd code becouse if I put it in the working directory then `go build ./...` invoked by Makefile fails due it try to build etcd

ese:
Ok its green now. There is another installation of etcd in `etcd/install.sh` invoked by makefile. I don't know if you want to mantain it or merge with this approach

aryszka:
@ese thanks for this. If you don't mind, I would actually prefer the solution with the existing install script: https://github.com/zalando/skipper/pull/314

PR #314, opened 2017-04-19
Etcd install script

aryszka:
- chaning the etcd install script to allow installation of a fixed version
- making it fail on the first error


hjacobs:
:+1:

PR #316, opened 2017-04-19
silence the command expression for the format check in the Makefile

aryszka:
silence the format check in the Makefile to avoid polluting the build output with all the source files


lmineiro:
👍

PR #317, opened 2017-04-20
Fix escaping when unmarshal json

jchanam:
We have found an issue when reading from kubernetes paths rules.

When having a regular expression on an ingress rule, you have to write it escaped (the backslashes):
```
- backend:
    serviceName: my-regex-backend
    servicePort: 80
  path: /regex-path/(\\d+)(/)?$
```

Note that here we have a double backslash.

Now, when skipper reads it from the api (on a json format) and unmarshal the content, it does it as a string LITERAL. This means that go reads it char by char and put it in a string, not converting the escaped double backslash into a single one (which would be correct inside go to check on a regex).

Because of this, skipper is not able to match a real URL with the path defined into the ingress.

I've did a hack, to unquote this string and convert the double backslash into a single one.

I now this is not the best solution, and it should it be done directly during the unmarshal, but I don't know how to do it.

Could you tell me if this is correct (and merge it), or give me a better solution to fix this bug.

Thanks!

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -373,11 +373,11 @@ func (c *Client) convertPathRule(ns, name, host string, prule *pathRule, service
 
 	var pathExpressions []string
 	if prule.Path != "" {
-		res, err := strconv.Unquote(`"` + prule.Path + `"`)
+		unquotedPath, err := strconv.Unquote(`"` + prule.Path + `"`)
 		if err != nil {
 			pathExpressions = []string{"^" + prule.Path}
 		} else {
-			pathExpressions = []string{"^" + res}
+			pathExpressions = []string{"^" + unquotedPath}
 		}
 	}
 
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -40,6 +40,7 @@ import (
 	"os"
 	"os/signal"
 	"regexp"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


aryszka:
Hi @jchanam ,

thanks for this PR. Are you confident that you need to write the regular expression this way:

```
path: /regex-path/(\\d+)(/)?$
```

?

I would have expected that it would be either, with a single backslash:

```
path: /regex-path/(\d+)(/)?$
```

or quoted:

```
path: "/regex-path/(\\d+)(/)?$"
```

i checked this:

http://yaml.org/spec/1.2/spec.html#id2786942

and to me it seems like this way of unqouting would be wrong, because it is possible that it would unqoute things that it shouldn't. I believe that the safe way would be to use the quoted string format for the regexp in the ingress rule yaml. Am I missing something?

i try it out in a bit...

aryszka:
@jchanam some update. I think we should not merge this. The reason is that the Ingress API accepts IEEE Std 1003.1 regular expressions for the path ("Path is an extended POSIX regex as defined by IEEE Std 1003.1, (i.e this follows the egrep/unix syntax, not the perl syntax)", https://kubernetes.io/docs/api-reference/v1.6/#httpingresspath-v1beta1-extensions). This standard doesn't know the \d shortcut for decimal characters. Now what happens in your example, using the double quotes this way means as if you wrote "/regex-path/(\\\\d+)(/)?$" and expecting an actual backslash and a 'd' character in the path to match.

what you can do instead is to write [0-9] instead of \d in the ingress rule. (And I still recommend to use the quoted format in the yaml for more complex regular expressions.)

jchanam:
@aryszka You're right.

Thanks!

Issue #271, opened 2017-02-13
improve the tee() filter


aryszka:
Let's use this entry to collect additional requirements to the tee() filter and consider them together to improve it with keeping backwards compatibility.

aryszka:
Improvement: allow the user to decide whether the branched requests should follow redirects or not.

szuecs:
@aryszka I close this because I do not understand what other options are required. We have now tee() and teenf() to have both supported, or do I miss something?
If so please reopen.

PR #318, opened 2017-05-05
tee filter does not follow redirects if user prefers so

haldun:
If tee filter is created with the name `teenf`, it won't follow redirects.

### DIFF
--- a/filters/builtin/builtin.go
+++ b/filters/builtin/builtin.go
@@ -77,6 +77,7 @@ func MakeRegistry() filters.Registry {
 		diag.NewBackendChunks(),
 		tee.NewTee(),
 		tee.NewTeeDeprecated(),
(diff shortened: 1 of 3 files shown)


hjacobs:
Thanks for the PR! :smile: 

@aryszka @lmineiro does this make sense to have?

aryszka:
@hjacobs , i think yes. I think now that the tee() filter shouldn't follow redirects by default, but we missed this aspect with the first version.

Does anyone have a good suggestion regarding the naming or using an argument for this behavior? The problem with argument would be that the tee() filter already has two optional arguments.

aryszka:
related to https://github.com/zalando/skipper/issues/271

aryszka:
lgtm

PR #323, opened 2017-05-10
remove failing docker build

aryszka:
removing failing docker builds until figuring a way to authenticate to a docker hub


NorthFury:
:+1:

Issue #274, opened 2017-02-17
Backend type: loopback


mo-gr:
Currently, skipper supports two types of backends: shunts and hosts. I suggest adding a third type of backend named `<loopback>` (name open for discussion).

What this backend type would do, is re-execute the whole routing and filterchain a second time for the given request. This of course only makes sense if the filters leading up to this endpoint have modified the request. Otherwise, the endpoint must throw an error to avoid internal redirect loops.

## Motivation

We observed quite a few use-cases, that are hard or impossible with the current skipper but could be solved elegantly with multiple cascading skippers. However, cascading skippers add unnecessary network hops. This feature could therefore be seen as an optimization to situations, where multiple skippers are cascaded behind each other. 
However, it is not as strong as cascading skippers. The loopback is not a complete substitute for cascading skippers, but it covers the majority of use-cases in an elegant and performant way.

## Example

Consider the use-case where the routing decision between two similar routes is based on the response of a remote service (for example an A/B test). Naturally, we do not want to call services for all routing decisions, as this would add cost to _all_ routes, not just the two routes in question.

But having a cascading setup of skipper would allow us to first isolate the two routes, then call the service to make the decisiosn, then pass on to the second skipper to actually handle the two routes. With the loopback, this could all happen within the same skipper. In eskip, this would look like the following:

```
pageUnderTest: Path("/foo")
  -> callABTestService() // this will call a service and add a header
  -> <loopback>;

pageVariantA: Path("/foo") && Header("X-Test-Variant", "A")
  -> filtersForVariantA()
  -> "https://a.example.org";

pageVariangB: Path("/foo") && Header("X-Test-Variant", "B")
  -> filtersForVariangB()
  -> "https://b.example.org";
```

Note: The filter `callABTestService()` must ensure that the incoming request gets the header `X-Test-Variant` appended to it in any case. If the filter would fail to add the header, the request would loop back in to the route `pageUnderTest`.

## Specification

- The `loopback` endpoint must reevaluate the routing and filter chain for the previously modified (by the filters) request.
- The `loopback` endpoint must be able to detect infinite loops and abort the request with an appropriate error (either invalid request or server error?)
- Skipper should support an arbitrary number of loopbacks as long as they are not looping

## Open Questions
- Should the loopback reuse the same `StateBag`/`RequestContext` or generate a new one? My gut feeling says reuse the `StateBag`, but this could violate assumptions made by other filters.
- [EDIT] What would be the expected behavior when collecting metrics for a request?

[...]

PR #327, opened 2017-05-15
Proxy refactor

aryszka:
refactoring proxy.go code for better maintainability and in preparation for the loopback backend type (#274)

### DIFF
--- a/proxy/context.go
+++ b/proxy/context.go
@@ -1,0 +1,204 @@
+package proxy
+
+import (
+	"bytes"
+	"io"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/zalando/skipper/routing"
+)
+
+// a byte buffer implementing the Closer interface
+type bodyBuffer struct {
+	*bytes.Buffer
+}
+
+type context struct {
+	responseWriter        http.ResponseWriter
+	request               *http.Request
+	response              *http.Response
+	route                 *routing.Route
+	deprecatedServed      bool
+	servedWithResponse    bool // to support the deprecated way independently
+	pathParams            map[string]string
+	stateBag              map[string]interface{}
+	originalRequest       *http.Request
+	originalResponse      *http.Response
+	outgoingHost          string
+	debugFilterPanics     []interface{}
+	outgoingDebugRequest  *http.Request
+	incomingDebugResponse *http.Response
+	loopCounter           int
+	startServe            time.Time
+}
+
+func defaultBody() io.ReadCloser {
+	return &bodyBuffer{&bytes.Buffer{}}
+}
+
+func defaultResponse(r *http.Request) *http.Response {
+	return &http.Response{
+		StatusCode: http.StatusNotFound,
+		Header:     make(http.Header),
+		Body:       defaultBody(),
+		Request:    r}
+}
+
+func (sb bodyBuffer) Close() error {
+	return nil
+}
+
+func cloneUrl(u *url.URL) *url.URL {
+	uc := *u
+	return &uc
+}
+
+func cloneRequestMetadata(r *http.Request) *http.Request {
+	return &http.Request{
+		Method:           r.Method,
+		URL:              cloneUrl(r.URL),
+		Proto:            r.Proto,
+		ProtoMajor:       r.ProtoMajor,
+		ProtoMinor:       r.ProtoMinor,
+		Header:           cloneHeader(r.Header),
+		Body:             &bodyBuffer{&bytes.Buffer{}},
+		ContentLength:    r.ContentLength,
+		TransferEncoding: r.TransferEncoding,
+		Close:            r.Close,
+		Host:             r.Host,
+		RemoteAddr:       r.RemoteAddr,
+		RequestURI:       r.RequestURI,
+		TLS:              r.TLS}
+}
+
+func cloneResponseMetadata(r *http.Response) *http.Response {
+	return &http.Response{
+		Status:           r.Status,
+		StatusCode:       r.StatusCode,
+		Proto:            r.Proto,
+		ProtoMajor:       r.ProtoMajor,
+		ProtoMinor:       r.ProtoMinor,
+		Header:           cloneHeader(r.Header),
+		Body:             &bodyBuffer{&bytes.Buffer{}},
+		ContentLength:    r.ContentLength,
+		TransferEncoding: r.TransferEncoding,
+		Close:            r.Close,
+		Request:          r.Request,
+		TLS:              r.TLS}
+}
+
+func mergeParams(to, from map[string]string) map[string]string {
+	if to == nil {
+		to = make(map[string]string)
+	}
+
+	for k, v := range from {
+		to[k] = v
+	}
+
+	return to
+}
+
+func newContext(w http.ResponseWriter, r *http.Request, preserveOriginal bool) *context {
+	c := &context{
+		responseWriter: w,
+		request:        r,
+		stateBag:       make(map[string]interface{}),
+		outgoingHost:   r.Host,
+	}
+
+	if preserveOriginal {
+		c.originalRequest = cloneRequestMetadata(r)
+	}
+
+	return c
+}
+
+func (c *context) incLoopCounter() {
+	c.loopCounter++
+}
+
+func (c *context) decLoopCounter() {
+	c.loopCounter--
+}
+
+func (c *context) applyRoute(route *routing.Route, params map[string]string, preserveHost bool) {
+	c.route = route
+	if preserveHost {
+		c.outgoingHost = c.request.Host
+	} else {
+		c.outgoingHost = route.Host
+	}
+
+	c.pathParams = mergeParams(c.pathParams, params)
+}
+
+func (c *context) ensureDefaultResponse() {
+	if c.response == nil {
+		c.response = defaultResponse(c.request)
+		return
+	}
+
+	if c.response.Header == nil {
+		c.response.Header = make(http.Header)
+	}
+
+	if c.response.Body == nil {
+		c.response.Body = defaultBody()
+	}
+}
+
+func (c *context) deprecatedShunted() bool {
+	return c.deprecatedServed
+}
+
+func (c *context) shunted() bool {
+	return c.servedWithResponse
+}
+
+func (c *context) isShuntRoute() bool {
+	return c.route.Shunt
+}
+
+func (c *context) setResponse(r *http.Response, preserveOriginal bool) {
+	c.response = r
+	if preserveOriginal {
+		c.originalResponse = cloneResponseMetadata(r)
+	}
+}
+
+func (c *context) ResponseWriter() http.ResponseWriter { return c.responseWriter }
+func (c *context) Request() *http.Request              { return c.request }
+func (c *context) Response() *http.Response            { return c.response }
+func (c *context) MarkServed()                         { c.deprecatedServed = true }
+func (c *context) Served() bool                        { return c.deprecatedServed || c.servedWithResponse }
+func (c *context) PathParam(key string) string         { return c.pathParams[key] }
+func (c *context) StateBag() map[string]interface{}    { return c.stateBag }
+func (c *context) BackendUrl() string                  { return c.route.Backend }
+func (c *context) OriginalRequest() *http.Request      { return c.originalRequest }
+func (c *context) OriginalResponse() *http.Response    { return c.originalResponse }
+func (c *context) OutgoingHost() string                { return c.outgoingHost }
+func (c *context) SetOutgoingHost(h string)            { c.outgoingHost = h }
+
+func (c *context) Serve(r *http.Response) {
+	r.Request = c.Request()
+
+	if r.Header == nil {
+		r.Header = make(http.Header)
+	}
+
+	if r.Body == nil {
+		r.Body = defaultBody()
+	}
+
+	c.servedWithResponse = true
+	c.response = r
+}
+
+func (c *context) clone() *context {
+	var cc context
+	cc = *c
+	return &cc
+}
--- a/proxy/context.go
+++ b/proxy/context.go
@@ -30,7 +30,6 @@ type context struct {
 	debugFilterPanics     []interface{}
 	outgoingDebugRequest  *http.Request
 	incomingDebugResponse *http.Response
-	loopCounter           int
 	startServe            time.Time
 }
 
@@ -43,14 +42,15 @@ func defaultResponse(r *http.Request) *http.Response {
 		StatusCode: http.StatusNotFound,
 		Header:     make(http.Header),
 		Body:       defaultBody(),
-		Request:    r}
+		Request:    r,
+	}
 }
 
 func (sb bodyBuffer) Close() error {
 	return nil
 }
 
-func cloneUrl(u *url.URL) *url.URL {
+func cloneURL(u *url.URL) *url.URL {
 	uc := *u
 	return &uc
 }
@@ -58,7 +58,7 @@ func cloneUrl(u *url.URL) *url.URL {
 func cloneRequestMetadata(r *http.Request) *http.Request {
 	return &http.Request{
 		Method:           r.Method,
-		URL:              cloneUrl(r.URL),
+		URL:              cloneURL(r.URL),
 		Proto:            r.Proto,
 		ProtoMajor:       r.ProtoMajor,
 		ProtoMinor:       r.ProtoMinor,
@@ -70,7 +70,8 @@ func cloneRequestMetadata(r *http.Request) *http.Request {
 		Host:             r.Host,
 		RemoteAddr:       r.RemoteAddr,
 		RequestURI:       r.RequestURI,
-		TLS:              r.TLS}
+		TLS:              r.TLS,
+	}
 }
 
 func cloneResponseMetadata(r *http.Response) *http.Response {
@@ -86,7 +87,8 @@ func cloneResponseMetadata(r *http.Response) *http.Response {
 		TransferEncoding: r.TransferEncoding,
 		Close:            r.Close,
 		Request:          r.Request,
-		TLS:              r.TLS}
+		TLS:              r.TLS,
+	}
 }
 
 func mergeParams(to, from map[string]string) map[string]string {
@@ -114,14 +116,6 @@ func newContext(w http.ResponseWriter, r *http.Request, preserveOriginal bool) *
 	}
 
 	return c
-}
-
-func (c *context) incLoopCounter() {
-	c.loopCounter++
-}
-
-func (c *context) decLoopCounter() {
-	c.loopCounter--
 }
 
 func (c *context) applyRoute(route *routing.Route, params map[string]string, preserveHost bool) {
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -22,8 +22,6 @@ const (
 	proxyErrorFmt   = "proxy: %s"
 	unknownRouteId  = "_unknownroute_"
 
-	DefaultMaxLoopbacks = 9
-
 	// The default value set for http.Transport.MaxIdleConnsPerHost.
 	DefaultIdleConnsPerHost = 64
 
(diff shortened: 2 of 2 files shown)


hjacobs:
This is relevant for Kubernetes Ingress, right? So we should review too @Raffo @szuecs @mikkeloscar @ideahitme @linki

aryszka:
@hjacobs yes, it is relevant for everything. I already assigned @szuecs , doing now for the others, too. And all reviews are especially welcome in this case because it can be a sensitive change.

proxy/proxy.go
```diff
@@ -315,134 +279,54 @@ func tryCatch(p func(), onErr func(err interface{})) {
...
-		start = time.Now()
-		tryCatch(func() { fi.Request(ctx) }, onErr)
+		start := time.Now()
+		tryCatch(func() { fi.Request(ctx) }, func(err interface{}) {
+			if p.flags.Debug() {
+				ctx.debugFilterPanics = append(ctx.debugFilterPanics, err)
+				return
+			}
+
+			log.Error("error while processing filters during request: ", err)
```
lmineiro: err msg seems to suggest all filters errored but it is only for a specific one. we should log which one instead

aryszka:
known changes to the previous behavior:
- metrics:route lookup:fix: in the current version the route lookup time is not measured for the cases when a route is not found. This refactor fixes this. (Note, that this change is not covered by tests, because the metrics collection in its entirety is not covered from the beginning. A debt. It would be a nice task to figure how to test the timing sensitive metrics collection in a reliable way. Suggestions welcome, but I think that should be a separate PR.)

proxy/proxy.go
```diff
@@ -315,134 +279,54 @@ func tryCatch(p func(), onErr func(err interface{})) {
...
-		start = time.Now()
-		tryCatch(func() { fi.Response(ctx) }, onErr)
+		start := time.Now()
+		tryCatch(func() { fi.Response(ctx) }, func(err interface{}) {
+			if p.flags.Debug() {
+				ctx.debugFilterPanics = append(ctx.debugFilterPanics, err)
+				return
+			}
+
+			log.Error("error while processing filters during response: ", err)
```
lmineiro: same comment as before - individual filter error

proxy/proxy.go
```diff
@@ -315,134 +279,54 @@ func tryCatch(p func(), onErr func(err interface{})) {
...
+		tryCatch(func() { fi.Request(ctx) }, func(err interface{}) {
+			if p.flags.Debug() {
+				ctx.debugFilterPanics = append(ctx.debugFilterPanics, err)
+				return
+			}
+
+			log.Error("error while processing filters during request: ", err)
+		})
+
 		p.metrics.MeasureFilterRequest(fi.Name, start)
```
lmineiro: I'm inclined to move this inside the closure above, after `fi.Request(ctx)`. Failed filter executions can be fast and skew the metric.

proxy/proxy.go
```diff
@@ -315,134 +279,54 @@ func tryCatch(p func(), onErr func(err interface{})) {
...
+		tryCatch(func() { fi.Response(ctx) }, func(err interface{}) {
+			if p.flags.Debug() {
+				ctx.debugFilterPanics = append(ctx.debugFilterPanics, err)
+				return
+			}
+
+			log.Error("error while processing filters during response: ", err)
+		})
+
 		p.metrics.MeasureFilterResponse(fi.Name, start)
```
lmineiro: Same as above about moving to the closure

proxy/proxy.go
```diff
@@ -462,177 +346,212 @@ func (p *Proxy) lookupRoute(r *http.Request) (rt *routing.Route, params map[stri
...
+		id,
+		c.request.Host,
+		c.request.Method,
+		code,
+		c.startServe,
+	)
+}
+
+func (p *Proxy) makeUpgradeRequest(ctx *context, route *routing.Route, req *http.Request) {
+	// have to parse url again, because path is not be copied by mapRequest
```
lmineiro: s/not be copied/not copied

proxy/proxy.go
```diff
@@ -462,177 +346,212 @@ func (p *Proxy) lookupRoute(r *http.Request) (rt *routing.Route, params map[stri
...
+	p.metrics.MeasureServe(
+		id,
+		c.request.Host,
+		c.request.Method,
+		code,
+		c.startServe,
+	)
+}
+
+func (p *Proxy) makeUpgradeRequest(ctx *context, route *routing.Route, req *http.Request) {
```
lmineiro: What's the goal for this one? HTTP/2?

proxy/context.go
```diff
@@ -0,0 +1,198 @@
...
+	"io"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/zalando/skipper/routing"
+)
+
+// a byte buffer implementing the Closer interface
+type bodyBuffer struct {
```
aryszka: true, thx

proxy/context.go
```diff
@@ -0,0 +1,198 @@
...
+		TransferEncoding: r.TransferEncoding,
+		Close:            r.Close,
+		Request:          r.Request,
+		TLS:              r.TLS,
+	}
+}
+
+func mergeParams(to, from map[string]string) map[string]string {
+	if to == nil {
+		to = make(map[string]string)
```
aryszka: well... https://github.com/zalando/pathmux/blob/master/tree.go#L335

i think you're right and we can handle it as given. I remove the safety, this code path is tested. (Debt: coverage to detect unused code ;) )

proxy/proxy.go
```diff
@@ -164,23 +144,7 @@ type Proxy struct {
...
-	originalRequest    *http.Request
-	originalResponse   *http.Response
-	backendUrl         string
-	outgoingHost       string
-}
-
-func (sb bodyBuffer) Close() error {
-	return nil
-}
+var errProxyCanceled = errors.New("proxy canceled")
```
aryszka: agree, comes handy when grepping, thx

proxy/proxy.go
aryszka: this is due to one of our design debts: currently the filter extension interface doesn't tell which filter this is. I would better postpone fixing this issue instead of using e.g. an index. Or shall we print the index for now?

proxy/proxy.go
aryszka: agree, thx

proxy/proxy.go
aryszka: no.. SPDY, web sockets. One known use case was the gcp client. Relevant: https://github.com/zalando/skipper/issues/159 and https://github.com/zalando/skipper/pull/167

I think we should decide about this feature independently from this PR.

proxy/proxy.go
aryszka: thx

proxy/proxy.go
lmineiro: Some lines below we do use the filter name - `fi.Name`

aryszka: sorry, my bad. I remember now that routing.Filter has the name. Thanks for pointing out the reference below.

hjacobs:
Regarding the branding: do we really need two headers (`X-Powered-By` and `Server`) for this? It looks like overkill and I expected you guys to be eager to save those transfer bytes :smirk:

aryszka:
@hjacobs i never really challenged but i would also be fine with only one header. Now it could break someone's client. Opinions? Which one do you think would be better to remove?

proxy/proxy.go
aryszka: actually, seems like we can keep this ;)

lmineiro: yes, my mistake ;-)

lmineiro:
The `Server` header was initially set to hide some backend signatures, only the `X-Powered-By` was really branding. I guess we could just keep one of them and delete the other one (even if already set).

lmineiro:
👍

proxy/context.go
```diff
@@ -0,0 +1,178 @@
...
+		ProtoMinor:       r.ProtoMinor,
+		Header:           cloneHeader(r.Header),
+		Body:             defaultBody(),
+		ContentLength:    r.ContentLength,
+		TransferEncoding: r.TransferEncoding,
+		Close:            r.Close,
+		Host:             r.Host,
+		RemoteAddr:       r.RemoteAddr,
+		RequestURI:       r.RequestURI,
+		TLS:              r.TLS,
```
mo-gr: Should this also clone the `Form` fields and `Trailer`? 

Also, copying the `ContentLength` but using an empty/default body seems dangerous, as it leaves the request in a potentially invalid state.

proxy/context.go
```diff
@@ -0,0 +1,178 @@
...
+		ProtoMajor:       r.ProtoMajor,
+		ProtoMinor:       r.ProtoMinor,
+		Header:           cloneHeader(r.Header),
+		Body:             defaultBody(),
+		ContentLength:    r.ContentLength,
+		TransferEncoding: r.TransferEncoding,
+		Close:            r.Close,
+		Request:          r.Request,
+		TLS:              r.TLS,
+	}
```
mo-gr: Same question as above for `cloneRequestMetadata`. Is this fine as it is?

proxy/context.go
```diff
@@ -0,0 +1,178 @@
...
+func (c *context) deprecatedShunted() bool {
+	return c.deprecatedServed
+}
+
+func (c *context) shunted() bool {
+	return c.servedWithResponse
+}
+
+func (c *context) isShuntRoute() bool {
+	return c.route.Shunt
```
mo-gr: maybe this function should also check for route being nil? I know it is only called after a route has been chosen at the moment. But as it is right now, it seems a bit dangerous...

proxy/context.go
```diff
@@ -0,0 +1,178 @@
...
+
+	if r.Body == nil {
+		r.Body = defaultBody()
+	}
+
+	c.servedWithResponse = true
+	c.response = r
+}
+
+func (c *context) clone() *context {
```
mo-gr: You have another function calle `cloneUrl`, so I think this one should be called `cloneContext`.

Also, this one has a slightly different implementation. I would prefer all the clone-helper to be mostly symmetrical and located closer to each other in the source file.

proxy/proxy.go
```diff
@@ -315,134 +279,57 @@ func tryCatch(p func(), onErr func(err interface{})) {
...
-		StatusCode: http.StatusNotFound,
-		Header:     make(http.Header),
-		Body:       &bodyBuffer{&bytes.Buffer{}},
-		Request:    r}
-}
-
 // applies all filters to a request
-func (p *Proxy) applyFiltersToRequest(f []*routing.RouteFilter, ctx *filterContext, onErr func(err interface{})) []*routing.RouteFilter {
-	var start time.Time
+func (p *Proxy) applyFiltersToRequest(f []*routing.RouteFilter, ctx *context) []*routing.RouteFilter {
```
mo-gr: if you name the function `applyAllFiltersToRequest`, the comment in line 282 can be deleted ;-)

proxy/proxy.go
```diff
@@ -462,177 +349,212 @@ func (p *Proxy) lookupRoute(r *http.Request) (rt *routing.Route, params map[stri
...
+
+	response, err := p.roundTripper.RoundTrip(req)
+	if err != nil {
+		code := http.StatusInternalServerError
+		if _, ok := err.(net.Error); ok {
+			code = http.StatusServiceUnavailable
 		}
+
+		p.sendError(ctx, code)
+		log.Error("error during backend roundtrip: ", err)
```
mo-gr: Might be helpful to have the route-id in the error log?

proxy/proxy.go
```diff
@@ -462,177 +349,212 @@ func (p *Proxy) lookupRoute(r *http.Request) (rt *routing.Route, params map[stri
...
+	if err != nil {
+		log.Errorf("could not map backend request, caused by: %v", err)
+		p.sendError(ctx, http.StatusInternalServerError)
+		return nil, errProxyCanceled
+	}
+
+	if p.experimentalUpgrade && isUpgradeRequest(req) {
+		p.makeUpgradeRequest(ctx, ctx.route, req)
+		// We are not owner of the connection anymore.
+		return nil, errProxyCanceled
```
mo-gr: We are not the owner anymore. But is this really an errProxyCanceled? Or should this have its own err like errConnectionUpgraded?

proxy/proxy.go
```diff
@@ -462,177 +349,212 @@ func (p *Proxy) lookupRoute(r *http.Request) (rt *routing.Route, params map[stri
...
-				}
+	processedFilters := p.applyFiltersToRequest(ctx.route.Filters, ctx)
 
-				sendError(w, http.StatusText(code), code)
-				p.metrics.MeasureServe(rt.Id, r.Host, r.Method, code, startServe)
-				log.Error("error during backend roundtrip: ", err)
-				return
-			}
+	if ctx.deprecatedShunted() {
+		return errProxyCanceled
```
mo-gr: Should we log here that deprecated shunting was used?

proxy/proxy.go
```diff
@@ -462,177 +349,212 @@ func (p *Proxy) lookupRoute(r *http.Request) (rt *routing.Route, params map[stri
...
-					err:          err,
-					filterPanics: filterPanics})
-				return
-			}
+	return response, nil
+}
 
-			rs = &http.Response{Header: make(http.Header)}
-		} else {
+func (p *Proxy) do(ctx *context) error {
```
mo-gr: Is there a better name for this than`do` ?

proxy/context.go
aryszka: Your point regarding the Trailer is valid. Copying the From field is not necessary, it's always empty at this point. From the docs:

```
    // Form contains the parsed form data, including both the URL
    // field's query parameters and the POST or PUT form data.
    // This field is only available after ParseForm is called.
    // The HTTP client ignores Form and uses Body instead.
```

The ContentLength: this object is only used to get information about the original request and not used for the actual request chain, that's why it is called cloneRequestMetadata. The reason for using a dummy body is that we don't buffer the content. I think it actually would be more dangerous to put the real body there.

Updating it with the Trailer, as that's a fix.

proxy/context.go
aryszka: Updating with the Trailer. For the Body, the same applies as in the cloneRequestMetadata.

proxy/context.go
aryszka: i'd rather remove the function now and call c.route.Shunt directly. I am not too much in favor of starting a full preventive style for package internal functions in case of this piece of code. What do you think?

proxy/proxy.go
aryszka: right. thanks

proxy/proxy.go
aryszka: errProxyCanceled means something like this: "It indicates to the caller that the current proxy mechanism is canceled due to varying reasons. If there were any errors, they are already handled."

Do you think it is OK to just put this comment next to the errProxyCanceled definition? Or can you suggest a better name?

Regarding a separate error object for the upgrade, it sounds legit, but it would come with the inconvenience that an additional case would need to be checked in the calling code.

proxy/proxy.go
aryszka: right, thx

proxy/proxy.go
aryszka: suggest one, i am open to change it. (I peronally really like this one :) )

proxy/proxy.go
aryszka: sure! warn or only debug? warn can flood someones logs..

aryszka:
removed the context.clone() function as it wasn't used. thx @Raffo

proxy/proxy.go
aryszka: not sure actually. this one comes in pair with the applyFiltersToResponse, and both apply the filters that are passed in as args, but the response one doesn't necessarily receives all the filters, because of the shunting.

proxy/context.go
aryszka: trailer added

proxy/context.go
aryszka: tailer added

proxy/proxy.go
```diff
@@ -315,134 +279,57 @@ func tryCatch(p func(), onErr func(err interface{})) {
...
-		start = time.Now()
-		tryCatch(func() { fi.Response(ctx) }, onErr)
-		p.metrics.MeasureFilterResponse(fi.Name, start)
+		start := time.Now()
+		tryCatch(func() {
+			fi.Response(ctx)
+			p.metrics.MeasureFilterResponse(fi.Name, start)
+		}, func(err interface{}) {
+			if p.flags.Debug() {
+				ctx.debugFilterPanics = append(ctx.debugFilterPanics, err)
```
szuecs: You have created a new side effect here (appending to ctx). I think it should be documented in the doc string of the method.

proxy/proxy.go
```diff
@@ -315,134 +279,57 @@ func tryCatch(p func(), onErr func(err interface{})) {
...
-		start = time.Now()
-		tryCatch(func() { fi.Request(ctx) }, onErr)
-		p.metrics.MeasureFilterRequest(fi.Name, start)
+		start := time.Now()
+		tryCatch(func() {
+			fi.Request(ctx)
+			p.metrics.MeasureFilterRequest(fi.Name, start)
+		}, func(err interface{}) {
+			if p.flags.Debug() {
+				ctx.debugFilterPanics = append(ctx.debugFilterPanics, err)
```
szuecs: You have created a new side effect here (appending to ctx). I think it should be documented

proxy/proxy.go
szuecs: To clarify the use case: It as used to be a transparent proxy for kubernetes API server and kubectl exec is using upgrade protocol to SPDY to have the bidirectional stream for the shell into container. We could provide audit logs with it and we had to convert oauth2 access to a different one (basic auth) to access GCP apiserver.

proxy/proxy.go
```diff
@@ -462,177 +349,213 @@ func (p *Proxy) lookupRoute(r *http.Request) (rt *routing.Route, params map[stri
...
-	if !c.served {
-		if p.flags.Debug() {
-			dbgResponse(w, &debugInfo{
-				route:        &rt.Route,
-				incoming:     c.OriginalRequest(),
-				outgoing:     debugReq,
-				response:     c.Response(),
-				filterPanics: filterPanics})
+	if err != nil {
+		if err == errProxyCanceled {
```
szuecs: Do you really want to have a silent return?

proxy/context.go
```diff
@@ -0,0 +1,170 @@
...
+		Request:    r,
+	}
+}
+
+func cloneURL(u *url.URL) *url.URL {
+	uc := *u
+	return &uc
+}
+
+func cloneRequestMetadata(r *http.Request) *http.Request {
```
boopathi: Why is clone useful ?

This is in context of `r.Body` can be read only once and you want to consume it multiple times. But here, since the body is always replaced with `defaultBody()`, why is cloning the metadata useful ? 

Also, same question for `cloneResponseMetadata`.

proxy/proxy.go
aryszka: thx, i'll add the docs

proxy/proxy.go
aryszka: from another comment:

errProxyCanceled: "It indicates to the caller that the current proxy mechanism is canceled due to varying reasons. If there were any errors, they are already handled."

i add this to the errProxyCanceled as docs. Does it make sense?

I am also not sure anymore if this is the best approach. Will have a look if it can be handled differently. The reason for this is that i am planning a feature whose implementation triggered this refactor: #274

proxy/context.go
aryszka: this is to provide information to the custom filters about the original incoming request before the filters could change it:

https://godoc.org/github.com/zalando/skipper/filters#FilterContext

```
    // The copy (deep) of the original incoming request or nil if the
    // implementation does not provide it.
    //
    // The object received from this method contains an empty body, and all
    // payload related properties have zero value.
```

proxy/proxy.go
aryszka: i'll add the docs, thx

to give an explanation, this was "practically" already like here, but instead of using a field on the context object, it was used as a variable. See the earlier filterPanics variable. The main reason of moving this variable to context object is that i am planning to open a pull request with a loopback backend. The context object is used only for a single request flow, and only evaluated by the debug scenario, so I believe it's just fine.

proxy/proxy.go
mo-gr: ok. a small comment to errProxyConaceled sounds reasonable to me

szuecs:
From my side lgtm

szuecs:
:+1:

haldun:
👍

proxy/proxy.go
aryszka: @szuecs thanks, cleaned up the error handling a bit

proxy/proxy.go
```diff
@@ -164,22 +143,32 @@ type Proxy struct {
...
-	originalRequest    *http.Request
-	originalResponse   *http.Response
-	backendUrl         string
-	outgoingHost       string
+// proxyError is used to wrap errors during proxying and to indicate
+// the required status code for the response sent from the main
+// ServeHTTP method. Alternatively, it can indicate that the request
+// was already handled, e.g. in case of deprecated shunting or the
+// upgrade request.
+type proxyError struct {
```
boopathi: don't you want to export this type ?

aryszka: no. This is only used internally from ServeHTTP. It doesn't return an error.

boopathi:
👍

PR #329, opened 2017-05-18
Fix eskip usage text for -innkeeper-url

DeTeam:
`eskip --help` -> `-innkeeper-url` part had usage text from the etcd url, updated it.

Not it looks like this:

```
  -innkeeper-url string
    	url for the innkeeper service
```


aryszka:
:+1:

hjacobs:
:+1:

Issue #324, opened 2017-05-11
Kubernetes Ingress: Weighted Traffic Switching


hjacobs:
Use the new traffic functionality (introduced with #319) to support weighted traffic switching for Kubernetes Ingress. This enables green/blue or canary deployments.

Idea: use different backends (services) in the Ingress and use relative weights to specify how much traffic each backend service should receive. The relative weights are annotated as a simple JSON object (backend service name -> relative weight).

Relative weights can be specified as arbitrary float values, examples for expressing exactly the same traffic distribution (20% to 80%):

*  {"my-app-green": 0.2, "my-app-blue": 0.8}
*  {"my-app-green": 2, "my-app-blue": 8}
*  {"my-app-green": 20, "my-app-blue": 80}
*  {"my-app-green": 1, "my-app-blue": 4}

Example `ingress.yaml` for switching 10% of the traffic to the new service `my-app-v2`.

```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    zalando.org/backend-weights: |
      {"my-app-v1": 90, "my-app-v2": 10}
  name: my-app
spec:
  rules:
  - host: my-app.example.org
    http:
      paths:
      - backend:
          serviceName: my-app-v1
          servicePort: 80
      - backend:
          serviceName: my-app-v2
          servicePort: 80
      - path: /some-sub-path-without-traffic-switching
        backend:
          serviceName: some-other-app
          servicePort: 80
```

Benefits:

* Simplicity: we only annotate the Ingress resource with the weights (if we could extend Ingress we would put the weight into the `backend` property as proposed in https://github.com/kubernetes/kubernetes/issues/25485)
* Usability: the traffic weights are easily understandable by humans and by using relative weights it's hard to make some error (nobody needs to make sure that the sum is exactly 1.0 or 100).

Drawbacks:

* Would need to be extended if we would need to support traffic switching on all levels with the same backends (i.e. differents paths routing to the same service) --- I consider it unlikely that we need this
* The Ingress resource needs to be changed (overwritten) for switching traffic --- this can be approached by using templating in the CD pipeline and/or extracting the weights to a different resource in the future (TPR, config map?)

Jan-M:
"Would need to be extended if we would need to support traffic switching on all levels with the same backends (i.e. differents paths routing to the same service) --- I consider it unlikely that we need this"

I would actually agree here, e.g. current zmon proxying and tuning kairosdb could hugely benefit from support weights on paths.

[...]

PR #330, opened 2017-05-19
Implement traffic switching in kubernetes dataclient

mikkeloscar:
This implements `backend-weights` (or traffic switching) using an annotation `zalando.org/backend-weights` as described in #324.

It works by parsing the annotation of format:

```
zalando.org/backend-weights: |
  {"my-app-v1": 90, "my-app-v2": 10}
```

and maps the weights defined in the annotation to the backend services defined in the ingress spec.

The traffic weights are computed relative to the number of backends, and subsequent backends for a path will get higher relative weight because of how the `Traffic` predicate in skipper works. E.g. if you have the following backends and weights:

```
* backend-1: 0.2
* backend-2: 0.6
* backend-3: 0.2
```

The computed traffic values will be:

```
* backend-1: 0.2
* backend-2: 0.75
* backend-3: 1.0
```

Which results in the following eskip rules:

```
backend-1:
  Traffic(0.2) ->

backend-2:
  Traffic(0.75) ->

backend-3:
  * ->
```

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -48,23 +48,25 @@ import (
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/filters/builtin"
 	"github.com/zalando/skipper/predicates/source"
+	"github.com/zalando/skipper/predicates/traffic"
 )
 
 // FEATURE:
 // - provide option to limit the used namespaces?
 
 const (
-	defaultKubernetesURL    = "http://localhost:8001"
-	ingressesURI            = "/apis/extensions/v1beta1/ingresses"
-	serviceURIFmt           = "/api/v1/namespaces/%s/services/%s"
-	serviceAccountDir       = "/var/run/secrets/kubernetes.io/serviceaccount/"
-	serviceAccountTokenKey  = "token"
-	serviceAccountRootCAKey = "ca.crt"
-	serviceHostEnvVar       = "KUBERNETES_SERVICE_HOST"
-	servicePortEnvVar       = "KUBERNETES_SERVICE_PORT"
-	healthcheckRouteID      = "kube__healthz"
-	httpRedirectRouteID     = "kube__redirect"
-	healthcheckPath         = "/kube-system/healthz"
+	defaultKubernetesURL        = "http://localhost:8001"
+	ingressesURI                = "/apis/extensions/v1beta1/ingresses"
+	serviceURIFmt               = "/api/v1/namespaces/%s/services/%s"
+	serviceAccountDir           = "/var/run/secrets/kubernetes.io/serviceaccount/"
+	serviceAccountTokenKey      = "token"
+	serviceAccountRootCAKey     = "ca.crt"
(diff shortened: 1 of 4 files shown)


dataclients/kubernetes/kube.go
```diff
@@ -310,12 +312,13 @@ func (c *Client) getServiceURL(namespace, name string, port backendPort) (string
...
 // TODO: find a nicer way to autogenerate route IDs
-func routeID(namespace, name, host, path string) string {
+func routeID(namespace, name, host, path, backend string) string {
 	namespace = nonWord.ReplaceAllString(namespace, "_")
 	name = nonWord.ReplaceAllString(name, "_")
 	host = nonWord.ReplaceAllString(host, "_")
 	path = nonWord.ReplaceAllString(path, "_")
-	return fmt.Sprintf("kube_%s__%s__%s__%s", namespace, name, host, path)
+	backend = nonWord.ReplaceAllString(backend, "_")
+	return fmt.Sprintf("kube_%s__%s__%s__%s__%s", namespace, name, host, path, backend)
```
hjacobs: this also means that the metrics change, right?

aryszka: the route based ones, yes. However, it should be possible to collect them the old way, because the metrics endpoint does a prefix match on them. Is that an option?

dataclients/kubernetes/kube.go
```diff
@@ -417,26 +438,95 @@ func (c *Client) ingressToRoutes(items []*ingressItem) []*eskip.Route {
...
+//      backend-3: 0.2
+//
+// then the weight will be calculated to:
+//
+//      backend-1: 0.2
+//      backend-2: 0.75
+//      backend-3: 1.0
+//
+// where for a weight of 1.0 no Traffic predicate will be generated.
+func computeBackendWeights(backendWeights map[string]float64, rule *rule) {
```
aryszka: if I understand correctly, the calculated weights assume that the leave routes on the same path are evaluated in the order of the input. Is this correct?

If so, then we need to change this:

https://github.com/zalando/skipper/blob/master/routing/matcher.go#L228

and this:

https://github.com/zalando/skipper/blob/master/routing/matcher.go#L308

to sort.Stable: https://godoc.org/sort#Stable

mikkeloscar: > if I understand correctly, the calculated weights assume that the leave routes on the same path are evaluated in the order of the input. Is this correct?

Yeah, this was my assumption, would changing to a stable sort cause problems elsewhere?

aryszka: no. Stable is fine. We simply didn't need it up until now

mikkeloscar: Changed to sort.Stable.

aryszka:
:+1:

dataclients/kubernetes/kube.go
mikkeloscar: @hjacobs you are right. But since we have to break it one way or another I think it's reasonable like it is. (I found one internal check relying on the old behavior).

hjacobs:
:+1:

Raffo:
👍

PR #331, opened 2017-05-22
Loop backend

aryszka:
implement the loop backend #274

### DIFF
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -80,6 +80,7 @@ const (
 	backendFlushIntervalUsage      = "flush interval for upgraded proxy connections"
 	experimentalUpgradeUsage       = "enable experimental feature to handle upgrade protocol requests"
 	versionUsage                   = "print Skipper version"
+	maxLoopbacksUsage              = "maximum number of loopbacks for an incoming request, set to -1 to disable loopbacks"
 )
 
 var (
@@ -129,6 +130,7 @@ var (
 	backendFlushInterval      time.Duration
 	experimentalUpgrade       bool
 	printVersion              bool
+	maxLoopbacks              int
 )
 
 func init() {
@@ -173,6 +175,7 @@ func init() {
 	flag.DurationVar(&backendFlushInterval, "backend-flush-interval", defaultBackendFlushInterval, backendFlushIntervalUsage)
 	flag.BoolVar(&experimentalUpgrade, "experimental-upgrade", defaultExperimentalUpgrade, experimentalUpgradeUsage)
 	flag.BoolVar(&printVersion, "version", false, versionUsage)
+	flag.IntVar(&maxLoopbacks, "max-loopbacks", proxy.DefaultMaxLoopbacks, maxLoopbacksUsage)
 	flag.Parse()
 }
 
@@ -256,6 +259,7 @@ func main() {
 		KeyPathTLS:                keyPathTLS,
 		BackendFlushInterval:      backendFlushInterval,
 		ExperimentalUpgrade:       experimentalUpgrade,
+		MaxLoopbacks:              maxLoopbacks,
 	}
 
 	if insecure {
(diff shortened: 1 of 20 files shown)


eskip/eskip.go
```diff
@@ -33,32 +18,41 @@ var (
 	invalidPredicateArgCountError   = errors.New("invalid predicate count arg")
 	duplicatePathTreePredicateError = errors.New("duplicate path tree predicate")
 	duplicateMethodPredicateError   = errors.New("duplicate method predicate")
+	invalidBackendErr               = errors.New("invalid backend")
```
haldun: maybe `errInvalidBackend` is better here. This also applies the other errors.

haldun:
👍

cmd/skipper/main.go
```diff
@@ -80,6 +80,7 @@ const (
 	backendFlushIntervalUsage      = "flush interval for upgraded proxy connections"
 	experimentalUpgradeUsage       = "enable experimental feature to handle upgrade protocol requests"
 	versionUsage                   = "print Skipper version"
+	maxLoopbacksUsage              = "maximum number of loopbacks for an incoming request, set to -1 to disable loopbacks"
```
mo-gr: Wouldn't `0` loopbacks also disable the feature?

doc.go
```diff
@@ -75,10 +61,20 @@ filters in reverse order. Finally, it is mapped as the response of the
...
 
 Besides the default proxying mechanism, it is possible to define routes
-without a real network backend endpoint. This is called a 'shunt'
-backend, in which case one of the filters needs to handle the request
-providing its own response (e.g. the 'static' filter). Actually, filters
-themselves can instruct the request flow to shunt.
+without a real network backend endpoint. One of these cases is called a
+'shunt' backend, in which case one of the filters needs to handle the
+request providing its own response (e.g. the 'static' filter). Actually,
+filters themselves can instruct the request flow to shunt.
```
mo-gr: Would be cool to also say how a filter can cause a shunt.

doc.go
```diff
@@ -137,6 +134,10 @@ A shunt route means that Skipper handles the request alone and doesn't
 make requests to a backend service. In this case, it is the
 responsibility of one of the filters to generate the response.
 
+A loopback route executes the routine mechanism on current state of
```
mo-gr: `s/routine/routing/` ?

eskip/doc.go
```diff
@@ -178,7 +178,7 @@ predicates start with upper case, while filters start with lower case.
 
 Backend
 
-There are two types of backends: a network endpoint address or a shunt.
+There are three types of backends: a network endpoint address or a shunt.
```
mo-gr: > There are three types

But you only mention the old two ;-)

eskip/example_test.go
```diff
@@ -38,12 +39,15 @@ func Example() {
...
+
 		// route definition directing requests to an api endpoint
 		route3:
 			Method("POST") && Path("/api") ->
 			requestHeader("X-Type", "ajax-post") ->
-			"https://api.example.org"
+			"https://api.example.org";
+
+		// route definition with a loopback (no backend address)
+		route4: Path("/some/other/path") -> static("/", "/var/www") -> <loopback>;
```
mo-gr: serve static with a loopback? Interesting example. What does this actually do? Maybe better to have a `setHeader` in the example to make it more meaningful

proxy/proxy.go
```diff
@@ -283,6 +297,12 @@ func WithParams(p Params) *Proxy {
 		m = metrics.Void
 	}
 
+	if p.MaxLoopbacks == 0 {
+		p.MaxLoopbacks = DefaultMaxLoopbacks
+	} else if p.MaxLoopbacks < 0 {
+		p.MaxLoopbacks = 0
+	}
```
mo-gr: Is this a common pattern in go?
```
if foo = 0
  foo = DEFAULT
else if foo < 0
  foo = 0
```

To me, this is an evil and confusing pattern

PR #333, opened 2017-05-25
add illustrator source and svg for skipper logo

dazld:
I vectorized™ the skipper logo, and there may be a little easter egg there too, for a bit of fun.

Any good?


dazld:
will talk offline.

PR #334, opened 2017-05-25
Add support for kubernetes.io/ingress.class annotation

slok:
Hi!

This PR adds support for the k8s `kubernetes.io/ingress.class` [annotation on the ingress rules](https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/multiple-ingress-controllers), this way you can have multiple ingress controllers.

In the end it will filter the ingresses obtained from k8s checking this annotation with these cases as result:
* If no annotation: valid
* If empty annotation: valid
* If different annotation from the desired one: invalid 
* If same annotarion as the desired one: valid

This way, the change is backwards compatible with all the ingresses that where previously being used.

By default if no ingress class is configured on run, it will be set `skipper` by default.

For example this would be a valid ingress for skipper:

```yaml
kind: Ingress
metadata:
  name: valid-ingress
  annotations:
    kubernetes.io/ingress.class: "skipper"
...
```
and this not:

```yaml
kind: Ingress
metadata:
  name: invalid-ingress
  annotations:
    kubernetes.io/ingress.class: "nginx"
...
```

### DIFF
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -47,6 +47,7 @@ const (
 	kubernetesURLUsage             = "kubernetes API base URL for the ingress data client; requires kubectl proxy running; omit if kubernetes-in-cluster is set to true"
 	kubernetesHealthcheckUsage     = "automatic healthcheck route for internal IPs with path /kube-system/healthz; valid only with kubernetes"
 	kubernetesHTTPSRedirectUsage   = "automatic HTTP->HTTPS redirect route; valid only with kubernetes"
+	kubernetesIngressClassUsage    = "ingress class used to filter ingress resources on kubernetes"
 	innkeeperUrlUsage              = "API endpoint of the Innkeeper service, storing route definitions"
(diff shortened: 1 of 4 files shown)


aryszka:
Thanks for the pull request!

:+1:

hjacobs:
Great, thanks for the PR! :smile:

hjacobs:
:+1:

mikkeloscar:
I think it maybe also should be possible to completely opt out of this annotation when running skipper.

Since skipper is not a real ingress controller at this point (it never writes to the ingress object) it should be possible to let skipper configure itself for all ingresses while still using and respecting the annotation in other ingress controllers running in the same environment IMO.

aryszka:
e.g. with a flag like -kubernetes-ignore-ingress-class? or using -kubernetes-ingress-class="*" to just generate all the routes?

mikkeloscar:
> e.g. with a flag like -kubernetes-ignore-ingress-class? or using -kubernetes-ingress-class="*" to just generate all the routes?

Yeah, something like that!

PR #336, opened 2017-05-26
flag for ignoring the ingress class

aryszka:
added a flag to explicitly ignore the Kubernetes ingress class.

See the discussion in: #334


mikkeloscar:
:+1:

hjacobs:
I'm not sure if this makes sense for me: we might need a more generic approach, e.g. with pattern/regex matching. Why? We have the `kube-ingress-aws-controller` and `skipper` which both need to watch Ingress rules and potentially act on them, i.e. the ingress class might need to specify more than one controller and pattern matching would be the most simple and generic approach to allow both "match all" and "match a substring":

* `skipper` should be able to ignore the ingress class annotation
* `skipper` should be able to only consider Ingress resources which have the "skipper" ingress class or a list of ingress classes including "skipper" (e.g. "alb,skipper")

slok:
I agree with @hjacobs a regex solution would be more flexible

mikkeloscar:
lgtm

PR #338, opened 2017-06-02
Add ca-certificates to Docker image

mikkeloscar:
Adds `ca-certificates` to the docker image so it can be used for routes with https backends:

```
route: * -> "https://google.com"
```

Without certs you get:

```
level=error msg="error during backend roundtrip: x509: failed to load system roots and no roots provided"
```


mikkeloscar:
Err dup of #288, sorry.

PR #342, opened 2017-06-13
use metrics keys only from known hosts and methods

aryszka:
since the metrics are created on demand, and some of them have keys based on request method and host, it is possible to create unwanted number of metrics from user input. With this PR:

- use common host based metrics for those requests that were not matched against a host predicate
- allow only standard methods in metrics keys

### DIFF
--- a/proxy/context.go
+++ b/proxy/context.go
@@ -11,6 +11,8 @@ import (
 	"github.com/zalando/skipper/routing"
 )
 
+const unknownHost = "_unknownhost_"
+
 type context struct {
 	responseWriter        http.ResponseWriter
 	request               *http.Request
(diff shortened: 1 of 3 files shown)


szuecs:
👍

proxy/context.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/zalando/skipper/routing"
 )
 
+const unknownHost = "_unknownhost_"
```
mo-gr: consistency. you have `_unknown_method_` and `_unknownhost_`. I would make it consistent and change to `_unknown_host_` - also reads nicer

mo-gr:
:+1:

Issue #326, opened 2017-05-11
Add flag for logs in json format


milosradovanovic:
It would be nice to have config param to enable logs in json format. 
That way logs shall look like:

{"duration":68,"host":"::1","level":"info","method":"GET","msg":"","proto":"HTTP/1.1","referer":"","requested-host":"localhost:8088","response-size":0,"status":304,"time":"2017-05-05T17:20:55+02:00","timestamp":"05/May/2017:17:20:55 +0200","uri":"/v2/customers/","user-agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.81 Safari/537.36"}

jreyeshdez:
There is no need to output "time":"2017-05-05T17:20:55+02:00" and "timestamp":"05/May/2017:17:20:55 +0200" ; I think it is enough to just have the latter ("timestamp":"23/Jun/2017:10:23:06 +0100")

The above can be achieved by using custom Timestamp Format and by disabling automatic timestamps in output in JSONFormatter.

aryszka:
@milosradovanovic @jreyeshdez can we considered this completed by https://github.com/zalando/skipper/pull/357 ? Please feel free to reopen it or create a new issue if something is missing.

jreyeshdez:
@aryszka I am happy with it. If there is any issue related to it, please let me know.

milosradovanovic:
Looks good. Thanks

PR #354, opened 2017-06-23
fix eskip patch append

aryszka:
fixing in bug in processing input args

fixes: #354

PR #357, opened 2017-06-23
Add flag for logs in json format #326

jreyeshdez:
Added a config param to enable logs in JSON format.
Solution for  #326

Log looks like:
{"duration":89,"host":"::1","level":"info","method":"GET","msg":"","proto":"HTTP/1.1","referer":"","requested-host":"localhost:9090","response-size":606,"status":404,"time":"2017-06-23T10:23:06+01:00","timestamp":"23/Jun/2017:10:23:06 +0100","uri":"/hello","user-agent":"Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36"}

### DIFF
--- a/cmd/eskip/args.go
+++ b/cmd/eskip/args.go
@@ -220,7 +220,7 @@ func processPatchArgs(pfilters, pfile, afilters, afile string) []*medium {
 	}
 
 	if afilters != "" {
-		media = append(media, &medium{typ: patchAppend, patchFilters: pfilters})
+		media = append(media, &medium{typ: patchAppend, patchFilters: afilters})
 	}
 
 	if pfile != "" {
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -75,7 +75,7 @@ const (
 	applicationLogPrefixUsage      = "prefix for each log entry"
 	accessLogUsage                 = "output file for the access log, When not set, /dev/stderr is used"
 	accessLogDisabledUsage         = "when this flag is set, no access log is printed"
-	accessLogJSONDisabledUsage     = "when this flag is set, custom access log format is used"
+	accessLogJSONEnabledUsage      = "when this flag is set, log in JSON format is used"
 	debugEndpointUsage             = "when this address is set, skipper starts an additional listener returning the original and transformed requests"
 	certPathTLSUsage               = "the path on the local filesystem to the certificate file (including any intermediates)"
 	keyPathTLSUsage                = "the path on the local filesystem to the certificate's private key file"
@@ -127,7 +127,7 @@ var (
 	applicationLogPrefix      string
 	accessLog                 string
 	accessLogDisabled         bool
-	accessLogJSONDisabled     bool
+	accessLogJSONEnabled      bool
 	debugListener             string
 	certPathTLS               string
 	keyPathTLS                string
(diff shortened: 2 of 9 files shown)


jreyeshdez:
After used custom Timestamp Format from access log and disabled auto Timestamp for JSONFormatter the JSON log output looks like the following:

{"duration":95,"host":"::1","level":"info","method":"GET","msg":"","proto":"HTTP/1.1","referer":"","requested-host":"localhost:9090","response-size":606,"status":404,"timestamp":"23/Jun/2017:11:41:03 +0100","uri":"/hello","user-agent":"Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36"}

There is no need to output "time":"2017-06-23T10:23:06+01:00" and "timestamp":"23/Jun/2017:10:23:06 +0100" ; I think it is enough to just have the latter ("timestamp":"23/Jun/2017:10:23:06 +0100")

cmd/eskip/args.go
```diff
@@ -220,7 +220,7 @@ func processPatchArgs(pfilters, pfile, afilters, afile string) []*medium {
 	}
 
 	if afilters != "" {
-		media = append(media, &medium{typ: patchAppend, patchFilters: afilters})
+		media = append(media, &medium{typ: patchAppend, patchFilters: pfilters})
```
aryszka: thanks for this catch :) :heart:

cmd/skipper/main.go
```diff
@@ -172,6 +174,7 @@ func init() {
 	flag.StringVar(&applicationLogPrefix, "application-log-prefix", defaultApplicationLogPrefix, applicationLogPrefixUsage)
 	flag.StringVar(&accessLog, "access-log", "", accessLogUsage)
 	flag.BoolVar(&accessLogDisabled, "access-log-disabled", false, accessLogDisabledUsage)
+	flag.BoolVar(&accessLogJSONDisabled, "access-log-json-disabled", true, accessLogJSONDisabledUsage)
```
aryszka: i would prefer to use accessLogJSONEnabled or just simply accessLogJSON.

the reason is that this way the default behavior for the binary distribution, built from cmd/skipper, stays the same, but not so for those projects that import skipper as a library, from the root of the project. This way it would not be backwards compatible for the lib, only for the cmd.

aryszka:
thanks for this PR. It generally looks ok, but see my comment pls about the default behavior of the library distribution.

jreyeshdez:
Changed PR to use _accessLogJSONEnabled_ instead as per comments above. Default value for this config param would be **false**.

cmd/skipper/main.go
jreyeshdez: Thanks for letting me know.
Used _accessLogJSONDisabled_ with default value as **true** just to follow same structure as seen in _accessLogDisabled_ where its default value is **false**.

cmd/eskip/args.go
jreyeshdez: Not sure if I messed up fix done in #354 ?

aryszka: ah.. ok, i see, don't worry, i'll check the merge locally

aryszka:
:+1:

cmd/eskip/args.go
aryszka: ah.. sorry, my bad before, i didn't watch carefully enough. This basically reverts #354. Please keep it as 'afilters'.

jreyeshdez: I was trying to sync repo with my fork and somehow it reverted that fix.

jreyeshdez:
Keeping #354 the way it was.

DeTeam:
👍

aryszka:
:+1:

DeTeam:
👍

PR #359, opened 2017-06-28
replace godeps with glide

aryszka:
Replacing godeps with glide:
- use glide in the common build
- allow building the lib with fixed dependency versions from glide
- keep it possible to build the lib from head with go get
- build the binaries with fixed versions in the common build


Raffo:
Yes!

Raffo:
👍

hjacobs:
:+1:

PR #362, opened 2017-07-03
Circuit breaker

aryszka:
OP:

In the simplest case, the proxy gets a breaker for the current backend host from the registry, if configured. The breakers are always bound to a backend host, even if they share their settings. There are two breaker types, one opens after N consecutive failures, the other opens when the failure rate crosses a threshold within a sliding window (numeric window, not time based). The idle breakers are automatically dropped from the registry when new breakers need to be created, e.g. when some backend hosts were removed from the configuration but new ones were added.

CONFIG:

1. it is possible to define global breaker settings (common settings for each backend host breaker)
2. it is possible to override or disable the global settings for each host individually
3. it is possible to override the global or host based settings for each route individually, using standard filters

DOCS:
https://github.com/zalando/skipper/blob/circuit-breaker/circuit/doc.go


aryszka:
:-1:

DeTeam:
Related issues in dobreaker:

- https://github.com/sony/gobreaker/issues/4
- https://github.com/sony/gobreaker/pull/6

DeTeam:
Read very briefly and looks really nice! Would like to check in more details, but just to get to know things.

👍

circuit/doc.go
```diff
@@ -0,0 +1,152 @@
...
+specific host or route.
+
+Command line name: type. Possible command line values: consecutive, rate, disabled.
+
+Settings - Host
+
+The Host field indicates to which backend host should the current set of settings be applied. Leaving it empty
+indicates global settings.
+
+Command line name: host.
```
DeTeam: \* The answer is no, but mb it's a featue worth considering..

aryszka: how would it work? currently if the host is not set, that means global settings (still with individual breaker instances for the hosts)

circuit/doc.go
```diff
@@ -0,0 +1,152 @@
...
+
+Settings - Window
+
+The window value sets the size of the sliding counter window of the failure rate breaker.
+
+Command line name: window. Possible command line values: any positive integer.
```
aryszka: there are.. but currently they are used from sony/gobreaker and may be replaced in a later wave with https://github.com/ideahitme/circuit-breaker, so considered not declaring the defaults in the docs atm. Basically, 5 consecutive failures, 1 minute timeout, 1 half-open request. The default IdleTTL is documented.

circuit/doc.go
```diff
@@ -0,0 +1,152 @@
...
+
+Settings - Failures
+
+The failures value sets the max failure count for both the "consecutive" and "rate" breakers.
+
+Command line name: failures. Possible command line values: any positive integer.
```
aryszka: see above

DeTeam:
Related issues in gobreaker:

- https://github.com/sony/gobreaker/issues/4
- https://github.com/sony/gobreaker/pull/6

aryszka:
sorry, didn't mean to delete those comments :face_with_head_bandage:

DeTeam:
👍

circuit/list.go
```diff
@@ -0,0 +1,77 @@
+package circuit
+
+// a simple list to keep breakers sorted by access order.
+// It can return the first consecutive items that match a
+// condition (in our case, the ones that were inactive for a
+// while).
+type list struct {
```
mo-gr: I think this is the first time ever since university, that I see a linked list implementation :-)

Anyway. Why not just use an array? This seems like premature optimisation and just adds fuckup-surface. Please consider just using an array. Yes, it might be slower, no it won't matter

proxy/proxy.go
```diff
@@ -470,6 +485,23 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, error) {
 	return response, nil
 }
 
+func (p *Proxy) checkBreaker(c *context) (func(bool), bool) {
+	if p.breakers == nil {
+		return nil, true
+	}
+
+	settings, _ := c.stateBag[circuitfilters.RouteSettingsKey].(circuit.BreakerSettings)
+	settings.Host = c.outgoingHost
```
mo-gr: Can I set breakers for `<shunt>` or `<loopback>`? Is that desirable behaviour? Or should we check and early escape in those cases?

mo-gr:
:+1:

mo-gr:
A general question, maybe more like a feature request: If a circuit breaker flips open, is it possible to purge the DNS cache for that host? Seems like a nice-to-have feature...

proxy/proxy.go
```diff
@@ -515,13 +547,26 @@ func (p *Proxy) do(ctx *context) error {
 		ctx.outgoingDebugRequest = debugReq
 		ctx.setResponse(&http.Response{Header: make(http.Header)}, p.flags.PreserveOriginal())
 	} else {
+		done, allow := p.checkBreaker(ctx)
+		if !allow {
+			return errCircuitBreakerOpen
```
mo-gr: Do I read this correct, that we will only log that a circuit breaker is open, but not which one? If a breaker flips open (or more general flips state), I would like to see a log line stating clearly, which breaker for which host goes from which old state into which new state.

Is this not there or am I missing something?

aryszka: valid point. It should be there.

Besides the logs, the plan is to add an endpoint to the "metrics" -> "support" listener that reports all the breaker states. After https://github.com/zalando/skipper/pull/364 gets finished.

aryszka:
purging the DNS cache. Will look into it.

proxy/proxy.go
aryszka: we cannot set breakers for <shunt> and <loopback> . I think the "shunt", by its name, already breaks the circuit by itself. For the loopback, not sure, but i believe it is enough to see where it loops to.

circuit/list.go
aryszka: i also been waiting long for the use case ;)

i think it is simply optimization, but not premature, since it is already in the PR. Using an array would be like not using anything, because then we could also just iterate over the timestamps to see which breaker is idle. The reason for the optimization here is that this is in a synchronized scope, happening on every request, and blocking every concurrent request. Why do you think that it won't matter?

aryszka: now that thinking about it.. maybe there is a way to bind the breakers to each route. Will play with that as a separate effort.

mo-gr:
:+1:

aryszka:
:+1:

DeTeam:
👍

Issue #320, opened 2017-05-10
List current rules


szuecs:
It would be nice to have an API endpoint to list the current rules from the management port for example:

    % curl localhost:9911/routes 
    ...skipper routes here...

Best would be if we could also specify the Content-Type: application/json for example.

jreyeshdez:
Skipper would need to have a new http Handler for this request but, what would it be its behavior if no route is defined in routes-file for /routes? Should it be bypassed somehow?

aryszka:
my understanding is that in this case it would just return an empty response, meaning there are no routes. This would be on a separate port and not on the main serving port, so no need to define this endpoint in the configuration for the routing.

szuecs:
Since we have :mgmtport/routes which lists routes in eskip format, I consider this as done.
In case you need json format please add a new issue referencing this one.

PR #364, opened 2017-07-07
List current rules 

haldun:
Initial work for for #320

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -110,8 +110,8 @@ func init() {
 	Default = Void
 }
 
-// Initializes the collection of metrics.
-func Init(o Options) http.Handler {
+// NewHandler returns a collection of metrics handlers.
+func NewHandler(o Options) http.Handler {
 	Default = New(o)
 
 	handler := &metricsHandler{registry: Default.reg, options: o}
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -9,18 +9,12 @@ import (
 	"time"
 
 	"github.com/rcrowley/go-metrics"
-	log "github.com/sirupsen/logrus"
 )
 
 type skipperMetrics map[string]interface{}
 
 // Options for initializing metrics collection.
 type Options struct {
-	// Network address where the current metrics values
-	// can be pulled from. If not set, the collection of
-	// the metrics is disabled.
-	Listener string
-
 	// Common prefix for the keys of the different
 	// collected metrics.
 	Prefix string
@@ -117,12 +111,7 @@ func init() {
 }
 
 // Initializes the collection of metrics.
-func Init(o Options) {
-	if o.Listener == "" {
-		log.Infoln("Metrics are disabled")
-		return
-	}
-
+func Init(o Options) http.Handler {
 	Default = New(o)
 
 	handler := &metricsHandler{registry: Default.reg, options: o}
@@ -136,8 +125,7 @@ func Init(o Options) {
 		handler.profile = mux
 	}
 
-	log.Infof("metrics listener on %s/metrics", o.Listener)
-	go http.ListenAndServe(o.Listener, handler)
+	return handler
 }
 
 func createTimer() metrics.Timer {
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -127,15 +127,15 @@ func applyIncoming(defs routeDefs, d *incomingData) routeDefs {
 
 // merges the route definitions from multiple data clients by route id
 func mergeDefs(defsByClient map[DataClient]routeDefs) []*eskip.Route {
-	mergeById := make(routeDefs)
+	mergeByID := make(routeDefs)
 	for _, defs := range defsByClient {
 		for id, def := range defs {
-			mergeById[id] = def
+			mergeByID[id] = def
 		}
 	}
 
 	var all []*eskip.Route
-	for _, def := range mergeById {
+	for _, def := range mergeByID {
 		all = append(all, def)
 	}
 
@@ -396,7 +396,7 @@ func receiveRouteMatcher(o Options, out chan<- *routeTable, quit <-chan struct{}
 
 			for _, err := range errs {
 				o.Log.Error(err)
-				invalidRouteIds[err.Id] = struct{}{}
+				invalidRouteIds[err.ID] = struct{}{}
 			}
 
 			for _, r := range routes {
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -369,9 +369,8 @@ func processRouteDefs(o Options, fr filters.Registry, defs []*eskip.Route) (rout
 
 type routeTable struct {
 	m             *matcher
-	validRoutes   []*Route
-	invalidRoutes []*Route
-	invalidDefs   []*eskip.Route
+	validRoutes   []*eskip.Route
+	invalidRoutes []*eskip.Route
 	created       time.Time
 }
 
@@ -389,12 +388,11 @@ func receiveRouteMatcher(o Options, out chan<- *routeTable, quit <-chan struct{}
 		select {
 		case defs := <-updatesRelay:
 			o.Log.Info("route settings received")
-			routes, invalidDefs := processRouteDefs(o, o.FilterRegistry, defs)
+			routes, invalidRoutes := processRouteDefs(o, o.FilterRegistry, defs)
 			m, errs := newMatcher(routes, o.MatchingOptions)
 
 			invalidRouteIds := make(map[string]struct{})
-			invalidRoutes := []*Route{}
-			validRoutes := []*Route{}
+			validRoutes := []*eskip.Route{}
 
 			for _, err := range errs {
 				o.Log.Error(err)
@@ -403,15 +401,14 @@ func receiveRouteMatcher(o Options, out chan<- *routeTable, quit <-chan struct{}
 
 			for _, r := range routes {
 				if _, found := invalidRouteIds[r.Id]; found {
-					invalidRoutes = append(invalidRoutes, r)
+					invalidRoutes = append(invalidRoutes, &r.Route)
 				} else {
-					validRoutes = append(validRoutes, r)
+					validRoutes = append(validRoutes, &r.Route)
 				}
 			}
 
 			rt = &routeTable{
 				m:             m,
-				invalidDefs:   invalidDefs,
 				validRoutes:   validRoutes,
 				invalidRoutes: invalidRoutes,
 				created:       time.Now().UTC(),
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -353,26 +353,26 @@ func mapPredicates(cps []PredicateSpec) map[string]PredicateSpec {
 }
 
 // processes a set of route definitions for the routing table
-func processRouteDefs(o Options, fr filters.Registry, defs []*eskip.Route) []*Route {
+func processRouteDefs(o Options, fr filters.Registry, defs []*eskip.Route) (routes []*Route, invalidDefs []*eskip.Route) {
 	cpm := mapPredicates(o.Predicates)
-
-	var routes []*Route
 	for _, def := range defs {
 		route, err := processRouteDef(cpm, fr, def)
 		if err == nil {
 			routes = append(routes, route)
 		} else {
+			invalidDefs = append(invalidDefs, def)
 			o.Log.Error(err)
 		}
 	}
-
-	return routes
+	return
 }
 
 type routeTable struct {
-	m       *matcher
-	routes  []*Route
-	created time.Time
+	m             *matcher
+	validRoutes   []*Route
+	invalidRoutes []*Route
+	invalidDefs   []*eskip.Route
+	created       time.Time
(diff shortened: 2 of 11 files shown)


metrics/metrics.go
```diff
@@ -117,12 +111,7 @@ func init() {
 }
 
 // Initializes the collection of metrics.
-func Init(o Options) {
-	if o.Listener == "" {
-		log.Infoln("Metrics are disabled")
-		return
-	}
-
+func Init(o Options) http.Handler {
```
aryszka: Init -> NewHandler?

routing/datasource.go
```diff
@@ -389,12 +394,12 @@ func receiveRouteMatcher(o Options, out chan<- *matcher, quit <-chan struct{}) {
 			for _, err := range errs {
 				o.Log.Error(err)
 			}
-
-			mout = m
+			// TODO: What about the errors? Should we exclude the routes which has errors?
```
aryszka: i think we should show first the applied routes, so yes. In another wave, we can add the routes that have errors.

aryszka: though, after thinking about it, we should be careful with the slices and maps, because of the memory. Do you know if this current routes slice is thrown away after the update? If not, then there is somewhere a map containing the routes, that could be used. and maybe with a container struct, each of them could be marked depending on whether they errored or succeeded.

routing/routing.go
```diff
@@ -154,19 +158,88 @@ func New(o Options) *Routing {
...
+			return
+		}
+		if lim < 0 {
+			http.Error(w, "invalid limit", http.StatusBadRequest)
+			return
+		}
+		limit = lim
+	}
+
+	accept := req.Header.Get("accept")
```
aryszka: we're generally trying to use the canonical form of the header names: "Accept"

routing/routing.go
```diff
@@ -154,19 +158,88 @@ func New(o Options) *Routing {
...
+			return
+		}
+		if off < 0 {
+			http.Error(w, "invalid offset", http.StatusBadRequest)
+			return
+		}
+		offset = off
+	}
+
+	limit := 75
```
aryszka: how about some larger value? 1024?

routing/datasource.go
haldun: How about changing `processRouteDefs` so that it will return a slice of `*Route` (i.e. no error) and a slice of `*eskip.Route` which contains the faulty ones?

aryszka: sounds good

haldun: There are two validations though, `newMatcher` also validates the results of `processRouteDefs`. Should we consider both?

aryszka: as i understand, there is the syntax validation and then the semantic to a certain extent, e.g. the reference to a missing filter. How the users would see it from the outside, i believe: i put in these routes, and i want to see which ones were applied, which ones failed. So yes, i think we need to consider both kinds of failures.

routing/datasource.go
```diff
@@ -353,48 +353,73 @@ func mapPredicates(cps []PredicateSpec) map[string]PredicateSpec {
...
+			routes, invalidDefs := processRouteDefs(o, o.FilterRegistry, defs)
 			m, errs := newMatcher(routes, o.MatchingOptions)
+
+			invalidRouteIds := make(map[string]struct{})
+			invalidRoutes := []*Route{}
+			validRoutes := []*Route{}
+
 			for _, err := range errs {
 				o.Log.Error(err)
+				invalidRouteIds[err.Id] = struct{}{}
```
aryszka: maybe we could just change the definitionError structure to contain the route and not only the id. Then the invalid routes could be taken from the error object

routing/routing.go
```diff
@@ -154,19 +161,71 @@ func New(o Options) *Routing {
...
+		http.Error(w, "invalid limit", http.StatusBadRequest)
+		return
+	}
+	accept := req.Header.Get("Accept")
+	routes := slice(rt.validRoutes, offset, limit)
+
+	if accept == "application/json" {
+		w.Header().Set(routeTableTimestampHeaderName, createdUnix)
+		w.Header().Set("content-type", "application/json")
+		if err := json.NewEncoder(w).Encode(routes); err != nil {
```
aryszka: does this work like this? I think that we should only serialize the eskip.route. The routing.route should stay in-memory object, because even if now we don't have anything that would prevent its serialization, this wouldl pose a limitation for the future changes.

routing/datasource.go
```diff
@@ -353,48 +353,73 @@ func mapPredicates(cps []PredicateSpec) map[string]PredicateSpec {
...
 		}
 	}
+	return
+}
 
-	return routes
+type routeTable struct {
+	m             *matcher
+	validRoutes   []*Route
+	invalidRoutes []*Route
```
aryszka: i think both the validRoutes and the invalidRoutes should be only the slice of the eskip.Route type. Invalid routes and invalidDefs could be a single slice

routing/routing.go
```diff
@@ -154,19 +161,71 @@ func New(o Options) *Routing {
...
+		http.Error(w, "invalid offset", http.StatusBadRequest)
+		return
+	}
+	limit, err := extractParam(req, "limit", defaultRouteListingLimit)
+	if err != nil {
+		http.Error(w, "invalid limit", http.StatusBadRequest)
+		return
+	}
+	accept := req.Header.Get("Accept")
+	routes := slice(rt.validRoutes, offset, limit)
```
lmineiro: Is this a safe copy of the "live" routing table?

routing/routing.go
```diff
@@ -154,19 +161,71 @@ func New(o Options) *Routing {
...
+	}
+
+	if accept == "application/eskip" {
+		w.Header().Set(routeTableTimestampHeaderName, createdUnix)
+		w.Header().Set("content-type", "application/eskip")
+		eskipRoutes := make([]*eskip.Route, len(routes))
+		for i, r := range routes {
+			eskipRoutes[i] = &r.Route
+		}
+		fmt.Fprint(w, eskip.String(eskipRoutes...))
```
lmineiro: Too late to implement Stringer in those types?

routing/datasource.go
haldun: However we still need to extract valid and invalid routes, so I'm not sure returning invalid Route will make the need for this map to go away.

aryszka: but the routes don't contain the invalid routes, or am i wrong? https://github.com/zalando/skipper/blob/master/routing/datasource.go#L363

aryszka: i see now, ok

routing/routing.go
aryszka: the route type implements the stringer, but here we stringify the slice, with separators. I believe this is fine.

routing/routing.go
aryszka: they are not deep copy. But since the last change, they are only the eskip definitions, and here they are only being read. Do you think there can be any issues with it?

routing/routing.go
```diff
@@ -154,19 +161,73 @@ func New(o Options) *Routing {
...
+	limit, err := extractParam(req, "limit", defaultRouteListingLimit)
+	if err != nil {
+		http.Error(w, "invalid limit", http.StatusBadRequest)
+		return
+	}
+	accept := req.Header.Get("Accept")
+	routes := slice(rt.validRoutes, offset, limit)
+
+	// TODO: Maybe this is too strict?
+	if accept == "application/json" {
```
aryszka: i think the accept mime should work like this:

1. if it contains application/json, return json
2. if it contains application/eskip, plain/text, plain/*, */* or not set at all, return eskip
3. if it is set to anything else return 406

routing/routing.go
```diff
@@ -154,19 +161,73 @@ func New(o Options) *Routing {
...
+	if err != nil {
+		http.Error(w, "invalid offset", http.StatusBadRequest)
+		return
+	}
+	limit, err := extractParam(req, "limit", defaultRouteListingLimit)
+	if err != nil {
+		http.Error(w, "invalid limit", http.StatusBadRequest)
+		return
+	}
+	accept := req.Header.Get("Accept")
```
aryszka: i think the accept header should be parsed. We can either use sg like: https://github.com/golang/gddo/blob/master/httputil/header/header.go
(which would mean a new dependency, but tbh it's nice and elaborate)

Or we could also just split by `,` then trimspace, then split by `;` then take the first and trim space, getting a slice this way and check which values it contains.

routing/routing.go
```diff
@@ -154,19 +161,73 @@ func New(o Options) *Routing {
...
+
+	// TODO: Maybe this is too strict? Also, what is the correct MIME type for eskip?
+	if accept == "application/eskip" {
+		w.Header().Set(routeTableTimestampHeaderName, createdUnix)
+		w.Header().Set("content-type", "application/eskip")
+		eskipRoutes := make([]*eskip.Route, len(routes))
+		for i, r := range routes {
+			eskipRoutes[i] = r
+		}
+		fmt.Fprint(w, eskip.String(eskipRoutes...))
```
aryszka: what do you think about pretty printing it? https://godoc.org/github.com/zalando/skipper/eskip#Print then we could just see it in a browser, too.

routing/routing.go
lmineiro: I'm willing to suggest just returning eskip by default (catch all)

routing/routing.go
lmineiro: This sounds like a good idea. I'd even go further and see how `ls` understands that it's output is being piped to another process and change the behavior so that chaining multiple commands work "well" in the unix world

routing/routing.go
haldun: What should be the content-type for eskip? Setting it to `application/eskip` makes browser to download the data into a file, so maybe `text/plain` is better.

routing/routing.go
haldun: Pretty print by default or optional controlled by a query param?

routing/routing.go
haldun: If we render eskip by default, then maybe we don't need to parse accept header. Checking whether it contains "application/json" is enough.

aryszka: yeah.. agree, ok. text/plain, and simple check for application/json

routing/routing.go
aryszka: i prefer default.

for piping, however, it is better when a rule is in a single line, so it's nice to have a query param for disabling it, like ?nopretty, or ?nopretty=, or ?nopretty=X where X can be anything but 0 or false, because then in those latter cases, it should be explicitly pretty. (No need to implement in full, these are only suggestions ;) )

lmineiro: Not a fan of the isatty() idea for piping?
Something like https://github.com/mattn/go-isatty would make it a no brainer...
ls changes into "ls -1" automatically when you pipe, I always liked that behavior without me being explicit about it

aryszka:
:+1:

aryszka:
:+1:

DeTeam:
👍

routing/routing.go
```diff
@@ -175,17 +225,59 @@ func (r *Routing) startReceivingUpdates(o Options) {
...
+func slice(r []*eskip.Route, offset int, limit int) []*eskip.Route {
+	if offset > len(r) {
+		offset = len(r)
+	}
+	end := offset + limit
+	if end > len(r) {
+		end = len(r)
+	}
+	result := r[offset:end]
+	if result == nil {
```
szuecs: I think this condition will never happen.

[2] go-pry> a = []string{"foo"}
=> []string{"foo"}
[1] go-pry> a[0:0]
=> []string{}
[4] go-pry> a[0:0] == nil
=> false

routing/routing.go
```diff
@@ -175,17 +225,59 @@ func (r *Routing) startReceivingUpdates(o Options) {
...
 
-// Closes routing, stops receiving routes.
+// Close closes routing, stops receiving routes.
 func (r *Routing) Close() {
 	close(r.quit)
 }
+
+func slice(r []*eskip.Route, offset int, limit int) []*eskip.Route {
+	if offset > len(r) {
+		offset = len(r)
```
szuecs: this should be an off by one issue

routing/routing.go
aryszka: it's always better to check len(a) == 0 instead of a == nil

aryszka: figured.. when i am trying just simply in go it seems like go-pry doesn't work exactly the same way:

```
var a []int
println(a == nil, a[0:0] == nil)
// Output: true true
```

and:

```
var a = []string{"foo"}
println(a[0:0] == nil, a[1:1] == nil)
// Output: false false
```

The reason to do this nil check, I assume, is to avoid sending null as the json response instead of an empty array in cases when the input routes are nil for any reason.

routing/routing.go
aryszka: see the a[1:1] example below

routing/routing.go
szuecs: True and since the caller has to check, that he does not access out of bounds, it's ok.

Issue #366, opened 2017-07-10
migrate skoap into skipper


aryszka:
Migrating Skoap [https://github.com/zalando-incubator](https://github.com/zalando-incubator) would make a lot of sense because many default scenarios of skipper could benefit from it, while after migration, skipper could be used as a drop-in replacement of Skoap, without loosing any feautres.

Issues to take care of:
- verify that the authentication api that Skoap is calling is following the oauth2 standards
- define a formal web hook interface to support the current team service dependency in Skoap so that any user can plug in their own implementation or provide an API translation layer if the team service is required
- rename 'team' to 'group' so that it has a more generic meaning

Plan:
1. in the first step migrate the tests and the implementation of Skoap with small changes to skipper, ideally the auth filter under the package filters/oauth (open for better naming), and the command under cmd/skoap (let's keep the legacy name). The audit filter should go to its own package, if there's no better suggestion then under filters/audit. Mark this phase as experimental, because it is possible that backwards incompatible changes will be required in the further steps. At the same time, give an early notice in the Skoap repository about the migration. Suggested deadline before Skoap gets deleted: 2017.08.15
2. consider saving over the work in this WIP PR of Skoap: https://github.com/zalando-incubator/skoap/pull/12 . This should provide browser support with authentication redirects to an oauth provider for human users. (Check other similar efforts: i am aware of a less general, but already working approach that can be useful to add to skipper, but infortunately not yet open-source.)
3. at this phase it should be possible to delete the Skoap repo. RIP
4. finalize the oauth API, and settle the group web hook definition for the group service

aryszka:
further improvements after phase 3:
- add proper timeouts for the service calls

aryszka:
WIP PR here: https://github.com/zalando/skipper/pull/367

szuecs:
I think we have everything we need for oauth in the mentioned cases and with https://github.com/zalando/skipper/issues/744 everyone can build there "group" filter.

PR #367, opened 2017-07-10
raw migration of the skoap code to skipper

aryszka:
migrating skoap [https://github.com/zalando/skoap](https://github.com/zalando/skoap) to skipper - Phase 1 https://github.com/zalando/skipper/issues/366

Missing from Phase 1:
- [x] update docs: main doc.go
- [x] fill missing tests
- [x] log/audit log: it is ok to do this as a separate feature
- [x] check which basic auth is better the one in Skoap or the one in Skipper
- [x] figure a good configuration system for the auth and group services: typically one is enough, but allow multiple ones
- [x] drop the separate skoap command

Dropping the skoap command is fine, i believe, because inline routing config was added recently as a new startup feature, and it enables the main use case os the skoap command: just start an auth proxy as a single command.


aryszka:
:-1:

aryszka:
@szuecs @Raffo FYI :smile:

cmd/skoap/main.go
```diff
@@ -0,0 +1,318 @@
...
+
+	experimentalUpgradeFlag = "experimental-upgrade"
+)
+
+const (
+	usageHeader = `
+skoap - Skipper based reverse proxy with authentication.
+
+Use the skoap proxy to verify authorization tokens before forwarding requests, and optionally check OAuth2 realms
+and scoap or team membership. In addition to check incoming requests, optionally set basic authorzation headers
```
lmineiro: and scope

lmineiro: s/authorzation/authorization

cmd/skoap/main.go
```diff
@@ -0,0 +1,318 @@
...
+	} else {
+		var filterArgs []interface{}
+		if realm != "" {
+			filterArgs = append(filterArgs, realm)
+		}
+
+		args := scopes
+		name := skoap.AuthName
+		if teamCheckMode {
+			args = teams
```
lmineiro: Does this mean scopes and teams can't be combined? The docs above aren't clear about this

aryszka: good catch! But at least it only applies to the "single-route" mode. When used with eskip files or etcd then they can be combined. I would add this fix to Phase 2 (https://github.com/zalando/skipper/issues/366)

aryszka:
Question to @lmineiro @Raffo @szuecs @gargravarr : i think we should use the term "group" instead of "team". Simply because it is somewhat more generic. Is this renaming fine?

cmd/skoap/main.go
aryszka: not even fully sure if we should keep this separate command. Does it provide that much value?

lmineiro: I agree that it can be confusing to provide both

lmineiro:
I'm neutral about the naming for teams/groups/communities. I'd go with the most common terminology out there for the same scope (pun intended)

cmd/skoap/main.go
```diff
@@ -0,0 +1,318 @@
...
+
+	experimentalUpgradeFlag = "experimental-upgrade"
+)
+
+const (
+	usageHeader = `
+skoap - Skipper based reverse proxy with authentication.
+
+Use the skoap proxy to verify authorization tokens before forwarding requests, and optionally check OAuth2 realms
+and scope or team membership. In addition to check incoming requests, optionally set basic authorization headers
```
Joneser: "check" -> "checking" sounds better here I think

cmd/skoap/main.go
```diff
@@ -0,0 +1,318 @@
...
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/filters/builtin"
+	"github.com/zalando/skipper/proxy"
+	"github.com/zalando/skipper/routing"
+)
+
+const (
+	addressFlag        = "address"
+	defaultEtcdPrefix  = "/skipper"
```
szuecs: Should skoap have a different default prefix?
Maybe you want to run skipper and skoap with one etcd?

cmd/skoap/main.go
```diff
@@ -0,0 +1,318 @@
...
+  authorization parameters are set in the routes file with the auth and authTeam filters.
+
+When used with eskip configuration files, it is possible to apply detailed augmentation of the requests and
+responses using Skipper rules.
+
+https://github.com/zalando/skipper
+
+`
+
+	addressUsage    = `network address that skoap should listen on`
```
szuecs: Why the quotes are different on this line and below?
I would use "" and not ``

cmd/skoap/main.go
```diff
@@ -0,0 +1,318 @@
...
+	auditBodyUsage = `set the limit of the audit log body`
+
+	routesFileUsage = `alternatively to the target address, it is possible to use a full eskip route
+configuration, and specify the auth() and authTeam() filters for the routes individually. See also:
+https://godoc.org/github.com/zalando/skipper/eskip`
+
+	insecureUsage = `when this flag set, skipper will skip TLS verification`
+
+	authUrlBaseUsage = `URL base of the authentication service. The authentication token found
+in the incoming requests will be validated agains this service. It will be passed as the Authorization Bearer
```
szuecs: agains -> against

cmd/skoap/main.go
```diff
@@ -0,0 +1,318 @@
...
+header`
+
+	teamUrlBaseUsage = `URL base of the team service. The user id received from the authentication service will
+be appended to this url, and the list of teams that the user is a member of will be requested`
```
szuecs: An example would be good here.
Example URL: https://my-group-svc.example.org/:user , where :user will be replaced by the user got from the authentication service, for example jdoe.

cmd/skoap/main.go
```diff
@@ -0,0 +1,318 @@
...
+
+	// TODO
+	certPathTLSUsage = "path of the certificate file"
```
szuecs: Should the cert also contain the CA?
This is important to document.

cmd/skoap/main.go
```diff
@@ -0,0 +1,318 @@
...
+authenticated against an OAuth2 realm, set the value of the realm with this flag. Note, that in case of a routes
+file is used, the realm can be set for each auth filter reference individually`
+
+	scopesUsage = `a comma separated list of the OAuth2 scopes to be checked in addition to the token validation
+and the realm check`
+
+	teamsUsage = `a comma separated list of the teams to be checked in addition to the token validation and the
+realm check`
+
+	auditUsage = `enable audit log in single route mode`
```
szuecs: Why this is only for single route mode?

cmd/skoap/main.go
szuecs: I am also not sure if you write in the docs about "single-route" mode, but no user can choose to run a "single-route" mode. What does a "single-route" mode even mean for our users?
The meaning in code (`singleRouteMode := targetAddress != ""`) is clear, but not how the user specifies it?

cmd/skoap/main.go
```diff
@@ -0,0 +1,318 @@
...
+			name = skoap.AuthTeamName
+		}
+
+		if args != "" {
+			if realm == "" {
+				// realm set to empty
+				filterArgs = append(filterArgs, "")
+			}
+
+			argss := strings.Split(args, ",")
```
szuecs: argss -> argsList or argsSlice :)

filters/auth/oauth.go
```diff
@@ -0,0 +1,343 @@
...
+
+	filter struct {
+		typ         roleCheckType
+		authClient  *authClient
+		groupClient *groupClient
+		realm       string
+		args        []string
+	}
+
+	// basic string
```
szuecs: Could be cleaned up

filters/auth/oauth.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	return ts, nil
+}
+
+func newSpec(typ roleCheckType) filters.Spec {
+	return &spec{typ: typ}
+}
+
+// Creates a new auth filter specification to validate authorization
```
szuecs: godoc forces you to start with function name, so:
"NewAuth creates a new filter specification..."

filters/auth/oauth.go
```diff
@@ -0,0 +1,343 @@
...
+// Authorization header and in case of a valid token, it expects it
+// to return the user id and the realm of the user associated with
+// the token ('uid' and 'realm' fields in the returned json document).
+// The token is set as the Authorization Bearer header.
+//
+func NewAuth() filters.Spec {
+	return newSpec(checkScope)
+}
+
+// Creates a new auth filter specification to validate authorization
```
szuecs: "NewAuthGroup creates a new filter specification to validate..."

filters/auth/oauth.go
```diff
@@ -0,0 +1,343 @@
...
+func (f *filter) Request(ctx filters.FilterContext) {
+	r := ctx.Request()
+
+	token, err := getToken(r)
+	if err != nil {
+		unauthorized(ctx, "", missingBearerToken)
+		return
+	}
+
+	a, err := f.authClient.validate(token)
```
szuecs: "a" is an "authDoc" right?
I would like to have more clear variable name here, because it's not an index variable and it's derived from a call to a method, where the sicgnature is not easy to understand from this point in code.

filters/auth/oauth.go
```diff
@@ -0,0 +1,343 @@
...
+	} else if !valid {
+		unauthorized(ctx, a.Uid, invalidGroup)
+	} else {
+		authorized(ctx, a.Uid)
+	}
+}
+
+func (f *filter) Response(_ filters.FilterContext) {}
+
+// // Creates basicAuth filter specification.
```
szuecs: Do you want to delete this commented code or do you want to provide this code in the PR?

filters/log/log.go
```diff
@@ -0,0 +1,145 @@
...
+	if wl >= tb.maxTee {
+		wl = tb.maxTee
+	}
+
+	n, err := tb.buffer.Write(b[:wl])
+	if err != nil {
+		return n, err
+	}
+
+	tb.maxTee -= n
```
szuecs: Why do you have to subtract n from maxTee and set it?
This seems to be the state pointer of teeBody, if Write will be called more than once. But on the other hand it's defining the upper limit read from b and write to buffer.

filters/log/log.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// lie to avoid short write
```
szuecs: Why is it needed to "lie" here?

filters/log/log.go
```diff
@@ -0,0 +1,145 @@
...
+	return len(b), nil
+}
+
+// Creates an auditLog filter specification. It expects a writer for
```
szuecs: "NewAuditLog creates an auditLog filter specification..."

filters/log/log.go
```diff
@@ -0,0 +1,145 @@
...
+
+		if tb.buffer.Len() > 0 {
+			doc.RequestBody = tb.buffer.String()
+		}
+	}
+
+	enc := json.NewEncoder(al.writer)
+	err := enc.Encode(&doc)
+	if err != nil {
+		log.Println(err)
```
szuecs: It would be nice to have a better error message, that you can grep more easily.
Maybe something like:

    log.Printf("Failed to json encode auditDoc: %v", err)

Another question is why you choose to use Printf and not Errorf?

cmd/skoap/main.go
aryszka: ok, thanks, won't keep the separate command

cmd/skoap/main.go
aryszka: right. Typically, there aren't a lot of etcds around, only one. On the other hand, I think it's better to remove the separate skoap command.

cmd/skoap/main.go
aryszka: good point

filters/auth/oauth.go
aryszka: right

filters/auth/oauth.go
aryszka: i think we can delete it, i don't think it is very useful. The same thing can be done with the setRequestHeader filter.

filters/log/log.go
aryszka: yeah, it could have a better name. It is used basically just for a countdown.

This whole teeBody could be written a bit smarter. On the other hand, I am not sure if it should be kept, or rather something shared should be figured with the existing tee() filter's internals.

filters/log/log.go
aryszka: true.

(the only reason that it was Print here is that skoap didn't use logrus, only the stdlib log, and it wasn't updated yet)

Raffo:
It looks like the automerge screwed up a bunch of things, I'll get a shot at fixing that later.

gargravarr:
There is also the need to implement this feature so we can replace the proprietary OpenIGs from the DCs.

aryszka:
@Raffo @gargravarr , i updated the tasks in the PR

szuecs:
@gargravarr can you update the branch with your local changes, please?

HerrmannHinz:
yes please +1

szuecs:
I talked with @gargravarr and he is working on another branch: https://github.com/zalando/skipper/tree/oauth-filter

@HerrmannHinz What kind of OAuth2/JWT IDP do you want to use?
I want to understand what makes sense for you.

HerrmannHinz:
@szuecs currently we are evaluating if keycloak can be used as an authn/authz endpoint/provider for our internal ci/cd infrastructure. it will act as an openidc/saml provider. 

have a look here: 
https://eclipsesource.com/blogs/2018/01/11/authenticating-reverse-proxy-with-keycloak/
... this is what we want to archive. **SSO via an Authentication Reverse Proxy** (auth token via http-header). 

for some of the common ci/cd tools there are openidc plugins, for some don't. 

aprox. these toolset needs to be tackled: 

- graylog >> http://docs.graylog.org/en/2.1/pages/users_and_roles/external_auth.html#single-sign-on
- jenkins >> https://plugins.jenkins.io/reverse-proxy-auth-plugin (something similar like the graylog approach) 
- nexus >> https://help.sonatype.com/repomanager2/configuration/authentication-via-remote-user-token
- sonarqube >> https://docs.sonarqube.org/display/SONAR/HTTP+headers
- maybe selenium-hub
- drone? portus? concourse?
- grafana >> https://grafana.com/blog/2015/12/07/grafana-authproxy-have-it-your-way/
- pcf >> https://docs.pivotal.io/pivotalcf/1-12/opsguide/auth-sso.html

i think u get the idea :)

szuecs:
@HerrmannHinz Does it makes sense to you to have a default auth provider set via cli flags and overwrite that using the filter annotation and use a filter parameter to choose a different one or should this be explicit for each one set in ingress?

HerrmannHinz:
@szuecs uhm. might be possible that we would need to name a specific authn realm for the different ingresses. setting an endpoint plus certain parameters for each ingress would give more flexibility. what do you think?

szuecs:
@HerrmannHinz I don't understand why you need to pass a secret. Normally the request has to sent a secret which you only want to validate, which does not need to be authenticated with another secret.
Which OAuth provider needs the client secret of the "OAuth proxy"?

szuecs:
The first goal of this PR will be to enable plain Bearer tokens in the Authorization header.
Implicit flow like shown in http://oauthlib.readthedocs.io/en/latest/oauth2/grants/implicit.html are not supported with this. I created the issues #624

szuecs:
We use https://godoc.org/golang.org/x/oauth2#Config to configure via CLI defaults (done TokenURL) that can be overridden by filter configuration (not done yet)

I refactored the filter code, not done is the audit log option and other things from skoap, that might be useful in general.

szuecs:
- base code is ready to review
- open question which basic auth filter to use (I did not looked at it)
- missing part are docs

szuecs:
@aryszka @vetinari if you like you can start to review this

filters/auth/doc.go
```diff
@@ -1,18 +1,128 @@
...
+check if it has at least one of the predefined scopes. If any of the
+expectations are not met, it doesn't forward the request to the target
+endpoint, but returns with status 401.
+
+As additional features, the package also supports audit logging.
+
+OAuth - Provider Configuration
+
+To enable OAuth2 filters you have to set the CLI argument
+-token-url=<TokenURL>.  Scopes and realms are dependend on your OAuth2
```
aryszka: "Scopes and realms depend on the OAuth2 provider."

filters/auth/doc.go
```diff
@@ -1,18 +1,128 @@
...
+infrastructure provider. Tokens can be OAuth2 or JWT as long as
```
aryszka: This part: "Tokens can be OAuth2 or JWT as long  as TokenURL returns the data being checked." is not clear to me. Does it mean the format? And that the oauth provider accepts them?

filters/auth/oauth.go
```diff
@@ -0,0 +1,429 @@
...
+}
+
+func (ac *authClient) getTokeninfo(token string) (map[string]interface{}, error) {
+	var a map[string]interface{}
+	err := jsonGet(ac.urlBase, token, &a)
+	return a, err
+}
+
+func newSpec(typ roleCheckType, cfg oauth2.Config) filters.Spec {
+	return &spec{typ: typ, cfg: cfg}
```
aryszka: i think it would be better if newSpec validated the `typ`, and returned an error if it is checkUnknown, because it never can be valid. Then the check in CreateFilter is not necessary.

filters/auth/oauth.go
```diff
@@ -0,0 +1,429 @@
...
+
+// Options to configure auth providers
+type Options struct {
+	// TokenURL is the tokeninfo URL able to return information
+	// about a token.
+	TokenURL string
+	// AuthType is the type of authnz function you want to
+	// use. Examples are the values "authAnyScope" or "authAllScope",
+	// defined in constants AuthAnyScopeName and AuthAllScopeName.
+	AuthType string
```
aryszka: We should make `roleCheckType` exported, and accept enumerated values in the options instead of strings, because the options are used programmatically.

Another way would to just have separate NewAuthAny, NewAuthAll... constructors. See the header filter as an example: https://github.com/zalando/skipper/blob/master/filters/builtin/headerfilter.go#L72

filters/auth/oauth.go
```diff
@@ -0,0 +1,429 @@
...
+	invalidScope       rejectReason = "invalid-scope"
+)
+
+// TODO: discuss these names, because these are the filter names used by the endusers
+const (
+	AuthAnyScopeName = "authAnyScope"
+	AuthAllScopeName = "authAllScope"
+	AuthAnyKVName    = "authAnyKV"
+	AuthAllKVName    = "authAllKV"
+	AuthUnknown      = "authUnknown"
```
mikkeloscar: I think `oauth2` should be in the names. `auth` is a bit too generic.

skipper.go
```diff
@@ -407,6 +409,12 @@ type Options struct {
 	// the last item of the string list of the X-Forwarded-For
 	// header, in this case you want to set this to true.
 	ReverseSourcePredicate bool
+
+	// TokenURL sets the TokenURL similar to https://godoc.org/golang.org/x/oauth2#Endpoint
```
vetinari: The endpoints in https://godoc.org/golang.org/x/oauth2#Endpoint are the authentication and token endpoints, but with this one you want the non-standard tokeninfo endpoint, right? What about servers without  a tokeninfo endpoint (e.g. https://github.com/coreos/dex/)

filters/auth/doc.go
szuecs: The TokenURL service has to accept them, but as @vetinari pointed out this might be a very specific choice to our environment. I try to talk with @vetinari to make this more generic.

filters/auth/doc.go
```diff
@@ -1,18 +1,128 @@
...
+TokenURL returns the data being checked.
+
+OAuth - authAnyScope() filter
+
+The filter authAnyScope allows access if the realm and one of the scopes
+are satisfied by the request.
+
+    Path("/uid") -> authAnyScope("/employees", "uid") -> "https://internal.example.org/";
```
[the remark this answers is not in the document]
aryszka: named parameters are not supported. We could use rather different filter names. One would ignore the realm, the other would always expect the realm as the first argument. Sg like:

```
authAnyScope("uid", "service-read")
```

and

```
authRealmAnyScope("/employees", "uid", "service-read")
```

What do you think?

szuecs:
I looked more deeply into tokeninfo and token_introspection. For the current PR I discussed with @vetinari and some others to understand possibilities to make tokeninfo usable for everyone. Google for example has also tokeninfo https://www.googleapis.com/oauth2/v1/tokeninfo?access_token=myaccessToken . It does not seem to be a standard, but more implementations exist. Common is that the accesstoken is passed as query string with key "access_token", which I would use in this PR. A default tokeninfo endpoint has to be passed as CLI argument to skipper or a user can pass it as argument into a filter to change the default.

filters/auth/oauth.go
```diff
@@ -0,0 +1,429 @@
...
+		return "", errInvalidAuthorizationHeader
+	}
+
+	return h[len(b):], nil
+}
+
+func unauthorized(ctx filters.FilterContext, uname string, reason rejectReason) {
+	ctx.StateBag()[logfilter.AuthUserKey] = uname
+	ctx.StateBag()[logfilter.AuthRejectReasonKey] = string(reason)
+	ctx.Serve(&http.Response{StatusCode: http.StatusUnauthorized})
```
aryszka: wondering if we want to fill the `WWW-Authenticate` header, too.

filters/auth/oauth.go
```diff
@@ -0,0 +1,429 @@
...
+// right. Right can be a superset of left.
+func all(left, right []string) bool {
+	if len(left) > len(right) {
+		return false
+	}
+
+	for _, l := range left {
+		for _, r := range right {
+			if l != r {
+				return false
```
aryszka: this will return false for any larger r than len=1

filters/auth/oauth.go
```diff
@@ -0,0 +1,429 @@
...
+	if !ok {
+		return false
+	}
+	v, ok := vI.([]interface{})
+	if !ok {
+		return false
+	}
+	var a []string
+	for i := range v {
+		a = append(a, v[i].(string))
```
aryszka: it should be checked with `vi, ok := v[i].(string)`

filters/auth/oauth.go
```diff
@@ -0,0 +1,429 @@
...
```
aryszka: it should be checked with vi, ok := v[i].(string)

filters/auth/oauth.go
aryszka: for handling if it's not a string: it even can be a good way to convert it into a string. What do you think?

filters/auth/oauth.go
szuecs: will be done in the next push

filters/auth/doc.go
szuecs: will be done in the next push

filters/auth/doc.go
szuecs: will be done in the next push

filters/auth/oauth.go
```diff
@@ -0,0 +1,429 @@
...
+		return fmt.Sprintf("%s(%s,%s)", AuthAllScopeName, f.realm, strings.Join(f.scopes, ","))
+	case checkAnyKV:
+		return fmt.Sprintf("%s(%s,%s)", AuthAnyKVName, f.realm, f.kv)
+	case checkAllKV:
+		return fmt.Sprintf("%s(%s,%s)", AuthAllKVName, f.realm, f.kv)
+	}
+	return AuthUnknown
+}
+
+func (f *filter) validateRealm(h map[string]interface{}) bool {
```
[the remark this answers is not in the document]
szuecs: This will addressed as @aryszka suggested above, create 2 checks (new name style):

    oauthTokeninfoAnyScope("uid", "service-read")
    oauthTokeninfoRealmAnyScope("/employees", "uid", "service-read")

filters/auth/oauth.go
szuecs: fixed

skipper.go
szuecs: I don't use this type anymore, thanks for pointing this out.

szuecs:
@aryszka @vetinari this is ready for review in my opinion.
I removed the realm checks completely, because they can be checked by the `*KV("realmKey", "realmVal")` filters. I also made `uid` non mandatory, but try to look it up in case we have it we can pass it to the `auditLog` filter via statebag.

vetinari:
LGTM

filters/auth/doc.go
```diff
@@ -1,18 +1,162 @@
...
-Usage
+Embedding the filter in routes:
+
+	generic: * -> basicAuth("/path/to/htpasswd") -> "https://internal.example.org";
+
+	myRealm: Host("my.example.org")
+	  -> basicAuth("/path/to/htpasswd", "My Website")
+	  -> "https://my-internal.example.org";
+
+Set Basic - Outgoing Authentication Header
```
aryszka: I think the outgoing basic auth was a small feature of skoap, and we are not migrating that in this phase. Please remove this 3 lines from the doc.

filters/auth/doc.go
```diff
@@ -1,18 +1,162 @@
...
+
+The auth filter takes the incoming request, and tries to extract the
+Bearer token from the Authorization header. Then it validates against
+a configured service. Depending on the settings, it also can check if
+the owner of the token belongs to a specific OAuth2 realm, and it can
+check if it has at least one of the predefined scopes. If any of the
+expectations are not met, it doesn't forward the request to the target
+endpoint, but returns with status 401.
+
+As additional feature, the package also supports audit logging.
```
aryszka: i think the audit logging could be mentioned separately after the oauth2 sections in the doc

filters/auth/doc.go
```diff
@@ -1,18 +1,162 @@
...
+To enable OAuth2 tokeninfo filters you have to set the CLI argument
+-oauth2-tokeninfo-url=<OAuthTokeninfoURL>. Scopes and key value pairs
+depend on the OAuth2 tokeninfo provider. AccessTokens has to be
+accepted by your OAuth2 provider's TokeninfoURL. Filter names starting
+with oauthTokeninfo will work on the returned data from
+TokeninfoURL. The request from skipper to TokeninfoURL will use the
+query string to do the request:
+?access_token=<access-token-from-authorization-header>.
+
+Example json output of this tokeninfo could be:
```
aryszka: "of this tokeninfo" -> "of the tokeninfo response"

filters/auth/doc.go
```diff
@@ -1,18 +1,162 @@
...
+        "bar-w",
+        "qux-rw"
+      ],
+      "token_type": "Bearer",
+      "uid": "jdoe"
+    }
+
+OAuth2 - oauthTokeninfoAnyScope filter
+
+The filter oauthTokeninfoAnyScope allows access if one of the scopes
```
aryszka: "one of the scopes are" -> "one of the scopes is"

filters/auth/doc.go
```diff
@@ -1,18 +1,162 @@
...
+are satisfied by the request.
+
+    Path("/a") -> oauthTokeninfoAnyScope("uid") -> "https://internal.example.org/";
```
aryszka: when we have multiple routes in the same eskip doc separated by `;`, then the route id is mandatory. E.g:

```
a: Path("/a") -> oauthTokeninfoAnyScope("uid") -> "https://internal.example.org/";
b: Path("/b") -> oauthTokeninfoAnyScope("uid", "bar") -> "https://internal.example.org/";
```

filters/auth/doc.go
```diff
@@ -1,18 +1,162 @@
...
+    Path("/b") -> oauthTokeninfoAnyScope("uid", "bar") -> "https://internal.example.org/";
+
+OAuth - oauthTokeninfoAllScope() filter
+
+The filter oauthTokeninfoAllScope allows access if all of the scopes
+are satisfied by the request:
+
+    Path("/a") -> oauthTokeninfoAllScope("uid") -> "https://internal.example.org/";
```
aryszka: route id:

```
a: Path("/a") -> oauthTokeninfoAnyScope("uid") -> "https://internal.example.org/";
b: Path("/b") -> oauthTokeninfoAnyScope("uid", "bar") -> "https://internal.example.org/";
```

filters/auth/doc.go
szuecs: ok

szuecs:
@aryszka I addressed your comments

filters/auth/doc.go
```diff
@@ -1,18 +1,155 @@
...
+are satisfied by the request:
+
+    a: Path("/a") -> oauthTokeninfoAllScope("uid") -> "https://internal.example.org/";
+    b: Path("/b") -> oauthTokeninfoAllScope("uid", "bar") -> "https://internal.example.org/";
+
+OAuth - oauthTokeninfoAnyKV() filter
+
+The filter oauthTokeninfoAnyKV allows access if the token information
+returned by OAuthTokeninfoURL has the given key and the given
+value. Additionally the second will check if there is
```
aryszka: "Additionally the second will check if there is
+the correct "realm" set." is not needed here, because it's there in the paragraph below.

filters/auth/doc.go
```diff
@@ -1,18 +1,155 @@
...
+        "bar-w",
+        "qux-rw"
+      ],
+      "token_type": "Bearer",
+      "uid": "jdoe"
+    }
+
+OAuth - oauthTokeninfoAllKV() filter
+
+The filter oauthTokeninfoAnyKV allows access if the token information
```
aryszka: "The filter oauthTokeninfoAnyKV..." -> "The filter oauthTokeninfoAllKV..."

filters/auth/oauth.go
```diff
@@ -0,0 +1,437 @@
...
+// String prints nicely the filter configuration based on the
+// configuration and check used.
+func (f *filter) String() string {
+	switch f.typ {
+	case checkOAuthTokeninfoAnyScopes:
+		return fmt.Sprintf("%s(%s)", OAuthTokeninfoAnyScopeName, strings.Join(f.scopes, ","))
+	case checkOAuthTokeninfoAllScopes:
+		return fmt.Sprintf("%s(%s)", OAuthTokeninfoAllScopeName, strings.Join(f.scopes, ","))
+	case checkOAuthTokeninfoAnyKV:
+		return fmt.Sprintf("%s(%s)", OAuthTokeninfoAnyKVName, f.kv)
```
aryszka: i think the format for the kv should be %v, therefore: `fmt.Sprintf("%s(%v)", OAuthTokeninfoAnyKVName, f.kv)` (also in the branch below)

filters/auth/oauth.go
```diff
@@ -0,0 +1,437 @@
...
+
+	vI, ok := h[scopeKey]
+	if !ok {
+		return false
+	}
+	v, ok := vI.([]interface{})
+	if !ok {
+		return false
+	}
+	var a []string
```
aryszka: the conversion of []interface{} to []string appears both in validateAnyScopes and validateAllScopes. It can be it's own function, or the intersect and all functions can accept []interface{} as the second argument and have the conversion in place, but it's also fine like this.

szuecs: I like it more like this.

filters/auth/oauth.go
```diff
@@ -0,0 +1,437 @@
...
+	if err != nil {
+		return err
+	}
+
+	defer rsp.Body.Close()
+	if rsp.StatusCode != 200 {
+		return errInvalidToken
+	}
+
+	d := json.NewDecoder(rsp.Body)
```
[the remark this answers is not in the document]
szuecs: the Unmarshal change I would not do, because it would mean more code, because of the []byte you need. It looked to ugly to me to change it.

szuecs:
From my side it's ready to go

szuecs:
:+1:

aryszka:
:+1:

PR #368, opened 2017-07-12
Add X-Flow-Id to the access logs

### LINKED ISSUE #243: FlowId support in logs
Currently skipper doesn't put FlowId in the access logs or in the application log, so this makes both correlating an access log entry with some downstream request pretty difficult and it makes impossible to correlate access log entry with other logging (e.g. some other filter's error report).

Would be nice if both things worked whenever a request has an `X-Flow-Id` header already or route has `flowId()` filter enabled.

DeTeam:
Fixes #243

Flow id is added to the access logs. Info is taken from request object which is either provided with the id from the caller or filled by the filter. Works for both json and apache [format](https://gist.github.com/DeTeam/ea043b9272060e3060bcc8853942dfc3)

Question here as well: should we make "reuse" flag a default on in the `flowId`. Right now without it we'd be re-generating flow Ids on every call, it kinda destroys the purpose.

cc @aryszka @lmineiro

### DIFF
--- a/logging/access.go
+++ b/logging/access.go
@@ -2,10 +2,13 @@ package logging
 
 import (
 	"fmt"
-	"github.com/sirupsen/logrus"
 	"net"
 	"net/http"
 	"time"
+
+	"github.com/sirupsen/logrus"
+
+	flowidFilter "github.com/zalando/skipper/filters/flowid"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


lmineiro:
Generating a new flow-id for each request made sense back then. Skipper would usually sit at the edge and, technically, no other system to correlate with should send a flow-id before skipper. This also allowed us to clearly identify a skipper generated FlowId from others.

Changing the default behavior of the filter doesn't sound good to me as it would cause unexpected surprises to other users.

lmineiro:
👍

aryszka:
:+1:

PR #372, opened 2017-07-18
add vendoring check

aryszka:
glide list returns with non-0 exit code if any of the imported dependencies is not vendored, 0 otherwise


szuecs:
:+1:

aryszka:
:+1:

Issue #375, opened 2017-07-18
path lookup issue when using pathsubtree() predicates


aryszka:
With the following routing:

```
foo1: PathSubtree("/foo/") && Method("PUT")
  -> setResponseHeader("X-Route", "foo1")
  -> status(200)
  -> <shunt>;

foo2: Path("/foo")
  -> setResponseHeader("X-Route", "foo2")
  -> status(200)
  -> <shunt>;
```

it is impossible to hit the route foo2. Most possibly because of how the PathSubtree predicates are constructed together with the effect of backtracking.

PR #383, opened 2017-07-27
Fix subtree conflict

aryszka:
Closes https://github.com/zalando/skipper/issues/375

### DIFF
--- a/pathmux/tree.go
+++ b/pathmux/tree.go
@@ -13,7 +13,7 @@ import (
 	"strings"
 )
 
-// Matcher objects, when using the LookupMatch function, can be used for additional checks and to override the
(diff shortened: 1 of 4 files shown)


szuecs:
@aryszka Is there a reason why you implemented the path-rewriting functions as lambda functions?
I don't see a reason, so I would argue it would be better to define them with a name.

aryszka:
@szuecs you are right! bad reflex for sparing on argument count. Fixing this. Thx

szuecs:
@aryszka travis job failed, otherwise lgtm

aryszka:
:+1:

szuecs:
:+1:

Issue #352, opened 2017-06-21
Provide an option to allow custom metrics inside filters.


chandu188:
The current metrics doesn't provide an option to measure custom metrics inside a filter. We have set of filters which process images (resize, converTojpg) and we would like to measure how long does each step take resize or convertTojpg.

szuecs:
Do I understand you correctly that you want to know the execution time of a filter?

chandu188:
Sorry if the above description is not clear. Let me try to explain with some examples.

Consider the following as an example of Response method in filter

func (_ *articlePath) Response(_ filters.FilterContext) {
   step1: convert to jpeg
   step2: resize image 

} 
currently, the default metrics provide the overall execution time of a filter. It's not possible to measure the execution time of the above 2 steps separately.


Another example could be from this https://github.com/zalando/skipper/blob/8fb107612ac6543f1c76d01bb3e8efe61dac5beb/filters/builtin/compress.go#L367. 
Here I don't think it's possible to measure the execution time for "encode" method. The default execution time will not be much help for this "compress" filter as it does not measure the time spent in encoding as it runs in a separate goroutine (correct me if I am wrong).

lmineiro:
This was actually discussed before. There's value on exposing "some" metrics registry to filters
We didn't come to a conclusion how to properly namespace that to avoid collisions in the metrics key space.

The suggestion, at the time was:
- Keep the metrics registry protected
- Expose methods to lazily get a metric (timer, counter, etc) under a namespace which should include the filter name
- Serialize filter specific metrics

PR #384, opened 2017-08-01
Provide option to collect custom metrics inside filters

chandu188:
Now one can collect the custom metrics inside the filter using the methods defined in the filters.Metrics interface.
Mentioned in this issue #352 
Please give your suggestions. 

Following is the idea the PR implemented:
1) Add an interface 'Metrics' in the filters package with following methods
        - MeasureSince(key, time)
        - IncCounter(key)
2) Add a method in the FilterContext interface to expose the above interface
3) add an extra field (metrics) in the context and implement the Metrics method added in the FilterContext interface
4) Implement the Metrics interface in metrics.Metrics struct.


aryszka:
:+1:

aryszka:
@lmineiro what do you think about this extension?

lmineiro:
This looks like one good start but I wouldn't leave it as it is.

Although the metrics registry seems to be properly protected, the namespacing [mentioned before](https://github.com/zalando/skipper/issues/352#issuecomment-310083869) is not. This can, potentially, lead to:
- A filter can use the same metric key as any other filter (breaking metrics for both)
- A filter can use a metric key that belongs to Skipper itself

One simple way could be to always prefix a filter's key with the filter name, which already needs to be unique in the filter registry.

What do you think?

aryszka:
prefixing is a good point, i agree. It can be done in the proxy where it is iterating over the filters.

filters/filters.go
```diff
@@ -117,7 +119,17 @@ func (r Registry) Register(s Spec) {
...
 
 // Metrics provides an option to collect custom metrics
-type Metrics interface {
-	IncCounter(key string)
-	MeasureSince(key string, start time.Time)
+const MetricKey = "%s.%s"
+
+type FilterMetrics struct {
+	Metrics    *metrics.Metrics
+	FilterName string
```
aryszka: i would prefer not using this as a package visible field. Let me contemplate a bit on a different approach..

chandu188: you are right. I should have thought a bit more before pushing the code.

aryszka:
@lmineiro we applied the prefixing. Could you please check if it looks ok like this?

aryszka:
:+1:

vigneshshanmugam:
👍

PR #385, opened 2017-08-02
use shared compiled regular expressions

aryszka:
using now shared compiled objects for the identical regular expressions in the host and path regexp predicates. `BenchmarkRegexp` was used to check whether this comes with degradation in the overall request processing. rx.Copy() is used according to the stdlib docs: http://godoc.org/regexp#Regexp.Copy

(to help testing, there is now a Mute/Umute function on the test logger)

### DIFF
--- a/eskip/eskip.go
+++ b/eskip/eskip.go
@@ -79,7 +79,7 @@ type Route struct {
 	// E.g. route1: ...
 	Id string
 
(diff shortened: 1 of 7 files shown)


DeTeam:
👍

lmineiro:
👍

grassator:
👍

PR #392, opened 2017-08-04
add option to strip query strings from request URIs in the access logs

aryszka:
add an option to strip query strings from the request URI field in the access logs


logging/access.go
```diff
@@ -44,7 +45,13 @@ type AccessEntry struct {
 	RequestTime time.Time
 }
 
-var accessLog *logrus.Logger
+// TODO: create individiual instances from the access log and
```
lmineiro: typo
s/individiual/individual

szuecs:
:+1:

lmineiro:
Starting to wonder if it wouldn't be better to just apply a formatter compatible with http://httpd.apache.org/docs/2.4/mod/mod_log_config.html#formats and give users the freedom on their access log format.

The reason why I'm suggesting it because we already extended it slightly to serve other purposes and this will be, yet another option to further customize it. Something like this https://github.com/lestrrat/go-apache-logformat ?

aryszka:
@lmineiro i agree. Honestly, the %U option missed my attention. However, I suggest to do it in a separate pull request, together with all the other standard formatter options.

Raffo:
The patch looks good, but I agree with @lmineiro that it would probably make sense at some point to support custom formatters.

logging/access.go
```diff
@@ -90,6 +97,14 @@ func (f *accessLogFormatter) Format(e *logrus.Entry) ([]byte, error) {
 	return []byte(fmt.Sprintf(f.format, values...)), nil
 }
 
+func stripQueryString(u string) string {
+	if i := strings.IndexRune(u, '?'); i < 0 {
```
Joneser: Is there any scenario in which we should maintain an anchor tag? I believe this will strip that too if it comes after the querystring

aryszka: I think the anchor tag is browser functionality and should not be sent to the server. So I believe no, there's no such sane scenario. Can someone think about any?

lmineiro: afaik, anchors/fragment parts are not sent to the server

aryszka:
:+1:

aryszka:
:-1:

aryszka:
thanks for the feedback. We agreed with @Raffo that if the solution suggested by @lmineiro can be shipped reasonably soon, then let's do rather that.

szuecs:
@aryszka @lmineiro there was no change here since the suggested better solution. I would rather merge and create a new issue for the evolved solution.

lmineiro:
I guess we'll look into the custom formatting some day ;-)

lmineiro:
👍

Raffo:
It took some time 😂

PR #397, opened 2017-08-07
Small README edits

LappleApple:
Fixed a few things. :)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -13,22 +13,20 @@ Skipper is an HTTP router and reverse proxy for service composition. It's design
 definitions with detailed lookup conditions, and flexible augmentation of the request flow with filters. It can be
 used out of the box or extended with custom lookup, filter logic and configuration sources.
 
-Main features:
+## Main features:
 
 - identifies routes based on the requests' properties, such as path, method, host and headers
 - allows modification of the requests and responses with filters that are independently configured for each route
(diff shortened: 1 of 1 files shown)


LappleApple:
👍

aryszka:
Hi @LappleApple thanks a lot for the fixes!!!

There's one thing that I'd like to bikeshed about, but just a bit: I was very conscious about writing out the full URLs of the links. This is an opinionated thing and the rationale behind it is that while we're using a desktop browser, we can see the URLs before clicking on a link when we hover on it. E.g. we can check if it's HTTPS. But when we're using a mobile browser, the URL stays hidden if only the link label is displayed, until we load the resource.

If I'm completely alone with this opinion and the single weirdo around here who cares about this, then of course, I'm not against changing the links, I'll do it myself for the few remaining ones. And it's totally OK to laugh at me for this ;)

Opinions anyone?

LappleApple:
Hey @aryszka -- someone else at Zalando has this style preference too/likes to show the full link for trust-building purposes. Is that part of your preference reasoning? If so, keep in mind that with Skipper being a company-owned repo, we are pretty trustworthy by default (we can't afford not to be). :)

aryszka:
@LappleApple ok, i'm convinced :)

aryszka:
:+1:

Issue #406, opened 2017-08-09
Kubernetes: Rules get removed in case of transient API server fault


linki:
We had an issue with Skipper as Kubernetes Ingress Proxy which resulted in Rules being removed due to a temporary outage of the Kubernetes control plane (specifically the API server). Skipper kept running but every request was answered with a `404`.

We identified the cause of the issue and explain it here:
* Skipper loads all Ingresses in `loadAndConvert()` (https://github.com/zalando/skipper/blob/v0.9.78/dataclients/kubernetes/kube.go#L590)
  * If this fails, Skipper won't change its rule set at all which is the correct behaviour for transient errors (https://github.com/zalando/skipper/blob/v0.9.78/routing/datasource.go#L75-L80 Note that it doesn't push into the `out` channel.)
* In case it successfully gets the Ingress list it converts them into a list of Skipper routes (https://github.com/zalando/skipper/blob/v0.9.78/dataclients/kubernetes/kube.go#L601)
  * For each referenced Ingress rule it looks up the corresponding Kubernetes Service by making another call to the Kubernetes API (aka n+1 query). This call is deeply hidden below https://github.com/zalando/skipper/blob/v0.9.78/dataclients/kubernetes/kube.go#L468.
  * This call might fail if the API server becomes unavailable at this point in time.
  * However, the error is handled in the wrong way with a `continue` and results in **no rule being added** for this particular ingress target to the desired rule set (https://github.com/zalando/skipper/blob/v0.9.78/dataclients/kubernetes/kube.go#L475-L479)
  * Since the list that's build up here is a *desired* list of routes, later on there is a difference calculation with the current list which will ultimately lead to this perfectly fine rule being actively deleted from skipper. (https://github.com/zalando/skipper/blob/v0.9.78/dataclients/kubernetes/kube.go#L704-L706)

Logs from Skipper (with some internal identifiers changed):

After fetching the Ingress objects, Skipper fails to retrieve the associated Services and removes the corresponding rules.

```console
{"log":"[APP]time=\"2017-08-08T10:04:11Z\" level=error msg=\"error while getting service: Get https://10.3.0.1:443/api/v1/namespaces/default/services/my-nginx: dial tcp 10.3.0.1:443: getsockopt: connection refused\" \n","stream":"stderr","time":"2017-08-08T10:04:11.104944951Z"}  
[...]

PR #409, opened 2017-08-09
Error early when apiserver is unreachable

linki:
This is a naive approach to fixing https://github.com/zalando/skipper/issues/406.

Skipper makes a series of requests in quick succession to the API server. If one request other than the first one fails the compiled list of rules is inconsistent and applying it does more harm than just doing nothing.

Therefore, instead of ignoring these errors we propagate them up which leads to no rules being changed in case of any failed request.

**However**, whilst doing it I discovered that this will also fail for errors such as referenced services that don't exist. Service names are provided by users and should not lead to skipper being permanently stuck with old rules. This needs to be fixed but I wanted to get this out early.

My idea would be that the returned error should be checked for whether it is *transient* or *permanent*.
* In case of *transient errors* we should skip this cycle and don't change the rules (e.g. temporary API server outage) hoping that the next cycle we'll have more luck. We should somehow report this *degraded* state via `/metrics` or similar immediately.
* However, *permanent errors* should just be ignored like before (such as broken rules due to Ingresses referencing Services that don't exist. Those errors cannot be fixed by retrying but only by users changing their manifests)

@aryszka @mikkeloscar @szuecs @hjacobs Let me know if this appraoch would make sense.

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -664,7 +664,7 @@ func (c *Client) LoadAll() ([]*eskip.Route, error) {
 	log.Debug("loading all")
 	r, err := c.loadAndConvert()
 	if err != nil {
-		log.Debugf("failed to load all: %v", err)
+		log.Errorf("failed to load all: %v", err)
 		return nil, err
 	}
 
@@ -689,7 +689,7 @@ func (c *Client) LoadUpdate() ([]*eskip.Route, []string, error) {
 	log.Debugf("polling for updates")
 	r, err := c.loadAndConvert()
 	if err != nil {
-		log.Debugf("polling for updates failed: %v", err)
+		log.Errorf("polling for updates failed: %v", err)
 		return nil, nil, err
 	}
 
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -423,7 +423,7 @@ func (c *Client) convertPathRule(ns, name, host string, prule *pathRule, service
 //
 // TODO:
 // - check how to set failures in ingress status
-func (c *Client) ingressToRoutes(items []*ingressItem) []*eskip.Route {
+func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
 	routes := make([]*eskip.Route, 0, len(items))
 	for _, i := range items {
 		if i.Metadata == nil || i.Metadata.Namespace == "" || i.Metadata.Name == "" ||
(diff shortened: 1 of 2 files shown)


mikkeloscar:
> However, whilst doing it I discovered that this will also fail for errors such as referenced services that don't exist. Service names are provided by users and should not lead to skipper being permanently stuck with old rules. This needs to be fixed but I wanted to get this out early.

Could we return a special error for service not found vs. connection problem? Then we could easily determine if it's a *real, don't continue error* or if it's a *service does not exist, and remove route error*.

Raffo:
I second what @mikkeloscar said, still with the goal of having a "quick" solution out of the door.

dataclients/kubernetes/kube.go
```diff
@@ -598,7 +597,11 @@ func (c *Client) loadAndConvert() ([]*eskip.Route, error) {
 	log.Debugf("all ingresses received: %d", len(il.Items))
 	fItems := c.filterIngressesByClass(il.Items)
 	log.Debugf("filtered ingresses by ingress class: %d", len(fItems))
-	r := c.ingressToRoutes(fItems)
+	r, err := c.ingressToRoutes(fItems)
+	if err != nil {
+		log.Debugf("converting ingresses to routes failed: %v", err)
```
szuecs: Debug?

aryszka: i also believe this could be an Errorf

szuecs:
As I discussed with @linki it would be good if we would have the net.Error interface that differentiate between permanent and temporary errors. 

On the other hand I think the whole thing should be changed to have an initial sync and then a watch loop for updates. This change would only makes sense if we would use Kubernetes client-go, which could clean up also some parts for the data retrieval. 

For the time now, we should just merge the quick fix, which may have degraded status. 
Another thing we have to do is alert based on an error log that shows that route updates are failing, caused by user input.

dataclients/kubernetes/kube.go
szuecs: It's done now in below in LoadAll and LoadUpdate, which is ok for me

szuecs:
:+1:

aryszka:
:+1:

aryszka:
i don't agree with using the client-go from kubernetes, prefer to just use the http api, but I strongly we could make the client smarter and less agressive with the polling

linki:
> Could we return a special error for service not found vs. connection problem? Then we could easily determine if it's a real, don't continue error or if it's a service does not exist, and remove route error.

@mikkeloscar @Raffo This is also what I had in mind, just in a structured form so that we can learn from it. For instance, Package net provides the following `Error` definition:

```go
package net

type Error interface {
    error
    Timeout() bool   // Is the error a timeout?
    Temporary() bool // Is the error temporary?
}
```

So that the client can decide what to do based on whether the error is temporary or not.

```go
if nerr, ok := err.(net.Error); ok && nerr.Temporary() {
    time.Sleep(1e9)
    continue
}
if err != nil {
    log.Fatal(err)
}
```

Handling them on a per-occurrence basis just means we'll add more and more types in the future, I believe.

linki:
Follow up https://github.com/zalando/skipper/pull/410

PR #410, opened 2017-08-09
Treat network errors as transient and defer updating routes

linki:
Follow up of https://github.com/zalando/skipper/pull/409#issuecomment-321256681

It restores the original test suite that tolerates misconfiguration such as Ingresses referencing missing services (404 from the API server).

However, this adds an early return in case of network errors such as an unavailable API server. This has the effect that Skipper won't remove its routes until it gets a successful response from the API server. (Note that successful includes 500s in this case. I guess we should treat 500s in a similar "don't change the routes" way)

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -140,7 +140,6 @@ var nonWord = regexp.MustCompile("\\W")
 
 var (
 	errServiceNotFound      = errors.New("service not found")
-	errServicePortNotFound  = errors.New("service port not found")
 	errAPIServerURLNotFound = errors.New("kubernetes API server URL could not be constructed from env vars")
 	errInvalidCertificate   = errors.New("invalid CA")
 )
@@ -281,6 +280,10 @@ func (c *Client) getJSON(uri string, a interface{}) error {
 	log.Debugf("request to %s succeeded", url)
 	defer rsp.Body.Close()
 
+	if rsp.StatusCode == http.StatusNotFound {
+		return errServiceNotFound
+	}
+
 	if rsp.StatusCode != http.StatusOK {
 		log.Debugf("request failed, status: %d, %s", rsp.StatusCode, rsp.Status)
 		return fmt.Errorf("request failed, status: %d, %s", rsp.StatusCode, rsp.Status)
@@ -330,7 +333,7 @@ func (c *Client) getServiceURL(namespace, name string, port backendPort) (string
 	}
 
 	log.Debugf("service port not found by name: %s", pn)
-	return "", errServicePortNotFound
+	return "", errServiceNotFound
 }
 
 // TODO: find a nicer way to autogenerate route IDs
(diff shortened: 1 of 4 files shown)


dataclients/kubernetes/kube.go
```diff
@@ -467,11 +467,19 @@ func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
 				if prule.Backend.Traffic > 0 {
 					r, err := c.convertPathRule(i.Metadata.Namespace, i.Metadata.Name, rule.Host, prule, servicesURLs)
 					if err != nil {
-						// tolerate single rule errors
+						// fail when there's a temporary network error so that we keep the last known rules
+						// active until the next sync loop.
+						if err, ok := err.(net.Error); ok {
```
mikkeloscar: I think this is too low-level. Imagine the apiserver respond with 500 then this would be treated as a permanent error.

mikkeloscar: err, you already noted this in the description, sorry.

>  I guess we should treat 500s in a similar "don't change the routes" way)

Yup :)

dataclients/kubernetes/kube.go
```diff
@@ -467,11 +467,19 @@ func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
...
+							log.Errorf("temporary error while getting service, deferring update: %v", err)
+							return nil, err
```
aryszka: not good. This will delete all the existing routes on update if there is a network error after the initial load was successful.

The checking of the error type should happen after loadAndConvert() was called in LoadAll and LoadUpdate. If it is a network error during LoadAll, then the routes in c.current should be returned. If it is a network error in LoadUpdate, then nil should be returned for the routes and the deleted ids, plus an error.

The reason for the error in the case of LoadUpdate() is that when LoadUpdate() fails, skipper will automatically call LoadAll() and repeat calling it with a delay (debounce without throttling), in order to recover when the source (the ingress API in this case) recovers. (We should document this for the routing.DataClient interface.)

dataclients/kubernetes/kube.go
aryszka: note that checking for net.Error is unfortunately only reliable when we're directly handling the net package. Remember the io.EOF errors.

https://github.com/golang/go/issues/13667
and maybe more interestingly: https://github.com/golang/go/issues/16465#issuecomment-234669772

i think it would be better to do this:
- add an additional return argument to this function, which contains all the route ids that had an error, and just log the error without checking it just like before
- in the loadall(), keep the previous version of the failed routes
- in the loadupdate(), ignore the failed routes (make sure not to delete them)

dataclients/kubernetes/kube.go
mikkeloscar: > not good. This will delete all the existing routes on update if there is a network error after the initial load was successful.

Are you sure? If `LoadAll` or `LoadUpdate` fails it will check if `err != nil` and not do any route updates if that's the case, but just try again: https://github.com/zalando/skipper/blob/master/routing/datasource.go#L68-L80

I don't see why this would delete the existing routes.

aryszka: true. my bad, sorry.

dataclients/kubernetes/kube.go
aryszka: maybe we can return our own type of error from getJSON() and use that for distinguishing

aryszka:
i suggest not to rely on net.Error but use our own error type for the cases where getJSON() fails. Otherwise looks good to me.

aryszka:
👍

dataclients/kubernetes/kube.go
szuecs: Please create a custom error type and an interface, such that we can decide if it's permanent or non permanent on error creation time. I think this will simplify the handling of the error.

mikkeloscar:
@szuecs @linki @aryszka Added a branch with a solution that uses the existing `errServiceNotFound` error variable to determine whether it was a temporary error of if the backend service doesn't exist/is invalid.

See the patch here (it's on top of this PR): https://github.com/zalando/skipper/compare/known-error

linki:
@mikkeloscar lgtm. I'm testing this and cherry pick.

linki:
Works fine for me in a local test: I let skipper fetch the ingress through `kube proxy` then I kill `kubectl proxy` and it keeps retrying while serving the old routes. Before it used to clean them up at this point. So 👍

mikkeloscar:
:+1:

linki:
Fixes https://github.com/zalando/skipper/issues/406

PR #416, opened 2017-08-13
remove outdated examples

szuecs:
Remove Examples, which have no reference anymore.


aryszka:
i'll fix these examples in a bit so that they will be reachable from godoc

PR #425, opened 2017-08-27
Make default HTTP status configurable

### LINKED ISSUE #408: kubeclient: Return 502 instead of 404 if route is not configured.
If you make a request with a hostname that is not configured in skipper you currently get a 404 i.e. a client error.
In reality it is more likely to be a server error that the route wasn't configured and thus the correct response should be 502 - Bad gateway.

mikkeloscar:
Makes the default 'no-routes-found' http status code configurable via
the flag `--default-http-status`.

This makes it possible to treat routes not found as server errors e.g.
502 Bad Gateway, instead of treating them as client errors (defaults to 404).

When using the Kubernetes dataclient a default catchall route which
returns 404 will be configured for any host rule which doesn't specify
its own catchall rule (`""` or `/`).

Fix #408

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -466,6 +466,8 @@ func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
 			// update Traffic field for each backend
 			computeBackendWeights(backendWeights, rule)
 
+			hostRoutes := make([]*eskip.Route, 0, len(rule.Http.Paths))
+
 			for _, prule := range rule.Http.Paths {
 				if prule.Backend.Traffic > 0 {
 					r, err := c.convertPathRule(i.Metadata.Namespace, i.Metadata.Name, rule.Host, prule, servicesURLs)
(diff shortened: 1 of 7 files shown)


dataclients/kubernetes/kube.go
```diff
@@ -481,15 +483,46 @@ func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
...
+
+			routes = append(routes, hostRoutes...)
+
+			// if routes were configured, but there is no catchall route defined,
+			// create a route which returns 404
+			if len(hostRoutes) > 0 && !catchAllRoutes(hostRoutes) {
+				catchAll := &eskip.Route{
+					Id:          routeID(i.Metadata.Namespace, i.Metadata.Name, rule.Host, "", ""),
+					HostRegexps: host,
+					BackendType: eskip.ShuntBackend,
```
mikkeloscar: @aryszka you mentioned in #408 to use the status filter. Is that better than using the `ShuntBackend` like I do here?

proxy/proxy.go
```diff
@@ -320,6 +324,10 @@ func WithParams(p Params) *Proxy {
 		p.MaxLoopbacks = 0
 	}
 
+	if p.DefaultHTTPStatus >= http.StatusContinue && p.DefaultHTTPStatus <= http.StatusNetworkAuthenticationRequired {
+		errRouteLookupFailed.code = p.DefaultHTTPStatus
```
mikkeloscar: It feels a bit wrong to modify the package level `errRouteLookupFailed` variable as a side-effect here. Alternatively I could add a `DefaultHTTPStatus` field on the proxy struct and reference that, when the error occur.

What do you think?

dataclients/kubernetes/kube.go
aryszka: @mikkeloscar the shunt backend is fine, there must be a backend, and shunt is like a "blind" backend. The status filter can be used in addition to that. But if the default status is good, then the status filter is not needed. So, I think it is good like this.

proxy/proxy.go
aryszka: agree. The errRouteLookupFailed was until now virtually a "const". Now that it depends on init data, we need a field on the Proxy struct.

aryszka:
apart from the package level errRouteLookupFailed thingy, it looks good to me

aryszka:
:+1:

szuecs:
:+1:

dataclients/kubernetes/kube.go
AlexanderYastrebov: >  But if the default status is good, then the status filter is not needed.

Since this does not use status filter it would use DefaultHTTPStatus - the same value as when route lookup failed which means that catchall routes are useless (unless I am missing something) - the client would get 404 (or configured value) either way.

szuecs: @AlexanderYastrebov the problem is something like "confused deputy". So you create a prepared call to x, but get application y because of no catchall host route. I think this was the reasoning why we have a catchall by host. IIRC one driver was, because of the ingress route annotation and predicate annotation.
Since RouteGroup I think we have everything to replace all these use cases, so I would also say we can drop the catchall for host routes.

Issue #296, opened 2017-03-09
accept integers in programmatically created filter/predicate arguments


aryszka:
eskip parses numbers as float64. Some filters and predicates relying this accept only float64 args in number positions. This is very inconvenient when creating filters programmatically. Where possible fix these filters to accepts integers, too.

PR #430, opened 2017-09-07
Accept integers in programmatically created filter/predicate arguments. #296

jreyeshdez:
Modified eskip to parse numbers as int64 too.

This is a possible solution for #296 and I'd like to get feedback if it's in the right path.

### DIFF
--- a/eskip/eskip_test.go
+++ b/eskip/eskip_test.go
@@ -302,7 +302,7 @@ func TestParseFilters(t *testing.T) {
 	}, {
 		"success",
 		`filter1(3.14) -> filter2("key", 42)`,
-		[]*Filter{{Name: "filter1", Args: []interface{}{3.14}}, {Name: "filter2", Args: []interface{}{"key", float64(42)}}},
+		[]*Filter{{Name: "filter1", Args: []interface{}{3.14}}, {Name: "filter2", Args: []interface{}{"key", int64(42)}}},
 		false,
 	}} {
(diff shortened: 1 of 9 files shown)


aryszka:
Thanks for raising this topic. Using float64 for all cases is an early design mistake on our side, and wish now the parser would have been created rather as it is in your PR.

The main issue with changing this behavior is that there can be extensions with custom filter and predicate implementations that only handle the float64 case and would get broken if such numbers would be parsed as int64 at some point. At the moment, I don't see a nice way to do this in a backwards compatible way, only with a breaking change.

A major release is something that we are planning for a while, but that should be done in a planned way that includes all the similar breaking changes like this and this means at least a couple of weeks.

If the breaking release doesn't happen early enough, would it be an OK temporary solution for your use case to parse only explicit literals to int64? E.g. `42L`? I have no better idea atm, but will think about it.

aryszka:
opened this: https://github.com/zalando/skipper/issues/432 and referenced the topic there

jreyeshdez:
I have tried to handle most of the scenarios as seen in the test cases, even though now all the tests go though I think this would need more testing, but it would be an OK if it is taken as a draft solution for what it could be a possible solution in future. As you have said, there are too many cases where float64 is expected.

aryszka:
Closing this as there was no progress. The distinction between integers and floats will be introduced in a later version of the eskip syntax, probably with breaking changes.

PR #431, opened 2017-09-07
add combined metrics for backend latency and response streaming

aryszka:
adding the non route based backend and response streaming metrics. The key of the metrics has an 'all.' prefix, so that existing filter with the prefix 'backend.' or 'response.' won't include them, and so avoid ending up in duplicate sum of the values.

What I am wondering if these should be enabled only if the route based variants of these metrics are disabled. Any advice?

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -77,17 +77,19 @@ type Options struct {
 }
 
 const (
-	KeyRouteLookup      = "routelookup"
-	KeyRouteFailure     = "routefailure"
-	KeyFilterRequest    = "filter.%s.request"
-	KeyFiltersRequest   = "allfilters.request.%s"
-	KeyProxyBackend     = "backend.%s"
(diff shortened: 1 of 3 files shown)


NorthFury:
:+1:

metrics/metrics.go
```diff
@@ -240,8 +243,9 @@ func (m *Metrics) MeasureAllFiltersResponse(routeId string, start time.Time) {
 }
 
 func (m *Metrics) MeasureResponse(code int, method string, routeId string, start time.Time) {
+	method = measuredMethod(method)
+	m.measureSince(fmt.Sprintf(KeyResponseCombined, code, method), start)
```
lmineiro: these actuator metrics also cause some explosion of metrics. I'd also add a flag for it. This is more useful with APIs than regular websites, hence the flag

aryszka:
:+1:

NorthFury:
:+1:

Issue #432, opened 2017-09-07
breaking release planning


aryszka:
This issue should collect the topics that we should address with the next major release that can contain breaking changes. API refactor items or changes to eskip, that would be much better and cleaner implemented with breaking backwards compatibility, should be listed here. I suggest to open an epic branch as an integration point and merge the individual pull requests there.

### Eskip

- support integers. See https://github.com/zalando/skipper/issues/296 and https://github.com/zalando/skipper/pull/430

### Metrics

- smarter way to enable/disable metrics types

...

aryszka:
Resurrecting this thread...

One of the plenty topics to address are the legacy predicates that have a named field representation on the eskip.Route object, e.g Path, Method etc.

Related: https://github.com/zalando/skipper/issues/1272

aryszka:
https://github.com/zalando/skipper/pull/1347

PR #433, opened 2017-09-14
Add opentracing to skipper

vetinari:
* enable with `-opentracing "$implementation"`
* currently hard coded implementations, these should
  be loadable modules


lmineiro:
I believe this is a good opportunity to explore [Go Plugins](https://golang.org/pkg/plugin/) and have the specific implementations outside of the main binary.

Otherwise we could consider build flags and only pull in specific dependencies and implementations explicitely

Thoughts @szuecs, @aryszka ?

szuecs:
@lmineiro  I like the idea of having a defined way to extend skipper with plugins to not pollute the main binary.

aryszka:
i like the plugin idea. Added it here, too: https://github.com/zalando/skipper/issues/432

tracing/tracing.go
```diff
@@ -0,0 +1,59 @@
...
+			AccessToken: token,
+			Tags: map[string]interface{}{
+				lightstep.ComponentNameKey: "skipper", // FIXME
+			},
+		})
+
+	default:
+		return ErrUnsupportedTracer
+	}
+	ot.SetGlobalTracer(tracer)
```
aryszka: On the long run, I have always regretted later when I chose to use global, shared objects. Based on this: https://github.com/opentracing/opentracing-go/blob/master/globaltracer.go , using the globalTracer of this libe doesn't bring much value, and it's basically just provides a global access to the same shared object. This is similar to the current log usage that I already regretted. Just a few examples that come to my mind:

- mocking it for some testing becomes harder, sometimes even resulting in false positives
- replacing the implementation with a different compatible one becomes harder
- someone with 'weird' use cases cannot so freely reuse the library, e.g. when one would want to run multiple skipper proxies in the same process

Maybe, these examples are too specific and we just should be pragmatic, like now with the logs, and if everybody's fine with the global shared object, then i'm just fine with it, too. But I'd still vote for explicitly passing the tracing to the objects that use it. E.g. the main initializes the tracer, passes it to the proxy object, the proxy object passes it to the context for every request, the filters get it from the context if they want to use it. Would this be too much more tedious than just using the package level var?

Opinions?

proxy/proxy.go
```diff
@@ -671,6 +699,22 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx := newContext(w, r, p.flags.PreserveOriginal(), p.metrics)
 	ctx.startServe = time.Now()
 
+	if p.openTracing {
+		carrier := ot.HTTPHeadersCarrier(r.Header)
+		tracer := ot.GlobalTracer()
+		wireContext, err := tracer.Extract(ot.HTTPHeaders, carrier)
+		var span ot.Span
+		if err == nil {
+			span = tracer.StartSpan("ingress", ext.RPCServerOption(wireContext))
```
aryszka: what does 'ingress' mean here? Is it a reference to the Kubernetes Ingress? If yes, then I think this should be a bit more general, or if needed should come from the settings, because ingress is not the only use case of skipper.

tracing/tracing.go
mikkeloscar: I agree! For logging it can be convenient to not have to pass the logger around every time you want to log something. But for this use case I think it would actually be more convenient/clear with a tracer field on the Proxy.

vetinari: Fine with me (and it's an easy change ;-))

proxy/proxy.go
vetinari: Yes, that's were it came from.

tracing/tracing.go
```diff
@@ -1,59 +1,81 @@
 // Package tracing handles opentracing support for skipper
+//
+// The tracers, except for "noop", are build as loadable modules. The modules must have
```
lmineiro: s/are build/are built

tracing/tracing.go
```diff
@@ -1,59 +1,81 @@
...
+// an "InitTracer" function with the function signature
```
lmineiro: a function named "InitTracer" with the following signature

tracing/tracing.go
lmineiro: Let's consistently call these Plugins. Here, since it's the first time, I'd:
s/loadable modules/Go Plugins
s/The modules/The plugins

tracing/tracing.go
```diff
@@ -1,59 +1,81 @@
...
+//
+//    func([]string) (opentracing.Tracer, error)
+//
+// The parameters passed are all arguments for the module, i.e. everything after the first
```
lmineiro: s/for the module/for the plugin
s/-opentracing parmeter/-opentracing parameter

tracing/tracing.go
```diff
@@ -1,59 +1,81 @@
...
+// word from skipper's -opentracing parameter. E.g. when the -opentracing parmeter is
+// "mytracer foo=bar token=xxx somename=bla:3" the "mytracer" plugin will receive
+//
+//    []string{"foo=bar", "token=xxx", "somename=bla:3"}
+//
+// as arguments.
+//
+// The tracer plugin is responsible for argument parsing.
```
lmineiro: The tracer plugin implementation is responsible to parse the received arguments

tracing/tracing.go
```diff
@@ -1,59 +1,81 @@
...
+//
+// An example plugin looks like
+//
+//     package main
+//
+//     import (
+//          instana "github.com/instana/golang-sensor"
```
lmineiro: Let's leave the example with the OpenTracing Basic

tracing/tracing.go
```diff
@@ -1,59 +1,81 @@
...
+//     )
+//
+//     func InitTracer(opts []string) (opentracing.Tracer, error) {
+//          return instana.NewTracerWithOptions(&instana.Options{
+//              Service:  "skipper",
+//              LogLevel: instana.Error,
+//          }), nil
+//     }
+//
+// This needs to be build with
```
lmineiro: s/needs to be build/should be built

tracing/tracing.go
```diff
@@ -1,59 +1,81 @@
...
+//
+//    go build -buildmode=plugin -o instana.so ./instana/instana.go
```
lmineiro: s/instana/basic

tracing/tracing.go
```diff
@@ -1,59 +1,81 @@
...
+//
+// and copied to the directory given as -plugindir (by default, ".").
+// Then it can be loaded with -opentracing "instana" as parameter to skipper.
```
lmineiro: s/instana/basic
s/as parameter/as startup parameter
Does it need the quotes?

tracing/tracing.go
```diff
@@ -1,59 +1,81 @@
...
-	ErrMissingArguments error = errors.New("no arguments passed")
-)
-
-// Init sets up opentracing
-func Init(opts []string) error {
+// LoadPlugin loads the given opentracing plugin and returns an opentracing.Tracer
+func LoadPlugin(pluginDir string, opts []string) (ot.Tracer, error) {
 	if len(opts) == 0 {
-		return ErrMissingArguments
+		return nil, errors.New("no arguments passed")
```
lmineiro: s/no arguments passed/opentracing: the implementation parameter is mandatory

tracing/tracing.go
lmineiro: We should also mention the note from Go's plugin package that this is currently incomplete and only supported on Linux

tracing/tracing.go
```diff
@@ -1,42 +1,49 @@
...
-//              LogLevel: instana.Error,
+//          return basic.NewTracerWithOptions(basic.Options{
+//              Recorder:       basic.NewInMemoryRecorder(),
+//              ShouldSample:   func(traceID uint64) bool { return traceID%64 == 0 },
+//              MaxLogsPerSpan: 25,
 //          }), nil
 //     }
 //
-// This needs to be build with
+// This should to be build with
```
lmineiro: s/be build/be built

cmd/skipper/main.go
```diff
@@ -258,57 +261,52 @@ func main() {
...
-		EnableRuntimeMetrics:                runtimeMetrics,
-		EnableServeRouteMetrics:             serveRouteMetrics,
-		EnableServeHostMetrics:              serveHostMetrics,
-		EnableBackendHostMetrics:            backendHostMetrics,
-		EnableAllFiltersMetrics:             allFiltersMetrics,
-		EnableRouteResponseMetrics:          routeResponseMetrics,
-		EnableRouteBackendErrorsCounters:    routeBackendErrorCounters,
-		EnableRouteStreamingErrorsCounters:  routeStreamErrorCounters,
-		EnableRouteBackendMetrics:           routeBackendMetrics,
-		DisableMetricsCompatibilityDefaults: disableMetricsCompat,
```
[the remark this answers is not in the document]
vetinari: Looks better now. Was probably dropped on merge of upstream/master ... :(

vetinari:
The travis build currently fails because `glide list` expects `github.com/opentracing/opentracing-go` to be in the vendor directory. For plugins outside the main skipper repo this is a requirement. I already made a PR to glide to have an `--ignore` which fixes this: https://github.com/Masterminds/glide/pull/909.

aryszka:
a temporary workaround that comes to my mind for the vendored plugin is to clone it into the vendor dir as a preparation step. Then, when it is fixed in glide, this step could be removed. Any better ideas?

vetinari:
Now we just need a way for a filter to access the tracer to set tags / operation name / baggage ...

proxy/loopback_test.go
```diff
@@ -176,7 +177,7 @@ func TestLoopbackShunt(t *testing.T) {
 			-> <shunt>;
 	`
 
-	testLoopback(t, routes, Params{}, http.StatusTeapot, http.Header{
+	testLoopback(t, routes, Params{OpenTracer: &opentracing.NoopTracer{}}, http.StatusTeapot, http.Header{
```
aryszka: i think that the context should be initialized with the nooptracer when the params don't contain any. That way we wouldn't need to set the nooptracer explicitly everywhere. For the filterstest package, it doesn't matter too much, but for the proxy package it does. See the newContext() "constructure" function, there we could check if the input tracer if nil, then we would set the nooptracer by default.

filters/filtertest/filtertest.go
```diff
@@ -61,6 +64,7 @@ func (fc *Context) BackendUrl() string                  { return fc.FBackendUrl
 func (fc *Context) OutgoingHost() string                { return fc.FOutgoingHost }
 func (fc *Context) SetOutgoingHost(h string)            { fc.FOutgoingHost = h }
 func (fc *Context) Metrics() filters.Metrics            { return fc.FMetrics }
+func (fc *Context) Tracer() opentracing.Tracer          { return &opentracing.NoopTracer{} }
```
aryszka: we can just return the FTracer field, and then the anyone writing tests can decide to set it or not. The other fields work also like this.

proxy/proxy.go
```diff
@@ -290,7 +290,9 @@ func New(r *routing.Routing, options Options, pr ...PriorityRoute) *Proxy {
 		Routing:              r,
 		Flags:                Flags(options),
 		PriorityRoutes:       pr,
-		CloseIdleConnsPeriod: -time.Second})
+		CloseIdleConnsPeriod: -time.Second,
+		OpenTracer:           &ot.NoopTracer{},
```
aryszka: i think we should move the setting of the default tracer to the WithParams function. Something like:

```
params.Tracer == nil {
   params.Tracer = &ot.NoopTracer{}
}
```

filters/filtertest/filtertest.go
vetinari: Anyone needs it in the FilterContext, so I went for returning a valid tracer. We could check that FTracer is non-nil and return that and fall back to the NoopTracer.

aryszka:
:+1:

lmineiro:
👍

szuecs:
@vetinari @lmineiro I don't see any tests provided that show a good case or a bad case. Can you provide it before you merge?

szuecs:
Thanks @vetinari

szuecs:
:+1:

lmineiro:
👍

Issue #424, opened 2017-08-20
Rate limit filter


szuecs:
Skipper should have a rate limit implemented as filter.
One use case is in #414, but it could be used in other configurations without Kubernetes.

It could be used as DDoS prevention ([slowloris](https://en.wikipedia.org/wiki/Slowloris_(software)) is one http based attack used in [DDoS](https://en.wikipedia.org/wiki/Denial_of_Service) attacks) and login enumeration (user enumeration or password guessing, dictionary attacks, ..)

The configuration should allow
* rate limit per global request per second, p.e. 100 req/s
* allow a burst factor, p.e. 4, meaning 400 req/s for one minute
* rate limit per client IP (and X-Forwarded-For content) request per second, p.e. 10 req/s with burs of 40 req/s
* rate limit per Authorization Header (Bearer Token or Basic auth)
* respond automatically with HTTP code 429 (or does it makes sense to make it configurable?)

szuecs:
Burst factor and global rate limit settings were not done.

I consider this as unnecessary, because global rate limit can only be done, if skipper-ingress is aware of the number of instances (~ service mesh required).
For burst factor, I am not sure if someone really wants this feature or if this is just a feature copy of nginx.

If someone really needs these features please create follow up issues.

PR #434, opened 2017-09-19
ratelimit support via args

szuecs:
example: skipper -ratelimits type=local,maxHits=50,timeWindow=60s
this would be applied before the filter chain

This partially implements #424

### DIFF
--- a/cmd/skipper/ratelimiterflags.go
+++ b/cmd/skipper/ratelimiterflags.go
@@ -8,11 +8,11 @@ import (
 	"github.com/zalando/skipper/ratelimit"
 )
 
-const ratelimitUsage = `set global rate limit settings, e.g. -ratelimit type=local,maxHits=20,timeWindow=60
+const ratelimitUsage = `set global rate limit settings, e.g. -ratelimit type=local,max-hits=20,time-window=60
 	possible ratelimit properties:
 	type: local/disabled (defaults to local)
-	maxHits: the number of hits a local (meaning per instance) ratelimiter can get
-	timeWindow: the duration of the sliding window for the rate limiter
+	max-hits: the number of hits a local (meaning per instance) ratelimiter can get
+	time-window: the duration of the sliding window for the rate limiter
 	(see also: https://godoc.org/github.com/zalando/skipper/ratelimit)`
 
 const enableRatelimitUsage = `enable ratelimit`
@@ -50,13 +50,13 @@ func (r *ratelimitFlags) Set(value string) error {
 			default:
 				return errInvalidRatelimitConfig
 			}
-		case "maxHits":
+		case "max-hits":
 			i, err := strconv.Atoi(kv[1])
 			if err != nil {
 				return err
 			}
 			s.MaxHits = i
-		case "timeWindow":
+		case "time-window":
 			d, err := parseDurationFlag(kv[1])
 			if err != nil {
 				return err
@@ -68,7 +68,7 @@ func (r *ratelimitFlags) Set(value string) error {
 		}
 	}
 
-	if s.Type == ratelimit.NonRatelimit {
+	if s.Type == ratelimit.NoRatelimit {
 		s.Type = ratelimit.DisableRatelimit
 	}
 
--- a/cmd/skipper/ratelimiterflags.go
+++ b/cmd/skipper/ratelimiterflags.go
@@ -1,0 +1,77 @@
+package main
+
+import (
+	"errors"
+	"strconv"
+	"strings"
+
+	"github.com/zalando/skipper/ratelimit"
+)
+
+const ratelimitUsage = `set global rate limit settings, e.g. -ratelimit type=local,maxHits=20,timeWindow=60
+	possible ratelimit properties:
+	type: local/disabled (defaults to local)
+	maxHits: the number of hits a local (meaning per instance) ratelimiter can get
+	timeWindow: the duration of the sliding window for the rate limiter
+	(see also: https://godoc.org/github.com/zalando/skipper/ratelimit)`
+
+const enableRatelimitUsage = `enable ratelimit`
+
+type ratelimitFlags []ratelimit.Settings
+
+var errInvalidRatelimitConfig = errors.New("invalid ratelimit config")
+
+func (r *ratelimitFlags) String() string {
+	s := make([]string, len(*r))
+	for i, ri := range *r {
+		s[i] = ri.String()
+	}
+
+	return strings.Join(s, "\n")
+}
+
+func (r *ratelimitFlags) Set(value string) error {
+	var s ratelimit.Settings
+
+	vs := strings.Split(value, ",")
+	for _, vi := range vs {
+		kv := strings.Split(vi, "=")
+		if len(kv) != 2 {
+			return errInvalidRatelimitConfig
+		}
+
+		switch kv[0] {
+		case "type":
+			switch kv[1] {
+			case "local":
+				s.Type = ratelimit.LocalRatelimit
+			case "disabled":
+				s.Type = ratelimit.DisableRatelimit
+			default:
+				return errInvalidRatelimitConfig
+			}
+		case "maxHits":
+			i, err := strconv.Atoi(kv[1])
+			if err != nil {
+				return err
+			}
+			s.MaxHits = i
+		case "timeWindow":
+			d, err := parseDurationFlag(kv[1])
+			if err != nil {
+				return err
+			}
+			s.TimeWindow = d
+			s.CleanInterval = d * 10
+		default:
+			return errInvalidRatelimitConfig
+		}
+	}
+
+	if s.Type == ratelimit.NonRatelimit {
+		s.Type = ratelimit.DisableRatelimit
+	}
+
+	*r = append(*r, s)
+	return nil
+}
--- a/filters/ratelimit/ratelimit.go
+++ b/filters/ratelimit/ratelimit.go
@@ -1,0 +1,130 @@
+/*
+Package ratelimit provides filters to control the rate limitter settings on the route level.
+
+For detailed documentation of the ratelimit, see https://godoc.org/github.com/zalando/skipper/ratelimit.
+*/
+package ratelimit
+
+import (
+	"time"
+
+	"github.com/zalando/skipper/filters"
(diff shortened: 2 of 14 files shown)


cmd/skipper/ratelimiterflags.go
```diff
@@ -0,0 +1,77 @@
...
+		case "type":
+			switch kv[1] {
+			case "local":
+				s.Type = ratelimit.LocalRatelimit
+			case "disabled":
+				s.Type = ratelimit.DisableRatelimit
+			default:
+				return errInvalidRatelimitConfig
+			}
+		case "maxHits":
```
aryszka: we're using dash casing already in the circuit breaker params. I think this should be "max-hits", and "time-window" below.

cmd/skipper/ratelimiterflags.go
```diff
@@ -0,0 +1,77 @@
...
+				return err
+			}
+			s.TimeWindow = d
+			s.CleanInterval = d * 10
+		default:
+			return errInvalidRatelimitConfig
+		}
+	}
+
+	if s.Type == ratelimit.NonRatelimit {
```
aryszka: NonRatelimit -> NoRatelimit maybe?

ratelimit/ratelimit.go
```diff
@@ -0,0 +1,115 @@
...
+	Header = "X-Rate-Limit"
+	// LocalRatelimitName is the name of the LocalRatelimit, which will be shown in log
+	LocalRatelimitName = "localRatelimit"
+	// DisableRatelimitName is the name of the DisableRatelimit, which will be shown in log
+	DisableRatelimitName = "disableRatelimit"
+)
+
+const (
+	// NonRatelimit is not used
+	NonRatelimit Type = iota
```
aryszka: NonRatelimit -> NoRatelimit maybe?

skipper.go
```diff
@@ -281,6 +282,16 @@ type Options struct {
...
 	BreakerSettings []circuit.BreakerSettings
 
+	// EnableRatelimiters enables the usage of the ratelimiter in the route definitions without initializing any
+	// by default. It is a shortcut for setting the RatelimitSettings to:
+	//
+	// 	[]ratelimit.Settings{{Type: DisableRatelimit}}
+	//
+	EnableRatelimiters bool
+
+	// RatelimitSettings contain global and host specific settings for the ratelimitters.
```
aryszka: ratelimitters -> ratelimiters

filters/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// NewLocalRatelimit creates a local measured rate limiting, that is
+// only aware of itself. If you have 5 instances with 20 req/s, then
+// it would allow 100 req/s to the backend.
+//
+// Example:
+//
+//    backendHealthcheck: Path("/healthcheck")
+//    -> localRatelimit(20, "1m")
```
aryszka: wondering if it would be enough to just call it "ratelimit". Are we preserving the symbol "ratelimit" for another variant of the rate limiters?

ratelimit/ratelimit.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (l voidRatelimit) Allow(string) bool {
+	return true
+}
+
+func newRatelimit(s Settings) *Ratelimit {
+	var impl implementation
+	switch s.Type {
+	case LocalRatelimit:
+		impl = circularbuffer.NewRateLimiter(s.MaxHits, s.TimeWindow, s.CleanInterval, nil)
```
aryszka: calling circularbuffer.NewRateLimiter(..., nil): i would prefer to preserve the possibility to cleanup the implementation's internal goroutine, even if the typical case running only one in the process' life cycle.

aryszka: the *Ratelimit type could have a Close method, that calls a close() on the passed in quitter channel.

cmd/skipper/ratelimiterflags.go
```diff
@@ -0,0 +1,77 @@
...
+				return err
+			}
+			s.MaxHits = i
+		case "timeWindow":
+			d, err := parseDurationFlag(kv[1])
+			if err != nil {
+				return err
+			}
+			s.TimeWindow = d
+			s.CleanInterval = d * 10
```
aryszka: i see this `window * 10` in multiple places. Isn't it enough to rely on `ratelimit.DefaultCleanInterval`?

ratelimit/ratelimit.go
szuecs: Where should the quit channel be passed from? proxy.go or ratelimit.go or even main package?

filters/ratelimit/ratelimit.go
szuecs: Good question. 
I thought about having "local" prefix, because it's instance local. If you think it makes sense to drop that prefix, I am ok.

cmd/skipper/ratelimiterflags.go
szuecs: the problem here is, that cleaninterval should be dependent on timewindow, ratelimit.DefaultCleanInterval is not

aryszka: ok, i see

filters/ratelimit/ratelimit.go
aryszka: ah! OK, i see, let's keep the prefix then.

ratelimit/ratelimit.go
aryszka: this `ratelimit.Ratelimit` is the "owner" of the circularbuffer.ratelimiter created by `circularbuffer.NewRateLimiter`, so it should have its own close method, and whoever creates instances of `ratelimit.Ratelimit`, can call ratelimit.Ratelimit.Close() when doesn't need it anymore.

proxy/proxy.go
```diff
@@ -555,13 +567,52 @@ func (p *Proxy) checkBreaker(c *context) (func(bool), bool) {
...
+		code: http.StatusTooManyRequests,
+		additionalHeader: http.Header{
+			ratelimit.Header: []string{strconv.Itoa(settings.MaxHits * int(time.Hour/settings.TimeWindow))},
+		},
+	}
+}
+
+// TODO(sszuecs) not in use for the global rate limiter, but should be
+// used in per route rate limiting
+func (p *Proxy) checkRatelimit(c *context) bool {
```
aryszka: this function is not called now, let's delete it, and document the idea rather next to the Lookuper.

proxy/proxy.go
```diff
@@ -693,6 +744,10 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 		ctx.responseWriter.Header().Set("X-Circuit-Open", "true")
 	case err == errRouteLookupFailed:
 		code = p.defaultHTTPStatus
+	case perr.err == errRatelimitError:
```
aryszka: the additionalHeader field on the proxyError type could be used in a more generic way, not only for the rate limit errors, even if currently we only use for that rate limiter response. How about copying all the headers from the additionalHeader with the copyHeader function to the response whenever the error is of type proxyError? (A small bonus refactoring could be to do the same to the circuit breaker error, by moving the above "X-Circuit-Open" header to the additionHeader field where the errCircuitBreakerOpen var is defined.)

ratelimit/ratelimit.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+// Type defines the type of the used breaker: consecutive, rate or
+// disabled.
+type Type int
+
+const (
+	// Header is
+	Header = "X-Rate-Limit"
+	// LocalRatelimitName is the name of the LocalRatelimit, which will be shown in log
```
aryszka: "LocalRatelimitName is the name of the LocalRatelimit, which..." ->
"LocalRatelimitName is the name of the LocalRatelimit filter, which..." ?

proxy/proxy.go
szuecs: I like the idea.
I think, I also have to check for "ok" if I access perr.err, which could be a nil ptr dereference, if not ok.

szuecs:
@aryszka I think I addressed all your requested changes

filters/ratelimit/ratelimit.go
mikkeloscar: @szuecs I'm still not completely sure what is meant by "instance local"? Do you mean it's local to one skipper instance in case you run multiple skippers as ingress?

In that case, then I would say it's obvious, and "local" is not needed.

proxy/proxy.go
```diff
@@ -555,13 +567,29 @@ func (p *Proxy) checkBreaker(c *context) (func(bool), bool) {
...
+}
+
 func (p *Proxy) do(ctx *context) error {
 	if ctx.loopCounter > p.maxLoops {
 		return errMaxLoopbacksReached
 	}
 
 	ctx.loopCounter++
 
+	if ok, settings := p.limiters.Check(ctx.request); !ok {
```
mikkeloscar: nit: it's normal in go to have the variable order switched like: `settings, ok := ..`

ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,98 @@
...
+// 	}
+
+// 	ss := s.String()
+// 	expect := "type=rate,host=www.example.org,window=300,failures=30,timeout=1m0s,half-open-requests=15,idle-ttl=1h0m0s"
+// 	if ss != expect {
+// 		t.Error("invalid ratelimit settings string")
+// 		t.Logf("got     : %s", ss)
+// 		t.Logf("expected: %s", expect)
+// 	}
+// }
```
mikkeloscar: Please remove commented code.

filters/ratelimit/ratelimit.go
szuecs: @mikkeloscar yes it's instance local and I am not sure if this is "obvious". I don't like the wording "obvious", because it always depends on your context.

szuecs:
Ratelimit per route is implemented in the last commit.
I also added the possibility to configure a way to select which part of an http.Request you can use to select which part is counted to be in the same bucket. For example http Headers "X.Forwarded-For" or "Authorization".

aryszka:
:+1:

szuecs:
:+1:

Issue #436, opened 2017-09-25
Proper 404 catchall for ingress routes.


mikkeloscar:
#425 introduced a 404 catchall route for ingresses without a default catchall path.

The catch all rule were per ingress object, causing a problem where the same hostname is configured in multiple different ingress objects.

Example of the problem:

```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: test-1
spec:
  rules:
  - host: example.org
    http:
      paths:
      - backend:
          serviceName: my-svc
          servicePort: 80
        path: /path1

---

apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: test-2
spec:
  rules:
  - host: example.org
    http:
      paths:
      - backend:
          serviceName: my-svc
          servicePort: 80
```

The ingress definition for `test-1` would generate a catchall 404 route since only `/path1` is specified. The catch all would shadow the route created from the `test-2` ingress definition.

The catch all route should instead be per hostname rather than per ingress object.

PR #437, opened 2017-09-25
Remove catchAll route for kube ingress

mikkeloscar:
Removes the `catchAll` route for Kubernetes ingress which was introduced
in #425.

The catchall was only per ingress object, but since the same hostname can be specified for multiple ingresses the catchall can overwrite a valid route in that case.

Example of the problem:

```yaml
kind: Ingress
metadata:
  name: test-1
spec:
  rules:
  - host: example.org
    http:
      paths:
      - backend:
          serviceName: my-svc
          servicePort: 80
        path: /path1

---

kind: Ingress
metadata:
  name: test-2
spec:
  rules:
  - host: example.org
    http:
      paths:
      - backend:
          serviceName: my-svc
          servicePort: 80
```

The ingress definition for `test-1` would generate a catchall 404 route since only `/path1` is specified. The catch all would shadow the route created from the `test-2` ingress definition.

I think it's ok to remove this catchall for now since the behaviour is the same as long as you don't run with a custom value for `--default-http-status`.

I have created #436 to track properly implementing a catchall which is per hostname and not per ingress object.

Let me know what you think @aryszka @szuecs

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -466,8 +466,6 @@ func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
 			// update Traffic field for each backend
 			computeBackendWeights(backendWeights, rule)
 
-			hostRoutes := make([]*eskip.Route, 0, len(rule.Http.Paths))
-
 			for _, prule := range rule.Http.Paths {
 				if prule.Backend.Traffic > 0 {
 					r, err := c.convertPathRule(i.Metadata.Namespace, i.Metadata.Name, rule.Host, prule, servicesURLs)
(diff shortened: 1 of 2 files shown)


aryszka:
@mikkeloscar another way might be to always set a path for the non-catchall routes, e.g. defaulting to PathRegexp(".*"). This way the catchall routes would have only a Host() while the explicit ones would have a Host and PathRegexp and would not get shadowed by the catchall. Can this way be better?

mikkeloscar:
@aryszka sorry I don't follow your suggestion. Could you illustrate what the resulting routes would be in eskip?

aryszka:
it would be something like:

```
explicit1: PathRegexp("/path1") && Host("example.org") -> ...
catchall_for_example_org: Host("example.org") -> ...

explicit2: PathRegexp(".*") && Host("example.org") -> ...
catchall_for_example_org: Host("example.org") -> ...
```

I'd agree, though, that the `PathRegexp(".*")` is not very nice.

(the double ids of catchall_for_example_org are automatically resolved to a single route, but it would be of course nicer to avoid them)

hjacobs:
@mikkeloscar I'm trying to assess the impact of this change, could you give me a brief overview of **why** we do this change and what the impact is? AFAIK we do not use any path rules (in Zalando) at the moment.

mikkeloscar:
@hjacobs We saw an issue with one service (sandbox iam) which uses two ingresses for a single hostname, one with no paths defined and one with *only* paths defined (tokeninfo part) similar to the example I showed.

The impact was that only tokeninfo part worked while all other paths got 404.

mikkeloscar:
Closing this in favor of #440 which has a real fix that creates a catchall route per host instead of per ingress object.

PR #439, opened 2017-09-26
Document rate limiters

szuecs:
add rate limiter package godocs and in the checkpoint list in readme.md

### DIFF
--- a/ratelimit/doc.go
+++ b/ratelimit/doc.go
@@ -21,19 +21,19 @@ one most users want to use.
 Usage
 
 When imported as a package, the Registry can be used to hold the rate
-limiters and their settings On a higher level, rate limiter settings
+limiters and their settings. On a higher level, rate limiter settings
 can be simply passed to skipper as part of the skipper.Options object,
 or defined as command line flags.
 
 The following command starts skipper with default X-Forwarded-For
 Lookuper, that will start to rate limit after 5 requests within 60s
-from the same requester
+from the same client
 
     % skipper -ratelimits type=local,max-hits=5,time-window=60s
 
 The following configuration will rate limit /foo after 2 requests
 within 90s from the same requester and all other requests after 20
-requests within 60s from the same requester
+requests within 60s from the same client
 
     % cat ratelimit.eskip
     foo: Path("/foo") -> localRatelimit(2,"1m30s") -> "http://www.example.org/foo"
@@ -74,9 +74,9 @@ used to group client requests. It accepts the default
 
 HTTP Response
 
-In case of rate limiting a client it send a HTTP status code 429 Too
-Many Requests and a header to the response, which shows the maximum
-request per hour (based on RFC 6585):
+In case of rate limiting, the HTTP response status will be 429 Too
+Many Requests, and a header will be set which shows the maximum
+requests per hour (based on RFC 6585):
 
      X-Rate-Limit: 6000
 
--- a/ratelimit/doc.go
+++ b/ratelimit/doc.go
@@ -1,7 +1,7 @@
 /*
 Package ratelimit implements rate limiting functionality for the proxy.
 
-It provides per instance based, local, rate limiting. It can be
+It provides per process rate limiting. It can be
 configured globally, or based on routes. Rate limiting can be lookuped
 based on HTTP headers like X-Forwarded-For or Authorization.
 
--- a/ratelimit/doc.go
+++ b/ratelimit/doc.go
@@ -1,0 +1,93 @@
+/*
+Package ratelimit implements rate limiting functionality for the proxy.
+
+It provides per instance based, local, rate limiting. It can be
+configured globally, or based on routes. Rate limiting can be lookuped
+based on HTTP headers like X-Forwarded-For or Authorization.
+
+Lookuper Type - Authorization Header
+
+This lookuper will use the content of the Authorization header to
+calculate rate limiting. This will work for Bearer tokens or Basic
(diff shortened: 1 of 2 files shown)


ratelimit/doc.go
```diff
@@ -0,0 +1,93 @@
+/*
+Package ratelimit implements rate limiting functionality for the proxy.
+
+It provides per instance based, local, rate limiting. It can be
+configured globally, or based on routes. Rate limiting can be lookuped
```
lmineiro: weird term. Does this stand for bucket?

ratelimit/doc.go
```diff
@@ -0,0 +1,93 @@
...
+based on HTTP headers like X-Forwarded-For or Authorization.
+
+Lookuper Type - Authorization Header
```
lmineiro: This seems to be bucketing based on Auth header, is that it?

ratelimit/doc.go
```diff
@@ -0,0 +1,93 @@
+/*
+Package ratelimit implements rate limiting functionality for the proxy.
+
+It provides per instance based, local, rate limiting. It can be
```
lmineiro: "It provides per process rate limiting."

ratelimit/doc.go
szuecs: It's not the bucket, but it is used to select a bucket. 
For example X-Forwarded-For Lookuper will select the bucket based on the content of this header or the remote ip, in case there is no such header.

I am open to other suggestions.

ratelimit/doc.go
szuecs: bucketing? This sounds more weird than lookup to me. 
You also "lookup" a value from a hashmap, not "bucket" or "bucketing" a value from a hashmap.

ratelimit/doc.go
```diff
@@ -0,0 +1,93 @@
...
+
+This lookuper will use the remote IP of the origin request to
+calculate rate limiting. If there is no such header it will use the
+remote IP of the request. This is the default Lookuper and may be the
+one most users want to use.
+
+Usage
+
+When imported as a package, the Registry can be used to hold the rate
+limiters and their settings On a higher level, rate limiter settings
```
aryszka: On -> on

ratelimit/doc.go
```diff
@@ -0,0 +1,93 @@
...
+or defined as command line flags.
+
+The following command starts skipper with default X-Forwarded-For
+Lookuper, that will start to rate limit after 5 requests within 60s
+from the same requester
+
+    % skipper -ratelimits type=local,max-hits=5,time-window=60s
+
+The following configuration will rate limit /foo after 2 requests
+within 90s from the same requester and all other requests after 20
```
aryszka: "requester" -> "client"? not sure though...

ratelimit/doc.go
```diff
@@ -0,0 +1,93 @@
...
+
+Settings - Lookuper
+
+Defines an optional configuration to choose which Header should be
+used to group client requests. It accepts the default
+"x-forwarded-for" or "auth"
+
+HTTP Response
+
+In case of rate limiting a client it send a HTTP status code 429 Too
```
aryszka: "In case of rate limiting, the HTTP response status will be 429 Too Many Requests, and a header will be set which shows the maximum requests per hour (based on RFC 6585):

    X-Rate-Limit: 6000
"

?

aryszka:
Lookuper: we simply used the go way of calling interfaces that contain a single method, because we couldn't figure a better name :) Suggestions very welcome if one comes to your mind

ratelimit/doc.go
szuecs: I lost the ".", but thanks to spot it

ratelimit/doc.go
szuecs: It was probably to late for me to write.

szuecs:
From my side, I addressed all concerns.
The Lookuper interface is from my side good enough documented.

aryszka:
:+1:

ratelimit/doc.go
```diff
@@ -0,0 +1,93 @@
...
+be "local". In case of a skipper swarm or service mesh this would be
+an interesting configuration option, for example "global" or "dc".
+
+Settings - MaxHits
+
+Defines the maximum number of requests per user within a TimeWindow.
+
+Settings - TimeWindow
+
+Defines the time window until rate limits will be forced, if maximum
```
Joneser: "forced" -> "enforced"?

ratelimit/doc.go
```diff
@@ -0,0 +1,93 @@
...
+number of requests are exceeded. This is defined as string
```
Joneser: "as string" -> "as a string"

ratelimit/doc.go
```diff
@@ -0,0 +1,93 @@
...
+Many Requests, and a header will be set which shows the maximum
+requests per hour (based on RFC 6585):
+
+     X-Rate-Limit: 6000
+
+Registry
+
+The active rate limiters are stored in a registry. They are created
+based on routes or command line flags. The registry synchronizes
+access to the shared rate limiters. A registry has default settings it
```
Joneser: "settings it will apply" -> "settings that it will apply"

szuecs:
:+1:

Joneser:
:+1:

szuecs:
thanks @Joneser, everything fixed

aryszka:
:+1:

PR #440, opened 2017-09-30
Create catchall per host instead of per ingress

### LINKED ISSUE #436: Proper 404 catchall for ingress routes.

mikkeloscar:
Instead of creating a catchall route per ingress object, aggregate all ingress objects and only create a catchall route per host. This solves the problem of having multiple ingress objects with the same hostname defined where a catchall for one ingress object can shadow a catchall of another ingress object as detailed in #437.

This is the real fix and makes #437 obsolete.

Fixes #436

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -428,6 +428,7 @@ func (c *Client) convertPathRule(ns, name, host string, prule *pathRule, service
 // - check how to set failures in ingress status
 func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
 	routes := make([]*eskip.Route, 0, len(items))
+	hostRoutes := make(map[string][]*eskip.Route)
 	for _, i := range items {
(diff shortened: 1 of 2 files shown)


lmineiro:
👍

szuecs:
👍

PR #442, opened 2017-10-02
Check length of flag.Args() to ensure the were parsed correctly

mikkeloscar:
This checks the length of `flag.Args()` which will be non-empty in case some flags were not parsed. This is to ensure that you don't accidentally run skipper with the wrong args.

We had an issue were the args were specified like: `./skipper skipper -kubernetes ..` which means non of the args are parsed because the `skipper` command is invalid meaning it's the same as running `./skipper`. With this check skipper would crash rather then running with the unintended default configuration.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #444, opened 2017-10-02
Provide access to backend metrics from filters

### LINKED ISSUE #443: Provide access to backend metrics from filters
Filters should have access to more information about the call to the backend, at response time. At least the following metrics should be available.

- `send` - Time spent in Skipper before sending the request to the backend
- `wait` - Time spent waiting for a response from the backend
- `receive` - Time spent receiving the response from the backend (including any posible parsing)

For example, I'd like to be able to write a filter that sends all request/response pairs to a third party service such as [Galileo](https://getgalileo.io/) and to fulfill their API I need access to the send, wait, receive timings of the backend call.

grahamar:
Record send, wait & receive timings for backend calls and provide them
in the filter context for filters to use if they wish.

- Fixes #443

### DIFF
--- a/proxy/context.go
+++ b/proxy/context.go
@@ -33,8 +33,15 @@ type context struct {
 	incomingDebugResponse *http.Response
 	loopCounter           int
 	startServe            time.Time
+	proxyMeta             *proxyMetadata
 	metrics               *filterMetrics
 	tracer                opentracing.Tracer
+}
+
+type proxyMetadata struct {
+	sendingTime   time.Duration
+	waitingTime   time.Duration
+	receivingTime time.Duration
 }
 
 type filterMetrics struct {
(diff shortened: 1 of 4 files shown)


proxy/context.go
```diff
@@ -216,6 +225,14 @@ func (c *context) clone() *context {
 	return &cc
 }
 
+func (c *context) setProxyMetadata(sent time.Duration, wait time.Duration, recieve time.Duration) {
```
szuecs: recieve -> receive

proxy/proxy.go
```diff
@@ -622,6 +622,7 @@ func (p *Proxy) do(ctx *context) error {
 		return rerr
 	}
 
+	var sent, wait, recieve time.Duration = 0, 0, 0
```
szuecs: recieve -> receive

proxy/proxy.go
```diff
@@ -680,16 +681,22 @@ func (p *Proxy) do(ctx *context) error {
 			return err
 		}
 
+		recieveStart := time.Now()
```
szuecs: revieve -> receive

PR #445, opened 2017-10-04
Add visualization Tree built from pathmux

ruiaraujo:
The visualization tree is basically an explosion of the radix tree used in pathmux.

It looks like a tree representation of a path broken down by its slashes. However, it is built from the original pathmux Tree.

It is included in the pathmux so it can access the inner representation of the tree.

### DIFF
--- a/pathmux/viztree.go
+++ b/pathmux/viztree.go
@@ -1,41 +1,53 @@
 package pathmux
 
-type vizNode struct {
-	path     string
-	children []*vizNode
+// Exploded version of the pathmux tree designed for being easy to use in a visualization.
+// Simple wildcard nodes are represented by the ':' prefix and free wildcard nodes with the '*' prefix.
+type VizTree struct {
+	Path     string // string representation of the node path
+	children []*VizTree
 	canMatch bool
 
 	// The wildcard names still waiting to be processed, should be always nil after the end of the conversion.
 	leafWildcardNames []string
 }
 
-type VizTree vizNode
-
+// Creates a new visualization tree from a pathmux.Tree.
 func NewVizTree(tree *Tree) *VizTree {
 	vizTree := aggregateTree((*node)(tree), "/")
-	return (*VizTree)(vizTree[0])
+	return vizTree[0]
 }
 
-func aggregateTree(tree *node, previousPath string) []*vizNode {
+// Return a child referenced by direct Path
+//
+// Example:
+//
+// - in a tree such as: / -> images -> abc
+//
+// looking for images in the root tree would yield the images subtree.
+func (n *VizTree) Child(path string) *VizTree {
+	return findNode(n.children, path)
+}
+
+func aggregateTree(tree *node, previousPath string) []*VizTree {
 	if tree == nil {
 		return nil
 	}
-	var currentNode, middleNode *vizNode
+	var currentNode, middleNode *VizTree
 	nextPath := previousPath + tree.path
-	// On every slash we create new nodes if the previous path has a value, otherwise we ignore it
+	// On every slash we create new nodes if the previous Path has a value, otherwise we ignore it
 	if tree.path == "/" {
 		if previousPath != "" {
-			currentNode = &vizNode{previousPath, nil, tree.leafValue != nil, tree.leafWildcardNames}
+			currentNode = &VizTree{previousPath, nil, tree.leafValue != nil, tree.leafWildcardNames}
 		}
 		nextPath = ""
 	} else {
 		// This is to handle intermediate nodes that are matched without a slash
 		if tree.leafValue != nil {
-			middleNode = &vizNode{nextPath, nil, true, tree.leafWildcardNames}
+			middleNode = &VizTree{nextPath, nil, true, tree.leafWildcardNames}
 		}
 	}
 	numberOfChildren := len(tree.staticChild)
-	children := make([]*vizNode, 0, numberOfChildren)
+	children := make([]*VizTree, 0, numberOfChildren)
 	for i := 0; i < numberOfChildren; i++ {
 		childrenNodes := aggregateTree(tree.staticChild[i], nextPath)
 		children = append(children, childrenNodes...)
@@ -46,18 +58,18 @@ func aggregateTree(tree *node, previousPath string) []*vizNode {
 	children = append(children, childrenNodes...)
 
 	if tree.catchAllChild != nil && tree.catchAllChild.leafValue != nil {
-		children = append(children, &vizNode{"*" + tree.catchAllChild.path, nil, true, sliceLeafNames(tree.catchAllChild.leafWildcardNames)})
+		children = append(children, &VizTree{"*" + tree.catchAllChild.path, nil, true, sliceLeafNames(tree.catchAllChild.leafWildcardNames)})
 	}
 	if currentNode != nil {
 		if leafWildcardNames != nil {
 			currentNode.leafWildcardNames = leafWildcardNames
 		}
 		currentNode.children = children
-		return []*vizNode{currentNode}
+		return []*VizTree{currentNode}
 	}
 	if middleNode != nil {
 		// Prevent duplication of intermediate nodes
-		if previousNode := getNode(children, middleNode.path); previousNode == nil {
+		if previousNode := findNode(children, middleNode.Path); previousNode == nil {
 			if leafWildcardNames != nil {
 				middleNode.leafWildcardNames = leafWildcardNames
 			}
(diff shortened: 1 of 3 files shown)


lmineiro:
go fmt

szuecs:
@ruiaraujo Can you fix the build?
I think you have to remove the opentracing workaround, which should not be used anymore.

ruiaraujo:
👍

ruiaraujo:
@szuecs Build fixed.

szuecs:
@ruiaraujo I also miss documentation :)

ruiaraujo:
@szuecs Docs added.

szuecs:
@ruiaraujo thanks for adding this docs, but I wonder how can I use this. I guess it's meant for library users or if you want to enhance the output of http://localhost:adminport/routes. Is my assumption correct? 
Can you find a place to add some documentation like this. Thanks!

ruiaraujo:
@szuecs It is not really defined yet how we are going to use this.

This was a first step in the investigation of visualising the skipper routes.
Since it was a first step, I have only done the code for the very basic data structure.

To be honest I am not aware of the output of locahost:adminport/routes so I have not looked into how this would fit with that.

szuecs:
:+1:

ruiaraujo:
+1

ruiaraujo:
👍

pathmux/viztree_test.go
```diff
@@ -0,0 +1,117 @@
...
+	addPathToTree(t, tree, "/:page")
+	addPathToTree(t, tree, "/:page/:index")
+	addPathToTree(t, tree, "/post/:post/page/:page")
+	addPathToTree(t, tree, "/plaster")
+	addPathToTree(t, tree, "/users/:pk/:related")
+	addPathToTree(t, tree, "/users/:id/updatePassword")
+	addPathToTree(t, tree, "/:something/abc")
+	addPathToTree(t, tree, "/:something/def")
+
+	vizTree := (*VizTree)(NewVizTree(tree))
```
aryszka: `NewVizTree(tree)` returns a `*VizTree`, so no need for casting

pathmux/viztree_test.go
```diff
@@ -0,0 +1,117 @@
...
+	testChildren(t, vizTree.Child("users"), []string{":pk", ":id"})
+	testChildren(t, vizTree.Child("users").Child(":pk"), []string{":related"})
+	testChildren(t, vizTree.Child("users").Child(":id"), []string{"updatePassword"})
+	testChildren(t, vizTree.Child("post").Child(":post").Child("page"), []string{":page"})
+	testChildren(t, vizTree.Child("date"), []string{":year"})
+	testChildren(t, vizTree.Child("date").Child(":year"), []string{":month", "month"})
+	testChildren(t, vizTree.Child("date").Child(":year").Child(":month"), []string{":post", "abc", "*post"})
+}
+
+func testChildren(t *testing.T, tree *VizTree, expectedChildren []string) {
```
aryszka: expectedChildren is a great use case for varargs, makes testChildren() calls simpler

pathmux/viztree.go
```diff
@@ -0,0 +1,148 @@
...
+			if removeFromChildren {
+				children[i].leafWildcardNames = nil
+			}
+		}
+	}
+	return leafWildcardNames
+}
+
+func sliceLeafNames(names []string) []string {
+	if names == nil || len(names) < 2 {
```
aryszka: no need to check for nil. len(names) will return 0 if the slice is nil. This applies to multiple similar occurences in the code.

pathmux/viztree_test.go
```diff
@@ -0,0 +1,117 @@
...
+	if t.Failed() {
+		t.FailNow()
+	}
+	t.Log("Testing children of", tree.Path)
+	children, err := childrenString(tree)
+	if err != nil || children == nil {
```
aryszka: it's better to use `len(children) == 0` instead of `children == nil`. That way it is possible everywhere to use var slice []something or slice := make([]something) without caring about the calling code or the need for an additional contract between the different functions

pathmux/viztree_test.go
```diff
@@ -0,0 +1,117 @@
...
+		if tree.children[i].leafWildcardNames != nil {
+			t.Errorf("leafWildcardNames different from nil")
+			return
+		}
+	}
+	sort.Strings(expectedChildren)
+	sort.Strings(children)
+	if !reflect.DeepEqual(children, expectedChildren) {
+		t.Errorf("Children  %v are different \n from expected children %v", children, expectedChildren)
+		return
```
aryszka: unnecessary return

pathmux/viztree_test.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+}
+func testIfChildMatch(t *testing.T, tree *VizTree, childPath string) {
+	if t.Failed() {
+		t.FailNow()
+	}
+	t.Logf("Looking for child %v of %v", childPath, tree.Path)
+	if tree.Child(childPath) == nil {
+		t.Errorf("No children %v found in %v", childPath, tree.Path)
+		return
```
aryszka: unnecessary return

pathmux/viztree.go
```diff
@@ -0,0 +1,148 @@
...
+func aggregateTree(tree *node, previousPath string) []*VizTree {
+	if tree == nil {
+		return nil
+	}
+	var currentNode, middleNode *VizTree
+	nextPath := previousPath + tree.path
+	// On every slash we create new nodes if the previous Path has a value, otherwise we ignore it
+	if tree.path == "/" {
+		if previousPath != "" {
+			currentNode = &VizTree{previousPath, nil, tree.leafValue != nil, tree.leafWildcardNames}
```
aryszka: with this many fields, it is more readable to write out the field names in the struct literal. Instead of:

```
&VizTree{nextPath, nil, true, tree.leafWildcardNames}
```

it would better if:

```
&VizTree{
    Path: nextPath,
    canMatch: true,
    leafWildcardNames: tree.leafWildcardNames,
}
```

pathmux/viztree.go
```diff
@@ -0,0 +1,148 @@
...
+	return nil
+}
+
+func getLeafNames(children []*VizTree, removeFromChildren bool) []string {
+	var leafWildcardNames []string
+	for i := 0; i < len(children); i++ {
+		if children[i].leafWildcardNames != nil {
+			leafWildcardNames = children[i].leafWildcardNames
+			if removeFromChildren {
+				children[i].leafWildcardNames = nil
```
aryszka: i think the removal from the children should be a separate function, or at least this one should have a better name than `getLeafNames`

pathmux/viztree.go
```diff
@@ -0,0 +1,148 @@
...
+		child := children[i]
+		if path == child.Path {
+			return child
+		}
+	}
+	return nil
+}
+
+func getLeafNames(children []*VizTree, removeFromChildren bool) []string {
+	var leafWildcardNames []string
```
aryszka: this function is not completely clear to me. Should return the leafWildcardNames of the last child whose leafWildcardnames is not empty and also remove them if required?

pathmux/viztree.go
```diff
@@ -0,0 +1,148 @@
...
+		if leafWildcardNames != nil {
+			currentNode.leafWildcardNames = leafWildcardNames
+		}
+		currentNode.children = children
+		return []*VizTree{currentNode}
+	}
+	if middleNode != nil {
+		// Prevent duplication of intermediate nodes
+		if previousNode := findNode(children, middleNode.Path); previousNode == nil {
+			if leafWildcardNames != nil {
```
aryszka: it is better to check for `len(slice) != 0`

pathmux/viztree.go
```diff
@@ -0,0 +1,148 @@
...
+				middleNode.leafWildcardNames = leafWildcardNames
+			}
+			children = append(children, middleNode)
+		} else {
+			previousNode.canMatch = true
+			if leafWildcardNames != nil {
```
aryszka: it is better to check for `len(slice) != 0`

pathmux/viztree.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	return children
+}
+
+func processWildCardNode(tree *node) []*VizTree {
+	if tree == nil {
+		return nil
+	}
+	numberOfChildren := len(tree.staticChild)
+	if tree.leafWildcardNames != nil && numberOfChildren == 0 {
```
aryszka: it is better to check for `len(slice) != 0`

pathmux/viztree.go
```diff
@@ -0,0 +1,148 @@
...
+			return child
+		}
+	}
+	return nil
+}
+
+func getLeafNames(children []*VizTree, removeFromChildren bool) []string {
+	var leafWildcardNames []string
+	for i := 0; i < len(children); i++ {
+		if children[i].leafWildcardNames != nil {
```
aryszka: it is better to check for `len(slice) != 0`

pathmux/viztree.go
```diff
@@ -0,0 +1,148 @@
...
+		currentPath, leafWildcardNames := tree.leafWildcardNames[len(tree.leafWildcardNames)-1], sliceLeafNames(tree.leafWildcardNames)
+		return []*VizTree{{":" + currentPath, nil, tree.leafValue != nil, leafWildcardNames}}
+	}
+	children := make([]*VizTree, 0, numberOfChildren)
+	for i := 0; i < numberOfChildren; i++ {
+		childrenNodes := aggregateTree(tree.staticChild[i], "")
+		for j := 0; j < len(childrenNodes); j++ {
+			child := childrenNodes[j]
+			if child.leafWildcardNames != nil {
+				currentPath, leafWildcardNames := ":"+child.leafWildcardNames[len(child.leafWildcardNames)-1], sliceLeafNames(child.leafWildcardNames)
```
aryszka: this looks like could also be just separate lines

pathmux/viztree.go
```diff
@@ -0,0 +1,148 @@
+package pathmux
+
+// Exploded version of the pathmux tree designed for being easy to use in a visualization.
+// Simple wildcard nodes are represented by the ':' prefix and free wildcard nodes with the '*' prefix.
+type VizTree struct {
+	Path     string // string representation of the node path
+	children []*VizTree
```
aryszka: wondering if it would be more useful to make the `children` field visible to the calling code. It would make the viztree walkable. Is there a reason not to do it?

aryszka:
:+1:

ruiaraujo:
👍

aryszka:
:+1:

PR #448, opened 2017-10-12
Add redirectToLower filter for case-insensitive redirects

### LINKED ISSUE #328: Skipper should allow URLs to be treated as case-insensitive using MatchingOption
The current default for `skipper` is to allow URLs to be case-sensitive. This creates a requirement for all the route definition to have all combinations of upper and lowercase for a given route. 

Skipper should allow `IgnoreCaseSensitive` as a `MatchingOptions` so that we don't need to define multiple routes.

E.g. The wishlist route works for www.zalando.de/wishlist but not for www.zalando.de/WISHLIST

addityasingh:
- Close #328
- Add `redirectToLower` filter for case-insensitive redirects
- Add enum `redirectType` to differentiate between redirect filter types
- Fix syntax and `gofmt` warnings

### DIFF
--- a/filters/builtin/builtin.go
+++ b/filters/builtin/builtin.go
@@ -31,17 +31,18 @@ const (
 	DropRequestHeaderName    = "dropRequestHeader"
 	DropResponseHeaderName   = "dropResponseHeader"
 
-	HealthCheckName  = "healthcheck"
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

aryszka:
:+1:

Issue #449, opened 2017-10-12
Allow non-TTY applications to execute eskip commands


last-ent:
**It would be great** if we can execute eskip commands in ssh sessions and as Vim commands.

**This would enable us** to include eskip checks as part of a CD/CI pipeline.

**Rationale:**
At the moment if we try to run `eskip check routes.eskip` within Jenkins/CDP, this would give us the error " too many inputs". This is caused by [processStdin](https://github.com/zalando/skipper/blob/8fb107612ac6543f1c76d01bb3e8efe61dac5beb/cmd/eskip/args.go#L205) which expects a `stdin` to be available but this is not always possible because 
* ssh sessions created as remote sessions are not allocated TTY, this is the case for Jenkins and CDP.
*  trying to execute eskip as shell commands from within Vim fails on Mac - `:!eskip check routes.eskip`. I couldn't reproduce the issue with Linux + Vim.

**Workaround:** Currently the way to tackle this issue is
```bash
@RESULT=$$(cat routes.eskip | eskip check 2>&1);\
if [ $$? != 0 ]; then\
	echo "not ok 1 - $$RESULT";\
	false;\
else\
	echo "ok 1 - eskip check";\
fi
```
This also works in Vim shell as `:! cat routes.eskip | eskip check 2>&1`

However this is quite inconvenient and I would like avoid using bash when possible. **Note** Following command will not stop CD/CI script - `:! cat does-not-exist | eskip check 2>&1`

**Proposed Solution:**
I would like to suggest introducing a flag which we can pass to eskip.
Something like this - `eskip -no-tty check routes.eskip`. I have already tested it with CDP and it works fine.
Following is the PR with proposed changes: https://github.com/zalando/skipper/pull/450

last-ent:
fixed

PR #450, opened 2017-10-12
Allow non-TTY applications to execute eskip commands

last-ent:
ref: https://github.com/zalando/skipper/issues/449

Introduce new flag: -no-tty

### DIFF
--- a/cmd/eskip/args.go
+++ b/cmd/eskip/args.go
@@ -17,10 +17,11 @@ package main
 import (
 	"errors"
 	"flag"
-	"golang.org/x/crypto/ssh/terminal"
 	"net/url"
 	"os"
 	"strings"
+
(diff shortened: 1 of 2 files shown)


aryszka:
:+1:

mikkeloscar:
Is there a use case for both reading from stdin *and* a filepath (given as argument)? Otherwise it might make sense to just not read from stdin when a filepath is passed as an argument?

aryszka:
@last-ent what do you think about @mikkeloscar 's suggestion? Would it work for your use case? In case yes, can you please update the PR? :)

szuecs:
@last-ent please delete also the flag in eskip to fix the build:

cmd/eskip/args.go:209: undefined: hasNoTTY

aryszka:
:+1:

mikkeloscar:
:+1:

PR #455, opened 2017-10-15
fix metrics timing issue in metrics test check

aryszka:
fix metrics check issue:
- look only for the expected key
- wait while the key gets registered in the background


szuecs:
👍

aryszka:
:+1:

ruiaraujo:
:+1:

PR #456, opened 2017-10-16
Allow using an exponentially decaying sample for histograms

### LINKED ISSUE #363: backend latency metrics don't fall back after spikes
after backend latency spikes, the metrics go back to normal only after a very long time.

### LINKED ISSUE #428: Use exponentially decaying sample for metrics
In 739f413, the metrics were changed to use a `UniformSample` with a hard-coded 1024 reservoir size. This makes latency metrics for services with non-uniform load very skewed. For example, if our service is under a very high load (100 rps) due to a misconfiguration in another service, and then the load drops to normal (1-2 rps) after the misconfiguration is fixed, it can take hours, days or even weeks for the metrics to go back to normal. Would you consider switching back to the default exponentially decaying sample? I can send a PR if necessary.

aermakov-zalando:
Add -metrics-exp-decay-sample option. If set, metrics will use an exponentially decaying sample for the histograms instead of the uniform one. Once we're satisfied with the results, we can switch the defaults.

Fix #428.
Fix #363.

### DIFF
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -80,6 +80,7 @@ const (
 	routeBackendErrorCountersUsage = "enables counting backend errors for each route"
 	routeStreamErrorCountersUsage  = "enables counting streaming errors for each route"
 	routeBackendMetricsUsage       = "enables reporting backend response time metrics for each route"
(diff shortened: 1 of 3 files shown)


aryszka:
:+1:

Issue #463, opened 2017-10-20
Define arbitrary skipper filters via Ingress


szuecs:
Based on an internal Friday demo, I got feedback to enable setting of arbitrary skipper filters https://godoc.org/github.com/zalando/skipper/filters via Ingress. This is a superset of ratelimitting and makes totally sense.

As an example we could specify it like this:

    apiVersion: extensions/v1beta1
    kind: Ingress
    metadata:
      annotations:
        zalando.org/skipper-filter: localRatelimit(50, "10m") -> requestCookie("test-session", "abc")
      name: app
    spec:
      rules:
      - host: app-default.example.org
        http:
          paths:
          - backend:
              serviceName: app-svc
              servicePort: 80

PR #466, opened 2017-10-20
set arbitrary skipper filters from ingress annotation

szuecs:
this implements #463


szuecs:
Added tests, but we should change the use of testIngresses(), such that each test is based on it's own table. We always have to adjust many other tests, if we add a new test case.

aryszka:
rebased and reopened here: https://github.com/zalando/skipper/pull/468

PR #469, opened 2017-10-22
create a flag to disable detailed route update logging

aryszka:
when the -suppress-route-update-logs is set, only a summary of the rout changes will be logged

### DIFF
--- a/logging/loggingtest/logger.go
+++ b/logging/loggingtest/logger.go
@@ -14,9 +14,9 @@ type logSubscription struct {
 	notify chan<- struct{}
 }
 
-type countRequest struct {
-	exp      string
-	response chan<- int
+type countMessage struct {
+	expression string
+	response   chan<- int
 }
 
 type logWatch struct {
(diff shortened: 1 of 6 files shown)


logging/loggingtest/logger.go
```diff
@@ -14,6 +14,11 @@ type logSubscription struct {
 	notify chan<- struct{}
 }
 
+type countRequest struct {
+	exp      string
```
szuecs: what is "exp"?
I think you can find a better name. Based on the use it could be "substr", but the question is on what data of the Request is that used?

logging/loggingtest/logger.go
```diff
@@ -64,6 +70,17 @@ func (lw *logWatch) notify(req logSubscription) {
 	}
 }
 
+func (lw *logWatch) count(req countRequest) {
```
szuecs: req? This is not a Request :)

szuecs:
@aryszka I only dislike the naming of variables. code lgtm

aryszka:
updated

aryszka:
:+1:

szuecs:
:+1:

PR #471, opened 2017-10-23
add eskip json unmarshalling

aryszka:


grassator:
Looks good (for my limited understanding of Go). As for the tests, it might not be very idiomatic, but it might be nice to have some kind generative tests, that ensure:

```
1. json -> internal -> json === json
2. internal -> json -> internal === internal
```

aryszka:
ok, thanks, will add such tests

grassator:
@aryszka is this still something we want to have, if so I can pick it up probably, otherwise lets close the PR.

aryszka:
@grassator yes, we need this :smile:

grassator:
@aryszka I'm picking this up then.

grassator:
@aryszka I've rebased on top of master and had to make a few changes on the go. Most of them are still keeping API compatible, the only really breaking change if someone was creating `WatchClient` on their own.

eskip/json.go
```diff
@@ -57,28 +75,147 @@ func marshalJsonPredicates(r *Route) []*Predicate {
...
+}
+
+// MarshalJSON returns the JSON representation of a predicate.
+//
+// (Typically used only as part of the unmarshalling complete routes.)
 func (p *Predicate) MarshalJSON() ([]byte, error) {
 	return marshalNameArgs(p.Name, p.Args)
 }
 
+// UnmarshalJSON returns the JSON representation of a predicate.
```
szuecs: Are you sure that you don’t do it the other way around?
Marshal and unmarshal have very similar doc strings.

aryszka: I agree that the doc comment should rather say something like this:

"UnmarshalJSON parses a predicate object from its JSON representation."

eskipfile/watch.go
```diff
@@ -26,9 +27,14 @@ type WatchClient struct {
 
 // Watch creates a route configuration client with file watching. Watch doesn't follow file system nodes, it
 // always reads from the file identified by the initially provided file name.
-func Watch(name string) *WatchClient {
+func Watch(fileName string) *WatchClient {
+	return WatchFile(fileName, eskip.ParseBytes)
+}
+
+func WatchFile(fileName string, parseFunc eskip.ParseFunc) *WatchClient {
```
aryszka: the name of this function should reflect what's the difference from the simple Watch(). My candidate is: WatchParser()

eskip/eskip.go
```diff
@@ -334,6 +336,11 @@ func Parse(code string) ([]*Route, error) {
 	return routeDefinitions, nil
 }
 
+// Parses a route expression or a routing document to a set of route definitions.
```
miry: According the https://golang.org/doc/effective_go.html#commentary comments should start with function name:

```suggestion
// ParseBytes parses a route expression or a routing document to a set of route definitions.
```

It some languages, to generate docs, function name is ignored, that's why doc comments starts with function name.

marcinzaremba:
https://github.com/zalando/skipper/tree/check-json-unmarshal

Issue #465, opened 2017-10-20
Define arbitrary skipper predicates via Ingress


szuecs:
Similar to https://github.com/zalando/skipper/issues/463 people might want to use Predicates https://godoc.org/github.com/zalando/skipper/predicates from ingress.

Requirements: https://github.com/zalando/skipper/issues/464

As an example we could specify it like this:

    apiVersion: extensions/v1beta1
    kind: Ingress
    metadata:
      annotations:
        zalando.org/skipper-predicate: Path("/myreport") && Between("2016-01-01T12:00:00+02:00", "2016-02-01T12:00:00+02:00")
      name: app
    spec:
      rules:
      - host: app-default.example.org
        http:
          paths:
          - backend:
              serviceName: app-svc
              servicePort: 80

PR #467, opened 2017-10-20
function for parsing predicates individually

aryszka:
an eskip.ParseFilters() function already exists, this is its pair for the predicates: ParsePredicates()

Closes: #464

PR #472, opened 2017-10-23
set predicates using ingress annotation

szuecs:
implements #465, set predicates for ingress via ingress annotation, test and docs added


aryszka:
i just realized that due to the legacy way of the core predicates like Method() and Host() are implemented, this won't fully work yet. This means that the parse functioin, implemented here: https://github.com/zalando/skipper/pull/467/files , is also wrong. I suggest to hold this PR and let's fix the parsing function first.

aryszka:
:+1:

szuecs:
:+1:

Issue #470, opened 2017-10-23
Rate limit for service backends


szuecs:
Right now the ratelimit can be used to rate limit individual clients based on Request data (for example X-Forwarded-For header). 
People might want to define an upper bound value for their service application and rate limit all requests if this is exceeding. The algorithm would be leaky bucket and choose per backend or per Host header.

PR #478, opened 2017-10-31
Feature/service ratelimit

szuecs:
Rate limit type service implements #470 a backend protection. 

It allows the definition of a maximum amount of requests that are passed to the backend without considering data from the client. It would be used like this:

    ratelimit(200, "1s")

The naming ratelimit() and not serviceRatelimit() indicates, that this is likely the default behaviour a user wants to have. Instead of the more complex localRatelimit(), which allows rate limiting based on client data.

### DIFF
--- a/filters/ratelimit/ratelimit.go
+++ b/filters/ratelimit/ratelimit.go
@@ -26,10 +26,10 @@ type filter struct {
 
 // NewLocalRatelimit creates a local measured rate limiting, that is
 // only aware of itself. If you have 5 instances with 20 req/s, then
-// it would allow 100 req/s to the backend. A third argument can be
-// used to set which part of the request should be used to find the
-// same user. Third argument defaults to XForwardedForLookuper,
-// meaning X-Forwarded-For Header.
+// it would allow 100 req/s to the backend from the same user. A third
+// argument can be used to set which part of the request should be
+// used to find the same user. Third argument defaults to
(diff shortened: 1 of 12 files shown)


filters/ratelimit/ratelimit.go
```diff
@@ -46,6 +46,19 @@ func NewLocalRatelimit() filters.Spec {
...
+//
+// Example:
+//
+//    backendHealthcheck: Path("/healthcheck")
+//    -> ratelimit(20, "1s")
+//    -> "https://foo.backend.net";
+func NewRatelimit() filters.Spec {
+	return &spec{typ: ratelimit.ServiceRatelimit, filterName: ratelimit.ServiceRatelimitName}
+}
+
```
ruiaraujo: ❤️

ratelimit/ratelimit.go
```diff
@@ -25,8 +27,13 @@ const (
 const (
 	// NoRatelimit is not used
 	NoRatelimit Type = iota
-	// LocalRatelimit is used to have a simple local rate limit,
-	// which is calculated and measured within each instance
+	// ServiceRatelimit is used to have a simple rate limit for a
```
aryszka: i would rephrase the scope of this rate limiter somehow. As i understand, it is a single global bucket, and it can be used in a backend specific way, if defined for a particular route.

aryszka:
:+1:

ruiaraujo:
:+1:

PR #479, opened 2017-11-01
create a data client for inline routes

aryszka:
- creates a DataClient implementation for taking routes as in-memory strings
- adds an InlineRoutes option in the root package
- adds command a line flag to set inline routes

### DIFF
--- a/dataclients/routestring/string.go
+++ b/dataclients/routestring/string.go
@@ -1,0 +1,31 @@
+// Package routestring provides a DataClient implementation for
+// setting route configuration in form of simple eskip string.
+package routestring
+
+import (
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/routing"
+)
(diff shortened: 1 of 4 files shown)


dataclients/routestring/string.go
```diff
@@ -0,0 +1,31 @@
+// Package routestring provides a DataClient implementation for
+// setting route configuration in form of simple eskip string.
```
szuecs: an example would be nice :)

aryszka: i'd like to add an example to this one using the inlineContent() filter from this PR: https://github.com/zalando/skipper/pull/480

so please let's have a look at the other one first

aryszka: done

ruiaraujo:
👍

aryszka:
:+1:

PR #480, opened 2017-11-02
create a filter for inline content

aryszka:
creates a filter that can be used to return static inline content, primarily for testing or demo


ruiaraujo:
👍

szuecs:
:+1:

PR #488, opened 2017-11-11
header copy filters 

### LINKED ISSUE #426: copy header filter
Header Filter should also have a CopyHeader(src, dest), such that the value of X-Foo can be copied to X-Bar. 
UseCase: change, for example open tracing, headers for migrations

szuecs:
requestCopyHeader() and responseCopyHeader() that can copy the content of header "src" to "dst" in the request or response pass of the proxy.

Closes #426


szuecs:
👍

aryszka:
:+1:

PR #492, opened 2017-11-18
Properly log the state of the circuits breakers.

### LINKED ISSUE #491: incorrect application logs on  rateBreakers
The skipper application logs do not correctly represent the actual state of the rate circuit breaker.
i.e. the log from line https://github.com/zalando/skipper/blob/master/circuit/ratebreaker.go#L80 is also issued when the underlying gobreaker circuitbreaker transitions to "half-open" state.

In our skipper instance the last observed log was:

```
circuit breaker closed: type=rate,host=our.host,window=300,failures=30,timeout=1m0s,half-open-requests=20,idle-ttl=30m0s
```

although, upon issueing requests, we received a 503 response with the `X-Circuit-Open: true` header.

It seems there is no logs when the circuit breaker transitions from "half-open" state to any other.
As there is currently no other way to observe the state of a circuit breaker, we, for now, need to back off from using it, until this is resolved.

Otherwise, skipper always served us well. Thank you!

ruiaraujo:
Closes #491

### DIFF
--- a/circuit/consecutivebreaker.go
+++ b/circuit/consecutivebreaker.go
@@ -7,7 +7,6 @@ import (
 
 type consecutiveBreaker struct {
 	settings BreakerSettings
-	open     bool
 	gb       *gobreaker.TwoStepCircuitBreaker
 }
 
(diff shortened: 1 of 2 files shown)


szuecs:
Looks much simpler now, good PR!

szuecs:
:+1:

ruiaraujo:
👍

aryszka:
❤️

PR #498, opened 2017-11-26
fix build for osx/win

szuecs:
use common git command to get version (sort -V) is not available on bsd like systems or you have to use gsort

added 2 build targets one for osx and one for windows


ruiaraujo:
👍

aryszka:
👍

Issue #490, opened 2017-11-14
Allow user to add custom metrics


last-ent:
While using skipper as a proxy, I would like to monitor some custom states that might change because of filters. I would like have them as part of the skipper's /metrics endpoint so that we can rely on a single endpoint to provide us with these new states as well.

For example consider that we have 3 filters calling on three external http endpoints as part of the filter logic and each of these filters has a circuit breaker around these http calls - CB1, CB2, CB3.
Let imagine that CB1 is in Closed State, CB2 is in Open State & CB3 is in Partially Open State.

As a part of my monitoring system, I would like to keep track of these CB states as well.
Right now I would need to expose them on another endpoint within my main proxy application. Instead having them on /metrics would be helpful.

Given that we create the `skipper.Options` and then  run the application as `skipper.Run(options)`. There doesn't seem to be a provision to add any extra metrics.

I propose that `skipper.Options` has a new field called `CustomMetrics` of type `map[string]interface{}` 

As a result, when I call on metrics endpoint I can expect something as follows.
```javascript
GET /metrics
{
// Standard set of metrics

// Custom Metrics
"CB1": "Closed",
"CB2": "Open",
"CB3": "Half Open",
}
```

lmineiro:
This has been [discussed before](https://github.com/zalando/skipper/issues/352#issuecomment-310083869).

Do you think what was contributed with https://github.com/zalando/skipper/pull/384 is enough for your needs?

last-ent:
@lmineiro the discussion in the ticket deals with adding counter & measuring duration for a given key. Unfortunately this does not fit our requirements as we need a way to update values for different keys/metrics.

aryszka:
the metrics.Default object allows custom keys for duration measures and counters. The same interface is exposed on the filter context, where the keys will be prefixed with the filter name. How I understand, what's missing:

1. allow other types of custom metrics, not only durations and counters

(2. not strictly belongs here, but it would be nice if the state of the breakers would be possible to monitor anyway)

@last-ent do you think 1. would cover what you want to achieve?

last-ent:
Yes, 1. is exactly what I am looking for and 2. is just a special case of 1. IMO.

Is this already available in Skipper? Otherwise I wouldn't mind working on this.

aryszka:
[...]

PR #501, opened 2017-11-27
(WIP) add methods DecCounter & GetOrRegister

last-ent:
GetOrRegister allows for custom metrics to be added that are not Counter or Timer.
DecCounter provides the alternative to IncCounter.

Closes https://github.com/zalando/skipper/issues/490

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 SOURCES            = $(shell find . -name '*.go' -not -path "./vendor/*")
 PACKAGES           = $(shell glide novendor || echo -n "./...")
-CURRENT_VERSION    = $(shell git tag | sort -V | tail -n1)
+CURRENT_VERSION    = $(shell git describe --tags --always --dirty)
 VERSION           ?= $(CURRENT_VERSION)
 NEXT_MAJOR         = $(shell go run packaging/version/version.go major $(CURRENT_VERSION))
 NEXT_MINOR         = $(shell go run packaging/version/version.go minor $(CURRENT_VERSION))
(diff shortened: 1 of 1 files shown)


szuecs:
Could you add tests please? :)

lmineiro:
This approach allows different filters to override each other metric, eventually causing non-safe operations like invalid casts. This is mostly due to issues [already discussed](https://github.com/zalando/skipper/issues/352#issuecomment-310083869) and the main reason why I think this should be changed.

A filter could create a Timer for the metric with key "foo" while another "gets or registers" a Counter with the same key - "foo".

Additionally, it will also allow virtually anything to be pushed as a metric because of the 'interface{}' signature, which can break metric serialization later. For this to be manually avoided, it leaks the metrics library, which is bad.

lmineiro:
The proposal is not calling [Register()](https://github.com/rcrowley/go-metrics/blob/master/registry.go#L129) is it? Isn't it calling [GetOrRegister()](https://github.com/rcrowley/go-metrics/blob/master/registry.go#L80) that will happily return the existing one?

lmineiro:
I'm thinking that we could explore the [PrefixedRegistry](https://github.com/rcrowley/go-metrics/blob/master/registry.go#L149) and add one for each unique filter on registration. The added memory overhead of separate registries shouldn't be too much.

This would add more complexity to the metric serialization though, where there's already a different prefix strategy, although a global one.

last-ent:
Ah, my mistake. You are right.
Can I suggest using another type?

```go
type CustomMetrics interface {
    metrics map[string]interface{}
    Snapshot() CustomMetrics
    Upsert(key, interface{})  // or Update
}
```

This should allow us to handle the custom metrics consistent with rest of the metrics.

lmineiro:
I still prefer the namespacing solution. Other alternatives have the potential of runtime errors (invalid cast).

Filters shouldn't care about raw metrics and we definitely don't want to leak the metrics library to filters. This would prevent us from evolving or changing which one is used.

@last-ent would you consider contributing the prefixed registry for each filter on filter registration and the corresponding serialization? I could help with further ideas around this approach.

last-ent:
👍

However one caveat is that I might not be able to work on this during December. If that's not going to be an issue, I am all up for it!

lmineiro:
I think the other maintainers will agree when I say that we're thankful for your help and would be delighted if you could contribute such an improvement, regardless of the timeline.

lmineiro:
Filter registry (at least builtin) happens [here](https://github.com/zalando/skipper/blob/master/skipper.go#L482). Seems like a reasonable place to create a filter specific Registry.

We could use a [PrefixedRegistry](https://github.com/rcrowley/go-metrics/blob/master/registry.go#L149) and hand it over to the Filter instance and, somehow appropriate, to the "/metrics" handler. We'd have to blacklist any possible prefixes already use by skipper itself.

Another alternative would be to use the normal [StandardRegistry](https://github.com/rcrowley/go-metrics/blob/master/registry.go#L58) for each filter and the metrics handler would then handle "/metrics/filterName/foo.bar" as the way to get filter specific metrics where "filterName" would be the unique filter name and "foo.bar" its own metric. The namespacing would be part of the path instead of polluting the key names.

Would be interesting to think about the implication of metrics collectors of both approaches.

PR #507, opened 2017-12-04
Prometheus metrics

### LINKED ISSUE #504: Prometheus metrics
Hi there!

At this moment Skipper has a bunch of good metrics, but they are in a custom format. The problem is that this format is not portable to other kind of monitoring stacks.

Could be nice to have this same metrics in Prometheus format, knowing that Prometheus is one of the "de facto" standards out there nowadays.

Is this feature something you see valuable to have it in Skipper?

Thank you!

slok:
This PR adds Prometheus metrics as a new metrics format and closes #504.

* Refactor Metrics package to be extensible by different backend formats.
* Adds Prometheus format metrics backend.
* Refactor metrics handler to be generic for metrics backends.
* Move default metrics format as CodaHale backend
* Add new flag on cmd toenable Prometheus metrics `enable-prometheus-metrics`.
* Add Prometheus go client to glide dependencies (it updates other dependencies).
* By default use CodaHale metrics backend.

As a final note I would like to add that, now that other metrics format will be supported some metrics are ver "format" oriented and in a future could be rethink in a more generic way :)


metrics/metrics.go
```diff
@@ -84,327 +140,56 @@ type Options struct {
...
-		t.Update(d)
 	}
-}
 
-func (m *Metrics) MeasureSince(key string, start time.Time) {
-	m.measureSince(key, start)
-}
+	// Root path should return 404.
+	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusBadRequest)
```
aryszka: i'd actually agree with the comment above, it's more like a 404 (http.StatusNotFound) than a 400

metrics/codahale.go
```diff
@@ -0,0 +1,319 @@
...
+		json.NewEncoder(w).Encode(metrics)
+	} else {
+		http.NotFound(w, nil)
+	}
+}
+
+// This listener is only used to expose the metrics
+func (c *codaHaleMetricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method == "POST" {
+		w.WriteHeader(http.StatusBadRequest)
```
aryszka: 405 (http.StatusMethodNotAllowed) could be a more precise response here

metrics/doc.go
```diff
@@ -1,11 +1,20 @@
...
+
+* Coda Hale format.
+* Prometheus format.
+
+For CodaHale format it uses the Go implementation of the Go Coda Hale metrics library:
 
 https://github.com/dropwizard/metrics
 https://github.com/rcrowley/go-metrics
 
+For Prometheus format is will use Prometheus Go  official client library:
```
aryszka: "...is will use..." -> "...uses..."

metrics/doc.go
```diff
@@ -24,15 +33,15 @@ respectively.
 REST API
 
 This listener accepts GET requests on the /metrics endpoint like any other REST api. A request to "/metrics" should
-return a JSON response including all the collected metrics. Please note that a lot of metrics are created lazily
-whenever a request triggers them. This means that the API response will depend on the current routes and
-the filters used. In the case there are no metrics due to inactivity, the API will return 404.
+return a JSON response including all the collected metrics if CodaHale format is used, or in Plain text if Prometheus
+format is use . Please note that a lot of metrics are created lazily whenever a request triggers them. This means that
```
aryszka: "...format is use ." -> "...format is used."

aryszka:
👍

ruiaraujo:
:+1:

slok:
Thank you! I'll make a new PR with the minor fixes that you point out :)

PR #521, opened 2017-12-23
proxy: simplify satisfying an interface

ferhatelmas:
- drop unnecessary returns
 - set arguments and receivers to `_` to make clear, they aren't used

This is more like personal/cosmetic changes. If you don't want to take it, let me know.

### DIFF
--- a/proxy/tracing_test.go
+++ b/proxy/tracing_test.go
@@ -167,9 +167,7 @@ func (t *tracer) Extract(format interface{}, carrier interface{}) (ot.SpanContex
 }
 
 // SpanContext interface
-func (s *span) ForeachBaggageItem(handler func(k, v string) bool) {
-	return
-}
+func (s *span) ForeachBaggageItem(_ func(k, v string) bool) {}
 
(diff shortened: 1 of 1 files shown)


proxy/tracing_test.go
```diff
@@ -167,9 +167,7 @@ func (t *tracer) Extract(format interface{}, carrier interface{}) (ot.SpanContex
 }
 
 // SpanContext interface
-func (s *span) ForeachBaggageItem(handler func(k, v string) bool) {
-	return
-}
+func (s *span) ForeachBaggageItem(_ func(k, v string) bool) {}
```
lmineiro: Just because you're at it, we could also _ the receiver `s`, like:

```Go
func (_ *span)  ForeachBaggageItem(_ func(k, v string) bool) {}
```

ferhatelmas: Yes, thanks, done so.

szuecs:
:+1:

aryszka:
:+1:

PR #522, opened 2017-12-24
LUA Scripting

vetinari:
These commits add lua scripting to skipper

It introduces a new filter "luaScript" which should be used like
`any: * -> luaScript("function request(ctx, param); print(ctx.request.url); end") -> <shunt>`.

See script/README.md what is currently working.


script/README.md
```diff
@@ -0,0 +1,148 @@
...
+**NOTE**: Any parameter starting with "lua-" should not be used to pass
+values for the script - those will be used for configuring the filter.
+
+A filter script needs at least one global function: `request` or `response`.
+If present, they are called with a skipper filter context and the params passed
+in the route as table like
+```lua
+-- route looks like
+--
+-- any: * -> lua("./test.lua", "myparam=foo", "other=bar") -> <shunt>
```
szuecs: What is the CWD in this case, used for ./test.lua ?
Can you add this to the docs please?

script/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+Request headers can be accessed by accessing the `ctx.request.header` map like
+```lua
+ua = ctx.request.header["user-agent"]
+```
+Header names are normalized by the `net/http` go module like usual. Setting
+header is done by assigning to the headers map. Setting a header to `nil` or
+an empty string deletes the header - setting to `nil` is preferred.
+```lua
+ctx.request.header["user-agent"] = "skipper.lua/0.1"
```
szuecs: better use semantic versioning: 0.1 -> v0.0.1

szuecs:
@vetinari Thanks for the PR. Seems to be a nice feature!
I added some documentation comments and have one additional request here.

Can you add some benchmarks for lua based redirect and base64 encode/decode, to show our users numbers of the dispatch to lua and calculation?
It would be really great to see these.

szuecs:
👍

szuecs:
I guess it's already ok to merge @vetinari or do you need something to add?

aryszka:
nice!

aryszka:
👍

vetinari:
@szuecs  I think it can be merged now. We can add some real live (example) scripts later to https://github.com/skipper-plugins/. The current state should be enough to e.g. re-implement skoap as lua script  - except for the logging part ;-)

szuecs:
@vetinari better document in ./docs/ in branch  gh-pages to show it somewhere at https://zalando.github.io/skipper/

PR #524, opened 2017-12-29
all: simplifications

ferhatelmas:
* `time.Since` instead of `time.Now().Sub`
 * `strings.Contains` instead of `strings.Index`
 * `bytes.Buffer.String` instead of `string(bytes.Buffer.Bytes())`
 * `copy` instead of for loop


szuecs:
Thanks for the PR and your reviews / refactoring efforts!

szuecs:
:+1:

lmineiro:
👍

PR #526, opened 2018-01-02
Improve error logs for invalid routes

mo-gr:
When processing of routes fail, we currently only output the error, but nothing to identify the broken route. This can be a problem when there are many thousands of routes.

This change improves the logging in case of broken routes by also outputting the id of the route that failed.


mo-gr:
:+1:

szuecs:
:+1:

PR #527, opened 2018-01-03
[lua] add body="string" to ctx.serve()

vetinari:
with this we can have a body with an error message like in
```lua
function request(ctx, args)
    local ok, token, err = pcall(extract_token, ctx)
    if not ok then
        -- in case pcall fails, the error is in the token var
        print("ERROR: " .. token)
        ctx.serve({status_code=500, body="Internal Server Error\n"})
        return
    end

    if err ~= nil then
        print("ERROR: " .. err)
        ctx.serve({status_code=401, body=tostring(err)})
        return
    end
    -- do more here
end
```

### DIFF
--- a/script/script.go
+++ b/script/script.go
@@ -2,8 +2,10 @@
 package script
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"net/http"
 	"net/url"
@@ -256,6 +258,7 @@ func serveTableWalk(res *http.Response) func(lua.LValue, lua.LValue) {
 				return
 			}
 			res.StatusCode = int(n)
+
 		case "header":
 			t, ok := v.(*lua.LTable)
 			if !ok {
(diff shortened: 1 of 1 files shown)


script/script.go
```diff
@@ -264,6 +267,16 @@ func serveTableWalk(res *http.Response) func(lua.LValue, lua.LValue) {
 			h := make(http.Header)
 			t.ForEach(serveHeaderWalk(h))
 			res.Header = h
+
+		case "body":
+			// would love to be able to pass a table to be then rendered as
+			// json body... but layeh.com/gopher-json does not export toJSON
```
szuecs: I hope you created an issue for that upstream :)

vetinari: not yet... will do a PR later

vetinari:
fixes #528

script/script.go
vetinari: upstream PR https://github.com/layeh/gopher-json/pull/2

vetinari:
:+1:

szuecs:
:+1:

Issue #531, opened 2018-01-04
hardcoded healthz check Source predicate


szuecs:
The hardcoded healthz check Source predicate is not possible to be overwritten by the user.
Possible workaround for users, use an inline route parameter to proxy the health check call:

    -inline-routes 'Path("/health") -> setPath("/kube-system/healthz") -> <loopback>'

lmineiro:
The hardcoded values include ~all private networks.
I'd be curious to know why the health check endpoint would support public networks. That being said, I agree that it could be user-defined since the Source predicate already supports it anyway...

mikkeloscar:
I'm wondering if we really need to define any source networks in skipper. It seems out of scope to define it in the application rather than restrict this on the security group/firewall level?

szuecs:
@lmineiro I think we should just drop the Source predicate, because you can easily fake it

    % curl -kv -H"X-Forwarded-For: 192.168.0.4" -H"Host: localhost" https://some-ingress-endpoint/kube-system/healthz
    < HTTP/1.1 200 OK
    < Date: Thu, 04 Jan 2018 14:44:12 GMT
    < Content-Type: text/plain; charset=utf-8
    < Content-Length: 0
    < Connection: keep-alive
    < Server: Skipper

lmineiro:
Can you try that behind the ELB/ALB?

szuecs:
@lmineiro It will work, because we use X-Forwarded-For header as remote ip if it's in the header and the host header is chosen by the user, so no difference anyways

lmineiro:
I'd still like to see how the ELB/ALB handles an existing `X-Forwarded-For` header. If it just appends itself the multivalue header or as the last comma-separated list of proxies.

The [source predicate takes the left most host as the client IP](https://github.com/zalando/skipper/blob/master/net/net.go#L29).

szuecs:
Amazon documentation is undefined for ALB and is different from ELB:

    curl -kv -H"X-Forwarded-For: 192.168.0.4" -H"Host: localhost" https://some-endpoint/kube-system/healthz

15:58:31.794845 IP 172.31.2.194.3768 > 172.31.20.165.9999: Flags [P.], seq 1502:1748, ack 159, win 10209, options [nop,nop,TS val 2407620323 ecr 572090078], length 246
E..*-L@...............'..F.     w6.>..'........
..^.".f.GET /kube-system/healthz HTTP/1.1
[...]

Issue #250, opened 2017-01-18
consider trimming the hop-by-hop http headers


aryszka:
See: https://tools.ietf.org/html/rfc2616#section-13.5.1

lmineiro:
This is mentioned and I'd expect it to be covered by #61 
We can probably detail the original ticket title, at a minimum

aryszka:
thx

closing because a dupe of #61

Issue #61, opened 2015-11-20
provide standard proxy headers


aryszka:

pc-alves:
Let's go into more details here. Is the idea to add this as a filter, predicate, all of the above? Is it meant to be a part of the built-in predicates and filters, or should it exist as separate methods?

lmineiro:
I think this is neither predicates or filters. If memory doesn't fail, this would be about much of the stuff found at https://golang.org/src/net/http/httputil/reverseproxy.go
For example, some headers shouldn't be forwarded (See `hopHeaders` around line 111) while others should be added like the `X-Forwarded-For` around line 211

pc-alves:
Ok, so far here is what I identified on this topic:

- Support for X-Forwarded-For header (https://en.wikipedia.org/wiki/X-Forwarded-For)
- Support for Forwarded header (https://tools.ietf.org/html/rfc7239)
- Prevent Hop-By-Hop headers from being forwarded (https://tools.ietf.org/html/rfc2616#section-13.5.1)

krism74:
Greetings ,

I actually wanted to create a separate thread but since this is a related topic thought I might as well ask here.

Is there a way to pass headers to a service GET/POST request for example : REMOTE-USER or X-FORWARDED-USER something similar to 
request.Header.Add("REMOTE-USER","user") or request.Header.Add("X-FORWARDED-USER","user")

thanks in advance
kris

lmineiro:
I believe those will be forwarded if they're already present.
Are you asking if skipper should be adding those headers?

aryszka:
@krism74 if it is about the incoming headers, then as @lmineiro said, they are passed through. If it is about adding additional headers, you can setRequestHeader or appendRequestHeader filters. See https://godoc.org/github.com/zalando/skipper/filters/builtin#NewSetRequestHeader

krism74:
Greetings
Here is my scenario, I am trying to use skipper as a reverse proxy(for services) and need to set header attributes on the request header before passing them off to backend services.

So I have the following 
`serviceA: PathSubtree("/serviceA")  -> requestHeader("X-Forwarded-User", "xxx.yyy") -> setRequestHeader("X-Forwarded-Groups", "Services") -> "http://xyz.abc.com:9000";`

I tried the following as well
[...]

PR #533, opened 2018-01-04
Source predicate can be by passed in AWS

szuecs:
elb and alb and probably will append all X-Forwarded-For header items they get in the request header to their own in the request they do to the backend. As result the right most is the real client IP.

TODO:
- [x] add an option to select left or right most matching in Source() predicate or skipper cli
- [x] do not change the default to not break implementations based on this

see https://github.com/zalando/skipper/issues/531#issuecomment-355320512


lmineiro:
Is this a breaking change?

szuecs:
I changed the description of the PR to make sure it is understood what is open todo to fix the issue, before merge

lmineiro:
For reference, since there were discussions around Skipper's proxy behavior in the past:
* https://github.com/zalando/skipper/issues/61
* https://github.com/zalando/skipper/issues/250

aryszka:
how about using a new predicate name for this behavior: e.g. LastSource() or s.g. like that?

lmineiro:
Or `CloserSource`?

szuecs:
@aryszka I am fine with that, but we have to pass an option for running in AWS anyways, because otherwise we hardcode the AWS way for us in the dataclient which does not makes sense.

@lmineiro CloserSource ? What? Why?

szuecs:
:+1:

szuecs:
@aryszka I created a second predicate with tests and an option to skipper that can be used to change the automatic healthz endpoint to use the not common AWS way of X-Forwarded-For header to find the client ip.

net/net.go
```diff
@@ -17,13 +17,20 @@ func stripPort(address string) string {
...
+		return res
 	}
 	return nil
 }
 
-// The remote address of the client. When the 'X-Forwarded-For'
-// header is set, then it is used instead.
+// RemoteHost returns the remote address of the client. When the
+// 'X-Forwarded-For' header is set, then it is used instead. This is
+// the most often behaviour for proxies. Wikipedia shows the format
```
aryszka: "This is how most often proxies behave."

predicates/source/source.go
```diff
@@ -86,11 +99,51 @@ func (s *spec) Create(args []interface{}) (routing.Predicate, error) {
...
 }
+
+func NewFromLast() routing.PredicateSpec { return &specLast{} }
+
+func (s *specLast) Name() string {
+	return Name
+}
+
+func (s *specLast) Create(args []interface{}) (routing.Predicate, error) {
+	if len(args) == 0 {
```
aryszka: this duplicate code could be avoided, if the spec would be the same for the two flavors with the bool field passed to the predicate instance. The separate New() and NewFromLast() functions should stay but return the same spec with this bool field difference.

aryszka:
@szuecs looks good to me. I only found that the two flavors of the predicate could share the Create() method if the spec had a bool field like fromLast

szuecs:
:+1:

szuecs:
:+1:

aryszka:
:+1:

PR #536, opened 2018-01-05
LoadBalancer feature

### LINKED ISSUE #499: Loadbalancer support
Why:

- This would enables `ingress -> endpoints` circumventing service (`ingress -> svc -> endpoints`). 
- This is also required to enable service mesh
- optimize backend dispatch
- find faster miss behaving backends

szuecs:
This will fix #499.

Description: This is not doing active health checks for all backends in skipper. The loadbalancer gets input to do healthchecking from the proxy and makes it possible to filter by healthiness from dataclients, that can delete routes if these are unhealthy. 

There are some major changes forming the loadbalancer feature:
- loadbalancer() predicate, which is doing a roundrobin loadbalancing implemented via route Matching
- loadbalancer package which has a healthchecking and filtering facility keeping track of all unhealthy or dead backends
- proxy package adds all backends with roundtrips that are returning a net.Error to the loadbalancer health checks lb.AddHealthcheck via a buffered channel that should not block (the hardcoded 100 items are hopefully enough)
- added flag to skipper and pass the ptr to loadbalancer.LB to dataclient kubernetes and proxy
- <s>changed http handling on both sides to have client and server timeouts configured, for now they are hardcoded and we have to set some defaults and create a lot of options somehow. TBD how this should be configurable and what are the best defaults.</s> this was done in a separate PR

open tasks:
- [x] cleanup logging
- [x] cleanup routeState map in loadbalancer package if the route for this backend does not exist anymore (should be done by the dataclient)
- [x] check if I have some TODO(sszuecs) left over
- [x] fix test cases
- [x] add unit tests 
- [x] manually test in kubernetes dev cluster scenarios (good case, terminate node, iptables drop comunication to a backend to show net.Error.Temporary()==False case) 
- [x] could be done later: add retry some net.Error with ctx.Route.Next and ctx.Route.Head linked list
- [x] make client and server timeouts configurable and set and agree on defaults https://github.com/zalando/skipper/pull/538
- [x] close channels on sigterm/shutdown
- [x] proxy test to reproduce a bug in the backend connection refused (sigkilled) condition


szuecs:
I tested with kill -9 backend to test the retry and loadbalancer behaviour  and it improved a lot.

proxy/proxy.go
```diff
@@ -703,14 +729,39 @@ func (p *Proxy) do(ctx *context) error {
...
 			if done != nil {
 				done(false)
 			}
 
 			p.metrics.IncErrorsBackend(ctx.route.Id)
-			return err
+
+			neterr := perr.NetError()
+			if neterr != nil && !neterr.Temporary() && ctx.route.IsLoadBalanced {
+				// here we do a transparent retry, because we know it's safe to do
```
aermakov-zalando: I would rather be safe than sorry and explicitly list the conditions where we can safely retry. Going to `syscall_unix.go` just to find out which errors are considered temporary isn't really maintainable in my opinion.

Additionally, I'm not sure how is the connection timeout case covered by this. We *should* retry on connection timeouts, since this is what was happening in the original issue, but we *mustn't* retry on read/write timeouts after the connection was established, since it might be dangerous if we retry a non-idempotent request. However, if I'm reading this correctly, both of these would be covered by `neterr.Timeout()` which marks the error as `Temporary()`.

szuecs: No the problem was "connection refused", which is not a neterr.Timeout() but a "not neterr.Temporary()"

aermakov-zalando: While an instance is rebooting, you wouldn't get connection resets, since there's nobody there to send a packet back, so you'll get connection timeout. You should be able to test this by blocking incoming packets with `-j DROP`, I think.

My actual concern is that since we retry all requests here, even `POST`, `PUT` and `DELETE` ones, we have to be very careful as to when we retry. This code makes the retry logic completely unclear and dependent on stuff that isn't really documented unless you go into Go source code, without any guarantees on whether it'll change or not. I think that the retry criteria should be made much more clear and documented.

szuecs: I share your concern, but the retry is only happening on permanent network errors, so this should be safe in any case. A retry strategy which is more fine grained have to be per route, because only the user knows if HTTP GET is safe to reapply for temporary errors or timeouts.
I agree that iptables -j DROP is also a valuable case, but in the VM based environment a EC2 instance can die and the hyp can return the connection refused. In any case I want to start with connection refused and later think about other cases we might want to retry. Feel free to write an issue in this repository to specify which else condition(s) you think we should retry.
Thanks for your feedback.

aermakov-zalando: I didn't want to suggest making the retry strategy fine grained, sorry if I wasn't clear enough. What I was trying to say is that since the retry strategy applies to all requests, not even those that should be safe enough to retry (and people also do crazy things with GET!), we should be as explicit as possible in the logic to avoid accidentally retrying something. The retry logic should be clear and understandable from the first glance, and shouldn't change whenever the Golang stdlib changes something. The current code which implicitly trusts the `NetError` implementation to provide the correct information isn't really suitable, especially considering that the error might be raised long after the connection was successful and we've sent the request.

aermakov-zalando: Also, the hypervisor won't return connection refused if the VM is down — this is only something the VM kernel can do. You can verify it yourself by rebooting or terminating an instance and trying to connect.

szuecs: I honestly think that something you have to trust and what would you suggest to fix it without a trust?

I see these possibilities, maybe you have opinions on these or have to add another one:
1. check error messages to filter for tcp handshake errors (for example "connection refused" or some connection timeout) or tls handshake errors and 
1. have a low timeout for tcp/tls connections, have high timeouts for all others, measure time and hope that if we have passed the tcp/tls timeouts it is safe to retry
1. create the connections pools ourselves
1. do not retry at all

The "check error messages" idea could be dropped as soon Go would have exported error types that help to clearly differentiate, but right now it would required to check strings in errors, which I consider a bit ugly but I would do it if you think it makes more sense.

szuecs:
@aermakov-zalando as also offline showed and tested we do not follow symlinks, if we use RoundTripper as we do. 
I wrapped the net.Dialer as you suggested and return a proxyError and check if this was caused while dialing to check if we can retry the call.

Thanks for your comments.

szuecs:
:+1:

szuecs:
I cleaned up code and test.
I did a last test with jmeter in one cluster with a failing pod to see if we are stable and it works well.
Follow up issues: 
- #556 small/medium refactoring of the lb lookup to get rid of lbmember routes and the double route lookup 
- #557 implement lb algorithms which should be straight forward after #556, dependent on the algorithm.

szuecs:
:+1:

PR #553, opened 2018-02-13
Fixed eskip examples to use camelCase

voidance:
Current examples on eskip file format are written using snake case but that doesn't work. Changed to use camelCase.


szuecs:
👍

voidance:
👍

PR #555, opened 2018-02-22
computeBackendWeights: avoid rounding issues

aermakov-zalando:
The last active backend (with a non-zero traffic weight) can get a weight of less than 1.0 due to floating point precision issues. Fix by forcibly setting a weight of 1.0.

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -716,22 +716,26 @@ func catchAllRoutes(routes []*eskip.Route) bool {
 //
 // where for a weight of 1.0 no Traffic predicate will be generated.
 func computeBackendWeights(backendWeights map[string]float64, rule *rule) {
-	type sumCount struct {
-		sum   float64
-		count int
+	type pathInfo struct {
+		sum        float64
+		lastActive *backend
+		count      int
 	}
 
 	// get backend weight sum and count of backends for all paths
-	pathSumCount := make(map[string]*sumCount)
+	pathInfos := make(map[string]*pathInfo)
 	for _, path := range rule.Http.Paths {
-		sc, ok := pathSumCount[path.Path]
+		sc, ok := pathInfos[path.Path]
 		if !ok {
-			sc = &sumCount{}
-			pathSumCount[path.Path] = sc
+			sc = &pathInfo{}
+			pathInfos[path.Path] = sc
 		}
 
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

szuecs:
:-1:

dataclients/kubernetes/kube.go
```diff
@@ -733,31 +733,41 @@ func catchAllRoutes(routes []*eskip.Route) bool {
...
 		if !ok {
-			sc = &sumCount{}
-			pathSumCount[path.Path] = sc
+			sc = &pathInfo{}
+			pathInfos[path.Path] = sc
 		}
 
 		if weight, ok := backendWeights[path.Backend.ServiceName]; ok {
 			sc.sum += weight
+			if weight > 0 {
```
szuecs: why do you need this check?

dataclients/kubernetes/kube.go
```diff
@@ -733,31 +733,41 @@ func catchAllRoutes(routes []*eskip.Route) bool {
 //
 // where for a weight of 1.0 no Traffic predicate will be generated.
 func computeBackendWeights(backendWeights map[string]float64, rule *rule) {
-	type sumCount struct {
-		sum   float64
-		count int
+	type pathInfo struct {
+		sum        float64
+		lastActive *backend
```
szuecs: I think you don't need this lastActive to achieve the goal of having the last set to 1.0

dataclients/kubernetes/kube.go
aermakov-zalando: To skip backends with weights explicitly set to 0.0.

dataclients/kubernetes/kube.go
aermakov-zalando: Any suggestions, or at least a comment on why you don't like this approach?

szuecs:
I am not convinced that this can happen right now and I think this could break other patterns.
Please provide examples or best test cases, which will show the problem that you found.

aermakov-zalando:
I've added a test that was failing before the fix.

szuecs:
Offline discussed with @aermakov-zalando, this will not break any ingress with predicate traffic and fixes a current float rounding issue.

szuecs:
:+1:

szuecs:
:+1:

mikkeloscar:
:+1:

PR #558, opened 2018-02-25
Remove hop headers

pc-alves:
Contributes to https://github.com/zalando/skipper/issues/61 .

This PR includes only the feature to remove hop headers as specified in [RFC 2616](https://tools.ietf.org/html/rfc2616#section-13.5.1).

This feature is disabled by default (to maintain current functionality) and enabled by using a flag.
That flag is extended to the command line client:

```
-remove-hop-headers
    	enables removal of Hop-Headers according to RFC-2616
```


proxy/proxy.go
```diff
@@ -230,12 +249,26 @@ func copyHeader(to, from http.Header) {
 	}
 }
 
+func copyHeaderExcluding(to, from http.Header, excludeHeaders map[string]bool) {
+	for k, v := range from {
+		if len(excludeHeaders) == 0 || !excludeHeaders[k] {
```
szuecs: Why 2 checks here?
This should be enough, not?:

    !excludeHeaders[k] {

The more Goish way would be:

    if _, ok := excludeHeaders[k]; !ok {

This would check only that there is a key within the map.

szuecs:
@pc-alves I don't really understand why this contributes to #61 .
As far as I understand the issue #61 should add and modify headers of the X-Forwarded-* family.
You added an option to drop these headers, which I am fine with according to rfc2616 as you showed.

The code looks in general fine to me, only a small nit. Thanks for the PR!

lmineiro:
@szuecs Part of #61 is to actually remove the hop-by-hop headers, as mentioned in the discussion on that issue. Adding the X-Forwarded-* headers is another requirement.

szuecs:
@lmineiro I see in the 4th comment it's written. Maybe it makes sense to rewrite the issue description to show what should be done.
Thanks for pointing it out.

Only the nit fixes should be done to be more readable.

szuecs:
please rebase and resolve the conflicts in  cmd/skipper/main.go

proxy/proxy.go
pc-alves: Thanks for the tip on the Go(ish) way of making that if-statement. I've change the code to reflect that. I will test the request with `ConNection` and get back to you on that.

pc-alves:
@szuecs My second comment (5th overall) provides a summary of the requirements for #61. Alas, I cannot change the description of the issue. @aryszka would you be so kind as to make that edit?

pc-alves:
@szuecs Regarding your concern about case sensitive comparison of header names, you cannot bypass it with `ConNection`, for example. I tested it and could not bypass. I then took a look why and the reason is because http package converts the header names to the canonical version. See for example https://golang.org/src/net/http/header.go?s=487:525#L13 and https://golang.org/src/net/textproto/header.go?s=403:445#L3

szuecs:
@pc-alves cool thanks for checking!

pc-alves:
@szuecs @aryszka I was wondering if you could give me a hint on what seems to be the issue with the failing Travis CI build. I tried the same make target (`make ci-trigger`) and everything worked locally. But that target has different executions, so it's possible I'm not running the correct one.
Any tips would be most welcome. Thank you.

szuecs:
@pc-alves just fetch current master and merge master into your branch, this was already fixed

pc-alves:
Conflicts fixed.
@ruiaraujo added the comment you suggested.

ruiaraujo:
👍

szuecs:
:+1:

PR #564, opened 2018-03-05
fix endpoint.Targets() to filter selected targets also by ports

### LINKED ISSUE #563: loadbalancer does not filter ports in endpoints
For example this service creates for one replica 2 endpoints and we would sent to both endpoints traffic, but maybe in the ingress we selected only service with port name "http" :

    spec:
      clusterIP: 10.3.0.69
      ports:
      - name: http
        port: 80
        protocol: TCP
        targetPort: 9090
      - name: metrics
        port: 9911
        protocol: TCP
        targetPort: 9911

szuecs:
This will fix  #563

Given an ingress spec:

    spec:
      rules:
      - host: foo.example.org
        http:
          paths:
          - backend:
              serviceName: foo
              servicePort: http
            path: /

And a service definition:

    spec:
      ports:
      - name: http
        port: 80
        protocol: TCP
        targetPort: 9090
      - name: metrics
         port: 9911
        protocol: TCP
        targetPort: 9911
      selector:
        application: foo
      type: ClusterIP

would result in 2x the replica count one for each pod :80 and :9911. 

This PR will fix it and make sure only selected endpoints will be used (with port :80).

### DIFF
--- a/dataclients/kubernetes/definitions.go
+++ b/dataclients/kubernetes/definitions.go
@@ -122,12 +122,15 @@ type endpoint struct {
 	Subsets []*subset `json:"subsets"`
 }
 
-func (ep endpoint) Targets() []string {
+func (ep endpoint) Targets(ingSvcPort string) []string {
+	epPort, _ := strconv.Atoi(ingSvcPort)
 	result := make([]string, 0)
 	for _, s := range ep.Subsets {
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

linki:
👍

Issue #566, opened 2018-03-06
500s because of division by 0


linki:
Using skipper v0.9.165 we get a division by zero which leads to 500s for all affected ingresses.

I see this routes in my skipper:

```console
[APP]time="2018-03-06T12:14:11Z" level=error msg="error while processing filter during request: lbDecide: runtime error: integer divide by zero"
[APP]time="2018-03-06T12:14:11Z" level=error msg="error while proxying, route kube_default__blabla__lb_group, status code 500: max loopbacks reached"
```

```console
$ curl localhost:9911/routes
kube_default__foobar_service__foobar_service_mycluster_test_my_domain____foobar_service__lb_group: Host(/^zmon-data-service[.]stups-test[.]zalan[.]do$/) && LBGroup("kube_default__foobar_service__foobar_service_mycluster_test_my_domain____foobar_service")
  -> lbDecide("kube_default__foobar_service__foobar_service_mycluster_test_my_domain____foobar_service", 0)
```
This leads to a division by zero here: 
* https://github.com/zalando/skipper/blob/v0.9.165/loadbalancer/filter.go#L77
* https://github.com/zalando/skipper/blob/v0.9.165/loadbalancer/filter.go#L33

Looks like the new `Targets` with port filter return an empty list: https://github.com/zalando/skipper/blob/v0.9.165/dataclients/kubernetes/kube.go#L441

The Targets function assumes the ports defined on the service match the ones defined on the Endpoints https://github.com/zalando/skipper/blob/v0.9.165/dataclients/kubernetes/definitions.go#L125-L138 but they don't:

```console
$ kubectl get ing myapp -o yaml
apiVersion: extensions/v1beta1
kind: Ingress
spec:
  rules:
  - host: foo.host
    http:
      paths:
      - backend:
          serviceName: foosvc
          servicePort: 80
```

```console
$ kubectl get endpoints myapp
NAME              ENDPOINTS                                                      AGE
myapp   10.2.42.42:8080,10.2.43.25:8080,10.2.45.245:8080 + 3 more...   279d
```

/cc @szuecs

linki:
We also have to take into account the mapping from `port` to `targetPort` in the `Service`.

```console
$ kubectl get svc myapp -o yaml
apiVersion: v1
kind: Service
spec:
  clusterIP: 10.3.0.119
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 8080
```

aryszka:
[...]

PR #567, opened 2018-03-06
fix dataclient part of the "no endpoints" problem

szuecs:
#566 dataclient part
Make sure the port filter of getEndpoints does not result in no endpoints being used.
In case of no endpoints it will fallback to the Kubernetes service as before.


aryszka:
:+1:

szuecs:
:-1:

szuecs:
It's more complicated than thought first. There are named ports and unnamed ports and we have to lookup services before endpoints to match the targetport from service to the endpoint port.

szuecs:
:+1:

aryszka:
:+1:

PR #569, opened 2018-03-06
Feature/add migration path for metrics

szuecs:
This change will make it possible to serve prometheus and codahale metrics.
To get codahale metrics you have to send an "Accept" header "application/json", else skipper will serve prometheus metrics.

### DIFF
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -51,66 +51,66 @@ const (
 	defaultMaxIdleConnsBackend             = 0
 	defaultLoadBalancerHealthCheckInterval = 0 // disabled
 
-	addressUsage                         = "network address that skipper should listen on"
-	etcdUrlsUsage                        = "urls of nodes in an etcd cluster, storing route definitions"
-	etcdPrefixUsage                      = "path prefix for skipper related data in etcd"
-	kubernetesUsage                      = "enables skipper to generate routes for ingress resources in kubernetes cluster"
-	kubernetesInClusterUsage             = "specify if skipper is running inside kubernetes cluster"
-	kubernetesURLUsage                   = "kubernetes API base URL for the ingress data client; requires kubectl proxy running; omit if kubernetes-in-cluster is set to true"
-	kubernetesHealthcheckUsage           = "automatic healthcheck route for internal IPs with path /kube-system/healthz; valid only with kubernetes"
-	kubernetesHTTPSRedirectUsage         = "automatic HTTP->HTTPS redirect route; valid only with kubernetes"
-	kubernetesIngressClassUsage          = "ingress class regular expression used to filter ingress resources for kubernetes"
-	innkeeperURLUsage                    = "API endpoint of the Innkeeper service, storing route definitions"
-	innkeeperAuthTokenUsage              = "fixed token for innkeeper authentication"
-	innkeeperPreRouteFiltersUsage        = "filters to be prepended to each route loaded from Innkeeper"
-	innkeeperPostRouteFiltersUsage       = "filters to be appended to each route loaded from Innkeeper"
-	oauthURLUsage                        = "OAuth2 URL for Innkeeper authentication"
-	oauthCredentialsDirUsage             = "directory where oauth credentials are stored: client.json and user.json"
-	oauthScopeUsage                      = "the whitespace separated list of oauth scopes"
-	routesFileUsage                      = "file containing static route definitions"
-	inlineRoutesUsage                    = "inline routes in eskip format"
-	sourcePollTimeoutUsage               = "polling timeout of the routing data sources, in milliseconds"
-	insecureUsage                        = "flag indicating to ignore the verification of the TLS certificates of the backend services"
-	proxyPreserveHostUsage               = "flag indicating to preserve the incoming request 'Host' header in the outgoing requests"
-	idleConnsPerHostUsage                = "maximum idle connections per backend host"
-	closeIdleConnsPeriodUsage            = "period of closing all idle connections in seconds or as a duration string. Not closing when less than 0"
-	devModeUsage                         = "enables developer time behavior, like ubuffered routing updates"
-	supportListenerUsage                 = "network address used for exposing the /metrics endpoint. An empty value disables support endpoint."
-	metricsListenerUsage                 = "network address used for exposing the /metrics endpoint. An empty value disables metrics iff support listener is also empty."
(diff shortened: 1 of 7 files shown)


szuecs:
:+1:

cmd/skipper/main.go
```diff
@@ -273,6 +275,7 @@ func init() {
 	flag.IntVar(&defaultHTTPStatus, "default-http-status", http.StatusNotFound, defaultHTTPStatusUsage)
 	flag.BoolVar(&suppressRouteUpdateLogs, "suppress-route-update-logs", false, suppressRouteUpdateLogsUsage)
 	flag.BoolVar(&enablePrometheusMetrics, "enable-prometheus-metrics", false, enablePrometheusMetricsUsage)
+	flag.BoolVar(&enableCodahaleMetrics, "enable-codahale-metrics", false, enableCodahaleMetricsUsage)
```
ruiaraujo: Isn't this a breaking change?

szuecs: No it isn't, because the flag is miss leading. ;)
This flag is only required for enabling codahale metrics, if prometheus metrics are enabled.

metrics/all_kind.go
```diff
@@ -0,0 +1,99 @@
...
+}
+func (a *All) RegisterHandler(path string, handler *http.ServeMux) {
+	a.prommHandler = a.prometheus.CreateHandler()
+	a.codaHaleHandler = a.codaHale.CreateHandler(path)
+	handler.Handle(path, a.newHandler())
+}
+
+func (a *All) newHandler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		if req.Header.Get("Accept") == "application/json" {
```
Raffo: This seems like a hidden feature to me that can easily get lost. Any specific reason why it is done like that?

metrics/all_kind.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"net/http"
+	"time"
+)
+
+type All struct {
+	prometheus      *Prometheus
+	codaHale        *CodaHale
+	prommHandler    http.Handler
```
mikkeloscar: nit: maybe we should call it `prometheusHandler` `promm` with two m's looks wrong :)

metrics/all_kind.go
szuecs: Do you have another suggestion how to do this?
For me it was the native http way of doing it.

metrics/all_kind.go
szuecs: done

cmd/skipper/main.go
szuecs: The truth is that if we enable this by default, then if we enable enablePrometheusMetrics, how to decide if enableCodahaleMetrics should be enabled, too?

Maybe it's better to have --metrics-flavour=<codahale|prometheus|all> opinions?

metrics/all_kind.go
Raffo: It is and you are not totally wrong, but it seems a bit hidden to me... in the end `application/json` doesn't map necessarily to `codehale`, but just to the json format, you know what I mean?

szuecs: I understand, but what else do you recommend?

cmd/skipper/main.go
ruiaraujo: How about instead?
--metrics-flavour=codahale, prometheus

metrics/all_kind.go
Raffo: Thinking about it twice, it is not that bad and I think actually the problem is just docs. I would do the following: 

- document the way metrics are exposed in Skipper, by saying that we use two formats.
- have flags such that it is always clear which metrics are enabled by default and how to use them. 

Also: is it actually right that in case you have `application/json` you do `codahale` and in the else prometheus? If I write `application/foobar` I get prometheus text format, is that intended? If so, it kinda clashes with the idea that `codahale` is the default... when the flags are not set for example you always get `codahale`, when they are both set, you have to be "very specific" to get `codahale` otherwise you get `prometheus`.

cmd/skipper/main.go
Raffo: +1 on @ruiaraujo suggestion, maybe by validating that the strings for the metrics type are actually supported.

szuecs: I agree, I will change that

aryszka: (sry, meant the plus one :D )

metrics/all_kind.go
aryszka: i am thinking about using different paths for the different metrics. E.g:

/ -> the current default
/codahale -> codahale if enabled or 404
/prometheus -> prometheus if enabled or 404

not sure if the above is the best, but i agree with @Raffo 's comment that the mime type doesn't represent the types that we're choosing from.

metrics/metrics.go
```diff
@@ -19,6 +19,7 @@ const (
 	UnkownKind Kind = iota
 	CodaHaleKind
 	PrometheusKind
+	AllKind
```
aryszka: it's good like this, but possible cosmetics change:

```
UnknownKind Kind = 0
CodaHaleKind Kind = 1 << iota
PrometheusKind
AllKind = CodaHaleKind | PrometheusKind
```

this way it is easier to add future kinds.

metrics/all_kind.go
Raffo: I had the same idea as @aryszka but then I didn't write it, I felt like I didn't want to be that specific to even put it in the path.

aermakov-zalando: If you want to go the content-type route I'd suggest doing something explicit like `application/codahale+json`.

szuecs: @aryszka the path change will also break the current monitoring, but content-type already does the same so for me it's not a problem to do the path routing approach.
Anyone against the path routing?
Someone prefer the Content-Type based solution?

Raffo: +1 for @aermakov-zalando

aryszka: @szuecs i also like @aermakov-zalando 's suggestion best now.

szuecs: I also talked with @mikkeloscar and we agreed to use application/codahale+json Content-Type based solution

szuecs:
I think I covered all comments :+1:

szuecs:
:+1:

aryszka:
:+1:

szuecs:
:+1:

PR #570, opened 2018-03-07
Fix/groupsize check in lb filter

aryszka:
adds check for non-positive group size in the lbDecide() filter

partially fixes: #566 
related PR: #567


mikkeloscar:
:+1:

szuecs:
:+1:

Issue #134, opened 2016-04-05
Using regex delimiter fails in character list in eskip


grassator:
This should work

``` js
PathRegexp(/[/]/)
```

but it fails in eskip

LappleApple:
@mo-gr @grassator @aryszka Should I add a "Help Wanted" label to this, or were one of you guys going to do this at some point?

MALPI:
Can this one be closed?

szuecs:
yes

PR #571, opened 2018-03-07
Using regex delimiter fails in character list in eskip #134

snurmine:
Modified scanning of regexp as it need to take into account special characters #134
Question 
from 
`PathRegexp(/\/\w+Id$/)` 
it seems that / is meant to be escaped. But inside [] I think / can be go without escaping? 

And master impl seems to parse `/\\/` as Go regexp `\` , I mean one cannot escape \ ? I removed this behaviour, in those cases, but I didn't see any issues.

### DIFF
--- a/eskip/lexer.go
+++ b/eskip/lexer.go
@@ -151,8 +151,47 @@ func scanEscaped(delimiter byte, code string) ([]byte, string) {
 	return b, code
 }
 
+func scanRegexp(code string) ([]byte, string) {
+	var b []byte
+	escaped := false
+	var insideGroup = false
+	for len(code) > 0 {
+		c := code[0]
+		isDelimiter := c == '/'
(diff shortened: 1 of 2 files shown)


aryszka:
@snurmine 

i agree that the /exp/ notation for regexps is not consistent with how this notation is used in other langauges, and a slash inside a [ ] should be allowed.

the same is true for \\. If we take JS /exp/ notation for a basis, then \\ should mean \\ for Go regexps.

snurmine:
For example, if I have String `\` and I want match to that string, I need Go regexp '\\' . But lexer seems to parse two `\` as single `\` ? https://github.com/zalando/skipper/blob/master/eskip/lexer.go#L130

At least https://regex101.com/ seems to require that `\` is escaped by `\`.

PR #574, opened 2018-03-09
initial tests that are based on the filesystem

szuecs:
This implements basic test infrastructure to test kubernetes objects defined as yaml files, which are automatically converted to json and then served by ServeHTTP within kube2_test.go.

needed a rebase to fix/566-kube-dc-part, because of a dependency of the data structure change

What do you think?


aryszka:
@szuecs i think this is a cool initiative. We need to improve our test harnesses, and the kube client is one of those areas where testing is currently not a very nice experience.

szuecs:
I won't follow up on this

Issue #544, opened 2018-01-22
legacy style predicates from kubernetes ingress annotations don't work


aryszka:
The "legacy" style predicates are added to a route definition as fields on the eskip.Route type. The newer style of predicates are appended to the Predicates slice field. The predicates defined as Kuberentes ingress annotations are all appended to the Predicates slice field. When the routing processes these predicates, it cannot find their specification, and cannot initialize these routes, logging the error and omitting these routes.

Short term solution: when the routing package processes the route definitions, convert the legacy predicates in the Predicates slice to their right field.

Long term solution: migrate away from the legacy solution and support only the Predicates slice field. This way also making it analog to the solution of the filters.

szuecs:
So as far as I understand #560 , this is only an issue for loadbalanced routes.

PR #576, opened 2018-03-13
Support legacy predicates in generic list

aryszka:
Part of: #544

### DIFF
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -250,18 +250,21 @@ func isTreePredicate(name string) bool {
 	}
 }
 
-func getFreeStringArg(count int, p *eskip.Predicate) ([]string, error) {
-	err := func() error { return fmt.Errorf("invalid predicate args in %s", p.Name) }
-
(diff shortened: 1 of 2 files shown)


aryszka:
:-1:

routing/datasource.go
```diff
@@ -241,6 +250,87 @@ func isTreePredicate(name string) bool {
 	}
 }
 
+func getFreeStringArg(count int, p *eskip.Predicate) ([]string, error) {
+	err := func() error { return fmt.Errorf("invalid predicate args in %s", p.Name) }
```
szuecs: I don't like that, because it does not show exactly what was wrong.
Can you please have different error messages for length and type?

szuecs:
@aryszka I am fine with both or will it break something?
The test failure in travis should be fixed "TestTimedoutEndpointsRotation".

aryszka:
@szuecs this change won't break anything, it's just not the full fix

aryszka:
:+1:

PR #578, opened 2018-03-14
Using regex delimiter fails in character list in eskip #134

snurmine:
Using regex delimiter fails in character list in eskip #134 
Tried to keep behaviour related to `\` in `/.../` same.


aryszka:
:+1:

szuecs:
@snurmine can you fix the test case that broke in the dataclient, please?

szuecs:
:+1:

Issue #559, opened 2018-02-27
CORS filter


DeTeam:
Just experienced a use-case when it would be nice to have a convenience `cors` filter which would take care of setting the right response headers.

szuecs:
@DeTeam Could you describe how this should be ideally looks like from the user perspective?
Please also specify which headers we should write into the response.

I think it makes a lot of sense, but please provide more information, such that we can build it fast. Thanks!

DeTeam:
In my particular use-case: `Access-Control-Allow-Origin: *` + any methods + any headers
But I guess it's only good for testing.

It could also in theory take care of setting up a separate "pre-flight route" which would response to OPTION and quickly return cors headers.

<sup>1</sup> https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS - nice reference

Thinking about it again, it might be enough to just use `setResponseHeader("Access-Control-Allow-Origin", "*")` which we can already do manually 🤔

DeTeam:
Another thing which this filter could do: take some request headers (access-control-allow-*) and just return them (for methods and headers) %)

ruiaraujo:
@DeTeam It should take options for allowed origins since the header only allows sending a single origin.

szuecs:
@ruiaraujo I assigned you according to your twitter message, you might want to do additional work, so I let this issue open.

vetinari:
@ruiaraujo @szuecs  is this done with #595?

ruiaraujo:
I believe so but I will check when I try to use it. I will close it now myself and once I have it in production, if something is missing I will reopen.

PR #581, opened 2018-03-15
Add annotation for specifying custom routes on ingress

mikkeloscar:
This is an idea of how we could allow specifying custom routes on an ingress resource, to make it very simple to extend the application at the ingress layer.

My use case is that I have a REST API which is now called by a client-side application so I need to respond with the correct `Access-Control-Allow-Origin` headers. Instead of extending my application, with this I can just add an annotation (`zalando.org/skipper-routes`) to my ingress and quickly extend the API to send the right headers. The filter or predicate annotations are not enough because I need to respond with just the headers for the `OPTIONS` request and I can't proxy to my backend because it doesn't handle the `OPTIONS` request.

This is not yet tested, I just wanted to show a PoC for you to see before I go further with the idea.

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -639,16 +639,19 @@ func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
 	routes := make([]*eskip.Route, 0, len(items))
 	hostRoutes := make(map[string][]*eskip.Route)
 	for _, i := range items {
+		logger := log.WithFields(log.Fields{
+			"ingress": fmt.Sprintf("%s/%s", i.Metadata.Namespace, i.Metadata.Name),
+		})
 		if i.Metadata == nil || i.Metadata.Namespace == "" || i.Metadata.Name == "" ||
 			i.Spec == nil {
-			log.Warn("invalid ingress item: missing metadata")
+			logger.Warn("invalid ingress item: missing metadata")
 			continue
 		}
 
 		if r, ok, err := c.convertDefaultBackend(i); ok {
 			routes = append(routes, r...)
 		} else if err != nil {
-			log.Errorf("error while converting default backend: %v", err)
+			logger.Errorf("error while converting default backend: %v", err)
 		}
 
 		// TODO: only apply the filters from the annotations if it
@@ -678,7 +681,7 @@ func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
 			var err error
 			extraRoutes, err = eskip.Parse(annotationRoutes)
 			if err != nil {
-				log.Errorf("failed to parse routes from %s, skipping: %v", skipperRoutesAnnotationKey, err)
+				logger.Errorf("failed to parse routes from %s, skipping: %v", skipperRoutesAnnotationKey, err)
 			}
 		}
 
@@ -687,7 +690,7 @@ func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
 		if backends, ok := i.Metadata.Annotations[backendWeightsAnnotationKey]; ok {
 			err := json.Unmarshal([]byte(backends), &backendWeights)
 			if err != nil {
-				log.Errorf("error while parsing backend-weights annotation: %v", err)
+				logger.Errorf("error while parsing backend-weights annotation: %v", err)
 			}
 		}
 
@@ -695,7 +698,7 @@ func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
 		endpointsURLs := make(map[string][]string)
 		for _, rule := range i.Spec.Rules {
 			if rule.Http == nil {
-				log.Warn("invalid ingress item: rule missing http definitions")
+				logger.Warn("invalid ingress item: rule missing http definitions")
 				continue
 			}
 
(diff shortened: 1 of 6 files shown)


szuecs:
Thanks, this is also related to fix #559

dataclients/kubernetes/kube.go
```diff
@@ -49,6 +49,7 @@ const (
 	ratelimitAnnotationKey        = "zalando.org/ratelimit"
 	skipperfilterAnnotationKey    = "zalando.org/skipper-filter"
 	skipperpredicateAnnotationKey = "zalando.org/skipper-predicate"
+	skipperRoutesAnnotationKey    = "zalando.org/skipper-routes"
```
szuecs: This should not be needed, but reuse the predicate key.  
We want to migrate also all routes magic, which are similar to predicates into predicates.

mikkeloscar: But this is supposed to work differently than how the predicate annotation currently works. The predicate annotation just takes the predicates and adds them to the existing routes generated by the rules section of the ingress.

The routes annotation on the other hand is meant to create custom routes additional to the ones created from the ingress rules section.

These are two different behaviors so I believe it should be split into two different annotations.

aryszka:
I think this should be an opt-in feature because if a cluster is used as a shared environment, it becomes to make big damage without the cluster operator's consent. I suggest to create cmd flag defaulting to false. Am I being too paranoid?

aryszka:
Replying to myself: since we haven't considered sofar all the cases where one ingress can harm others anway, the feature is fine, and we probably have to come up with a better validation system than just the cmd flags if we want to address those concerns. So: this is fine.

mikkeloscar:
@aryszka The extra routes will still get the Host predicate(s) from the ingress rules such that they are local to the ingress resource. It's not global routes which could overwrite routes from other ingresses.

So it's only a problem if multiple ingresses has the same host names defined, but this was always a problem as you point out!

aryszka:
@mikkeloscar you are right. I only started reading the PR now :)

dataclients/kubernetes/kube.go
```diff
@@ -670,6 +671,17 @@ func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
...
 		}
 
+		// parse routes annotation
+		var extraRoutes []*eskip.Route
+		annotationRoutes := i.Metadata.Annotations[skipperRoutesAnnotationKey]
+		if annotationRoutes != "" {
+			var err error
+			extraRoutes, err = eskip.Parse(annotationRoutes)
+			if err != nil {
+				log.Errorf("failed to parse routes from %s, skipping: %v", skipperRoutesAnnotationKey, err)
```
aryszka: the namespace and the name of the current ingress can be a useful info in the error.

dataclients/kubernetes/kube.go
```diff
@@ -692,6 +704,13 @@ func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
 			// currently handled as mandatory
 			host := []string{"^" + strings.Replace(rule.Host, ".", "[.]", -1) + "$"}
 
+			// add extra routes from optional annotation
+			for _, route := range extraRoutes {
+				route.HostRegexps = host
+				route.Id = routeID("", route.Id, rule.Host, "", "")
```
aryszka: i would keep the namespace in the route id.

since a single route without an id is a valid eskip expression, i would try to be more specific when generating this id here. A single route expression looks like this (no id and no semicolon):

```
Method("OPTIONS") -> setResponseHeader("Foo", "bar") -> status(200) -> <shunt>
```

eskip.Parse will return a slice with one route that has an empty Id.

How about generating the route id here so:

```
routeID(namespace, name + "_" + route.Id + "_" + indexString, rule.Host, "", "")
```

where the indexString comes from the index in the enclosing loop. Would it work?

If yes, then of course it would be nice the create a new function, s.g. like routeIDForCustom() or a similar name where the args are more expressive regarding what the function does.

dataclients/kubernetes/kube_test.go
```diff
@@ -185,13 +193,16 @@ func testIngresses() []*ingressItem {
...
-		testIngress("namespace2", "svcwith2ports", "service4", "", "", "", backendPort{4444}, 1.0),
+		testIngress("namespace1", "ratelimit", "service1", "localRatelimit(20,\"1m\")", "", "", "", backendPort{8080}, 1.0),
+		testIngress("namespace1", "ratelimitAndBreaker", "service1", "", "localRatelimit(20,\"1m\") -> consecutiveBreaker(15)", "", "", backendPort{8080}, 1.0),
+		testIngress("namespace2", "svcwith2ports", "service4", "", "", "", "", backendPort{4444}, 1.0),
 	}
 }
 
 func checkRoutes(t *testing.T, r []*eskip.Route, expected map[string]string) {
+	for _, route := range r {
+		fmt.Println(route.Id)
```
aryszka: forgotten tracing or a case for t.Log()? I vote for removing it.

docs/kubernetes/ingress-usage.md
```diff
@@ -97,6 +98,44 @@ This example shows how to add predicates and filters:
               serviceName: app-svc
               servicePort: 80
 
+## Custom Routes
+
+Custom routes is a way of extending the default routes configured for your
```
aryszka: your -> an

mikkeloscar: fixed

dataclients/kubernetes/kube_test.go
mikkeloscar: Ups, that's not supposed to be there :) Removed!

dataclients/kubernetes/kube.go
mikkeloscar: Yeah I thought about this but thought having the host name there would be enough uniqueness. Still it makes sense to have more context in the id for debugging. I have done something like you suggested.

aryszka:
:+1:

dataclients/kubernetes/kube.go
szuecs: Thanks for your explanation. Then a new key makes sense.

szuecs:
super awesome, thanks!

szuecs:
:+1:

PR #582, opened 2018-03-15
Fix: empty group when no target endpoints + fix: id formatting for default backends with target endpoints + small improvements

aryszka:
- fix: remove load balancing when using service fallback, therefore removing invalid route with 0 size group
- fix: ID formatting, therefore remove duplicate IDs, therefore broken load balancing for default backends
- improve: remove unnecessary load balancing loopback when there's just a single target endpoint, but keep using the target endpoint
- refactor: remove unused method (listRoutes)


szuecs:
:+1:

mikkeloscar:
:+1:

Issue #585, opened 2018-03-17
Fix unstable tests: httpsredirect_test.go


mikkeloscar:
The tests in `dataclients/kubernetes/httpsredirect_test.go` are currently failing quite often, making PRs with unrelated changes fail.

I believe the failure is caused by a race in setting up the router and running tests against it before it has got the initial routes from the dataclient.

Some ideas for fixing this:
1. Drop the whole integration test and just test that we create the expected redirect route.
2. Make it possible to load routes from the dataclient in a synchronized way for testing.
3. Add a sleep after the router initialization (I don't like this).

/cc @aryszka @szuecs

aryszka:
@mikkeloscar can you confirm that this changeset was already applied when you ran this test? https://github.com/zalando/skipper/commit/2773183b47f7fcca0d25039b3b8bf09fff4bd829#diff-9ce10fec44f49e925acea5a07258d3b7 . Because it's not failing for me anymore.

1. dropping the integration: the reason i am not a big fan of it is that we had recently a couple of bugs that would have been immediately spotted if having at least one layer of integration

2. currently i often use this monitored logger for synchronizing tests with the tested behavior. Seemed like a good idea before, but now I think it is rather a hack. I would support adding some synchronization hooks, primarily in the routing.Routing object.

3. no sleeps. We've been there, did once a major cleanup session.

szuecs:
Thanks for sharing ideas how to fix the tests.
There was once a very good talk from Mitchel Hashimoto how to test in Go.

For timing dependent tests he presented a solution which we could try to use:

- do not use fake time!
- add multiplier to increase timeouts as CLI parameter and default it to 1

```
var int timeMultiplier // set from CLI with default = 1
func TestThing(t *testing.T) {
  // ..

  timeout := 3 * time.Minute * timeMultiplier

  select {
    case <-thingHappened:
    case <-time.After(timeout):
      t.Fatal("timeout")
  }
}
```

mikkeloscar:
> @mikkeloscar can you confirm that this changeset was already applied when you ran this test? 2773183#diff-9ce10fec44f49e925acea5a07258d3b7 . Because it's not failing for me anymore.

Yes, unfortunately I still see it fail when I run tests on master now. I also saw it on both travis and CDP when working on #581 where I also rebased on master after we merged your changes.

I have tried on both my old laptop and my desktop :) and it happens roughly every 1/10 runs.

> For timing dependent tests he presented a solution which we could try to use:

[...]

PR #587, opened 2018-03-17
remove concurrent test specs from https redirect test

aryszka:
Fixes: #585 

This PR only fixes the tests but not the root cause.

As it turns out, the failing tests were not caused here due to a concurrency race condition, but because of the current ingress client accepts multiple default backends from different ingress specs, therefore there are multiple default routes that can match a request that doesn't match any host and path rules.

The problem is referenced here: https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/kube.go#L352

The reason, why the test only occasionally fails, is that the routing package uses maps internally during processing the routes, and this introducses a certain level of randomness.

This PR removes the competing ingress specs from the https redirect tests.

### DIFF
--- a/dataclients/kubernetes/httpsredirect_test.go
+++ b/dataclients/kubernetes/httpsredirect_test.go
@@ -19,7 +19,7 @@ type redirectTest struct {
 	backend         string
 	fallbackBackend string
 	t               *testing.T
-	l *loggingtest.Logger
+	l               *loggingtest.Logger
 }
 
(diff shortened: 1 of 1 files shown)


mikkeloscar:
Awesome @aryszka thanks for looking into this and fixing it!

I always forget about the stupid default backend and I wonder if anyone is even using it (not just in skipper ingress, but ingress in general).

mikkeloscar:
:+1:

szuecs:
👍

PR #588, opened 2018-03-18
File watch

aryszka:
Replacing #257 and #275 

This is a much more primitive implementation of file watching than the other two efforts. This doesn't rely on system events, but just simple polls the file at the initially defined path. This way it is most useful only when experimenting/testing with route configurations on a local dev environment, but I assume that this would be the primary use case for this feature.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,7 @@ lint: build
 
 clean:
 	go clean -i ./...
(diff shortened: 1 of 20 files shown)


szuecs:
:+1:

aryszka:
:+1:

PR #592, opened 2018-03-21
load dataclients, filters and predicates from plugins

vetinari:
this one allows filters to be loaded as plugins, similar to the
opentracing ones.

A "noop" plugin would look like
```go
package main

import (
)

type noopSpec struct{}

func InitFilter(opts []string) (filters.Spec, error) {
	return noopSpec{}, nil
}

func (s noopSpec) Name() string {
	return "noop"
}
func (s noopSpec) CreateFilter(config []interface{}) (filters.Filter, error) {
	return noopFilter{}, nil
}

type noopFilter struct{}

func (f noopFilter) Request(filters.FilterContext)  {}
func (f noopFilter) Response(filters.FilterContext) {}
```

This needs to be compiled with

```sh
go build -buildmode=plugin -o noop.so ~/noop.go
```

and can then be used with

```sh
./bin/skipper -filter-plugins noop -inline-route 'any: * -> noop() -> status(200) -> <shunt>'
```

The -filter-plugin option accepts multiple filters separated by a "," like
```
  -filter-plugins "noop,geoip datafile=/path/to/file.db"
```

In this example, the "geoip" plugin will get an argument of
`datafile=/path/to/file.db` passed as init parameter (not to be confused
with a route paremeter).

Untested is if one could load the plugin multiple times with setting the value returned
by `Name()`


szuecs:
:+1:

lmineiro:
I'd prefer a default directory to scan for filter plugins and the flag would allow users to set a custom folder to scan for them. This way users would only need to package their filters and they would be picked up automatically.

If this sounds like a good idea, I propose:
`./plugins` as the root directory for plugins and inside it there could be:
* `/predicates` for all custom predicates plugins
* `/filters` for all custom filter plugins
* `/tracing` for opentracing implementations

What do you think?

vetinari:
The default plugin dir is the one set from skipper's -plugindir=/path/to/plugins (with a default of "./", i.e. skipper's working directory). Currently you can drop any plugin there and load it with either `-opentracing="tracerName opt=foo opt2=bar"` or  with this PR `-filter-plugins=noop,geoip db=/path/to/file"`.

skipper.go
```diff
@@ -352,6 +352,9 @@ type Options struct {
 	// PluginDir defines the dir to load plugins from
 	PluginDir string
 
+	// FilterPlugins loads additional filters from modules
```
aryszka: there could be a hint here in the codedoc, how to use the slice of slices

aryszka:
my preference for the flags would be:

```
skipper -filter-plugin=noop -filter-plugin="name=geoip,db=/path/to/file"
```

...meaning that one flag declares only one plugin, so there is only level of nesting inside a flag. (See the metrics for an example, of flag list with the same name.)

@lmineiro @vetinari @szuecs what do you think? does it make sense?

szuecs:
@aryszka I like your suggestion with multi args appending

aryszka:
regarding @lmineiro 's suggestion of the conventional ./plugins dir, i think the plugins from this dir should be tried to load automatically, and the flags should be necessary only if the plugin has mandatory arguments or someone wants override the defaults. E.g. to load the above noop, you just would simply drop it in the plugins dir.

Regarding the dir structure of the plugin dir, it is a nice convention to store the different plugins in different subdirectories by type, but we consider it only as a convention, and recursively load all plugins from each subdir, and detect which type of plugin it is simply by checking which interface it implements. Is this possible? Does it make sense?

szuecs:
I also like the default directory as suggested by @lmineiro. This will not be possible to add arguments, right or do you have an idea for that?

aryszka:
for the arguments, i think we can use the flags. I expect that a significant number of plugins can just be initialized with default arguments, and if it is routing related then the real args would come as part of the route configuration.

vetinari:
```
./bin/skipper -filter-plugins=noop -filter-plugins "geoip,db=./GeoLite2-Country.mmdb" -inline-routes 'any: * -> geoip() -> status(200) -> <shunt>'
```
works now

vetinari:
***NOTE*** this is a breaking change to the plugins dir, docker images and deployment filesneed to be updated to reflect the "tracing/" subdir in the plugin dir...

vetinari:
With the latest commits the breaking change warning does not apply anymore :)

... predicates and filters can be found in the newly created repos https://github.com/skipper-plugins/predicates and https://github.com/skipper-plugins/filters 

...currently this is just a `GeoIP` predicate and a `geoip` and a `noop` filter

skipper.go
```diff
@@ -352,10 +352,13 @@ type Options struct {
 	// PluginDir defines the directories to load plugins from
 	PluginDirs []string
 
-	// FilterPlugins loads additional filters from modules
+	// FilterPlugins loads additional filters from modules. The first value in each []string
+	// needs to be the plugin name (as on disk, without path, without ".so" suffix). The
+	// following values are passed as arguments to the plugin while loading, see docs/plugins.md
```
szuecs: docs/plugins.md ?

Please create dev docs here in godoc and if you like to link the new docs: 

1. mkdocs.yml to create a new page
2. create the new page in docs/plugins.md
3. link base: https://zalando.github.io/skipper/ path as defined in mkdocs.yaml

vetinari: fixed

vetinari:
https://github.com/zalando/skipper/compare/filter-plugins...multi-plugins ...maybe that's also worth merging

szuecs:
@aryszka are you fine with the plural flag or do you want to have it singular?

plugins.go
```diff
@@ -0,0 +1,194 @@
...
+			}
+			return nil
+		})
+	}
+
+	for _, fltr := range o.FilterPlugins {
+		name := fltr[0]
+		path, ok := found[name]
+		if !ok {
+			fmt.Printf("filter plugin %s not found in plugin dirs\n", name)
```
szuecs: log.Fatal

plugins.go
```diff
@@ -0,0 +1,194 @@
...
+			continue
+		}
+		spec, err := LoadFilterPlugin(path, fltr[1:])
+		if err != nil {
+			fmt.Printf("failed to load plugin %s: %s\n", path, err)
```
szuecs: log.Fatal

plugins.go
```diff
@@ -0,0 +1,194 @@
...
+		o.CustomFilters = append(o.CustomFilters, spec)
+		fmt.Printf("loaded plugin %s (%s) from %s\n", name, spec.Name(), path)
+		delete(found, name)
+	}
+
+	for _, pred := range o.PredicatePlugins {
+		name := pred[0]
+		path, ok := found[name]
+		if !ok {
+			fmt.Printf("predicate plugin %s not found in plugin dirs\n", name)
```
szuecs: log.Fatal

plugins.go
```diff
@@ -0,0 +1,194 @@
...
+			continue
+		}
+		spec, err := LoadPredicatePlugin(path, pred[1:])
+		if err != nil {
+			fmt.Printf("failed to load plugin %s: %s\n", path, err)
```
szuecs: log.Fatal

plugins.go
```diff
@@ -0,0 +1,194 @@
...
+		o.CustomPredicates = append(o.CustomPredicates, spec)
+		fmt.Printf("loaded plugin %s (%s) from %s\n", name, spec.Name(), path)
+		delete(found, name)
+	}
+
+	for _, pred := range o.DataClientPlugins {
+		name := pred[0]
+		path, ok := found[name]
+		if !ok {
+			fmt.Printf("data client plugin %s not found in plugin dirs\n", name)
```
szuecs: log.Fatal

plugins.go
```diff
@@ -0,0 +1,194 @@
...
+			continue
+		}
+		spec, err := LoadDataClientPlugin(path, pred[1:])
+		if err != nil {
+			fmt.Printf("failed to load plugin %s: %s\n", path, err)
```
szuecs: log.Fatal

plugins.go
```diff
@@ -0,0 +1,194 @@
...
+		mod, err := plugin.Open(path)
+		if err != nil {
+			fmt.Printf("open plugin %s from %s: %s\n", name, path, err)
+			continue
+		}
+
+		if sym, err := mod.Lookup("InitFilter"); err == nil {
+			spec, err := loadFilterPlugin(sym, path, []string{})
+			if err != nil {
+				fmt.Printf("filter plugin %s returned: %s\n", path, err)
```
szuecs: log.Fatal

plugins.go
```diff
@@ -0,0 +1,194 @@
...
+		o.CustomDataClients = append(o.CustomDataClients, spec)
+		fmt.Printf("loaded plugin %s from %s\n", name, path)
+		delete(found, name)
+	}
+
+	for name, path := range found {
+		fmt.Printf("attempting to load plugin from %s\n", path)
+		mod, err := plugin.Open(path)
+		if err != nil {
+			fmt.Printf("open plugin %s from %s: %s\n", name, path, err)
```
szuecs: log.Fatal

plugins.go
```diff
@@ -0,0 +1,194 @@
...
+				continue
+			}
+			o.CustomFilters = append(o.CustomFilters, spec)
+			fmt.Printf("filter plugin %s loaded from %s\n", name, path)
+		}
+
+		if sym, err := mod.Lookup("InitPredicate"); err == nil {
+			spec, err := loadPredicatePlugin(sym, path, []string{})
+			if err != nil {
+				fmt.Printf("predicate plugin %s returned: %s\n", path, err)
```
szuecs: log.Fatal

plugins.go
```diff
@@ -0,0 +1,194 @@
...
+			}
+			o.CustomPredicates = append(o.CustomPredicates, spec)
+			fmt.Printf("predicate plugin %s loaded from %s\n", name, path)
+		}
+
+		fmt.Printf("checking %s for data client in %s\n", name, path)
+		if sym, err := mod.Lookup("InitDataClient"); err == nil {
+			spec, err := loadDataClientPlugin(sym, path, []string{})
+			if err != nil {
+				fmt.Printf("data client plugin %s returned: %s\n", path, err)
```
szuecs: log.Fatal

skipper.go
```diff
@@ -349,8 +349,22 @@ type Options struct {
 	// OpenTracing enables opentracing
 	OpenTracing []string
 
-	// PluginDir defines the dir to load plugins from
-	PluginDir string
```
szuecs: let it and mark as deprecated. Use append(PluginDirs, PluginDir) then this will not be a breaking change and we can do later the get rid of deprecated stuff.

tracing/tracing.go
```diff
@@ -59,7 +59,7 @@ import (
 )
 
 // LoadPlugin loads the given opentracing plugin and returns an opentracing.Tracer
-func LoadPlugin(pluginDir string, opts []string) (ot.Tracer, error) {
+func LoadPlugin(pluginDirs []string, opts []string) (ot.Tracer, error) {
```
szuecs: An internal function or exposed "LoadPlugins" could do the file traversal and call the exposed LoadPlugin(), such that you do not need a breaking change

cmd/skipper/main.go
```diff
@@ -271,7 +274,7 @@ func init() {
 	flag.BoolVar(&enableRatelimiters, "enable-ratelimits", false, enableRatelimitUsage)
 	flag.Var(&ratelimits, "ratelimits", ratelimitUsage)
 	flag.StringVar(&openTracing, "opentracing", "noop", opentracingUsage)
-	flag.StringVar(&pluginDir, "plugindir", ".", pluginDirUsage)
+	flag.StringVar(&pluginDir, "plugindir", "", pluginDirUsage)
```
szuecs: pluginDirUsage should have a deprecation mark, can you check if this is in the usage string?

szuecs:
:+1:

cmd/skipper/filterFlags.go
```diff
@@ -0,0 +1,89 @@
...
+	var ret []string
+	for _, val := range f.values {
+		ret = append(ret, strings.Join(val, ","))
+	}
+	return strings.Join(ret, " ")
+}
+
+func (f *dataclientFlags) Set(value string) error {
+	if f == nil {
+		f = &dataclientFlags{}
```
aryszka: i don't think this works. We're not setting here a value for the *f, but we're creating this way a new pointer, replacing f. This way the calling code will still have the old nil value. https://play.golang.org/p/ttasuPgILeI

if we were setting *f, that still would not be good, because it's nil, so we cannot dereference it.

I would simple remove this nil check and expect the calling code to initialize the pointer as usual.

cmd/skipper/filterFlags.go
```diff
@@ -0,0 +1,89 @@
+package main
+
+import (
+	"strings"
+)
+
+const filterPluginUsage = "set a custom filter plugins to load, a comma separated list of name and arguments"
```
aryszka: "...a custom filter plugin..." (singular)

cmd/skipper/filterFlags.go
```diff
@@ -0,0 +1,89 @@
...
+	var ret []string
+	for _, val := range f.values {
+		ret = append(ret, strings.Join(val, ","))
+	}
+	return strings.Join(ret, " ")
+}
+
+func (f *filterFlags) Set(value string) error {
+	if f == nil {
+		f = &filterFlags{}
```
aryszka: see my comment for the `dataclientFlags`

cmd/skipper/filterFlags.go
```diff
@@ -0,0 +1,89 @@
...
+		f.values = append(f.values, strings.Split(v, ","))
+	}
+	return nil
+}
+
+func (f *filterFlags) Get() [][]string {
+	return f.values
+}
+
+const predicatePluginUsage = "set a custom predicate plugins to load, a comma separated list of name and arguments"
```
aryszka: "...a custom predicate plugin..." (singular)

cmd/skipper/filterFlags.go
```diff
@@ -0,0 +1,89 @@
...
+	var ret []string
+	for _, val := range f.values {
+		ret = append(ret, strings.Join(val, ","))
+	}
+	return strings.Join(ret, " ")
+}
+
+func (f *predicateFlags) Set(value string) error {
+	if f == nil {
+		f = &predicateFlags{}
```
aryszka: see my comment for the `dataclientFlags`

cmd/skipper/filterFlags.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (f *filterFlags) String() string {
+	var ret []string
+	for _, val := range f.values {
+		ret = append(ret, strings.Join(val, ","))
+	}
+	return strings.Join(ret, " ")
+}
+
+func (f *filterFlags) Set(value string) error {
```
aryszka: is the logic exactly the same for filterFlags, predicateFlags and dataclientFlags? If yes, is it possible to create only type for them, s.g. like pluginFlags, and use it in each case?

plugins.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/zalando/skipper/routing"
+)
+
+func findAndLoadPlugins(o *Options) error {
+	found := make(map[string]string)
+
+	for _, dir := range o.PluginDirs {
+		filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+			if err != nil || info.IsDir() {
+				return nil
```
aryszka: maybe it would be worth to at least print a warning in case err != nil

aryszka:
@vetinari i hope it's fine, i added some simplifications to the flag parsing

aryszka:
:+1:

lmineiro:
What do you think about a convention such as a best effort attempt to load a `foo.conf` file with configuration for the `foo` plugin in the same directory as the plugin itself `foo.so`? It would look tidy enough for me.

plugins.go
szuecs: I would even suggest to do a log.Fatal, because if you roll this out with systems like Kubernetes, this would automatically stop the roll out. If you ignore errors, but the production system thinks to have the activated plugin, this would be a much bigger incident, then that the roll out is stopped.

plugins.go
```diff
@@ -0,0 +1,185 @@
...
+	"os"
+	"path/filepath"
+	"plugin"
+	"strings"
+
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/routing"
+)
+
+func findAndLoadPlugins(o *Options) error {
```
szuecs: This function could be split up into 3-4 functions, because the loops are independent as far as I understand.
Another idea would be to have these functions as methods (function with receiver *Options).

    func (o *Options) walkPlugindir() error
    func (o *Options) loadFilterPlugins() error
    func (o *Options) loadPredicatePlugins() error
    func (o *Options) loadDataclientPlugins() error

szuecs:
@lmineiro I like your idea, but I am also fine with creating an issue to enhance this feature later.
From my side 👍 , besides the ignore error in filepath.Walk usage.

aryszka:
the conf file is a good idea, but i also think that it's good enough to add it as an enhancement in a separate PR

szuecs:
:+1:

vetinari:
:+1:

PR #593, opened 2018-03-21
Swarm ratelimiter

szuecs:
We (@vetinari @myhro @aryszka @szuecs) started to have the first use case implementation for cluster based data exchange in our hackweek. 

The use case is: clusterRatelimit

We use memberlist from hashicorp as swim implementation as data exchange protocol.
Skipper running as Kubernetes ingress-controller is able to share information for rate limiting (backend or client based) for the whole cluster per route. 
The current PR is already tested and works. To make it production ready, we have to:
- [x] expose some metrics (number of swarm values and exchange rate)
- [x] drop dead code
- [x] rework logging
- [x] refactor the code
- [x] add unit tests
- [ ] test in a dev cluster
- [x] document architecture decision (why swim was chosen)
- [x] document how to use it in a skipper deployment
- [x] document dev docs in godoc


Raffo:
I haven't reviewed the PR (and I guess it's not ready), but why using a gossip protocol when you could rely on a central coordination which in case of Kubernetes could be the API server? What are the considerations that brought you to this decision? Documenting them is probably valuable :-)

szuecs:
@Raffo I promise more docs, but in short we need a concurrent writeable "datastore" with low latency, but we are fine with eventual consistent data.
We discussed to use simply redis, but redis is not good for concurrent data access, it will just block the data path.
Kubernetes API is good for single write multiple reads. We need multiple writes and as you also know kubernetes is not really good for low latency data exchange.

Last hackweek we started a POC using swim/memberlist and this hackweek we implemented this ratelimit use case. Any other ideas are welcome.

szuecs:
@Raffo is the part in docs.go good enough to understand the why swim was chosen in favor of kubernetes API ?

swarm/doc.go
```diff
@@ -0,0 +1,24 @@
+// Package swarm implements functionality to exchange information
+// between skipper instances. It is meant for multiple readers and
+// writers, with low latency, weakly consistent data. As an example
+// tere is the filter clusterRatelimit implementation, that uses swarm
```
lmineiro: typo: s/tere/there

lmineiro:
I think this can be further improved if the member discovery is made abstract and we can plug multiple/different ways of finding members.

Although Kubernetes is an obvious one, I'm confident that other, non-Kubernetes methods can also be helpful and increase Skipper's usability outside of this particular environment.

An idea that we had in the past was also to allow more customization of the bucket key for the rate limiting. The IP address is the de-facto way to do this but, in the past, we had a couple of times where going higher in the model could be helpful, like the Oauth2 client_id. This, obviously, could come later as another improvement.

swarm/doc.go
szuecs: ty fixed

szuecs:
@lmineiro I agree that an abstract interface could be the goal, but I am not sure if this should be done right now. In the end it should be also pluggable, but it does not need to be in the first merge to master.

I also would like to have something like bronze, silver, gold tokens to be able to sell N number of API req/s per token information for example. All of this requires #367 (or similar) to be done and this is not the only GH issue waiting for this PR to be done.

szuecs:
@Raffo thanks for your help! 
I modified it slightly, but used in general your modifications.

filters/ratelimit/ratelimit.go
```diff
@@ -109,6 +133,55 @@ func serviceRatelimitFilter(args []interface{}) (filters.Filter, error) {
...
 }
 
+func clusterRatelimitFilter(args []interface{}) (filters.Filter, error) {
+	if !(len(args) == 2 || len(args) == 3) {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	var err error
+	var maxHits int
+	if len(args) > 0 {
```
aryszka: this will be always true

filters/ratelimit/ratelimit.go
```diff
@@ -109,6 +133,55 @@ func serviceRatelimitFilter(args []interface{}) (filters.Filter, error) {
...
+		maxHits, err = getIntArg(args[0])
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	var timeWindow time.Duration
+	if len(args) > 1 {
```
aryszka: this will be always true, because of:

```
	if !(len(args) == 2 || len(args) == 3) {
	}
```

ratelimit/registry.go
```diff
@@ -23,10 +24,11 @@ type Registry struct {
 	global   Settings
 	//routeSettings map[string]Settings
 	lookup map[Settings]*Ratelimit
+	swarm  *swarm.Swarm
 }
 
 // NewRegistry initializes a registry with the provided default settings.
-func NewRegistry(settings ...Settings) *Registry {
+func NewRegistry(sw *swarm.Swarm, settings ...Settings) *Registry {
```
aryszka: on one hand, we typically try to avoid breaking, on the other hand, i don't think someone uses the ratelimit package separately. So the question is whether it is fine to refactor a bit the exported interface of this package? Because then i would have other suggestions, too.

If not, then we can just create a new function called NewSwarmRegistry, and the NewRegistry would simply call it with nil for the sw param.

ratelimit/cluster.go
```diff
@@ -0,0 +1,133 @@
...
+
+type Swarmer interface {
+	ShareValue(string, interface{}) error
+	Values(string) map[string]interface{}
+}
+
+type ClusterLimit struct {
+	name    string
+	local   implementation
+	maxHits float64
```
aryszka: maxHits should be an integer

ratelimit/registry.go
szuecs: fixed

docs/operation/operation.md
```diff
@@ -398,3 +413,53 @@ In order to control this limits, there are two parameters: `limit` and
 to get the results paginated or getting all of them at the same time.
 
     % curl localhost:9911/routes?offset=200&limit=100
+
+# Memory consumption
+
+Memory consumption is in general no problem, but some features can if
```
aryszka: "Memory consumption is in general no problem, but some features can if
not carefully used create more memory consumption that you might
thought.

Heavy memory consumers:
- Metrics
- Filters
- Slow Backends and chatty clients"

could be:

"While Skipper is generally not memory bound, some features may require
some attention and planning regarding the memory consumption.

Potentially high memory consumers:
- Metrics
- Filters
- Slow Backends and chatty clients"

swarm/doc.go
```diff
@@ -0,0 +1,33 @@
+/*
+Package swarm implements the exchange of information between
+Skipper instances using a gossip protocol with name swim.  This
```
aryszka: "with name swim" => "called SWIM"

skipper.go
```diff
@@ -760,13 +771,36 @@ func Run(o Options) error {
...
+				KubernetesAPIBaseURL: o.KubernetesURL,
+				Namespace:            o.SwarmKubernetesNamespace,
+				LabelSelectorKey:     o.SwarmKubernetesLabelSelectorKey,
+				LabelSelectorValue:   o.SwarmKubernetesLabelSelectorValue,
+			}
+		}
+		theSwarm, err = swarm.NewSwarm(swops)
+		if err != nil {
+			log.Errorf("failed to init swarm with options %+v: %v", swops, err)
+		}
```
aryszka: shouldn't we call here `defer theSwarm.Leave()`?

swarm/kubernetes.go
```diff
@@ -0,0 +1,262 @@
...
+type ClientKubernetes struct {
+	httpClient *http.Client
+	apiURL     string
+	token      string
+}
+
+// Get does the http GET call to kubernetes API to find the initial
+// peers of a swarm.
+func (c *ClientKubernetes) Get(s string) (*http.Response, error) {
+	req, err := c.createRequest("GET", s, nil)
```
aryszka: also could be just `c.httpClient.Get(s)`

swarm/kubernetes.go
```diff
@@ -0,0 +1,262 @@
...
+
+	bToken, err := ioutil.ReadFile(tokenFilePath)
+	if err != nil {
+		return "", err
+	}
+
+	return string(bToken), nil
+}
+
+func (c *ClientKubernetes) getEndpoints(ns, name string) ([]string, error) {
```
aryszka: where is this called from?

swarm/swarm.go
```diff
@@ -0,0 +1,382 @@
...
+		getValues:        getValues,
+		listeners:        listeners,
+		leave:            leave,
+		shared:           shared,
+		mlist:            ml,
+		cleanupF:         cleanupF,
+		metrics:          metrics.Default,
+	}
+
+	// TODO(sszuecs): maybe we should wrap it in a recover for panic, but we need to close the channels
```
aryszka: what would be the recovery strategy in case of panic?

we don't need to close the channels, but we need to return from the goroutine. This can be done if we call the Leave() function. Suggested in skipper.go

swarm/swarm.go
```diff
@@ -0,0 +1,382 @@
...
+		ret: make(chan map[string]interface{}),
+	}
+	s.getValues <- req
+	return <-req.ret
+}
+
+// Leave sends a signal for the local node to leave the Swarm.
+func (s *Swarm) Leave() {
+	close(s.leave)
+	s.cleanupF()
```
aryszka: i can't find it, is it sure that s.cleanupF is not nil? I would make sure to set a default noop in Join() like `if cleanupF == nil { cleanupF = func() {} }`

swarm/kubernetes.go
szuecs: deleted

skipper.go

ratelimit/cluster.go
```diff
@@ -0,0 +1,136 @@
...
+	_ = c.local.Allow(s) // update local rate limit
+
+	if err := c.swarm.ShareValue(swarmPrefix+s, t0); err != nil {
+		log.Errorf("clusterRatelimit failed to share value: %v", err)
+	}
+
+	swarmValues := c.swarm.Values(swarmPrefix + s)
+	log.Debugf("%s: clusterRatelimit swarmValues(%d) for '%s': %v", c.name, len(swarmValues), swarmPrefix+s, swarmValues)
+
+	c.resize <- resizeLimit{s: s, n: len(swarmValues)}
```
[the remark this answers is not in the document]
szuecs: I think the implementer did want to decouple it from the hot path. Should I change it to call Resize()?

ratelimit/cluster.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	// TODO(sszuecs): we might want to have one goroutine for all of these
+	go func() {
+		for {
+			select {
+			case size := <-rl.resize:
+				log.Debugf("resize clusterRatelimit: %v", size)
+				// TODO(sszuecs): call with "go" ?
+				rl.Resize(size.s, rl.maxHits/size.n)
```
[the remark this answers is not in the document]
szuecs: I think it was, because of not having Resize() in the hot path of the request.
Why does a request needs to wait for the resize of a datastructure?

ratelimit/cluster.go
```diff
@@ -0,0 +1,136 @@
...
+
+	log "github.com/sirupsen/logrus"
+	circularbuffer "github.com/szuecs/rate-limit-buffer"
+)
+
+// Swarmer interface defines the requirement for a Swarm, for use as
+// an exchange method for cluster ratelimits:
+// ratelimit.ClusterServiceRatelimit and
+// ratelimit.ClusterClientRatelimit.
+type Swarmer interface {
```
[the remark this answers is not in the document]
szuecs: I changed it to a).

swarm/kubernetes.go
szuecs: No because of the potentially used token. We read the token from the secret of the serviceAccount we run.
Maybe I missunderstand your point?

swarm/swarm.go
szuecs: I dropped the TODO and did the Leave as suggested.
Something open here?

ratelimit/cluster.go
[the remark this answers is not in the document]
aryszka: i see. Then let's keep it, and measure the effect as a follow up.

The reason why I am still not sure if we need it is that we normally have more pending requests than CPUs. So if we continue the goroutine of the request without waiting for the resize to finish, my asumption is, the entire throughput of requests won't be higher, because we are calling the resize on every single request, and so there's still the same amount of tasks to complete. Actually, it is even possible that the total throughput can get a little lower because of the higher number synchronization steps. Just guessing.

Both the current solution and the proposed change are optimizations and need proof, while the logic looks correct to me, so let's keep it now, and measure separately.

swarm/kubernetes.go
aryszka: true, my bad, missed the token part.

ratelimit/cluster.go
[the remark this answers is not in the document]
aryszka: as discussed in the other comment, it's fine, let's get numbers first.

swarm/swarm.go
aryszka: lgtm

aryszka:
:+1:

szuecs:
:+1:

docs/operation/operation.md
```diff
@@ -398,3 +413,52 @@ In order to control this limits, there are two parameters: `limit` and
...
+
+    % curl localhost:9911/metrics
+
+## Filters
+
+Ratelimit filters `clientRatelimit` and `clusterClientRatelimit`, both consume
+roughly 15MB per filter for 100.000 individual clients and 10 maximum
+hits. Make sure you monitor Go metrics.
+
+## Slow Backends
```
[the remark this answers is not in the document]
szuecs: What is the difference between a slow backend and slow writes from a backend?

Issue #438, opened 2017-09-26
rate limit should set header Retry-After


szuecs:
Based on https://tools.ietf.org/html/rfc6585#section-4 you should set HTTP header "Retry-After".

szuecs:
https://github.com/zalando/skipper/pull/596 was started but not finished.

conorclifford:
@szuecs do you know if there is any planned timeline for finishing #596 ?

szuecs:
@conorclifford PR is done by @myhro, so I don't know. Maybe he can enlighten us if someon else should take it.
I did not plan this to be done. When do you need it?

myhro:
@szuecs Unfortunately I wasn't able to finish the PR. I see no problem if somebody else takes it, as I can't promise that I'll be able to continue its development soon.

conorclifford:
@szuecs I believe it is being looked at as being useful nearly immediately, and will allow off-lifting similar work across several applications.

tkrop:
Our [API Guidelines](https://opensource.zalando.com/restful-api-guidelines/#153) are open how to implement this. However, we should make sure, that it does not provide a the same answer future request time (see https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/ for reasoning) for multiple rejected requests.

A constant jitter interval depending on the allowed rate should be the minimum. However, I'm thinking whether there is a smarter solution that also takes the rejected request rate into consideration to determine the jitter. My thinking for a simple approach is to create jitter in the order of the rejected request rate.

E.g. if you have rejected the same number of requests in the last period as you have served, the jitter interval to choose from should be as big as 2 (factor guessed) times the interval. This leaves 50% space for new requests in average, but creates an uneven distribution with the bulk to the begin.

While this is surly an improvement to any unjittered implementation, the resulting distribution should still be checked whether it is creating to much pressure on the short end. In this case the factor should be adapted or even change to a none-linear function.

dehora:
> While this is surly an improvement to any unjittered implementation

@tkrop I don't understand, 1) why jitter is a feature of middleware like skipper and not a recommendation on origin client implementations, 2) why this is relevant guidance on the API interface level. Can you elaborate?

tkrop:
@dehora I think, `Retry-After` is and should not be a client issue - clients should use the value just as it is. In my opinion, using `Retry-After` shifts the responsibility for coordinating subsequent request to the server side (Skipper).

[...]

PR #596, opened 2018-03-23
Retry after

myhro:
Hi,

I'm working on implementing the `Retry-After` header described on #438. This is what I get when I run some requests against the Skipper version in this PR:

```
(...)
[APP]INFO[0000] route settings, reset, route: foo: * -> ratelimit(1, "5s") -> "https://www.google.de/"
::1 - - [23/Mar/2018:16:21:53 +0100] "HEAD / HTTP/1.1" 200 0 "" "curl/7.54.0" 222 localhost:9090
::1 - - [23/Mar/2018:16:21:54 +0100] "HEAD / HTTP/1.1" 429 18 "" "curl/7.54.0" 0 localhost:9090
```

In the client we can see the headers properly set:

```
$ curl -I localhost:9090
HTTP/1.1 200 OK
(...)

$ curl -I localhost:9090
HTTP/1.1 429 Too Many Requests
Retry-After: 5
Server: Skipper
X-Rate-Limit: 720
Date: Fri, 23 Mar 2018 15:21:54 GMT
Content-Length: 18
```

The problem is that the test I wrote isn't passing, even when getting a `429` response:

```
$ go test -run TestRetryAfterHeader proxy/ratelimit_test.go
--- FAIL: TestRetryAfterHeader (0.00s)
        ratelimit_test.go:369: should set retry header Retry-After
FAIL
FAIL    command-line-arguments  0.018s
```

With @vetinari help, we realised that the coding is returning on the check after the line `settings, ok := ctx.stateBag[ratelimitfilters.RouteSettingsKey].(ratelimit.Settings)` of the `Proxy.checkRatelimit()` method.

I though that the only place where the rate limit was actually checked was at the end of the same method, where we have `return settings, rl.Allow(s)`, but there's also [the `Registry.Check()` method of the ratelimit][1] that does something similar.

So, what should we do here? Change the behaviour of the test, ensuring that `ratelimitfilters.RouteSettingsKey` is set, to mimic the behaviour of the manual test? Or the `Ratelimit.RetryAfter()` method should not be called on the `Proxy.checkRatelimit()` but on `Registry. Check()`?

Regards,
Tiago.

[1]: https://github.com/zalando/skipper/blob/62ad0f8/ratelimit/registry.go#L74

### DIFF
--- a/filters/ratelimit/ratelimit.go
+++ b/filters/ratelimit/ratelimit.go
@@ -12,6 +12,9 @@ import (
 	"github.com/zalando/skipper/ratelimit"
 )
 
+// RetryAfterKey is used as key in the context state bag
+const RetryAfterKey = "#ratelimitretryafter"
+
 // RouteSettingsKey is used as key in the context state bag
 const RouteSettingsKey = "#ratelimitsettings"
 
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: ed4c9832f72de1aeb33488d81331aa379adc61ba07337c51e15bada9c8299b2d
-updated: 2018-01-04T08:00:57.348776525+01:00
+hash: 9553ba72abe52ef7ffb117fe347b1a8530cc0315d011f5b944e0035551511c67
+updated: 2018-03-22T23:34:00.457358076+01:00
 imports:
 - name: github.com/abbot/go-http-auth
   version: efc9484eee77263a11f158ef4f30fcc30298a942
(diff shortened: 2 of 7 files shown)


aryszka:
my understanding is that in the case of this test, the retry-after value is missing. Without a clear idea at the moment how, I would try to do the following:

- since this retry-after value is set and read only by the proxy, it could be a package internal field on the context
- i would try to refactor the global check (registry.Check) to make it similar to the route specific, where we have a handle to the ratelimiter instance, and so we get the retry-after value

myhro:
Hi Arpad,

Yes, I agree that the `RetryAfter` value could be stored somewhere else to avoid the `stateBag`, but I think I wasn't clear describing the problem. The value that isn't present there in my tests is `ratelimitfilters.RouteSettingsKey`, not `ratelimitfilters.RetryAfterKey`, that will be set later. So the state bag is not really the problem in the retry after case, only the `RouteSettingsKey` that isn't present in there - and this is probably happening because the test case doesn't have defined routes.

I tried check for the [value of `s.Type` in the `Registry.Check()` method][1]. When I ran the test suite, its value is `LocalRatelimit` (`1`), but when I running Skipper in the command line (with `./bin/skipper -enable-ratelimits -routes-file ...`), its value is `DisableRatelimit` (`3`). That gives us a hint on why in the test suite the values are returned in this `switch/case` in there (and not in `Proxy.checkRatelimit()`).

What is not clear to me why it keeps working on the command line, as I wasn't expecting for `DisableRatelimit` when `-enable-ratelimits` is set. I need to debug this further to have a better picture of the whole settings definition between these two different executions (test suite and command line).

[1]: https://github.com/zalando/skipper/blob/62ad0f8/ratelimit/registry.go#L80

aryszka:
Hi Tiago,

I believe that the difference is between when you run it in the command line and in the test is that in the command line you use a route specific ratelimiter specification while in the test a global one. And the checkRatelimit() function is executed only for the route specific cases, while the global case simply uses the registry.Check() function, and the settings are not set to the state bag in that case. I tried this by setting a fake value where the registry.Check() is called, and then the test passes. That's why suggesting to refactor the registry.Check() function to something, e.g. GetGlobal(), so that it would return the handle to Ratelimit object, and then the RetryAfter() function could be called in this case, too.

PR #597, opened 2018-03-26
Multi plugins

vetinari:
so https://github.com/aryszka/configfilter could be converted to a plugin ;-)

*  when loading a `.so` file, it is checked if there is also a `.conf` file
  with the same base, e.g. for `./plugins/noop.so` a config of
  `./plugins/noop.conf` is attempted to open. If it exists, each line
  in this file is passed as argument to the plugin loader function
  (leading and trailing spaces stripped, lines beginning with a `#`
  after space stripping are ignored).
* add support for multitype plugins complete auto detection and `.conf`
 file support
* add some noop plugins in _test_plugins and add a make target to build
  and try loading as test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ vendor
 linter.log
 site/
 plugins/
+_test_plugins/*.so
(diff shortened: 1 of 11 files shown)


vetinari:
err ... not to this branch...

PR #598, opened 2018-03-26
Multi plugins, plugin .conf filles

vetinari:
so https://github.com/aryszka/configfilter could be converted to a plugin ;-)

*  when loading a `.so` file, it is checked if there is also a `.conf` file
  with the same base, e.g. for `./plugins/noop.so` a config of
  `./plugins/noop.conf` is attempted to open. If it exists, each line
  in this file is passed as argument to the plugin loader function
  (leading and trailing spaces stripped, lines beginning with a `#`
  after space stripping are ignored).
* add support for multitype plugins complete auto detection and `.conf`
 file support
* add some noop plugins in _test_plugins and add a make target to build
  and try loading as test

### DIFF
--- a/plugins.go
+++ b/plugins.go
@@ -29,6 +29,31 @@ func findAndLoadPlugins(o *Options) {
 		})
 	}
 
+	for _, plug := range o.MultiPlugins {
+		name := plug[0]
+		path, ok := found[name]
+		if !ok {
+			fmt.Printf("mutlitype plugin %s not found in plugin dirs\n", name)
+			continue
+		}
+		fltr, pred, dc, err := LoadMultiPlugin(path, plug[1:])
+		if err != nil {
+			fmt.Printf("failed to load plugin %s: %s\n", path, err)
+			continue
+		}
+		if fltr != nil {
+			o.CustomFilters = append(o.CustomFilters, fltr)
+		}
+		if pred != nil {
+			o.CustomPredicates = append(o.CustomPredicates, pred)
+		}
+		if dc != nil {
+			o.CustomDataClients = append(o.CustomDataClients, dc)
+		}
+		fmt.Printf("loaded plugin %s from %s\n", name, path)
+		delete(found, name)
+	}
+
 	for _, fltr := range o.FilterPlugins {
 		name := fltr[0]
 		path, ok := found[name]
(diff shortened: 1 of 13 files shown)


vetinari:
looking at https://github.com/zalando-stups/skrop I wonder if we should now change the signature of the `InitFilter` from `(opts []string) (filters.Spec, error)` to `(opts []string) ([]filters.Spec, error)`...

aryszka:
@vetinari about `InitFilters(opts []string) ([]filters.Spec, error)`..

It would be nice, but how would the config arguments be assigned to the different filters? `opts [][]string`?

plugins.go
```diff
@@ -116,16 +107,98 @@ func findAndLoadPlugins(o *Options) error {
 	return nil
 }
 
+func (o *Options) LoadMultiPlugins(found map[string]string) error {
```
aryszka: I think that the methods matching `Load.+Plugins?` here should not be exported. Is there a reason for exporting them, or a use case for applications that import skipper? I see that they are already exported in the master, but maybe it's not late to change this, if there is no reasonable use case for it. What may it break if unexporting them all?

plugins.go
```diff
@@ -116,16 +107,98 @@ func findAndLoadPlugins(o *Options) error {
...
+
+	sym, err := mod.Lookup("InitPlugin")
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("lookup module symbol failed for %s: %s", path, err)
+	}
+	return loadMultiPlugin(sym, path, append(conf, args...))
+}
+
+func loadMultiPlugin(sym plugin.Symbol, path string, args []string) (filters.Spec, routing.PredicateSpec, routing.DataClient, error) {
+	fn, ok := sym.(func([]string) (filters.Spec, routing.PredicateSpec, routing.DataClient, error))
```
aryszka: How would it be to have an `InitPlugin(opts []string) ([]filters.Spec, []routing.Predicate, []routing.DataClient)` generic method with slices of return values, that can return multiple of any filters, predicates and data clients?  This could be the generic plugin interface, while we could still keep the InitFilter, InitPredicate and InitDataClient interfaces to support the implementations of simpler plugins, with a single return value.

plugins.go
```diff
@@ -85,8 +54,31 @@ func findAndLoadPlugins(o *Options) error {
...
+			if fltr != nil {
+				o.CustomFilters = append(o.CustomFilters, fltr)
+			}
+			if pred != nil {
+				o.CustomPredicates = append(o.CustomPredicates, pred)
+			}
+			if dc != nil {
+				o.CustomDataClients = append(o.CustomDataClients, dc)
+			}
+			fmt.Printf("mutlitype plugin %s loaded from %s (filter: %t, predicate: %t, dataclient: %t)\n",
```
aryszka: We follow the convention of printing log messages to the stderr. Please, change 'fmt.Println' to log.Println, using logrus. (Logrus allows the user to override the output.)

skipper.go
```diff
@@ -368,6 +368,10 @@ type Options struct {
 	// what the []string should contain.
 	DataClientPlugins [][]string
 
+	// MultiPlugins combine multiple types of the above plugin types in one plugin (where
+	// necessary because of shared data between e.g. a filter and a data client).
+	MultiPlugins [][]string
```
aryszka: If we consider the multi plugin the generic form of plugins, we could also just call this option simply: `Plugins`. What do you think?

aryszka:
ok, see my other comment. What I suggest basically, there should be a possibility to define a generic `InitPlugin(opts []string) ([]filters.Spec, []routing.PredicateSpec, []routing.DataClient)`, and we can leave the `InitFilter`, `InitPredicate` and `InitDataClient` cases as they are, only to support the easy implementation of the simplest plugins. Also, I would make it fail when more than one of these methods are implemented by a single plugin.

vetinari:
:+1:

aryszka:
:+1:

PR #599, opened 2018-03-27
prevent startup failure when default plugin dir is missing

aryszka:
skipper currently cannot be started if the plugin dir is missing. This is fine if an explicit plugin dir is specified, but not when plugins are not intended to be used, and breaks compatibility this way.

The fix is a cheap one: ignore if the error is a path error and the dirname is the default. Open for better suggestions.

This is related to: https://github.com/zalando/skipper/pull/598


szuecs:
:+1:

szuecs:
CDP has an error which is independent of skipper.

aryszka:
:+1:

PR #600, opened 2018-03-27
plugins dir does not exist

szuecs:
fix: if not plugins are mounted as volume, skipper should have an empty plugins directory to not fail to start

```
% ./bin/skipper
2018/03/27 17:33:22 failed to search for plugins: lstat ./plugins: no such file or directory
```


szuecs:
in favour of #599

PR #607, opened 2018-04-03
Add basic auth

### LINKED ISSUE #487: ETCD data client doesn't support basic authentication
https://coreos.com/etcd/docs/latest/v2/auth_api.html#v2-auth

MALPI:
Fixes #487 

Added basic auth as additional mechanism to authenticate to etcd client. @szuecs What do you think about the prioritization vs OAuth?

### DIFF
--- a/etcd/etcd.go
+++ b/etcd/etcd.go
@@ -31,8 +31,8 @@ package etcd
 import (
 	"bytes"
 	"crypto/tls"
-	"encoding/json"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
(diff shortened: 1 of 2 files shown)


szuecs:
@MALPI if you want to do more skipepr PRs, then we can add you to the repository.

In case of prio: Let people running skipper the choice and it's fine. 

It would be great to have a test for this, code looks good to me.
In case it's too complicated to test, I think we can also merge it like this.

hjacobs:
:+1:

aryszka:
:+1:

PR #609, opened 2018-04-09
document inlineContent filter in ingress-usage

szuecs:
with inlineContent you can return arbitrary content and with <shunt> you can respond early

### DIFF
--- a/docs/kubernetes/ingress-usage.md
+++ b/docs/kubernetes/ingress-usage.md
@@ -240,6 +240,26 @@ a unique value. Read more about this in our
 
      flowId("reuse")
 
+# Filters - return fast
+
+Sometimes you just want to return a header, redirect or even static
+html content. You can return from skipper without doing a proxy call
+to a backend, if you end your filter chain with `<shunt>`.
+
+## Return static content
+
+To set a response header `X: bar`, return
+`<html><body>hello</body></html>` in the HTTP response body and return
+HTTP status code 200 from ingress itself, you can use the following
+filter chain:
+
+    zalando.org/skipper-filter: |
+      setResponseHeader("X", "bar") -> inlineContent("<html><body>hello</body></html>") -> status(200) -> <shunt>
+
+Keep in mind you need a valid backend definition to backends which are
+available, otherwise Skipper would not populate the route for safety
+reasons.
+
 # Filters - reliability features
 
 Filters can modify http requests and responses. There are plenty of
--- a/docs/kubernetes/ingress-usage.md
+++ b/docs/kubernetes/ingress-usage.md
@@ -248,17 +248,16 @@ to a backend, if you end your filter chain with `<shunt>`.
 
 ## Return static content
 
-To set a response header `X: bar`, return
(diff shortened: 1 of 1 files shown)


docs/kubernetes/ingress-usage.md
```diff
@@ -240,6 +240,26 @@ a unique value. Read more about this in our
...
 
+# Filters - return fast
+
+Sometimes you just want to return a header, redirect or even static
+html content. You can return from skipper without doing a proxy call
+to a backend, if you end your filter chain with `<shunt>`.
+
+## Return static content
+
+To set a response header `X: bar`, return
```
lmineiro: I'd prefer is this sentence started more like "The following example sets a response..."

docs/kubernetes/ingress-usage.md
```diff
@@ -240,6 +240,26 @@ a unique value. Read more about this in our
...
+`<html><body>hello</body></html>` in the HTTP response body and return
+HTTP status code 200 from ingress itself, you can use the following
```
lmineiro: "from the ingress itself". This doesn't seem to be specific to skipper's usage as an ingress. Should we rather promote a general documentation entry for the features that work as well when used as a simple proxy and reference them from the Kubernetes ingress specific entries?

docs/kubernetes/ingress-usage.md
```diff
@@ -240,6 +240,26 @@ a unique value. Read more about this in our
...
+filter chain:
+
+    zalando.org/skipper-filter: |
+      setResponseHeader("X", "bar") -> inlineContent("<html><body>hello</body></html>") -> status(200) -> <shunt>
+
+Keep in mind you need a valid backend definition to backends which are
+available, otherwise Skipper would not populate the route for safety
+reasons.
```
lmineiro: I'd simplify the statement to:
"Keep in mind that you need a valid backend definition otherwise Skipper will not accept the entire route definition"

Is it a "safety"/security concern or just plain syntax failure?

szuecs: 1. It's not security, but safety
2. if you have no backend this might be a syntax error
3. if you have a backend definition without having a kubernetes service for that, skipper will not create a route for safety reasons.

docs/kubernetes/ingress-usage.md
szuecs: @lmineiro I don't think it makes sense, because it is right now mostly used as ingress.
I am open to get PRs that refactor the docs, but I do not think it is a good way to split these. In the end we have godoc for all and basically only a few documentation for other dataclients. If you like to provide more docs, feel free to provide them. I don't agree spreading the docs for users makes sense.

szuecs:
@lmineiro thanks for your comments, I changed it as suggested, but the last I changed to make more sense to me. I hope it's ok for you.

aryszka:
:+1:

ruiaraujo:
:+1:

hjacobs:
@szuecs thanks!

PR #615, opened 2018-04-12
do not leak decision header to backends

szuecs:
The decision header will disappear in the future, but we should not leak it to backends


aryszka:
:+1:

szuecs:
👍

Issue #614, opened 2018-04-12
kubernetes uneven loadbalanced members


szuecs:
I assume a proper fix would be to implement https://github.com/zalando/skipper/issues/556.
I do not fully understand the problem, yet but we see in 3 ingress uneven balanced distribution of backend calls.

We tried to:

- [X] restart skipper instances which did not help to spread again
- [X] checked with tcpdump and saw that backend calls with X-lb-member was passed to backends
 

As a workaround you can use an older version, that does not have this issue: https://github.com/zalando/skipper/releases/tag/v0.9.152.

aryszka:
related #608

Issue #608, opened 2018-04-04
loadbalancer does not balance load evenly


tkrop:
We have a strong evidence, that the Skipper's loadbalancer does not distribute load evenly between endpoints. Our setup contains two different paths of a single service provided by two service instances. Path (a) is processing a very low number of long running requests, while path (b) is serving a huge number of short requests. When we look into the logs, we can see that requests are not evenly distributed between our two service instances. We can see the following numbers for the last two days: 

| Paths    |    a + b    |    a    |
|------------|--------------|---------|
| Node A | 2,579,111 |   126 |
| Node B | 2,412,315 | 2246 |

While this may look acceptable from request perspective, it is unacceptable from processor load perspective because node A is serving 95% of the requests that are responsible for 75% of the overall service load.

szuecs:
I checked it with @tkrop in this specific situation and the caller targets only one skipper instance. In front of skipper we run ALBs with targetGroup skipper, so the caller has an open connection to skipper, I expect that it uses something like h2 which would transparently stream to one endpoint in case you only open one connection this will be most probably the reason for this issue.
We can't do anything, besides having a filter that downgrades connections to http/1.1 which also could be done by the backend service.
If you want to create a feature request to be able to downgrade connections, then please open a new issue.

tkrop:
@szuecs Not sure that this is the way to go. When I look at the behavior of the ALB it seems to not fall in such a trap. The ALB seems to forward request randomly to one of the available endpoints on skipper. May be Skipper should act similar.

szuecs:
@tkrop This I don't understand, because we looked both into it with tcpdump and watched the requests coming into one instance only. Let's discuss it in person again and check the nodes.

PR #616, opened 2018-04-12
add a concurrent test case with multple routes

szuecs:
this should show the bug mentioned in #614 and #608.


szuecs:
@aryszka sadly not, the diff for 686 calls is at most 1, which is expected and in our spec

szuecs:
👍

szuecs:
It complements tests, but it does not show the uneven balance from the mentioned issues

aryszka:
:+1:

PR #618, opened 2018-04-14
Fix/unnecessary healthcheck updates

aryszka:
- preserve health state
- in LoadUpdate, provide a healthcheck route only when it has changed


szuecs:
👍

szuecs:
This will partially fix #614 and #608.

aryszka:
:+1:

PR #620, opened 2018-04-16
Fix balancer route update

vecmezoni:
* Added failing test with the meaninglessly updating routing;
* Starting round robin with a random member, not with the first one;

Fixes https://github.com/zalando/skipper/issues/614


aryszka:
:+1:

PR #636, opened 2018-04-28
Document how to deploy with enabled RBAC

baez90:
Added sample manifests for:

* ServiceAccount
* ClusterRole
* ClusterRoleBinding
* adopted DaemonSet

see issue in [kube-ingress-aws-controller](https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/153)

### DIFF
--- a/docs/kubernetes/ingress-controller.md
+++ b/docs/kubernetes/ingress-controller.md
@@ -309,7 +309,7 @@ ClusterRole:
     apiVersion: rbac.authorization.k8s.io/v1beta1
     kind: ClusterRole
     metadata:
-      name: aws-ingress-controller
+      name: skipper-ingress 
     rules:
     - apiGroups: ["extensions"]
       resources: ["ingresses", ]
@@ -332,14 +332,14 @@ ClusterRoleBinding:
     apiVersion: rbac.authorization.k8s.io/v1beta1
     kind: ClusterRoleBinding
     metadata:
-      name: aws-ingress-controller
+      name: skipper-ingress 
     roleRef:
       apiGroup: rbac.authorization.k8s.io
       kind: ClusterRole
-      name: aws-ingress-controller
+      name: skipper-ingress 
     subjects:
     - kind: ServiceAccount
-      name: aws-ingress-controller
+      name: skipper-ingress 
       namespace: kube-system
 
 Last but not least the `ServiceAccount` has to be assigned to the Skipper daemonset.
--- a/docs/kubernetes/ingress-controller.md
+++ b/docs/kubernetes/ingress-controller.md
@@ -194,7 +194,7 @@ We deploy a service type ClusterIP that we will select from ingress:
     apiVersion: v1
     kind: Service
     metadata:
-      name: sszuecs-demo
+      name: skipper-demo
       labels:
         application: skipper-demo
     spec:
@@ -205,7 +205,7 @@ We deploy a service type ClusterIP that we will select from ingress:
           targetPort: 9090
           name: external
       selector:
-        application: sszuecs-demo
+        application: skipper-demo
 
 To deploy both, you have to run:
 
--- a/docs/kubernetes/ingress-controller.md
+++ b/docs/kubernetes/ingress-controller.md
@@ -300,7 +300,7 @@ First create a new `ServiceAccount` which will be assigned to the Skipper pods:
     apiVersion: v1
     kind: ServiceAccount
     metadata:
-      name: skipper-ingress 
+      name: skipper-ingress
       namespace: kube-system
 
 the required permissions are defined within a `ClusterRole` resource.
@@ -312,20 +312,16 @@ ClusterRole:
     apiVersion: rbac.authorization.k8s.io/v1beta1
     kind: ClusterRole
     metadata:
-      name: skipper-ingress 
+      name: skipper-ingress
     rules:
     - apiGroups: ["extensions"]
       resources: ["ingresses", ]
       verbs: ["get", "list"]
-    - apiGroups: ["extensions"]
-      resources: ["ingresses/status", ]
-      verbs: ["get", "list", "patch"]
     - apiGroups: [""]
       resources: ["namespaces", "services", "endpoints"]
       verbs: ["get", "list"]
 
 This `ClusterRole` defines access to `get` and `list` all created ingresses, namespaces, services and endpoints.
-Furthermore it enables Skipper to `patch` the state of an ingress (like updating the endpoint of an ingress).
 
 To assign the defined `ClusterRole` to the previously created `ServiceAccount`
 a `ClusterRoleBinding` has to be created:
@@ -335,14 +331,14 @@ ClusterRoleBinding:
     apiVersion: rbac.authorization.k8s.io/v1beta1
     kind: ClusterRoleBinding
     metadata:
-      name: skipper-ingress 
+      name: skipper-ingress
     roleRef:
       apiGroup: rbac.authorization.k8s.io
       kind: ClusterRole
-      name: skipper-ingress 
+      name: skipper-ingress
     subjects:
     - kind: ServiceAccount
-      name: skipper-ingress 
+      name: skipper-ingress
       namespace: kube-system
 
 Last but not least the `ServiceAccount` has to be assigned to the Skipper daemonset.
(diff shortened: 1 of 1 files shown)


hjacobs:
Thanks for the PR!

docs/kubernetes/ingress-controller.md
```diff
@@ -281,3 +281,119 @@ Example ingress:
...
+              - "-kubernetes"
+              - "-kubernetes-in-cluster"
+              - "-address=:9999"
+              - "-proxy-preserve-host"
+              - "-serve-host-metrics"
+              - "-enable-ratelimits"
+              - "-experimental-upgrade"
+              - "-metrics-exp-decay-sample"
+            resources:
+              limits:
```
szuecs: I had a discussion with @aermakov-zalando on setting only limits or only requests. I think we agreed on having only requests, because it will remove the CFS settings at cgroup level, such that no throttling will happen.
Do I remember this correctly @aermakov-zalando or did we agreed on having only limits?

docs/kubernetes/ingress-controller.md
```diff
@@ -281,3 +281,119 @@ Example ingress:
...
+Kubernetes cluster with RBAC enabled but it's highly recommended to read the
+[RBAC docs](https://kubernetes.io/docs/admin/authorization/rbac/) to get a better
+understanding which permissions are delegated to Skipper within your Kubernetes cluster.
+
+First create a new `ServiceAccount` which will be assigned to the Skipper pods:
+
+    apiVersion: v1
+    kind: ServiceAccount
+    metadata:
+      name:  aws-ingress-controller
```
szuecs: Why this is called aws-ingress-controller?
Skipper can perfectly run in any cluster type.

docs/kubernetes/ingress-controller.md
```diff
@@ -281,3 +281,119 @@ Example ingress:
...
+    kind: ClusterRole
+    metadata:
+      name: aws-ingress-controller
+    rules:
+    - apiGroups: ["extensions"]
+      resources: ["ingresses", ]
+      verbs: ["get", "list"]
+    - apiGroups: ["extensions"]
+      resources: ["ingresses/status", ]
+      verbs: ["get", "list", "patch"]
```
szuecs: Are you sure we need "patch", I think we don't need "patch". Skipper only reads, and let for example the kube-aws-ingress-controller do the update of the status field.

docs/kubernetes/ingress-controller.md
```diff
@@ -281,3 +281,119 @@ Example ingress:
...
+    - apiGroups: [""]
+      resources: ["namespaces", "services", "endpoints"]
+      verbs: ["get", "list"]
+
+This `ClusterRole` defines access to `get` and `list` all created ingresses, namespaces, services and endpoints.
+Furthermore it enables Skipper to `patch` the state of an ingress (like updating the endpoint of an ingress).
```
szuecs: Are you sure? I think patching status is done by the kube-aws-ingress-controller.

docs/kubernetes/ingress-controller.md
```diff
@@ -281,3 +281,119 @@ Example ingress:
...
+
+To assign the defined `ClusterRole` to the previously created `ServiceAccount`
+a `ClusterRoleBinding` has to be created:
+
+ClusterRoleBinding:
+
+    apiVersion: rbac.authorization.k8s.io/v1beta1
+    kind: ClusterRoleBinding
+    metadata:
+      name: aws-ingress-controller
```
szuecs: naming maybe skipper-ingress-controller ?

docs/kubernetes/ingress-controller.md
```diff
@@ -281,3 +281,119 @@ Example ingress:
...
+            args:
+              - "skipper"
+              - "-kubernetes"
+              - "-kubernetes-in-cluster"
+              - "-address=:9999"
+              - "-proxy-preserve-host"
+              - "-serve-host-metrics"
+              - "-enable-ratelimits"
+              - "-experimental-upgrade"
+              - "-metrics-exp-decay-sample"
```
szuecs: Add these flags please:

          - "-lb-healthcheck-interval=3s"
          - "-metrics-flavour=codahale,prometheus"
          - "-enable-connection-metrics"

baez90:
I haven't thought of that aspect. I'll adopt the requested changes when I'm back from vacations.

I will also double check whether Skipper needs the `patch` access right. I'm also not sure if all `list` permissions are required, I'll check that, too.

Thanks for the great and immediate feedback!

docs/kubernetes/ingress-controller.md
baez90: Do you really want to enable the metrics API by default?

I just copied the flags of the deployment sample in the same file, shall I adjust the other yaml too?

szuecs: Yes please do.
I think it makes a lot of sense for production and also show features to pitch the project.

docs/kubernetes/ingress-controller.md
baez90: Renamed to `skipper-ingress`

docs/kubernetes/ingress-controller.md
baez90: Renamed to `skipper-ingress`, I thought it should be clear enough.

docs/kubernetes/ingress-controller.md
baez90: I added the flags and updated the original DaemonSet where I copied the flags from.

baez90:
I renamed the resources as requested but I still have to validate the required permissions of Skipper without `kube-aws-ingress-controller` (`patch` and `list`).

I hope I'll be able to do that in the beginning of the next week.

Plase let me know if you agree with the renames or if I forgot something.

docs/kubernetes/ingress-controller.md
aermakov-zalando: For the ingress controller I think it's fine to set both because latency doesn't really affect it.

docs/kubernetes/ingress-controller.md
szuecs: I think we also do not need ingresses/status api group.

szuecs:
@baez90 I am fine with your changes.
You can also just delete the "patch", because we do not have any "write" call to the kube-apiserver. We only get all ingress, svc, endpoints. We only use plain http without using client-go for these calls, so there is no magic involved.
So in my opinion you do not need to test it and I am fine with merging like this.

baez90:
I removed the permissions for `ingresses/status` and `patch` as mentioned.

I'll validate if the permissions are sufficient as soon as possible but I don't expect that to happen before the upcoming Monday. It's up to you if you'd like to wait till then or if you merge it right now and I'll open another PR if something isn't working as expected.

szuecs:
:+1:

docs/kubernetes/ingress-controller.md
baez90: Removed `ingresses/status` I think it's only required for kube-ingress-aws-controller

docs/kubernetes/ingress-controller.md
baez90: Yeah I think you're right.
I'll take that into account for the Helm chart.

Do you think it's better to have 2 service accounts (one for Skipper and one for the kube-ingress-aws-controller)?

szuecs:
:+1:

baez90:
I think that's it.
I validated the shown set of permissions with Minikube to ensure, that Skipper is still running correctly.
I could add some explanations why which permission is needed but I don't expect a beginner to be curious why it's `get` and `list` instead of just `get`.
If no one requests other changes I'm fine with merging :wink:

szuecs:
:+1:

aryszka:
:+1:

so0k:
Thanks a lot for getting this in!

baez90:
To be honest, I haven't thought of that.

Right now I'm stabilizing the chart to be sure that it's flexible enough, fulfills all requirements and follows at least most of the best practices and then I might PR to kubernetes/charts repo.

so0k:
I'm bootstrapping it using kops channels - I really need to do a blogpost about it

this is suitable to bootstrap tiller, basic rbac and also deployments that do not have upstream charts

PR #637, opened 2018-04-30
unverifiedAuditLog - fix #631 

### LINKED ISSUE #631: Extract and log "sub" claim from JWT token from Authorization header
To be discussed: In order to have authorization information in the Skipper access log, we could consider extracting the `sub` claim from the JWT access token passed in `Authorization` header (OAuth Bearer token).

szuecs:
implement #631 unverifiedAuditLog logs "sub" from 2nd part of decoded base64 jwt token if possible.

@hjacobs this is what you meant? Any idea for the prefix? I just did "Audit sub: " for now but I guess we want something else.

    Audit sub: c4ddfe9d-a0d3-4afb-bf26-24b9588731a0 ::1 - - [30/Apr/2018:09:39:59 +0200] "GET /ffff09 HTTP/1.1" 200 0 "" "curl/7.49.0" 0 localhost:9999
    Audit sub: c4ddfe9d-a0d3-4afb-bf26-24b9588731a0 ::1 - - [30/Apr/2018:09:39:59 +0200] "GET /ffff10 HTTP/1.1" 200 0 "" "curl/7.49.0" 2 localhost:9999

### DIFF
--- a/filters/log/log.go
+++ b/filters/log/log.go
@@ -32,9 +32,10 @@ const (
 	// UnverifiedAuditLogName is th filtername seend by the user
 	UnverifiedAuditLogName = "unverifiedAuditLog"
 
-	authHeaderName      = "Authorization"
-	authHeaderPrefix    = "Bearer "
-	accessTokenQueryKey = "access_token"
+	UnverifiedAuditHeader = "X-Unverified-Audit"
+	authHeaderName        = "Authorization"
+	authHeaderPrefix      = "Bearer "
+	accessTokenQueryKey   = "access_token"
 )
 
 type auditLog struct {
@@ -166,12 +167,10 @@ func (al *auditLog) Response(ctx filters.FilterContext) {
 	}
 }
 
-type unverifiedAuditLog struct {
-	writer io.Writer
-}
+type unverifiedAuditLog struct{}
 
 // NewUnverifiedAuditLog logs "Sub" of the middle part of a JWT Token.
-func NewUnverifiedAuditLog() filters.Spec { return &unverifiedAuditLog{writer: os.Stderr} }
+func NewUnverifiedAuditLog() filters.Spec { return &unverifiedAuditLog{} }
 
 func (ual *unverifiedAuditLog) Name() string { return UnverifiedAuditLogName }
 
@@ -209,7 +208,7 @@ func (ual *unverifiedAuditLog) Request(ctx filters.FilterContext) {
 		if err != nil {
 			return
 		}
-		ual.writer.Write([]byte("Audit sub: " + j.Sub + " "))
+		req.Header.Add(UnverifiedAuditHeader, j.Sub)
 	}
 }
 
--- a/filters/log/log.go
+++ b/filters/log/log.go
@@ -7,9 +7,11 @@ package log
 
 import (
 	"bytes"
+	"encoding/base64"
 	"encoding/json"
 	"io"
 	"os"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 
@@ -27,6 +29,12 @@ const (
 	// reject reason information into the state bag to pass the
 	// information to the auditLog filter.
 	AuthRejectReasonKey = "auth-reject-reason"
+	// UnverifiedAuditLogName is th filtername seend by the user
+	UnverifiedAuditLogName = "unverifiedAuditLog"
+
+	authHeaderName      = "Authorization"
+	authHeaderPrefix    = "Bearer "
+	accessTokenQueryKey = "access_token"
 )
 
 type auditLog struct {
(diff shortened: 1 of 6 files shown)


hjacobs:
Yes, that's what I meant --- I just realize that our Zalando Platform IAM contains somer user UUID which needs to be mapped to the username when reading the audit log (not something Skipper has to do).

filters/log/log.go
```diff
@@ -157,3 +165,52 @@ func (al *auditLog) Response(ctx filters.FilterContext) {
 		log.Errorf("Failed to json encode auditDoc: %v", err)
 	}
 }
+
+type unverifiedAuditLog struct {
+	writer io.Writer
+}
+
+// NewUnverifiedAuditLog logs "Sub" of the middle part of a JWT Token.
+func NewUnverifiedAuditLog() filters.Spec { return &unverifiedAuditLog{writer: os.Stderr} }
```
aryszka: it would be nicer if the writer was a parameter, and the caller side could decide what to pass in

filters/log/log.go
```diff
@@ -157,3 +165,52 @@ func (al *auditLog) Response(ctx filters.FilterContext) {
...
+		if err != nil {
+			return
+		}
+
+		var j jwtMidPart
+		err = json.Unmarshal(sDec, &j)
+		if err != nil {
+			return
+		}
+		ual.writer.Write([]byte("Audit sub: " + j.Sub + " "))
```
aryszka: alternatively:

```
fmt.Fprintf(ual.writer, "Audit sub: %s ", j.Sub)
```

filters/log/log.go
```diff
@@ -27,6 +29,12 @@ const (
 	// reject reason information into the state bag to pass the
 	// information to the auditLog filter.
 	AuthRejectReasonKey = "auth-reject-reason"
+	// UnverifiedAuditLogName is th filtername seend by the user
+	UnverifiedAuditLogName = "unverifiedAuditLog"
```
aryszka: not sure about this name. Maybe "subClaim" or "oauth2SubClaim"?

filters/log/log.go
aryszka: not sure how this works without the newline. I think we need a new line at the end, because even if we rely on the some other following request related log output, there's no guarantee that from the many concurrent requests not something else will print something first.

filters/log/log.go
szuecs: dropped writer

filters/log/log.go
szuecs: No, because it's not getting a claim. It's also not a verified audit log, just a shorthand base64 decode of the middle part of a jwt token to make it possible to find the correlate logs. It's not meant for security auditlogging.
These are the reasons why I don't want to write down miss leading information to users.

filters/log/log.go
szuecs: I dropped the write, thanks for your hints in the chat!

filters/log/log.go
```diff
@@ -27,6 +29,13 @@ const (
...
+
+	UnverifiedAuditHeader = "X-Unverified-Audit"
```
szuecs: @aryszka what do you think about this idea?
Passing this information by header to the accessLog.

szuecs:
With the last change the accessLog is modified to append this unverifiedAuditLog, like this:

    ::1 - - [30/Apr/2018:19:29:39 +0200] "GET /foo HTTP/1.1" 200 0 "" "curl/7.49.0" 0 localhost:9999  c4ddfe9d-a0d3-4afb-bf26-24b9588731a0

The double whitespace is, because of the missing FlowID. I consider this as correct, because if we have a single whitespace you can not decide if this is unverifiedAuditLog or flowID.

filters/log/log.go
aryszka: from the two alternatives that we discussed today (globally enabling this access log field or allowing it to be controlled in the routes), i find this approach better on the long term.

I still have the concern about the naming, but right now I don't have a better name, either. The reason why I don't like the naming is that it doesn't call the thing what it is but it refers to a concept that it can be used for ("using the sub claim from the auth header for auditing" instead of just "printing the sub claim from the auth header"). Let's park this for a day, maybe someone can give it a good name. @hjacobs do you have a suggestion for a good name for this?

logging/access.go
```diff
@@ -17,8 +18,8 @@ const (
 	// format:
 	// remote_host - - [date] "method uri protocol" status response_size "referer" "user_agent"
 	combinedLogFormat = commonLogFormat + ` "%s" "%s"`
-	// We add the duration in ms, a requested host and a flow id
-	accessLogFormat = combinedLogFormat + " %d %s %s\n"
+	// We add the duration in ms, a requested host and a flow id and audit log
+	accessLogFormat = combinedLogFormat + " %d %s %s %s\n"
```
aryszka: if we do it so, then there will be a trailing space in the access log of those deployments that don't use this field. there's a chance that some already used parsers may consider this extra trailing space as part of the flow id. I would play it safe and use the new format only when there is a 'X-Unverified-Audit' header found.

logging/access_test.go
```diff
@@ -5,10 +5,12 @@ import (
 	"net/http"
 	"testing"
 	"time"
+
+	logFilter "github.com/zalando/skipper/filters/log"
 )
 
 const logOutput = `127.0.0.1 - - [10/Oct/2000:13:55:36 -0700] "GET /apache_pb.gif HTTP/1.1" 418 2326 "" "" 42 example.com `
-const logJSONOutput = `{"duration":42,"flow-id":"","host":"127.0.0.1","level":"info","method":"GET","msg":"","proto":"HTTP/1.1","referer":"","requested-host":"example.com","response-size":2326,"status":418,"timestamp":"10/Oct/2000:13:55:36 -0700","uri":"/apache_pb.gif","user-agent":""}`
+const logJSONOutput = `{"audit":"-","duration":42,"flow-id":"-","host":"127.0.0.1","level":"info","method":"GET","msg":"","proto":"HTTP/1.1","referer":"","requested-host":"example.com","response-size":2326,"status":418,"timestamp":"10/Oct/2000:13:55:36 -0700","uri":"/apache_pb.gif","user-agent":""}`
```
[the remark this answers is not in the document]
szuecs: I changed it to be empty string in case of json and empty string in text formatter are replaced by "-".

aryszka:
:+1:

logging/access.go
aermakov-zalando: This will cause the log format to flip based on whether the request is authenticated or not, I'm not sure if it's a good idea. If keeping backwards compatibility is a must, then the only thing that can be done is a command-line argument, otherwise I'd just flip the order of flow id and sub claim and hope that most of the parsers can deal with an extra value. This will of course require missing flow ids to be represented as "-" instead of just a space, but it's probably a good idea anyway.

szuecs: @aermakov-zalando see the output of https://github.com/zalando/skipper/pull/637#issuecomment-385473261 all whitespaced fields will be replaced by "-" and we agreed on having this change a fix, which was overseen and a bug before.

szuecs:
:+1:

logging/access.go
aermakov-zalando: Then just swapping it should be enough, I guess.

Issue #630, opened 2018-04-23
support openidc with JWT claim support 


szuecs:
## get config

The OIDC config is located at `<issuer>/.well-known/openid-configuration`.

For example
- authorization_endpoint: oauth2.Config.Endpoint.AuthURL
- token_endpoint: oauth2.Config.Endpoint.TokenURL

One **problem** is that `/.well-known/openid-configuration` , is not mandatory returning the introspection_endpoint by https://tools.ietf.org/html/draft-ietf-oauth-discovery-06 (marked as OPTIONAL ).

Example call to introspection_endpoint, it has to support POST by https://tools.ietf.org/html/rfc7662

    % curl -X POST -d "token=$(ztoken)" localhost:9021/oauth2/introspect

### Google

    { 
      "issuer": "https://accounts.google.com",
      "authorization_endpoint": "https://accounts.google.com/o/oauth2/v2/auth",
      "token_endpoint": "https://www.googleapis.com/oauth2/v4/token",
      "userinfo_endpoint": "https://www.googleapis.com/oauth2/v3/userinfo",
      "revocation_endpoint": "https://accounts.google.com/o/oauth2/revoke",
      "jwks_uri": "https://www.googleapis.com/oauth2/v3/certs",
      "response_types_supported": [
        "code",
        "token",
        "id_token",
        "code token",
        "code id_token",
        "token id_token",
        "code token id_token",
        "none"
      ],
      "subject_types_supported": [
        "public"
      ],
      "id_token_signing_alg_values_supported": [
        "RS256"
      ],
      "scopes_supported": [
        "openid",
        "email",
        "profile"
      ],
      "token_endpoint_auth_methods_supported": [
        "client_secret_post",
        "client_secret_basic"
      ],
      "claims_supported": [
        "aud",
        "email",
        "email_verified",
        "exp",
        "family_name",
        "given_name",
        "iat",
        "iss",
        "locale",
        "name",
        "picture",
        "sub"
      ],
      "code_challenge_methods_supported": [
        "plain",
        "S256"
      ]
    }

### Zalando

    {
      "issuer": "https://identity.zalando.com",
      "authorization_endpoint": "https://identity.zalando.com/oauth2/authorize",
      "token_endpoint": "https://identity.zalando.com/oauth2/token",
      "revocation_endpoint": "https://identity.zalando.com/oauth2/revoke",
      "jwks_uri": "https://identity.zalando.com/.well-known/jwk_uris",
      "response_types_supported": [
        "token",
[...]

PR #638, opened 2018-05-02
Feature/oauth2 token introspection RFC7662

szuecs:
This PR implements oauth2 token introspection as defined in RFC7662 including auto configuration based on  https://tools.ietf.org/html/draft-ietf-oauth-discovery-06
<s>Decided to also include https://github.com/zalando/skipper/issues/630</s>

- [x] tokenintrospection svc-to-svc similar to tokeninfo
- [x] <s>UI workflow authCode
  - [x]  full handshake flow works
  - [x] a working "session" afterwards</s>
- [x] refactoring
- [x] cleanups and TODOs
- [x] tests
- [x] docs


filters/auth/oauth.go
```diff
@@ -432,4 +585,238 @@ func (f *filter) Request(ctx filters.FilterContext) {
...
+	return newOAuthTokenintrospectionFilter(checkOAuthTokenintrospectionAnyClaims, oauthIssuerURL, oauthIntrospectionURL)
+}
+func NewOAuthTokenintrospectionAllClaims(oauthIssuerURL, oauthIntrospectionURL string) filters.Spec {
+	return newOAuthTokenintrospectionFilter(checkOAuthTokenintrospectionAllClaims, oauthIssuerURL, oauthIntrospectionURL)
+}
+
+func newOAuthTokenintrospectionFilter(typ roleCheckType, oauthIssuerURL, oauthIntrospectionURL string) filters.Spec {
+	cfg, err := getOpenIDConfig(oauthIssuerURL)
+	if err != nil {
+		return &tokenIntrospectionSpec{
```
aryszka: We should move the open-id-config file checking/reading to the calling code (the root package), and here only use the instrospection url from it. There, only the case when the file doesn't exist should be ignored, otherwise the startup should fail.

skipper.go
```diff
@@ -637,6 +651,13 @@ func Run(o Options) error {
 	}
 	o.CustomFilters = append(o.CustomFilters, logfilter.NewAuditLog(o.MaxAuditBody))
 
+	if o.OAuthIssuerURL != "" {
+		o.CustomFilters = append(o.CustomFilters, auth.NewOAuthTokenintrospectionAnyClaims(o.OAuthIssuerURL, o.OAuthTokenintrospectionURL))
```
aryszka: alt way:

```
o.CustomFilters = append(o.CustomFilters, 
    auth.NewOAuthTokenintrospectionAnyClaims(o.OAuthIssuerURL, o.OAuthTokenintrospectionURL),
    auth.NewOAuthTokenintrospectionAllClaims(o.OAuthIssuerURL, o.OAuthTokenintrospectionURL),
    auth.NewOAuthTokenintrospectionAnyKV(o.OAuthIssuerURL, o.OAuthTokenintrospectionURL),
    auth.NewOAuthTokenintrospectionAllKV(o.OAuthIssuerURL, o.OAuthTokenintrospectionURL),
)
```

filters/auth/oauth.go
```diff
@@ -183,10 +276,78 @@ func newAuthClient(baseURL string) (*authClient, error) {
 
 func (ac *authClient) getTokeninfo(token string) (map[string]interface{}, error) {
 	var a map[string]interface{}
-	err := jsonGet(ac.url, token, &a)
+	err := jsonGet(ac.url, token, a)
```
aryszka: we can't remove the `&` here

filters/auth/oauth.go
```diff
@@ -173,6 +235,37 @@ func jsonGet(url *url.URL, auth string, doc interface{}) error {
 	return d.Decode(doc)
 }
 
+// jsonPost requests url with access token in the body, if auth was given and writes into doc.
+func jsonPost(u *url.URL, auth string, doc *tokenIntrospectionInfo) error {
+	if auth == "" {
+		return fmt.Errorf("invalid without auth")
```
aryszka: this error currently doesn't happen because the token is checked before. If we want to keep this error here, then a better message would be required. E.g: fmt.Errorf("missing auth token for %v", u)

filters/auth/oauth.go
```diff
@@ -173,6 +235,37 @@ func jsonGet(url *url.URL, auth string, doc interface{}) error {
...
+	if err != nil {
+		return err
+	}
+
+	defer rsp.Body.Close()
+	if rsp.StatusCode != 200 {
+		return errInvalidToken
+	}
+	buf := make([]byte, rsp.ContentLength)
+	n, _ := rsp.Body.Read(buf)
```
aryszka: Why not just returning the error from rsp.Body.Read(buf) if there is any, instead of the content-length checking?

filters/auth/oauth.go
```diff
@@ -183,10 +276,78 @@ func newAuthClient(baseURL string) (*authClient, error) {
...
+func (tii tokenIntrospectionInfo) getStringValue(k string) (string, error) {
+	s, ok := tii[k].(string)
+	if !ok {
+		return "", errInvalidTokenintrospectionData
+	}
+	return s, nil
+}
+
+func (tii tokenIntrospectionInfo) getUNIXTimeValue(k string) (time.Time, error) {
+	ts, ok := tii[k].(string)
```
aryszka: I think the json parser will parse this number as a float64 and not a string.

filters/auth/oauth.go
```diff
@@ -432,4 +585,238 @@ func (f *filter) Request(ctx filters.FilterContext) {
...
+	return &tokenIntrospectionSpec{
+		typ:              typ,
+		issuerURL:        oauthIssuerURL,
+		introspectionURL: cfg.IntrospectionEndpoint,
+		config:           cfg,
+	}
+}
+
+func getOpenIDConfig(issuerURL string) (*openIDConfig, error) {
+	u, err := url.Parse(issuerURL + "/.well-known/openid-configuration")
```
aryszka: a constant above is defined for this path: `tokenIntrospectionConfigPath`

filters/auth/oauth.go
aryszka: update: actually, if the calling code specifies the `oauthIntrospectionURL` explicitly, then we probably should call the getOpenIDConfig at all. What do you think?

filters/auth/oauth.go
```diff
@@ -234,13 +395,13 @@ func (s *tokeninfoSpec) Name() string {
...
 // CreateFilter creates an auth filter. All arguments have to be
-// strings. Depending on the variant of the auth filter, the arguments
+// strings. Depending on the variant of the auth tokeninfoFilter, the arguments
 // represent scopes or key-value pairs to be checked in the tokeninfo
 // response. How scopes or key value pairs are checked is based on the
 // type. The shown example for checkOAuthTokeninfoAllScopes will grant
 // access only to tokens, that have scopes read-x and write-y:
 //
-//     s.CreateFilter(read-x", "write-y")
+//     s.CreateTokeninfoFilter(read-x", "write-y")
```
aryszka: this doc entry should stay s.CreateFilter

szuecs: oh this was a bigger replace Filter with TokeninfoFilter ...

filters/auth/oauth.go
szuecs: see below at the dispatch of f.typ to checkOAuthTokenintrospectionAnyClaims, it's checking s.config.ClaimsSupported, which is only got from the open-id-config.

On the other hand @vetinari asked me why not to use https://github.com/coreos/go-oidc ?
I see some problems with that library, but I want to try it before I do not consider it. It can more easily provide some nice flow integrations, not sure if this is easily possible in our case, but let's see.

aryszka: i see. let's try it.

i would still try to find a way to return the error during the startup. I see that the filter spec creation in the builtin.MakeRegistry() doesn't allow returning errors, which I guess was the main problem. Maybe, if getOpenIDConfig or the go-oidc sutff is called from outside the NewOAuth... functions, in the root package, and the available options are passed to the NewOAuth... functions with the available info, including the open id config, it is possible ensure failing early.

szuecs: Yes exactly, maybe this makes sense. I am also not sure if we want to poll for updates of this configuration from time to time. Right now I don't know and maybe this means pulling it into the skipper bootstrapping process makes most sense than overcomplicate this.

szuecs:
@aryszka this is now ready to review. I removed the OIDC UI flow to make this PR faster to release and less lines to review.

docs/filters.md
```diff
@@ -448,6 +448,74 @@ Examples:
 oauthTokeninfoAllKV("k1", "v1", "k2", "v2")
 ```
 
+## oauthTokenintrospectionAnyClaims
+
+If skipper is started with `-oauth2-issuer-url` flag, you can use this
+filter.
+
+The filter accepts variable arguments of strings, which are used to
```
aryszka: "...variable arguments of strings, ..." might be better as "...variable number of string arguments, ..."

filters/auth/auth.go
```diff
@@ -0,0 +1,141 @@
...
+func (kv kv) String() string {
+	var res []string
+	for k, v := range kv {
+		res = append(res, k, v)
+	}
+	return strings.Join(res, ",")
+}
+
+func getToken(r *http.Request) (string, error) {
+	if tok := r.FormValue(accessTokenKey); tok != "" {
```
aryszka: i don't think we can use FormValue in a filter, because it consumes the request body, which in most cases should be forwarded to the backend.

filters/auth/authclient.go
```diff
@@ -0,0 +1,119 @@
...
+	defer rsp.Body.Close()
+	if rsp.StatusCode != 200 {
+		return errInvalidToken
+	}
+	buf := make([]byte, rsp.ContentLength)
+	_, err = rsp.Body.Read(buf)
+	if err != nil && err != io.EOF {
+		return err
+	}
+	err = json.Unmarshal(buf, &doc)
```
aryszka: here we could write alternatively:

```
return json.Unmarshal(buf, &doc)
```

filters/auth/tokenintrospection.go
```diff
@@ -0,0 +1,342 @@
...
+)
+
+type (
+	tokenIntrospectionSpec struct {
+		typ              roleCheckType
+		issuerURL        string
+		introspectionURL string
+		config           *OpenIDConfig
+		timeout          time.Duration
+		authClient       *authClient
```
aryszka: is the `authClient` field of the filter spec used?

filters/auth/tokenintrospection.go
```diff
@@ -0,0 +1,342 @@
...
+func (s *tokenIntrospectionSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	sargs, err := getStrings(args)
+	if err != nil {
+		return nil, err
+	}
+	if len(sargs) == 0 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	ac, err := newAuthClient(s.introspectionURL, s.timeout)
```
aryszka: i think that the authClient should be created at the time when the filter spec is created, which means once on starting up skipper. If we create the authClient every time a new filter instance is created, there will be a new goroutine created in createHTTPClient and never be cleaned up.

filters/auth/tokenintrospection.go
```diff
@@ -0,0 +1,342 @@
...
+	f := &tokenintrospectFilter{
+		typ:        s.typ,
+		authClient: ac,
+		kv:         make(map[string]string),
+	}
+	switch f.typ {
+	case checkOAuthTokenintrospectionAllClaims:
+		fallthrough
+	case checkOAuthTokenintrospectionAnyClaims:
+		f.claims = sargs[:]
```
aryszka: wondering why this sargs[:] is required here, since we don't modify the sargs slice. Is there a reason?

filters/auth/tokenintrospection_test.go
```diff
@@ -0,0 +1,460 @@
...
+		args:        []interface{}{"wrongKey", "wrongValue", testKey, testValue},
+		hasAuth:     true,
+		auth:        testToken,
+		expected:    http.StatusUnauthorized,
+	}} {
+		t.Run(ti.msg, func(t *testing.T) {
+			if ti.msg == "" {
+				t.Fatalf("unknown ti: %+v", ti)
+			}
+			backend := httptest.NewServer(http.HandlerFunc(func(http.ResponseWriter, *http.Request) {}))
```
aryszka: the backend server should be closed.

```
defer backend.Close()
```

filters/auth/tokenintrospection_test.go
```diff
@@ -0,0 +1,460 @@
...
+			default:
+				t.Fatalf("FATAL: authType '%s' not supported", ti.authType)
+			}
+
+			args = append(args, ti.args...)
+			fr := make(filters.Registry)
+			fr.Register(spec)
+			r := &eskip.Route{Filters: []*eskip.Filter{{Name: spec.Name(), Args: args}}, Backend: backend.URL}
+
+			proxy := proxytest.New(fr, r)
```
aryszka: the proxy should be closed:

```
defer proxy.Close()
```

filters/auth/auth.go
szuecs: @aryszka It creates a separate request, so why should this matter?

filters/auth/tokenintrospection.go
szuecs: seems to be a leftover

filters/auth/tokenintrospection.go
szuecs: You are right. If I now think about, it might also makes sense to have multiple `issuer` allowed, so I could initialize atuhClient per issuer. The problem is how to have this accessible for the filter.

In the single issuer case, I can intialize the might failing part in skipper.go and put the client initialization into the spec. If we create a filter from the spec I would just copy the pointer.
In this way we would have for N filters of these type N times called close on the channel, which would panic.
Any idea how to resolve this easily?

filters/auth/auth.go
aryszka: i see it here that it uses the incoming request from the filter context: https://github.com/zalando/skipper/pull/638/files#diff-97a96747f71322fa840a549316da3a1bR278

aryszka: the reason why this is a problem is in the cases when someone makes a post request with a body that should be sent to the backend service, and puts the auth token in the authorization header. What would happen in this case is that the FormValue consumes the incoming request body, and then the proxy won't be able to send it to the backend.

szuecs:
I found the missing close also in tokeninfo_test and authServer was also not closed.

The N issuer M spec L filters problem with authclient and quit channel might need more discussion.
Let's discuss it in our public slack channel.

filters/auth/tokenintrospection.go
aryszka: i would put it into skipper.go, and pointer can be passed from the spec to the filter instance.

szuecs:
@aryszka as discussed the changes were done:

- issuer is now the first argument to the filter (multi authnz possible)
- authclient is created, but cached, in the filter, such that we have 1 authclient per issuer shared across all filters with the same issuer, but only created if you have such a filter.

filters/auth/tokenintrospection.go
```diff
@@ -333,9 +341,12 @@ func (f *tokenintrospectFilter) Request(ctx filters.FilterContext) {
 
 func (f *tokenintrospectFilter) Response(filters.FilterContext) {}
 
-// Close cleans-up the quit channel used for this spec
+// Close cleans-up the quit channel used for this filter
 func (f *tokenintrospectFilter) Close() {
+	f.authClient.mu.Lock()
```
aryszka: (not important, i think it is fair to expect that the Close is called only from one goroutine)

aryszka:
the FormValue is still consuming the request body. I think that should be fixed.

filters/auth/tokenintrospection.go
szuecs: better being safe in this case!
I think on teardown we have plenty of time, no problem for production traffic if we make sure we do not panic because of the Close()

aryszka: Accepted.

szuecs:
@aryszka I deleted the FormValue support at the incoming proxy side. I think we do not need to support it and only support `Authorization: Bearer <token>` in the header at the incomin proxy side.
FormValue is only used in the client call to the RFC7662 compliant auth server.

szuecs:
:+1:

szuecs:
@aryszka I think I fixed your comment. Do you agree and we can merge?

aryszka:
:+1:

szuecs:
:+1:

PR #639, opened 2018-05-03
fix prepared wrong garbage in unverified auditlog

szuecs:
To not log user specified random data this proofs that Sub claim contains "valid" string information and not random data chosen by the user sending the request.

@aermakov-zalando thanks for the hint!


szuecs:
:+1:

aryszka:
:+1:

Issue #654, opened 2018-05-27
Don't respond with status 500 for timeout errors


mikkeloscar:
See #652 for full context.

When skipper fails to route to a backend e.g. because the pod no longer exists as shown in #652 then it will result in the following error:

```
[APP]time=\"2018-05-25T14:59:56Z\" level=error msg=\"error during backend roundtrip: kube_default__cluster_registry__cluster_registry_stups_zalan_do____cluster_registry_0: context canceled\"
[APP]time=\"2018-05-25T14:59:56Z\" level=error msg=\"Failed to do backend request to http://10.2.108.118:8080: proxy error with code 0, dialing failed false, unwrap: context canceled\"
[APP]time=\"2018-05-25T14:59:56Z\" level=error msg=\"error while proxying, route kube_default__cluster_registry__cluster_registry_stups_zalan_do____cluster_registry__lb_group, status code 500: proxy error with code 0, dialing failed false, unwrap: context canceled\"
<client> - - [25/May/2018:14:59:46 +0000] \"GET /kubernetes-clusters/<cluster_id>?verbose=false HTTP/1.1\" 500 22 \"\" \"zmon-worker/0.1\" 10010 cluster-registry.stups.zalan.do
```

The error `context canceled` means that skipper failed to establish a connection with the backend and timed out after 10s. Responding with a 500 in this case seems wrong to me since it's not really an internal server error but rather a problem routing to the backend. It should either be 504 or 503 to better express what is happening IMO.

WDYT?

For reference the 500 comes from here: https://github.com/zalando/skipper/blob/master/proxy/proxy.go#L690-L694

hjacobs:
I would definitely consider this a bug, we had something similar in the past: #269 

Skipper should never return any "internal server error" for proxy failures.

szuecs:
@mlarsen my assumption was that nerr.Timeout() would catch that condition.
Did you see who is canceling what context to choose the right error code?
I am not sure if you are right here and it should be clear what condition should result in which error code. 
For sure there is a bug in some other code part, because it should result in connection refused and not in a timeout. I bet it’s yet another race condition in dataclient  kubernetes on tear down of a pod.

szuecs:
@mikkeloscar I don't see the log message from https://github.com/zalando/skipper/blob/master/proxy/proxy.go#L690 in your issue. Why you are sure it comes from there?

szuecs:
[...]

PR #656, opened 2018-05-28
return proxyError after retry of a loadbalanced route 

szuecs:
one part of the fix for #654 is to make sure we return a 502 in case of "No route to host".

Reply also with proxyError after retry of a loadbalanced route to make sure we create a 502 and not a 500. We have to be consistent in returning proxyError.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #658, opened 2018-05-29
fix goroutine leak in kubernetes dataclient

szuecs:
- goroutine leak is caused by IdleConnTimeout value being the same as a sleep value internally used by the go runtime
- tune some values to be more reasonable to production load


szuecs:
:+1:

mikkeloscar:
:+1:

PR #661, opened 2018-05-31
create filter for setting custom span name for distributed tracing

aryszka:
Creating a filter to set the name of the proxy span for distributed tracing.

- [x] Create a filter
- [x] Apply the span name in the proxy
- [x] Add a parameter to override the initial span name (currently "ingress")
- [x] docs, mark the filter experimental

### DIFF
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -107,7 +107,6 @@ const (
 	experimentalUpgradeUsage       = "enable experimental feature to handle upgrade protocol requests"
 	versionUsage                   = "print Skipper version"
 	maxLoopbacksUsage              = "maximum number of loopbacks for an incoming request, set to -1 to disable loopbacks"
-	opentracingUsage               = "list of arguments for opentracing (space separated), first argument is the tracer implementation"
 	defaultHTTPStatusUsage         = "default HTTP status used when no route is found for a request"
 	pluginDirUsage                 = "set the directory to load plugins from, default is ./"
 	suppressRouteUpdateLogsUsage   = "print only summaries on route updates/deletes"
@@ -129,6 +128,9 @@ const (
 	enableHopHeadersRemovalUsage         = "enables removal of Hop-Headers according to RFC-2616"
 	oauth2TokeninfoURLUsage              = "sets the default tokeninfo URL to query information about an incoming OAuth2 token in oauth2Tokeninfo filters"
 	maxAuditBodyUsage                    = "sets the max body to read to log inthe audit log body"
+
+	opentracingUsage           = "list of arguments for opentracing (space separated), first argument is the tracer implementation"
+	opentracingIngressSpanName = "set the name of the initial, pre-routing, tracing span"
 )
 
 var (
@@ -195,6 +197,7 @@ var (
 	enableRatelimiters              bool
 	ratelimits                      ratelimitFlags
 	openTracing                     string
+	openTracingInitialSpan          string
 	defaultHTTPStatus               int
 	pluginDir                       string
 	suppressRouteUpdateLogs         bool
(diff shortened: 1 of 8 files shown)


lmineiro:
:+1:

cmd/skipper/main.go
```diff
@@ -129,6 +128,9 @@ const (
 	enableHopHeadersRemovalUsage         = "enables removal of Hop-Headers according to RFC-2616"
 	oauth2TokeninfoURLUsage              = "sets the default tokeninfo URL to query information about an incoming OAuth2 token in oauth2Tokeninfo filters"
 	maxAuditBodyUsage                    = "sets the max body to read to log inthe audit log body"
+
```
szuecs: nit: hmm separate it from the rest?
If you do this we should probably do it in blocks in a separate PR to make this more readable for all the things.

aryszka: definitely, i just wanted to keep the changeset in this PR small. Is it fine if I just create a separate PR for the formatting?

filters/tracing/spanname.go
```diff
@@ -0,0 +1,53 @@
...
+
+type filter struct {
+	spanName string
+}
+
+// NewSpanName creates a filter spec for setting the name of the outgoing span. (By default "proxy".)
+//
+// 	tracingSpanName("example-operation")
+//
+// WARNING: this filter is experimental, and the name and the arguments can change until marked as stable.
```
szuecs: Do we have a policy documented somewhere how to mark things as experimental or deprecated?
I like the idea, maybe it's enough like this and we should create a separate issue to document how to do this with introducing new features and how to deal with old stuff, which I personally mark as DEPRECATED in our code base.

proxy/proxy.go
```diff
@@ -482,22 +488,28 @@ func WithParams(p Params) *Proxy {
 		defaultHTTPStatus = p.DefaultHTTPStatus
 	}
 
+	openTracingInitialSpan := p.OpenTracingInitialSpan
+	if openTracingInitialSpan == "" {
```
szuecs: This can not happen as we have a default from CLI set, or is this helping the skipper use as library?

skipper.go
```diff
@@ -792,12 +796,17 @@ func Run(o Options) error {
...
 		proxyParams.OpenTracer = tracer
+		proxyParams.OpenTracingInitialSpan = o.OpenTracingInitialSpan
 	} else {
 		// always have a tracer available, so filter authors can rely on the
 		// existence of a tracer
 		proxyParams.OpenTracer, _ = tracing.LoadTracingPlugin(o.PluginDirs, []string{"noop"})
 	}
 
+	if proxyParams.OpenTracingInitialSpan != "" {
+		proxyParams.OpenTracingInitialSpan = o.OpenTracingInitialSpan
```
szuecs: this should catch the skipper use as library, not?

filters/tracing/spanname.go
aryszka: we don't have currently such a policy, and yes it would be useful. I like the godoc for this, because that's something that we can assume someone always reads. How about using all-caps labels: EXPERIMENTAL, DEPRECATED, and the reason, in its own paragraph of the related godoc? e.g:

```
// NewFoo creates and initializes a new foo object. Use the Close() method to release
// resources, once it's not required anymore.
//
// EXPERIMENTAL: this feature is in an experimental state, and can be removed or
// changed in the near future.
//
func NewFoo() foo { return 42 }
```

proxy/proxy.go
aryszka: yes, it's for the library use case, potentially for testing, too. Not particularly useful, but it's minimal effort and code.

szuecs:
I added one nit and a couple of questions. Code is fine, only one generic question:
Do you think to have "ingress" and "proxy" span not to be too generic?
Maybe it would be better to use "skipper-ingress" or "skipper" and "skipper-proxy-call" .
Thoughts?

skipper.go
aryszka: well, not if someone uses the proxy package as a library. In theory, it is possible.

aryszka:
I like the terms "ingress" and "proxy", because there's a plethora of use cases and I couldn't find a better name for these operations. They are good defaults. True that they are generic, but the potential use cases of skipper are also a lot. I prefer not to put "skipper-" in the span name, because I think the span names should be about the operation, not the entities. We should try to go with the current way, and collect some experience.

The reason for this PR is that current use case raised the need to customize the name "ingress", because for one we wanted to avoid confusion with Kubernetes ingress. And we also want to give route specific operation names for the proxy spans based on the feature. At least to experiment, how it works out.

aryszka:
see my question regarding the EXPERIMENTAL and DEPRECATED labels ;)

cmd/skipper/main.go
szuecs: yes

szuecs:
:+1:

filters/tracing/spanname.go
szuecs: I like EXPERIMENTAL (better than WARNING) and DEPRECATED as godoc

szuecs:
@aryszka then lets change the godoc from WARNING to EXPERIMENTAL

lmineiro:
👍🏻

vigneshshanmugam:
👍

lmineiro:
:+1:

aryszka:
:+1:

PR #663, opened 2018-06-01
group command line options

aryszka:
this is only a formatting of the command line options, primarily to reduce realigning too many lines when adding a new cmd line option


szuecs:
👍

PR #665, opened 2018-06-04
fix #654 - return gw timeout

### LINKED ISSUE #654: Don't respond with status 500 for timeout errors

szuecs:
fix #654 
if stdlib cancels or time exceeded, then return proxy timeout instead of 500

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -721,6 +721,12 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
 			}
 		}
 		p.log.Errorf("error during backend roundtrip: %s: %v", ctx.route.Id, err)
+
+		if cerr := req.Context().Err(); cerr != nil {
+			p.log.Errorf("Failed to do request, because of context: %v", cerr)
+			return nil, &proxyError{err: cerr, code: http.StatusGatewayTimeout}
+		}
+
 		return nil, &proxyError{err: err}
 	}
 	ext.HTTPStatusCode.Set(proxySpan, uint16(response.StatusCode))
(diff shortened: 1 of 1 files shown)


aryszka:
:+1:

szuecs:
@mikkeloscar added another check as discussed.
@aryszka yet another part which could be the reason.

aryszka:
:+1:

szuecs:
:+1:

Issue #653, opened 2018-05-26
document kubernetes apps deployment strategy for terminating


szuecs:
As described in #652, there is a problem that exists in kubernetes, while terminating endpoints.
Mitigation strategies can be different and we should document strategies for application developers.

An application that is target of an ingress can circumvent 500s with these strategies:

1. use pod lifecycle hooks https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/#define-poststart-and-prestop-handlers:

```yaml
lifecycle:
  preStop:
    exec:
      command: ["sleep","20"]
```

2. application implements a SIGTERM handler that changes the readinessProbe target to unready for the instance to get automatically deleted from the endpoints list and hang until SIGKILL will cleanup the instance after 60s.

```go
var sigs chan os.Signal
sigs = make(chan os.Signal, 1)
signal.Notify(sigs, syscall.SIGTERM)
for {
    select {
        <-sigs:
           healthCheck = unhealthy
    }
    doWork()
}
```

3. <add missing idea>

PR #666, opened 2018-06-05
document teardown strategy for applications

szuecs:
docuement #653


docs/kubernetes/ingress-backends.md
```diff
@@ -0,0 +1,70 @@
+# Kubernetes Backend Deployments
+
+## Kubernetes Race Condition problem
+
+As described in [#652](https://github.com/zalando/skipper/issues/652),
+there is a problem that exists in kubernetes, while terminating Pods.
+Terminating Pods could be graceful, but the nature of distributed
```
aryszka: In the Kubernetes docs, 'pod' is written in lowercase. https://kubernetes.io/docs/concepts/workloads/pods/pod/

szuecs: here the do not: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/

szuecs: I think Pod is a name, so I decided to do the first character uppercase. I don't mind, I can change. What do you think?

ruiaraujo: +1 in favor of Pod.

aryszka: \me marked it as resolved :+1:

docs/kubernetes/ingress-backends.md
```diff
@@ -0,0 +1,70 @@
...
+strategies for application developers to mitigate the problem.
+
+## Teardown strategies
+
+An application that is target of an ingress can circumvent HTTP code
+504s Gateway Timeouts with these strategies:
+
+1. use [Pod lifecycle hooks](#pod-lifecycle-hooks)
+2. use a SIGTERM handler to switch `readinessProbe` to unhealthy and
+teardown your application later, probably just waiting until SIGKILL
```
aryszka: The lead sentence starts like this: "An application that is target of...". So instead of "...and teardown your application, ...", we could write "...and exit later, or just wait for SIGKILL terminating the process."

docs/kubernetes/ingress-backends.md
```diff
@@ -0,0 +1,70 @@
...
+list and from skipper's routing table. Similar to [Pod Lifecycle
+Hooks](#pod-lifecycle-hooks) you could sleep 20 seconds and after that
+terminate your application or you just wait until SIGKILL will cleanup
+the instance after 60s.
+
+```go
+var sigs chan os.Signal
+sigs = make(chan os.Signal, 1)
+signal.Notify(sigs, syscall.SIGTERM)
+for {
```
aryszka: What suppose to do this code snippet? To me what this means that "wait for SIGTERM, call doWork(), and then wait for the next SIGTERM".

aryszka: as discussed, basically the default branch is missing here. So it could sg like this:

```go
var sigs chan os.Signal
sigs = make(chan os.Signal, 1)
signal.Notify(sigs, syscall.SIGTERM)
for {
    select {
    case <-sigs:
        healthCheck = unhealthy
    default:
    }
    doWork()
}
```

szuecs:
:+1:

Issue #662, opened 2018-06-01
Nondeterministic route matching ?


beenokle:
I have an ingress. Lets say this simplified:
```
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: my-ingress
  annotations:
spec:
  rules:     
  - host: "myhost.com"
    http:
      paths:
      - path: "/path"
         backend:
          serviceName: upstream1
          servicePort: 80
      - backend:
          serviceName: upstream2
          servicePort: 80
```

I have three nodes in cluster.
When I run `curl -H "host: myhost.com" localhost:9999/path/foo/bar/long/path/1` locally on each node I got different results, for instance two times expected result from `upstream1`, and on another node I got unexpected result from `upstream2`
But If I do the same request directly to clusterIP of the service , I got expected result on each node. This narrows down the source of my issue to Skipper. 

Version is `0.10.13`

hjacobs:
This looks like a bug, but can you first check whether skipper correctly updated all routes on both nodes?  I could suspect that one skipper instance (on of the nodes) has outdated routes.

szuecs:
Hi!

We had a similar case yesterday. The current workaround is to use 2 ingress and to specify also a Predicate annotation on the path ingress.   

```
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: my-ingress-path
  annotations:
    zalando.org/skipper-predicate: PathSubtree("/path")
spec:
  rules:     
  - host: "myhost.com"
    http:
      paths:
      - path: "/path"
         backend:
          serviceName: upstream1
          servicePort: 80
```

```
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: my-ingress-root
  annotations:
spec:
  rules:     
  - host: "myhost.com"
    http:
      paths:
      - backend:
          serviceName: upstream2
          servicePort: 80
```

szuecs:
@hjacobs this was also found internally, yesterday. We had a discussion and I think we agreed on this is a bug for the use in kubernetes at least. We will defintively fix it.

beenokle:
>  skipper correctly updated all routes on both nodes? I could suspect that one skipper instance (on of the nodes) has outdated routes.

Yes, `localhost:9911/routes` shows all the nodes are aware of these routes.
I think the problem appears, when you have:

`path: "/some/path/longer"`

and

`path: "/some/path"`

Then skipper can't pick one deterministically

ese:
[...]

Issue #360, opened 2017-06-30
improve Kubernetes ingress path matching


aryszka:
the ingress API defines only path regular expressions. This can lead to overlapping matching rules, like /.* and /foo, and with multiple, independently configured services this can lead to unexpected routing conflicts. Investigate the possibility to use a combination of the PathSubtree() and PathRegexp() to mitigate this issue.

aryszka:
see https://github.com/zalando/skipper/issues/304 for previous discussions.

jchanam:
Hi @aryszka,

We've been struggling with this for a long time and, after checking the code and how the score is calculated, we want to propose a solution about defaults routes.

We think that a weight-setter would be a good solution for those very specific cases that maybe you'd want to be the very last or very first route to use.

I dont't know if this should be created as a filter, predicate or a new thing like "weight".

The idea is that this will be checked on: https://github.com/zalando/skipper/blob/edb1a2d64027d9af4e41bfe608935d410bba499f/routing/matcher.go#L42

If it's set, that weight will be taken into account, causing matching as we desire. This piece of code could be something like this:
```go
func leafWeight(l *leafMatcher) int {
	w := l.Weight // 0 by default

	if l.method != "" {
		w++
	}

	w += len(l.hostRxs)
	w += len(l.pathRxs)
	w += len(l.headersExact)
	w += len(l.headersRegexp)
	w += len(l.predicates)

	return w
}
```

aryszka:
Hi @jchanam ,

I like this proposal with the weights.

I think it definitely should be a Predicate, so that we don't need to extend the config syntax when we define routes in eskip format.

It could be called "Weight" or maybe "Priority".
(@lmineiro , @szuecs any naming suggestions?).

An example in eskip then would look like:

```
foo: Weight(-1) && PathRegexp("^/foo") -> "https://foo";
fooExact: Weight(9) && PathRegexp("^/foo$") -> "https://foo-exact";
```
or with "Priority":

```
foo: Priority(-1) && PathRegexp("^/foo") -> "https://foo";
fooExact: Priority(9) && PathRegexp("^/foo$") -> "https://foo-exact";
```

Thanks for the proposal! :smile:

szuecs:
https://opensource.zalando.com/skipper/reference/predicates/#weight-priority exists

PR #667, opened 2018-06-07
provide path modes for Kubernetes ingress paths

aryszka:
fixes: #662 and #360

introducing path handling modes in order to allow control over how ingress paths are interpreted by default and per-ingress.

- [x] global path handling mode for the Kubernetes data client
- [x] ingress specific path handling mode
- [x] command line option for the global path handling mode
- [x] documentation
- [x] drop ExactPath

### DIFF
--- a/dataclients/kubernetes/path_test.go
+++ b/dataclients/kubernetes/path_test.go
@@ -332,3 +332,106 @@ func TestPathMatchingModes(t *testing.T) {
 		}
 	})
 }
+
+func TestPathModeParsing(t *testing.T) {
+	for _, test := range []struct {
+		str  string
+		mode PathMode
+		fail bool
+	}{{
+		str:  "foo",
+		fail: true,
+	}, {
+		str:  defaultPathModeString,
+		mode: DefaultPathMode,
+	}, {
+		str:  pathRegexpString,
+		mode: PathRegexp,
+	}, {
+		str:  exactPathString,
+		mode: ExactPath,
+	}, {
+		str:  pathPrefixString,
+		mode: PathPrefix,
+	}} {
+		t.Run(test.str, func(t *testing.T) {
+			m, err := ParsePathMode(test.str)
+
+			if err == nil && test.fail {
+				t.Fatal("failed to fail")
+			} else if err != nil && !test.fail {
+				t.Fatal(err)
+			} else if err != nil {
+				return
+			}
+
+			if m != test.mode {
+				t.Errorf(
+					"failed to parse the right mode, got: %v, expected: %v",
+					m,
+					test.mode,
+				)
+			}
+		})
+	}
+}
+
+func TestIngressSpecificMode(t *testing.T) {
+	s := testServices()
+	api := newTestAPI(t, s, &ingressList{})
+	defer api.Close()
+
+	ingressWithDefault := testIngress(
+		"namespace1", "ingress1", "service1", "", "", "", "", backendPort{8080}, 1.0,
+		testRule("www.example.org", testPathRule("^/foo", "service1", backendPort{8080})),
+	)
+
+	ingressWithCustom := testIngress(
+		"namespace1", "ingress1", "service1", "", "", "", "", backendPort{8080}, 1.0,
+		testRule("www.example.org", testPathRule("/bar", "service1", backendPort{8080})),
+	)
+	ingressWithCustom.Metadata.Annotations[pathModeAnnotationKey] = pathPrefixString
+
+	api.ingresses.Items = []*ingressItem{ingressWithDefault, ingressWithCustom}
+
+	c, err := New(Options{
+		KubernetesURL: api.server.URL,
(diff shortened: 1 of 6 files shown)


aryszka:
Some naming advice would be very welcome. The name ExactPath is not precise, because it accepts wildcards, like `/users/:id` or `/api/**`. In fact, the only difference between `Path("/api/**")`, aka ExactPath, and `PathSubtree("/api")`, aka PathPrefix, is only in the handling of the trailing "/". Shall ExactPath maybe called SkipperPath?

szuecs:
In the original issue we described 3 different cases, now we have 4 in the PR. I am not sure if this is not too complicated for the user, or we should guide them that they select the most obvious one first and might test later others. I am not sure about the behaviour of ExactPath. What happens with these matches:

- /foo  does it match  /foo/ ?
- /foo/
- /foo*   valid? if so does it match /foobar and /foo/bar ?
- /foo/*  does it match /foo/bar and /foo/a/b or is this globbing and ** is the only subtree match?
- /foo**   valid?
- /foo/**  does it match /foo ?

Here the copy of the current godoc for the mode constants:
```
+	// KubernetesIngressMode is the default path mode. Expects regular expressions
+	// with a mandatory leading "/". The expressions are automatically prepended by
+	// the "^" control character.
+	KubernetesIngressMode PathMode = iota
+
+	// PathRegexp is like KubernetesIngressMode but is not prepended by the "^"
+	// control character.
+	PathRegexp
+
+	// PathPrefix is like the PathSubtree predicate. E.g. "/foo/bar" will match
+	// "/foo/bar" or "/foo/bar/baz", but won't match "/foo/barooz".
+	//
+	// In this mode, when a Path or a PathSubtree predicate is set in an annotation,
+	// the value from the annotation has precedence over the standard ingress path.
+	PathPrefix
+
+	// ExactPath is like the Path predicate. E.g. "/foo/bar" will only match
+	// "/foo/bar".
+	//
+	// In this mode, when a Path or a PathSubtree predicate is set in an annotation,
+	// the value from the annotation has precedence over the standard ingress path.
+	ExactPath
+)
```

dataclients/kubernetes/path_test.go
```diff
@@ -0,0 +1,392 @@
...
+
+		if p != nil {
+			return ri, nil
+		}
+	}
+
+	return nil, nil
+}
+
+func findRouteWithExactPath(r []*eskip.Route) (*eskip.Route, error) {
```
szuecs: this can be dropped and all tests depdening on that, correct?

dataclients/kubernetes/path_test.go
```diff
@@ -0,0 +1,392 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if p.Args[0] != "/foo" {
+			t.Error("invalid path prefix value", p.Args[0])
+		}
+	})
+
+	t.Run("additional exact path from annotation", func(t *testing.T) {
```
szuecs: this does not have an implementation anymore

docs/kubernetes/ingress-usage.md
```diff
@@ -75,6 +75,51 @@ Cloud migration.
...
+to the default kubernetes ingress specification way, but is not prepended by the "^"
+control character.
+
+### Path prefix
+
+When the path mode is set to "path-prefix", the ingress path is not a regular
+expression. As an example, "/foo/bar" will match "/foo/bar" or "/foo/bar/baz", but
+won't match "/foo/barooz".
+
+### Path prefix behavior
```
szuecs: this is a subesction of Path prefix, so either no header or do 4x #

docs/kubernetes/ingress-usage.md
```diff
@@ -75,6 +75,51 @@ Cloud migration.
...
+
+When PathPrefix is used, the path matching becomes deterministic when
+a request could match more than one ingress routes otherwise.
+
+### Path prefix with predicate annotations
```
szuecs: this is a subesction of Path prefix, so either no header or do 4x #

dataclients/kubernetes/path_test.go
aryszka: this is used in the test where we use a path prefix for the ingress path, but the annotation contains an explicit path, and we verify that the explicit path overrules the path prefix

dataclients/kubernetes/path_test.go
aryszka: currently we don't prevent somebody adding an path predicate (Path() or PathSubtree()) via annotation, and I think this is fine. With these tests we just verify that both regexp based path and the predicate path is there, because when use path prefix mode, only one of the should be kept.

aryszka:
Thanks for the review! See my responses regarding the tests.

szuecs:
:+1:

szuecs:
As offline discussed I agree for having these tests

aryszka:
:+1:

PR #672, opened 2018-06-10
move the traffic predicate to the decision route

aryszka:
Currently, when the traffic switching is used together with load balanced routes, the right route often may not be found. When there multiple load balanced route groups with traffic control, the route decision may keep looping, until the maxLoopback is reached.

The reason is that the Traffic() predicate is set for the member routes in the load balancing group, and it causes the already selected route match only with a certain chance.

This PR would move the Traffic() predicate to the loadbalancing group decision route from the member routes.

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -740,15 +740,6 @@ func (c *Client) convertPathRule(
 		}
 
 		setPath(pathMode, r, prule.Path)
-
-		// add traffic predicate if traffic weight is between 0.0 and 1.0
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

mikkeloscar:
@aryszka While this makes sense, I'm not completely sure it fixes the RC of the issues we saw. I tried this exact change myself the other day and I were still able to produce the errors. I did not investigate more so I'm not 100% sure why.

Either way we should add this as it makes more sense to do the traffic switching at this level.

mikkeloscar:
:+1:

aryszka:
@mikkeloscar right. This was one way I could reproduce the described behavior. So let's merge it, and find the remaining issues separately.

PR #675, opened 2018-06-11
include the response streaming duration in the proxy span

aryszka:
Currently, we set two spans in the proxy for opentracing:

- server span ("ingress")
- client span ("proxy")

The proxy span starts before making the backend request. However it is closed as soon as the response headers are parsed, or the request failed. After discussing it with @lmineiro, I believe the proxy span should include the duration while the response content is fully copied onto the incoming connection. This PR moves finishing the proxy span after the stream copying is done or aborted.

### DIFF
--- a/proxy/proxy_test.go
+++ b/proxy/proxy_test.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/zalando/skipper/filters"
 	"github.com/zalando/skipper/filters/builtin"
+	"github.com/zalando/skipper/loadbalancer"
 	"github.com/zalando/skipper/logging/loggingtest"
 	"github.com/zalando/skipper/routing"
 	"github.com/zalando/skipper/routing/testdataclient"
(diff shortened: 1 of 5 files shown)


proxy/proxy_test.go
```diff
@@ -135,7 +136,12 @@ func newTestProxyWithFiltersAndParams(fr filters.Registry, doc string, params Pa
 		FilterRegistry: fr,
 		PollTimeout:    sourcePollTimeout,
 		DataClients:    []routing.DataClient{dc},
-		Log:            tl})
```
szuecs: Are you sure you want to add these predicates and these are not from another pr?

aryszka: yes, they are used in the test `TestTracingProxySpanWithRetry`. Do you think they can cause an issue in other tests?

aryszka: actually, i think it should be fine, it's just additional predicates, so yes :smile:

szuecs:
:+1:

proxy/proxy_test.go
szuecs: I just ask, because I wondered that you modified a test and I always ask by test modifications that are not obvious.

vetinari:
:+1:

PR #676, opened 2018-06-12
do less route updates

szuecs:
sort ingress and endpoints to be more predictable with route updates

This will not entirely fix the problem of unstable blue-green traffic switching.


aryszka:
added a change. For some reason, we deal with the cases of missing metadata, but ignoring those specs.

aryszka:
:+1:

szuecs:
:+1:

PR #679, opened 2018-06-14
prevent multiple load balancer groups to match

aryszka:
when a previous load balancer group was already matched, the LBGroup() predicate will always return false

### DIFF
--- a/loadbalancer/predicate.go
+++ b/loadbalancer/predicate.go
@@ -39,7 +39,7 @@ func getGroupDecision(h http.Header, group string) (string, bool) {
 		}
 	}
 
-	return "", false
+	return "", len(h[DecisionHeader]) > 0
 }
 
 // NewGroup creates a predicate spec identifying the entry route
--- a/loadbalancer/predicate.go
+++ b/loadbalancer/predicate.go
@@ -27,7 +27,7 @@ type memberPredicate struct {
 	indexString string
 }
 
-func getGroupDecision(h http.Header, group string) (string, bool) {
+func getGroupDecision(h http.Header, group string) string {
 	for _, header := range h[DecisionHeader] {
 		decision := strings.Split(header, "=")
 		if len(decision) != 2 {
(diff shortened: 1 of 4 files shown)


aryszka:
The original problem with the Traffic control was hard to test so that it always covers the underlying problem, so the test is now simplified to use groups without members. This way it is always reproduce the difference between not matching the wrong route, represented by a 404, vs ending up in an infinite loopback.

loadbalancer/predicate.go
```diff
@@ -39,7 +39,7 @@ func getGroupDecision(h http.Header, group string) (string, bool) {
 		}
 	}
 
-	return "", false
+	return "", len(h[DecisionHeader]) > 0
```
mikkeloscar: nit: I think a doc string for this function would be helpful to understand what is really going on here.

proxy/proxytest/proxytest.go
```diff
@@ -42,6 +43,7 @@ func newTestProxy(fr filters.Registry, routingOptions routing.Options, proxyPara
 		routingOptions.Predicates = []routing.PredicateSpec{
 			loadbalancer.NewGroup(),
 			loadbalancer.NewMember(),
+			traffic.New(),
```
mikkeloscar: Given your last comment, is this needed still? I don't fully understand what it's used for.

aryszka:
thanks @mikkeloscar 
removed the unused test changes. For the getGroupDecision() function, instead of doc string, I split the function into two with single responsibility and descriptive names.

loadbalancer/predicate.go
```diff
@@ -27,19 +27,23 @@ type memberPredicate struct {
...
-			return decision[1], true
+			return decision[1]
 		}
 	}
 
-	return "", false
+	return ""
+}
+
+func hasGroupDecision(h http.Header, group string) bool {
```
mikkeloscar: `group` arg is not used.

mikkeloscar:
@aryszka thanks for digging into this!

szuecs:
@mikkeloscar we also tested it in playground, that it fixes the problem. I could not stop the investigation. O:)
@aryszka thanks for helping with investigation, opening the PR and writing the tests, awesome to see you always write tests!

szuecs:
fixed comment @mikkeloscar

szuecs:
:+1:

ese:
:100: We were facing some issues due this

PR #681, opened 2018-06-15
sort routing table output on support endpoint

szuecs:
sort routing table output to create simply diffs if you want to investigate differences of 2 routing tables

### DIFF
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -3,6 +3,7 @@ package routing
 import (
 	"fmt"
 	"net/url"
+	"sort"
 	"time"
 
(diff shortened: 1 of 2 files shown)


aryszka:
:+1:

aryszka:
it would be more reasonable to sort on creation, not on each fetch. I think around this code block: https://github.com/zalando/skipper/blob/master/routing/datasource.go#L504

Also, pls add some tests.

aryszka:
:+1:

aryszka:
moved the sorting to where the reporting routes are created

szuecs:
:+1:

PR #684, opened 2018-06-20
Testing/export test tracer

aryszka:
This PR is exporting the testtracer from the proxy package into its own package to allow other packages and extensions to use it for their tests.


lmineiro:
I am wondering if it wouldn't be better to use the [upstream mock tracer](https://github.com/opentracing/opentracing-go/blob/master/mocktracer/mocktracer.go)

This could give us the benefit of having it maintained upstream and updated whenever the API changes.

aryszka:
@lmineiro true, thanks for pointing this out. The reason why have this mock now is basically that it evolved step-by-step from a small state together with the changes to the tracing code. For the long term, it is worth to switch to the upstream one, and since both implement the same interface, it should not be a big effort to migrate from this one and deprecate it.

aryszka:
:+1:

Issue #686, opened 2018-06-21
Scope Kubernetes Dataclient to a Single Namespace


daviddyball:
Hi Guys,

I can see in https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/kube.go#L1100 that Skipper currently uses the cluster-scoped URI for listing ingresses. Is it possible to have a flag that allows switching between the two types of ingress list endpoints?

Cluster scoped: `/apis/extensions/v1beta1/ingresses`
Namespaced: `/apis/extensions/v1beta1/namespaces/{namespace}/ingresses`

Perhaps something like `-kubernetes-namespace={namespace}` that would allow switching between the two types of behaviour?

Cheers,
David

szuecs:
What is the use case for that?
We have the ownership model via annotation, so what is the advantage of having it per namespace separated?
I am in general fine with that, but the use case should be clear and documented.

daviddyball:
Hi @szuecs ,

We have multiple environments separated by namespace (and taints/tolerations for hardware segregation). If, for example, we have `development`, `staging` and `production` namespaces we would have the following ingresses configured: `web`, `api` and `internal` and each of these has an AWS load-balancer. For ease of configuration, we'd like to re-use the same `ingress.class` values for each, otherwise it's more templating that needs to be done when we deploy into each environments namespace (e.g. using `ingress.class: api` is better than having to do `development-admin`, `staging-admin` and `production-admin`).

Thoughts?

I've also started writing a PR to add support for it, but my Go unit-testing experience is very limited and it feels bad-form to submit a PR without unit-tests 😖 

Cheers,
David

szuecs:
Sounds like a good use case.
Can you add this to the docs/kubernetes/ingress-deployment.md, please? (I hope I remember the filename correctly)

daviddyball:
Docs updated in the PR 👍

daviddyball:
Added a commit for basic unit-test as well 😄

aryszka:
done in https://github.com/zalando/skipper/pull/688

PR #688, opened 2018-06-22
Support for Limiting Ingress Lookups to a Single Namespace

daviddyball:
Issue Reference: https://github.com/zalando/skipper/issues/686

**Purpose**
We have a use-case where we deploy identical environments on the same cluster, segregating them by `Namespace`. To aid in configuration it'd be nice to use the same `ingress.class` in each namespace (e.g. `ingress.class: api`) without having to add the namepsace to the class.

This also has the added benefit of being able to further restrict RBAC permissions by combining a `ClusterRole` and `RoleBinding` to ensure that the associated `ServiceAccount` can't list ingresses in another namespace.

**Implementation**

I've added a `-kubernetes-namepsace` option that defaults to `""`.

* When the argument is omitted (empty) it will use `/apis/extensions/ingresses`
* When a namespace is provided it will use `apis/extensions/namepsaces/{namespace}/ingresses`

I'm still unfamiliar with how go does unit-testing, so I haven't included tests. If anyone can give me a pointer on which of the `kube_test.go` functions is most suited to include this test I'll make sure to add some in another commit.

### DIFF
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -125,6 +125,7 @@ const (
 	kubernetesIngressClassUsage     = "ingress class regular expression used to filter ingress resources for kubernetes"
 	whitelistedHealthCheckCIDRUsage = "sets the iprange/CIDRS to be whitelisted during healthcheck"
 	kubernetesPathModeUsage         = "controls the default interpretation of Kubernetes ingress paths: kubernetes-ingress/path-regexp/path-prefix"
+	kubernetesNamespaceUsage        = "limit ingress monitoring to this namespace"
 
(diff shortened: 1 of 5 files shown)


daviddyball:
I refactored the dataclient to use a new function `getIngressesURI` that allowed me to make a tiny unit-test for it. Let me know how it looks 👍

daviddyball:
@szuecs Comments addressed 👍 Thanks

szuecs:
:+1:

aryszka:
:+1:

aryszka:
thanks a lot! :smile:

PR #690, opened 2018-06-25
Fix Header and HeaderRegexp examples

NorthFury:
* The slashes for HeaderRegexp are misleading. Could be the initial example was using this format: `HeaderRegexp("Server", /skipper/)`.
* Use a request header for the examples.

### DIFF
--- a/docs/predicates.md
+++ b/docs/predicates.md
@@ -134,8 +134,8 @@ Parameters:
 Examples:
 
 ```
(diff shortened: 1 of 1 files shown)


NorthFury:
:+1:

aryszka:
THX!

maybe we should also change it something other than the Server header because that's a response header while the predicate is for requests. Do you mind updating it to sg like `HeaderRegexp("Accept", "^text/html")`?

NorthFury:
:+1:

aryszka:
:+1:

PR #699, opened 2018-07-03
fix invalid connection header issue reported via slack channel

szuecs:
upstream issue: https://github.com/golang/go/issues/23699

Tested with:

```
% ./bin/skipper -inline-routes '* -> "https://www.example.org"'
% curl -H Connection:\ Keep-Alive http://localhost:9090
```


aryszka:
:+1:

PR #707, opened 2018-07-16
Add flag for Oauth request timeout

### LINKED ISSUE #633: Flag for timeouts to oauth filters
Allow to define timeouts to oauth filters. 

- [ ] `-oauth2-tokeninfo-timeout` as a command-line
- [ ] change the authClient to use the specified request timeout

addityasingh:
Close #633 
- [x] Add flag for Oauth request timeout
- [x] Fix test for correct HTTP status code
- [x] Revisit the HTTPClient used

### DIFF
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -57,6 +57,9 @@ const (
 	defaultTLSHandshakeTimeoutBackend = 60 * time.Second
 	defaultMaxIdleConnsBackend        = 0
 
+	// OAuth2:
+	defaultOAuthTokeninfoTimeout = 1500 * time.Millisecond // Not sure if this makes sense
+
 	// generic:
 	addressUsage                         = "network address that skipper should listen on"
 	ignoreTrailingSlashUsage             = "flag indicating to ignore trailing slashes in paths when routing"
@@ -128,12 +131,11 @@ const (
 	kubernetesNamespaceUsage        = "watch only this namespace for ingresses"
 
 	// OAuth2:
-	defaultOAuthTokeninfoTimeout = 100 * time.Millisecond // Not sure if this makes sense
-	oauthURLUsage                = "OAuth2 URL for Innkeeper authentication"
-	oauthCredentialsDirUsage     = "directory where oauth credentials are stored: client.json and user.json"
-	oauthScopeUsage              = "the whitespace separated list of oauth scopes"
-	oauth2TokeninfoURLUsage      = "sets the default tokeninfo URL to query information about an incoming OAuth2 token in oauth2Tokeninfo filters"
-	oauth2TokeninfoTimeoutUsage  = "sets the default tokeninfo request timeout duration"
+	oauthURLUsage               = "OAuth2 URL for Innkeeper authentication"
+	oauthCredentialsDirUsage    = "directory where oauth credentials are stored: client.json and user.json"
+	oauthScopeUsage             = "the whitespace separated list of oauth scopes"
+	oauth2TokeninfoURLUsage     = "sets the default tokeninfo URL to query information about an incoming OAuth2 token in oauth2Tokeninfo filters"
+	oauth2TokeninfoTimeoutUsage = "sets the default tokeninfo request timeout duration to 1500ms"
 
 	// connections, timeouts:
 	idleConnsPerHostUsage           = "maximum idle connections per backend host"
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -128,10 +128,12 @@ const (
 	kubernetesNamespaceUsage        = "watch only this namespace for ingresses"
 
 	// OAuth2:
-	oauthURLUsage            = "OAuth2 URL for Innkeeper authentication"
-	oauthCredentialsDirUsage = "directory where oauth credentials are stored: client.json and user.json"
-	oauthScopeUsage          = "the whitespace separated list of oauth scopes"
-	oauth2TokeninfoURLUsage  = "sets the default tokeninfo URL to query information about an incoming OAuth2 token in oauth2Tokeninfo filters"
+	defaultOAuthTokeninfoTimeout = 100 * time.Millisecond // Not sure if this makes sense
+	oauthURLUsage                = "OAuth2 URL for Innkeeper authentication"
+	oauthCredentialsDirUsage     = "directory where oauth credentials are stored: client.json and user.json"
+	oauthScopeUsage              = "the whitespace separated list of oauth scopes"
+	oauth2TokeninfoURLUsage      = "sets the default tokeninfo URL to query information about an incoming OAuth2 token in oauth2Tokeninfo filters"
+	oauth2TokeninfoTimeoutUsage  = "sets the default tokeninfo request timeout duration"
 
 	// connections, timeouts:
 	idleConnsPerHostUsage           = "maximum idle connections per backend host"
@@ -235,10 +237,11 @@ var (
 	kubernetesNamespace        string
 
 	// OAuth2:
-	oauthURL            string
-	oauthScope          string
-	oauthCredentialsDir string
-	oauth2TokeninfoURL  string
+	oauthURL               string
+	oauthScope             string
+	oauthCredentialsDir    string
+	oauth2TokeninfoURL     string
+	oauth2TokeninfoTimeout time.Duration
 
 	// connections, timeouts:
 	idleConnsPerHost           int
@@ -344,6 +347,7 @@ func init() {
 	flag.StringVar(&oauthScope, "oauth-scope", "", oauthScopeUsage)
 	flag.StringVar(&oauthCredentialsDir, "oauth-credentials-dir", "", oauthCredentialsDirUsage)
 	flag.StringVar(&oauth2TokeninfoURL, "oauth2-tokeninfo-url", "", oauth2TokeninfoURLUsage)
+	flag.DurationVar(&oauth2TokeninfoTimeout, "oauth2-tokeninfo-timeout", defaultOAuthTokeninfoTimeout, oauth2TokeninfoTimeoutUsage)
 
 	// connections, timeouts:
 	flag.IntVar(&idleConnsPerHost, "idle-conns-num", proxy.DefaultIdleConnsPerHost, idleConnsPerHostUsage)
@@ -499,10 +503,11 @@ func main() {
 		KubernetesNamespace:        kubernetesNamespace,
 
 		// OAuth2:
-		OAuthUrl:            oauthURL,
-		OAuthScope:          oauthScope,
-		OAuthCredentialsDir: oauthCredentialsDir,
-		OAuthTokeninfoURL:   oauth2TokeninfoURL,
+		OAuthUrl:              oauthURL,
+		OAuthScope:            oauthScope,
+		OAuthCredentialsDir:   oauthCredentialsDir,
+		OAuthTokeninfoURL:     oauth2TokeninfoURL,
+		OAuthTokeninfoTimeout: oauth2TokeninfoTimeout,
 
 		// connections, timeouts:
 		IdleConnectionsPerHost:     idleConnsPerHost,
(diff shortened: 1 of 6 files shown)


cmd/skipper/main.go
```diff
@@ -128,10 +128,12 @@ const (
...
-	oauthURLUsage            = "OAuth2 URL for Innkeeper authentication"
-	oauthCredentialsDirUsage = "directory where oauth credentials are stored: client.json and user.json"
-	oauthScopeUsage          = "the whitespace separated list of oauth scopes"
-	oauth2TokeninfoURLUsage  = "sets the default tokeninfo URL to query information about an incoming OAuth2 token in oauth2Tokeninfo filters"
+	defaultOAuthTokeninfoTimeout = 100 * time.Millisecond // Not sure if this makes sense
+	oauthURLUsage                = "OAuth2 URL for Innkeeper authentication"
+	oauthCredentialsDirUsage     = "directory where oauth credentials are stored: client.json and user.json"
+	oauthScopeUsage              = "the whitespace separated list of oauth scopes"
+	oauth2TokeninfoURLUsage      = "sets the default tokeninfo URL to query information about an incoming OAuth2 token in oauth2Tokeninfo filters"
+	oauth2TokeninfoTimeoutUsage  = "sets the default tokeninfo request timeout duration"
```
lmineiro: Let's add some info what a value of 0 means - no timeout, stdlib default (how much?), etc.

cmd/skipper/main.go
```diff
@@ -128,10 +128,12 @@ const (
 	kubernetesNamespaceUsage        = "watch only this namespace for ingresses"
 
 	// OAuth2:
-	oauthURLUsage            = "OAuth2 URL for Innkeeper authentication"
-	oauthCredentialsDirUsage = "directory where oauth credentials are stored: client.json and user.json"
-	oauthScopeUsage          = "the whitespace separated list of oauth scopes"
-	oauth2TokeninfoURLUsage  = "sets the default tokeninfo URL to query information about an incoming OAuth2 token in oauth2Tokeninfo filters"
+	defaultOAuthTokeninfoTimeout = 100 * time.Millisecond // Not sure if this makes sense
```
lmineiro: This seems too short as a default to me. It should be more relaxed by default and be able to accommodate, let's say, the redirects the default client will follow - which count towards the timeout. I propose something more in the range of 1~2 sec as default. End users should customize that for their own use case. If your use case considers more appropriate 100ms, that's the value you should use.

skipper.go
```diff
@@ -433,6 +433,9 @@ type Options struct {
 	// OAuthTokeninfoURL sets the OAuthTokeninfoURL similar to https://godoc.org/golang.org/x/oauth2#Endpoint
 	OAuthTokeninfoURL string
 
+	// OAuthTokeninfoTimeout sets timwout duration while calling oauth token service
```
lmineiro: typo s/timwout/timeout

filters/auth/oauth.go
```diff
@@ -251,7 +261,9 @@ func (s *tokeninfoSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 		return nil, filters.ErrInvalidFilterParameters
 	}
 
-	ac, err := newAuthClient(s.tokeninfoURL)
+	log.Info("Log the timeout ", s.tokenInfoTimeout)
```
lmineiro: Please remove this log line

cmd/skipper/main.go
szuecs: +1 and please move the defaults to the defaults section, thanks

filters/auth/oauth.go
```diff
@@ -159,7 +162,14 @@ func jsonGet(url *url.URL, auth string, doc interface{}) error {
 		return err
 	}
 
-	rsp, err := http.DefaultClient.Do(req)
+	var client *http.Client
+	if timeout != 0 {
+		client = &http.Client{Timeout: timeout}
```
szuecs: This will not failover DNS if someone will change the traffic to another loadbalancer for example.
We need to have the workaround as in the kubernetes dataclient or the proxy.
See https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/kube.go#L310-L337.

lmineiro: @szuecs Is this still true since the introduction of Transport.IdleConnTimeout in go 1.7? I think that new setting which defaults to 90secs is good enough to mitigate that problem.

addityasingh: @szuecs do I need to create the same Transport dialer in the auth filter?

cmd/skipper/main.go
addityasingh: I have updated this to 1500ms

cmd/skipper/main.go
addityasingh: Is it enough to add the timeout value of 1500ms from `defaultOAuthTokeninfoTimeout `?

filters/auth/oauth.go
szuecs: @addityasingh you don't need the same, but you need the `for {}` loop below the transport to close idle connections to make the http client to do a DNS call to make DNS based failover happen.

filters/auth/oauth.go
```diff
@@ -145,9 +149,35 @@ func intersect(left []string, right []string) bool {
 	return false
 }
 
+func createHttpClient(timeout time.Duration) *http.Client {
+	var client *http.Client
+	if timeout != 0 {
```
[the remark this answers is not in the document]
szuecs: Yes use always timeouts

skipper.go
```diff
@@ -655,10 +658,10 @@ func Run(o Options) error {
 	}
 
 	if o.OAuthTokeninfoURL != "" {
```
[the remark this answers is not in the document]
szuecs: No you should not check timeout here

filters/auth/oauth.go
```diff
@@ -145,9 +149,35 @@ func intersect(left []string, right []string) bool {
...
+		transport := &http.Transport{
+			DialContext: (&net.Dialer{
+				Timeout: timeout,
+			}).DialContext,
+		}
+
+		client = &http.Client{Transport: transport}
+
+		go func() {
```
szuecs: We need to have a quit channel to not leak this goroutine

readme.md
```diff
@@ -100,15 +100,20 @@ logging. See the documentation at:
...
+```sh
     go get -t github.com/zalando/skipper/...
-
-Build and test all packages:
-
     cd src/github.com/zalando/skipper
-    make deps
-    make install
+```
+2. Install all dependencies using `glide`
```
aryszka: we have a make task, 'deps', that installs the dependencies

readme.md
```diff
@@ -100,15 +100,20 @@ logging. See the documentation at:
...
+```sh
+glide install
+```
+3. To build the packages and skipper cli, run `make`
```
aryszka: i'd prefer not using the term 'cli'.

readme.md
```diff
@@ -100,15 +100,20 @@ logging. See the documentation at:
 
 Getting the code with the test dependencies (`-t` switch):
 
+1. Get skipper using `go get`
```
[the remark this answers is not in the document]
aryszka: @addityasingh have you checked the 'Working with the code' section below? The reason for this 'Installation' section is left more open, is that 1/ 'go get' actually installs the skipper command in GOPATH/bin, 2/ it leaves the choice between using the latest or the bound version of the dependencies to the user.

This doesn't mean I am not open to improve the readme, but I think it should be a discussion in a separate PR. Can you please move the changes to the readme into a new, separate PR?

filters/auth/oauth.go
aryszka: i wonder if this is required. If it is required, then we should make it possible to return from this goroutine.

filters/auth/oauth.go
```diff
@@ -159,7 +189,8 @@ func jsonGet(url *url.URL, auth string, doc interface{}) error {
 		return err
 	}
 
-	rsp, err := http.DefaultClient.Do(req)
+	client := createHttpClient(timeout)
```
aryszka: we should not create a new client on every request.

cmd/skipper/main.go
```diff
@@ -57,6 +57,9 @@ const (
 	defaultTLSHandshakeTimeoutBackend = 60 * time.Second
 	defaultMaxIdleConnsBackend        = 0
 
+	// OAuth2:
+	defaultOAuthTokeninfoTimeout = 1500 * time.Millisecond // Not sure if this makes sense
```
aryszka: It makes sense, but still a bit opinionated in my view :smile: . How about time.Minute?

filters/auth/oauth.go
addityasingh: At which part of the lifecycle should the `quit` channel be closed?

cmd/skipper/main.go
addityasingh: I have changed it to `1.5 * time.Second`

filters/auth/oauth.go
addityasingh: @szuecs  I did not find any lifecycle method in `oauth` filter similar to `Close()` method in the kubernetes ingress controller. I have added the `quit` channel to return from the `goroutine` but still not sure at which point to send to the `quit` channel.

szuecs: We have a sigterm handler

addityasingh: I checked the quit channel implementations in those 2 but the `kubernetes/dataclients` and `proxy/proxy.go` have `Close()` methods, but a filter doesn't. Should I implement a custom dataclient with Close() method for the oauth filter?

filters/auth/oauth.go
```diff
@@ -173,50 +200,56 @@ func jsonGet(url *url.URL, auth string, doc interface{}) error {
...
 
-func newAuthClient(baseURL string) (*authClient, error) {
+func newAuthClient(baseURL string, timeout time.Duration) (*authClient, error) {
 	u, err := url.Parse(baseURL)
 	if err != nil {
 		return nil, err
 	}
-	return &authClient{url: u}, nil
+
+	quit := make(chan struct{})
```
aryszka: to call `close(quit)`, I would the following:

1. add a Close() method to the filterspec
2. call `close(spec.authClient.quit)` from this Close() method
3. for normal skipper operation do nothing, do not call close, because there's going to be only a single instance of the spec
4. for tests, always call Close() on the spec
5. document that in case someone uses filter spec as a library in a custom build, they need to call Close() if they may create an arbitrary number of spec instances

@addityasingh @szuecs what do you think?

szuecs: sounds good to me

addityasingh: This approach looks clean to me, since we wouldn't need to pass down `quit` channel all the way down from `skipper` instance creation. And since the `Close()` is on `filterspec`, we wouldn't need to bother about the leaking goroutine as well. @szuecs Thoughts?

addityasingh: Since the `FilterSpec` interface doesn't have the `Close` method, this approach wouldn't work. Check this https://play.golang.org/p/LkpQRoatvvn

szuecs: You might be able to type cast it. See: https://play.golang.org/p/BK0kM25jnjb

addityasingh: It works in the playground, but not in the skipper test

szuecs: I see you found out. An interface can only be casted to, if you use the pointer IIRC.

filters/auth/oauth.go
```diff
@@ -280,6 +314,13 @@ func (s *tokeninfoSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 	return f, nil
 }
 
+// Close cleans-up the quit channel used for this spec
+func (s *tokeninfoSpec) Close() {
+	if s.authClient.quit != nil {
```
szuecs: A spec can create filters.
I think you have to close it from the `filter`, that's why you get a nil ptr panic.
See https://github.com/zalando/skipper/pull/707/files#diff-82513d1c973748c3c6b7fa0677fd3fd4L259, which shows that you store it in the filter and not in the spec.

lmineiro:
:+1:

aryszka:
:+1:

aryszka:
Thanks a lot for this PR! :smile:

PR #708, opened 2018-07-17
client closed request should not be reported as 5xx

### LINKED ISSUE #687: 500s replaced with 504s since v0.10.16
Putting this here to not forget about it..

I have a suspicion that we may be "hiding" some errors since introducing: https://github.com/zalando/skipper/pull/665

The reason I think this is that we saw 500s with the following errors:

```
[APP]time="2018-06-13T00:19:33Z" level=error msg="error during backend roundtrip: kube_default__carrier_network__carrier_network_fulfillment_zalan_do____carrier_network_0: http: server closed idle connection"
[APP]time="2018-06-13T00:19:33Z" level=error msg="Failed to do backend request to http://10.2.231.92:8080: proxy error with code 0, dialing failed false, unwrap: http: server closed idle connection"
[APP]time="2018-06-13T00:19:33Z" level=error msg="error while proxying, route kube_default__carrier_network__carrier_network_fulfillment_zalan_do____carrier_network__lb_group, status code 500: proxy error with code 0, dialing failed false, unwrap: http: server closed idle connection"
```

When running v0.10.13 in our clusters and since we switched to v0.10.26: https://github.com/zalando-incubator/kubernetes-on-aws/pull/1165 we no longer see these 500s but instead 504s:

```
[APP]time="2018-06-16T16:27:47Z" level=error msg="error during backend roundtrip: kube_default__si_hyperspace__si_hyperspace_fulfillment_zalan_do____si_hyperspace_master_88_2: context canceled"
[APP]time="2018-06-16T16:27:47Z" level=error msg="Failed to do request, because of context: context canceled"
[APP]time="2018-06-16T16:27:47Z" level=error msg="Failed to do backend request to http://10.2.232.90:9000: proxy error with code 504, dialing failed false, unwrap: context canceled"
[APP]time="2018-06-16T16:27:47Z" level=error msg="error while proxying, route kube_default__si_hyperspace__si_hyperspace_fulfillment_zalan_do____si_hyperspace_master_88__lb_group, status code 504: proxy error with code 504, dialing failed false, unwrap: context canceled"
```

The error is always `context cancelled` but I suspect some of those were `http: server closed idle connection` before.

I will try to validate my theory by running latest skipper with https://github.com/zalando/skipper/pull/665 reverted.

szuecs:
fix #687: client closed request should not be reported as 5xx

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -29,9 +29,10 @@ import (
 )
 
 const (
-	proxyBufferSize = 8192
-	proxyErrorFmt   = "proxy: %s"
-	unknownRouteID  = "_unknownroute_"
+	proxyBufferSize     = 8192
+	proxyErrorFmt       = "proxy: %s"
+	unknownRouteID      = "_unknownroute_"
+	unknownRouteBackend = "<unknown>"
 
 	// Number of loops allowed by default.
 	DefaultMaxLoopbacks = 9
@@ -256,7 +257,7 @@ type proxyError struct {
 
 func (e proxyError) Error() string {
 	if e.err != nil {
-		return fmt.Sprintf("proxy error with code %d, dialing failed %v, unwrap: %v", e.code, e.DialError(), e.err.Error())
+		return fmt.Sprintf("dialing failed %v: %v", e.DialError(), e.err.Error())
 	}
 
 	if e.handled {
@@ -408,9 +409,9 @@ func (dc *skipperDialer) DialContext(ctx stdlibcontext.Context, network, addr st
 			dialingFailed: true, // indicate error happened before http
 		}
 	} else if cerr := ctx.Err(); cerr != nil {
-		// deadline exceeded or canceled in stdlib
+		// unclear when this is being triggered
 		return nil, &proxyError{
-			err:  cerr,
+			err:  fmt.Errorf("err from dial context: %v", cerr),
 			code: http.StatusGatewayTimeout,
 		}
 	}
@@ -723,11 +724,11 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
 				}
 			}
 		}
-		p.log.Errorf("error during backend roundtrip: %s: %v", ctx.route.Id, err)
 
 		if cerr := req.Context().Err(); cerr != nil {
-			p.log.Errorf("Failed to do request, because of context: %v", cerr)
-			return nil, &proxyError{err: cerr, code: http.StatusGatewayTimeout}
+			// deadline exceeded or canceled in stdlib, client closed request
+			// see https://github.com/zalando/skipper/issues/687#issuecomment-405557503
+			return nil, &proxyError{err: cerr, code: 499}
 		}
 
 		return nil, &proxyError{err: err}
@@ -895,7 +896,6 @@ func (p *Proxy) do(ctx *context) error {
 				}
 				p.log.Infof("successful retry to %v, orig %v, code: %d", ctx.route.Backend, origRoute.Backend, rsp.StatusCode)
 			} else {
-				p.log.Errorf("Failed to do backend request to %s: %v", ctx.route.Backend, perr)
 				return perr
 			}
 		}
@@ -950,8 +950,10 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 	}
 
 	id := unknownRouteID
+	backend := unknownRouteBackend
 	if ctx.route != nil {
 		id = ctx.route.Id
+		backend = ctx.route.Backend
 	}
 
 	code := http.StatusInternalServerError
@@ -994,7 +996,7 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 	case ok && perr.err == errRatelimitError:
 		code = perr.code
 	default:
-		p.log.Errorf("error while proxying, route %s, status code %d: %v", id, code, err)
+		p.log.Errorf("error while proxying, route %s with backend %s, status code %d: %v", id, backend, code, err)
 	}
 
 	p.sendError(ctx, id, code)


aermakov-zalando:
Deadline exceeded sounds more like 502 gateway timeout, no? Only client cancels should be reported as 499.

proxy/proxy.go
```diff
@@ -408,10 +408,11 @@ func (dc *skipperDialer) DialContext(ctx stdlibcontext.Context, network, addr st
 			dialingFailed: true, // indicate error happened before http
 		}
 	} else if cerr := ctx.Err(); cerr != nil {
-		// deadline exceeded or canceled in stdlib
+		// deadline exceeded or canceled in stdlib, client closed request
+		// see https://github.com/zalando/skipper/issues/687#issuecomment-405557503
 		return nil, &proxyError{
 			err:  cerr,
-			code: http.StatusGatewayTimeout,
+			code: 499,
```
aryszka: this is not a standard code. Do we have other options?

aermakov-zalando: This is used by nginx (and I think apache, but not sure).

aryszka: maybe 0? indicating that the request was not served, therefore there is no status code at all?

mikkeloscar:
I don't see a 499 logged when running skipper build from this branch. I do as described in https://github.com/zalando/skipper/issues/687#issuecomment-405557503 (Thanks for a very good description by the way).

I only see 499 if I replace the 504 here: https://github.com/zalando/skipper/blob/e9c64ec7e2a8de4ad85f7163807683c446ba80f5/proxy/proxy.go#L730

proxy/proxy.go
szuecs: 0 is specially handled in skipper and returns in the end 500 @aryszka

szuecs:
@aermakov-zalando maybe my message was missleading, but it's exactly client cacneled the request, see https://github.com/zalando/skipper/issues/687#issuecomment-405557503

@mikkeloscar gotcha, thanks for testing!

szuecs:
I fixed the issue reported by @mikkeloscar.
I additionally cleaned up the logging and enhanced the error message: dropped duplicates and added the backend to the log, which was an addition from the middle log entry.

proxy:
```
 % ./bin/skipper --address :9999 -inline-routes 'p0: * -> "http://127.0.0.1:9090/"'
[APP]INFO[0000] proxy listener on :9999
[APP]INFO[0000] route settings, reset, route: p0: * -> "http://127.0.0.1:9090/"
[APP]ERRO[0007] error while proxying, route p0 with backend http://127.0.0.1:9090/, status code 499: dialing failed false: context canceled
::1 - - [17/Jul/2018:18:28:36 +0200] "GET / HTTP/1.1" 499 1 "-" "curl/7.49.0" 1000 localhost:9999 - -
::1 - - [17/Jul/2018:18:28:35 +0200] "GET / HTTP/1.1" 200 14 "-" "curl/7.49.0" 5001 localhost:9999 - -
```

backend:
```
% ./bin/skipper -inline-routes 'p0: * -> latency(5000) -> inlineContent("<h1>Hello</h1>") -> <shunt>'
[APP]INFO[0000] route settings, reset, route: p0: * -> latency(5000) -> inlineContent("<h1>Hello</h1>") -> <shunt>


127.0.0.1 - - [17/Jul/2018:18:28:35 +0200] "GET / HTTP/1.1" 200 14 "-" "curl/7.49.0" 5000 127.0.0.1:9090 - -
127.0.0.1 - - [17/Jul/2018:18:28:36 +0200] "GET / HTTP/1.1" 200 14 "-" "curl/7.49.0" 5000 127.0.0.1:9090 - -
```

client calls:
```
% curl localhost:9999/ &
[1] 29258
[sszuecs@sandor-lab:~]% curl -m 1 localhost:9999/
curl: (28) Operation timed out after 1000 milliseconds with 0 bytes received
zsh: exit 28    curl -m 1 localhost:9999/
[sszuecs@sandor-lab:~]% <h1>Hello</h1>
[1]  + done       curl localhost:9999/
```

proxy/proxy.go
```diff
@@ -408,10 +409,11 @@ func (dc *skipperDialer) DialContext(ctx stdlibcontext.Context, network, addr st
 			dialingFailed: true, // indicate error happened before http
 		}
 	} else if cerr := ctx.Err(); cerr != nil {
-		// deadline exceeded or canceled in stdlib
+		// deadline exceeded or canceled in stdlib, client closed request
+		// see https://github.com/zalando/skipper/issues/687#issuecomment-405557503
 		return nil, &proxyError{
-			err:  cerr,
-			code: http.StatusGatewayTimeout,
+			err:  fmt.Errorf("err from dial contrext: %v", cerr),
```
mikkeloscar: `contrext` -> `context`.

proxy/proxy.go
```diff
@@ -408,10 +409,11 @@ func (dc *skipperDialer) DialContext(ctx stdlibcontext.Context, network, addr st
...
+			code: 499,
```
mikkeloscar: Are we sure this is always caused by client closing the connection?

I'm just afraid that we are masking some `http: server closed idle connection` errors as mentioned in https://github.com/zalando/skipper/issues/687 but I'm also not sure where they came from.

szuecs:
fixed comments and squashed commits.

The only open question is if 499 is fine or someone disagrees. 

My reasoning:
The error is raised, because the client has a timeout. The proxy did not timedout and get's the value from the backend and will respond to the client which already send a packet with a FIN flag (socket half closed), which will result to the client will respond the proxy response with a TCP RST packet, because it already closed the socket from the client and the kernel will send a RST packet. From the proxy point of view everything worked fine, but the client hang up. For me this is a 4xx and since nginx uses 499 I would propose to reuse their http status code, because having another value does not makes sense to me and probably some folks are already familiar of the meaning of 499.

proxy/proxy.go
```diff
@@ -723,11 +724,11 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
...
 			}
 		}
-		p.log.Errorf("error during backend roundtrip: %s: %v", ctx.route.Id, err)
 
 		if cerr := req.Context().Err(); cerr != nil {
-			p.log.Errorf("Failed to do request, because of context: %v", cerr)
-			return nil, &proxyError{err: cerr, code: http.StatusGatewayTimeout}
+			// deadline exceeded or canceled in stdlib, client closed request
+			// see https://github.com/zalando/skipper/issues/687#issuecomment-405557503
+			return nil, &proxyError{err: cerr, code: 499}
```
aryszka: if this is a request that the client has already cancelled, then this status code is used only for logging. Is this correct? If yes, then i'd prefer leaving the status code 0, because it is something that we use only for logging.

mikkeloscar: It's also added to the metrics, so that's why a 5xx is a problem because it gives the wrong impression of the service error rate.

Counting this as a 4xx seems right because we can exclude it from errors caused by the proxy/backend, but we can still monitor the 4xx rate and be aware of the error rate.

It would be interesting to know what AWS ELBs report in this case.

szuecs: Do you think the AWS ELB is better for the status code handling then nginx?
I mean AWS ELB also have a very different way of handling X-For-* headers and set the clientip last instead of all others that put this first as shown in an example of the RFC.

mikkeloscar: > Do you think the AWS ELB is better for the status code handling then nginx?

AWS ELB is just interesting because this is what our users are used to.

> For me the biggest drawback to status code 0 is that:

Note that status code `0` will be reported as `500` in logs and metrics. This is my biggest concern because it will give the wrong impression to users.

I agree with the benefit of `499` being fairly simple to search for.

aryszka: i got convinced, thanks for explaining it. Let's use 499.

aryszka:
:+1:

mikkeloscar:
:+1:

PR #710, opened 2018-07-20
add predicates to match based on jwt content

szuecs:
JwtAnyKV("iss", "https://accounts.google.com")
JwtAllKV("iss", "https://accounts.google.com", "email", "skipper-router@googlegroups.com")

### DIFF
--- a/predicates/auth/jwt.go
+++ b/predicates/auth/jwt.go
@@ -1,0 +1,182 @@
+/*
+Package auth implements custom predicates to match based on content
+of the HTTP Authorization header.
+
+This predicate can be used to match a route based on data in the 2nd
+part of a JWT token, for example based on the issuer.
+
+Examples:
+
+    // one key value pair has to match
+    example1: JWTPayloadAnyKV("iss", "https://accounts.google.com", "email", "skipper-router@googlegroups.com")
+	-> "http://example.org/";
+    // all key value pairs have to match
+    example2: * && JWTPayloadAllKV("iss", "https://accounts.google.com", "email", "skipper-router@googlegroups.com")
+	-> "http://example.org/";
+*/
+package auth
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"net/http"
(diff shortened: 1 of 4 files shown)


szuecs:
:+1:

mikkeloscar:
@szuecs regarding the naming, would it make sense to include `Payload` in the name, because this is what it looks into afaict.? I use the term payload because this is what's used here: https://jwt.io/

njuettner:
@szuecs JWT should be written in uppercase WDYT?

szuecs:
@mikkeloscar @njuettner addressed your comments

szuecs:
:+1:

mikkeloscar:
:+1:

predicates/auth/jwt.go
```diff
@@ -0,0 +1,183 @@
+/*
+Package auth implements custom predicates to match based on content
+of the HTTP Authorization header.
+
+You can use this predicate to match a route based on data in the 2nd part of a JWT token, for example based on the issuer.
```
aryszka: I think that it is nicer to use the passive form of "This predicate can be used..." in code docs.

predicates/auth/jwt.go
```diff
@@ -0,0 +1,183 @@
...
+
+	"github.com/zalando/skipper/routing"
+)
+
+const (
+	authHeaderName   = "Authorization"
+	authHeaderPrefix = "Bearer "
+)
+
+var InvalidArgsError = errors.New("invalid arguments")
```
aryszka: errors should be called conventionally ErrSomething. -> ErrInvalidArgsError. For some reason we also have this error here: https://github.com/zalando/skipper/blob/master/predicates/predicates.go#L6

aryszka:
:+1:

aryszka:
formatting fails:

```
make[1]: *** [check-fmt] Error 1
```

aryszka:
:+1:

PR #711, opened 2018-07-20
Fix custom routes for multiple hosts

mikkeloscar:
This fixes an issue, where given an ingress with the custom routes
annotation `zalando.org/skipper-routes` and multiple rules defined, the
route would only be created for the last rule.

I.e. the following ingress:

```yaml
kind: Ingress
metadata:
  annotations:
    zalando.org/skipper-routes: |
      Method("OPTIONS") -> status(200) -> <shunt>
  name: app
spec:
  rules:
  - host: app-default.example.org
    http:
      paths:
      - backend:
          serviceName: app-svc
          servicePort: 80
  - host: app-alt.example.org
    http:
      paths:
      - backend:
          serviceName: app-svc
          servicePort: 80
```

Would produce the custom route:

```
Host(/^app-alt[.]example[.]org$/) && Method("OPTIONS") ->
  status(200) -> <shunt>
```

When it should create routes for both host names:

```
1: Host(/^app-default[.]example[.]org$/) && Method("OPTIONS") ->
  status(200) -> <shunt>;
2: Host(/^app-alt[.]example[.]org$/) && Method("OPTIONS") ->
  status(200) -> <shunt>;
```

The fix is to copy the base route instead of overwriting the same route
pointer.

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -905,10 +905,11 @@ func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
 			host := []string{"^" + strings.Replace(rule.Host, ".", "[.]", -1) + "$"}
 
 			// add extra routes from optional annotation
-			for idx, route := range extraRoutes {
+			for idx, r := range extraRoutes {
+				route := *r
 				route.HostRegexps = host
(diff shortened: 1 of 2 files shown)


hjacobs:
@mikkeloscar can you add a test please? This sounds like a perfect opportunity to increase Skipper's test suite :smirk:

mikkeloscar:
Added a test case which shows it's fixed :)

szuecs:
:+1:

mikkeloscar:
:+1:

PR #715, opened 2018-07-23
Improve the routing table information documentation

### LINKED ISSUE #713: Routes endpoint not giving every route loaded
We're using Skipper on our k8s cluster. It's loading automatically all ingresses and converting them to it's own routes.

We sometimes need to get all the routes loaded. We do that executing a curl inside the pod:
```
curl "http://localhost:9911/routes"
```

The problem we're facing is that even though this call returns a list of routes, it's not complete. If we check the number of lines, it's a changing number near to 4300 lines.

Is there any way of not limiting this response?

jchanam:
This PR adds information about getting the routes from Skipper. It explains the limit of routes given by default and how to configure it.

It fixes #713

### DIFF
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -379,7 +379,7 @@ of the last route table update `X-Timestamp`, using a HEAD request:
     Date: Sun, 04 Feb 2018 20:54:31 GMT
 
 The number of routes given is limited (1024 routes by default).
-In order to control this limits, there are two filters: `limit` and
+In order to control this limits, there are two parameters: `limit` and
 `offset`. The `limit` defines the number of routes to get and
 `offset` where to start the list. Thanks to this, it's possible
 to get the results paginated or getting all of them at the same time.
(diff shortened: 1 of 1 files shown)


docs/operations.md
```diff
@@ -377,3 +377,11 @@ of the last route table update `X-Timestamp`, using a HEAD request:
     X-Count: 1
     X-Timestamp: 1517777628
     Date: Sun, 04 Feb 2018 20:54:31 GMT
+
+The number of routes given is limited (1024 routes by default).
+In order to control this limits, there are two filters: `limit` and
```
aryszka: what do you think of using the wording "parameter" or "query param" instead of "filter" in order to avoid overloading the word "filter", since it is used very often in skipper for something very different?

aryszka:
:+1:

aryszka:
:+1:

hjacobs:
:+1:

PR #717, opened 2018-07-23
fix #714 ingress backends same svc name but different ports

### LINKED ISSUE #714: ingress backends can not have different ports from one service
```
spec:
  rules:
  - host: foo.example.org
    http:
      paths:
      - backend:
          serviceName: foo
          servicePort: 8080
        path: /
  - host: bar.example.org
    http:
      paths:
      - backend:
          serviceName: foo
          servicePort: 8081   // <-- skipper will not find this and create routes with 8080 for it
        path: /
```

My first assumption is that https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/kube.go#L644 is limiting it and we should add the port, but I have to check it.
We also have the comment https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/kube.go#L894 , which might did this on purpose. I remember that we had the problem if a servuce has 2 ports but in ingress we only use one it would find both and create 50% wrong routes with this.

szuecs:
fix #714 ingress backends can not have different ports from one service

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -641,7 +641,6 @@ func (c *Client) convertPathRule(
 		return nil, fmt.Errorf("invalid path rule, missing backend in: %s/%s/%s", ns, name, host)
 	}
 
-	endpointKey := ns + prule.Backend.ServiceName
 	var (
 		eps    []string
 		err    error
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

PR #722, opened 2018-07-25
Add -histogram-metric-buckets option

### LINKED ISSUE #721: support custom Prometheus buckets
The default bucket setup is not very fine granular and result in p99 values of 10s kept for example.
We should support to choose a different bucket setup then the default, configurable by CLI flag.

Default in prometheus client go library:

	DefBuckets = []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}

Can be changed by `HistogramOpts.Buckets` option in all `NewHistogramVec` calls

aermakov-zalando:
If set, use a custom set of buckets for Prometheus histogram metrics instead of the default one.

Example: 

    ./bin/skipper -enable-prometheus-metrics -histogram-metric-buckets=.005,.01,.02,.03,.04,.05,.06,.07,.08,.09,.1,.2,.3,.4,.5,.6,.7,.8,.9,1,1.5,2,2.5,3,3.5,4,4.5,5,6,7,8,9,10,15,20,25,30,40,50,60,70,80,90,100,125,150,200,275,425,600,1000

Fixes #721.

### DIFF
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -19,10 +19,12 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
 	"github.com/zalando/skipper"
 	"github.com/zalando/skipper/dataclients/kubernetes"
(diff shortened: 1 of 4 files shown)


szuecs:
:+1:

aryszka:
:+1:

Issue #723, opened 2018-07-25
setQuery(string)


szuecs:
It should be possible to set just a key in the query.
 setQuery("foo") will result in "?foo", which is often used in old WSDL implementations.

PR #724, opened 2018-07-25
add setQuery(string) #723

szuecs:
needed for some old application that do not support ?foo=, but want to have ?foo


szuecs:
:+1:

aryszka:
:+1:

PR #726, opened 2018-07-26
Adds Gauges to metrics

topicus:
- Add Gauges and their corresponding tests

### DIFF
--- a/metrics/codahale.go
+++ b/metrics/codahale.go
@@ -44,6 +44,7 @@ type CodaHale struct {
 	reg           metrics.Registry
 	createTimer   func() metrics.Timer
 	createCounter func() metrics.Counter
+	createGauge   func() metrics.GaugeFloat64
 	options       Options
 	handler       http.Handler
 }
@@ -64,6 +65,7 @@ func NewCodaHale(o Options) *CodaHale {
 	c.createTimer = func() metrics.Timer { return createTimer(createSample()) }
 
 	c.createCounter = metrics.NewCounter
+	c.createGauge = metrics.NewGaugeFloat64
 	c.options = o
 
 	if o.EnableDebugGcMetrics {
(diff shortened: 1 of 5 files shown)


aryszka:
:+1:

aryszka:
Is it possible to apply these changes for the prometheus variant as well? If yes, then it would also be possible to declare these methods on `interface Metrics`, and they would become available in the filters as custom metrics. It's also for consistency.

topicus:
I tried to make it for both but prometheus doesn't provide Meters. I can remove Meters from this PR and implement the prometheus version of Gauges.

I think this can provide a good balance between adding value and keeping consistency.

aryszka:
It would be really nice to have them, or at least the Gauges on the shared Metrics interface. Do you think it is better to do it as part of this PR, or is it more reasonable to do it as a follow-up?

topicus:
I have updated the PR. It includes the Gauges implementation for Prometheus and CodeHale with their respective tests.

Later we can figure out how to implement Meters in a way that works for both clients.

metrics/codahale.go
```diff
@@ -282,6 +295,13 @@ func (sm skipperMetrics) MarshalJSON() ([]byte, error) {
 		case metrics.Gauge:
 			metricsFamily = "gauges"
 			values["value"] = m.Value()
+		case metrics.Meter:
```
szuecs: this should be dropped for this PR, right?

topicus:
@szuecs I fixed those items. Could you review again?

aryszka:
:+1:

topicus:
:+1:

szuecs:
@topicus can you also add some documentation in https://github.com/zalando/skipper/blob/master/docs/operations.md#monitoring please

topicus:
:+1:

szuecs:
@topicus please do https://github.com/zalando/skipper/pull/726#issuecomment-408409361, before merge, thanks!

topicus:
@aryszka  @szuecs Is there any section like "extending skipper" or similar? All the documentation I found is about using skipper out-of-the-box except by the plugins section.

szuecs:
@topicus I am not sure what you mean by that.
I you like to chat about that, please join our hipchat channel, I guess it'sbetter to discuss it and document if there is a need to.

szuecs:
As discussed offline there are no additional metrics, only a facilty to add Gauges if skipper is used as library.

szuecs:
:+1:

PR #733, opened 2018-07-31
Fix Metrics interface

topicus:
- Fix Metrics struct
- Use snapshot before get Gauge value
- Fix Gauge type during marshalling

### DIFF
--- a/metrics/codahale.go
+++ b/metrics/codahale.go
@@ -294,6 +294,8 @@ func (sm skipperMetrics) MarshalJSON() ([]byte, error) {
 
 		switch m := metric.(type) {
 		case metrics.Gauge:
+			metricsFamily = "gauges"
+			values["value"] = m.Value()
 		case metrics.GaugeFloat64:
 			t := m.Snapshot()
 			metricsFamily = "gauges"
--- a/metrics/codahale.go
+++ b/metrics/codahale.go
@@ -291,10 +291,13 @@ func (sm skipperMetrics) MarshalJSON() ([]byte, error) {
 	for name, metric := range sm {
 		values := make(map[string]interface{})
 		var metricsFamily string
+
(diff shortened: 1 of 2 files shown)


metrics/metrics.go
```diff
@@ -69,6 +69,7 @@ type Metrics interface {
 	MeasureBackend5xx(t time.Time)
 	IncErrorsStreaming(routeId string)
 	RegisterHandler(path string, handler *http.ServeMux)
+	UpdateGauge(key string, value float64)
```
szuecs: this will break prometheus provider, please implement a dummy in the prometheus provider, too

topicus: Good catch!

topicus: Sorry I got confused. It is already implemented https://github.com/zalando/skipper/blob/master/metrics/prometheus.go#L297

metrics/codahale.go
```diff
@@ -291,10 +291,15 @@ func (sm skipperMetrics) MarshalJSON() ([]byte, error) {
 	for name, metric := range sm {
 		values := make(map[string]interface{})
 		var metricsFamily string
+
 		switch m := metric.(type) {
 		case metrics.Gauge:
 			metricsFamily = "gauges"
 			values["value"] = m.Value()
+		case metrics.GaugeFloat64:
```
addityasingh: Does it make sense to also add a test similar to this https://github.com/zalando/skipper/blob/master/metrics/prometheus_test.go#L695?

szuecs:
:+1:

metrics/codahale.go
szuecs: It’s not needed as long as you implemented the interface and you did, so it is fine.
Thanks

aryszka:
:+1:

Issue #739, opened 2018-08-14
oauthTokeninfo* memory leak


szuecs:
If you heavily use oauthTokeninfoAllScope("uid"), we see unexpected memory growth, that will not stop.
A test with about 150 ingress and this filter configuration you can see the problem: 
 
![image](https://user-images.githubusercontent.com/50872/44104387-0812c83a-9fef-11e8-8a18-5ec6f0870927.png)

Without this particular filter everything is fine.

PR #740, opened 2018-08-15
cache authclient per tokeninfoURL

szuecs:
#739 tokeninfo filter should cache authclient per tokeninfoURL and reuse it instead of creating one for each filter

Right now a hashmap for the cache is not really needed, but we can easily create filters, that can pass the target tokeninfoURL to be able to change this per route.

```
% go test -benchmem -bench=BenchmarkOAuthTokeninfoFilter -run='^$' ./filters/auth/ -v
goos: darwin
goarch: amd64
pkg: github.com/zalando/skipper/filters/auth
BenchmarkOAuthTokeninfoFilter-8           300000              4481 ns/op            1739 B/op         17 allocs/op
PASS
ok      github.com/zalando/skipper/filters/auth 1.499s
% go test -benchmem -bench=BenchmarkOAuthTokeninfoFilter -run='^$' ./filters/auth/ -v
goos: darwin
goarch: amd64
pkg: github.com/zalando/skipper/filters/auth
BenchmarkOAuthTokeninfoFilter-8          5000000               412 ns/op             219 B/op          5 allocs/op
PASS
ok      github.com/zalando/skipper/filters/auth 2.514s
```

### DIFF
--- a/filters/auth/tokeninfo.go
+++ b/filters/auth/tokeninfo.go
@@ -33,6 +33,8 @@ type (
 	}
 )
 
+var tokeninfoAuthClient map[string]*authClient = make(map[string]*authClient)
+
 // NewOAuthTokeninfoAllScope creates a new auth filter specification
 // to validate authorization for requests. Current implementation uses
 // Bearer tokens to authorize requests and checks that the token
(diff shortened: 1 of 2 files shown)


szuecs:
I tried this in a test cluster and you can see it fixed the leak:

image

szuecs:
:+1:

mikkeloscar:
:+1:

PR #742, opened 2018-08-16
Simplify CDP build with golang overlay

mikkeloscar:
CDP now has an 'official' Golang overlay which simplifies the build a lot. Use that instead of installing Go manually.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #747, opened 2018-08-22
implment a simple webhook

szuecs:
This implemetns a simple webhook filter, that clones all headers from the original request and send it to the webhook, which returns with either >= 400 status code or error to be unauthorized otherwise authorized.
 
missing:

- [ ] docs
- [ ] tests

### DIFF
--- a/filters/auth/webhook.go
+++ b/filters/auth/webhook.go
@@ -1,0 +1,70 @@
+package auth
+
+import (
+	"net/http"
+
+	"github.com/zalando/skipper/filters"
+)
+
+const WebhookName = "webhook"
+
+type (
+	webhookSpec   struct{}
+	webhookFilter struct {
+		authClient *authClient
+	}
+)
+
+// NewWebhook creates a new auth filter specification
+// to validate authorization for requests.
+func NewWebhook() filters.Spec {
+	return &webhookSpec{}
+}
+
+func (s *webhookSpec) Name() string {
+	return WebhookName
+}
+
+// CreateFilter creates an auth filter. The first argument is an URL
+// string.
+//
+//     s.CreateFilter("https://my-auth-service.example.org/auth")
+//
+func (s *webhookSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if l := len(args); l == 0 || l > 2 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	s, ok := args[0].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	ac, err := newAuthClient(s)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


filters/auth/webhook.go
```diff
@@ -0,0 +1,70 @@
+package auth
+
+import (
+	"net/http"
+
+	"github.com/zalando/skipper/filters"
+)
+
+const WebhookName = "webhook"
```
aryszka: less is more, how about just simply "hook"?

filters/auth/webhook.go
```diff
@@ -0,0 +1,70 @@
+package auth
```
aryszka: i think the hook should be a completely independent filter.

filters/auth/webhook.go
aryszka: webhook is also fine, of course

filters/auth/webhook.go
szuecs: I disagree, because it only makes sense for auth. As we discussed in the chat you also had no use case for non IAM webhooks.

szuecs:
Reday to be reviewed

szuecs:
:+1:

docs/filters.md
```diff
@@ -382,6 +382,24 @@ basicAuth("/path/to/htpasswd")
...
+is being done to the webhook endpoint. Responses from the webhook with
+status code less than 300 will be authorized, rest unauthorized.
+
+Examples:
+
+```
+webhook("https://custom-webhook.example.org/auth")
+```
+
+The webhook timeout has a default of 2 seconds and can be globally
```
addityasingh: Why not also specifying the timeout as an optional second param to the filter?

szuecs: From parameter point of view it's fine, but from client connection pool you would have per `webhook * timeout` times connection pools. We would rather like to have less connection pools, than more.

:)

filters/auth/webhook_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+				tok := r.Header.Get(authHeaderName)
+				tok = tok[len(authHeaderPrefix):len(tok)]
+				switch tok {
+				case testToken:
+					w.WriteHeader(200)
+					fmt.Fprintln(w, "OK - Got token: "+tok)
+					return
+				}
+				w.WriteHeader(402)                            //http.StatusUnauthorized)
```
linki: this seems to be a leftover

filters/auth/webhook.go
aryszka: +1 we agreed not to make it generic

aryszka:
:+1:

szuecs:
:+1:

aryszka:
:+1:

PR #752, opened 2018-08-24
Feature/access log filter

### LINKED ISSUE #750: Kubernetes ingress: allow to reduce request logging
Currently, ingress skipper produces log entry for each request and for some of the services we want to be able to switch off extensive request logging.

vecmezoni:
Fixes #750

* Removed access log http handler and moved logic to proxy 
* Added `accessLog` filter to give chance to override global AccessLogDisabled setting


vecmezoni:
Can be considered just as proof of concept because I'm not sure that `proxy` is a good place for it.

aryszka:
i like the idea, it's aligned with other features like the preserve host.

please, use `make fmt` before pushing. It's makes the build currently failing.

vecmezoni:
@aryszka done

filters/accesslog/log_test.go
```diff
@@ -0,0 +1,25 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var ctx filtertest.Context
+	ctx.FStateBag = make(map[string]interface{})
+
+	f.Request(&ctx)
+	bag := ctx.StateBag()
+	if bag[AccessLogEnabledKey] != false {
```
szuecs: can you also test the negative?
This test would also be fine if this is the default.

filters/accesslog/log.go
```diff
@@ -0,0 +1,50 @@
...
+
+type accessLog struct {
+	status bool
+}
+
+// NewAccessLog creates a filter spec for overriding the state of the access log. (By default global setting is used.)
+//
+// 	accessLog("false")
+//
+// EXPERIMENTAL: this filter is experimental, and the name and the arguments can change until marked as stable.
```
aryszka: I would go as far as to not even mark it as experimental. @vecmezoni @szuecs , do you foresee any reason to change the behavior/name/args in the future?

szuecs: I don't see it no.

vecmezoni: I would think of a more generic approach when you can even provide log format here. Maybe it can be done with the separate filter.

filters/accesslog/log_test.go
vecmezoni: sure, I'll update the test and also add the docs as you mentioned.

vecmezoni:
@szuecs @aryszka 
So, I've renamed `accessLog` to `accessLogDisabled` to follow naming convention for Skipper setting `access-log-disabled` and remove possible confusion.
Added more tests and documentation.

filters/accesslog/disable.go
```diff
@@ -0,0 +1,45 @@
...
+//
+// 	accessLogDisabled("false")
+func NewAccessLogDisabled() filters.Spec {
+	return &accessLogDisabled{}
+}
+
+func (al *accessLogDisabled) Name() string { return AccessLogDisabledName }
+
+func (al *accessLogDisabled) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) < 1 {
```
szuecs: `!= 1` shoud be used here

filters/accesslog/disable.go
```diff
@@ -0,0 +1,45 @@
...
+		return nil, filters.ErrInvalidFilterParameters
+	}
+}
+
+func (al *accessLogDisabled) Request(ctx filters.FilterContext) {
+	bag := ctx.StateBag()
+	bag[AccessLogDisabledKey] = al.disabled
+}
+
+func (al *accessLogDisabled) Response(_ filters.FilterContext) {}
```
szuecs: I think this can be used: `func (*accessLogDisabled) Response(filters.FilterContext) {}`

filters/accesslog/disable_test.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+}
+
+func TestUnknownValue(t *testing.T) {
+	const state = "unknownValue"
+
+	_, err := NewAccessLogDisabled().CreateFilter([]interface{}{state})
+
+	if err == nil || err != filters.ErrInvalidFilterParameters {
+		t.Error("should throw error on unknown value")
```
szuecs: throw -> return

filters/accesslog/disable_test.go
```diff
@@ -0,0 +1,62 @@
...
+	bag := ctx.StateBag()
+	if bag[AccessLogDisabledKey] != true {
+		t.Error("failed to set access log state")
+	}
+}
+
+func TestUnknownValue(t *testing.T) {
+	const state = "unknownValue"
+
+	_, err := NewAccessLogDisabled().CreateFilter([]interface{}{state})
```
szuecs:
```
if _, err := NewAccessLogDisabled().CreateFilter([]interface{}{state}); err == nil || err != filters.ErrInvalidFilterParameters {
...
}
```

filters/accesslog/disable_test.go
```diff
@@ -0,0 +1,62 @@
...
+	ctx.FStateBag = make(map[string]interface{})
+
+	f.Request(&ctx)
+	bag := ctx.StateBag()
+	if bag[AccessLogDisabledKey] != false {
+		t.Error("failed to set access log state")
+	}
+}
+
+func TestDisabled(t *testing.T) {
```
szuecs: If you like you can rewrite all these tests to table driven tests.
https://opensource.zalando.com/skipper/development/#writing-tests shows how to do table driven tests, which will shorten the code.

readme.md
```diff
@@ -100,10 +100,10 @@ Getting the code with the test dependencies (`-t` switch):
...
 Build and test all packages:
 
-  cd src/github.com/zalando/skipper
-  make deps
-  make install
-  make shortcheck
+    cd src/github.com/zalando/skipper
+    make deps
+    make install
+    make shortcheck
```
szuecs: please drop these changes

vecmezoni: But it makes it look better:

default

filters/accesslog/disable.go
vecmezoni: 👍

filters/accesslog/disable.go
vecmezoni: 👍

filters/accesslog/disable_test.go
vecmezoni: 👍

filters/accesslog/disable_test.go
vecmezoni: made table tests, not needed anymore

filters/accesslog/disable_test.go
vecmezoni: 👍

vecmezoni:
@szuecs done

readme.md
szuecs: Oh, I did not know, that it's not correctly rendered right now. :)

szuecs:
👍

vecmezoni:
👍

aryszka:
:+1:

mo-gr:
I know I'm too late to the party. But boolean flags are an interface smell. And abusing strings as boolean flags doesn't make the smell nicer. Have you considered to just turn this into two filters `accessLogDisabled()` and `accessLogEnabled()`? Also to avoid the unnecessary double negation of `accessLogDisabled("false")`...

aryszka:
retrospectively, i agree. Didn't think of this when discussed the approach. Neither in case of this one https://godoc.org/github.com/zalando/skipper/filters/builtin#PreserveHost , long time ago.

vecmezoni:
Unfortunatelly I haven’t considered this option. We can fix it until this gets wide usage. 

I was trying to follow configuration parameter name to remove confusion.

szuecs:
@vecmezoni deprecate it and recreate the filter and split it into 2.

szuecs:
Later we can drop everything we deprecate without breaking

aryszka:
we can use new names: disableAccessLog() and enableAccessLog() for the new split filters, so that it doesn't conflict with the deprecated one.

tkrop:
To late for the party, too. How about substituting the *true|false* concept against a real access log configuration, that allows to enable, disable, and in limits add information, e.g. a specific header. Disabling than could be expressed by an *empty|nil* configuration.

ruiaraujo:
@tkrop You should comment on the new PR

PR #756, opened 2018-08-29
add RequestContext(ctx context.Context) to filtercontext

vetinari:
this allows filters to set the context of the request. This is needed when e.g. an opentracing span
should be set as new current span and this span is then retrieved at a later point as parent span

### DIFF
--- a/filters/filters.go
+++ b/filters/filters.go
@@ -21,7 +21,7 @@ type FilterContext interface {
 	Request() *http.Request
 
 	// Updates the context in the Request
-	RequestContext(context.Context)
+	SetRequestContext(context.Context)
 
 	// The response object. It is returned to the client with its
 	// properties changed by the filters.
--- a/filters/filters.go
+++ b/filters/filters.go
@@ -1,6 +1,7 @@
 package filters
 
 import (
+	"context"
 	"errors"
 	"net/http"
 	"time"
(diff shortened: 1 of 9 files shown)


proxy/context.go
```diff
@@ -8,6 +8,8 @@ import (
 	"net/url"
 	"time"
 
+	corectx "context"
```
szuecs: stdlibctx

szuecs:
@vetinari please add:
- [ ] a test for a filter, for example modifying tokeninfo filter to pass context to authclient calls.
- [ ] user documentation in docs/developer.md
- [ ] developer documentation as godoc

filters/filters.go
```diff
@@ -19,6 +20,9 @@ type FilterContext interface {
 	// with its properties changed by the filters.
 	Request() *http.Request
 
+	// Updates the context in the Request
+	RequestContext(context.Context)
```
aryszka: I would call this `SetRequestContext` to indicate the direction.

szuecs:
please remove the label if you did all open tasks, thanks

aryszka:
@vetinari is there any update on this PR, is it ready for review and merge?

szuecs:
We already have this instrumented https://github.com/zalando/skipper/blob/master/filters/auth/authclient.go#L70-L77 and no follow up since 8 months, so close.

PR #761, opened 2018-08-30
Build the lightstep image with ubuntu base image.

arjunrn:
The lightstep image needs libc to load plugins which is missing from the standard skipper Docker image. This PR changes the dockerfile to use a ubuntu base image.

### DIFF
--- a/packaging/Dockerfile.lightstep
+++ b/packaging/Dockerfile.lightstep
@@ -1,5 +1,7 @@
-ARG BASE_IMAGE
-FROM $BASE_IMAGE
+FROM registry.opensource.zalan.do/stups/ubuntu:latest
+MAINTAINER Skipper Maintainers <team-pathfinder@zalando.de>
+ADD skipper eskip /usr/bin/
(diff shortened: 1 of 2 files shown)


mikkeloscar:
I would assume it only needs libc because it's being linked against it on the Ubuntu build VM. If we instead build it in a multistage alpine based Dockerfile, then we can keep using the smaller alpine image.

arjunrn:
@mikkeloscar Already tried something like that, didn't work. I believe that the go plugins need to be build with `CGO_ENABLED=1` and then the plugins are loaded by the main skipper binary and this steps needs some(not all) libraries installed. These libraries are not present on the standard `alpine` image. The other option is to extend the alpine image with only the libraries needed by the `skipper` binary but I'm not sure about this. @szuecs maybe you have some suggestions.

mikkeloscar:
@arjunrn @szuecs fine from my side :+1: I might take a look at using alpine, because I'm interested in how the pluginsin go work anyway. :)

PR #764, opened 2018-09-02
Use Go 1.11 modules

### LINKED ISSUE #737: Use go1.11 for vendoring and go modules, once stable
I have been trying about go modules which is expected to be part of soon to be released go1.11 version. Does it makes sense to try using `go mod` using the mod files or still wait for this version to be mature?

### LINKED ISSUE #763: Switch to Go 1.11 modules (and streamline build)
Currently the skipper Makefiles consists of a mix of `go get` and `glide install --strip-vendor`. It's a bit confusing this way and hard to understand what version of packages skipper really is being build against. We also have some hacks in place to be able to build plugins: https://github.com/zalando/skipper/blob/5d7ae30e7f71a566b2562906f86959a6044ed319/Makefile#L89-L94

Since we have Go modules in Go 1.11 I would like to take this as an opportunity to streamline the skipper build/dependency management a bit.

What do you think? Does it make sense to try out Go Modules already or is it too early? It will put a hard depedency on Go 1.11 for skipper. IMO this is fine, and I believe Go Modules are much better than glide already!

/cc @szuecs @aryszka @vetinari

mikkeloscar:
Introduce Go 1.11 modules as a replacement for glide.

Fix #763
Fix #737


packaging/Makefile
```diff
@@ -26,21 +26,20 @@ docker-push:
 	docker push $(IMAGE)
 
 plugins:
-	CUR=$(shell pwd) ; GOPATH=$(shell pwd)/build ;\
-		go get -d github.com/skipper-plugins/opentracing ;\
-		cd build/src/github.com/skipper-plugins/opentracing ;\
-		glide install -v;\
+	mkdir -p build/; cd build ;\
+	git clone --branch=go-modules https://github.com/mikkeloscar/opentracing.git ;\
```
mikkeloscar: This should be changed once https://github.com/skipper-plugins/opentracing/pull/10 is merged.

szuecs: please change it

mikkeloscar: Done

szuecs:
:+1:

mikkeloscar:
:+1:

PR #766, opened 2018-09-03
Go1.11 modules enabled tracer build

szuecs:
This is how it works right now:
- build with alpine build image using Go1.11
- go.mod for skipper as library does not have lightstep
- ./cmd/go.mod has lightstep
-  ./opentracingplugin/ is a new folder to import the plugins into skipper repository, which is another package main to build the plugin with a separate go.mod
- beware to use: `opentracing="tracing_lightstep .."` not "lightstep"
``` 
[sszuecs@sandor-lab:~/go/src/github.com/zalando/skipper/packaging]% make docker-lightstep-build           
...
Successfully tagged registry-write.opensource.zalan.do/pathfinder/skipper-lightstep:1ba0430e56a51feb9d08061f776c49177d00710a

% docker run registry-write.opensource.zalan.do/pathfinder/skipper-lightstep:1ba0430e56a51feb9d08061f776c49177d00710a -opentracing="tracing_lightstep token=abc"
[APP]time="2018-09-03T16:32:14Z" level=info msg="found plugin tracing_lightstep at plugins/tracing_lightstep.so"
[APP]time="2018-09-03T16:32:14Z" level=info msg="found plugin tracing_lightstep at /plugins/tracing_lightstep.so"
[APP]time="2018-09-03T16:32:14Z" level=info msg="attempting to load plugin from /plugins/tracing_lightstep.so"
[APP]time="2018-09-03T16:32:14Z" level=warning msg="no route source specified"
[APP]time="2018-09-03T16:32:14Z" level=info msg="Expose metrics in codahale format"
[APP]time="2018-09-03T16:32:14Z" level=info msg="support listener on :9911"
[APP]time="2018-09-03T16:32:14Z" level=info msg="proxy listener on :9090"
[APP]time="2018-09-03T16:32:14Z" level=info msg="certPathTLS or keyPathTLS not found, defaulting to HTTP"
2018/09/03 16:32:15 LS Tracer error: (rpc error: code = NotFound desc = AccessToken abc not found). NOTE: Set the SetGlobalEventHandler handler to log events.
``` 


szuecs:
:+1:

mikkeloscar:
I'm wondering if it would be better to treat the whole repo as a single module i.e. one `go.mod` in the root. Otherwise we still have to keep two `go.mod` files in sync like when the repository was split.

mikkeloscar:
The problem is that Go modules map 1:1 to git repositories since they aim to depend on the tags as versions.

While it technically can work that all three modules are different but just happen to have the same version I don't think it's the right way to do it, and it complicates things because you need to align the versions used for all three modules. Instead of just using the same versions for all, defined in a single place.

szuecs:
@mikkeloscar I just pushed the one go.mod file solution. @arjunrn clarified how it works. It only uses the main go.mod file anyways, such that it makes sense to only have one.

szuecs:
:+1:

mikkeloscar:
@szuecs Great!

mikkeloscar:
:+1:

szuecs:
:+1:

PR #769, opened 2018-09-04
handle connection reset by peer 

### LINKED ISSUE #768: Random 500s on Ingress
We are seeing some 500s on the ingress level that do not appear on the application logs.

This is the log from the ingress.

```
{"log":"[APP]time=\"2018-09-04T15:54:54Z\" level=error msg=\"net.Error during backend roundtrip to http://10.2.38.130:3000: timeout=false temporary=true: read tcp 10.2.11.0:44362->10.2.38.130:3000: read: connection reset by peer\"\n","stream":"stderr","time":"2018-09-04T15:54:54.290610471Z"}
{"log":"[APP]time=\"2018-09-04T15:54:54Z\" level=error msg=\"error while proxying, route kube_default__REDACTED__lb_group with backend , status code 500: dialing failed false: read tcp 10.2.11.0:44362->10.2.38.130:3000: read: connection reset by peer\"\n","stream":"stderr","time":"2018-09-04T15:54:54.290631962Z"}
```

szuecs:
fix #768 handle connection reset by peer with a retry to another endpoint if it is loadbalanced


proxy/proxy.go
```diff
@@ -717,6 +718,16 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
...
 					code: http.StatusServiceUnavailable,
 				}
+			} else if !nerr.Timeout() && nerr.Temporary() {
+				ext.HTTPStatusCode.Set(ctx.proxySpan, uint16(http.StatusServiceUnavailable))
+				if strings.Contains(err.Error(), "read: connection reset by peer") {
+					// retrieable case https://github.com/zalando/skipper/issues/768
+					return nil, &proxyError{
+						err:           err,
+						code:          http.StatusServiceUnavailable,
+						dialingFailed: true,
```
ruiaraujo: Does this trigger a retry?

The application never received the request so this seems just like a better error code.

aryszka: it retries only if it is a "loadbalanced" route, with multiple available endpoints. Applies to Kubernetes ingress scenarios.

ruiaraujo: @aryszka got it, that is the case here so all good.

ruiaraujo:
:+1:

proxy/proxy.go
szuecs: I had to fix the fallback problem, because we would not return an error if it would bre no timeout and temporary without connection reset by peer...

szuecs:
:+1:

aryszka:
:+1:

aermakov-zalando:
This is potentially dangerous if one backend starts handling the request and then drops the connection. Or am I missing something?

PR #770, opened 2018-09-04
Speedup CI buildtime

mikkeloscar:
Speed up CI build time in two ways.

1. Download etcd binary instead of pulling all sources and building from source.
2. Reduce `delivery.yaml` configuration to a single build step with multiple commands. Each build step is run on a separate VM so this reduces the build time a lot because we build all images on a single VM and avoid pulling the same dependencies three times on three different build VMs.

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -15,12 +15,8 @@ pipeline:
       export IMAGE
       make deps shortcheck
       cd packaging && make docker-build docker-push
-- id: build-lightstep
-  overlay: ci/golang
-  type: script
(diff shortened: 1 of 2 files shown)


szuecs:
Can you also add that we have the auto generated tag populated, before merge to master build is triggered?
Then we don't need to retrigger the pipeline for a nice version.

szuecs:
@mikkeloscar we somehow have to move it from travis to the delivery.yaml as executed command after the successful build+test, before docker image creation.

szuecs:
👍

mikkeloscar:
:+1:

PR #775, opened 2018-09-07
change connection-reset-by-peer handling to be not retried

szuecs:
https://github.com/zalando/skipper/issues/768
fix connection-reset-by-peer handling to be not retried but logged with an error message to hint the user what the problem is


szuecs:
:+1:

PR #762, opened 2018-08-31
Build lightstep image based on alpine

mikkeloscar:
Build lightstep image based on alpine. This is done to reduce the image size. It's a reduction from `179MB` -> `45.6MB` compared to the ubuntu based image.

```
REPOSITORY                                                           TAG                                        IMAGE ID            CREATED             SIZE
registry.opensource.zalan.do/pathfinder/skipper-lightstep-test       pr-762-9                                   de824efae48c        4 minutes ago       45.6MB
registry.opensource.zalan.do/pathfinder/skipper-lightstep            v0.10.68                                   61a3b1c0a562        24 hours ago        179MB
```

It's implemented by adding new make targets `alpine.build` and `alpine.plugins` that builds skipper and plugins in a `golang:alpine` based image such that both skipper and the tracing plugin is linked against `musl` instead of `libc`.

Ref: https://github.com/zalando/skipper/pull/761

PR #779, opened 2018-09-08
Proposal: build opentracing tracers into main binary

mikkeloscar:
This is a proposal to change how we distribute tracers with skipper. I'm
creating the proposal in the form of a PR to show how I would change it.

### Problem statement

Currently we provide opentracing tracers by building separate plugins and
distributing those along with the main skipper binary such that the plugins can
be loaded at runtime if needed. This is implemented via [go
plugins](https://golang.org/pkg/plugin/) and the idea is to provide felxibility
for people wanting to use skipper e.g. with proprietary tracer plugins without
forking skipper. This however, also comes with a set of tradeoffs.

* Skipper binary and plugin must be built agains the same dependencies. This
    requires manual management of versions if the plugin lives outside of the
    skipper repository.
* Both binary and plugins must be built with `CGO_ENABLED=1` and against the same
  `libc` (or `musl`). I.e. the build must be done in an alpine container if you
  want to distribute the final skipper binary and plugins via alpine to save on
  the total image size (#762).
* By building skipper and plugins as separate binaries the total
    artifact size about doubles.

    Size of skipper and `lightstep` plugin linked against libc on my machine:

    ```
    $ du -hs packaging/skipper
    15M	bin/skipper
    $ du -hs packaging/build/tracing_lightstep.so
    17M	build/tracing_lightstep.so
    ```

    vs. statically linked skipper

    ```
    $ du -hs bin/skipper
    17M	bin/skipper
    ```
* We build 3 different images to provide a slimmed down version for those not
    needing opentracing support i.e. `skipper`,
    `skipper-lightstep` and `skipper-opentracing`.

#### Solution #1

Instead of building the tracers as plugins I propose we just include them in
the main skipper binary and accept the few additional MBs.

```
$ du -hs bin/skipper
22M	bin/skipper
```

This increases the skipper binary with `~29.5%` compared to being build without
opentracing support.
But reduces the size by `~58.8%` compared to `skipper` + `lightstep`.

##### Pros

* Single, **statically linked**, binary with everything included.
  * No need to build in a custom alpine docker image.
* One skipper docker image with everything.
* Size of skipper image with e.g. lightstep support would be smaller.

##### Cons

* Running skipper with a proprietary tracer would require maintaining a
    fork/importing skipper as library.
* The default docker image will increase with about 5 MB.

This PR implements this solution.

#### Solution #2

Similar to **Solution #1** but using [build tags](https://golang.org/pkg/go/build/#hdr-Build_Constraints) to only include the desired tracers at build time.

Here's an example of how we could do it: https://github.com/zalando/skipper/compare/2f4b6ae...build-flag-tracers

If we only build skipper e.g. with `lightstep` support it would be a bit
smaller compared to building with all tracers supported.

```
$ GO111MODULE=on go build -tags "opentracing lightstep" -ldflags "-X main.version=v0.10.80-1-gae10a32 -X main.commit=ae10a32" -o bin/skipper ./cmd/skipper/*.go
$ du -hs bin/skipper
20M	bin/skipper
```

##### Pros

* **Statically linked** binaries with opentracing support.
  * No need to build in a custom alpine docker image.
* Size of skipper image with e.g. lightstep support would be smaller.

##### Cons

* Running skipper with a proprietary tracer would require maintaining a
    fork/importing skipper as library.
* Need multiple docker images to support different builds e.g. one without
    opentracing support, and one with `lightstep` and one with all.

### Conclusion

I'm in favor of **Solution #1** because it simplifies the build the most at the
cost of a few extra megabytes and maybe making it harder to use skipper with
proprietary plugins.


szuecs:
Skipper is a first class library, therefore I am against option 1 as far as I understand your description it would be a must to create a fork of the library to enable the library use case.

mikkeloscar:
> Skipper is a first class library, therefore I am against option 1 as far as I understand your description it would be a must to create a fork of the library to enable the library use case.

@szuecs no, you would still be able to use skipper as a library, you would just implicitly pull in the extra opentracing libraries even if you don't use them at run time, but the functionality wouldn't change.

What I meant would change is the situation where you want to use a custom opentracing library not included in skipper. In this case you would have to either fork skipper and change this: https://github.com/zalando/skipper/pull/779/files#diff-f88a13755a4b2ad7cfcd12f7defdd3a2R74 or use skipper as a library and manually configure the tracer.

Right now you can build `myproprietary_opentracing_plugin.so` and put it in the plugin folder and run the upstream skipper binary. This is what you wouldn't be able to do anymore.

I hope this makes sense.

aryszka:
from the two options, I prefer 1, as well, because overall simpler. My concerns are not so much about the size of binary, but rather the number of dependencies. But let's err for the simpler build.

In ideal case, we could include all the tracer implementations only in the executable binary, while leaving them out from the library representation and letting the user choose which tracer to build with skipper, but unfortunately this is only possible with a breaking change in the skipper.Options type [OpenTracing []string](https://github.com/zalando/skipper/blob/master/skipper.go#L380), and the breaking change would be to use a concrete, configured tracer instance that is set up in cmd/skipper/main.go instead of the strings. We can deprecate the []string style option and provide an alternative way of setup, but it is also fine to do it in a separate PR.

aryszka:
:+1:

szuecs:
:+1:

mikkeloscar:
> In ideal case, we could include all the tracer implementations only in the executable binary, while leaving them out from the library representation and letting the user choose which tracer to build with skipper, but unfortunately this is only possible with a breaking change in the skipper.Options type OpenTracing []string, and the breaking change would be to use a concrete, configured tracer instance that is set up in cmd/skipper/main.go instead of the strings. We can deprecate the []string style option and provide an alternative way of setup, but it is also fine to do it in a separate PR.

I also thought about this, and I think it makes sense to do in a separate PR.

szuecs:
:+1:

aryszka:
:+1:

PR #782, opened 2018-09-10
avoid re-calculating the -1 of the reverse loop

ddufourboivin-zalando:
This `[count-1-i]` triggered my OCD. Let's calculate the `-1` part just once, shall we?


szuecs:
:+1:

PR #790, opened 2018-09-14
Added CDP build step for publishing docs.

arjunrn:
Added a mkdocs theme. Also CDP build step to publish docs.


arjunrn:
The docs cannot be published from CDP because credentials for github are not present. Closing this PR.

Issue #802, opened 2018-09-20
Skipper not logging 5xxs


mikkeloscar:
I will follow up with more information later. Putting this as a note so I don't forget!

I believe we might have introduced a breaking change in https://github.com/zalando/skipper/pull/752 because we no longer see 5xx error logs in our skipper-ingress logs like we used to.

This shows the log volume for a specific hostname:

![image](https://user-images.githubusercontent.com/128566/45813851-081ce480-bcd4-11e8-9f46-eead3d2cc463.png)

and this shows the metrics for the same hostname at the same time:

![image](https://user-images.githubusercontent.com/128566/45813870-17039700-bcd4-11e8-8ce4-45f596dd685b.png)

What we see is that log volume goes down when the 5xx metrics go up and we can't find 5xx in the skipper logs for that time frame.

szuecs:
@vecmezoni we have to revert your change. #803
I think only the happy path works with the current filter idea. I am not sure if we can easily fix it, so I created #803 to revert the whole filter log system.

vecmezoni:
If you switch off access log globally you won’t see errors in logs. Why filter should have different behavior?

szuecs:
@vecmezoni we don't switched off the logging and if we would how can you explain the less logs during the error period?

vecmezoni:
I haven't got the problem in the issue from the first read.
I've got that "accessLogDisabled() won't produce log messages for errors". 
Actually, it was "because of bug in 'accessLog' filter we've lost error messages". 
Sorry for the confusion.

szuecs:
@vecmezoni can you look into this issue?
I expect that if we have a GW timeout (for example backend `nc -l 8080`) or something like that will trigger this bug.

PR #803, opened 2018-09-23
Revert/logging filter

szuecs:
The logging filter only supports the happy path. Logging 500s will not happen anymore.
Therefore we need to revert this change. This is a breaking change as before.

See #802 for details


szuecs:
:+1:

szuecs:
Closed in favor of https://github.com/zalando/skipper/pull/806

PR #806, opened 2018-09-25
Fix logging regression after #752

### LINKED ISSUE #802: Skipper not logging 5xxs

vecmezoni:
Fixes #802 

During logging refactoring error condition was missed
```
	if err != nil {
		p.errorResponse(ctx, err)
		return
	}
```

### DIFF
--- a/logging/loggingwriter.go
+++ b/logging/loggingwriter.go
@@ -10,6 +10,7 @@ import (
 type LoggingWriter struct {
 	writer http.ResponseWriter
 	bytes  int64
+	code   int
 }
 
 func NewLoggingWriter(writer http.ResponseWriter) *LoggingWriter {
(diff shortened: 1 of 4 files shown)


szuecs:
:+1:

szuecs:
Thanks for your investigation @vecmezoni !

aryszka:
:+1:

PR #812, opened 2018-10-03
fix dropRequestHeaders in kubernetes should be applied before user filters

### LINKED ISSUE #808: tee() filter between two kubernetes clusters with skipper load balancing between the endpoints
in load balanced routes, the X-Load-Balancer-Member is dropped at the end of the route. This causes a problem in case of the tee() filter, when the tee requests go to another similar Kubernetes cluster.

szuecs:
fix #808 and add dropRequestHeaders filter before the filters from the annotations applied by users

Signed-off-by: Sandor Szuecs <sandor.szuecs@zalando.de>


aryszka:
:+1:

mikkeloscar:
:+1:

PR #823, opened 2018-10-24
Activate compatibility defaults on Prometheus metrics

jchanam:
I've found that the Prometheus metrics does not have the default compatibility options activated by default.

If you see the comment for `DisableCompatibilityDefaults`, you can see the following:
https://github.com/zalando/skipper/blob/7d1e7ba519fc513bae8a7ecb8a5faec398c24ef1/metrics/metrics.go#L141-L146

So, with this PR, the compatibility defaults are also applied on Prometheus metrics.


aryszka:
:+1:

szuecs:
:+1:

szuecs:
👍

aryszka:
:+1:

Issue #767, opened 2018-09-04
document auth filters


szuecs:
The auth filters need a bit more documentation to make them understandable.
We have to point out that you need some tokeninfo implementation and tokenintrospection implemenetation and create pictures to explain how the setup looks like.

Issue #701, opened 2018-07-06
Ingress redirectTo doesn't keep the url suffix


fpozzobon:
Hi,

when adding the annotation : redirectTo(301, "https://myUrl")
for the host : redirected.url.do

I was expecting that https://redirected.url.do/aSufix?prop=value
would redirect to https://myUrl/**aSufix?prop=value**
but instead redirect to https://myUrl/**?prop=value**

Is there a way to keep the full url sufix when redirecting ?

Thanks by advance,

Fabien

fpozzobon:
Got the answer from @szuecs : 

modPath("^/", "/")
      -> redirectTo(301, "https://myUrl")

Note : in my case I wanted a redirect like : 
 - redirectTo(301, "https://myUrl/foo")
so it should be like 
 - modPath("^/", "/foo")
      -> redirectTo(301, "https://myUrl")

addityasingh:
@fpozzobon Seems you found the answer to your issue. @szuecs Is this issue still valid for documentation purpose?

szuecs:
@addityasingh https://opensource.zalando.com/skipper/filters/#redirectto does not show it so I would say, yes

fpozzobon:
Hi @addityasingh, I kept the issue open as a reminder to update the documentation.
But from my point of view the issue is solved ;)

addityasingh:
@szuecs I will try to add somethings soon so that we can close it

Issue #511, opened 2017-12-15
document <shunt>


szuecs:
Document what is <shunt> and when to use it

PR #832, opened 2018-11-02
Doc/tutorial 

szuecs:
topics:

- basics
- ratelimit
- authnz
- development
- common use cases (redirect only)

closes: #767 #701 #511


docs/index.md
```diff
@@ -1,13 +1,19 @@
 # Introduction
 
-This is the documentation page of [Skipper](https://github.com/zalando/skipper). Skipper is an HTTP router and reverse proxy for service composition. It's designed to handle >100k HTTP route definitions with detailed lookup conditions, and flexible augmentation of the request flow with filters. It can be used out of the box or extended with custom lookup, filter logic and configuration sources.
+This is the documentation page of
+[Skipper](https://github.com/zalando/skipper). Skipper is an HTTP
+router and reverse proxy for service composition. It's designed to
+handle large amounts of HTTP route definitions (>800000 routes) with
```
aryszka: what do you think about adding here something like "dynamically configured routes", or something to emphasized that there's no need to restart on config?

docs/tutorials/auth.md
```diff
@@ -0,0 +1,141 @@
...
+
+Install htpasswd command line tool, we assume Debian based
+system. Please refer the documentation of your Operating System or
+package management vendor how to install `htpasswd`:
+
+```
+% apt-get install apache2-utils
+```
+
+Create a htpasswd file `foo.passwd` and a use `catpain` with password `apassword`:
```
aryszka: "...and a use `catpain` with password..." => "...and use `captain` with password..."

docs/tutorials/auth.md
```diff
@@ -0,0 +1,141 @@
...
+< Server: Skipper
+< Www-Authenticate: Basic realm="Basic Realm"
+< Date: Thu, 01 Nov 2018 21:27:18 GMT
+< Content-Length: 0
+<
+* Connection #0 to host localhost left intact
+
+```
+
+A client request with correct specifying credentials:
```
aryszka: A client request with the correct credentials:|

docs/tutorials/auth.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Tokeninfo
+
+Tokeninfo is a common, but not specified protocol, only supporting
+Bearer tokens in the Authorization header.
+
+In most cases you would have to have your own OAuth2 token
+infrastructure, that can return JWT or OAuth2 access tokens to authenticated parties
+and validate tokens with their custom tokeninfo endpoint. In case of
+JWT the access token is signed and can be validated without central
```
aryszka: "...without central..." => "...without a central..."

docs/tutorials/auth.md
```diff
@@ -0,0 +1,141 @@
...
+```
+all: Path("/")
+     -> oauthTokeninfoAnyScope("read-X", "readwrite-X")
+     -> "http://localhost:9090/"
+```
+
+The access token should be passed from the client as Bearer token in
+the Authorization header. Skipper will send this token unchanged as
+Bearer token in the Authorization header to the Tokeninfo endpoint.
+Skipper with Tokeninfo setup, request flow is shown in the following
```
aryszka: ... The request flow with a Tokeninfo setup is shown on the following picture:

docs/tutorials/auth.md
```diff
@@ -0,0 +1,141 @@
...
+        -> oauthTokenintrospectionAnyKV("https://identity.example.com/managed-id", "jdoe")
+        -> "http://localhost:9090/";
+```
+
+The access token should be passed from the client as Bearer token in
+the Authorization header. Skipper will send this token as
+defined in [RFC7662](https://tools.ietf.org/html/rfc7662#section-2.1)
+in a POST request "application/x-www-form-urlencoded" as value for key
+`token` to the Tokenintrospection endpoint.
+Skipper with Tokenintrospection setup, request flow is shown in the
```
aryszka: The request flow with Tokenintrospection setup is shown on the following picture:

docs/tutorials/basics.md
```diff
@@ -0,0 +1,346 @@
...
+```
+
+### Predicate
+
+A Predicate adds a matching rule to a route.
+For example the Cookie predicate, `Cookie("yandex", "true")`, matched
+if there is a cookie in the request with name "yandex" and the value
+is "true", else the route processing will go one and try to find
+another matching route for the given request. Multiple predicates can
+be concatenated by `&&` which means a logical **AND**. If you need a
```
aryszka: i would use "combined" instead of "concatenated"

docs/tutorials/basics.md
```diff
@@ -0,0 +1,346 @@
...
+HTTP/1.1 200 OK
+Content-Type: text/plain
+X-Count: 1
+X-Timestamp: 1541086036
+Date: Fri, 02 Nov 2018 00:30:43 GMT
+```
+
+For skipper operators the number of routes can be interesting for
+statistics and the timestamp to detect skipper instances that have not
+updated its routing table.
```
aryszka: I would add something like:

If there is more than 1024 routes used, then the paging the results is possible with the offset and limit query parameters:

```
curl locahost:9911/routes?offset=2048&limit=512
```

docs/tutorials/development.md
```diff
@@ -0,0 +1,151 @@
...
+Dataclients are the way how to integrate new route
+sources. Dataclients pull information from a source and create routes
+for skipper's routing table.
+
+You have to implement `routing.DataClient`, which is an interface that defines
+function signatures `LoadAll() ([]*eskip.Route, error)` and
+`LoadUpdate() ([]*eskip.Route, []string, error)`.
+
+A complete example is the [routestring implementation](https://github.com/zalando/skipper/blob/master/dataclients/routestring/string.go), which fits in
+less than 50 lines of code.
```
aryszka: Do you it can be relevant how both polling and blocking are supported? Sg. like: 

The LoadUpdate() method can be implemented either in a way that returns immediately, or blocks until there is a change. The routing package will regularly call the LoadUpdate() method with a small delay between the calls.

docs/tutorials/development.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Opentracing
+
+Your custom Opentracing implementations needs to satisfy the `opentracing.Tracer` interface from
```
aryszka: "implementations needs to satisfy " => "implementations need to satisfy "

docs/tutorials/ratelimit.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Instance local Ratelimit
+
+Filters `ratelimit()` and `clientRatelimit()` calculate the ratelimit
+in a local view having no information about other skipper instances.
+
+### Backend Ratelimit
+
+The backend ratelimit filter is `ratelimit()` and it is the most
+simplest one. You can define how many requests a route allows for a
```
aryszka: "it is the most simplest one": drop "most"

docs/tutorials/auth.md
```diff
@@ -0,0 +1,141 @@
...
+
+Install htpasswd command line tool, we assume Debian based
+system. Please refer the documentation of your Operating System or
+package management vendor how to install `htpasswd`:
+
+```
+% apt-get install apache2-utils
+```
+
+Create a htpasswd file `foo.passwd` and use `catpain` with password `apassword`:
```
aryszka: it is still catpain, if it is for a joke, i buy it ;)

aryszka:
:+1:

PR #833, opened 2018-11-02
Fix/disable audit log in experimental upgrade

aryszka:
the experimental web socket upgrade feature currently includes an audit logging functionality logs the request line and the messages without opt out. This PR makes this behavior configurable. In addition to that, it sets not logging the messages as the default behavior because, I believe, that is more expected be the operators.


aryszka:
:+1:

szuecs:
👍

Issue #835, opened 2018-11-05
Skipper starts serving before the routes are initialized


aermakov-zalando:
When Skipper is running in Kubernetes, if it restarts for any reason (e.g. OOM kill) before the ELB health check can notice that a node is down, it'll start responding 404s to all client requests. This is dangerous because applications can and will react to 404s, e.g. by expiring tokens or doing other irreversible changes.
There is some code that's supposed to check if the data client was initialised or not, but this only affects the health check status. Routing doesn't check it.

PR #837, opened 2018-11-06
wait for first routes loaded before starting the listener

aryszka:
Closes: https://github.com/zalando/skipper/issues/835

If the -wait-first-route-load flag is set, then the listener will be started only after the first batch of routes are received.

- [x] implement signalling the first route load
- [x] test the routing package that the signal channel only blocks when the option is enabled


szuecs:
:+1:

aryszka:
:+1:

Issue #420, opened 2017-08-14
service mesh


szuecs:
Placeholder to reference, work in progress

aryszka:
how about using the standard http proxy setting configured in the environment? Then the clients just need to use the proxy.

szuecs:
I changed the description to reflect your point. Separated cluster internal and external communication and 2 options for external communication.

PR #841, opened 2018-11-06
east-west svc-to-svc communication in kubernetes

### LINKED ISSUE #791: Create routes for ingress east/west communication (cluster internal svc-to-svc)
I just had a discussion with my colleagues how to enable cluster internal "east/west" communication to be proxied and loadbalanced through skipper, too. This is an enhancement for cluster internal communication through Kubernetes service, because of more functionality for east/west calls (used as internal API GW) and reliability, for example transparent connection retry.

External-DNS or [CoreDNS](https://github.com/coredns/coredns) (or maybe /etc/hosts, but wildcards seem not to work) can create records for skipper into coredns derived from ingress name and namespace: `<name>.<namespace>.skipper.cluster.local`
A flag in skipper would enable `-enable-kubernetes-east-west` (better name?) duplication of additional routes for all ingress with `Host()` predicates and change the `Host()` predicate for the cluster internal route.

## Example

```
kind: Ingress
metadata:
  name: demo
  namespace: default
spec:
  rules:
  - host: demo.example.org
    http:
      paths:
      - backend:
          serviceName: example
          servicePort: 80
```

The shown ingress normally creates a frontend route and for all endpoints corresponding to the service backend routes. All of these have to be duplicated if these contain the `Host()` predicate, which is automatically set for  `demo.example.org`.
If enabled, skipper would additionally to these duplicate the routes and change the `Host()` predicate to `demo.default.skipper.cluster.local`.  (Thanks @aermakov-zalando for the better hostname hint!)

## additional things to solve to enable this

From skipper and HTTP Host header point of view this would already work. We only need to have a documented idea how to create the DNS entries pointing to the skippers.

### Kubernetes service to access skipper

Skipper should be deployed additionally with kubernetes service type ClusterIP to have a fixed, loadbalanced IP for all the skippers.

### populate DNS

Creating the DNS records is simple with the `template` plugin from CoreDNS.

Corefile example:
```
.:1053 {
    errors
    health
    kubernetes cluster.local in-addr.arpa ip6.arpa {
        pods insecure
        upstream
        fallthrough in-addr.arpa ip6.arpa
    }
    template IN A skipper.cluster.local  {
      match "^.*[.]skipper[.]cluster[.]local"
      answer "{{ .Name }} 60 IN A 10.3.11.28"
      fallthrough
    }
    prometheus :9153
    proxy . /etc/resolv.conf
    cache 30
    reload
}
```

### resolve DNS

DNS resolution should just work as the normal flow, no changes needed.

#### optional add custom search in resolv.conf

https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-config shows that you can pass a custom `dnsConfig` with `searches` to add also `skipper.cluster.local` domain to the resolv.conf and add one additional `ndots` to resolve it. 

## Old idea to create routes for service type ClusterIP

If you create a service type ClusterIP `foo` in namespace `bar` you get service name in DNS: `foo.bar.svc.cluster.local`
If skipper would create automatically for all of these a service type clusterIP with prefix "skipper-"  and points to skipper-ingress (by label selector), we would get `skipper-foo.bar.svc.cluster.local` pointing to the skipper ingress controller.

Skipper would then need to create for all these "skipper-" prefixed services the routes with endpoints from the endpoints without the "skipper-" prefix.

Like this we could support cluster internal traffic to pass through skipper and users only need to send traffic to the "skipper-" prefixed DNS to call their destination service.

How do you like the idea or do you have other similar ideas?

Cc @mikkeloscar @aermakov-zalando @aryszka @arjunrn  @ese @jchanam

szuecs:
This feature enables east-west svc-to-svc communication in kubernetes

Closes #791 #420

TODO:

- [x] catchall routes that have a host header needs to be added
- [x] test in dev cluster if this works correctly
- [x] add docs
  - [x] deployment docs
  - [x] user docs


dataclients/kubernetes/kube_test.go
```diff
@@ -2931,7 +3341,258 @@ func TestSkipperCustomRoutes(t *testing.T) {
...
+			"", backendPort{}, 1.0,
+			testRule("www1.example.org", testPathRule("/a/path", "bar", backendPort{"baz"})),
+		)},
+		expectedRoutes: map[string]string{
+			"kube_foo__qux__www1_example_org___a_path__bar":   "Host(/^www1[.]example[.]org$/) && PathRegexp(/^\\/a\\/path/) -> \"http://1.2.3.4:8181\"",
+			"kube_foo__qux__0__www1_example_org_a_path____":   "Host(/^www1[.]example[.]org$/) && PathRegexp(/^\\/a\\/path/) && Method(\"OPTIONS\") -> <shunt>",
+			"kube___catchall__www1_example_org____":           "Host(/^www1[.]example[.]org$/) -> <shunt>",
+			"kubeew_foo__qux__www1_example_org___a_path__bar": "Host(/^qux[.]foo[.]skipper[.]cluster[.]local$/) && PathRegexp(/^\\/a\\/path/) -> \"http://1.2.3.4:8181\"",
+			"kubeew_foo__qux__0__www1_example_org_a_path____": "Host(/^qux[.]foo[.]skipper[.]cluster[.]local$/) && PathRegexp(/^\\/a\\/path/) && Method(\"OPTIONS\") -> <shunt>",
+			//"kubeew___catchall__www1_example_org____":         "Host(/^qux[.]foo[.]skipper[.]cluster[.]local$/) -> <shunt>",
```
aryszka: i think we need the catchall for the east-west routes, as well. Is it WIP?

szuecs: yes

szuecs: I added a todo list

szuecs:
:+1:

szuecs:
@mikkeloscar @aryszka this can be merged in my opinion, thoughts?

aryszka:
yes, agree

aryszka:
:+1:

dataclients/kubernetes/kube.go
```diff
@@ -37,6 +37,8 @@ const (
 	ingressesNamespaceFmt         = "/apis/extensions/v1beta1/namespaces/%s/ingresses"
 	ingressClassKey               = "kubernetes.io/ingress.class"
 	defaultIngressClass           = "skipper"
+	ingressRouteIDPrefix          = "kube"
+	defaultEastWestDomainFmt      = "%s.%s.skipper.cluster.local"
```
mikkeloscar: would it make sense to use go templating and require the string to be something like: `{{ .Name }}.{{ .Namespace }}.skipper.cluster.local`. This would allow having namespace before service name if desired and be easier to understand from a user point of view I think.

This idea is inspired by how we did it in external-dns: https://github.com/kubernetes-incubator/external-dns/blob/c0968ab857201b61a300aeb1a0b1b41a6f0daad3/pkg/apis/externaldns/types.go#L226

szuecs: discussed offline

szuecs:
:+1:

ese:
Sorry I'm a little bit late here but i have some concerns.

It's a little bit weird take a totally different way to build host filter that could lead to multiple corner cases.

For example:
 * A ingress with two different domains in their set of rules
´´´yaml
kind: Ingress
metadata:
  name: demo
  namespace: default
spec:
  rules:
  - host: demo.example.org
    http:
      paths:
      - backend:
          serviceName: example
          servicePort: 80
  - host: demo2.example.org
    http:
      paths:
      - backend:
          serviceName: example2
          servicePort: 80

´´´
I think this would work for external traffic through `demo.example.org` and `demo2.example.org` but would be a conflict for internal `name.ns...`

 * A domain with rules in multiple ingresses
```
kind: Ingress
metadata:
  name: demo
  namespace: api
spec:
  rules:
  - host: demo.example.org
    http:
      paths:
      - path: /api
        backend:
          serviceName: api
          servicePort: 80
---
kind: Ingress
metadata:
  name: demo
  namespace: app
spec:
  rules:
  - host: demo.example.org
    http:
      paths:
      - path: /app
        backend:
          serviceName: app
          servicePort: 80
```

In this case what it works with one endpoint `demo.example.org` suddenly is splitted to two `demo.app...` and `demo.api` which can break the client implementation exposing infrastructure details that you had behind the ingress abstraction.

szuecs:
True the first case would be ambiguous.
The second would work as documented. I know it might be not obvious, but how do you think it should work in case 2 for example?

Issue #853, opened 2018-11-17
External DNS Collision causes 404


jruhland:
Deployed via kubernetes with ingress controller, we use a ingress that maps two very similar DNS entries, which causes intermittent "Not Found" errors, removing either of the entries will result in fixing, or adding them as separate ingresses, but together, ~20% error rate with `Not Found`

Produced on both 0.10.0 and 0.10.121

Ingress Definition Used to Reproduce:
```
{
  "kind": "Ingress",
  "apiVersion": "extensions/v1beta1",
  "metadata": {
    "name": "svc",
    "namespace": "default",
    "labels": {
      "app": "svc",
      "chart": "svc-1.0.0",
      "heritage": "Tiller",
      "release": "prd-svc"
    },
    "annotations": {
      "kubernetes.io/ingress.class": "skipper-internet-facing"
    }
  },
  "spec": {
    "rules": [
      {
        "host": "svc.prd.myapp.com",
        "http": {
          "paths": [
            {
              "path": "/",
              "backend": {
                "serviceName": "svc",
                "servicePort": "http"
              }
            }
          ]
        }
      }
      {
        "host": "svc-prd.myapp.com",
        "http": {
          "paths": [
            {
              "path": "/",
              "backend": {
                "serviceName": "svc",
                "servicePort": "http"
              }
            }
          ]
        }
      }
    ]
  }
}
```

It appears to be related to `dataclients/kubernetes/kube.go:470` on master and the fact that all non-words are replaced with an underscore

szuecs:
@jruhland would you mind testing to change the `host: value` to `host: ^value` in both of your cases?
I think it’s because we use regexp matching to host headers as the ingress spec tells and there are multiple use cases that compete.
I will check on Monday the code line you mentioned.

greenboxal:
Pushed a fix here: https://github.com/zalando/skipper/pull/854

szuecs:
I was wrong for the kubernetes dataclient we create Host header matches like this. that match the dot correctly:

    Host(/a[.]b[.]localhost/)

PR #854, opened 2018-11-18
Fix kubernetes rule name hashing

greenboxal:
This fixes a ID generation issue when a single ingress contains two hostnames that would collide given the previous `routeID` function.

The problem was caused because of the sanitization rule, that replaced any non-word with an underscore. Given `my-api.prd.company.com` and `my-api-prd.company.com`, they would be sanitized to the same name, `my_api_prd_company_com`.

Somewhere down the code, one would effectively replace the other rule, allowing only one of the domains to be actually registered. (probably by putting the routes on a map using `eskip.Route.Id` as key)

My solution was to include the index of the ingress `host` in the name. Anything else that could make the ID unique would work as well (hashing, etc)

### DIFF
--- a/dataclients/kubernetes/httpsredirect_test.go
+++ b/dataclients/kubernetes/httpsredirect_test.go
@@ -311,13 +311,13 @@ func TestEnableHTTPSRedirectFromIngress(t *testing.T) {
 	}
 
 	const expectEskip = `
-		kube_namespace1__ingress1______: * -> "http://1.2.3.4:8080";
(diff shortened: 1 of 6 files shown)


szuecs:
I think this will fix the routeID and in the kubernetes dataclient we also mask `.` with `[.]` so it should be fine in general. So my mentioned comment here https://github.com/zalando/skipper/issues/853#issuecomment-439677447 does not apply.

szuecs:
:+1:

szuecs:
@aryszka please review, thanks

szuecs:
:+1:

PR #855, opened 2018-11-19
Clearer JSON Serialization

ddufourboivin-zalando:
Interested in knowing here what you prefer in terms of customised JSON serialization.

Before: implementing `json.Marshaller` on the complete `pathInfo` struct, using a temporary anonymous struct to bypass only the `Matcher`, serialising the regular expression's string instead of the `regexp.Regexp` struct itself.

After: using a type definition `Regex` based on `regexp.Regexp` and implementing the `json.Marshaller` interface just for itself, marshalling its string. Using that type definition inside of `pathInfo`.

### DIFF
--- a/filters/apiusagemonitoring/pathinfo.go
+++ b/filters/apiusagemonitoring/pathinfo.go
@@ -6,13 +6,22 @@ import (
 	"regexp"
 )
 
+type Regex struct{ *regexp.Regexp }
+
+func (r *Regex) MarshalJSON() ([]byte, error) {
+	if r == nil {
+		return json.Marshal(nil)
+	}
+	return json.Marshal(r.String())
+}
+
 // pathInfo contains the tracking information for a specific path.
 // The exported members are marshaled as JSON.
 type pathInfo struct {
 	ApplicationId  string
 	ApiId          string
 	PathTemplate   string
-	Matcher        *regexp.Regexp
+	Matcher        *Regex
 	ClientTracking *clientTrackingInfo
 	CommonPrefix   string
 	ClientPrefix   string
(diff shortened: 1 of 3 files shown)


filters/apiusagemonitoring/pathinfo.go
```diff
@@ -6,13 +6,22 @@ import (
...
+
 // pathInfo contains the tracking information for a specific path.
 // The exported members are marshaled as JSON.
 type pathInfo struct {
 	ApplicationId  string
 	ApiId          string
 	PathTemplate   string
-	Matcher        *regexp.Regexp
+	Matcher        *Regex
 	ClientTracking *clientTrackingInfo
```
aryszka: wouldn't we need to use the same approach on the clientTrackingInfo type, which also has a Regexp field?

filters/apiusagemonitoring/pathinfo.go
```diff
@@ -6,13 +6,22 @@ import (
 	"regexp"
 )
 
+type Regex struct{ *regexp.Regexp }
```
aryszka: small consistency nitpick: it's called regexp everywhere in the stdlib and the app, let's keep that 'p'

aryszka:
"in terms of customised JSON serialization": I find this approach fine :smile:

filters/apiusagemonitoring/pathinfo.go
ddufourboivin-zalando: Was also wondering if I should call it something like `JsonRegexp` or `JSRegexp` or something more specifically linked to why it is being extended.

aryszka: following the rule of 'less is more', better keep it just regexp. Oh! Is this type referenced anywhere outside of the filter? Because if not, then we don't need to export it (meaning no need for the capital 'R')

ddufourboivin-zalando:
Another alternative I read is to bypass ONLY the fields to be custom-marshalled in an anonymous struct.
The rest of the marshalling will be the default for all other fields.

```go
func (p *pathInfo) MarshalJSON() ([]byte, error) {
	type Alias pathInfo
	return json.Marshal(&struct{
		*Alias
		Matcher string
	}{
		Alias: (*Alias)(p),
		Matcher: p.Matcher.String(),
	})
}
```

@aryszka @szuecs

filters/apiusagemonitoring/pathinfo.go
ddufourboivin-zalando: image

Mmmm ... well then that makes a name clash with the import of the `regexp` package...

aryszka: "less is more", it could be then:

```
type rexp struct { *regexp.Regexp }
```

PR #857, opened 2018-11-21
preserve only the span not the entire context of the incoming request

aryszka:
instead of preserving the context.Context of the incoming request to the outgoing request, preserve only the opentracing span. `skptesting/benchmark-proxy.sh` showed that this eliminates the 'context cancelled' errors.

- [x] preserve only the span between the incoming and outgoing request instead of the entire context.Context object
- [ ] prove with a unit test that this is the cause of the the 'context cancelled' errors


vetinari:
:+1:

szuecs:
:+1:

Issue #797, opened 2018-09-19
filters that call external endpoints should add opentracing spans


szuecs:
Yesterady, I saw a trace that took 60ms in skipper and I bet it was the `compress()` filter.
I think it makes sense to show the filter runtime individually, which is also good for webhook, oauth* filters, calling external services.

vetinari:
depends on #756

vetinari:
```
diff --git a/proxy/proxy.go b/proxy/proxy.go
index e05844f..a90dd67 100644
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -543,6 +543,10 @@ func (p *Proxy) applyFiltersToRequest(f []*routing.RouteFilter, ctx *context) []
        for _, fi := range f {
                start := time.Now()
                tryCatch(func() {
+                       parentSpan := opentracing.SpanFromContext(ctx.Request().Context())
+                       span := p.Tracer.StartSpan("filter_"+fi.Name, opentracing.ChildOf(parentSpan))
+                       defer span.Finish()
+                       ctx.SetRequestContext(opentracing.ContextWithSpan(ctx.Request().Context(), span))
                        ctx.setMetricsPrefix(fi.Name)
                        fi.Request(ctx)
                        p.metrics.MeasureFilterRequest(fi.Name, start)
```
should do, without #756 it's hard to test :)

szuecs:
IMHO we should only instrument with more span for all external calls we do from filters, unlike it was done in #860.

PR #860, opened 2018-11-22
Add OpenTracing spans to filter req/resp

mikkeloscar:
This adds open tracing spans to each filter on both the request part and response part. I have named the spans `filter_request_<filter-name>` and `filter_response_<filter-name>`.

The point of this is to add visibility especially to filters taking a long time which is currently "hidden" in the overarching span for the whole request trough skipper.

I have tested it by running skipper like this:

```bash
./bin/skipper "-opentracing=lightstep component-name=skipper-ingress-mlarsen token=<token> collector=tracing.zmon.zalan.do:8444 cmd-line=skipper-ingress max-buffered-spans=4096" -inline-routes 'p0: * -> compress("text/html") -> status(200) -> <shunt>'
```
This includes two filters `compress` and `status` and looks like this in a trace:

image

And each filter span has the following tags

image

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -558,7 +558,20 @@ func (p *Proxy) applyFiltersToRequest(f []*routing.RouteFilter, ctx *context) []
 		start := time.Now()
 		tryCatch(func() {
 			ctx.setMetricsPrefix(fi.Name)
+
+			var filterSpan ot.Span
+			spanName := "filter_request_" + fi.Name
+			ingressSpan := ot.SpanFromContext(ctx.request.Context())
+			if ingressSpan == nil {
+				filterSpan = p.openTracer.StartSpan(spanName)
+			} else {
+				filterSpan = p.openTracer.StartSpan(spanName, ot.ChildOf(ingressSpan.Context()))
+			}
+			filterSpan.SetTag("skipper.filter", fi.Name)
+			filterSpan.SetTag("skipper.filter.index", fi.Index)
 			fi.Request(ctx)
+			filterSpan.Finish()
+
 			p.metrics.MeasureFilterRequest(fi.Name, start)
 		}, func(err interface{}) {
 			if p.flags.Debug() {
@@ -591,7 +604,20 @@ func (p *Proxy) applyFiltersToResponse(filters []*routing.RouteFilter, ctx *cont
 		start := time.Now()
 		tryCatch(func() {
 			ctx.setMetricsPrefix(fi.Name)
+
+			var filterSpan ot.Span
+			spanName := "filter_response_" + fi.Name
+			ingressSpan := ot.SpanFromContext(ctx.request.Context())
+			if ingressSpan == nil {
+				filterSpan = p.openTracer.StartSpan(spanName)
+			} else {
+				filterSpan = p.openTracer.StartSpan(spanName, ot.ChildOf(ingressSpan.Context()))
+			}
+			filterSpan.SetTag("skipper.filter", fi.Name)
+			filterSpan.SetTag("skipper.filter.index", fi.Index)
 			fi.Response(ctx)
+			filterSpan.Finish()
+
 			p.metrics.MeasureFilterResponse(fi.Name, start)
 		}, func(err interface{}) {
 			if p.flags.Debug() {


szuecs:
:+1:
@aryszka

mikkeloscar:
Maybe also @lmineiro or @vetinari  has some thoughts on the span names and tags added?

proxy/proxy.go
```diff
@@ -591,7 +604,20 @@ func (p *Proxy) applyFiltersToResponse(filters []*routing.RouteFilter, ctx *cont
 		start := time.Now()
 		tryCatch(func() {
 			ctx.setMetricsPrefix(fi.Name)
+
```
aryszka: the duplicate code can be avoided by using a function: 

```
func (p *proxy) traceFilter(ctx *context, f *routing.RouteFilter, filterMethod func(eskip.FilterContext)) {
    // ...
}
```

aryszka:
from the above trace screenshot, the status looks quite interesting: the request of the status filter is a noop, how can it take 10 microseconds?

aryszka:
I am not completely sure of the all the consequences of this change, but I believe we should not do this.

We have metrics for individual filters. When they are enabled, then we have detailed timers for each filter.  Since the majority of the filters doesn't make any network calls, these timers can give a good enough insight about their performance contribution. Even for filters that are more expensive like the compress() filter. There are Skipper deployments with a quite high number of filters and routes and traffic, and multiplying the collected spans by a factor of 10 to 20 can be very harmful for those deployments.

Distributed tracing brings tons of value when we want to observe how multiple connected services operate together. But most of Skipper's filters have nothing to do with that, only some exceptions, like the auth filters. There is some pragmatism applied when having these network-dependent filters, primarily because their great utility.

So, I think it is a more cautious approach to first add opentracing to specific filters only: those that depend on network services, like auth and tee. That way we can make a decision later easier to generalize the solution or not, without breaking someone's use case who started depending on a generic opentracing instrumentation around all filters.

However, if this helps us to learn more about our system, then let's do it, but for that case we need to put it behind a feature toggle, and clearly document its experimental nature.

What do you think? Am I being too cautious? I also really wonder about @lmineiro 's opinion on this topic.

szuecs:
@aryszka, I agree that we should have a separate configuration to enable it. Then we can learn from that in our smaller deployments, before we enable it as default.

aryszka:
sounds good! @mikkeloscar can you add an opt-in style config for this?

lmineiro:
This looks pretty good.

I'd name the operations `filter_request` and `filter_response` and rely on the tags to filter for specific filters. It's no longer redundant and, besides the per-filter metrics, it will also enable an "overall filtering" overhead.

The filter index may not be needed as the route definition is usually present on the routing span. I couldn't think of any useful aggregation for latencies/errors/throughput of the n-th filter.

I also agree with some configuration that allows opt-in into this. I'd use a whitelist of filters that we'd like to trace. Eventually, it could accept a wildcard for all filters but not really a regex as the matching could easily take more processing than it's worth

mikkeloscar:
Thanks for all the feedback!

@aryszka I agree we should weigh potential performance impact against added visibility and definitely have it behind a flag for opt-in.

Regarding your comment on the time taken for the status filter. seems like this is the overhead of open tracing:

I tried to get the time like this:

```go
startPreSpan := time.Now()
var filterSpan ot.Span
spanName := "filter_request_" + fi.Name
ingressSpan := ot.SpanFromContext(ctx.request.Context())
if ingressSpan == nil {
		filterSpan = p.openTracer.StartSpan(spanName)
} else {
		filterSpan = p.openTracer.StartSpan(spanName, ot.ChildOf(ingressSpan.Context()))
}
filterSpan.SetTag("skipper.filter", fi.Name)
filterSpan.SetTag("skipper.filter.index", fi.Index)
startInnerSpan := time.Now()
fi.Request(ctx)
fmt.Printf("inner span time: %s: %s\n", fi.Name, time.Since(startInnerSpan))
filterSpan.Finish()
fmt.Printf("span time: %s: %s\n", fi.Name, time.Since(startPreSpan))
```

And we clearly see where the microseconds are spent:

```
inner span time: compress: 443ns
span time: compress: 33.195µs
inner span time: status: 785ns
span time: status: 10.166µs
```

With this in mind we could consider only instrumenting filters which result in external calls or use a white list like @lmineiro suggest. I like the white list approach and think we could enable open tracing for white listed filters at skipper startup so we don't have to check it in every request.

This could be done by wrapping the selected filters in a `OpenTracingEnabledFilter` type which implements the Filter interface with added open tracing spans around the inner filter. WDYT?

lmineiro:
> This could be done by wrapping the selected filters in a OpenTracingEnabledFilter type which implements the Filter interface with added open tracing spans around the inner filter. WDYT?

Good idea

aryszka:
i really like both the whitelisting idea and the OpenTracingEnabledFilter type.

lmineiro:
I wouldn't expect `Finish()` to be a "heavy action". I guess it depends what are we comparing it against. The usual implementation adds the Span record to the circular buffer. This is often protected by some concurrency primitives, like Mutexes if the most popular Go Tracers. No sane tracer should do more than that and reporting/shipping spans happens in a separate thread/routine.

Measuring the time `Finish()` takes can, very easily, have the same complexity as `Finish()` itself. Keeping a measurement histogram will have very similar cost (thread-safe set of samples to report the percentile).

szuecs:
See also my comment in https://github.com/zalando/skipper/issues/797#issuecomment-449038228.

szuecs:
@mikkeloscar I think we could close this pr if we would merge #890. Do you agree?

mikkeloscar:
@szuecs I think it could be valuable to have a whitelist which can cover any filter when enabled as discussed in this PR, instead of only implementing the spans per filter.

I'm not going to pick this up right now, since we have #890 but maybe I would like to pick it up in the future unless you don't think it makes sense? I'm fine with closing this for now if you like.

szuecs:
@mikkeloscar I added the wip label

mikkeloscar:
Closing for now as I have not worked on this at all. Can re-open if it becomes relevant.

PR #875, opened 2018-12-08
fix swarm/clusterClient reatlimit integration

szuecs:
fix: clusterClientRatelimit was no initialized correctly
fix: kubernetes-swarm retry connection to kube-apiserver
fix: kubernetes-swarm add connection timeouts to http client
fix: add missing configuration to intialize swarm
fix: kubernetes-swarm tear down handling stops goroutine
fix: clusterClientRatelimit filter parsing
log: reduce log volume caused by swarm

### DIFF
--- a/filters/ratelimit/ratelimit.go
+++ b/filters/ratelimit/ratelimit.go
@@ -281,7 +281,7 @@ func localRatelimitFilter(args []interface{}) (filters.Filter, error) {
 	}, nil
 }
 
-func disableFilter(args []interface{}) (filters.Filter, error) {
+func disableFilter([]interface{}) (filters.Filter, error) {
 	return &filter{
 		settings: ratelimit.Settings{
 			Type: ratelimit.DisableRatelimit,
@@ -298,7 +298,7 @@ func (s *spec) CreateFilter(args []interface{}) (filters.Filter, error) {
 	case ratelimit.ClusterServiceRatelimit:
 		return clusterRatelimitFilter(args)
 	case ratelimit.ClusterClientRatelimit:
-		return clusterRatelimitFilter(args)
+		return clusterClientRatelimitFilter(args)
 	default:
 		return disableFilter(args)
 	}
--- a/filters/ratelimit/ratelimit.go
+++ b/filters/ratelimit/ratelimit.go
@@ -30,7 +30,7 @@ type filter struct {
 // NewLocalRatelimit creates a local measured rate limiting, that is
 // only aware of itself. If you have 5 instances with 20 req/s, then
 // it would allow 100 req/s to the backend from the same user. A third
-// argument can be used to set which part of the request should be
+// argument can be used to set which HTTP header of the request should be
 // used to find the same user. Third argument defaults to
 // XForwardedForLookuper, meaning X-Forwarded-For Header.
 //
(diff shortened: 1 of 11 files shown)


szuecs:
:+1:

swarm/kubernetes.go
```diff
@@ -95,10 +118,12 @@ func NewClientKubernetes(kubernetesInCluster bool, kubernetesURL string) (*Clien
 		httpClient: httpClient,
 		apiURL:     apiURL,
 		token:      token,
+		retry:      backoff.NewConstantBackOff(5 * time.Second),
```
aryszka: i think we should use the backoff.WithMaxRetries because otherwise the backoff.Retry will never return in case of permanent failure

filters/ratelimit/ratelimit.go
```diff
@@ -72,9 +71,10 @@ func NewRatelimit() filters.Spec {
 //    -> clusterRatelimit(200, "1m")
 //    -> "https://foo.backend.net";
 //
-// The above example behaves like the "ratelimit", i.e. per backend. To create a client limit
-// like in "localRatelimit" a third parameter is mandatory. Currently known parameters are
-// "auth" (limit by "Authorization" header) and "xfwd" (client ip from X-Forwarded-For header)
+// The above example behaves like the "ratelimit", i.e. per backend.
+// There is an optional third parameter, which is one of "auth"
+// (limit by "Authorization" header) and "xfwd" (client ip from
```
grassator: Would it be bad if we allow to specify arbitrary header here instead of using limited number of magic strings?

filters/ratelimit/ratelimit.go
```diff
@@ -226,10 +296,10 @@ func (s *spec) CreateFilter(args []interface{}) (filters.Filter, error) {
 	case ratelimit.LocalRatelimit:
 		return localRatelimitFilter(args)
 	case ratelimit.ClusterServiceRatelimit:
-		log.Debugf("SWARM: create cluster ratelimit: %v", args)
+		return clusterRatelimitFilter(args)
```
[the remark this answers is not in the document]
szuecs: no this one is correct the one below is wrong

filters/ratelimit/ratelimit.go
aryszka: Good idea. This should be a separate pull request. Would you like to implement it?

szuecs: I can do it in this PR, but xfwd, will be separate, because it needs a separate path (think about the client ip handling)

szuecs: fixed

filters/ratelimit/ratelimit.go
```diff
@@ -145,25 +161,43 @@ func clusterRatelimitFilter(args []interface{}) (filters.Filter, error) {
...
-			s.Lookuper = ratelimit.NewXForwardedForLookuper()
-			s.CleanInterval = 10 * timeWindow
-		default:
-			return nil, filters.ErrInvalidFilterParameters
-		}
+		s.Lookuper = ratelimit.NewHeaderLookuper(headerName)
+		s.CleanInterval = 10 * timeWindow
 	} else {
-		s.Lookuper = ratelimit.NewSameBucketLookuper()
+		s.Lookuper = ratelimit.NewXForwardedForLookuper()
```
aryszka: does this mean that when we specify the only the time window and the max requests, then we will always use the X-Forwarded-For header for bucketing by default? This is not the same as the previous behavior.

szuecs: yes

aryszka:
:+1:

szuecs:
:+1:

filters/ratelimit/ratelimit.go
grassator: It looks like this is already merged, but as far as I understand in current implementation if have `clusterClientRatelimit(20, "1h", "X-Forwarded-For")` it is different from not providing a header.

Also probably makes sense to either normalize header casing names unless I'm missing this part.

szuecs: Please create an issue for the normalizing and yes the header thing is different from the default.

PR #881, opened 2018-12-13
858 added flowid

emyphan:


lmineiro:
This doesn't follow the established conventions https://github.bus.zalan.do/SRE/opentracing/blob/master/wg-semantic-conventions/semantic-conventions.md

PR #884, opened 2018-12-14
static swarm

szuecs:
fix: data exchange for swarms (not fixed kubernetes issues)
add: static swarm

### DIFF
--- a/swarm/nodeinfo.go
+++ b/swarm/nodeinfo.go
@@ -3,6 +3,8 @@ package swarm
 import (
 	"fmt"
 	"net"
+	"strconv"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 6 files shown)


swarm/nodeinfo.go
```diff
@@ -25,6 +27,26 @@ type NodeInfo struct {
 	Port uint16
 }
 
+func NewStaticNodeInfo(name, addr string) *NodeInfo {
+	a := strings.Split(addr, ":")
```
aryszka: this split won't work with IPv6 addresses. We should use https://godoc.org/net#SplitHostPort

swarm/swarm.go
```diff
@@ -262,8 +277,7 @@ func (s *Swarm) control() {
 		case req := <-s.getOutgoing:
 			s.messages = takeMaxLatest(s.messages, req.overhead, req.limit)
 			if len(s.messages) <= 0 {
-				time.Sleep(10 * time.Millisecond)
-				continue
+				log.Warning("SWARM: getOutgoing with 0 messages, should not happen")
```
aryszka: this warning text may be misleading because the reason for 0 messages can be simply that the underlying messaging cannot send out the next message if it is too large. I think a better warning text would be: "0 outgoing messages with the current overhead and queued message size"

szuecs:
fixed as commented @aryszka

aryszka:
:+1:

PR #886, opened 2018-12-18
fix ratelimit filter header configuration

szuecs:
fix for filter configuration to automatically use canonical headers as Go parses the headers in to the Request in skippers FilterContext


szuecs:
:+1:

PR #890, opened 2018-12-23
Added parent span to context so that it can be used in filters

arjunrn:
Added the filter span to filter context so that it can used in the filters. Also implemented opentracing for webhook authentication so that a span is created and propagated forward to the downstream.

The result trace looks like as show below:

filter-span

### DIFF
--- a/filters/auth/authclient.go
+++ b/filters/auth/authclient.go
@@ -4,12 +4,17 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/opentracing/opentracing-go"
+	"github.com/zalando/skipper/filters"
 	"io"
 	"net"
 	"net/http"
 	"net/url"
 	"sync"
 	"time"
+)
+
+const (
+	webhookSpanName = "webhook"
 )
 
 type authClient struct {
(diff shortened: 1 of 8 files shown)


filters/auth/authclient.go
```diff
@@ -122,12 +125,16 @@ func jsonPost(u *url.URL, auth string, doc *tokenIntrospectionInfo, client *http
 
 // doClonedGet requests url with the same headers and query as the
 // incoming request and returns with http statusCode and error.
-func doClonedGet(u *url.URL, client *http.Client, incoming *http.Request) (int, error) {
+func doClonedGet(u *url.URL, client *http.Client, incoming *http.Request, tracer opentracing.Tracer,
+	parentSpan opentracing.Span, childSpanName string) (int, error) {
+	span := tracer.StartSpan(childSpanName, opentracing.ChildOf(parentSpan.Context()))
+	defer span.Finish()
```
szuecs: I think it would be nicer to pass 'span` and move the 3 lines of code to `getWebhook`.

szuecs:
From code point of view it's nice and the feature is very good to have.
@aryszka we should plan the breaking change, because now we have the feature requesting it. #432 

@aryszka @mo-gr @greenboxal @ese what do you think?

szuecs:
@ddufourboivin-zalando the FilterContext Interface change adds one function to the interface and all implementations would be broken. I am not sure if there’s someone using its own implementation of that.

filters/auth/authclient.go
aryszka: the tracer can be a field of the authClient. The doClonedGet could be a method if the *authClient. Since we alwyas use the webhookSpanName constant, it is not required as a parameter. This way doClonedGet would have only the following parameters: incoming and parentSpan.

aryszka: sry: i see that the tracer cannot be the field of the client without bigger changes, so let's not do that. But the incoming request, the tracer and the parent span name are all coming from the filter context, so why not just passing down the filter context?

filters/auth/webhook.go
```diff
@@ -60,7 +60,9 @@ func copyHeader(to, from http.Header) {
 }
 
 func (f *webhookFilter) Request(ctx filters.FilterContext) {
-	statusCode, err := f.authClient.getWebhook(ctx.Request())
+	tracer := ctx.Tracer()
+	parentSpan := ctx.ParentSpan()
+	statusCode, err := f.authClient.getWebhook(ctx.Request(), tracer, parentSpan)
```
aryszka: here we could also just pass the ctx?

aryszka:
@szuecs i believe it's fine to add the new method to the interface. From experience, there's only testing code that implements its own filtercontext.

aryszka:
:+1:

szuecs:
:+1:

PR #892, opened 2018-12-29
Fix the Host predicate regex example

pierreprinetti:
Apparently, the Javascript regex literal syntax somehow slipped into the
documentation. I have tried and failed to make it work. Removing the
slashes did fix the predicate in my local instance.

### DIFF
--- a/docs/reference/predicates.md
+++ b/docs/reference/predicates.md
@@ -13,7 +13,7 @@ Predicates can be chained using the double ampersand operator `&&`.
 Example route with a Host, Method and Path match predicates and a backend:
 
 ```
-all: Host("^my-host-header\.example\.org$") && Method("GET") && Path("/hello") -> "http://127.0.0.1:1234/";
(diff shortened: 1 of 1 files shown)


pierreprinetti:
Actually, both `Host(/^example\.com$/)` and `Host("^example\.com$")` seem to work, but putting both quotes and slashes may be too much.

pierreprinetti:
OK now I get it. The parser expects slashes, but is liberal about double quotes. Here is the commit description of my latest take.

This PR fixes the example in the `Host` section.

```
The docs state that regular expressions should be surrounded by slashes
(`/`), but just below, the `Host` example is surrounded by double
quotes.

In the `Host` paragraph, there are both slashes and quotes; this example
will fail. I have changed them to just slashes, for consistency with the
description at the top of the page.

However, I will not touch the `PathRegexp` examples, as I have not
tested them yet. Moreover, having a slash-intense literal surrounded by
slashes might not be the clearest syntax.
```

pierreprinetti:
Superseded by #893

PR #893, opened 2018-12-30
docs: Fix the Host predicate regex example

pierreprinetti:
The docs state that regular expressions should be surrounded by slashes
(`/`), but just below, the `Host` example was surrounded by double
quotes.

In the `Host` paragraph, there were both slashes and quotes; the example
would fail. I have changed them to just slashes for consistency.


szuecs:
:+1:

jrake-revelant:
👍

szuecs:
thanks @pierreprinetti

PR #895, opened 2019-01-02
feature: derive default application from api usage monitoring filter configuration (#888)

### LINKED ISSUE #888: Add default application for non configured paths in API usage monitoring
Currently, all non configured paths in the API usage monitoring filter are tracked with an unknown `application_id`. This makes it difficult to find out which filter configuration is source of the unknown events. To improve this, we can derive a default configuration for the configuration, if all API configurations of the API usage monitoring filter setup point to the same `application_id`.

The assumption is, that the configuration of the filter is usually routing against the same target.

tkrop:
fixes #888.

### DIFF
--- a/filters/apiusagemonitoring/filter.go
+++ b/filters/apiusagemonitoring/filter.go
@@ -37,6 +37,7 @@ const (
 type apiUsageMonitoringFilter struct {
 	Spec  *apiUsageMonitoringSpec
 	Paths []*pathInfo
+	UnknownPath *pathInfo // shouldn't this parameter(s) be privated?
 }
 
 func (f *apiUsageMonitoringFilter) Request(c filters.FilterContext) {
@@ -161,7 +162,7 @@ func (f *apiUsageMonitoringFilter) resolvePath(req *http.Request) *pathInfo {
 			return p
 		}
 	}
-	return f.Spec.unknownPath
+	return f.UnknownPath
 }
 
 // getEndpointMetricsNames returns the structure with names of the metrics for this specific context.
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -1091,13 +1091,6 @@ apiUsageMonitoring(`
 `)
 ```
 
-NOTE: Non configured paths will be tracked with `<unknown>` application ID, API ID
-and path template.
-
-```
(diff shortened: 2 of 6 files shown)


filters/apiusagemonitoring/filter.go
```diff
@@ -37,6 +37,7 @@ const (
 type apiUsageMonitoringFilter struct {
 	Spec  *apiUsageMonitoringSpec
 	Paths []*pathInfo
+	UnknownPath *pathInfo // shouldn't this parameter(s) be privated?
```
szuecs: No comment please, better set it as a private member and later decide to have it public in case you need it for library users

szuecs:
Please use `make fmt` to fix the format rules enforced by our CI pipeline @tkrop , thanks

filters/apiusagemonitoring/filter.go
tkrop: Is a question to the reviewer(s), whether I should change the visibility.

filters/apiusagemonitoring/filter.go
```diff
@@ -37,6 +37,7 @@ const (
 type apiUsageMonitoringFilter struct {
 	Spec  *apiUsageMonitoringSpec
 	Paths []*pathInfo
+	UnknownPath *pathInfo
```
tkrop: @szuecs @ddufourboivin-zalando shouldn't this be private properties?

maxim-tschumak:
:+1:

maxim-tschumak:
:+1:

filters/apiusagemonitoring/filter.go
tkrop: Talked with @maxim-tschumak: for the moment we leave it as it is.

szuecs: You should add json struct tags to the members and have them exported to make the JSON marshal/unmarshal or encode/decode work correctly and reliably

szuecs:
:-1: please change the struct members to have json tags, in case the struct is used in json Unmarshal/Marshal

szuecs:
:-1:

tkrop:
@szuecs It is not used in this context, and we do not intend to use it this way. @ddufourboivin-zalando's comment was misleading.

ddufourboivin-zalando:
To clarify: At the moment, it is marshalled to JSON exclusively for debug-logging purposes.

Issue #912, opened 2019-01-15
Ratelimit based on loadbalancer members / ratelimit group


szuecs:
Let's say you know your backend can handle 100 req/s per instance and you have some autoscaling in place, but you want to make sure you do not hit backends too much. You want to ratelimit, if you have more than 100 req/s for the backend. 
A backend could be an `endpoint` in kubernetes for example.

We should think about how this looks like if we rebuild the loadbalancing route, but we should be able to ratelimit a route based on the backend or loadbalancer member.

aryszka:
i agree that this would make configuring the rate limits easier, because it makes the rate limit configuration independent of the particular routes.

A step more abstract approach that I am thinking about can be grouping by custom labels:

```
rateLimitGroup("groupA", 300, "1m")
```

Where a group can represent a backend or just be a label, and any set of routes could belong this way to the same rate limit group, even ones with multiple backends. The drawback is when we really want to configure rate limits for a given backend, then this labelled configuration becomes somewhat redundant. Probably this should be an additional possibility besides the backend based rate limiting.

szuecs:
I like the idea of a rateLimitGroup via label

PR #913, opened 2019-01-15
fix clusterRatelimit with more than one ratelimit instance

### LINKED ISSUE #903: Multiple Rate Limits for Same Backend are clashing
For the scenario where you try to apply two different rate limits to two separate routes for a single backend. The rate limits appear to be clashing.

Two skipper routes are created:
Route A:
Path("/routeA") && Method("GET") -> clusterClientRatelimit(1800, "1m")
Route B:
Path("/routeB") && Method("GET") -> clusterClientRatelimit(450, "1m")

**Load Test when only one rate limit is present. (~30 req/s going to /routeA)**
screen shot 2019-01-10 at 11 41 45

**Load Test when both rate limits are present. (~30 req/s evenly spread between /routeA and /routeB)**
screen shot 2019-01-10 at 11 41 52

szuecs:
fix #903 clusterRatelimit with more than one ratelimit instance, which overwrote the shared data from other ratelimit instances.


ratelimit/ratelimit.go
```diff
@@ -299,8 +300,9 @@ func newRatelimit(s Settings, sw Swarmer) *Ratelimit {
 		s.CleanInterval = 0
 		fallthrough
 	case ClusterClientRatelimit:
+		f := fnv.New32()
 		if sw != nil {
-			impl = newClusterRateLimiter(s, sw, "CRL")
+			impl = newClusterRateLimiter(s, sw, string(f.Sum([]byte(s.String()))))
```
aryszka: looks fine, just why the hashing, why not simply s.String()?

aryszka:
:+1:

aryszka:
as we discussed the PR in its current form doesn't fully fix the problem

aryszka:
:-1:

szuecs:
After an internal discussion we agreed on having the implementation changed to https://github.com/zalando/skipper/issues/912#issuecomment-454781662 which enables full control of grouping ratelimiters across routes.
The user interface would get another option to set the `group`.

szuecs:
@aryszka the code was changed as discussed with internal customers and based on your comment https://github.com/zalando/skipper/issues/912#issuecomment-454781662.

If the code is good in the review and tested successfully in a dev cluster, I would fix all docs according to the implementation.

proxy/proxy.go
```diff
@@ -815,30 +815,37 @@ func (p *Proxy) checkRatelimit(ctx *context) (ratelimit.Settings, int) {
...
-	}
+		s := setting.Lookuper.Lookup(ctx.Request())
+		if s == "" {
+			p.log.Errorf("Lookuper found no data in request for setting: %s and request: %v", setting, ctx.Request())
+			continue
+		}
 
-	if rl.Allow(s) {
-		return settings, 0
+		allow = allow && rl.Allow(s)
```
aryszka: i think the `allow` var is not required. It could be just:

```
if !rl.Allow(s) {
    return setting, rl.RetryAfter(s)
}
```

aryszka:
lgtm

please update the branch once #932 gets merged

proxy/proxy.go
szuecs: fixed

szuecs:
:+1:

aryszka:
:+1:

szuecs:
:+1:

PR #917, opened 2019-01-17
Full TLS Config for proxy server in skipper options

UltraNemesis:
Added support to pass in a tls.Config in skipper options which would be used for the proxy HTTP server instance. 

- Allows fine grained control over TLS settings for the proxy listener including ability to setup two way SSL.
- Existing CertPathTLS and KeyPathTLS settings left unchanged and will work as fallback options so that current usages do not break.
- No changes done in skipper main.go to make use of the new config.

Signed-off-by: UltraNemesis

### DIFF
--- a/skipper.go
+++ b/skipper.go
@@ -371,6 +371,9 @@ type Options struct {
 	// multiple keys, the order must match the one given in CertPathTLS
 	KeyPathTLS string
 
+	// TLS Settings for Proxy Server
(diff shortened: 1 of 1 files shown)


szuecs:
@UltraNemesis please sign your git commits, see DCO check.

UltraNemesis:
Going to remove and create as I am having trouble amending the existing PR

Issue #908, opened 2019-01-15
Support forwarding requests


jmcs:
I have a MITM proxy that adds some credentials conditionally.

I need to have a catch all rule that modifies the request headers and then forwards the request to the requested location.

Currently the only way I found to implement this was set the backend for the catch all rule to a local nginx proxy that forwards the request.

This could be done fully in skipper if skipper had a `<forward>` backend for this purpose.

jmcs:
I added a clarification of my use case to the description.

I want to have something like:
```
* -> setRequestHeader("X-Passed-Skipper", "true") -> <forward>;
```

So that if the user requests an arbitrary domain, it gets forwarded to that arbitrary domain.

jmcs:
This seems to be a common use case as can be seen by stackoverflow questions on how to do this nginx, for example:

- https://stackoverflow.com/questions/46060028/how-to-use-nginx-as-forward-proxy-for-any-requested-location
- https://stackoverflow.com/questions/49459903/nginx-as-a-forward-proxy

szuecs:
Forward proxy can't be done right now and would need to have a backend like you suggest. 
Normally this also means, that we need to implement the CONNECT Method.
https://en.wikipedia.org/wiki/HTTP_tunnel#HTTP_CONNECT_method.

Issue #556, opened 2018-02-23
loadbalancer optimization


szuecs:
#536 will do a double lookup of routes. As optimization we should change the lbgroup/members.

lbmembers should not be a route, so also not part of the routing tree, but should be a slice of members within a loadbalanced route.

        hello_lb_group: Path("/foo") && LBGroup("hello")
                -> lbDecide("hello", 3)
                -> <loopback>;
        hello_1: Path("/foo") && LBMember("hello",0)
                -> "http://127.0.0.1:12345";
        hello_2: Path("/foo") && LBMember("hello",1)
                -> "http://127.0.0.1:12346";
        hello_3: Path("/foo") && LBMember("hello",2)
                -> "http://127.0.0.1:12347";

The member routes hello_1 hello_2 hello_3 would not be part of the routing tree, but part of the route  hello_lb_group. Everything else, for example <loopback>, should stay the same.

This will also make it easy to build different loadbalancer algorithms on top of the member slice, because these can be implemented on the current slice of members and chosen per route.

As a proposed solution we would have only one route and not use `<loopback>`.
We would use something similar to this as target, options: 
- `<lbgroup algorithm=roundrobin(args_if_required) [ep1, ep2, ..]>`
- shorter version: `<RR [ep1 ep2 ep3]>`

Another idea from @aermakov-zalando was to enhance this also with options like `retry` that could be enabled to the `lbgroup`.

aermakov-zalando:
I'd suggest adding support for more dynamic forms of this. For Kubernetes, for example, the users would want to just specify the endpoint name & namespace in the rule and have it expand into an automatically refreshed list of backends inside Skipper. This was requested a few times already and is impossible to do in the current setup without going through the Service IP that has tons of issues. It would also look nicer in Skipper route table dump, because you'd see something like `<lbgroup algorithm=… target=kube_endpoint(default, foo)>` instead of `<lbgroup algorithm=… target=addresses(<list of 100 IPs)>`.

PR #918, opened 2019-01-17
feature: support requests forwarding

Traumeel:
Introduces requests forwarding with a new backend `<forward>`

Example usage:
- `* -> <forward>`
- `* -> setRequestHeader("X-Passed-Skipper", "true") -> <forward>;`

Fixes https://github.com/zalando/skipper/issues/908

TCP tunnelling is not covered.


hjacobs:
Thanks for the PR!

szuecs:
@Traumeel can you also add user documentation in `docs/`, please?

proxy/doc.go
```diff
@@ -70,6 +70,10 @@ Routing will happen again with an augmented request.
 This can happen several times in a row until maxLoopbacks
 limit is reached.
 
+3.d forward:
+
+In case of a `forward` route, requests are forwarded to the original destination.
```
szuecs: original destination got from `Host: destination`, right?
I think this should be in the docs

proxy/proxy.go
```diff
@@ -372,9 +372,13 @@ func copyStream(to flusherWriter, from io.Reader) error {
...
 func mapRequest(r *http.Request, rt *routing.Route, host string, removeHopHeaders bool) (*http.Request, error) {
+
 	u := r.URL
-	u.Scheme = rt.Scheme
-	u.Host = rt.Host
+
+	if rt.BackendType != eskip.ForwardBackend {
+		u.Scheme = rt.Scheme
+		u.Host = rt.Host
+	}
```
szuecs: I think the Scheme is never set. In my test:

```
/bin/skipper -inline-routes '* -> setRequestHeader("X-Passed-Skipper", "true") -> <forward>'
[APP]INFO[0000] route settings, reset, route: : * -> setRequestHeader("X-Passed-Skipper", "true") -> <forward>

[APP]ERRO[0004] error while proxying, route  with backend , status code 500: dialing failed false: unsupported protocol scheme ""
::1 - - [18/Jan/2019:14:44:53 +0100] "GET / HTTP/1.1" 500 22 "-" "curl/7.49.0" 0 localexample.org - -
[APP]ERRO[0013] error while proxying, route  with backend , status code 500: dialing failed false: unsupported protocol scheme ""
::1 - - [18/Jan/2019:14:45:02 +0100] "GET / HTTP/1.1" 500 22 "-" "curl/7.49.0" 0 www.google.com - -

[APP]ERRO[0022] error while proxying, route  with backend , status code 500: dialing failed false: unsupported protocol scheme ""
::1 - - [18/Jan/2019:14:45:11 +0100] "GET / HTTP/1.1" 500 22 "-" "curl/7.49.0" 0 www.google.com - -
```

Client calls:
```
% curl localhost:9090 -H"X-Sandor: yes" -H"Host: www.google.com" -v
* Rebuilt URL to: localhost:9090/
*   Trying ::1...
> GET / HTTP/1.1
> Host: www.google.com
> User-Agent: curl/7.49.0
> Accept: */*
> X-Sandor: yes
>
< HTTP/1.1 500 Internal Server Error
< Content-Type: text/plain; charset=utf-8
< Server: Skipper
< X-Content-Type-Options: nosniff
< Date: Fri, 18 Jan 2019 13:45:11 GMT
< Content-Length: 22
<
Internal Server Error
[sszuecs@sandor-lab:~/go/src/github.com/zalando/skipper/skptesting]% curl localhost:9090 -H"X-Sandor: yes" -H"Host: http://www.google.com" -v
* Rebuilt URL to: localhost:9090/
*   Trying ::1...
> GET / HTTP/1.1
> Host: http://www.google.com
> User-Agent: curl/7.49.0
> Accept: */*
> X-Sandor: yes
>
< HTTP/1.1 400 Bad Request: malformed Host header
< Content-Type: text/plain; charset=utf-8
< Connection: close
<
* Closing connection 0
400 Bad Request: malformed Host header
```

szuecs: Ok did it wrong as offline discussed and tested it works with `--proxy`:

backend:
```
% sudo ./bin/skipper -address :80 -inline-routes 'r: * -> inlineContent("OK") -> <shunt>'
[APP]INFO[0000] proxy listener on :80
[APP]INFO[0000] route settings, reset, route: r: * -> inlineContent("OK") -> <shunt>
127.0.0.1 - - [18/Jan/2019:15:04:41 +0100] "GET / HTTP/1.1" 200 2 "-" "curl/7.49.0" 0 localexample.org - -

```

proxy:
```
% ./bin/skipper -inline-routes '* -> setRequestHeader("X-Passed-Skipper", "true") -> <forward>'
[APP]INFO[0000] route settings, reset, route: : * -> setRequestHeader("X-Passed-Skipper", "true") -> <forward>
::1 - - [18/Jan/2019:15:05:35 +0100] "GET http://localexample.org/ HTTP/1.1" 200 2 "-" "curl/7.49.0" 0 localexample.org - -
```

client
```
% curl --proxy localhost:9090 http://localexample.org -v
* Rebuilt URL to: http://localexample.org/
*   Trying ::1...
> GET http://localexample.org/ HTTP/1.1
> Host: localexample.org
> User-Agent: curl/7.49.0
> Accept: */*
>
< HTTP/1.1 200 OK
< Content-Length: 2
< Content-Type: text/plain; charset=utf-8
< Date: Fri, 18 Jan 2019 14:05:35 GMT
< Server: Skipper
<
OK
```

grassator:
This extends `eskip` language with a completely new keyword. In my view if we are changing the language a more robust strategy for the future would be to allow filters to act as backends. I remember discussing this with @aryszka briefly. The change required would be much bigger of course, but would help generalize the setup and also allow for nice features such as dynamic backends.

aryszka:
thanks a lot for the nice PR!

aryszka:
:+1:

aryszka:
@grassator right, dynamic backends would be a nice addition. Though I would still distinguish the backends from the filters in the syntax. Anyway, a POC is always welcome :smile:

aryszka:
@grassator this discussion https://github.com/zalando/skipper/issues/556 is also related to the dynamic backends.

Traumeel:
@szuecs let's merge, covers my use case

szuecs:
:+1:

UltraNemesis:
Why does the scheme for forwarding need to be set separately at a global level? 

As I see it, this is a very basic forward proxy scenario being baked into what has so far been a reverse proxy. The fundamental difference with the request meant for a forward proxy is that the RequestURI will contain the full target URI including scheme and host while in normal cases, only path will be present. So, both target host and scheme should be obtained from RequestURI and used to make the forward request.

Also, I don't see the utility of handling "CONNECT" to enable tunneling even in future. The main USP of skipper is being able to conditionally modify request/responses in the process of routing. With tunneling, that ability will be removed as all traffic passing through will be encrypted with skipper only maintaining a tunnel.

UltraNemesis:
@szuecs What you say about CONNECT makes perfect sense from that perspective. In fact, if skipper is going be usable as a forward proxy, I think you should also think about having an MITM mode for this as an option so that requests can be intercepted and predicates/filters can be applied before forwarding even in the HTTPS mode. 

What I don't still understand is the need for the default scheme flag. The RequestURI will have full URI including scheme and in any case, without CONNECT handling, all requests will have to be HTTP.

Also, with regard to my own use case, my scenario about was about being able to select the target host from a filter based on headers or body while the requests are still made with skipper as a reverse proxy. I had setup something like this at my work using an F5 BIG IP iRule and evaluating whether I can replace it with skipper and a custom filter.

The addition of the "forward" backend type makes this easier, but what we will still need to have is a method like SetTargetHost in the filter context that can be used to set a target host for a request and skipper should make the request to that host.

szuecs:
@UltraNemesis The following should work with `forward` in your case:

```
r: * -> setRequestHeader("Host", "foo.example.org") -> <forward>
```

With the use of `setRequestHeader` filter and instead of `*`, [Predicates](https://opensource.zalando.com/skipper/reference/predicates/) you can also build more complex routes as you wish, the question is if you need `<forward>` in this case, because you have to already know in advance what kind of Host header you need.

Answering inline:

> @szuecs What you say about CONNECT makes perfect sense from that perspective. In fact, if skipper is going be usable as a forward proxy, I think you should also think about having an MITM mode for this as an option so that requests can be intercepted and predicates/filters can be applied before forwarding even in the HTTPS mode.

Sounds good to me, CONNECT does not need to be only TCP pass through.

> What I don't still understand is the need for the default scheme flag. The RequestURI will have full URI including scheme and in any case, without CONNECT handling, all requests will have to be HTTP.

The problem is that the RequestURI only contains path and query, there is no Scheme if you get a Request object passed.
See this example:

```
% cat >main.go
package main

import (
        "fmt"
        "io/ioutil"
        "log"
        "net/http"
)

func runServer() {

        http.HandleFunc("/foo", func(w http.ResponseWriter, r *http.Request) {
                fmt.Fprintf(w, "Hello, '%q' '%s'", r.URL.Scheme, r.URL)
        })

        log.Fatal(http.ListenAndServe(":8081", nil))
}

func main() {
        go runServer()

        resp, err := http.Get("http://127.0.0.1:8081/foo?q=hello")

        if err != nil {
                log.Fatalf("Failed to Get: %v", err)
        }
        defer resp.Body.Close()
        body, err := ioutil.ReadAll(resp.Body)
        if err != nil {
                log.Fatalf("Failed to Get: %v", err)
        }
        fmt.Printf("%s\n", body)
}
[sszuecs@uho:/tmp/foo]% go run main.go 
Hello, '""' '/foo?q=hello'
```

> 
> Also, with regard to my own use case, my scenario about was about being able to select the target host from a filter based on headers or body while the requests are still made with skipper as a reverse proxy. I had setup something like this at my work using an F5 BIG IP iRule and evaluating whether I can replace it with skipper and a custom filter.

Can you explain your use case in a separate issue?
We can discuss there how this could be done.

> 
> The addition of the "forward" backend type makes this easier, but what we will still need to have is a method like SetTargetHost in the filter context that can be used to set a target host for a request and skipper should make the request to that host.

-> `setRequestHeader("myheader", "myValue")`

szuecs:
👍

UltraNemesis:
@szuecs : The problem in the code you posted is that you are not making a forward proxy compatible HTTP request. Try setting the server as forward proxy

```
	proxyURL, _ := url.Parse("http://127.0.0.1:8081")
	client := &http.Client{}

	if proxyURL != nil {
		client.Transport = &http.Transport{
			Proxy: http.ProxyURL(proxyURL),
		}
	}
        resp, err := client.Get("http://127.0.0.1:8080/foo?q=hello")
```

I recommend reading through this

https://parsiya.net/blog/2016-07-28-thick-client-proxying---part-6-how-https-proxies-work/

When you make the request to a forward proxy, the specs require that the full Request URI including the scheme is sent instead of just relative URI like in normal cases.

Here is what the dump looks like

**Direct Request**

```
GET /foo?q=hello HTTP/1.1
Host: 127.0.0.1:8080
Accept-Encoding: gzip
User-Agent: Go-http-client/1.1
```
**Proxy Request**

```
GET http://127.0.0.1:8080/foo?q=hello HTTP/1.1
Accept-Encoding: gzip
User-Agent: Go-http-client/1.1
```

If skipper is going to act as a forward proxy with this implementation, it should go with the established standards.

UltraNemesis:
Here is code including a fully functional forward proxy implementation

```
package main

import (
	"fmt"
	"io"
	"io/ioutil"
	"log"
	"net"
	"net/http"
	"net/http/httputil"
	"net/url"
	"time"
)

type targetHandler struct{}

type proxyHandler struct{}

func main() {
	log.Println("Starting Proxy Server")
	go runProxyServer()

	log.Println("Starting Target Server")
	go runServer()

	proxyURL, _ := url.Parse("http://127.0.0.1:8888")

	log.Println("Make Direct HTTP Request")
	makeRequest(nil)

	log.Println("Make Proxy HTTP Request")
	makeRequest(proxyURL)
}

func makeRequest(proxyURL *url.URL) {
	client := &http.Client{}

	if proxyURL != nil {
		client.Transport = &http.Transport{
			Proxy: http.ProxyURL(proxyURL),
		}
	}

	resp, err := client.Get("http://127.0.0.1:8080/foo?q=hello")

	if err != nil {
		log.Fatalf("Failed to Get: %v", err)
	}
	defer resp.Body.Close()
	body, err := ioutil.ReadAll(resp.Body)
	if err != nil {
		log.Fatalf("Failed to Get: %v", err)
	}
	log.Printf("%s\n", body)
}

func runServer() {
	log.Fatal(http.ListenAndServe(":8080", &targetHandler{}))
}

func runProxyServer() {

	http.ListenAndServe(":8888", &proxyHandler{})
}

func (h *targetHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
	bytes, _ := httputil.DumpRequest(req, true)

	fmt.Println(string(bytes))

	fmt.Fprintf(rw, "Hello, '%q' '%s'", req.URL.Scheme, req.URL)
}

func (h *proxyHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
	if req.Method == http.MethodConnect {
		handleTunneling(rw, req)
	} else {
		handleHTTP(rw, req)
	}

}

func handleHTTP(rw http.ResponseWriter, req *http.Request) {
	log.Printf("Request URI : %s", req.RequestURI)

	bytes, _ := httputil.DumpRequest(req, true)

	fmt.Println(string(bytes))

	newReq, _ := http.NewRequest(req.Method, req.RequestURI, req.Body)

	copyHeader(newReq.Header, req.Header)

	resp, err := http.DefaultClient.Do(newReq)

	if err != nil {
		log.Println(err)
	} else {
		copyHeader(rw.Header(), resp.Header)

		io.Copy(rw, resp.Body)

	}

}

func handleTunneling(rw http.ResponseWriter, req *http.Request) {
	target, err := net.DialTimeout("tcp", req.Host, 10*time.Second)

	if err != nil {
		http.Error(rw, err.Error(), http.StatusServiceUnavailable)
		return
	}

	rw.WriteHeader(http.StatusOK)

	hijacker, ok := rw.(http.Hijacker)

	if !ok {
		http.Error(rw, "Hijacking not supported", http.StatusInternalServerError)
		return
	}

	source, _, err := hijacker.Hijack()

	if err != nil {
		http.Error(rw, err.Error(), http.StatusServiceUnavailable)
	}
	go pipe(target, source)
	go pipe(source, target)
}

func pipe(target io.WriteCloser, source io.ReadCloser) {
	defer target.Close()
	defer source.Close()
	io.Copy(target, source)
}

func copyHeader(dst, src http.Header) {
	for k, vv := range src {
		for _, v := range vv {
			dst.Add(k, v)
		}
	}
}

```

Traumeel:
@UltraNemesis thx for the nice read. 

> https://parsiya.net/blog/2016-07-28-thick-client-proxying---part-6-how-https-proxies-work/

> When you make the request to a forward proxy, the specs require that the full Request URI including the scheme is sent instead of just relative URI like in normal cases.

So this should hold true for a proxy aware client, in case it is not URIs will be relative. If it is the case we need to reconstruct it.

Going to change the PR to handle it

UltraNemesis:
@Traumeel Yes, a properly implemented forward proxy aware client will always send the request in that format. You can count on it being that way.

To be honest, you don't really need to read the scheme at all. Any forward proxy aware client will send a CONNECT request to establish a tunnel if the target URL scheme is HTTPS . So, if you get anything other than a CONNECT, it will be HTTP.

Traumeel:
So, by default forwarding will work for HTTP only.

To access HTTPS URLs `forward-scheme-https` flag can be used, it will upgrade all requests with `<forward>` backends to HTTPS. This basically mimics the current behaviour of HTTP endpoint backend, where you can set the scheme explicitly. Proxy receives HTTP request from a client, adds headers (optionally) and sends HTTPS request to the target.

UltraNemesis:
Then, it is no longer a proper forward or reverse proxy scenario. 

If you configure skipper as a forward proxy, A forward proxy aware client will automatically send a CONNECT request for HTTPS URLs. So, if you have HTTPS endpoint, you will have to manually change the request to HTTP first at the client side before sending it to skipper just so that skipper will accept it then change it HTTPS based on the flag. Further more, you have no control as its a global scheme. 

It is no good as a reverse proxy case either as you would need to construct a custom HTTP request with Host header set to actual target rather than skipper and then send it to skipper. Such an implementation would screw up any chance to support SNI and Virtual Host based TLS in skipper in future.

So, I highly recommend sticking to the specs for out of the box behavior be it as a forward proxy ore reverse proxy. If any custom/overriding behavior like this is required, provision for that should be made as part of the filters framework. 

In fact, I too have a pretty similar custom requirement that would fit on top of this. In my case, I want to make a request to skipper and I want to decide based on headers or body content which destination I want to send it to. Only difference from your case is that you just want to modify the scheme in the URL rather than whole URI. This is best fit into the filter system.  Have a method in filter context which allows setting the Target Backend for the request.

szuecs:
@UltraNemesis as far as I get @Traumeel, the use case is to have a "malicious" DNS server that responds to all lookups to point to skipper, so clients are not aware to have a proxy MITM.
I think `<forward>` should be more the backend for the `CONNECT` case and we should have some other name for it. The Problem with the `Scheme` in the request is that skipper acting as MITM proxy does not know which backend protocol to use. 

There are multiple options here:
1. We could have a filter that copies a header to the `statebag` and the backend `<mitm>` could restore the value from the `statebag`. 
2. We could have a filter that copies a header to the `statebag` and a filter `statebag` to header and the backend `<mitm>` could restore the value from a hardcoded header, which could enable similar use cases as for example linkerd has with the overwrite route from a request (in our case we would have to set the backend to enable this behavior)

Traumeel:
@aryszka 

> As I understand, @Traumeel 's use case, it would be required to accept HTTP requests and forward them to HTTPS. Is this correct?

Correct. 
My current setup works in the following way: 
- Hosts configured with bind9 which routes request to Skipper
- Skipper is behind ELB (HTTPS is terminated)
- Skipper injects headers and forwards

The setups works nicely for a predefined set of hosts, but it is cumbersome to update it with new hosts.

> For the scheme, I would avoid using a global default scheme, instead I would create a filter that overrides the scheme of the incoming request:
> 
> ```
> * -> setForwardScheme("https") -> <forward>
> ```
> 
> @Traumeel would this approach work for your use case?

If I understand this correctly `setForwardScheme` would be a builtin filter. In general I think this approach might work, but I need to propagate this setting to request mapping somehow. As far as my understanding goes there are two possibilities: extend `FilterContext` interface or use `StateBag`. In both cases it would be possible to reuse this functionality from a custom filter.

Thus I was thinking to go with extending `FilterContext` interface to provide a possibility to redefine both host and scheme from a filter. Which will probably work in the same way `SetOutgoingHost` works. In this way filters will get full control of the final destination. Not sure if there are any implications  which I don't currently see with this approach or if there is a technical blocker somewhere, but looks good to me at the moment.

@aryszka @szuecs What do you think? This will solve my use case and @UltraNemesis 

The functionality can optionally be enabled for only a special backend type e.g. `<filter-defined-host>`

szuecs:
After an offline discussion with  @Traumeel I have the following idea:
We restore all request data from headers of the `http.Request`. 
Like this everyone can use `filters` to modify the backend target, if the backend is type `<filter-defined-host>` (or maybe better `<request-defined-backend>` ?).

What do we need to do the call to do the `http.Request` similar to the call the original client?
- `http.Request` has all data + data from a proxy in between (`X-Forwarded-*`), but it has no `http.Request.Scheme` defined which is OSI layer below HTTP so the parser of the http data does not get data ti reconstruct `http.Request.Scheme`.
- `http.Request.Scheme` can be restored by the value of `X-Forwarded-Proto` header [rfc link](https://tools.ietf.org/html/rfc7239#section-5.4), if there is a http-proxy in between. For other cases you can use plain filter to modify the `http.Request.Headers` to achieve the same goal. 

Wit this implementation you can even write your custom filter and lookup a database to change the final request being made to an arbitrary chosen backend.

Any drawbacks with this you see @aryszka @UltraNemesis ?

aryszka:
@szuecs where would be the backend host taken from in this case? From the Host header?

szuecs:
@aryszka yes we would use the target from the Host header

aryszka:
In the mean time, I also collected my thoughts on this and here's an alternative suggestion. This is not completely different from the one @szuecs layed out.

Supported scenarios (and one unsupported):

**Scenario 1: Skipper listens on HTTP, the client sends a proxy request**

In this case the RequestURI will contain the right backend scheme and host. If the route has a `<forward>` backend, then we can just take the scheme and the host from the RequestURI.

**Scenario 2: Skipper listens on HTTPS, the client sends a proxy request**

The proxy request in this case will be a CONNECT request. I suggest not to support this scenario in this PR.

**Scenario 3: Skipper listens on either HTTP or HTTPS, the client sends a normal request, e.g. because of DNS hijack**

In these cases the RequestURI will contain only the path and the query but the request will contain the Host header. If the route has a `<forward>` backend, we can take the host from the Host header.

The more problematic part is the right scheme in this case. Since skipper can listen only either on HTTP or HTTPS, and the client in this case hits Skipper directly, for the most cases it should be fine to use the same scheme for the backend as Skipper listens on. Note that the client doesn't "know" that Skipper is in the request chain, so they certainly had to mean the same scheme as Skipper listens on, otherwise the request will fail before reaching the routing.

If we know that the backend listens on a different scheme than Skipper, then we could override this default with a filter setForwardScheme("https") explicitly in the route config, using a Host() predicate to distinguish those routes that require a special handling.

**Scenario 4: Skipper listens on HTTP, behind a load balancer, the client sends a proxy request**

Questions:
- is this a valid scenario?
- will the LB forward the request as a proxy request?

Depending on the second question, if yes, then the same should apply as in Scenario 1. If not then we can take the host from the Host header, and we can take the scheme from the X-Forwarded-Proto header.

**Scenario 5: Skipper listens on HTTP, behind a load balancer, the client sends a normal request**

This case is the same as Scenario 3, but we can use here the X-Forwarded-Proto header to decide about the right scheme.

**Suggested mechanism:**

If we can agree about the above scenarios, then I believe the following logic should support each of them:

Scheme decision, if the backend is `<forward>`:

1. If a filter has set the forward scheme, use it, done
2. If the RequestURI contains the scheme, use it, done
3. If there is an X-Forwarded-Proto header, use it, done
4. Use the same scheme as what Skipper listens on: HTTP or HTTPS

Host decision, if the backend is `<forward>`:

1. If the RequestURI contains the host, use it, done
2. Use the host from the Host header

aryszka:
@szuecs As you may see, my suggestion is not much different from yours, except that it also considers the proxy requests, where the RequestURI contains the scheme and the host. You're right that setting the X-Forwarded-Proto with a filter is equivalent to the setForwardScheme(), maybe just not as explicit.

Not that when a client sends a real proxy request, then the Host header will be missing.

aryszka:
@UltraNemesis see my updated proposal. It is very similar to the last suggestion of @szuecs , I just believe it is more complete in some of the details.

Regarding the way the setForwardScheme() filter would work: using the StateBag, or having an explicit method on the FilterContext, or just using the setRequestHeader("X-Forwarded-Proto"), I have no strong preferences, either way is fine with me.

szuecs:
@aryszka I am fine with your suggestion and checked that in Scenario 4 the answer of Q2 in case of ALB is: no

UltraNemesis:
Also just to clarify one more thing, A forward proxy's listener scheme has nothing to with the scheme of the forwarded requests as was suggested in earlier posts.

A forward proxy listening on HTTP can handle both HTTP and HTTPS outgoing requests. When a proxy request comes in for a HTTP URI, it will take the RequestURI and makes request to it and and returns the response. For HTTPS URI, the client will send a CONNECT request to the proxy which will then make a connection to the target pointed in the CONNECT and returns a 200 response. Then the client will send a TLS handshake which the proxy will simply pipe to the target connection. That is why its a tunnel and the proxy simply acts as a packet forwarder. 

Even if we enable HTTPS for a forward proxy, it simply means that the traffic between the client and the proxy is also encrypted. But this also means that when a HTTPS URL is requested, it will  end up being a double wrapping of TLS encryption.

szuecs:
@UltraNemesis Thanks for your valuable comments, but I am not sure if I completely understand/follow your idea. 

- How would you support the use case of @Traumeel, which was nicely written by @aryszka [Scenario 4](https://github.com/zalando/skipper/pull/918#issuecomment-456454259)?
- What does the suggested `SetBackend` filter get as argument to make it different from the `<forward>` backend?

szuecs:
@UltraNemesis thanks for your writing. I agree skipper should (was and will be) implement all the features that are publicly standardized conform to the standard (for example forward proxy).

Skipper right now is only a http reverse proxy, we do not have forward proxy nor a TCP (l4) proxy implementation, yet.
I created [a new issue to support forward proxy](https://github.com/zalando/skipper/issues/929) in the future.

The current PR from @Traumeel tries to implement a reverse proxy with forwarding requests, basically MITM transparent to the client, because of the **malicious** DNS the directs all clients to skipper. It is a special use case which is not defined in an RFC as far as I know.

### MITM use case implementation proposed by @aryszka 

@aryszka wrote the implementation could be like this, which should not hurt anyone, because it has to be selected by the user for a particular route that has the backend `<forward>`:

> Scheme decision, if the backend is `<forward>`:
> 
> 1. If a filter has set the forward scheme, use it, done
> 2. If the RequestURI contains the scheme, use it, done
> 3. If there is an X-Forwarded-Proto header, use it, done
> 4. Use the same scheme as what Skipper listens on: HTTP or HTTPS
> 
> Host decision, if the backend is `<forward>`:
> 
> 1. If the RequestURI contains the host, use it, done
> 2. Use the host from the Host header

### scenario from @UltraNemesis 

> Also, with regard to my own use case, my scenario about was about being able to select the target host from a filter based on headers or body while the requests are still made with skipper as a reverse proxy. I had setup something like this at my work using an F5 BIG IP iRule and evaluating whether I can replace it with skipper and a custom filter.

For me the use case is valid and also not a forward proxy, but a reverse proxy as @UltraNemesis already wrote, so we can consider forward proxy out of scope and try to get an agreement how to make the custom forwarding work for both use cases.

### changed proposal by @szuecs 

I don't like the current name of the backend `<forward>`, because it interferes with the idea forward-proxy, so we should use some other name that is maybe more clear: `<dynamic>`

I would suggest to have a step 0 in the decision for the backend target from @aryszka's proposal:

Scheme decision, if the backend is `<dynamic>`:
 
0. If there is a `statebag` value for key `"backend-url"`, use it, done 
1. If a filter has set the forward scheme, use it, done
2. If the RequestURI contains the scheme, use it, done
3. If there is an X-Forwarded-Proto header, use it, done
4. Use the same scheme as what Skipper listens on: HTTP or HTTPS

Host decision, if the backend is `<dynamic>`:
0. If we could restore the target URL from statebag, use it, done
1. If the RequestURI contains the host, use it, done
2. Use the host from the Host header

With the statebag entry you are free to do anything with your customer filter and set the dynamic backend. Maybe we can even drop all the other steps and implement filters that can copy http data to this statebag key or to an arbitrary statebag key.

aryszka:
@UltraNemesis thanks for the explanation. I see the point of separating the forward proxy scenario from the one where the reverse proxy simply sets the backend from a filter, and how relying on the Host header in the forward proxy scenario is not standard.

I also like the `SetBackend()` suggestion.

OTH, there's @Traumeel 's use case, where they want to proxy requests, without the client knowning about that there is a proxy. This means that it's not a real forward proxy scenario. In addition to that, they ideally want to avoid maintaining every possible backend address in the configuration, therefore they need to infer the backends from the request 'somehow'. (If I got the use case correctly.)

So we have the following use cases:

1. forward proxy,
2. dynamic backends set by filters,
3. dynamic backends taken from the request 'somehow'.

I think the latest proposal from @szuecs supports all of these. It just has this potential conceptual problem that it can automatically fall back from a forward proxy mode to a reverse proxy with dynamic backends. Do you think this can lead to practical problems?

If it can be fine, then @szuecs 's suggestion sounds fine to me. I also like the `<dynamic>` notation.

// --

Anecdotically, we resisted dynamic backends for a long time by design and I was one of the main opponents, but we have learned about too many possible use cases :)

UltraNemesis:
1. For a standard reverse proxy case, Request will never contain Host or scheme in Request URI. They are only available in Forward proxy request.  So, we can throw those options right away if we keep forward proxy scenario out of scope.

2. I would much rather that the decision to use host header or a scheme be explicit rather than implicit inside skipper code. 

Here is the much simplified approach that I propose instead

* If there is a `statebag` value for key `"backend-url"`, use it, done.
* If backend-url key is not found, return a 500 or 501 response, done.

Optionally, include a message body to indicate what went wrong along with the 500/501 response.

Basically "dynamic" will be a backend type that requires at least one filter to be present to set the backend-url. No assumptions from skipper core and it fits what "dynamic" represents. If no filter is configured, it will throw a HTTP 500/501 error since it doesn't know where to sent the request.

In addition, We can have a out of box filter like useHostHeaderAsBackend() that will set the backend-url to the host header value and maybe scheme can be passed as an option to it. Another filter could be something like useBackend

In summary

`route1: * ->  dynamic;`

Will always result in 500 or 501 response to the client.

`route1: * -> useHostHeaderAsBackend("https") -> dynamic;`

Use the host header as backend and take the scheme passed as option.

`route2: * -> useBackend("https://custom.backend.com") -> dynamic;`

Set the backend to a custom value passed in.

Traumeel:
Modified PR with the version which uses SetBackend(url) with `<dynamic>` backend.

todo:
- fail during route parsing if a route with `<dynamic>` backend has no filters
- implement built-in filter(s) with a default scheme/host decision

aryszka:
@Traumeel I don't think this is important for this particular feature:

> fail during route parsing if a route with <dynamic> backend has no filters

we have other cases with possibly invalid routes. The semantic validation should be a separate general feature. I like the idea though!

UltraNemesis:
Agreed @aryszka  

I think it is good enough for now, if skipper just returns a HTTP 500/501 code when backend url is not set for the dynamic backend.

aryszka:
@UltraNemesis your proposal sounds good to me. If it supports @Traumeel 's use case then let's go with it.

szuecs:
I agree your suggestion is simpler and cleaner.

I would think we should not have a filter `useHostHeaderAsBackend`, but maybe better these examples:

```
* -> copyHeaderToStatebag("Host", "backend-host") -> copyHeaderToStatebag("X-Forwarded-Proto", "backend-scheme") -> <dynamic>

* -> copyHeaderToStatebag("X-My-Custom-Choice", "backend-url") -> <dynamic>

```

- backend-url is a parsed URL object
- backend-host is hostname as string
- backend-scheme is the URL scheme as string

Traumeel:
Usage:

```
foo: * -> setDynamicBackendHostFromHeader("X-Forwarded-Host") -> <dynamic>;
foo: * -> setDynamicBackendSchemeFromHeader("X-Forwarded-Proto") -> <dynamic>;
foo: * -> setDynamicBackendUrlFromHeader("X-Custom-Url") -> <dynamic>;
foo: * -> setDynamicBackendHost("example.com") -> <dynamic>;
foo: * -> setDynamicBackendScheme("https") -> <dynamic>;
foo: * -> setDynamicBackendUrl("https://example.com") -> <dynamic>;
```

Value set by `setDynamicBackendUrl` or `setDynamicBackendUrlFromHeader` would have a priority if used in combination with filters which set target url partially.

docs/reference/architecture.md
```diff
@@ -16,11 +16,12 @@ always a current routing table which will be replaced on change.
 A route is one entry in the routing table. A route consists of one or
 more `predicate`, that are used to find a route for a given HTTP
 request. A route can also have one or more `filter`, that can modify
-the content of the request or response.  A route always points to a
-backend, or it's a `<shunt>`, meaning that skipper serves the requests
-for the route, or a `<loopback>`, meaning that the request will be
+the content of the request or response.  A route can points to a
```
szuecs: A route can points to -> A route can point to

docs/reference/architecture.md
```diff
@@ -16,11 +16,12 @@ always a current routing table which will be replaced on change.
...
+backend, it can be a `<shunt>`, meaning that skipper serves the requests
+for the route, a `<loopback>`, meaning that the requests will be
 matched against the routing table again after filters have modified
-it.
+them, or a `<dynamic>`, meaning that the target url can be set dynamically 
+by a filter.
```
szuecs: I think here should be also a reference to the relevant filters

docs/reference/filters.md
```diff
@@ -931,6 +931,104 @@ Example:
 auditLog()
 ```
 
+
+## setDynamicBackendHostFromHeader
+
+Filter sets the backend host for a route, value is taken from the provided header.
+Can be used only with `<dynamic>` backend. Meant to be used together with [setDynamicBackendSchemeFromHeader](#setdynamicbackendschemefromheader)
+or [setDynamicBackendScheme](#setdynamicbackendscheme). If this filter chained together with [setDynamicBackendUrlFromHeader](#setdynamicbackendurlfromheader) 
+or [setDynamicBackendUrl](#setdynamicbackendurl) filters, the latter ones would have a priority.
```
szuecs: would have a priority ->  would have priority

docs/reference/filters.md
```diff
@@ -931,6 +931,104 @@ Example:
...
+
+```
+foo: * -> setDynamicBackendHost("example.com") -> <dynamic>;
+```
+
+## setDynamicBackendScheme
+
+Filter sets the backend scheme for a route. Can be used only with `<dynamic>` backend.
+Meant to be used together with [setDynamicBackendHostFromHeader](#setdynamicbackendhostfromheader)
+or [setDynamicBackendHost](#setdynamicbackendhost).If this filter chained together with
```
szuecs: t).If -> t). If

docs/tutorials/basics.md
```diff
@@ -20,11 +20,11 @@ always a current routing table which will be replaced on change.
 A route is one entry in the routing table. A route consists of one or
 more `predicate`, that are used to find a route for a given HTTP
 request. A route can also have one or more `filter`, that can modify
-the content of the request or response.  A route always points to a
-backend, or it's a `<shunt>`, meaning that skipper serves the requests
-for the route, or a `<loopback>`, meaning that the request will be
+the content of the request or response.  A route can points to a
```
szuecs: A route can points -> A route points

eskip/doc.go
```diff
@@ -186,38 +186,43 @@ predicates start with upper case, while filters start with lower case.
 
 Backend
 
-There are three types of backends: a network endpoint address, a shunt or
-a loopback.
+There are four backend types: HTTP endpoint, shunt, loopback and dynamic.
 
-A network endpoint address example:
+HTTP endpoint:
```
szuecs: I would not change this, because it highlights more the "network", that maps in the code to `NetworkBackend` for example and this document is meant for library users that want to understand the code.

filters/builtin/dynamicbackendfilter.go
```diff
@@ -0,0 +1,159 @@
...
+	setDynamicBackendHostFromHeader dynamicBackendFilterType = iota
+	setDynamicBackendSchemeFromHeader
+	setDynamicBackendUrlFromHeader
+	setDynamicBackendHost
+	setDynamicBackendScheme
+	setDynamicBackendUrl
+)
+
+// common structure for dynamic backend filter
+// filters
```
szuecs: please drop the generic comment

eskip/eskip.go
```diff
@@ -271,16 +275,25 @@ func newRouteDefinition(r *parsedRoute) (*Route, error) {
...
+	backendFlags := []bool{shunt, loopback, dynamic}
+	var counter int
+	for _, flag := range backendFlags {
+		if flag {
+			counter++
+			if counter > 1 {
+				err = errInvalidBackend
+				return
+			}
+		}
```
szuecs: I don't see why we need this code validation for the flags here.

filters/filters.go
```diff
@@ -8,6 +8,17 @@ import (
...
 
+const (
+	// DynamicBackendHostKey is the key used in the state bag to pass host to the proxy.
+	DynamicBackendHostKey = "statebag:backend:dynamic:host"
+
+	// DynamicBackendSchemeKey is the key used in the state bag to pass scheme to the proxy.
+	DynamicBackendSchemeKey = "statebag:backend:dynamic:scheme"
+
+	// DynamicBackendURLKey is the key used in the state bag to pass url to the proxy.
+	DynamicBackendURLKey = "statebag:backend:dynamic:url"
```
szuecs: Please drop the prefix `statebag:`, which is clear in the context of use

filters/filters.go
```diff
@@ -57,7 +68,7 @@ type FilterContext interface {
 	StateBag() map[string]interface{}
 
 	// Gives filters access to the backend url specified in the route or an empty
-	// value in case it's a shunt or loopback
+	// value in case it's a shunt, loopback. In case of dynamic backend initially empty.
```
szuecs: In case of dynamic backend initially empty. -> In case of dynamic backend initially empty and set by a filter.

filters/filtertest/filtertest.go
```diff
@@ -5,10 +5,9 @@ FilterContext interfaces used during tests.
 package filtertest
 
 import (
-	"net/http"
-
 	"github.com/opentracing/opentracing-go"
 	"github.com/zalando/skipper/filters"
+	"net/http"
```
szuecs: please drop this change

proxy/doc.go
```diff
@@ -70,6 +70,10 @@ Routing will happen again with an augmented request.
 This can happen several times in a row until maxLoopbacks
 limit is reached.
 
+3.d forward:
```
szuecs: forward is now `dynamic`

proxy/doc.go
szuecs: `forward` is now `dynamic`

proxy/proxy.go
```diff
@@ -376,11 +377,33 @@ func copyStream(to flusherWriter, from io.Reader) error {
...
 // creates an outgoing http request to be forwarded to the route endpoint
 // based on the augmented incoming request
-func mapRequest(r *http.Request, rt *routing.Route, host string, removeHopHeaders bool) (*http.Request, error) {
+func mapRequest(r *http.Request, rt *routing.Route, host string, removeHopHeaders bool, stateBag map[string]interface{}) (*http.Request, error) {
+
 	u := r.URL
 	u.Scheme = rt.Scheme
 	u.Host = rt.Host
 
+	if rt.BackendType == eskip.DynamicBackend {
```
szuecs: body of this if should be refactored into a function

docs/operation/operation.md
```diff
@@ -78,7 +78,6 @@ implementation of DialContext, which is the TCP connection pool used in the
     -enable-dualstack-backend
         enables DualStack for backend connections (default true)
 
-
```
szuecs: please delete this change

szuecs:
@Traumeel the filters you implemented should also be registered in `filters/builtin/builtin.go`

Traumeel:
@szuecs new filters are registered in `filters/builtin/builtin.go`. Fixed comments.

szuecs:
@UltraNemesis maybe you want to check this again, if it's ok for you?

aryszka:
:+1:

aryszka:
@Traumeel thanks a lot for the PR

filters/builtin/dynamicbackendfilter.go
```diff
@@ -0,0 +1,157 @@
...
+	case setDynamicBackendUrlFromHeader:
+		return SetDynamicBackendUrlFromHeader
+	case setDynamicBackendHost:
+		return SetDynamicBackendHost
+	case setDynamicBackendScheme:
+		return SetDynamicBackendScheme
+	case setDynamicBackendUrl:
+		return SetDynamicBackendUrl
+	default:
+		panic("invalid type")
```
szuecs: panic is not good at runtime, better unknown type

filters/builtin/dynamicbackendfilter.go
```diff
@@ -0,0 +1,157 @@
...
+	case setDynamicBackendUrl:
+		ctx.StateBag()[filters.DynamicBackendURLKey] = f.input
+		bu, err := url.ParseRequestURI(f.input)
+		if err == nil {
+			ctx.SetOutgoingHost(bu.Host)
+		}
+	}
+}
+
+func (f *dynamicBackendFilter) Response(ctx filters.FilterContext) {}
```
szuecs: you don't need the names only the types in an empty function

eskip/eskip.go
```diff
@@ -151,6 +152,8 @@ func (t BackendType) String() string {
 		return "shunt"
 	case LoopBackend:
 		return "loopback"
+	case DynamicBackend:
+		return "dynamic"
 	default:
 		return "unknown"
```
szuecs: "unknown" should be a const and might be reused at  https://github.com/zalando/skipper/pull/918/files#diff-f983044d09504437f499749382321472R110 (maybe I am wrong and it's something else but looks similar)

filters/builtin/dynamicbackendfilter.go
UltraNemesis: Do we need to define so many state bag keys? I would say that we should have only one key for setting  Backend Url and all other filters should just set this value accordingly. 

To set scheme, all you need to do is take current backend url, change the scheme and set the new value back to backend url. 

To take backend url form host, you read its value and set the value to Backend url key.

UltraNemesis:
Do we need to define so many state bag keys? I would say that we should have only one key for setting Backend Url and all other filters should just map and set this value accordingly.

To set scheme, all you need to do is take current backend url, change the scheme and set the new value back to backend url.

To take backend url from host, you read its value and set the value to Backend url key.

Every thing can be handled through a single key.

filters/builtin/dynamicbackendfilter.go
aryszka: I think this one should be fine.

I didn't comment on this panic because the spec the values of typ can be set only from inside the package, but the spec initializers, so this panic basically cannot happen. Since something is required for the default, it is simply a matter of preference whether the impossible panic should happen here or delegated to somewhere else.

I quickly went through the other panics in the filters, most of them are similar to this case. Excaptions are the flowid and the compress packages, where the panics, if happen, happen only during the startup phase, meaning they prevent the program from starting up with an invalid state. So they should be fine, too.

aryszka:
about the two statebag keys: @Traumeel @UltraNemesis , I don't have a strong opinion on it, I find both ways fine.

szuecs:
:+1:

szuecs:
Thanks all reviewers and @Traumeel for this PR !

PR #919, opened 2019-01-18
Full TLS Config for proxy server in skipper options

UltraNemesis:
Added support to pass in a tls.Config in skipper options which would be used for the proxy HTTP server instance. 

- Allows fine grained control over TLS settings for the proxy listener including ability to setup two way SSL.
- Existing CertPathTLS and KeyPathTLS settings left unchanged and will work as fallback options so that current usages do not break.
- No changes done in skipper main.go to make use of the new config.

Signed-off-by: UltraNemesis <dev.ajayabraham@gmail.com>

### DIFF
--- a/skipper.go
+++ b/skipper.go
@@ -371,6 +371,9 @@ type Options struct {
 	// multiple keys, the order must match the one given in CertPathTLS
 	KeyPathTLS string
 
+	// TLS Settings for Proxy Server
+	ProxyTLS *tls.Config
+
 	// Flush interval for upgraded Proxy connections
 	BackendFlushInterval time.Duration
 
@@ -632,7 +635,7 @@ func initLog(o Options) error {
 }
 
 func (o *Options) isHTTPS() bool {
-	return o.CertPathTLS != "" && o.KeyPathTLS != ""
+	return (o.ProxyTLS != nil) || (o.CertPathTLS != "" && o.KeyPathTLS != "")
 }
 
 func listenAndServe(proxy http.Handler, o *Options) error {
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

szuecs:
@UltraNemesis I would be also happy if you would add it to the flags in main.go to make it usable from the CLI, but `tls.Config` is quite a big struct, which would require quite a lot of options to add.

@aryszka what do you think about the name `ProxyTLS` in `skipper.Options`?

UltraNemesis:
@szuecs, As you mentioned, tls.Config is quite big and there are way too many things that can be customized through it for them all to be mapped to flags in main.go. Apart form the normal fields, it also contains a number of func fields like GetCertificate, GetClientCertificate, GetConfigForClient, VerifyPeerCertificate which are quite useful when you want to embed skipper and build a custom version with heavy customization to TLS behavior.  I am of the opinion that only a few important settings should be exposed in flags and the rest should be seen as flexibility to a developer making a custom build.

szuecs:
@UltraNemesis ywah the functions you should let to the lib users to customize, but you might want to set certificates,keys,CAs, best as filepath params to be able to set and roteate tls keys/certs/ca in an automated way via for example Kubernetes secrets later.

aryszka:
@szuecs @UltraNemesis I think that we don't need to add the cmd line flags this time. If some of them turns out to be important, we add them in a separate PR.

adding to skipper.Options as ProxyTLS, however, is a good idea and this would be consistent with how skipper.Options and skipper.Run are built.

aryszka:
:+1:

PR #921, opened 2019-01-18
Test/idle connection

szuecs:
test a bug I saw in one cluster

### DIFF
--- a/proxy/upgrade.go
+++ b/proxy/upgrade.go
@@ -61,6 +61,23 @@ type auditLog struct {
 // write data to an auditlog. It will not return until the connection
 // is closed.
 func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
+	// The following 2 checks are based on
+	// https://tools.ietf.org/html/rfc2616#section-14.42
+	// https://tools.ietf.org/html/rfc7230#section-6.7
+	// and https://tools.ietf.org/html/rfc6455 (websocket)
+	if req.Method != http.MethodGet {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		w.Write([]byte(http.StatusText(http.StatusMethodNotAllowed)))
+		return
+	}
+	if (req.ProtoMajor <= 1 && req.ProtoMinor < 1) ||
+		req.Header.Get("Connection") != "Upgrade" ||
+		req.Header.Get("Upgrade") == "" {
+		w.WriteHeader(http.StatusBadRequest)
+		w.Write([]byte(http.StatusText(http.StatusBadRequest)))
+		return
+	}
+
 	backendConn, err := p.dialBackend(req)
 	if err != nil {
 		log.Errorf("Error connecting to backend: %s", err)
@@ -105,6 +122,13 @@ func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
 		log.Errorf("Got unauthorized error from backend for: %s %s", req.Method, req.URL)
 		w.WriteHeader(http.StatusUnauthorized)
 		w.Write([]byte(http.StatusText(http.StatusUnauthorized)))
+		return
+	}
+
+	if resp.StatusCode != http.StatusSwitchingProtocols {
+		log.Errorf("Got invalid status code from backend: %d", resp.StatusCode)
+		w.WriteHeader(http.StatusServiceUnavailable)
+		w.Write([]byte("Backend do not allow to switch protocols"))
 		return
 	}
 
--- a/proxy/upgrade.go
+++ b/proxy/upgrade.go
@@ -126,9 +126,9 @@ func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
 	}
 
 	if resp.StatusCode != http.StatusSwitchingProtocols {
-		log.Errorf("Got invalid status code from backend: %d", resp.StatusCode)
+		log.Debugf("Got invalid status code from backend: %d", resp.StatusCode)
 		w.WriteHeader(http.StatusServiceUnavailable)
-		w.Write([]byte("Backend do not allow to switch protocols"))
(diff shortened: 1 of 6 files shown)


proxy/upgrade.go
```diff
@@ -108,6 +125,13 @@ func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	if resp.StatusCode != http.StatusSwitchingProtocols {
+		log.Errorf("Got invalid status code from backend: %d", resp.StatusCode)
```
aryszka: we don't control the backend behavior, we should log this only as Debug(), because it is not our error

skptesting/slow-calls-multi-proxy-lbroutes.sh
```diff
@@ -0,0 +1,114 @@
+#! /bin/bash
+
+# build go test binaries
+go build -o gohttpclient gohttpclient.go
+#go build -o gorilla-ws-example gorilla-ws-example.go
```
aryszka: please drop commented code

skptesting/slow-calls-multi-proxy-lbroutes.sh
```diff
@@ -0,0 +1,114 @@
...
+pids="$pids $!"
+log [servers started, wait 1 sec]
+
+sleep 1
+
+#log; log [start connections wrk]
+#wrk -H "Host: test.example.org"  -c "5" -d "12s" http://127.0.0.1:9090/  &
+
+SEC=8
+log; log [start connections vegeta]
```
aryszka: as i see we don't use vegeta here, but the custom client, change vegeta -> client

skptesting/slow-calls-multi-proxy-lbroutes.sh
```diff
@@ -0,0 +1,114 @@
...
+# sleep 1
+# printf "GET / HTTP/1.1\r\nHost: test.example.org\r\n\r\n" | nc 127.0.0.1 9090 &
+# echo ""
+# sleep 1
+
+#wrk -H "Host: test.example.org"  -c "5" -d "3" http://127.0.0.1:9090/
+#slow :9090 "Host: test.example.org"
+
+./gohttpclient &
+clipid="$!"
```
aryszka: why not add the pid to the existing `pids` var, then `cleanup` would kill it together with the others (https://github.com/zalando/skipper/blob/master/skptesting/benchmark.inc#L48)

skptesting/gohttpclient.go
```diff
@@ -0,0 +1,77 @@
...
+
+	ureq, _ := http.NewRequest("GET", target, nil)
+	ureq.Host = "wrong.example.org"
+	ureq.Header.Set("Upgrade", "websocket")
+	ureq.Header.Set("Connection", "upgrade")
+	uresp, err := tr.RoundTrip(ureq)
+	if err != nil {
+		logrus.Errorf("Failed to do ugrade roundtrip: %v", err)
+	}
+	defer uresp.Body.Close()
```
aryszka: in case err != nil, uresp may be nil

skptesting/benchmark.inc
```diff
@@ -82,6 +82,18 @@ ngx() {
 	pids="$pids $!"
 }
 
+slow() {
```
aryszka: this function is not used, we can remove it

proxy/upgrade.go
```diff
@@ -108,6 +125,13 @@ func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
...
+		w.WriteHeader(http.StatusServiceUnavailable)
+		w.Write([]byte("Backend do not allow to switch protocols"))
```
aryszka: "Backend _does_ not allow..."

proxy/upgrade.go
```diff
@@ -61,6 +61,23 @@ type auditLog struct {
 // write data to an auditlog. It will not return until the connection
 // is closed.
 func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
+	// The following 2 checks are based on
+	// https://tools.ietf.org/html/rfc2616#section-14.42
+	// https://tools.ietf.org/html/rfc7230#section-6.7
+	// and https://tools.ietf.org/html/rfc6455 (websocket)
+	if req.Method != http.MethodGet {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		w.Write([]byte(http.StatusText(http.StatusMethodNotAllowed)))
```
aryszka: i think go server automatically write the status text. Is there a reason to do it here?

skptesting/slow-calls-multi-proxy-lbroutes.sh
szuecs: done

proxy/upgrade.go
szuecs: but then we should also not error on Auth check or maybe just delete the `http.StatusUnauthorized` check just above this if check?

aryszka: i think we should remove logging an error on unauthorized. It's not an error and we return the right status code.

proxy/upgrade.go
szuecs: I just follow the same pattern as it was done before in this function.

proxy/upgrade.go
szuecs: done

skptesting/benchmark.inc
szuecs: done

skptesting/gohttpclient.go
szuecs: fixed

proxy/upgrade.go
aryszka: true, i didn't check the surroundings

aryszka:
:+1:

aryszka:
:+1:

szuecs:
:+1:

PR #925, opened 2019-01-20
Access log disable filter

### LINKED ISSUE #909: Access logging based on response code
We love the disableAccessLog() filter. It would be nicer if we could choose to disable/enable logs based on response code. That way we can do away with logging all the `2xx`s, and only log the errors in access logs.

May look something like `Path("/api/frequently-requested-resource") -> disableAccessLog(200, 201) -> "https://related-backend.com`.

I've already gone through the code a bit and can work on it.

avb1989:
Adds ability to enable/disable logs based on response code returned by the backend. This would allow more fine grained control for high throughput systems which don't want to log 2xx responses (~99.99% of responses) but still would like to log errors.

Fixes #909

### DIFF
--- a/filters/accesslog/control.go
+++ b/filters/accesslog/control.go
@@ -40,6 +40,8 @@ func extractFilterValues(args []interface{}, enable bool) (filters.Filter, error
 		}
 		if ok {
 			prefixes = append(prefixes, intPref)
+		} else {
+			return nil, filters.ErrInvalidFilterParameters
 		}
 	}
 	return &AccessLogFilter{Enable: enable, Prefixes: prefixes}, nil
--- a/filters/accesslog/control.go
+++ b/filters/accesslog/control.go
@@ -13,11 +13,46 @@ const (
 	AccessLogEnabledKey = "statebag:access_log:proxy:enabled"
 )
 
+// Common filter struct for holding access log state
+type AccessLogFilter struct {
+	Enable   bool
+	Prefixes []int
+}
+
+func (al *AccessLogFilter) Request(ctx filters.FilterContext) {
+	bag := ctx.StateBag()
+	bag[AccessLogEnabledKey] = al
+}
+
+func (*AccessLogFilter) Response(filters.FilterContext) {}
+
+func extractFilterValues(args []interface{}, enable bool) (filters.Filter, error) {
+	prefixes := make([]int, 0)
+	for _, prefix := range args {
+		intPref, ok := 0, false
+		switch prefix.(type) {
+		case float32:
+			intPref, ok = int(prefix.(float32)), true
+		case float64:
+			intPref, ok = int(prefix.(float64)), true
+		default:
+			intPref, ok = prefix.(int)
+		}
+		if ok {
+			prefixes = append(prefixes, intPref)
+		}
+	}
+	return &AccessLogFilter{Enable: enable, Prefixes: prefixes}, nil
+}
+
 type disableAccessLog struct{}
 
-// NewDisableAccessLog creates a filter spec to disable access log for specific route
+// NewDisableAccessLog creates a filter spec to disable access log for specific route.
+// Optionally takes in response code prefixes as arguments. When provided, access log is disabled
+// only if response code matches one of the arguments.
 //
-// 	disableAccessLog()
+//  	disableAccessLog() or
+//  	disableAccessLog(1, 20, 301)  to disable logs for 1xx, 20x and 301 codes
 func NewDisableAccessLog() filters.Spec {
 	return &disableAccessLog{}
 }
@@ -25,21 +60,17 @@ func NewDisableAccessLog() filters.Spec {
 func (*disableAccessLog) Name() string { return DisableAccessLogName }
 
 func (al *disableAccessLog) CreateFilter(args []interface{}) (filters.Filter, error) {
-	return al, nil
+	return extractFilterValues(args, false)
 }
-
-func (al *disableAccessLog) Request(ctx filters.FilterContext) {
-	bag := ctx.StateBag()
-	bag[AccessLogEnabledKey] = false
-}
-
-func (*disableAccessLog) Response(filters.FilterContext) {}
 
 type enableAccessLog struct{}
 
 // NewEnableAccessLog creates a filter spec to enable access log for specific route
+// Optionally takes in response code prefixes as arguments. When provided, access log is enabled
+// only if response code matches one of the arguments.
 //
 // 	enableAccessLog()
+// 	enableAccessLog(1, 20, 301)  to enable logs for 1xx, 20x and 301 codes
 func NewEnableAccessLog() filters.Spec {
 	return &enableAccessLog{}
 }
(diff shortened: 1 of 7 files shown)


filters/accesslog/control.go
```diff
@@ -13,46 +13,70 @@ const (
...
+	bag[AccessLogEnabledKey] = al
+}
+
+func (*AccessLogFilter) Response(filters.FilterContext) {}
+
+func extractFilterValues(args []interface{}, enable bool) (filters.Filter, error) {
+	prefixes := make([]int, 0)
+	for _, prefix := range args {
+		intPref, ok := 0, false
+		switch prefix.(type) {
```
szuecs: I don't see why the switch case block is required.
Can't you always use `int()` ?
Maybe also `strconv.Atoi()` could be used as replacement.

proxy/proxy.go
```diff
@@ -1070,6 +1070,25 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 	p.sendError(ctx, id, code)
 }
 
+func shouldLog(statusCode int, filter al.AccessLogFilter) bool {
+	if len(filter.Prefixes) == 0 {
+		return filter.Enable
```
szuecs: Zero value of bool is false, so the default would be not to log

proxy/proxy.go
```diff
@@ -1092,17 +1111,18 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		accessLogEnabled, ok := ctx.stateBag[al.AccessLogEnabledKey].(*al.AccessLogFilter)
 
 		if !ok {
-			accessLogEnabled = !p.accessLogDisabled
+			accessLogEnabled = &al.AccessLogFilter{Enable: !p.accessLogDisabled, Prefixes: nil}
 		}
+		statusCode := lw.GetCode()
 
-		if accessLogEnabled {
+		if shouldLog(statusCode, *accessLogEnabled) {
```
szuecs: Is there a reason to not pass the `accessLogEnabled` as a pointer?

filters/accesslog/control.go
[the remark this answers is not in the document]
szuecs: We should not rely on the eskip parser returning only float64, IMO better to test while creating the filter for all these types.

szuecs:
I would like to have also some go test benchmarks, because for me this looks like we could introduce easily a memory leak.
 @avb1989 could you create some benchmark tests https://golang.org/pkg/testing/#hdr-Benchmarks

filters/accesslog/control.go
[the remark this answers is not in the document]
lmineiro: This is not how we created many of the predicates and filters until now. If this is not documented as such, we assumed `float64` as the specification. That was the base for not advancing with https://github.com/zalando/skipper/pull/430 some time ago.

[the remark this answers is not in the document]
aryszka: i think it's fine to allow and check for integers, because it helps with programmatically defined eskip objects, e.g. during testing. On the other hand, I would return an error when an argument is not a number.

proxy/proxy.go
aryszka: to me this seems fine, because below, before calling shouldLog, the argument is initialized with !p.accessLogDisabled

aryszka: i think this is fine, because shouldLog is called with an initialized Enable value, by default taking !p.accessLogDisabled

avb1989: Yep. The value of `Enable` is set in `createFilter`.

proxy/proxy.go
avb1989: No specific reason. My go-fu is weak :sweat_smile:

filters/accesslog/control.go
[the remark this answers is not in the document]
avb1989: Ah. Didn't know this. I just assumed it was returning float64 because I'm running a 64 bit system. I can leave it in if it helps with testcases.

proxy/proxy.go
szuecs: Ok I am fine with it

avb1989:
@szuecs Added benchmark tests for routes without filters and with filters (+ additional options).   
On my machine, results seem to be comparable (but slightly worse) compared to master. 
Test : 
`go test -cpu 4 -benchmem -bench ^BenchmarkAccess github.com/zalando/skipper/proxy`

On master 
```
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/proxy
BenchmarkAccessLogNoFilter-4       	  100000	     13678 ns/op	   13960 B/op	      84 allocs/op
BenchmarkAccessLogDisable-4	   	  100000	     16633 ns/op	   12097 B/op	      70 allocs/op
BenchmarkAccessLogEnable-4    		  100000	     16077 ns/op	   14547 B/op	      95 allocs/op
PASS
ok  	github.com/zalando/skipper/proxy	8.712s
```

On feature branch 
```
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/proxy
BenchmarkAccessLogNoFilter-4       	  100000	     15514 ns/op	   13962 B/op	      84 allocs/op
BenchmarkAccessLogDisablePrint-4   	  100000	     16204 ns/op	   14549 B/op	      95 allocs/op
BenchmarkAccessLogDisable-4        	  100000	     17723 ns/op	   12096 B/op	      70 allocs/op
BenchmarkAccessLogEnablePrint-4    	  100000	     16606 ns/op	   14298 B/op	      95 allocs/op
BenchmarkAccessLogEnable-4         	  200000	     16434 ns/op	   12074 B/op	      70 allocs/op
PASS
ok  	github.com/zalando/skipper/proxy	14.422s
```

avb1989:
@szuecs Caching struct values in case of filter not being used improved timings for the "NoFilter" case. 

Master : 
```
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/proxy
BenchmarkAccessLogNoFilter-4   	  100000	     12614 ns/op	   13963 B/op	      84 allocs/op
BenchmarkAccessLogDisable-4    	  100000	     12591 ns/op	   12078 B/op	      70 allocs/op
BenchmarkAccessLogEnable-4     	  100000	     14794 ns/op	   14548 B/op	      95 allocs/op
PASS
ok  	github.com/zalando/skipper/proxy	7.980s
```

Feature branch : 
```
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/proxy
BenchmarkAccessLogNoFilter-4       	  100000	     13204 ns/op	   13960 B/op	      84 allocs/op
BenchmarkAccessLogDisablePrint-4   	  100000	     12746 ns/op	   14548 B/op	      95 allocs/op
BenchmarkAccessLogDisable-4        	  100000	     11827 ns/op	   12082 B/op	      70 allocs/op
BenchmarkAccessLogEnablePrint-4    	  100000	     14114 ns/op	   14298 B/op	      95 allocs/op
BenchmarkAccessLogEnable-4         	  200000	     12843 ns/op	   12071 B/op	      70 allocs/op
PASS
ok  	github.com/zalando/skipper/proxy	12.058s
```

szuecs:
@avb1989 thanks great to see!

szuecs:
@avb1989 did you resolved all comments from @aryszka? 
If so we can merge this in my opinion.

avb1989:
I did make the changes that @aryszka suggested. Adding string value as parameters results in invalid filter error now. Waiting for him to give a thumbs up.

szuecs:
:+1:

aryszka:
:+1:

aryszka:
@avb1989 thank you!

aryszka:
:+1:

szuecs:
:+1:

PR #932, opened 2019-01-25
fix static check for type switch

aryszka:
from the failing check:

"assigning the result of this type assertion to a variable (switch prefix := prefix.(type)) could eliminate the following type assertions:"


aryszka:
:+1:

Issue #819, opened 2018-10-11
Ingress Path Matching - Default Backends and Catch-All Regex's Trump All Other Path Matches


kailam77:
I'm trying to apply a 'blacklist' of paths that should 404.  My intent is to define path rules that send matching requests to a 404 service, and have all other requests route to the app.  I've tried a few approaches to accomplish this, documented in detail below.  What I'm finding is that if there is a default backend defined, all the path rules are ignored and everything gets routed to the default backend.  And, if a 'catch-all' path rule is define, e.g. '/.*', again the other path rules are ignored, and all requests route to the catch-all.

The only way to make the blacklist path rules work is to have specific 'whitelist' paths defined that route to the app (see cases below for details).  This is undesirable since we have a lot of different URL's and don't want to have to whitelist them all.

The behavior here is the opposite of what I'd expect.  I would think that if a path matches a path rule, that should take precedence over the default or the catch-all generic regex.  Instead, it seems like the broadest match is the one that wins.  This also seems contrary to the [documentation](http://opensource.zalando.com/skipper/reference/architecture/#route-processing).

We can get the blacklist logic to work using custom routes, specifying \<shunt\> for the blacklisted routes, but we'd rather avoid skipper-specific annotations if possible.  Is there a reason for the current path matching behavior, or is this a bug?

## Case 1: (Control case) Default backend, no blacklist
* *Result:* Works as expected
* *Expected:*
  * /heartbeat/ returns "OKOKOK"
  * /health/ returns "UP"
* *Actual:* Matches Expected

#### ingress.yml:
    ...
    spec:
      backend:
        serviceName: landers2
        servicePort: http

#### Output:
    $ curl https://site.com/heartbeat/
    {"status":"OKOKOK"}
    
    $ curl https://site.com/health/
    {"status":"UP"}

## Case 2: Path blacklist and default backend
* *Result:* Doesn't work - path match rule is ignored.
* *Expected:*
  * /heartbeat/ returns 404
  * /health/ returns "UP"
* *Actual:* /heartbeat/ returns "OKOKOK"

#### ingress.yml:
    ...
    spec:
      backend:
        serviceName: landers2
        servicePort: http
      rules:
      - host: site.com
        http:
          paths:
          - path: /heartbeat/
            backend:
              serviceName: default-404-service
              servicePort: web-port

#### Output:
    $ curl https://site.com/heartbeat/
    {"status":"OKOKOK"}
    
    $ curl https://site.com/health/
[...]

Issue #689, opened 2018-06-23
Weight() predicate - former: https redirect with kubernetes dataclient is radomly bypassed


ese:
It seems related with #662 

Requests to http endpoints are supposed to be responded with a 301 to https but are radomly bypassed directly to the service. This seems only happened in services with several replicas and the loadbalancer feature is creating the routes.

I guess it can be related with how routes are gen and stored in memory and how the predicates are weight.

As much skipper replicas and ingresses you have more probabilities this happens.

example of routes with this problem:
```
kube_default__nginx_1__test_nginx______nginx_slash_1: Host(/^test-nginx$/) && PathRegexp(/^\//) && LBMember("kube_default__nginx_1__test_nginx______nginx_slash", 1)
  -> dropRequestHeader("X-Load-Balancer-Member")
  -> "http://100.96.3.5:8080";


kube_default__nginx_1__test_nginx______nginx_slash_2: Host(/^test-nginx$/) && PathRegexp(/^\//) && LBMember("kube_default__nginx_1__test_nginx______nginx_slash", 2)
  -> dropRequestHeader("X-Load-Balancer-Member")
  -> "http://100.96.3.8:8080";

kube_default__nginx_1__test_nginx______nginx_slash_0: Host(/^test-nginx$/) && PathRegexp(/^\//) && LBMember("kube_default__nginx_1__test_nginx______nginx_slash", 0)
  -> dropRequestHeader("X-Load-Balancer-Member")
  -> "http://100.96.1.27:8080";

kube_default__nginx_1__test_nginx______nginx_slash__lb_group: Host(/^test-nginx$/) && PathRegexp(/^\//) && LBGroup("kube_default__nginx_1__test_nginx______nginx_slash")
  -> lbDecide("kube_default__nginx_1__test_nginx______nginx_slash", 3)
  -> <loopback>;

kube__redirect: PathRegexp(/.*/) && Header("X-Forwarded-Proto", "http") && HeaderRegexp("X-Forwarded-Port", /.*/)
  -> redirectTo(301, "https:")
  -> <shunt>;
```

szuecs:
It looks like we created a race condition in the predicate between the lb group lookup and the redirect.
@aryszka can you check that?
For me it looks like 3 predicates vs. 3 predicates.

aryszka:
checking

aryszka:
[...]

Issue #674, opened 2018-06-11
retry() request filter


Traumeel:
I have a custom _Filter_ which injects additional headers into requests. Occasionally requests fail with 5xx. I would like to have an ability to retry failed requests and return the final successful response to a client.

I want to be able to retry the backend call on specific http status codes. For example I want to retry on status code 502. Example filter setting: `retry(3, 502)` would retry 3 times if HTTP status code is 502.

aryszka:
@Traumeel thanks for the input, good idea!

As for the implementation, I believe the filter should 'mark' the request flow to be retried, and the execution itself should rely on the existing loadbalancer fallback mechanism, except that in this case, it could retry the request to the same backend, and not only on dial errors. The filter could also instrument the retry mechanism for the "usual" reliability patterns. @szuecs, @lmineiro, what do you think about this approach?

aryszka:
this approach would also mean that the filter only instruments the retry, regarless where it is in the filter chain. E.g. the tee() filter calls won't be retried.

aryszka:
what i meant was to support all these cases:

1. if not load balanced, but marked as retriable: retry to the same backend
2. if load balanced, and marked as retriable: retry to the next load balanced backend
3. if load balanced, but not marked as retriable: retry to the next load balanced backend, but only when it is a dial error

and I believe this should be implemented by extending the current load balancing mechanism.

szuecs:
Follow up is #1473

Issue #557, opened 2018-02-23
loadbalancer algorithms


szuecs:
- #536 implements a loadbalancer feature with only roundrobin algorithm. 
- After #556 we should have a solid code base to easily implement different algorithms per route. 

We want to have some other algorithms being possible and able to choose per route (for example from ingress). Algorithm ideas we could support:

- [ ] random
- [ ] LRU (adds memory usage)
- [ ] closest (by ip on the local node)
- [ ] fastest response (maybe complicated to do, because metrics lookup required)
- [ ] consistent hashing (supports sticky sessions across one TCP/IP connection)
- [ ] [Maglev](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/44824.pdf) (if possible)

ruiaraujo:
Weighted round robin would also be a good alternative, where the weight is dependent of the size of the nodes.

szuecs:
@ruiaraujo can you elaborate on how to pass the data you need from nodes to the load balancer?

ruiaraujo:
Could be something like a weight on each target.

Use case
I deploy my application using Spot of multiple sizes. My apps reacts and takes advantage of the added CPUs. 

I have Skipper as LB where the route has the meta information on each node relative size and the roundRobin would take advantage of this information to generate the right amount of load for each target.

I have no idea how this would look and how it should be specified.

szuecs:
I think a simpler algorithm that could have a similar effect is the best (least) of N choices, but we would need to have data per backend in lbcontext for the lb algorithm.

szuecs:
@ruiaraujo I added your suggestion to the list

ardaguclu:
I would like to work on "least requests". Because round robin is not uniformly distributed between nodes especially if long connections exist. Do you have any concern or objection before that?

szuecs:
As discussed in slack there is https://github.com/zalando/skipper/pull/1507 that you can test or also check the code to make a new PR, if you like.

AshrafAhmed9:
Hi @szuecs, following up on your pointer from #4096. Looking at the current state of loadbalancer/algorithm.go, several items from the list are already done: random, consistent hashing (incl. bounded loads via consistentHashBalanceFactor), least-of-N (powerOfRandomNChoices), and slow ramp-up via the fadein filter (#1207).

From what's left, weighted round robin looks like the most useful next step — @ruiaraujo's use case above (heterogeneous Spot node sizes) is still a real one.

[...]

PR #938, opened 2019-01-30
refactor loadbalancer to be a single route

szuecs:
This PR will fix https://github.com/zalando/skipper/issues/556 and make https://github.com/zalando/skipper/issues/557 and https://github.com/zalando/skipper/issues/674 simple to implement.

This PR will probably fix some edge cases of https://github.com/zalando/skipper/issues/689 and https://github.com/zalando/skipper/issues/819.

feature: add roundrobin() and endpoints() filters
refactor: apply retry in the proxy using the new filters
refactor: get rid of roundrobin routing hacks
refactor: switch kubernetes dataclient to use new loadbalancer implementation
refactor: drop old code that is not necessary anymore
refactor: fix tests


dataclients/kubernetes/kube.go
```diff
@@ -969,12 +896,12 @@ func (c *Client) ingressToRoutes(items []*ingressItem) ([]*eskip.Route, error) {
 			// update Traffic field for each backend
 			computeBackendWeights(backendWeights, rule)
 
-			for idx, prule := range rule.Http.Paths {
+			for _, prule := range rule.Http.Paths {
 				// add extra routes from optional annotation
 				for extraIndex, r := range extraRoutes {
 					route := *r
 					route.HostRegexps = host
-					route.Id = routeIDForCustom(i.Metadata.Namespace, i.Metadata.Name, route.Id, rule.Host+strings.Replace(prule.Path, "/", "_", -1), idx, extraIndex)
```
aryszka: i think we cannot drop the index of the path rule from the generated id of the custom route, because we make a copy of the custom routes for every path rule.

aryszka: i get it: the path itself will make it unique

loadbalancer/lbalgorithm.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/zalando/skipper/filters"
+)
+
+type lbAlgorithmType int
+
+const (
+	RoundRobin         = "roundrobin"
+	unknownLbAlgorithm = "unknownLbAlgorithm"
+
+	roundrobin = iota
```
aryszka: this constant could be marked with `lbAlgorithmType` as

```
roundrobin lbAlgorithmType = iota
```

proxy/proxy.go
```diff
@@ -975,32 +976,27 @@ func (p *Proxy) do(ctx *context) error {
...
-				}
-
 				if ctx.proxySpan != nil {
 					ctx.proxySpan.Finish()
 					ctx.proxySpan = nil
 				}
 
 				tracing.LogKV("retry", ctx.route.Id, ctx.Request().Context())
 
+				// lbalgorithm has to be the last route
```
aryszka: this limitation is a bit of a leaky abstraction. There could be an interface:

```
type LBAlgorithmFilter interface { ApplyLB(filters.FilterContext) }
```

and we could find the last filter from the end of the filter chain that implements this interface, and explicitly call it.

routing/fallbackgroup.go
```diff
@@ -5,51 +5,16 @@ package routing
...
-
-		current := head
-		for _, route := range group[1:] {
-			current.Next = route
-			current = route
-			current.Head = head
-			current.Me = current
-			current.Group = name
-			current.IsLoadBalanced = true
+		f := r[i].Filters[l-1]
```
aryszka: see my comment in the proxy.go, the same applies here, too

loadbalancer/lbalgorithm.go
```diff
@@ -0,0 +1,82 @@
...
+	"sync"
+
+	"github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/filters"
+)
+
+type lbAlgorithmType int
+
+const (
+	RoundRobin         = "roundrobin"
```
aryszka: style: for me it would be more consistent if it was called "roundRobin"

proxy/proxy.go
szuecs: If we can do this before the proxy.go, I am fine.

routing/fallbackgroup.go
szuecs: yes, but only if it's at route parse time and not in the hot path

loadbalancer/lbalgorithm.go
szuecs: changed

loadbalancer/lbalgorithm.go
szuecs: done

dataclients/kubernetes/kube.go
szuecs: I think we should move this change to another PR

aryszka:
missing:

- [ ] documentation
- [ ] testing

loadbalancer/algorithm.go
```diff
@@ -0,0 +1,118 @@
...
+	return endpoints[r.index]
+}
+
+func NewAlgorithmProvider() routing.PostProcessor {
+	return &algorithmProvider{}
+}
+
+func algorithmTypeFromString(a string) (algorithmType, error) {
+	switch a {
+	case "":
```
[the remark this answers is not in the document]
aryszka: The idea is that the algorithm was not specified. The user doesn't care and we use a default one. In eskip, this looks like this:

```
* -> <"http://10.0.0.1", "http://10.0.0.2">
```

loadbalancer/algorithm.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return err
+	}
+
+	a := defaultAlgorithm
+	if t != none {
+		a = algorithms[t]
+	}
+
+	r.LBAlgorithm = a(len(r.LBEndpoints))
```
szuecs: I think it would be better to work on endpoints and not on a number, because not every algorithm does only work on a number.

loadbalancer/algorithm.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+type roundRobin struct {
+	mx    sync.Mutex
+	index int
+}
+
+type algorithmProvider struct{}
+
+var algorithms = map[algorithmType]func(int) routing.LBAlgorithm{
```
szuecs: I think I don't like the idea of this signature, because it binds it to simple algorithms only (roundrobin and random are fine but everything that needs to lookup metrics data for an endpoint would be not).

loadbalancer/algorithm.go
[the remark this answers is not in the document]
szuecs: Ok, but I think it needs a comment telling "user does not specify, choose default"

szuecs:
@aryszka thanks for the changes, looks much nicer with the new backend IMO.

szuecs:
@aryszka PR is ready to review, I only changed the `docs/` and rebased to fix signoff which was missing in a commit.

szuecs:
:+1:

aryszka:
@szuecs thanks for the docs!

aryszka:
:+1:

PR #947, opened 2019-02-08
fix non rfc compliant websocket requests

szuecs:
fix non compliant websocket requests that are made by many libraries out there

### DIFF
--- a/proxy/upgrade.go
+++ b/proxy/upgrade.go
@@ -71,7 +71,8 @@ func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 	if (req.ProtoMajor <= 1 && req.ProtoMinor < 1) ||
-		req.Header.Get("Connection") != "Upgrade" ||
+		// RFC states "Upgrade", but reality is different in many libraries out in the world
+		strings.ToLower(req.Header.Get("Connection")) != "upgrade" ||
 		req.Header.Get("Upgrade") == "" {
 		w.WriteHeader(http.StatusBadRequest)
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

aryszka:
:+1:

arjunrn:
:+1:

proxy/upgrade.go
```diff
@@ -71,7 +71,8 @@ func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 	if (req.ProtoMajor <= 1 && req.ProtoMinor < 1) ||
-		req.Header.Get("Connection") != "Upgrade" ||
+		// RFC states "Upgrade", but reality is different in many libraries out in the world
```
Baishan: There may be a misunderstanding here. Please see review [rfc6455](https://tools.ietf.org/html/rfc6455) with particular attention to criteria 2 and 3 on page 18.

```
2.  If the response lacks an |Upgrade| header field or the |Upgrade|
       header field contains a value that is not an ASCII case-
       insensitive match for the value "websocket", the client MUST
       _Fail the WebSocket Connection_.

   3.  If the response lacks a |Connection| header field or the
       |Connection| header field doesn't contain a token that is an
       ASCII case-insensitive match for the value "Upgrade", the client
       MUST _Fail the WebSocket Connection_.
```
It would seem that the HTTP headers should be case insensitive which may explain why many clients are different.

aryszka: Indeed. I think we can drop this comment. Thanks for the review.

PR #951, opened 2019-02-12
kubernetes: fetch endpoints/services once

aermakov-zalando:
Instead of fetching individual services/endpoints for every ingress, fetch them all at once when refreshing. In clusters with lots of ingresses and/or skippers this will bring down the apiserver RPS by Skipper from 1500+ to  15.

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -38,8 +38,10 @@ const (
 	defaultIngressClass           = "skipper"
 	ingressRouteIDPrefix          = "kube"
 	defaultEastWestDomainFmt      = "%s.%s.skipper.cluster.local"
-	endpointURIFmt                = "/api/v1/namespaces/%s/endpoints/%s"
-	serviceURIFmt                 = "/api/v1/namespaces/%s/services/%s"
+	endpointsClusterURI           = "/api/v1/endpoints"
+	endpointsNamespaceFmt         = "/api/v1/namespaces/%s/endpoints"
+	servicesClusterURI            = "/api/v1/services"
+	servicesNamespaceFmt          = "/api/v1/namespaces/%s/services"
 	serviceAccountDir             = "/var/run/secrets/kubernetes.io/serviceaccount/"
 	serviceAccountTokenKey        = "token"
 	serviceAccountRootCAKey       = "ca.crt"
@@ -188,9 +190,11 @@ type Client struct {
 	reverseSourcePredicate   bool
 	pathMode                 PathMode
 	quit                     chan struct{}
-	namespace                string
 	kubernetesEnableEastWest bool
 	eastWestDomainFmt        string
+	ingressesURI             string
+	servicesURI              string
+	endpointsURI             string
 }
 
 var nonWord = regexp.MustCompile(`\W`)
@@ -267,7 +271,7 @@ func New(o Options) (*Client, error) {
 		eastWestDomainFmt = "%s.%s." + o.KubernetesEastWestDomain
 	}
 
-	return &Client{
+	result := &Client{
 		httpClient:               httpClient,
 		apiURL:                   apiURL,
 		provideHealthcheck:       o.ProvideHealthcheck,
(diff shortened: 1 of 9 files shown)


aermakov-zalando:
Performance improvements:

image

image

aermakov-zalando:
An even better idea would be adding WATCH for ingresses/services/endpoints and only doing refreshing if there are changes, but this will require embedding a proper client.

dataclients/kubernetes/kube.go
```diff
@@ -639,65 +617,57 @@ func (c *Client) convertPathRule(
...
+			}}, r.Predicates...)
+			log.Infof("Traffic weight %.2f for backend '%s'", prule.Backend.Traffic, svcName)
 		}
-		log.Debugf("%d new routes for %s/%s/%s", len(eps), ns, svcName, svcPort)
-	} else {
-		eps = val
-		log.Debugf("%d routes for %s/%s/%s already known", len(eps), ns, svcName, svcPort)
+		return r, nil
+
+	} else if err != nil {
```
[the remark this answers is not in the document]
aermakov-zalando: I think some of the above code might still assign to `err`, so I'd keep this.

arjunrn:
lgtm

go.mod
```diff
@@ -4,6 +4,8 @@ require (
 	cloud.google.com/go v0.28.0 // indirect
 	github.com/DataDog/datadog-go v0.0.0-20180822151419-281ae9f2d895 // indirect
 	github.com/abbot/go-http-auth v0.0.0-20150922224136-efc9484eee77
+	github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc // indirect
```
aryszka: is it worth to import all these dependencies for those few assertions in the tests? They slow down the builds.

aryszka:
:+1:

aryszka:
WATCH is indeed a good idea. Maybe even it is possible to implement it without an additional client library. Is the Kube API more complex regarding the watch functionality than the one of etcd.v2?

go.mod
aermakov-zalando: Fixed, not sure where they came from.

aermakov-zalando:
No, it's pretty complicated so I'd rather embed `client-go`.

aryszka:
:+1:

dataclients/kubernetes/kube.go
```diff
@@ -38,8 +38,10 @@ const (
 	defaultIngressClass           = "skipper"
 	ingressRouteIDPrefix          = "kube"
 	defaultEastWestDomainFmt      = "%s.%s.skipper.cluster.local"
-	endpointURIFmt                = "/api/v1/namespaces/%s/endpoints/%s"
-	serviceURIFmt                 = "/api/v1/namespaces/%s/services/%s"
+	endpointsClusterURI           = "/api/v1/endpoints"
+	endpointsNamespaceFmt         = "/api/v1/namespaces/%s/endpoints"
+	servicesClusterURI            = "/api/v1/services"
+	servicesNamespaceFmt          = "/api/v1/namespaces/%s/services"
```
szuecs: Do we need the namespaces versions?

dataclients/kubernetes/kube.go
```diff
@@ -188,9 +190,11 @@ type Client struct {
 	reverseSourcePredicate   bool
 	pathMode                 PathMode
 	quit                     chan struct{}
-	namespace                string
```
szuecs: If we drop namespace, do we break any filter by namespace or so?
I don’t remember if we have such a feature and can not check it right now, but maybe something to check.

szuecs:
In general looks good to me and the improvements are great!
I am against using client-go and watch, because of the current broken state, where you can hang forever until tcp timeout. We could work on a client that works with long poll and test that we have timeouts that work.

dataclients/kubernetes/kube.go
aermakov-zalando: There is support for requesting only ingresses in a certain namespace, and it makes sense to do the same for endpoints/services.

dataclients/kubernetes/kube.go
aermakov-zalando: I don't think so, because then compilation would fail. It's a private field as well so I don't think it's used by any filters.

aryszka: as far as I can tell from the changes, the namespace handling is good.

szuecs:
:+1:

dataclients/kubernetes/kube.go
szuecs: thanks

dataclients/kubernetes/kube.go
szuecs: I just wondered, because of the delete of `namespace`.

aryszka:
:+1:

szuecs:
Thanks @aermakov-zalando !

PR #963, opened 2019-02-22
Fix rendering of filter docs and add a little more description

mikkeloscar:
Fix rendering of filter docs and add a little more description


mikkeloscar:
:+1:

szuecs:
:+1:

Issue #949, opened 2019-02-11
API usage monitoring filter improvements


maxim-tschumak:
## Acceptance criteria

* Use `{no-match}` placeholder for non-matching paths and clients
* Use curly brackets to identify path parameters and placeholders
* Configure `realms` globally (which realms have to be monitored)
* Configure `client` patterns on the filter config level
* Filter creation should fail if not all mandatory parameters are supplied

In order to not break the compatibility to previous global configurations, the `api-usage-monitoring-default-client-tracking-pattern` global parameter is still allowed but will be ignored. It will be completely removed in the near future.

PR #966, opened 2019-02-25
Create noop filter in case of invalid monitoring filter config

maxim-tschumak:
We don't want Skipper route creation to fail if a monitoring filter is provided with an invalid configuration. Let's make it less strict for now and return a `noop` filter if some information is missing.

Related to #949


aryszka:
:+1:

maxim-tschumak:
:+1:

PR #971, opened 2019-02-26
sync.Map to protect `metricPrefixedPerClient` against concurrent calls

### LINKED ISSUE #968: [apiusagemonitoring] Unprotected access to map causes panic
Here: https://github.com/zalando/skipper/pull/950/files#diff-a3889b1942785e41a55bf77d4653e68cR103

maxim-tschumak:
Fixes #968

I'm also happy to provide a fix based on lower-level locks. Depends on how fast we want to roll out the fix and your preferences.


aryszka:
@maxim-tschumak have you already checked if any other fields can be written concurrently or only this map?

maxim-tschumak:
@aryszka I'll double check

maxim-tschumak:
I'll prepare another PR with lower-level sync mechanisms.

PR #972, opened 2019-02-26
968 fix unprotected map access in monitoring

### LINKED ISSUE #968: [apiusagemonitoring] Unprotected access to map causes panic

maxim-tschumak:
Fixes #968

### DIFF
--- a/filters/apiusagemonitoring/filter_clientmetrics_test.go
+++ b/filters/apiusagemonitoring/filter_clientmetrics_test.go
@@ -3,6 +3,9 @@ package apiusagemonitoring
 import (
 	"encoding/base64"
 	"net/http"
+	"regexp"
+	"strconv"
+	"sync"
 	"testing"
 )
 
@@ -215,7 +218,7 @@ func Test_Filter_ClientMetrics_NoAuthHeader(t *testing.T) {
 		realmKeyName:          "realm",
 		clientKeyName:         "client",
 		clientTrackingPattern: clientTrackingPatternJustSomeUsers,
-		header:                http.Header{
+		header: http.Header{
 			/* no Authorization header */
 		},
 		expectedEndpointMetricPrefix: "apiUsageMonitoring.custom.my_app.my_api.GET.foo/orders.*.*.",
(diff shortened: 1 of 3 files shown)


filters/apiusagemonitoring/pathinfo.go
```diff
@@ -16,8 +17,9 @@ type pathInfo struct {
 	CommonPrefix   string
 	ClientPrefix   string
 
-	metricPrefixesPerMethod [methodIndexLength]*endpointMetricNames
-	metricPrefixedPerClient map[string]*clientMetricNames
+	metricPrefixesPerMethod [methodIndexLength]*endpointMetricNames // endpoint metric names cache
+	metricPrefixedPerClient map[string]*clientMetricNames           // client metric names cache
```
tkrop: Let us use `sync.Map` as replacement here?

The code to maintain will be shorter and easier to understand using a standard pattern, and there are smart guys that have thought about the performance implications of locks and formed it into this code.

aryszka: I would rather try to put the endpointMetricNames and the clientMetricNames into a single combined structure and protect it with a single RW lock scope, if the logic allows. As an alternative approach, I would even consider putting the scope of the Response() method of the filter into a single shared RW lock scope. The reason why the latter may be better, is that the endpointMetricNames is always accessed, and this way the code can be safer for future changes. For performance optimizations, we should create benchmarks first.

filters/apiusagemonitoring/filter_clientmetrics_test.go
```diff
@@ -215,7 +218,7 @@ func Test_Filter_ClientMetrics_NoAuthHeader(t *testing.T) {
 		realmKeyName:          "realm",
 		clientKeyName:         "client",
 		clientTrackingPattern: clientTrackingPatternJustSomeUsers,
-		header:                http.Header{
+		header: http.Header{
```
szuecs: please run `make fmt`

filters/apiusagemonitoring/pathinfo.go
szuecs: sync.Map is not simpler to maintain. 
+1 for @aryszka suggestion

maxim-tschumak:
replaced by #973

PR #977, opened 2019-03-01
k8s dataclient default filter config via config map

### LINKED ISSUE #976: [Kubernetes dataclient] Default filter configurations via ConfigMap
**As** Skipper Ingress operator
**I want** to have a possibility to define and use default filter configurations
**So that** I can easily preconfigure filters for specific services

With this feature in place, Skipper operators can define global filters centrally 
(e.g. authorization/logging/etc.). It is also practical for complex filter configurations
which are safer to generate automatically.

## Acceptance criteria
* Default filter configurations are optional
* Configurations are stored in a [ConfigMap](https://cloud.google.com/kubernetes-engine/docs/concepts/configmap)
* The key format: `{service}.{namespace}`
* The value is a string with one or more filter configurations (as you would specify them in Ingress annotation)
* The default filters are appended to those which are defined in the Ingress

maxim-tschumak:
Closes #976 

Related to [Out-of-the-box API Usage monitoring (internal link)](https://docs.google.com/document/d/1gPbsorpVwK3foG776-Gh-ks4p3bKtOruwZ_q7muFyVw/edit)

### DIFF
--- a/dataclients/kubernetes/kube_test.go
+++ b/dataclients/kubernetes/kube_test.go
@@ -41,6 +41,7 @@ type testAPI struct {
 	services  *serviceList
 	endpoints *endpointList
 	ingresses *ingressList
+	configmap *configMap
 	server    *httptest.Server
 	failNext  bool
 }
@@ -407,6 +408,7 @@ func newTestAPIWithEndpoints(t *testing.T, s *serviceList, i *ingressList, e *en
 		services:  s,
 		ingresses: i,
 		endpoints: e,
+		configmap: nil,
 	}
 
 	api.server = httptest.NewServer(api)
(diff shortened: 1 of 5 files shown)


dataclients/kubernetes/kube_test.go
```diff
@@ -3617,3 +3626,99 @@ func TestCreateEastWestRouteOverwriteDomain(t *testing.T) {
 		})
 	}
 }
+
+func TestSkipperDefaultFilters(t *testing.T) {
```
szuecs: I miss a test, that does not apply the default filters to optionally created redirects.
@aryszka what do you think should this be applied to redirects?

tkrop:
Substituted by #983.

Issue #820, opened 2018-10-19
Provide additional information for a route from Kubernetes


wndhydrnt:
I'm using Skipper do to a Canary deployment of an application on Kubernetes.
After the deployment of the Canary, I want to know how it is doing by checking the results of HTTP requests that the Canary has served.
Prometheus scrapes the metrics from Skipper and I want to use Grafana to create two graphs. One graph should show the HTTP requests of the Current deployment of the application and one for the Canary of the application.

When looking at the metrics exported by Skipper, I don't know how to accomplish this.
In my test setup, the metrics are exported like this:
```
skipper_serve_route_duration_seconds_bucket{code="200",method="GET",route="kube_default__whoami_0__whoami_dev_local_____whoami_5740a829",le="0.005"} 12
skipper_serve_route_duration_seconds_bucket{code="200",method="GET",route="kube_default__whoami_0__whoami_dev_local_____whoami_5740a829",le="0.01"} 12
skipper_serve_route_duration_seconds_bucket{code="200",method="GET",route="kube_default__whoami_0__whoami_dev_local_____whoami_5740a829",le="0.025"} 12
skipper_serve_route_duration_seconds_bucket{code="200",method="GET",route="kube_default__whoami_0__whoami_dev_local_____whoami_5740a829",le="0.05"} 12
skipper_serve_route_duration_seconds_bucket{code="200",method="GET",route="kube_default__whoami_0__whoami_dev_local_____whoami_5740a829",le="0.1"} 12
skipper_serve_route_duration_seconds_bucket{code="200",method="GET",route="kube_default__whoami_0__whoami_dev_local_____whoami_5740a829",le="0.25"} 12
skipper_serve_route_duration_seconds_bucket{code="200",method="GET",route="kube_default__whoami_0__whoami_dev_local_____whoami_5740a829",le="0.5"} 12
skipper_serve_route_duration_seconds_bucket{code="200",method="GET",route="kube_default__whoami_0__whoami_dev_local_____whoami_5740a829",le="1"} 12
skipper_serve_route_duration_seconds_bucket{code="200",method="GET",route="kube_default__whoami_0__whoami_dev_local_____whoami_5740a829",le="2.5"} 12
skipper_serve_route_duration_seconds_bucket{code="200",method="GET",route="kube_default__whoami_0__whoami_dev_local_____whoami_5740a829",le="5"} 12
[...]

PR #978, opened 2019-03-03
Add metric that exposes the name of Ingress and Service that are the source of a route

wndhydrnt:
The metric is only added if Prometheus is activated. It can be used in a join to add more information to other metrics exposed by Skipper. 
See #820 for a detailed description.

Metrics are exported at the Prometheus endpoint like this:
```
# HELP skipper_route_kube_resources Kubernetes resources that are the source of a route.
# TYPE skipper_route_kube_resources gauge
skipper_route_kube_resources{ingress="web-ingress",route="kube_default__web_ingress__web_dev_local_____web_service_prod",service="web-service-prod"} 1
```

A feature flag has been added and exposing this metric is disabled by default.


metrics/all_kind.go
```diff
@@ -93,6 +93,7 @@ func (a *All) IncErrorsStreaming(routeId string) {
 	a.codaHale.IncErrorsStreaming(routeId)
 
 }
+func (a *All) ObserveRouteKubeResources(ingress string, routeId string, service string) {}
```
szuecs: this should call both implementations

szuecs:
A tiny change is required. Everything else looks good to me. 
Thanks @wndhydrnt

wndhydrnt:
Thanks for the review @szuecs. The issue has been fixed.

szuecs:
:+1:

lmineiro:
Isn't this leaking dataclient details?

szuecs:
@lmineiro the whole change is about passing metrics data from dataclient to metrics, such that you can decide which backend version is hit. I think there are a lot of use cases for having this.
How would you do this @lmineiro in a "not leaking" way?

lmineiro:
A good way would be for us to do the same as https://github.com/zalando/skipper/pull/384

wndhydrnt:
@szuecs @lmineiro 
I took a look at the implementation of #384 and I cannot come up with a good generic interface that lets me set three labels and a value for a metric.

What I can think of:
```
# dataclients/dataclients.go
type Metrics interface {
  SetGaugeWithThreeLabels(label1 string, label2 string, label3 string, value float64)
}
```

```
# metrics/prometheus.go
# passing and registering of metrics omitted
customGaugeThreeLabels := prometheus.NewGaugeVec(prometheus.GaugeOpts{
		Namespace: namespace,
		Subsystem: promCustomSubsystem,
		Name:      "gauges_three_labels",
		Help:      "Gauges number of custom metrics.",
}, []string{"label1", "label2", "label3"})

func (p *Prometheus) SetGaugeWithThreeLabels(label1 string, label2 string, label3 string, value float64) {
	p.customCounterM.WithLabelValues(label1, label2, label3).Set(value)
}
```

Did you have something like this in mind?
Please advice.

szuecs:
@wndhydrnt I am not 100% sure why you need 3 labels. I understand that you want to have a measure by backend group metric, which I guess we need also for github.com/zalando-incubator/stackset-controller later to enable automatic rollback to the old version.
For me loadbalancing has a frontend vip and a backend pool and maybe this can be reflected in the metric naming. In kubernetes "frontend" is ingress defined Predicates+Path+HostRegexp and "backend" is the service name.
I also don't think we should have a super generic metric. :)

Do I miss something why you need 3 labels?
Does my comment help you?

wndhydrnt:
No worries @szuecs . I'm with you on not making the metric too generic. I just tried to make sense of the conversation between you two, the code in #384 and the content of this PR and that was the result :).

I used three labels in the example since this PR also uses three labels.

Reading your last comment, I'd go with an interface that looks like this:

```
# dataclients/dataclients.go
type Metrics interface {
  ObserveFrontendInfo(frontend string)
}
```

Does that sound about right?

szuecs:
How about https://github.com/zalando/skipper/issues/820#issuecomment-431593002 and adding `metrics.MeasureServeWithLabel(map[string]string)` and replace the current measurement in proxy/proxy.go?
If label is empty have the current behavior.
If label `len(label) > 0 ` add the labels to the metric data.
Example label: {"version": "v1.2.4"} 

An open question is of course how to set label and here we need input from the dataclient. 
1) We could add a filter that sets and entry into the statebag that we use to get the label settings from the dataclient to a route and check statebag before we use the metrics.MeasureServe() and it non empty we use the new MeasureServeWithLabel().
2) We could add a member to eskip.Route Label: map[string]string, set it in the dataclient and use it in the proxy as in version 1), but without the statebag. 

I think we should implement version 2), because 1) seems to me like a workaround.

@aryszka, @mikkeloscar, @lmineiro what do you think about 1) or 2)?

wndhydrnt:
I'm not sure if this is relevant for the implementation you have in mind, but please note that setting an arbitrary number of labels is hard to do in Prometheus.

From [the docs](https://godoc.org/github.com/prometheus/client_golang/prometheus#GaugeVec.GetMetricWithLabelValues):
> An error is returned if the number of label values is not the same as the number of VariableLabels in Desc (minus any curried labels).

`GaugeVec.WithLabelValues()` [panics](https://github.com/prometheus/client_golang/blob/master/prometheus/gauge.go#L214-L220) if the number of label values do not match the number of label keys set in the description of the metric vector.

szuecs:
@wndhydrnt then maybe the map is overkill and we should go for "version": <value> ?
I am not so much into the prometheus client, so please suggest something that works. :)

wndhydrnt:
Hello @szuecs and sorry for the late reply. I was on vacation.

I would go with what is already in this PR:
* Define a metric and register it with Prometheus
* Expose a method to set the metric via the `Metrics` interface
* Set labels in the Kubernetes dataclient

The naming of the metric could be adjusted to make it clear where it is coming from, e.g. `skipper_dataclient_kube_routeinfo`.

---

A thought that came to my mind just now is to be able to set the ID of a route via an ingress annotation. That would make this whole PR obsolete. Is this an options worth exploring?

szuecs:
> Hello @szuecs and sorry for the late reply. I was on vacation.
> 
> I would go with what is already in this PR:
> 
>     * Define a metric and register it with Prometheus
> 
>     * Expose a method to set the metric via the `Metrics` interface
> 
>     * Set labels in the Kubernetes dataclient

sgtm.

> 
> 
> The naming of the metric could be adjusted to make it clear where it is coming from, e.g. `skipper_dataclient_kube_routeinfo`.

I guess @lmineiro would say it would leak details of the dataclient.
@aryszka & @lmineiro what do you think?

> 
> A thought that came to my mind just now is to be able to set the ID of a route via an ingress annotation. That would make this whole PR obsolete. Is this an options worth exploring?

I would consider route.Id as skipper internal, that is only exposed in a few cases, for example opentracing. Setting the route.Id, I do not like, because it would be easy to mess up as a user.

aryszka:
as I understand, a generic metric method with a `map[string]string` set of labels is not possible because of how Prometheus metrics are defined, and this is why we have now the notion of "kube" in the metrics package, and this indeed couples a particular data client with the metrics package. Some _potentially_ better ways that are on my mind:

- a metrics filter that would rely on https://godoc.org/github.com/zalando/skipper/filters#Metrics , and could be injected via 'zalando.org/skipper-filter' ingress annotations
- a way to support injecting metrics objects, like a wrapper around the prometheus collector (https://godoc.org/github.com/prometheus/client_golang/prometheus#Collector) and the Codahale equivalent, into the metrics registry from other packages

In case of the former, I am not 100% sure if it is good enough for the currently addressed use case.

In case of the latter, I don't know at the moment if it is possible to implement it in a clean way, and what caveats, conflicts it could introduce.

So, I'm suggesting to mark this current approach as experimental, and see if more, similar use cases arise.

szuecs:
@aryszka what about adding application tag and version tag to the errors and latency metrics. A tag could be implemented as label values in Prometheus and default to something in case not used. For codahale metrics we could add something to the dot notation to implement this and use feature flag to enable it.

@wndhydrnt would this solve your issue?
Or maybe @aryszka‘s ideas?

wndhydrnt:
The underlying problem this PR tries to solve is that the route ID itself does not allow identification of the service/backend that is being proxied. After reading your comments on this PR, I've come to the conclusion that the route ID should be treated as a string consisting of random characters. The fact that the Kubernetes data client uses the names of the Ingress and Service to generate the route ID lead me to believe that the route ID carries meaning beyond differentiating one route from another. I was on the wrong track there.

Given the realization that the route ID is just a string of random chars, here is my take on your suggestions:

**Metrics Filter** (@aryszka first idea)
I assume that the name of the service can be set similar to this on an Ingress:
```
zalando.org/skipper-filter: 'routeInfo("myservice")'
```
Since this is a filter, the code that sets the metric would be called on every request, right? Seems like a lot of unnecessary since the actual metric value would not change.
However, that would work in general I think.

**Prometheus Collector** (@aryszka second idea)
I share the concern that this could introduce lots of changes to the code. Since Prometheus and CodaHale are so different in the way they can be used, it could also not be possible at all. Don't think this is worth it.

**Tags** (@szuecs idea)
A label on the latency metrics would solve my problem. How would that be implemented since it cannot be inferred from the route ID?

szuecs:
I suggest to add a map[string]string as Tags member of eskip.Route with that it would be possible to have application and version passed to metrics.

aryszka:
OK, let's drop the collector wrapper approach for now.

I'd like to understand the pros and cons of tagging and the routeInfo filter better. One pro for the filter that I see is: there's no need to introduce a new concept, e.g. we don't need to extend the eskip syntax.

szuecs:
@aryszka I was thinking about tags as an internal vehicle that can be used by our machinery and not accessed by users, so I would not expose this to eskip.
For now I would only add keys "application" and "version" with values from the route creation to the route.Route which we can use in the proxy to target the metrics with Labels. Maybe the filter is the simpler approach let's discuss in person tomorrow how to deal with it.

The goal is to have tuple `application,version` queryable for the metrics. 
Something like this we also need if stackset-controller should be fully automated traffic switching, which should be the goal anyways to free up the deployer from looking at graphs or get a page.

szuecs:
We talked with some nice folks from the gopher community and they hinted us to use Const metrics for the use case.

https://godoc.org/github.com/prometheus/client_golang/prometheus#NewConstMetric

metrics/metrics.go
```diff
@@ -74,6 +74,7 @@ type Metrics interface {
 	IncErrorsStreaming(routeId string)
 	RegisterHandler(path string, handler *http.ServeMux)
 	UpdateGauge(key string, value float64)
+	ObserveRouteKubeResources(ingress string, routeId string, service string)
```
mlvzk: this could be shortened into:
```suggestion
	ObserveRouteKubeResources(ingress, routeId, service string)
```

metrics/codahale.go
```diff
@@ -225,6 +225,8 @@ func (c *CodaHale) IncErrorsStreaming(routeId string) {
 	}
 }
 
+func (c *CodaHale) ObserveRouteKubeResources(ingress string, routeId string, service string) {}
```
mlvzk: this could be shortened into:
```suggestion
func (c *CodaHale) ObserveRouteKubeResources(ingress, routeId, service string) {}
```

metrics/all_kind.go
```diff
@@ -93,6 +93,11 @@ func (a *All) IncErrorsStreaming(routeId string) {
 	a.codaHale.IncErrorsStreaming(routeId)
 
 }
+func (a *All) ObserveRouteKubeResources(ingress string, routeId string, service string) {
```
mlvzk: this could be shortened into:
```suggestion
func (a *All) ObserveRouteKubeResources(ingress, routeId, service string) {
```

metrics/prometheus.go
```diff
@@ -409,3 +419,9 @@ func (p *Prometheus) MeasureBackend5xx(start time.Time) {
 func (p *Prometheus) IncErrorsStreaming(routeID string) {
 	p.proxyStreamingErrorsM.WithLabelValues(routeID).Inc()
 }
+
+func (p *Prometheus) ObserveRouteKubeResources(ingress string, routeId string, service string) {
```
mlvzk: this could be shortened into:
```suggestion
func (p *Prometheus) ObserveRouteKubeResources(ingress, routeId, service string) {
```

metrics/metrics.go
szuecs: @mlvzk thanks for the review, but there are more serious issues with the current PR.

PR #983, opened 2019-03-07
k8s dataclient default filter config via config map

### LINKED ISSUE #976: [Kubernetes dataclient] Default filter configurations via ConfigMap

tkrop:
fixes #976.

Substitutes #977.

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -92,6 +92,8 @@ const (
 	pathPrefixString            = "path-prefix"
 )
 
+const maxFileSize = 1024 * 1024 // 1MB
+
 var internalIPs = []interface{}{
 	"10.0.0.0/8",
 	"192.168.0.0/16",
@@ -173,7 +175,8 @@ type Options struct {
 	// KubernetesEastWestDomain sets the DNS domain to be used for east west traffic, defaults to "skipper.cluster.local"
 	KubernetesEastWestDomain string
 
-	// DefaultFiltersDir enables default filters mechanism and sets the location of the default filters
+	// DefaultFiltersDir enables default filters mechanism and sets the location of the default filters.
+	// The provided filters are then applied to all routes.
 	DefaultFiltersDir string
 }
 
@@ -1402,8 +1405,6 @@ func (c *Client) fetchDefaultFilterConfigs() map[resourceId]string {
 	return filters
 }
 
-const maxFileSize = 1024 * 1024 // 1MB
-
 func (c *Client) getDefaultFilterConfigurations() (map[resourceId]string, error) {
 	files, err := ioutil.ReadDir(c.defaultFiltersDir)
 	if err != nil {
@@ -1413,7 +1414,7 @@ func (c *Client) getDefaultFilterConfigurations() (map[resourceId]string, error)
 	filters := make(map[resourceId]string)
 	for _, f := range files {
 		r := strings.Split(f.Name(), ".") // format: {service}.{namespace}
-		if len(r) != 2 || f.IsDir() || f.Size() > maxFileSize {
+		if len(r) != 2 || notRegularFile(f) || f.Size() > maxFileSize {
 			log.WithError(err).WithField("file", f.Name()).Debug("incompatible file")
 			continue
 		}
@@ -1429,4 +1430,10 @@ func (c *Client) getDefaultFilterConfigurations() (map[resourceId]string, error)
 	}
 
 	return filters, nil
+}
+
+func notRegularFile(f os.FileInfo) bool {
+	mode := f.Mode()
+	return f.IsDir() || mode == os.ModeIrregular || mode == os.ModeSymlink || mode == os.ModeDevice ||
+		mode == os.ModeNamedPipe || mode == os.ModeSocket
 }
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"regexp"
 	"sort"
 	"strconv"
@@ -171,6 +172,9 @@ type Options struct {
 
 	// KubernetesEastWestDomain sets the DNS domain to be used for east west traffic, defaults to "skipper.cluster.local"
 	KubernetesEastWestDomain string
+
+	// DefaultFiltersDir enables default filters mechanism and sets the location of the default filters
+	DefaultFiltersDir string
 }
 
 // Client is a Skipper DataClient implementation used to create routes based on Kubernetes Ingress settings.
@@ -194,6 +198,7 @@ type Client struct {
 	ingressClass                *regexp.Regexp
 	pathMode                    PathMode
 	quit                        chan struct{}
+	defaultFiltersDir           string
 }
 
 var (
@@ -288,6 +293,7 @@ func New(o Options) (*Client, error) {
 		ingressesURI:                ingressesClusterURI,
 		servicesURI:                 servicesClusterURI,
 		endpointsURI:                endpointsClusterURI,
+		defaultFiltersDir:           o.DefaultFiltersDir,
 	}
 	if o.KubernetesNamespace != "" {
 		result.setNamespace(o.KubernetesNamespace)
@@ -767,7 +773,7 @@ func applyAnnotationPredicates(m PathMode, r *eskip.Route, annotation string) er
 // valid ones.  Reporting failures in Ingress status is not possible,
 // because Ingress status field is v1.LoadBalancerIngress that only
 // supports IP and Hostname as string.
-func (c *Client) ingressToRoutes(state *clusterState) ([]*eskip.Route, error) {
+func (c *Client) ingressToRoutes(state *clusterState, defaultFilters map[resourceId]string) ([]*eskip.Route, error) {
 	routes := make([]*eskip.Route, 0, len(state.ingresses))
 	hostRoutes := make(map[string][]*eskip.Route)
 	redirect := createRedirectInfo(c.provideHTTPSRedirect, c.httpsRedirectCode)
@@ -801,6 +807,7 @@ func (c *Client) ingressToRoutes(state *clusterState) ([]*eskip.Route, error) {
 			}
 			annotationFilter += val
 		}
+
 		// parse predicate annotation
 		var annotationPredicate string
 		if val, ok := i.Metadata.Annotations[skipperpredicateAnnotationKey]; ok {
@@ -898,6 +905,16 @@ func (c *Client) ingressToRoutes(state *clusterState) ([]*eskip.Route, error) {
 						}
 					}
 
+					// add pre-configured default filters
+					if defFilter, ok := defaultFiltersOf(prule.Backend.ServiceName, i.Metadata.Namespace, defaultFilters); ok {
+						defaultFilters, err := eskip.ParseFilters(defFilter)
+						if err != nil {
+							logger.Errorf("Can not parse default filters: %v", err)
+						} else {
+							endpointsRoute.Filters = append(defaultFilters, endpointsRoute.Filters...)
+						}
+					}
+
 					err = applyAnnotationPredicates(pathMode, endpointsRoute, annotationPredicate)
 					if err != nil {
 						logger.Errorf("failed to apply annotation predicates: %v", err)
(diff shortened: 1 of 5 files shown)


tkrop:
@szuecs asked to make this feature available for redirects https://github.com/zalando/skipper/pull/977#discussion_r262839270.

szuecs:
https://github.com/zalando/skipper/pull/1003/ implemented a PreProcessor facility that works on []eskip.Route and can add []eskip.Filters as default, like in the mentioned PR.
This could be used to integrate without the dataclient `LoadUpdate()` and `LoadAll()`.

maxim-tschumak:
@tkrop @zeitlinger @szuecs @aryszka @aermakov-zalando please have a look!

dataclients/kubernetes/kube.go
```diff
@@ -788,6 +794,15 @@ func (c *Client) ingressToRoutes(state *clusterState) ([]*eskip.Route, error) {
 			}
 			annotationFilter += val
 		}
+
+		// add pre-configured default filters
+		if defFilter, ok := defaultFiltersOf(i, defaultFilters); ok {
+			if annotationFilter != "" {
+				annotationFilter += " -> "
+			}
+			annotationFilter += defFilter
```
tkrop: Do we really want to add this filters at the end?

We originally talked about putting monitoring at the incoming edge of the API gateway, i.e. at the begin of the filter chain.

dataclients/kubernetes/kube.go
```diff
@@ -957,6 +972,29 @@ func (c *Client) ingressToRoutes(state *clusterState) ([]*eskip.Route, error) {
...
+	services := make(map[string]struct{})
+	for _, r := range i.Spec.Rules {
+		for _, p := range r.Http.Paths {
+			services[p.Backend.ServiceName] = struct{}{}
+		}
+	}
+
+	for s := range services {
+		if val, ok := defaultFilters[resourceId{name: s, namespace: i.Metadata.Namespace}]; ok {
+			filters = append(filters, val)
```
tkrop: This loop is against my intuition. I currently can't prove, but I think this function is doing the wrong thing and is called from the wrong place in the execution path.

docs/operation/operation.md
```diff
@@ -470,6 +471,14 @@ using the Redis ring based solution, adds 2 additional roundtrips to
 redis per hit. Make sure you monitor redis closely, because skipper
 will fallback to allow traffic if redis can not be reached.
 
+## Default filters
+
+Kubernetes dataclient supports default filters. You can enable this feature by
+specifying `default-filters-dir`. The defined directory must contain per-service
+filter configurations, with file name following the pattern `${service}.${namespace}`.
```
aermakov-zalando: There will be race conditions between the ingress/service/etc being created and the default filter for it being added to the configmap. If the default filters configure something like authentication, this can cause serious issues. It also makes it impossible for an administrator to just setup a bunch of filters for the whole cluster, because they now have to write a cronjob or a dedicated controller to watch the services. Are there actually any use cases for this, or was it just added for flexibility?

maxim-tschumak: In general, you are right.

Our use-case behind the change is the following: we want to provide out-of-the-box monitoring, i.e. we will have a facility in place which generates API monitoring filter configurations based on the (published) API specifications. These configs are then persisted as "default filters".

The api-monitoring-controller, which is responsible for generating the default filters, takes into account which services/ingresses are already manually configured and doesn't create default configurations in such cases. No race conditions.

We should extend the documentation with the hint about the possible race conditions.

tkrop:
@maxim-tschumak Found the flaw in the default filter concept: Our `apiUsageMonitoring`-filter requires us to add a single filter configuration matching a *backend service*. This requires us to add the filter inside the `Host`/`Path`-loop between https://github.com/zalando/skipper/pull/983/files#diff-901a01c39ef9601977a8ff66e7e30c54R857 and https://github.com/zalando/skipper/pull/983/files#diff-901a01c39ef9601977a8ff66e7e30c54R890.

If not done this way, we would potentially add multiple `apiUsageMonitoring`-filters which would result in counting every request multiple times. This is also a reason why the `api-monitroing-controller` must suppress its default filter creation, if the ingress configuration adds an own `apiUsageMonitoring`-filter.

We really should add a warning, that the `apiUsageMonitoring`-filter must not be configured more than once per route, unless we find a way to merge multiple `apiUsageMonitoring`-filter configurations in the skipper filter chain.

docs/operation/operation.md
aermakov-zalando: Why not add global default filters as well, which seems to me a more common use case?

maxim-tschumak: @szuecs already implemented **global** filters (#1002). This proposal is to have default filters **per service**.

tkrop: We have this on our radar and are very careful about which filters we plan to add here. The current premise is, that each filter chain/route works perfect without this additional filters. If we change this premise, we will make sure that skipper routes will not work unless a default filter configuration is available :smile:.

aermakov-zalando: Ah, I've missed the first part. Sorry then!

aermakov-zalando: Ah, I've missed the first part. But then why not have the controller update the filters in the service object (or the corresponding ingress) directly, instead of going through the configmap and then trying to map it back to the services in the data client?

szuecs: @aermakov-zalando 

1. This will create a race condition and probably flapping with stackset-controller
2. defaults should work for all dataclients, not only in kubernetes
3. it's more efficient if we read from a file, then to query ingress and get a lot of more or less static data.

maxim-tschumak: @szuecs We are going to move the filter configuration to the Service resource (opt-out and custom config). This will be a higher level configuration and will be the basis for `api-monitoring-controller` to generate the effective `apiUsageMonitoring` filter configurations.

tkrop: @szuecs I would say the service resource is the more precise location for API filter configurations, because the setup is related to target service and not some ingress definition. There surely exist other situations - and it is a valid option to implement the same mechanism against ingress resources, but this is not our use case. Anyhow, using the service as reference and key in the configuration makes sure, that we do not have conflicts in situations where one ingress configuration is actually leading to two different service - where we would require two different filters.

tkrop:
:+1:

dataclients/kubernetes/kube.go
maxim-tschumak: changed to prepend

dataclients/kubernetes/kube.go
maxim-tschumak: refactored to add the default filters per route (instead of doing it per ingress)

maxim-tschumak:
FYI: I've squashed the commits and rebased to the current master.

dataclients/kubernetes/kube.go
```diff
@@ -171,6 +172,9 @@ type Options struct {
 
 	// KubernetesEastWestDomain sets the DNS domain to be used for east west traffic, defaults to "skipper.cluster.local"
 	KubernetesEastWestDomain string
+
+	// DefaultFiltersDir enables default filters mechanism and sets the location of the default filters
```
szuecs: I miss something like: default filters will be applied to all routes

dataclients/kubernetes/kube.go
```diff
@@ -898,6 +905,16 @@ func (c *Client) ingressToRoutes(state *clusterState) ([]*eskip.Route, error) {
 						}
 					}
 
+					// add pre-configured default filters
+					if defFilter, ok := defaultFiltersOf(prule.Backend.ServiceName, i.Metadata.Namespace, defaultFilters); ok {
+						defaultFilters, err := eskip.ParseFilters(defFilter)
+						if err != nil {
+							logger.Errorf("Can not parse default filters: %v", err)
```
szuecs: nitpick: "Can not" -> "Failed to"

dataclients/kubernetes/kube.go
```diff
@@ -974,6 +991,13 @@ func (c *Client) ingressToRoutes(state *clusterState) ([]*eskip.Route, error) {
 	return routes, nil
 }
 
+func defaultFiltersOf(service string, namespace string, defaultFilters map[resourceId]string) (string, bool) {
+	if filters, ok := defaultFilters[resourceId{name: service, namespace: namespace}]; ok {
```
szuecs: I think this should work and would be the same:
```
return defaultFilters[resourceId{name: service, namespace: namespace}]
```

dataclients/kubernetes/kube.go
```diff
@@ -1356,3 +1383,50 @@ func (c *Client) Close() {
...
+		log.WithError(err).Error("could not fetch default filter configurations")
+		return make(map[resourceId]string)
+	}
+
+	log.WithField("#configs", len(filters)).Debug("default filter configurations loaded")
+
+	return filters
+}
+
+const maxFileSize = 1024 * 1024 // 1MB
```
szuecs: please move the const to the top of the file

dataclients/kubernetes/kube.go
```diff
@@ -1356,3 +1383,50 @@ func (c *Client) Close() {
...
+func (c *Client) getDefaultFilterConfigurations() (map[resourceId]string, error) {
+	files, err := ioutil.ReadDir(c.defaultFiltersDir)
+	if err != nil {
+		return nil, err
+	}
+
+	filters := make(map[resourceId]string)
+	for _, f := range files {
+		r := strings.Split(f.Name(), ".") // format: {service}.{namespace}
+		if len(r) != 2 || f.IsDir() || f.Size() > maxFileSize {
```
szuecs: f could be also a fifo or a socket, which both would block.
I think you should check for file or symlink (symlink could also target a socket, which is bad, but maybe ok enough)

dataclients/kubernetes/kube_test.go
```diff
@@ -3639,3 +3640,147 @@ func TestCreateEastWestRouteOverwriteDomain(t *testing.T) {
...
+		})
+		if err != nil {
+			t.Error(err)
+		}
+
+		defer dc.Close()
+
+		r, err := dc.LoadAll()
+
+		assert.Nil(t, err)
```
szuecs: Please do not use `assert`, we would like to drop the library in the future.

tkrop: Interesting. Why?

dataclients/kubernetes/kube.go
tkrop: Thinking about it, wouldn't be 16k enough.

dataclients/kubernetes/kube.go
tkrop: Looks like.

dataclients/kubernetes/kube.go
maxim-tschumak: It's consistent to already implemented `Can not parse annotation filters` error (line 903).

dataclients/kubernetes/kube.go
tkrop: Agreed. May be we should also add the term target service to make the context clearer.

dataclients/kubernetes/kube.go
maxim-tschumak: It doesn't if we not explicitly say that we expect 2 return values.

dataclients/kubernetes/kube_test.go
szuecs: Because it adds maintenance work for using dependencies and bloat the binary and compile time

dataclients/kubernetes/kube.go
maxim-tschumak: Good point!

dataclients/kubernetes/kube.go
```diff
@@ -1356,3 +1386,54 @@ func (c *Client) Close() {
...
+
+		filters[resourceId{name: r[0], namespace: r[1]}] = string(config)
+	}
+
+	return filters, nil
+}
+
+func notRegularFile(f os.FileInfo) bool {
+	mode := f.Mode()
+	return f.IsDir() || mode == os.ModeIrregular || mode == os.ModeSymlink || mode == os.ModeDevice ||
```
aryszka: q: what is the reason we're not allowing symlinks here?

aryszka:
:+1:

dataclients/kubernetes/kube.go
maxim-tschumak: Actually, we have to allow symlinks. Good catch, thank you!

aryszka:
:+1:

dataclients/kubernetes/kube.go
tkrop: Change both, because else it would be `cannot` :smile_cat:

szuecs:
:+1:

PR #984, opened 2019-03-07
feat: add cache for retry after value

tkrop:
Have not tested it completely, but this outlines the simple idea to cache the `retry` time stamp to optimize the number of `redis` request under heavy request load with minimal overhead.


szuecs:
@tkrop I will close this now

PR #986, opened 2019-03-10
Support loadbalancing routes for HTTP Upgrade handling.

### LINKED ISSUE #982: Support Websockets.
Currently Skipper support the Upgrade connection which is enough for a server usecase.

When a browser client tries to connect, it enforces the usage of the scheme `ws` and `wss` for TLS based connections.

Currently Skipper throws an error here: https://github.com/zalando/skipper/blob/master/proxy/upgrade.go#L193

The RFC for this protocol is https://tools.ietf.org/html/rfc6455

angelaigreja:
Fixes #982.

The url in the request has the selected host for any type of routes.
I am not sure if any more changes are needed but with a local websocket server, this fix was enough.

This area of the code is not tested at all and this PR does not try to improve upon the situation. The added test just cover the bug that is fixing.

### DIFF
--- a/proxy/proxy_test.go
+++ b/proxy/proxy_test.go
@@ -739,6 +739,8 @@ func TestProcessesRequestWithHTTPUpgradeWithLoadBalancing(t *testing.T) {
 	backend := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusNoContent) // bad header on purpose for the test
 	}))
+	defer backend.Close()
+
 	u, _ := url.ParseRequestURI("wss://www.example.org/ws")
 	r := &http.Request{
 		URL:    u,
--- a/proxy/proxy_test.go
+++ b/proxy/proxy_test.go
@@ -735,6 +735,32 @@ func TestProcessesRequestWithShuntBackend(t *testing.T) {
 	}
 }
 
+func TestProcessesRequestWithHTTPUpgradeWithLoadBalancing(t *testing.T) {
+	backend := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusNoContent) // bad header on purpose for the test
+	}))
+	u, _ := url.ParseRequestURI("wss://www.example.org/ws")
+	r := &http.Request{
+		URL:    u,
+		Method: "GET",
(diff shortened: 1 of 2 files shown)


szuecs:
If I test the change with: 

```
./bin/skipper -inline-routes='r1: Host("foo") -> inlineContent("OK") -> status(200) -> <shunt>; r2: Host("bar") -> inlineContent("OK") -> status(201) -> <shunt>; r0: * -> "http://127.0.0.1:8080"' -address=:9999
```

I get the following result:
skipper proxy: 
```
[APP]ERRO[0010] error while copying the response streamhttp: request method or response status code does not allow body
```

ws server:
```
% % wscat -l 8080
listening on port 8080 (press CTRL+C to quit)
client connected
> a
disconnected
```

ws client:
```
% wscat -c http://127.0.0.1:9999/
connected (press CTRL+C to quit)
>
>
> d
```

ruiaraujo:
You need to enable the upgrade support.

angelaigreja:
If you run skipper like this:
`./bin/skipper -inline-routes='r1: Host("foo") -> inlineContent("OK") -> status(200) -> <shunt>; r2: Host("bar") -> inlineContent("OK") -> status(201) -> <shunt>; r0: * -> "http://127.0.0.1:8080"' -address=:9999 -experimental-upgrade`

your example works correctly.

Maybe we should break the if check apart here: https://github.com/zalando/skipper/blob/7d388ef92ce7add3bada17de92cece49da0dc653/proxy/proxy.go#L761

We could check first if it is a upgradable request and then warn in case the support is not enabled, it would be a better UX IMO with the side effect of more overhead for those who do not have it enabled.

szuecs:
@angelaigreja @ruiaraujo true and I wondered how this works, but the request object is already created. I also checked that other routes are working, so from this perspective it's fine.
There is a bit of cleanup needed. You do not need the route anymore in the `makeUpgradeRequest`, so please clean this up, too

proxy/proxy_test.go
```diff
@@ -735,6 +735,32 @@ func TestProcessesRequestWithShuntBackend(t *testing.T) {
 	}
 }
 
+func TestProcessesRequestWithHTTPUpgradeWithLoadBalancing(t *testing.T) {
+	backend := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
aryszka: please close the test server: `defer backend.Close()`

aryszka:
:+1:

ruiaraujo:
:+1:

szuecs:
:+1:

Issue #867, opened 2018-11-29
Response filter to redirect if backend response has status code >=500


futare1:
I want to have a chosen redirect if the backend response status code is >= 500. 

A solution could be a specific responsefilter `redirectToGtStatus("https://example.org", 500)`, that would redirect to https://example.org with a temporary redirect `307`, only if the response status code is >= 500 from the backend.

## Original issue
Hi - I'm new to Skipper.  Can anyone please point me in the direction of an example to configure/redirect to static outage message page for when a backend received a >=500 status code instead of the "Bad Gateway" message?

**Scenario:**
outage.domain.local (httpd app serving outage message html content)
app.domain.local (httpd app)

app Ingress:
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: "app-ingress"
  namespace: default
spec:
  rules:
  - host: "app.domain.local"
    http:
      paths:
      - path: /
        backend:
          serviceName: app-service
          servicePort: 80
```
Any help is appreciated.
Wikus

szuecs:
Can you share the output of `curl -v https://app.domain.local/` please?
Is `domain.local` inside the cluster or a valid domain, that you hide from us?

futare1:
Hi szuecs, `domain.local` is only an example.  'outage' is a simple web app with a static page for customers when we do maintenace.  `app` is the production web app.  Say during maintenance of production `app` , a >=500 is returned and I wish to know how to redirect to the `outage` 

(in hindsight I should have referred to it as maintenance instead of outage).

szuecs:
From what I understand you would like to do:

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: "app-ingress"
  namespace: default
    zalando.org/skipper-routes: |                                                                                       
       simpleredirect: PathRegexp("/") -> redirectTo(307, "https://outage.domain.local/") -> <shunt>;      
spec:
  rules:
  - host: "app.domain.local"
    http:
      paths:
      - path: /
        backend:
          serviceName: app-service
          servicePort: 80
```

You would get another route which is more specific (+1 Predicate PathRegexp()) then the route from the ingress spec. The more specific route would do the redirect as specified.

futare1:
Hi szuecs - thanks for the example of the redirect.  Unfortunately, I'm trying to do the opposite.  

[...]

PR #995, opened 2019-03-18
Log access filter

njuettner:
This PR allows to set an log access filter global.
By passing the flag --access-log-filter it is possible to log by HTTP status codes.

Example: --access-log-filter=2,301

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -15,12 +15,13 @@ import (
 	"strconv"
 	"time"
 
+	al "github.com/zalando/skipper/filters/accesslog"
+
 	ot "github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
 	"github.com/zalando/skipper/circuit"
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/filters"
-	al "github.com/zalando/skipper/filters/accesslog"
 	circuitfilters "github.com/zalando/skipper/filters/circuit"
 	ratelimitfilters "github.com/zalando/skipper/filters/ratelimit"
 	tracingfilter "github.com/zalando/skipper/filters/tracing"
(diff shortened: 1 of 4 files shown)


arjunrn:
Personally I would avoid the use of the term filter here because it's ambiguous wrt to what filter means in a Skipper context. I would recommend calling it something else.

njuettner:
I'm open here 😃. What ever we can agree on. 
TBH I'm currently out of ideas how to name it better.

arjunrn:
How about `access-log-matching-enabled`?

skipper.go
```diff
@@ -1060,3 +1065,21 @@ func Run(o Options) error {
 
 	return listenAndServe(proxy, &o)
 }
+
+func parseAccessLogFilter(accessLogDisabled bool, filter string) al.AccessLogFilter {
```
mikkeloscar: Maybe add a test for this function? :)

proxy/proxy.go
```diff
@@ -120,8 +121,8 @@ type Params struct {
 	// and the response messages during web socket upgrades.
 	ExperimentalUpgradeAudit bool
 
-	// When set, no access log is printed.
-	AccessLogDisabled bool
```
szuecs: This is a breaking change and I am sure we can have a better solution if we change it like this which might support also https://github.com/zalando/skipper/issues/867

skipper.go
szuecs: There is already filter parser in `eskip` package, that you can use, why not use this instead of this function?

proxy/proxy.go
mikkeloscar: I wouldn't consider this a breaking change since it's a change of what is essentially an internal package in skipper. I understand that packages are public in Go, but I would not expect anyone to use the proxy package directly. Or do you think there is a use case for using the proxy package directly?

szuecs: I will close the PR as discussed with @njuettner . I doubt, that the idea is more a hack than needed. If we would have a default filter applied in routing package it would make more sense and create more value.

szuecs:
As discussed we will check and plan this change.

PR #998, opened 2019-03-20
fix close-idle-conns-period flag

szuecs:
fix -close-idle-conns-period flag handling to use plain time.Duration parsing and not adjusted string handling which does not work

### DIFF
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -177,7 +177,7 @@ const (
 	// connections, timeouts:
 	waitForHealthcheckIntervalUsage   = "period waiting to become unhealthy in the loadbalancer pool in front of this instance, before shutdown triggered by SIGINT or SIGTERM"
 	idleConnsPerHostUsage             = "maximum idle connections per backend host"
-	closeIdleConnsPeriodUsage         = "period of closing all idle connections in seconds or as a duration string. Not closing when less than 0"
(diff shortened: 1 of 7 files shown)


cmd/skipper/main.go
```diff
@@ -697,7 +691,7 @@ func main() {
 		// connections, timeouts:
 		WaitForHealthcheckInterval:   waitForHealthcheckInterval,
 		IdleConnectionsPerHost:       idleConnsPerHost,
-		CloseIdleConnsPeriod:         time.Duration(clsic) * time.Second,
+		CloseIdleConnsPeriod:         closeIdleConnsPeriod,
```
aryszka: i think to fix the bug, this single change is the right one. All the rest of the changes are not required and they also would break compatibility.

cmd/skipper/main.go
```diff
@@ -569,12 +569,6 @@ func main() {
 		eus = strings.Split(etcdUrls, ",")
 	}
 
-	clsic, err := parseDurationFlag(closeIdleConnsPeriod)
```
aryszka: if we want to stay consistent, then we should remove the parseDurationFlag usage from the breakerflags.go and ratelimiterflags.go, as well, and delete the parseDurationFlag function.

szuecs:
@aryszka I also changed the docs, that were related to ratelimits, which were wrong or missing.

aryszka:
:+1:

szuecs:
:+1:

Issue #1002, opened 2019-03-21
apply default filter set as prefix filters


szuecs:
**Is your feature request related to a problem? Please describe.**

We want to be able to set a default filter to enable access logs only for errors, but want to be able that users can override based on their needs this global decision.

**Describe the solution you would like**

A solution would be to read default filters from some input, for example CLI flags and patch in our default filters to all routes in the routing package which call `LoadUpdate()` and `LoadAll()`.

**Describe alternatives you've considered (optional)**

I thought about putting it into dataclients, which would be not really maintainable.

**Would you like to work on it?**

Yes

aryszka:
At first read, this sounds good to me. Could be a preprocessor in the routing, similar to the postprocessors, but not on *routing.Routes but on *eskip.Routes.

topicus:
Is it possible to use a config file instead of CLI flags (not just for this but for all the options)? Once skipper is running I didn't find a simple way to capture the current configuration and a config file may mitigate this issue.

szuecs:
@topicus why not file an issue for that?

topicus:
Yes, I wanted to get your feedback first.

PR #1003, opened 2019-03-21
default filters as routing.PreProcessors

szuecs:
Implements #1002

### DIFF
--- a/cmd/skipper/defaultfilterflags.go
+++ b/cmd/skipper/defaultfilterflags.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/zalando/skipper/eskip"
 )
@@ -17,11 +16,7 @@ type defaultFiltersFlags struct {
 }
 
 func (dpf *defaultFiltersFlags) String() string {
-	res := make([]string, len(dpf.filters))
-	for _, f := range dpf.filters {
-		res = append(res, f.Name)
-	}
-	return strings.Join(res, " -> ")
+	return "<default filters>"
 }
 
 func (dpf *defaultFiltersFlags) Set(value string) error {
--- a/cmd/skipper/defaultfilterflags.go
+++ b/cmd/skipper/defaultfilterflags.go
@@ -1,0 +1,43 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/zalando/skipper/eskip"
+)
+
+const (
+	defaultPrependFiltersUsage = "set of default filters to apply to prepend to all filters of all routes"
+	defaultAppendFiltersUsage  = "set of default filters to apply to append to all filters of all routes"
+)
+
+type defaultFiltersFlags struct {
+	filters []*eskip.Filter
+}
+
+func (dpf *defaultFiltersFlags) String() string {
+	res := make([]string, len(dpf.filters))
+	for _, f := range dpf.filters {
+		res = append(res, f.Name)
+	}
+	return strings.Join(res, " -> ")
+}
+
+func (dpf *defaultFiltersFlags) Set(value string) error {
+	if dpf == nil {
+		dpf = &defaultFiltersFlags{}
+	}
+
+	fs, err := eskip.ParseFilters(value)
+	if err != nil {
+		return fmt.Errorf("failed to parse default filters: %v", err)
+	}
+
+	dpf.filters = append(dpf.filters, fs...)
+	return nil
+}
+
+func (dpf *defaultFiltersFlags) Get() []*eskip.Filter {
+	return dpf.filters
+}
--- a/eskip/eskip.go
+++ b/eskip/eskip.go
@@ -20,6 +20,20 @@ var (
 	duplicateMethodPredicateError   = errors.New("duplicate method predicate")
 )
 
+type DefaultFilters struct {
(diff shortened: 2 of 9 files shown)


cmd/skipper/defaultfilterflags.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+type defaultFiltersFlags struct {
+	filters []*eskip.Filter
+}
+
+func (dpf *defaultFiltersFlags) String() string {
+	res := make([]string, len(dpf.filters))
+	for _, f := range dpf.filters {
+		res = append(res, f.Name)
```
aryszka: this will print sg like: `foo -> bar -> baz`. Where is it used?

skipper.go
```diff
@@ -861,6 +865,7 @@ func Run(o Options) error {
 			loadbalancer.HealthcheckPostProcessor{LB: lbInstance},
 			loadbalancer.NewAlgorithmProvider(),
 		},
+		PreProcessors:   []routing.PreProcessor{o.DefaultFilters},
```
aryszka: we should not add the o.DefaultFilters when it is nil

cmd/skipper/defaultfilterflags.go
szuecs: only if you print the flags, so I don't really care

eskip/eskip.go
```diff
@@ -20,6 +20,20 @@ var (
 	duplicateMethodPredicateError   = errors.New("duplicate method predicate")
 )
 
+type DefaultFilters struct {
+	Prepend []*Filter
+	Append  []*Filter
+}
+
+func (df *DefaultFilters) Do(r []*Route) []*Route {
+	for i := range r {
```
aryszka: we should clone the routes objects. The reason is that a DataClient may cache parsed routes, and this logic changes the Filters field on those routes, too. This way we may end up prepending and appending to the same slice of filters.

skipper.go

routing/datasource.go
```diff
@@ -494,6 +494,11 @@ func receiveRouteMatcher(o Options, out chan<- *routeTable, quit <-chan struct{}
 		select {
 		case defs := <-updatesRelay:
 			o.Log.Info("route settings received")
+
+			for i := range o.PreProcessors {
```
aryszka: it would be nice to have a test checkes that they are applied

eskip/eskip.go
```diff
@@ -20,6 +20,20 @@ var (
 	duplicateMethodPredicateError   = errors.New("duplicate method predicate")
 )
 
+type DefaultFilters struct {
+	Prepend []*Filter
+	Append  []*Filter
+}
+
+func (df *DefaultFilters) Do(r []*Route) []*Route {
```
aryszka: it would be nice to have a test that the filters are prepended and appended

eskip/eskip.go
```diff
@@ -20,6 +20,20 @@ var (
 	duplicateMethodPredicateError   = errors.New("duplicate method predicate")
 )
 
+type DefaultFilters struct {
```
aryszka: A short godoc could explain that this is a PreProcessor and to be used with routing package.

eskip/eskip.go
szuecs: done

eskip/eskip.go
szuecs: done

cmd/skipper/defaultfilterflags.go
```diff
@@ -0,0 +1,43 @@
...
+	defaultPrependFiltersUsage = "set of default filters to apply to prepend to all filters of all routes"
+	defaultAppendFiltersUsage  = "set of default filters to apply to append to all filters of all routes"
+)
+
+type defaultFiltersFlags struct {
+	filters []*eskip.Filter
+}
+
+func (dpf *defaultFiltersFlags) String() string {
+	res := make([]string, len(dpf.filters))
```
lmineiro: shouldn't this be `make([]string, 0, len(dpf.filters))`?

lmineiro: The append below will start appending elements after `len(dpf.filters)`

aryszka: :+1:

routing/datasource.go
szuecs: done

cmd/skipper/defaultfilterflags.go
szuecs: it's not different if you have no `if ... continue` or `break`

aryszka: it should be also fine if we just remove this method

aryszka: my bad, must implement Stringer

szuecs: we decided to drop this and just return a static string instead

aryszka: @szuecs if it was only about the initialization, you'd be right. The comment is about how the append() will behave afterwards

szuecs: https://play.golang.org/p/GSEH2W1uANR @lmineiro you are right

szuecs:
:+1:

aryszka:
:+1:

PR #1007, opened 2019-03-26
A / B test example

fpozzobon:
Added stickyness to Traffic on the cookie, without it we noticed an unstable behaviour (traffic routed partially to A and B)
Suggested in https://github.com/zalando/skipper/blob/master/docs/reference/predicates.md


szuecs:
Closed in favor of https://github.com/zalando/skipper/pull/1017

PR #1008, opened 2019-03-26
Make unverifiedAuditLog configurable

irishshagua:
Adds an optional single input value to the unverifiedAuditLog filter
which acts as a lookup key for the Auth token in case you need to
log something other than the sub. If no parameter is provided for
the unverifiedAuditLog filter then it will default to "sub"

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -951,6 +951,10 @@ be wrote to the log file. By default, the value of the audit header will be equa
 the Authorization token. This can be changed by providing a `string` input to the filter which matches another key from the
 token.
 
+*N.B.* It is important to note that, if the content of the `X-Unverified-Audit` header does not match the following regex, then
+a default value of `invalid-sub` will be populated in the header instead:
+    `^[a-zA-z0-9_/:?=&%@.#-]*$`    
+
 Examples:
 
 ```
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -947,7 +947,7 @@ auditLog()
 ## unverifiedAuditLog
 
 Filter `unverifiedAuditLog()` adds a Header, `X-Unverified-Audit`, to the request, the content of which, will also
-be wrote to the log file. By default, the value of the audit header will be equal to the value of the `sub` key, from
+be written to the log file. By default, the value of the audit header will be equal to the value of the `sub` key, from
 the Authorization token. This can be changed by providing a `string` input to the filter which matches another key from the
 token.
 
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -944,6 +944,22 @@ Example:
 auditLog()
 ```
 
+## unverifiedAuditLog
+
+Filter `unverifiedAuditLog()` adds a Header, `X-Unverified-Audit`, to the request, the content of which, will also
+be wrote to the log file. By default, the value of the audit header will be equal to the value of the `sub` key, from
+the Authorization token. This can be changed by providing a `string` input to the filter which matches another key from the
+token.
+
+Examples:
+
+```
+unverifiedAuditLog()
+```
+
+```
+unverifiedAuditLog("azp")
+```
 
 ## setDynamicBackendHostFromHeader
 
--- a/filters/log/log.go
+++ b/filters/log/log.go
@@ -30,13 +30,15 @@ const (
 	// reject reason information into the state bag to pass the
 	// information to the auditLog filter.
 	AuthRejectReasonKey = "auth-reject-reason"
-	// UnverifiedAuditLogName is th filtername seend by the user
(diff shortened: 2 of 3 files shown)


szuecs:
:+1:

filters/log/log.go
```diff
@@ -208,16 +221,21 @@ func (ual *unverifiedAuditLog) Request(ctx filters.FilterContext) {
...
+		var j map[string]interface{}
 		err = json.Unmarshal(sDec, &j)
 		if err != nil {
 			return
 		}
-		req.Header.Add(UnverifiedAuditHeader, cleanSub(j.Sub))
+
+		if k, ok := j[ual.TokenKey]; ok {
+			if v, ok2 := k.(string); ok2 {
+				req.Header.Add(UnverifiedAuditHeader, cleanSub(v))
```
herojan: Does cleanSub work for non-sub parts?

irishshagua: It's some kind of sanitizing regex that the value needs to match:
`re = regexp.MustCompile("^[a-zA-z0-9_/:?=&%@.#-]*$")`

Looks like it should be good for anything that we'd want though...

herojan: Ah I see it just has to match `^[a-zA-z0-9_/:?=&%@.#-]*$`, but then it defaults the header to <invalid-sub> when it may not be sub they were looking for

szuecs: Maybe this should be added to the documentation, that the data has to match this regular expression.

irishshagua: Adding key specific defaults might be a bit overkill. If the key is not present, it will not use the default. So it would only end up being set if the value of the dynamic was invalid as per the regex.... 🤔

irishshagua: @szuecs this particular filter isnt documented. Not sure what the reason is for that.
I'm happy to throw in some documentation as part of this PR though

szuecs: Yes, sorry for that. It was my fault not to document it in the first place.

szuecs: We should have defaults to not break compatibility. I don't know if someone is using it, but we want to do our best to not break filter configurations out there.

irishshagua: Ok @szuecs / @herojan 
Documentation added for the `unverifiedAuditLog` filter.
I was careful to try and not break backward compatibility. If there is no parameter passed to the filter, as per previously, it will default to using the sub key in the lookup...

szuecs:
:+1:

irishshagua:
👍

herojan:
👍

docs/reference/filters.md
```diff
@@ -944,6 +944,26 @@ Example:
 auditLog()
 ```
 
+## unverifiedAuditLog
+
+Filter `unverifiedAuditLog()` adds a Header, `X-Unverified-Audit`, to the request, the content of which, will also
+be wrote to the log file. By default, the value of the audit header will be equal to the value of the `sub` key, from
```
cjbooms: written

irishshagua:
Any chance of another thumb on this?

szuecs:
@irishshagua there is another open comment

irishshagua:
> @irishshagua there is another open comment

Apologies, had missed that comment. Updated the docs now...

szuecs:
:+1:

docs/reference/filters.md
```diff
@@ -944,6 +944,26 @@ Example:
...
+## unverifiedAuditLog
+
+Filter `unverifiedAuditLog()` adds a Header, `X-Unverified-Audit`, to the request, the content of which, will also
+be written to the log file. By default, the value of the audit header will be equal to the value of the `sub` key, from
+the Authorization token. This can be changed by providing a `string` input to the filter which matches another key from the
+token.
+
+*N.B.* It is important to note that, if the content of the `X-Unverified-Audit` header does not match the following regex, then
+a default value of `invalid-sub` will be populated in the header instead:
+    `^[a-zA-z0-9_/:?=&%@.#-]*$`
```
whiskeysierra: What's the reasoning behind this restriction?

irishshagua: This was an existing restriction that was in place. Maybe @szuecs can shine some light.

szuecs: The commit message was:
```
commit 81013aefcfa07d4fac9b3c8f75d4752dbcb1f0a4 (tag: v0.10.2)
Author: Sandor Szücs <sandor.szuecs@zalando.de>
Date:   Thu May 3 12:01:01 2018 +0200

    fix prepared wrong garbage in unverified auditlog (#639)

    * fix prepared wrong garbage in unveirfied auditlog

    * add URI Sub test and change the defaultSub if we do not match Sub data
```

aryszka:
:+1:

PR #1016, opened 2019-03-29
use hostname from host header

### LINKED ISSUE #1000: Question: Open Id Connect Cookie Domain
Hi everyone. Thank you for working on and providing skipper!

I'm currently about to setup the pretty new Open Id Connect feature of skipper. The issue is that the hostname is used for setting the cookie-domain, see https://github.com/zalando/skipper/blob/master/filters/auth/oidc.go#L345 . Since my skipper sits behind another proxy this is no valid option in my case. 
Is there some configuration option I miss where I can override this or do I need to fork because of that? For example I think about having a flag on startup to overwrite would be great.
What do you think?

Pindar:
fixes #1000

### DIFF
--- a/filters/auth/oidc.go
+++ b/filters/auth/oidc.go
@@ -341,13 +341,23 @@ func extractDomainFromHost(host string) string {
 	return strings.Join(strings.Split(h, ".")[1:], ".")
 }
 
+func getHost(request *http.Request) string {
+	if h := request.Header.Get("host"); h != "" {
+		return h
+	} else {
+		return request.Host
+	}
+}
+
 func (f *tokenOidcFilter) doDownstreamRedirect(ctx filters.FilterContext, oidcState []byte, redirectUrl string) {
 	log.Debugf("Doing Downstream Redirect to :%s", redirectUrl)
+	host := getHost(ctx.Request())
+
 	r := &http.Response{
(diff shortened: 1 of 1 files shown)


szuecs:
@Pindar can you please sign-off your commits ?
You have to do it with git and it won't work in the browser.

Pindar:
Yes, I will do it later today. Sorry for the inconveniences.

Sandor Szücs <notifications@github.com> schrieb am Fr., 29. März 2019,
14:05:

Pindar:
@szuecs sign-off done and I extracted the implementation to a dedicated function.

filters/auth/oidc.go
```diff
@@ -341,13 +341,23 @@ func extractDomainFromHost(host string) string {
 	return strings.Join(strings.Split(h, ".")[1:], ".")
 }
 
+func getHost(request *http.Request) string {
+	if request.Header.Get("host") != "" {
```
szuecs: You can optimize the double lookup:
```
if h := request.Header.Get("host"); h != "" {
     return h
```

Pindar: thank you, I'm totally new to Go…

szuecs:
:+1:

aryszka:
:+1:

PR #1025, opened 2019-04-09
increase codecoverage to 80% required by silver cii bestpractices

szuecs:
see https://bestpractices.coreinfrastructure.org/ silver

- on parsing cli flags
- add some trivia

fix https://github.com/golang/go/issues/31293


szuecs:
:+1:

szuecs:
not sure how much is it now...

filters/auth/oidc_test.go
```diff
@@ -137,3 +139,103 @@ func TestExtractDomainFromHost(t *testing.T) {
...
+			name:    "test wrong number of args",
+			args:    []interface{}{"s", "d", "a"},
+			wantErr: true,
+		},
+		{
+			name:    "test wrong args",
+			args:    []interface{}{"s", "d", "a", "f"},
+			wantErr: true,
+		},
+		// cert error for provider
```
aryszka: is this a TODO comment?

Please use example domains, like example.org, because identity-provider.com and callback.com are existing, unrelated domain names.

szuecs: I would like to have a test case that also tests the happy path but I was not able to think about how to achieve it in this case. That's why it's commented.
Drop it?

aryszka: if I understand correctly, this would require a quite elaborate mocking. Let's drop it.

szuecs: done

szuecs:
:+1:

skptesting/enckey
```diff
@@ -0,0 +1 @@
+very secure
```
aryszka: where is this file used? Was it added by accident?

szuecs: ah this was used in the dropped test

szuecs: I pushed the test

aryszka:
:+1:

szuecs:
:+1:

PR #1027, opened 2019-04-10
Update authclient.go

justinrcs:
Don't send the auth header to the OAuth provider if the Access Token is not set


justinrcs:
Not familiar with Travis CI so I don't understand the failure message; let me know if there's something I need to check on my end.

szuecs:
@justinrcs thanks for your PR!
The travis pipeline says that you have to fix gofmt, which you can do with `make fmt`.

filters/auth/authclient.go
```diff
@@ -125,8 +125,11 @@ func jsonGet(
 	if err != nil {
 		return err
 	}
-	req.Header.Set(authHeaderName, authHeaderPrefix+accessToken)
-
+	
+	if len(accessToken > 0) {
```
szuecs: Please use this instead:
```
if accessToken != "" {
```

filters/auth/authclient.go
```diff
@@ -42,9 +43,9 @@ func newAuthClient(baseURL string, timeout time.Duration) (*authClient, error) {
 	return &authClient{url: u, client: client, quit: quit}, nil
 }
 
-func (ac *authClient) getTokenintrospect(token string, ctx filters.FilterContext) (tokenIntrospectionInfo, error) {
+func (ac *authClient) getTokenintrospect(token string, ctx filters.FilterContext, clientId string, clientSecret string) (tokenIntrospectionInfo, error) {
 	info := make(tokenIntrospectionInfo)
-	err := jsonPost(ac.url, token, &info, ac.client, ctx.Tracer(), ctx.ParentSpan(), tokenIntrospectionSpan)
+	err := jsonPost(ac.url, token, clientId, clientSecret, &info, ac.client, ctx.Tracer(), ctx.ParentSpan(), tokenIntrospectionSpan)
```
szuecs: please use `ac.url.User` to set the UserInfo, see https://golang.org/pkg/net/url/#URL and https://golang.org/pkg/net/url/#UserPassword, like this you do not need to pass 2 additional arguments around

filters/auth/authclient.go
```diff
@@ -159,6 +163,8 @@ func injectSpan(tracer opentracing.Tracer, parentSpan opentracing.Span, childSpa
 func jsonPost(
 	u *url.URL,
 	auth string,
+	clientId string,
+	clientSecret string,
```
szuecs: these parameters are not needed if you follow the suggestion above

justinrcs:
I added 4 additional TokenIntrospection filters:
secureOauthTokenintrospectionAnyClaims
secureOauthTokenintrospectionAllClaims
secureOauthTokenintrospectionAnyKV
secureOauthTokenintrospectionAllKV

They take a client-id and client-secret to authenticate with the token introspection provider.
If the values are sent in blank, it will try to use the client-id and client-secret set through env vars.

Updated the docs and tests

filters/auth/tokenintrospection.go
```diff
@@ -185,6 +241,12 @@ func (s *tokenIntrospectionSpec) CreateFilter(args []interface{}) (filters.Filte
 		issuerAuthClient[issuerURL] = ac
 	}
 
+	if s.secure && clientId != "" && clientSecret != "" {
+		ac.url.User = url.UserPassword(clientId, clientSecret)
+	} else {
+		ac.url.User = nil
```
szuecs: I think the default is nil already, so else part could be dropped.

filters/auth/tokenintrospection.go
```diff
@@ -327,6 +405,14 @@ func (f *tokenintrospectFilter) Request(ctx filters.FilterContext) {
...
 	case checkOAuthTokenintrospectionAllKV:
 		allowed = f.validateAllKV(info)
+	case checkSecureOAuthTokenintrospectionAnyClaims:
+		allowed = f.validateAnyClaims(info)
+	case checkSecureOAuthTokenintrospectionAnyKV:
+		allowed = f.validateAnyKV(info)
+	case checkSecureOAuthTokenintrospectionAllClaims:
+		allowed = f.validateAllClaims(info)
+	case checkSecureOAuthTokenintrospectionAllKV:
+		allowed = f.validateAllKV(info)
```
szuecs: please group the switch case block: `case checkOAuthTokenintrospectionAnyClaims, checkSecureOAuthTokenintrospectionAnyClaims: ..`

https://play.golang.org/p/Oh7ot21eWrH

szuecs:
@justinrcs great PR, only some minors left.
@aryszka please check it.

filters/auth/tokenintrospection.go
justinrcs: I added this to fix a caching issue. 

In testing, I found that if I created a filter with credentials:
eg. secureOauthTokenintrospectionAllKV("https://provider.url", "client-id", "client-secret", "key", "value")

and then subsequently updated the filter to use one without credentials:
oauthTokenintrospectionAllKV("https://provider.url", "key", "value")

if would cache the config and allow the request to go through even though the introspection endpoint requires authentication.

It was one of those use-cases that might not be very common but could lead to some unexplained behavior.

justinrcs:
Hi - the Travis CI is failing, but when I run the filter tests locally they pass. Any idea what could be wrong?

szuecs:
@justinrcs there is a race condition found:

```
--- FAIL: TestOAuth2Tokenintrospection (0.50s)
    --- FAIL: TestOAuth2Tokenintrospection/allKV():_valid_token,_one_valid_key_value_pair (0.01s)
        tokenintrospection_test.go:590: unexpected status code: 401 != 200
FAIL
FAIL	github.com/zalando/skipper/filters/auth	13.515s
/bin/sh: 1: break: Illegal number: -1
Makefile:68: recipe for target 'check-race' failed
```

justinrcs:
Thanks - looks like the race condition only occurs on Linux. On Mac build, the tests pass. 
I'll rebase and dig into see what's happening.

szuecs:
@justinrcs could be only happening on slow machines (CI runs with lowest costs possible, I bet).

szuecs:
:+1:

aryszka:
:+1:

PR #1031, opened 2019-04-12
enhance opentracing

szuecs:
add logs to possibly identify slow skipper client vs. slow backend
add tag client.request=canceled on 499, such that we could query this case


szuecs:
:+1:

lmineiro:
:+1:

Issue #1035, opened 2019-04-16
OAuth Scope Validation is returning an incorrect response code


irishshagua:
## Describe the bug
If you use any of the oauthTokeninfo* family of filters on a route and the token is successfully authenticated but does not pass the scope check, a 401 response code is returned. I think this should return a 403 to signify the difference between an authentication issue and an authorization issue.

### To Reproduce
An example 
**K8s Resources**:
```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: scoping-auth-example
  labels:
    application: scoping-auth-example
spec:
  replicas: 1
  selector:
    matchLabels:
      application: scoping-auth-example
  template:
    metadata:
      labels:
        application: scoping-auth-example
    spec:
      containers:
      - name: scoping-auth-example
        image: jmalloc/echo-server:latest
        ports:
        - containerPort: 8080

---
apiVersion: v1
kind: Service
metadata:
  name: scoping-auth-example
  labels:
    application: scoping-auth-example
spec:
  selector:
    application: scoping-auth-example
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 8080
  type: ClusterIP

---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    zalando.org/skipper-filter: oauthTokeninfoAllScope("uid", "service.scope")
    zalando.org/skipper-predicate: Path("/api/resource") && Method("GET")
  name: scoping-auth-example
spec:
  rules:
  - host: scoping-auth-example.example.com
    http:
      paths:
      - backend:
          serviceName: scoping-auth-example
          servicePort: http
``` 

**Sample Requests**
```
curl -i https://scoping-auth-example.example.com/api/resource
HTTP/2 401
date: Tue, 16 Apr 2019 10:41:34 GMT
content-length: 0
server: Skipper
www-authenticate: auth.example.com

curl -i -H "Authentication: Bearer $TOKEN"  https://scoping-auth-example.example.com/api/resource
HTTP/2 401
date: Tue, 16 Apr 2019 10:42:01 GMT
content-length: 0
server: Skipper
www-authenticate: auth.example.com
```

**Expected behavior**
I would expect a 401 for the first request and a 403 for the second request

**Observed behavior**
I get a 401 for both scenarios

szuecs:
RFC says nothing about not having the current behavior, but I agree that it might be interesting to the user in understanding the difference.

[...]

PR #1037, opened 2019-04-16
Differentiate authentication and authorization issues in response codes

irishshagua:
This addresses https://github.com/zalando/skipper/issues/1035

If the OAuthTokeninfo* family of filters reject a request
due to invalid scopes, then we will return a 403 response
code instead of the 401 response code to indicate that the
token was successfully authenticated but access to the resource
was rejected due to missing scopes in the token.


szuecs:
:+1:

szuecs:
:+1:

vetinari:
:+1:

PR #1030, opened 2019-04-11
lifo filters, concept

aryszka:
**[WIP]**

This PR integrates [this jobstack solution](https://github.com/aryszka/jobstack) to provide control over the maximum concurrency that skipper is allowed to handle, primarily to protect against chatty clients, and slow backends as noisy neighbors. It's a form of load shedding.

This PR is currently only a concept.

The plan is to support the following independent alternatives that allow:

- filters to define route specific LIFO stacks with custom configuration
- filters to assign a route to a LIFO group with a shared stack with globally defined configuration
- protect the entire process with a global, more permissive LIFO stack, as a last resort

In a second phase, shared counters across the nodes in a Skipper cluster will be also considered.

Signed-off-by: Arpad Ryszka <arpad.ryszka@gmail.com>

PR #1038, opened 2019-04-16
limit number of inflight requests per route

szuecs:
limit request per route, such that we do not get overrun by one slow backend

tested with 500Mi memory and 500m cpu, backend latency 5s, 25s and 50s, 500req/s to the mentioned backend
memory usage ~350Mi
The problem we saw with pprof is in the stdlib bufio.NewReaderSize and bufio.NewWriteterSize are spiking in memory usage


szuecs:
This will be closed, because we will integrate #1030 , which works similar, but is already more mature than the hack showing how to fix the problem.

PR #1042, opened 2019-04-26
Set URL from Host and TLS in dynamic backend

stepanbujnak:
When using dynamic backend, it is expected to use one of the
`setDynamicBackend*` filters to fill the host and scheme for the request
URL. However, there are cases in which we don't want to set it explicitly,
for example:

    foo: * -> compress("text/html") -> <dynamic>

Now if I do something like:

    curl -x http://skipper:9090 http://httpbin.org/anything

it complains about URL scheme and host not being set.

In this commit, the host and scheme part are automatically set for the
request URL before they're set from the statebag. This still allows the
user to modify host and scheme using the `setDynamicBackend*` filters
but if let unset, it will fallback to some sane defaults.

### DIFF
--- a/docs/reference/backends.md
+++ b/docs/reference/backends.md
@@ -238,6 +238,11 @@ $ curl -sv localhost:9090/ >/dev/null
 * Connection #0 to host localhost left intact
 ```
 
+When no filters modifying the target are set (e.g. `r0: * -> <dynamic>;`), the
+target host defaults to either the `Host` header or the host name given in the
+URL, and the target scheme defaults to either `https` when TLS is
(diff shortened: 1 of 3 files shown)


szuecs:
Thanks for your PR.
Please add tests and documentation for this change.

szuecs:
Code looks good to me, @Traumeel can you check if it could break your use case, I guess not, but better to check.

Traumeel:
@szuecs looks good to me as well.

szuecs:
:+1:

aryszka:
:+1:

PR #1046, opened 2019-05-07
add tag= parameter for lightstep plugin

vetinari:
this can be used to set tags present on any span, example

```
./bin/skipper -opentracing="lightstep tag=cluster=localhost token=<myToken> tag=foo=bar"
```

will set the tags `foo=bar` and `cluster=localhost` on any span

Signed-off-by: Hanno Hecker <hanno@zalando.de>

### DIFF
--- a/tracing/tracers/lightstep/lightstep.go
+++ b/tracing/tracers/lightstep/lightstep.go
@@ -22,6 +22,7 @@ func InitTracer(opts []string) (opentracing.Tracer, error) {
 	var cmdLine string
 	var logCmdLine bool
 	var maxBufferedSpans int
+	globalTags := make(map[string]string)
 
 	for _, o := range opts {
 		parts := strings.SplitN(o, "=", 2)
@@ -32,6 +33,13 @@ func InitTracer(opts []string) (opentracing.Tracer, error) {
 			}
 		case "token":
 			token = parts[1]
+		case "tag":
+			if len(parts) > 1 {
+				tags := strings.SplitN(parts[1], "=", 2)
+				if len(tags) == 2 {
+					globalTags[tags[0]] = tags[1]
+				}
+			}
 		case "collector":
 			var err error
 			var sport string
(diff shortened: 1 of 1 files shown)


szuecs:
@vetinari any specific reason why not having `tags=t1=v1,t2=v2,..` ?
I am fine with the current solution.

vetinari:
I think it's more readable in the parameter than having to check the many `=` in one value if there's a `,` in between or not

szuecs:
:+1:

tracing/tracers/lightstep/lightstep.go
```diff
@@ -32,6 +33,13 @@ func InitTracer(opts []string) (opentracing.Tracer, error) {
 			}
 		case "token":
 			token = parts[1]
+		case "tag":
+			if len(parts) > 1 {
+				tags := strings.SplitN(parts[1], "=", 2)
```
aryszka: currently we ignore those tags that don't have the form of `foo=bar`. I'd prefer:

a) fail early
b) consider such tags as boolean tags with the value of `true`

What do you think?

szuecs: a) is ok if we do not want to simply ignore the error as we do right now
b) can create more weirdness to the enduser, which I don't like. For example if you have complex tag settings and forgot one =foo then where ever in the chain the tags are specified this can create unexpected results

aryszka: between a) and c), I prefer a) because c) requires manual troubleshooting

vetinari: pushed a changed for a)

szuecs:
:+1:

szuecs:
:+1:

aryszka:
:+1:

tracing/tracers/lightstep/lightstep.go
aryszka: thx @vetinari

PR #1047, opened 2019-05-07
feat: provide the first stack trace on filter panic (#1045)

### LINKED ISSUE #1045: Provide stack trace if a filter panics
In case a filter instance panics, we only know which filter failed (name), but we don't know where (file, line number, method, etc.)

A log output containing stack trace. Logging stack traces for all requests can be expensive. A good starting point would be to log the first panic stack trace only.

In order to not pollute the logs, we can implement a HTTP endpoint which can provide run-time information about Skipper (configured routes, errors, stats, etc.). This solution is more complex and requires further discussions.

The `apiUsageMonitoringFilter` panics in one of our clusters. Without having this feature in place, we can only spot the problem by analyzing the code and/or trying to reproduce the error, which is less efficient.

Yes

maxim-tschumak:
Closes #1045 

Signed-off-by: Maxim Tschumak <maxim.tschumak@gmail.com>

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -12,6 +12,7 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"os"
+	"runtime"
 	"strconv"
 	"time"
 
@@ -632,10 +633,22 @@ func WithParams(p Params) *Proxy {
 	}
 }
 
-func tryCatch(p func(), onErr func(err interface{})) {
+var caughtPanic = false
+
+// tryCatch executes function `p` and `onErr` if `p` panics
+// onErr will receive a stack trace string of the first panic
+// further panics are ignored for efficiency reasons
+func tryCatch(p func(), onErr func(err interface{}, stack string)) {
 	defer func() {
 		if err := recover(); err != nil {
-			onErr(err)
+			s := ""
+			if !caughtPanic {
+				buf := make([]byte, 1024)
+				l := runtime.Stack(buf, false)
+				s = string(buf[:l])
+				caughtPanic = true
+			}
+			onErr(err, s)
 		}
 	}()
 
@@ -661,7 +674,7 @@ func (p *Proxy) applyFiltersToRequest(f []*routing.RouteFilter, ctx *context) []
 			ctx.setMetricsPrefix(fi.Name)
 			fi.Request(ctx)
 			p.metrics.MeasureFilterRequest(fi.Name, start)
-		}, func(err interface{}) {
+		}, func(err interface{}, stack string) {
 			if p.flags.Debug() {
 				// these errors are collected for the debug mode to be able
 				// to report in the response which filters failed.
@@ -669,7 +682,7 @@ func (p *Proxy) applyFiltersToRequest(f []*routing.RouteFilter, ctx *context) []
 				return
 			}
 
-			p.log.Errorf("error while processing filter during request: %s: %v", fi.Name, err)
+			p.log.Errorf("error while processing filter during request: %s: %v (%s)", fi.Name, err, stack)
 		})
 		filtersSpan.LogKV(fi.Name, "done")
 
@@ -695,7 +708,7 @@ func (p *Proxy) applyFiltersToResponse(filters []*routing.RouteFilter, ctx *cont
 			ctx.setMetricsPrefix(fi.Name)
 			fi.Response(ctx)
 			p.metrics.MeasureFilterResponse(fi.Name, start)
-		}, func(err interface{}) {
+		}, func(err interface{}, stack string) {
 			if p.flags.Debug() {
 				// these errors are collected for the debug mode to be able
 				// to report in the response which filters failed.
@@ -703,7 +716,7 @@ func (p *Proxy) applyFiltersToResponse(filters []*routing.RouteFilter, ctx *cont
 				return
 			}
 
-			p.log.Errorf("error while processing filters during response: %s: %v", fi.Name, err)
+			p.log.Errorf("error while processing filters during response: %s: %v (%s)", fi.Name, err, stack)
 		})
 	}
 
--- a/proxy/proxy_test.go
+++ b/proxy/proxy_test.go
@@ -1861,6 +1861,49 @@ func TestHopHeaderRemovalDisabled(t *testing.T) {
 	}
 }
 
+func thisOneWillPanic() {
+	panic("oops")
+}
+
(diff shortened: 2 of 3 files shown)


szuecs:
:+1:

tkrop:
:+1:

proxy/proxy.go
```diff
@@ -632,10 +633,22 @@ func WithParams(p Params) *Proxy {
...
+func tryCatch(p func(), onErr func(err interface{}, stack string)) {
 	defer func() {
 		if err := recover(); err != nil {
-			onErr(err)
+			s := ""
+			if !caughtPanic {
+				buf := make([]byte, 1024)
+				l := runtime.Stack(buf, false)
+				s = string(buf[:l])
+				caughtPanic = true
```
aryszka: Because of using a package level variable, I think there is a data race condition on the caughtPanic variable. It can kind crash the process. An equivalent solution without the data race would be using a buffered channel:

```
var caughtPanic = make(chan bool, 1)

// this will return false only on the first panic:
func hasCaughtPanic() bool {
    select {
    case caughtPanic <- true:
        return false
    default:
        return true
    }
}
```

As a possible improvement, the logging could be even nicer, if we used a synced map, where we store max N hashes of the error message and the panic stack trace.

maxim-tschumak: This is a very important point. My understanding is the following:

We don't want to synchronize on every request - this is just too expensive for Skipper.

Regarding the panic: Concurrent writes/read to a `map` can (and will) panic since the underlying structure is dynamic. Concurrent writes/reads to a `bool` do not panic: it is a single value, which is being copied for different processes / cached in different CPU caches. In the worst case a copy of the value is not up-to-date which causes multiple outputs of the whole stack trace - this should be acceptable.

Here is also a test which I used to be a bit more sure about my reasoning (it's not a proof):

```go
func Test_ConcurrentReadsAndWritesToBool(t *testing.T) {
	var wg sync.WaitGroup
	wg.Add(20000000)

	b := false
	counter:=0
	for i := 0; i < 10000000; i++ {
		func () {
			b = counter % 2 == 0
			counter++
			wg.Done()
		}()
		func () {
			b = counter % 2 == 1
			counter++
			wg.Done()
		}()
	}

	wg.Wait()
	fmt.Println(b)
	fmt.Println(counter)
}
```

Please correct me if I'm wrong.

aryszka: I also wrote a small test to check, and I couldn't create a data race:

```
package main

func main() {
	var boolean bool
	flip := func(n int) {
		for {
			boolean = !boolean
		}
	}

	for i := 0; i < 100; i++ {
		go flip(i)
	}

	select {}
}
```

So while this works fine, I am still somewhat concerned, because this reasoning considers implementation details which are not specified by the language: "Concurrent writes/reads to a bool do not panic: it is a single value, which is being copied for different processes / cached in different CPU caches.". I may be completely wrong, though, and I don't know under what condition Go panics on data races.

OTOH: what I proposed, wouldn't synchronize on every request, only for those when there is a panic from a filter, which should not happen in normal operation.

Thx, reflecting on this, I leave the decision to you :)

aryszka:
:+1:

proxy/proxy.go
maxim-tschumak: > So while this works fine, I am still somewhat concerned, because this reasoning considers implementation details which are not specified by the language

I think they don't specify the behaviour because they cannot guarantee that the updated value will be propagated to all the processes. My _assumption_ is that this kind of operations cannot panic. @tkrop once shared a great article about processor implementations details regarding this topic.

> OTOH: what I proposed, wouldn't synchronize on every request, only for those when there is a panic from a filter, which should not happen in normal operation.

You are right, it's done only for panicking requests. In our case it were **all** requests on a specific route though..

I would go with proposed solution if nobody disagrees.

lmineiro: Have we considered `atomic` operations?

lmineiro: Not that I'm recommending the particular implementation, but this one has simple benchmarks: https://github.com/tevino/abool

maxim-tschumak: @lmineiro Unfortunately, I don't have any experience with this package. The numbers look great in comparison with the `Mutex` implementation, but it's definitely slower than any implementation with no sync at all.

I would say, the owners have to decide (if we want to sync or not, and if yes - how) :) @aryszka @szuecs I'm happy to implement the change.

tkrop: @aryszka @lmineiro Using an atomic operations would be surely a suitable alternative and from semantic perspective preferable alternative as it makes clear how race conditions are handled. However, with respect to the goal the provided solution is absolute fine and most efficient I would go for here.

Note: the only situation where this race condition could result in a `panic` is on a `numa` architecture with a cache consistency protocol that exposes conflicts instead of resolving it in the micro code. I'm not aware of any processor having ever implemented it this way and would count this as a processor bug.

tkrop:
:+1:

proxy/proxy.go
aryszka: so to sum up, our options:

a) atomic var (in this case please rely on the suggested library only as a guide, but don't introduce it as a new dependency)
b) a mutex next to the var
c) a channel
d) just merge it as is

I understand now that d) is safe, thanks for the explanations. I don't have any strong preference but the order looks like this for me:

c > a > b > d

@maxim-tschumak please decide :smiley:

aryszka:
:+1:

proxy/proxy.go
maxim-tschumak: we want to catch the next (release) train :)

proxy/proxy_test.go
```diff
@@ -1861,6 +1861,49 @@ func TestHopHeaderRemovalDisabled(t *testing.T) {
...
+			t.Error("should provide function names and file names in the stack trace")
+			return
+		}
+	})
+}
+
+func TestTryCatchProvidesStackTraceOnlyOnce(t *testing.T) {
+	caughtPanic = false
+	tryCatch(thisOneWillPanic, func(err interface{}, stack string) {
+		fmt.Printf("HERE: %v %s", err, stack)
```
szuecs: Remove your logs, prints from the tests that should be most of the time leftovers.

proxy/proxy.go
szuecs: I am fine with any of it but channels seem to be overkill. From my side a bool and rely on the processor is fine.

PR #1054, opened 2019-05-14
Fix/lifo self healing

szuecs:
fix: if lifo filter response is not called, the proxy should execute done() to release the queue


aryszka:
:+1:

szuecs:
:+1:

PR #1061, opened 2019-05-21
Log important events emitted by LightStep tracer

sarslanhan:
The recommended way to log information regarding issues is to use `SetGlobalEventHandler` (see [this](https://github.com/lightstep/lightstep-tracer-go/blob/master/options.go#L165) and the default logger logs only the first error event (see [this](https://github.com/lightstep/lightstep-tracer-go/blob/master/event_handlers.go#L72)).

I actually wanted to inject `logging.Logger`on [`InitTracer`](https://github.com/zalando/skipper/blob/master/tracing/tracing.go#L68). However, this would break the behavior described [here](https://github.com/zalando/skipper/blob/e20191f29c77e0eb8487356545f65d699ae75bf6/docs/reference/plugins.md#opentracing-plugins). Let me know you prefer to do this another way.

### DIFF
--- a/tracing/tracers/lightstep/lightstep.go
+++ b/tracing/tracers/lightstep/lightstep.go
@@ -3,6 +3,7 @@ package lightstep
 import (
 	"errors"
 	"fmt"
+	"github.com/zalando/skipper/logging"
 	"net"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


tracing/tracers/lightstep/lightstep.go
```diff
@@ -3,6 +3,7 @@ package lightstep
 import (
 	"errors"
 	"fmt"
+	"github.com/zalando/skipper/logging"
```
szuecs: I think you should use plain github.com/sirupsen/logrus, instead.

sarslanhan: Sure, I was not sure which one as it is used in a mixed way in different parts of the code. Will send an update.

szuecs:
:+1:

aryszka:
:+1:

PR #1064, opened 2019-05-22
Drop `skipper.route` tag to reduce the size of the messages shipped

sarslanhan:
Currently, skipper creates spans that serializes to ~2.5 KB of payload in protobuf format based on my experiments. The main contributors to this are `skipper.route` tag, logs for marking filter start & done events, and the response stream events.

We can reduce the size by dropping these attributes however we may lose debugging abilities. To get around this, I propose we add feature flags under `OpenTracingParams` that would be passed to `Proxy`. In this way, we can enable logging these attributes when it is needed and keep it turned off by
default.

For now, we can drop `skipper.route` tag to reduce the size of the messages shipped for OpenTracing client.

In the future, we can have feature toggles to have the logs & additional tags enabled/disabled when necessary.

Relates to #1065 

Signed-off-by: Serbay Arslanhan <serbay.arslanhan@zalando.de>


szuecs:
:+1:

sarslanhan:
👍

sarslanhan:
👎

sarslanhan:
👍

Issue #1065, opened 2019-05-22
Reduce the payload shipped per request for OpenTracing instrumentation


sarslanhan:
**Is your feature request related to a problem? Please describe.**
Currently, skipper creates spans that serializes to ~2.5 KB of payload in protobuf format based on my experiments (I took an example request and serialized it into protobuf to see the size). It looks like the main contributors to this are skipper.route tag, logs for marking filter start & done events, and the response stream events.

**Describe the solution you would like**
We can reduce the size by dropping these attributes however we may lose debugging abilities. To get around this, I propose we add feature flags under OpenTracingParams that would be passed to Proxy. In this way, we can enable logging these attributes when it is needed and keep it turned off by default.

**Would you like to work on it?**
Sure.

PR #1066, opened 2019-05-23
Add options to reduce the payload shipped per request

sarslanhan:
Overall the PR adds 2 CLI options that are used in `Proxy`: `opentracing-excluded-proxy-tags` & `opentracing-log-filter-lifecycle-events`.

Some notes on additional changes:
- As a part of the PR I also attempted to unify how we set tags on `Proxy` level. 
- We could also extend functionality offered by `proxyTracing` by moving `tracing.CreateSpan` there as well as it is not used anywhere outside of `Proxy`.

Relates to #1065

### DIFF
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -98,34 +98,36 @@ const (
 	maxAuditBodyUsage                    = "sets the max body to read to log in the audit log body"
 
 	// logging, metrics, tracing:
-	enablePrometheusMetricsUsage    = "switch to Prometheus metrics format to expose metrics. *Deprecated*: use metrics-flavour"
-	opentracingUsage                = "list of arguments for opentracing (space separated), first argument is the tracer implementation"
-	opentracingIngressSpanNameUsage = "set the name of the initial, pre-routing, tracing span"
-	metricsListenerUsage            = "network address used for exposing the /metrics endpoint. An empty value disables metrics iff support listener is also empty."
-	metricsPrefixUsage              = "allows setting a custom path prefix for metrics export"
-	enableProfileUsage              = "enable profile information on the metrics endpoint with path /pprof"
-	debugGcMetricsUsage             = "enables reporting of the Go garbage collector statistics exported in debug.GCStats"
-	runtimeMetricsUsage             = "enables reporting of the Go runtime statistics exported in runtime and specifically runtime.MemStats"
-	serveRouteMetricsUsage          = "enables reporting total serve time metrics for each route"
-	serveHostMetricsUsage           = "enables reporting total serve time metrics for each host"
-	backendHostMetricsUsage         = "enables reporting total serve time metrics for each backend"
-	allFiltersMetricsUsage          = "enables reporting combined filter metrics for each route"
-	combinedResponseMetricsUsage    = "enables reporting combined response time metrics"
-	routeResponseMetricsUsage       = "enables reporting response time metrics for each route"
-	routeBackendErrorCountersUsage  = "enables counting backend errors for each route"
-	routeStreamErrorCountersUsage   = "enables counting streaming errors for each route"
-	routeBackendMetricsUsage        = "enables reporting backend response time metrics for each route"
-	metricsUseExpDecaySampleUsage   = "use exponentially decaying sample in metrics"
-	histogramMetricBucketsUsage     = "use custom buckets for prometheus histograms, must be a comma-separated list of numbers"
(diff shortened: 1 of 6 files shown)


cmd/skipper/main.go
```diff
@@ -406,6 +410,8 @@ func init() {
 	flag.BoolVar(&enablePrometheusMetrics, "enable-prometheus-metrics", false, enablePrometheusMetricsUsage)
 	flag.StringVar(&openTracing, "opentracing", "noop", opentracingUsage)
 	flag.StringVar(&openTracingInitialSpan, "opentracing-initial-span", "ingress", opentracingIngressSpanNameUsage)
+	flag.StringVar(&openTracingExcludedProxyTags, "opentracing-excluded-proxy-tags", "", openTracingExcludedProxyTagsUsage)
+	flag.BoolVar(&opentracingLogFilterLifecycleEvents, "opentracing-log-filter-lifecycle-events", true, opentracingLogFilterLifecycleEventsUsage)
```
szuecs: I think this should be false as default

proxy/proxytest/proxytest.go
```diff
@@ -45,9 +44,9 @@ func newTestProxy(fr filters.Registry, routingOptions routing.Options, proxyPara
 	rt := routing.New(routingOptions)
 
 	proxyParams.Routing = rt
-	if proxyParams.OpenTracer == nil {
-		proxyParams.OpenTracer = &opentracing.NoopTracer{}
-	}
+	//if proxyParams.OpenTracing == nil {
+	//	proxyParams.OpenTracing = &proxy.OpenTracingParams{ Tracer: &opentracing.NoopTracer{} }
+	//}
```
szuecs: please clean up

proxy/proxy.go
```diff
@@ -196,6 +206,8 @@ type Params struct {
 
 	// Client TLS to connect to Backends
 	ClientTLS *tls.Config
+
+	OpenTracing *OpenTracingParams
```
szuecs: should have godoc comment

proxy/proxy.go
```diff
@@ -702,11 +703,14 @@ func (p *Proxy) applyFiltersToRequest(f []*routing.RouteFilter, ctx *context) []
 // applies filters to a response in reverse order
 func (p *Proxy) applyFiltersToResponse(filters []*routing.RouteFilter, ctx *context) {
 	filtersStart := time.Now()
+	filtersSpan := tracing.CreateSpan("response_filters", ctx.request.Context(), p.tracing.tracer)
```
szuecs: :+1: thanks for adding this

cmd/skipper/main.go
sarslanhan: I thought about it as well, but decided to set to `true` to keep the current behavior.

szuecs: As discussed default true is fine

proxy/tracing.go
```diff
@@ -0,0 +1,111 @@
...
+	span.LogKV(eventName, eventValue)
+}
+
+func (t *proxyTracing) setTag(span ot.Span, key string, value interface{}) *proxyTracing {
+	if span == nil {
+		return t
+	}
+
+	_, excluded := t.excludeTags[key]
+	if !excluded {
```
aryszka: this could be simply:

```
if !t.excludeTags[key] {
```

because the map returns the zero value of the map value's type when the key doesn't exist, and the zero value of bool is false.

aryszka:
:+1:

proxy/tracing.go
```diff
@@ -0,0 +1,111 @@
...
+	SkipperRouteIDTag     = "skipper.route_id"
+	SpanKindTag           = "span.kind"
+
+	ClientRequestCanceled = "canceled"
+	SpanKindClient        = "client"
+	SpanKindServer        = "server"
+
+	EndEvent           = "start"
+	StartEvent         = "start"
+	StreamHeadersEvent = "stream_Headers"
```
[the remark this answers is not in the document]
sarslanhan: I tried to keep everything the same as before but extend the control over what was shipped.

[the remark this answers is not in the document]
sarslanhan: But regarding the naming, you are definitely right that there is room for improvement. Operation names can also be changed to be inline with [OpenTracing spec](https://github.com/opentracing/specification/blob/master/specification.md#start-a-new-span)

proxy/tracing.go
sarslanhan: Updated.

aryszka:
:+1:

szuecs:
:+1:

PR #1070, opened 2019-05-27
Enrich forwardToken filter

irishshagua:
* Add ability to mask keys in the JSON payload for the forwardToken filter
  * Update docs in relation to forwardToken filter enhancements
  * Add tests covering new functionality

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -687,10 +687,9 @@ secureOauthTokenintrospectionAllKV("issuerURL", "", "", "k1", "v1", "k2", "v2")
 
 ## forwardToken
 
-The filter accepts a variable number of string arguments but there must be at least 1 input which represents the header
-name where the result of token info or token introspection is added when the request is passed to the backend. Input
-arguments after the first, are considered to be keys within the JSON that you want to be excluded from forwarding to
-the backend service.
+The filter takes the (string) header name as its first argument. The result of token info or token introspection is added to 
+this header when the request is passed to the backend. Any subsequent arguments are keys within the JSON to be excluded from the 
+header payload when forwarding to the backend service.
 
 If this filter is used when there is no token introspection or token info data
 then it does not have any effect.
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -688,7 +688,8 @@ secureOauthTokenintrospectionAllKV("issuerURL", "", "", "k1", "v1", "k2", "v2")
 ## forwardToken
 
 The filter takes the (string) header name as its first argument. The result of token info or token introspection is added to 
-this header when the request is passed to the backend. Any subsequent arguments are keys within the JSON to be excluded from the 
+this header when the request is passed to the backend. If there are additional arguments, these
+values are treated as a whitelisted of JSON keys to be included in the 
 header payload when forwarding to the backend service.
 
 If this filter is used when there is no token introspection or token info data
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -687,9 +687,10 @@ secureOauthTokenintrospectionAllKV("issuerURL", "", "", "k1", "v1", "k2", "v2")
 
 ## forwardToken
 
-The filter accepts a single string as an argument. The argument is the header
-name where the result of token info or token introspection is added when the
-request is passed to the backend.
+The filter accepts a variable number of string arguments but there must be at least 1 input which represents the header
+name where the result of token info or token introspection is added when the request is passed to the backend. Input
+arguments after the first, are considered to be keys within the JSON that you want to be excluded from forwarding to
+the backend service.
 
 If this filter is used when there is no token introspection or token info data
 then it does not have any effect.
@@ -698,6 +699,7 @@ Examples:
 
 ```
 forwardToken("X-Tokeninfo-Forward")
+forwardToken("X-Tokeninfo-Forward", "access_token")
 ```
 
 ## oauthOIDCUserInfo
--- a/filters/auth/forwardtoken.go
+++ b/filters/auth/forwardtoken.go
@@ -19,8 +19,8 @@ type (
 	forwardTokenSpec struct {
 	}
 	forwardTokenFilter struct {
-		HeaderName    string
-		StripJsonKeys []string
+		HeaderName     string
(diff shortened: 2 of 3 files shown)


docs/reference/filters.md
```diff
@@ -687,9 +687,10 @@ secureOauthTokenintrospectionAllKV("issuerURL", "", "", "k1", "v1", "k2", "v2")
 
 ## forwardToken
 
-The filter accepts a single string as an argument. The argument is the header
-name where the result of token info or token introspection is added when the
-request is passed to the backend.
+The filter accepts a variable number of string arguments but there must be at least 1 input which represents the header
+name where the result of token info or token introspection is added when the request is passed to the backend. Input
+arguments after the first, are considered to be keys within the JSON that you want to be excluded from forwarding to
+the backend service.
```
cjbooms: Maybe reword to:
```
The filter takes the (string) header name as its first argument. The result of token info or token introspection is added to this header when the request is passed to the backend. Any subsequent arguments are keys within the JSON to be excluded from the header payload when forwarding to the backend service.
```

irishshagua:
@aryszka / @szuecs 
Maybe one of you can help here? I made a relatively small change to one of the filters but it seems to have some knock-on effect to some unrelated tests. I can't reproduce these fails locally. Is it possible these could be false-negatives? Or is there some issue introduced with this code change?

szuecs:
@irishshagua for whatever reason travis tests are very flaky and get worse now over time.

filters/auth/forwardtoken.go
```diff
@@ -73,3 +98,20 @@ func (f *forwardTokenFilter) Request(ctx filters.FilterContext) {
...
+func stringInSlice(a string, list []string) bool {
+	for _, b := range list {
+		if b == a {
+			return true
+		}
+	}
+	return false
+}
+
+func removeKeys(data map[string]interface{}, keys []string) {
```
szuecs: this can be simplified: https://play.golang.org/p/bvrzCWpiBHI

filters/auth/forwardtoken.go
```diff
@@ -62,6 +76,17 @@ func (f *forwardTokenFilter) Request(ctx filters.FilterContext) {
 		return
 	}
 
+	if len(f.StripJsonKeys) > 0 {
+		switch typedTiMap := tiMap.(type) {
+		case map[string]interface{}:
+			removeKeys(typedTiMap, f.StripJsonKeys)
```
szuecs: `fallthrough` seems to be the better option here

filters/auth/forwardtoken_test.go
```diff
@@ -169,6 +198,7 @@ func TestForwardTokenIntrospection(t *testing.T) {
 			clientServer := httptest.NewServer(http.HandlerFunc(func(_ http.ResponseWriter, r *http.Request) {
 				if ti.oauthFilterPresent {
 					tokenInfo := r.Header.Get(ti.headerName)
+					fmt.Printf("Header Value is %s\n", tokenInfo)
```
szuecs: this seems to be debug logging, so please drop

szuecs:
@irishshagua small changes necessary, but thanks for the enhancement. :)

filters/auth/forwardtoken.go
irishshagua: compiler is complaining that `fallthrough` isn't supported for type switch.

szuecs: ah ok, my parser was not good enough :)

szuecs:
:+1:

szuecs:
:+1:

tkrop:
@szuecs I think it would be better to support a list of fields that should be forwarded instead of a list that should be dropped.

tkrop:
:+1:

szuecs:
@irishshagua do you want to change the PR to do whitelisting instead of blacklisting, as @tkrop suggested?
I think it might makes sense from security point of view, if you think it's a security feature.

irishshagua:
Hey @szuecs and @tkrop 
My only conern about changing to a whitelist as opposed to a blacklist is that it would be a backward breaking change. The logic would be a little funky if no additional params means forward all fields and some params means only forward the listed fields. Not a major issue though, I can clearly document if this is the desired behaviour. What ye think?

szuecs:
I think it’s fine to have:
- no Params -> send all
- params -> only send these

Because no filter is the: send nothing case, not?

irishshagua:
> I think it’s fine to have:
> 
>     * no Params -> send all
> 
>     * params -> only send these
> 
> 
> Because no filter is the: send nothing case, not?

Ok, canged from blacklist to whitelist... cheers for the feedback.  :)

filters/auth/forwardtoken.go
```diff
@@ -62,6 +76,17 @@ func (f *forwardTokenFilter) Request(ctx filters.FilterContext) {
...
 	}
 
+	if len(f.RetainJsonKeys) > 0 {
+		switch typedTiMap := tiMap.(type) {
+		case map[string]interface{}:
+			tiMap = retainKeys(typedTiMap, f.RetainJsonKeys)
+		case tokenIntrospectionInfo:
+			tiMap = retainKeys(typedTiMap, f.RetainJsonKeys)
+		default:
+			log.Errorf("Unexpected input type[%v] for `forwardToken` filter. Unable to apply mask", reflect.TypeOf(typedTiMap))
```
szuecs: I think instead of `reflect.TypeOf` you can just use `%T` in the format string.

docs/reference/filters.md
```diff
@@ -687,9 +687,10 @@ secureOauthTokenintrospectionAllKV("issuerURL", "", "", "k1", "v1", "k2", "v2")
 
 ## forwardToken
 
-The filter accepts a single string as an argument. The argument is the header
-name where the result of token info or token introspection is added when the
-request is passed to the backend.
+The filter takes the (string) header name as its first argument. The result of token info or token introspection is added to 
+this header when the request is passed to the backend. If there are additional arguments, these
+values are treated as a whitelisted of JSON keys to be included in the
```
aryszka: "...whitelisted set of JSON keys..."?

aryszka:
:+1:

szuecs:
:+1:

PR #1078, opened 2019-05-31
fix source from last predicate

szuecs:
fix source from last predicate was based on wrong assumptions, we do not get parsed headers into string slice, but a plain string, so use Header.Get and work on the string ourselves

reported by a user

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -50,16 +50,14 @@ func RemoteHost(r *http.Request) net.IP {
 //
 //     X-Forwarded-For: ip-address-1, ip-address-2, client-ip-address
 func RemoteHostFromLast(r *http.Request) net.IP {
-	a := r.Header["X-Forwarded-For"]
-	if a == nil {
-		return parse(r.RemoteAddr)
+	ffs := r.Header.Get("X-Forwarded-For")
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

net/net.go
```diff
@@ -50,16 +50,14 @@ func RemoteHost(r *http.Request) net.IP {
...
-	l := len(a) - 1
-	if l < 0 {
-		l = 0
-	}
-	if ffh := parse(a[l]); ffh != nil {
-		return ffh
+	ffs := r.Header.Get("X-Forwarded-For")
+	ffa := strings.Split(ffs, ",")
+	ff := ffa[len(ffa)-1]
+	if ff != "" {
```
aryszka: i think it would be better to trim the space before checking for empty. What do you think?

aryszka:
:+1:

net/net.go
szuecs: We need to trim the words as far as I tested these have `, `

szuecs:
:+1:

szuecs:
:+1:

szuecs:
:+1:

aryszka:
:+1:

PR #1088, opened 2019-06-10
Fix/rfc patch path

### LINKED ISSUE #1083: path predicates don't handle URL escaped paths
**Describe the bug**
When a URL escaped path is used in a path predicate, the predicate doesn't match valid requests.

**To Reproduce**
Use the following route:

```
Path("/api/something%2A") -> status(200) -> <shunt>
```

**Expected behavior**

It should return 200 for the requests with the path: /api/something%2A.

**Observed behavior**

The route doesn't match.

aryszka:
Closes #1083 

- provide a filter for preserving the reserved characters escaped in the path if the incoming request had them escaped.
- provide a command line flag to enforce this behavior centrally for a Skipper instance


docs/reference/filters.md
```diff
@@ -1450,3 +1450,35 @@ lifoGroup("mygroup", 100, 150, "10s")
...
+
+In case we want to use the id while routing the request, we can use the <loopback>
+backend. Example:
+
+```
+api: Path("/api/:id") -> setPath("/api/${id}/summary") -> "http://api-backend";
+patch: Path("/api/*id") -> rfcPath() -> <loopback>;
+```
+
+In the above case, if the incoming request path is /api/foo%2Fbar, it will match
```
szuecs: Why this will match the `patch:` route?

rfc/patchpath.go
```diff
@@ -0,0 +1,130 @@
...
+// PatchPath attempts to patch a request path based on an interpretation of the standards
+// RFC 2616 and RFC 2396 where the reserved characters should not be unescaped. Currently
+// the Go stdlib does unescape these characters (v1.12.5).
+//
+// It expects the parsed path as found in http.Request.URL.Path and the raw path as found
+// in http.Request.URL.RawPath. It returns a path where characters e.g. like '/' have the
+// escaped form of %2F, if it was detected that they are unescaped in the raw path.
+//
+// It only returns the patched variant, if the only difference between the parsed and raw
+// paths are the encoding of the chars, according to RFC2 396. If it detects any other
```
szuecs: RFC2 396 -> RFC 2396

proxy/pathpatch_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+		if rsp.StatusCode != expectedStatus {
+			t.Error()
+		}
+	})
+}
+
+func TestPathPatch(t *testing.T) {
+	testPatch(t, "not patched", FlagsNone, http.StatusNotFound)
+	testPatch(t, "patched", PatchPath, http.StatusOK)
```
szuecs: Can you also test a utf-8 character, which should be allowed in the path?
Because you work on []byte slices later and I want to make sure that this will work.

rfc/patchpath_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}, {
+		title:    "reserved and unreserved",
+		raw:      "/foo%2Fbar/%2A",
+		parsed:   "/foo/bar/*",
+		expected: "/foo%2Fbar/*",
+	}, {
+		title:    "unreserved and reserved",
+		raw:      "/foo%2Abar%2F",
+		parsed:   "/foo*bar/",
+		expected: "/foo*bar%2F",
```
szuecs: one other test case with utf-8 character please :)

docs/reference/filters.md
aryszka: because first it's parsed as /api/foo/bar, and it's matched by /api/*id, where $id will contain /foo/bar (or foo/bar, not sure currently).

proxy/pathpatch_test.go
aryszka: Sure, I can. Why it will work, though, is that we only compare bytes in the ASCII range ('%', and '%xx'), and we exit on the first difference.

aryszka: just tried with /世%2F界. The rfc.PatchPath() function works as expected, but the test in the proxy doesn't. The reason is that in this case the http.Request.URL.RawPath remains empty. I'll investigate why.

rfc/patchpath.go
szuecs: RFC 2396 was updated to https://tools.ietf.org/html/rfc3986

proxy/pathpatch_test.go
szuecs: We agreed on ignoring it on the proxy and drop the suggested test, but have it tested in rfc filter package.

aryszka:
Thanks for the review!

aryszka:
:+1:

szuecs:
:+1:

PR #1095, opened 2019-06-18
fix: use path from request chain (#1094)

### LINKED ISSUE #1094: apiusagemonitoring is using modified path from response chain
**Describe the bug**
When monitoring APIs using the API usage monitoring filter, the filter is using the modified path from the response chain instead of the original path from the begin of the request chain - assuming that the API usage monitoring filter is the first installed in the filter chain.

**To Reproduce**
Add context, for example:
1. create an `apiUsageMonitoring` filter followed by a `modPath` filter changing the path.
2. execute a request on the original path
3. look for the metrics collected by the monitoring filter, which will match against the modified path.

**Expected behavior**
Collect a monitoring event for the original path before filter modification.

**Observed behavior**
Collected a monitoring event for the modified path after filter modification.

tkrop:
fixes #1094.

### DIFF
--- a/filters/apiusagemonitoring/filter.go
+++ b/filters/apiusagemonitoring/filter.go
@@ -25,8 +25,7 @@ const (
 )
 
 const (
-	stateBagKeyPrefix = "filter." + Name + "."
-	stateBagKeyBegin  = stateBagKeyPrefix + "begin"
+	stateBagKey = "filter." + Name
 )
 
 const (
@@ -42,14 +41,27 @@ type apiUsageMonitoringFilter struct {
 	UnknownPath *pathInfo
 }
 
+type apiUsageMonitoringStateBag struct {
+	URL   *url.URL
+	Begin time.Time
+}
+
 func (f *apiUsageMonitoringFilter) Request(c filters.FilterContext) {
-	c.StateBag()[stateBagKeyBegin] = time.Now()
+	URL := *c.Request().URL
+	c.StateBag()[stateBagKey] = apiUsageMonitoringStateBag{
+		URL:   &URL,
+		Begin: time.Now(),
+	}
 }
 
 func (f *apiUsageMonitoringFilter) Response(c filters.FilterContext) {
 	request, response, metrics := c.Request(), c.Response(), c.Metrics()
-	begin, beginPresent := c.StateBag()[stateBagKeyBegin].(time.Time)
-	path := f.resolveMatchedPath(request.URL)
+	stateBag, stateBagPresent := c.StateBag()[stateBagKey].(apiUsageMonitoringStateBag)
+	URL := request.URL
+	if stateBagPresent && stateBag.URL != nil {
+		URL = stateBag.URL
+	}
+	path := f.resolveMatchedPath(URL)
 
 	classMetricsIndex := response.StatusCode / 100
 	if classMetricsIndex < 1 || classMetricsIndex > 5 {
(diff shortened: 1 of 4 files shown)


zeitlinger:
:+1:

aryszka:
:+1:

aryszka:
@tkrop please run `go fmt ./...` and also sign off the commits :) (https://github.com/zalando/skipper/pull/1095/checks?check_run_id=150714005)

aryszka:
If I see it right, then this PR allows the right level of flexibility: https://github.com/zalando/skipper/issues/1094

if the path is modified before the usage monitoring filter, then the modified path is monitored.

szuecs:
@tkrop please sign all commits

filters/apiusagemonitoring/filter.go
```diff
@@ -42,14 +41,27 @@ type apiUsageMonitoringFilter struct {
...
 }
 
+type apiUsageMonitoringStateBag struct {
+	URL   *url.URL
+	Begin time.Time
+}
+
 func (f *apiUsageMonitoringFilter) Request(c filters.FilterContext) {
-	c.StateBag()[stateBagKeyBegin] = time.Now()
+	URL := *c.Request().URL
```
szuecs: please use lower case `u` as local var

filters/apiusagemonitoring/filter.go
```diff
@@ -42,14 +41,27 @@ type apiUsageMonitoringFilter struct {
...
+		Begin: time.Now(),
+	}
 }
 
 func (f *apiUsageMonitoringFilter) Response(c filters.FilterContext) {
 	request, response, metrics := c.Request(), c.Response(), c.Metrics()
-	begin, beginPresent := c.StateBag()[stateBagKeyBegin].(time.Time)
-	path := f.resolveMatchedPath(request.URL)
+	stateBag, stateBagPresent := c.StateBag()[stateBagKey].(apiUsageMonitoringStateBag)
+	URL := request.URL
```
szuecs: please use lower case `u` as local var

tkrop:
Changed the logic to use the response path as fallback accepting the performance drawback for unmatched paths after discussion with @maxim-tschumak.

filters/apiusagemonitoring/filter.go
```diff
@@ -42,14 +41,29 @@ type apiUsageMonitoringFilter struct {
...
 }
 
+type apiUsageMonitoringStateBag struct {
+	URL   *url.URL
+	Begin time.Time
+}
+
 func (f *apiUsageMonitoringFilter) Request(c filters.FilterContext) {
-	c.StateBag()[stateBagKeyBegin] = time.Now()
+	uRL := *c.Request().URL
```
szuecs: please use `u` instead of `uRL`, thanks

filters/apiusagemonitoring/filter.go
```diff
@@ -42,14 +41,29 @@ type apiUsageMonitoringFilter struct {
 	UnknownPath *pathInfo
 }
 
+type apiUsageMonitoringStateBag struct {
+	URL   *url.URL
+	Begin time.Time
```
szuecs: Do you need to expose URL and Begin?

tkrop: no :+1:

szuecs:
:+1:

tkrop:
:+1:

PR #1097, opened 2019-06-20
Add roundtrip start/end span logs

mikkeloscar:
This adds span logs for when roundtrip to backend is started and ended. This should enable us to better understand when a backend was called and how long it took to get the response.

Without this log we only see `dial_context` logs in some cases where a new connection has to be established but often we don't see it when a connection is re-used meaning we only see the `stream_Headers` logs when we start to stream the response from the backend back to the client calling skipper.

Example of how this looks:

image

Here the backend has a latency of `50ms`


sarslanhan:
👍

szuecs:
:+1:

Issue #549, opened 2018-02-07
Kubernetes service ExternalName should be possible to target from ingress


szuecs:
As a user, I want to specify an external proxy target within the ingress object.

In Kubernetes this could be done with [service type ExternalName](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services---service-types), but the skipper kubernetes dataclient does not support this, yet.

daviddyball:
Hey @szuecs, did you ever find a way around this? We're running Skipper and I want to do exactly this, but I'm getting `error during backend roundtrip: kube xxxxxxxxxxxxxxxx service: dial tcp :443: getsockopt: connection refused`

daviddyball:
Ahhh right, this bypasses my need to use `ExternalName` completely 😛 

Currently I have this (with your additions):

```
apiVersion: v1
kind: Service
metadata:
  name: logs
spec:
  type: ExternalName
  externalName: logs.internal
  ports:
  - name: https
    port: 443
    targetPort: 443
    protocol: TCP
  selector: {}
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: logs
  annotations:
    kubernetes.io/ingress.class: skipper
    zalando.org/skipper-routes: |
      Host("logs.example.com") -> "https://logs.internal";
spec:
  rules:
  - host: logs.example.com
    http:
      paths:
      - backend:
          serviceName: logs
          servicePort: https
```

When I apply it I get this in the skipper logs:
```
[APP]time="2018-05-17T10:14:29Z" level=info msg="route settings, update, route: kube_default__logs__logs_example_com____logs: Host(/^logs[.]example[.]com$/) -> \"http://:443\""

And this is what I get when I try to access:
```
[APP]time="2018-05-17T10:15:28Z" level=error msg="error during backend roundtrip: kube_default__logs__logs_example_com____logs: dial tcp :443: getsockopt: connection refused"
[APP]time="2018-05-17T10:15:28Z" level=error msg="error while proxying, route kube_default__logs__logs_example_com____logs, status code 503: dial tcp :443: getsockopt: connection refused"
81.139.52.202 - - [17/May/2018:10:15:28 +0000] "GET /_plugin/kibana/ HTTP/1.1" 503 20 "" "curl/7.59.0" 0 logs.example.com
```

Any ideas?

daviddyball:
There's no option to ignore certificate validation errors?

szuecs:
`-insecure` :)

[...]

PR #1098, opened 2019-06-20
support ingress with service type ExternalName

szuecs:
feature: #549 ingress with backend svc type ExternalName to do a proxy call to the externalName defined in the target service

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -559,15 +559,27 @@ func (c *Client) convertDefaultBackend(state *clusterState, i *ingressItem) (*es
 		)
 		log.Debugf("convertDefaultBackend: Found %d endpoints for %s: %v", len(eps), svcName, err)
 	}
+
+	if svc.Spec.Type == "ExternalName" {
+		scheme := "https"
+		if targetPort != "443" {
+			scheme = "http"
+		}
+		u := fmt.Sprintf("%s://%s:%s", scheme, svc.Spec.ExternalName, targetPort)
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

dataclients/kubernetes/kube.go
```diff
@@ -559,15 +559,27 @@ func (c *Client) convertDefaultBackend(state *clusterState, i *ingressItem) (*es
 		)
 		log.Debugf("convertDefaultBackend: Found %d endpoints for %s: %v", len(eps), svcName, err)
 	}
+
+	if svc.Spec.Type == "ExternalName" {
```
aryszka: wouldn't it be better to do this check before checking the target port and the endpoints?

dataclients/kubernetes/kube.go
```diff
@@ -656,15 +668,27 @@ func (c *Client) convertPathRule(
 		eps, err = state.getEndpoints(ns, svcName, svcPort.String(), targetPort)
 		log.Debugf("convertPathRule: Found %d endpoints %s for %s", len(eps), targetPort, svcName)
 	}
+
+	if svc.Spec.Type == "ExternalName" {
```
aryszka: same question as above: wouldn't it be better to do this check before checking the target port and the endpoints?

dataclients/kubernetes/kube.go
aryszka: I see that the current approach relies on the target port.

dataclients/kubernetes/kube.go
aryszka: I see that the current approach relies on the target port.

aryszka:
:+1:

aryszka:
:+1:

szuecs:
:+1:

PR #1104, opened 2019-06-24
fix chunked transfer-encoding

### LINKED ISSUE #1103: Transfer-Encoding: chunked blocks until first body was sent
**Describe the bug**
If a response with `Transfer-Encoding: chunked` is sent without body, skipper response is blocked until first body is sent from the backend.

**To Reproduce**

Client: 
- `curl -v localhost:9090`
- see the difference: `curl -v localhost:8080`
Run skipper: `./bin/skipper -inline-routes='foo: * -> "http://127.0.0.1:8080"'`
Run backend like this:
```
package main

import (
	"fmt"
	"log"
	"net/http"
	"time"
)

var data []byte = []byte(`{
    "apiVersion": "extensions/v1beta1",
    "kind": "Ingress",
    "metadata": {
        "creationTimestamp": "2018-11-20T21:01:23Z",
        "generation": 3,
        "labels": {
            "application": "foo"
        },
        "name": "nytest",
        "namespace": "default",
        "resourceVersion": "352055402",
        "selfLink": "/apis/extensions/v1beta1/namespaces/default/ingresses/nytest",
        "uid": "703ead8e-ed07-11e8-bbde-06405a32eeee"
    },
    "spec": {
        "rules": [
            {
                "host": "mytest",
                "http": {
                    "paths": [
                        {
                            "backend": {
                                "serviceName": "extname-test",
                                "servicePort": 443
                            }
                        }
                    ]
                }
            }
        ]
    },
    "status": {
        "loadBalancer": {
            "ingress": [
                {
                    "hostname": "kube-ing-lb-zaitko7kr3qb-702506834.us-west-1.elb.amazonaws.com"
                }
            ]
        }
    }
}`)

func main() {

	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		flusher, ok := w.(http.Flusher)
		if !ok {
			panic("expected http.ResponseWriter to be an http.Flusher")
		}

		w.Header().Set("Transfer-Encoding", "chunked")
		w.Header().Set("Content-Type", "application/json")
		for i := 0; i < 50; i++ {
			w.Header().Set(fmt.Sprintf("X-Foo%d", i), "application/json")
		}

		w.WriteHeader(200)
		flusher.Flush() // sent headers before data

		for i := 0; i < 2; i++ {
			time.Sleep(2 * time.Second)
			w.Write(data)
			flusher.Flush()
			log.Printf("write body")
		}
	})

	s := &http.Server{
		Addr:           ":8080",
		ReadTimeout:    10 * time.Second,
		WriteTimeout:   10 * time.Second,
		MaxHeaderBytes: 1 << 20,
	}
	log.Fatal(s.ListenAndServe())
}
```

**Expected behavior**
Send headers before body

**Observed behavior**
Headers are sent with the first body

szuecs:
fix #1103 http headers should be sent before the body

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -291,11 +291,6 @@ type PriorityRoute interface {
 	Match(*http.Request) (*routing.Route, map[string]string)
 }
 
-type flusherWriter interface {
-	http.Flusher
-	io.Writer
-}
-
 // Proxy instances implement Skipper proxying functionality. For
 // initializing, see the WithParams the constructor and Params.
 type Proxy struct {
@@ -395,7 +390,7 @@ func cloneHeaderExcluding(h http.Header, excludeList map[string]bool) http.Heade
 
 // copies a stream with flushing on every successful read operation
 // (similar to io.Copy but with flushing)
-func copyStream(to flusherWriter, from io.Reader, tracing *proxyTracing, span ot.Span) error {
+func copyStream(to flushedResponseWriter, from io.Reader, tracing *proxyTracing, span ot.Span) error {
 	b := make([]byte, proxyBufferSize)
 
 	for {
@@ -1117,9 +1112,8 @@ func (p *Proxy) serveResponse(ctx *context) {
 	}
 
 	ctx.responseWriter.WriteHeader(ctx.response.StatusCode)
-	fw := ctx.responseWriter.(flusherWriter)
-	fw.Flush()
-	err := copyStream(fw, ctx.response.Body, p.tracing, ctx.proxySpan)
+	ctx.responseWriter.Flush()
+	err := copyStream(ctx.responseWriter, ctx.response.Body, p.tracing, ctx.proxySpan)
 	if err != nil {
 		p.metrics.IncErrorsStreaming(ctx.route.Id)
 		p.log.Error("error while copying the response stream", err)
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1117,7 +1117,9 @@ func (p *Proxy) serveResponse(ctx *context) {
 	}
 
 	ctx.responseWriter.WriteHeader(ctx.response.StatusCode)
-	err := copyStream(ctx.responseWriter.(flusherWriter), ctx.response.Body, p.tracing, ctx.proxySpan)
+	fw := ctx.responseWriter.(flusherWriter)
+	fw.Flush()
+	err := copyStream(fw, ctx.response.Body, p.tracing, ctx.proxySpan)
 	if err != nil {
 		p.metrics.IncErrorsStreaming(ctx.route.Id)
 		p.log.Error("error while copying the response stream", err)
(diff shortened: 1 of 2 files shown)


linki:
For anyone interested: This flushes the HTTP headers if the body is empty (i.e. `copyStream()` doesn't flush) and it's the expected behaviour by `kubectl` when making watch requests that don't send any events initially.

proxy/proxy.go
```diff
@@ -1117,7 +1117,9 @@ func (p *Proxy) serveResponse(ctx *context) {
 	}
 
 	ctx.responseWriter.WriteHeader(ctx.response.StatusCode)
-	err := copyStream(ctx.responseWriter.(flusherWriter), ctx.response.Body, p.tracing, ctx.proxySpan)
+	fw := ctx.responseWriter.(flusherWriter)
```
linki: @szuecs Would it make sense to change the interface of [responseWriter in context](https://github.com/zalando/skipper/blob/3c9d0559f917ca7e35057103e367b38dc2dc47e5/proxy/context.go#L20) to something that extends `http.ResponseWriter` with the `Flush()` function? It should still satisfy everything that expects a plain `http.ResponseWriter`, shouldn't it?

aryszka:
:+1:

proxy/proxy.go
szuecs: Should this be a new public interface or not?

aryszka: no, not a public. I think flushing should be controlled only in the proxy package.

szuecs: Do you think the last commited change makes sense?

aryszka:
:+1:

proxy/proxy.go
```diff
@@ -395,7 +390,7 @@ func cloneHeaderExcluding(h http.Header, excludeList map[string]bool) http.Heade
 
 // copies a stream with flushing on every successful read operation
 // (similar to io.Copy but with flushing)
-func copyStream(to flusherWriter, from io.Reader, tracing *proxyTracing, span ot.Span) error {
+func copyStream(to flushedResponseWriter, from io.Reader, tracing *proxyTracing, span ot.Span) error {
```
linki: You could even keep the `flushWriter` here because its interface seems smaller than `flushedResponseWriter`. Makes  `copyStream` function more flexible. We just need Write and Flush?

proxy/proxy.go
linki: Yes, this is what I meant. A ResponseWriter that can also Flush. But you should judge if it makes sense and how it effects other parts of skipper.

proxy/proxy.go
aryszka: depends on the scope, but if we consider the entire proxy.go, then we need the full net/http.ResponseWriter, including the Header and WriteHeader functions

szuecs:
:+1:

PR #1106, opened 2019-06-25
fix format check to be as strict as format

aryszka:
- update the check-fmt Makefile task to match fmt
- fix formatting of code file that was formatted without the `-s` flag 


szuecs:
:+1:

aryszka:
:+1:

PR #1108, opened 2019-06-30
refactor OIDC filters enc/dec handling into secrets module

szuecs:
This breaks the signature of the OIDC Filters. IMO it's fine to do so, but wanted to highlight it.

This refactoring is a start to have some more features possible related to secrets rotation and dynamically adding secrets, for example to support mTLS or reasonable TLS termination in dynamic environments. It should be able to add cert+key to a TLSListener. 
I want to have letsencrypt support, such that I can get rid of my Apache+cronjobs. 
As first step I want to read from files, which would enable passing TLS and mTLS certs/keys via  Kubernetes secrets.

refactor: extract the Encrypt/Decrypt parts out of the oidc filter and use secrets package from the filter
refactor: secrets.Registry to have no global anymore


szuecs:
👍

secrets/encrypter.go
```diff
@@ -48,31 +48,31 @@ func newFileSecretSource(file string) secretSource {
 	return &fileSecretSource{fileName: file}
 }
 
-type encrypter struct {
+type Encrypter struct {
 	cipherSuites []cipher.AEAD
 	mux          sync.RWMutex
-	sSource      secretSource
-	closer       chan struct{}
+	SecSource    secretSource
```
aryszka: I think the SecSource and the Closer fields should stay encapsulated. There could be a public interface for the SecretSource, and a `func WithSource(SecretSource) *Encrypter {}` constructor could accept custom implementations when necessary (I see this use case in the secretstest). The closer never needs to be public, the constructor can just initialize the channel, so that it stays hidden and always available when the Close() method is called.

szuecs: done

szuecs:
:+1:

aryszka:
:+1:

Issue #1082, opened 2019-06-05
cron style predicate for time based route match


szuecs:
**Is your feature request related to a problem? Please describe.**

In case of scheduled maintenance window, you want to be able to schedule daily, weekly, monthly a fixed maintenance window and show a "come back later" message.

**Describe alternatives you've considered (optional)**

You can workaround the problem with a cronjob that changes the routes-file or the routes in etcd, for example

aryszka:
See also: https://opensource.zalando.com/skipper/reference/predicates/#interval

szuecs:
@aryszka from what I understand interval predicates can’t do something like everyday from 10-18 for example.

aryszka:
no, they cannot. They're just a closely related feature.

sarslanhan:
As discussed with both of you, I created a small library for doing this: https://github.com/sarslanhan/cronmask

@szuecs @aryszka let me know what you think about it. In the meantime, I can work on adding a predicate that makes use of this in Skipper.

szuecs:
Thanks @sarslanhan for providing the merged PR and the library.

PR #1111, opened 2019-07-07
Add "Cron" pred. that supports cron-like expressions to match routes

sarslanhan:
Link to the small library that I created for this: https://github.com/sarslanhan/cronmask

Summary of the features removed because they don't match the use case:
- [Non-standard characters](https://en.wikipedia.org/wiki/Cron#Non-standard_characters)
- Year field - We could probably add this as well if you see the need.
- Command section
- Text representation of the fields "month" and "day of week"

Relates to https://github.com/zalando/skipper/issues/1082

### DIFF
--- a/predicates/cron/cron.go
+++ b/predicates/cron/cron.go
@@ -49,14 +49,14 @@ func (*spec) Create(args []interface{}) (routing.Predicate, error) {
 }
 
 type predicate struct {
-	mask    cronmask.CronMask
+	mask    *cronmask.CronMask
 	getTime clock
 }
 
 func (p *predicate) Match(r *http.Request) bool {
 	now := p.getTime()
 
-	return p.mask.Matches(now)
+	return p.mask.Match(now)
 }
 
 func New() routing.PredicateSpec {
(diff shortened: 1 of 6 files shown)


predicates/cron/cron.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	mask, err := cronmask.New(expr)
+	if err != nil {
+		return nil, err
+	}
+
+	return &predicate{
+		mask:    mask,
+		getTime: getSystemTime,
```
szuecs: I am not sure, but can’t you use:
```
getTime: time.Now,
```

sarslanhan: Thank you, updated.

sarslanhan:
@szuecs both of those look good. Are they not running in CI build?

sarslanhan:
@aryszka had some comments and I updated the library based on those: https://github.com/sarslanhan/cronmask/pull/1. After I merge it, I would like to update it here as well.

aryszka:
:+1:

aryszka:
lgtm. Thanks for this PR!

szuecs:
:+1:

PR #1114, opened 2019-07-10
Fix/build release

szuecs:
Problem was that we enter multiple times to the same directory in the same step.
It should be done for each step only once


mikkeloscar:
:+1:

aryszka:
:+1:

PR #1116, opened 2019-07-10
differentiate error from backend and error during roundtrip

szuecs:
log error from stdlib that we do not handle good enough and we want to differentiate error from backend and error during roundtrip


szuecs:
:+1:

mikkeloscar:
:+1:

PR #1118, opened 2019-07-11
Apiusagemonitoring context

zeitlinger:
- proper support for API usage monitoring tag
- support for SLO monitoring for API usage monitoring

### DIFF
--- a/filters/apiusagemonitoring/spec.go
+++ b/filters/apiusagemonitoring/spec.go
@@ -2,11 +2,13 @@ package apiusagemonitoring
 
 import (
 	"encoding/json"
-	"github.com/sirupsen/logrus"
-	"github.com/zalando/skipper/filters"
 	"regexp"
 	"sort"
 	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/zalando/skipper/filters"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


filters/apiusagemonitoring/spec.go
```diff
@@ -32,6 +35,7 @@ func NewApiUsageMonitoring(
 	realmKeys string,
 	clientKeys string,
 	realmsTrackingPattern string,
+	tracer opentracing.Tracer,
```
aryszka: We could keep backwards compatibility, if we created a new function, e.g. `NewApiUsageMonitoringWithTracing`, and the current one called the new with the NopTracer. Even if there's a low chance that somebody uses this function directly, it's nicer not to assume that nobody does.

szuecs:
:-1: 

I am not really convinced that it makes sense to have opentracing integration in a filter that does not create costs, like call to external services. Therefore I would vote against following this idea. Spans are not free and we already have start stop logs for all filters.

szuecs:
:-1:

szuecs:
We talked internally about this and the decision with @zeitlinger was that there is a simpler way to do this, which will be followed up on. In case this needs to be done like this for whatever reason please reopen this PR.

PR #1125, opened 2019-07-15
Feature/secrets from files

szuecs:
I had some fun yesterday evening and implemented secrets rotation and an example `bearerinjector()` filter. Naming might be a bit rough, but otherwise seems to be ok to open the PR. The filter is meant for egress traffic, such that if we would implement the CONNECT headers, we can run applications that don't know IAM infra and use skipper to inject tokens for the application.

TODO:
- [x] remove fsnotify and replace with polling
- [x] update docs


cmd/skipper/secretsFlags.go
```diff
@@ -0,0 +1,28 @@
...
+	values []string
+}
+
+func (sf *secretsFlags) String() string {
+	return strings.Join(sf.values, ",")
+}
+
+func (sf *secretsFlags) Set(value string) error {
+	if sf == nil {
+		sf = &secretsFlags{}
```
aryszka: I have seen this in other places, but I don't think that it is useful to initialized this pointer here this way, because once Set exits, the reference on the caller side will still remain nil.

skipper.go
```diff
@@ -888,8 +892,13 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	}
 	defer o.SecretsRegistry.Close()
 
+	sp := secrets.NewSecretPaths()
```
aryszka: just to not forget to add a `defer sp.Close()`

secrets/file.go
```diff
@@ -0,0 +1,179 @@
...
+				dat, err := ioutil.ReadFile(event.Name)
+				if err != nil {
+					log.Errorf("Failed to read file (%s): %v", event.Name, err)
+					continue
+				}
+				log.Infof("update secret file: %s", event.Name)
+				sp.updateSecret(filepath.Base(event.Name), dat)
+			}
+		case err, ok := <-sp.watcher.Errors:
+			if !ok {
```
aryszka: The docs say that only the Events channel will be closed, so probably this condition is not required: https://godoc.org/gopkg.in/fsnotify/fsnotify.v1#Watcher .

secrets/file.go
```diff
@@ -0,0 +1,179 @@
...
+	known   map[string][]byte
+}
+
+func NewSecretPaths() *SecretPaths {
+	sp := &SecretPaths{
+		known: make(map[string][]byte),
+	}
+
+	if watcher, err := fsnotify.NewWatcher(); err != nil {
+		log.Errorf("Failed to create watcher, probably no support for fsnotify: %v", err)
```
aryszka: I would return here the error and then the calling code can decide not to start the process at all.

secrets/file.go
```diff
@@ -0,0 +1,179 @@
...
+		sp.watcher = watcher
+		go sp.startRefresher()
+	}
+
+	return sp
+}
+
+// GetSecret returns secret and if found or not for a given name.
+func (sp *SecretPaths) GetSecret(s string) ([]byte, bool) {
+	dat, ok := sp.known[s]
```
aryszka: the scopes where we access the known field should be protected, because multiple goroutines will use it. An RWMutex is fine.

secrets/file.go
```diff
@@ -0,0 +1,179 @@
...
+	watcher *fsnotify.Watcher
+	known   map[string][]byte
+}
+
+func NewSecretPaths() *SecretPaths {
+	sp := &SecretPaths{
+		known: make(map[string][]byte),
+	}
+
+	if watcher, err := fsnotify.NewWatcher(); err != nil {
```
aryszka: I am wondering if file system notifications are really required for this use case, and whether simple polling file stats changes in the date and size wouldn't be a more safe and robust solution. What do you think?

skipper.go
```diff
@@ -888,8 +892,13 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
+	for _, p := range o.CredentialsPaths {
+		sp.Add(p)
```
aryszka: I think we should handle the error that `*fsnotify.Watcher.Add()` may return.

szuecs:
we offline discussed that this might not work with kubernetes secrets, because inotify notifications might not be propagated

secrets/file.go
szuecs: will change this to a polling facility instead of event based, because of multiple people showed me that inotify is not really safe and we need a very safe way here.

secrets/file.go
szuecs: fixed

secrets/file.go
szuecs: As we discussed I will change it to polling

skipper.go
szuecs: not needed anymore

szuecs:
:+1:

filters/auth/bearer.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	secretName, ok := args[0].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	return newbearerInjectorFilter(secretName, b.secretsReader), nil
+}
+
+func newbearerInjectorFilter(s string, sr secrets.SecretsReader) *bearerInjectorFilter {
```
aryszka: `newbearerInjectorFilter` -> `newBearerInjectorFilter`, with capital b?

secrets/file.go
```diff
@@ -0,0 +1,179 @@
...
+	SecretsReader
+	// Add adds the given source that contains a secret to the
+	// automatically updated secrets store
+	Add(string) error
+}
+
+type SecretPaths struct {
+	mu      sync.RWMutex
+	quit    chan struct{}
+	secrets map[string][]byte
```
aryszka: if we use sync.Map for the files, we could also use it for the secrets, or the other way around, both maps could share the mutex

aryszka:
:+1:

filters/auth/bearer.go
szuecs: done

secrets/file.go
szuecs: Files can not easily handled with a mutex, because read to loop and each item has a single write, which breaks with the race detector.
I also looked into changing secrets to sync.Map, but it had also its difficulties, so I dropped that change.

szuecs:
:+1:

aryszka:
:+1:

PR #1127, opened 2019-07-17
add route id to filter context

Shareed2k:
i need a way inside plugin to understand to what route, request is related. In my situation i have plugin for route policy so i need a way to fetch specific rules from service to validate the request, so the easiest way i'm see here is get route id from filter context.

[feature request](https://github.com/zalando/skipper/issues/1128)

### DIFF
--- a/filters/filters.go
+++ b/filters/filters.go
@@ -71,6 +71,9 @@ type FilterContext interface {
 	// value in case it's a shunt, loopback. In case of dynamic backend is empty.
 	BackendUrl() string
 
+	// Gives filters access to the route id.
+	RouteId() string
+
 	// Returns the host that will be set for the outgoing proxy request as the
 	// 'Host' header.
 	OutgoingHost() string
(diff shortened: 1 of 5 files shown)


Shareed2k:
in my situation i need to implement plugin related to auth make route policy i need to create cache with routes related to specific service and i what to use route id as a key for cache and maybe also create special ratelimit rule with route id...

szuecs:
Can you elaborate on the need of the route ID?
What do you achieve with that and why it is required?

Shareed2k:
> Can you elaborate on the need of the route ID?
> What do you achieve with that and why it is required?
> 
> The best for this change would be to create a feature request to explain more background about this issue you solve. We have templates to guide with what information we need to understand the change better.

https://github.com/zalando/skipper/issues/1128

szuecs:
Thanks for creating the issue.

szuecs:
:-1:

szuecs:
Please reopen in case the proposed solution does not work. Thanks for your understanding. #1128

Issue #1115, opened 2019-07-10
add proxy errors to the opentracing logs


aryszka:
**Is your feature request related to a problem? Please describe.**
It is hard to relate errors in the application log to opentracing spans.

**Describe the solution you would like**
When the proxy logs an error, it could also add it as a log entry to the current opentracing span.

**Would you like to work on it?**
Yes

szuecs:
I am not sure if this volume of data makes sense. We already see problems with the amount of data we send to our lightstep satellites, I don't think it makes sense to add even more data in this case.
Maybe better to generate a uuid that corresponds to a request and pass this to all logs and all spans as tags.

aryszka:
The UUID is a good idea independently. Maybe we can even attach the existing trace ID to the logs.

OTOH, I don't think it would significantly increase the size of data sent out, because this is only about the errors, which should affect only a small portion of the traces.

I think that the problems with a particular proprietary provider have only limited relevance, but are we aware of any evidence that the message size causes problems? Or can it be only the rate of the messages?

szuecs:
Error logs are fine, sorry did not read the issue correctly.
Trace ID to logs seem to be good enough to me.

Issue #1130, opened 2019-07-19
Ratelimiting potentially spamming the error logs


aryszka:
We may report errors in the logs from rate limiting, even if the condition is rooted in the incoming request and neither in the code or the configuraiton. We should only log those cases where the rate limiting has a problem in the code or the config.

PR #1131, opened 2019-07-19
Refactor/proxy error logging

aryszka:
Refactoring error handling in the proxy and adding errors to tracing logs.

- [x] abort processing when the client context was cancelled (!!! needs extra attention)
- [ ] test cancelled client context (!!! needs extra attention)
- [x] move logging of proxying errors to the root of the call chain
- [ ] test that previously logged errors are still logged
- [ ] add errors to opentracing as log entries
- [ ] test errors in opentracing
- [ ] use flow ID when available both in the logs and in opentracing
- [ ] test flow ID

Closes: https://github.com/zalando/skipper/issues/1115
Related but addressed in the current PR: https://github.com/zalando/skipper/issues/1130

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -331,7 +331,7 @@ type proxyError struct {
 
 func (e proxyError) Error() string {
 	if e.err != nil {
-		return fmt.Sprintf("dialing failed %v: %v", e.DialError(), e.err.Error())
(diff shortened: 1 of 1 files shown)


AlexanderYastrebov:
@aryszka This one is pretty old, part of it was merged within #1563, client cancellation was test-covered in #1634, flow id opentracing tag is addressed in #1666

Should we close this and proceed in small steps maybe?

aryszka:
agree, yes, we can address the remaining ones from scratch

PR #1142, opened 2019-07-29
Build/improve performance

szuecs:
- speedup builds with GOPROXY and newer version of staticcheck
- make sure we use GO111 everywhere instead of the hardcoded value


aryszka:
:+1:

szuecs:
:+1:

PR #1146, opened 2019-07-31
Docs/egress features

szuecs:
document the egress features


docs/reference/egress.md
```diff
@@ -0,0 +1,128 @@
...
+In Kubernetes you can write
+[Secrets](https://kubernetes.io/docs/concepts/configuration/secret/)
+with an API and read them using a rotated, mounted files from skipper
+for example.
+
+To specify files or directories to find secrets, you can use
+`-credentials-paths` command line flag. Filenames are used to define
+the name of the secret, which will be used as a lookup key.
+
+Files need to be created before skipper started, and as of today
```
aryszka: The files need to be created before skipper is started...

docs/reference/egress.md
```diff
@@ -0,0 +1,128 @@
...
+start fake service
+
+```
+nc -l 8080
+```
+
+start skipper forward proxy
+
+```
+skipper -inline-routes='Host("host1") -> bearerinjector("sec1") -> "http://127.0.0.1:8080/"' -credentials-paths=/tmp/secrets -credentials-update-interval=10s
```
aryszka: I wouldn't call this yet a forward proxy, because it is not. For that we would need the dynamic backend set from the incoming scheme and host. Let's just call it a proxy here, and explain it in a separate paragraph that this behavior is missing.

aryszka: I see that it's metnioned below, so probably it's just enough not to call it a forward proxy.

docs/reference/filters.md
```diff
@@ -1484,3 +1490,29 @@ the api-backend service will receive a request with the path /api/foo%2Fbar/summ
...
+egress2: Method("GET") && Host("api.example.com") -> bearerinjector("read-token") -> "https://api.example.com/shoes";
+```
+
+To integrate with the `bearerinjector` filter you need to run skipper
+with `-credentials-paths=/tmp/secrets` and specify an update interval
+`-credentials-update-interval=10s`. Files in the credentials path can
+be a directory, which will be able to find all files within this
+directory, but it won't walk subtrees. For the example case, there
+have to be filenames `write-token` and `read-token` within the
+specified credential paths `/tmp/secrets/`, so
```
aryszka: "so" -> "resulting in"

szuecs:
:+1:

aryszka:
:+1:

PR #1149, opened 2019-08-01
monitor route creation time

zeitlinger:
Adds a metric that measures the route creation time.

The end time of the measurement is the successful creation of the route.
The start time of the measurement is taken from..
1. the creation time of ingresses for the kube DataClient - this creates the `routeCreationTime.default` metric
2. a provided creation time of apiUsageMonitoring filter - this creates the `routeCreationTime.apiUsageMonitoring` metric
3. other DataClients are not supported yet

### DIFF
--- a/filters/apiusagemonitoring/spec.go
+++ b/filters/apiusagemonitoring/spec.go
@@ -77,7 +77,7 @@ func NewApiUsageMonitoring(
 	}
 	unknownPath := newPathInfo(
 		unknownPlaceholder,
-		unknownPlaceholder,
+		noTagPlaceholder,
 		unknownPlaceholder,
 		noMatchPlaceholder,
 		noMatchPlaceholder,
--- a/skipper.go
+++ b/skipper.go
@@ -18,6 +18,7 @@ import (
 
 	ot "github.com/opentracing/opentracing-go"
 	log "github.com/sirupsen/logrus"
+
 	"github.com/zalando/skipper/circuit"
 	"github.com/zalando/skipper/dataclients/kubernetes"
 	"github.com/zalando/skipper/dataclients/routestring"
@@ -344,6 +345,9 @@ type Options struct {
 	// enabled by default.
 	EnableRouteBackendMetrics bool
 
+	// EnableRouteCreationMetrics enables the OriginMarker to track route creation times. Disabled by default
+	EnableRouteCreationMetrics bool
+
 	// When set, makes the histograms use an exponentially decaying sample
 	// instead of the default uniform one.
 	MetricsUseExpDecaySample bool
@@ -655,6 +659,7 @@ func createDataClients(o Options, auth innkeeper.Authentication) ([]routing.Data
 			KubernetesEnableEastWest:   o.KubernetesEnableEastWest,
 			KubernetesEastWestDomain:   o.KubernetesEastWestDomain,
 			DefaultFiltersDir:          o.DefaultFiltersDir,
+			OriginMarker:               o.EnableRouteCreationMetrics,
 		})
 		if err != nil {
 			return nil, err
@@ -1006,6 +1011,7 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 			loadbalancer.HealthcheckPostProcessor{LB: lbInstance},
 			loadbalancer.NewAlgorithmProvider(),
 			scheduler.NewRegistry(),
+			builtin.NewRouteCreationMetrics(mtr),
 		},
 		SignalFirstLoad: o.WaitFirstRouteLoad,
 	}
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
+
 	"github.com/zalando/skipper"
 	"github.com/zalando/skipper/dataclients/kubernetes"
 	"github.com/zalando/skipper/eskip"
@@ -121,6 +122,7 @@ const (
 	routeBackendErrorCountersUsage           = "enables counting backend errors for each route"
 	routeStreamErrorCountersUsage            = "enables counting streaming errors for each route"
 	routeBackendMetricsUsage                 = "enables reporting backend response time metrics for each route"
+	routeCreationMetricsUsage                = "enables reporting for route creation times"
 	metricsFlavourUsage                      = "Metrics flavour is used to change the exposed metrics format. Supported metric formats: 'codahale' and 'prometheus', you can select both of them"
 	metricsUseExpDecaySampleUsage            = "use exponentially decaying sample in metrics"
 	histogramMetricBucketsUsage              = "use custom buckets for prometheus histograms, must be a comma-separated list of numbers"
@@ -295,6 +297,7 @@ var (
 	accessLogJSONEnabled                bool
 	accessLogStripQuery                 bool
 	suppressRouteUpdateLogs             bool
+	routeCreationMetrics                bool
 
 	// route sources:
 	etcdUrls                  string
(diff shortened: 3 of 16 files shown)


filters/apiusagemonitoring/spec.go
```diff
@@ -169,6 +183,23 @@ func (s *apiUsageMonitoringSpec) buildUnknownPathInfo(paths []*pathInfo) *pathIn
 	return s.unknownPath
 }
 
+func (s *apiUsageMonitoringSpec) configInfo(apis []*apiConfig) apiConfigContext {
+	info := apiConfigContext{}
+
+	for _, api := range apis {
+		context := api.Context
+		if context.ConfigID == "" || context.ConfigCreated.IsZero() {
```
maxim-tschumak: Context is optional. This code doesn't look to support this case. (nil reference on `context.ConfigID`)

filters/apiusagemonitoring/spec.go
```diff
@@ -169,6 +183,23 @@ func (s *apiUsageMonitoringSpec) buildUnknownPathInfo(paths []*pathInfo) *pathIn
 	return s.unknownPath
 }
 
+func (s *apiUsageMonitoringSpec) configInfo(apis []*apiConfig) apiConfigContext {
```
maxim-tschumak: This can be a static function (not coupled to the `apiUsageMonitoringSpec` struct).

routing/filter_creation_metrics.go
```diff
@@ -0,0 +1,101 @@
...
+	metricsPrefix       = "routeCreationTime."
+	defaultCreationTime = "default"
+)
+
+type FilterCreationMetrics struct {
+	metrics   filters.Metrics
+	configIds map[string]map[string]int
+}
+
+type ConfigInfo interface {
```
maxim-tschumak: We should also document this interface.

routing/filter_creation_metrics.go
```diff
@@ -0,0 +1,103 @@
...
+func (m *FilterCreationMetrics) pruneCache() {
+	for _, fc := range m.configIds {
+		for id, age := range fc {
+			age++
+			if age > maxAge {
+				delete(fc, id)
+			} else {
+				fc[id] = age
+			}
+		}
```
tkrop: We should remove `_` from the map if `fc` is empty at after aging and pruning single maps to prevent a memory leak.

tkrop:
:+1:

zeitlinger:
:+1:

filters/apiusagemonitoring/spec.go
```diff
@@ -77,7 +78,7 @@ func NewApiUsageMonitoring(
 	}
 	unknownPath := newPathInfo(
 		unknownPlaceholder,
-		unknownPlaceholder,
+		noTagPlaceholder,
```
zeitlinger: @tkrop the rationale for using `{unknown}` instead of `{no-tag}`: We don't know what the right tag is, so saying it has no tag is more than we know.

tkrop: Okay. Feel free to role this back. I did this because of an request of @aryszka.

zeitlinger: Also fine for me.

zeitlinger:
:+1:

filters/apiusagemonitoring/spec.go
```diff
@@ -96,11 +97,17 @@ func NewApiUsageMonitoring(
...
-	Tag                   string   `json:"tag"`
-	ApiId                 string   `json:"api_id"`
-	PathTemplates         []string `json:"path_templates"`
-	ClientTrackingPattern string   `json:"client_tracking_pattern"`
+	ApplicationId         string           `json:"application_id"`
+	Tag                   string           `json:"tag"`
+	ApiId                 string           `json:"api_id"`
+	PathTemplates         []string         `json:"path_templates"`
+	ClientTrackingPattern string           `json:"client_tracking_pattern"`
+	Context               apiConfigContext `json:"context"`
```
aryszka: context may be not the right name for this, just don't have a good suggestion at the moment

filters/apiusagemonitoring/spec.go
aryszka: it is guaranteed. The `apiConfig.Context` field is not a pointer.

routing/filter_creation_metrics.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/zalando/skipper/filters"
+)
+
+const (
+	maxAge              = 2
+	metricsPrefix       = "routeCreationTime."
+	defaultCreationTime = "default"
+)
+
+type FilterCreationMetrics struct {
```
aryszka: As we discussed, the notion of "filter creation" is misleading because it is not the filter whose creation we can track but the route. This naming should be changed across the entire PR.

filters/filters.go
```diff
@@ -128,6 +128,14 @@ type Filter interface {
 	Response(FilterContext)
 }
 
+// Filters can optionally also implement the ConfigInfo interface
+// If they do, the route will get an additional 'routeCreationTime' metric entry
+// see filter_creation_metrics.go
```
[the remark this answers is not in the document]
aryszka: as we discussed IRL: this interface most probably won't be required if we use a generic `originMarker()` filter.

routing/filter_creation_metrics.go
```diff
@@ -0,0 +1,103 @@
...
+	startTimes := map[string]time.Time{}
+
+	t := m.configInfoStartTime(route.ConfigInfo, defaultCreationTime)
+	if !t.IsZero() {
+		startTimes[defaultCreationTime] = t
+	}
+
+	for _, f := range route.Filters {
+		t := m.filterStartTime(f)
+		o, exists := startTimes[f.Name]
```
aryszka: i don't see a reason to use the filter name as a key, because the creation time should depend only the route. This also can explode the metrics space.

aryszka: I think we should use the config ID as a key and not the filter name. This can also reduce the metrics space in case of grouped routes, like the ones coming from the same ingress object, in case of a Kubernetes scenario.

aryszka: (This also applies to the `default` key, I think it's not required.)

routing/filter_creation_metrics.go
```diff
@@ -0,0 +1,103 @@
...
+
+const (
+	maxAge              = 2
+	metricsPrefix       = "routeCreationTime."
+	defaultCreationTime = "default"
+)
+
+type FilterCreationMetrics struct {
+	metrics   filters.Metrics
+	configIds map[string]map[string]int
```
aryszka: what's the goal of this cache here? Can you give an explanation?

skipper.go
```diff
@@ -978,6 +978,7 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 			loadbalancer.HealthcheckPostProcessor{LB: lbInstance},
 			loadbalancer.NewAlgorithmProvider(),
 			scheduler.NewRegistry(),
+			routing.NewFilterCreationMetrics(mtr),
```
aryszka: Considering that this change has a potentially significant change to the memory usage, please make it an opt-in feature for both the library and the command line distributions by creating an option and a flag for it.

routing/filter_creation_metrics.go
zeitlinger: this is possible in the new design - so I added this pruning step as well

routing/filter_creation_metrics.go
zeitlinger: I added the documentation to RouteCreationMetrics

// RouteCreationMetrics reports metrics about the time it took to create metrics.
// It looks for filters of type OriginMarker to determine when the source object (e.g. ingress) of the route
// was created.
// If an OriginMarker with the same type and id is seen again, the creation time is not reported again, because
// a route with the same configuration already existed before.

filters/builtin/origin_marker.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/zalando/skipper/filters"
+)
+
+const (
+	OriginMarkerName = "originMarker"
+)
+
+type originMarkerSpec struct{}
+
+// OriginMarker carries information about the origin of a route
```
zeitlinger: @aryszka I'm not sure if we should also document it in filters.md - since it's a rather internal thing. what do you think?

tkrop:
:+1:

zeitlinger:
@aryszka I've managed to make this PR independent of #1155 - which hopefully makes this one easier.

zeitlinger:
:+1:

aryszka:
:+1:

szuecs:
Can you also add some documentation @zeitlinger , please?

zeitlinger:
> Can you also add some documentation @zeitlinger , please?

done

aryszka:
:+1:

szuecs:
:+1:

zeitlinger:
thanks :smile:

PR #1152, opened 2019-08-09
add option to provide a metrics backend instance (#1151)

### LINKED ISSUE #1151: Support custom Metrics implementations
I'm currently using skipper and I have a use case to export the metrics to NewRelic as custom events, I managed to make it work by polling the Codahale registry and sending these events to NewRelic but is just a quick hack that is not very reliable as the data is polled and also for some metric types the aggregation already took place.

Skipper should support a way to give users the ability to provide their own implementation of Metrics. This way there's no need to support any new metric kind out of the box and give the users real time flexibility to extend how their metrics are being stored.

Polling the registry is the way I'm doing it right now.
For this specific problem you could also provide a new relic custom events metrics provider and one would configure it in a similar way to how this is done for prometheus.

Yes

mtrovo:
Provide a way to specify the `metrics.Metrics` backend used by skipper to report metrics.

Fix #1151

### DIFF
--- a/skipper.go
+++ b/skipper.go
@@ -358,6 +358,11 @@ type Options struct {
 	// for these options can be set to false.
 	DisableMetricsCompatibilityDefaults bool
 
+	// Implementation of a Metrics handler. If provided this is going to be used
+	// instead of creating a new one based on the Kind of metrics wanted. This
+	// is useful in case you want to report metrics to a custom aggregator.
+	MetricsBackend metrics.Metrics
+
 	// Output file for the application log. Default value: /dev/stderr.
 	//
 	// When /dev/stderr or /dev/stdout is passed in, it will be resolved
@@ -845,7 +850,10 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 		DisableCompatibilityDefaults:       o.DisableMetricsCompatibilityDefaults,
 	}
 
-	mtr := metrics.NewMetrics(mtrOpts)
+	mtr := o.MetricsBackend
+	if mtr == nil {
+		mtr = metrics.NewMetrics(mtrOpts)
+	}
 	metrics.Default = mtr
 
 	// *DEPRECATED* client tracking parameter


metrics/metrics.go
```diff
@@ -172,6 +175,11 @@ func NewDefaultHandler(o Options) http.Handler {
 
 // NewMetrics creates a metrics collector instance based on the Format option.
 func NewMetrics(o Options) Metrics {
+	// custom metrics implementation has priority over Format configuration
+	if o.MetricsBackend != nil {
```
aryszka: this feels a bit like magic from the calling code's perspective. Since the calling case provides the options, it also just could us the MetricsBackend when available, without calling metrics.NewMetrics. It can be misleading also because in this case, all the other options are ignored.

skipper.go
```diff
@@ -827,6 +832,7 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	log.Infof("Expose metrics in %s format", metricsKind)
 	mtrOpts := metrics.Options{
 		Format:                             metricsKind,
+		MetricsBackend:                     o.MetricsBackend,
```
aryszka: (see my other comment for the proposed solution in metrics.NewMetrics, too)

I think we should rather check in this scope whether o.MetricsBackend (or simply o.Metrics is also a possible name) is nil, and if not, then use it, or if yes, only then call metrics.NewMetrics. Sg like:

```
mtr := o.MetricsBackend
if mtr == nil {
    mtrOpts := metrics.Options{/* ... all without the metrics backend */}
    mtr = metrics.NewMetrics(mtrOpts)
}
```

mtrovo: I'm gonna review my code but are you sure this is the preferred way? 

For me it's going to be a little weird because is going to take place on the `skipper` package and will be transparent on the `metrics` package itself. Which means other functions that call `NewMetrics` directly (right now it's just `NewDefaultHandler`) could have a totally different behaviour.

aryszka: Of course, cannot be 100% sure, but how I read this change, what happens is that we only create a metrics object from the metrics package, if we don't provide one from on our own. The current state reflects exactly this, while the previous one kind of made this more hidden. This is why I like it better like this, after you updated.

aryszka:
:+1:

mtrovo:
👍

njuettner:
👍

PR #1155, opened 2019-08-15
refactor kube data client

zeitlinger:
split the methods in the kube data client

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	log "github.com/sirupsen/logrus"
+
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/filters/builtin"
 	"github.com/zalando/skipper/predicates/source"
@@ -202,6 +203,20 @@ type Client struct {
 	pathMode                    PathMode
 	quit                        chan struct{}
 	defaultFiltersDir           string
+}
+
+type ingressContext struct {
+	state               *clusterState
+	ingress             *ingressItem
+	logger              *log.Entry
+	annotationFilter    string
+	annotationPredicate string
+	extraRoutes         []*eskip.Route
+	backendWeights      map[string]float64
+	pathMode            PathMode
+	redirect            *redirectInfo
+	hostRoutes          map[string][]*eskip.Route
+	defaultFilters      map[resourceId]string
 }
 
 var (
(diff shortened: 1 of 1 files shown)


tkrop:
:+1:

dataclients/kubernetes/kube.go
```diff
@@ -802,217 +802,275 @@ func (c *Client) ingressToRoutes(state *clusterState, defaultFilters map[resourc
...
-		if c.kubernetesEnableEastWest {
-			for _, rule := range i.Spec.Rules {
-				if rs, ok := hostRoutes[rule.Host]; ok {
-					rs = append(rs, createEastWestRoutes(c.eastWestDomainRegexpPostfix, i.Metadata.Name, i.Metadata.Namespace, rs)...)
-					hostRoutes[rule.Host] = rs
-				}
+func (c *Client) addSpecRule(ru *rule, logger *log.Entry, backendWeights map[string]float64,
+	extraRoutes []*eskip.Route, i *ingressItem, pathMode PathMode, hostRoutes map[string][]*eskip.Route,
+	redirect *redirectInfo, state *clusterState, annotationFilter string,
+	defaultFilters map[resourceId]string, annotationPredicate string) error {
```
szuecs: 4 Lines of parameters?
Looks like a smell to me.

zeitlinger:
What do you think about parameter structs that group related parameters?
I wasn't sure if this ideomatic.

szuecs:
I am not sure if it makes it better. Normally the data should be on the object itself and only a few parameters. Maybe you can identify an object that covers most of it and probably other methods would be also of this receiver type and get reduced number of parameters, too.
How would you suggest to group them?

zeitlinger:
@szuecs I've gave it a shot with the new `ingressContext` struct. Please tell me what you think.

dataclients/kubernetes/kube.go
```diff
@@ -802,217 +817,258 @@ func (c *Client) ingressToRoutes(state *clusterState, defaultFilters map[resourc
...
+	}
 
-		// parse predicate annotation
-		var annotationPredicate string
-		if val, ok := i.Metadata.Annotations[skipperpredicateAnnotationKey]; ok {
-			annotationPredicate = val
+	return routes, nil
+}
+
+func (c *Client) ingressRoute(i *ingressItem, redirect *redirectInfo, state *clusterState,
```
aryszka: while extracting this block to a method looks safe, it still could be cleaner. What makes a bit hard to comprehend is that it returns one route, but potentially it creates more routes which are added to the hostRoutes argument as a side effect.

aryszka:
:+1:

szuecs:
:+1:

szuecs:
Thanks for the effort @zeitlinger

PR #1160, opened 2019-08-26
feat: Add baggageItemToTag filter

addityasingh:
Add baggageItemToTag filter

This filter adds an opentracing tag for a given baggage item in the trace. 

Syntax:
```
baggageItemToTag("<baggage_item_name>", "<tag_name>")
```

Example: If a trace consists of baggage item named `foo` with a value `bar`. Adding below filter will add a tag named `baz` with value `bar` 
```
baggageItemToTag("foo", "baz")
```

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -1503,7 +1503,7 @@ It is also possible to enable this behavior centrally for a Skipper instance wit
 the -rfc-patch-path flag. See
 [URI standards interpretation](../../operation/operation/#uri-standards-interpretation).
 
-## Bearerinjector
+## bearerinjector
 
 This filter injects `Bearer` tokens into `Authorization` headers read
 from file providing the token as content. This is only for use cases
@@ -1528,3 +1528,17 @@ directory, but it won't walk subtrees. For the example case, there
 have to be filenames `write-token` and `read-token` within the
 specified credential paths `/tmp/secrets/`, resulting in
 `/tmp/secrets/write-token` and `/tmp/secrets/read-token`.
+
+## baggageItemToTag
+
+This filter adds an opentracing tag for a given baggage item in the trace. 
+
+Syntax:
+```
+baggageItemToTag("<baggage_item_name>", "<tag_name>")
+```
+
+Example: If a trace consists of baggage item named `foo` with a value `bar`. Adding below filter will add a tag named `baz` with value `bar` 
+```
+baggageItemToTag("foo", "baz")
+```
--- a/filters/builtin/builtin.go
+++ b/filters/builtin/builtin.go
@@ -128,6 +128,7 @@ func MakeRegistry() filters.Registry {
 		cors.NewOrigin(),
 		logfilter.NewUnverifiedAuditLog(),
 		tracing.NewSpanName(),
+		tracing.NewBaggageToTagFilter(),
 		accesslog.NewAccessLogDisabled(),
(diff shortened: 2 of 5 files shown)


szuecs:
@addityasingh thanks for the PR, can you please also add something in the documentation?
`docs/reference/filters.md` would need an update

aryszka:
:+1:

szuecs:
:+1:

docs/reference/filters.md
```diff
@@ -1528,3 +1528,17 @@ directory, but it won't walk subtrees. For the example case, there
 have to be filenames `write-token` and `read-token` within the
 specified credential paths `/tmp/secrets/`, resulting in
 `/tmp/secrets/write-token` and `/tmp/secrets/read-token`.
+
+## baggageItemToTag
```
mikkeloscar: Is `baggage` a general term used in OpenTracing? Just looking at the filter name `baggageItemToTag` I had no idea it had anything to do with OpenTracing but maybe it's just a convention that I'm not familiar with?

It's documented so I can learn that it's an OpenTracing thing but ideally filter names should be somewhat descriptive on their own.

szuecs: See https://godoc.org/github.com/opentracing/opentracing-go#Span it’s a member of this struct

aryszka:
thx @mikkeloscar . I think I get your concern. From Skipper's perspective it may not be necessary clear, what kind of 'baggage' this suppose to be. My suggestion now is to rename it to: `tracingBaggageToSpan'. Wdyt?

szuecs:
@aryzska that is even more weird to me.
We copy from span.BagageItem to span.Tag.
Do I miss something?

docs/reference/filters.md
addityasingh: Baggage is indeed a term used in opentracing to represent information valid throughout the process boundaries, unlike Span or Logs https://opentracing.io/docs/overview/tags-logs-baggage/

addityasingh:
IMO the name is self explanatory if we consider that BaggageItem and Tags are part of the core components of opentracing(https://opentracing.io/docs/overview/tags-logs-baggage/)

szuecs:
I agree, but maybe we should link in our docs to the opentracing docs

lmineiro:
I'm late to this party. For consistency, it probably would've been better to prefix with `tracing` just like the filter [`tracingSpanName`](https://github.com/zalando/skipper/blob/13c09daed036f5fb3de529a9e3ec5bb4acecce1f/filters/tracing/spanname.go#L12)

mikkeloscar:
Thanks I understand now that there were some meaningful reason behind the name in the context of open tracing, however I would still argue that we should avoid using such names in a "global" namespace of filter names. It's not unlikely that other parts of the system could have similar terminology internally as OpenTracing. I agree with @lmineiro that having it prefixed with `tracing` would make sense, similar to how the `oauth` filters also have the `oauth` prefix :)

aryszka:
prefixing was my intention, too, while keeping it short. Maybe a good name would be: `tracingBaggageToTag` (@szuecs you're right, tag is better than span)

addityasingh:
Cool. I will update the name and create a PR later today with `tracingBaggageToTag`

addityasingh:
@mikkeloscar This PR(https://github.com/zalando/skipper/pull/1168) addresses the renaming

PR #1164, opened 2019-08-28
Fix link to auth tutorial

AverageMarcus:
Fix link to auth tutorial

### DIFF
--- a/docs/kubernetes/ingress-usage.md
+++ b/docs/kubernetes/ingress-usage.md
@@ -453,7 +453,7 @@ Set a Cookie in the response path of your clients.
 
 ### Authorization
 
-Our [autnetication and authorization tutorial](../tutorials/auth/)
+Our [autnetication and authorization tutorial](../../tutorials/auth/)
 or [filter auth godoc](https://godoc.org/github.com/zalando/skipper/filters/auth)
 shows how to use filters for authorization.
 


szuecs:
can you sign-off the commit please?
We need DCO for our compliance rules

AverageMarcus:
Done. 

Out of curiosity, do you know if there's a way to sign commits when editing in GitHub's web UI?

szuecs:
no it's not possible

szuecs:
:+1:

docs/kubernetes/ingress-usage.md
```diff
@@ -453,7 +453,7 @@ Set a Cookie in the response path of your clients.
 
 ### Authorization
 
-Our [autnetication and authorization tutorial](../tutorials/auth/)
+Our [autnetication and authorization tutorial](../../tutorials/auth/)
```
mikkeloscar: `../tutorials/auth.md` would produce the right link no matter what path the docs are served under.

mikkeloscar:
:+1:

docs/kubernetes/ingress-usage.md
szuecs: we should redo the links ourselves

mikkeloscar: Thanks for the PR we will fix up all links at a later point since there's probably more than just this one. Your change makes it work so it's better than not having it :)

PR #1165, opened 2019-08-28
Refactor lifo

aryszka:
Refactor LIFO:
- [x] increase jobqueue dependency version, supporting reconfiguring an existing queue (depends on https://github.com/aryszka/jobqueue/pull/4)
- [x] fix the configuration of grouped lifo filters
- [x] fix the reconfiguration of any lifo filters
- [x] refactor the lifo filter interface
- [x] rename the stack -> queue
- [x] test the config updates
- [x] global lifo queue
- [x] provide metrics for the queue
- [x] update the documentation


szuecs:
As internally discussed I don't see a value in a global LIFO with the provided option.
I see only 2 options to not :-1: it:

1. You can drop the global lifo
2. You can make the settings for the global one based on memory setting. Basically what I want as operator is not guess values which can change based on user provided routes, but describe skipper has 200Mi and global lifo can limit memory consumption of it to 180Mi.

szuecs:
:-1:

aryszka:
@szuecs limiting the number of concurrently handled requests based on the memory consumption is a nice idea.

> Basically what I want as operator is not guess values which can change based on user provided routes...

Which values and how would they change based by user provided routes? The intention of the global LIFO is exactly that it is independent of the routes.

> guess values

measure? ;)

cmd/skipper/main.go
```diff
@@ -99,6 +99,7 @@ const (
 	enableHopHeadersRemovalUsage         = "enables removal of Hop-Headers according to RFC-2616"
 	rfcPatchPathUsage                    = "patches the incoming request path to preserve uncoded reserved characters according to RFC 2616 and RFC 3986"
 	maxAuditBodyUsage                    = "sets the max body to read to log in the audit log body"
+	enableRouteLIFOMetricsUsage          = "enalbe metrics for the individual route LIFO queues"
```
szuecs: enalbe -> enable

filters/scheduler/lifo.go
```diff
@@ -207,39 +210,28 @@ func (s *lifoGroupSpec) CreateFilter(args []interface{}) (filters.Filter, error)
...
-// SetStack binds the stack to the current filter context
-func (l *lifoFilter) SetStack(s *scheduler.Stack) {
-	l.stack = s
-}
-
-// GetStack is only used in tests
-func (l *lifoFilter) GetStack() *scheduler.Stack {
-	return l.stack
+// SetQueue binds the queue to the current filter context
+func (l *lifoFilter) SetQueue(s *scheduler.Queue) {
```
szuecs: maybe s should be renamed to q

filters/scheduler/lifo.go
```diff
@@ -248,37 +240,37 @@ func (l *lifoFilter) Response(ctx filters.FilterContext) {
...
+// Config returns the scheduler configuration for the given filter
+func (l *lifoGroupFilter) Config() scheduler.Config {
+	return l.config
 }
 
-// Key returns the scheduler group string
-func (l *lifoGroupFilter) Key() string {
-	return l.key
+// SetQueue binds the queue to the current filter context
+func (l *lifoGroupFilter) SetQueue(s *scheduler.Queue) {
```
szuecs: s -> q

filters/scheduler/lifo.go
```diff
@@ -298,11 +290,17 @@ func request(s *scheduler.Stack, key string, ctx filters.FilterContext) {
 		// TODO: replace the log with metrics
 		switch err {
 		case jobqueue.ErrStackFull:
-			log.Errorf("Failed to get an entry on to the stack to process StackFull: %v for host %s", err, ctx.Request().Host)
-			ctx.Serve(&http.Response{StatusCode: http.StatusServiceUnavailable, Status: "Stack Full - https://opensource.zalando.com/skipper/operation/operation/#scheduler"})
+			log.Errorf("Failed to get an entry on to the queue to process QueueFull: %v for host %s", err, ctx.Request().Host)
+			ctx.Serve(&http.Response{
+				StatusCode: http.StatusServiceUnavailable,
+				Status:     "Queue Full - https://opensource.zalando.com/skipper/operation/operation/#scheduler",
```
szuecs: :+1:

scheduler/scheduler_test.go
```diff
@@ -128,48 +131,194 @@ func TestScheduler(t *testing.T) {
...
 				}
 			}
-			// check pointers to stack are the same for same group
-			for k, stacks := range stacksMap {
-				firstStack := stacks[0]
-				for _, stack := range stacks {
-					if stack != firstStack {
-						t.Errorf("Unexpected different stack in group: %s", k)
+			// check pointers to queue are the same for same group
+			for k, stacks := range queuesMap {
```
szuecs: why stacks and not q ?

scheduler/scheduler_test.go
```diff
@@ -128,48 +131,194 @@ func TestScheduler(t *testing.T) {
...
+			FilterRegistry:  builtin.MakeRegistry(),
+			DataClients:     []routing.DataClient{cli},
+			PostProcessors: []routing.PostProcessor{
+				reg,
+			},
+		}
+
+		rt := routing.New(ro)
+		<-rt.FirstLoad()
+		return reg, rt, cli
```
szuecs: could also return a func close, that close reg and rt, then you don't have a double defer

skptesting/benchmark.inc
```diff
@@ -72,6 +72,7 @@ skp() {
 	fi
 
 	skipper -access-log-disabled -address "$1" -routes-file "$cwd"/"$2" -insecure \
+		-enable-global-lifo -global-lifo-max-concurrency 256 \
```
szuecs: does not exist anymore

scheduler/scheduler_test.go
aryszka: fixed

docs/operation/operation.md
```diff
@@ -207,6 +207,30 @@ It will expose them in /metrics, for example json structure looks like this exam
       /* stripped a lot of metrics here */
     }
 
+## LIFO metrics
+
+When enabled in the routes, LIFO queues can control the maximum concurrency level
+proxied to the backends and mitigate the impact of traffic pikes. The current
```
szuecs: pikes -> spikes

docs/operation/operation.md
```diff
@@ -207,6 +207,30 @@ It will expose them in /metrics, for example json structure looks like this exam
...
+filters, use the command line option:
+
+    -enable-route-lifo-metrics
+
+When queried, it will return metrics like:
+
+
+    {
+      "gauges": {
+        "skipper.lifo.route.active": {
```
szuecs: is `route` here the string `route` or is this per route and use the `routeid`?

docs/reference/filters.md
```diff
@@ -1475,9 +1478,19 @@ Example:
...
+The above configuration will set MaxConcurrency to 100, MaxQueueSize
 to 150 and Timeout to 10 seconds for the lifoGroup "mygroup", that can
-be shared between more than routes.
+be shared between multiple routes.
+
+It is enough to set the concurrency, queue size and timeout parameters for one instance of
+the filter in the group, and only the group name for the rest. Setting these values for
+multiple instances is fine, too. While only one of them will be used as the source for the
+applied settings, if there is accidentally a difference between the settings in the same
+group, a warning will be logged.
```
szuecs: Does this mean the first occurence will be used`Is this stable or results in undefined behavior?

scheduler/scheduler.go
```diff
@@ -1,122 +1,328 @@
...
-	return s.stack.Wait()
+// Config returns the configuration that the queue was created with.
+func (q *Queue) Config() Config {
+	return q.config
 }
 
-func (s *Stack) close() {
-	s.stack.Close()
+func (q *Queue) reconfigure() {
+	// renaming Stack -> Queue in the jobqueue project will follow
```
szuecs: comment is not clear to me
Does it mean that a rename would be a recreate (I suppose it would be)?

scheduler/scheduler.go
```diff
@@ -1,122 +1,328 @@
...
+		queue: jobqueue.With(jobqueue.Options{
+			MaxConcurrency: c.MaxConcurrency,
+			MaxStackSize:   c.MaxQueueSize,
+			Timeout:        c.Timeout,
+		}),
+	}
 
-func (r *Registry) setStack(name string, s *Stack) {
-	r.stacks[name] = s
+	if measure {
```
szuecs: do we need the local var?
It could be also `if r.options.EnableRouteLIFOMetrics`

scheduler/scheduler.go
```diff
@@ -1,122 +1,328 @@
...
+		if name == "" {
+			name = "unknown"
+		}
+
+		q.activeRequestsMetricsKey = fmt.Sprintf("lifo.%s.active", name)
+		q.queuedRequestsMetricsKey = fmt.Sprintf("lifo.%s.queued", name)
```
szuecs: As far as I see this would be a shared key for all "unknown" lifo queues.  -> user docs should be more clear in this case

scheduler/scheduler.go
```diff
@@ -1,122 +1,328 @@
...
+	}
+
+	return q
 }
 
-// Do implements routing.PostProcessor and sets the stack for the scheduler filters.
-//
-// It preserves the existing stack when available.
-func (r *Registry) Do(routes []*routing.Route) []*routing.Route {
+func (r *Registry) initLIFOFilters(routes []*routing.Route) []*routing.Route {
```
szuecs: For me it would be nicer to have this not renamed, if `Do` is only a single line function it does not provide value IMO.

scheduler/scheduler.go
```diff
@@ -1,122 +1,328 @@
...
 	rr := make([]*routing.Route, len(routes))
+	existingKeys := make(map[string]bool)
+	groups := make(map[string][]GroupedLIFOFilter)
+
 	for i, ri := range routes {
 		rr[i] = ri
 		for _, fi := range ri.Filters {
+			// TODO: warn on multiple lifos in the same route
```
szuecs: It sounds like a local `var count int` just before in the outer for loop with increase and if within the inner for loop. 
So should be done just now.

scheduler/scheduler.go
```diff
@@ -1,122 +1,328 @@
...
+			if glf, ok := fi.Filter.(GroupedLIFOFilter); ok {
+				groupName := glf.Group()
+				groups[groupName] = append(groups[groupName], glf)
+				continue
+			}
+
 			lf, ok := fi.Filter.(LIFOFilter)
+			if !ok {
+				continue
```
szuecs: Is this an error that we should report?

docs/operation/operation.md
aryszka: it's per route, and using the route ID. If it sounds fine, then I'll update the doc to reflect this better.

scheduler/scheduler.go
```diff
@@ -1,122 +1,328 @@
...
+			}
+
+			var q *Queue
+			key := fmt.Sprintf("lifo::%s", ri.Id)
+			existingKeys[key] = true
```
szuecs: PostProcessors are not running concurrently, but the tests could fail, if we run with race detector, which we do.
Maybe change to sync.Map ?

docs/reference/filters.md
aryszka: While post-processing, the first occurrence will be used, but the routes are not handled in a sorted way, so any route may become the first one. Therefore undefined.

scheduler/scheduler.go
```diff
@@ -1,122 +1,328 @@
...
+
+		var q *Queue
+		key := fmt.Sprintf("group-lifo::%s", name)
+		existingKeys[key] = true
+		qi, ok := r.queues.Load(key)
+		if ok {
+			q = qi.(*Queue)
+		}
+
+		if !ok {
```
szuecs: Maybe better to move it to he else from above for readability.
The `else if`  could be a separate if.

scheduler/scheduler.go
aryszka: I'm removing the comment. It's simply that in the jobqueue package, while the project was renamed from jobstack to jobqueue, the exported structure is still called Stack. It should be Queue.

scheduler/scheduler.go
aryszka: Indeed.

scheduler/scheduler.go
```diff
@@ -1,122 +1,328 @@
...
+
 	return rr
 }
 
-func (r *Registry) Close() {
-	for _, s := range r.stacks {
-		s.close()
+// Do implements routing.PostProcessor and sets the queue for the scheduler filters.
+//
+// It preserves the existing queue when available.
```
szuecs: Please mention side effect in the docs: Starts goroutine to create gauges, if `r.options.Metrics` is not nil and `measuring` is not true. (why do we need r.measuring? Can't we just use the nil check?)

scheduler/scheduler.go
aryszka: The name argument is either the route ID or the group name. The only case when the name is unknown, is when we use a single route, typically during experimenting. E.g:

```
skipper -inline-routes '* -> lifo(300, 150) -> static("/", ".") -> <shunt>
```

Question: is it worth to put this rare use case in the docs?

filters/scheduler/lifo.go
```diff
@@ -248,37 +239,37 @@ func (l *lifoFilter) Response(ctx filters.FilterContext) {
...
-	return r.Config(l.key)
+func (l *lifoGroupFilter) Group() string {
+	return l.name
 }
 
-// SetStack binds the stack to the current filter context
-func (l *lifoGroupFilter) SetStack(s *scheduler.Stack) {
-	l.stack = s
+func (l *lifoGroupFilter) HasConfig() bool {
+	return l.hasConfig
```
szuecs: I guess you want to differentiate a chosen empty scheduler.Config with a not configured one, but do we actually need this?
As far as I see this would be also ok to test `l.config != scheduler.Config{}`.

docs/reference/filters.md
aryszka: will mention it in the godoc, too.

filters/scheduler/lifo.go
aryszka: the difference is that it is also possible to set the defaults explicitly: `lifoGroup("groupname", 1, 0)`, and then it may occur that we don't get a config at all during the post-processing. This field tracks the difference.

scheduler/scheduler.go
aryszka: Based on my understanding about how the race detector works, which may not be complete, this is not a problem. I tested with -race, and this here seems to be fine. Let's fix it once it is a problem.

scheduler/scheduler.go
aryszka: Clarified: no

scheduler/scheduler.go
aryszka: the `r.measuring` is used to avoid starting multiple goroutines and starting them lazily only when they're required for the first time. Since the post-processor can run only on a single goroutine for other reasons anyway, this approach seems OK here.

docs/operation/operation.md
szuecs: it's fine

filters/scheduler/lifo.go
szuecs: ok

scheduler/scheduler.go
szuecs: ok

scheduler/scheduler.go
szuecs: ok

szuecs:
:+1:

aryszka:
:+1:

szuecs:
:+1:

PR #1168, opened 2019-08-29
feat: Update baggageItemToTag filter name

addityasingh:
Update `baggageItemToTag` filter name to `tracingBaggageToTag `. See this PR for details https://github.com/zalando/skipper/pull/1160

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -1511,7 +1511,7 @@ It is also possible to enable this behavior centrally for a Skipper instance wit
 the -rfc-patch-path flag. See
 [URI standards interpretation](../../operation/operation/#uri-standards-interpretation).
 
-## bearerinjector
+## bearerInjector
 
 This filter injects `Bearer` tokens into `Authorization` headers read
 from file providing the token as content. This is only for use cases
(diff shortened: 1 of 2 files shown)


ruiaraujo:
:+1:

docs/reference/filters.md
```diff
@@ -1511,7 +1511,7 @@ It is also possible to enable this behavior centrally for a Skipper instance wit
 the -rfc-patch-path flag. See
 [URI standards interpretation](../../operation/operation/#uri-standards-interpretation).
 
-## bearerinjector
+## bearerInjector
```
aryszka: this was actually cased like that, not saying that it's the right way ;)
https://github.com/zalando/skipper/blob/master/filters/auth/bearer.go#L9

aryszka:
:+1:

docs/reference/filters.md
addityasingh: In that case, I will revert this change from current PR as its not relevant. I will create an issue to discuss if its possible to change the name by keeping it backward compatible or would it be a breaking change

addityasingh:
```suggestion
## bearerinjector
```

mikkeloscar:
:+1:

szuecs:
:+1:

Issue #745, opened 2018-08-19
Add tls1.3 support 


szuecs:
https://github.com/golang/go/issues/9671 will add tls1.3 to Go.
Code review: https://go-review.googlesource.com/c/go/+/147638/
TLS1.3 is now a standard https://tools.ietf.org/html/rfc8446.

`srv.TLSConfig.MaxVersion = tls.VersionTLS12`

We need to:
- [x] build skipper with the new Go version  1.13
- [ ] Test
- [ ] Add metrics for tls version usage 
- [ ] add configuration support, maybe similar to traefik https://github.com/containous/traefik/blob/master/tls/certificate.go

szuecs:
https://github.com/golang/go/issues/30055 Go1.14 should have it as default

szuecs:
We have now `-tls-min-version`

PR #1175, opened 2019-09-04
Update skipper build for Go 1.13

mikkeloscar:
Updates skipper build for `go 1.13`

I basically ran `go mod tidy` with Go 1.13 to ensure that there are no changes to `go.mod` during build which would make the build "dirty" in git. :)


mikkeloscar:
Added an update which moves the flag parsing to be called in `func main` rather than in `func init`. Without this change the following fails:

```
go test -race -test.short -run ^Test github.com/zalando/skipper/cmd/skipper
flag provided but not defined: -test.testlogfile
Usage of /tmp/go-build634583681/b001/skipper.test:
```

I'm not completely sure on the impact of this change but considering it's bad practice to do a lot of things in `init` causing problems like this, it probably makes sense :)

The `test.testlogfile` flag comes from: https://github.com/golang/go/blob/befa17c5f4591e180e5034fd4e8ca508f52dc7bc/src/cmd/go/internal/test/test.go#L1122

szuecs:
Disabled staticcheck for now

szuecs:
Go 1.13 has TLS1.3 https://github.com/zalando/skipper/issues/745

szuecs:
:+1:

mikkeloscar:
:+1:

Issue #660, opened 2018-05-31
skipper route CRD


vetinari:
We're currently running skoap with a single route like this
```
   Path("/*") && Method("GET")
              -> auth("/services")
              -> dropRequestHeader("Authorization")
              -> basicAuth("username", "secret")
              -> "https://special.service.example.com";
```
with `auth("/services")` -> `oauthTokeninfoAllKV("realm","/services")` for newer skipper versions.

We would now need an ingress definition with incoming host names defined and one disconnected kubernetes service, that is not deleted in later and has a special meaning for this route. We have no pods involved in this route. The current `zalando.org/skipper-routes` annotation can be used as workaround, but needs to have a kubernetes service that exists, too.
https://github.com/zalando/skipper/issues/512

szuecs:
We might want to think about having a customer skipper route CRD for the use case of custom skipper routes.

szuecs:
Request for comments:

```
apiVersion: zalando.org/v1
kind: SkipperRoute
metadata:
  name: "{{{APPLICATION}}}-blogposts-put"
  labels:
    application: "{{{APPLICATION}}}"
    version: "{{{BUILD_VERSION}}}"
spec:
  ingress: "name-reference"  # optional, copy base predicates and filters from ingress
  filters: # optional
  - localRatelimit(“1m”, 10)
  - enableAccessLog()
  - flowId("reuse")
  predicates:
  - JWTPayloadAllKV("iss", "https://idp.example.com", “https://identity.example.com/role”, “publisher”)
  - Method(“PUT")
  - PathSubtree(“/blogposts/:id”)
status:
  applied: <timestamp>
```

My Questions:
- [ ] How should the behavior of the `status` field be?
  - first applied?
  - should all instances add themselves, but how to delete them again?
  - does it makes sense to have a status or do we need some other data?
- [ ] Anything you miss in this CRD to support your use cases?

szuecs:
@aermakov-zalando do you have ideas for the status field, because you talked about "the usual pattern is that the controller writes the latest value from metadata.generation into status.observedGeneration when it does stuff". One problem is that we have N skippers and only one object status field, but maybe you have an idea.

aermakov-zalando:
No idea. I'd either create a separate controller that would get the results from all Skipper instances and update the status, or use leader election in Skippers and collect the data in one of the instances. Both of these would be annoying to implement, though.

aermakov-zalando:
A couple of suggestions:
[...]

PR #1180, opened 2019-09-05
Feature/routegroup crd

szuecs:
This PR is ~~work in progress~~ ready for review.
The PR is used to comment on the proposal (crd.md).

General discussion in Slack or https://github.com/zalando/skipper/issues/660


dataclients/kubernetes/crd.md
```diff
@@ -0,0 +1,601 @@
...
+
+```yaml
+apiVersion: zalando.org/v1
+kind: RouteGroup
+spec:
+  hosts:
+  - <string>
+  backends:
+  - <backend>
+  paths:
```
szuecs: @aermakov-zalando suggested `routes`, instead of `paths`. 
I could also imagine `routes_template`, but I think `routes` sounds more nice. Template, because dependent on number of `hosts`, the routes will be multiplied.

dataclients/kubernetes/crd.md
```diff
@@ -0,0 +1,601 @@
...
+  - <path>
```
szuecs: If we change to `routes`, this should be changed to `<route>`

dataclients/kubernetes/crd.md
szuecs: changed `paths` to `routes`

dataclients/kubernetes/crd.md
szuecs: changed `<path>` to `<route>`

dataclients/kubernetes/crd.md
```diff
@@ -0,0 +1,596 @@
...
+
+The purpose of the document is to define a CRD, that solves issues
+we observed, while using Ingress.
+
+There are some other solutions out there, like Istio, SMI and others,
+but all of these have drawbacks for skipper provided use cases. One
+common example is an application that needs a redirect to somewhere
+else for a specific path in an Ingress type of object. This example is
+only provided with multiple objects or annotations, both are unsafe
+for concurrent use.
```
mikkeloscar: I think we need to expand on this section and show more clearly why some other existing solutions doesn't provide what we need or make it overly complex. The statement: "both are unsafe for concurrent use" is hard to validate as a reader without some small example.

I think it's important to focus on the reason why we want a single CRD which the other solutions don't provide and what we believe the benefit is of that, because feature wise we probably could implement most of it with SMI (I'm not sure) and if we can't implement all the features then this should also be highlighted as it makes it more reasonable to go with a custom solution.

dataclients/kubernetes/crd.md
```diff
@@ -0,0 +1,596 @@
...
+
+## Problem statement
+
+Provide a format for routing input that makes it easier to benefit
+more from Skipper's routing features, without workarounds, than the
+generic Ingress specs. Primarily targeting more complex Kubernetes
+Ingress scenarios.
```
mikkeloscar: This reads more like the beginning of a solution. The problem statement should IMO make it clear how we provide features using multiple ingresses and why this is a problem.

This makes it more clear that a single resource is a benefit in the proposed solution section.

dataclients/kubernetes/crd.go
```diff
@@ -0,0 +1,171 @@
...
+
+type pathSpec struct {
+	Path       string   `json:"path"`
+	Method     string   `json:"method,omitempty"`
+	Filters    []string `json:"filters,omitempty"`
+	Predicates []string `json:"predicates,omitempty"`
+	Backend    string   `json:"backend,omitempty"`
+}
+
+type algorithm int
```
szuecs: @aryszka Why we don't reuse these from the loadbalancer package?

dataclients/kubernetes/fixtures/skipper_route_crd_complex_example.yaml
```diff
@@ -0,0 +1,40 @@
...
+kind: RouteGroup
+metadata:
+  name: my-routes
+spec:
+  backends:
+  - serviceName: foo-service-v1
+    servicePort: 80
+  paths:
+  - path: /api/resource
+    method: post
```
cjbooms: `operation: post` would be more aligned with OpenApi3

dataclients/kubernetes/crd.go
```diff
@@ -0,0 +1,171 @@
...
+)
+
+// Shall we support dynamic backend? Does it make sense in this
+// scenario?
+func backendTypeFromString(s string) (eskip.BackendType, error) {
+	switch s {
+	case "", "service":
+		return serviceBackend, nil
+	case "external":
+		return externalURL, nil
```
szuecs: why not eskip.NetworkBackend?

dataclients/kubernetes/fixtures/skipper_route_crd_complex_example.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - f1("foo")
+      - f2("bar", "baz")
+      predicates:
+      - P1("serviceUser")
+      - P2("bar", "baz")
+    - filters:
+      - f1("dosomething")
+      predicates:
+      - P1("adminusers")
+  - path: /api/resource/*
```
cjbooms: Rather than re-declare path over and over again, operations could be listed under a path as in OpenApi:
```
        - path: /api/resource/*
          - operation:  get
            configs:
              - filters:
                  - f("foo")
                predicates:
                  - P("b", 0.4)
                backend:  # overrides default
                  serviceName: migrate-to-bar-service
                  servicePort: 8080
          - operation:  put
            configs:
              - filters:
                  - f("foo")
                predicates:
                  - P("b", 0.4)
                backend:  # overrides default
                  serviceName: migrate-to-bar-service
                  servicePort: 8080
          - operation:  post
            configs:
              - filters:
                  - f("foo")
                predicates:
                  - P("b", 0.4)
                backend:  # overrides default
                  serviceName: migrate-to-bar-service
                  servicePort: 8080
```

dataclients/kubernetes/crd.go
```diff
@@ -0,0 +1,197 @@
...
+// The main reason is to differentiate between service and external, in a way
+// where we can also use the current global option to decide whether the service
+// should then be converted to LB. Or shall we expect the route group already
+// contain the pod endpoints, and ignore the global option for skipper?
+const (
+	serviceBackend = eskip.LBBackend + iota
+	externalURL
+)
+
+// Shall we support dynamic backend? Does it make sense in this
```
herojan: At least the shunt is useful for our purposes, since we use it to reject tls requests or to only give 404 if the client is authenticated and 403 otherwise. Our catchall routes use PathSubtree("/"), is this rule equivalent:

```yaml
- path: /*
  configs:
    - predicates:
      - ...
    - filters:
      - <shunt>
    
```

dataclients/kubernetes/fixtures/skipper_route_crd_complex_example.yaml
cjbooms: Ignore this if `method:` better aligns with Skipper's nomenclature

dataclients/kubernetes/crd.go
szuecs: `<shunt>` is already supported

dataclients/kubernetes/crd.md
szuecs: SMI does not provide access to predicates and filters, but on the other hand yes you are right I will add this.

dataclients/kubernetes/fixtures/skipper_route_crd_complex_example.yaml
szuecs: I like the suggestion.
@aryszka and @aermakov-zalando what do you think? Any opinions on that?

dataclients/kubernetes/crd.md
szuecs: Changed the purpose to make it more focused on us

dataclients/kubernetes/crd.md
szuecs: I changed purpose and problem statement to focus on what we want to achieve, not on weaknesses of other solutions.

dataclients/kubernetes/crd.md
```diff
@@ -0,0 +1,14 @@
+# RouteGroup CRD
+
+This document is a temporary design document, continously updated as this development branch evolves.
+
+## Goal of the RouteGroup CRD
+
+Provide a format for routing input that makes it easier to benefit more from Skipper's routing features, without
+workarounds, than the generic Ingress specs. Primarily targeting the Kubernetes Ingress scenario.
```
aryszka: "..., without workarounds that are required when using the generic Ingress specs..."

dataclients/kubernetes/crd.go
aryszka: I am not against reusing them. One reason can be less coupling, but I think it's also fine to use them in this case.

dataclients/kubernetes/crd.go
aryszka: It doesn't have the same meaning. When we generate the actual routes, and a service uses ClusterIP, then both the external and cluster IP would be network backends in eskip terms. I wanted to avoid mixing them here. It's just a package internal temporary phase currently. Would it still be better to just go and use eskip.NetworkBackend?

dataclients/kubernetes/fixtures/skipper_route_crd_complex_example.yaml
aermakov-zalando: If the `method` selector is changed to support multiple values (I think I proposed it in the other discussion), does it actually lead to an improvement? Or is it just because the "path" predicated is required in Skipper?

aryszka: Agree, more DRY and also familiar for the users.

dataclients/kubernetes/crd.md
```diff
@@ -0,0 +1,634 @@
...
+  hosts:
+  - api.example.org
+  - api.service.example.net
+  backends:
+  - name: api-svc
+    type: service
+    serviceName: api-service-v1
+    servicePort: 80
+    default: true
+    weight: 80
```
aermakov-zalando: Copying my comments from the other proposal document.

The semantics of this are very unclear and I think this needs to be split. Here, we define the implementation of the backend, whether it's a default or not, and its traffic weight in one declaration, but these are completely different things. For example, I might want to define three different services, have two of them be default ones with traffic weights 70 and 30, but then want to use all three in a dedicated route with different traffic weights, and how would I do that? Or, what exactly is going to happen if one of the service is no longer marked as default but still has a traffic weight?

I propose to change this as follows:

1) Have one section to define the backends:

       backends:
         - name: api-svc-v1
           type: service
           serviceName: api-service-v1
           servicePort: 80
         - name: api-svc-v2
           type: service
           serviceName: api-service-v2
           servicePort: 80

2) Then add another section to configure the default backends. This is a simple list of backend names and their traffic weights. If unset, and there's just one backend in the backend list, make it default to that backend only:

       defaultBackends:
         - name: api-svc-v1
           weight: 70
         - name: api-svc-v2
           weight: 30

3) Change the per-route backend to the same structure as the default backends (since the users should have freedom to define custom backends per-route):
    
       - path: /api
         backends:
          - name: api-svc-v2
            weight: 100

dataclients/kubernetes/crd.md
```diff
@@ -0,0 +1,634 @@
...
+  - api.service.example.net
+  backends:
+  - name: api-svc
+    type: service
+    serviceName: api-service-v1
+    servicePort: 80
+    default: true
+  routes:
+  - path: /api/resource
+    method: post
```
aermakov-zalando: This should be a list. It's highly unlikely that someone would want to setup different policies for e.g. `put` vs `post` vs `patch`, so I'd rather cover the 99% case in something that wouldn't require copy-pasting the route.

dataclients/kubernetes/crd.md
```diff
@@ -0,0 +1,634 @@
...
+  default: <bool> optional  // TODO: maybe separate object
+  weight: <number> optional // TODO: maybe separate object
+```
+
+`<stringarray>` is an array of `<string>`
+`<number>` is an unsigned integer, example: 80
+`<bool>` is a either `true` or `false`
+
+```
+<route>
```
aermakov-zalando: One thing that I'd like to see addressed is the route specificity issue. For me as a user, it'd make sense if the routes were matched in the order they were declared in the CRD, but IIUC Skipper matches based on the number of predicates. Since a lot of these predicates will be hidden from the users, I don't think it'll be easy to reason about the ordering, which might lead to nasty surprises (e.g. a route requiring authentication silently skipped).

dataclients/kubernetes/crd.go
szuecs: Good question, I think at service parse time it makes sense to differentiate between service types, on intermediate object to create routes, I would prefer the skipper specific names.

dataclients/kubernetes/crd.md
szuecs: this part was rewritten in a concurrent change :)

dataclients/kubernetes/crd.md
szuecs: So you suggest:

```yaml
apiVersion: zalando.org/v1
kind: RouteGroup
spec:
  hosts:
  - <string>
  backends:
  - <backend>
  default_backends: optional
  - <default_backend>
  routes:
  - <route>

<backend>
  name: <string>
  type: <string> that is one of "service|shunt|loopback|lb|url"
  url: <string> required for type=url
  algorithm: <string> optional, only user for type=lb
  endpoints: <stringarray> required for type=lb
  serviceName: <string> required for type=service
  servicePort: <number> required for type=service

<default_backend>
  default: <bool>
  weight: <number>

<route>
  path: <string>
  exactPath: <bool> optional
  backend: optional, overrides defaults
  - <backend_ref>
  filters: <stringarray>, optional
  predicates: <stringarray>, optional
  method: <string>, optional, one of the HTTP methods "GET|HEAD|PATCH|POST|PUT|DELETE|CONNECT|OPTIONS|TRACE", defaults to all

<backend_ref>
  - backend_name: <string>
    weight: <number>, optional


<string>  an arbitrary string
<stringarray> is an array of <string>
<number> is an unsigned integer, example 80
<bool> is a either true or false
```

dataclients/kubernetes/crd.md
szuecs: So you mean, that it's likely to have the same rules for the `writes` and others for the `reads` like different scopes in auth filters, for example

dataclients/kubernetes/crd.md
aermakov-zalando: `default_backends` and `route.backends` should use the same structure that's a list of `backend_ref`.

dataclients/kubernetes/crd.md
szuecs: This will be out of scope and not really covered. 
This would need a major change in the routing tree and I don't see too much value of this.
Also beware, that the Path and PathSubtree are reducing the routes in a lookup tree to not have to do a linear search based on predicates on all routes, but the reduced number of routes.

dataclients/kubernetes/crd.md
```diff
@@ -0,0 +1,634 @@
...
+    - modPath("/api", "/")
+  - path: /login
+    method: get
+    backend: redirect
+    filters:
+    - redirectTo(308, "https://login.example.org/)
+```
+
+
+#### Complex routes with ratelimits based on tokens
```
irishshagua: 💯

dataclients/kubernetes/crd.md
aermakov-zalando: @szuecs Please fix the casing to `camelCase` (default_backends, backend_name, etc).

How does `path_regexp` work? Does it override `path`/`path_subtree`, or can you use it at the same time as those?

aermakov-zalando: Also, `route.backend` -> `route.backends`.

szuecs: Thanks will push in a bit

szuecs: fixed all the things

dataclients/kubernetes/crd.md
```diff
@@ -0,0 +1,898 @@
...
+Skipper's `kubernetes` dataclient will also support Ingress in the
+future.
+
+## Purpose
+
+The purpose of the document is to define a RouteGroup CRD, that allows
+configuring Skipper for Kubernetes such that it benefits from all the
+Skipper features in a straightforward way.
+
+## Problem statement
```
mikkeloscar: @szuecs I still think the problem statement is too "weak" for the big change we want to make.
It focuses on a specific use which may be the initial motivation for this change, but it's not the only reason afaik.

IMO the problem is the following (and possible more)

* Multiple ingresses are complicated for users. It's easier for users to create overlapping ingresses when they can't define everything in a single resource.
* Orchestrating traffic switching is complicated when multiple ingresses are involved.
* Configuring everything via annotations is not as nice as configuring via fields in the spec.

Additionally we should make it clear why we don't go with one of the other designs already in the community like SMI. Maybe this doesn't belong in the problem statement but at least in the background information.

dataclients/kubernetes/crd.md
```diff
@@ -0,0 +1,898 @@
...
+        path: /
+      - backend:
+          serviceName: my-app-2
+          servicePort: http
+        path: /
+```
+
+The problem with having multiple Ingress having this kind of
+annotation is that it needs to be in-sync across all Ingress belonging
+to a group of routes. This has of course a race condition, which is a
```
mikkeloscar: It's not clear why their is a race condition in this case. I actually don't know what the race condition is?

dataclients/kubernetes/crd.md
```diff
@@ -0,0 +1,898 @@
...
+```yaml
+<route>
+  path: <string>            either path or pathSubtree is allowed
+  pathSubtree: <string>     either path or pathSubtree is allowed
+  pathRegexp: <string>      optional
+  backends:                 optional, overrides defaults
+  - <backendRef>
+  filters: <stringarray>    optional
+  predicates: <stringarray> optional
+  methods: <stringarray>     optional, one of the HTTP methods "GET|HEAD|PATCH|POST|PUT|DELETE|CONNECT|OPTIONS|TRACE", defaults to all
```
mikkeloscar: "one or more" ? since it's an array.

dataclients/kubernetes/crd.md
szuecs: Let's talk, I think I covered all your points already in the last section of the problem statement.

I started a longer explanation what I think is wrong in x and y, but there was a good point of being nicer and telling what we want to achieve: better skipper integration in Kubernetes.

dataclients/kubernetes/crd.md
szuecs: race condition: multiple, not global synchronized, writes and a list read

dataclients/kubernetes/crd.md
szuecs: fixed

dataclients/kubernetes/crd.md
szuecs: I think I covered now everything more clearly.
I also added in the bottom a  "why not adapt X" section

dataclients/kubernetes/crd.md
```diff
@@ -0,0 +1,1032 @@
...
+    - JWTPayloadAllKV("iss", "https://accounts.google.com", "email", "skipper-router@googlegroups.com")
+  - path: /api/resource
+    filters:
+    - ratelimit(20, "1m")
+    - oauthTokeninfoAllKV("iss", "https://accounts.google.com")
+```
+
+#### A/B test
+
+A/B test via cookie `canary`, used for sticky sessions.
```
aryszka: this is probably just a naming issue: we should try not to mix A/B tests and canary tests because they are quite different high level concepts. We use canary tests to make sure that the new versions of an application are stable. A/B tests may span over multiple versions of an application, e.g. when a variant or each variants need a bugfix, but we want to continue with the A/B test.

szuecs: Feel free to change it, but IIRC I used canary, because it was actually testing that the B branch works, before switching 10% customers to it.

aryszka: I think 'canary' is the better name. Replacing 'A/B' with 'canary'.

aryszka: Actually... not in every case. Ignore the above.

dataclients/kubernetes/crd.go
aryszka: reusing

dataclients/kubernetes/api_test.go
```diff
@@ -0,0 +1,428 @@
...
+			var o map[string]interface{}
+			if err := d.Decode(&o); err == io.EOF || len(o) == 0 {
+				break
+			} else if err != nil {
+				return nil, err
+			}
+
+			kind, ok := o["kind"].(string)
+			if !ok {
+				println("from here")
```
aryszka: remove tracing

dataclients/kubernetes/definitions.go
```diff
@@ -4,14 +4,11 @@ import (
...
 	"fmt"
-	"sort"
 	"strconv"
 	"time"
-
-	log "github.com/sirupsen/logrus"
 )
 
-type resourceId struct {
+type resourceID struct {
```
aryszka: let's rename definitions to ingressdefinitions and routegroupcrd to routegroupdefinitions

dataclients/kubernetes/ingress.go
```diff
@@ -0,0 +1,775 @@
...
+				path.Backend.Traffic = 1.0 / float64(sc.count)
+			}
+			// reduce count by one in order to give subsequent
+			// backends for the path a higher relative weight.
+			sc.count--
+		}
+	}
+}
+
+// TODO: default filters not applied to 'extra' routes from the custom route annotations. Is it on purpose?
```
szuecs: I think this should be fixed, if it was done like that. I remember, that I changed it to apply everything from the original ingress including filters and predicates, such that custom routes are bound to the ingress and can not increase the scope, but always have to reduce the scope.

dataclients/kubernetes/ingress.go
```diff
@@ -0,0 +1,775 @@
...
+		eps, err = state.getEndpoints(
+			ns,
+			svcName,
+			svcPort.String(),
+			targetPort,
+		)
+		log.Debugf("convertDefaultBackend: Found %d endpoints for %s: %v", len(eps), svcName, err)
+	}
+
+	if len(eps) == 0 || err == errEndpointNotFound {
```
szuecs: Based on our discussion: do we want to drop this service handling if we have a service type ClusterIP?

`svc.Spec.Type == "ClusterIP"`

see also above "ExternalName" https://github.com/zalando/skipper/pull/1180/files#diff-cc26174bd99fb74da18dffde42bd2efcR502

dataclients/kubernetes/ingress.go
```diff
@@ -0,0 +1,775 @@
...
+			Filters:     f,
+		}, nil
+	} else {
+		// err handled below
+		eps, err = state.getEndpoints(ns, svcName, svcPort.String(), targetPort)
+		log.Debugf("convertPathRule: Found %d endpoints %s for %s", len(eps), targetPort, svcName)
+	}
+
+	// TODO: errors may get ignored if err != nil && err != errEndpointNotFound && len(eps) == 0
+	if len(eps) == 0 || err == errEndpointNotFound {
```
szuecs: similar to https://github.com/zalando/skipper/pull/1180/files#r369044301

dataclients/kubernetes/ingress.go
```diff
@@ -0,0 +1,775 @@
...
+	} else if err != nil {
+		return nil, err
+	}
+	log.Debugf("%d routes for %s/%s/%s", len(eps), ns, svcName, svcPort)
+
+	if len(eps) == 0 {
+		return nil, nil
+	}
+
+	// Consider: if there is only a single endpoint, wouldn't it be better to use the cluster IP?
```
szuecs: no because of DNAT/SNAT it's not safe for us to use the service IP

dataclients/kubernetes/clusterclient.go
```diff
@@ -0,0 +1,374 @@
...
+	bToken, err := ioutil.ReadFile(tokenFilePath)
+	if err != nil {
+		return "", err
+	}
+
+	return string(bToken), nil
+}
+
+func newClusterClient(o Options, apiURL, ingCls string, quit <-chan struct{}) (*clusterClient, error) {
+	token, err := readServiceAccountToken(serviceAccountDir+serviceAccountTokenKey, o.KubernetesInCluster)
```
szuecs: we need to update tokens regularly, soon

dataclients/kubernetes/clusterclient.go
```diff
@@ -0,0 +1,374 @@
...
+}
+
+// filterIngressesByClass will filter only the ingresses that have the valid class, these are
+// the defined one, empty string class or not class at all
+func (c *clusterClient) filterIngressesByClass(items []*ingressItem) []*ingressItem {
+	validIngs := []*ingressItem{}
+
+	for _, ing := range items {
+		// No metadata is the same as no annotations for us
+		if ing.Metadata != nil {
```
szuecs: Aren't we already sure we have only valid ingress objects?

dataclients/kubernetes/clusterclient.go
```diff
@@ -0,0 +1,374 @@
...
+			cls, ok := ing.Metadata.Annotations[ingressClassKey]
+			// Skip loop iteration if not valid ingress (non defined, empty or non defined one)
+			if ok && cls != "" && !c.ingressClass.MatchString(cls) {
+				continue
+			}
+		}
+		validIngs = append(validIngs, ing)
```
szuecs: So if we have no metadata we add it anyways (if we do not delete the metadata check, then we have to put this into the if branch.

dataclients/kubernetes/ingress.go
aryszka: yes, on it

dataclients/kubernetes/ingress.go
aryszka: got it.

dataclients/kubernetes/ingress.go
aryszka: https://github.com/zalando/skipper/issues/1287

dataclients/kubernetes/clusterclient.go
aryszka: the point is that it is coming from an external process, even if we expect it to not contain bugs, we cannot be sure, and we don't want to crash our process.

Note that in the case of the route gropus, we are adding now full validation in a single step, but that also comes only after the cluster client has loaded the current route groups.

dataclients/kubernetes/clusterclient.go
aryszka: well, I think that for filtering on ingress classes, this is actually fine. I would like to have though full validation of the ingress objects, in a PR specifically for this. This code here is just being moved from kube.go with the current change. (https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/kube.go#L1246)

dataclients/kubernetes/ingress.go
aryszka: i mean do we want to change this behavior in this PR?

szuecs: yes

dataclients/kubernetes/clusterclient.go
szuecs: I realized it later, that this is basically the earliest validation

dataclients/kubernetes/clusterclient.go
szuecs: ok

dataclients/kubernetes/defaultfilters.go
```diff
@@ -0,0 +1,80 @@
...
+		fs.err = fmt.Errorf("[eskip] default filters: %v", fs.err)
+	}
+
+	fs.parsed = true
+}
+
+func (df defaultFilters) get(serviceID resourceID) ([]*eskip.Filter, error) {
+	fs, ok := df[serviceID]
+	if !ok {
+		return nil, nil
```
szuecs: I guess the nil error is for readability on the user side, right?

dataclients/kubernetes/definitions.go
```diff
@@ -174,20 +193,52 @@ type endpointList struct {
...
 	result := make([]string, 0)
 	for _, s := range ep.Subsets {
 		for _, port := range s.Ports {
+			// TODO: verify if the comparison of port.Name == svcPortName is valid,
+			// considering that the svcPortName is the name of service port specified
+			// in ingress. The right way probably is not to use the svcPortName here,
+			// since that's a reference from the ingress backend to the service, but
+			// use the svcPortTarget, which can be a name or a number, to compare it
+			// with the subset port.Name and port.Port, which is referenced by the
+			// service target port, which can also be either a name or a port value.
```
szuecs: `port.Name == svcPortName` is valid.

dataclients/kubernetes/fixtures_test.go
```diff
@@ -0,0 +1,269 @@
...
+		expectedRoutes, err := eskip.Parse(string(b))
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if !eskip.EqLists(routes, expectedRoutes) {
+			t.Error("Failed to convert the resources to the right routes.")
+			t.Logf("routes: %d, expected: %d", len(routes), len(expectedRoutes))
+			t.Logf("got:\n%s", eskip.String(eskip.CanonicalList(routes)...))
+			t.Logf("expected:\n%s", eskip.String(eskip.CanonicalList(expectedRoutes)...))
```
szuecs: not sure if `cmp.Diff(eskip.String(eskip.CanonicalList(routes)...), eskip.String(eskip.CanonicalList(expectedRoutes)...))` would not help to show the error better

dataclients/kubernetes/hosts.go
```diff
@@ -0,0 +1,70 @@
...
+		hrx[i] = rxDots(h[i])
+	}
+
+	return fmt.Sprintf("^(%s)$", strings.Join(hrx, "|"))
+}
+
+// hostCatchAllRoutes creates catch-all routes for those hosts that only have routes with
+// a Host predicate and at least one additional predicate.
+//
+// currently only used for RG
```
szuecs: RG -> RouteGroup

dataclients/kubernetes/hosts.go
```diff
@@ -0,0 +1,70 @@
...
+func hostCatchAllRoutes(hostRoutes map[string][]*eskip.Route, createID func(string) string) []*eskip.Route {
+	var catchAll []*eskip.Route
+	for h, r := range hostRoutes {
+		var hasHostOnlyRoute bool
+		for _, ri := range r {
+			ct := eskip.Canonical(ri)
+			var hasNonHostPredicate bool
+			for _, p := range ct.Predicates {
+				if p.Name != "Host" {
```
szuecs: Maybe: `&& p.Name != "True"` ?

dataclients/kubernetes/ingress.go
szuecs: We found a bug, if servicePort is used with only referencing the name, the endpoint does not have a port number and you have to lookup from pods, which is a really bad choice in kubernetes.
The bug is not visible to the user, because we fallback to serviceIP if we can not get endpoints with port as number.
The question is, if we should change it anyways and document that we do not completely follow the K8s spec because of safety and efficiency reasons or we just let it like that.
wdyt @aryszka ?

dataclients/kubernetes/ingress.go
```diff
@@ -0,0 +1,775 @@
...
+	}
+
+	return routes
+}
+
+// catchAllRoutes returns true if one of the routes in the list has a catchAll
+// path expression.
+//
+// TODO: this should also consider path types exact and subtree
+func catchAllRoutes(routes []*eskip.Route) bool {
```
szuecs: Maybe `hasCatchAllRoute`

dataclients/kubernetes/ingress.go
```diff
@@ -0,0 +1,775 @@
...
+			}
+		}
+	}
+
+	return false
+}
+
+// convert logs if an invalid found, but proceeds with the
+// valid ones.  Reporting failures in Ingress status is not possible,
+// because Ingress status field is v1.LoadBalancerIngress that only
```
szuecs: it's not v1, but v1beta1

dataclients/kubernetes/routegroup.go
```diff
@@ -0,0 +1,554 @@
...
+	"github.com/zalando/skipper/loadbalancer"
+)
+
+// TODO:
+// - consider catchall for east-west routes
+// - document how route group errors are handled
+// - document in the CRD that the service type must be ClusterIP when using service backends
+// - document the implicit routes, or clarify: spec.routes is not optional, but an example doesn't have any
+// - document the rules and the loopholes with the host catch-all routes
+// - document the behavior of the weight implementation
```
szuecs: do you already documented all of these?
If not do you plan to do this in this PR?

dataclients/kubernetes/routegroup.go
```diff
@@ -0,0 +1,554 @@
...
+
+	return t
+}
+
+func trafficBalance(t *calculatedTraffic) []*eskip.Predicate {
+	if t.balance <= 0 {
+		return nil
+	}
+
+	p := eskip.Predicate{Name: "True"}
```
szuecs: :+1:

dataclients/kubernetes/routegroup.go
```diff
@@ -0,0 +1,554 @@
...
+func storeHostRoute(ctx *routeGroupContext, r *eskip.Route) {
+	for _, h := range ctx.hosts {
+		ctx.hostRoutes[h] = append(ctx.hostRoutes[h], r)
+	}
+}
+
+func appendEastWest(ctx *routeGroupContext, routes []*eskip.Route, current *eskip.Route) []*eskip.Route {
+	// how will the route group name for the domain name play together with
+	// zalando.org/v1/stackset and zalando.org/v1/fabricgateway? Wouldn't it be better to
+	// use the service name instead?
```
szuecs: It should just work with stackset and and fabricgateway. We don't depend on versioned stacks in the traffic switching case, so we also do not have this in RouteGroup nor in the east-west case.

dataclients/kubernetes/routegroup.go
```diff
@@ -0,0 +1,554 @@
...
+	}
+
+	return routes, nil
+}
+
+func transformExplicitGroupRoute(ctx *routeContext) (*eskip.Route, error) {
+	gr := ctx.groupRoute
+	r := &eskip.Route{Id: ctx.id}
+
+	// Path or PathSubtree, prefer Path if we have, becasuse it is more specifc
```
szuecs: becasuse -> because

dataclients/kubernetes/routegroupcrd.go
```diff
@@ -0,0 +1,427 @@
...
+	b.Name = p.Name
+	b.Type = bt
+	b.Address = p.Address
+	b.ServiceName = p.ServiceName
+	b.ServicePort = p.ServicePort
+	b.Algorithm = a
+	b.Endpoints = p.Endpoints
+	b.parseError = perr
+
+	*sb = b
```
szuecs: Is this not a nil ptr deref similar to https://play.golang.org/p/dT7-sRilS1y

docs/kubernetes/routegroups.md
```diff
@@ -0,0 +1,628 @@
...
+
+The definition file of the CRD can be found as part of Skipper's source code, at:
+
+https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/deploy/apply/routegroups_crd.yaml
+
+To install it manually in a cluster, assuming the current directory is the root of Skipper's source, call this
+command:
+
+```
+kubectl apply -f dataclients/kubernetes/deploy/apply/routegroups_crd.yaml
```
szuecs: % kubectl apply -f https://raw.githubusercontent.com/zalando/skipper/feature/routegroup-crd/dataclients/kubernetes/deploy/apply/routegroups_crd.yaml
customresourcedefinition.apiextensions.k8s.io/routegroups.zalando.org created

docs/kubernetes/routegroups.md
```diff
@@ -0,0 +1,628 @@
...
+
+### type=lb
+
+This backend provides load balancing between multiple network endpoints. Keep in mind that the service type
+backend automatically generates load balanced routes for the service endpoints, so this backend type typically
+doesn't need to be used for services.
+
+### type=network
+
+This backend type results in routes that forward incoming requests to the defined network address, regardless of
```
szuecs: forward -> proxy

The problem is that forward is not equal to proxy. As an example, think about NLB. NLB is a packet forwarder, such that the client IP is the real IP, not the NLB. This would be miss leading in the context for people that know more network or load balancers

docs/kubernetes/routegroups.md
```diff
@@ -0,0 +1,628 @@
...
+### type=shunt, type=loopback, type=dynamic
+
+These backend types allow advanced routing setups. Please check the [reference
+manual](../../reference/backends/) for more details.
+
+## Default Backends
+
+- *[Format](../routegroup-crd/#backend-reference)*
+
+A default backend is a references to one of the defined backends. When a route doesn't specify which backend(s)
```
szuecs: a references -> a reference

docs/kubernetes/routegroups.md
```diff
@@ -0,0 +1,628 @@
...
+to use, the ones referenced in the default backends will be used.
+
+In case there are no individual routes at all in the route group, a default set of routes (one or more) will be
+generated and will forward the incoming traffic to the default backends.
```
szuecs: forward -> proxy

docs/kubernetes/routegroups.md
```diff
@@ -0,0 +1,628 @@
...
+
+The reason, why multiple backends can be referenced as default, is that this makes it easy to execute gradual
+traffic switching between different versions, even more than two, of the same application. [See
+more](#gradual-traffic-switching).
+
+## Routes
+
+- *[Format](../routegroup-crd/#route_1)*
+
+Routes define where to and how the incoming requests will be forwarded. The predicates, including the path,
```
szuecs: forwarded -> proxied

docs/kubernetes/routegroups.md
```diff
@@ -0,0 +1,628 @@
...
+pathSubtree, pathRegexp and methods fields, and any free-form predicate listed under the predicates field,
+control which requests are matched by a route, the filters can apply changes to the forwarded requests and the
+returned responses, and the backend refs, if defined, override the default backends, where the requests will be
+forwarded to. If a route group doesn't contain any explicit routes, but it contains default backends, a default
```
szuecs: forwarded -> proxied

docs/kubernetes/routegroups.md
```diff
@@ -0,0 +1,628 @@
...
+  - pathSubtree: /
+    predicates:
+    - Header("X-Forwarded-Proto", "http")
+    filters:
+    - redirectTo(302, "https:")
+    backends:
+    - redirectShunt
+```
+
+### zalando.org/skipper-loadbalancer
```
szuecs: Maybe link to https://opensource.zalando.com/skipper/reference/backends/#load-balancer-backend

docs/kubernetes/routegroups.md
```diff
@@ -0,0 +1,628 @@
...
+    type: service
+    serviceName: my-service
+    servicePort: 80
+    algorithm: consistentHash
+```
+
+### zalando.org/skipper-ingress-path-mode
+
+The route objects support the different path lookup modes, by using the path, pathSubtree or the
+pathRegexp field. See also the [route matching](http://localhost:8000/reference/architecture/#route-matching)
```
szuecs: Please fix the link to not localhost :)

szuecs:
Great work thanks!
I have only a few comments that should be addressed.

dataclients/kubernetes/api_test.go
aryszka: done

dataclients/kubernetes/definitions.go
aryszka: done

dataclients/kubernetes/ingress.go
aryszka: we have to postpone this, the changes are stored in this branch:

https://github.com/zalando/skipper/tree/improvement/service-port-fallback-handling

the majority of the kubernetes tests are not written with endpoints and target ports in mind. This means that a huge chunk of the tests would need to be changed, which makes the whole PR quite unsafe

dataclients/kubernetes/ingress.go
aryszka: I think we should change it, but not in this PR, because of the additional testing needs. Currently, we fall back to using the service cluster IP and service port, which will cover those cases when the target port is referenced by value, however, will generate useless routes for those services that don't have endpoints, e.g. because of the deployment being scaled down to 0. In the future solution, we will need to distinguish between these two cases. Branch with a possible improvement can be found here: https://github.com/zalando/skipper/tree/improvement/service-port-fallback-handling

dataclients/kubernetes/defaultfilters.go
aryszka: not sure I completely get the question. Returning nil, nil means that we don't have default filters and everything is fine.

dataclients/kubernetes/definitions.go
aryszka: related to other comments and this branch: https://github.com/zalando/skipper/tree/improvement/service-port-fallback-handling

dataclients/kubernetes/defaultfilters.go
szuecs: thanks

dataclients/kubernetes/hosts.go
aryszka: not necessary, because then that would act as a catch-all, so in worst case we generated an unnecessary catch-all route. I'd rather wouldn't scatter the code with the True predicate, since I consider it rather a diagnostic tool, even if we use it currently for the traffic control. For the valid use cases, we need to find better solutions.

dataclients/kubernetes/routegroup.go
aryszka: yes, removing these todo items.

dataclients/kubernetes/routegroupcrd.go
aryszka: https://godoc.org/encoding/json#Unmarshal

the doc can be confusing, because the input arg cannot be nil, but the fields can be nil. I can add here some guards, but there's on reasonable way to test it :)

docs/kubernetes/routegroups.md
aryszka: good to know that this is possible with apply -f, but i would not promote it in the docs. Note that we may new releases, while some want to use the edge from master, etc.

dataclients/kubernetes/hosts.go
szuecs: Ok

docs/kubernetes/routegroups.md
aryszka: done, thx

docs/kubernetes/routegroups.md
aryszka: good point, done

docs/kubernetes/routegroups.md
aryszka: thx, done

dataclients/kubernetes/routegroupcrd.go
aryszka: added a guard

dataclients/kubernetes/routegroup.go
aryszka: done

dataclients/kubernetes/ingress.go
aryszka: thx, done

dataclients/kubernetes/ingress.go
aryszka: right, done

dataclients/kubernetes/fixtures_test.go
aryszka: added diff, but I left the current one, too, because I found it useful to have an overview while working on these tests, as sometimes the diff was too big.

szuecs:
:+1:

aryszka:
:+1:

universam1:
@szuecs seeing errors probably when RouteGroups are not used:
````
 
level=error msg="Error while checking known resource types: resource not found."
```

Issue #1040, opened 2019-04-23
Report Failures to generate routes to Ingress


arjunrn:
**Is your feature request related to a problem? Please describe.**
Currently when the service or ingress objects are misconfigured there is no way to report those issues back to the user other than reading the logs. 

**Describe the solution you would like**
The ingress object itself does not support anything other than the loadbalancer names in it's status. So when backend routes cannot be correctly generated then the load balancer should not be set. Or there is a separate resource called `IngressStatus` or something like that where skipper could report configuration issues.

**Would you like to work on it?**
Yes

szuecs:
Maybe we can use Kubernetes events, such that it would show up in `describe ingress`?

arjunrn:
We could but IMO events are not the right choice because there would be a lot of events, every time the ingress is processed and this would simply create a lot of noise. So I suggested a new object which could be used to report the status. But if that's excessive then events should be sufficient.

PR #1181, opened 2019-09-06
Use Errorf when port defition is wrong

ruudk:
I spend 2 hours yesterday fighting with Skipper because it didn't log this error:
```
Failed to find target port for service xx, fallback to service: GetTargetPort: target port not found [xx 80 http] given http
```

Only after enabling debug mode I discovered it.

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -681,7 +681,7 @@ func (c *Client) convertPathRule(
 	targetPort, err := svc.GetTargetPort(svcPort)
 	if err != nil {
 		// fallback to service, but service definition is wrong or no pods
-		log.Debugf("Failed to find target port for service %s, fallback to service: %v", svcName, err)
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

szuecs:
Thanks for the PR. It can create some noise, because people might have leftovers in the cluster, but I think it's a valid point to make it easier to understand these kind of problems.
Best would be to report this on the ingress as event, such that describe ingress would show it: https://github.com/zalando/skipper/issues/1040

njuettner:
👍

ruudk:
Thanks for merging this. Yes, the noise will be good. If you are annoyed by it, you should fix it, it's a bad configuration anyway :)

szuecs:
If you have 1-10 clusters fine, we run 150. It’s just to consider. Sometimes you get users that clean only 50% of the things on tear down....

PR #1080, opened 2019-05-31
Remove X-Zalando-team from open source skipper

addityasingh:
close #1079 

Remove X-Zalando-team from open source skipper

Issue #1079, opened 2019-05-31
Remove X-Zalando-team from open source skipper


addityasingh:
Remove X-Zalando-team from open source skipper 
https://github.com/zalando/skipper/blob/3c504ec104ddc6f3171bc79e575b0c4a2cc3c0c7/.zappr.yaml#L9

addityasingh:
Closing as this is not valid anymore

PR #1184, opened 2019-09-10
chore: Update team name in zappr

addityasingh:
Refer #1079, #1080

Update owning team for skipper in zappr

### DIFF
--- a/.zappr.yaml
+++ b/.zappr.yaml
@@ -6,4 +6,4 @@ approvals:
       from:
         orgs:
           - "zalando"
-X-Zalando-Team: "pathfinder"
+X-Zalando-Team: "teapot"


oporkka:
:+1:

ruiaraujo:
:+1:

mikkeloscar:
@szuecs @aryszka are we ready for this? Biggest "impact" as I see it would be the docker image name changes, which people need to be aware of in new releases.

hjacobs:
:+1:

mikkeloscar:
Since this PR is from a fork, we can't verify that it works on CDP (I expect the docker push to fail because of the team change, like @szuecs pointed out about the docker image renaming).

@addityasingh do you have access to push a branch directly here? Then we could test this on CDP in the PR already.

szuecs:
We can also take over this PR to test and change everything which is required. Right now all docs are pointing to the wrong image. This might also be coordinated with PRs to kube-ingress-aws-controller.

addityasingh:
>@addityasingh do you have access to push a branch directly here? Then we could test this on CDP in the PR already.
I don't

szuecs:
:+1:

PR #1186, opened 2019-09-11
doc: add missing annotation and some minor fixes in docs

szuecs:
doc: add missing annotation
doc: change "" to backticks to render more nicely
doc: separate path mode options with | instead of /, because we use the same pattern elsewhere in the cli docs


szuecs:
:+1:

PR #1192, opened 2019-09-20
monitor route creation time

zeitlinger:
cleanup of testing logic in route creation

### DIFF
--- a/filters/builtin/creationmetrics.go
+++ b/filters/builtin/creationmetrics.go
@@ -38,7 +38,7 @@ func (m *RouteCreationMetrics) reportRouteCreationTimes(routes []*routing.Route,
 	for _, r := range routes {
 		for origin, start := range m.startTimes(r) {
 			if m.initialized {
-				m.metrics.MeasureFloat(metricsPrefix+origin, now.Sub(start).Seconds()/100)
+				m.metrics.MeasureDuration(metricsPrefix+origin, now.Sub(start)/100)
 			}
 		}
 	}
--- a/filters/builtin/creationmetrics.go
+++ b/filters/builtin/creationmetrics.go
@@ -31,14 +31,14 @@ func NewRouteCreationMetrics(metrics filters.Metrics) *RouteCreationMetrics {
 
 // Do implements routing.PostProcessor and records the filter creation time.
 func (m *RouteCreationMetrics) Do(routes []*routing.Route) []*routing.Route {
-	return m.reportRouteCreationTimes(routes, time.Now())
+	return m.reportRouteCreationTimes(routes)
 }
 
-func (m *RouteCreationMetrics) reportRouteCreationTimes(routes []*routing.Route, now time.Time) []*routing.Route {
+func (m *RouteCreationMetrics) reportRouteCreationTimes(routes []*routing.Route) []*routing.Route {
 	for _, r := range routes {
 		for origin, start := range m.startTimes(r) {
 			if m.initialized {
-				m.metrics.MeasureDuration(metricsPrefix+origin, now.Sub(start)/100)
+				m.metrics.MeasureSince(metricsPrefix+origin, start)
 			}
 		}
 	}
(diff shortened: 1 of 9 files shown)


metrics/codahale.go
```diff
@@ -100,10 +103,24 @@ func (c *CodaHale) updateTimer(key string, d time.Duration) {
...
+		t.Update(int64(value))
+	}
+}
+
 func (c *CodaHale) MeasureSince(key string, start time.Time) {
 	c.measureSince(key, start)
 }
 
+func (c *CodaHale) MeasureFloat(key string, value float64) {
+	go c.updateHistogram(key, value)
```
szuecs: why do you spawn a goroutine?

filters/builtin/creationmetrics.go
```diff
@@ -31,10 +31,14 @@ func NewRouteCreationMetrics(metrics filters.Metrics) *RouteCreationMetrics {
...
 func (m *RouteCreationMetrics) Do(routes []*routing.Route) []*routing.Route {
+	return m.reportRouteCreationTimes(routes, time.Now())
+}
+
+func (m *RouteCreationMetrics) reportRouteCreationTimes(routes []*routing.Route, now time.Time) []*routing.Route {
 	for _, r := range routes {
 		for origin, start := range m.startTimes(r) {
 			if m.initialized {
-				m.metrics.MeasureSince(metricsPrefix+origin, start)
+				m.metrics.MeasureFloat(metricsPrefix+origin, now.Sub(start).Seconds()/100)
```
szuecs: Can't you shift the duration instead of converting to float?
See https://play.golang.org/p/eXWe0zFXYge

metrics/codahale.go
zeitlinger: that was the same as in measure since

filters/builtin/creationmetrics.go
zeitlinger: done

zeitlinger:
:+1:

metrics/codahale.go
szuecs: Where do we call MeasureFloat ?
If this is not in the hot path of the proxy package, then I would say drop the goroutine.

zeitlinger:
I dropped measure float as suggested.

metrics/codahale.go
```diff
@@ -104,6 +104,10 @@ func (c *CodaHale) MeasureSince(key string, start time.Time) {
 	c.measureSince(key, start)
 }
 
+func (c *CodaHale) MeasureDuration(key string, duration time.Duration) {
+	go c.updateTimer(key, duration)
```
szuecs: The goroutine @zeitlinger

zeitlinger: OK, removed it - the goroutine in measure since is used for `MeasureFilterResponse` (the critical proxy path), so I left it.

filters/builtin/creationmetrics.go
```diff
@@ -31,10 +31,14 @@ func NewRouteCreationMetrics(metrics filters.Metrics) *RouteCreationMetrics {
...
 func (m *RouteCreationMetrics) Do(routes []*routing.Route) []*routing.Route {
+	return m.reportRouteCreationTimes(routes, time.Now())
+}
+
+func (m *RouteCreationMetrics) reportRouteCreationTimes(routes []*routing.Route, now time.Time) []*routing.Route {
 	for _, r := range routes {
 		for origin, start := range m.startTimes(r) {
 			if m.initialized {
-				m.metrics.MeasureSince(metricsPrefix+origin, start)
+				m.metrics.MeasureDuration(metricsPrefix+origin, now.Sub(start)/100)
```
tkrop: I find a factor of `100` here very confusing. If we do any factor, it should be `1000`. However, I do not see that we gain a real value by this change. It make the values in the buckets intransparent and it actually violates the contract of the measurement bucket, which is to receive values in seconds.

szuecs: I agree that it might be confusing, maybe 1000 is better than 100.
It's a workaround to shift to buckets, which are more fine grained.
If we want to drop the workaround we would need to increase the buckets (IIRC the step is 60s,120s, so 80s would be in one of these buckets and not counted individually), which costs, memory. Skipper memory in our setup is fine, but Prometheus is already suffering.
Maybe you can enhance the apiMonitoring metrics in general to have them in a separate keyspace.

zeitlinger:
:+1:

zeitlinger:
@szuecs this is just a refactoring - can you approve it?

aryszka:
:+1:

PR #1193, opened 2019-09-22
delete routes only on ENOEXIST when using file watching

aryszka:
delete routes only on ENOEXIST when using file watching


szuecs:
:+1:

aryszka:
:+1:

PR #1197, opened 2019-09-30
set origin markers only when there were any changes to the routes

aryszka:
set the origin marker after the diff of the routes was taken. (Currently the originMarkers are added to a different route on every poll, therefor an invalid update is detected.)


zeitlinger:
:+1:

zeitlinger:
thanks for spotting this one!

szuecs:
:+1:

PR #1202, opened 2019-10-02
Note that inlineContent filter is special

mikkeloscar:
Note in the docs that `inlineContent` filter is special and must be last in the chain.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #1205, opened 2019-10-02
Using YAML configuration file to pass args for skipper binary

### LINKED ISSUE #1039: Add option to pull cli args/switches from configfile
There are numerous command line args for skipper, if you have a complex deployment the command line arg gets very long.

Add an option(yes I get the irony) like `-configfile=/path/to/skipper.yaml`.  Doesn't have to be yaml, just an example.  In this file you can define all the switches and arg values.

Right now I just put all the switches and args on a single line in my systemd service definition file.

None.

No.  I don't know the language well enough yet to be of any use in this regard.

ghophp:
Hello :) as a good first issue to work on Skipper I took the chance to work on the issue #1039 which proposed to introduce a `yaml` to pass the long list of flags that are possible to configure the skipper command.

I started by evaluating the possible formats that we could use to receive the parameters and used as reference links like:
- https://www.zionandzion.com/json-vs-xml-vs-toml-vs-cson-vs-yaml/
- https://medium.com/@robmuh/yaml-has-won-ba5dae37e740
- https://gohugohq.com/howto/toml-json-yaml-comparison/

After much deliberation and back and forth on how to implement such a feature with compatibility to the current flags binding, my main reasoning to chose `yaml` was: skipper already has a way of receiving configurations (flags), so functionality-wise it is delivering, the only aspect that is missing is readability when providing such arguments, and since the strongest format for readability is `yaml`, `yaml` was chosen.

I encourage also the maintainers and other members to discuss if we want to have another format. Since now this is implemented, it is not complicated to replace the format that we will use, it is only a matter of a bit of refactoring on the method receivers and tests. In resume, do not feel attached to the format, I just wanted to check the feasibility of it, and already come with a concrete solution that can be discussed.

About the implementation itself, basically the main changes are focused on the separation of concerns from the main file to a `Config` struct aimed to encapsulate the logic for parsing the arguments from flags or with a config file. The method `NewConfig` that created the config struct binds the flags to the struct, and the `Parse` method is the only one that can return an error in case of parsing anything wrong. After that, you can get the configuration as `skipper.Options` by calling `ToOptions`. This made the main method very clean and easy to follow.

I introduced the necessary tests and even took the chance to cover the `pluginFlag` struct with tests. Correct some link that was broken on the docblock of a field, and extended the documentation on two sections, please feel free to suggest other sections that should be updated.

Looking forward to your feedback. closes #1039


szuecs:
Thanks for the PR @ghophp , I add @aryszka to it to check if he also agrees on the format and idea.
I see a lot of value in this and I tried to get some people from the community to check it, too.
My code review is ongoing, but from 7 / 23 files, looks good to me.

aryszka:
Hi @ghophp ,

thanks for the PR!

Regarding the format: I think YAML is a good choice because of wide adoption, stability of the tooling, and the readability argument that you pointed out. Even if it's not my favorite, it's very reasonable to support it. For the alternatives, I believe that we can support other formats in the future as an addition instead of a replacement.

aryszka:
This looks nice after a quick glance. I am still reviewing it, but I'd like to point out a problem that I see with the approach: currently the values in the config file override the according values set in the flags. E.g, having the following config file:

```
address: :9999
```

and starting skipper with the following flags:

```
bin/skipper -config-file config.yaml -address :8888
```

skipper will take :9999 as the address. To me this feels somewhat unexpected. What I would generally expect is that I can set the main preferences in the config file, and in special cases, I can provide an overriding flag but only for a specific values.

What do you think, does this make sense?

ghophp:
@aryszka thanks for the feedback, now that I visualized the case with the config and flags together indeed I agree that it makes more sense to have the flags having more weight than the configuration file. I will check how to implement the desired behaviour with a clean code 👍 commit will follow

ghophp:
@aryszka gladly the solution is quite simple to give priority to the flags, the `flag.Parse` actually only overwrite the values if provided, meaning that if I do provide an address on the config, and reparse the flags but address is not provided, the value of the variable will not be changed. I adjusted the test for it, and you can observe the correct behaviour there.

please let me know if anything else 👍

ghophp:
@aryszka can you retrigger the PR build? this test seems flaky

szuecs:
@ghophp done, we will take care about the travis ci test run

szuecs:
What about https://play.golang.org/p/P0zG9JyqCYw for parsing once and reading config file before flags?

aryszka:
while the approach in the playfield is also mostly fine, I rather prefer using the flag.Parse approach from this two. I'll finish my review today. Cheers

cmd/skipper/config_test.go
```diff
@@ -0,0 +1,155 @@
...
+			os.Args = tt.args
+
+			err := cfg.Parse()
+			if (err != nil) != tt.wantErr {
+				t.Errorf("config.NewConfig() error = %v, wantErr %v", err, tt.wantErr)
+			}
+
+			if !tt.wantErr {
+				if cmp.Equal(cfg, tt.want, cmp.AllowUnexported(listFlag{}, pluginFlag{}, defaultFiltersFlags{})) == false {
+					t.Errorf("config.NewConfig() got v, want v, %v", cmp.Diff(cfg, tt.want))
```
szuecs: -> `t.Errorf("config.NewConfig() got vs. want:\n%v", cmp.Diff(cfg, tt.want))`

cmd/skipper/listflag.go
```diff
@@ -36,16 +36,40 @@ func (lf *listFlag) Set(value string) error {
...
+	lf.values = values
+
+	if err := lf.hasOnlyAllowedValues(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (lf *listFlag) hasOnlyAllowedValues() error {
```
szuecs: The name is not good, because functions with `has` prefix should return bool.

cmd/skipper/ratelimiterflags.go
```diff
@@ -20,15 +20,11 @@ const enableRatelimitUsage = `enable ratelimit`
 
 type ratelimitFlags []ratelimit.Settings
 
-var errInvalidRatelimitConfig = errors.New("invalid ratelimit config")
+var errInvalidRatelimitConfig = errors.New("invalid ratelimit config (allowed values are: local, client, service or disabled)")
```
szuecs: please drop the `local` it's deprecated anyways

skipper.go
```diff
@@ -479,7 +479,7 @@ type Options struct {
 	// FilterPlugins loads additional filters from modules. The first value in each []string
 	// needs to be the plugin name (as on disk, without path, without ".so" suffix). The
 	// following values are passed as arguments to the plugin while loading, see also
-	// https://zalando.github.io/skipper/plugins/
+	// https://opensource.zalando.com/skipper/reference/plugins/
```
szuecs: :+1:

cmd/skipper/listflag.go
ghophp: what about rename to `validate() error`?

szuecs: fine for me

cmd/skipper/defaultfilterflags.go
```diff
@@ -20,10 +20,6 @@ func (dpf defaultFiltersFlags) String() string {
 }
 
 func (dpf *defaultFiltersFlags) Set(value string) error {
-	if dpf == nil {
```
aryszka: :heart:

aryszka:
:+1:

aryszka:
looks good to me, thanks!

szuecs:
:+1:

Issue #1210, opened 2019-10-07
PathRegexp() problem with trailing slashes in the path


aryszka:
**Describe the bug**
When using PathRegexp with ignore trailing slash option, it doesn't match the path when the regexp enforces the trailing slash.

**To Reproduce**

1. start skipper with: `skipper -ignore-trailing-slash -inline-routes 'PathRegexp("^/foo/bar/baz-[0-9-]+/$") -> status(200) -> <shunt>'` 
2. make request like: `curl localhost:9090/foo/bar/baz-42-0/`

**Expected behavior**

The status should be 200, as the request path matches the expression in the predicate.

**Observed behavior**

The status is 404.

aryszka:
fixed in #1211

PR #1211, opened 2019-10-07
keep the trailing slash in the path for regexp matching

aryszka:
Fixing: #1210 

### DIFF
--- a/routing/matcher.go
+++ b/routing/matcher.go
@@ -11,8 +11,9 @@ import (
 )
 
 type leafRequestMatcher struct {
-	r    *http.Request
-	path string
+	r         *http.Request
+	path      string
+	exactPath string
 }
 
 func (m *leafRequestMatcher) Match(value interface{}) (bool, interface{}) {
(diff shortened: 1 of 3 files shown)


NorthFury:
:+1:

routing/matcher.go
```diff
@@ -326,7 +315,8 @@ func newMatcher(rs []*Route, o MatchingOptions) (*matcher, []*definitionError) {
 		}
 
 		if r.pathSubtree != "" {
-			path, _ := cleanPath(r.pathSubtree, o|IgnoreTrailingSlash)
+			path := httppath.Clean(r.pathSubtree)
+			path = trimTrailingSlash(path)
```
aryszka: don't we need to use the `if o.ignoreTrailingSlash()` condition here?

aryszka: oops.. my bad for not watching :man_facepalming:

aryszka:
:+1:

routing/matcher.go
NorthFury: `(o|IgnoreTrailingSlash).ignoreTrailingSlash()` is always true

szuecs:
:+1:

PR #1218, opened 2019-10-09
Clarify that flowIds must be of a certain format to be reusable

mikkeloscar:
Clarify that flowIds must be of a certain format to be reusable

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -264,8 +264,9 @@ Sets an X-Flow-Id header, if it's not already in the request.
 This allows you to have a trace in your logs, that traces from
 the incoming request on the edge to all backend services.
 
-Flow IDs must be in a [certain format][flowid_validation] to be reusable in
-skipper.
+Flow IDs must be in a certain format to be reusable in skipper. Valid formats
+depend on the generator used in skipper. Default generator creates IDs of
+length 16 matching the following regex: `^[0-9a-zA-Z+-]+$`
 
 Paramters:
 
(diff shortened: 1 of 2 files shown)


docs/reference/filters.md
```diff
@@ -276,6 +279,8 @@ Example:
 * -> flowId("reuse") -> "https://some-backend.example.org";
 ```
 
+[flowid_validation]: https://github.com/zalando/skipper/blob/b2d74584a5753b21034be452743ae242f86fd513/filters/flowid/ulid.go#L55-L58
```
szuecs: from the code pov, there are 3 conditions, but only the regex is important.
Regexp says `^[0123456789ABCDEFGHJKMNPQRSTVWXYZ]{26}$`, which means `^[0-9A-Z]{26}$`, so numbers and upper ascii alphabet with exact 26 characters. Maybe the len check is done to make sure we have no slow regex, but then we could do `len(fid) == 26 && regexCheck`

mikkeloscar: Changed PTAL

mikkeloscar: I was even wrong in which generator is used by default. Now it should be more clear and correct in the docs, thanks!

mikkeloscar: Could not simplify the regex to `^[0-9A-Z]{26}$ ` because the original doesn't have a `U` in the list of characters and tests are failing with that change. I reverted that part.

szuecs: ah oh :)

szuecs:
:+1:

aryszka:
:+1:

docs/reference/filters.md
```diff
@@ -264,10 +264,14 @@ Sets an X-Flow-Id header, if it's not already in the request.
 This allows you to have a trace in your logs, that traces from
 the incoming request on the edge to all backend services.
 
+Flow IDs must be in a certain format to be reusable in skipper. Valid formats
+depend on the generator used in skipper. Default generator creates IDs of
+length 16 matching the following regex: `^[0-9a-zA-Z+-]+$`
+
 Paramters:
```
jjst: Typo: Parameters

szuecs: https://github.com/zalando/skipper/pull/1228

PR #1219, opened 2019-10-09
trim origin markers after evaluated

aryszka:
the origin markers are used to measure the time when the incoming routes were applied but they don't change the request flow. Therefore, it is possible to remove them after the metrics were updated.


szuecs:
:+1:

aryszka:
:+1:

PR #1220, opened 2019-10-09
Feature to include additional data to the access log

### LINKED ISSUE #1209: Add ability to expand access log
Structured JSON logs are great to analyise Skipper usage. But it would be even better if there was a way to extend what the structured access log actually logs.

I'd like some sort of _hook_ to provide an override to what an accessLog entry contains.

Adding a filter that does the type of accessLog that I want, but I'm not sure how to properly do that so it works properly.

Yes/No/Yes, but no time

xavivars:
Fixes #1209.

### DIFF
--- a/logging/access.go
+++ b/logging/access.go
@@ -113,7 +113,8 @@ func stripQueryString(u string) string {
 }
 
 // Logs an access event in Apache combined log format (with a minor customization with the duration).
-func LogAccess(entry *AccessEntry) {
+// Additional allows to provide extra data that may be also logged, depending on the specific log format.
+func LogAccess(entry *AccessEntry, additional map[string]interface{}) {
 	if accessLog == nil || entry == nil {
 		return
 	}
@@ -151,7 +152,7 @@ func LogAccess(entry *AccessEntry) {
 		auditHeader = entry.Request.Header.Get(logFilter.UnverifiedAuditHeader)
 	}
 
-	accessLog.WithFields(logrus.Fields{
+	logData := logrus.Fields{
 		"timestamp":      ts,
 		"host":           host,
 		"method":         method,
@@ -165,5 +166,11 @@ func LogAccess(entry *AccessEntry) {
 		"duration":       duration,
 		"flow-id":        flowId,
 		"audit":          auditHeader,
-	}).Infoln()
+	}
+
+	for k, v := range additional {
+		logData[k] = v
+	}
+
+	accessLog.WithFields(logData).Infoln()
 }
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1251,7 +1251,9 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				Duration:     time.Since(ctx.startServe),
 			}
 
-			logging.LogAccess(entry)
+			additionalData, _ := ctx.stateBag[al.AccessLogAdditionalDataKey].(map[string]interface{})
+
+			logging.LogAccess(entry, additionalData)
 		}
 	}()
 
--- a/filters/accesslog/control.go
+++ b/filters/accesslog/control.go
@@ -11,6 +11,9 @@ const (
 
 	// AccessLogEnabledKey is the key used in the state bag to pass the access log state to the proxy.
 	AccessLogEnabledKey = "statebag:access_log:proxy:enabled"
+
+	// AccessLogAdditionalDataKey is the key used in the state bag to pass extra data to access log
+	AccessLogAdditionalDataKey = "statebag:access_log:additional"
 )
 
 // Common filter struct for holding access log state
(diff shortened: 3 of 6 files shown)


logging/access.go
```diff
@@ -151,7 +151,7 @@ func LogAccess(entry *AccessEntry) {
 		auditHeader = entry.Request.Header.Get(logFilter.UnverifiedAuditHeader)
 	}
 
-	accessLog.WithFields(logrus.Fields{
+	nativeData := map[string]interface{}{
```
szuecs: Can this be of type `logrus.Fields` such that we do not need to copy the nativeData in a loop?

logging/access.go
```diff
@@ -165,5 +165,17 @@ func LogAccess(entry *AccessEntry) {
 		"duration":       duration,
 		"flow-id":        flowId,
 		"audit":          auditHeader,
-	}).Infoln()
+	}
+
+	logData := make(logrus.Fields)
+
+	for k, v := range metadata {
+		logData[k] = v
```
szuecs: this can be then copied into the `logrus.Fields` typed `map[string]interface{}`

proxy/proxy.go
```diff
@@ -1184,7 +1184,12 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				Duration:     time.Since(ctx.startServe),
 			}
 
-			logging.LogAccess(entry)
+			extraData := ctx.stateBag[al.AccessLogMetadataKey]
+			if extraData == nil {
+				extraData = make(map[string]string)
+			}
+
+			logging.LogAccess(entry, extraData.(map[string]string))
```
szuecs: I would rather do:
```
extraData := ctx.stateBag[al.AccessLogMetadataKey]  // can be nil is fine
logging.LogAccess(entry, extraData)
```

logging/access.go
```diff
@@ -113,7 +113,7 @@ func stripQueryString(u string) string {
 }
 
 // Logs an access event in Apache combined log format (with a minor customization with the duration).
-func LogAccess(entry *AccessEntry) {
+func LogAccess(entry *AccessEntry, metadata map[string]string) {
```
szuecs: ->
`func LogAccess(entry *AccessEntry, metadata logrus.Fields) {`

Maybe rename `metadata` to `additional`

logging/access.go
xavivars: I was just trying to avoid any additional data overriding the builtin one. But now I think it makes sense.

logging/access.go
xavivars: Wouldn't change here `metadata` type to `logrus.Fields`, because I don't think it's a good idea expose the internals of the internal logger library used to places like proxy.  Changing name to `additional` makes a lot of sense

proxy/proxy.go
xavivars: If we do this, we're saying type of `extraData` is `interface{}`. I don't really like that idea...

szuecs: logrus.Fields is also map[string]interface{}, so why converting everything to map[string]string and later pass it into map[string]interface{}?
This will either be changed in the compiler, if it's smart enough or slow down the performance in the critical path.

logging/access.go
szuecs: Then please do map[string]interface{}

proxy/proxy.go
xavivars: I've changed it to map[string]interface{}. I just didn't want to use `logrus.Fields`. But if you think that's a better approach, I'm happy to do it.

xavivars:
@szuecs Added some documentation for using the StateBag key.

I'd love feedback on where and what's the documentation missing, if any. 

Thanks!

logging/access.go
```diff
@@ -113,7 +113,7 @@ func stripQueryString(u string) string {
 }
 
 // Logs an access event in Apache combined log format (with a minor customization with the duration).
-func LogAccess(entry *AccessEntry) {
+func LogAccess(entry *AccessEntry, additional map[string]interface{}) {
```
szuecs: Godoc missing here

proxy/proxy.go
```diff
@@ -1184,7 +1184,9 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				Duration:     time.Since(ctx.startServe),
 			}
 
-			logging.LogAccess(entry)
+			additionalData, _ := ctx.stateBag[al.AccessLogAdditionalDataKey].(map[string]interface{})
```
szuecs: Do we need to cast? As far as I remember, it’s already the type map[string]interface{}.
On my phone so can not look easily 😇

xavivars: No, the "value" of the StateBag is `interface{}`, so if not casted, compilation fails.

go.mod
```diff
@@ -81,3 +81,5 @@ require (
 	launchpad.net/gocheck v0.0.0-20140225173054-000000000087 // indirect
 	layeh.com/gopher-json v0.0.0-20180103211521-1aab82196e3b
 )
+
+go 1.13
```
szuecs: please rebase on master, it's already there https://github.com/zalando/skipper/blob/master/go.mod#L49

proxy/proxy.go
szuecs: yes, true you are right!

szuecs:
From my side LGTM only `rebase master` needed.

xavivars:
@szuecs: rebase (+ squash) done

szuecs:
:+1:

aryszka:
:+1:

aryszka:
Thanks! :smile:

PR #1221, opened 2019-10-15
remove reporting deprecated filter from startup

aryszka:
removing this warning, because it's unactionable for those users who don't compile the binary on their own. When the filter is actually referenced in the route config, a warning still will be logged.


szuecs:
:+1:

aryszka:
:+1:

PR #1227, opened 2019-10-16
Allow wildcard name collision

NorthFury:
Changes:
* removed `VizTree` implementation because it was broken by this change, it is very unlikely somebody is using it, and as such not worth the effort of fixing it (I may put the effort if required)
* removed wildcard names from the `pathmux.Tree` and moved the responsibility to the matcher
  ** the wildcard param names are now stored on the leaf matcher
* cleaned up the high complexity code related to `PathSubtree` when building the tree, in part present in order to deal with wildcard name conflicts

As a side benefit, I believe this change improves the performance when building the tree, both in terms of memory usage and CPU utilization. I haven't benchmarked this.


pathmux/viztree.go
```diff
@@ -6,162 +6,9 @@ type VizTree struct {
...
-	}
-	return leafWildcardNames
-}
-
-func sliceLeafNames(names []string) []string {
-	if len(names) < 2 {
-		return nil
-	}
-	return names[:len(names)-1]
+	panic("note implemented")
```
jeremycolin:
```suggestion
	panic("not implemented")
```

routing/matcher.go
```diff
@@ -96,7 +90,7 @@ func (err *definitionError) Error() string {
 }
 
 // rx identifying the 'free form' wildcards at the end of the paths
-var freeWildcardRx = regexp.MustCompile("/[*][^/]+$")
+var freeWildcardRx = regexp.MustCompile("/[*][^/]*$")
```
szuecs: This means, that `Path("/*")` is fine (unnamed wildcards), right?

routing/matcher.go
```diff
@@ -268,101 +267,28 @@ func addTreeMatchers(pathTree *pathmux.Tree, matchers map[string]*pathMatcher, s
...
-		}
-
-		for _, l := range pm.leaves {
-			l.subtreeMergeControl.ignoreSubtreeRoot = true
-			l.subtreeMergeControl.subtreeRoot = subRoot
-			l.subtreeMergeControl.mapParamNameFrom = fromParam
-			l.subtreeMergeControl.mapParamNameTo = pm.freeWildcardParam
+func addSubtreeLeafsToPath(pms map[string]*pathMatcher, path string, l *leafMatcher, o MatchingOptions) {
+	basePath := freeWildcardRx.ReplaceAllLiteralString(path, "")
+	basePath = strings.TrimSuffix(basePath, "/")
```
szuecs: I am not sure if the function works like expected. Maybe yes, maybe not.

https://play.golang.org/p/_tTUtfhO5q1
```
	freeWildcardRx := regexp.MustCompile("/[*][^/]*$")
	s := freeWildcardRx.ReplaceAllLiteralString("/foo/*bar/*baz", "")
	fmt.Println(s) // "/foo/*bar"
```

routing/matcher.go
NorthFury: After the refactoring, this is used in:
* `addSubtreeLeafsToPath` for removing the free wildcard param from the end
* `extractWildcardParamNames` to see if a `Path` predicate ends with a free wildcard param
* `hasFreeWildcardParam`  to see if a `Path` predicate ends with a free wildcard param

The code using this regexp cares for both named and unnamed free wildcard params.

routing/matcher.go
NorthFury: This is the intended behavior. Remove the free wildcard from the end in order to get a base path, without free wildcard params.
For a `PathSubtree`  we need to add matchers for the `basePath` and for `basePath + free wildcard`. You also add `basePath + "/"` depending on the trailing slash flag.

NorthFury: In your example, `/foo/*bar/*baz`, `baz` is a free wildcard param, while `bar` is a normal wildcard.
`/foo/*bar/*baz` is equivalent to `/foo/:bar/*baz`

routing/matcher.go
szuecs: thanks

routing/matcher.go
szuecs: ok thanks!

routing/matcher.go
```diff
@@ -139,6 +133,69 @@ func canonicalizeHeaderRegexps(hrx map[string][]*regexp.Regexp) map[string][]*re
...
+		} else {
+			nextSlash := strings.IndexByte(path[i:], '/')
+			if nextSlash == -1 {
+				nextSlash = len(path)
+			} else {
+				nextSlash += i
+			}
+			if c == ':' || c == '*' {
+				sb.WriteByte(c)
+				sb.WriteByte('*')
```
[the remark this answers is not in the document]
NorthFury: I added tests for it and it's working as expected.
It's good to have tests for this particular part; I hope I didn't miss any.

szuecs:
:+1:

routing/matcher.go
```diff
@@ -139,6 +133,75 @@ func canonicalizeHeaderRegexps(hrx map[string][]*regexp.Regexp) map[string][]*re
...
+			nextSlashExists := true
+			if nextSlash == -1 {
+				nextSlash = len(path)
+				nextSlashExists = false
+			} else {
+				nextSlash += i
+			}
+			if c == ':' || c == '*' {
+				if nextSlashExists {
+					sb.WriteByte(':')
```
aryszka: will it accept paths like `/foo/*bar/baz` this way? I think if nextSlashExists is true, and `c == '*'` then it should return an error.

NorthFury: I thought that `:` and `*` is interchangeable for non-free wildcard params. It's worth confirming this was the previous behavior.

NorthFury: I pushed a change to match the previous behavior where only free wildcard params are prefixed by `*` and are only last.

NorthFury:
:+1:

aryszka:
:+1:

PR #1231, opened 2019-10-21
Move config code to own package

### LINKED ISSUE #1230: Move config code of Skipper into its own package
We are adding some custom filters to Skipper. To be able to do this, we are using Skipper as a library. We copy the files in [cmd/skipper](https://github.com/zalando/skipper/tree/master/cmd/skipper) into our project each time we update the version of Skipper. This is done to have all flags available and set their proper default values.
Copying the files requires manual work and is prone to errors.

With the refactoring released in v0.11.1, all options/flags are hidden behind a nice API (`NewConfig()`, `Config.Parse()` and `Config.ToOptions()`). However, when using Skipper as a library we cannot make use of that API because it is defined in package `main` and that is not importable:
```
import "github.com/zalando/skipper/cmd/skipper" is a program, not an importable package
```
It would be perfect for us if all config-related code would be moved into its own package that can be imported.

Can't think of any suitable alternative right now.

/

Yes

bl1nk:
Moves config code to its own package to allow third party code to import the package and use upstream options parsing instead of implementing it themselves.

Fixes #1230 

Also fixes a bug in `Test_NewConfig` where unexported fields were compared but neither allowed nor ignored when printing the diff on failure causing a panic.


szuecs:
:+1:

aryszka:
:+1:

PR #1239, opened 2019-10-31
fix bug when having at least 3 prepend filters

NorthFury:
The issue is that as soon as one has at least 3 prepend filters, all routes get the filters from the last route.

### DIFF
--- a/eskip/eskip.go
+++ b/eskip/eskip.go
@@ -31,12 +31,31 @@ type DefaultFilters struct {
 // prepends filters stored to incoming routes and returns the modified
 // version of routes.
 func (df *DefaultFilters) Do(routes []*Route) []*Route {
+	pn := len(df.Prepend)
+	an := len(df.Append)
+	if pn == 0 && an == 0 {
+		return routes
+	}
+
 	nextRoutes := make([]*Route, len(routes))
 	for i, r := range routes {
 		nextRoutes[i] = new(Route)
 		*nextRoutes[i] = *r
-		nextRoutes[i].Filters = append(df.Prepend, nextRoutes[i].Filters...)
-		nextRoutes[i].Filters = append(nextRoutes[i].Filters, df.Append...)
+
+		fn := len(r.Filters)
+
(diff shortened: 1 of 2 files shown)


eskip/eskip.go
```diff
@@ -31,12 +31,31 @@ type DefaultFilters struct {
...
 	for i, r := range routes {
 		nextRoutes[i] = new(Route)
 		*nextRoutes[i] = *r
-		nextRoutes[i].Filters = append(df.Prepend, nextRoutes[i].Filters...)
-		nextRoutes[i].Filters = append(nextRoutes[i].Filters, df.Append...)
+
+		fn := len(r.Filters)
+
+		filters := make([]*Filter, fn+pn+an)
+		for i, f := range df.Prepend {
```
aryszka:
```
copy(filters[:pn], df.Prepend)
copy(filters[pn:pn+fn], r.Filters)
copy(filters[pn+fn:], df.Append)
```

addityasingh:
👍

eskip/eskip_test.go
```diff
@@ -562,3 +562,27 @@ func TestDefaultFiltersDo(t *testing.T) {
...
 }
+
+func TestDefaultFiltersDoCorrectPrependFilters(t *testing.T) {
+	filters, err := ParseFilters("status(1) -> status(2) -> status(3)")
+	if err != nil {
+		t.Errorf("Failed to parse filter: %v", err)
+	}
+
+	routes, err := Parse(`
+r1: Method("GET") -> inlineContent("r1") -> <shunt>;
```
aryszka: just for the record: the eskip format is not whitespace sensitive

aryszka:
:+1:

szuecs:
:+1:

szuecs:
Thanks interesting behavior of slice allocation. I bet a slice of cap=2 is allocated on the stack, so the append created a new slice on the heap.
With 3 it was re-using the allocated slice.
I have to check that.

PR #1254, opened 2019-12-06
enhance opentracing support in auth filters 

szuecs:
feature: add opentracing tag "target" to show target url for auth calls
feature: wrap authclient to add dial_context log to tokeninfo, tokenintrospection and webhook span


filters/auth/authclient.go
```diff
@@ -92,9 +95,39 @@ func (ac *authClient) doClonedGet(ctx filters.FilterContext) (int, error) {
...
+		f:      d.DialContext,
+	}
+}
+
+// DialContext wraps net.Dialer's DialContext to add logs to the
+// current opentracing span and returns an error, if there is an error
+// while dialing.
+func (acdd *authClientDialerDialer) DialContext(ctx context.Context, network, addr string) (net.Conn, error) {
+	span := opentracing.SpanFromContext(ctx)
+	if span != nil {
```
szuecs: span is always nil as far as I tested :(

filters/auth/authclient.go
```diff
@@ -74,6 +76,7 @@ func (ac *authClient) doClonedGet(ctx filters.FilterContext) (int, error) {
 	parentSpan := ctx.ParentSpan()
 	request := ctx.Request()
 	span := tracer.StartSpan(webhookSpanName, opentracing.ChildOf(parentSpan.Context()))
+	span.SetTag("target", ac.url.String())
```
lmineiro: I believe there are some [established conventions](https://github.com/opentracing/specification/blob/master/semantic_conventions.md) for this.
Either `http.url` or `peer.service` seem like good candidates.

filters/auth/authclient.go
```diff
@@ -159,6 +193,7 @@ func jsonGet(
 func injectSpan(tracer opentracing.Tracer, parentSpan opentracing.Span, childSpanName string, req *http.Request) opentracing.Span {
 	if tracer != nil && parentSpan != nil && childSpanName != "" {
 		span := tracer.StartSpan(childSpanName, opentracing.ChildOf(parentSpan.Context()))
+		span.SetTag("target", req.URL.String())
```
lmineiro: See [comment](https://github.com/zalando/skipper/pull/1254/files#r355376337).

filters/auth/authclient.go
szuecs: the problem was caused by not having:

	req = req.WithContext(opentracing.ContextWithSpan(req.Context(), parentSpan))

filters/auth/authclient.go
```diff
@@ -92,9 +94,39 @@ func (ac *authClient) doClonedGet(ctx filters.FilterContext) (int, error) {
 	return rsp.StatusCode, nil
 }
 
-func createHTTPClient(timeout time.Duration, quit chan struct{}, maxIdleConns int) (*http.Client, error) {
+type authClientDialer struct {
```
sarslanhan: @szuecs after on our offline discussion, I remembered that there is a way to make this more granular to see how long DNS resolution takes. What do you think about using https://golang.org/pkg/net/http/httptrace?

I think this would be valuable for the `proxy` operation as well.

You can find an example [here](https://github.com/opentracing-contrib/go-stdlib/blob/master/nethttp/client.go#L239).

szuecs: I will measure the impact, but sounds good.

szuecs:
OpenTracing Tag: `http.url`

OpenTracing logs:
image

filters/auth/authclient.go
```diff
@@ -113,17 +113,46 @@ func createHTTPClient(timeout time.Duration, quit chan struct{}, maxIdleConns in
...
-		Transport: transport,
-	}, nil
+	return transport
 }
 
-// jsonGet does a get to the url with accessToken as the bearer token in the authorization header. Writes response body into doc.
+func injectClientTrace(req *http.Request, span opentracing.Span) *http.Request {
+	trace := &httptrace.ClientTrace{
+		DNSStart: func(httptrace.DNSStartInfo) {
+			span.LogKV("dial_auth", "DNS lookup start")
```
sarslanhan: I have a suggestion:

Instead of putting everyr event under `dial_auth`, what do you think about having separate events denoted with `start` and `end` like `dns: start`, `dns: end`, `connect: start`, `tls_handshake: end` and so on?

szuecs: fixed

szuecs:
I added documentation for opentracing spans, tags, logs

szuecs:
:+1:

filters/auth/authclient.go
```diff
@@ -21,13 +24,16 @@ const (
 	tokenIntrospectionSpan = "tokenintrospection"
 )
 
-const defaultMaxIdleConns = 64
+const (
+	defaultMaxIdleConns = 64
+	tracingTagURL       = "http.url"
```
lmineiro: You can reuse the "constant" from https://github.com/opentracing/opentracing-go/blob/master/ext/tags.go#L91

lmineiro: I actually mean, not to declare this one and use `ext.HTTPUrl` wherever `tracingTagURL` was referenced

sarslanhan: @szuecs if you do another iteration on this, tagging token validation calls as `span.kind: client` would be good to have. I think that's also part of `ext/tags` @lmineiro posted.

szuecs: will do

szuecs: I did this in https://github.com/zalando/skipper/pull/1255 which I will merge into this pr. I will push it tomorrow, when I added tests.

szuecs: merged #1255 into this one

szuecs:
Tested changes in a dev cluster. The documentation is in sync with all of the changes that were made.

szuecs:
:+1:

aryszka:
:+1:

PR #1255, opened 2019-12-10
implement skipper httpclient and use it in all auth filters

szuecs:
feature: skipper httpclient 
refactor: use it in all svc2svc auth filters

### DIFF
--- a/filters/auth/authclient.go
+++ b/filters/auth/authclient.go
@@ -32,7 +32,7 @@ type authClient struct {
 
 func newAuthClient(baseURL, spanName string, timeout time.Duration, maxIdleConns int, tracer opentracing.Tracer) (*authClient, error) {
 	if tracer == nil {
-		tracer = tracing.NoopTracer{}
+		tracer = opentracing.NoopTracer{}
 	}
 	if maxIdleConns <= 0 {
 		maxIdleConns = defaultMaxIdleConns
--- a/filters/auth/authclient.go
+++ b/filters/auth/authclient.go
@@ -71,7 +71,7 @@ func (ac *authClient) getTokenintrospect(token string, ctx filters.FilterContext
 	}
 	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
 
-	rsp, err := ac.tr.Do(req)
+	rsp, err := ac.tr.RoundTrip(req)
 	if err != nil {
 		return info, err
 	}
@@ -103,7 +103,7 @@ func (ac *authClient) getTokeninfo(token string, ctx filters.FilterContext) (map
 		req.Header.Set(authHeaderName, authHeaderPrefix+token)
 	}
 
-	rsp, err := ac.tr.Do(req)
+	rsp, err := ac.tr.RoundTrip(req)
 	if err != nil {
 		return doc, err
 	}
@@ -125,7 +125,7 @@ func (ac *authClient) getWebhook(ctx filters.FilterContext) (int, error) {
 	}
 	copyHeader(req.Header, ctx.Request().Header)
 
-	rsp, err := ac.tr.Do(req)
+	rsp, err := ac.tr.RoundTrip(req)
 	if err != nil {
 		return -1, err
 	}
--- a/filters/auth/authclient.go
+++ b/filters/auth/authclient.go
@@ -31,6 +31,9 @@ type authClient struct {
 }
 
 func newAuthClient(baseURL, spanName string, timeout time.Duration, maxIdleConns int, tracer opentracing.Tracer) (*authClient, error) {
+	if tracer == nil {
+		tracer = tracing.NoopTracer{}
+	}
 	if maxIdleConns <= 0 {
 		maxIdleConns = defaultMaxIdleConns
 	}
--- a/filters/auth/authclient.go
+++ b/filters/auth/authclient.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strings"
@@ -60,12 +61,11 @@ func (ac *authClient) Close() {
 }
 
 func (ac *authClient) getTokenintrospect(token string, ctx filters.FilterContext) (tokenIntrospectionInfo, error) {
-	info := make(tokenIntrospectionInfo)
 	body := url.Values{}
 	body.Add(tokenKey, token)
 	req, err := http.NewRequest("POST", ac.url.String(), strings.NewReader(body.Encode()))
 	if err != nil {
-		return info, err
+		return nil, err
 	}
 
 	if ac.url.User != nil {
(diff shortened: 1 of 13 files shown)


szuecs:
@aryszka what do you think?

net/httpclient.go
```diff
@@ -0,0 +1,185 @@
...
+
+	go func() {
+		for {
+			select {
+			case <-time.After(options.IdleConnTimeout):
+				htransport.CloseIdleConnections()
+			case <-quit:
+				return
+			}
+		}
```
miry: @szuecs Would make sense to catch panics in such goroutins?

szuecs: @miry In these kind of control loops, I don't think so. Why do you think there could be a panic?

miry: I checked the library, and found overuse of `panic`. 
Found one example:

```golang
func (pc *persistConn) closeLocked(err error) {
	if err == nil {
		panic("nil error")
```

lmineiro: All filters are "expected" to panic since the idea was always to allow the community to contribute filters. Skipper itself handles eventual panics from filters on the [upper proxy layer](https://github.com/zalando/skipper/blob/master/proxy/proxy.go#L735).

net/httpclient.go
```diff
@@ -0,0 +1,234 @@
...
+
+func (t *Transport) shallowCopy() *Transport {
+	tt := *t
+	return &tt
+}
+
+// Do the request with tracing, bearer token injection and add client
+// tracing: DNS, TCP/IP, TLS handshake, connection pool access. Client
+// traces are added as logs into the created span.
+func (t *Transport) Do(req *http.Request) (*http.Response, error) {
```
aryszka: since this is more similar in behavior to the net/http.RoundTripper.RoundTrip(), let's name this function RoundTrip, and in a next pass we can wrap it then with a client that behaves like net/http.Client.Do(), e.g. following redirects

net/httpclient.go
aryszka: It's true that panics are recovered from in the proxy, but it's also true that this is a separate goroutine, so panics from would not be recovered from this way.

OTOH, the snippet above is not going to ever panic, because the function is always called with a non-nil error (https://github.com/golang/go/blob/master/src/net/http/transport.go#L772) I assume the panic is there only in case a test would call it in an invalid way.

There can be of course other causes for panics, like the system calls through the atomic package. We don't handle these, it's better to crash.

filters/auth/tokeninfo.go
```diff
@@ -85,12 +90,13 @@ func NewOAuthTokeninfoAllKV(OAuthTokeninfoURL string, OAuthTokeninfoTimeout time
 // to validate authorization for requests. Current implementation uses
 // Bearer tokens to authorize requests and checks that the token
 // contains at least one key value pair provided.
-func NewOAuthTokeninfoAnyKV(OAuthTokeninfoURL string, OAuthTokeninfoTimeout time.Duration) filters.Spec {
+func NewOAuthTokeninfoAnyKV(OAuthTokeninfoURL string, OAuthTokeninfoTimeout time.Duration, tracer opentracing.Tracer) filters.Spec {
```
aryszka: changing the signature of these initializers makes it hard for applications using Skipper as a library to migrate to the new version. In other parts of the code base, we tend to use an Options structure to flexibly extend the parameters of exported initializers. I'd suggest to stick to this approach. This would become something like this:

```
// extend the existing options
type TokeninfoOptions struct {
	OAuthTokeninfoURL string
	OAuthTokeninfoTimeout time.Duration
	Tracer opentracing.Tracer
	// ...the existing fields
}

func OAuthTokeninfoAnyKVWitOptions(o TokeninfoOptions) filters.Spec {
	return &tokeninfoSpec{
		typ: checkOAuthTokeninfoAnyKV,
		options: o,
	}
}

func NewOAuthTokeninfoAnyKV(OAuthTokeninfoURL string, OAuthTokeninfoTimeout time.Duration) {
	return OAuthTokeninfoAnyKVWitOptions(TokeninfoOptions{
		OAuthTokeninfoURL string
		OAuthTokeninfoTimeout time.Duration
		Tracer opentracing.Tracer
	})
}
```

there are other similar occurences, please change them, too

szuecs: fixed

net/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package net provides generic functions that might be useful in
```
aryszka: Maybe this is not the main message the package docs should send. How about:

```
// Package net provides generic network related functions used across Skipper, which
// might be useful also in other contexts than Skipper.
```

?

filters/auth/authclient.go
```diff
@@ -46,201 +43,96 @@ func newAuthClient(baseURL string, timeout time.Duration, maxIdleConns int) (*au
...
-			span.LogKV("TLS", "end")
-		},
-		GetConn: func(string) {
-			span.LogKV("get_conn", "start")
-		},
-		GotConn: func(httptrace.GotConnInfo) {
-			span.LogKV("get_conn", "end")
-		},
+	buf := make([]byte, rsp.ContentLength)
+	_, err = rsp.Body.Read(buf)
```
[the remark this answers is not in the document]
szuecs: Good catch, the use case before had only a limitted number of bytes in return.

filters/auth/authclient.go
```diff
@@ -46,201 +43,96 @@ func newAuthClient(baseURL string, timeout time.Duration, maxIdleConns int) (*au
...
+	if err != nil && err != io.EOF {
+		return info, err
+	}
+	err = json.Unmarshal(buf, &info)
+	if err != nil {
```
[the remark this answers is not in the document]
szuecs: true

filters/auth/authclient.go
```diff
@@ -46,201 +43,96 @@ func newAuthClient(baseURL string, timeout time.Duration, maxIdleConns int) (*au
...
 
 func (ac *authClient) getTokenintrospect(token string, ctx filters.FilterContext) (tokenIntrospectionInfo, error) {
 	info := make(tokenIntrospectionInfo)
-	err := jsonPost(ac.url, token, &info, ac.rt, ctx.Tracer(), ctx.ParentSpan(), tokenIntrospectionSpan)
+	body := url.Values{}
+	body.Add(tokenKey, token)
+	req, err := http.NewRequest("POST", ac.url.String(), strings.NewReader(body.Encode()))
 	if err != nil {
-		return nil, err
+		return info, err
```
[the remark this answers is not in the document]
szuecs: true

net/doc.go
szuecs: +1

filters/auth/authclient.go
```diff
@@ -76,22 +76,19 @@ func (ac *authClient) getTokenintrospect(token string, ctx filters.FilterContext
...
-
 	defer rsp.Body.Close()
 	if rsp.StatusCode != 200 {
-		return info, errInvalidToken
+		return nil, errInvalidToken
 	}
-	buf := make([]byte, rsp.ContentLength)
-	_, err = rsp.Body.Read(buf)
+
+	buf, err := ioutil.ReadAll(rsp.Body)
```
aryszka: ioutil.ReadAll doesn't return io.EOF, but nil

aryszka:
:+1:

szuecs:
:+1:

PR #1256, opened 2019-12-12
Remove excplict set of GOPROXY

miry:
The default setting for GOPROXY is "https://proxy.golang.org,direct" since go version 1.13.
With current implementation, the option direct is missing, that is fallback to original sources.


aryszka:
:+1:

szuecs:
:+1:

PR #1260, opened 2019-12-20
fix two small issues in redis usage by cluster*Ratelimit filters

szuecs:
- upstream issue was fixed in go-redis v7 (beta), so we switch to TxPipeline to be safe
- add EXPIRE call to remove keys automatically, if we have no writes since specified time window + 1s


aryszka:
:+1:

szuecs:
:+1:

PR #1261, opened 2019-12-30
Forward headers from webhook filter responses

dlmiddlecote:
Update the `webhook` filter so that it can take a second argument, a comma-separated list of header names, that should be copied from the webhook response to the continuing request, if the webhook response is authorized.

i.e. `webhook("auth-endpoint", "Headers-To-Copy,And-More")`

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -453,13 +453,17 @@ The `webhook` filter makes it possible to have your own authentication and
 authorization endpoint as a filter.
 
 Headers from the incoming request will be copied into the request that
-is being done to the webhook endpoint. Responses from the webhook with
-status code less than 300 will be authorized, rest unauthorized.
+is being done to the webhook endpoint. It is possible to copy headers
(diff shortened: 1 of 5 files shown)


dlmiddlecote:
Not sure why the tests failed, can't reproduce locally.

szuecs:
From the docs and or message, I see that it’s a breaking change.
Is it a breaking change?
If so please do it non breaking.
I did not read the code yet, I can probably do it on Friday.

dlmiddlecote:
Hey @szuecs, this shouldn't be a breaking change, as the second argument to the webhook filter doesn't need to be specified, and if it's not, the filter will work as it does today. I guess the documentation doesn't convey that well enough.

szuecs:
Can you please have both as examples, such that it’s clear in the docs?
Thanks!

szuecs:
:+1:

aryszka:
:+1:

Issue #1266, opened 2020-01-09
rainbow traffic switching flap route matches


szuecs:
If you specify more than 2 versions of the same ingress, like in the example you get unepected results:

```
kind: Ingress
metadata:
  name: my-app
  labels:
    application: my-app
  annotations:
    zalando.org/backend-weights: |
      {"my-app-1": 50, "my-app-2": 25, "my-app-3": 25}
```

This will result in 3 routes:

```
r1: Traffic(0.5) -> "my-app-1" // 50%
r2: Traffic(0.5) -> "my-app-2" // 50% out of the lasting 50% -> 25%
r3: * -> "my-app-3" // rest -> 25%
```

This looks good, but we now have r1 and r2 with the same priority for the routing match!
Therefore r1 and r2 can flap, which results in sometimes having unexpected r1 25% and r2 50%.

PR #1267, opened 2020-01-09
Fix/kube multicolor traffic switch

szuecs:
fix https://github.com/zalando/skipper/issues/1266

- fix: multi color (aka rainbow) traffic switching in kubernetes
- test: Traffic() predicate
- feature: True() predicate always return true for "match"

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/filters/builtin"
+	"github.com/zalando/skipper/predicates/primitive"
 	"github.com/zalando/skipper/predicates/source"
 	"github.com/zalando/skipper/predicates/traffic"
 )
@@ -718,14 +719,7 @@ func (c *Client) convertPathRule(
 		}
 
 		setPath(pathMode, r, prule.Path)
-
-		if 0.0 < prule.Backend.Traffic && prule.Backend.Traffic < 1.0 {
-			r.Predicates = append([]*eskip.Predicate{{
-				Name: traffic.PredicateName,
-				Args: []interface{}{prule.Backend.Traffic},
-			}}, r.Predicates...)
-			log.Debugf("Traffic weight %.2f for backend '%s'", prule.Backend.Traffic, svcName)
-		}
+		setTraffic(r, svcName, prule.Backend.Traffic, prule.Backend.noopCount)
 		return r, nil
 
 	} else if err != nil {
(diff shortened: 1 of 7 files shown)


dataclients/kubernetes/kube.go
```diff
@@ -1193,6 +1192,7 @@ func computeBackendWeights(backendWeights map[string]float64, rule *rule) {
 			if weight > 0 {
 				sc.lastActive = path.Backend
 			}
+			sc.weightsCount++
```
mikkeloscar: Probably better to only add this if `weight > 0` because we only need to create `noop` predicates for those cases.

dataclients/kubernetes/kube.go
```diff
@@ -1213,6 +1213,15 @@ func computeBackendWeights(backendWeights map[string]float64, rule *rule) {
 				// give subsequent backends a higher relative
 				// weight.
 				sc.sum -= weight
+
+				// noops are required to make sure that routes are in order selected by routing tree
+				noopCount := sc.weightsCount - 2
```
mikkeloscar: Not sure what the 2 is here? I guess this is related to my comment above about only counting positive weights and wouldn't be needed in that case?

szuecs: Basically we only need a Noop() for >2 traffic switching routes, because the last  has no Traffic() and the other one has 1 Traffic(), so 1 predicate more than the one without Traffic().
For the >1 with Traffic() routes we need to be sure we have a decreasing number of matching predicates.

mikkeloscar: Then I think a check like `sc.weightsCount > 2` would be simpler to understand in the code.

szuecs: I guess done :)

dataclients/kubernetes/kube.go
szuecs: fixed

szuecs:
:+1:

szuecs:
:+1:

dataclients/kubernetes/kube.go
```diff
@@ -881,6 +876,7 @@ func (c *Client) ingressRoute(i *ingressItem, redirect *redirectInfo, state *clu
 		ic.logger.Errorf("error while converting default backend: %v", err)
 	}
 	for _, rule := range i.Spec.Rules {
+		ic.trafficNoopCount = 0 // TODO(sszuecs): maybe move to path route
```
aryszka: the `ic` var is invariant inside the loop, it can be outside of the loop

predicates/primitive/true.go
```diff
@@ -0,0 +1,31 @@
...
+const (
+	NameTrue = "True"
+)
+
+type trueSpec struct{}
+
+type truePredicate struct{}
+
+// NewTrue provides a predicate spec to create a Predicate instance that evaluates to true
+func NewTrue() routing.PredicateSpec { return &trueSpec{} }
```
aryszka: just for general symmetry, can we add a `False()`, as well?

dataclients/kubernetes/kube.go
```diff
@@ -222,6 +223,7 @@ type ingressContext struct {
 	redirect            *redirectInfo
 	hostRoutes          map[string][]*eskip.Route
 	defaultFilters      map[resourceId]string
+	trafficNoopCount    int
```
aryszka: i think we don't use this field

szuecs: dropped

aryszka:
:+1:

szuecs:
:+1:

PR #1278, opened 2020-01-19
Feature/skipper http client

szuecs:
net.Client implements the current net/http.Ciient interface with skipper's net.Transport with opentracing and secrets injection on top.
I changed the secrets module interface to simplify the implementation. I think it's fine to do, because no one should have used this besides our OpenIDConnect filters.

### DIFF
--- a/filters/auth/authclient.go
+++ b/filters/auth/authclient.go
@@ -27,7 +27,7 @@ const (
 
 type authClient struct {
 	url *url.URL
-	tr  *net.Transport
+	cli *net.Client
 }
 
 func newAuthClient(baseURL, spanName string, timeout time.Duration, maxIdleConns int, tracer opentracing.Tracer) (*authClient, error) {
@@ -43,20 +43,20 @@ func newAuthClient(baseURL, spanName string, timeout time.Duration, maxIdleConns
 		return nil, err
 	}
 
-	tr := net.NewTransport(net.Options{
-		ResponseHeaderTimeout: timeout,
-		TLSHandshakeTimeout:   timeout,
-		MaxIdleConnsPerHost:   maxIdleConns,
-		Tracer:                tracer,
+	cli := net.NewClient(net.Options{
+		ResponseHeaderTimeout:   timeout,
+		TLSHandshakeTimeout:     timeout,
+		MaxIdleConnsPerHost:     maxIdleConns,
+		Tracer:                  tracer,
+		OpentracingComponentTag: "skipper",
+		OpentracingSpanName:     spanName,
 	})
-	tr = net.WithSpanName(tr, spanName)
-	tr = net.WithComponentTag(tr, "skipper")
 
-	return &authClient{url: u, tr: tr}, nil
+	return &authClient{url: u, cli: cli}, nil
 }
 
 func (ac *authClient) Close() {
-	ac.tr.Close()
+	ac.cli.Close()
 }
 
 func (ac *authClient) getTokenintrospect(token string, ctx filters.FilterContext) (tokenIntrospectionInfo, error) {
@@ -73,7 +73,7 @@ func (ac *authClient) getTokenintrospect(token string, ctx filters.FilterContext
 	}
 	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
 
-	rsp, err := ac.tr.RoundTrip(req)
+	rsp, err := ac.cli.Do(req)
 	if err != nil {
 		return nil, err
 	}
@@ -103,7 +103,7 @@ func (ac *authClient) getTokeninfo(token string, ctx filters.FilterContext) (map
 		req.Header.Set(authHeaderName, authHeaderPrefix+token)
 	}
 
-	rsp, err := ac.tr.RoundTrip(req)
+	rsp, err := ac.cli.Do(req)
 	if err != nil {
 		return doc, err
 	}
@@ -125,7 +125,7 @@ func (ac *authClient) getWebhook(ctx filters.FilterContext) (*http.Response, err
 	}
 	copyHeader(req.Header, ctx.Request().Header)
 
-	rsp, err := ac.tr.RoundTrip(req)
+	rsp, err := ac.cli.Do(req)
 	if err != nil {
 		return nil, err
 	}
--- a/net/httpclient.go
+++ b/net/httpclient.go
@@ -2,17 +2,129 @@ package net
 
 import (
 	"crypto/tls"
+	"io"
 	"net/http"
 	"net/http/httptrace"
+	"net/url"
+	"strings"
 	"time"
 
 	"github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
+	"github.com/zalando/skipper/logging"
+	"github.com/zalando/skipper/secrets"
 )
 
 const (
 	defaultIdleConnTimeout = 30 * time.Second
+	defaultRefreshInterval = 5 * time.Minute
 )
+
+// Client adds additional features like Bearer token injection, and
+// opentracing to the wrapped http.Client with the same interface as
+// http.Client from the stdlib.
+type Client struct {
+	client http.Client
+	tr     *Transport
+	log    logging.Logger
+	sr     secrets.SecretsReader
+}
+
+// NewClient creates a wrapped http.Client and uses Transport to
+// support OpenTracing. On teardown you have to use Close() to
+// not leak a goroutine.
+//
+// If secrets.SecretsReader is nil, but BearerTokenFile is not empty
+// string, it creates StaticDelegateSecret with a wrapped
+// secrets.SecretPaths, which can be used with Kubernetes secrets to
+// read from the secret an automatically updated Bearer token.
+func NewClient(o Options) *Client {
+	if o.Log == nil {
+		o.Log = &logging.DefaultLog{}
+	}
+
+	tr := NewTransport(o)
+
+	sr := o.SecretsReader
+	if sr == nil && o.BearerTokenFile != "" {
+		if o.BearerTokenRefreshInterval == 0 {
+			o.BearerTokenRefreshInterval = defaultRefreshInterval
+		}
+		sp := secrets.NewSecretPaths(o.BearerTokenRefreshInterval)
+		if err := sp.Add(o.BearerTokenFile); err != nil {
+			o.Log.Errorf("failed to read secret: %v", err)
+		}
+		sr = secrets.NewStaticDelegateSecret(sp, o.BearerTokenFile)
+	}
+
+	c := &Client{
+		client: http.Client{
+			Transport: tr,
+		},
+		tr:  tr,
+		log: o.Log,
+		sr:  sr,
+	}
+
+	return c
+}
+
+func (c *Client) Close() {
+	c.tr.Close()
+	if c.sr != nil {
+		c.sr.Close()
+	}
+}
+
+func (c *Client) Head(url string) (*http.Response, error) {
+	req, err := http.NewRequest("HEAD", url, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return c.Do(req)
+}
+
+func (c *Client) Get(url string) (*http.Response, error) {
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+	return c.Do(req)
+}
+
+func (c *Client) Post(url, contentType string, body io.Reader) (*http.Response, error) {
+	req, err := http.NewRequest("POST", url, body)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Content-Type", contentType)
+
+	return c.Do(req)
+}
+
+func (c *Client) PostForm(url string, data url.Values) (*http.Response, error) {
+	return c.Post(url, "application/x-www-form-urlencoded", strings.NewReader(data.Encode()))
+}
+
+// Do delegates the given http.Request to the underlying http.Client
+// and adds a Bearer token to the authorization header, if Client has
+// a secrets.SecretsReader and the request does not contain an
+// Authorization header.
+func (c *Client) Do(req *http.Request) (*http.Response, error) {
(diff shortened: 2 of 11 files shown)


net/httpclient_example_test.go
```diff
@@ -30,3 +42,83 @@ func ExampleTransport() {
...
+
+	cli := net.NewClient(net.Options{
+		Tracer:                  tracer,
+		OpentracingComponentTag: "testclient",
+		OpentracingSpanName:     "clientSpan",
+		SecretsReader:           sp,
+		IdleConnTimeout:         2 * time.Second,
+	})
+	defer func() {
+		println("closing...")
```
aryszka: `println`: it may be good for the example, but if we want to keep them, then why not using `log.Println` here?

net/httpclient.go
```diff
@@ -2,18 +2,114 @@ package net
...
+		sr:          sr,
+		secretsPath: o.BearerTokenFile,
+		quit:        quit,
+	}
+
+	return c
+}
+
+func (c *Client) Close() {
+	close(c.quit)
```
aryszka: I don't see c.quit being used, I think we can delete it. OTOH, I think we should call c.client.Transport.Close() somehow (maybe by keeping a reference to our Transport implementation), because NewClient() calls NewTransport() which starts its goroutine.

net/httpclient.go
```diff
@@ -2,18 +2,114 @@ package net
...
+func (c *Client) Head(url string) (*http.Response, error) {
+	req, err := http.NewRequest("HEAD", url, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return c.Do(req)
+}
+
+func (c *Client) Get(url string) (*http.Response, error) {
```
tkrop: What about content negotiation and other headers?

aryszka: my two cents: content negotiation is out of scope for this PR

net/httpclient.go
szuecs: good point, done

net/httpclient_example_test.go
szuecs: dropped println

net/httpclient.go
```diff
@@ -61,6 +157,21 @@ type Options struct {
 	ExpectContinueTimeout time.Duration
 	// Tracer instance, can be nil to not enable tracing
 	Tracer opentracing.Tracer
+
+	// BearerTokenFile injects bearer token read from file, which file path is the given string
+	BearerTokenFile string
```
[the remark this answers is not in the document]
szuecs: good point

net/httpclient.go
```diff
@@ -2,18 +2,114 @@ package net
...
+
+	tr := NewTransport(o)
+
+	sr := o.SecretsReader
+	if sr == nil {
+		if o.BearerTokenRefreshInterval == 0 {
+			o.BearerTokenRefreshInterval = defaultRefreshInterval
+		}
+		sp := secrets.NewSecretPaths(o.BearerTokenRefreshInterval)
+		if err := sp.Add(o.BearerTokenFile); err != nil {
```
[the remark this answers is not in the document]
szuecs: thanks, was an error, fixed

net/httpclient.go
```diff
@@ -2,18 +2,114 @@ package net
...
 
+type Client struct {
+	client      http.Client
+	log         logging.Logger
+	sr          secrets.SecretsReader
+	secretsPath string
+	quit        chan struct{}
+}
+
+func NewClient(o Options) *Client {
```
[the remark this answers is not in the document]
szuecs: done

net/httpclient.go
```diff
@@ -2,18 +2,114 @@ package net
...
+	"github.com/zalando/skipper/logging"
+	"github.com/zalando/skipper/secrets"
 )
 
 const (
 	defaultIdleConnTimeout = 30 * time.Second
+	defaultRefreshInterval = 5 * time.Minute
 )
 
+type Client struct {
```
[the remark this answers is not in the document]
szuecs: done

szuecs:
@aryszka I added a bit more dynamic secret lookup possibility.  wdyt?

net/httpclient.go
tkrop: It is not only content negotiation, but the inability to communicate with many of our services without setting the correct (required) content type. @szuecs I just mentioned this because you announced to build a non-sucking net/http.Client replacement in the golang-chat room.

szuecs: @tkrop Why this is important to communicate with our services?
Maybe sent me a private message with links to show it.

tkrop: If I want to use this as a replacement of the net/http.Client ... and this is what I understood you were working on.

tkrop: If I want to use this as a replacement of the net/http.Client ... and this is what I understood you were working on.

tkrop: If I want to use this as a replacement of the net/http.Client ... and this is what I understood you were working on.

tkrop: And no, default headers on any request are usually not flexible enough to support our [RESTful API Guidelines](https://opensource.zalando.com/restful-api-guidelines/)

szuecs: We won't handle the complete API guidelines. I don't see why default headers can not provide value, if you only interact with json APIs in most cases then apply this as `Accept` default header for example.

tkrop: The reason is, that I would expect one client to support multiple concurrent request to different endpoints. So, *iff* you intend to write the wrapper generic, you should support this use case by providing a request specific header map.

net/httpclient.go
```diff
@@ -21,14 +21,66 @@ const (
...
+// Lookuper is an indirection and used to map calls to
+// secrets.SecretsReader's GetSecret to the right parameter.
+type Lookuper interface {
+	// Lookup maps an URL string to the parameter passed into
+	// SecretsReader.GetSecret().
+	Lookup(*url.URL) string
+}
+
+// SingleStaticSecretLookuper stores the string to statically lookup
+type SingleStaticSecretLookuper string
```
aryszka: name SingleStaticSecretLookuper, how about calling it StaticSecret? Singular indicates that it's single, lookuper derives from the fact that it implements it.

net/httpclient.go
```diff
@@ -21,14 +21,66 @@ const (
...
+
+// NewSingleStaticSecretLookuper creates a SingleStaticSecretLookuper
+// that lookups always to the given s.
+func NewSingleStaticSecretLookuper(s string) SingleStaticSecretLookuper {
```
aryszka: NewStaticSecret?

aryszka: TBH, not sure if this initializer is needed at all, it's equivalent to StaticSecret(s)

net/httpclient.go
```diff
@@ -21,14 +21,66 @@ const (
 	defaultRefreshInterval = 5 * time.Minute
 )
 
+// Lookuper is an indirection and used to map calls to
+// secrets.SecretsReader's GetSecret to the right parameter.
+type Lookuper interface {
```
aryszka: from the 'net' packages perspective, 'net.Lookuper' is not very descriptive. It could be rather 'SecretLookuper'

aryszka: I can see how the Lookuper solves the previous limitation caused by using the 'secretsPath' field. But we always use now the req.URL as the input for the lookup, and then whatever the Lookup returns, we use that as the key for the GetSecret. Since both the SecretReader and the SecretLookuper can be implemented by the user code, have you considered to just use the SecretReader and use the req.URL as the key? I would probably try to do it that way, because it appears simpler to me.

net/httpclient.go
szuecs: true

net/httpclient.go
szuecs: changed

net/httpclient.go
szuecs: I agree, changed

szuecs: Maybe the `SecretLookuper.Lookup` has the wrong parameter. I guess we should pass the http.Request instead, such that it can differentiate read and write paths.
For example you have 2 tokens "r" and "w", you could implement POST,PATCH,PUT,DELETE to pass token "w" and rest "r".
What do you think?

szuecs:
@aryszka I changed the code to implement secrets.SecretsReader, looks much better than having new interfaces defined, thanks for the suggestion!

net/httpclient.go
```diff
@@ -2,18 +2,219 @@ package net
...
+
+// StaticDelegateSecret delegates with a static string to the wrapped
+// SecretsReader
+type StaticDelegateSecret struct {
+	sr  secrets.SecretsReader
+	key string
+}
+
+// NewStaticDelegateSecret creates a wrapped secrets.SecretsReader,
+// that use given s to the underlying secrets.SecretsReader to return
```
aryszka: s -> key to be more descriptive.

aryszka: I am not sure if this StaticDelegateSecret is useful, tbh, because using a fixed key is not a complicated thing for the user code.

net/httpclient.go
```diff
@@ -2,18 +2,219 @@ package net
...
+	sr     secrets.SecretsReader
+	secMap map[string]string
+}
+
+// NewHostSecret create a SecretsReader that returns a secret for
+// given host. The given map is used to map hostname to the secrets
+// reader key to read the secret from.
+func NewHostSecret(sr secrets.SecretsReader, h map[string]string) *HostSecret {
+	if h == nil {
+		h = make(map[string]string)
```
aryszka: it's unnecessary to initialize the map value, a nil map will behave the same way when only reading from it.

net/httpclient.go
```diff
@@ -2,18 +2,219 @@ package net
...
+// not leak a goroutine.
+//
+// If secrets.SecretsReader is nil, but BearerTokenFile is not empty
+// string, it creates StaticDelegateSecret with a wrapped
+// secrets.SecretPaths, which can be used with Kubernetes secrets to
+// read from the secret an automatically updated Bearer token.
+func NewClient(o Options) *Client {
+	quit := make(chan struct{})
+	if o.Log == nil {
+		o.Log = logrus.New()
```
aryszka: not sure, but I would consider to use `github.com/zalando/skipper/logging.DefaultLogger{}` here instead of logrus.New(). It will use currently the globally preconfigured logrus, e.g. to enforce the '[APP]' prefix for non-access-log entries, which can be important when splitting the log output into different files. While using the global is regrettable, using the DefaultLogger{} may make it easier later for us to refactor the logging cetrally.

filters/auth/authclient.go
```diff
@@ -43,20 +43,20 @@ func newAuthClient(baseURL, spanName string, timeout time.Duration, maxIdleConns
 		return nil, err
 	}
 
-	tr := net.NewTransport(net.Options{
-		ResponseHeaderTimeout: timeout,
-		TLSHandshakeTimeout:   timeout,
-		MaxIdleConnsPerHost:   maxIdleConns,
-		Tracer:                tracer,
+	cli := net.NewClient(net.Options{
```
aryszka: we either should set the logger here, or use a different default in net.NewClient, because otherwise we will mix together access and application logs

net/httpclient.go
```diff
@@ -2,18 +2,219 @@ package net
...
+
+// Client adds additional features like Bearer token injection, and
+// opentracing to the wrapped http.Client with the same interface as
+// http.Client from the stdlib.
+type Client struct {
+	client http.Client
+	tr     *Transport
+	log    logging.Logger
+	sr     secrets.SecretsReader
+	quit   chan struct{}
```
aryszka: i think we don't need this quit attribute

net/httpclient.go
```diff
@@ -2,18 +2,219 @@ package net
...
+	defaultRefreshInterval = 5 * time.Minute
 )
 
+// StaticSecret implements SecretsReader interface. Example:
+//
+//    sec := []byte("mysecret")
+//    sss := StaticSecret(sec)
+//    b,_ := sss.GetSecret("")
+//    string(b) == sec // true
+type StaticSecret []byte
```
aryszka: I wonder if it would be better to move the StaticSecret and the HostSecret to the secrets package. WDYT?

net/httpclient.go
aryszka: for these cases, I would rather use two SecretReaders (or lookupers), and would leave to the calling code which one to use. The reason is that the secrets package may be good for other things, too, besides only http requests.

aryszka:
thanks for follow-up @szuecs  . See my comments. The most important remark would be the one about the logging, to make sure that we can differentiate between the application logs and the access logs.

net/httpclient.go
szuecs: StaticDelegateSecret can be used with kubernetes secrets, if you only have one, that you use for all client calls, for example.

net/httpclient.go
szuecs: Yeah, I also had the idea, but not yet the time to do it!

net/httpclient.go
szuecs: Internally discussed with @tkrop, that we do not want to support such highlevel http client features. The purpose is more to reduce the boilerplate and add common functionality: regular DNS lookups, tracing with detailed data, Bearer token injection updated from files

More highlevel clients could use this Client in order to have a solid base

net/httpclient.go
szuecs: moved

net/httpclient.go
szuecs: Thanks! Cleaned up this was used by an experiment

filters/auth/authclient.go
szuecs: changed in NewClient

net/httpclient.go
szuecs: fixed

net/httpclient.go
szuecs: fixed

net/httpclient.go
aryszka: fair enough

szuecs:
:+1:

aryszka:
:+1:

Issue #1089, opened 2019-06-12
Cookies set by oauthOidc -filters are occasionally too big


josafat:
**Describe the bug**
oauthOidc* -filters like oauthOidcAnyClaims set a cookie from the client. The cookie size reaches easily the maximum size that the browsers accept (4096 bytes on Chrome?) causing the cookie to be dropped. With a basic OIDC setup e.g. with Keycloak, the size of the cookie easily reaches 5100bytes.

**To Reproduce**
1. Any route with oauthOidcAnyClaims -filter, which passes OIDC auth.
```
test:
      PathSubtree("/test/")
      -> oauthOidcAnyClaims("https://test.net/auth/realms/testrealm", "test", "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxxx", "https://test.net/test/", "email profile", "name email")
      -> "http://test:9912";
```

Callback request sent after successful authentication to oidcp
```
Request URL: https://test.net/test/?state=e30123f4c2164...(truncated)...d791826ebce&session_state=614e2f07-8c31-...(truncated)...-d215d8569e70
Request Method: GET
Status Code: 307 
```

Response headers:
```
content-length: 0
date: Thu, 06 Jun 2019 09:14:34 GMT
location: /test
server: Skipper
set-cookie: skipperOauthOidcec3dfd82=d6a93c3bc69cae1201503...(truncated 5.75KB of data)...1a1c0b72ddd32577ed2b443d990891558930364288da0835; Path=/; HttpOnly; MaxAge=3600; Domain=test.net
status: 307
```

**Expected behavior**
The cookie(s) fit to browsers. E.g. the big ones are split into smaller cookies and merged when read. As an example keycloak-gatekeeper, which acts in a similar role from OIDC point of view splits the big cookies into smaller ones https://github.com/keycloak/keycloak-gatekeeper/blob/master/cookies.go

**Observed behavior**
Cookies are rejected by Chrome, Safari due to too big size.

szuecs:
@josafat thanks for the detailed explanation!

@arjunrn do you want to work on a fix?

arjunrn:
@szuecs yes I will assign the issue to myself

universam1:
Looks like we are hitting the same issue

szuecs:
It would be nice to have a test bed here, such that we can fix it more easy. 

[...]

PR #1289, opened 2020-01-28
implementing cookie chunks for OIDC

universam1:
### adding OIDC Cookie chunk and merge functions

This solves a cookie size overflow issue #1089 where at some point the cookie is not accepted by the browser due to its size

* adding a `chunkCookie` function that splits an existing cookie based on limits defined at `cookieMaxSize`
* the value of `cookieMaxSize` is taken by well-known projects and verified with Chrome
* the approach is `non-deterministic` as the signature of cookies might change over time
* mergerCookies return a single (virtual) cookie, ordered by the suffix. This approach should be least invasive to the current code
* switching serialization to Base64 instead of Base16 since it reduces the transaction data by 33% and is considered common practice. It does not influence other cookies of Skipper

### DIFF
--- a/filters/auth/oidc.go
+++ b/filters/auth/oidc.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"sort"
 	"strings"
 	"time"
 
@@ -28,6 +29,7 @@ const (
 	oauthOidcCookieName = "skipperOauthOidc"
 	stateValidity       = 1 * time.Minute
 	oidcInfoHeader      = "Skipper-Oidc-Info"
+	cookieMaxSize       = 4069
 )
 
 type (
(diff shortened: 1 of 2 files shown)


filters/auth/oidc.go
```diff
@@ -28,6 +30,7 @@ const (
 	oauthOidcCookieName = "skipperOauthOidc"
 	stateValidity       = 1 * time.Minute
 	oidcInfoHeader      = "Skipper-Oidc-Info"
+	cookieMaxSize       = 4069
```
szuecs: I searched a bit and indeed it's everywhere, but no documentation, that explicitly states this.
It would be nice to add a note `// known max cookie size for common browsers` or something like that

szuecs:
@universam1 thanks! Lgtm, only a note as comment, that this weird cookie size is documented.

filters/auth/oidc.go
universam1: I've verified this assumption using this service: http://browsercookielimits.squawky.net/

szuecs:
:+1:

universam1:
@szuecs I performed now a full integration test with my idP and I can confirm the issue #1089 is solved with this PR!

image

szuecs:
:+1:

aryszka:
:+1:

PR #1290, opened 2020-01-28
cancel the outgoing when the incoming request got canceled

aryszka:
The current approach relies on using a shared context (context.Context). All consequences of this need yet to be verified, but it's open for testing in deployments.


dataclients/routestring/example_test.go
```diff
@@ -39,7 +39,4 @@ func Example() {
 	}
 
 	fmt.Println(string(content))
-
-	// Output:
-	// Hello, world!
```
szuecs: Was this on purpose or some linter recommended it?

proxy/longpoll_test.go
```diff
@@ -0,0 +1,288 @@
...
+			}
+		}
+
+		if sendHeader {
+			// in practice, this doesn't block
+			w.WriteHeader(http.StatusOK)
+			w.(interface{ Flush() }).Flush()
+		}
+
+		for loop := true; loop; {
```
szuecs: why do you need `loop` and `for {}` is not enough ?

dataclients/routestring/example_test.go
aryszka: for some reason the tests failed here too often. It's unrelated from this PR, I should put it back.

proxy/longpoll_test.go
aryszka: not needed, at some point this loop had a break in a select, but even then, this is not the right way, it's WIP garbage

aryszka:
eventually, i was able to repro the failing build in the master branch, but seems like it happens more often here. 'Seems', maybe not.

dataclients/routestring/example_test.go
aryszka: i see now. There is a race condition in the example and we don't know when will skipper return 404 as no routes have been configured yet, or the expected output. We can either make the test less readable by adding the necessary synchronization or we can just skip checking the output.

szuecs: fine for me

aryszka:
see this issue and the reference PR: https://github.com/zalando/skipper/pull/857

the underlying case was never proven, or that it is actually a production issue and not only log spam

proxy/longpoll_test.go
aryszka: fixed

szuecs: this was not changed

szuecs:
:+1:

aryszka:
thanks for the review!

aryszka:
:+1:

PR #1291, opened 2020-01-30
Forward body from backend on connection upgrade errors

mikkeloscar:
In case the backend rejects a connection upgrade and returns something other than `101` we just responded with the HTTP status code as the body e.g. `Bad Request`

Instead we should respond with the body of the response which may have important information on why it failed.

### DIFF
--- a/proxy/upgrade.go
+++ b/proxy/upgrade.go
@@ -124,7 +124,12 @@ func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
 	if resp.StatusCode != http.StatusSwitchingProtocols {
 		log.Debugf("Got invalid status code from backend: %d", resp.StatusCode)
 		w.WriteHeader(resp.StatusCode)
-		w.Write([]byte(http.StatusText(resp.StatusCode)))
+		_, err := io.Copy(w, resp.Body)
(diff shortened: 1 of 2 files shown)


proxy/upgrade.go
```diff
@@ -124,7 +124,12 @@ func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
 	if resp.StatusCode != http.StatusSwitchingProtocols {
 		log.Debugf("Got invalid status code from backend: %d", resp.StatusCode)
 		w.WriteHeader(resp.StatusCode)
-		w.Write([]byte(http.StatusText(resp.StatusCode)))
+		_, err := io.Copy(w, resp.Body)
+		if err != nil {
+			log.Errorf("Error writing body to client: %s", err)
+			return
+		}
+		resp.Body.Close()
```
aryszka: we should call resp.Body.Close regardless if the copy succeeded or failed.

Nice to have: we could close the response body in each case where we return from here.

mikkeloscar: Fixed! Thanks! :)

aryszka:
:+1:

mikkeloscar:
:+1:

PR #1292, opened 2020-01-30
feature: add optional startup checks

szuecs:
feature: add optional startup checks, that are checked to return 200 before we start listener -startup-checks="https://www.zalando.de/,https://www.google.com/,http://127.0.0.1:12345/health"

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -41,6 +41,7 @@ type Config struct {
 	DebugListener                   string         `yaml:"debug-listener"`
 	CertPathTLS                     string         `yaml:"tls-cert"`
 	KeyPathTLS                      string         `yaml:"tls-key"`
+	StartupChecks                   *listFlag      `yaml:"startup-checks"`
 	PrintVersion                    bool           `yaml:"version"`
 	MaxLoopbacks                    int            `yaml:"max-loopbacks"`
 	DefaultHTTPStatus               int            `yaml:"default-http-status"`
@@ -247,6 +248,7 @@ const (
 
 	// generic:
 	addressUsage                         = "network address that skipper should listen on"
+	startupChecksUsage                   = "URLs to check before reporting healthy on startup"
 	enableTCPQueueUsage                  = "enable experimental TCP listener queue"
 	expectedBytesPerRequestUsage         = "bytes per request, that is used to calculate concurrency limits to buffer connection spikes"
 	maxTCPListenerConcurrencyUsage       = "sets hardcoded max for TCP listener concurrency, normally calculated based on available memory cgroups with max TODO"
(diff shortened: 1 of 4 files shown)


mikkeloscar:
I think it would make more sense if the checks were continuous rather than only at startup. Imagine our case with tokeninfo sidecar. If tokeninfo crashes for whatever reason after running for 30 min. then it would not be detected by this check.

aryszka:
@mikkeloscar I agree, addressed this idea with @szuecs . We decided to start with this one, in small steps. It can be extended to the solution you proposed. I think the ideal design would be to be able to pass around some kind of a 'health context', and the different components could use it to set healthy or unhealthy, and also could dynamically add dependencies to be regularly verified.

But I think we can proceed with this PR as the first step. WDYT?

config/config.go
```diff
@@ -41,6 +41,7 @@ type Config struct {
 	DebugListener                   string         `yaml:"debug-listener"`
 	CertPathTLS                     string         `yaml:"tls-cert"`
 	KeyPathTLS                      string         `yaml:"tls-key"`
+	StartupChecks                   *listFlag      `yaml:"startup-checks"`
```
aryszka: I believe we should think about extending this functionality in the future into continuous checks, and in that case this name would become deprecated very quick. Can you think about a better name that doesn't suggest that it's currently only for the startup?

Maybe HealthcheckHooks? StatusHooks? StatusChecks?

skipper.go
```diff
@@ -1288,6 +1291,24 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
+		for {
+			/* #nosec */
+			resp, err := http.Get(startupCheckURL)
+			if err != nil {
+				log.Infof("%s unhealthy", startupCheckURL)
+				time.Sleep(1 * time.Second)
+				continue
+			}
+			if resp.StatusCode == 200 {
+				log.Infof("%s got healthy", startupCheckURL)
```
aryszka: "got healthy": maybe it was healthy in the beginning. Let's only log the success if it was unhealthy before.

mikkeloscar:
In that case I would maybe just mark the flag as experimental or so, so it doesn't become "official" API and harder to remove at a later point.

aryszka:
@mikkeloscar good point, thx

skipper.go
szuecs: Does it matter?
I would like to see if there was a check that it actually checked it.

config/config.go
szuecs: I am not sure if I would change to a generic name, because we might just drop this in favor for a more generic solution as we discussed.

aryszka: dropping may cause unnecessary maintenance cost. I see that the interface may change, and that's why we're suggesting to mark it experimental, but since there is a good chance that we keep it, it is worth making it more generic with this anticipation in this case.

skipper.go
aryszka: then please change it from "got health" to "is healthy"

szuecs: I drop "got"

szuecs:
Addressed the comments Naming TBD

config/config.go
szuecs: We decided to go with StatusChecks

szuecs:
:+1:

aryszka:
:+1:

PR #1294, opened 2020-02-03
fix: default filters via configmap are not applied

szuecs:
fix: default filters via configmap are not applied, because we filtered symlinks


hjacobs:
Should we have a test for this?

szuecs:
@hjacobs done

hjacobs:
:+1:

szuecs:
:+1:

Issue #1277, opened 2020-01-17
[Kubernetes] Refresh Service Account token before it expires


linki:
**Is your feature request related to a problem? Please describe.**

We would like to transition to issuing ServiceAccount tokens that expire, aka [BoundServiceAccountTokenVolumes](https://github.com/kubernetes/kubernetes/issues/70679).

Skipper's Kubernetes data client is currently not aware that ServiceAccount tokens can expire and therefore only reads the token once when the client is initialized. (e.g. [here](https://github.com/zalando/skipper/blob/v0.11.26/dataclients/kubernetes/kube.go#L250) and [here](https://github.com/zalando/skipper/blob/v0.11.26/swarm/kubernetes.go#L108))

We would like for skipper to refresh the token from the mounted volume so that it always has a valid token.

**Describe the solution you would like**

The token should be re-read from the mounted volume while skipper is running. The path to the token within the container doesn't change. This can either be done in fixed intervals or by taking the expiry time within the token into account.

Looking at Kubernetes' TokenManager itself [it refreshes tokens](https://github.com/kubernetes/kubernetes/blob/v1.18.0-alpha.1/pkg/kubelet/token/token_manager.go#L167-L184) once they are past 80% of their lifetime or at least once per day. The default validity of ServiceAccount tokens is 1h so in practice they will be rotated once per hour.

**Describe alternatives you've considered (optional)**

We could use [client-go](https://github.com/kubernetes/client-go) for talking to the Kubernetes API (or a suitable subpackage). Token refresh is handled within the client automatically and it might have other benefits (or drawbacks).

**Would you like to work on it?**

Yes

szuecs:
Sounds good, but we don’t like client-go.
We have already a [secrets module](https://godoc.org/github.com/zalando/skipper/secrets) which you can use to read and update secrets.
I also work on providing an http client in a branch https://github.com/zalando/skipper/blob/feature/skipper-http-client/net/httpclient.go that should also use the secrets module and you can use this client to replace the current kubernetes dataclient http client.

linki:
Thanks, I'll take a look.

linki:
Duplicate https://github.com/zalando/skipper/issues/1283

Issue #1283, opened 2020-01-22
Kubernetes dataclient should read service account token regularly


szuecs:
When we create a new kubernetes dataclient we use `readServiceAccountToken()` once to fetch the token to access k8s API. We should support rotating access tokens, so reading every now and then the token on `LoadUpdate()` an/or `LoadAll()`.

As we want to enable token rotation, we need to update tokens regularly, soon. Maybe we could use our secrets.SecretsReader implementation https://godoc.org/github.com/zalando/skipper/secrets#SecretPaths to fetch them.

PR #1295, opened 2020-02-03
Kubernetes: Refresh ServiceAccount token

linki:
This PR adds support for expiring service account tokens by refreshing them from their mount path regularly.

Fixes https://github.com/zalando/skipper/issues/1277 and https://github.com/zalando/skipper/issues/1283

The implementation is inspired by [Kubernetes way of handling it](https://github.com/kubernetes/kubernetes/blob/v1.17.2/staging/src/k8s.io/client-go/transport/token_source.go#L48-L61).

### How it works

It changes `clusterClient` to use a [`oauth2.TokenSource`](https://godoc.org/golang.org/x/oauth2#TokenSource) instead of a plain string `token`. It then uses a custom `fileTokenSource` that fetches the token from a file, basically rebuilding the current functionality with the exception that it reads the file every single time. Finally, it wraps the file-based TokenSoure into a [`oauth2.ReuseTokenSource`](https://godoc.org/golang.org/x/oauth2#ReuseTokenSource) which takes care of reading the token from cache and only reloading it once it's expired.

Interesting to note here is that the actual expiration time from the token read from the file is ignored and statically set to 1 minute instead, resulting in re-reading the token once per minute. This is consistent with `client-go`'s implementation but doesn't have to be good since `client-go` assumes a token lifetime of 10 minutes in future Kubernetes clusters.

I assume periodically re-reading regardless of the token's expiration time is an attempt to work well in case tokens get revoked (by deleting the ServiceAccount). A local client that only has the token isn't able to detect that case. Of cource, one can try reading a new token if there's an unexpected `401` but that would complicate the code further.

The current implementation strikes a balance between simple and correct. The tokens are mounted as `tmpfs` so the overhead of re-reading should be rather small.

Let me know what you think.

@szuecs 
@aryszka


szuecs:
@linki code looks good but check travis output please

vet: dataclients/kubernetes/kube_test.go:2355:15: undeclared name: readServiceAccountToken

linki:
Thanks, I missed that.

I changed the test to cover the missing file case as before but I removed the case where it's "out-of-cluster" because that's not handled by the TokenSource anymore.

szuecs:
One question would be why not using the internal secrets module, but use a dependency from x?
We can also later change it to our net.Client, that is in development.

dataclients/kubernetes/token_source.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+type fileTokenSource struct {
+	path   string
+	period time.Duration
+}
+
+func (ts *fileTokenSource) Token() (*oauth2.Token, error) {
+	token, err := ioutil.ReadFile(ts.path)
+	if err != nil {
```
aryszka: in case of IO errors, would it make sense to return the previous token? Or is this handled by oauth2.ReuseTokenSource?

aryszka:
generally lgtm. See my question about the IO errors.

I also wonder if it would make sense to make this a part of the "secrets" package, and use it with the net.Client implementation. We can also do that in a separate PR, too.

dataclients/kubernetes/token_source.go
aryszka: discussed IRL, it's fine

aryszka:
:+1:

linki:
Changed implementation to use skipper's `secrets` package.

szuecs:
:+1:

aryszka:
:+1:

PR #1296, opened 2020-02-05
Authorization support for OIDC

universam1:
# AuthZ filter OIDCQueryClaimsFilter

This filter is intended to be chained after one of `OAuthOidc*Claims` OIDC flow provider, making use of the token information saved in `stateBag` to classify **authorization**

* It leverages the query syntax of `GJSON` to express a matching of the token.
* It allows validation per path request, matching one or more queries.
* Each argument of the filter expresses a path/query relation, one or several paths can be defined
* The first positive match wins

Format:
`<path>:[<query>]`

Following example statements would be possible:
```
"/some/path:@_:sub==\"somesub\")"
"/some/path:[@this].#(sub==\"somesub\")"
"/another/path:groups.#(%\"CD-*\")"
"/asdf/:groups.#(%\"CD-*\") 'groups.#(%\"*-Test-Users\")' groups.#(%\"Purchasing-Department\")"
"/:groups.#(%\"Purchasing-Department\")"
```

These would match each one of these:
```json
{
    "sub": "somesub",
    "groups": [
        "CD-Administrators",
        "Purchasing-Department",
        "AppX-Test-Users",
    ]
}
```
ref: https://github.com/tidwall/gjson/issues/151

## generalize and remove cognitive challenge

 * Cognitive challenge reduced by splitting out the stage 5 of the OIDC flow into distinct method
 * generlizing duplicate code paths
 * making use of one struct `tokenContainer` as they can be united through `omitempty` annotation. This is necessary for later filter chain parsing of the statebag

PR #1303, opened 2020-02-12
performance issue sync.Pool flushed by GC

universam1:
Since `sync.Pools` are garbage collected, the current implementation where the pools are warmed up at `init()` are purged after few cycles! This leads to empty pools after around 2 GC cycles (ref: https://go-review.googlesource.com/c/go/+/162919/)

This implementation changes to not pre-create pools but rather use the constructor to create them on demand.
Regarding memory consumption, currently it would be `CPUs x 4` or `num of concurrent threads` whichever is bigger. New implementation would be also maximum number of concurrent threads, but degrades if not used.

The existing benchmark already shows that, regarding the amount of concurrent calls, this implemenation is faster.
```
BenchmarkCompress6-8 (currently)        234	   5687061 ns/op	 3315737 B/op	      61 allocs/op
BenchmarkCompress6-8 (this PR)          292	   3843183 ns/op	 2139794 B/op	      43 allocs/op
```

It is also possible to simulate the `runtime.GC()` which shows the loss of the pool.

refs:
https://go-review.googlesource.com/c/go/+/162919/
https://medium.com/a-journey-with-go/go-understand-the-design-of-sync-pool-2dde3024e277

### DIFF
--- a/filters/builtin/compress.go
+++ b/filters/builtin/compress.go
@@ -7,12 +7,12 @@ import (
 	"io"
 	"math"
 	"net/http"
-	"runtime"
 	"sort"
 	"strconv"
 	"strings"
 	"sync"
 
+	log "github.com/sirupsen/logrus"
 	"github.com/zalando/skipper/filters"
 )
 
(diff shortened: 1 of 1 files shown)


aryszka:
:+1:

szuecs:
:+1:

szuecs:
@universam1 thanks!
Did you saw my [last comment](https://github.com/zalando/skipper/pull/1296#issuecomment-584523595) on your question about the "T-shirt bounty program"?

PR #1307, opened 2020-02-14
Header logger

aryszka:
create a logHeader() filter

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -418,7 +418,9 @@ It logs with INFO level and a unique ID per request:
 ## logHeader
 
 The logHeader filter prints the request line and the header, but not the boddy, to
-stderr.
+stderr. Note that this filter should be used only in diagnositcs setup and with care,
+since the request headers may contain sensitive data, and they also can explode the
+amount of logs.
 
 ## tee
 
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -418,7 +418,7 @@ It logs with INFO level and a unique ID per request:
 ## logHeader
 
 The logHeader filter prints the request line and the header, but not the body, to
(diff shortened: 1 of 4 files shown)


docs/reference/filters.md
```diff
@@ -415,6 +415,13 @@ It logs with INFO level and a unique ID per request:
 - the finishing event of the request
 - any read errors other than EOF
 
+## logHeader
+
+The logHeader filter prints the request line and the header, but not the boddy, to
+stderr. Note that this filter should be used only in diagnositcs setup and with care,
```
szuecs: diagnositcs -> diagnostics

szuecs:
@aryszka maybe drop the Authorization header from the output or truncate the value?

aryszka:
thought it, but it can be sometimes the goal of the diagnostics, and also there can be other headers containing sensitive data. What do you think about a more drastic safeguard, e.g. a global switch as `-enable-sensitive-diagnostics`? This probably would be the safest way, if we want to keep this filter, but I'm also fine with dropping the PR.

docs/reference/filters.md
```diff
@@ -415,6 +415,13 @@ It logs with INFO level and a unique ID per request:
 - the finishing event of the request
 - any read errors other than EOF
 
+## logHeader
+
+The logHeader filter prints the request line and the header, but not the boddy, to
```
szuecs: boddy -> body

aryszka: right, thx!

docs/reference/filters.md
aryszka: thx!

szuecs:
:+1:

szuecs:
:+1:

aryszka:
:+1:

Issue #1310, opened 2020-02-17
Routes using tee() combined with <shunt> cause an increase in goroutines


topicus:
**Describe the bug**
Bodies of tee'd requests are not cleaned up when shunting responses. This causes the following goroutine to being executed indefinitely https://github.com/zalando/skipper/blob/master/filters/tee/tee.go#L155-L169

**To Reproduce**
1. route or ingress definition:
```
example_route: *
   -> tee("http://localhost:8080")
   -> <shunt>;
```
2. request with a body
```
curl -d "{}" http://localhost:9990/
```
3. response with headers 
Depending on how strict is the backend configuration the request may fail or succeed. However, this does not affect the leaking behavior.

**Expected behavior**
Body of tee'd requests should be closed. This will avoid the increase in the goroutines.

**Observed behavior**
The body of the tee'd request is never closed causing a memory leak and depending on the traffic to the route, this can lead to goroutines' spike. This could lead to skipper crashing.

szuecs:
@topicus thanks for the issue, we could reproduce it. Problem is that `<shunt>` does not read the body and `tee()` filter use a Pipe to read the body when the the backend processes the request body.

aryszka:
Fixed in: https://github.com/zalando/skipper/pull/1314

PR #1311, opened 2020-02-19
Close request body on shunted requests

topicus:
Exposes Shunt() in the filter context. This method is used to check is the route is shunted in the tee filter. If that is the case then we close the req.body to prevent leaks.

Fixes: https://github.com/zalando/skipper/issues/1310

### DIFF
--- a/filters/tee/tee.go
+++ b/filters/tee/tee.go
@@ -160,6 +160,12 @@ func (r *tee) Request(fc filters.FilterContext) {
 		}()
 
 		rsp, err := r.client.Do(copyOfRequest)
+
+		// Prevent to leak goroutines when request body is never read.
+		if fc.Shunted() {
+			defer copyOfRequest.Body.Close()
+		}
+
 		if err != nil {
 			log.Warn("tee: error while tee request", err)
 			return
--- a/filters/filters.go
+++ b/filters/filters.go
@@ -96,6 +96,9 @@ type FilterContext interface {
 
 	// Allow filters to create their own spans
 	ParentSpan() opentracing.Span
+
+	// Returns if the current response is shunted.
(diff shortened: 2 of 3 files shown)


filters/tee/tee.go
```diff
@@ -160,6 +160,12 @@ func (r *tee) Request(fc filters.FilterContext) {
 		}()
 
 		rsp, err := r.client.Do(copyOfRequest)
+
+		// Prevent to leak goroutines when request body is never read.
+		if fc.Shunted() {
+			defer copyOfRequest.Body.Close()
```
szuecs: As far as I remember the pipe is written to, as soon the original body is read.
Therefore I think you should read the body, because of the pipe and you also want to execute the tee() call and stream the body, right?

topicus: I think that path works fine. If you have a backend, then the req.body is read and closed properly. 
On the other hand, if you shunt the response, it means that you are not interested in the response. Hence, we can close it. 

Also, I am now thinking that the defer does not have so much sense as the intention is to close the request immediately. 

**To summarize the intention:**
- If the route is shunted then, as soon as we applied this filter, we close the request body. In other words, the tee filter triggers a side effect. The client receives a 404 but there is a request executed from skipper to the backend we are teeing.

- If the route is not shunted then the original request is read and pipes the body to the copy of the request. It means that goes through the Read method of teeTie which consumes the body, pipes the contents and closes the pipe. This is the most common flow.

Let me know if this does it make sense.

topicus: I will close this PR as we discussed better ways to fix this with @aryszka.

topicus:
Closing this PR as we discussed better ways to address this problem with @aryszka

Issue #1312, opened 2020-02-20
compress filter can panic


aryszka:
**Describe the bug**
GZip compression may panic sometimes.

**To Reproduce**

1. use a backend that returns uncompressed data, with Content-Type: application/octet-stream 
2. create a route to this backend that includes the compress() filter
3. generate high concurrent load to this route where the requests have the Accept-Encoding: gzip header

**Expected behavior**

compressed responses without problems

**Observed behavior**

rarely, depending on the concurrency level, the following panics are logged from skipper and damaged responses returned:

```
panic: runtime error: index out of range [108] with length 35

goroutine 294066 [running]:
compress/flate.byFreq.Less(...)
	/home/aryszka/go/src/compress/flate/huffman_code.go:339
sort.insertionSort(0x1258840, 0xc00ed08e50, 0x6b, 0x70)
	/home/aryszka/go/src/sort/sort.go:27 +0xc4
sort.quickSort(0x1258840, 0xc00ed08e50, 0x6b, 0x70, 0x9)
	/home/aryszka/go/src/sort/sort.go:209 +0x201
sort.quickSort(0x1258840, 0xc00ed08e50, 0x6b, 0x7c, 0x9)
	/home/aryszka/go/src/sort/sort.go:194 +0xf8
sort.quickSort(0x1258840, 0xc00ed08e50, 0x6b, 0x96, 0xa)
	/home/aryszka/go/src/sort/sort.go:194 +0xf8
sort.Sort(0x1258840, 0xc00ed08e50)
	/home/aryszka/go/src/sort/sort.go:218 +0x79
compress/flate.(*byFreq).sort(...)
	/home/aryszka/go/src/compress/flate/huffman_code.go:333
compress/flate.(*huffmanEncoder).generate(0xc00ed08dc0, 0xc0100e2000, 0x11e, 0x11e, 0xf)
	/home/aryszka/go/src/compress/flate/huffman_code.go:306 +0x188
compress/flate.(*huffmanBitWriter).writeBlockHuff(0xc00ed0ad20, 0xc01012c000, 0xc0100e4000, 0x328, 0xffff)
	/home/aryszka/go/src/compress/flate/huffman_bit_writer.go:639 +0xf7
compress/flate.(*compressor).encSpeed(0xc010038000)
	/home/aryszka/go/src/compress/flate/deflate.go:363 +0x285
compress/flate.(*compressor).syncFlush(0xc010038000, 0x328, 0x2000)
[...]

PR #1313, opened 2020-02-20
close encoder before putting back to the pool

aryszka:

fixing: https://github.com/zalando/skipper/issues/1312

### DIFF
--- a/filters/builtin/compress.go
+++ b/filters/builtin/compress.go
@@ -298,9 +298,11 @@ func encode(out *io.PipeWriter, in io.ReadCloser, enc string, level int) {
 		err error
 	)
 
+	backToPool := func() {}
 	defer func() {
 		if e != nil {
 			e.Close()
+			backToPool()
 		}
 
 		if err == nil {
@@ -316,7 +318,7 @@ func encode(out *io.PipeWriter, in io.ReadCloser, enc string, level int) {
 		pe := pool.Get()
 		if pe != nil {
 			e = pe.(encoder)
-			defer pool.Put(pe)
+			backToPool = func() { pool.Put(pe) }
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


filters/builtin/compress.go
```diff
@@ -316,7 +318,7 @@ func encode(out *io.PipeWriter, in io.ReadCloser, enc string, level int) {
 		pe := pool.Get()
 		if pe != nil {
 			e = pe.(encoder)
-			defer pool.Put(pe)
+			backToPool = func() { pool.Put(pe) }
```
aermakov-zalando: This whole function is rather complicated to follow. I'd simplify it a bit:
 * Drop the `backToPool` variable.
 * Move the `level` check and putting the encoder to the main closing defer. It's also a good idea to check the error from `e.Close()` and refuse to pool the encoder back.
 * Use explicit `if/else` statements instead of re-checking for `if e == nil`.
 * `pool.Get()` can't return a `nil` if used with `p.New`, so just drop this thing altogether (or simply panic).

aermakov-zalando: It might also be a good idea to `Reset` _before_ putting things back into the pool instead of the other way around.

filters/builtin/compress.go
```diff
@@ -312,17 +315,17 @@ func encode(out *io.PipeWriter, in io.ReadCloser, enc string, level int) {
...
-	if e == nil {
+		// if the pool.New failed to create an encoder,
+		// then we already have logged the error
+		if e == nil {
+			return
+		}
+	} else {
 		e, err = newEncoder(enc, level)
 		if err != nil {
+			log.Error(err)
```
aermakov-zalando: Shouldn't this return the error?

aermakov-zalando:
👍

aryszka:
:+1:

PR #1314, opened 2020-02-22
Read and discard request body on shunt

topicus:
As we discussed with @aryszka , reading the request body matches better with the expected behavior of the shunt backend. In other words, default to reading the request body on any case. 

I am not sure if this is the most efficient way to read and discard the request body.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"net/http/httputil"
@@ -1045,6 +1046,7 @@ func (p *Proxy) do(ctx *context) error {
 		p.log.Debugf("deprecated shunting detected in route: %s", ctx.route.Id)
 		return &proxyError{handled: true}
 	} else if ctx.shunted() || ctx.route.Shunt || ctx.route.BackendType == eskip.ShuntBackend {
+		io.Copy(ioutil.Discard, ctx.request.Body)
 		ctx.ensureDefaultResponse()
 	} else if ctx.route.BackendType == eskip.LoopBackend {
 		loopCTX := ctx.clone()
--- a/proxy/proxy_test.go
+++ b/proxy/proxy_test.go
@@ -883,9 +883,11 @@ func TestNilFilterIsNotCalledAndDoesNotBreakFilterChain(t *testing.T) {
 
 func TestProcessesRequestWithShuntBackend(t *testing.T) {
 	u, _ := url.ParseRequestURI("https://www.example.org/hello")
+	reqBody := strings.NewReader("sameple request body")
 	r := &http.Request{
 		URL:    u,
-		Method: "GET",
+		Method: "POST",
+		Body:   ioutil.NopCloser(reqBody),
 		Header: http.Header{"X-Test-Header": []string{"test value"}}}
 	w := httptest.NewRecorder()
 
@@ -906,6 +908,11 @@ func TestProcessesRequestWithShuntBackend(t *testing.T) {
 	if h, ok := w.Header()["X-Test-Response-Header"]; !ok || h[0] != "response header value" {
 		t.Error("wrong response header")
 	}
+	_, err = reqBody.ReadByte()
+	if io.EOF != err {
+		t.Error("request body was not read")
+	}
+
 }
 
 func TestProcessesRequestWithPriorityRoute(t *testing.T) {


proxy/proxy.go
```diff
@@ -1045,6 +1046,10 @@ func (p *Proxy) do(ctx *context) error {
 		p.log.Debugf("deprecated shunting detected in route: %s", ctx.route.Id)
 		return &proxyError{handled: true}
 	} else if ctx.shunted() || ctx.route.Shunt || ctx.route.BackendType == eskip.ShuntBackend {
+		// consume the body to prevent goroutine leaks
+		if ctx.request.Body != nil {
+			io.Copy(ioutil.Discard, ctx.request.Body)
```
szuecs: body should be closed, too.

topicus: Good catch!

szuecs:
@aryszka and I talked about the change and it's also possible with some other backends to trigger the leak, for example `tee() -> <loopback>`. 
Maybe the possible code paths are more seldom, and I would say let's fix it like this.

wdyt @aryszka and @topicus ?

proxy/proxy_test.go
```diff
@@ -883,9 +883,11 @@ func TestNilFilterIsNotCalledAndDoesNotBreakFilterChain(t *testing.T) {
 
 func TestProcessesRequestWithShuntBackend(t *testing.T) {
 	u, _ := url.ParseRequestURI("https://www.example.org/hello")
+	reqBody := strings.NewReader("sameple request body")
```
addityasingh:
```suggestion
	reqBody := strings.NewReader("sample request body")
```

topicus:
@szuecs 
>  I would say let's fix it like this.
Do you mean to merge this PR as is or also add the loopback to this PR?

aryszka:
I think that the current approach actually covers the cases with the different backend types, because even with multiple <loopback>s, eventually there can be only a single <shunt>. What is not covered in the current PR, is the different cases of errors, e.g. when the backend connection fails. So, we can say it's still incomplete, and I would prefer to have these cases handled in this PR.

proxy/proxy.go
aryszka: actually, no :)

See the docs for the Body field of the Request type: https://pkg.go.dev/net/http/?tab=doc#Request

aryszka:
please see the DCO, error: https://github.com/zalando/skipper/pull/1314/checks?check_run_id=467220735

(we basically need to use the --signoff flag when committing)

addityasingh:
> (we basically need to use the --signoff flag when committing)

@aryszka The DCO blocks people from using github suggestions and committing directly in the web UI. The only workflow I found to provide signed off commits is https://github.com/todogroup/gh-issues/issues/50#issuecomment-330144236 but this needs to be documented

aryszka:
@addityasingh true, and I am sorry about this inconvenience. I am not a great fan of it, either, it's just more like a compliance thing.

aryszka:
@topicus i tried the following: I put the below code snippet in proxy.go, in the ServeHTTP method, line 1323:

```
	if _, err := io.Copy(ioutil.Discard, ctx.request.Body); err != nil {
		p.log.Errorf("Error while discarding remainder request body: %v.", err)
	}
```

and the it works in most of the cases: shunt, connection error, with and without tee(). I would suggest to start from this approach, and let's verify if this covers every case and doesn't break anything.

Then in a next step, let's also think about if need to do discard the remainder body in each case. Probably, if the request body on the context is the original, then we don't need to this, and we can save some io and also returning earlier in cases like a 404.

WDYT?

proxy/proxy_test.go
```diff
@@ -957,6 +959,11 @@ func TestProcessesRequestWithShuntBackend(t *testing.T) {
 	if h, ok := w.Header()["X-Test-Response-Header"]; !ok || h[0] != "response header value" {
 		t.Error("wrong response header")
 	}
+	_, err = reqBody.ReadByte()
+	if io.EOF != err {
```
szuecs, line 963: Please swap the condition:

err != io.EOF

topicus, line 963:
```suggestion
	if err != io.EOF {
```

topicus:
👍

queuelistener/listener.go
```diff
@@ -101,12 +101,15 @@ func (c *connection) SetReadDeadline(t time.Time) error  { return c.net.SetReadD
 func (c *connection) SetWriteDeadline(t time.Time) error { return c.net.SetWriteDeadline(t) }
 
 func (c *connection) Close() error {
-	select {
-	case c.release <- token:
-	case <-c.quit:
-	}
+	c.once.Do(func() {
```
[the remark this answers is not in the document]
szuecs: close should always be exactly once, otherwise we risk panic at runtime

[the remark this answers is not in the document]
szuecs: I guess there is a missing rebase master in the PR or so

aryszka:
:+1:

szuecs:
:+1:

PR #1316, opened 2020-02-25
fixed type in ingress-usage.md

ResuBaka:
I just found while reading the docs that **authentication** was a bit miss typed as `autnetication`. 

So here is the fix for it.


szuecs:
:+1:

szuecs:
Thanks !

aryszka:
:+1:

PR #1323, opened 2020-02-29
Add modRequestHeader filter

ruudk:
This allows you to do things like this:
```
add_www: Host("^example\.") -> modHost("example.(\w+)", "www.example.$1") -> redirectTo(301) -> <shunt>;
```
or
```
add_www: Host("^example\.") -> modHost("(.*)", "www.$1") -> redirectTo(301) -> <shunt>;
```

```
curl -I -H 'Host: example.com' localhost:9090/hello
HTTP/1.1 301 Moved Permanently
Location: https://www.example.com/hello
Server: Skipper
Date: Sat, 29 Feb 2020 10:26:56 GMT
```

## Todo

- [x] Add more tests
- [x] Add documentation

### DIFF
--- a/filters/builtin/mod_header.go
+++ b/filters/builtin/mod_header.go
@@ -1,0 +1,75 @@
+package builtin
+
+import (
+	"net/http"
+	"regexp"
+	"strings"
+
+	"github.com/zalando/skipper/filters"
+)
+
+type modRequestHeader struct {
+	headerName  string
+	rx          *regexp.Regexp
+	replacement string
+}
+
+// NewModRequestHeader returns a new filter Spec, whose instances execute
+// regexp.ReplaceAllString on the request host. Instances expect three
+// parameters: the header name, the expression to match and the replacement string.
+// Name: "modRequestHeader".
+func NewModRequestHeader() filters.Spec { return &modRequestHeader{} }
+
+func (spec *modRequestHeader) Name() string {
+	return ModRequestHeaderName
+}
+
+//lint:ignore ST1016 "spec" makes sense here and we reuse the type for the filter
+func (spec *modRequestHeader) CreateFilter(config []interface{}) (filters.Filter, error) {
+	if len(config) != 3 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	headerName, ok := config[0].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	expr, ok := config[1].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	replacement, ok := config[2].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	rx, err := regexp.Compile(expr)
+	if err != nil {
+		return nil, err
+	}
+
+	return &modRequestHeader{headerName: headerName, rx: rx, replacement: replacement}, nil
+}
+
+func (f *modRequestHeader) Request(ctx filters.FilterContext) {
+	req := ctx.Request()
+
+	if strings.ToLower(f.headerName) == "host" {
+		nh := f.rx.ReplaceAllString(getRequestHost(req), f.replacement)
+
+		req.Header.Set(f.headerName, nh)
+		ctx.SetOutgoingHost(nh)
+
+		return
+	}
+
+	if _, ok := req.Header[http.CanonicalHeaderKey(f.headerName)]; !ok {
+		return
+	}
+
+	req.Header.Set(f.headerName, f.rx.ReplaceAllString(req.Header.Get(f.headerName), f.replacement))
+}
+
+func (*modRequestHeader) Response(filters.FilterContext) {}
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -63,17 +63,9 @@ Example:
 foo: * -> setRequestHeader("X-Passed-Skipper", "true") -> "https://backend.example.org";
 ```
 
-## setResponseHeader
-
-Same as [setRequestHeader](#setrequestheader), only for responses
-
 ## appendRequestHeader
 
 Same as [setRequestHeader](#setrequestheader), does not remove a possibly existing value, but adds a new header value
-
-## appendResponseHeader
-
-Same as [appendRequestHeader](#appendrequestheader), only for responses
 
 ## dropRequestHeader
 
@@ -88,6 +80,14 @@ Example:
 ```
 foo: * -> dropRequestHeader("User-Agent") -> "https://backend.example.org";
 ```
+
+## setResponseHeader
+
+Same as [setRequestHeader](#setrequestheader), only for responses
+
+## appendResponseHeader
+
+Same as [appendRequestHeader](#appendrequestheader), only for responses
 
 ## dropResponseHeader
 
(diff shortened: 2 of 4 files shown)


filters/builtin/host.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// Modifies the host with regexp.ReplaceAllString.
+func (f *modHost) Request(ctx filters.FilterContext) {
+	req := ctx.Request()
+	switch f.behavior {
+	case regexpReplaceHost:
+		req.Host = f.rx.ReplaceAllString(req.Host, f.replacement)
+	case fullReplaceHost:
+		req.Host = f.template.Apply(ctx.PathParam)
```
ruudk: What to do with this? Should we remove this whole `fullReplaceHost` functionality? Or can we somehow get `ctx.HostParam`?

szuecs:
@ruudk I wonder if it would make more sense to add modRequestHeader (and maybe modResponseHeader ?).
Syntax could be: `modRequestHeader("Host", "(.*)", "www.$1")`.
What do you think?

filters/builtin/host.go
szuecs: I guess it should be dropped, in case you decide to go with my `modeRequestHeader` idea.

ruudk:
Changed it to `ModRequestHeader`. But for `Host` I need to do a special trick, because by default the `Host` header does not exist, and it can be either part of `req.Host` or `req.URL.Host`. Is this how it should be done?

filters/builtin/mod_header.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	return &modRequestHeader{headerName: headerName, rx: rx, replacement: replacement}, nil
+}
+
+// Modifies the host with regexp.ReplaceAllString.
+func (f *modRequestHeader) Request(ctx filters.FilterContext) {
+	req := ctx.Request()
+
+	if f.headerName == "Host" {
```
aryszka: In other places, we ignore the case of the header names coming from filter parameters. Can you please change it to behave similar?

filters/builtin/mod_header.go
```diff
@@ -0,0 +1,80 @@
...
+		h := req.Header.Get("Host")
+
+		if h == "" && req.Host != "" {
+			req.Host = f.rx.ReplaceAllString(req.Host, f.replacement)
```
aryszka: we have the SetOutgoingHost() on the filter context. Do you think it would work here, too? See: https://github.com/zalando/skipper/blob/master/filters/builtin/headerfilter.go#L153

and the SetOutgoingHost() documentation here: https://godoc.org/github.com/zalando/skipper/filters#FilterContext

I think using the SetOutgoingHost() will also make it consistent with the behavior of the -proxy-preserve-host global option and the preserveHost() predicate.

filters/builtin/mod_header.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/filters"
+)
+
+type modRequestHeader struct {
+	headerName  string
+	rx          *regexp.Regexp
+	replacement string
+	template    *eskip.Template
```
aryszka: This field doesn't seem to be used. Can you please delete it? We can improve the filter to use templates in a separate pass.

(`make staticcheck` detects the unused field, not sure now if this is only since we upgraded to go1.14)

aryszka:
thx for the PR, @ruudk ! Please see my comment regarding the ctx.SetOutgoingHost() usage and the linting problem.

filters/builtin/mod_header.go
ruudk: Removed.

filters/builtin/mod_header.go
ruudk: Applied your feedback. WDYT?

ruudk:
@aryszka Applied your feedback in a1bc08e. Would this work?

filters/builtin/mod_header.go
aryszka: lgtm!

aryszka:
thanks for the update!

aryszka:
:+1:

ruudk:
I'm flattered that it already looks good to you, but there must be some things that I can improve? Should I add more tests? Should I add documentation?

aryszka:
Indeed, I missed the documentation :) This filter should be described in the reference section, next to the existing header changing filters: https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#setrequestheader

About testing, I think we can address more cases:

- non-host headers
- the case when for some reason only the req.Host is set
- the case when for some reason only the req.URL.Host is set

Btw, I think about these latter two cases rather as defensive logic, because normally they won't happen, unless some custom filter or other extension provides such request objects.

ruudk:
I had to do a rebase to get my #1324 feature in. 

Applied all feedback.

I'm still a bit in doubt how things should be handled when you want to replace a header that does not exist. 

In case of `Host` header it will take the Host from [getRequestHost](https://github.com/zalando/skipper/blob/5c5bf42b4b5a0d5d821a50b9c3a08d8b4866e960/filters/builtin/redirect.go#L107), try to replace it, and will insert the new host as Header + OutgoingHost. So if the replacement is empty string, it will be an empty string Host header and empty string OutgoingHost.

But for other headers, it will early return when the header does not exist, because it doesn't match then anyway. 

This is not consistent. What do you suggest?

aryszka:
The handling of the Host header sounds fine to me, primarily because there's no such request that doesn't have a host header, up from HTTP/1.1. I hope that the current approach also supports your original use case.

Regarding the other headers, I'd make a small change. Instead of using net/http.Header.Get, I'd check whether the key exists. Sg like this:

```
if _, ok := req.Header[http.CanonicalHeaderKey(f.headerName)]; !ok {
	return
}
```

This way, we only change the header if it was defined and matches the pattern, even if it's set to empty for some reason.

Since nobody uses HTTP/1.0, and it really only can happen with non-host headers that they are not defined, I think this approach can be considered consistent. WDYT?

ruudk:
@aryszka Good one. Applied it :)

ruudk:
Do you want me to  rebase this PR and make 1 nice commit or do you squash it?

filters/builtin/mod_header.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/zalando/skipper/filters"
+)
+
+type modRequestHeader struct {
+	headerName  string
+	rx          *regexp.Regexp
+	replacement string
+}
+
+// Returns a new modRequestHeader filter Spec, whose instances execute
```
szuecs: Go doc wants that you start all doc comments by the function name.
So:

NewModRequestHeader returns a new filter Spec, whose instances execute...

filters/builtin/mod_header.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	if _, ok := req.Header[http.CanonicalHeaderKey(f.headerName)]; !ok {
+		return
+	}
+
+	req.Header.Set(f.headerName, f.rx.ReplaceAllString(req.Header.Get(f.headerName), f.replacement))
+}
+
+// Noop.
```
szuecs: just drop the comment :)

ruudk: Done

filters/builtin/mod_header.go
ruudk: Done

ruudk:
Test is failing because of race issue.

szuecs:
:+1:

aryszka:
:+1:

szuecs:
:+1:

PR #1325, opened 2020-02-29
Add FastCGI backend

### LINKED ISSUE #1322: Route requests directly to PHP via FastCGI
When running a PHP application you need `php-fpm`, a fast CGI server. On order to serve traffic to it, you need something that is able to convert HTTP requests to FastCGI requests, like Nginx or Apache.

But having Nginx running only to pass the request coming from Skipper directly to PHP feels like a waste of resources. 

It would be super cool if Skipper would support [FastCGI](https://github.com/yookoala/gofast) so that I can directly connect Skipper to PHP-FPM.

Yes, with some guidance.

ruudk:
Fixes #1322 

**Work in Progress**

## How to test?

Start an example FastCGI server:
```
docker-compose --file=fastcgiserver/docker-compose.yml up --build
```

Then run Skipper like this:
```
go run ./cmd/skipper -inline-routes='* -> "fastcgi://127.0.0.1:9000"'
```
or
```
go run ./cmd/skipper -inline-routes='* -> <roundRobin, "fastcgi://127.0.0.1:9000">'
```

And then:
```
$ curl -i localhost:9090
HTTP/1.1 200 OK
Content-Type: text/html; charset=UTF-8
Server: Skipper
X-Powered-By: PHP/7.4.3
Date: Wed, 04 Mar 2020 08:40:08 GMT
Transfer-Encoding: chunked

Hello, World!⏎
```

## Todo

- [x] Find better solution for httptest.NewRecorder 
- [x] Add tests
- [ ] Add docs


ruudk:
I think the `FastCGIAddress` should be dynamic and not hard coded like this.

proxy/proxy.go
```diff
@@ -882,7 +885,24 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
...
+		connFactory := gofast.SimpleConnFactory("tcp", ctx.route.FcgiAddress)
+
+		rr := httptest.NewRecorder()
+
+		gofast.NewHandler(
+			gofast.NewFileEndpoint("index.php")(gofast.BasicSession),
+			gofast.SimpleClientFactory(connFactory, 0),
+		).ServeHTTP(rr, req)
+
+		response = rr.Result()
```
szuecs: Is it possible to use client connections pooling and then: https://godoc.org/github.com/yookoala/gofast#ClientFunc.Do instead of this hack (use of httptest.Recorder to get the response from an inline created handler )?

szuecs: You already highlighted as todo. :)

ruudk: Yeah I was wrestling with this. The `gofast` library uses a [ResponsePipe](https://godoc.org/github.com/yookoala/gofast#ResponsePipe) and Skipper wants a regular Response. Here my knowledge of Go limited me. How would you pull this off?

ruudk:
I refactored it and made it much simpler. Instead of creating a new backend type, I now rely on the scheme. This makes it much easier to work with. Plus it should also work out of the box with Kubernetes. WDYT @szuecs @aryszka ?

proxy/proxy.go
szuecs: Then have a type that makes it possible to get the http.Response from the ResponsePipe.
Best would be to use the pooled connections provided by https://godoc.org/github.com/yookoala/gofast#NewClientPool

proxy/proxy.go
```diff
@@ -882,7 +883,22 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
...
 	p.metrics.IncCounter("outgoing." + req.Proto)
 	ctx.proxySpan.LogKV("http_roundtrip", StartEvent)
-	response, err := p.roundTripper.RoundTrip(req)
+
+	var response *http.Response
+	switch req.URL.Scheme {
+	case "fastcgi":
+		rt, err := NewFastCgiRoundTripper(p.log, req.URL.Host)
+		if err != nil {
+			p.log.Errorf("could not create fastcgi roundtripper: %v", err)
```
szuecs: In case you log, start with uppercase letter, in case you return error you start with lowercase.
I suggest:
`"Failed to create fastcgi roundtripper: %v"`

proxy/fastcgi.go
```diff
@@ -0,0 +1,53 @@
...
+
+func NewFastCgiRoundTripper(log logging.Logger, addr string) (FastCgiRoundTripper, error) {
+	connFactory := gofast.SimpleConnFactory("tcp", addr)
+
+	c, err := gofast.SimpleClientFactory(connFactory, 0)()
+	if err != nil {
+		return nil, fmt.Errorf("gofast: failed creating client: %w", err)
+	}
+
+	handler := gofast.NewFileEndpoint("index.php")(gofast.BasicSession)
```
szuecs: "index.php" should be parameterized

szuecs:
@ruudk I am fine with the Scheme introspection style dispatch.

@aryszka wdyt?

proxy/fastcgi.go
ruudk: Fixed

proxy/proxy.go
ruudk: Fixed

aryszka:
I am not sure about the scheme, but only because I don't know the fastcgi possibilities well enough. E.g in this documentation: https://www.digitalocean.com/community/tutorials/understanding-and-implementing-fastcgi-proxying-in-nginx , i can also see an example like `unix:/var/run/php5-fpm.sock` where unix is the schema.

So my questions would be:

- are there any solutions out there that use this schema 'fastcgi:'?
- is it possible that we prevent some future additions/improvements to this feature if we settle on this schema?
- why is the solution simpler with the 'fastcgi:' schema, than with a backend type?

I definitely need to read up a bit about the fastcgi topic, to be able to confidently judge this approach. However, it doesn't seem to be bad altogether.

szuecs:
I thought about it and I think the backend type seems to be cleaner. We could also think in the future of implementing other backend types of the type loadbalancer backend for example grpc, tls, tcp and use fastcgi as a first one.
I would propose a change in the LB backend definition, such that we have a `type`:

```
 <roundRobin, type="http", "http://10.2.0.234:9090", "http://10.2.1.227:9090">
```

`type` would be optional to fallback to "http" as default and the first non default implementation would be "fcgi".

For fastcgi with unix domain socket we could even do:
```
 <roundRobin, type="fcgi", "unix://my/socket1", "unix://my/socket2">
```

@aryszka @ruudk what do you think?

ruudk:
Nice! I didn’t think about adding the type to the backend. That was my initial problem: if I create `<fastcgi>` backend how to do round robin for example.

I’ll work on some improvements :)

aryszka:
I partially like @szuecs 's proposal, but not entirely. My main concern is that it introduces some kind of a subtype, or more precisely, a notation where the type is represented in a different way as before.

What I really agree with is to move from handling the load balancing as a backend type, and making it general. This is really good. Then we can say that whenever more than one endpoints are defined, we always make load balancing, by default roundRobin, unless another algorithm is specified explicitly.

But there are backends that don't have endpoints and therefore don't have algorithm, so I think the algorithm should not be in the main position of the notation, but rather the type which is most often the protocol. So better some of these:

```
<fastcgi, roundRobin, "http://10.2.0.234:9090", "http://10.2.1.227:9090">
<http, roundRobin, "http://10.2.0.234:9090", "http://10.2.1.227:9090">
<roundRobin, "http://10.2.0.234:9090", "http://10.2.1.227:9090">
```

```
<fastcgi, lbalgorithm="roundRobin", "http://10.2.0.234:9090", "http://10.2.1.227:9090">
<http, lbalgorithm="roundRobin", "http://10.2.0.234:9090", "http://10.2.1.227:9090">
<lbalgorithm="roundRobin", "http://10.2.0.234:9090", "http://10.2.1.227:9090">
```

http backend could be the default. WDYT?

Also a question, when using fastcgi, isn't it still http in some way?

szuecs:
@aryszka I like your first proposed syntax better than the second, so +1 for having this style:

``` 
<fastcgi, roundRobin, "http://10.2.0.234:9090", "http://10.2.1.227:9090">
<http, roundRobin, "http://10.2.0.234:9090", "http://10.2.1.227:9090">
<roundRobin, "http://10.2.0.234:9090", "http://10.2.1.227:9090">
```

aryszka:
ok

i forgot to mention, but it may be obvious. If we handle it this way, then:

```
"http://10.1.0.1:8080"
```

becomes just a syntax sugar for:

```
<http, roundRobin, "http:10.1.0.1:8080">
```

ruudk:
> Also a question, when using fastcgi, isn't it still http in some way?

I don't think so, see https://en.wikipedia.org/wiki/FastCGI.

Btw, instead of `(backend) type` we should call this `protocol`. By default it will be `http` protocol. But you can change it to `fastcgi` protocol.

I wonder why you want to even support sockets? Skipper can be opinionated right? Right now there is no fastcgi, and this PR makes it possible to have it but with host:port. 

That's why I liked my `fastcgi://1.1.1.1:222` idea. It makes it super clear that we are not talking about `http` protocol anymore. 

```
<fastcgi, roundRobin, "http://10.2.0.234:9090", "http://10.2.1.227:9090">
```

Kinda weird that it says `fastcgi` protocol and then starts lists `http://` later.

ruudk:
If in the future somebody really wants fastcgi over sockets, we can introduce `fastcgi-unix://my/socket1`?

aryszka:
ok, I see, thx. I am in for "fastcgi://1.1.1.1:222"

szuecs:
To summarize, the decision as far as I understand is:
```
<roundRobin, "http://10.2.0.234:9090", "http://10.2.1.227:9090">
<roundRobin, "fascgi://10.2.0.234:9090", "fascgi://10.2.1.227:9090">
```

Do we want to support mixed protocol scheme (no seems more natural to me)?

aryszka:
@szuecs the summarized approach sounds fine to me.

> mixed protocols
I think it's better not to support this, at least until somebody explicitly wants to use it, as we may end up with hard-to-solve situations in the implementation.

ruudk:
What do you think about this d2c36e1e5e1485d41f0e481b5001e199eb660728 to prevent mixed protocols?

eskip/eskip.go
```diff
@@ -75,6 +76,8 @@ const (
 	LBBackend
 )
 
+var errLBEndpointsShouldHaveSameProtocol = errors.New("loadbalancer endpoints should have the same protocol")
```
ruudk:
```suggestion
var errMixedProtocols = errors.New("loadbalancer endpoints should have the same protocol")
```

proxy/proxy.go
ruudk: I decided to not do this. My first approach was to copy `httptest.NewRecorder` but what's the point of completely copying the httptest recorder? Can't we just use this? It gets the job done, and is battle tested.

ruudk:
@szuecs @aryszka The PR is now ready for review. I added the tests. Right now, there are no docs that explain on how to use this. This is intentional. Let's first test this in production :)

ruudk:
Todo:
- [ ] test Kubernetes integration, how to change the protocol? Introduce annotation?

szuecs:
@ruudk I think an annotation is fine for this. Like `zalando.org/backend-protocol: fastcgi` and defaults to http

proxy/context.go
```diff
@@ -200,6 +201,17 @@ func (c *context) SetOutgoingHost(h string)            { c.outgoingHost = h }
...
+
+func (c *context) FastCgiFilename() string {
+	if c.fastCgiFilename == "" {
+		return "index.php"
+	}
+
+	return c.fastCgiFilename
+}
+
+func (c *context) SetFastCgiFilename(f string) { c.fastCgiFilename = f }
```
szuecs: I think it would be better not to change the FilterContext (see travis errors), but use a default filename as you have and use filters to configure it.
Just add the filename into the stateBag in the request filter and dispatch in the proxy.

filters/builtin/fastcgi.go
```diff
@@ -0,0 +1,47 @@
...
+		return setFastCgiFilenameSpec{a}, nil
+	}
+
+	return nil, filters.ErrInvalidFilterParameters
+}
+
+func (s setFastCgiFilenameSpec) Response(_ filters.FilterContext) {}
+
+func (s setFastCgiFilenameSpec) Request(ctx filters.FilterContext) {
+	ctx.SetFastCgiFilename(s.fileName)
```
szuecs: `c.StateBag()["fcgiFilename"] = s.fileName`

proxy/proxy.go
```diff
@@ -882,7 +884,27 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
 
 	p.metrics.IncCounter("outgoing." + req.Proto)
 	ctx.proxySpan.LogKV("http_roundtrip", StartEvent)
-	response, err := p.roundTripper.RoundTrip(req)
+
+	var response *http.Response
+	switch req.URL.Scheme {
+	case "fastcgi":
+		rt, err := fastcgi.NewRoundTripper(p.log, req.URL.Host, ctx.FastCgiFilename())
```
szuecs: fcgiFilename should be a const.
```
fname := builtin.DefaultFCGIName
if s, ok := c.StateBag()["fcgiFilename"]; ok {
    fname = string(s)
}
rt, err := fastcgi.NewRoundTripper(p.log, req.URL.Host, fname)
```

szuecs:
:+1:

proxy/proxy.go
```diff
@@ -882,7 +884,31 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
...
 	p.metrics.IncCounter("outgoing." + req.Proto)
 	ctx.proxySpan.LogKV("http_roundtrip", StartEvent)
-	response, err := p.roundTripper.RoundTrip(req)
+
+	var response *http.Response
+	switch req.URL.Scheme {
+	case "fastcgi":
+		f := "index.php"
+		if sf, ok := ctx.StateBag()["fastCgiFilename"]; ok {
+			f = sf.(string)
```
aryszka: it would be safer to check whether it is a string, because this way a buggy filter can set something else in the state bag.

aryszka: discussed: failing fast via crashing is also fine in this case

aryszka:
:+1:

Issue #1233, opened 2019-10-23
Skipper Helm Chart


kolbis:
Is there anyone which already created a production ready helm chart for skipper?
If so, can you please share it?

szuecs:
We don’t like helm and probably all of them in the internet are outdated.

https://github.com/zalando-incubator/kubernetes-on-aws/tree/dev/cluster/manifests/skipper has all manifests that you need to build a production setup. It’s used in >140 clusters.

kolbis:
I appreciate the answer.
Just out of curiosity, why dont you like helm?

szuecs:
Tiller is known to be unsafe.
I don’t like, if I have a templated mess that I can’t easily understand.
Plain manifests with a little template to replace numbers of CPUs or memory is much easier.
I understand that helm is good for, just trying something, but you could also provide a directory that has some yaml files which you can deploy with `kubectl create -f dir`. Maybe something that can help. What do you think?

kolbis:
I understand your point. That said, I think it has many benefits which overcome the downfalls such as "hard to read templating engine".
For us, microservice may have dependency in redis, elastic, mssql and etc.
Coordinating such a deployment is easier with helm as you are reusing charts. 

I think supporting helm for skipper would be a nice add-on.

szuecs:
@kolbis I wrote documentation. Would you mind checking if this helps you in order to get faster started?

szuecs:
It's now available at https://opensource.zalando.com/skipper/kubernetes/ingress-controller/#install-skipper-as-ingress-controller

kolbis:
Thanks!
I will have a look.
If we will decide to create a chart, we will open source it.
I Will update

szuecs:
If you like create a doc PR, maybe someone find it useful, too

szuecs:
I had a discussion about helm v3 an @shaunzie (?) convinced me that it makes sense to create an official helm chart to simplify testing the ingress stack.
We can have some knobs to change to make sure you can easily test a simple setup and one with all batteries included.

PR #1326, opened 2020-03-01
add helm chart

szuecs:
fix https://github.com/zalando/skipper/issues/1233 and add a helm chart which can optionally enable clusterRatelimits, eastwest, and use kube2iam or kube-aws-iam-controller to get AWS credentials for external-dns and kube-ingress-aws-controller

### DIFF
--- a/packaging/helm/skipper-aws/README.md
+++ b/packaging/helm/skipper-aws/README.md
@@ -1,0 +1,126 @@
+# Helm chart
+
+## Pre-requisites
+
+### IAM Policies
+
+If using [Kube2IAM](https://github.com/jtblin/kube2iam) make sure you read their documentation on IAM roles (specifically about trust relationships) on how to create the roles needed for these policies.
+
+If using IAM Roles for Service Accounts please take a look at the [AWS Documentation](https://docs.aws.amazon.com/eks/latest/userguide/create-service-account-iam-policy-and-role.html) on how to create the roles needed for these policies.
+
+**external-dns**
+
+```json
+{
+  "Version": "2012-10-17",
+  "Statement": [
+      {
+          "Sid": "",
+          "Effect": "Allow",
+          "Action": "route53:ChangeResourceRecordSets",
+          "Resource": "arn:aws:route53:::hostedzone/*"
+      },
+      {
+          "Sid": "",
+          "Effect": "Allow",
+          "Action": [
+              "route53:ListResourceRecordSets",
+              "route53:ListHostedZones"
+          ],
+          "Resource": "*"
+      }
+  ]
+}
+```
+
+**kube-ingress-aws-controller**
+
+```json
+{
+  "Version": "2012-10-17",
+  "Statement": [
+      {
+          "Sid": "",
+          "Effect": "Allow",
+          "Action": [
+              "iam:ListServerCertificates",
+              "iam:GetServerCertificate",
+              "iam:CreateServiceLinkedRole",
+              "elasticloadbalancingv2:*",
+              "elasticloadbalancing:*",
+              "ec2:DescribeVpcs",
+              "ec2:DescribeSubnets",
+              "ec2:DescribeSecurityGroups",
+              "ec2:DescribeRouteTables",
+              "ec2:DescribeInternetGateways",
+              "ec2:DescribeInstances",
+              "ec2:DescribeAccountAttributes",
+              "cloudformation:*",
+              "autoscaling:DetachLoadBalancers",
+              "autoscaling:DetachLoadBalancerTargetGroups",
+              "autoscaling:DescribeLoadBalancerTargetGroups",
+              "autoscaling:DescribeAutoScalingGroups",
+              "autoscaling:AttachLoadBalancers",
+              "autoscaling:AttachLoadBalancerTargetGroups",
+              "acm:ListCertificates",
+              "acm:GetCertificate",
+              "acm:DescribeCertificate"
+          ],
+          "Resource": "*"
+      }
+  ]
+}
+```
+
+## Installation
+
+```sh
+git clone https://github.com/zalando/skipper.git
+cd skipper/packaging/helm/skipper-aws
+helm install --name skipper-aws .
+```
+
+## Values
+
+| Value | Description | Default |
+| --- | --- | --- |
+| clusterID | The name of your Kubernetes cluster | mycluster |
+| region | The region in which your Kubernetes cluster resides | eu-central-1 |
+| vpa | Enable vertical pod autoscaling | false |
+| kube2iam.install | Install kube2iam | true |
+| kube2iam.enable | Make use of kube2iam | true |
+| kube2iam.aws_role | The kube2iam role that can assume other roles | kube2iam-role |
+| kube2iam.version | The version of kube2iam to install | 0.10.7 |
+| kube2iam.image | The image of kube2iam to install | registry.opensource.zalan.do/teapot/kube2iam |
+| kube_aws_iam_controller.enable | Enable use of kube-aws-iam-controller | false |
+| kube_aws_iam_controller.install | Enable installation of kube-aws-iam-controller | false |
+| kube_aws_iam_controller.aws_role | The role used by kube-aws-iam-controller | kube-aws-iam-controller-role |
+| kube_aws_iam_controller.version | The version of kube_aws_iam_controller  to install | 0.10.7 |
+| kube_aws_iam_controller.image | The image of kube_aws_iam_controller to install | registry.opensource.zalan.do/teapot/kube2iam |
+| external_dns.ownership_prefix | The txt prefix to use for DNS entries | skipper-test |
+| external_dns.aws_role | The role created for the external-dns IAM policy | external-dns-role |
+| external_dns.version | The version of external-dns to use | 0.5.18 |
+| external_dns.image | The image of external-dns to use | registry.opensource.zalan.do/teapot/external-dns |
+| kube_ingress_aws_controller.idle_timeout | The idle connection timeout for kube-ingress-aws-controller | 60s |
+| kube_ingress_aws_controller.aws_role | The role used by kube-ingress-aws-controller | kube-ingress-aws-controller-role |
+| kube_ingress_aws_controller.version | The version of kube-ingress-aws-controller to install | 0.10.1 |
+| kube_ingress_aws_controller.image | The image of kube-ingress-aws-controller to install | registry.opensource.zalan.do/teapot/kube-ingress-aws-controller |
+| kube_ingress_aws_controller.ssl_policy | The SSL policy to use | ELBSecurityPolicy-TLS-1-2-2017-01 |
+| skipper.version | The version of Skipper to install | 0.11.48 |
+| skipper.image | The image of Skipper to install | registry.opensource.zalan.do/pathfinder/skipper |
+| skipper.cluster_ratelimit | Enable rate limiting in Skipper | false |
+| skipper.redis.version | The version of Redis to install | 4.0.9-master-6 |
+| skipper.redis.image | The image of Redis to install | registry.opensource.zalan.do/zmon/redis |
+| skipper.svc_ip |  | 10.3.11.28 |
+| skipper.east_west | Enable East-West support in Skipper | false |
+| skipper.east_west_domain | The East-West domain to use | .ingress.cluster.local |
+
+See [values.yaml](values.yaml) for more.
+
+
+## Documentation
+
+Before running production you should consider reading all important
+[tutorials](https://opensource.zalando.com/skipper/tutorials/basics/)
+and check the [operations
+reference](https://opensource.zalando.com/skipper/operation/operation/).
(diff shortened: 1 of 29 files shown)


packaging/helm/skipper-aws/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+See values.yaml to change installation options.
+
+If you want to setup skipper, kube-ingress-aws-controller,
+external-dns and kube2iam via helm in your test environment:
+
+```
+git clone https://github.com/zalando/skipper.git
+cd skipper/packaging/helm
+kubectl create -f skipper-aws/manifests/
```
MarcusNoble: What's the reasoning for having these outside of Helm?

packaging/helm/skipper-aws/templates/external-dns/aws-iam-role.yaml
```diff
@@ -0,0 +1,9 @@
+{{ if .Values.kube_aws_iam_controller }}
+apiVersion: zalando.org/v1
+kind: AWSIAMRole
+metadata:
+  name: external-dns-aws-iam-credentials
+  namespace: kube-system
```
MarcusNoble: This will allow users to install into another namespace if they choose.
```suggestion
  namespace: {{ .Release.Namespace }}
```

packaging/helm/skipper-aws/templates/external-dns/aws-iam-role.yaml
```diff
@@ -0,0 +1,9 @@
+{{ if .Values.kube_aws_iam_controller }}
+apiVersion: zalando.org/v1
+kind: AWSIAMRole
+metadata:
+  name: external-dns-aws-iam-credentials
```
MarcusNoble: Each Helm install gets a release name (either autogenarated or provided by the user) that allows multiple instances of a single chart to be installed. It's good practice to use this name in the name of resources.
```suggestion
  name: {{ .Release.Name }}-external-dns-aws-iam-credentials
```

Docs: https://helm.sh/docs/chart_template_guide/builtin_objects/

packaging/helm/skipper-aws/templates/external-dns/deployment.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: external-dns
+  namespace: kube-system
+  labels:
```
MarcusNoble: It might also be useful to include the chart version as a label (`{{ .Chart.Version }}`)

packaging/helm/skipper-aws/templates/external-dns/deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+    spec:
+      dnsConfig:
+        options:
+          - name: ndots
+            value: "1"
+      priorityClassName: system-cluster-critical
+      serviceAccountName: external-dns
+      containers:
+      - name: external-dns
+        image: registry.opensource.zalan.do/teapot/external-dns:{{ .Values.externalDnsVersion }}
```
MarcusNoble: I recommend moving the image to a values property also as I know people (us included) pull all images via a local proxy (e.g. Artifactory) so would want to be able to override this.

packaging/helm/skipper-aws/templates/external-dns/deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+        - --aws-batch-change-size=100
+        resources:
+          limits:
+            cpu: 50m
+            memory: 100Mi
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 7979
+{{- if .Values.kube_aws_iam_controller }}
```
MarcusNoble: It would be good to also have the choice of using IAM for service accounts on EKS (https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html).

Happy to help with this as I know you don't use EKS. (It's just an annotation on the Service similar to the kube2iam annotation)

packaging/helm/skipper-aws/templates/kube-aws-iam-controller/crd.yaml
```diff
@@ -0,0 +1,64 @@
+{{ if .Values.install_kube_aws_iam_controller }}
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
MarcusNoble: When using CRDs with Helm make sure you're aware of the limitations: https://helm.sh/docs/topics/chart_best_practices/custom_resource_definitions/

Basically, updates / deleted of CRDs isn't supported and to ensure CRDs are installed first (before any resources that rely on them) they either need to go in a directory called `crds` or manually applied to the cluster prior to the helm install.

packaging/helm/skipper-aws/templates/kube-aws-iam-controller/deployment.yaml
```diff
@@ -0,0 +1,45 @@
+{{ if .Values.install_kube_aws_iam_controller }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kube-aws-iam-controller
+  namespace: kube-system
+  labels:
+    application: kube-aws-iam-controller
+    version: v0.1.0
```
MarcusNoble: Would be good to pull this out into the values file.

packaging/helm/skipper-aws/templates/kube-aws-iam-controller/vpa.yaml
```diff
@@ -0,0 +1,21 @@
+{{ if .Values.install_kube_aws_iam_controller }}
```
MarcusNoble:
```suggestion
{{ if and (.Values.install_kube_aws_iam_controller) (.Values.vpa) }}
```

packaging/helm/skipper-aws/templates/kube2iam/daemonset.yaml
```diff
@@ -0,0 +1,67 @@
...
+    metadata:
+      labels:
+        application: kube2iam
+        version: {{ .Values.kube2iamVersion }}
+    spec:
+      dnsConfig:
+        options:
+          - name: ndots
+            value: "1"
+      priorityClassName: system-node-critical
```
MarcusNoble: If you change the namespace to be configurable you'll need some check here to see if it's set to `kube-system` or not.

packaging/helm/skipper-aws/values.yaml
```diff
@@ -0,0 +1,33 @@
+clusterID: mycluster
+
+# AWS IAM integration kube2iam or kube-aws-iam-controller
+install_kube2iam: true
```
MarcusNoble: This is my personal preference but I tend to like nested yaml for this sort of thing, e.g.

```yaml
kube2iam:
  enabled: true
  role: kube2iam-role
  etc...
```

The Helm best practice docs don't suggest one way over another so don't feel like you need to change this.

packaging/helm/skipper-aws/templates/skipper/service-internal.yaml
```diff
@@ -0,0 +1,16 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: skipper-internal
+  namespace: kube-system
+  labels:
+    application: skipper-ingress
+spec:
+  type: ClusterIP
+  clusterIP: {{ .Values.skipperSvcIP }}
```
MarcusNoble: Do we need to specify this? Wouldn't it be better to let Kubernetes auto-assign it? (Or make it optional)

MarcusNoble:
I do wonder if this chart might be including too much (making it more work to maintain). Do we need to include kube2iam or kube-aws-iam-controller in this?

packaging/helm/skipper-aws/README.md
szuecs: Reasoning was, that these don't contain template code

packaging/helm/skipper-aws/templates/external-dns/deployment.yaml
szuecs: makes sense, will change it

packaging/helm/skipper-aws/templates/external-dns/deployment.yaml
szuecs: makes sense, will change it

packaging/helm/skipper-aws/templates/kube-aws-iam-controller/crd.yaml
szuecs: will move them to crds/

packaging/helm/skipper-aws/templates/kube-aws-iam-controller/deployment.yaml
szuecs: true all image versions should be there

packaging/helm/skipper-aws/values.yaml
szuecs: makes sense, will change it

packaging/helm/skipper-aws/templates/skipper/service-internal.yaml
szuecs: No, because you need to add this to the CoreDNS template, too.
If you have 100 clusters and the IP is dynamically created, you have to change the Corefile for all of your clusters.

packaging/helm/skipper-aws/README.md
MarcusNoble: It also makes it easier to upgrade and uninstall everything. (e.g. `helm uninstall skipper` will remove everything that was installed with helm)

szuecs: good point, fixed

packaging/helm/skipper-aws/templates/external-dns/deployment.yaml
szuecs: changed

packaging/helm/skipper-aws/templates/kube-aws-iam-controller/deployment.yaml
szuecs: fixed

packaging/helm/skipper-aws/values.yaml
szuecs: fixed

packaging/helm/skipper-aws/templates/external-dns/deployment.yaml
szuecs: It would be nice to get help here, because I have no experience with that.

MarcusNoble: Sure. It's very similar to kube2iam (from the Kubernetes side) but rather than annotating the pod you annotate the associated service account.
e.g.
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: example
  annotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::AWS_ACCOUNT_ID:role/IAM_ROLE_NAME
```

The role then needs a spcific trust relationship set up but I don't think that needs documenting in this as if they're using IAM for service accounts then they'd have already done it for any other roles.

packaging/helm/skipper-aws/templates/external-dns/deployment.yaml
szuecs: fixed

packaging/helm/skipper-aws/templates/kube2iam/daemonset.yaml
szuecs: We could pass an if `namespace == kube-system` apply this line

MarcusNoble: yeah that's pretty much what I'm thinking.

szuecs: fixed

szuecs:
Forced push because of DCO

packaging/helm/skipper-aws/templates/external-dns/deployment.yaml
szuecs: I added this, too

packaging/helm/skipper-aws/templates/external-dns/rbac.yaml
```diff
@@ -7,6 +7,11 @@ metadata:
   labels:
     application: external-dns
     chart: {{ .Chart.Version }}
+{{- if .Values.eks_iam }}
+  annotations:
+    eks.amazonaws.com/role-arn:
+      arn:aws:iam::AWS_ACCOUNT_ID:role/{{ .Values.external_dns.aws_role }}
```
MarcusNoble:
```suggestion
      arn:aws:iam::{{ .Values.awsAccountID }}:role/{{ .Values.external_dns.aws_role }}
```

szuecs: ah shit, will do a bulk change

szuecs: fixed

MarcusNoble: 👍 😄

szuecs:
:+1:

aryszka:
:+1:

aryszka:
thx!

PR #1331, opened 2020-03-04
Explain when using predicate as filter

ruudk:
For beginners, it's very easy to mix predicates and filters. Also when using Kubernetes annotations, sometimes a mistake is easy to make.

Instead of returning `failed to process route (kube_x): filter not found: 'PathRegexp'` it will now return `failed to process route (kube_x): trying to use 'PathRegexp' as filter, but it is only available as predicate`


szuecs:
:+1:

aryszka:
:+1:

PR #1335, opened 2020-03-04
Support hash comments

ruudk:
In Go it's only possible to comment with `//` but in the context of Kubernetes (YAML) it's pretty common to use hashes `#`. So let's support both.

### DIFF
--- a/eskip/lexer.go
+++ b/eskip/lexer.go
@@ -215,6 +215,12 @@ func scanRegexpOrComment(code string) (t token, rest string, err error) {
 		return
 	}
 
+	if code[0] == '#' {
+		rest = scanComment(code)
+		err = void
+		return
+	}
+
 	if code[1] == '/' {
 		rest = scanComment(code)
 		err = void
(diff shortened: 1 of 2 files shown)


eskip/lexer.go
```diff
@@ -291,6 +297,7 @@ func selectVaryingScanner(code string) scanner {
 	var sf scannerFunc
 	switch code[0] {
 	case '/':
+	case '#':
```
aryszka: this would be really `case '/', '#':` in go, while this way nothing happens in `case '/':`

ruudk: Oh wow, I guess I'm too used to PHP where it will fallthrough.

aryszka:
I think we rather should apply 'less is more' in this case, and for eskip the available commenting syntax is enough.

When it comes to yaml, we actually have eskip embedded in yaml, because yaml allows it, and I think it is even better to distinguish whether something is a yaml or an eskip comment.

So thanks for the suggestion, but if you don't mind, let's close this PR.

ruudk:
@aryszka Fair enough :)

PR #1336, opened 2020-03-05
Feature/stream editor

aryszka:
create a familiy of stream editing filters

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -459,6 +459,56 @@ incoming requests.
 
 The same as [tee filter](#tee), but does not follow redirects from the backend.
 
+## sed
+
+This filter () expects a regexp pattern and a replacement string as arguments.
+During the streaming of the response body, every occurence of the pattern will
+be replaced with the replacement string. The editing doesn't happen right when
+the filter is executed, only later when the streaming normally happens, after
+all response filters were called. The sed() filter accepts an optional third
+argument, the max editor buffer size in bytes. This argument limits how much
+data can be buffered at a given time by the editor. The default value is 2MiB.
+See more details below.
+
+The filter uses the go regular expression implementation:
+https://github.com/google/re2/wiki/Syntax
+
+### Memory handling and limitations
+
+In order to avoid unbound buffering of unprocessed data, the sed* filters need to
+apply some limitations. Some patterns, e.g. `.*` would allow to match the complete
+payload, and it could result in trying to buffer it all and potentially causing
+running out of available memory. Similarly, in case of certain expressions, when
+they don't match, it's impossible to tell if they would match without reading more
+data from the source, and so would potentially need to buffer the entire payload.
+
+To prevent too high memory usage, the max buffer size is limited in case of each
+variant of the filter, by default to 2MiB, which is the same limit as the one we
+apply when reading the request headers by default. When the limit is reached, and
+the buffered content matches the pattern, then it is processed by replacing it,
+when it doesn't match the pattern, then it is forwarded unchanged. This way, e.g.
+`sed(".*", "")` can be used safely to consume and discard the payload.
+
+As a result of this, with large payloads, it is possible that the resulting content
+will be different than if we had run the replacement on the entire content at once.
+If we have enough preliminary knowledge about the payload, then it may be better to
+use the delimited variant of the filters, e.g. for line based editing.
+
+## sedDelim
+
+Like sed(), but it expects an additional argument, before the optional max buffer
+size argument, that is used to delimit chunks to be processed at once. The pattern
+replacement is executed only within the boundaries of the chunks defined by the
+delimiter, and matches across the chunk boundaries are not considered.
+
+## sedRequest
+
+Like sed(), but for the request content.
+
+## sedRequestDelim
+
+Like sedDelim(), but for the request content.
+
 ## basicAuth
 
 Enable Basic Authentication
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -522,7 +522,7 @@ and the rest of the payload is dropped.
 
 ## sedDelim
 
-Like sed(), but it expects an additional argument, before the optional max buffer
+Like [sed()](#sed), but it expects an additional argument, before the optional max buffer
 size argument, that is used to delimit chunks to be processed at once. The pattern
 replacement is executed only within the boundaries of the chunks defined by the
 delimiter, and matches across the chunk boundaries are not considered.
@@ -535,7 +535,7 @@ editorRoute: * -> sedDelim("foo", "bar", "\n") -> "https://www.example.org";
 
 ## sedRequest
 
-Like sed(), but for the request content.
+Like [sed()](#sed), but for the request content.
 
 Example:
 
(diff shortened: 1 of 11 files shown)


filters/sed/editor.go
```diff
@@ -0,0 +1,279 @@
...
+// - we can document that it is currently not supported, together
+//   with the templating
+
+const (
+	readBufferSize             = 8 << 10
+	defaultMaxEditorBufferSize = 2 << 20
+)
+
+// editor provides a reader that wraps a source reader, and replaces each occurence of
+// the provided search pattern with the provided replacment. It can be used with a
```
szuecs: replacment -> replacement

docs/reference/filters.md
```diff
@@ -459,6 +459,56 @@ incoming requests.
 
 The same as [tee filter](#tee), but does not follow redirects from the backend.
 
+## sed
+
+This filter () expects a regexp pattern and a replacement string as arguments.
```
szuecs: () ?

filters/sed/editor.go
```diff
@@ -0,0 +1,474 @@
...
+)
+
+type maxBufferHandling int
+
+const (
+	maxBufferBestEffort maxBufferHandling = iota
+	maxBufferAbort
+)
+
+// editor provides a reader that wraps a input reader, and replaces each occurence of
```
szuecs: a input -> an input

filters/sed/editor.go
```diff
@@ -0,0 +1,474 @@
...
+)
+
+// TODO:
+// - we need to document that only non-zero matches are replaced
+// - we can document that it is currently not supported, together
+//   with the templating
+
+const (
+	readBufferSize             = 8 << 10
+	defaultMaxEditorBufferSize = 2 << 20
```
szuecs: I am not so into bit shifting, if it's not necessary for performance.
I like much more:
readBufferSize = 8192
defaultMaxEditorBufferSize = 256 * readBufferSize

I guess that defaultMaxEditorBufferSize and readBufferSize are dependent on each other makes it also more clear.
If you really like `8 << 10`, then please use `defaultMaxEditorBufferSize = 256 * readBufferSize`.

filters/sed/editor.go
```diff
@@ -0,0 +1,474 @@
...
+	mbh maxBufferHandling,
+) *editor {
+	if maxBufferSize <= 0 {
+		maxBufferSize = defaultMaxEditorBufferSize
+	}
+
+	rsize := readBufferSize
+	if maxBufferSize < rsize {
+		rsize = maxBufferSize
+	}
```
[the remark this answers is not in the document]
aryszka: Not sure. I don't really see the reason for that. `rsize` normally will be way smaller than `maxBufferSize`. They are different concerns. `rsize` controls how big chunks we try to read from the network at a time, and from experience 8k is a good value. `maxBufferSize` controls how much we allow to reside in the memory at a given time. The above safety check is only there because it never makes sense to read more than `maxBufferSize`. Does this make sense to you like this?

filters/sed/editor.go
```diff
@@ -0,0 +1,474 @@
...
+}
+
+func (e *editor) finalizeEdit(pendingMatches bool) {
+	if pendingMatches {
+		e.ready.Write(e.replacement)
+		return
+	}
+
+	if len(e.delimiter) == 0 {
+		io.CopyBuffer(e.ready, e.pending, e.readBuffer)
```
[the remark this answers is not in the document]
aryszka: thx! :smile:

filters/sed/sed.go
```diff
@@ -0,0 +1,272 @@
...
+		if len(args) >= 4 {
+			maxBuf = args[3]
+		}
+
+		if len(args) == 5 {
+			maxBufHandling = args[4]
+		}
+
+		// Temporary solution, see eskip tokenizer bug: ...
+		delimiterString = unescape(delimiterString)
```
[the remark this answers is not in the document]
aryszka: used as a placeholder, thanks for catching this. Eventually, it was fixed here: https://github.com/zalando/skipper/pull/1341 , so we don't need it.

filters/sed/editor.go
aryszka: i picked up this notation in the Go standard library, where I seen it used as:

`<< 10` means Ki
`<< 20` means Mi
`<< 30` means Gi

So the meaning of these values should be 8KiB and 2MiB. I'll replace them though if it is really not intuitive.

aryszka: Calculating one from the other doesn't make too much sense in my view, because these are not related values.

szuecs: It makes sense, because you read x times readBufferSize and it would be weird to be able to read only 3.43 times readBufferSize as max.
This is what I am trying to argue for. :)

filters/sed/editor.go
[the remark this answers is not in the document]
szuecs: fine for me, was just a suggestion.

docs/reference/filters.md
```diff
@@ -459,6 +459,91 @@ incoming requests.
 
 The same as [tee filter](#tee), but does not follow redirects from the backend.
 
+## sed
+
```
szuecs: I miss an introduction to the filter

docs/reference/filters.md
```diff
@@ -459,6 +459,91 @@ incoming requests.
...
+If the max buffer handling is set to "abort", then the stream editing is stopped
+and the rest of the payload is dropped.
+
+## sedDelim
+
+```
+editorRoute: * -> sedDelim("foo", "bar", "\n") -> "https://www.example.org";
+```
+
+Like sed(), but it expects an additional argument, before the optional max buffer
```
szuecs: sed() -> [sed()](#sed)

docs/reference/filters.md
```diff
@@ -459,6 +459,91 @@ incoming requests.
...
+replacement is executed only within the boundaries of the chunks defined by the
+delimiter, and matches across the chunk boundaries are not considered.
+
+## sedRequest
+
+```
+editorRoute: * -> sedRequest("foo", "bar") -> "https://www.example.org";
+```
+
+Like sed(), but for the request content.
```
szuecs: see https://github.com/zalando/skipper/pull/1336/files#r392334483

docs/reference/filters.md
```diff
@@ -459,6 +459,91 @@ incoming requests.
...
+
+## sedRequestDelim
+
+```
+editorRoute: * -> sedRequestDelim("foo", "bar", "\n") -> "https://www.example.org";
+```
+
+Like sedDelim(), but for the request content.
```
szuecs: similar to https://github.com/zalando/skipper/pull/1336/files#r392334483

szuecs:
:+1:

aryszka:
I restructured the filters doc a bit and added an introduction to the main `sed` entry. @szuecs WDYT?

szuecs:
@aryszka I like it!
What about the comments like https://github.com/zalando/skipper/pull/1336/commits/e9a7da85a5e85093905789d7b6f8c604027286e7#r392334483?
You can link the sed() filter in the markdown docs!

aryszka:
set links

szuecs:
:+1:

aryszka:
:+1:

PR #1338, opened 2020-03-05
Introduce Weight predicate

ruudk:
I noticed this odd looking double Regexp in my Kubernetes Skipper instance:
```eskip
route: 
    Host(/^www[.]xxx[.]no$/) && 
    PathRegexp(/.*/) && 
    PathRegexp(/.*/) && 
    Header("X-Forwarded-Proto", "http")
    -> <roundRobin, "http://xxx:80">;
```

Turns out it's by design:
https://github.com/zalando/skipper/blob/ae2a3bfe499c888154a630bbab008de4a84b11c3/dataclients/kubernetes/redirect.go#L90-L99

Let's fix this.

## Todo

- [x] Add more tests
- [x] Add documentation

### DIFF
--- a/dataclients/kubernetes/redirect.go
+++ b/dataclients/kubernetes/redirect.go
@@ -7,6 +7,7 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/routing"
 )
 
 const (
@@ -87,16 +88,11 @@ func initRedirectRoute(r *eskip.Route, code int) {
 	}
 	r.Headers[forwardedProtoHeader] = "http"
 
-	// the below duplicate any-path (.*) is set to make sure that
-	// the redirect route has a higher priority during matching than
-	// the normal routes that may have max 2 predicates: path regexp
-	// and host.
-	//
-	// A better solution might be to implement a Weight() predicate
-	// and apply it here.
-	//
-	r.PathRegexps = append(r.PathRegexps, ".*")
-	r.PathRegexps = append(r.PathRegexps, ".*")
+	// Give this route a higher weight so that it will get precedence over existing routes
+	r.Predicates = append([]*eskip.Predicate{{
+		Name: routing.WeightPredicateName,
+		Args: []interface{}{float64(1000)},
+	}}, r.Predicates...)
 
 	r.Filters = append(r.Filters, &eskip.Filter{
 		Name: "redirectTo",
@@ -113,16 +109,11 @@ func initDisableRedirectRoute(r *eskip.Route) {
 	}
 	r.Headers[forwardedProtoHeader] = "http"
 
-	// the below duplicate any-path (.*) is set to make sure that
-	// the redirect route has a higher priority during matching than
-	// the normal routes that may have max 2 predicates: path regexp
-	// and host.
-	//
-	// A better solution might be to implement a Weight() predicate
-	// and apply it here.
-	//
-	r.PathRegexps = append(r.PathRegexps, ".*")
-	r.PathRegexps = append(r.PathRegexps, ".*")
+	// Give this route a higher weight so that it will get precedence over existing routes
+	r.Predicates = append([]*eskip.Predicate{{
+		Name: routing.WeightPredicateName,
+		Args: []interface{}{float64(1000)},
+	}}, r.Predicates...)
 }
 
 func globalRedirectRoute(code int) *eskip.Route {
--- a/docs/reference/predicates.md
+++ b/docs/reference/predicates.md
@@ -128,6 +128,30 @@ Host(/^my-host-header\.example\.org$/)
 Host(/header\.example\.org$/)
 ```
 
+## Weight (priority)
+
+By default, the weight (priority) of a route is determined by the number of defined predicates.
+
+If you want to give a route more priority, you can give it more weight.
+
+Parameters:
+
+* Weight (int)
+
+Example where `route2` has more priority because it has more predicates:
+
+```
+route1: Path("/test") -> "http://www.zalando.de";
+route2: Path("/test") && True() -> "http://www.zalando.de";
+```
+
+Example where `route1` has more priority because it has more weight:
+
+```
+route1: Path("/test") && Weight(100) -> "http://www.zalando.de";
+route2: Path("/test") && True() && True() -> "http://www.zalando.de";
+```
+
 ## Method
 
 The HTTP method that the request must match. HTTP methods are one of
--- a/routing/matcher.go
+++ b/routing/matcher.go
@@ -33,6 +33,7 @@ type leafMatcher struct {
 	hasFreeWildcardParam bool
 	exactPath            string
 	method               string
+	weight               int
 	hostRxs              []*regexp.Regexp
 	pathRxs              []*regexp.Regexp
 	headersExact         map[string]string
@@ -44,7 +45,7 @@ type leafMatcher struct {
 type leafMatchers []*leafMatcher
 
 func leafWeight(l *leafMatcher) int {
-	w := 0
+	w := l.weight
 
 	if l.method != "" {
 		w++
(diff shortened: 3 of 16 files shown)


eskip/eskip.go
```diff
@@ -133,6 +133,10 @@ type Route struct {
 	// E.g. PathRegexp(/\/api\//)
 	PathRegexps []string
 
+	// Base weight of the route.
+	// E.g. Weight(50)
+	Weight int
```
aryszka: We'd like to make these explicit predicate fields deprecated, and use only the Predicate slice field for new predicates. Please remove this field and create a predicate spec like the ones here: https://godoc.org/github.com/zalando/skipper/predicates

ruudk: I cannot get this to work. See my last commit, especially this: https://github.com/zalando/skipper/pull/1338/commits/c79e86d3ac8aef7a1a438f53685b1ccbf7cec75a#diff-ae803bdbe486828b2ecb3c8a0b06a9d9R139-R147

Any advice on how to structure this?

routing/matcher.go
```diff
@@ -135,6 +136,15 @@ func canonicalizeHeaderRegexps(hrx map[string][]*regexp.Regexp) map[string][]*re
 	return chrx
 }
 
+func extractWeight(r *Route) int {
+	//for _, p := range r.Predicates {
+	//	if w, ok := p.(weight.Predicate); ok {
+	// 		THIS IS IMPOSSIBLE >> import cycle not allowed
```
aryszka: indeed. Let me think of an approach...

aryszka: what i am thinking about is that this not so much the functionality of the predicate, but the routing package itself. We can see that the Match() method is always true. So we can either:

1. move the predicate to the routing package
2. use only the eskip name to identify the predicate while processing the route configuration and parse the arguments in the routing package

no better idea right now. @szuecs , what do you think?

eskip/eskip.go
ruudk: Btw, should all these predicates be extracted to separate predicates? Seem like a nice refactoring that I can work on if you give me the OK.

routing/matcher.go
aryszka: thought of it a bit, and I like now 2. better.

szuecs: I am with you, do 2.

szuecs:
Beware that in general you don't need the PathRegexp for this, you can also use `True()`

routing/matcher.go
aryszka: for example, it could look like this:

https://github.com/ruudk/skipper/pull/1

ruudk:
Applied your code changes and made everything working again. What do you think? :)

docs/reference/predicates.md
```diff
@@ -128,6 +128,30 @@ Host(/^my-host-header\.example\.org$/)
...
+```
+route1: Path("/test") -> "http://www.zalando.de";
+route2: Path("/test") && PathRegexp(".*") -> "http://www.zalando.de";
+```
+
+Example where `route1` has more priority because it has more weight:
+
+```
+route1: Path("/test") && Weight(100) -> "http://www.zalando.de";
+route2: Path("/test") && PathRegexp(".*") && PathRegexp(".*") -> "http://www.zalando.de";
```
szuecs: Please replace PathRegexp() with True().
Can you also show an example where Weight() does not apply, for example a better Path/PathSubtree match?

dataclients/kubernetes/redirect.go
```diff
@@ -87,16 +88,11 @@ func initRedirectRoute(r *eskip.Route, code int) {
...
-	//
-	// A better solution might be to implement a Weight() predicate
-	// and apply it here.
-	//
-	r.PathRegexps = append(r.PathRegexps, ".*")
-	r.PathRegexps = append(r.PathRegexps, ".*")
+	// Give this route a higher weight so that it will get precedence over existing routes
+	r.Predicates = append([]*eskip.Predicate{{
+		Name: routing.WeightPredicateName,
+		Args: []interface{}{float64(1000)},
```
aryszka: float64 appears in many places, because some time ago we decided that we parse numbers always as float. While this should be revisited at some point, there are already filters and predicates that also accept numbers as ints, and if we define in-memory eskip objects, and the filter or predicate accepts ints, then it's unnecessary to pass in floats. The weight predicates accepts ints, so let's just use an int here, too.

dataclients/kubernetes/redirect.go
```diff
@@ -113,16 +109,11 @@ func initDisableRedirectRoute(r *eskip.Route) {
...
```
aryszka: see previous comment, no need to use float64

aryszka:
thx! Generally, lgtm. I have only those minor comments.

xavivars:
It would be great if this change allowed the custom predicates to also have a custom weight.

aryszka:
@xavivars I see how the weights for custom predicates can be useful, e.g. based on one of the examples here: https://docs.google.com/document/d/1fBrCTcvDGFVmCnmcwTevOfCa9j86TewAnn-D1U-Xtas/edit?ts=5e676b6f# . But at this stage I vote for merging this PR without custom predicates having a weight, and discuss this suggestion separately. Maybe a followup feature.

xavivars:
I'm happy with your approach, as long as we keep the conversation going.

@szuecs's proposal of evaluating always `Host` before any other predicate makes a lot of sense to me, and would make sure in cases like the one below, the right route gets selected.

Routes file:
```
Host('example.org') && PathPrefix('backend/') -> http://backend.example.org
Host('example.org') -> status(404) -> http://404page.example.org

Path(“/foobar”) -> http://my-second-backend;
```
With this setup, I'd would expect a request going to 'example.org/foobar' to show a 404. Instead, it'll show the content of http://my-second-backend/foobar.

Does it make sense?

dataclients/kubernetes/redirect.go
ruudk: If I don't add the `float64` it will become an `int`, and then a lot of tests start to fail. Example:
```
    TestEnableHTTPSRedirectFromIngress: httpsredirect_test.go:372:   []*eskip.Route{
          	&{Id: "kube___catchall__api_example_org____", HostRegexps: []string{"^api[.]example[.]org$"}, BackendType: s"shunt"},
          	&{Id: "kube___catchall__www_example_org____", HostRegexps: []string{"^www[.]example[.]org$"}, BackendType: s"shunt"},
          	&{
          		... // 5 identical fields
          		Headers:       map[string]string{"X-Forwarded-Proto": "http"},
          		HeaderRegexps: nil,
          		Predicates: []*eskip.Predicate{
          			&{
          				Name: "Weight",
        - 				Args: []interface{}{int(1000)},
        + 				Args: []interface{}{float64(1000)},
          			},
          		},
```

How to fix?

docs/reference/predicates.md
ruudk: > Please replace PathRegexp() with True().

Done

> Can you also show an example where Weight() does not apply, for example a better Path/PathSubtree match?

Is this needed? I think this is clear enough?

aryszka: IMHO, it's fine like this. @szuecs is there a specific example on your mind?

dataclients/kubernetes/redirect.go
aryszka: my bad, didn't realize that in the tests we compare the routing output to routing parsed from text. :+1:

aryszka:
@xavivars Yes, it makes sense.

I think this PR is good enough for the Weight() predicate, and I'd like to steer the further conversation to this doc: https://docs.google.com/document/d/1fBrCTcvDGFVmCnmcwTevOfCa9j86TewAnn-D1U-Xtas/edit?ts=5e686b13#heading=h.874ji3wxbpuc

aryszka:
:+1:

eskip/eskip.go
aryszka: Absolutely. But please open the pull request against the release/v1 branch. Thanks in advance!

docs/reference/predicates.md
szuecs: I thought like this:

```
r1: PathSubtree("/foo") && Weight(100) -> status(200) -> "<shunt>"
r2: PathSubtree("/foo/bar") -> status(201) -> "<shunt>"
```
This request will match r2: `GET /foo/bar HTTP/1.1`

szuecs:
:+1:

xavivars:
Thanks for merging! This already allows us to solve all our use cases as long as we make sure host is in all routes.

ruudk:
Great to hear :)

PR #1339, opened 2020-03-05
InlineContent > Replace \n with newline

ruudk:
```eskip
test_robots:
        Path("/robots.txt")
        -> setResponseHeader("Content-Type", "text/plain")
        -> inlineContent("User-agent: *\nDisallow: /")
        -> <shunt>;
```

```bash
$ http get localhost:9090/robots.txt
HTTP/1.1 200 OK
Content-Length: 25
Content-Type: text/plain
Date: Thu, 05 Mar 2020 19:04:35 GMT
Server: Skipper

User-agent: *
Disallow: /
```

I think we only need this in `inlineContent` filter. If you think it's better, we can also try to get this support in [eskip lexer](https://github.com/zalando/skipper/blob/master/eskip/lexer.go#L127) so that it works everywhere.

### DIFF
--- a/filters/builtin/inlinecontent.go
+++ b/filters/builtin/inlinecontent.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/zalando/skipper/filters"
 )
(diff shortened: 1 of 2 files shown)


szuecs:
Why not use the setResponseHeader filter, instead?

ruudk:
How would that help? I want to return inline content with new lines. It only works when I insert real line breaks. But sometimes that looks weird and ugly so I want to just type `\n` for every line break and the filter should convert it to a real line break.

szuecs:
@ruudk inlineContent is about the body, not the header and I would not mix that.
https://github.com/zalando/skipper/pull/1336 might solve the problem of the problem to replace \n with \r\n. wdyt?

ruudk:
this is a robots.txt format. User-Agent is no header but part of the inline content.

ruudk:
See https://m.zalando.de/robots.txt as an example

szuecs:
Then the sed filter would solve your problem.

PR #1340, opened 2020-03-06
Refactor/origin marker

aryszka:
the origin markers are now set on an arbitrary route that doesn't necessarily gets updated/deleted when a partial update happens. Here using a dedicated route, that has always the same ID, and is only set when there is a change or a full reset.


szuecs:
:+1:

aryszka:
:+1:

Issue #562, opened 2018-03-02
eskip parsing of string literals is different than expected


NorthFury:
The only cases where the escape char is removed is when it precedes an escape char or a delimiter (`"`, `/` or `).
[Here](https://github.com/zalando/skipper/blob/master/eskip/lexer.go#L130) is the code location I'm referring to. 

Examples:
`"\."` gets parsed as `\.`

szuecs:
@NorthFury can you show what you tested as input and what was the result of your test and what you might want to see.
Probably you are right, but I can not understand you.

NorthFury:
Here is an example of a route that is currently valid:
`one: HostRegexp("^/one\.two$") -> <shunt>;`
Here is what would be a correct definition (still valid in the current implementation):
`one: HostRegexp("^/one\\.two$") -> <shunt>;`

What would be expected in a string literal is that the only characters following `\` would be `\` and `"`.  The Regexp literal has the same parsing model.

Fixing this would be a breaking change.

aryszka:
my understanding is:

`"\\"` in eskip is parsed to `"\\"` go internal, which is fine
`"\."` in eskip is parsed to `"\\."` go internal, which is wrong

i agree that this is a bug

aryszka:
another example where the parsing doesn't work as expected:

```
      Path("/robots.txt") ->
      setResponseHeader("Content-Type", "text/plain") ->
      inlineContent("User-agent: *\r\nAllow: /graphql\r\nAllow: /graphql/*\r\nDisallow: /\r\n") ->
      status(200) -> <shunt>
```

In the above route \r and \n sequences are used without unescaping.

PR #1341, opened 2020-03-06
Support newlines in lexer

ruudk:
This is a different approach to solve #1339.

## Problem

I have the following route:
```eskip
test:
    Path("/test")
    -> inlineContent("Line 1\nLine 2")
    -> <shunt>;
```

When I run this, Skipper will return my content wrong:
```
$ http get localhost:9090/test
HTTP/1.1 200 OK
Content-Length: 14
Date: Fri, 06 Mar 2020 14:52:13 GMT
Server: Skipper

Line 1\nLine 2

$ http get localhost:9090/test | hexdump
0000000 4c 69 6e 65 20 31 5c 6e 4c 69 6e 65 20 32
000000e

// 5c = \
// 6e = n
```

As you can see, the `\n` is not a real newline char (10). 

In the log of Skipper I see this:
```
[APP]INFO[0156] route settings, update, route: test: Path("/test") -> inlineContent("Line 1\\nLine 2") -> <shunt> 
```

It escaped the escape char.

The only way to get this working is by creating a route like this and insert a newline:

```eskip
test:
    Path("/test")
    -> inlineContent("Line 1
Line 2")
    -> <shunt>;

$ http get localhost:9090/test
HTTP/1.1 200 OK
Content-Length: 13
Date: Fri, 06 Mar 2020 14:53:49 GMT
Server: Skipper

Line 1
Line 2

$ http get localhost:9090/test | hexdump
0000000 4c 69 6e 65 20 31 0a 4c 69 6e 65 20 32
000000d

// 0a = \n
```

In the log I now see an actual newline:
```
[APP]INFO[0210] route settings, update, route: test: Path("/test") -> inlineContent("Line 1
Line 2") -> <shunt> 
```

## Solution

Correctly convert `\` + `n` to `char(10)` and `\` + `r` to `char(13)`.

Now, the log will always show it like this:
```
[APP]INFO[0000] route settings, reset, route: test: Path("/test") -> inlineContent("Line 1\nLine 2") -> <shunt> 
```

No more real newlines in the log!

### DIFF
--- a/eskip/lexer.go
+++ b/eskip/lexer.go
@@ -133,7 +133,24 @@ func scanEscaped(delimiter byte, code string) ([]byte, string) {
 		isEscapeChar := c == escapeChar
 
 		if escaped {
-			if !isDelimiter && !isEscapeChar {
+			switch c {
+			case 'a':
+				c = '\a'
+			case 'b':
+				c = '\b'
+			case 'f':
+				c = '\f'
+			case 'n':
+				c = '\n'
+			case 'r':
+				c = '\r'
+			case 't':
+				c = '\t'
+			case 'v':
+				c = '\v'
+			case delimiter:
+			case escapeChar:
+			default:
 				b = append(b, escapeChar)
 			}
 
--- a/eskip/string.go
+++ b/eskip/string.go
@@ -16,7 +16,7 @@ type PrettyPrintInfo struct {
 func escape(s string, chars string) string {
 	for i := 0; i < len(chars); i++ {
 		c := chars[i : i+1]
-		s = strings.Replace(s, c, "\\"+c, -1)
+		s = strings.Replace(s, c, `\`+c, -1)
 	}
 
 	return s
--- a/eskip/string.go
+++ b/eskip/string.go
@@ -14,6 +14,13 @@ type PrettyPrintInfo struct {
 }
 
 func escape(s string, chars string) string {
+	s = strings.Replace(s, "\a", `\a`, -1)
+	s = strings.Replace(s, "\b", `\b`, -1)
+	s = strings.Replace(s, "\f", `\f`, -1)
+	s = strings.Replace(s, "\n", `\n`, -1)
+	s = strings.Replace(s, "\r", `\r`, -1)
+	s = strings.Replace(s, "\t", `\t`, -1)
+	s = strings.Replace(s, "\v", `\v`, -1)
 	for i := 0; i < len(chars); i++ {
 		c := chars[i : i+1]
 		s = strings.Replace(s, c, `\`+c, -1)
--- a/eskip/string.go
+++ b/eskip/string.go
@@ -14,8 +14,6 @@ type PrettyPrintInfo struct {
 }
 
 func escape(s string, chars string) string {
-
-	s = strings.Replace(s, "\\", "\\\\", -1)
 	for i := 0; i < len(chars); i++ {
 		c := chars[i : i+1]
 		s = strings.Replace(s, c, "\\"+c, -1)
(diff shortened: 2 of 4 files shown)


aryszka:
see also: https://github.com/zalando/skipper/issues/562

eskip/lexer.go
```diff
@@ -131,12 +131,20 @@ func scanEscaped(delimiter byte, code string) ([]byte, string) {
 		c := code[0]
 		isDelimiter := c == delimiter
 		isEscapeChar := c == escapeChar
+		isN := c == 'n'
+		isR := c == 'r'
 
 		if escaped {
-			if !isDelimiter && !isEscapeChar {
+			if !isDelimiter && !isEscapeChar && !isN && !isR {
```
aryszka: can we support the full classic range of:

\b \f \n \r \t \v

?

And in addition, \" which is only a slightly different case

eskip/string.go
```diff
@@ -14,12 +14,15 @@ type PrettyPrintInfo struct {
 }
 
 func escape(s string, chars string) string {
-
-	s = strings.Replace(s, "\\", "\\\\", -1)
+	s = strings.Replace(s, `\r`, "\r", -1)
```
aryszka: I believe, since we are escaping here, this should be the opposite:

```
s = strings.Replace(s, "\r", `\\r`, -1)
```

the line for the `\` is correct.

aryszka:
thanks for this PR. The unescaping looks good to me, but I think we need to change the escaping.

eskip/string.go
ruudk: Actually, this is a trick. I replace the literals with the actual newline char 10, then continue the escaping process, and then I convert the newline char back (10) to the `\` + `n` chars. That way, we don’t get double escapes.

eskip/lexer.go
ruudk: Is it only these? Or are there more?

szuecs: Maybe \0 would be nice for some hacks

ruudk: I checked some Go source code and did not see how `\0` was done. So decided to leave that out for now.

szuecs:
Do we need to document the behavior?

aryszka:
> Do we need to document the behavior?

I think we are fixing a bug here, so it's fine not to. OTOH...

> Do we want to increase the minor version to show the lexer change happened?

I agree, and we need to document in the release notes, that in case someone depended on the broken behavior, then they may need to revisit their configurations. E.g: it may be possible to detect problems with the eskip tool:

```
eskip print < current-routes.eskip > verify.eskip && diff current-routes.eskip verify.eskip
```

aryszka:
thx for updating the PR!

aryszka:
:+1:

szuecs:
:+1:

eskip/string.go
```diff
@@ -14,11 +14,16 @@ type PrettyPrintInfo struct {
 }
 
 func escape(s string, chars string) string {
-
-	s = strings.Replace(s, "\\", "\\\\", -1)
```
AlexanderYastrebov, line 18: The escaping of backslash was lost here

szuecs: interesting finding

eskip/string_test.go
```diff
@@ -100,6 +100,16 @@ func TestRouteString(t *testing.T) {
 			Filters:     []*Filter{{"filter0", []interface{}{"arg"}}},
 			BackendType: DynamicBackend},
 		`* -> filter0("arg") -> <dynamic>`,
+	}, {
+		&Route{
+			Filters:     []*Filter{{"filter0", []interface{}{`Line 1\r\nLine 2`}}},
```
AlexanderYastrebov, line 105: This is wrong due to https://github.com/zalando/skipper/pull/1341/files#r773092391, the raw `Line 1\r\nLine 2` should be escaped as `"Line 1\\r\\nLine 2"`

eskip/string.go
AlexanderYastrebov: I have created a https://github.com/zalando/skipper/pull/1938 to fix

PR #1343, opened 2020-03-07
Remove escaping backslashes

ruudk:
I was looking at this code and it doesn't make sense to me what the purpose is. There is also no comment on why this is needed.

To me, the only thing that needs escaping are the `chars` that are provided as input. That is done in the `for` loop. Let's see if the tests fail on this.

### DIFF
--- a/eskip/string.go
+++ b/eskip/string.go
@@ -14,8 +14,6 @@ type PrettyPrintInfo struct {
 }
 
 func escape(s string, chars string) string {
-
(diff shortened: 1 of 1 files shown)


aryszka:
I think we can close this in favor of https://github.com/zalando/skipper/pull/1341

@ruudk wdyt?

ruudk:
Yes this was to prove that it was doing nothing :)

PR #1351, opened 2020-03-14
Release v1 move predicates

ruudk:
This extracts the `Path` predicate to `predicates` package.

I also introduced a mechanism to guard against duplicate predicates like `Weight`, `Path` and maybe more in the future. See `AppendPredicate` and `PrependPredicate` in `eskip.go`.

### DIFF
--- a/dataclients/kubernetes/ingress.go
+++ b/dataclients/kubernetes/ingress.go
@@ -113,22 +113,27 @@ func routeIDForCustom(namespace, name, id, host string, index int) string {
 	return routeID(namespace, name, host, "", "")
 }
 
-func setPath(m PathMode, r *eskip.Route, p string) {
+func setPath(m PathMode, r *eskip.Route, p string) error {
 	if p == "" {
-		return
+		return nil
 	}
 
 	switch m {
 	case PathPrefix:
-		r.Predicates = append(r.Predicates, &eskip.Predicate{
+		err := r.AppendPredicate(&eskip.Predicate{
 			Name: "PathSubtree",
 			Args: []interface{}{p},
 		})
+		if err != nil {
+			return err
+		}
 	case PathRegexp:
 		r.PathRegexps = []string{p}
 	default:
 		r.PathRegexps = []string{"^" + p}
 	}
+
+	return nil
 }
 
 func convertPathRule(
@@ -210,8 +215,14 @@ func convertPathRule(
 			HostRegexps: hostRegexp,
 		}
 
-		setPath(pathMode, r, prule.Path)
-		setTraffic(r, svcName, prule.Backend.Traffic, prule.Backend.noopCount)
+		err := setPath(pathMode, r, prule.Path)
+		if err != nil {
+			return nil, err
+		}
+		err = setTraffic(r, svcName, prule.Backend.Traffic, prule.Backend.noopCount)
+		if err != nil {
+			return nil, err
+		}
 		return r, nil
 
 	} else if err != nil {
@@ -231,8 +242,14 @@ func convertPathRule(
 			HostRegexps: hostRegexp,
 		}
 
-		setPath(pathMode, r, prule.Path)
-		setTraffic(r, svcName, prule.Backend.Traffic, prule.Backend.noopCount)
+		err := setPath(pathMode, r, prule.Path)
+		if err != nil {
+			return nil, err
+		}
+		err = setTraffic(r, svcName, prule.Backend.Traffic, prule.Backend.noopCount)
+		if err != nil {
+			return nil, err
+		}
 		return r, nil
 	}
 
@@ -243,26 +260,41 @@ func convertPathRule(
 		LBAlgorithm: getLoadBalancerAlgorithm(metadata),
 		HostRegexps: hostRegexp,
 	}
-	setPath(pathMode, r, prule.Path)
-	setTraffic(r, svcName, prule.Backend.Traffic, prule.Backend.noopCount)
+	err = setPath(pathMode, r, prule.Path)
+	if err != nil {
+		return nil, err
+	}
+	err = setTraffic(r, svcName, prule.Backend.Traffic, prule.Backend.noopCount)
+	if err != nil {
+		return nil, err
+	}
 	return r, nil
 }
 
-func setTraffic(r *eskip.Route, svcName string, weight float64, noopCount int) {
+func setTraffic(r *eskip.Route, svcName string, weight float64, noopCount int) error {
 	// add traffic predicate if traffic weight is between 0.0 and 1.0
 	if 0.0 < weight && weight < 1.0 {
-		r.Predicates = append([]*eskip.Predicate{{
+		err := r.AppendPredicate(&eskip.Predicate{
 			Name: traffic.PredicateName,
 			Args: []interface{}{weight},
-		}}, r.Predicates...)
+		})
+		if err != nil {
+			return err
+		}
 		log.Debugf("Traffic weight %.2f for backend '%s'", weight, svcName)
 	}
+
 	for i := 0; i < noopCount; i++ {
-		r.Predicates = append([]*eskip.Predicate{{
+		err := r.AppendPredicate(&eskip.Predicate{
 			Name: primitive.NameTrue,
 			Args: []interface{}{},
-		}}, r.Predicates...)
+		})
+		if err != nil {
+			return err
+		}
 	}
+
+	return nil
 }
 
 func applyAnnotationPredicates(m PathMode, r *eskip.Route, annotation string) error {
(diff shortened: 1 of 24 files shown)


eskip/eskip.go
```diff
@@ -185,6 +179,85 @@ type RouteInfo struct {
 	ParseError error
 }
 
+// HasPredicate checks if the route has predicate with the given name.
+func (r *Route) HasPredicate(name string) bool {
```
aryszka: this function could be more generally useful and would better reflect what it does, if it worked on any slice of predicates, and not only on those that are on the same route. E.g:

```
func PredicatesContain(p []*Predicate, name string) bool {}
```

eskip/eskip.go
```diff
@@ -185,6 +179,85 @@ type RouteInfo struct {
...
+		if p.Name == name {
+			return true
+		}
+	}
+
+	return false
+}
+
+// GetPredicates returns all predicates that match the given name.
+func (r *Route) GetPredicates(name string) []*Predicate {
```
aryszka: similarly to HasPredicate, this function could be more generally useful and would better reflect what it does, if it worked on any slice of predicates, and not only on those that are on the same route. E.g:

```
func PredicatesByName(p []*Predicate, name string) []*Predicate {}
```

eskip/eskip.go
```diff
@@ -185,6 +179,85 @@ type RouteInfo struct {
...
+		if p.Name == name {
+			pp = append(pp, p)
+		}
+	}
+
+	return pp
+}
+
+// GetPredicate returns matching predicate or error when multiple predicates are given.
+func (r *Route) GetPredicate(name string) (*Predicate, error) {
```
aryszka: This could be derived from the above suggested `PredicatesByName`.

eskip/eskip.go
```diff
@@ -185,6 +179,85 @@ type RouteInfo struct {
...
+	}
+
+	if len(pp) > 1 {
+		return nil, fmt.Errorf("multiple predicates of the same name: %d %s", len(pp), name)
+	}
+
+	return pp[0], nil
+}
+
+func (r *Route) PrependPredicate(pp ...*Predicate) error {
```
aryszka: Performance nitpicking: this would be possible to do by just appending once to a temp var, checking the temp var, and then if it is fine, setting the reference on the route to the var. This way there is only one allocation. Pseudo:

```
next := append(pp, r.Predicates...)
if len(PredicatesByName(next, "Path")) > 1 || len(PredicatesByName(next, "Weight")) > 1 {
    return fmt.Errorf()
}

r.Predicates = next
return nil
```

eskip/eskip.go
```diff
@@ -185,6 +179,85 @@ type RouteInfo struct {
...
+	for _, p := range pp {
+		switch p.Name {
+		case "Path", "Weight":
```
aryszka: the "PathSubtree" is missing.

eskip/eskip.go
aryszka: please also check for "PathSubtree"

aryszka: Please add godoc for exported functions.

eskip/eskip.go
```diff
@@ -185,6 +179,85 @@ type RouteInfo struct {
...
+			}
+		}
+
+		r.Predicates = append([]*Predicate{p}, r.Predicates...)
+	}
+
+	return nil
+}
+
+func (r *Route) AppendPredicate(pp ...*Predicate) error {
```
aryszka: See my comments for PrependPredicate. And please add godoc.

eskip/eskip.go
```diff
@@ -185,6 +179,85 @@ type RouteInfo struct {
...
+		}
+
+		r.Predicates = append(r.Predicates, p)
+	}
+
+	return nil
+}
+
+func (r *Route) ReplacePredicates(pp ...*Predicate) error {
+	r.Predicates = make([]*Predicate, 0, len(pp))
```
aryszka: Shouldn't we first check whether there is an error, and only then replace the existing predicates. To my that would make more sense. With the proposed way of making PredicatesByName, this could become:

```
if len(PredicatesByName(pp, "Path")) > 1 ||
    len(PredicatesByName(pp, "PathSubtree")) > 1 ||
    len(PredicatesByName(pp, "Weight")) > 1 {
    return fmt.Errorf()
}

r.Predicates = pp
return nil
```

eskip/eskip.go
```diff
@@ -185,6 +179,85 @@ type RouteInfo struct {
...
+			}
+		}
+
+		r.Predicates = append(r.Predicates, p)
+	}
+
+	return nil
+}
+
+func (r *Route) ReplacePredicates(pp ...*Predicate) error {
```
aryszka: Please add godoc for exported functions.

predicates/path/path.go
```diff
@@ -0,0 +1,48 @@
...
+
+// New creates a new Path predicate specification.
+func New() routing.PredicateSpec { return &spec{} }
+
+func (s *spec) Name() string {
+	return name
+}
+
+func (s *spec) Create(args []interface{}) (routing.Predicate, error) {
+	if len(args) == 0 || len(args) > 2 {
```
aryszka: `len(args) != 1`?

predicates/path/path.go
```diff
@@ -0,0 +1,48 @@
...
+const name = "Path"
+
+// New creates a new Path predicate specification.
+func New() routing.PredicateSpec { return &spec{} }
+
+func (s *spec) Name() string {
+	return name
+}
+
+func (s *spec) Create(args []interface{}) (routing.Predicate, error) {
```
aryszka: I haven't checked, but am wondering what happens if this spec is not defined at all? The PathSubtree predicate works somehow.

dataclients/kubernetes/ingress.go
```diff
@@ -291,7 +322,12 @@ func applyAnnotationPredicates(m PathMode, r *eskip.Route, annotation string) er
 		}
 	}
 
-	r.Predicates = append(r.Predicates, predicates...)
+	for _, p := range predicates {
```
aryszka: nitpicking: would it be nicer to first append into a tmp var, and then use ReplacePredicates?

eskip/eskip_test.go
```diff
@@ -123,11 +126,11 @@ func TestParseRouteExpression(t *testing.T) {
 			Backend: "https://www.example.org"},
 		false,
 	}, {
-		"double path predicates",
-		`Path("/one") && Path("/two") -> "https://www.example.org"`,
-		nil,
-		true,
-	}, {
+		//	"double path predicates",
```
aryszka: let's delete this, we don't need it

dataclients/kubernetes/ingress.go
ruudk: Handled differently :)

eskip/eskip.go
ruudk: This PR doesn't touch PathSubtree (yet) so that will be done in another PR.

predicates/path/path.go
ruudk: It's used here https://github.com/zalando/skipper/blob/79cbd3e1c15c438dfcaf0e79c7ee34dc232596b1/skipper.go#L1055

eskip/predicates.go
```diff
@@ -0,0 +1,98 @@
+package eskip
+
+import (
+	"fmt"
+)
+
+// PredicatesContains checks if the route has predicate with the given name.
+func PredicatesContains(p []*Predicate, name string) bool {
```
aryszka: PredicatesContain (plural)

ruudk: Done

predicates/path/path.go
aryszka: Right. What I mean what if we remove it from there, too. The point would be to make it analogous to PathSubtree.

Another option would be to make PathSubtree analogous to the suggested one, I'm just not sure if we benefit from noop types if they are avoidable.

eskip/predicates.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	return pp[0], nil
+}
+
+// ValidatePredicates returns an error when certain predicates are added multiple times.
+func ValidatePredicates(pp []*Predicate) error {
+	for _, p := range pp {
+		switch p.Name {
+		case "Path", "Weight":
```
aryszka: Can include the PathSubtree handling in this PR, too? The rules should be sg like this:

- 0 or 1 Weight predicates
- 0 or 1 of either Path or PathSubtree predicates

ruudk: Let's do this in a new PR,

ruudk: Maybe you can create an issue to track it then I'll follow up on it.

predicates/path/path.go
ruudk: Let's see if it works https://github.com/zalando/skipper/pull/1351/commits/499e5da57203c4b31ecb6ed5fcd054a4877918df

ruudk:
@aryszka @szuecs Ready for review :) Please merge so that I can rebase https://github.com/zalando/skipper/pull/1354 and continue.

PR #1354, opened 2020-03-17
Remove legacy Method predicate and add support for singular/plural in Method(s) predicate

ruudk:
Remove legacy Method predicate and add support for singular/plural in Method(s) predicate

### DIFF
--- a/routing/matcher_test.go
+++ b/routing/matcher_test.go
@@ -11,15 +11,16 @@ import (
 
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/pathmux"
+	"github.com/zalando/skipper/predicates"
 	"github.com/zalando/skipper/predicates/methods"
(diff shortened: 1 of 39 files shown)


routing/matcher_test.go
```diff
@@ -11,6 +11,7 @@ import (
 
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/pathmux"
+	"github.com/zalando/skipper/predicates/methods"
```
ruudk: @aryszka this causes a import cycle:
```
# github.com/zalando/skipper/routing
import cycle not allowed in test
package github.com/zalando/skipper/routing (test)
	imports github.com/zalando/skipper/predicates/methods
	imports github.com/zalando/skipper/routing
```

szuecs: you would need to move routing.Predicate and routing.PredicateSpec to predicates package

aryszka:
:+1:

PR #1379, opened 2020-03-23
feat: added support for multiple variables in paths (#1345)

### LINKED ISSUE #1345: apiusagemonitoring does not support multiple variables in path segment
**Describe the bug**
apiusagemonitoring filter does not support multiple variables in path segment.

Consider path with multiple variables, e.g. `/exchange-rates/{source-currency}-{target-currency}`

Current implementation splits path into segments by `/`:
https://github.com/zalando/skipper/blob/12212234a5ce108bd98bd46fefc230b68e5443ba/filters/apiusagemonitoring/spec.go#L296
and assumes single variable in the segment:
https://github.com/zalando/skipper/blob/12212234a5ce108bd98bd46fefc230b68e5443ba/filters/apiusagemonitoring/spec.go#L304
https://github.com/zalando/skipper/blob/12212234a5ce108bd98bd46fefc230b68e5443ba/filters/apiusagemonitoring/spec.go#L334-L347

therefore example path would not be properly handled.

**Expected behavior**
Metrics are collected for path with multiple variables in path segment.

**Observed behavior**
Metrics are not collected for path with multiple variables in path segment.

tkrop:
fixes #1345.

- Implemented a new logic to create `pathTemplates` and `pathPatterns` for regular expression matching supporting multiple variables in a path segment.
- Unified `pathTemplate` and `pathLabel` to exposed the label syntax only.
- Implemented a new regular expression schema for matching request paths.

The general problem is how to effectively match paths. The main approach is based on the OpenAPI template language which allows more sophisticated matching as the Skipper internal language. The following rules are applied:
- Trailing slashes are ignored mapped to the same end-point as stated in our [API Guidelines](https://opensource.zalando.com/restful-api-guidelines/#136)
- Missing leading slashes in path templates and path requests are gracefully ignored, however, URI normalization and path rules should take care that this is not necessary.
- Multiple intermediate slashes in a template and path are treated as a single slashes as suggests but not enforces by path normalization.
- Path variables can now be marked (`?` at the end) to allow for empty matching - default is still on not empty. Note: path matching conflicts are majorly excluded by the precedence model that favors longer path templates. However, as result, the ambiguous request path `GET /orders/` would match `GET /orders/{id}` instead of as might be expected `GET /orders`.

This set of rules takes into account, that many service implementations are automatically treat request paths as normalized and even are able to match path templates, if request contain empty path variables. Usually, this is not use however, we can now add automated discovery of such path variables in the `api-monitoring-controller`, on user request.


filters/apiusagemonitoring/spec.go
```diff
@@ -17,15 +17,15 @@ const (
...
 var (
-	log                         = logrus.WithField("filter", Name)
-	regexVarPathPartCurlyBraces = regexp.MustCompile("^{([^:{}]+)}$")
-	regexVarPathPartColon       = regexp.MustCompile("^:([^:{}]+)$")
+	log                   = logrus.WithField("filter", Name)
+	regexpMultiSlash      = regexp.MustCompile("/+")
+	regexpLeadingSlash    = regexp.MustCompile("^/?")
+	regexpTrailingSlash   = regexp.MustCompile("/?$")
+	rexexpSlashColumnVar  = regexp.MustCompile("/:([^:{}/]*)")
+	rexexpCurlyBracketVar = regexp.MustCompile("{([^{}]*)}")
```
tkrop: Should we exclude `/` (slashes) from variable names? Is it likely or possible, that pattern matcher libraries implement it simila?

filters/apiusagemonitoring/spec_test.go
```diff
@@ -222,7 +222,8 @@ func Test_CreateFilter_FullConfigSingleApi(t *testing.T) {
 		"path_templates": [
```
tkrop: We can simplify the path_template, since we now test the generation of labels and regular expressions in its own unit test.

filters/apiusagemonitoring/spec_test.go
```diff
@@ -306,7 +314,8 @@ func Test_CreateFilter_FullConfigMultipleApis(t *testing.T) {
 			"path_templates": [
```
tkrop: Can be simplified as above.

filters/apiusagemonitoring/spec.go
```diff
@@ -292,56 +290,23 @@ func (s *apiUsageMonitoringSpec) buildClientTrackingInfo(apiIndex int, api *apiC
...
-		return matches[1]
-	}
-	// it is not a placeholder
-	return ""
+func createPathPattern(path string) string {
+	path = regexpMultiSlash.ReplaceAllString(path, "/")
+	path = regexpLeadingSlash.ReplaceAllString(path, "^/*")
+	path = regexpTrailingSlash.ReplaceAllString(path, "/*$")
+	path = rexexpSlashColumnVar.ReplaceAllString(path, "/.+")
+	path = rexexpCurlyBracketVar.ReplaceAllString(path, ".+")
```
tkrop: Currently, we are missing a step to enable matching of multiple slashes within paths.

aryszka:
thanks! It generally looks good to me, but I still would like to have a deeper look. I'll review the PR asap

filters/apiusagemonitoring/spec_test.go
tkrop: Done.

filters/apiusagemonitoring/spec_test.go
tkrop: Done.

filters/apiusagemonitoring/spec.go
tkrop: Done.

filters/apiusagemonitoring/spec_test.go
tkrop: Done.

filters/apiusagemonitoring/spec_test.go
tkrop: Done.

tkrop:
@aryszka thanks for the first review. I have changed a bit after discussing the details within the API guild. I'm quite confident, that everything is as we could best do it for now. Please have a look.

I introduced a [matrix test pattern](https://github.com/zalando/skipper/pull/1379/files#diff-5a01fadd086235c20dec7277aa9b54f6R497-R546), that we use in our projects for some time and that you might find interesting to adopt in skipper to create more expressive tests.

tkrop:
:+1:

GoWind:
:+1:

.project
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
aryszka: please drop the editor configuration

filters/apiusagemonitoring/pathinfo.go
```diff
@@ -23,7 +22,7 @@ type pathInfo struct {
 	metricPrefixedPerClient sync.Map
 }
 
-func newPathInfo(applicationId, tag, apiId, pathTemplate, pathLabel string, clientTracking *clientTrackingInfo) *pathInfo {
+func newPathInfo(applicationId, tag, apiId, pathLabel string, clientTracking *clientTrackingInfo) *pathInfo {
```
aryszka: would it be worth renaming `pathLabel` to `pathTemplate`?

filters/apiusagemonitoring/spec.go
```diff
@@ -290,58 +286,61 @@ func (s *apiUsageMonitoringSpec) buildClientTrackingInfo(apiIndex int, api *apiC
...
-	rawRegEx.WriteString(strings.Join(matcherPathParts, `\/`))
-	rawRegEx.WriteString(regexOptionalSlashes)
-	rawRegEx.WriteString("$")
+var (
+	regexpMultipleSlashes   = regexp.MustCompile(`/+`)
+	regexpLeadingSlashes    = regexp.MustCompile(`^/*`)
+	regexpTrailingSlashes   = regexp.MustCompile(`/*$`)
+	regexpMiddleSlashes     = regexp.MustCompile(`([^/^])/+([^/*])`)
+	rexexpSlashColumnVar    = regexp.MustCompile(`/:([^:{}/]*)`)
+	rexexpCurlyBracketVar   = regexp.MustCompile(`{([^{}]*?)([?]?)}`)
```
aryszka: the question mark after the star is not needed, if I am not mistaken

filters/apiusagemonitoring/spec.go
```diff
@@ -290,58 +286,61 @@ func (s *apiUsageMonitoringSpec) buildClientTrackingInfo(apiIndex int, api *apiC
...
-	rawRegEx.WriteString("^")
-	rawRegEx.WriteString(regexOptionalSlashes)
-	rawRegEx.WriteString(strings.Join(matcherPathParts, `\/`))
-	rawRegEx.WriteString(regexOptionalSlashes)
-	rawRegEx.WriteString("$")
+var (
+	regexpMultipleSlashes   = regexp.MustCompile(`/+`)
+	regexpLeadingSlashes    = regexp.MustCompile(`^/*`)
+	regexpTrailingSlashes   = regexp.MustCompile(`/*$`)
+	regexpMiddleSlashes     = regexp.MustCompile(`([^/^])/+([^/*])`)
```
aryszka: if I am not mistaken, we can just replace every occurence of multiple consecutive slashes. Then this expression could become simply: `/{2,}`

filters/apiusagemonitoring/spec.go
```diff
@@ -290,58 +286,61 @@ func (s *apiUsageMonitoringSpec) buildClientTrackingInfo(apiIndex int, api *apiC
...
+	path = regexpMiddleSlashes.ReplaceAllString(path, "$1/+$2")
+	path = regexpEscapeAfterChars.ReplaceAllString(path, "\\$1")
+	return path
+}
+
+// selectPathVarPattern select the correct path variable pattern depending
+// on the path variable syntax. A trailing question mark is interpreted as
+// a path variable that is allowed to be empty.
+func (ph defaultPathHandler) SelectPathVarPattern(match string) string {
+	log.Infof("match %s", match)
```
aryszka: I think this log entry should be enabled only in debug mode. Please change it to Debugf.

filters/apiusagemonitoring/spec.go
```diff
@@ -290,58 +286,61 @@ func (s *apiUsageMonitoringSpec) buildClientTrackingInfo(apiIndex int, api *apiC
...
+	path = regexpTrailingSlashes.ReplaceAllString(path, "/*$")
+	path = regexpMiddleSlashes.ReplaceAllString(path, "$1/+$2")
+	path = regexpEscapeAfterChars.ReplaceAllString(path, "\\$1")
+	return path
+}
+
+// selectPathVarPattern select the correct path variable pattern depending
+// on the path variable syntax. A trailing question mark is interpreted as
+// a path variable that is allowed to be empty.
+func (ph defaultPathHandler) SelectPathVarPattern(match string) string {
```
aryszka: SelectPathVarPattern doesn't depend on defaultPathHandler, it should be just a simple function instead of a method.

filters/apiusagemonitoring/spec.go
```diff
@@ -290,58 +286,61 @@ func (s *apiUsageMonitoringSpec) buildClientTrackingInfo(apiIndex int, api *apiC
...
+	rexexpCurlyBracketVar   = regexp.MustCompile(`{([^{}]*?)([?]?)}`)
+	regexpEscapeBeforeChars = regexp.MustCompile(`([.*+?\\])`)
+	regexpEscapeAfterChars  = regexp.MustCompile(`([{}[\]()|])`)
+)
 
-	matcher = rawRegEx.String()
-	normalizedPathTemplate = strings.Join(normalizedPathTemplateParts, "/")
-	pathLabel = strings.Join(pathLabelParts, "/")
+// PathHandler path handler interface.
+type PathHandler interface {
```
aryszka: please don't export this type, change its name to lowercase. The same for it's methods.

.project
tkrop: Sorry, will do.

filters/apiusagemonitoring/pathinfo.go
tkrop: Overlooked :)

filters/apiusagemonitoring/spec.go
tkrop: Yes `{2,}` is exactly the meaning of `+`, however, I would always prefer the latter.

filters/apiusagemonitoring/spec.go
tkrop: In this case it is, because else the match would change in an undesired way. E.g. `/resource/{a}-{b}`using `{([^{}]*?)([?]?)}` creates two matches `{a}` and `{c}`, while `{([^{}]*)([?]?)}` would result in a greedy behavior matching only once `{a}-{b}`.

filters/apiusagemonitoring/spec.go
tkrop: Yes, you are right. It is internal for testing.

tkrop:
@aryszka thanks for the great review. Have changed all as requested except for the regular expression :). See comment.

tkrop:
:+1:

filters/apiusagemonitoring/spec.go
aryszka: I didn't mean `/+` -> `/{2,}` (the meaning of `+` is `{1,}`. What i meant was: `([^/^])/+([^/*])` -> `/{2,}`. Then the following expression could be changed like this. From:

```
regexpMiddleSlashes.ReplaceAllString(path, "$1/+$2")
```

to:

```
regexpMiddleSlashes.ReplaceAllString(path, "")
```

Which is somewhat simpler.

aryszka:
thanks for the updates, checking them shortly

aryszka:
:+1:

filters/apiusagemonitoring/spec.go
tkrop: As far as I can see your suggestion would remove all occurrences of more than one two slash, meaning there is no slash left. However, I want to replace any number of slashes in the middle against a regular expression pattern `\+`.

I agree, that my condition looks a bit difficult, because I need to identify middle slashes (which requires the regular expression to be context sensitive) after I already did some replacements for leading and trailing slashes. May the expression and the whole algorithm can be simplified by first replacing all slashes against `\+` and than modifying the first and the last to match `\*`.

I think, I give this a try.

tkrop: The changes would look as follows: 

* leading slashes `^/*` to `^/*\+?`
* trailing slashes `/*$` to `(|/+\+?)$`
* middle slashes `([^/^])/+([^/*])` to `/+`

The reason for the increasing complexity in leading and trailing slashes is, that they are reused for creating the template. The change would also lead to unexpected side effects on the template, that I could compensate by reintroducing the before regular expressions in addition.

@aryszka what do you think.

aryszka: I see. Thanks for the explanation.

tkrop:
:+1:

PR #1380, opened 2020-03-24
Add `inlineContentIfStatus` filter

### LINKED ISSUE #1355: Custom response body for auth reject response
Currently if you use [auth](https://github.com/zalando/skipper/blob/cce10ed4a000ee627df5d7988b658745505d41a9/filters/auth/auth.go) filters, you end up with a 401 response and an empty page. This ok in most situation, but in a user facing scenario, or when protecting an endpoint whose clients expect a JSON response this can cause issues.

The most simple implementation would allow to specify a string (template?) for the response in [filter options](https://github.com/zalando/skipper/blob/cce10ed4a000ee627df5d7988b658745505d41a9/filters/auth/tokeninfo.go#L21-L26). Potentially might be great to have another filter like `oauthOnRejectedResponseBody('{"error": "Not Allowed"}')` so that the response can be set per-route.

Can not think of any.

Maybe, but no time

grassator:
Fixes #1355

### DIFF
--- a/filters/builtin/inlinecontentifstatus.go
+++ b/filters/builtin/inlinecontentifstatus.go
@@ -1,0 +1,92 @@
+package builtin
+
+import (
+	"bytes"
+	log "github.com/sirupsen/logrus"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+
+	"github.com/zalando/skipper/filters"
+)
+
+type inlineContentIfStatus struct {
+	statusCode int
+	text       string
+	mime       string
+}
+
+// Creates a filter spec for the inlineContent() filter.
+//
+//     * -> inlineContentIfStatus(401, "{\"foo\": 42}", "application/json") -> "https://www.example.org"
+//
+// It accepts three arguments: the statusCode code to match, the content and the optional content type.
+// When the content type is not set, it tries to detect it using http.DetectContentType.
+//
+// The filter replaces the response coming from the backend or the following filters.
+//
+func NewInlineContentIfStatus() filters.Spec {
+	return &inlineContentIfStatus{}
+}
+
+func (c *inlineContentIfStatus) Name() string { return InlineContentIfStatusName }
+
+func (c *inlineContentIfStatus) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) < 2 || len(args) > 3 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	var (
+		f  inlineContentIfStatus
+		ok bool
+	)
+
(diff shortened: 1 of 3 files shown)


filters/builtin/inlinecontentifstatus.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (c *inlineContentIfStatus) Request(filters.FilterContext) {}
+
+func (c *inlineContentIfStatus) Response(ctx filters.FilterContext) {
+	if ctx.Response().StatusCode != c.statusCode {
+		return
+	}
+	rsp := ctx.Response()
+
+	err := rsp.Body.Close()
```
grassator: I am not sure if I need to do this or something else with the original request...

ruiaraujo: We definitely need to do this. we learned the hard way. :smile:

grassator: @ruiaraujo Is this sufficient though or do I need to explicitly consume the body to avoid weird result for the backend server?

aryszka: we need to close the response body as for the contract with the library used on our side (net/http, see the Body field docs in https://godoc.org/net/http#Response)

about cosuming the response body: that's not mandatory, a client can just refuse to consume it. Sometimes, with long streams, it may result in logspam on the server side, but it's still better not to put that load on the pipe, IMHO. In addition to this, I expect that in the typical use case of this filter, the body won't be too long, which means that probably all the bytes are already buffered on the client side, so we can just safely discard it. So I suggest to not consume it.

filters/builtin/inlinecontentifstatus.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		log.Error(err)
+	}
+
+	contentLength := len(c.text)
+	rsp.ContentLength = int64(contentLength)
+	rsp.Header.Set("Content-Type", c.mime)
```
aryszka: can we make the mime type optional? The server can automatically detect the common mime types.

aryszka: oh, i just checked the CreateFilter() function. It's nicer like this :+1:

aryszka:
:+1:

ruiaraujo:
:+1:

ruudk:
@grassator Can you please also update the docs? 

@aryszka This should be mandatory for new features imho, maybe a template with checklist can help with this?

szuecs:
@ruudk I agree every PR should have tests and docs, always to keep it in sync. I am not sure about a template checklist. Maybe we can have a PR template in GH?!

PR #1382, opened 2020-03-25
Feature/decompress

aryszka:
add a decompress counterpart of the compress filter

### DIFF
--- a/filters/builtin/decompress.go
+++ b/filters/builtin/decompress.go
@@ -14,9 +14,6 @@ import (
 	"github.com/zalando/skipper/filters"
 )
 
-// precursor for an open source counterpart of the compress() filter
-// https://godoc.org/github.com/zalando/skipper/filters/builtin#NewCompress
-
 const (
 	// DecompressionNotPossible is the state-bag key to indicate
 	// to the subsequent filters during response processing that the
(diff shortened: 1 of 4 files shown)


filters/builtin/decompress.go
```diff
@@ -0,0 +1,231 @@
...
+	log "github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/filters"
+)
+
+const (
+	// DecompressionNotPossible is the state-bag key to indicate
+	// to the subsequent filters during response processing that the
+	// content is compressed, but decompression was not possible, e.g
+	// because the encoding is not supported.
+	DecompressionNotPossible = "filter::decompress::impossible"
```
gargravarr, line 22: nit: either call it DecompressionImpossible, or its value be something like `"filter::decompress::notpossible"` or `"filter::decompress::not-possible"`

filters/builtin/decompress.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// DecompressionError is the state-bag key to indicate to the
+	// subsequent filters during response processing that th
```
gargravarr, line 25: th -> the

filters/builtin/decompress.go
```diff
@@ -0,0 +1,231 @@
...
+
+	return err
+}
+
+func (e decodingError) Error() string {
+	switch {
+	case e.decoder == nil:
+		return e.original.Error()
+	case e.original == nil:
+		return e.decoder.Error()
```
gargravarr, line 145: I think it would be less confusing if you used != nil, like this:
```
case e.decoder != nil:
		return e.decoder.Error()
	case e.original != nil:
		return e.original.Error()
```

aryszka:
:+1:

PR #1383, opened 2020-03-25
create filter for setting the X-Forwarded-* headers

aryszka:
add a filter to support the standard proxy headers explicitly

### DIFF
--- a/filters/xforward/xforward.go
+++ b/filters/xforward/xforward.go
@@ -1,0 +1,75 @@
+package xforward
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/zalando/skipper/filters"
+)
+
+const (
+	Name      = "xforward"
+	NameFirst = "xforwardFirst"
+)
+
+type filter struct {
+	reverse bool
+}
+
+// New creates a specification for the xforward filter
+// that appends the remote IP of the incoming request to the
+// X-Forwarded-For header, and sets the X-Forwarded-Host header
+// to the value of the incoming request's Host header.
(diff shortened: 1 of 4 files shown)


filters/xforward/xforward.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/zalando/skipper/filters"
+)
+
+const (
+	Name      = "xforward"
```
gargravarr, line 11: My linter is complaining about this: exported const Name should have comment (or a comment on this block) or be unexported

filters/xforward/xforward.go
```diff
@@ -0,0 +1,75 @@
...
+	NameFirst = "xforwardFirst"
+)
+
+type filter struct {
+	reverse bool
```
gargravarr, line 16: I think it would be more clear if this variable was named `prepend`

filters/xforward/xforward_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+func getForwardedFor(headerValue string, reverse bool) []string {
+	values := strings.Split(headerValue, ",")
+	for i := range values {
+		values[i] = strings.TrimSpace(values[i])
+	}
+
+	if len(values) == 1 && values[0] == "" {
+		return nil
+	}
```
gargravarr, line 27: I'd put this before the previous loop.

gargravarr:
:+1:

aryszka:
:+1:

PR #1387, opened 2020-03-27
create grant flow filter

aryszka:
**[WIP]**


filters/auth/grantflowstate.go
```diff
@@ -0,0 +1,100 @@
...
+		return "", err
+	}
+
+	nonce, err := encrypter.CreateNonce()
+	if err != nil {
+		return "", err
+	}
+
+	state := state{
+		Rand:       randString(randomStringLength),
```
muaazsaleem, line 51: We should use `crypto` based `rand` here

filters/auth/grantflowstate.go
```diff
@@ -0,0 +1,100 @@
...
+	"fmt"
+	"time"
+
+	"github.com/zalando/skipper/secrets"
+)
+
+const randomStringLength = 20
+
+type state struct {
+	Rand       string `json:"rand"`
```
szuecs, line 15: As discussed we can remove this field

soceanainn:
As a backend engineer - the possibility of returning some raw HTML from an API endpoint and having Skipper handling all of the OAuth2 login flow is something that's very exciting as it opens up a lot of possibilities for our team internally 👍

aryszka:
closing this as it was finalized in: https://github.com/zalando/skipper/pull/1609
Thanks for the contributors!

PR #1391, opened 2020-03-31
Log call duration, backendType + backend on proxy error

ruudk:
Instead of logging:
```
time="2020-03-31T14:44:16Z" level=error msg="error while proxying, route kube_frontend__frontend__www_ticketswap_nl with backend , status code 499: dialing failed false: context canceled"
```

it will now log:
```
time="2020-03-31T14:44:16Z" level=error msg="error while proxying after 5m41s, route kube_frontend__frontend__www_ticketswap_nl with backend , status code 499: dialing failed false: context canceled"
```

Also, discovered that when you use loadbalanced backends, the error log will be an empty string:
```
time="2020-03-31T14:44:16Z" level=error msg="error while proxying, route kube_frontend__frontend__www_ticketswap_nl with backend , status code 499: dialing failed false: context canceled"
```

Look at `with backend , `. 

This is because the single `Backend` is selected there. If you have a route `* -> "https://zalando.de"` it would show it, but if you have `* -> <roundRobin, "ip:port">` it won't.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1241,8 +1241,10 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 		code = p.defaultHTTPStatus
 	case ok && perr.err == errRatelimit:
 		code = perr.code
+	case code == 499:
+		p.log.Errorf("client canceled after %v, route %s with backend %s, status code %d: %v", time.Since(ctx.startServe), id, backend, code, err)
 	default:
-		p.log.Errorf("error while proxying, route %s with backend %s, status code %d: %v", id, backend, code, err)
+		p.log.Errorf("error while proxying after %v, route %s with backend %s, status code %d: %v", time.Since(ctx.startServe), id, backend, code, err)
 	}
 
 	p.sendError(ctx, id, code)
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -37,10 +37,11 @@ import (
 )
 
 const (
-	proxyBufferSize      = 8192
-	unknownRouteID       = "_unknownroute_"
-	unknownRouteBackend  = "<unknown>"
-	backendIsProxyHeader = "X-Skipper-Proxy"
+	proxyBufferSize         = 8192
+	unknownRouteID          = "_unknownroute_"
+	unknownRouteBackendType = "<unknown>"
+	unknownRouteBackend     = "<unknown>"
+	backendIsProxyHeader    = "X-Skipper-Proxy"
 
 	// Number of loops allowed by default.
 	DefaultMaxLoopbacks = 9
@@ -1196,10 +1197,12 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 	}
 
 	id := unknownRouteID
+	backendType := unknownRouteBackendType
 	backend := unknownRouteBackend
 	if ctx.route != nil {
 		id = ctx.route.Id
-		backend = ctx.route.Backend
+		backendType = ctx.route.BackendType.String()
+		backend = fmt.Sprintf("%s://%s", ctx.request.URL.Scheme, ctx.request.URL.Host)
 	}
 
 	code := http.StatusInternalServerError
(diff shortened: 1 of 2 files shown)


proxy/proxy.go
```diff
@@ -1196,10 +1197,12 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 	}
 
 	id := unknownRouteID
+	backendType := unknownRouteBackendType
 	backend := unknownRouteBackend
 	if ctx.route != nil {
 		id = ctx.route.Id
-		backend = ctx.route.Backend
+		backendType = ctx.route.BackendType.String()
+		backend = ctx.request.URL.Host
```
aryszka: can we please include the scheme from ctx.request.URL?

ruudk: Done

proxy/proxy.go
```diff
@@ -1241,8 +1244,10 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 		code = p.defaultHTTPStatus
 	case ok && perr.err == errRatelimit:
 		code = perr.code
+	case code == 499:
+		p.log.Errorf("client canceled after %s, route %s with backend %s %s, status code %d: %v", time.Since(ctx.startServe), id, backendType, backend, code, err)
 	default:
-		p.log.Errorf("error while proxying, route %s with backend %s, status code %d: %v", id, backend, code, err)
+		p.log.Errorf("error while proxying after %s, route %s with backend %s %s, status code %d: %v", time.Since(ctx.startServe), id, backendType, backend, code, err)
```
aryszka: the first parameter should be %v because we feed it with a time.Duration value

proxy/proxy.go
```diff
@@ -1241,8 +1244,10 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 		code = p.defaultHTTPStatus
 	case ok && perr.err == errRatelimit:
 		code = perr.code
+	case code == 499:
+		p.log.Errorf("client canceled after %s, route %s with backend %s %s, status code %d: %v", time.Since(ctx.startServe), id, backendType, backend, code, err)
```
aryszka: he first parameter should be %v because we feed it with a time.Duration value

ruudk: Nope, I want to show the string representation, see example: https://play.golang.org/p/ZQW0Pl-UnlA

proxy/proxy.go
ruudk: Nope, I want to show the string representation, see example: https://play.golang.org/p/ZQW0Pl-UnlA

proxy/proxy.go
aryszka: i updated the example a bit: https://play.golang.org/p/ZBwdgnJLmZh
plus, here's the doc, seek for %s: https://godoc.org/fmt
;)

proxy/proxy.go
aryszka: i updated the example a bit: https://play.golang.org/p/ZBwdgnJLmZh
plus, here's the doc, seek for %s: https://godoc.org/fmt
;)

proxy/proxy.go
ruudk: Ah, so in this case, the output will still be the same:
```
2009-11-10 23:00:00 +0000 UTC m=+0.000000001
2009-11-10 23:05:41 +0000 UTC m=+341.000000001
5m41s
5m41s
```

but it's good practice to always use `%v` because that way, it will default to the correct type depending on the input?

aryszka:
:+1:

proxy/proxy.go
aryszka: indeed, %v is a safe default. But it's a best practice to use %s %d or similar well known cases, where we know the matched input. There are various reasons, e.g. static checking can detect potential mistakes. It's worth having a peek sometimes in the stdlib implementation for examples about the intended usage.

ruudk: Changed.

ruudk:
The build is green (https://travis-ci.org/github/zalando/skipper/builds/669333506?utm_source=github_status&utm_medium=notification) but it didn't update Github. Can this be merged? :pray:

szuecs:
I try to close reopen to trigger the event again

ruudk:
Retry

aryszka:
:+1:

PR #1393, opened 2020-04-01
Change default value for no flow id

ruudk:
Looks better in the logs.

```
level=error msg="client canceled after 48.961234ms, route x with backend lb http://100.109.163.119:80, flow id -, status code 499: dialing failed false: context canceled"
```

Kinda weird, that single dash.

Much better:
```
level=error msg="client canceled after 48.961234ms, route x with backend lb http://100.109.163.119:80, flow id not set, status code 499: dialing failed false: context canceled"
```

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -39,7 +39,7 @@ import (
 
 const (
 	proxyBufferSize         = 8192
-	unknownFlowId           = "-"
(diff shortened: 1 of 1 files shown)


aryszka:
:+1:

szuecs:
:+1:

PR #1394, opened 2020-04-01
remove early close of the response body

aryszka:
removing early close of the response body as it is closed by the enclosing function


szuecs:
:+1:

aryszka:
:+1:

PR #1395, opened 2020-04-02
Log proxy errors in access log

ruudk:
Logging the proxy errors without the request context is pretty useless. The only way to find out which request failed is to enable flowId's and try to map them together. On a production server with lots of requests this is almost impossible.

To make it easier to find out which request failed, the proxy errros are now stored in the Context.

When the access log is enabled, the proxy error is appended to the end of the access log.

Normal access log looks like this (watch the `"-"` at the end):
```
::1 - - [02/Apr/2020:09:34:38 +0200] "GET /robots.txt HTTP/1.1" 200 11 "-" "HTTPie/2.0.0" 0 localhost:9090 - - "-"
```

When an error occurred while proxying, it looks like this:
```
::1 - - [02/Apr/2020:09:37:11 +0200] "GET / HTTP/1.1" 500 22 "-" "HTTPie/2.0.0" 0 localhost:9090 - - "error while proxying after 118.044µs, route php with backend network fastcgi://0.0.0.0:9123: dialing failed false: gofast: failed creating client: dial tcp 0.0.0.0:9123: connect: connection refused"
```

Much better, isn't it?

### DIFF
--- a/logging/access.go
+++ b/logging/access.go
@@ -19,8 +19,8 @@ const (
 	// format:
 	// remote_host - - [date] "method uri protocol" status response_size "referer" "user_agent"
 	combinedLogFormat = commonLogFormat + ` "%s" "%s"`
-	// We add the duration in ms, a requested host and a flow id and audit log
-	accessLogFormat = combinedLogFormat + " %d %s %s %s\n"
+	// We add the duration in ms, a requested host and a flow id and audit log and a possible error
+	accessLogFormat = combinedLogFormat + ` %d %s %s %s "%s"` + "\n"
 )
 
 type accessLogFormatter struct {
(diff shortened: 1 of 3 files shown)


aryszka:
I think the proxy errors don't belong to the access logs. With many requests, the notion of "which" request is not very palpable either. The flowID should be a good approach to connect application logs with access logs.

If we want to make it easier to investigate the proxy errors, maybe we can do this the other way around, and add information about the current request to the proxy error.

aryszka:
:-1:

ruudk:
> If we want to make it easier to investigate the proxy errors, maybe we can do this the other way around, and add information about the current request to the proxy error.

I will do it like that :)

ruudk:
Can you re-open this one? I want to update this PR and it allows me to keep track of it easily.

ruudk:
Nevermind, created a new PR https://github.com/zalando/skipper/pull/1407

PR #1396, opened 2020-04-02
Improvements for FastCgi

ruudk:
Please review commit-by-commit.

### DIFF
--- a/proxy/fastcgi/fastcgi.go
+++ b/proxy/fastcgi/fastcgi.go
@@ -30,6 +30,7 @@ func NewRoundTripper(log logging.Logger, addr, filename string) (*RoundTripper, 
 		gofast.MapEndpoint(filename),
 		func(handler gofast.SessionHandler) gofast.SessionHandler {
 			return func(client gofast.Client, req *gofast.Request) (*gofast.ResponsePipe, error) {
+				req.Params["HTTP_HOST"] = req.Params["SERVER_NAME"]
 				req.Params["SERVER_SOFTWARE"] = "Skipper"
 
 				// Gofast sets this param to `fastcgi` which is not what the backend will expect.
--- a/proxy/fastcgi/fastcgi.go
+++ b/proxy/fastcgi/fastcgi.go
@@ -24,12 +24,23 @@ func NewRoundTripper(log logging.Logger, addr, filename string) (*RoundTripper, 
 		return nil, fmt.Errorf("gofast: failed creating client: %w", err)
 	}
 
-	handler := gofast.NewFileEndpoint(filename)(gofast.BasicSession)
+	chain := gofast.Chain(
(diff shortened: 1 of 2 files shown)


proxy/proxy.go
```diff
@@ -502,6 +502,7 @@ func mapRequest(r *http.Request, rt *routing.Route, host string, removeHopHeader
 		rr.Header = cloneHeader(r.Header)
 	}
 	rr.Host = host
+	rr.RemoteAddr = r.RemoteAddr
```
ruudk, line 505: @aryszka What do you think about this? If you don't agree, then I will try to do this only for `fastcgi` requests. It's important, because if `REMOTE_ADDR` is empty, the PHP backend cannot properly handle the other headers as well.

proxy/fastcgi/fastcgi.go
```diff
@@ -24,12 +24,27 @@ func NewRoundTripper(log logging.Logger, addr, filename string) (*RoundTripper,
 		return nil, fmt.Errorf("gofast: failed creating client: %w", err)
 	}
 
-	handler := gofast.NewFileEndpoint(filename)(gofast.BasicSession)
+	chain := gofast.Chain(
+		gofast.BasicParamsMap,
+		gofast.MapHeader,
+		gofast.MapEndpoint(filename),
+		func(inner gofast.SessionHandler) gofast.SessionHandler {
```
szuecs, line 31: inner -> handler

proxy/proxy.go
```diff
@@ -902,6 +903,10 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
 			return nil, &proxyError{err: err}
 		}
 
+		if _, _, err := net.SplitHostPort(req.Host); err != nil {
+			req.Host = req.Host + ":" + req.URL.Port()
```
szuecs, line 907: Do you need the port in the host header?

proxy/proxy.go
szuecs: I don’t think it will break something.

aryszka: for outgoing requests, this field is ignored, so it doesn't break anything, but TBH it would be nicer to solve it somehow for only the fastcgi, that relies on it and leave it blank for the rest of the logic. If possible.

proxy/proxy.go
aryszka: It is valid to put it there. Can you give some explanation about the reason as inline comment?

ruudk:
@szuecs @aryszka Applied feedback :)

aryszka:
:+1:

proxy/proxy.go
szuecs: The comment is showing the why, thanks @ruudk

szuecs:
:+1:

PR #1397, opened 2020-04-02
Use ReverseSourcePredicate in access log

ruudk:
Skipper doesn't reverse the host in the access log when it's running with `-reverse-source-predicate`.

I'm not running Skipper behind multiple proxies so I have no idea if this is actually an issue, but when I just look at the code, I think it just takes the `X-Forwarded-For` header's value, when printing it in the access log. 

## Todo

- [ ] Apply to https://github.com/zalando/skipper/pull/1407


szuecs:
The current code could print the whole x-forwarded-for headers, but it would need a manual test to confirm.

aryszka:
the current access log logs the the x-forwarded-for header when present, and the remote addr when not. We should not change it such that it limits to only the first host, as users, and log parsers may rely on it.

We can prepend/append the host from the remote addr, though.

PR #1398, opened 2020-04-02
fix streaming documents through the editor

aryszka:
fix streaming with large read buffers, where the bug is that EOF is returned too early


aryszka:
:+1:

gargravarr:
:+1:

PR #1403, opened 2020-04-03
Allow Traffic to receive values >= 1

ruiaraujo:
If a user adds a filter with Traffic >= 1 it will drop the route due to invalid parameters. Specially when setting it to 1, the current implementation seems to have a consequence too severe.

### DIFF
--- a/predicates/traffic/traffic.go
+++ b/predicates/traffic/traffic.go
@@ -102,7 +102,7 @@ func (s *spec) Create(args []interface{}) (routing.Predicate, error) {
 
 	p := &predicate{}
 
-	if c, ok := args[0].(float64); ok && 0.0 <= c && c < 1.0 {
+	if c, ok := args[0].(float64); ok && 0.0 <= c {
 		p.chance = c
 	} else {
 		return nil, predicates.ErrInvalidPredicateParameters
--- a/predicates/traffic/traffic_test.go
+++ b/predicates/traffic/traffic_test.go
@@ -52,10 +52,10 @@ func TestCreate(t *testing.T) {
 		predicate{chance: .3},
 		false,
 	}, {
-		"wrong chance, bigger than 1",
+		"bigger than 1 is acceptable",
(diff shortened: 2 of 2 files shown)


predicates/traffic/traffic.go
```diff
@@ -102,7 +102,7 @@ func (s *spec) Create(args []interface{}) (routing.Predicate, error) {
 
 	p := &predicate{}
 
-	if c, ok := args[0].(float64); ok && 0.0 <= c && c < 1.0 {
+	if c, ok := args[0].(float64); ok && 0.0 <= c {
```
addityasingh, line 105: We should default the chance to `1.0` in case `c >= 1` to ensure the chance is at the maximum since `rand.Float64() is in the range [0.0, 1.0)` 
https://golang.org/pkg/math/rand/#Float64

NorthFury:
There are tests failing because of this changed and tests missing to account for this change.

The documentation should also be updated both in the code file and in the predicates.md documentation file.

ruudk:
Please rebase with master.

ruiaraujo:
I forgot to mention that I wanted some feedback if the change makes sense from @aryszka or @szuecs . 

If yes, I would finish it. This changes the approach from `fail fast` to `user resilient`, where we can at most add some logs for education.

szuecs:
@ruiaraujo I think this can break functionality and can be unsafe.

aryszka:
I think having Traffic(1) is like not having it, and so better fail early, because there certainly something in the configuration that doesn't make sense.

aryszka:
@szuecs I don't find it as dangerous, though. Can you give some more details about what you meant?

szuecs:
@aryszka, if we don’t apply a catchall route.
We can also do: if >=1, then not apply the predicate.

ruiaraujo:
Discuss internally, fail fast is important for non human consumers.
For humans, validation should be in place to ensure proper usage of the filter.

PR #1405, opened 2020-04-04
V1 > Remove deprecated Proxy.New

ruudk:
Now that `Proxy.New` is removed, can we rename `Proxy.WithParams` to `Proxy.New`? I changed this in 3e353e1. We can remove the commit if this is not what you want.


aryszka:
could we also rename Params to Options and use just flags instead of the current Options?

ruudk:
@aryszka Like this? :)

aryszka:
:+1:

PR #1406, opened 2020-04-04
V1 > Use Weight predicate to give Traffic more priority

ruudk:
This is nicer than the `True() && True() && True()` hack.


aryszka:
:+1:

PR #1407, opened 2020-04-04
Log request info on proxy error

ruudk:
This makes it easier to debug why a proxy request failed.

## How does it look?
```
[APP]ERRO[0040] error while proxying after 225.477µs, route php with backend network fastcgi://0.0.0.0:55212, flow id not set, status code 500: dialing failed false: gofast: failed creating client: dial tcp 0.0.0.0:55212: connect: connection refused, remote host: 1.2.3.4, request: "GET /hello HTTP/1.1", user agent: "HTTPie/2.0.0" 
```

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"runtime"
 	"strconv"
+	"strings"
 	"time"
 
 	ot "github.com/opentracing/opentracing-go"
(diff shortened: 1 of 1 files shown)


proxy/proxy.go
```diff
@@ -1243,22 +1244,89 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
...
+func remoteAddr(r *http.Request) string {
+	ff := r.Header.Get("X-Forwarded-For")
+	if ff != "" {
+		return ff
+	}
+
+	return r.RemoteAddr
+}
+
+func remoteHost(r *http.Request) string {
```
aryszka, line 1325: aren't the function from this net package fitting here? https://godoc.org/github.com/zalando/skipper/net

aryszka:
The message looks generally good.

i know that it's not part of this PR, but the "flow id not set" fragment may appear to somebody less familiar with the app, as it is not set as a consequence of a problem. I can envision repeated support questions. Can we use two formats, one for flow id is set and one for when it's not?

proxy/proxy.go
ruudk: For now, I copied it, to have the same behaviour as accesslog package. But I want to make this better in https://github.com/zalando/skipper/pull/1397. So I'd rather merge this as is, and fix it later :)

ruudk:
@aryszka Applied feedback.

aryszka:
:+1:

szuecs:
:+1:

PR #1424, opened 2020-05-04
fix measuring the connection count in TCP-LIFO

aryszka:
fix measuring the connection count in cases when the server sides closes connections multiple times

### DIFF
--- a/queuelistener/listener.go
+++ b/queuelistener/listener.go
@@ -101,12 +101,15 @@ func (c *connection) SetReadDeadline(t time.Time) error  { return c.net.SetReadD
 func (c *connection) SetWriteDeadline(t time.Time) error { return c.net.SetWriteDeadline(t) }
 
 func (c *connection) Close() error {
-	select {
-	case c.release <- token:
-	case <-c.quit:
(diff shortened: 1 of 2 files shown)


mikkeloscar:
@aryszka just to make sure I fully understand this fix. Basically before we could risk executing `case c.release <- token:` multiple times which translates to `concurrency--` on the other end: https://github.com/zalando/skipper/blob/5be4371b21b6285021857a4ccdd58d6f66dcd8d7/queuelistener/listener.go#L341

And thus we would be able to accept more connections than intended because the `concurrency` value was lower than expected.

Is this the right understanding?

aryszka:
@mikkeloscar yes, it's exactly what happened.

I remember crawling through skipper and stdlib/net code for possibles cases where this can happen. Apparently, we were too confident that it won't happen. (Saving that synchronization step that is done now with c.once was a pretty good example of premature optimization.)

mikkeloscar:
:+1:

mikkeloscar:
:+1:

aryszka:
:+1:

PR #1425, opened 2020-05-06
Traffic(): allow weight of 1.0

aermakov-zalando:
This is not optimal (because the user should just delete the other routes), but in some cases it makes it a lot easier to define the routes. An example is defining a route where the weight passed to the `Traffic()` predicate is parameterised and then changing it between `0.0` and `1.0`. The users would expect `1.0` to work, but currently it doesn't.

Update the documentation and the comments to mention that both values are now allowed (`0.0` used to work but was documented as not working).


aryszka:
:+1:

aryszka:
:+1:

aermakov-zalando:
👍

PR #1429, opened 2020-05-27
use cryptographic secure random to generate random secrets

szuecs:
use cryptographic secure random and do not have a rand source that is created at compile time


szuecs:
I will move it to secrets module

szuecs:
this was only used to add rand data for encrypting a not to static payload so it's fine
The use and idea should be documented

PR #1432, opened 2020-05-28
flexible upstream headers

universam1:
Similar to [Keycloak Gatekeeper Custom Claim Headers](https://github.com/keycloak/keycloak-documentation/blob/master/securing_apps/topics/oidc/keycloak-gatekeeper.adoc#custom-claim-headers) we need Skipper to inject headers to upstream endpoints to provide user identity.

As of now the "Skipper-Oidc-Info" header provides this information but in a custom, not ready to use format which would require to adapt each backend application.

Like Gatekeeper this information should be possible to be specified per use case on the ingress annotation level.

This implementation uses the `GJSON` syntax already used for other purposes to define and parse the information that is requested and inject as the defined header name.

For backward compatibilty, this additional field is optional, falling back to outputting the "Skipper-Oidc-Info" as before.

### DIFF
--- a/filters/auth/oidc.go
+++ b/filters/auth/oidc.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/coreos/go-oidc"
 	log "github.com/sirupsen/logrus"
+	"github.com/tidwall/gjson"
 	"github.com/zalando/skipper/filters"
 	"github.com/zalando/skipper/secrets"
 	"golang.org/x/oauth2"
@@ -55,6 +56,7 @@ type (
 		encrypter       secrets.Encryption
 		authCodeOptions []oauth2.AuthCodeOption
 		compressor      cookieCompression
+		upstreamHeaders map[string]string
 	}
 
 	tokenContainer struct {
@@ -173,7 +175,7 @@ func (s *tokenOidcSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 	}
 
 	f.authCodeOptions = make([]oauth2.AuthCodeOption, 0)
-	if len(sargs) > 6 {
+	if len(sargs) > 6 && sargs[6] != "" {
 		extraParameters := strings.Split(sargs[6], " ")
 
 		for _, p := range extraParameters {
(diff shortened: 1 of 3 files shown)


universam1:
@szuecs FYI

webframp:
Nice! This would enable some simplifications for us.

szuecs:
Thanks will check it, but can you fix dco, thanks

filters/auth/oidc.go
```diff
@@ -186,6 +188,21 @@ func (s *tokenOidcSpec) CreateFilter(args []interface{}) (filters.Filter, error)
...
 	log.Debugf("Auth Code Options: %v", f.authCodeOptions)
+
+	// inject additional headers from the access token for upstream applications
+	if len(sargs) > 7 && sargs[7] != "" {
+		f.upstreamHeaders = make(map[string]string)
+
+		for _, header := range strings.Split(sargs[7], " ") {
+			sl := strings.SplitN(header, ":", 2)
+			if len(sl) != 2 || sl[0] == "" || sl[1] == "" {
+				return nil, fmt.Errorf("%v: malformatted filter for upstream headers %s", filters.ErrInvalidFilterParameters, sl)
```
szuecs, line 199: Do you know if you do %w instead of %v, the error.Is() function can check if it’s a ErrInvalidFilterParameters?
I always wanted to move the errors in skipper from %v to %w and maybe you already have experience with it.

filters/auth/oidc.go
```diff
@@ -613,15 +630,42 @@ func (f *tokenOidcFilter) Request(ctx filters.FilterContext) {
...
 	}
-	// saving token info for chained filter
-	ctx.StateBag()[oidcClaimsCacheKey] = container
-	ctx.Request().Header.Add(oidcInfoHeader, string(oidcInfoJson))
+}
+
+func (f *tokenOidcFilter) setHeaders(ctx filters.FilterContext, container tokenContainer) (err error) {
+	oidcInfoJson, err := json.Marshal(container)
+	if err != nil || !gjson.ValidBytes(oidcInfoJson) {
+		return fmt.Errorf("Failed to serialize OIDC token info: %w", err)
```
szuecs, line 648: Nitpick: errors start always with lowercase, logs with uppercase character

filters/auth/oidc.go
```diff
@@ -613,15 +630,42 @@ func (f *tokenOidcFilter) Request(ctx filters.FilterContext) {
...
+		return
+	}
+
+	parsed := gjson.ParseBytes(oidcInfoJson)
+
+	for key, query := range f.upstreamHeaders {
+		match := parsed.Get(query)
+		log.Debugf("header: %s results: %s", query, match.String())
+		if !match.Exists() {
+			log.Errorf("lookup failed for upstream header '%s'", query)
```
szuecs, line 663: Nitpick as above but now upper case

filters/auth/oidc.go
universam1: Great point! Actually I also work on migrating all `fmt.Errorf ...%v to %w` syntax, it is cost-free.
`error.Is()` is one new possibility, but actually the wrapping includes the underlying error type which can be retrieve via normal switch case type statements again. Cannot promote its use case enough as it allows way more safer error tests.

filters/auth/oidc.go
universam1: thanks :+1:

filters/auth/oidc.go
universam1: thanks :+1:

universam1:
> Thanks will check it, but can you fix dco, thanks

fixed, thanks for review  @szuecs

szuecs:
:+1:

aryszka:
:+1:

PR #1433, opened 2020-06-01
Deduplicate OIDC cookies

universam1:
For each ingress Skipper will hash a new cookie. If several ingresses for the same domain are defined, either by several hosts or sub-paths, it will accumulate cookies on the client browser which can lead to side effects like header limits exceeded or heavy request payloads.

Example which generates several cookies when the annotations are identical besides of the `RedirectURL` parameter. This as of now results into 3 cookies which would not be necessary.:

```
path: "/" --> ingress: portal
path: "/subpathA" --> ingress: serviceA
path: "/subpathB" --> ingress: serviceB
```

There might be better approaches to tackle this issue by specifying the actual path of its ingress on the cookie, however this information is apparently not available at the filter level for now.

This simple approach 
1. removes the `CallbackURL` out of the hashing, which keeps the idP, client-ID, client-secret, scopes, claims and optional parameters as a sufficient source of integrity signature.
1. Expects that  `CallbackURL` is not integral part of the JWT token distinction as otherwise the idP connection creds would need to change anyway

It will allow Skipper to reuse the cookie where applicable and deduplicate where avoidable.

### DIFF
--- a/filters/auth/oidc.go
+++ b/filters/auth/oidc.go
@@ -37,6 +37,22 @@ const (
 	cookieMaxSize       = 4093 // common cookie size limit http://browsercookielimits.squawky.net/
 )
 
+// Filter parameter:
+//
+//  oauthOidc...("https://oidc-provider.example.com", "client_id", "client_secret",
+//               "http://target.example.com/subpath/callback", "email profile", "name email picture",
+//               "parameter=value", "X-Auth-Authorization:claims.email")
+const (
+	paramIdpURL int = iota
+	paramClientID
+	paramClientSecret
+	paramCallbackURL
+	paramScopes
+	paramClaims
+	paramAuthCodeOpts
+	paramUpstrHeaders
+)
+
 type (
 	tokenOidcSpec struct {
 		typ             roleCheckType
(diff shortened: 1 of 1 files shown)


dhohengassner:
@universam1  nice idea - we have seen this issue as well

universam1:
@aryszka @szuecs related to slack discussion https://gophers.slack.com/archives/C82Q5JNH5/p1591001354207800

szuecs:
Thanks @universam1 I will check it tomorrow.

@dhohengassner please share things in an issue if you see miss behavior like this, such that we can think about a fix. 😊

szuecs:
:+1:

filters/auth/oidc.go
```diff
@@ -37,6 +37,22 @@ const (
 	cookieMaxSize       = 4093 // common cookie size limit http://browsercookielimits.squawky.net/
 )
 
+// Filter parameter:
+//
+//  oauthOidc...("https://oidc-provider.example.com", "client_id", "client_secret",
+//               "http://target.example.com/subpath/callback", "email profile", "name email picture",
+//               "parameter=value", "X-Auth-Authorization:claims.email")
+const (
+	paramIdpURL int = iota
```
szuecs, line 46: Nice way of making the code more readable!

aryszka:
:+1:

Issue #1428, opened 2020-05-27
The flag -ratelimits doesn't accept cluster-based strategies


rafaellima:
**Is your feature request related to a problem? Please describe.**
I want to enable global rate-limiting (using Redis ring shards) by using the command line `-ratelimits` flag. Although cluster-based options are available in the implementation[1] and documentation[2], they can't be set via command line because there is a validation that accepts only one of the following options: (client/service/disabled)[3]. 

**Describe the solution you would like**
Add the missing implementations (`ClusterClientRatelimit` and `ClusterServiceRatelimit`) as valid options.

**Would you like to work on it?**
Yes.

1. https://github.com/zalando/skipper/blob/master/ratelimit/ratelimit.go#L121
2. https://godoc.org/github.com/zalando/skipper/ratelimit#hdr-Settings___Type
3. https://github.com/zalando/skipper/blob/master/config/ratelimiterflags.go#L47

szuecs:
Sounds great!

PR #1434, opened 2020-06-01
Fix -ratelimit flag to accept cluster based options

rafaellima:
Relates to #1428 

The cluster-based rate-limiting implementations are available but couldn't be passed in via command-line flags.

This PR enables `clusterClient` and `clusterService` as valid types for rate-limiting as well as the `group` flag.

### DIFF
--- a/config/ratelimiterflags.go
+++ b/config/ratelimiterflags.go
@@ -11,11 +11,12 @@ import (
 	"github.com/zalando/skipper/ratelimit"
 )
 
-const ratelimitUsage = `set global rate limit settings, e.g. -ratelimit type=local,max-hits=20,time-window=60s
+const ratelimitUsage = `set global rate limit settings, e.g. -ratelimit type=client,max-hits=20,time-window=60s
 	possible ratelimit properties:
-	type: local/service/disabled (defaults to disabled)
+	type: client/service/clusterClient/clusterService/disabled (defaults to disabled)
 	max-hits: the number of hits a ratelimiter can get
 	time-window: the duration of the sliding window for the rate limiter
+	group: defines the ratelimit group, which can be the same for different routes.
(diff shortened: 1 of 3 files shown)


config/ratelimiterflags.go
```diff
@@ -11,11 +11,12 @@ import (
...
 )
 
-const ratelimitUsage = `set global rate limit settings, e.g. -ratelimit type=local,max-hits=20,time-window=60s
+const ratelimitUsage = `set global rate limit settings, e.g. -ratelimit type=client,max-hits=20,time-window=60s
 	possible ratelimit properties:
-	type: local/service/disabled (defaults to disabled)
+	type: client/service/clusterClient/clusterService/disabled (defaults to disabled)
 	max-hits: the number of hits a ratelimiter can get
 	time-window: the duration of the sliding window for the rate limiter
+	group: defines the ratelimit group, which can be the same for different routes.
```
szuecs, line 19: Only valid for cluster ratelimit

aryszka:
thanks for the PR, @rafaellima . It looks good to me, except for we should mention the limitation of the group attribute as @szuecs pointed out.

rafaellima:
Hi, thanks for the comments.

>A question would be how we can enforce that the required settings for cluster Ratelimits are set, for example redis URLs.
What do you think?

From what I can see in the configuration, there is no validation based on key combinations atm. One possible solution would be to have a function in the config that either check both (rate limit type and Redis URLs) or calls a function that does this. That could be called in [cmd/skipper/main.go after the parsing](https://github.com/zalando/skipper/blob/be390155f94dc290a4fb5ccc0e80a099495b0a30/cmd/skipper/main.go#L30)

szuecs:
@rafaellima you are right combinations are not checked at cli or config, so let’s merge it.

szuecs:
:+1:

aryszka:
:+1:

PR #1435, opened 2020-06-02
OIDC-introspection concurrency crashes

universam1:
Turns out the `gjson` internal map access for "modifiers" is not thread safe:
https://github.com/tidwall/gjson/issues/173

It accesses a map unsyncronized causing application crashes during RW operations:

```
fatal error: concurrent map writes
goroutine 4555 [running]:
runtime.throw(0x1041950, 0x15)
	/usr/local/go/src/runtime/panic.go:1116 +0x72 fp=0xc001dc5338 sp=0xc001dc5308 pc=0x434b42
runtime.mapassign_faststr(0xecda00, 0xc000118f60, 0x102fe84, 0x1, 0x0)
	/usr/local/go/src/runtime/map_faststr.go:291 +0x3de fp=0xc001dc53a0 sp=0xc001dc5338 pc=0x41375e
github.com/tidwall/gjson.AddModifier(...)
	/go/pkg/mod/github.com/tidwall/gjson@v1.4.0/gjson.go:2612
```

This workaround uses `Mutex` semaphores in order to achieve thread safety, that might be slower however as an improvement the write operation is only conditional now which makes it rare.

### DIFF
--- a/filters/auth/oidc_introspection.go
+++ b/filters/auth/oidc_introspection.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"sync"
 
 	"github.com/tidwall/gjson"
 
@@ -15,6 +16,8 @@ const (
 	OidcClaimsQueryName = "oidcClaimsQuery"
 	oidcClaimsCacheKey  = "oidcclaimscachekey"
 )
+
+var gjsonModifierMutex = sync.RWMutex{}
 
 type (
 	oidcIntrospectionSpec struct {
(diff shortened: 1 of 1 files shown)


universam1:
@aryszka @szuecs

filters/auth/oidc_introspection.go
```diff
@@ -120,10 +123,23 @@ func (filter *oidcIntrospectionFilter) Request(ctx filters.FilterContext) {
...
+
+	gjsonModifierMutex.RLock()
+	// method is not thread safe
+	modExists := gjson.ModifierExists("_", gjsonThisModifier)
+	gjsonModifierMutex.RUnlock()
+
+	if !modExists {
+		gjsonModifierMutex.Lock()
+		// method is not thread safe
+		gjson.AddModifier("_", gjsonThisModifier)
```
aryszka, line 140: Isn't the problem rather that we call AddModifier in every call to validateClaimsQuery? Would it work if we simply just moved gjson.AddModifier to the package init function. If yes, I'd rather prefer that approach over syncing.

(In theory, the ideal would be if the auth package could use its own independent instance of the syntax, because then there is no danger that another package, own or imported, would try to use the same modifier name "_". But I see how this is currently not possible with the gjson library.)

universam1: Very good point - I think you are on something here!
Makes sense to move this part into the init section of the filter rather then the execution. Let me try that.

Agreed, the lib should actually support instancing in order to provide a clear boundary, I've opened an issue there in this regard

universam1: @aryszka your suggestion applied :+1:

szuecs:
:+1:

aryszka:
:+1:

Issue #1440, opened 2020-06-08
tee() routes for advanced shadow traffic


szuecs:
**Is your feature request related to a problem? Please describe.**

Shadow traffic is used more and more and sometimes you need additional features, which can create complex routes for the user.
People want for example ratelimit the backend service that is targeted by `tee()`.

**Describe the solution you would like**

A solution could be to add a Predicate `Tee`, that matches a statebag entry (or a special header) with value `"A"` in the `rTee` route. We also have to add a filter `teeLoopback` that will set the value of our special statebag entry and execute an internal `<loopback>` backend to reach the `rTee` route.

Example:
```
rTee: Tee("A") -> "https://teeed-backend/";
r0: Host("foo.example.org") -> teeLoopback("A") -> "https://my-backend"
```

This will also close https://github.com/zalando/skipper/issues/997

szuecs:
This was done

PR #1443, opened 2020-06-10
feat: add Tee predicate and teeLoopback for advanced shadow traffic

topicus:
Ref: https://github.com/zalando/skipper/issues/1440

I've created a flow diagram to make the implementation more understandable. 

skipper-tee-filter


filters/tee/tee.go
```diff
@@ -144,7 +144,10 @@ func (r *tee) Response(filters.FilterContext) {}
 // Request is copied and then modified to adopt changes in new backend
 func (r *tee) Request(fc filters.FilterContext) {
 	req := fc.Request()
-	copyOfRequest, tr, err := cloneRequest(r, req)
+	u := cloneURL(r, req)
+	copyOfRequest, tr, err := cloneRequest(u, req)
```
szuecs, line 148: can you rename `tr` to `body`?
`tr` is most often used for net/http.Transport

filters/tee/teeloopback.go
```diff
@@ -0,0 +1,65 @@
...
+type teeLoopbackFilter struct {
+	teeKey string
+}
+
+func (t *teeLoopbackSpec) Name() string {
+	return FilterName
+}
+
+func (t *teeLoopbackSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) < 1 {
```
szuecs, line 23: `<` -> `!=`

filters/tee/teeloopback.go
```diff
@@ -0,0 +1,65 @@
...
+func NewTeeLoopback() filters.Spec {
+	return &teeLoopbackSpec{}
+}
+
+func (f *teeLoopbackFilter) Request(ctx filters.FilterContext) {
+	origRequest := ctx.Request()
+	// prevent the loopback to be executed indefinitely
+	teeRegistry, registryExists := ctx.Request().Context().Value(teePredicate.ContextTeeKey).(map[string]bool)
+	if !registryExists {
+		teeRegistry = map[string]bool{}
```
szuecs, line 44: To be more efficient for memory usage please use this as a set, which saves a bool from being stored (empty struct is a zero byte `struct{}` to get an instance of empty struct you use `struct{}{}`:
```
set := make(map[string]struct{})
set["foo"] = struct{}{}
if _, ok := set["foo"]; ok { .. } // has the key "foo"
if _, ok := set["bar"]; !ok { .. } // has no key "bar"
```

filters/tee/teeloopback.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	if _, ok := teeRegistry[f.teeKey]; ok {
+		return
+	}
+	u := new(url.URL)
+	*u = *origRequest.URL
+	cr, body, err := cloneRequest(u, origRequest)
+	if err != nil {
+		log.Error("teeloopback: failed to clone request")
```
szuecs, line 53: log.Errorf("teeloopback: failed to clone request: %v", err)

filters/tee/teeloopback_test.go
```diff
@@ -0,0 +1,113 @@
...
+	go waitForAll([]backendtest.Done{
+		shadow.Done,
+		split.Done,
+	}, allDone)
+	select {
+	case <-allDone:
+	case <-time.After(time.Second * 1):
+		t.Error("teeloopback: time exceeded while waiting for requests")
+	}
+	if shadow.GetServedRequests() != 0 && split.GetServedRequests() != 2 {
```
szuecs, line 110: `&&` -> `||`, because we want to have both cases to be true and if one is false we want to error

filters/tee/teeloopback_test.go
```diff
@@ -0,0 +1,113 @@
...
+		shadow.Done,
+		split.Done,
+		original.Done,
+	}, allDone)
+	select {
+	case <-allDone:
+	case <-time.After(time.Second * 1):
+		t.Error("teeloopback: time exceeded while waiting for requests")
+	}
+	if shadow.GetServedRequests() != 1 && split.GetServedRequests() != 1 {
```
szuecs, line 69: `&&` -> `||`, because we want to have both cases to be true and if one is false we want to error

predicates/tee/tee.go
```diff
@@ -0,0 +1,46 @@
...
+	}, nil
+}
+
+func (p *predicate) Match(r *http.Request) bool {
+	ctx := r.Context()
+	teeRegistry, ok := ctx.Value(ContextTeeKey).(map[string]bool)
+	if !ok {
+		return false
+	}
+	if _, ok:=teeRegistry[p.key]; ok {
```
szuecs, line 42: We could also return the value of ok:
```
_, ok:=teeRegistry[p.key]
return ok
```

proxy/backendtest/backendrecorder.go
```diff
@@ -0,0 +1,94 @@
...
+	mutex            sync.RWMutex
+	expectedRequests int
+	pendingRequests  int
+	Done             Done
+}
+
+func (rec *backendRecorderHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		log.Error("backendrecorder: error while reading request body")
```
szuecs, line 31: log.Errorf("backendrecorder: error while reading request body: %v". err)

proxy/backendtest/backendrecorder.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	// return request body in the response
+	_, err = w.Write(body)
+	if err != nil {
+		log.Error("backendrecorder: error while writing the response body")
```
szuecs, line 36: log.Errorf("backendrecorder: error while writing the response body: %v", err)

filters/tee/teeloopback.go
```diff
@@ -0,0 +1,65 @@
...
+		return
+	}
+	origRequest.Body = body
+	teeRegistry[f.teeKey] = true
+	newReqContext := context.WithValue(cr.Context(), teePredicate.ContextTeeKey, teeRegistry)
+	newReqWithContext := cr.WithContext(newReqContext)
+	cc, _ := ctx.SplitWithRequest(newReqWithContext)
+	go cc.Loopback()
```
szuecs, line 61: Should we add a timeout, if not we could have a piling up of goroutines

szuecs:
Besides https://github.com/zalando/skipper/pull/1443/files#r440690522, I found only small things to change.
The cloned request won't show up in access logs as far as I understand, which is IMO fine.

filters/filters.go
```diff
@@ -96,6 +96,12 @@ type FilterContext interface {
 
 	// Allow filters to create their own spans
 	ParentSpan() opentracing.Span
+
+	// Returns a clone of the FilterContext including a brand new request object.
+	SplitWithRequest(cr *http.Request) (FilterContext, error)
```
aryszka, line 101: I thought the request cloning would happen in the context. What's the reason for not doing it there?

filters/tee/teeloopback.go
```diff
@@ -0,0 +1,65 @@
...
+	}, nil
+}
+
+func NewTeeLoopback() filters.Spec {
+	return &teeLoopbackSpec{}
+}
+
+func (f *teeLoopbackFilter) Request(ctx filters.FilterContext) {
+	origRequest := ctx.Request()
+	// prevent the loopback to be executed indefinitely
```
aryszka, line 41: this protection already exists with a configurable limit in the proxy loopback implementation, with a configurable counter. I think we should reuse it.

filters/tee/teeloopback.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	u := new(url.URL)
+	*u = *origRequest.URL
+	cr, body, err := cloneRequest(u, origRequest)
+	if err != nil {
+		log.Error("teeloopback: failed to clone request")
+		return
+	}
+	origRequest.Body = body
+	teeRegistry[f.teeKey] = true
```
aryszka, line 57: i think there is a race condition this way, because the teeRegistry is shared across multiple goroutines

filters/tee/teeloopback.go
```diff
@@ -0,0 +1,65 @@
...
+	newReqContext := context.WithValue(cr.Context(), teePredicate.ContextTeeKey, teeRegistry)
+	newReqWithContext := cr.WithContext(newReqContext)
+	cc, _ := ctx.SplitWithRequest(newReqWithContext)
```
aryszka, line 60: let's not ignore this error. If there is an error, then let's not make a loopback. OTOH, I think the split should be possible to implement without an error

predicates/tee/tee.go
```diff
@@ -0,0 +1,46 @@
...
+	if len(args) != 1 {
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+	return &predicate{
+		key: args[0].(string),
+	}, nil
+}
+
+func (p *predicate) Match(r *http.Request) bool {
+	ctx := r.Context()
```
aryszka, line 37: :+1: I like the context is used instead of a header!

filters/tee/teeloopback_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/zalando/skipper/proxy/proxytest"
+	"github.com/zalando/skipper/routing"
+	"net/http"
+	"sync"
+	"testing"
+	"time"
+)
+
+func waitForAll(channels []backendtest.Done, done chan struct{}) {
+	var wg sync.WaitGroup
```
aryszka, line 19: this is possible without many goroutines and without the waitgroup:

```
func waitForAll(channels []backendtest.Done, done chan struct{}) {
	for _, ch := range channels {
		<-ch
	}

	close(done)
}
```

filters/filters.go
topicus: To augment the request context from outside. I tried adding SplitWithOptions and SplitOptions to support overrides but I can't use SplitOptions struct from filters because it creates a circular import dependency.

filters/tee/teeloopback_test.go
topicus: Nice!

filters/tee/teeloopback.go
topicus: You are right, I will use an RW lock for this.

filters/tee/teeloopback.go
topicus: I think this is already handled in p.do. I can provide a timeout to make it more robust but I think the resilience mechanisms should be inside the p.do method instead.

filters/tee/teeloopback.go
topicus: This is no longer relevant since we are using the request for sharing.

filters/tee/teeloopback.go
topicus: I have tested this works properly I need to figure out how to write a test for this.

filters/filters.go
topicus: Rolled back to the Split signature.

filters/tee/teeloopback.go
```diff
@@ -0,0 +1,48 @@
...
+func (t *teeLoopbackSpec) Name() string {
+	return FilterName
+}
+
+func (t *teeLoopbackSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) != 1 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+	teeKey, ok := args[0].(string)
+	if !ok || teeKey == "" {
```
aryszka, line 25: Not sure why we wouldn't accept "" or no params and then use "". If somebody just playing with the routing, it may make sense.

If not however, then this can be simplified:

```
teeKey, _ := args[0].(string)
if teeKey == "" {
}
```

filters/tee/teeloopback.go
```diff
@@ -0,0 +1,48 @@
+package tee
+
+import (
+	log "github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/filters"
+	teePredicate "github.com/zalando/skipper/predicates/tee"
```
aryszka, line 6: Please use all lowercase for package identifiers: `teepredicate`. See the stdlib (strconv, maphash, etc...)

filters/tee/teeloopback.go
aryszka: I agree, it's handled in the Proxy.do().

predicates/tee/tee.go
```diff
@@ -0,0 +1,37 @@
...
+func New() routing.PredicateSpec { return &spec{} }
+
+func (s *spec) Name() string { return PredicateName }
+
+func (s *spec) Create(args []interface{}) (routing.Predicate, error) {
+	if len(args) != 1 {
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+	return &predicate{
+		key: args[0].(string),
```
aryszka, line 30: please check if it's a string, otherwise it may panic. Similar to the teeLoopback filter args.

proxy/context.go
```diff
@@ -233,10 +237,104 @@ func (c *context) clone() *context {
 	return &cc
 }
 
+func (c *context) wasExecuted()bool {
```
aryszka, line 240: is there any missing formatting? Wondering if there suppose to be a whitespace before the bool return type.

proxy/context.go
```diff
@@ -35,14 +36,19 @@ type context struct {
...
 	metrics              *filterMetrics
 	tracer               opentracing.Tracer
 	proxySpan            opentracing.Span
 	parentSpan           opentracing.Span
+	proxy                *Proxy
+	routeLookup          *routing.RouteLookup
+}
 
-	routeLookup *routing.RouteLookup
+type teeTie struct {
```
aryszka, line 49: Can we please move the teeTie type to its own file? There's quite some concept bundled already in context.go.

filters/filters.go
```diff
@@ -96,6 +96,12 @@ type FilterContext interface {
 
 	// Allow filters to create their own spans
 	ParentSpan() opentracing.Span
+
+	// Returns a clone of the FilterContext including a brand new request object.
```
aryszka, line 100: Can we please mention here that the stream of the request body is shared?

proxy/context.go
```diff
@@ -233,10 +237,104 @@ func (c *context) clone() *context {
...
+	h := make(http.Header)
+	for k, v := range req.Header {
+		h[k] = v
+	}
+
+	var teeBody io.ReadCloser
+	mainBody := req.Body
+
+	// see proxy.go:231
+	if req.ContentLength != 0 {
```
aryszka, line 282: we cannot rely on ContentLength, because it is not always set, like in case of chunked content in the request. I think the teeTie body should work for 0 length body, too, so we could just skip the condition.

proxy/context.go
```diff
@@ -233,10 +237,104 @@ func (c *context) clone() *context {
...
+func cloneRequest(u *url.URL, req *http.Request) (*http.Request, io.ReadCloser, error) {
+	h := make(http.Header)
+	for k, v := range req.Header {
+		h[k] = v
+	}
+
+	var teeBody io.ReadCloser
+	mainBody := req.Body
+
+	// see proxy.go:231
```
aryszka, line 281: Please don't reference line numbers, the code can change over time. Better reference concrete function names.

proxy/context.go
```diff
@@ -233,10 +237,104 @@ func (c *context) clone() *context {
...
+func (c *context) Split() (filters.FilterContext, error) {
+	originalRequest := c.Request()
+	cc := c.clone()
+	c.stateBag = map[string]interface{}{}
+	cc.metrics = &filterMetrics{
+		prefix: cc.metrics.prefix,
+		impl:   cc.proxy.metrics,
+	}
+	u := new(url.URL)
+	*u = *originalRequest.URL
```
aryszka, line 308: I think we need to set the Host field of `u`, to originalRequest.Host. See the docs of the URL field: https://godoc.org/net/http#Request and also the host handling in the old tee implementation.

proxy/context.go
```diff
@@ -233,10 +237,104 @@ func (c *context) clone() *context {
...
+		}
+		err := c.response.Body.Close()
+		if err != nil {
+			c.proxy.log.Error("context: error during closing the response body", err)
+		}
+	}
+	if c.proxySpan != nil {
+		c.proxySpan.Finish()
+	}
+	if err != nil {
```
aryszka, line 333: don't log the error if it was handled:

```
	perr, ok := err.(*proxyError)
	if ok && perr.handled {
		return
	}
```

proxy/context.go
```diff
@@ -233,10 +237,104 @@ func (c *context) clone() *context {
...
+		log.Error("context: failed to execute loopback request", err)
```
aryszka, line 334: we have some logging enhancements in the Proxy.errorResponse function. Some of it should also be applied here (flow ID. It's fine with me in a second pass, though.

proxy/context.go
```diff
@@ -233,10 +237,104 @@ func (c *context) clone() *context {
...
+	}
+
+	clone.Header = h
+	clone.ContentLength = req.ContentLength
+
+	return clone, mainBody, nil
+}
+
+func (c *context) Split() (filters.FilterContext, error) {
+	originalRequest := c.Request()
```
aryszka, line 300: we should preserve the opentracing server span, if available. Sg like (pseudo):

```
serverSpan := opentracing.SpanFromContext(originalRequest.Context())
cr = cr.WithContext(opentracing.ContextWithSpan(cr.Context(), otspan))
```

this way the cloned request's stdlib context and the original request's stdlib is not tied together, but both of them have the same server span, and the opentracing graph will show the tee request nicely.

proxy/context.go
```diff
@@ -233,10 +237,104 @@ func (c *context) clone() *context {
...
+	cc := c.clone()
```
aryszka, line 301: we should replace the context.responseWriter field with a dummy writer, because there can be legacy filters that use it directly

proxy/context.go
```diff
@@ -233,10 +237,104 @@ func (c *context) clone() *context {
...
+		return nil, nil, err
+	}
+
+	clone.Header = h
+	clone.ContentLength = req.ContentLength
+
+	return clone, mainBody, nil
+}
+
+func (c *context) Split() (filters.FilterContext, error) {
```
aryszka, line 299: there is the experimentalUpgrade feature, that I think we should not allow when doing Splitting. It doesn't make sense and it's dangerous, potentially causing hard to debug leaks. Let's return an error here when we detect that it's an upgrade request && the experimental upgrade is enabled. Conditions: `c.proxy.experimentalUpgrade && isUpgradeRequest(req)`

filters/tee/teeloopback.go
topicus: Mark as resolved based on @aryszka comment https://github.com/zalando/skipper/pull/1443#discussion_r443418863

filters/tee/teeloopback.go
topicus: For this one, I prefer to require a non-empty key. If we default to "" then it is more likely to end up in a scenario of circular unwanted loopbacks. WDYT?

proxy/context.go
topicus: Missign formatting. Done.

proxy/context.go
topicus: Removed

proxy/context.go
topicus: I added the snippet right above.

filters/filters.go
topicus: I added more context here.

proxy/context.go
topicus: Moved.

topicus: This is done.

proxy/context.go
szuecs: True originalRequest.URL won't have the Host.
Basically because we get a request like this:
```
GET /foo HTTP/1.1
Host: example.org
...
```
The parsed request URL will only have information of the first line, so `req.URL.String()` will be `/foo` and `Host` won't be in `req.Headers`, but in `req.Host`.

proxy/context.go
```diff
@@ -233,10 +232,59 @@ func (c *context) clone() *context {
...
+
+func (c *context) Loopback() {
+	err := c.proxy.do(c)
+	if c.response != nil && c.response.Body != nil {
+		if _, err := io.Copy(ioutil.Discard, c.response.Body); err != nil {
+			c.proxy.log.Errorf("context: error while discarding remainder response body: %v.", err)
+		}
+		err := c.response.Body.Close()
+		if err != nil {
+			c.proxy.log.Error("context: error during closing the response body", err)
```
szuecs, line 271: Error -> Errorf and add `: %v`

proxy/proxy.go
```diff
@@ -35,7 +36,6 @@ import (
 	"github.com/zalando/skipper/ratelimit"
 	"github.com/zalando/skipper/rfc"
 	"github.com/zalando/skipper/routing"
-	"github.com/zalando/skipper/scheduler"
```
szuecs, line 38: Can you run `make fmt` please

proxy/context.go
topicus: We can postpone this for the next iteration.

proxy/context.go
topicus: Fixed.

proxy/context.go
topicus: Is SpanFromContext cloning the context or returning a reference? If it is a reference I assume the span to be closed in the main request. If not, should I close the cloned version of the span?

proxy/context.go
topicus: I have pushed this, Could you check if the implementation is correct?

proxy/context.go
topicus: I pushed this change. Could you review it?

proxy/context.go
aryszka: my previous comment was wrong. See: https://github.com/zalando/skipper/blob/master/proxy/proxy.go#L488

filters/tee/teeloopback_test.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func TestLoopbackAndMatchPredicate(t *testing.T) {
+	// Test split and shadow routes are used:
+	// the backend set in the split route should serve the request issued by the client
+	// the backend set in the shadow route should serve the request issued by the teeLoopback
+	// the backend set in the original route should not be called
+	const routeDoc = `
+		original: Path("/foo") -> "%v";
+	 	split: Path("/foo") && True()  -> teeLoopback("A") ->"%v";
```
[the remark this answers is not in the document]
topicus: Fixed.

proxy/context.go
aryszka: lgtm

proxy/context.go
aryszka: lgtm

proxy/context.go
topicus: Done, could you check?

aryszka: lgtm

aryszka:
:+1:

topicus:
👍

ruiaraujo:
:+1:

addityasingh:
👍

proxy/context.go
```diff
@@ -233,10 +236,66 @@ func (c *context) clone() *context {
...
+	u := new(url.URL)
+	*u = *originalRequest.URL
+	u.Host = originalRequest.Host
+	cr, body, err := cloneRequest(u, originalRequest)
+	if err != nil {
+		c.proxy.log.Errorf("context: failed to clone request: %v", err)
+		return nil, err
+	}
+	serverSpan := opentracing.SpanFromContext(originalRequest.Context())
+	cr = cr.WithContext(opentracing.ContextWithSpan(cr.Context(), serverSpan))
```
szuecs, line 268: What if we would set some meta information that would show up in the new proxy Span for the shadow traffic, for example a `shadow: true` tag ?

Not sure if here or better at https://github.com/zalando/skipper/pull/1443/files#diff-969f319a10987eafd414f7c4d57e91e6R286

szuecs: I tested that it's better to have it just before finish the proxy span, such that we have the tag only in the proxy and not in the ingress span.

filters/tee/teeloopback.go
```diff
@@ -0,0 +1,49 @@
...
+	return FilterName
+}
+
+func (t *teeLoopbackSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+
+	if len(args) != 1 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+	teeKey, _ := args[0].(string)
+	if teeKey == "" {
```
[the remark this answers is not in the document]
aryszka: I raised the question, whether we should actually allow it. Then we settled on not allowing it to limit potential manual mistakes.

szuecs:
:+1:

aryszka:
:+1:

Issue #1207, opened 2019-10-04
Slowly ramp up traffic to new backend-ips


otrosien:
**Is your feature request related to a problem? Please describe.**

A service that scales up because of load might have a significant amount of RPS per backend-ip. If a new pod reports ready it will directly get its full share of RPS. This can overload services that don't pre-warm fully. It also requires each service to properly implement its pre-warming properly. Having a slow ramp-up of traffic will make it easier for services to scale.

**Describe the solution you would like**

I would imagine the round-robin table to add a weight per backend-ip, by default it's equal. If a new backend-ip joins, it will get a lower weight (probability), and time-based the weight increases until equal to the rest. 

**Describe alternatives you've considered (optional)**

Alternatives would be creating a new stack and gradually shifting traffic to it. But the downside is of course manual effort, and with every scaling step a full shift between two stacks.

**Additional context (optional)**

AWS ELBs call this "slow_start.duration_seconds".

**Would you like to work on it?**

No

mikkeloscar:
This feature was also expressed by other users (Team in @Zalando).

szuecs:
@mikkeloscar in general sounds good, but I can also imagine that it might be dangerous if you take too long time and you have a rolling update with not so many endpoints ongoing.

szuecs:
For implementation see also https://github.com/zalando/skipper/issues/1249#issuecomment-558633258

aryszka:
This design is intended to describe the approach supporting gradually increasing the traffic sent to new backend endpoints.

**Problem:**

When using the load balanced backends, and the backend service is scaling out horizontally, new backend endpoints are added, and once Skipper detects the new endpoint, it immediately starts sending the proportional traffic rate to the new endpoint. In certain cases, the backend endpoints can tolerate the proportional traffic only after a "warm-up" period. Example:

1. a backend is running with 2 endpoints and handling 1000rps traffic, each endpoint 500rps.
2. the load increases by 200rps, and the backend service needs to scale out
3. Skipper detects the new endpoint, and immediately starts sending the proportional traffic to each backend: 400rps
4. this can cause problems for the instance if it requires a warm-up period, and failure in serving 33% of the traffic can also cause significant amount of retried requests if the clients are programmed this way, making the problem even more severe

**Proposed solution:**

Skipper would gradually, "fade-in", increase the traffic to new endpoints, starting from zero, with a configurable duration, after which the traffic to the endpoint would reach the current proportional rate.

[...]

PR #1447, opened 2020-06-14
Fadein for new backend endpoints

aryszka:
addressing: #1207 

- [x] general mechanism
- [x] configuration with filters
- [x] LB algorithm
- [x] configuration testing
- [x] mechanism test harness
- [x] LB algorithm testing
- [x] move mechanism testing to the algorithm package (loadbalancer)
- [x] test full mechanism
- [x] documentation
- [ ] ~~capture Kubernetes pod created times~~ (moving to a separate PR)

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -2019,7 +2019,7 @@ fadeInDuration("3m")
 This filter, when used together with the [fadeInDuration](#fadeInDuration) filter, can be used to adjust the
 shape of the fade-in function, based on the following equation:
 
-current-rate = proportional-rate * (time-since-detection / duration) ^ exponent
+current-rate = proportional-rate * min((now - detected) / duration, 1) ^ exponent
 
 Parameters:
 
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -1993,3 +1993,57 @@ Example:
 ```
 originMarker("apiUsageMonitoring", "deployment1", "2019-08-30T09:55:51Z")
 ```
+
+## fadeInDuration
+
+When this filter is set, and the route has a load balanced backend, then the newly added endpoints will receive
+the traffic in a gradually increasing way, starting from their detection for the specified duration, after which
+they receive equal amount traffic as the previously existing routes. By default, the fade-in function is linear,
+but it can be adjusted with the [fadeInExponent](#fadeInExponent) filter. The detection time of an load balanced
+backend endpoint is preserved over multiple generations of the route configuration (over route changes). This
+filter can be used to saturate the load of autoscaling applications that require a warm-up time and therefore a
+smooth ramp-up. The fade-in feature can be used together with the round-robin and random LB algorithms.
+
+Parameters:
+
+* the duration of the fade-in in milliseconds or as a duration string
+
+Example:
+
+```
+fadeInDuration("3m")
+```
+
+## fadeInExponent
+
+This filter, when used together with the [fadeInDuration](#fadeInDuration) filter, can be used to adjust the
+shape of the fade-in function, based on the following equation:
+
+current-rate = proportional-rate * (time-since-detection / duration) ^ exponent
+
+Parameters:
+
+* a floating point number, default: 1
+
+Example:
+
+```
+fadeInExponent(2)
+```
+
+## endpointCreated
+
+This filter marks the creation time of a load balanced endpoint. When used together with the fadeInDuration
+filter, it prevents missing the detection of a new backend instance with the same hostname. This filter is
+typically automatically appended, and it's parameters are based on external sources, e.g. the Kubernetes API.
+
+Parameters:
+
+* the address of the endpoint
+* timestamp, either as a number of seconds since the unix epocs, or a string in RFC3339 format
+
+Example:
+
+```
+endpointCreated("http://10.0.0.1:8080", "2020-12-18T15:30:00Z01:00")
+```
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -2006,7 +2006,7 @@ smooth ramp-up. The fade-in feature can be used together with the round-robin an
 While the default fade-in curve is linear, the optional exponent parameter can be used to adjust the shape of
 the fade-in curve, based on the following equation:
 
-current-rate = proportional-rate * min((now - detected) / duration, 1) ^ exponent
+current_rate = proportional_rate * min((now - detected) / duration, 1) ^ exponent
 
 Parameters:
 
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -1994,46 +1994,35 @@ Example:
 originMarker("apiUsageMonitoring", "deployment1", "2019-08-30T09:55:51Z")
 ```
 
-## fadeInDuration
+## fadeIn
 
 When this filter is set, and the route has a load balanced backend, then the newly added endpoints will receive
 the traffic in a gradually increasing way, starting from their detection for the specified duration, after which
-they receive equal amount traffic as the previously existing routes. By default, the fade-in function is linear,
-but it can be adjusted with the [fadeInExponent](#fadeInExponent) filter. The detection time of an load balanced
+they receive equal amount traffic as the previously existing routes. The detection time of an load balanced
 backend endpoint is preserved over multiple generations of the route configuration (over route changes). This
 filter can be used to saturate the load of autoscaling applications that require a warm-up time and therefore a
 smooth ramp-up. The fade-in feature can be used together with the round-robin and random LB algorithms.
 
-Parameters:
-
-* the duration of the fade-in in milliseconds or as a duration string
-
-Example:
-
-```
-fadeInDuration("3m")
-```
-
-## fadeInExponent
-
-This filter, when used together with the [fadeInDuration](#fadeInDuration) filter, can be used to adjust the
-shape of the fade-in function, based on the following equation:
+While the default fade-in curve is linear, the optional exponent parameter can be used to adjust the shape of
+the fade-in curve, based on the following equation:
 
 current-rate = proportional-rate * min((now - detected) / duration, 1) ^ exponent
 
 Parameters:
 
-* a floating point number, default: 1
+* duration: duration of the fade-in in milliseconds or as a duration string
+* fade-in curve exponent - optional: a floating point number, default: 1
 
-Example:
+Examples:
 
 ```
-fadeInExponent(2)
+fadeIn("3m")
+fadeIn("3m", 1.5)
 ```
 
 ## endpointCreated
 
-This filter marks the creation time of a load balanced endpoint. When used together with the fadeInDuration
+This filter marks the creation time of a load balanced endpoint. When used together with the fadeIn
 filter, it prevents missing the detection of a new backend instance with the same hostname. This filter is
 typically automatically appended, and it's parameters are based on external sources, e.g. the Kubernetes API.
 
--- a/filters/fadein/fadein.go
+++ b/filters/fadein/fadein.go
@@ -1,0 +1,249 @@
+package fadein
+
+import (
+	"fmt"
+	"net"
+	"net/url"
+	"strings"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/routing"
+)
+
+const (
+	DurationName        = "fadeInDuration"
+	EaseName            = "fadeInEase"
+	EndpointCreatedName = "endpointCreated"
+)
+
+type (
+	duration time.Duration
+	ease     float64
+
+	endpointCreated struct {
+		when  time.Time
+		which string
+	}
+
+	detectedFadeIn struct {
+		when       time.Time
+		duration   time.Duration
+		lastActive time.Time
+	}
+
+	postProcessor struct {
+		detected map[string]detectedFadeIn
+	}
+)
+
+func NewDuration() filters.Spec {
+	return duration(0)
+}
+
+func (duration) Name() string { return DurationName }
+
+func (duration) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) != 1 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	switch v := args[0].(type) {
+	case int:
+		return duration(v * int(time.Millisecond)), nil
+	case float64:
+		return duration(int(v) * int(time.Millisecond)), nil
+	case string:
+		d, err := time.ParseDuration(v)
+		return duration(d), err
+	default:
+		return nil, filters.ErrInvalidFilterParameters
+	}
+}
+
+func (duration) Request(filters.FilterContext)  {}
+func (duration) Response(filters.FilterContext) {}
+
+func NewEase() filters.Spec {
+	return ease(0)
+}
+
+func (ease) Name() string { return EaseName }
+
+func (ease) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) != 1 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	switch v := args[0].(type) {
+	case int:
+		return ease(v), nil
+	case float64:
+		return ease(v), nil
+	default:
+		return nil, filters.ErrInvalidFilterParameters
+	}
+}
+
+func (ease) Request(filters.FilterContext)  {}
+func (ease) Response(filters.FilterContext) {}
+
+func NewEndpointCreated() filters.Spec {
+	var ec endpointCreated
+	return ec
+}
+
+func (endpointCreated) Name() string { return EndpointCreatedName }
+
+func normalizeSchemeHost(s, h string) (string, string, error) {
+	// endpoint address cannot contain path, the rest is not case sensitive
+	s, h = strings.ToLower(s), strings.ToLower(h)
+
+	h, p, err := net.SplitHostPort(h)
+	if err != nil {
+		return "", "", err
+	}
+
+	switch {
+	case p == "" && s == "http":
+		p = "80"
+	case p == "" && s == "https":
+		p = "443"
+	}
+
+	h = net.JoinHostPort(h, p)
+	return s, h, nil
+}
+
+func normalizeEndpoint(e string) (string, string, error) {
+	u, err := url.Parse(e)
+	if err != nil {
+		return "", "", err
+	}
+
+	return normalizeSchemeHost(u.Scheme, u.Host)
+}
+
+func endpointKey(scheme, host string) string {
+	return fmt.Sprintf("%s://%s", scheme, host)
+}
+
+func (endpointCreated) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) != 2 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	e, ok := args[0].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	s, h, err := normalizeEndpoint(e)
+	if err != nil {
+		return nil, err
+	}
+
+	ec := endpointCreated{which: endpointKey(s, h)}
+	switch v := args[0].(type) {
+	case int:
+		ec.when = time.Unix(int64(v), 0)
+	case float64:
+		ec.when = time.Unix(int64(v), 0)
+	case string:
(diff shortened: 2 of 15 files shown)


loadbalancer/algorithm.go
```diff
@@ -41,24 +42,81 @@ var (
...
+	rnd   *rand.Rand
+}
+
+func fadeIn(now time.Time, duration time.Duration, ease float64, detected time.Time) float64 {
+	rel := now.Sub(detected)
+	if rel >= duration {
+		return 1
+	}
+
+	return math.Pow(float64(rel)/float64(duration), ease)
```
AlexanderYastrebov, line 64: Here I like that `ease` may be <1, not sure if that is useful though:
x1/8, x1/4, x1/2, x1, x2, x4, x8

aryszka: I'm not entirely sure about the usefulness either, but there was no reason to prevent it.

filters/fadein/fadein.go
```diff
@@ -0,0 +1,252 @@
...
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	switch v := args[0].(type) {
+	case int:
+		return duration(v * int(time.Millisecond)), nil
+	case float64:
+		return duration(int(v) * int(time.Millisecond)), nil
+	case string:
+		d, err := time.ParseDuration(v)
```
szuecs, line 59: I wonder why we need the other cases and can't just always use string

filters/fadein/fadein.go
```diff
@@ -0,0 +1,252 @@
...
+	EaseName            = "fadeInEase"
+	EndpointCreatedName = "endpointCreated"
+)
+
+type (
+	duration time.Duration
+	ease     float64
+
+	endpointCreated struct {
+		when  time.Time
```
szuecs, line 27: "when": what does it mean? Creation time?

filters/fadein/fadein.go
```diff
@@ -0,0 +1,252 @@
...
+		which string
+	}
+
+	detectedFadeIn struct {
+		when       time.Time
```
szuecs, line 32: "when": what does it mean? Creation time or start of fade-in or time we detected a fade-in?

filters/fadein/fadein.go
```diff
@@ -0,0 +1,252 @@
...
+		duration   time.Duration
```
szuecs, line 33: what is duration in the context of a detected fade-in ?

filters/fadein/fadein.go
```diff
@@ -0,0 +1,252 @@
...
+
+const (
+	DurationName        = "fadeInDuration"
+	EaseName            = "fadeInEase"
+	EndpointCreatedName = "endpointCreated"
+)
+
+type (
+	duration time.Duration
+	ease     float64
```
szuecs, line 24: Is `ease` used to wait before starting a fade-in to let JIT compilers do their work before we start to fade-in traffic?

loadbalancer/algorithm.go
```diff
@@ -41,24 +42,81 @@ var (
...
+}
+
+func chance(rnd *rand.Rand, c float64) bool {
+	return rnd.Float64() < c
+}
+
+// expects len(ctx.Route.LBEndpoints) to be >= 2
+func shiftToRemaining(rnd *rand.Rand, ctx *routing.LBContext, now time.Time, from int) routing.LBEndpoint {
+	fd, fe, ep := ctx.Route.LBFadeInDuration, ctx.Route.LBFadeInEase, ctx.Route.LBEndpoints
```
szuecs, line 73: I would prefer fd -> d and fe -> ease. For me this would be more readable

proxy/fadein_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+Available data:
+- fade-in configuration: this contains the fade-in duration and optionally the fade-in degree.
+- detection time: this is the time when Skipper detected an endpoint that it didn't know about.
+- created time, optional: this is external information about the created of an endpoint, when available. E.g.
+  it can be the time when a Kubernetes pod became ready. It can be used as a fallback value for the detection
+  time in cases when Skipper (re)starts. When present, it renders any detection time past this value as invalid.
+- synced detection time, optional: it's the detection time of an endpoint shared across multiple Skipper
+  instances, stored in Redis or shared as a Skipper swarm datagram. It can be used to continue the fade-in when
+  Skipper gets (re)started. It has a precedence over the external created time in these cases.
```
szuecs, line 17: If we would store it in swarm/redis we also need the case when it's not present, because of errors of the redis/swarm.

filters/fadein/fadein.go
aryszka: It supports programmatic definition of routes when you use skipper as a library and want to skip the parsing. I think about it as a nicety for library usage, but also useful in the test code.

routing/routing.go
```diff
@@ -144,6 +144,7 @@ type RouteFilter struct {
 // backends.
 type LBEndpoint struct {
 	Scheme, Host string
+	Detected     time.Time
```
szuecs, line 147: can you please add godoc?

filters/fadein/fadein.go
aryszka: yes. The 'created' term is already noted in the name of the type, I thought it's simpler to call it this way. It also could be 'value', maybe that's the best.

routing/routing.go
```diff
@@ -199,6 +200,9 @@ type Route struct {
 	// LBAlgorithm is the selected load balancing algorithm
 	// of a load balanced route.
 	LBAlgorithm LBAlgorithm
+
+	LBFadeInDuration time.Duration
+	LBFadeInEase     float64
```
szuecs, line 205: can you please add godoc for both?

filters/fadein/fadein.go
aryszka: See similar, though: https://github.com/golang/go/blob/master/src/time/sleep.go#L15

filters/fadein/fadein.go
aryszka: The term 'detected' is in the name of the type. See similar, though: https://github.com/golang/go/blob/master/src/time/sleep.go#L15

filters/fadein/fadein.go
aryszka: It's the fade-in duration.

filters/fadein/fadein.go
aryszka: No, see: https://github.com/zalando/skipper/pull/1447/files#diff-8ef30ff734de1bf682a3b0d530f9f663R64

proxy/fadein_test.go
aryszka: isn't that covered by the combination of state and event: proxies without syncing + proxy started during fade-in?

routing/routing.go
aryszka: sure. (You can ignore the missing godoc momentarily, the PR is just a draft currently, and there is no godoc yet anywhere.)

filters/fadein/fadein.go
szuecs: Ok thanks!

filters/fadein/fadein.go
szuecs: I see ok makes sense "detected fade in -> when", but maybe `start` sounds more good?
Not sure, I will think about this while working on the code

filters/fadein/fadein.go
szuecs: I see ok makes sense "detected fade in -> when", but maybe `start` sounds more good?
Not sure, I will think about this while working on the code

filters/fadein/fadein.go
szuecs: Yeah this "when" makes sense.

filters/fadein/fadein.go
szuecs: starting from `when` until `duration`, so maybe `start` and `until`?

filters/fadein/fadein.go
szuecs: Ok, JIT apps can also use startupProbe, but `ease` in its use: y in this function: `f(x,y) : x ** y`

filters/fadein/fadein.go
```diff
@@ -0,0 +1,252 @@
...
+				duration:   ri.LBFadeInDuration,
+				lastActive: now,
+			}
+		}
+	}
+
+	// cleanup old detected, considering last active
+	for key, d := range p.detected {
+		// this allows tolerating when a fade-in endpoint temporarily disappears:
+		if d.lastActive.Add(d.duration).Before(now) {
```
szuecs, line 246: I am not sure if I understand.
The loop is to clean up the state map of the fade-in endpoints as far as I understand.
For me the natural way would be: `now > when + duration`
Would this not make more sense to have this if condition:
```
if when.Add(d.duration).After(now) { delete...}
```

filters/fadein/fadein.go
aryszka: i think this should be fine. The type name is detectedFadeIn, when stands for when was it detected, while the duration stands for whats the duration of the fade-in. 'Until' would not be good, because that rather should represent a point in time and not a duration.

filters/fadein/fadein.go
aryszka: the point of considering the lastActive is to smoothen out the effect of possible flakiness, e.g. when the infrastructure temporarily doesn't detect an endpoint. By using the last active time, we can avoid resetting the fade-in unless the endpoint disappears for at least the duration of the fade-in.

loadbalancer/algorithm.go
```diff
@@ -41,48 +42,201 @@ var (
...
+		i = (i + 1) % len(ep)
+		if i == from {
+			break
+		}
+
+		if _, fadingIn := fadeInState(now, fd, ep[i].Detected); !fadingIn {
+			sum++
+		}
+
+		w[i] = sum
```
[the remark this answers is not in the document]
aryszka: right, it was overengineered. Clean up :+1:

loadbalancer/algorithm.go
```diff
@@ -41,48 +42,201 @@ var (
...
+	// TODO: use an alternative in this case
+	if sum == 0 {
+		return ep[from]
+	}
+
+	r := rand.Intn(sum)
+	i = from
+	for {
+		i = (i + 1) % len(ep)
+		if i == last || r < w[i] {
```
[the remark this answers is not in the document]
aryszka: same

aryszka:
@szuecs you had a suggestion in the code comments to consider rerunning Apply when the fade-in passes an endpoint. The problem with that approach would be that it would be bring uneven distribution in the case of the round-robin algorithm, because it would shift too many choices to the endpoints that follow one with an early stage of fading in. Tested, I think the current approach is cheap and simple enough. I had to improve the previous approach with handling the case when there are only fading in endpoints but in different stages of fading in.

loadbalancer/algorithm.go
```diff
@@ -40,49 +41,149 @@ var (
 	defaultAlgorithm = newRoundRobin
 )
 
-func newRoundRobin(endpoints []string) routing.LBAlgorithm {
-	i := time.Now().UnixNano()
-	rand.Seed(i)
-	return &roundRobin{
-		index: rand.Intn(len(endpoints)),
+func fadeInState(now time.Time, duration time.Duration, detected time.Time) (time.Duration, bool) {
+	rel := now.Sub(detected)
```
AlexanderYastrebov, line 45: Can this be negative?

docs/reference/filters.md
```diff
@@ -1993,3 +1993,57 @@ Example:
...
+```
+fadeInDuration("3m")
+```
+
+## fadeInExponent
+
+This filter, when used together with the [fadeInDuration](#fadeInDuration) filter, can be used to adjust the
+shape of the fade-in function, based on the following equation:
+
+current-rate = proportional-rate * (time-since-detection / duration) ^ exponent
```
AlexanderYastrebov, line 2022: Should it reflect what the rate is when elapsed > duration? 
E.g. like `rate = min((now - detected)/duration, 1) ^ exponent`?

loadbalancer/algorithm.go
aryszka: In practice, not really, because we capture the detected timestamp, in-process. OTOH, the structure defining the input field is public, so it is a good point to validate it. Thx, will add a check.

docs/reference/filters.md
aryszka: Thanks, makes sense!

filters/fadein/fadein.go
szuecs: Ok!

loadbalancer/algorithm.go
```diff
@@ -40,49 +41,149 @@ var (
...
 
-type consistentHash struct{}
+type consistentHash struct {
+	rnd *rand.Rand
+}
 
 func newConsistentHash(endpoints []string) routing.LBAlgorithm {
-	return &consistentHash{}
+	return &consistentHash{
+		rnd: rand.New(rand.NewSource(time.Now().UnixNano())),
```
mikkeloscar, line 177: Doesn't seem to be used anywhere?

docs/reference/filters.md
```diff
@@ -1993,3 +1993,57 @@ Example:
...
+
+* the duration of the fade-in in milliseconds or as a duration string
+
+Example:
+
+```
+fadeInDuration("3m")
+```
+
+## fadeInExponent
```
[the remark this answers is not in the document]
aryszka: true it could be. I actually don't remember why it was separated, I don't have it in my notes anymore. Then the filter could also just be called "fadeIn".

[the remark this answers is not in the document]
aryszka: I will change this.

filters/fadein/fadein.go
```diff
@@ -0,0 +1,267 @@
...
+		if ri.Route.BackendType != eskip.LBBackend {
+			continue
+		}
+
+		ri.LBFadeInDuration = 0
+		ri.LBFadeInExponent = 1
+		endpointsCreated := make(map[string]time.Time)
+		for _, f := range ri.Filters {
+			if d, ok := f.Filter.(duration); ok {
+				ri.LBFadeInDuration = time.Duration(d)
```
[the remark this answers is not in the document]
aryszka: true

filters/fadein/fadein.go
```diff
@@ -0,0 +1,267 @@
...
+	EndpointCreatedName = "endpointCreated"
+)
+
+type (
+	duration time.Duration
+	exponent float64
+
+	endpointCreated struct {
+		when  time.Time
+		which string
```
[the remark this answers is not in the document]
aryszka: Maybe, I just like it how the names 'when' and 'which' fit together with the type name that is 'endpoint created'.

filters/fadein/fadein.go
```diff
@@ -0,0 +1,267 @@
...
+			s, h, err := normalizeSchemeHost(ep.Scheme, ep.Host)
+			if err != nil {
+				log.Errorf(configErrFmt, ep.Scheme, ep.Host, err)
+				continue
+			}
+
+			key := endpointKey(s, h)
+			detected := p.detected[key].when
+			if detected.IsZero() || endpointsCreated[key].After(detected) {
+				detected = now
```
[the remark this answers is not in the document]
aryszka: No, that's not the purpose of the endpoint created value. The endpoints created value is used simply to detect when a new endpoint has the same hostname as an old one, or it was restarted. The reason why using it instead of 'now' would be bad is that there can be a large time between the created value and when we actually start sending traffic, and also an external API may send wrong values, too, e.g. due to bugs, while have full control over the internal timestamp.

loadbalancer/algorithm.go
aryszka: true, thx!

filters/fadein/fadein.go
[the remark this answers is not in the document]
mikkeloscar: I see, thanks for clarifying.

docs/reference/filters.md
```diff
@@ -1993,3 +1993,46 @@ Example:
...
+the traffic in a gradually increasing way, starting from their detection for the specified duration, after which
+they receive equal amount traffic as the previously existing routes. The detection time of an load balanced
+backend endpoint is preserved over multiple generations of the route configuration (over route changes). This
+filter can be used to saturate the load of autoscaling applications that require a warm-up time and therefore a
+smooth ramp-up. The fade-in feature can be used together with the round-robin and random LB algorithms.
+
+While the default fade-in curve is linear, the optional exponent parameter can be used to adjust the shape of
+the fade-in curve, based on the following equation:
+
+current-rate = proportional-rate * min((now - detected) / duration, 1) ^ exponent
```
szuecs, line 2009: Please don't use name-with-minus better to read is current_rate, because no one will first read - like a math term.

filters/fadein/fadein.go
```diff
@@ -0,0 +1,260 @@
...
+	if len(args) == 0 || len(args) > 2 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	var f fadeIn
+	switch v := args[0].(type) {
+	case int:
+		f.duration = time.Duration(v * int(time.Millisecond))
+	case float64:
+		f.duration = time.Duration(int(v) * int(time.Millisecond))
```
szuecs, line 61: `int(v)` truncates the float64 value. See also https://play.golang.org/p/3wFYVtpac9j
You want this:
```
	time.Duration(int(v * float64(time.Millisecond)))
```

filters/fadein/fadein.go
```diff
@@ -0,0 +1,260 @@
...
+
+	return normalizeSchemeHost(u.Scheme, u.Host)
+}
+
+func endpointKey(scheme, host string) string {
+	return fmt.Sprintf("%s://%s", scheme, host)
+}
+
+// endpointCreated("", <int,float64,string,time>)
+// endpointCreated("http://10.2.4.123:2156", $time)
```
szuecs, line 142: Better to have godoc format :)

filters/fadein/fadein.go
```diff
@@ -0,0 +1,260 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	ec := endpointCreated{which: endpointKey(s, h)}
+	switch v := args[1].(type) {
+	case int:
+		ec.when = time.Unix(int64(v), 0)
+	case float64:
+		ec.when = time.Unix(int64(v), 0)
```
szuecs, line 163: Because you drop here the float value anyways, can we drop this case?

filters/fadein/fadein.go
```diff
@@ -0,0 +1,260 @@
...
+		ec.when = v
+	default:
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	// mitigate potential flakyness caused by clock skew. When the created time is in the future based on
+	// the local clock, we ignore it.
+	now := time.Now()
+	if ec.when.After(now) {
+		log.Errorf("Endpoint created time in the future: %v. Potential clock skew.", ec.when)
```
szuecs, line 181: Maybe add what will happen in the message to understand impact from the message itself.

filters/fadein/fadein.go
```diff
@@ -0,0 +1,260 @@
...
+	for _, ri := range r {
+		if ri.Route.BackendType != eskip.LBBackend {
+			continue
+		}
+
+		ri.LBFadeInDuration = 0
+		ri.LBFadeInExponent = 1
+		endpointsCreated := make(map[string]time.Time)
+		for _, f := range ri.Filters {
+			if fi, ok := f.Filter.(fadeIn); ok {
```
szuecs, line 212: Let's use a type switch here

filters/fadein/fadein.go
szuecs: So `p.detected` stores endpoints which do not need to be passed via dataclient.
And we clean up endpoints that are hitting this case here, such that we tolerate temporary failures.
Ok got it now.

filters/fadein/fadein.go
aryszka: not even sure how this made it here, it might have been just some notes :)

filters/fadein/fadein.go
aryszka: No, because the parser returns currently only floats.

filters/fadein/fadein.go
aryszka: ok (the fade-in will happen, we just ignore the created time)

szuecs:
👍

szuecs (approved):

szuecs:
👍

aryszka:
:+1:

PR #1455, opened 2020-06-25
OIDC header integrity

universam1:
This is a security issue mitigation, where a potential attack would be that a person is sending manipulated header that are parsed for authentication.

Currently the OIDC header are `.Add` that is an `append` operation, which keeps the user generated values as precedence.
This PR solves the issue by changing the header output to a `.Set` operation which replaces potential existing headers accordingly.


universam1:
FYI @szuecs @aryszka

szuecs:
:+1:

aryszka:
:+1:

PR #1456, opened 2020-06-29
Filter RouteGroups via an annotation

### LINKED ISSUE #1452: Support multiple ingress controllers with different routing requirements in RouteGroup
We would like to utilise two ingresses, one internal-only ingress for inter-cluster communication, and one external-only ingress for customer products. These require different routing logic to be applied depending on the ingress that it comes from

An ability to apply rules defined in RouteGroups to traffic from a specific ingress.

Copy & pasted Slack message;

```
not yet. Please create an issue for that.
It should not be that complicated clusterclient.go line 280 is the ingress filtering 
and just below is the loadRouteGroups which should have a similar 
filtering in the for loop before we append the i into rgs.
```

Yes, but no time

dlmiddlecote:
Closes #1452.

Few todos that I'd love some pointers on, if possible.

todo:
- [x] Confirm annotation: `zalando.org/routegroup.class`
- [x] Add documentation
- [x] Add more tests?

### DIFF
--- a/dataclients/kubernetes/clusterclient.go
+++ b/dataclients/kubernetes/clusterclient.go
@@ -25,6 +25,7 @@ const (
 	clusterZalandoResourcesURI = "/apis/zalando.org/v1"
 	routeGroupsName            = "routegroups"
 	routeGroupsClusterURI      = "/apis/zalando.org/v1/routegroups"
+	routeGroupClassKey         = "zalando.org/routegroup-class"
 	servicesClusterURI         = "/api/v1/services"
 	endpointsClusterURI        = "/api/v1/endpoints"
 	defaultKubernetesURL       = "http://localhost:8001"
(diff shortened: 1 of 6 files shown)


dataclients/kubernetes/clusterclient.go
```diff
@@ -283,16 +291,35 @@ func (c *clusterClient) loadIngresses() ([]*ingressItem, error) {
 	return fItems, nil
 }
 
+// skipRouteGroup decides whether a found RouteGroup should be skipped or not.
+func (c *clusterClient) skipRouteGroup(i *routeGroupItem) bool {
```
aryszka, line 295: it would be more expressive if the validation and the filtering would stay separate. The name skipRouteGroup doesn't nicely express what it does. I think the two statements just could go inline.

szuecs:
I guess what rest is only user documentation.
We have something for ingress.class in docs/kubernetes/ingress-controller.md, so I think it would be great to have a small section with headline in docs/kubernetes/routegroups.md.

dlmiddlecote:
> I guess what rest is only user documentation.
> 
> We have something for ingress.class in docs/kubernetes/ingress-controller.md, so I think it would be great to have a small section with headline in docs/kubernetes/routegroups.md.

Cool, will get to this today. 

Is the annotation key okay?

szuecs:
The annotation key is great and documentation is done, so I think we can merge.

szuecs:
:+1:

aryszka:
:+1:

PR #1458, opened 2020-06-30
Allow to wrap http.Handler and http.RoundTripper

mariusgrigaitis:
Expose ability to wrap http.Handler and http.RoundTripper
when using skipper in library mode. It was previously internal.

This allows to inject additional logic (like tracing) into skipper.

To make this possible also depend on http.RoundTripper interface
internally.
https://github.com/zalando/skipper/pull/1458/commits/ffd302f53a4de9101bade3611ccb0609c833eeb3#diff-3ac192716a1df66bbd1008d327af8792R318

See examples on how it could be used:

https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-go-handler.html
https://docs.newrelic.com/docs/agents/go-agent/instrumentation/instrument-go-transactions#http-handler-txns

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -230,6 +230,12 @@ type Params struct {
 	// OpenTracing contains parameters related to OpenTracing instrumentation. For default values
 	// check OpenTracingParams
 	OpenTracing *OpenTracingParams
+
+	// CustomHttpRoundTripperWrap provides ability to wrap http.RoundTripper created by skipper.
+	// http.RoundTripper is used for making outgoing requests (backends)
+	// It allows to add additional logic (for example tracing) by providing a wrapper function
+	// which accepts original skipper http.RoundTripper as an argument and returns a wrapped roundtripper
+	CustomHttpRoundTripperWrap func(http.RoundTripper) http.RoundTripper
 }
 
 type (
@@ -309,7 +315,7 @@ type Proxy struct {
 	maxLoops                 int
 	defaultHTTPStatus        int
 	routing                  *routing.Routing
-	roundTripper             *http.Transport
+	roundTripper             http.RoundTripper
 	priorityRoutes           []PriorityRoute
 	flags                    Flags
 	metrics                  metrics.Metrics
@@ -323,6 +329,7 @@ type Proxy struct {
 	upgradeAuditLogOut       io.Writer
 	upgradeAuditLogErr       io.Writer
 	auditLogHook             chan struct{}
+	clientTLS                *tls.Config
 }
 
 // proxyError is used to wrap errors during proxying and to indicate
@@ -605,6 +612,13 @@ func WithParams(p Params) *Proxy {
 		p.ExpectContinueTimeout = DefaultExpectContinueTimeout
 	}
 
+	if p.CustomHttpRoundTripperWrap == nil {
+		// default wrapper which does nothing
+		p.CustomHttpRoundTripperWrap = func(original http.RoundTripper) http.RoundTripper {
+			return original
+		}
+	}
+
 	tr := &http.Transport{
 		DialContext: newSkipperDialer(net.Dialer{
 			Timeout:   p.Timeout,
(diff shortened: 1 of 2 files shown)


proxy/proxy.go
```diff
@@ -230,6 +230,12 @@ type Params struct {
 	// OpenTracing contains parameters related to OpenTracing instrumentation. For default values
 	// check OpenTracingParams
 	OpenTracing *OpenTracingParams
+
+	// CustomHttpRoundTripperWrap provides ability to wrap http.RoundTripper created by skipper.
+	// http.RoundTripper is used for making outgoing requests (backends)
+	// It allows to add additional logic (for example tracing) by providing a wrapper function
+	// which accepts original skipper http.RoundTripper as an agrument and returns a wrapped roundtripper
```
szuecs, line 237: agrument -> argument

mariusgrigaitis: Fixed in all places

skipper.go
```diff
@@ -310,6 +309,18 @@ type Options struct {
 	// Custom data clients to be used together with the default etcd and Innkeeper.
 	CustomDataClients []routing.DataClient
 
+	// CustomHttpHandlerWrap provides ability to wrap http.Handler created by skipper.
+	// http.Handler is used for accepting incoming http requests.
+	// It allows to add additional logic (for example tracing) by providing a wrapper function
+	// which accepts original skipper handler as an argument and returns a wrapped handler
+	CustomHttpHandlerWrap func(http.Handler) http.Handler
```
szuecs, line 316: Maybe it's then better to call it `WithHttpHandler`?
@aryszka do you have a name suggestion?

skipper.go
```diff
@@ -310,6 +309,18 @@ type Options struct {
...
+
+	// CustomHttpRoundTripperWrap provides ability to wrap http.RoundTripper created by skipper.
+	// http.RoundTripper is used for making outgoing requests (backends)
+	// It allows to add additional logic (for example tracing) by providing a wrapper function
+	// which accepts original skipper http.RoundTripper as an argument and returns a wrapped roundtripper
+	CustomHttpRoundTripperWrap func(http.RoundTripper) http.RoundTripper
```
szuecs, line 322: Maybe it's then better to call it `WithRoundTripper`?
@aryszka do you have a name suggestion?

skipper.go
mariusgrigaitis: I would expect `WithHttpHandler` to take just http.Handler object aka signature:

```
WithHttpHandler http.Handler
```

I don't like the naming part myself but suggestions are welcome ;)

proxy/proxy.go
```diff
@@ -673,7 +687,7 @@ func WithParams(p Params) *Proxy {
 
 	return &Proxy{
 		routing:                  p.Routing,
-		roundTripper:             tr,
+		roundTripper:             p.CustomHttpRoundTripperWrap(tr),
```
aryszka, line 690: looks good to me, just started thinking whether we should make more dynamic by calling the wrapper on every case where the roundtripper is used, and not only on initialization. Not sure yet what would be the implications. Also, we can do that in a subsequent changeset.

skipper.go
aryszka: I would drop the "Custom..." prefix, and just would call it WrapHandler and WrapTransport, but this is also fine with me.

proxy/proxy.go
mariusgrigaitis: which cases do you mean by "every case"?

skipper.go
aryszka: Or simply WrapTransport? I don't have a strong opinion, though.

skipper.go
mariusgrigaitis: The reason I put `Custom` in the name is because currently there are CustomPredicates, CustomFilters and those are things that you pass in as structs (interfaces) instead of cli-arguments.

aryszka:
Thanks for the PR!

It looks good to me. Please check the suggested names, not necessarily better than the current ones, but maybe.

aryszka:
:+1:

proxy/proxy.go
aryszka: by every case I meant every outgoing request, but currently this is just a thought.

szuecs:
:+1:

szuecs:
We discussed it and have no better idea about naming so merge

Issue #965, opened 2019-02-24
Prometheus route metrics still available after deleting route


wndhydrnt:
While working with Skipper, I recognized that metrics of deleted routes are still exposed on the Prometheus endpoint "/metrics".

traefik faced a similar issue and documented their solution in the source code[1].
There is also a question regarding obsolete metrics in the prometheus-users group[2].

1: https://github.com/containous/traefik/blob/a09dfa3ce10f222ab6e10ac11386d330698ff6f8/metrics/prometheus.go#L48-L59
2: https://groups.google.com/d/msg/prometheus-users/EbumSJYsJyc/wvdL_NqqAAAJ

---

## Steps to reproduce

1. I'm running Skipper Version 2c45dcac04f8e6e4da6891eacd0f9592942192cc (built and executed locally) and Kubernetes v1.11.6 on minikube
2. Deploy an application:
```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: web-deployment-prod
  labels:
    app: web
    release: prod
spec:
  replicas: 1
  selector:
    matchLabels:
      app: web
      release: prod
  template:
    metadata:
      labels:
        app: web
        release: prod
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80

---
kind: Service
apiVersion: v1
metadata:
  name: web-service-prod
spec:
  selector:
    app: web
    release: prod
  ports:
  - protocol: TCP
    port: 80

---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: skipper
  name: web-ingress
spec:
  rules:
  - host: web.dev.local
    http:
      paths:
      - backend:
          serviceName: web-service-prod
          servicePort: 80
        path: /
```
3. Send a request to the application: `curl -i -H'Host: web.dev.local' 127.0.0.1:9999`
4. Check the "/metrics" endpoint of Skipper. Metrics like the following are exposed:
```
skipper_response_duration_seconds_sum{code="200",method="GET",route="kube_default__web_ingress__web_dev_local_____web_service_prod"} 0.000525136
skipper_response_duration_seconds_count{code="200",method="GET",route="kube_default__web_ingress__web_dev_local_____web_service_prod"} 1
```
5. Delete the Ingress: `kubectl delete ingress web-ingress`
6. Check the "/metrics" endpoint of Skipper again.

### Expected result

[...]

Issue #202, opened 2016-08-30
Allow to cleanup resources for filters


grassator:
Currently filters instantiated using a `createFilter` method can leak memory or other resources if they are requisitioned on creation and the corresponding route is consequently removed from a skipper instance.

The proposal is to add an additional interface `filter.ManagedFilter` that would have a single `Close()` method that is called whenever a route is removed from a skipper instance.

LappleApple:
Hey @grassator @aryszka @Raffo @szuecs

aryszka:
this is relevant. At least one service being implemented may need it in the near future. Adding the Help Wanted label.

AlexanderYastrebov:
> Currently filters instantiated using a createFilter method can leak memory or other resources if they are requisitioned on creation and the corresponding route is consequently removed from a skipper instance.

All routes (and hence all filters) are recreated on any route update because whole route table is re-instantiated, as can be demonstrated by:
```diff
diff --git a/routing/datasource.go b/routing/datasource.go
index 3eaa44c..c772e94 100644
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -230,7 +230,7 @@ func createFilter(fr filters.Registry, def *eskip.Filter, cpm map[string]Predica
 
                return nil, fmt.Errorf("filter not found: '%s'", def.Name)
        }
-
+       fmt.Printf("create filter: %v %v\n", def.Name, def.Args)
        return spec.CreateFilter(def.Args)
 }
```

This means that all filters from previous version of routing table should be cleaned up. 
In order to perform safe cleanup there must be no in-flight requests that use route from previous version of route table. To achieve that route table (or particular route within it) has to track request start and completion. This in turn would probably require a move away from simple `atomic.Value` load/store
https://github.com/zalando/skipper/blob/233e1d33f24e0f3ef062488fbb99bfb187c1cebf/routing/routing.go#L390-L393
https://github.com/zalando/skipper/blob/233e1d33f24e0f3ef062488fbb99bfb187c1cebf/routing/routing.go#L338-L344
to some critical section that would not only guard table access but also track acquisitions (e.g. with a `sync.WaitGroup`)

It is also interesting that unimplemented `Close()` filter method is mentioned in the docs:
https://github.com/zalando/skipper/blame/master/docs/reference/development.md#L171-L174
https://github.com/zalando/skipper/blame/master/docs/tutorials/development.md#L38-L41

and some filters even define it
[...]

PR #1463, opened 2020-07-02
custom metrics from filters

aryszka:


AlexanderYastrebov:
This is an interesting feature but I guess it would require unsolved #202 (and similar #965)

aryszka:
good point!

deprioritizing, and for keeping things tidier, i will close the PR for now.

PR #1467, opened 2020-07-06
logHeader filter to have options "request" and "response"

szuecs:
compatible version of logHeader() filter which logs the request headers.
logHeader support 2 args: "request" and "response"
The new version also logs if we have a non nil body

### DIFF
--- a/filters/diag/logheader.go
+++ b/filters/diag/logheader.go
@@ -66,10 +66,16 @@ func (lh logHeader) Response(ctx filters.FilterContext) {
 	buf.WriteString(resp.Status)
 	buf.WriteString("\r\n")
 	for k, v := range resp.Header {
-		buf.WriteString(k)
-		buf.WriteString(": ")
-		buf.WriteString(strings.Join(v, " "))
-		buf.WriteString("\r\n")
+		if strings.ToLower(k) == "authorization" {
+			buf.WriteString(k)
+			buf.WriteString(": ")
+			buf.WriteString("TRUNCATED\r\n")
+		} else {
+			buf.WriteString(k)
+			buf.WriteString(": ")
+			buf.WriteString(strings.Join(v, " "))
+			buf.WriteString("\r\n")
+		}
 	}
 	buf.WriteString("\r\n")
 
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

topicus:
👍

filters/diag/logheader.go
```diff
@@ -2,33 +2,103 @@ package diag
...
-		log.Println(err)
+	buf.WriteString(req.Method)
+	buf.WriteString(" ")
+	buf.WriteString(req.URL.Path)
+	buf.WriteString(" ")
+	buf.WriteString(req.Proto)
+	buf.WriteString("\r\n")
+	buf.WriteString(resp.Status)
+	buf.WriteString("\r\n")
+	for k, v := range resp.Header {
```
aermakov-zalando, line 68: This should skip the `Authorization` and the `Cookie` headers (maybe with an explicit opt-in), or do you think that's not necessary?

filters/diag/logheader.go
```diff
@@ -2,33 +2,103 @@ package diag
...
+	buf := bytes.NewBuffer(nil)
+	buf.WriteString(req.Method)
+	buf.WriteString(" ")
+	buf.WriteString(req.URL.Path)
+	buf.WriteString(" ")
+	buf.WriteString(req.Proto)
+	buf.WriteString("\r\nHost: ")
+	buf.WriteString(req.Host)
+	buf.WriteString("\r\n")
+	for k, v := range req.Header {
```
aermakov-zalando, line 95: This should skip the `Authorization` and the `Cookie` headers (maybe with an explicit opt-in), or do you think that's not necessary?

filters/diag/logheader.go
szuecs: Do you think because of leaking secrets like Authorization header?

filters/diag/logheader.go
szuecs: ah ok I see, we can skip that

filters/diag/logheader.go
aermakov-zalando: Here I'm worried about logging the session IDs and other information from the cookies, I don't think we have other sensitive information in response headers.

szuecs: Cookie is not part of the headers

aermakov-zalando:
Looks like there's a failing test?

filters/diag/logheader_test.go
```diff
@@ -39,12 +41,164 @@ func TestLogHeader(t *testing.T) {
...
+		t.Error("failed to log the request header")
+		t.Log("expected:")
+		t.Log(loggerVerify.String())
+		t.Log("got:")
+		t.Log(loggerTest.String())
+	}
+}
+
+// TODO: test better with proxy, because we want to check if we log
+// Headers and in REsponse we can check these. If we chec raw string
```
addityasingh, line 61:
```suggestion
// Headers and in Response we can check these. If we check raw string
```

addityasingh:
👍

szuecs:
:+1:

PR #1469, opened 2020-07-07
Tracing/add stream body errors

szuecs:
To investigate with opentracing we need to get the error in our Span

### DIFF
--- a/proxy/tracing.go
+++ b/proxy/tracing.go
@@ -27,6 +27,8 @@ const (
 	EndEvent           = "end"
 	StartEvent         = "start"
 	StreamHeadersEvent = "stream_Headers"
+	StreamBodyEvent    = "streamBody.byte"
+	StreamBodyError    = "streamBody error"
 )
 
 type proxyTracing struct {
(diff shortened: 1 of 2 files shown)


proxy/tracing.go
```diff
@@ -27,6 +27,8 @@ const (
 	EndEvent           = "end"
 	StartEvent         = "start"
 	StreamHeadersEvent = "stream_Headers"
+	StreamBodyEvent    = "streamBody.byte"
```
aermakov-zalando, line 30: `stream_Body` so it's consistent with `StreamHeadersEvent`?

szuecs: I refactored this from the already used internal loop while streaming and I think it should be in-sync, wdyt?

szuecs: I can also create a different Name as you suggested for the error case

aermakov-zalando: Right, then let's keep it like this.

aermakov-zalando:
👍

szuecs:
:+1:

PR #1485, opened 2020-07-10
Implements lua scripting improvements

AlexanderYastrebov:
* Updates gopher-lua dependency to fix [checkptr: unsafe pointer arithmetic](https://github.com/yuin/gopher-lua/issues/254)
* Adds context path_param projection as table
* Adds request url_query projection as table
* Adds request url_raw_query getter and setter
* Adds request url_path getter and setter
* Adds response status_code getter and setter
* Implements header and url_query iteration
* Passes indexed script params along with key-values
* Uses proper log module
* Makes 'print' builtin write into log
* Uses lua's RaiseError for error handling
* Returns less results instead of pushing lua.LNil
* Updates documentation and tests

See script_test.go for usage examples.


AlexanderYastrebov:
I have looked at the `diff script/README.md docs/reference/scripts.md` and reference docs seemed more up to date. It makes no sense to maintain two documentation copies hence I've converted it to symlink.

docs/reference/scripts.md
```diff
@@ -200,6 +232,16 @@ function request(ctx, params)
...
 
+## set request header from params
+```lua
+function request(ctx, params)
+	ctx.request.header[params[1]] = params[2]
+	if params[1]:lower() == "host" then
+		ctx.request.outgoing_host = params[2]
+	end
+end
+```
```
szuecs, line 243: you can also use `preserveHost("false")` in this case, but for doc is nice to see that you can modify outgoing_host!

script/script.go
```diff
@@ -189,18 +198,19 @@ func (s *script) Response(f filters.FilterContext) {
 func (s *script) runFunc(name string, f filters.FilterContext) {
 	L, err := s.getState()
 	if err != nil {
-		log.Printf("ERROR: %s", err)
+		log.Errorf("ERROR: %v", err)
```
szuecs, line 201: I guess `ERROR:` we don't need and better to add more context where we are, such that users can understand that this is likely a lua script `getState` error (whatever the error message will look like

script/script.go
```diff
@@ -232,33 +242,36 @@ func serveRequest(f filters.FilterContext) func(*lua.LState) int {
 		t := s.Get(-1)
 		r, ok := t.(*lua.LTable)
 		if !ok {
-			s.Push(lua.LString("invalid type, need a table"))
-			return 1
+			s.RaiseError("unsupported type %v, need a table", t.Type())
```
szuecs, line 245: You could also use `%T` with `t`, but it's also fine like this if t is not nil (?)

script/script.go
```diff
@@ -232,33 +242,36 @@ func serveRequest(f filters.FilterContext) func(*lua.LState) int {
...
 	}
 }
 
-func serveTableWalk(res *http.Response) func(lua.LValue, lua.LValue) {
+func serveTableWalk(s *lua.LState, res *http.Response) func(lua.LValue, lua.LValue) {
 	return func(k, v lua.LValue) {
-		s, ok := k.(lua.LString)
+		sk, ok := k.(lua.LString)
 		if !ok {
+			s.RaiseError("unsupported key type %v, need a string", k.Type())
```
szuecs, line 259: same like https://github.com/zalando/skipper/pull/1485/files#r453719722

script/script.go
```diff
@@ -232,33 +242,36 @@ func serveRequest(f filters.FilterContext) func(*lua.LState) int {
...
 			return
 		}
-		switch string(s) {
+		switch string(sk) {
 		case "status_code":
 			n, ok := v.(lua.LNumber)
 			if !ok {
+				s.RaiseError("unsupported status_code type %v, need a number", v.Type())
```
szuecs, line 266: same like https://github.com/zalando/skipper/pull/1485/files#r453719722

script/script.go
```diff
@@ -232,33 +242,36 @@ func serveRequest(f filters.FilterContext) func(*lua.LState) int {
...
 				return
 			}
 			res.StatusCode = int(n)
 
 		case "header":
 			t, ok := v.(*lua.LTable)
 			if !ok {
+				s.RaiseError("unsupported header type %v, need a table", v.Type())
```
szuecs, line 274: same like https://github.com/zalando/skipper/pull/1485/files#r453719722

script/script.go
```diff
@@ -385,12 +422,16 @@ func setRequestValue(f filters.FilterContext) func(*lua.LState) int {
...
+				return 0
 			}
 			f.Request().URL = u
+		case "url_path":
+			f.Request().URL.Path = s.ToString(-1)
+		case "url_raw_query":
+			f.Request().URL.RawQuery = s.ToString(-1)
 		default:
-			// do nothing for now
+			s.RaiseError("unsupported request field %s", key)
```
szuecs, line 434: this could break production for people, not?

script/script.go
```diff
@@ -454,10 +516,9 @@ func setStateBag(f filters.FilterContext) func(*lua.LState) int {
 		case lua.LTNumber:
 			res = float64(val.(lua.LNumber))
 		default:
-			s.Push(lua.LString("unsupported type for state bag"))
-			return 1
+			s.RaiseError("unsupported state bag value type %v, need a string or a number", val.Type())
```
szuecs, line 519: same like https://github.com/zalando/skipper/pull/1485/files#r453719722

script/script.go
```diff
@@ -405,27 +446,48 @@ func getResponseValue(f filters.FilterContext) func(*lua.LState) int {
...
 
+func setResponseValue(f filters.FilterContext) func(*lua.LState) int {
+	return func(s *lua.LState) int {
+		key := s.ToString(-2)
+		switch key {
+		case "status_code":
+			v := s.Get(-1)
+			n, ok := v.(lua.LNumber)
+			if !ok {
+				s.RaiseError("unsupported status_code type %v, need a number", v.Type())
```
szuecs, line 474: same like https://github.com/zalando/skipper/pull/1485/files#r453719722

szuecs:
Thanks @AlexanderYastrebov I really like this PR only some nits to fix

docs/reference/scripts.md
AlexanderYastrebov: It is basically 1-1 clone of `setRequestHeader("header-name", "header-value")` filter

script/script.go
AlexanderYastrebov: It produces e.g. `unsupported type string, need a table`, `unsupported type nil, need a table`

And
```go
s.RaiseError("unsupported type %T, need a table", t)
```
produces `unsupported type lua.LString, need a table`, `unsupported type *lua.LNilType, need a table`.

I like as it is but will add a test anyway.

script/script.go
AlexanderYastrebov: This will break if people have assignment to non-existing request property.
I am not sure how to asses such a risk, could be reverted in worst case.

script/script.go
```diff
@@ -385,12 +422,16 @@ func setRequestValue(f filters.FilterContext) func(*lua.LState) int {
 		case "url":
 			u, err := url.Parse(s.ToString(-1))
 			if err != nil {
-				s.Push(lua.LString(err.Error()))
-				return 1
+				s.RaiseError("%v", err)
```
AlexanderYastrebov, line 425: @szuecs Here is another breaking change. Before the change it was a no-op in case of failure and also returned string error which is not possible to use in lua since assignment operator does not return value.

script/script.go
```diff
@@ -494,12 +558,23 @@ func setRequestHeader(f filters.FilterContext) func(*lua.LState) int {
...
+	}
+}
+
 func getResponseHeader(f filters.FilterContext) func(*lua.LState) int {
 	return func(s *lua.LState) int {
 		hdr := s.ToString(-1)
 		res := f.Response().Header.Get(hdr)
-		s.Push(lua.LString(res))
-		return 1
+		if res != "" {
```
AlexanderYastrebov, line 573: Here is also a breaking change. Before it would return empty string if header is missing and now it returns `nil` which is usual in lua return value from table when key is missing.

script/script.go
AlexanderYastrebov: For now I have just added a test until we decide how to deal with backward compatibility which I guess depends on the size of lua userbase

script/script.go
szuecs: ok then let's merge it like it is now

script/script.go
szuecs: I would prefer to comment the code and add a `TODO(sszuecs): <link>` with link to the comment here.

script/script.go
szuecs: I resolved the rest of the comments refering to here.
I think you have your point in showing me the log message and I agree the current one with using .Type() is nicer.

script/script.go
szuecs: Let's add a TODO as commented in https://github.com/zalando/skipper/pull/1485#discussion_r453818921

script/script.go
szuecs: Let's add a TODO as commented in https://github.com/zalando/skipper/pull/1485#discussion_r453818921

script/script.go
szuecs: We can change this in the v1 branch which is doing the breaking change

AlexanderYastrebov:
@szuecs Thank you for review. I have marked all the places where I think it makes sense to change behavior (in the v1 branch as you've suggested) with a link to the #1487

AlexanderYastrebov:
For the tests I have reused testable examples but overlooked `go vet` failures like
```
# github.com/zalando/skipper/script
script/script_test.go:324:1: ExampleLoadFileOK refers to unknown identifier: LoadFileOK
script/script_test.go:332:1: ExampleLoadOK refers to unknown identifier: LoadOK
...
```
:(

szuecs:
:+1:

szuecs:
please rebase

AlexanderYastrebov:
@szuecs 
> please rebase

Done, thank you

szuecs:
:+1:

muaazsaleem:
👍

PR #1489, opened 2020-07-15
Allow setting a custom Prometheus registry

wndhydrnt:
We are maintaining several custom filters and thus use skipper as a library. Prometheus is our monitoring system of choice.
The default API methods of the `metrics` package like `IncCounter()` or `UpdateGauge()` sometimes do not cut it for us.
In some cases we would like to be able to change the name of the metric instead of the label `key` to aid in discovery. Or it would be useful to add more than one label to a metric.

This change allows passing a registry down to the Prometheus implementation of the `metrics` package. This makes it possible to create custom metrics for filters, register them with the custom registry and instruct skipper to use that custom registry.
The code falls back to the current behavior of creating a new registry if no custom registry is passed from the outside.

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -6,6 +6,8 @@ import (
 	"net/http/pprof"
 	"strings"
 	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 const (
@@ -152,6 +154,11 @@ type Options struct {
 	// EnableProfile exposes profiling information on /pprof of the
 	// metrics listener.
 	EnableProfile bool
+
+	// An instance of a Prometheus registry. It allows registering and serving custom metrics when skipper is used as a
+	// library.
+	// A new registry is created if this option is nil.
+	PrometheusRegistry *prometheus.Registry
 }
 
 var (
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

szuecs:
I thought about if we could add an interface in between, such that we don't need Prometheus in the metrics/metrics.go Options, but I have no good idea how to do this better, besides having an empty interface...

szuecs:
please rebase

wndhydrnt:
I can see the argument regarding passing the registry through all layers of the configuration stack.
What about using a variable local to the package?
Something like this:

```go
package metrics

var (
    prometheusRegistry *prometheus.Registry
)

func SetPrometheusRegistry(r *prometheus.Registry) error {
    if prometheusRegistry != nil {
        return fmt.Errorf("prometheus registry already set")
    }

    prometheusRegistry = r
}
```

The function `NewPrometheus()` can then be changed to use `prometheusRegistry` or initialize a new Registry.

szuecs:
I like passing it better so let's rebase and merge

szuecs:
@wndhydrnt , @aryszka will be back next week to review the PR.
Just FYI, the change was not forgotten.

szuecs:
@aryszka ^

aryszka:
:+1:

aryszka:
Thanks a lot for the PR!

PR #1490, opened 2020-07-16
Improve RouteGroup CRD schema

aermakov-zalando:
* Add a regex for hostnames (matching one used for Ingress validation).
 * Fix array validation to use `minItems` (`minLength` doesn't do anything).
 * Make `hosts` required; having a RouteGroup that matches everything is an extremely unlikely case, so I'd rather optimise for the users not being able to break anything in a cluster by deploying a match-all RG.

### DIFF
--- a/dataclients/kubernetes/deploy/apply/routegroups_crd.yaml
+++ b/dataclients/kubernetes/deploy/apply/routegroups_crd.yaml
@@ -41,6 +41,7 @@ spec:
               type: array
               items:
                 type: string
+                pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'
             backends:
               type: array
               minLength: 1
--- a/dataclients/kubernetes/deploy/apply/routegroups_crd.yaml
+++ b/dataclients/kubernetes/deploy/apply/routegroups_crd.yaml
@@ -35,6 +35,7 @@ spec:
           required:
           - backends
           - routes
+          - hosts
           type: object
           properties:
             hosts:
@@ -42,6 +43,7 @@ spec:
               items:
                 type: string
                 pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'
+              minItems: 1
             backends:
               type: array
               minItems: 1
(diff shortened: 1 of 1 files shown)


szuecs:
@aermakov-zalando how do you create a catchall route to show a nice 404 page for example?

aermakov-zalando:
@szuecs By specifying the hosts it should apply to, similar to ingress. Would that not work?

aermakov-zalando:
Although it turns out that 1.18 [supports wildcards in hostnames](https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#support-for-hostname-wildcards), so maybe we should do the same in RouteGroups. They should still be explicit, though, to prevent incidents.

szuecs:
Yeah wildcard hostnames can work for this, good idea.

szuecs:
@universam1 I would like to merge this one, but as you wrote in Slack, that you want to use the one without specifying `hosts`, you would be impacted and we would need to test wildcard match. Are you fine if we postpone the wildcard to not immediately fix and merge this one?

aermakov-zalando:
@szuecs We could also drop this commit for now and keep a slightly different schema internally (see https://github.com/zalando-incubator/kubernetes-on-aws/pull/3407).

szuecs:
please rebase

universam1:
> @universam1 I would like to merge this one, but as you wrote in Slack, that you want to use the one without specifying `hosts`, you would be impacted and we would need to test wildcard match. Are you fine if we postpone the wildcard to not immediately fix and merge this one?

Thank you for asking @szuecs  - we were able to use regex wildcards with `hosts` as of today, so I don't have objections to make `hosts` mandatory.

```yaml
spec:
  hosts:
   - \w+.example.org
```

aermakov-zalando:
Shouldn't the route groups use the same syntax and semantics as Ingress hosts (only allowing *.foo.bar)?

szuecs:
@aermakov-zalando I think we should not change the behavior. We could add that wildcard prefix is allowed, like `*.example.org`.

szuecs:
:+1:

aermakov-zalando:
👍

vincentbaeten:
Hello
Any update on adding wildcard support to route groups? A wildcard prefix like @szuecs said would be perfect.

szuecs:
@vincentbaeten id it's not possible now then please create an issue instead of posting on a merged pull request.

PR #1491, opened 2020-07-24
Fix precedence issues with prepending ^ to regex.

tomlarkworthy:
User expressions like "/api|/$" need to become "^(/api|/$)" to operate correctly.

There are two major regex based ingress path filter [1] styles supported in skipper. Plain regex, and kubernetes ingress. This bug fix is for kubernetes ingress. To quote the docs on how this feature is supposed to work

"By default, the ingress path mode is set to kubernetes-ingress, which is interpreted as a regular expression with a mandatory leading /, and is automatically prepended by a ^ control character, **enforcing that the path has to be at the start of the incoming request path.**"

Note the intent that the ^ and leading / must be present and thus its intent as a prefix based filtering but with regex features. However, the implementation just does a string concatenation of the ^ character to the user supplied regex, which means that the start of path anchor can become detached when the user uses an operator such as "|". This comes up in practice when using a single regex to enforce two leading path prefixes using the "|" operator.

For example:
`/pathA`
`/pathB`

Given the explanation of the feature[1] one might expect path: `/pathA|/pathB` to permit traffic only to the two routes. However, the implementation expands to  `^/pathA|/pathB`. This actually permit any path that contains `/pathB` as a substring! e.g. `/**evil_route**/pathB`. The "^" is only relevant to the `/pathA` clause

This is clearly not the intent of the feature as this behavior compromises the feature description "**that the path has to be at the start of the incoming request path.**" [1] 

The implementation mistake is using string concatenation of "^" to the user supplied regex. String concatenation has "middle-of-the-pack" operator precedence in regex, notably "|" has higher precedence. To ensure the "^" anchor is logically the first thing matched, we must wrap the user supplied regex in "(", ")" before concatenation.

[1] https://opensource.zalando.com/skipper/kubernetes/ingress-usage/#ingress-path-handling


szuecs:
@tomlarkworthy please change the test in case you think this change makes sense, thanks!

tomlarkworthy:
yeah just struggling to get the tests running locally... It is definitely a clear bug to me.

tomlarkworthy:
@szuecs So its a several hours of work to fix all the tests. I am happy to do it as long as there is a good chance it will be merged at the end. I don't want to waste my time.

szuecs:
@tomlarkworthy fair enough
@aryszka what do you think to this change?
I remember you looked into the Kubernetes API spec to implement the Kubernetes style path regexp.

tomlarkworthy:
The implementation is https://github.com/zalando/skipper/pull/1491/files#diff-cc26174bd99fb74da18dffde42bd2efc

aryszka:
the change makes sense to me. The interesting thing about the spec is not something that is conflict with this. That was about that it declares regular expression, but the API validates (or validated then) the ingress objects slightly differently, expecting the expression to start with a '/', and describing that it's meant to be '^/'. That's why we prepend the expression. The example in the PR description is something that missed our attention. Thanks for the PR! Let's fix the tests, and please sign the commits (see: https://github.com/zalando/skipper/pull/1491/checks?check_run_id=919773375).

tomlarkworthy:
ok cool. Will get to it next week

tomlarkworthy:
@szuecs @aryszka w00t PTAL

szuecs:
I asked in our slack channel to check this out. Let’s wait some days such that everyone has time to check and think.

dataclients/kubernetes/ingress.go
```diff
@@ -128,7 +128,11 @@ func setPath(m PathMode, r *eskip.Route, p string) {
 	case PathRegexp:
 		r.PathRegexps = []string{p}
 	default:
-		r.PathRegexps = []string{"^" + p}
+		if p == "/" {
```
szuecs, line 131: What if p is empty?

tomlarkworthy: Well before it was "^" + p so I guess weirdest hypothetical case is when p == "" which would go to "^()" which in the regex world is equivalent to "^" so I don't think it will change behavior and will work ok. 

I had to add the p == "/" case today to fix some of the tests. Without that special case clause it seemed to be generating additional "catchall" cases in the route tables. The is coz the hasCatchAllRoutes(routes []*eskip.Route) looks for "^/" (see https://github.com/zalando/skipper/blob/dd70bd65e7f99cfb5dd6b6f71885d9fe3b2707f6/dataclients/kubernetes/ingress.go#L746).

szuecs:
:+1:

szuecs:
Let’s merge it next week, such that others can share their opinions in case someone has a different sense of the regexp. I don’t think so, but let’s see.

mariusgrigaitis:
Not sure I understand what kind of issue this PR solves. Seems dangerous.

tomlarkworthy:
@mariusgrigaitis I update the PR text a bit better to explain the bug. Do you agree that this is a bug now I have explained it better?

aryszka:
:+1:

szuecs:
Thanks @tomlarkworthy !

PR #1492, opened 2020-07-24
setting the parentSpan for response filters

ghophp:
When trying to access the ParentSpan() inside a Response filter like this:

```go
fc.ParentSpan()
```

Which is common when trying to instrument filters like this:

```go
filterSpan := fc.Tracer().StartSpan("myCoolFilter", opentracing.ChildOf(fc.ParentSpan().Context()))
```

The ParentSpan still returns the request_filters span, since the filtersSpan created in the `applyFiltersToResponse` didn't set the parentSpan variable before passing the context.

In the image bellow, the `headerSize` is a Response filter, but the span ends up in the request_filter span:

Screen Shot 2020-07-24 at 2 18 34 PM


szuecs:
:+1:

szuecs:
Thanks!

aryszka:
:+1:

PR #1493, opened 2020-07-27
Adding an AdmissionWebhook server to validate RouteGroups

muaazsaleem:
Adding Kubernetes [ValidatingAdmissionWebhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook) to validate RouteGroups.


cmd/webhook/main.go
```diff
@@ -0,0 +1,74 @@
...
+	log.Infof("Starting server on %s", address)
+
+	sig := make(chan os.Signal, 1)
+	signal.Notify(sig, syscall.SIGTERM)
+	go func() {
+		<-sig
+		server.Shutdown(context.Background())
+	}()
+
+	err := server.ListenAndServeTLS(certFile, keyFile)
```
aryszka, line 68: I think we should support non-HTTPS, too, at least for local testing. I think the logic should be something like this:

- if only one of TLS cert or key is provided => error
- if both TLS cert and key are provided, and no address is provided => listenTLS on 9443
- if both TLS cert and key are provided, and address is provided, too => listenTLS on the provided address
- if no TLS cert is provided, and no address is provided => listenPlain on 9080
- if no TLS cert is provided, and address is provided => listenPlain on the provided address

cmd/webhook/admission/admission.go
```diff
@@ -0,0 +1,218 @@
...
+
+func (r RouteGroupAdmitter) Name() string {
+	return "routegroup"
+}
+
+func (r RouteGroupAdmitter) Admit(req *admissionsv1.AdmissionRequest) (*admissionsv1.AdmissionResponse, error) {
+	rgItem := definitions.RouteGroupItem{}
+	err := json.Unmarshal(req.Object.Raw, &rgItem)
+	if err != nil {
+		emsg := fmt.Errorf("could not parse RouteGroup, %w", err)
```
aryszka, line 79: as I see, we never use emsg as an error object. We could just simply do:

```
emsg := fmt.Sprintf("could not parse RouteGroup, %v", err)
```

...and then use it as a string.

cmd/webhook/admission/admission.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func Handler(admitter Admitter) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		admitterName := admitter.Name()
+		totalRequests.WithLabelValues(admitterName).Inc()
+
+		if r.Method != "POST" || r.Header.Get("Content-Type") != "application/json" {
+			invalidRequests.WithLabelValues(admitterName).Inc()
+			w.WriteHeader(http.StatusBadRequest)
```
aryszka, line 116: a return is missing from here

cmd/webhook/admission/admission.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+}
+
+func extractName(request *admissionsv1.AdmissionRequest) string {
+	if request.Name != "" {
+		return request.Name
+	}
+
+	obj := metav1.PartialObjectMetadata{}
+	if err := json.Unmarshal(request.Object.Raw, &obj); err != nil {
```
aryszka, line 206: Question: when parsing the metadata from the raw object, shouldn't we wrap the metadata with the actual object? Do we have a test for the extract name?

cmd/webhook/admission/admission.go
```diff
@@ -0,0 +1,218 @@
...
+
+		operationInfo := fmt.Sprintf(
+			"%s %s %s/%s",
+			review.Request.Operation,
+			review.Request.Kind,
+			review.Request.Namespace,
+			extractName(review.Request),
+		)
+
+		gvr := review.Request.Resource
```
aryszka, line 144: the acronym `gvr` is not clear to me what it means

cmd/webhook/admission/admission.go
```diff
@@ -0,0 +1,218 @@
...
+
+		admResp, err := admitter.Admit(review.Request)
+		if err != nil {
+			log.Errorf("Rejected %s: %v", operationInfo, err)
+			writeResponse(w, errorResponse(review.Request.UID, err))
+			rejectedRequests.With(labelValues).Inc()
+			return
+		}
+
+		log.Infof("Allowed %s", operationInfo)
```
aryszka, line 171: I would consider if it is enough to log the successful admissions only with Debug, to avoid log spam

cmd/webhook/admission/admission.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+}
+
+func writeResponse(writer http.ResponseWriter, response *admissionsv1.AdmissionResponse) {
+	resp, err := json.Marshal(admissionsv1.AdmissionReview{
+		Response: response,
+	})
+	if err != nil {
+		log.Errorf("unable to serialize response: %v", err)
+		writer.WriteHeader(http.StatusInternalServerError)
```
aryszka, line 183: since marshalling the response failed here, we should return from this point

cmd/webhook/admission/admission.go
muaazsaleem: It's Group Version Resource https://godoc.org/k8s.io/apimachinery/pkg/runtime/schema#GroupVersionResource happy to make it more clear by a comment or rename

cmd/webhook/admission/admission.go
muaazsaleem: > when parsing the metadata from the raw object, shouldn't we wrap the metadata with the actual object?
sorry, what do you mean by wrapping the metadata with an actual object?
> Do we have a test for the extract name?
No, but I should add one

muaazsaleem: Added the test

cmd/webhook/admission/admission.go
aryszka: thx! TIL :+1:

cmd/webhook/admission/admission.go
aryszka: > sorry, what do you mean by wrapping the metadata with an actual object?

I was just not aware of the structure of [PartialObjectMetadata](https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#PartialObjectMetadata)

:+1:

cmd/webhook/main.go
```diff
@@ -0,0 +1,101 @@
...
+
+func healthCheck(writer http.ResponseWriter, _ *http.Request) {
+	writer.WriteHeader(http.StatusOK)
+	if _, err := writer.Write([]byte("ok")); err != nil {
+		log.Errorf("failed to write health check: %v", err)
+	}
+
+}
+
+func serve(address, certFile, keyFile string, handler http.Handler) {
```
aryszka, line 74: since we already have the config type, why just passing a config object to serve instead of the address, certFile and keyFile parameters? It would be easier and nicer to extend it in the future.

cmd/webhook/main.go
```diff
@@ -0,0 +1,101 @@
...
+		c.address = defaultHTTPAddress
+	}
+
+	if c.debug {
+		log.SetLevel(log.DebugLevel)
+	}
+}
+
+func main() {
+	var cfg config
```
aryszka, line 52: `cfg := &config{}`

aryszka:
:+1:

muaazsaleem:
👍

cmd/webhook/admission/admission.go
```diff
@@ -0,0 +1,220 @@
...
+
+		if r.Method != "POST" || r.Header.Get("Content-Type") != "application/json" {
+			invalidRequests.WithLabelValues(admitterName).Inc()
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+
+		body, err := ioutil.ReadAll(r.Body)
+		if err != nil {
+			log.Errorf("Unable to read request: %v", err)
```
szuecs, line 122: Normally we use `Failed to ...` instead of `Unable to...`, I think this should be fixed.

cmd/webhook/admission/admission.go
```diff
@@ -0,0 +1,220 @@
...
+		writeResponse(w, admResp)
+	}
+}
+
+func writeResponse(writer http.ResponseWriter, response *admissionsv1.AdmissionResponse) {
+	resp, err := json.Marshal(admissionsv1.AdmissionReview{
+		Response: response,
+	})
+	if err != nil {
+		log.Errorf("unable to serialize response: %v", err)
```
szuecs, line 183: unable -> Failed

cmd/webhook/admission/admission_test.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	rb, err := json.Marshal(expectedRg)
+	assert.NoError(t, err)
+
+	review := &admissionsv1.AdmissionReview{
+		Request: &admissionsv1.AdmissionRequest{
+			Name:      "r1",
+			Namespace: "n1",
+			// TODO: why doesnt runtime.RawExtension{Object: rg} work here?
```
szuecs, line 81: rg or expectedRg ?

cmd/webhook/main.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// One can use generate_cert.go in https://golang.org/pkg/crypto/tls
+	// to generate cert.pem and key.pem.
+	serve(cfg, handler)
+}
+
+func healthCheck(writer http.ResponseWriter, _ *http.Request) {
+	writer.WriteHeader(http.StatusOK)
+	if _, err := writer.Write([]byte("ok")); err != nil {
+		log.Errorf("failed to write health check: %v", err)
```
szuecs, line 69: failed -> Failed

cmd/webhook/main.go
```diff
@@ -0,0 +1,101 @@
...
+func (c *config) parse() {
+	kingpin.Flag("debug", "Enable debug logging").BoolVar(&c.debug)
+	kingpin.Flag("tls-cert-file", "File containing the certificate for HTTPS").Envar("CERT_FILE").StringVar(&c.certFile)
+	kingpin.Flag("tls-key-file", "File containing the private key for HTTPS").Envar("KEY_FILE").StringVar(&c.keyFile)
+	kingpin.Flag("address", "The address to listen on").Default(defaultHTTPSAddress).StringVar(&c.address)
+
+	kingpin.Parse()
+
+	if (c.certFile != "" || c.keyFile != "") && !(c.certFile != "" && c.keyFile != "") {
+		log.Fatal("config parse error: both of TLS cert & key must be provided or neither (for testing )")
```
szuecs, line 37: config parse error: .. -> Failed to parse config: ...

cmd/webhook/main.go
```diff
@@ -0,0 +1,101 @@
...
+		Addr:    cfg.address,
+		Handler: handler,
+	}
+
+	log.Infof("Starting server on %s", cfg.address)
+
+	sig := make(chan os.Signal, 1)
+	signal.Notify(sig, syscall.SIGTERM)
+	go func() {
+		<-sig
```
szuecs, line 85: Maybe add a log that we are shutting down would be nice

cmd/webhook/main.go
```diff
@@ -0,0 +1,101 @@
...
+
+	var err error
+	if cfg.certFile != "" && cfg.keyFile != "" {
+		err = server.ListenAndServeTLS(cfg.certFile, cfg.keyFile)
+	} else {
+		// support non-HTTPS for local testing
+		err = server.ListenAndServe()
+	}
+	if err != nil {
+		if err != http.ErrServerClosed {
```
szuecs, line 97: sounds more like you want to have `&&` in the if above

szuecs:
I think little changes should be made to adjust the logging, other than that it looks good to me.

aryszka:
:+1:

mikkeloscar:
:+1:

PR #1495, opened 2020-08-04
tracingTag with a backend name for RouteGroup-generated routes

marcinzaremba:
This is a simple change that adds an idiomatic OpenTracing tag (`skipper.backend_name`) for all routes originating from RouteGroups containing a backend reference name. That can be very helpful in debugging (clearly visible which logical backend was serving instead of seeing just actual service IPs) and metrics gathering/comparison (imagine comparing metrics between a primary and canary or AB test).

It's implemented adding an implicit `tracingTag` filter to all generated routes. (implemented this way for RouteGroups only as by default Skipper backends do not have names so it is not universal to all sources of routes).

### DIFF
--- a/skipper.go
+++ b/skipper.go
@@ -527,6 +527,10 @@ type Options struct {
 	// times when response headers & payload are streamed to the client
 	OpenTracingLogStreamEvents bool
 
+	// OpenTracingBackendNameTag enables an additional tracing tag containing a backend name
+	// for a route when it's available (e.g. for RouteGroups)
+	OpenTracingBackendNameTag bool
+
 	// PluginDir defines the directory to load plugins from, DEPRECATED, use PluginDirs
 	PluginDir string
 	// PluginDirs defines the directories to load plugins from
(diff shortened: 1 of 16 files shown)


marcinzaremba:
@aryszka Can you please suggest what's the preferred approach to correct tests for the change? Currently all the fixtures in `dataclients/kubernetes/testdata/routegroups` that tests are generated from are static files and any global change like this one (affecting all of the generated eskip files) enforces the author to correct all the expected, static `eskip` files.

marcinzaremba:
Another question, should this behavior (adding tracingTag) be enabled globally or not?

szuecs:
@marcinzaremba add a flag to enable it by default, such that you can opt-in to enable it.

marcinzaremba:
Done. Please review. Feel free to point out if I missed anything.

aryszka:
:+1:

aryszka:
Thanks for the PR!

marcinzaremba:
:+1:

skipper.go
```diff
@@ -720,6 +724,7 @@ func createDataClients(o Options, auth innkeeper.Authentication) ([]routing.Data
 			KubernetesEastWestDomain:   o.KubernetesEastWestDomain,
 			DefaultFiltersDir:          o.DefaultFiltersDir,
 			OriginMarker:               o.EnableRouteCreationMetrics,
+			BackendNameTracingTag:      o.OpenTracingBackendNameTag,
```
szuecs, line 727: Why not DefaultFilters PreProcessor as in https://github.com/zalando/skipper/blob/master/skipper.go#L1179

PR #1500, opened 2020-08-08
Unify detection of upgrade requests.

ohgnoes:
When firefox sends a websocket request, it sends it with a connection header of

Connection: keep-alive, Upgrade

Due to skipper having multiple definitions of an upgrade request, these requests are rejected with 400 bad request. This patch unifies skippers idea of what an upgrade request is.

### DIFF
--- a/proxy/upgrade.go
+++ b/proxy/upgrade.go
@@ -67,7 +67,7 @@ func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
 	// https://tools.ietf.org/html/rfc7230#section-6.7
 	// and https://tools.ietf.org/html/rfc6455 (websocket)
 	if (req.ProtoMajor <= 1 && req.ProtoMinor < 1) ||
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

proxy/upgrade.go
```diff
@@ -67,7 +67,7 @@ func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
 	// https://tools.ietf.org/html/rfc7230#section-6.7
 	// and https://tools.ietf.org/html/rfc6455 (websocket)
 	if (req.ProtoMajor <= 1 && req.ProtoMinor < 1) ||
-		strings.ToLower(req.Header.Get("Connection")) != "upgrade" ||
+		!isUpgradeRequest(req) ||
 		req.Header.Get("Upgrade") == "" {
```
AlexanderYastrebov, line 71: Is empty check still required?

AlexanderYastrebov: Its an another key

aryszka:
:+1:

PR #1502, opened 2020-08-11
Add WAF Filter plugin

jptosso:
Coraza WAF is a modsecurity based Golang WAF integrated with skipper and usable as a standalone plugin


szuecs:
Please sign off your commit to fix dco @jptosso 😊

PR #1503, opened 2020-08-11
feat: Add IncGauge and DecGauge for metrics

topicus:
## Context
This change is a step toward the least-request algorithm implementation. 

## Description
Gauges and Counter are implemented differently on the metrics providers we support. Codahale allows to decrement counters but only allows to set values on gauges. Conversely, prometheus supports increment and decrement values on gauges but it doesn't allow decrement counters.

## Solution
After some research seems like the Prometheus semantic is more accurate than Codahale. As a consequence, Codahale implementation uses counters under-the-hood. 

I added the IncGauge and DecGauge methods to the metrics interface to hide these inconsistencies.

## Additional changes
For some reason, the incCounter implementation wraps the increment in a goroutine while this does not happen in the prometheus version. I tried to find documentation in the PR that introduced the change but I could not find it. IMHO spawning a new goroutine on every increment seems overkill.

Moreover, the internal implementation of then codahale library and it only use atomic operations which most likely are cheaper than creating a new gorotine although I am not completely sure about this. I decided to remove the goroutine wrapper in this PR. Let me know if there is a reason to keep it. If so, I can exclude it.

Signed-off-by: Mariano Carballal <marianocarballal@gmail.com>

### DIFF
--- a/metrics/all_kind.go
+++ b/metrics/all_kind.go
@@ -39,6 +39,14 @@ func (a *All) UpdateGauge(key string, v float64) {
 	a.prometheus.UpdateGauge(key, v)
 	a.codaHale.UpdateGauge(key, v)
 }
+func (a *All) IncGauge(key string) {
+	a.prometheus.IncGauge(key)
+	a.codaHale.IncGauge(key)
+}
+func (a *All) DecGauge(key string) {
+	a.prometheus.DecGauge(key)
+	a.codaHale.DecGauge(key)
+}
 func (a *All) MeasureRouteLookup(start time.Time) {
 	a.prometheus.MeasureRouteLookup(start)
 	a.codaHale.MeasureRouteLookup(start)
--- a/metrics/codahale.go
+++ b/metrics/codahale.go
@@ -114,6 +114,16 @@ func (c *CodaHale) UpdateGauge(key string, v float64) {
 	}
 }
 
+// Codahale supports decrements in counters but not on Gauges
+func (c *CodaHale) IncGauge(key string) {
+	c.incCounter(key, 1)
+}
+
+// Codahale supports decrements in counters but not on Gauges
+func (c *CodaHale) DecGauge(key string) {
+	c.decCounter(key, 1)
+}
+
 func (c *CodaHale) IncCounter(key string) {
 	c.incCounter(key, 1)
 }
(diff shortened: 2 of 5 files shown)


szuecs:
I would like more to have a Prometheus style solution. I think we should not use gauge in the first place but counters and queries. I try to get some people that understand Prometheus better than me to understand if we could use it as library with a bit of effort. I think least connections is not the only algorithm that we should want to have. See also https://brooker.co.za/blog/2012/01/17/two-random.html
I also recommend checking envoy source code to understand how others do this. 

My question would be how do you want to use (how do you query the data?) the data you write with inc() dec() gauge?

topicus:
> I would like more to have a Prometheus style solution. I think we should not use gauge in the first place but counters and queries.

The PR implementation is aligned to the Prometheus style. They don't provide a decrement method for counters because they consider (and most of the metrics libraries) counters only can increment.  However, they provide a decrement and increment method for gauges. 

> I think least connections is not the only algorithm that we should want to have. See also https://brooker.co.za/blog/2012/01/17/two-random.html
I also recommend checking envoy source code to understand how others do this.

I actually did that: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#weighted-least-request

In the PC2 of Envoy, the LB picks two endpoints randomly and then select the one with least request.

I agree we should not only implement least request. For the most efficient algorithms we need metrics to choose between endpoints. 

> My question would be how do you want to use (how do you query the data?) the data you write with inc() dec() gauge?

I discussed with @aryszka to store the metrics of each endpoint (so far just pending requests) in the metrics object and pass it to the LBAlgorithm by LBContext.

As a long term solution I think we should provide hooks (like Request and Response implemented by filters) in the proxy that will be triggered an each phase of the request-response lifecycle so cross-cutting concerns like metrics and tracing can be move out of the proxy implementation. Similar to [httptrace](https://golang.org/pkg/net/http/httptrace/)

topicus:
I think can use the metrics object as it prometheus driver is optimized for fast updates but not fast reads. (https://github.com/prometheus/client_golang/issues/486#issuecomment-433345239)

szuecs:
@topicus see https://github.com/zalando/skipper/blob/master/net/httpclient.go#L370 has already httptrace and maybe we can think of how to add the hooks and use the net.Client in the proxy package.

Gauge for each endpoint, I don't like.
How do you clean up the memory?

I rather would like to have a prometheus in memory implementation and we use just counters and `sum(rate())` queries. An example how we could wire things is https://github.com/logicalhan/instrumentation-tools which executes promql queries on http metrics endpoint like skipper has. Maybe this can be used and my idea is an optimization. Optimization: implement the TSDB store interface as in memory data structure with a ring buffer or so to limit max memory.

topicus:
Sounds good. Don't merge this PR. I would work in a PoC so we can test it before deciding.

szuecs:
Thanks @topicus !!

aryszka:
@topicus first thanks for the PR

I just wanted to add to the goroutine topic: it was a very early decision, based on very rudimentary throwaway local benchmarks, with a much earlier Go runtime version. Things have changed a lot, and also we learned a lot since then. It would make sense to test with more sophisticated load, or just production load, whether it is better to drop those additional goroutines. This should be separate effort, independent from the the least-request algorithm.

So, for the new metrics methods, we don't need to create new goroutines.

PR #1507, opened 2020-08-17
feat: Add power-of-N-choices load balancing algorithm

topicus:
## Description
The PR introduces the power of N choices algorithm for load balancing.

The algorithm works like this: first selects N elements from the list of endpoints attached to the route and then, from those N elements picks the one with least outstanding requests.

The algorithm work improves the load distribution compare to random and round robin when there are multiple load balancer instance that not share state. This is, however, the most common setup for Skipper.

## References 
- https://www.eecs.harvard.edu/~michaelm/postscripts/mythesis.pdf
- https://www.haproxy.com/blog/power-of-two-load-balancing/
- https://blog.envoyproxy.io/examining-load-balancing-algorithms-with-envoy-1be643ea121c
- https://netflixtechblog.com/netflix-edge-load-balancing-695308b5548c?gi=eb9e22cdebc4

## Tests 
I think we should refactor them as they are very unstable. Random and P2C algorithms distributes better with a big number of request which we can't reach due to max open files restrictions.

I will leave them broken until we come to a conclusion.

### DIFF
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -65,7 +65,7 @@ func (r *roundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 }
 
 type random struct {
-	rand *rand.Rand
+	rand  *rand.Rand
 	mutex sync.Mutex
 }
 
@@ -105,10 +105,10 @@ func (*consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 	return ctx.Route.LBEndpoints.At(choice)
 }
 
-type powerOfChoices struct{
-	rand *rand.Rand
+type powerOfChoices struct {
+	rand            *rand.Rand
 	numberOfChoices int
-	mutex sync.Mutex
+	mutex           sync.Mutex
 }
 
 func (a *powerOfChoices) GetScore(e routing.LBEndpoint) int {
@@ -121,7 +121,7 @@ func (a *powerOfChoices) GetScore(e routing.LBEndpoint) int {
 func newPowerOfChoices(endpoints []string) routing.LBAlgorithm {
 	t := time.Now().UnixNano()
 	return &powerOfChoices{
-		rand: rand.New(rand.NewSource(t)),
+		rand:            rand.New(rand.NewSource(t)),
 		numberOfChoices: 2, // TODO: make this the value part of skipper configuration.
 	}
 }
@@ -147,7 +147,7 @@ func (a *powerOfChoices) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 		return bestEndpoint
 	}
 	var currEndpoint routing.LBEndpoint
-	for i := 1 ; i < a.numberOfChoices ; i++  {
+	for i := 1; i < a.numberOfChoices; i++ {
 		candidateIdx = a.GetRandomIndex(numEndpoints)
 		currEndpoint = ctx.Route.LBEndpoints.At(candidateIdx)
 		if a.GetScore(currEndpoint) > a.GetScore(bestEndpoint) {
@@ -167,7 +167,6 @@ type (
 func NewAlgorithmProvider() routing.PostProcessor {
 	return &algorithmProvider{}
 }
-
 
 // AlgorithmFromString parses the string representation of the algorithm definition.
 func AlgorithmFromString(a string) (Algorithm, error) {
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -29,16 +29,16 @@ const (
 	// ConsistentHash indicates choice between the backends based on their hashed address.
 	ConsistentHash
 
-	// RandomNChoices selects N random endpoints and picks the one with least outstanding requests from them.
-	RandomNChoices
+	// PowerOfRandomNChoices selects N random endpoints and picks the one with least outstanding requests from them.
+	PowerOfRandomNChoices
 )
 
 var (
 	algorithms = map[Algorithm]initializeAlgorithm{
-		RoundRobin:     newRoundRobin,
-		Random:         newRandom,
-		ConsistentHash: newConsistentHash,
-		RandomNChoices: newRandomNChoices,
+		RoundRobin:            newRoundRobin,
+		Random:                newRandom,
+		ConsistentHash:        newConsistentHash,
+		PowerOfRandomNChoices: newPowerOfRandomNChoices,
 	}
 	defaultAlgorithm = newRoundRobin
 )
@@ -105,34 +105,34 @@ func (*consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 	return ctx.Route.LBEndpoints.At(choice)
 }
 
-type randomNChoices struct {
+type powerOfRandomNChoices struct {
 	rand            *rand.Rand
 	numberOfChoices int
 	mutex           sync.Mutex
 }
 
-func (a *randomNChoices) GetScore(e routing.LBEndpoint) int {
+func (a *powerOfRandomNChoices) GetScore(e routing.LBEndpoint) int {
 	// TODO: support more metrics. E.g., least connections.
 	// endpoints with higher inflight request should have lower score
 	return -e.Metrics.GetInflightRequests()
 }
 
-// newRandomNChoices selects N random backends and picks the one with less outstanding requests.
-func newRandomNChoices(endpoints []string) routing.LBAlgorithm {
+// newPowerOfRandomNChoices selects N random backends and picks the one with less outstanding requests.
+func newPowerOfRandomNChoices(endpoints []string) routing.LBAlgorithm {
 	t := time.Now().UnixNano()
-	return &randomNChoices{
+	return &powerOfRandomNChoices{
 		rand:            rand.New(rand.NewSource(t)),
 		numberOfChoices: 2, // TODO: make this the value part of skipper configuration.
 	}
 }
 
-func (a *randomNChoices) getRandomIndex(length int) int {
+func (a *powerOfRandomNChoices) getRandomIndex(length int) int {
 	a.mutex.Lock()
 	defer a.mutex.Unlock()
 	return a.rand.Intn(length)
 }
 
-func (a *randomNChoices) Apply(ctx *routing.LBContext) routing.LBEndpoint {
+func (a *powerOfRandomNChoices) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 	numEndpoints := ctx.Route.LBEndpoints.Length()
 	candidateIdx := a.getRandomIndex(numEndpoints)
 	bestEndpoint := ctx.Route.LBEndpoints.At(candidateIdx)
@@ -170,8 +170,8 @@ func AlgorithmFromString(a string) (Algorithm, error) {
 		return Random, nil
 	case "consistentHash":
 		return ConsistentHash, nil
-	case "randomNChoices":
-		return RandomNChoices, nil
+	case "powerOfRandomNChoices":
+		return PowerOfRandomNChoices, nil
 	default:
 		return None, errors.New("unsupported algorithm")
 	}
@@ -186,8 +186,8 @@ func (a Algorithm) String() string {
 		return "random"
 	case ConsistentHash:
 		return "consistentHash"
-	case RandomNChoices:
-		return "randomNChoices"
+	case PowerOfRandomNChoices:
+		return "powerOfRandomNChoices"
 	default:
 		return ""
 	}
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"hash/fnv"
 	"math/rand"
-	"net/url"
 	"sync"
 	"time"
 
@@ -29,13 +28,17 @@ const (
 
 	// ConsistentHash indicates choice between the backends based on their hashed address.
 	ConsistentHash
+
+	// PowerOfChoices selects N random endpoints and picks the one with least outstanding requests from them.
+	PowerOfChoices
 )
 
 var (
-	algorithms = map[Algorithm]initializeAgorithm{
+	algorithms = map[Algorithm]initializeAlgorithm{
 		RoundRobin:     newRoundRobin,
 		Random:         newRandom,
 		ConsistentHash: newConsistentHash,
+		PowerOfChoices: newPowerOfChoices,
 	}
 	defaultAlgorithm = newRoundRobin
 )
@@ -57,12 +60,13 @@ type roundRobin struct {
 func (r *roundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 	r.mx.Lock()
 	defer r.mx.Unlock()
-	r.index = (r.index + 1) % len(ctx.Route.LBEndpoints)
-	return ctx.Route.LBEndpoints[r.index]
+	r.index = (r.index + 1) % ctx.Route.LBEndpoints.Length()
+	return ctx.Route.LBEndpoints.At(r.index)
 }
 
 type random struct {
 	rand *rand.Rand
+	mutex sync.Mutex
 }
 
 func newRandom(endpoints []string) routing.LBAlgorithm {
@@ -72,7 +76,10 @@ func newRandom(endpoints []string) routing.LBAlgorithm {
 
 // Apply implements routing.LBAlgorithm with a stateless random algorithm.
 func (r *random) Apply(ctx *routing.LBContext) routing.LBEndpoint {
-	return ctx.Route.LBEndpoints[r.rand.Intn(len(ctx.Route.LBEndpoints))]
+	r.mutex.Lock()
+	index := r.rand.Intn(ctx.Route.LBEndpoints.Length())
+	r.mutex.Unlock()
+	return ctx.Route.LBEndpoints.At(index)
 }
 
 type consistentHash struct{}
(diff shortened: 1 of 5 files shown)


szuecs:
Max open files with Linux in CI?

loadbalancer/algorithm.go
```diff
@@ -89,19 +96,70 @@ func (*consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
-		choice = len(ctx.Route.LBEndpoints) + choice
+		choice = ctx.Route.LBEndpoints.Length() + choice
+	}
+	return ctx.Route.LBEndpoints.At(choice)
+}
+
+type powerOfChoices struct{
+	rand *rand.Rand
+	numberOfChoices int
+	mutex sync.Mutex
```
szuecs, line 112: We should add here a scoring function

loadbalancer/algorithm.go
```diff
@@ -89,19 +96,70 @@ func (*consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
+	for i := 0 ; i < a.numberOfChoices ; i++  {
+		candidate := a.GetRandomChoice(ctx.Route.LBEndpoints.Length())
+		// Pick a different endpoint if was already selected
+		for contains(chosen, candidate) {
+			candidate = a.GetRandomChoice(ctx.Route.LBEndpoints.Length())
+		}
+		chosen = append(chosen, candidate)
+	}
+	bestEndpoint := ctx.Route.LBEndpoints.At(chosen[0])
+	bestScore := bestEndpoint.Metrics.GetInflightRequests()
```
szuecs, line 149: These specific calls we can replace with our stored func in our struct.

proxy/proxy.go
```diff
@@ -871,7 +871,14 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
 		p.log.Errorf("could not map backend request, caused by: %v", err)
 		return nil, &proxyError{err: err}
 	}
-
+	if ctx.route.BackendType == eskip.LBBackend && ctx.route.LBEndpoints != nil {
+		// TODO: check req.URL vs req.Host
+		endpoint := ctx.route.LBEndpoints.Get(req.URL.Host)
+		endpoint.Metrics.IncInflightRequest()
+		defer func() {
```
szuecs, line 878: You can drop the wrapping func, just pretend function call with defer, if you need multi line defer it’s how you do it now

routing/routing.go
```diff
@@ -140,12 +142,86 @@ type RouteFilter struct {
 	Index int
 }
 
+// Contains metrics used by LB algorithms
```
szuecs, line 145: Godoc please: LBMetrics contain ...
Same style required for all functions below

routing/routing.go
```diff
@@ -140,12 +142,86 @@ type RouteFilter struct {
...
+
 // LBEndpoint represents the scheme and the host of load balanced
 // backends.
 type LBEndpoint struct {
 	Scheme, Host string
+	Metrics *LBMetrics
+}
+
+type LBEndpointCollection struct {
+	endpoints []LBEndpoint
```
szuecs, line 180: We could drop this and use `Itoa`

loadbalancer/algorithm.go
```diff
@@ -72,7 +76,10 @@ func newRandom(endpoints []string) routing.LBAlgorithm {
 
 // Apply implements routing.LBAlgorithm with a stateless random algorithm.
 func (r *random) Apply(ctx *routing.LBContext) routing.LBEndpoint {
-	return ctx.Route.LBEndpoints[r.rand.Intn(len(ctx.Route.LBEndpoints))]
+	r.mutex.Lock()
+	index := r.rand.Intn(ctx.Route.LBEndpoints.Length())
+	r.mutex.Unlock()
```
AlexanderYastrebov, line 81: I think unlock should be in done in defer otherwise mutex will stay locked if `Intn` panics.
`Intn` panics if argument is <= 0, here it assumes `LBEndpoints.Length()` is never zero.

proxy/proxy.go
```diff
@@ -871,7 +871,14 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
 		p.log.Errorf("could not map backend request, caused by: %v", err)
 		return nil, &proxyError{err: err}
 	}
-
+	if ctx.route.BackendType == eskip.LBBackend && ctx.route.LBEndpoints != nil {
+		// TODO: check req.URL vs req.Host
+		endpoint := ctx.route.LBEndpoints.Get(req.URL.Host)
```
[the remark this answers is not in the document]
topicus: You remember correctly.

loadbalancer/algorithm.go
topicus: Sounds good.

loadbalancer/algorithm.go
```diff
@@ -89,19 +96,70 @@ func (*consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
 	}
-	return ctx.Route.LBEndpoints[choice]
+	return false
+}
+
+func (a *powerOfChoices) Apply(ctx *routing.LBContext) routing.LBEndpoint {
+	chosen := make([]int, 0, a.numberOfChoices)
+	for i := 0 ; i < a.numberOfChoices ; i++  {
+		candidate := a.GetRandomChoice(ctx.Route.LBEndpoints.Length())
+		// Pick a different endpoint if was already selected
```
AlexanderYastrebov, line 142: The paper [The Power of Two Choices in Randomized Load Balancing](https://www.eecs.harvard.edu/~michaelm/postscripts/mythesis.pdf) explores the algorithm based on the assumption that _each ball is placed sequentially into the least full of d bins chosen independently and uniformly at random_ (ball is the request and bin is the route in case of skipper). Here the choice of bins (routes) is not independent.

Moreover, consider the corner case n=1 (only one route available) and d=2 (hardcoded two draws): in this case `for contains(chosen, candidate)` becomes infinite loop on the second draw.

With the above said, proposed implementation could be simplified to draw routes independently and would require O(1) memory and O(d) cpu instead of O(d) memory and O(d^d) cpu needed to remember previous choices.

It also seems to be a good idea to extend `Random` algorithm with a draw count parameter like [it was done in HAProxy](https://github.com/haproxy/haproxy/commit/21c741a665f) instead of introducing new algorithm.

topicus: Very good observation. I missed the `independently and uniformly` part of the algorithm. I am fixing the issue and trying to unify it to the random implementation.

loadbalancer/algorithm.go
topicus: Added.

loadbalancer/algorithm.go
topicus: Done.

proxy/proxy.go
topicus: Done.

routing/routing.go
topicus: Done.

loadbalancer/algorithm.go
topicus: Done.

loadbalancer/algorithm.go
topicus: I fixed the algo however unifying random and power of N choices could be address in a separate PR as it requires to change the initialisation mechanism for LB algorithms. Currently there is no way to parametrise algorithms other than with endpoints.

routing/routing.go
topicus: But for this I would need to keep somehow the order of the elements.

loadbalancer/algorithm.go
```diff
@@ -89,19 +95,71 @@ func (*consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
+// newPowerOfChoices selects N random backends and picks the one with less outstanding requests.
+func newPowerOfChoices(endpoints []string) routing.LBAlgorithm {
+	t := time.Now().UnixNano()
+	return &powerOfChoices{
+		rand:            rand.New(rand.NewSource(t)),
+		numberOfChoices: 2, // TODO: make this the value part of skipper configuration.
+	}
+}
+
+func (a *powerOfChoices) GetRandomIndex(length int) int {
```
AlexanderYastrebov, line 129: Is there a reason to make it public?

loadbalancer/algorithm.go
```diff
@@ -89,19 +95,71 @@ func (*consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
+	a.mutex.Lock()
+	defer a.mutex.Unlock()
+	return a.rand.Intn(length)
+}
+
+func (a *powerOfChoices) Apply(ctx *routing.LBContext) routing.LBEndpoint {
+	numEndpoints := ctx.Route.LBEndpoints.Length()
+
+	// no need to compute a random endpoint if only one is given
+	if numEndpoints == 1 && a.numberOfChoices == 1 {
```
AlexanderYastrebov, line 139: `a.numberOfChoices == 1` is redundant - there is no need to draw random if there is only one endpoint

loadbalancer/algorithm.go
```diff
@@ -89,19 +95,71 @@ func (*consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
+		return ctx.Route.LBEndpoints.At(0)
+	}
+	candidateIdx := a.GetRandomIndex(numEndpoints)
+	bestEndpoint := ctx.Route.LBEndpoints.At(candidateIdx)
+
+	// no need to compute the scores when only one endpoint can be chosen
+	if a.numberOfChoices == 1 {
```
AlexanderYastrebov, line 146: This check is also redundant as it will be performed by for loop

loadbalancer/algorithm.go
```diff
@@ -89,19 +95,71 @@ func (*consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
+		return bestEndpoint
+	}
+	var currEndpoint routing.LBEndpoint
```
AlexanderYastrebov, line 149: Should it be moved into the loop to reduce scope?

loadbalancer/algorithm.go
topicus: Not at all, hiding it.

loadbalancer/algorithm.go
```diff
@@ -89,19 +95,71 @@ func (*consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
-	choice := int(sum) % len(ctx.Route.LBEndpoints)
+	choice := int(sum) % ctx.Route.LBEndpoints.Length()
 	if choice < 0 {
-		choice = len(ctx.Route.LBEndpoints) + choice
+		choice = ctx.Route.LBEndpoints.Length() + choice
+	}
+	return ctx.Route.LBEndpoints.At(choice)
+}
+
+type powerOfChoices struct {
```
AlexanderYastrebov, line 108: I think "power" in the  "The Power of Two Choices in Randomized Load Balancing" implies strength of two or N random choices over single one.

I personally would go with something casual like `randomN` or in this particular case even with `random2` as number of choices is hardcoded.

loadbalancer/algorithm.go
topicus: I am not sure how the routes are created in skipper ingress. If could happen a created route would have a single endpoint. Also, what happens if an endpoint is marked as sick and the route starts to have a single endpoint.
In any case, this is a micro-op. I removed that for the sake of readability.

loadbalancer/algorithm.go
topicus: You are right.

loadbalancer/algorithm.go
AlexanderYastrebov: Note that I am a bystander here, not a codeowner so consider this a nitpick :)

topicus:
> The algorithm can have a scoring function that gets an interface type as arg that implements the GetMetric() which in the current case returns the inflight requests. Like this we can create more scoring functions that would implement maybe latency based choice.

@szuecs  This is not easy to achieve without refactoring the signature of the initialize function. In other words, what would be the noobs to control the algorithm behavior. This is also related to the number of choices.

Since this is just a PoC, I would leave that for a second round when we are sure it makes sense to invest more time on the implementation.

What would be the process to test this with traffic?

loadbalancer/algorithm.go
topicus: I was wondering to change it to `powerOfNChoices` or `randomNChoices`. I prepare the code for experimenting with the number of choices. I would like to keep the word `choices` as it is a popular word reference for the algorithm.

loadbalancer/algorithm.go
AlexanderYastrebov: I mean `if numEndpoints == 1` is fine but `&& a.numberOfChoices == 1` is redundant - it does not make sense to draw random N times if you can pick only from one endpoint.
 
> In any case, this is a micro-op. I removed that for the sake of readability.

This is the best outcome in my opinion :)

loadbalancer/algorithm.go
topicus: I renamed it to PowerOfRandomNChoices. It is better to be explicit. If people want to research it is a googleable.

routing/routing.go
```diff
@@ -140,10 +142,83 @@ type RouteFilter struct {
 	Index int
 }
 
+// LBMetrics contains metrics used by LB algorithms
+type LBMetrics struct {
+	inflightRequests int
```
AlexanderYastrebov, line 147: Hard to tell for sure without a benchmark but I suspect that using atomic would be faster then mutex, see 
* https://golang.org/pkg/sync/atomic/#AddInt32
* https://stackoverflow.com/questions/47445344/is-there-a-difference-in-go-between-a-counter-using-atomic-operations-and-one-us

routing/routing.go
```diff
@@ -140,10 +142,83 @@ type RouteFilter struct {
...
+		if err != nil {
+			return nil, err
+		}
+		collection.Set(eu.Host, LBEndpoint{Scheme: eu.Scheme, Host: eu.Host, Metrics: &LBMetrics{}})
+	}
+	return collection, nil
+}
+
+// Stores an endpoint indexed by key
+func (e *LBEndpointCollection) Set(key string, item LBEndpoint) {
```
AlexanderYastrebov, line 204: As `LBEndpointCollection` seems to be immutable datastructure, should this be private or even inlined?

loadbalancer/algorithm.go
AlexanderYastrebov: > If people want to research it is a googleable.

Or just add a link to the paper in the comment :)

routing/routing.go
topicus: Yes, I though about using atomics and it is likely I will, but I wanted to do it once we test the algorithm with some shadow traffic. I don't think the impact will be huge although it might be meaningful. Let's see if I can allocate some time tomorrow.

routing/routing.go
topicus: Totally, I just replaced the existing data structure but I will change this tomorrow.

loadbalancer/algorithm.go
topicus: I will also do that.

proxy/proxy.go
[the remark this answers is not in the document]
AlexanderYastrebov: Alternatively chosen endpoint could be returned from `mapRequest` to avoid creation of a new `LBEndpointCollection` datastructure.

proxy/proxy.go
```diff
@@ -871,7 +871,11 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
 		p.log.Errorf("could not map backend request, caused by: %v", err)
 		return nil, &proxyError{err: err}
 	}
-
+	if ctx.route.BackendType == eskip.LBBackend && ctx.route.LBEndpoints != nil {
+		endpoint := ctx.route.LBEndpoints.Get(req.URL.Host)
```
AlexanderYastrebov, line 875: Alternatively chosen endpoint could be returned from mapRequest to avoid creation of a new `LBEndpointCollection` datastructure.

topicus: I agree this is not elegant. We discussed a bunch of options regarding how to handle the metrics collection for LBs with @aryszka . I tried to keep the code as less intrusive as possible otherwise, the refactoring would take so much.

routing/routing.go
```diff
@@ -140,10 +141,76 @@ type RouteFilter struct {
 	Index int
 }
 
+// LBMetrics contains metrics used by LB algorithms
+type LBMetrics struct {
+	inflightRequests int32
```
AlexanderYastrebov, line 146: Why not `int64`?

proxy/proxy.go
AlexanderYastrebov: > ... the refactoring would take so much

```diff
diff --git a/proxy/proxy.go b/proxy/proxy.go
index 58f7115..f65fedb 100644
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -470,22 +470,20 @@ func setRequestURLForDynamicBackend(u *url.URL, stateBag map[string]interface{})
        }
 }
 
-func setRequestURLForLoadBalancedBackend(u *url.URL, rt *routing.Route, lbctx *routing.LBContext) {
-       e := rt.LBAlgorithm.Apply(lbctx)
-       u.Scheme = e.Scheme
-       u.Host = e.Host
-}
-
 // creates an outgoing http request to be forwarded to the route endpoint
 // based on the augmented incoming request
-func mapRequest(r *http.Request, rt *routing.Route, host string, removeHopHeaders bool, stateBag map[string]interface{}) (*http.Request, error) {
+func mapRequest(r *http.Request, rt *routing.Route, host string, removeHopHeaders bool, stateBag map[string]interface{}) (*http.Request, *routing.LBEndpoint, error) {
+       var endpoint *routing.LBEndpoint
        u := r.URL
        switch rt.BackendType {
        case eskip.DynamicBackend:
                setRequestURLFromRequest(u, r)
                setRequestURLForDynamicBackend(u, stateBag)
        case eskip.LBBackend:
-               setRequestURLForLoadBalancedBackend(u, rt, routing.NewLBContext(r, rt))
+               e := rt.LBAlgorithm.Apply(routing.NewLBContext(r, rt))
+               u.Scheme = e.Scheme
+               u.Host = e.Host
+               endpoint = &e
        default:
                u.Scheme = rt.Scheme
                u.Host = rt.Host
@@ -498,7 +496,7 @@ func mapRequest(r *http.Request, rt *routing.Route, host string, removeHopHeader
 
        rr, err := http.NewRequest(r.Method, u.String(), body)
        if err != nil {
-               return nil, err
+               return nil, endpoint, err
        }
 
        rr = rr.WithContext(r.Context())
@@ -527,7 +525,7 @@ func mapRequest(r *http.Request, rt *routing.Route, host string, removeHopHeader
                rr = rr.WithContext(ot.ContextWithSpan(rr.Context(), ctxspan))
        }
 
-       return rr, nil
+       return rr, endpoint, nil
 }
 
 func forwardToProxy(incoming, outgoing *http.Request) {
@@ -866,13 +864,12 @@ func (p *Proxy) makeUpgradeRequest(ctx *context, req *http.Request) error {
 
 func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
        var err error
-       req, err := mapRequest(ctx.request, ctx.route, ctx.outgoingHost, p.flags.HopHeadersRemoval(), ctx.StateBag())
+       req, endpoint, err := mapRequest(ctx.request, ctx.route, ctx.outgoingHost, p.flags.HopHeadersRemoval(), ctx.StateBag())
        if err != nil {
                p.log.Errorf("could not map backend request, caused by: %v", err)
                return nil, &proxyError{err: err}
        }
-       if ctx.route.BackendType == eskip.LBBackend && ctx.route.LBEndpoints != nil {
-               endpoint := ctx.route.LBEndpoints.Get(req.URL.Host)
+       if endpoint != nil {
                endpoint.Metrics.IncInflightRequest()
                defer endpoint.Metrics.DecInflightRequest()
        }
@@ -1129,7 +1126,7 @@ func (p *Proxy) do(ctx *context) error {
                ctx.setResponse(loopCTX.response, p.flags.PreserveOriginal())
                ctx.proxySpan = loopCTX.proxySpan
        } else if p.flags.Debug() {
-               debugReq, err := mapRequest(ctx.request, ctx.route, ctx.outgoingHost, p.flags.HopHeadersRemoval(), ctx.StateBag())
+               debugReq, _, err := mapRequest(ctx.request, ctx.route, ctx.outgoingHost, p.flags.HopHeadersRemoval(), ctx.StateBag())
                if err != nil {
                        return &proxyError{err: err}
                }
```

AlexanderYastrebov: The change from above will not work because Algorithm.Apply returns endpoint value so later modification of inflight request counter would modify the copy.

routing/routing.go
topicus: I will change this.

proxy/proxy.go
topicus: Today I am testing the algorithm with some synthetic load tests. The diff you posted looks good to me. I think we can applied once I prove it is valuable and brings value.

szuecs:
@topicus you can get access to our dev cluster and use our ghe project loadtest (or similar name). You need just create a or and feel free to change the ingress/RouteGroup and scale deployments that are also called loadtest-*.
@aryszka can help to figure out the name.

routing/routing.go
```diff
@@ -140,10 +141,76 @@ type RouteFilter struct {
...
+			return nil, err
+		}
+		collection.Set(eu.Host, LBEndpoint{Scheme: eu.Scheme, Host: eu.Host, Metrics: &LBMetrics{}})
+	}
+	return collection, nil
+}
+
+// Stores an endpoint indexed by key
+func (e *lbEndpointCollection) Set(key string, item LBEndpoint) {
+	e.endpoints = append(e.endpoints, item)
```
aryszka, line 197: this potentially can lead to inconsistent internal state, when the key already exists in the map. If they key exists, we should find the currently stored item's index in the slice, and replace the item in the slice instead of appending.

Another approach, and maybe better, would be if we didn't export the Set method, and keep the collection immutable, only populated by the NewEndpointCollection initializer.

WDYT?

loadbalancer/algorithm.go
```diff
@@ -89,19 +95,61 @@ func (*consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
+}
+
+// Paper http://www.eecs.harvard.edu/~michaelm/NEWWORK/postscripts/twosurvey.pdf
+type powerOfRandomNChoices struct {
+	rand            *rand.Rand
+	numberOfChoices int
+	mutex           sync.Mutex
+}
+
+func (a *powerOfRandomNChoices) GetScore(e routing.LBEndpoint) int {
```
aryszka, line 115: we don't need to export the GetScore method, it is part of the algorithm. Please rename it to getScore

aryszka (changes requested):

aryszka:
@topicus looks good to me, I only have those minor requests in the comments.

@AlexanderYastrebov thanks for the review!

loadbalancer/algorithm.go
```diff
@@ -89,19 +95,61 @@ func (*consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
+	// TODO: support more metrics. E.g., least connections.
+	// endpoints with higher inflight request should have lower score
+	return -e.Metrics.GetInflightRequests()
```
addityasingh, line 118: @topicus Is there any specific reason to expose the inflight requests as negative integers? What about 0?

szuecs: @addityasingh see comment above and line 143

addityasingh: I understand the comparison is correct but I am specifically interested in understanding why comparing negative integers for inflight request count

szuecs: Because if you read the code in Algorithm the comparison (L143) looks nicer imo.

ruiaraujo:
fixed by #1548

PR #1512, opened 2020-08-26
add state bag to tag filter

### LINKED ISSUE #1510: add stateBagToTag filter

I would like to identify callers in open tracing.

I know that this should be possible just by analyzing the parent spans, but this is not very useful due to

1. vendor tooling which makes this inconvenient
2. skipper itself being often being shown as the caller

Hence the idea to convert the calling user to a tag.
The calling user is already present in the "auth-user" state bag.

see last question.

Yes

zeitlinger:
Fixes #1510

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -121,7 +121,7 @@ Parameters:
 * header name (string)
 * key in the state bag (string)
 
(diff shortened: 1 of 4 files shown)


aryszka:
:+1:

zeitlinger:
:+1:

PR #1513, opened 2020-08-26
Adds watch_test.go error handling

AlexanderYastrebov:
Adds test file error handling to simplify [flaky watch_test.go](https://github.com/zalando/skipper/runs/1028484357?check_suite_focus=true) debugging when it fails next time as it is hard to reproduce the failure reliably.

### DIFF
--- a/eskipfile/watch_test.go
+++ b/eskipfile/watch_test.go
@@ -36,30 +36,36 @@ type watchTest struct {
 	routing *routing.Routing
 }
 
-func deleteFile() {
-	os.RemoveAll(testWatchFile)
+func deleteFile(t *testing.T) {
+	err := os.Remove(testWatchFile)
+	if err != nil {
+		t.Logf("Ignoring %v", err)
+	}
 }
 
-func createFileWith(content string) {
+func createFileWith(content string, t *testing.T) {
 	f, err := os.Create(testWatchFile)
 	if err != nil {
-		return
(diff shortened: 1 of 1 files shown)


eskipfile/watch_test.go
```diff
@@ -36,30 +36,36 @@ type watchTest struct {
...
 	f, err := os.Create(testWatchFile)
 	if err != nil {
-		return
+		t.Fatal(err)
 	}
 
 	defer f.Close()
-	f.Write([]byte(content))
+	_, errw := f.WriteString(content)
+	if errw != nil {
```
szuecs, line 54: You can also do 2x and reuse the var name `err`:
```
if _, err := f.WriteString(content); err != nil {
```

AlexanderYastrebov: Changed to `writeErr`.

szuecs: I would prefer just `err`. If you use the mentioned line, then `err := ...` can be used more than once in a function.

AlexanderYastrebov: What is the point to reuse variable name (when we have more names available than there're atoms in universe)? I know that it is possible to redefine the name in if block but in this case it is only possible because first return value is ignored.
If you really _prefer_ them both to read the same - they are now ;)

aryszka:
:+1:

szuecs:
:+1:

PR #1515, opened 2020-08-27
Validate compression level in compress filter by comparing integers

wndhydrnt:
On a cluster with a lot of routes where many of those routes activate the `compress()` filter, skipper uses significantly more CPU on route updates.

We saw an increase of 100% - 400% of CPU usage during route updates, e.g. a Kubernetes worker joining or leaving or a new Deployment resource being created. Sometimes skipper pods also started to OOM.
This issue can be traced down to the check for validity of the compression level in the spec of the `compress()` filter.

This change replaces the creation of writers to check for validity of the compression level with comparing the levels directly without creating a writer for each type of compression.

### DIFF
--- a/filters/builtin/compress.go
+++ b/filters/builtin/compress.go
@@ -140,13 +140,11 @@ func (c *compress) CreateFilter(args []interface{}) (filters.Filter, error) {
 
 	if lf, ok := args[0].(float64); ok && math.Trunc(lf) == lf {
 		f.level = int(lf)
-		_, err := gzip.NewWriterLevel(nil, f.level)
-		if err != nil {
+		if f.level < gzip.HuffmanOnly || f.level > gzip.BestCompression {
 			return nil, filters.ErrInvalidFilterParameters
 		}
 
-		_, err = flate.NewWriter(nil, f.level)
-		if err != nil {
+		if f.level < flate.HuffmanOnly || f.level > flate.BestCompression {
 			return nil, filters.ErrInvalidFilterParameters
 		}
 


filters/builtin/compress.go
```diff
@@ -140,13 +140,11 @@ func (c *compress) CreateFilter(args []interface{}) (filters.Filter, error) {
 
 	if lf, ok := args[0].(float64); ok && math.Trunc(lf) == lf {
 		f.level = int(lf)
-		_, err := gzip.NewWriterLevel(nil, f.level)
-		if err != nil {
+		if f.level < gzip.HuffmanOnly || f.level > gzip.BestCompression {
```
AlexanderYastrebov, line 143: :+1: This is what [NewWriterLevel is doing](https://golang.org/src/compress/gzip/gzip.go?s=1836:1896#L50) internally

AlexanderYastrebov: OTOH filter documentation states:
> The possible values are integers between 0 and 9 (inclusive), where 0 means no-compression, 1 means best-speed and 9 means best-compression.

so the alternative would be to simply use 0 and 9 respectively (i.e. disable undocumented -1 aka DefaultCompression and -2 aka HuffmanOnly)

szuecs:
@wndhydrnt this sounds more like flapping route problem, in case you tested this PR and it fixed the problem.
Why? Because `CreateFilter` runs only in case of a create/update operation of a route.

Do you see that the initialization of gzip/deflate is using a lot of CPU in your case?

wndhydrnt:
Hello @szuecs 
I reproduced it like this:
1. Deploy v0.11.141. Result: CPU spikes appear each routes change, ranging from small spikes to big spikes to OOM.
2. Deploy a version of skipper where the code that checks the compression level is commented out. Result: No spikes in CPU usage anymore, no OOM.

We have a lot of route updates during the day due to deployments and EC2 spot instance terminations.

filters/builtin/compress.go
aryszka: interestingly, it's indeed what the gzip.NewWriterLevel does, but it's not the case with the flate.NewWriter. Traced this far: https://golang.org/src/compress/flate/huffman_bit_writer.go#L93

aryszka:
Thanks for the PR @wndhydrnt !

aryszka:
:+1:

aryszka:
thanks @AlexanderYastrebov for pointing out the difference with the documentation. We should fix it either mentioning the possibility for the default, or as you suggested, validating for 0..9, which I prefer more.

Fun fact, the Huffman check in the stdlib was added only a couple of months after the validation in skipper, but skipper was already wrong in comparison to its docs because of the default compression :smile: 
(https://github.com/zalando/skipper/commit/afb280b6576287cb9fa6f0283fae81c95912b0cf#diff-b456713d789e597a413279e2a6d97c78R151 and https://github.com/golang/go/commit/c9517b1ffe7c99a3fd2a748fab51645bd674ad75#diff-c6ddb466414bc294f416623ffb37af83R60)

aryszka:
https://github.com/zalando/skipper/issues/1517

filters/builtin/compress.go
wndhydrnt: I found this [function](https://golang.org/src/compress/flate/deflate.go#L578) which does the validation of the levels in a `switch` statement. That seemed a bit too elaborate for what the filterspec wants to achieve so I went for the simpler version from the `gzip` package.

I'm a bit confused: Do you want me to change the validation so that the filterspec only accepts 0 - 9 as valid inputs or will this be taken care of in #1517 ?
I can do the change, no problem, just tell me.

AlexanderYastrebov: @wndhydrnt @aryszka Yes, flate uses the aforementioned switch that does the same as far as I can tell (i.e. accepts valid level in range [-2, 9], see default case) 

I hardly can image how users could come up with using -2 or -1 as they are not documented and docs make no connection to go's implementation constants.
So I personally would narrow acceptable values to [0, 9] as documented and close followup #1517.
Do we consider fixing an inconsistency/undocumented behavior a breaking change?

gargravarr:
:+1:

szuecs:
@wndhydrnt thanks and interesting case!

PR #1519, opened 2020-09-11
Logging requests to webhook during debug logging

muaazsaleem:
To help debug ValidatingWebhookConfiguration, logging requests to the
webhook. Note that this will only be requests from the API server since
the webhook is not accessible otherwise.

Signed-off-by: Muaaz Saleem <muhammad.muaaz.saleem@zalando.de>


muaazsaleem:
👍

PR #1524, opened 2020-09-24
mask users in tracing tags

zeitlinger:
mask users in tracing tags

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -149,6 +150,7 @@ type Config struct {
 	OidcSecretsFile                 string        `yaml:"oidc-secrets-file"`
 	CredentialPaths                 *listFlag     `yaml:"credentials-paths"`
 	CredentialsUpdateInterval       time.Duration `yaml:"credentials-update-interval"`
+	MaskRealms                      string        `yaml:"mask-realms"`
 
 	// TLS client certs
 	ClientKeyFile  string            `yaml:"client-tls-key"`
@@ -240,6 +242,7 @@ const (
 	defaultOAuthTokenintrospectionTimeout = 2 * time.Second
 	defaultWebhookTimeout                 = 2 * time.Second
 	defaultCredentialsUpdateInterval      = 10 * time.Minute
+	defaultMaskRealms                     = "employees"
 
 	// API Monitoring
 	defaultApiUsageMonitoringRealmKeys                    = ""
@@ -355,6 +358,7 @@ const (
 	oidcSecretsFileUsage                 = "file storing the encryption key of the OID Connect token"
 	credentialPathsUsage                 = "directories or files to watch for credentials to use by bearerinjector filter"
 	credentialsUpdateIntervalUsage       = "sets the interval to update secrets"
+	maskRealmsUsage                      = "sets the pattern for OAuth realms which should be masked in logging and tracing"
 
 	// TLS client certs
 	clientKeyFileUsage  = "TLS Key file for backend connections, multiple keys may be given comma separated - the order must match the certs"
@@ -541,6 +545,7 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.OidcSecretsFile, "oidc-secrets-file", "", oidcSecretsFileUsage)
 	flag.Var(cfg.CredentialPaths, "credentials-paths", credentialPathsUsage)
 	flag.DurationVar(&cfg.CredentialsUpdateInterval, "credentials-update-interval", defaultCredentialsUpdateInterval, credentialsUpdateIntervalUsage)
+	flag.StringVar(&cfg.MaskRealms, "mask-realms", defaultMaskRealms, maskRealmsUsage)
 
 	// TLS client certs
 	flag.StringVar(&cfg.ClientKeyFile, "client-tls-key", "", clientKeyFileUsage)
@@ -795,6 +800,7 @@ func (c *Config) ToOptions() skipper.Options {
 		OIDCSecretsFile:                c.OidcSecretsFile,
 		CredentialsPaths:               c.CredentialPaths.values,
 		CredentialsUpdateInterval:      c.CredentialsUpdateInterval,
+		MaskRealms:                     regexp.MustCompile(c.MaskRealms),
 
 		// connections, timeouts:
 		WaitForHealthcheckInterval:   c.WaitForHealthcheckInterval,
--- a/filters/auth/tokeninfo.go
+++ b/filters/auth/tokeninfo.go
@@ -2,12 +2,15 @@ package auth
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
 	"github.com/opentracing/opentracing-go"
 	log "github.com/sirupsen/logrus"
+
 	"github.com/zalando/skipper/filters"
+	logfilter "github.com/zalando/skipper/filters/log"
 )
 
 const (
@@ -23,6 +26,7 @@ type TokeninfoOptions struct {
 	Timeout      time.Duration
 	MaxIdleConns int
 	Tracer       opentracing.Tracer
+	MaskRealms   *regexp.Regexp
 }
 
 type (
@@ -36,6 +40,7 @@ type (
 		authClient *authClient
 		scopes     []string
 		kv         kv
+		maskRealms *regexp.Regexp
 	}
 )
 
@@ -183,7 +188,7 @@ func (s *tokeninfoSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 		tokeninfoAuthClient[s.options.URL] = ac
 	}
 
-	f := &tokeninfoFilter{typ: s.typ, authClient: ac, kv: make(map[string][]string)}
+	f := &tokeninfoFilter{typ: s.typ, authClient: ac, kv: make(map[string][]string), maskRealms: s.options.MaskRealms}
 	switch f.typ {
 	// all scopes
 	case checkOAuthTokeninfoAllScopes:
@@ -353,6 +358,12 @@ func (f *tokeninfoFilter) Request(ctx filters.FilterContext) {
 	}
 
 	authorized(ctx, uid)
+
+	realm, _ := authMap[realmKey].(string)
+	if f.maskRealms.MatchString(realm) {
+		ctx.StateBag()[logfilter.MaskedAuthUserKey] = realm
+	}
+
 	ctx.StateBag()[tokeninfoCacheKey] = authMap
 }
 
--- a/filters/tracing/statebagtotag.go
+++ b/filters/tracing/statebagtotag.go
@@ -6,6 +6,7 @@ import (
 	"github.com/opentracing/opentracing-go"
 
 	"github.com/zalando/skipper/filters"
+	logfilter "github.com/zalando/skipper/filters/log"
 )
 
 const (
@@ -57,6 +58,15 @@ func (f stateBagToTagFilter) Request(ctx filters.FilterContext) {
 	if span == nil {
 		return
 	}
+
+	if f.stateBagItemName == logfilter.AuthUserKey {
+		value, ok := ctx.StateBag()[logfilter.MaskedAuthUserKey]
+		if ok {
+			span.SetTag(f.tagName, value.(string))
+			return
+		}
+	}
+
 	value, ok := ctx.StateBag()[f.stateBagItemName]
 	if !ok {
 		return
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -72,6 +72,7 @@ func Test_NewConfig(t *testing.T) {
 				WebhookTimeout:                          2 * time.Second,
 				CredentialPaths:                         commaListFlag(),
 				CredentialsUpdateInterval:               10 * time.Minute,
+				MaskRealms:                              "users",
 				ApiUsageMonitoringClientKeys:            "sub",
 				ApiUsageMonitoringRealmsTrackingPattern: "services",
 				WaitForHealthcheckInterval:              45 * time.Second,
--- a/filters/auth/auth.go
+++ b/filters/auth/auth.go
@@ -55,6 +55,7 @@ const (
 	// tokenKey defined at https://tools.ietf.org/html/rfc7662#section-2.1
 	tokenKey = "token"
 	scopeKey = "scope"
+	realmKey = "realm"
 	uidKey   = "uid"
 )
 
--- a/filters/log/log.go
+++ b/filters/log/log.go
@@ -26,6 +26,8 @@ const (
 	// information into the state bag to pass the information to
 	// the auditLog filter.
 	AuthUserKey = "auth-user"
+	// MaskedAuthUserKey is the same as AuthUserKey, but some realms are masked
+	MaskedAuthUserKey = "masked-auth-user"
 	// AuthRejectReasonKey is used by the auth package to set the
 	// reject reason information into the state bag to pass the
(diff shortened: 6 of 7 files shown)


filters/auth/tokeninfo.go
```diff
@@ -353,6 +358,12 @@ func (f *tokeninfoFilter) Request(ctx filters.FilterContext) {
 	}
 
 	authorized(ctx, uid)
+
+	realm, _ := authMap[realmKey].(string)
+	if f.maskRealms.MatchString(realm) {
```
szuecs, line 363: What if this is nil?
If nil we should not branch into the masking code.

config/config.go
```diff
@@ -541,6 +545,7 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.OidcSecretsFile, "oidc-secrets-file", "", oidcSecretsFileUsage)
 	flag.Var(cfg.CredentialPaths, "credentials-paths", credentialPathsUsage)
 	flag.DurationVar(&cfg.CredentialsUpdateInterval, "credentials-update-interval", defaultCredentialsUpdateInterval, credentialsUpdateIntervalUsage)
+	flag.StringVar(&cfg.MaskRealms, "mask-realms", defaultMaskRealms, maskRealmsUsage)
```
szuecs, line 548: I don't think we should change the current default.

config/config.go
```diff
@@ -240,6 +242,7 @@ const (
 	defaultOAuthTokenintrospectionTimeout = 2 * time.Second
 	defaultWebhookTimeout                 = 2 * time.Second
 	defaultCredentialsUpdateInterval      = 10 * time.Minute
+	defaultMaskRealms                     = "employees"
```
szuecs, line 245: We should not have a default

tkrop:
Not sure whether masking of users in logs is necessary. Afaik, the privacy setup for logs should usually be sufficient to allow logging of user names.

Anyhow, why didn't you add the masking directly in the filter?

szuecs (changes requested):
- Please handle nil case and use this as default 
- What about oidc* filters?
- What about tokenintrospection filters?

I think the other filters should be done in the same PR, such that we have feature parity and not friction (not sure if these apply but IIRC yes).

zeitlinger:
> Not sure whether masking of users in logs is necessary. Afaik, the privacy setup for logs should usually be sufficient to allow logging of user names.

this PR is not about logs

> Anyhow, why didn't you add the masking directly in the filter?

I did (as much as possible): https://github.com/zalando/skipper/pull/1524/files#diff-50793c74bbe4ccdd0d38b7b8296a0860R62

zeitlinger:
> * Please handle nil case and use this as default

yes, that's the idea

> * What about oidc* filters?
> * What about tokenintrospection filters?
> 
> I think the other filters should be done in the same PR, such that we have feature parity and not friction (not sure if these apply but IIRC yes).

yes, I'm still wondering how to do it - I didn't find a realm in https://tools.ietf.org/html/rfc7662#section-2.2. Can you help?

filters/auth/tokeninfo.go
```diff
@@ -353,6 +358,12 @@ func (f *tokeninfoFilter) Request(ctx filters.FilterContext) {
...
+		ctx.StateBag()[logfilter.MaskedAuthUserKey] = realm
```
AlexanderYastrebov, line 364: This looks wrong according to 
`// MaskedAuthUserKey is the same as AuthUserKey, but some realms are masked`

What do you think about pivoting design in the following way:
1. Use allowlist instead of masklist - i.e. user explicitly specifies for which realms auth user should be set on the span
2. Is there a real necessity to use regexp for realm match? How may realms are there in total afterall (or is it a way to specify a list `(employee|whatever)` maybe)?
3. Get rid of conditional logic here but rather store `realm` along with `auth-user` in statebag (in `auth.go` I guess) and move all the decisions to the filter(s).

AlexanderYastrebov: > yes, I'm still wondering how to do it - I didn't find a realm in https://tools.ietf.org/html/rfc7662#section-2.2. Can you help?

`auth.go` used by all filters to set `auth-user` at the moment is only aware about `username`. One way out may be to extend it and make it also aware of a new concept, say `auth-realm`. Filters that do not support realm may set it to empty or to some predefined constant placeholder.

zeitlinger: > 1. Use allowlist instead of masklist - i.e. user explicitly specifies for which realms auth user should be set on the span. If not set, realm is not considered.
> 2. Is there a real necessity to use regexp for realm match? How may realms are there in total afterall (or is it a way to specify a list `(employee|whatever)` maybe)?

no strong reason from my side.
I chose this approach after a discussion with @aryszka that masking values might be a concept that can be re-used in other places.

> 3. Get rid of conditional logic here but rather store `realm` along with `auth-user` in statebag (in `auth.go` I guess) and move all the decisions to the `statebagtotag` filter. This would allows building similar logic based on realm in other filters as well.

sounds good

tkrop:
@zeitlinger `realm` is not in the spec because it is a Zalando extension, since we have `users`, `services`, and at lease one realm for *partners*. Checkout: `ztoken | sed -e "s/\(.*\)\.\(.*\)\.\(.*\)/\2/" | base64 -d | jq '.'`

filters/tracing/statebagtotag.go
```diff
@@ -57,6 +58,15 @@ func (f stateBagToTagFilter) Request(ctx filters.FilterContext) {
 	if span == nil {
 		return
 	}
+
+	if f.stateBagItemName == logfilter.AuthUserKey {
```
AlexanderYastrebov, line 62: Is this special treatment really necessary? 
Won't
```
stateBagToTag("masked-auth-user", "client_id")
```
just work?

filters/auth/tokeninfo.go
```diff
@@ -353,6 +362,14 @@ func (f *tokeninfoFilter) Request(ctx filters.FilterContext) {
 	}
 
 	authorized(ctx, uid)
+
+	if f.maskOAuthRealms != nil {
+		realm, ok := authMap[realmKey].(string)
+		if ok && f.maskOAuthRealms.MatchString(realm) {
+			ctx.StateBag()[logfilter.MaskedAuthUserKey] = realm
```
AlexanderYastrebov, line 369: Should it rather be `uid` (not `realm`) that is stored under `MaskedAuthUserKey`?

config/config.go
```diff
@@ -541,6 +545,7 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.OidcSecretsFile, "oidc-secrets-file", "", oidcSecretsFileUsage)
 	flag.Var(cfg.CredentialPaths, "credentials-paths", credentialPathsUsage)
 	flag.DurationVar(&cfg.CredentialsUpdateInterval, "credentials-update-interval", defaultCredentialsUpdateInterval, credentialsUpdateIntervalUsage)
+	flag.StringVar(&cfg.MaskOAuthRealms, "mask-oauth-realms", defaultMaskOAuthRealms, maskOAuthRealmsUsage)
```
AlexanderYastrebov, line 548: As an alternative to the flag, tokeninfo filter could store `realm`ed `auth-user` like:
```go
ctx.StateBag()["auth-user" + realm] = uid
```
which then could be used like:
```
stateBagToTag("auth-user/employee", "client_id")
``` 
or chained as a sort of allowlist:
```
stateBagToTag("auth-user/employee", "client_id") -> stateBagToTag("auth-user/whatever", "client_id")
```

filters/tracing/statebagtotag.go
zeitlinger: yes, but then we have to make sure that nobody uses `stateBagToTag("auth-user", "client_id")`

zeitlinger:
> @zeitlinger `realm` is not in the spec because it is a Zalando extension

makes sense

for tokeninfo it's `curl -s -H "Authorization: Bearer $(ztoken)" -X POST https://info.services.auth.zalando.com/oauth2/tokeninfo` and there the key is `realm`.

filters/auth/tokeninfo.go
zeitlinger: idea was to at least see which realm the caller came from.

config/config.go
zeitlinger: seems rather hard to use IMO

filters/tracing/statebagtotag.go
AlexanderYastrebov: Oh, I see now -  it is not a feature to allow user to set tags only for certain auth realms but rather mask uids even though user wants `stateBagToTag("auth-user", "client_id")`, i.e. they will get realm instead of uid in masked realms.

Still it should be cleaner to persist realm unconditionally into statebag in tokeninfo filter and concentrate all logic of realm match and uid->realm substitution in the `stateBagToTag` filter.

It also changes simple `put this there` semantics of `stateBagToTag` into `put this there unless it is not auth-user and realm matches commandline switch regexp - in such a case substitute auth-user to ream` o_O

zeitlinger:
Thought about how to 
1. move more logic into the filter 
2. not hard code `realm` - as it's a Zalando extension
3. make masking work for the oidc, token inspection and webhook

This is my idea
1. store the entire auth map (with uid, realm, etc) in a new cache key ("auth-map")
2. rename `mask-oauth-realm` to `mask-oauth-user` with content e.g. `employees:realm=employees` which would look for `realm` = `employees` and replace it with `employees` (this could be a comma-separated list)

WDYT?

szuecs:
I think for oidc we already have stored everything in `ctx.StateBag()[oidcClaimsCacheKey]`, tokenintrospection uses state bag key `tokenintrospectionCacheKey` and tokeninfo `tokeninfoCacheKey`. I think you won't need to do anything for webhook, because basically it works only on response code of the http request of the webhook call, so there is no data to log.

In case of implementation we could also add a filter parameter that decides on which key to use: `filter("<tokeninfo|oidc|tokenintrospection>")`. Do you think it's not possible like this or does it have a drawback?

zeitlinger:
> I think for oidc we already have stored everything in `ctx.StateBag()[oidcClaimsCacheKey]`, tokenintrospection uses state bag key `tokenintrospectionCacheKey` and tokeninfo `tokeninfoCacheKey`. I think you won't need to do anything for webhook, because basically it works only on response code of the http request of the webhook call, so there is no data to log.

yes, we can also let the filter read directly from those keys instead of adding a new one.

> In case of implementation we could also add a filter parameter that decides on which key to use: `filter("<tokeninfo|oidc|tokenintrospection>")`. Do you think it's not possible like this or does it have a drawback?

That makes it harder to understand - hence easier to get wrong. 
So my preference would be to let the filter look for these 3 keys.

szuecs:
> > I think for oidc we already have stored everything in `ctx.StateBag()[oidcClaimsCacheKey]`, tokenintrospection uses state bag key `tokenintrospectionCacheKey` and tokeninfo `tokeninfoCacheKey`. I think you won't need to do anything for webhook, because basically it works only on response code of the http request of the webhook call, so there is no data to log.
> 
> yes, we can also let the filter read directly from those keys instead of adding a new one.
> 
> > In case of implementation we could also add a filter parameter that decides on which key to use: `filter("<tokeninfo|oidc|tokenintrospection>")`. Do you think it's not possible like this or does it have a drawback?
> 
> That makes it harder to understand - hence easier to get wrong.
> So my preference would be to let the filter look for these 3 keys.

sgtm

zeitlinger:
> > > I think for oidc we already have stored everything in `ctx.StateBag()[oidcClaimsCacheKey]`, tokenintrospection uses state bag key `tokenintrospectionCacheKey` and tokeninfo `tokeninfoCacheKey`. I think you won't need to do anything for webhook, because basically it works only on response code of the http request of the webhook call, so there is no data to log.
> > 
> > 
> > yes, we can also let the filter read directly from those keys instead of adding a new one.
> > > In case of implementation we could also add a filter parameter that decides on which key to use: `filter("<tokeninfo|oidc|tokenintrospection>")`. Do you think it's not possible like this or does it have a drawback?
> > 
> > 
> > That makes it harder to understand - hence easier to get wrong.
> > So my preference would be to let the filter look for these 3 keys.
> 
> sgtm

@szuecs changed accordingly - please review

aryszka:
I think this feature should not be a global flag, but rather a filter. Is there a reason or a limitation why it is a global setting?

zeitlinger:
> I think this feature should not be a global flag, but rather a filter. Is there a reason or a limitation why it is a global setting?

yes: the intention is to make it impossible to leak the user name. If we would make it dependent on the filter configuration, then we would need to validate that all filters are configured in a way to mask users (which is harder to do)

aryszka:
>> I think this feature should not be a global flag, but rather a filter. Is there a reason or a limitation why it is a global setting?

> yes: the intention is to make it impossible to leak the user name. If we would make it dependent on the filter configuration, then we would need to validate that all filters are configured in a way to mask users (which is harder to do)

We should not mistake our role as vendors for the one as users. As vendors, we should respect everybody's use case even if it's harder. The primary way of configuring Skipper's routing is using, well, the route configuration. To apply global routing constraints, we can rely on the default filters. See: https://opensource.zalando.com/skipper/operation/operation/#default-filters . Please, convert this functionality into a filter.

zeitlinger:
> Please, convert this functionality into a filter.

IIUC, you're saying we could append a global default filter that has the same effect.

Would this filter modify what `stateBagToTag` is doing?
Not sure I understand correctly.

aryszka:
> Would this filter modify what stateBagToTag is doing?

the current proposal works more or less like this:

1. defines a set of masks globally, on startup
2. the stateBagToTag applies this mask when the statebagitem is the auth user

the first step could be replaced by a filter that defines the same mask, and if the user wants, they can even set this filter only for specific routes or with the default filters option for all routes.

And then comes the question that was already raised in an earlier comment. Why limit this functionality to the auth user? I see no reason for that, masking could be applied to any statebag key at will of the user.

zeitlinger:
> the first step could be replaced by a filter that defines the same mask, and if the user wants, they can even set this filter only for specific routes or with the default filters option for all routes.

that would also work - would we actually be fine to have such a filter in all routes (given that it consumes some memory)? 

> And then comes the question that was already raised in an earlier comment. Why limit this functionality to the auth user? I see no reason for that, masking could be applied to any statebag key at the will of the user.

The logic for extracting the realm (or any other property) is specific to the [token caches](https://github.com/zalando/skipper/pull/1524/files#diff-80ada5524758cbbdf172afb3e2ad9b5beed163f1e96337aab386e90132ee46c2R63) - I don't know how to generalize it further.

aryszka:
> that would also work - would we actually be fine to have such a filter in all routes (given that it consumes some memory)?

No that's not a problem. This filter would not be much different from the other filters.

> The logic for extracting the realm (or any other property) is specific to the token caches - I don't know how to generalize it further.

Consider what it could do. It could apply a mask for values that are used by the stateBagToTag() filter. If we have `maskStateBag(stateBagKey, mask) -> stateBagToTag(stateBagKey, tag)`, then maskStateBag could store a masked value with the key prefixed by 'masked:...", and then stateBagToTag could check whether a masked value exists for the same key, or just use the masked key, as well. Then we don't need to modify (and couple functionality) in the auth filters for (with) masking.

ePaul:
This PR is linked from the our Zalando-internal [production readiness review template](https://docs.google.com/document/d/1F2Q1YlV40dsuZ3E0bWgrpZuK-GDleemEvKOwb8d5_24/edit#heading=h.4tpcplrvl7cc):

> (can also be done using [skipper](https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#statebagtotag) (deployed - but don’t use it yet until _this issue_ is resolved))

Is there any ETA on when this might be done?

szuecs:
@ePaul I think nobody is working on it.

Issue #1444, opened 2020-06-12
Support filter context (state bag) value templates


AlexanderYastrebov:
**Is your feature request related to a problem? Please describe.**

It is often needed to transfer/reuse values within route. Currently special-purpose filters are created, e.g.:
- `queryToHeader`, `headerToQuery`
- `setContextRequestHeader`
- `setDynamicBackendHostFromHeader`

**Describe the solution you would like**

* Support template resolution in filter values from context (state bag).
* Add *getter* filter variants to store value into context, e.g. `getRequestHeader("X-Foo-Header", "foo")`
* Add context *setter*, e.g. `set("first", "value")` or `set("second", "from ${first}")`. Might also be `setOnRequest`+`setOnResponse` pair.
* Optional/future: Add context-only filters to work with values stored in context, e.g. `replace("myVar", "^zalando\.(\w+)$", "www.zalando.$1")`

Examples:
* `setContextRequestHeader("X-Uid", "auth-user")` becomes `setRequestHeader("X-Uid", "${auth-user}")`
* `queryToHeader("foo-query-param", "X-Foo-Header")` could be `getQuery("foo-query-param", "myParam") -> setResponseHeader("X-Foo-Header", "${myParam}")`
* `queryToHeader("access_token", "Authorization", "Bearer %s")` (format string variant) could be `getQuery("access_token", "tok") -> setResponseHeader("Authorization", "Bearer ${tok}")`
* `headerToQuery("X-Foo-Header", "foo-query-param")` could be `getRequestHeader("X-Foo-Header", "hh") -> setQuery("foo-query-param", "${hh}")`
* `cookieToHeader("jwt", "Authorization", "Bearer ")` (#1006) would be `getCookie("jwt", "token") -> setResponseHeader("Authorization", "Bearer ${token}")`
* `setDynamicBackendHostFromHeader("X-Forwarded-Host") -> <dynamic>;` (#908) could be `getRequestHeader("X-Forwarded-Host", "backend:dynamic:host") -> <dynamic>;`
* `setDynamicBackendHost("example.com") -> <dynamic>;` could be `set("backend:dynamic:host", "example.com") -> <dynamic>;`
* Check token against path/header usecase (https://github.com/zalando/skipper/issues/804#issuecomment-427301435):
`
[...]

PR #1462, opened 2020-07-02
Implements variable template resolution in header filters

AlexanderYastrebov:
Template resolution is implemented for `setRequestHeader`, `appendRequestHeader`, `setResponseHeader` and `appendResponseHeader` filters.
Uses filter context path params and string state bag values to resolve template variables.

See #1453, #1444

Signed-off-by: Alexander Yastrebov <alexander.yastrebov@zalando.de>

PR #1528, opened 2020-09-30
Unifies filter template placeholder resolution

AlexanderYastrebov:
Introduces `eskip.template.ApplyRequestContext` and `eskip.template.ApplyResponseContext` functions and uses them for template placeholder resolution in `(set|append)(Request|Response)Header`, `(set|drop)Query` and `setPath` filters.

Removes placeholder resolution from state bag in header filters, see https://github.com/zalando/skipper/pull/1528#discussion_r526505822

See https://github.com/zalando/skipper/issues/1444#issuecomment-700967310


eskip/template.go
```diff
@@ -37,17 +39,38 @@ func NewTemplate(template string) *Template {
 }
 
 // Apply evaluates the template using a TemplateGetter function to resolve the
-// placeholders.
-func (t *Template) Apply(get TemplateGetter) string {
+// placeholders. Returns true if all placeholders resolved to non-empty values.
+func (t *Template) Apply(get TemplateGetter) (string, bool) {
```
szuecs, line 43: Because we don't use `Apply()` anymore other than in `ApplyContext()`, I would rather not change the signature here (breaking change for library users).

AlexanderYastrebov: Ok, I've kept signature intact.

docs/reference/filters.md
```diff
@@ -16,6 +16,17 @@ Example route with a match all, 2 filters and a backend:
 all: * -> filter1 -> filter2 -> "http://127.0.0.1:1234/";
 ```
 
+## Template variables
```
AlexanderYastrebov, line 19: Should it rather be called `Template placeholders` (also elsewhere)?

AlexanderYastrebov: Changed to `placeholder`

eskip/template.go
```diff
@@ -39,15 +41,40 @@ func NewTemplate(template string) *Template {
...
-		return result
+		return t.template
 	}
+	result, _ := t.apply(get)
+	return result
+}
 
+// ApplyContext evaluates the template using a filter context to resolve the
+// placeholders. Returns true if all placeholders resolved to non-empty values.
+func (t *Template) ApplyContext(ctx filters.FilterContext) (string, bool) {
```
AlexanderYastrebov, line 53: With a future extension in mind (e.g. support `${request.header.X-Foo}` placeholder) I think here we should have a pair `ApplyRequestContext` and `ApplyResponseContext` to be used in filter's `Request` and `Response` methods respectively.
`ApplyRequestContext` would have access to context and request properties and `ApplyResponseContext` would have access to context, request and response properties.

AlexanderYastrebov: Done

eskip/template.go
```diff
@@ -39,15 +41,54 @@ func NewTemplate(template string) *Template {
...
+func (t *Template) ApplyResponseContext(ctx filters.FilterContext) (string, bool) {
+	return t.apply(responseGetter(ctx))
+}
+
+func requestGetter(ctx filters.FilterContext) func(key string) string {
+	return func(key string) string {
+		if v := ctx.PathParam(key); v != "" {
+			return v
+		}
+		if v, ok := ctx.StateBag()[key]; ok {
```
aryszka, line 68: What do you think of supporting here additional primitve types? Basically only: integers and floats.

aryszka:
lgtm, I just would consider accepting primitive types from the state bag, namely: integers and floats

eskip/template.go
AlexanderYastrebov: I have updated PR to support numbers from statebag.

TBH I have a bit of a doubt regarding statebag here, maybe we can discuss it.
**In short**: do we actually need to give access to state bag at all?

Within #1444 I initially thought about using statebag as a kind of buffer between filter that copies e.g. header to statebag (e.g. `getRequestHeader("X-Foo-Header", "foo")`) and filter that uses stored value to e.g. set header (e.g `setRequestHeader("X-Bar-Header", "${foo}")`). With this in mind the ticket was outlined and initial templating #1462 implemented. 

Some time later the _projection_ idea [came up](https://github.com/zalando/skipper/issues/1444#issuecomment-700566455) - use placeholder prefix to denote datasource, e.g. `${request.header.X-Foo-Header}`, which I like and plan as a [subsequent extension](https://github.com/zalando/skipper/issues/1444#issuecomment-700967310) to this.

Prefixed placeholders obviate the need to have filters to store values into statebag.

`PathParam` placeholder replacement was done in #1462  with this PR (unified templating across filters) in mind - path params are resolved without a prefix to stay compatible with already existing `setQuery` and `setPath`.

#1462 added path params and statebag resolution but did not touch template regexp
https://github.com/zalando/skipper/blob/79de18de90fd338afae2a6ca6ac3e3bac8c46741/eskip/template.go#L12
which only allows words as placeholders, so actually many existing statebag keys like `auth-user` or `backend:isproxy` can not be used unless placeholder regexp is updated (which is necessary to support prefixed placeholders) - `${auth-user}` or `${backend:isproxy}` are not recognized as placeholders.

Since we did not hear complains from users about non-word placeholders, as well as, there are no means to store something into statebag in eskip we have a chance to retract and remove statebag support from templating. And maybe introduce it later as a prefixed placeholder `${statebag.whatever}` if needed. 

I just don't like two things: stagebag has internal use to pass data between skipper components (filters, proxy) and this PR exposes it without prefix (because it was done without prefix initially in #1462)
Or maybe I've just overthought all of this and having primitives from stagebag without a prefix is fine.

eskip/template_test.go
```diff
@@ -57,3 +61,101 @@ func TestTemplateGetter(t *testing.T) {
...
+				"uint":    uint(1),
+				"uint8":   uint8(1<<8 - 1),
+				"uint16":  uint16(1<<16 - 1),
+				"uint32":  uint32(1<<32 - 1),
+				"uint64":  uint64(1<<64 - 1),
+				"float32": float32(2.7182818284_5904523536_0287471352),
+				"float64": float64(3.1415926535_8979323846_2643383279),
+			},
+		},
+		"hello s -1 -128 -32768 -2147483648 -9223372036854775808 1 255 65535 4294967295 18446744073709551615 2.7182817 3.141592653589793",
```
AlexanderYastrebov, line 107: https://stackoverflow.com/questions/28045787/how-many-decimal-places-does-the-primitive-float-and-double-support

eskip/template.go
aryszka: i like this approach, and I think the templating could work well without the statebag. :+1:

aryszka:
@AlexanderYastrebov I am ok with the state of the PR, but I like it even more what you proposed regarding the statebag usage. Since we already allow referencing statebag values in the templating, I think it is fine to merge this PR as is, and decide on the 'statebag values in templates' topic later.

aryszka:
:+1:

AlexanderYastrebov:
@aryszka It is sitting for a while, can wait a bit more. Let me get rid of the statebag for the peace of mind, would be much easier to re-introduce it than to eliminate after we change placeholder pattern.

aryszka:
sgtm

AlexanderYastrebov:
I have rebased PR and removed statebag support as discussed

szuecs (approved):

szuecs:
:+1:

szuecs:
If you like you can change as I suggested the string concat looks a bit ugly imo :)

eskip/template_test.go
```diff
@@ -57,3 +61,62 @@ func TestTemplateGetter(t *testing.T) {
...
+		false,
+		"hello X ",
+		false,
+	},
+	} {
+		t.Run(ti.name, func(t *testing.T) {
+			template := NewTemplate(ti.template)
+			result, ok := template.ApplyRequestContext(ti.context)
+			if result != ti.requestExpect || ok != ti.requestOk {
+				t.Error(`Apply request error: "` + result + `" != "` + ti.requestExpect + `"`)
```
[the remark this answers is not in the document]
AlexanderYastrebov: Copy-Paste driven development :facepalm: 
Fixed.

szuecs:
👍

aryszka:
:+1:

PR #1530, opened 2020-10-01
use main request context for tracing tokeninfo requests

aryszka:
bind main request context to tokeninfo requests so that the tokeninfo opentracing spans would be connected to the main request span


aryszka:
:+1:

PR #1534, opened 2020-10-03
Pass the logrus JSONFormatter along so options can be changed

peterklijn:
Expose Logrus JSON formatter in `skipper.Options` so the defaults can be overwritten.

Issue: [feature request 1535](https://github.com/zalando/skipper/issues/1535)

### DIFF
--- a/logging/log.go
+++ b/logging/log.go
@@ -37,6 +37,11 @@ type Options struct {
 	// AccessLogStripQuery, when set, causes the query strings stripped
 	// from the request URI in the access logs.
 	AccessLogStripQuery bool
+
+	// LogrusJsonFormatter is passed along to to the underlying Logrus logging if JSON logging is enabled.
+	// To enable structured logging, use the ApplicationLogJSONEnabled and AccessLogJSONEnabled settings.
+	// In the case of access logs, the timestamp format will be overwritten and timestamps will be disabled.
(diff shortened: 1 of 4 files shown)


szuecs:
:+1:

logging/log.go
```diff
@@ -37,6 +37,11 @@ type Options struct {
 	// AccessLogStripQuery, when set, causes the query strings stripped
 	// from the request URI in the access logs.
 	AccessLogStripQuery bool
+
+	// LogrusJsonFormatter is passed along to to the underlying Logrus logging if JSON logging is enabled.
+	// To enable structured logging, use the ApplicationLogJSONEnabled and AccessLogJSONEnabled settings.
+	// In the case of access logs, the timestamp format will be overwritten and timestamps will be disabled.
```
AlexanderYastrebov, line 43: May I suggest to pass pointer here so that when it is `nil` defaults are used but when it's set it is used as is - this way lib user would have complete control over formatting and not worry about internal overrides.

Another thing to consider is to have two formatters right away, one for application log and one for access log (we know that it is a valid usecase since default configuration already uses separate formatters, also other options come in pairs)

peterklijn: Fair point, I've updated the PR accordingly

aryszka:
:+1:

szuecs:
:+1:

PR #1536, opened 2020-10-04
Added helpers for some predicates and filters

peterklijn:
Signed-off-by: Peter Klijn <p.klijn@thebeat.co>


predicates/core/core.go
```diff
@@ -0,0 +1,64 @@
...
+			len(p.Args),
+			count,
+		)
+	}
+
+	var a []string
+	for i := range p.Args {
+		s, ok := p.Args[i].(string)
+		if !ok {
+			return nil, fmt.Errorf("expected argument of type string, %s", p.Name)
```
szuecs, line 57: Please create a separate error, such that `error.Is()` can match the formated error for example you can do: `fmt.Errorf( "%w, %s", errNotAString, p.Name)`
and later check for `error.Is(errNotAString, err)` (positions of args could be different, did not checked them)

szuecs:
From my side looks good. What do others think @aryszka @AlexanderYastrebov ?

aryszka:
Thanks! As the direction, it looks good to me. I have currently two rather burocratic remarks:

- let's not create packages called like 'helpers', too generic and invites future maintainers to overload the functionality provided by them by unrelated concepts and features. I would be fine, if the proposed helpers/filters and helpers/predicates packages would be placed under the current eskip package, as eskip/filters and eskip/predicates. (As I believe they won't ever have any heavy dependencies on other parts of Skipper, therefore there's no real risk of ending up with circular deps.)
- the functionality in the helpers package (KV* etc), seems to me a bit of an overkill. How would the proposed helpers work without them? Are they critical from the perspective of convenience of routes defined in code?

peterklijn:
Hi @aryszka,

Thanks for your feedback!

> Thanks! As the direction, it looks good to me. I have currently two rather burocratic remarks:
> 
> * let's not create packages called like 'helpers', too generic and invites future maintainers to overload the functionality provided by them by unrelated concepts and features. I would be fine, if the proposed helpers/filters and helpers/predicates packages would be placed under the current eskip package, as eskip/filters and eskip/predicates. (As I believe they won't ever have any heavy dependencies on other parts of Skipper, therefore there's no real risk of ending up with circular deps.)

I agree, I'm not super happy with the `helpers` package. I actually started with having the helpers in the `predicates` and `filters` packages directly [in the first commit](https://github.com/zalando/skipper/pull/1536/commits/f6c54834765b59ff439c0f36249011dd18155154). However this didn't work due to cyclic dependencies. Sadly I can't find the build results as I think they're too old and Github removed them.

For predicates I think it's easy to fix, as it's just the error that's causing it, but for the `filters` package it's a bit more involved, and my initial approach was to not change the Skipper codebase too much while adding these helpers. Do you have any suggestions?

> * the functionality in the helpers package (KV* etc), seems to me a bit of an overkill. How would the proposed helpers work without them? Are they critical from the perspective of convenience of routes defined in code?

I agree that the KVPair is a bit overkill, the only benefit it gives is that it guarantees an even number of arguments. I added them to have consistency with the `JWTPayloadAnyKVRegexp` filter, where the arguments are any number of string, regex tuples. I think there's a benefit of using Golang native types like `regexp.Regexp` instead of using strings.
This way you have compiler guarantees that it are valid regexes. Like how [the 'Before' predicate](https://github.com/zalando/skipper/pull/1536/files#diff-f884a48c797bf65b32241a2d8d1a386341f83d8b01e4bb0565ff886c48e97bfaR171-R177) uses golang's `date.Time` and not a string in RFC3339 format.

But using these kind of constructs are something think we shouldn't have, I can remove them, you just lose some compiler help. What do you think?

```go
// JWTPayloadAnyKVRegexp
 func JWTPayloadAnyKVRegexp(pairs ...helpers.KVRegexPair) *eskip.Predicate {
 	return &eskip.Predicate{
 		Name: auth.MatchJWTPayloadAnyKVRegexpName,
 		Args: helpers.KVRegexPairToArgs(pairs),
 	}
 }
```

PR #1539, opened 2020-10-05
log: degrade level to debug in case no lookuper

szuecs:
degrade log level to debug because lookuper is called in the hot path and provide not much value at runtime, because it is likely dependent on user configuration

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1022,7 +1022,7 @@ func (p *Proxy) checkRatelimit(ctx *context) (ratelimit.Settings, int) {
 
 		s := setting.Lookuper.Lookup(ctx.Request())
 		if s == "" {
-			p.log.Errorf("Lookuper found no data in request for setting: %s and request: %v", setting, ctx.Request())
+			p.log.Debugf("Lookuper found no data in request for setting: %s and request: %v", setting, ctx.Request())
 			continue
 		}
 


szuecs:
:+1:

proxy/proxy.go
```diff
@@ -1022,7 +1022,7 @@ func (p *Proxy) checkRatelimit(ctx *context) (ratelimit.Settings, int) {
 
 		s := setting.Lookuper.Lookup(ctx.Request())
 		if s == "" {
-			p.log.Errorf("Lookuper found no data in request for setting: %s and request: %v", setting, ctx.Request())
+			p.log.Debugf("Lookuper found no data in request for setting: %s and request: %v", setting, ctx.Request())
```
AlexanderYastrebov, line 1025: Why not `Warnf`? Also what about other two above?

szuecs: 1) We don't dump the full request in the others.
2) IMO warnings should only be used for highlighting deprecation to users (library and runtime)
3) this case is, right now, by design no error, but "work as intended". Not sure if this should be considered as bug, but people use it as a feature.

gargravarr:
:+1:

proxy/proxy.go
AlexanderYastrebov: > but people use it as a feature.

I see now what you mean: 
```
... -> clusterClientRatelimit("x", 1, "1m", "X-header") -> clusterClientRatelimit("y", 1, "1m", "Y-header") -> ...
```
i.e. there may be requests either with `X-header` or `Y-header` and one or another would produce log message.

Issue #1451, opened 2020-06-17
Rename inlineContent filter to finallyInlineContent


burimshala:
**Is your feature request related to a problem? Please describe.**
There is no clear indication that this filter should be placed as the last one beside the documentation.

**Describe the solution you would like**
I want to make it obvious that this filter should be placed as the last filter which will be indicated by the finally keyword.

**Would you like to work on it?**
Yes, but no time

szuecs:
I would prefer changing documentation, because if it is missing it will be ok and I don't think a rename would make it obvious.

AlexanderYastrebov:
Generally to use a filter user consults documentation and [documentation for the `inlineContent`](https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#inlinecontent) states 
> !!! note inlineContent filter is special and must be the last in the filter chain.

IMO this should be closed to reduce backlog.

szuecs:
@AlexanderYastrebov Yes thanks documentation was done

PR #1544, opened 2020-10-07
Updates `inlineContent` filter documentation

AlexanderYastrebov:
* Adds inline content with status example
* Adds note on automatic json content type detection (https://github.com/golang/go/issues/10630)
* Explains why it must be the last in the chain

Improves on #1202 and #1451


szuecs:
:+1:

aryszka:
:+1:

PR #1545, opened 2020-10-07
add profiling docs

szuecs:


docs/operation/operation.md
```diff
@@ -879,3 +879,55 @@ predicates and filters involved in the route processing:
...
+To enable profiling in skipper you have to use `-enable-profile`. This
+will start a profiling route at `/debug/pprof/profile` on the support
+listener, which defaults to `:9911`.
+
+### Profiling example
+
+Start skipper with enabled profiling:
+
+```
+./bin/skipper -routes-file examples/shadow-2-routes.eskip -enable-profile
```
AlexanderYastrebov, line 897: It would be nice to use an existing file (e.g from `./skptesting/`) so that users may actually try it out

szuecs: Before I thought to just use inline route, which does not depend on any CWD, but I missed to replace it :D

szuecs:
:+1:

aryszka:
:+1:

PR #1547, opened 2020-10-09
Normalizes URLs for upstream

tomlarkworthy:
Currently the matching logic normalizes paths, but these transformation are not applied when proxied upstream. So /malicious/../nice will pass because the matcher sees /nice but upstream will see /malicious/../nice instead!

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/dimfeld/httppath"
 	ot "github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
 	"github.com/zalando/skipper/circuit"
@@ -37,8 +38,6 @@ import (
 	"github.com/zalando/skipper/routing"
 	"github.com/zalando/skipper/scheduler"
 	"github.com/zalando/skipper/tracing"
-
-	"github.com/dimfeld/httppath"
 )
 
 const (
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1440,12 +1440,12 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}()
 
+	// Change /foo/../bar to /bar for matching and passing upstream
+	r.URL.Path = httppath.Clean(r.URL.Path)
+
 	if p.flags.patchPath() {
 		r.URL.Path = rfc.PatchPath(r.URL.Path, r.URL.RawPath)
 	}
-
-	// Change /foo/../bar to /bar for matching and passing upstream
-	r.URL.Path = httppath.Clean(r.URL.Path)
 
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

tomlarkworthy:
Logic explained here https://play.golang.org/p/pR-0yMr0HTQ

routing/matcher.go
```diff
@@ -487,9 +487,8 @@ func matchLeaves(leaves leafMatchers, req *http.Request, path, exactPath string)
 // returns the associated value, and the wildcard parameters from the path definition,
 // if any.
 func (m *matcher) match(r *http.Request) (*Route, map[string]string) {
-	// normalize path before matching
 	// in case ignoring trailing slashes, match without the trailing slash
-	path := httppath.Clean(r.URL.Path)
```
AlexanderYastrebov, line 492: There are quite some other places where `httppath.Clean()` is used in the matcher. Should they be considered as well?

proxy/proxy.go
```diff
@@ -1443,6 +1444,9 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		r.URL.Path = rfc.PatchPath(r.URL.Path, r.URL.RawPath)
 	}
 
+	// Change /foo/../bar to /bar for matching and passing upstream
+	r.URL.Path = httppath.Clean(r.URL.Path)
```
AlexanderYastrebov, line 1448: Isn't this a breaking change? Should it be feature option maybe?

szuecs, line 1448: this would override the rfc.PatchPath() call above

tomlarkworthy: maybe, although there is a problem for my team accessing these kinda of feature flags. Also, one could also argue this is a bug fix and therefore everyone should get it.

tomlarkworthy: good catch. Yeah we need to apply it early as the pathPath might be swapping a / for a %2F and we probably want to operate on the / version. Though the critical thing for me is the proxied upstream path and the path used in matcher do not diverge

routing/matcher.go
tomlarkworthy: I think the other ones are cleaning a path derived from the *route which is config and not something the change in the proxy.go affects. (maybe we should clean earlier in the route definition too but thats a bit out of scope for this issue)

proxy/proxy.go
aryszka: the alternative would be to use an explicit cleanPath() filter, but I am not aware of a scenario that would get broken this way.

aryszka:
:+1:

szuecs:
:+1:

proxy/proxy.go
NorthFury: An explicit `cleanPath()` filter is probably the way to go. I don't expect a proxy to mutate my request unless explicitly expressed with a filter on the route.

PR #1548, opened 2020-10-10
Introduce powerofnchoices algorithm in Load Balancer

ardaguclu:
This pull request is rewritten version of https://github.com/zalando/skipper/pull/1507. 

I tested in locally;

- Run 5 simple Go HTTP Servers(same code, different ports and random sleeps 0-2000 milliseconds)
- Skipper eskip configuration, `hello: * -> <powerOfRandomNChoices, "http://localhost:8001", "http://localhost:8002", "http://localhost:8003", "http://localhost:8004", "http://localhost:8005">;`
- Python script sending 2000 request with asyncio.

it distributes traffic better than round robin. However, this is just a first phase trial, it is better to test algorithm in real time traffic or suggested any other method does not come to my mind.

### DIFF
--- a/docs/reference/backends.md
+++ b/docs/reference/backends.md
@@ -258,6 +258,7 @@ Current implemented algorithms:
 - `roundRobin`: backend is chosen by the round robin algorithm, starting with a random selected backend to spread across all backends from the beginning
 - `random`: backend is chosen at random
 - `consistentHash`: backend is chosen by a consistent hashing algorithm with the client X-Forwarded-For header with remote IP as the fallback as input to the hash function
+- `powerOfRandomNChoices`: backend is chosen by powerIfRandomNChoices algorithm with randomly select N backends and choose less active one.
 - __TODO__: https://github.com/zalando/skipper/issues/557
 
 Route example with 2 backends and the `roundRobin` algorithm:
@@ -273,6 +274,11 @@ r0: * -> <random, "http://127.0.0.1:9998", "http://127.0.0.1:9997">;
 Route example with 2 backends and the `consistentHash` algorithm:
 ```
 r0: * -> <consistentHash, "http://127.0.0.1:9998", "http://127.0.0.1:9997">;
+```
+
+Route example with 2 backends and the `powerOfRandomNChoices` algorithm:
+```
+r0: * -> <powerOfRandomNChoices, "http://127.0.0.1:9998", "http://127.0.0.1:9997">;
 ```
 
 Proxy with `roundRobin` loadbalancer and two backends:
--- a/docs/reference/backends.md
+++ b/docs/reference/backends.md
@@ -258,7 +258,7 @@ Current implemented algorithms:
 - `roundRobin`: backend is chosen by the round robin algorithm, starting with a random selected backend to spread across all backends from the beginning
 - `random`: backend is chosen at random
 - `consistentHash`: backend is chosen by a consistent hashing algorithm with the client X-Forwarded-For header with remote IP as the fallback as input to the hash function
-- `powerOfRandomNChoices`: backend is chosen by powerIfRandomNChoices algorithm with randomly select N backends and choose less active one.
+- `powerOfRandomNChoices`: backend is chosen by powerOfRandomNChoices algorithm with randomly select N backends and choose less active one.
 - __TODO__: https://github.com/zalando/skipper/issues/557
 
 Route example with 2 backends and the `roundRobin` algorithm:
--- a/docs/reference/backends.md
+++ b/docs/reference/backends.md
@@ -258,7 +258,7 @@ Current implemented algorithms:
 - `roundRobin`: backend is chosen by the round robin algorithm, starting with a random selected backend to spread across all backends from the beginning
 - `random`: backend is chosen at random
 - `consistentHash`: backend is chosen by a consistent hashing algorithm with the client X-Forwarded-For header with remote IP as the fallback as input to the hash function
-- `powerOfRandomNChoices`: backend is chosen by powerOfRandomNChoices algorithm with randomly select N backends and choose less active one.
+- `powerOfRandomNChoices`: backend is chosen by powerOfRandomNChoices algorithm with selecting N random endpoints and picking the one with least outstanding requests from them. (http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf)
 - __TODO__: https://github.com/zalando/skipper/issues/557
 
 Route example with 2 backends and the `roundRobin` algorithm:
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -1,4 +1,4 @@
-package algorithm
+package loadbalancer
 
 import (
 	"errors"
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -71,7 +71,7 @@ func shiftToOldest(ctx *routing.LBContext) routing.LBEndpoint {
 	return oldest
 }
 
-func shiftToRemaining(rnd *rand.Rand, ctx *routing.LBContext, w []int, now time.Time) routing.LBEndpoint {
+func shiftToRemaining(sr *safeRand, ctx *routing.LBContext, w []int, now time.Time) routing.LBEndpoint {
 	notFadingIndexes := w
 	ep := ctx.Route.LBEndpoints
 	for i := 0; i < len(ep); i++ {
@@ -89,10 +89,10 @@ func shiftToRemaining(rnd *rand.Rand, ctx *routing.LBContext, w []int, now time.
 	}
 
 	// otherwise equally distribute between the old endpoints
-	return ep[notFadingIndexes[rnd.Intn(len(notFadingIndexes))]]
+	return ep[notFadingIndexes[sr.getIntn(len(notFadingIndexes))]]
 }
 
-func withFadeIn(rnd *rand.Rand, ctx *routing.LBContext, w []int, choice int) routing.LBEndpoint {
+func withFadeIn(sr *safeRand, ctx *routing.LBContext, w []int, choice int) routing.LBEndpoint {
 	now := time.Now()
 	f := fadeIn(
 		now,
@@ -101,25 +101,53 @@ func withFadeIn(rnd *rand.Rand, ctx *routing.LBContext, w []int, choice int) rou
 		ctx.Route.LBEndpoints[choice].Detected,
 	)
 
-	if rnd.Float64() < f {
+	if sr.getFloat64() < f {
 		return ctx.Route.LBEndpoints[choice]
 	}
 
-	return shiftToRemaining(rnd, ctx, w, now)
+	return shiftToRemaining(sr, ctx, w, now)
+}
+
+// safeRand is a struct for managing concurrent safe random access
+// with the custom source.
+type safeRand struct {
+	rnd *rand.Rand
+	mx  sync.Mutex
+}
+
+func newSafeRand() *safeRand {
+	src := rand.NewSource(time.Now().UnixNano())
+	return &safeRand{
+		rnd: rand.New(src),
+	}
+}
+
+// getIntn returns random int within the given range.
+func (s *safeRand) getIntn(lbe int) int {
+	s.mx.Lock()
+	defer s.mx.Unlock()
+	return s.rnd.Intn(lbe)
+}
+
+// getFloat64 returns random float64
+func (s *safeRand) getFloat64() float64 {
+	s.mx.Lock()
+	defer s.mx.Unlock()
+	return s.rnd.Float64()
 }
 
 type roundRobin struct {
 	mx         sync.Mutex
 	index      int
-	rnd        *rand.Rand
+	sr         *safeRand
 	fadeInCalc []int
 }
 
 func newRoundRobin(endpoints []string) routing.LBAlgorithm {
-	rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
+	sr := newSafeRand()
 	return &roundRobin{
-		index: rnd.Intn(len(endpoints)),
-		rnd:   rnd,
+		index: sr.getIntn(len(endpoints)),
+		sr:    sr,
 
 		// preallocating frequently used slice
 		fadeInCalc: make([]int, 0, len(endpoints)),
@@ -140,18 +168,17 @@ func (r *roundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 		return ctx.Route.LBEndpoints[r.index]
 	}
 
-	return withFadeIn(r.rnd, ctx, r.fadeInCalc, r.index)
+	return withFadeIn(r.sr, ctx, r.fadeInCalc, r.index)
 }
 
 type random struct {
-	rand       *rand.Rand
+	sr         *safeRand
 	fadeInCalc []int
 }
 
 func newRandom(endpoints []string) routing.LBAlgorithm {
-	t := time.Now().UnixNano()
 	return &random{
-		rand: rand.New(rand.NewSource(t)),
+		sr: newSafeRand(),
 
 		// preallocating frequently used slice
 		fadeInCalc: make([]int, 0, len(endpoints)),
@@ -164,12 +191,12 @@ func (r *random) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 		return ctx.Route.LBEndpoints[0]
 	}
 
-	i := r.rand.Intn(len(ctx.Route.LBEndpoints))
+	i := r.sr.getIntn(len(ctx.Route.LBEndpoints))
 	if ctx.Route.LBFadeInDuration <= 0 {
 		return ctx.Route.LBEndpoints[i]
 	}
 
-	return withFadeIn(r.rand, ctx, r.fadeInCalc, i)
+	return withFadeIn(r.sr, ctx, r.fadeInCalc, i)
 }
 
 type consistentHash struct{}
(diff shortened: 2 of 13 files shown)


loadbalancer/algorithm.go
```diff
@@ -197,9 +201,66 @@ func (ch *consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
+		}
+	}
+	return best
+}
+
+// newPowerOfRandomNChoices selects N random backends and picks the one with less outstanding requests.
+func newPowerOfRandomNChoices(endpoints []string) routing.LBAlgorithm {
+	t := time.Now().UnixNano()
+	return &powerOfRandomNChoices{
+		rand:            rand.New(rand.NewSource(t)),
```
AlexanderYastrebov, line 256: From https://golang.org/pkg/math/rand/
>  The default Source is safe for concurrent use by multiple goroutines, but Sources created by NewSource are not.

loadbalancer/algorithm.go
```diff
@@ -197,9 +201,66 @@ func (ch *consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
+
+// getScore returns negative value of inflightrequests count.
+func (p *powerOfRandomNChoices) getScore(e routing.LBEndpoint) int {
+	// endpoints with higher inflight request should have lower score
+	return -e.Metrics.GetInflightRequests()
+}
+
+// getRandomIndex returns random integer with the range of 0 and l
+// and skips excluded integers. If the length is equal to excluded length
+// it behaves as if excluded list is empty
```
AlexanderYastrebov, line 217: Please have a look at https://github.com/zalando/skipper/pull/1507#discussion_r472041811 regarding exclusions.

Please consider borrowing [core algorithm](https://github.com/zalando/skipper/pull/1507/files#diff-8ef30ff734de1bf682a3b0d530f9f663R130-R147) from the previous attempt.

It might also be interesting [how it is implemented in haproxy](https://github.com/haproxy/haproxy/commit/21c741a665f)

loadbalancer/algorithm.go
```diff
@@ -30,13 +30,17 @@ const (
 
 	// ConsistentHash indicates choice between the backends based on their hashed address.
 	ConsistentHash
+
+	// PowerOfRandomNChoices selects N random endpoints and picks the one with least outstanding requests from them.
+	PowerOfRandomNChoices
```
AlexanderYastrebov, line 35: Please have a look at discussion [regarding naming](https://github.com/zalando/skipper/pull/1507#discussion_r472289180)

docs/reference/backends.md
```diff
@@ -258,6 +258,7 @@ Current implemented algorithms:
 - `roundRobin`: backend is chosen by the round robin algorithm, starting with a random selected backend to spread across all backends from the beginning
 - `random`: backend is chosen at random
 - `consistentHash`: backend is chosen by a consistent hashing algorithm with the client X-Forwarded-For header with remote IP as the fallback as input to the hash function
+- `powerOfRandomNChoices`: backend is chosen by powerOfRandomNChoices algorithm with randomly select N backends and choose less active one.
```
AlexanderYastrebov, line 261: Shouldn't this description, this https://github.com/zalando/skipper/pull/1548/files#diff-861ba78658f4e6d2a609b032484d23b1R22
and this https://github.com/zalando/skipper/pull/1548/files#diff-8ef30ff734de1bf682a3b0d530f9f663R34 all match?

AlexanderYastrebov: Also it might make sense to add a link to algorithm paper

AlexanderYastrebov:
> or suggested any other method does not come to my mind.

Here is an idea of local test setup for testing algorithms and edge cases

1. Make N (e.g. 10) test domains:
```
$ tail -n 10 /etc/hosts
127.0.0.1 h0.test
127.0.0.1 h1.test
127.0.0.1 h2.test
127.0.0.1 h3.test
127.0.0.1 h4.test
127.0.0.1 h5.test
127.0.0.1 h6.test
127.0.0.1 h7.test
127.0.0.1 h8.test
127.0.0.1 h9.test
```

2. Create route file to simulate backends with various latencies:
```
$ cat ~/tmp/algo.eskip 
r0: Host("h0.test") ->                inlineContent("00") -> <shunt>;
r1: Host("h1.test") -> latency(10) -> inlineContent("10") -> <shunt>;
r2: Host("h2.test") -> latency(20) -> inlineContent("20") -> <shunt>;
r3: Host("h3.test") -> latency(30) -> inlineContent("30") -> <shunt>;
r4: Host("h4.test") -> latency(40) -> inlineContent("40") -> <shunt>;
r5: Host("h5.test") -> latency(50) -> inlineContent("50") -> <shunt>;
r6: Host("h6.test") -> latency(60) -> inlineContent("60") -> <shunt>;
r7: Host("h7.test") -> latency(70) -> inlineContent("70") -> <shunt>;
r8: Host("h8.test") -> latency(80) -> inlineContent("80") -> <shunt>;
r9: Host("h9.test") -> latency(90) -> inlineContent("90") -> <shunt>;

test: Path("/test") -> setPath("/") -> 
<   roundRobin, // algorithm
    "http://h0.test:9090",
    "http://h1.test:9090",
    "http://h2.test:9090",
    "http://h3.test:9090",
    "http://h4.test:9090",
    "http://h5.test:9090",
    "http://h6.test:9090",
    "http://h7.test:9090",
    "http://h8.test:9090",
    "http://h9.test:9090"
>;
``` 
and run skipper
```
bin/skipper -routes-file ~/tmp/algo.eskip -access-log-disabled
```

3. Perform requests

```
$ grep algorithm ~/tmp/algo.eskip 
<   roundRobin, // algorithm
$ for i in {1..100}; do curl -s localhost:9090/test; echo; done | sort | uniq -c
     10 00
     10 10
     10 20
     10 30
     10 40
     10 50
     10 60
     10 70
     10 80
     10 90
```
change algorithm
```
$ grep algorithm ~/tmp/algo.eskip 
<   random, // algorithm
$ for i in {1..100}; do curl -s localhost:9090/test; echo; done | sort | uniq -c
     11 00
      9 10
     12 20
     13 30
      9 40
      9 50
     10 60
     13 70
      7 80
      7 90
```
once more
```
$ grep algorithm ~/tmp/algo.eskip 
<   consistentHash, // algorithm
$ for i in {1..100}; do curl -s localhost:9090/test; echo; done | sort | uniq -c
    100 60
```

4. Adjust parameters, rinse and repeat

ardaguclu:
Thanks @AlexanderYastrebov, My test flow is nearly same with your proposal, I send traffic concurrently rather than in a for loop to measure the performance.

AlexanderYastrebov:
@ardaguclu I have modified [my setup](https://github.com/zalando/skipper/pull/1548#issuecomment-706626861) to return different status codes per backend and used [vegeta](https://github.com/tsenart/vegeta) to load and show latencies:
```
$ cat ~/tmp/algo.eskip 
r0: Host("h0.test") ->                status(200) -> <shunt>;
r1: Host("h1.test") -> latency(10) -> status(210) -> <shunt>;
r2: Host("h2.test") -> latency(20) -> status(220) -> <shunt>;
r3: Host("h3.test") -> latency(30) -> status(230) -> <shunt>;
r4: Host("h4.test") -> latency(40) -> status(240) -> <shunt>;
r5: Host("h5.test") -> latency(50) -> status(250) -> <shunt>;
r6: Host("h6.test") -> latency(60) -> status(260) -> <shunt>;
r7: Host("h7.test") -> latency(70) -> status(270) -> <shunt>;
r8: Host("h8.test") -> latency(80) -> status(280) -> <shunt>;
r9: Host("h9.test") -> latency(90) -> status(290) -> <shunt>;

test: Path("/test") -> setPath("/") -> 
<   random, // algorithm
    "http://h0.test:9090",
    "http://h1.test:9090",
    "http://h2.test:9090",
    "http://h3.test:9090",
    "http://h4.test:9090",
    "http://h5.test:9090",
    "http://h6.test:9090",
    "http://h7.test:9090",
    "http://h8.test:9090",
    "http://h9.test:9090"
>;

$ grep algorithm ~/tmp/algo.eskip 
<   random, // algorithm
$ echo "GET http://localhost:9090/test" | vegeta attack -duration=10s -rate=1000/s | vegeta report
Requests      [total, rate, throughput]         10000, 1000.11, 991.24
Duration      [total, attack, wait]             10.088s, 9.999s, 89.486ms
Latencies     [min, mean, 50, 90, 95, 99, max]  322.005µs, 46.356ms, 46.002ms, 87.973ms, 91.287ms, 91.827ms, 97.281ms
Bytes In      [total, mean]                     0, 0.00
Status Codes  [code:count]                      200:987  210:1020  220:1003  230:998  240:1018  250:1002  260:976  270:969  280:1017  290:1010  
Error Set:

$ grep algorithm ~/tmp/algo.eskip 
<   powerOfRandomNChoices, // algorithm
$ echo "GET http://localhost:9090/test" | vegeta attack -duration=10s -rate=1000/s | vegeta report
Requests      [total, rate, throughput]         10000, 1000.10, 991.52
Duration      [total, attack, wait]             10.086s, 9.999s, 86.53ms
Latencies     [min, mean, 50, 90, 95, 99, max]  251.764µs, 34.357ms, 30.902ms, 80.736ms, 89.416ms, 91.68ms, 105.422ms
Bytes In      [total, mean]                     0, 0.00
Status Codes  [code:count]                      200:1933  210:1589  220:1264  230:1070  240:931  250:803  260:709  270:626  280:565  290:510  
Error Set:
```

As expected algorithm routes more requests towards faster backends and improves mean latency.

loadbalancer/algorithm.go
ardaguclu: This is the same implementation of newRandom.

loadbalancer/algorithm.go
ardaguclu: I agree about name of algorithm is misleading. But according to the paper http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf official name is this.

loadbalancer/algorithm.go
AlexanderYastrebov: `newRandom` also has this problem unfortunately. 
> The default Source is safe for concurrent use by multiple goroutines, but Sources created by NewSource are not.

Top-level functions, such as Float64 and Int use mutex, see https://golang.org/src/math/rand/rand.go?#L382

ardaguclu: I think, it is better different algorithms managing their own custom random objects(instead using global rand source) and I agree about concurrent access should be safe. That's why, I created safeRand object to manage these locking operations and inject it into random and powerofrandomnchoices algorithms.

loadbalancer/algorithm.go
```diff
@@ -97,25 +101,53 @@ func withFadeIn(rnd *rand.Rand, ctx *routing.LBContext, w []int, choice int) rou
...
 	mx         sync.Mutex
 	index      int
-	rnd        *rand.Rand
+	sr         *safeRand
 	fadeInCalc []int
 }
 
 func newRoundRobin(endpoints []string) routing.LBAlgorithm {
-	rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
+	sr := newSafeRand()
```
AlexanderYastrebov, line 147: `roundRobin` already has mutex that guards both index update and random generation in `Apply` and it is safe to call it without mutex in constructor, so it is not needed here.

loadbalancer/algorithm.go
```diff
@@ -186,7 +220,7 @@ func (ch *consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 	key := net.RemoteHost(ctx.Request).String()
 	if _, err := h.Write([]byte(key)); err != nil {
 		log.Errorf("Failed to write '%s' into hash: %v", key, err)
-		return ctx.Route.LBEndpoints[rand.Intn(len(ctx.Route.LBEndpoints))]
+		return ctx.Route.LBEndpoints[ch.sr.getIntn(len(ctx.Route.LBEndpoints))]
```
AlexanderYastrebov, line 223: `Write` actually never returns error https://golang.org/src/hash/fnv/fnv.go?#L109 so this path is a kind of fallback that is never executed, Therefore there is no need to have `safeRandom` in every `consistentHash` since is never used.

loadbalancer/algorithm.go
AlexanderYastrebov: IMO there is no benefit for each algorithm to have its own random over using global random unless algorithm uses one mutes to guard random and something else (like `roundRobin` does). 

See also my comments on why `roundRobin` and `consistentHash` do not need `safeRand`.

proxy/proxy.go
```diff
@@ -866,12 +868,17 @@ func (p *Proxy) makeUpgradeRequest(ctx *context, req *http.Request) error {
...
 	var err error
-	req, err := mapRequest(ctx.request, ctx.route, ctx.outgoingHost, p.flags.HopHeadersRemoval(), ctx.StateBag())
+	req, endpoint, err := mapRequest(ctx.request, ctx.route, ctx.outgoingHost, p.flags.HopHeadersRemoval(), ctx.StateBag())
 	if err != nil {
 		p.log.Errorf("could not map backend request, caused by: %v", err)
 		return nil, &proxyError{err: err}
 	}
 
+	if endpoint != nil {
+		endpoint.Metrics.IncInflightRequest()
```
AlexanderYastrebov, line 878: What do you think about changing it to something more generic?
E.g. with
```
start := endpoint.Metrics.RequestStarted()
defer endpoint.Metrics.RequestFinished(start)
```
it may count inflight requests as well as measure latency (tbd).

docs/reference/backends.md
```diff
@@ -258,6 +258,7 @@ Current implemented algorithms:
 - `roundRobin`: backend is chosen by the round robin algorithm, starting with a random selected backend to spread across all backends from the beginning
 - `random`: backend is chosen at random
 - `consistentHash`: backend is chosen by a consistent hashing algorithm with the client X-Forwarded-For header with remote IP as the fallback as input to the hash function
+- `powerOfRandomNChoices`: backend is chosen by powerOfRandomNChoices algorithm with selecting N random endpoints and picking the one with least outstanding requests from them. (http://www.eecs.harvard.edu/~michaelm/postscripts/handbook2001.pdf)
```
AlexanderYastrebov, line 261: s/endpoints/backends/
s/outstanding/inflight/

Also it may worth mentioning that inflight request are counted per route, i.e. the same backend may have different inflight request count in different routes.

ardaguclu:
@AlexanderYastrebov Thanks for your code review and feedback. I reverted safeRand from package, I think these enhancement can belong to an another pull request.

proxy/proxy.go
aryszka: `metrics.Metrics.MeasureSince` can be used.

currently we use a timer for the backend host, but it doesn't consider the endpoints. Another problem of it is that it includes more than just what it strictly needs to, e.g. the retries with a different endpoint.

loadbalancer/algorithm.go
AlexanderYastrebov: After some thought I have to correct myself. It might make sense to use random+mutex instance per algorithm instance. This way only requests following the route would compete for the mutex as opposed to global random where requests from all routes (that use the same algorithm) compete for it. I guess that is the reason we see random instance per algorithm.

szuecs: yes random instance per algorithm instance makes sense

ardaguclu: I added rand and mutex into powerOfRandomNChoices algorithm like you proposal.

szuecs:
@ardaguclu I didn't have time in the last weeks. Are you awaiting a review or do you have more work to do with this pull request?

ardaguclu:
Thanks @szuecs. I completed my work and waiting for review. My tests and @AlexanderYastrebov tests seem to show that algorithm is working propoerly.

loadbalancer/algorithm.go
szuecs: I would rather go with the current one, than diverging too much from the original paper

loadbalancer/doc.go
```diff
@@ -29,6 +35,7 @@ Eskip example:
         r1: * -> <roundRobin, "http://127.0.0.1:9998", "http://127.0.0.1:9997">;
         r2: * -> <consistentHash, "http://127.0.0.1:9998", "http://127.0.0.1:9997">;
         r3: * -> <random, "http://127.0.0.1:9998", "http://127.0.0.1:9997">;
+		r4: * -> <powerOfRandomNChoices, "http://127.0.0.1:9998", "http://127.0.0.1:9997">;
```
szuecs, line 38: please indent correctly TAB vs spaces?

szuecs (commented):
lgtm in general.
There are some open questions:
1) algorithm naming (seems fine to me)
2) do we want to parameterize the value of N in the future? If not Why we call it N and not 2?
3) the metrics discussion https://github.com/zalando/skipper/pull/1548/files#r503193171

ardaguclu:
@szuecs  Thanks for your review. I applied the changes you requested. In the future, I'm planning to make N value parameterize instead 2 (although according to the researches, 2 is the optimal value, but algorithm is dynamic and we can provide this dynamicity).
From the metrics discussion, what I understand is not to change the current behavior.

proxy/proxy.go
aryszka: sorry, revisiting my comment about MeasureSince here, because it is a 'heavier' tool. The current atomic counter is a good start.

aryszka:
:+1:

szuecs:
:+1:

szuecs:
Thanks @ardaguclu our automation will release this soon!

Issue #1238, opened 2019-10-30
consistent handling of the filter chain on the response path


aryszka:
The filters are applied first to the requests in normal order, then a backend request, a shunt, rate limiting, or loopback happens. Then in the case of a response received from the backend (regardless of its status), in the case of the shunt and loopback, the response is passed to the filter chain in reverse order before sending it back to the client. On the other hand the response filters are skipped in case of a connection problem or rate limiting. This is not consistent, and the filters should be applied to the response in these cases, too. In order to achieve this, we need to create a dummy response object with the right status (see for an example https://github.com/zalando/skipper/blob/master/proxy/context.go#L57), set it to the context and pass the context to the filter chain.

To make the rate limiting consistent, since it's configured as a filter, it should also act as a filter, and when rate limiting is applied then it should be just like any shunting filter. See the redirectTo() filter as an example. Ideally, it should use the FilterContext.Serve() method and be implemented purely in the filter package. (See: https://github.com/zalando/skipper/blob/master/filters/filters.go#L60)

Thanks @NorthFury for volunteering for this improvement.

AlexanderYastrebov:
A related example of inconsistency is that [status](https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#status) filter docs say
> Sets the response status code to the given value, with no regards to the backend response.

It is not true in case of backend error:
```
[APP]INFO[0000] route settings, reset, route: r: * -> status(200) -> "http://unavailable.example" 
...               
[APP]ERRO[0008] Failed to do backend roundtrip to http://unavailable.example: dialing failed true: dial tcp: lookup unavailable.example: no such host 
[APP]ERRO[0008] error while proxying after 1.245688ms, route r with backend network http://unavailable.example, status code 502: dialing failed true: dial tcp: lookup unavailable.example: no such host, remote host: 127.0.0.1, request: "GET / HTTP/1.1", user agent: "curl/7.58.0" 
127.0.0.1 - - [09/Oct/2020:23:08:11 +0200] "GET / HTTP/1.1" 502 12 "-" "curl/7.58.0" 1 localhost:9090 - -
```

szuecs:
I am not sure if we should replace the status in case of a connection error to the backend.

AlexanderYastrebov:
@szuecs True, probably status docs should be clarified

tkrop:
[...]

PR #1549, opened 2020-10-10
Refactors per-route ratelimit handling

AlexanderYastrebov:
At present per-route ratelimit filters do not perform actual ratelimiting but rather
store a configuration marker in request context. Ratelimits are
considered by proxy after request path of filter chain is executed but
before backend request.

For example in `clientRatelimit(3, "1m") -> oauthTokeninfoAnyScope("s1")` route
`oauthTokeninfoAnyScope` filter will request tokeninfo even if ratelimit is reached.

This change:

* refactors ratelimit filters to actually consult ratelimits and
shunt request path of filter chain with `429 Too Many Requests` if limit
is reached
* removes per-route ratelimit handling from proxy

Part of #1238

### DIFF
--- a/filters/ratelimit/ratelimit.go
+++ b/filters/ratelimit/ratelimit.go
@@ -6,6 +6,7 @@ For detailed documentation of the ratelimit, see https://godoc.org/github.com/za
 package ratelimit
 
 import (
+	"context"
 	"net/http"
 	"strings"
 	"time"
@@ -15,24 +16,51 @@ import (
 	"github.com/zalando/skipper/ratelimit"
 )
 
-// RetryAfterKey is used as key in the context state bag
-const RetryAfterKey = "#ratelimitretryafter"
-
-// RouteSettingsKey is used as key in the context state bag
-const RouteSettingsKey = "#ratelimitsettings"
-
 type spec struct {
 	typ        ratelimit.RatelimitType
+	provider   RatelimitProvider
 	filterName string
 }
 
 type filter struct {
 	settings ratelimit.Settings
+	provider RatelimitProvider
+}
+
+// RatelimitProvider returns a limit instance for provided Settings
+type RatelimitProvider interface {
+	get(s ratelimit.Settings) limit
+}
+
+type limit interface {
+	// AllowContext is used to decide if call is allowed to pass
+	AllowContext(context.Context, string) bool
+
+	// RetryAfter is used to inform the client how many seconds it
+	// should wait before making a new request
+	RetryAfter(string) int
+}
+
+// RegistryAdapter adapts ratelimit.Registry to RateLimitProvider interface.
+// ratelimit.Registry is not an interface and its Get method returns
+// ratelimit.Ratelimit which is not an interface either
+// RegistryAdapter narrows ratelimit interfaces to necessary minimum
+// and enables easier test stubbing
+type registryAdapter struct {
+	registry *ratelimit.Registry
+}
+
+func (a *registryAdapter) get(s ratelimit.Settings) limit {
+	return a.registry.Get(s)
+}
+
+func NewRatelimitProvider(registry *ratelimit.Registry) RatelimitProvider {
+	return &registryAdapter{registry}
 }
 
 // NewLocalRatelimit is *DEPRECATED*, use NewClientRatelimit, instead
-func NewLocalRatelimit() filters.Spec {
-	return &spec{typ: ratelimit.LocalRatelimit, filterName: ratelimit.LocalRatelimitName}
+func NewLocalRatelimit(provider RatelimitProvider) filters.Spec {
+	return &spec{typ: ratelimit.LocalRatelimit, provider: provider, filterName: ratelimit.LocalRatelimitName}
 }
 
 // NewClientRatelimit creates a instance based client rate limit.  If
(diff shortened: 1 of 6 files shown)


skipper.go
```diff
@@ -1111,6 +1112,84 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 		),
 	)
 
+	var swarmer ratelimit.Swarmer
```
AlexanderYastrebov, line 1115: This block of code moved up before custom filters are registered

skipper.go
```diff
@@ -1111,6 +1112,84 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
+	}
+
+	var ratelimitRegistry *ratelimit.Registry
+	if o.EnableRatelimiters || len(o.RatelimitSettings) > 0 {
+		log.Infof("enabled ratelimiters %v: %v", o.EnableRatelimiters, o.RatelimitSettings)
+		ratelimitRegistry := ratelimit.NewSwarmRegistry(swarmer, redisOptions, o.RatelimitSettings...)
+		defer ratelimitRegistry.Close()
+
+		provider := ratelimitfilters.NewRatelimitProvider(ratelimitRegistry)
+		o.CustomFilters = append(o.CustomFilters,
```
AlexanderYastrebov, line 1183: Registers ratelimit filters as customs filters

proxy/proxy.go
```diff
@@ -1099,12 +1054,6 @@ func (p *Proxy) do(ctx *context) error {
 
 	processedFilters := p.applyFiltersToRequest(ctx.route.Filters, ctx)
 
-	// per route rate limit
-	if settings, retryAfter := p.checkRatelimit(ctx); retryAfter > 0 {
```
AlexanderYastrebov, line 1103: With this change, global rate limits are handled as proxy errors while route ratelimits are handled as regular response with 429 code. Not sure if it is important.

ratelimit/ratelimit.go
```diff
@@ -422,3 +423,11 @@ func newRatelimit(s Settings, sw Swarmer, redisRing *ring) *Ratelimit {
 		impl:     impl,
 	}
 }
+
+func Headers(s *Settings, retryAfter int) http.Header {
+	limitPerHour := int64(s.MaxHits) * int64(time.Hour) / int64(s.TimeWindow)
```
AlexanderYastrebov, line 428: Calculates limit in `int64`. Before it was 
```go
settings.MaxHits * int(time.Hour/settings.TimeWindow)
```
which would produce incorrect result when `settings.TimeWindow > time.Hour`, e.g. `MaxHits = 2, TimeWindow = "2h"` would produce 0, whilte it should be 1.

filters/ratelimit/ratelimit.go
```diff
@@ -343,13 +379,30 @@ func getDurationArg(a interface{}) (time.Duration, error) {
...
+	}
+
+	s := f.settings.Lookuper.Lookup(ctx.Request())
+	if s == "" {
+		log.Debugf("Lookuper found no data in request for settings: %s and request: %v", f.settings, ctx.Request())
+		return
+	}
+
+	if !rateLimiter.AllowContext(ctx.Request().Context(), s) {
+		ctx.Serve(&http.Response{
```
AlexanderYastrebov, line 402: There is no body here. Global ratelimiter uses [http.Error](https://golang.org/pkg/net/http/#Error) that sends text/plain body.

filters/ratelimit/ratelimit.go
```diff
@@ -15,24 +16,51 @@ import (
...
 	filterName string
 }
 
 type filter struct {
 	settings ratelimit.Settings
+	provider RatelimitProvider
+}
+
+// RatelimitProvider returns a limit instance for provided Settings
+type RatelimitProvider interface {
```
aryszka, line 31: I am not completely convinced about the public interface with a private method, but I see the problem and the reason.

Alternatively:

1. we could move the RatelimitProvider interface to the ratelimit package (from filters/ratelimit)
2. the interface could have a public GetRatelimit method, returning an interface with the same methods, or better the necessarily required ones, as the ratelimit.Ratelimit structure, called e.g. Ratelimiter
3. the ratelimit.Registry would implement the ratelimit.RatelimitProvider interface, and would mark the current Get method deprecated

Then we could still conveniently stub the provider. WDYT?

AlexanderYastrebov: At first I have tried to adjust existing interfaces from `ratelimit` package but quickly fell into a tarpit. As I am and was not sure about deprecation and compatibility policy I have decided to workaround it in `filters/ratelimit` package.

The only downside of _sealed_ `RatelimitProvider` interface I see is that it prevents users from creating ratelimit filters from _outside_ of the `filters/ratelimit` package _and without_ `ratelimit.Registry` at hand. If that is needed (not sure why) it could be achieved by having an extra interface and making `get` public (though registry adapter would still be needed).

There is another option without new interfaces at all - make filters require something like `func CheckLimit(*http.Request, *Settings) (retryAfter int)` and create registry adapter for it.

The interface mismatch between what filters need and what ratelimit package provides feels to big.

I would like to address it later if and when we see this change has chances to land into mainline (i.e. there are no showstoppers).

aryszka: well, ok, though it still scratches my aesthetics a bit :smile:

szuecs:
:+1:

PR #1552, opened 2020-10-15
Implements filter resources cleanup support

### LINKED ISSUE #202: Allow to cleanup resources for filters

AlexanderYastrebov:
Extends routing to track requests that use actual route table.
After route table update it waits for completion of requests that use old route table
and calls `Close` method for each filter from old route table that implements `io.Closer`.

Updates development documentation and removes outdated `Close` methods -
Tokeninfo, tokenOidc and tokenintrospection filters reuse shared
authclient and encrypter and thus do not need to implement cleanup.

Fixes #202

### DIFF
--- a/docs/reference/development.md
+++ b/docs/reference/development.md
@@ -97,7 +97,7 @@ index 10d5769..da46fe0 100644
 
 Documentation for users should be done in `docs/`.
 
-```
+````
 diff --git a/docs/filters.md b/docs/filters.md
 index d3bb872..a877062 100644
 --- a/docs/filters.md
@@ -127,7 +127,7 @@ index d3bb872..a877062 100644
  ## oauthTokeninfoAnyScope
 
  If skipper is started with `-oauth2-tokeninfo-url` flag, you can use
-```
+````
 
 ### Add godoc
 
@@ -170,8 +170,8 @@ process. A spec has to satisfy the `Spec` interface `Name() string` and
 
 The actual filter implementation has to satisfy the `Filter`
 interface `Request(filters.FilterContext)` and `Response(filters.FilterContext)`.
-If you need to clean up for example a goroutine you can do it in
-`Close()`, which will be called on filter shutdown.
+If you need to clean up for example a goroutine then filter should implement
+`Close() error` method of `io.Closer` which will be called on filter shutdown.
 
 ```
 diff --git a/filters/auth/webhook.go b/filters/auth/webhook.go
@@ -256,13 +256,14 @@ index 0000000..f0632a6
 +func (*webhookFilter) Response(filters.FilterContext) {}
 +
 +// Close cleans-up the quit channel used for this filter
-+func (f *webhookFilter) Close() {
++func (f *webhookFilter) Close() error {
 +	f.authClient.mu.Lock()
 +	if f.authClient.quit != nil {
 +		close(f.authClient.quit)
 +		f.authClient.quit = nil
 +	}
 +	f.authClient.mu.Unlock()
++	return nil
 +}
 ```
 
--- a/docs/tutorials/development.md
+++ b/docs/tutorials/development.md
@@ -37,8 +37,8 @@ process. A spec has to satisfy the `filters.Spec` interface `Name() string` and
 
 The actual filter implementation has to satisfy the `filter.Filter`
 interface `Request(filters.FilterContext)` and `Response(filters.FilterContext)`.
-If you need to clean up for example a goroutine you can do it in
-`Close()`, which will be called on filter shutdown.
+If you need to clean up for example a goroutine then filter should implement
+`Close() error` method of `io.Closer` which will be called on filter shutdown.
 
 The simplest filter possible is, if `filters.Spec` and
 `filters.Filter` are the same type:
--- a/routing/routing.go
+++ b/routing/routing.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -259,7 +260,8 @@ type PreProcessor interface {
 // Routing ('router') instance providing live
 // updatable request matching.
 type Routing struct {
-	routeTable        atomic.Value // of struct routeTable
+	mx                sync.RWMutex
+	routeTable        *routeTable
 	log               logging.Logger
 	firstLoad         chan struct{}
 	firstLoadSignaled bool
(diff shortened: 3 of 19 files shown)


aryszka:
closing and reopening in order to retrigger the tests

routing/routing.go
```diff
@@ -238,7 +238,8 @@ type PreProcessor interface {
 // Routing ('router') instance providing live
 // updatable request matching.
 type Routing struct {
-	routeTable        atomic.Value // of struct routeTable
```
aryszka, line 241: when this atomic value was introduce, there was an actual performance improvement in the route lookup that made it worth it. Since then, many things have changed, including the Go runtime scheduler. Can we run a benchmark comparison before and after this change?

aryszka: found it! Checking... :smile:

aryszka: Nice! :+1:

```
BenchmarkRoutingGetAtomicValue/goroutines-1200-4         	45787898	        25.7 ns/op
BenchmarkRoutingGetAtomicValue/goroutines-1400-4         	54809168	        25.2 ns/op
BenchmarkRoutingGetAtomicValue/goroutines-1600-4         	43664787	        26.6 ns/op
BenchmarkRoutingGetAtomicValue/goroutines-1800-4         	43672563	        26.6 ns/op
BenchmarkRoutingGetAtomicValue/goroutines-2000-4         	41791792	        25.8 ns/op
BenchmarkRoutingGetAtomicValue/goroutines-2200-4         	45046342	        24.3 ns/op
BenchmarkRoutingGetAtomicValue/goroutines-2400-4         	45706029	        25.7 ns/op
BenchmarkRoutingGetAtomicValue/goroutines-2600-4         	40851520	        26.3 ns/op
BenchmarkRoutingGetMutex/goroutines-1200-4               	 9082144	       132 ns/op
BenchmarkRoutingGetMutex/goroutines-1400-4               	 9578566	       129 ns/op
BenchmarkRoutingGetMutex/goroutines-1600-4               	10937642	       132 ns/op
BenchmarkRoutingGetMutex/goroutines-1800-4               	16323981	       129 ns/op
BenchmarkRoutingGetMutex/goroutines-2000-4               	 9657950	       127 ns/op
BenchmarkRoutingGetMutex/goroutines-2200-4               	10629856	       130 ns/op
BenchmarkRoutingGetMutex/goroutines-2400-4               	12391550	       126 ns/op
BenchmarkRoutingGetMutex/goroutines-2600-4               	10572556	       123 ns/op
BenchmarkRoutingGetRWMutex/goroutines-1200-4             	17687150	        67.4 ns/op
BenchmarkRoutingGetRWMutex/goroutines-1400-4             	17627328	        66.8 ns/op
BenchmarkRoutingGetRWMutex/goroutines-1600-4             	17076651	        68.0 ns/op
BenchmarkRoutingGetRWMutex/goroutines-1800-4             	16905411	        66.4 ns/op
BenchmarkRoutingGetRWMutex/goroutines-2000-4             	16989332	        66.5 ns/op
BenchmarkRoutingGetRWMutex/goroutines-2200-4             	17167993	        66.4 ns/op
BenchmarkRoutingGetRWMutex/goroutines-2400-4             	17821053	        65.1 ns/op
BenchmarkRoutingGetRWMutex/goroutines-2600-4             	17468914	        67.1 ns/op
```

aryszka:
There is a possibility that some long lasting requests, eg long poll or websocket, would cause a leak, in case of frequent routing updates. Not sure atm how to prevent this situation.

aryszka:
@AlexanderYastrebov Can you please rebase this PR? I think that would fix the build failures.

routing/routing.go
AlexanderYastrebov: Yeah, with this ns/op values it does not look as a big deal to me

AlexanderYastrebov:
@aryszka 
Yes, it seems there is a problem with long running requests and it is a show stopper. I will think about what we can do about it.

Another thing is that filter support for `Close` was documented but functionality itself was missing, moreover I had to remove `Close` methods from 
* `tokenOidcFilter` - uses shared encrypter
* `tokenintrospectFilter` - uses shared authclient from the start #638
* `tokeninfo` - uses shared authclient from #740

It is only `webhookFilter` that needs it and it seems it could(should) also be also improved to reuse authclient which would obviate the need for `Close`as well.

So it turns out that no filter needs it at the moment which questions the whole functionality's value.

I think we can do one of the following:
* solve long running request issue and merge
* or cleanup docs and never-called `Close` methods and attempt to implement filter cleanup again someday in the future

AlexanderYastrebov:
I have moved `routeLookup.Release()` before possibly long-running (error)response is served.

docs/reference/development.md
```diff
@@ -97,7 +97,7 @@ index 10d5769..da46fe0 100644
 
 Documentation for users should be done in `docs/`.
 
-```
+````
```
aryszka, line 100: do four backquotes have a special meaning?

docs/reference/development.md
```diff
@@ -170,8 +170,8 @@ process. A spec has to satisfy the `Spec` interface `Name() string` and
 
 The actual filter implementation has to satisfy the `Filter`
 interface `Request(filters.FilterContext)` and `Response(filters.FilterContext)`.
-If you need to clean up for example a goroutine you can do it in
-`Close()`, which will be called on filter shutdown.
+If you need to clean up for example a goroutine then filter should implement
```
aryszka, line 173: "...then the filter..."

docs/tutorials/development.md
```diff
@@ -37,8 +37,8 @@ process. A spec has to satisfy the `filters.Spec` interface `Name() string` and
 
 The actual filter implementation has to satisfy the `filter.Filter`
 interface `Request(filters.FilterContext)` and `Response(filters.FilterContext)`.
-If you need to clean up for example a goroutine you can do it in
-`Close()`, which will be called on filter shutdown.
+If you need to clean up for example a goroutine then filter should implement
```
aryszka, line 40: "...then the filter..."

aryszka:
> - solve long running request issue and merge
> - or cleanup docs and never-called Close methods and attempt to implement filter cleanup again someday in the future

Unfortunately, I could not come up now with a solution for the long running requests. So we can move forward for now with the second option.

docs/reference/development.md
AlexanderYastrebov: The content of the block has three itself, four allow it

aryszka: OK, I have never seen it before. The rendered docs here: https://opensource.zalando.com/skipper/reference/development/#user-documentation look good to me, generated with mkdocs. Let's try how this will work out! :+1:

AlexanderYastrebov:
Closing for now, needs more thinking and understanding of long running request and routing table update interplay

Issue #1157, opened 2019-08-21
[Kubernetes] ExternalName Service is served for wrong Ingress


linki:
**Describe the bug**
Normally, when an Ingress references a wrong or missing Service then `skipper-ingress` serves 404 as the fallback. However, if there is at least one Ingress in the cluster that points to an ExternalName Service then `skipper-ingress` serves that ExternalName instead of 404.

**To Reproduce**
1. create an Ingress `foo` with a rule for `foo.playground.zalan.do` that points to a non-existing Service
2. curl `foo.playground.zalan.do` and see that it returns a 404
3. create another Ingress `bar` with a rule for `bar.playground.zalan.do` that points to Service `bar`
4. create Service `bar` to be an ExternalName for `google.com`
5. curl `bar.playground.zalan.do` and see that it serves google's search page
6. also curl again `foo.playground.zalan.do` and see that it now *also* serves google's search page

**Expected behavior**

Step 6 should not happen and `foo.playground.zalan.do` should still be a 404.

**Observed behavior**

`foo.playground.zalan.do` displays google's search page.

szuecs:
should be fixed in https://github.com/zalando/skipper/releases/tag/v0.11.166

PR #1554, opened 2020-10-16
Set HostRegexps for Service of type ExternalName

### LINKED ISSUE #1553: Non-matching requests are sent to route with Service of type ExternalName
**Describe the bug**
When an Ingress points to a Service of type `ExternalName`, skipper creates a route that matches every host. Requests that do not match a host/path configuration of any other route are then sent to that route.

**To Reproduce**
Ingress and Service definitions:
```yaml
---
kind: Ingress
metadata:
  name: my-external-service
  annotations:
  labels:
    app: my-external-service
spec:
  rules:
    - host: my-external-service.domain.local
      http:
        paths:
          - backend:
              serviceName: my-external-service
              servicePort: 443
---
kind: Service
apiVersion: v1
metadata:
  name: my-external-service
  labels:
    app: my-external-service
spec:
  type: ExternalName
  externalName: google.com
  ports:
    - port: 443
      name: https
      targetPort: 443
```

The resulting route:
```
kube_default__my_external_service______google_com: *
  -> setRequestHeader("Host", "google.com")
  -> "https://google.com:443";
```

**Expected behavior**

skipper only proxies requests to the host `my-external-service.domain.local` to `google.com`.

**Observed behavior**

skipper proxies all non-matching requests, e.g. `http://127.0.0.1:9090/foo` to `google.com`.

wndhydrnt:
Fixes #1553

### DIFF
--- a/dataclients/kubernetes/kube_test.go
+++ b/dataclients/kubernetes/kube_test.go
@@ -614,7 +614,7 @@ func TestIngressData(t *testing.T) {
 			),
 		)},
 		map[string]string{
-			"kube_foo__qux______www_zalando_de": "https://www.zalando.de:443",
+			"kube_foo__qux__www_zalando_de____www_zalando_de": "https://www.zalando.de:443",
 		},
 	}, {
 		"ignore ingress entries with missing Metadata",
--- a/dataclients/kubernetes/ingress.go
+++ b/dataclients/kubernetes/ingress.go
@@ -186,10 +186,11 @@ func convertPathRule(
 			return nil, e
 		}
 		return &eskip.Route{
-			Id:          routeID(ns, name, "", "", svc.Spec.ExternalName),
+			Id:          routeID(ns, name, host, "", svc.Spec.ExternalName),
 			BackendType: eskip.NetworkBackend,
 			Backend:     u,
 			Filters:     f,
+			HostRegexps: hostRegexp,
 		}, nil
 	} else {
 		protocol := "http"


dataclients/kubernetes/kube_test.go
```diff
@@ -614,7 +614,7 @@ func TestIngressData(t *testing.T) {
 			),
 		)},
 		map[string]string{
-			"kube_foo__qux______www_zalando_de": "https://www.zalando.de:443",
+			"kube_foo__qux__www_zalando_de____www_zalando_de": "https://www.zalando.de:443",
```
AlexanderYastrebov, line 617: The test is testing route id change but not the addition of `Host` predicate. 

I would be nice to modify test harness to check not only backend but rather whole route like, e.g.:
```go
map[string]string{
    "kube_foo__qux__www_zalando_de____www_zalando_de": "Host(/^www[.]zalando[.]de$/) -> setRequestHeader("Host", "www.zalando.de") -> "https://www.zalando.de:443""
```
Not sure if this is within the scope of this ticket though.

szuecs: out of scope, we have to change the tests similar to routegroup IMO

szuecs:
:+1:

szuecs:
@wndhydrnt thanks, I never had time to investigate this!
related: https://github.com/zalando/skipper/issues/1157

dataclients/kubernetes/kube_test.go
AlexanderYastrebov: There is `checkPrettyRoutes` that does the job.

aermakov-zalando:
👍

PR #1560, opened 2020-10-21
flush stream to client also on error path

szuecs:

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1458,12 +1458,12 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	defer func() {
 		if ctx.response != nil && ctx.response.Body != nil {
-			ctx.responseWriter.Flush()
 			err := ctx.response.Body.Close()
 			if err != nil {
 				p.log.Errorf("error during closing the response body: %v", err)
 			}
 		}
+		ctx.responseWriter.Flush()
 	}()
 
 	err = p.do(ctx)
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -836,6 +836,7 @@ func (p *Proxy) lookupRoute(ctx *context) (rt *routing.Route, params map[string]
 func (p *Proxy) sendError(c *context, id string, code int) {
 	addBranding(c.responseWriter.Header())
 	http.Error(c.responseWriter, http.StatusText(code), code)
+	c.responseWriter.Flush()
 	p.metrics.MeasureServe(
 		id,
 		c.metricsHost(),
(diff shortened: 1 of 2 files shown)


proxy/proxy.go
```diff
@@ -1231,6 +1232,7 @@ func (p *Proxy) serveResponse(ctx *context) {
 		p.tracing.setTag(ctx.proxySpan, ErrorTag, true)
 		p.tracing.setTag(ctx.proxySpan, StreamBodyEvent, StreamBodyError)
 		p.tracing.logStreamEvent(ctx.proxySpan, StreamBodyEvent, fmt.Sprintf("Failed to stream response: %v", err))
+		ctx.responseWriter.Flush()
```
AlexanderYastrebov, line 1235: Why not flush within `copyStream` itself?

ruiaraujo: @AlexanderYastrebov copyStream already flushes but it is not used for error conditions.

AlexanderYastrebov: Yes, and I propose to flush there right before returning the error

szuecs: I changed it to the defer in ServeHTTP, which seems to be most safe case for both cases.
The problem is that we have it also in sendError which is a different call path than streaming the body.

proxy/proxy.go
```diff
@@ -1458,6 +1458,7 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	defer func() {
 		if ctx.response != nil && ctx.response.Body != nil {
+			ctx.responseWriter.Flush()
```
AlexanderYastrebov, line 1461: I think there will be no context response in case of net error (that is handled by sendError), might worth to have separate unconditional defer...

proxy/proxy.go
szuecs: as discussed internally, use the one in defer of ServeHTTP

ruiaraujo:
:+1:

szuecs:
:+1:

andrewhowdencom:
:+1:

aryszka:
:+1:

AlexanderYastrebov:
@szuecs @aryszka  We need to test how this flush works with upgrade requests. Upgrade as far as I understand hijacks connection and [nullifies underlying response writer](https://github.com/golang/go/blob/2ad44158af373f68c1aef528738a5baade77d316/src/net/http/server.go#L2015)

Build tests show that [`Flush()`](https://github.com/golang/go/blob/2ad44158af373f68c1aef528738a5baade77d316/src/net/http/server.go#L1670) produces
```
2020/10/22 21:07:08 http: response.WriteHeader on hijacked connection from github.com/zalando/skipper/logging.(*LoggingWriter).Flush (loggingwriter.go:39)
```
[here](https://github.com/golang/go/blob/2ad44158af373f68c1aef528738a5baade77d316/src/net/http/server.go#L1137) and panics on nil writer [here]( https://github.com/golang/go/blob/2ad44158af373f68c1aef528738a5baade77d316/src/net/http/server.go#L1674)

szuecs:
@AlexanderYastrebov hijacked connections are not owned by us anymore so we should not call Flush() on them.

PR #1563, opened 2020-10-23
Refactors makeBackendRequest error handling

AlexanderYastrebov:
* Removes unreachable branches from net.Error status code mapping
* Splits roundTripper creation (fastcgi) and actual backend round trip

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -881,7 +881,6 @@ func (p *Proxy) makeUpgradeRequest(ctx *context, req *http.Request) error {
 }
 
 func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
-	var err error
 	req, endpoint, err := mapRequest(ctx.request, ctx.route, ctx.outgoingHost, p.flags.HopHeadersRemoval(), ctx.StateBag())
 	if err != nil {
 		p.log.Errorf("could not map backend request, caused by: %v", err)
@@ -900,6 +899,12 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
 
 		// We are not owner of the connection anymore.
 		return nil, &proxyError{handled: true}
+	}
+
+	roundTripper, err := p.getRoundTripper(ctx, req)
+	if err != nil {
+		p.log.Errorf("Failed to get roundtripper: %v", err)
+		return nil, &proxyError{err: err, code: http.StatusBadGateway}
 	}
 
 	bag := ctx.StateBag()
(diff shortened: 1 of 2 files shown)


proxy/proxy.go
```diff
@@ -963,31 +937,16 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
...
-					err:  err,
-					code: http.StatusInternalServerError,
-				}
+				status = http.StatusServiceUnavailable
+			}
+			p.tracing.setTag(ctx.proxySpan, HTTPStatusCodeTag, uint16(status))
+			return nil, &proxyError{
+				err:  err,
+				code: status,
 			}
```
AlexanderYastrebov, line 950: Pruning branches:
```go
// Status quo
if nerr.Timeout() {
	// http.StatusGatewayTimeout
} else if !nerr.Temporary() {
	// http.StatusServiceUnavailable
} else if !nerr.Timeout() && nerr.Temporary() {
	// http.StatusServiceUnavailable
} else {
	// http.StatusInternalServerError
}

// Remove always true !nerr.Timeout() from else branch 

if nerr.Timeout() {
	// http.StatusGatewayTimeout
} else if !nerr.Temporary() {
	// http.StatusServiceUnavailable
} else if nerr.Temporary() {
	// http.StatusServiceUnavailable
} else {
	// http.StatusInternalServerError
}

// Remove unreachable last else

if nerr.Timeout() {
	// http.StatusGatewayTimeout
} else if !nerr.Temporary() {
	// http.StatusServiceUnavailable
} else if nerr.Temporary() {
	// http.StatusServiceUnavailable
}

// Both !nerr.Temporary() and nerr.Temporary() lead to http.StatusServiceUnavailable

if nerr.Timeout() {
	// http.StatusGatewayTimeout
} else
	// http.StatusServiceUnavailable
}
```

proxy/proxy.go
```diff
@@ -915,38 +920,7 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
 	p.metrics.IncCounter("outgoing." + req.Proto)
 	ctx.proxySpan.LogKV("http_roundtrip", StartEvent)
 
-	var response *http.Response
```
AlexanderYastrebov, line 918: Moving switch into `getRoundTripper` does not change anything for `http` but brings back clear roundtrip point as it was before fastcgi was introduced:
https://github.com/zalando/skipper/blob/a0c9feef7d932305c6f0f9406a362d6063e92605/proxy/proxy.go#L883-L887

AlexanderYastrebov:
Just found #1131 that touches the same part though without switch branch pruning

proxy/proxy.go
```diff
@@ -1007,6 +966,33 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
...
+			f = sf.(string)
+		}
+		rt, err := fastcgi.NewRoundTripper(p.log, req.URL.Host, f)
+		if err != nil {
+			return nil, err
+		}
+
+		// FastCgi expects the Host to be in form host:port
+		// It will then be split and added as 2 separate params to the backend process
+		if _, _, err := net.SplitHostPort(req.Host); err != nil {
```
aryszka, line 983: I would move back these last two statements with changing the request host and the remote address to the original scope, because it doesn't really belong to "getting a roundtripper"

aryszka:
lgtm apart of the side effect in the `getRoundTripper` function

proxy/proxy.go
AlexanderYastrebov: This is an adhoc for fastcgi, originally it was within fastcgi switch case as well

aryszka: i see :+1:

aryszka:
:+1:

szuecs:
:+1:

Issue #1564, opened 2020-10-23
Flush on upgraded connections lead to panic 


szuecs:
**Describe the bug**
Flush socket also in error cases [PR](https://github.com/zalando/skipper/pull/1560) result in panic as 
[commented](https://github.com/zalando/skipper/pull/1560#issuecomment-714710929).

The bug was introduced in https://github.com/zalando/skipper/releases/tag/v0.11.167

**To Reproduce**
backend
```
wscat -l 1234
listening on port 1234 (press CTRL+C to quit)
client connected
< asd
< asd
C-c
```

proxy
```
% bin/skipper -experimental-upgrade -inline-routes '* -> "http://localhost:1234"'
[APP]INFO[0000] Expose metrics in codahale format
[APP]INFO[0000] support listener on :9911
[APP]INFO[0000] route settings, reset, route: : * -> "http://localhost:1234"
[APP]INFO[0000] route settings received
[APP]INFO[0000] route settings applied
[APP]INFO[0000] proxy listener on :9090
[APP]INFO[0000] TLS settings not found, defaulting to HTTP
[APP]ERRO[0021] error proxying data from src to dst: read tcp 127.0.0.1:38202->127.0.0.1:1234: read: connection reset by peer
127.0.0.1 - - [23/Oct/2020:14:50:02 +0200] "GET / HTTP/1.1" 0 0 "-" "-" 8922 localhost:9090 - -
2020/10/23 14:50:11 http: response.WriteHeader on hijacked connection from github.com/zalando/skipper/logging.(*LoggingWriter).Flush (loggingwriter.go:39)
2020/10/23 14:50:11 http: panic serving 127.0.0.1:33576: runtime error: invalid memory address or nil pointer dereference
goroutine 19 [running]:
net/http.(*conn).serve.func1(0xc000338140)
        /usr/share/go/src/net/http/server.go:1801 +0x147
panic(0x139ed20, 0x1df21f0)
        /usr/share/go/src/runtime/panic.go:975 +0x3e9
bufio.(*Writer).Flush(0x0, 0xc8, 0x7fb5481ae4a8)
        /usr/share/go/src/bufio/bufio.go:601 +0x26
net/http.(*response).Flush(0xc0003122a0)
        /usr/share/go/src/net/http/server.go:1673 +0x35
github.com/zalando/skipper/logging.(*LoggingWriter).Flush(...)
        /home/sszuecs/go/src/github.com/zalando/skipper/logging/loggingwriter.go:39
[...]

PR #1565, opened 2020-10-23
fix upgrade connections create panic logs on connection tear down

szuecs:
fix https://github.com/zalando/skipper/issues/1564
hijacked connections panic if we flush on so mark via statebag that we are upgraded and if not we flush

### DIFF
--- a/proxy/context.go
+++ b/proxy/context.go
@@ -29,6 +29,7 @@ type context struct {
 	route                *routing.Route
 	deprecatedServed     bool
 	servedWithResponse   bool // to support the deprecated way independently
+	successfulUpgrade   bool
 	pathParams           map[string]string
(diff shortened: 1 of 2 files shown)


andrewhowdencom:
With limited context, LGTM. :+1:

AlexanderYastrebov:
There are two response branches: sendError and serveResponse. serveResponse flushes within copyStream so why not simply flush in sendError?

AlexanderYastrebov:
offtopic: IMO we should strive to eliminate ServeHTTP defers rather then pile into them and couple with statebag

szuecs:
:+1:

aryszka:
:+1:

szuecs:
:+1:

Issue #1041, opened 2019-04-24
More fine-grained timeout handling


otrosien:
**Is your feature request related to a problem? Please describe.**

Our backend is expected to fail with a timeout after 450ms, but in case of massive traffic it starts garbage collecting pauses etc, so ideally skipper in front of my application should react and send a 504 Gateway Timeout to the client. Unfortunately there is only a global backend timeout configurable in skipper, and we're sharing skippers ingress with other applications.

**Describe the solution you would like**

Ideally I want to configure a timeout to be configurable per route, or per backend...

**Describe alternatives you've considered (optional)**

As a work-around we could invest in hardening the backend to ensure it fails faster, instead of trying to handle more connections and getting into GC issues.

Another alternative could be a dedicated skipper-ingress for our application that hard-codes our desired timeout setting.

**Additional context (optional)**

Why failing fast is important here? If we're not responding within the expected timeout, we're burdening our upstream clients with our increased latency, causing them to handle more concurrent connections, and making it harder for them to trigger their circuit breakers.

**Would you like to work on it?**

Maybe.

szuecs:
Thanks for the issue. I think we could start to work on it if we do configurable timeouts per backend (group).

AlexanderYastrebov:
I will post a comment instead of a new issue.

## Problem

**TL;DR**: There is no "response read timeout" available and therefore skipper hangs if backend hangs until client cancels the request. 

Skipper has the following timeout configurations (look for `*timeoutBackend*`) available regarding backend:
https://github.com/zalando/skipper/blob/5a92e9fd9abc86e2fbf925afb08d13932fec74e4/config/config.go#L178-L183
https://github.com/zalando/skipper/blob/5a92e9fd9abc86e2fbf925afb08d13932fec74e4/config/config.go#L386-L391

For backend requests skipper uses `http.Transport`:
https://github.com/zalando/skipper/blob/5a92e9fd9abc86e2fbf925afb08d13932fec74e4/proxy/proxy.go#L623-L637

* `TimeoutBackend` is used to configure `net.Dialer` timeout:
```
    // Timeout is the maximum amount of time a dial will wait for
    // a connect to complete. If Deadline is also set, it may fail
    // earlier.
    //
    // The default is no timeout.
    //
    // When using TCP and dialing a host name with multiple IP
    // addresses, the timeout may be divided between them.
    //
    // With or without a timeout, the operating system may impose
    // its own earlier timeout. For instance, TCP timeouts are
[...]

PR #1566, opened 2020-10-23
Implements route backend timeout

AlexanderYastrebov:
* Adds `backendTimeout` filter to configure route backend timeout
* Proxy sets up request context with configured timeout and responds
with 504 status on timeout (note: if response streaming has already
started it will be terminated, client will receive backend status and
truncated response body).

See #1041

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -521,6 +521,23 @@ Example:
 * -> repeatContent("I will not waste chalk. ", 1000) -> <shunt>;
 ```
 
+## backendTimeout
+
+Configure backend timeout. Skipper responds with `504 Gateway Timeout` status if obtaining a connection,
+sending the request, and reading the backend response headers and body takes longer than the configured timeout.
+However, if response streaming has already started it will be terminated, i.e. client will receive backend response
+status and truncated response body.
+
+Parameters:
+
+* timeout [(duration string)](https://godoc.org/time#ParseDuration)
+
+Example:
+
+```
+* -> backendTimeout("10ms") -> "https://www.example.org";
+```
+
 ## latency
 
 Enable adding artificial latency
--- a/filters/builtin/timeout.go
+++ b/filters/builtin/timeout.go
@@ -1,0 +1,45 @@
+package builtin
+
+import (
+	"time"
+
+	"github.com/zalando/skipper/filters"
+)
+
+type timeout struct {
+	timeout time.Duration
+}
+
+func NewBackendTimeout() filters.Spec {
+	return &timeout{}
+}
+
+func (*timeout) Name() string { return BackendTimeoutName }
+
+func (*timeout) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) != 1 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	var tf timeout
+	switch v := args[0].(type) {
+	case string:
+		d, err := time.ParseDuration(v)
+		if err != nil {
+			return nil, err
+		}
+		tf.timeout = d
+	case time.Duration:
+		tf.timeout = v
+	default:
+		return nil, filters.ErrInvalidFilterParameters
+	}
+	return &tf, nil
+}
+
+func (t *timeout) Request(ctx filters.FilterContext) {
+	// allows overwrite
+	ctx.StateBag()[filters.BackendTimeout] = t.timeout
+}
+
+func (t *timeout) Response(filters.FilterContext) {}
--- a/proxy/backendtimeout_test.go
+++ b/proxy/backendtimeout_test.go
@@ -1,0 +1,218 @@
+package proxy
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+)
+
+func TestSlowService(t *testing.T) {
+	wait := make(chan struct{})
+
+	service := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		<-wait
+	}))
+	defer func() {
+		close(wait)
+		service.Close()
+	}()
+
+	doc := fmt.Sprintf(`* -> backendTimeout("1ms") -> "%s"`, service.URL)
+	tp, err := newTestProxy(doc, FlagsNone)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tp.close()
+	if testing.Verbose() {
+		tp.log.Unmute()
+	}
+
+	ps := httptest.NewServer(tp.proxy)
+	defer ps.Close()
+
+	rsp, err := http.Get(ps.URL)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if rsp.StatusCode != http.StatusGatewayTimeout {
+		t.Errorf("expected 504, got: %v", rsp)
+	}
+}
+
+func TestFastService(t *testing.T) {
+	service := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		time.Sleep(1 * time.Millisecond)
+	}))
+	defer service.Close()
+
+	doc := fmt.Sprintf(`* -> backendTimeout("10ms") -> "%s"`, service.URL)
+	tp, err := newTestProxy(doc, FlagsNone)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tp.close()
+	if testing.Verbose() {
+		tp.log.Unmute()
+	}
+
+	ps := httptest.NewServer(tp.proxy)
+	defer ps.Close()
+
+	rsp, err := http.Get(ps.URL)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if rsp.StatusCode != http.StatusOK {
+		t.Errorf("expected 200, got: %v", rsp)
+	}
+}
+
+func TestBackendTimeoutInTheMiddleOfServiceResponse(t *testing.T) {
+	service := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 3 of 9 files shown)


AlexanderYastrebov:
If response truncation is not acceptable I see three ways out:
* declare it a feature for limited usecases :)
* try to workaround/disable streaming termination
* attempt another approach - consider timeout as `Transport.ResponseHeaderTimeout`, keep a pool of transports per configured timeout (cleanup?) and wrap roundtripper around here https://github.com/zalando/skipper/blob/8f32650afa693de7642e87026af0c6f39af179e9/proxy/proxy.go#L951

AlexanderYastrebov:
There is a general problem to notify client about backend that has stuck. At the moment `copyStream` stops writing on backend read error: 
https://github.com/zalando/skipper/blob/8f32650afa693de7642e87026af0c6f39af179e9/proxy/proxy.go#L411-L421
which leads to truncated response. 

`httputil.ReverseProxy` uses [`panic(http.ErrAbortHandler)`](https://github.com/golang/go/blob/d75b7f9648bb4968a5d980c4a4b7aa6ca22c4de5/src/net/http/httputil/reverseproxy.go#L328-L339), that [closes client connection](https://github.com/golang/go/blob/d75b7f9648bb4968a5d980c4a4b7aa6ca22c4de5/src/net/http/server.go#L1813-L1824), see https://github.com/golang/go/issues/23643 (for http/2 it seems to [close all client streams](https://github.com/golang/go/blob/d75b7f9648bb4968a5d980c4a4b7aa6ca22c4de5/src/net/http/h2_bundle.go#L7405-L7411))

aryszka:
> If response truncation is not acceptable I see three ways out:
> ...
> try to workaround/disable streaming termination

I believe this is possible if we implement our variant of context.Context and signal to it various phases like request headers sent, request stream done, response headers received, response stream done. But in my opinion, the current implementation is a good start, and covers a reasonable majority of the use cases.

filters/builtin/timeout.go
```diff
@@ -0,0 +1,47 @@
...
+		tf.timeout = v
+	default:
+		return nil, filters.ErrInvalidFilterParameters
+	}
+	return &tf, nil
+}
+
+func (t *timeout) Request(ctx filters.FilterContext) {
+	sb := ctx.StateBag()
+	if _, ok := sb[filters.BackendTimeout]; !ok { // once
```
aryszka, line 42: I would allow overriding. Eg it may make sense when the a default timeout is set via the global default filters. Just drop the if.

docs/reference/filters.md
```diff
@@ -466,6 +466,23 @@ Example:
 * -> randomContent(42) -> <shunt>;
 ```
 
+## backendTimeout
+
+Configure backend timeout. Skipper responds with `504 Gateway Timeout` status if obtaining a connection,
+sending the request, and reading the backend response headers and body takes longer than the configured timeout.
+However, if response streaming has already started it will be terminated, i.e. client will receive backend response
+status and truncated response body.
```
aryszka, line 474: I think we should include here that if the request body is received too slowly, that also cancels the flow.

aryszka:
I would allow multiple filters where the last one would take effect (see my comment for the `timeout.Request` method. Apart of that, lgtm

docs/reference/filters.md
AlexanderYastrebov: Doesn't `sending the request` covers it?

filters/builtin/timeout.go
AlexanderYastrebov: Done

proxy/backendtimeout_test.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+}
+
+func TestFastService(t *testing.T) {
+	service := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		time.Sleep(2 * time.Millisecond)
+	}))
+	defer service.Close()
+
+	doc := fmt.Sprintf(`* -> backendTimeout("10ms") -> "%s"`, service.URL)
```
szuecs, line 44: this might always fail in CI tests

proxy/backendtimeout_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"time"
+)
+
+func TestSlowService(t *testing.T) {
+	service := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		time.Sleep(2 * time.Millisecond)
+	}))
+	defer service.Close()
+
+	doc := fmt.Sprintf(`* -> backendTimeout("1ms") -> "%s"`, service.URL)
```
szuecs, line 18: this might always be ok in CI tests

proxy/proxy.go
```diff
@@ -1237,6 +1246,7 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 	}
 
 	if span := ot.SpanFromContext(ctx.Request().Context()); span != nil {
+		p.tracing.setTag(span, ErrorTag, true)
```
szuecs, line 1249: this is an unrelated change, which can hurt some of our production teams

szuecs: ah got it, was moved from ServeHTTP

szuecs:
Code looks good to me.
I am pretty sure tests will hurt in the future (1ms/2ms sleeps to separate cases will break too often)

docs/reference/filters.md
aryszka: true, sgtm

proxy/backendtimeout_test.go
aryszka: one way to test this better could be a service that never responds:

```
<-testDone
```

proxy/backendtimeout_test.go
aryszka: here we can set a higher TO, e.g. 1s. I don't see a really good way to test this in a more "scientifically" exact way.

AlexanderYastrebov:
I have rebased the PR. I have to think about tests, at the moment I don't quite see how it will hurt. But I can assure you I don't like flaky tests at all.

proxy/backendtimeout_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+)
+
+func TestSlowService(t *testing.T) {
+	service := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		time.Sleep(10 * time.Millisecond)
```
szuecs, line 14: replace with:
<-testDone

proxy/backendtimeout_test.go
```diff
@@ -0,0 +1,200 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+)
+
+func TestSlowService(t *testing.T) {
```
szuecs, line 12: create testDone channel:
testDone := make(chan struct{})

proxy/backendtimeout_test.go
```diff
@@ -0,0 +1,200 @@
...
+	defer ps.Close()
+
+	rsp, err := http.Get(ps.URL)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if rsp.StatusCode != http.StatusGatewayTimeout {
+		t.Errorf("expected 504, got: %v", rsp)
+	}
```
szuecs, line 35: close(testChan)

proxy/backendtimeout_test.go
szuecs: true let's keep like this and see

proxy/backendtimeout_test.go
AlexanderYastrebov: Fixed

proxy/proxy.go
```diff
@@ -924,10 +927,16 @@ func (p *Proxy) makeBackendRequest(ctx *context) (*http.Response, *proxyError) {
...
 
 		// Check if the request has been cancelled or timed out
-		// The roundtrip error `err` may be different
+		// The roundtrip error `err` may be different:
+		// - for `Canceled` it could be either the same `context canceled` or `unexpected EOF` (net.OpError)
+		// - for `DeadlineExceeded` it is net.Error(timeout=true, temporary=true) wrapping this `context deadline exceeded`
 		if cerr := req.Context().Err(); cerr != nil {
 			ctx.proxySpan.LogKV("event", "error", "message", cerr.Error())
-			return nil, &proxyError{err: cerr, code: 499}
+			if cerr == stdlibcontext.Canceled {
```
AlexanderYastrebov, line 935: This now covers client cancellation and backend request timeout

proxy/proxy.go
```diff
@@ -1153,6 +1167,7 @@ func (p *Proxy) serveResponse(ctx *context) {
 	copyHeader(ctx.responseWriter.Header(), ctx.response.Header)
 	p.tracing.logStreamEvent(ctx.proxySpan, StreamHeadersEvent, EndEvent)
 
+	// TODO differentiate client cancelation and backend timeout before streaming
```
AlexanderYastrebov, line 1170: Needs to be implemented and test covered, see similar https://github.com/zalando/skipper/pull/1566#discussion_r538839840

AlexanderYastrebov: Added the test for client cancellation. This codepath is not a subject of backendTimeout as it configures timeout on backend request, not on client request.

AlexanderYastrebov:
I have rebased and updated tests as suggested

proxy/proxy.go
szuecs: nice!

szuecs:
:+1:

aryszka:
Thanks a lot!

aryszka:
:+1:

PR #1568, opened 2020-10-25
Fixes copyStream edge case

AlexanderYastrebov:
Writes n > 0 bytes returned before considering the read error (see https://golang.org/pkg/io/#Reader)

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -413,24 +413,21 @@ func copyStream(to flushedResponseWriter, from io.Reader, tracing *proxyTracing,
 
 	for {
 		l, rerr := from.Read(b)
-
-		tracing.logStreamEvent(span, StreamBodyEvent, fmt.Sprintf("%d", l))
-
-		if rerr != nil && rerr != io.EOF {
-			return rerr
-		}
-
 		if l > 0 {
+			tracing.logStreamEvent(span, StreamBodyEvent, strconv.Itoa(l))
+
 			_, werr := to.Write(b[:l])
 			if werr != nil {
 				return werr
 			}
-
 			to.Flush()
 		}
-
-		if rerr == io.EOF {
-			return nil
+		if rerr != nil {
+			if rerr != io.EOF {
+				return rerr
+			} else {
+				return nil
+			}
 		}
 	}
 }


szuecs:
:+1:

ruiaraujo:
:+1:

szuecs:
@AlexanderYastrebov needs rebase master

proxy/proxy.go
```diff
@@ -413,24 +413,21 @@ func copyStream(to flushedResponseWriter, from io.Reader, tracing *proxyTracing,
 
 	for {
 		l, rerr := from.Read(b)
-
-		tracing.logStreamEvent(span, StreamBodyEvent, fmt.Sprintf("%d", l))
-
-		if rerr != nil && rerr != io.EOF {
-			return rerr
```
aryszka, line 420: I am not so sure which one is better. My line of thinking here is that we have no reason to attempt a write if the read has failed with a non-EOF error.

aryszka: am I missing a real-life consequence?

AlexanderYastrebov: Read may fail with non-eof and still return some data according to its contract https://golang.org/pkg/io/#Reader
> Callers should always process the n > 0 bytes returned before considering the error err. Doing so correctly handles I/O errors that happen after reading some bytes and also both of the allowed EOF behaviors.

IMO Skipper as a proxy should attempt to pass as much as possible back to the client even if the whole roundtrip ends up with error.

TBH this loop is what `io.CopyBuffer` does and I personally would prefer to use library implementation like so
```diff
diff --git a/proxy/proxy.go b/proxy/proxy.go
index cb9f072..b59ea3d 100644
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -404,37 +404,28 @@ func cloneHeaderExcluding(h http.Header, excludeList map[string]bool) http.Heade
        return hh
 }
 
+type flusher struct {
+       w flushedResponseWriter
+}
+
+func (f *flusher) Write(p []byte) (n int, err error) {
+       n, err = f.w.Write(p)
+       if n > 0 && err != nil {
+               f.w.Flush()
+       }
+       return
+}
+
 // copies a stream with flushing on every successful read operation
 // (similar to io.Copy but with flushing)
 func copyStream(to flushedResponseWriter, from io.Reader, tracing *proxyTracing, span ot.Span) error {
        b := make([]byte, proxyBufferSize)
 
-       var bytesCopied int
-       defer func() {
-               tracing.logStreamEvent(span, StreamBodyEvent, fmt.Sprintf("%d", bytesCopied))
-       }()
-
-       for {
-               l, rerr := from.Read(b)
+       n, err := io.CopyBuffer(&flusher{to}, struct{ io.Reader }{from}, b) // https://github.com/golang/go/issues/16474
 
-               if rerr != nil && rerr != io.EOF {
-                       return rerr
-               }
-
-               if l > 0 {
-                       _, werr := to.Write(b[:l])
-                       if werr != nil {
-                               return werr
-                       }
+       tracing.logStreamEvent(span, StreamBodyEvent, strconv.FormatInt(n, 10))
 
-                       to.Flush()
-                       bytesCopied += l
-               }
-
-               if rerr == io.EOF {
-                       return nil
-               }
-       }
+       return err
 }
```

I can PR from the above diff, this PR I would close once we finish the discussion, makes no sense to rebase.

aryszka: I think your proposal makes sense.

I wonder if we just missed the release of CopyBuffer by a couple of months back then: https://github.com/golang/go/commit/2c89992f445a631da250517d6f9b9fcd7852872e :smile:

AlexanderYastrebov:
Closing in favour of https://github.com/zalando/skipper/pull/1627

PR #1570, opened 2020-10-26
Improves randomContent filter performance

AlexanderYastrebov:
* Uses `Context.Serve` and `io.LimitReader` to use constant memory regardless of generated length
* Uses local `rand.Rand` to avoid global rand contention and locking per character generated

Signed-off-by: Alexander Yastrebov <alexander.yastrebov@zalando.de>

PR #1578, opened 2020-10-30
use an optimized way to generate a less random content

aryszka:

### DIFF
--- a/filters/diag/diag.go
+++ b/filters/diag/diag.go
@@ -9,7 +9,6 @@ package diag
 
 import (
 	"io"
-	"math/rand"
 	"net/http"
 	"time"
 
@@ -42,7 +41,8 @@ const (
 )
 
 type random struct {
-	len int
+	reader io.Reader
+	len    int
 }
 
 type throttle struct {
@@ -50,8 +50,6 @@ type throttle struct {
 	chunkSize int
 	delay     time.Duration
 }
-
-var randomChars = []byte("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789")
 
 func kbps2bpms(kbps float64) float64 {
 	return kbps * 1024 / 1000
@@ -120,31 +118,15 @@ func (r *random) CreateFilter(args []interface{}) (filters.Filter, error) {
 	}
 
 	if l, ok := args[0].(float64); ok {
-		return &random{int(l)}, nil
+		return &random{reader: weakRandom(), len: int(l)}, nil
 	} else {
 		return nil, filters.ErrInvalidFilterParameters
 	}
 }
 
 func (r *random) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
-	for n := 0; n < r.len; {
-		l := defaultChunkSize
-		if n+l > r.len {
-			l = r.len - n
-		}
-
-		b := make([]byte, l)
-		for i := 0; i < l; i++ {
-			b[i] = randomChars[rand.Intn(len(randomChars))]
-		}
-
-		ni, err := w.Write(b)
-		if err != nil {
-			log.Error("error while writing random content", err)
-			return
-		}
-
-		n += ni
+	if _, err := io.Copy(w, io.LimitReader(r.reader, int64(r.len))); err != nil {
+		log.Error("error while writing random content", err)
 	}
 }
 
--- a/filters/diag/weakrand.go
+++ b/filters/diag/weakrand.go
@@ -1,0 +1,46 @@
+package diag
+
+import (
+	"math/rand"
+	"sync"
+)
+
+type reader struct {
+	c  []byte
+	i  int
+	mx *sync.Mutex
+}
+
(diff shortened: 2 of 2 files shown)


szuecs:
no tests?
Interesting test would be reading more than `len(randomChars)`.

szuecs:
👍

filters/diag/diag.go
```diff
@@ -120,31 +118,15 @@ func (r *random) CreateFilter(args []interface{}) (filters.Filter, error) {
...
-		}
-
-		ni, err := w.Write(b)
-		if err != nil {
-			log.Error("error while writing random content", err)
-			return
-		}
-
-		n += ni
+	if _, err := io.Copy(w, io.LimitReader(r.reader, int64(r.len))); err != nil {
```
AlexanderYastrebov, line 128: Is it better than
```go
func (r *random) Request(ctx filters.FilterContext) {
	ctx.Serve(&http.Response{
		StatusCode: http.StatusOK,
		Body:       ioutil.NopCloser(io.LimitReader(r.reader, r.len)),
	})
}
```

filters/diag/weakrand.go
```diff
@@ -0,0 +1,46 @@
...
+var randomChars = []byte("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789")
+
+func weakRandom() *reader {
+	c := make([]byte, len(randomChars))
+	copy(c, randomChars)
+	rand.Shuffle(len(c), func(i, j int) {
+		c[i], c[j] = c[j], c[i]
+	})
+
+	return &reader{c: c, mx: &sync.Mutex{}}
```
AlexanderYastrebov, line 23: I think there is no need to use mutex pointer

filters/diag/weakrand.go
```diff
@@ -0,0 +1,46 @@
...
+		l := copy(p, r.c[i:])
+		p = p[l:]
+		n += l
+		i += l
+		if i >= len(r.c) {
+			i = 0
+		}
+	}
+
+	r.mx.Lock()
```
AlexanderYastrebov, line 42: I think we know that returned length is always `len(p)` and also do not need second mutex lock:
```diff
diff --git a/filters/diag/weakrand.go b/filters/diag/weakrand.go
index 4ad7185..4c92ff5 100644
--- a/filters/diag/weakrand.go
+++ b/filters/diag/weakrand.go
@@ -24,23 +24,20 @@ func weakRandom() *reader {
 }
 
 func (r *reader) Read(p []byte) (int, error) {
+       n := len(p)
+
        r.mx.Lock()
        i := r.i
+       r.i = (r.i + n) % len(r.c)
        r.mx.Unlock()
 
-       var n int
        for len(p) > 0 {
                l := copy(p, r.c[i:])
                p = p[l:]
-               n += l
                i += l
-               if i >= len(r.c) {
+               if i == len(r.c) {
                        i = 0
                }
        }
-
-       r.mx.Lock()
-       r.i = i
-       r.mx.Unlock()
        return n, nil
 }
```

AlexanderYastrebov:
Surely gives better performance than https://github.com/zalando/skipper/pull/1570#issuecomment-716853285
```
$ echo "GET http://localhost:9090/" | vegeta attack -duration=1m -rate=0 --max-workers=500 | vegeta report
Requests      [total, rate, throughput]         1058794, 17641.55, 17639.92
Duration      [total, attack, wait]             1m0s, 1m0s, 5.533ms
Latencies     [min, mean, 50, 90, 95, 99, max]  118.308µs, 14.795ms, 11.721ms, 32.977ms, 39.923ms, 55.059ms, 225.154ms
Bytes In      [total, mean]                     8673640448, 8192.00
Status Codes  [code:count]                      200:1058794  
Error Set:
```

filters/diag/weakrand.go
```diff
@@ -0,0 +1,46 @@
...
+
+type reader struct {
+	c  []byte
+	i  int
+	mx *sync.Mutex
+}
+
+var randomChars = []byte("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789")
+
+func weakRandom() *reader {
```
AlexanderYastrebov, line 16: It is really weak, better not ask for more than 62 characters :)

aryszka: :smile: 

yeah.. So the thing is that we don't want to turn Skipper into a real random generator, we just want to generate some data that is not too predicable for diagnostic purposes. https://xkcd.com/221/

filters/diag/weakrand.go
aryszka: nice! Thanks a lot!

filters/diag/diag.go
aryszka: No, your suggestion is better because it involves less moving parts. Thanks!

aryszka:
@AlexanderYastrebov 

I need to apologize here. I have a couple of hasty days behind me, and I completely missed that you opened a pull request doing the same thing. Sorry!

Let's discuss tomorrow day-time, which approach to go with. Your PR looks good to me, too.

aryszka:
Closing this in favor of: https://github.com/zalando/skipper/pull/1570

PR #1589, opened 2020-11-04
Support multiple files in `-routes-file`

ruudk:
Support multiple files in `-routes-file`

### DIFF
--- a/skipper.go
+++ b/skipper.go
@@ -221,10 +221,11 @@ type Options struct {
 	// *DEPRECATED* The whitespace separated list of OAuth2 scopes.
 	OAuthScope string
 
-	// File containing static route definitions.
+	// File containing static route definitions. Multiple may be given comma separated.
 	RoutesFile string
 
-	// File containing route definitions with file watch enabled. (For the skipper
+	// File containing route definitions with file watch enabled.
+	// Multiple may be given comma separated. (For the skipper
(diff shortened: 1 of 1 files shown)


szuecs:
This is a breaking change and maybe it would be better to have a `RoutesFiles`, that is used by `RoutesFile`, such that it is a non breaking change.

ruudk:
@szuecs Hmm, but that would mean that I also have to introduce a new Config property like this?:
```golang
RoutesFile                string            `yaml:"routes-file"`
RoutesFiles              *listFlag        `yaml:"routes-files"`
```
?

That would make it a huge mess imho.

szuecs:
@ruudk yes the other you can add a deprecation warning and add a log in case it is used with warning level, such that skipper users get a notice that they should migrate to the new flag.

ruudk:
And how do you suggest that the CLI interface works?

szuecs:
If possible:
`skipper -routes-file foo.eskip -routes-file bar.eskip`
or use comma separated (we have this already)
`skipper -routes-file foo.eskip,bar.eskip`

ruudk:
@szuecs @aryszka Changed the implementation. It's now much simpler.

szuecs:
:+1:

aryszka:
:+1:

Issue #1593, opened 2020-11-06
Hijacked connection panics when flush


ardaguclu:
**Describe the bug**
We are using skipper with upgrading and hijacking request to our websocket service. It was working without any problem with the 0.11.1 version. After upgrading to 0.11.182(with trying 0.11.177, etc.), it started getting panics while closing the connection.

`gateway           | 2020/11/05 07:32:26 http: response.WriteHeader on hijacked connection from github.com/zalando/skipper/logging.(*LoggingWriter).Flush (loggingwriter.go:39)
gateway           | 2020/11/05 07:32:26 http: panic serving 172.18.0.4:53380: runtime error: invalid memory address or nil pointer dereference
gateway           | goroutine 6346 [running]:
gateway           | net/http.(*conn).serve.func1(0xc0001377c0)
gateway           | 	/usr/local/go/src/net/http/server.go:1801 +0x147
gateway           | panic(0x137d5c0, 0x1bcf1f0)
gateway           | 	/usr/local/go/src/runtime/panic.go:975 +0x47a
gateway           | bufio.(*Writer).Flush(0x0, 0xc8, 0x7f56a8a88ad8)
gateway           | 	/usr/local/go/src/bufio/bufio.go:601 +0x26
gateway           | net/http.(*response).Flush(0xc00053e8c0)
gateway           | 	/usr/local/go/src/net/http/server.go:1673 +0x35
gateway           | github.com/zalando/skipper/logging.(*LoggingWriter).Flush(...)
gateway           | 	/go/pkg/mod/github.com/zalando/skipper@v0.11.182/logging/loggingwriter.go:39
gateway           | github.com/zalando/skipper/proxy.(*Proxy).ServeHTTP.func2(0xc000435b60, 0xc0003f61c0, 0xc0005510a0, 0xc000435c08)
gateway           | 	/go/pkg/mod/github.com/zalando/skipper@v0.11.182/proxy/proxy.go:1461 +0xfb
gateway           | github.com/zalando/skipper/proxy.(*Proxy).ServeHTTP(0xc0003f61c0, 0x16162c0, 0xc00053e8c0, 0xc00080b900)
gateway           | 	/go/pkg/mod/github.com/zalando/skipper@v0.11.182/proxy/proxy.go:1491 +0x5a9
gateway           | net/http.serverHandler.ServeHTTP(0xc0003f62a0, 0x16162c0, 0xc00053e8c0, 0xc00080b800)
[...]

PR #1594, opened 2020-11-06
Remove hijacked connection flush

ardaguclu:
https://github.com/zalando/skipper/issues/1593

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1456,6 +1456,11 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 			logging.LogAccess(entry, additionalData)
 		}
+
+		// This flush is required in I/O error
+		if !ctx.successfulUpgrade {
+			lw.Flush()
+		}
 	}()
 
 	if p.flags.patchPath() {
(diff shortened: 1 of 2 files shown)


proxy/proxy.go
```diff
@@ -1456,10 +1456,6 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 			logging.LogAccess(entry, additionalData)
 		}
-		// This flush is required in I/O error
-		if ctx.successfulUpgrade {
-			lw.Flush()
```
szuecs, line 1461: then we need to make sure we have a Flush in the error path, I think in line https://github.com/zalando/skipper/blob/master/proxy/proxy.go#L847

AlexanderYastrebov: Hi. The full story goes like this: we needed to flush error response and [it was introduced](https://github.com/zalando/skipper/pull/1560/files). Then it was found that [upgrade connections suffer](https://github.com/zalando/skipper/issues/1564) and [this fix](https://github.com/zalando/skipper/pull/1565/files) was done. Apparently the last fix passed review in rush because it simply needs to be negated to work
```go
		if !ctx.successfulUpgrade {
			lw.Flush()
		}
```

ardaguclu: @szuecs @AlexanderYastrebov, is this the correct fix for this problem?

ardaguclu:
@szuecs @AlexanderYastrebov, is this the correct fix for this problem?

szuecs:
:+1:

aryszka:
:+1:

PR #1595, opened 2020-11-06
drop excessive span logs

vetinari:
this is removing the `streamBody.byte:` logs which make up the biggest chunk of logs per proxy span.
Errors are still logged with the `streamBody.byte:` log.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -410,6 +410,11 @@ func cloneHeaderExcluding(h http.Header, excludeList map[string]bool) http.Heade
 func copyStream(to flushedResponseWriter, from io.Reader, tracing *proxyTracing, span ot.Span) error {
 	b := make([]byte, proxyBufferSize)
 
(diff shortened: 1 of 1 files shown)


proxy/proxy.go
```diff
@@ -413,8 +413,6 @@ func copyStream(to flushedResponseWriter, from io.Reader, tracing *proxyTracing,
 	for {
 		l, rerr := from.Read(b)
 
-		tracing.logStreamEvent(span, StreamBodyEvent, fmt.Sprintf("%d", l))
```
aryszka, line 416: Could you please log the total number of copied bytes in a single log entry instead?

vetinari: done

aryszka:
:+1:

vetinari:
:+1:

pitr:
👍

szuecs:
:+1:

PR #1599, opened 2020-11-12
Feature/eskip2routegroup command

aryszka:
- [x] implement converter logic
- [ ] test with Kubernetes with and without host validation
- [ ] integrate the temporary command with the existing eskip supplementary tool as a subcommand
- [ ] move the main logic to the dataclients/kubernetes/definitions as library function `FromEskip`
- [ ] documentation and usage


szuecs:
Great tool!
Do you want to put the binary into the docker container?
I am not sure if we should bloat the image, wdyt?

aryszka:
we can add it. The amd64 docker image already has it. I don't think it is _too_ much bloat, because the skipper command 37M, while the eskip is only 7.7M

PR #1600, opened 2020-11-12
Adds Redis ratelimit sampler

AlexanderYastrebov:
Shows Redis ratelimit groups statistics by checking given number of keys using RANDOMKEY.

- [x] implement parallel sampling

### DIFF
--- a/cmd/rsample/main.go
+++ b/cmd/rsample/main.go
@@ -77,7 +77,8 @@ func main() {
 	if err != nil {
 		log.Fatalln(err)
 	}
-	histograms := measure(ctx, ring, *samples, *threads)
+
+	histograms := measure(sampler(ctx, ring), *samples, *threads)
 
 	report(keyspaces, histograms)
 }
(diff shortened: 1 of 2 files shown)


AlexanderYastrebov:
Inspired by https://github.com/antirez/redis-sampler

cmd/rsample/main.go
```diff
@@ -0,0 +1,209 @@
...
+// total                               2468
+//
+// group                            samples    %total       min       max     50.0%     75.0%     95.0%     99.0%     99.9%
+// app-one-two-three                    381      38.1         1         3         2         3         3         3         3
+// app-one-two                          347      34.7         1         3         2         3         3         3         3
+// app-three                            149      14.9         1         1         1         1         1         1         1
+// app-two                              123      12.3         1         1         1         1         1         1         1
+// ```
+//
+// When sampling multiple Redis shards it is important to configure shard list in exactly the same order
```
AlexanderYastrebov, line 18: This is because of the ring, for the sampling the ring is not necessary so the improvement would be to ger rid of the ring and use a list of redis clients

cmd/rsample/main.go
```diff
@@ -0,0 +1,209 @@
...
+				group, ok := group(key)
+				if !ok {
+					// ignore unrecognized keys
+					continue
+				}
+
+				count, err := ring.ZCard(ctx, key).Result()
+				if err != nil {
+					continue
+				}
```
AlexanderYastrebov, line 143: Use `func getSample() (name string, count int64, err error)` parameter. Also makes this function testable

cmd/rsample/main.go
AlexanderYastrebov: Needs verification as ring consistent hash should not depend on the member list order

AlexanderYastrebov:
Closing for now due to lack of demand

Issue #1501, opened 2020-08-10
Make query parameter to be passed to OIDC provider on oauthOidc filters redirect


rraraujo:
**Is your feature request related to a problem? Please describe.**
No. It is just a feature that would help us.

**Describe the solution you would like**
We need a way to pass dynamically a query parameter and value to the OIDC provider when the filters do the redirect to the provider.
Our OIDC provider accepts the query parameter "login_hint". The value passed in the parameter will be already set as the username, avoiding the need for it to be typed again.
For a possible simple solution, I was thinking about having a logic near [here](https://github.com/zalando/skipper/blob/master/filters/auth/oidc.go#L198) that could load every query parameter from the Request into authCodeOptions.
As a more elaborated option, we could have a filter field that would have a list of parameters that, if present, should be sent to the provider.

**Describe alternatives you've considered (optional)**
The user will need to add its username on the OIDC provider page although we already have it.

**Would you like to work on it?**
Yes

szuecs:
The link to the code is for configuring a filter, so here you can’t read the query from a request but do you suggest that you add an authcodeoption that if set would read a query from the request to pass it to the auth server?
If so sounds reasonable to me.
I am not sure what the spec says about the usage of authcode options but sounds reasonable to me.

rraraujo:
Thank you @szuecs. You are right. Maybe it should be inserted [here](https://github.com/zalando/skipper/blob/master/filters/auth/oidc.go#L329) then?

Something like this:

        queryValues := ctx.Request().URL.Query()
        for k := range queryValues {
          f.authCodeOptions = append(f.authCodeOptions, oauth2.SetAuthURLParam(k, queryValues.Get(k)))
        }

szuecs:
@rraraujo sounds good to me.
@universam1 what do you think about it?

universam1:
Sounds like a great idea!
I wonder if this would need to be guarded by a feature flag in case the idP does not deal well with potentially included application queries?

szuecs:
@universam1 as far as I understand the solution will use authCodeOptions to pass it as user input to the filter.
@rraraujo did I understand correctly?

rraraujo:
Thank you @szuecs and @universam1. Yes, using authCodeOptions was one of the options I mentioned. But, @universam1 is right, the code on my last post was not reflecting this option.

I will create a PR using authCodeOptions then. But just to see if you both have any suggestions, let me explain a possible solution with authCodeOptions.

[...]

PR #1602, opened 2020-11-12
Make authCodeOptions dynamic on OIDC auth filters

rraraujo:
On OIDC auth filters, Auth Code Options will be able to be dynamically set by a query parameter.

Enhancement Request Issue #1501

### DIFF
--- a/filters/auth/oidc.go
+++ b/filters/auth/oidc.go
@@ -71,6 +71,7 @@ type (
 		redirectPath    string
 		encrypter       secrets.Encryption
 		authCodeOptions []oauth2.AuthCodeOption
+		queryParams     []string
 		compressor      cookieCompression
 		upstreamHeaders map[string]string
 	}
@@ -205,7 +206,11 @@ func (s *tokenOidcSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 			if len(splitP) != 2 {
 				return nil, filters.ErrInvalidFilterParameters
 			}
-			f.authCodeOptions = append(f.authCodeOptions, oauth2.SetAuthURLParam(splitP[0], splitP[1]))
+			if splitP[1] == "skipper-request-query" {
+				f.queryParams = append(f.queryParams, splitP[0])
+			} else {
+				f.authCodeOptions = append(f.authCodeOptions, oauth2.SetAuthURLParam(splitP[0], splitP[1]))
+			}
 		}
 	}
 	log.Debugf("Auth Code Options: %v", f.authCodeOptions)
@@ -327,7 +332,18 @@ func (f *tokenOidcFilter) doOauthRedirect(ctx filters.FilterContext) {
 		return
 	}
 
-	oauth2URL := f.config.AuthCodeURL(fmt.Sprintf("%x", stateEnc), f.authCodeOptions...)
+	opts := f.authCodeOptions
+	if f.queryParams != nil {
+		opts = make([]oauth2.AuthCodeOption, len(f.authCodeOptions), len(f.authCodeOptions)+len(f.queryParams))
+		copy(opts, f.authCodeOptions)
+		for _, p := range f.queryParams {
+			if v := ctx.Request().URL.Query().Get(p); v != "" {
+				opts = append(opts, oauth2.SetAuthURLParam(p, v))
+			}
+		}
+	}
+
+	oauth2URL := f.config.AuthCodeURL(fmt.Sprintf("%x", stateEnc), opts...)
 	rsp := &http.Response{
 		Header: http.Header{
 			"Location": []string{oauth2URL},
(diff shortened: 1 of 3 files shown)


filters/auth/oidc.go
```diff
@@ -327,6 +328,25 @@ func (f *tokenOidcFilter) doOauthRedirect(ctx filters.FilterContext) {
 		return
 	}
 
+	for i, param := range f.authCodeOptions {
+		r := reflect.ValueOf(param)
```
AlexanderYastrebov, line 332: May I suggest a way out of reflection?

We may have two slices `f.authCodeOptions` and `f.queryParams` and move  `if v == "skipper-request-query"` into `CreateFilter` so that here we create new slice, copy  `f.authCodeOptions`  into it and for each `p` in  `f.queryParams` we append new `oauth2.SetAuthURLParam` to this new slice...

AlexanderYastrebov: Of course `if f.queryParams) == nil` we can skip the copy altogether.

rraraujo: That is a very good suggestion. I changed the code. Is that what you had in mind? Thank you @AlexanderYastrebov !

filters/auth/oidc.go
```diff
@@ -196,6 +197,7 @@ func (s *tokenOidcSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 	}
 
 	f.authCodeOptions = make([]oauth2.AuthCodeOption, 0)
+	f.queryParams = make([]string, 0)
```
AlexanderYastrebov, line 200: There is no need to init it, append supports nil slices

filters/auth/oidc.go
```diff
@@ -327,6 +333,12 @@ func (f *tokenOidcFilter) doOauthRedirect(ctx filters.FilterContext) {
 		return
 	}
 
+	for _, param := range f.queryParams {
+		if queryValue := ctx.Request().URL.Query().Get(param); len(queryValue) > 0 {
+			f.authCodeOptions = append(f.authCodeOptions, oauth2.SetAuthURLParam(param, queryValue))
```
AlexanderYastrebov, line 338: Here filter setting are modified so it will grow on each call, it should be a copy, something like 
```diff
-       oauth2URL := f.config.AuthCodeURL(fmt.Sprintf("%x", stateEnc), f.authCodeOptions...)
+       opts := f.authCodeOptions
+       if f.queryParams != nil {
+               opts = make([]oauth2.AuthCodeOption, 0, len(f.authCodeOptions)+len(f.queryParams))
+               copy(opts, f.authCodeOptions)
+               for _, p := range f.queryParams {
+                       if v := ctx.Request().URL.Query().Get(p); v != "" {
+                               opts = append(opts, oauth2.SetAuthURLParam(p, v))
+                       }
+               }
+       }
+
+       oauth2URL := f.config.AuthCodeURL(fmt.Sprintf("%x", stateEnc), opts...)
```

filters/auth/oidc.go
AlexanderYastrebov: Yes, I have posted some comments below.

filters/auth/oidc.go
rraraujo: Great! That is much better. I pushed the changes. Thank you.

filters/auth/oidc.go
```diff
@@ -327,7 +332,18 @@ func (f *tokenOidcFilter) doOauthRedirect(ctx filters.FilterContext) {
 		return
 	}
 
-	oauth2URL := f.config.AuthCodeURL(fmt.Sprintf("%x", stateEnc), f.authCodeOptions...)
+	opts := f.authCodeOptions
+	if f.queryParams != nil {
+		opts = make([]oauth2.AuthCodeOption, 0, len(f.authCodeOptions)+len(f.queryParams))
+		copy(opts, f.authCodeOptions)
```
AlexanderYastrebov, line 338: The `copy` is tricky, it copies [`min(len(src), len(dst))`](https://golang.org/pkg/builtin/#copy) so we need to init `opts` to the length of src which is `f.authCodeOptions`
```go
opts = make([]oauth2.AuthCodeOption, len(f.authCodeOptions), len(f.authCodeOptions)+len(f.queryParams))
```

See https://play.golang.org/p/-c9w6AxpN0-

rraraujo: Yeah, that is a good solution. Thank you. Is there anything missing now?

szuecs: code lgtm, a test would be nice :)

AlexanderYastrebov: @rraraujo Just to be clear 
```go
opts = make([]oauth2.AuthCodeOption, 0, len(f.authCodeOptions)+len(f.queryParams))
```
should be changed to
```go
opts = make([]oauth2.AuthCodeOption, len(f.authCodeOptions), len(f.authCodeOptions)+len(f.queryParams))
```

szuecs:
:+1:

citosid:
Hello! We have an issue that would be fixed by this PR. Do you know when this could be released?

szuecs:
@citosid as soon as a second review accept this, too. After merge everything is automated.

aryszka:
:+1:

PR #1609, opened 2020-11-16
Fully functional OAuth2 grant flow.

oasalonen:
Fixes and extends the original grant PR for a fully functional OAuth2
authorization code grant flow.

Add support for loading and automatic reloading of client credentials
from the filesystem. This is accomplished by using the preexisting
`SecretsProvider`.

Make the feature more configurable. For example:
* provide arbitrary key-value pairs which should be set for all calls
to authorize and token endpoints. Accomplished with the help of a new
`mapFlags` option type.
* name of the header where to place the user's access token.
* name of the cookie where to store encrypted tokens.

Create a new `grantClaimsQuery()` filter. This is an alias for
`oidcClaimsQuery()` and enables the combination of access control rules
with the `oauthGrant()` filter.

Increase OAuth grant flow state validity time from 1 minute -> 1 hour.
One minute is a short time. The user's login is disrupted and must be
restarted if it takes them longer than a minute to actually log in.
By increasing this to 1 hour the user experience improves without
undermining security much.

Fix invalid redirect URI given to the token callback.

Fix broken token refresh. When token is valid, but token refreshing
fails, the code crashes with a segfault. The fix involves not setting
the state bag oauth2 token to a `nil` value.

Fix cookie expiration. Expiration relies on access token expiry,
meaning cookie is deleted by the browser before refresh token expires.
This is fixed by adding an extra 30 days to the cookie expiration time.

Reuse the `TokeninfoClient` for fetching token claims and remove the
duplicated code.

Refactor token validity checking and token refreshing code for improved
readability.

Add the `tokeninfo` payload to the state bag under the `tokeninfoCacheKey`
so the `oauthGrant()` filter can be used with the `forwardToken()`
filter.

Add support for handling multiple cookies with the same name. It's
possible that the browser calls `oauthGrant()` with multiple cookies
under the same name as the grant token cookie. The code handles it by
attempting to decode each cookie with the correct name, until it finds
one that decodes properly.

Make sure the grant token cookie is dropped from all requests that it
forwards downstream to prevent the cookie from leaking.

The oauth grant feature depends on a `randString()` function that was
removed in `oidc.go` in master. The function is revived in 
`grantflowstate.go`.

Add tutorial documentation for using the OAuth2 auth code grant flow.

Add reference documentation for all grant-related filters.

Add unit tests for new code and unit test the existing token
refresh functionality.


delivery.yaml
```diff
@@ -6,6 +6,9 @@ pipeline:
   commands:
   - desc: build-push
     cmd: |
+      echo "CDP_BUILD_VERSION: $CDP_BUILD_VERSION"
+      echo "CDP PR Number: $CDP_PULL_REQUEST_NUMBER"
+
```
szuecs, line 11: this should be cleaned up

docs/reference/filters.md
```diff
@@ -1084,6 +1084,98 @@ forwardToken("X-Tokeninfo-Forward")
 forwardToken("X-Tokeninfo-Forward", "access_token")
 ```
 
+## oauthGrant
+
+Enables authentication and authorization with an OAuth2 authorization code grant flow.
```
szuecs, line 1089: should we add a reference to the RFC https://tools.ietf.org/html/rfc6749#section-1.3.1

filters/auth/grant.go
```diff
@@ -0,0 +1,239 @@
...
+
+	log "github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/filters"
+	"golang.org/x/oauth2"
+)
+
+const (
+	OAuthGrantName = "oauthGrant"
+
+	bearerPrefix              = "Bearer "
```
szuecs, line 17: this is a duplicate of `authHeaderPrefix`

filters/auth/grant.go
```diff
@@ -0,0 +1,239 @@
...
+	req := ctx.Request()
+	redirect, original := config.RedirectURLs(req)
+
+	if originalOverride != "" {
+		original = originalOverride
+	}
+
+	state, err := config.flowState.createState(original)
+	if err != nil {
+		log.Errorf("failed to create login redirect: %v", err)
```
szuecs, line 70: nitpick: logs start with an upper case char

filters/auth/grant.go
```diff
@@ -0,0 +1,239 @@
...
+			"Location": []string{authConfig.AuthCodeURL(state, config.GetAuthURLParameters(redirect)...)},
+		},
+	})
+}
+
+func (f grantFilter) refreshToken(c cookie) (*oauth2.Token, error) {
+	token := &oauth2.Token{
+		AccessToken:  c.AccessToken,
+		RefreshToken: c.RefreshToken,
+		Expiry:       time.Now().Add(-time.Minute),
```
szuecs, line 88: maybe add a comment: `enforce refresh`

filters/auth/grant.go
```diff
@@ -0,0 +1,239 @@
...
+
+	token, err := f.refreshTokenIfRequired(*c)
+	if err != nil && c.isAccessTokenExpired() {
+		// Refresh failed and we no longer have a valid access token.
+		loginRedirect(ctx, f.config)
+		return
+	}
+
+	if token == nil {
+		// Token can be null if:
```
szuecs, line 183: nitpick: null -> nil

filters/auth/grantconfig.go
```diff
@@ -0,0 +1,308 @@
...
+
+	// Secrets is a secret registry to access secret keys used for encrypting
+	// auth flow state and auth cookies.
+	Secrets *secrets.Registry
+
+	// SecretFile contains the filename with the encryption key for the authentication
+	// cookie and grant flow state stored in Secrets.
+	SecretFile string
+
+	// AuthURL, the url to redirect the requests to when login is require.
```
szuecs, line 34: require -> required

filters/auth/grantcookie.go
```diff
@@ -0,0 +1,126 @@
...
+	if d < time.Minute {
+		d = time.Minute
+	}
+
+	return now.Add(d)
+}
+
+func decodeCookie(cookieHeader string, config OAuthConfig) (c *cookie, err error) {
+	var eb []byte
+	if eb, err = base64.StdEncoding.DecodeString(cookieHeader); err != nil {
```
szuecs, line 37: Note: StdEncoding seems to be fine because of https://stackoverflow.com/questions/1969232/what-are-allowed-characters-in-cookies/1969339#1969339

filters/auth/grantcookie.go
```diff
@@ -0,0 +1,126 @@
...
+func (c cookie) isAccessTokenExpired() bool {
+	now := time.Now()
+	return now.After(c.Expiry)
+}
+
+// GetCookie extracts the OAuth Grant token cookie from a HTTP request.
+// The function supports multiple cookies with the same name and returns
+// the best match (the one that decodes properly).
+// The client may send multiple cookies if a parent domain has set a
+// cookie of the same name.
```
szuecs, line 64: Note: seems to be fine, because cookie size is limited. If not it would be cheaper to attack the cookie for attackers

filters/auth/grantflowstate.go
```diff
@@ -0,0 +1,132 @@
+package auth
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"math/rand"
```
szuecs, line 7: this should be crypto/rand, not?

filters/auth/grantflowstate.go
```diff
@@ -0,0 +1,132 @@
...
+	"time"
+
+	"github.com/zalando/skipper/secrets"
+)
+
+const randomStringLength = 20
```
szuecs, line 13: ignore: the const name is funny, because the string length is quite static :)

go.sum
```diff
@@ -307,6 +307,9 @@ github.com/yuin/gopher-lua v0.0.0-20200603152657-dc2b0ca8b37e/go.mod h1:gqRgreBU
 go.opencensus.io v0.21.0/go.mod h1:mSImk1erAIZhrmZN+AvHh14ztQfjbGwt4TtuofqLduU=
 go.opentelemetry.io/otel v0.13.0 h1:2isEnyzjjJZq6r2EKMsFj4TxiQiexsM04AVhwbR/oBA=
 go.opentelemetry.io/otel v0.13.0/go.mod h1:dlSNewoRYikTkotEnxdmuBHgzT+k/idJSfDv/FxEnOY=
+github.com/yuin/goldmark v1.1.27/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
+github.com/yuin/gopher-lua v0.0.0-20190514113301-1cd887cd7036 h1:1b6PAtenNyhsmo/NKXVe34h7JEZKva1YB/ne7K7mqKM=
+github.com/yuin/gopher-lua v0.0.0-20190514113301-1cd887cd7036/go.mod h1:gqRgreBUhTSL0GeU64rtZ3Uq3wtjOa/TB2YfrtkCbVQ=
```
szuecs, line 312: these changes above look weird to me.

filters/auth/grant_test.go
```diff
@@ -0,0 +1,466 @@
...
+	testToken                = "foobarbaz"
+	testRefreshToken         = "refreshfoobarbaz"
+	testAccessCode           = "quxquuxquz"
+	testSecretFile           = "testdata/authsecret"
+	testAccessTokenExpiresIn = int(time.Hour / time.Second)
+	testCookieName           = "testcookie"
+)
+
+func newGrantTestTokeninfo(validToken string, tokenInfoJSON string) *httptest.Server {
+	const prefix = "Bearer "
```
szuecs, line 32: we can reuse `authHeaderPrefix`

szuecs:
Thanks, that you picked this up!
It's great to read the code. There are only some small things to clean up and some nitpicks to fix, because you already have to cleanup some things. :)
Please also rebase to master (resolve conflicts)

filters/auth/grantcookie.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	return nil, http.ErrNoCookie
+}
+
+// Drops the grant token cookie from the request
+func dropCookie(request *http.Request, config OAuthConfig) {
+	cookies := request.Header.Get("Cookie")
+	cookies = config.TokenCookieRegexp.ReplaceAllString(cookies, "")
+	request.Header.Set("Cookie", cookies)
+}
```
AlexanderYastrebov, line 82: Here I feel a bit of mismatch. getCookie tries best effort to find one and dropCookie removes all with the same name.
Should it be rather something like
```go
// extractCookie removes and returns the OAuth Grant token cookie from a HTTP request.
// The function supports multiple cookies with the same name and returns
// the best match (the one that decodes properly).
// The client may send multiple cookies if a parent domain has set a
// cookie of the same name.
func extractCookie(request *http.Request, config OAuthConfig) (cookie *cookie, err error) {
	old := request.Cookies()
	new := make([]*http.Cookie, 0, len(old))

	for i, c := range old {
		if c.Name == config.TokenCookieName {
			cookie, _ = decodeCookie(c.Value, config)
			if cookie != nil {
				new = append(new, old[i+1:]...)
				break
			}
		}
		new = append(new, c)
	}

	if cookie != nil {
		request.Header.Del("Cookie")
		for _, c := range new {
			request.AddCookie(c)
		}
		return cookie, nil
	}
	return nil, http.ErrNoCookie
}
```

filters/auth/grant.go
```diff
@@ -0,0 +1,239 @@
...
+		// We need to fetch tokeninfo again if we refreshed the token.
+		tokeninfo, err = f.config.TokeninfoClient.getTokeninfo(token.AccessToken, ctx)
+		if err != nil {
+			if err == errInvalidToken {
+				log.Errorf("Error while calling tokeninfo: %v.", err)
+			}
+			loginRedirect(ctx, f.config)
+			return
+		}
+	}
```
AlexanderYastrebov, line 206: This codepath feels complicated. If `checkAccessTokenValidity` is inlined it becomes apparent that there is `f.config.TokeninfoClient.getTokeninfo` twice in the code path. Maybe if `refreshTokenIfRequired` is changed to make token from cookie when refresh is not needed like:
```go
func (f grantFilter) refreshTokenIfRequired(c cookie) (*oauth2.Token, error) {
	now := time.Now()
	isAccessTokenExpired := c.isAccessTokenExpired()
	canRefresh := !f.config.DisableRefresh && c.RefreshToken != ""
	shouldRefresh := isAccessTokenExpired || now.After(c.RefreshAfter)

	if canRefresh && shouldRefresh {
		log.Debugf(
			"Refreshing token. Can refresh: %v, Should refresh: %v, Expired: %v",
			canRefresh, shouldRefresh, isAccessTokenExpired,
		)

		return f.refreshToken(c)
	}
	return &oauth2.Token{
		AccessToken:  c.AccessToken,
		TokenType:    "Bearer",
		RefreshToken: c.RefreshToken,
		Expiry:       c.Expiry,
	}, nil
}
```
this codepath could be simplified to
```go
        token, err := f.refreshTokenIfRequired(*c)
	if err != nil {
		// Refresh failed and we no longer have a valid access token.
		loginRedirect(ctx, f.config)
		return
	}

	tokeninfo, err := f.config.TokeninfoClient.getTokeninfo(token.AccessToken, ctx)
	if err != nil {
		if err == errInvalidToken {
			log.Errorf("Error while calling tokeninfo: %v.", err)
		}
		loginRedirect(ctx, f.config)
		return
	}
```
WDYT?

filters/auth/grantflowstate.go
```diff
@@ -0,0 +1,132 @@
...
+	letterIdxBits = 6                    // 6 bits to represent a letter index
+	letterIdxMask = 1<<letterIdxBits - 1 // All 1-bits, as many as letterIdxBits
+	letterIdxMax  = 63 / letterIdxBits   // # of letter indices fitting in 63 bits
+)
+
+var (
+	src = rand.NewSource(time.Now().UnixNano())
+)
+
+// https://stackoverflow.com/questions/22892120/how-to-generate-a-random-string-of-a-fixed-length-in-golang
```
AlexanderYastrebov, line 41: This is a famous answer but I would go with
> Paul's solution provides a simple, general solution.

IMO the performance difference does not make sense in the context of back-and-forth redirects while readability suffers

filters/auth/grantflowstate.go
```diff
@@ -0,0 +1,132 @@
...
+		return "", err
+	}
+
+	nonce, err := encrypter.CreateNonce()
+	if err != nil {
+		return "", err
+	}
+
+	state := state{
+		Rand:       randString(randomStringLength),
```
AlexanderYastrebov, line 83: What is this Rand for? The state is encrypted and has a random Nonce every time it is created...

docs/reference/filters.md
```diff
@@ -1084,6 +1084,98 @@ forwardToken("X-Tokeninfo-Forward")
...
+    -> grantCallback()
+    -> <shunt>;
+
+all:
+    *
+    -> oauthGrant()
+    -> "http://localhost:9090";
+```
+
+Program arguments:
```
AlexanderYastrebov, line 1120: Skipper arguments?

delivery.yaml
oasalonen: Done

docs/reference/filters.md
oasalonen: Done

filters/auth/grant.go
oasalonen: Removed. It was unused.

filters/auth/grant.go
oasalonen: Done

filters/auth/grant.go
oasalonen: Added a comment to clarify.

filters/auth/grant.go
oasalonen: Done

filters/auth/grantconfig.go
oasalonen: Done

filters/auth/grantflowstate.go
oasalonen: Removed as a result of removing the `randString()` function

filters/auth/grantflowstate.go
oasalonen: Removed. Unused after removing the `randString()` function

go.sum
oasalonen: Removed.

filters/auth/grant_test.go
oasalonen: I left this as is. The reason is that `authHeaderPrefix` lives in a different package (`auth`) than the test (`auth_test`). Or should these be in the same package?

Also, it might be better to validate against the expected string `Bearer ` here in order to break the test if someone accidentally modifies the constant to an invalid value.

filters/auth/grantcookie.go
oasalonen: Agreed. Your approach is better. Fixed.

filters/auth/grant.go
oasalonen: Yes makes sense. I modified it to what you suggested with two differences:

1. Right after refreshing token, I keep the second condition `c.isAccessTokenExpired()` here:
```go
	token, err := f.refreshTokenIfRequired(*c)
	if err != nil && c.isAccessTokenExpired() {
```
The reason is that if the access token is not yet expired, but the refresh call failed (because the refresh is done well before the expiry), then we don't want to force the user to relogin at this point since they still have a valid access token (or if it is revoked, that gets caught by the later tokeninfo call).

2. Right after doing the tokeninfo check, I added a condition to check that the returned `tokeninfo` is not nil
```go
	tokeninfo, err := f.config.TokeninfoClient.getTokeninfo(token.AccessToken, ctx)
	if err != nil || tokeninfo == nil {
```
To be honest, I can't think of a scenario where it would happen, but if it does, it will break downstream filters unexpectedly.

filters/auth/grantflowstate.go
oasalonen: This code was removed as a result of removing the `randString()` function.

filters/auth/grantflowstate.go
oasalonen: This code was removed. As we discussed, it's not necessary.

docs/reference/filters.md
oasalonen: Fixed.

filters/auth/grant.go
AlexanderYastrebov: Nice!

Regarding 1. `if err != nil && c.isAccessTokenExpired() ` I can not easily get it. 
> The reason is that if the access token is not yet expired, but the refresh call failed 

Ok, that means we are refreshing because of `now.After(c.RefreshAfter)`. Is that `c.RefreshAfter` not reliable?

Regarding 2. `if err != nil || tokeninfo == nil` I would leave just error check and rather fix tokeninfoclient to provide strict contract "either error or valid result" if we ever observe nil result (I don't see how that is possible).

IMO we should strive for strict method contracts in general such that it is either error or valid return value to have simple error checking `if err != nil`; otherwise it is hard to reason and reader stumbles and has to parse error checking conditions.

AlexanderYastrebov:
It seems some rebase is necessary as I can see in the diff the changes that are in master already

filters/auth/grant_test.go
szuecs: You are right, I missed the package _test.

filters/auth/grantcookie.go
```diff
@@ -0,0 +1,134 @@
...
+
+func refreshAfter(expiry time.Time) time.Time {
+	now := time.Now()
+	d := expiry.Sub(now)
+	if d <= 0 {
+		return now
+	}
+
+	d /= 10
```
AlexanderYastrebov, line 27: Why is it divided by 10 here?

filters/auth/grantcookie.go
```diff
@@ -0,0 +1,134 @@
...
+type cookie struct {
+	AccessToken  string    `json:"access_token"`
+	RefreshToken string    `json:"refresh_token"`
+	Expiry       time.Time `json:"expiry,omitempty"`
+	RefreshAfter time.Time `json:"refresh_after,omitempty"`
+}
+
+func refreshAfter(expiry time.Time) time.Time {
+	now := time.Now()
+	d := expiry.Sub(now)
```
AlexanderYastrebov, line 22: https://golang.org/pkg/time/#Time.Before

config/mapflags.go
```diff
@@ -0,0 +1,59 @@
...
+	m.values = make(map[string]string)
+
+	vs := strings.Split(value, ",")
+	for _, vi := range vs {
+		kv := strings.Split(vi, "=")
+
+		k := strings.TrimSpace(kv[0])
+		v := strings.TrimSpace(kv[1])
+
+		if len(kv) != 2 || k == "" || v == "" {
```
AlexanderYastrebov, line 39: The length is checked after slice is addressed, also https://golang.org/pkg/strings/#SplitN might be useful

config/mapflags.go
```diff
@@ -0,0 +1,59 @@
...
+			message := fmt.Sprint("invalid map key-value pair, expected format key=value but got: '", vi, "'")
+			return errors.New(message)
```
AlexanderYastrebov, line 41: https://golang.org/pkg/fmt/#Errorf

filters/auth/grantcookie.go
oasalonen: @aryszka could you give some background on this one?

filters/auth/grantcallback.go
```diff
@@ -0,0 +1,92 @@
...
+
+	queryState := q.Get("state")
+	if queryState == "" {
+		badRequest(ctx)
+		return
+	}
+
+	state, err := f.config.flowState.extractState(queryState)
+	if err != nil {
+		log.Errorf("Error when extracting flow state: %v.", err)
```
AlexanderYastrebov, line 51: `state` is under external control so maybe we should not spam log here

filters/auth/grantconfig.go
```diff
@@ -0,0 +1,296 @@
...
+	if secret, exists := c.SecretsProvider.GetSecret(c.ClientSecretFile); exists {
+		return string(secret)
+	}
+
+	return ""
+}
+
+// RedirectURLs constructs the redirect URI based on the request and the
+// configured CallbackPath.
+func (c OAuthConfig) RedirectURLs(req *http.Request) (redirect, original string) {
```
AlexanderYastrebov, line 273: Should be pointer `(c *OAuthConfig)`

filters/auth/grantconfig.go
```diff
@@ -0,0 +1,296 @@
...
+	ErrMissingClientSecret    = errors.New("missing client secret")
+	ErrMissingSecretsRegistry = errors.New("missing secrets registry")
+	ErrMissingSecretFile      = errors.New("missing secret file")
+	ErrMissingTokeninfoURL    = errors.New("missing tokeninfo URL")
+	ErrMissingProviderURLs    = errors.New("missing provider URLs")
+)
+
+func (c *OAuthConfig) init() error {
+	if c.initialized {
+		return c.initErr
```
AlexanderYastrebov, line 115: `return nil` as initialized = true only if there was not error. Or alternatively on every early error return initialized should be set to true as well.

Overall this feels a bit defensive, I think alternative could be to have public `Init() error` that user of  `OAuthConfig` should call once. Then there is no need to call `err := c.init()` in every `NewGrant*` constructor and 
```go
func initGrant(c *auth.OAuthConfig, o *Options) error {
	grant, err := c.NewGrant()
	if err != nil {
		return err
	}

	grantCallback, err := c.NewGrantCallback()
	if err != nil {
		return err
	}

	grantClaimsQuery, err := c.NewGrantClaimsQuery()
	if err != nil {
		return err
	}

	o.CustomFilters = append(o.CustomFilters, grant, grantCallback, grantClaimsQuery)
	return nil
}
```
could be reduced to
```go
		if err := oauthConfig.Init(); err != nil {
			log.Errorf("Error while initializing oauth grant filters: %v.", err)
		} else {
                        o.CustomFilters = append(o.CustomFilters, oauthConfig.NewGrant(), oauthConfig.NewGrantCallback(), grantClaimsQuery.NewGrantClaimsQuery())
                }
```

filters/auth/grantconfig.go
AlexanderYastrebov: I think makes sense to check elsewhere (e.g. `cookie`) that pointer to struct is used, see https://dave.cheney.net/2016/03/19/should-methods-be-declared-on-t-or-t

> In short, I think that you should prefer declaring methods on *T unless you have a strong reason to do otherwise.

skipper.go
```diff
@@ -1198,9 +1373,20 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
+		ro.PreProcessors = append(ro.PreProcessors, o.DefaultFilters)
+	}
+
+	if o.EnableOAuth2GrantFlow /* explicitly enable grant flow when callback route was not disabled */ {
+		grantPrep, err := oauthConfig.NewGrantPreprocessor()
+		if err != nil {
+			log.Errorf("Error while initializing oauth grant preprocessor: %v.", err)
+		}
+
+		ro.PreProcessors = append(ro.PreProcessors, grantPrep)
```
AlexanderYastrebov, line 1387: With `ro.PreProcessors = append(ro.PreProcessors, o.DefaultFilters)` change there is not need to have second `if o.EnableOAuth2GrantFlow` - grant preprocessor could be appended in the filter construction block to keep all grant init in one place.

oasalonen: Is it safe to move this to the filter construction block? It would then result in adding the grant preprocessor to the `ro.PreProcessors` _before_ `o.DefaultFilters`, which currently is right above:
```go
	if o.DefaultFilters != nil {
		ro.PreProcessors = append(ro.PreProcessors, o.DefaultFilters)
	}
```
Wouldn't it affect the order in which the default filters are applied?

filters/auth/grant.go
oasalonen: > Ok, that means we are refreshing because of now.After(c.RefreshAfter). Is that c.RefreshAfter not reliable?

As you saw the `RefreshAfter` is the expiry duration / 10. So if the token is set to expire in 1h, it will actually try the first refresh after 6 mins. I don't know exactly why `/ 10` was chosen, but if the refresh call fails then a user's session will be invalidated a lot before the actual token expiry.

> Regarding 2. if err != nil || tokeninfo == nil I would leave just error check

I reviewed the tokeninfoclient code and there is no way tokeninfo can be null. Also, an error is returned in all cases. So I agree the `tokeninfo == nil` is not needed. I will remove it.

oasalonen: 2. Fixed

filters/auth/grantcookie.go
oasalonen: Fixed

config/mapflags.go
oasalonen: Fixed with additional unit tests as well.

config/mapflags.go
oasalonen: Fixed

filters/auth/grantcallback.go
oasalonen: Fixed

filters/auth/grantconfig.go
oasalonen: Fixed

filters/auth/grantconfig.go
oasalonen: Fixed. I simplified it as you suggested.

skipper.go
```diff
@@ -955,6 +1004,16 @@ func listenAndServe(proxy http.Handler, o *Options) error {
 	return listenAndServeQuit(proxy, o, nil, nil, nil)
 }
 
+func initGrant(c *auth.OAuthConfig, o *Options) error {
+	if err := c.Init(); err != nil {
+		log.Errorf("Error while initializing oauth grant filters: %v.", err)
```
szuecs, line 1009: Error while initializing .. -> Failed to initialize ..

skipper.go
```diff
@@ -1200,6 +1259,32 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
+		oauthConfig.Secrets = secrets.NewRegistry()
+		oauthConfig.AccessTokenHeaderName = o.OAuth2AccessTokenHeaderName
+		oauthConfig.TokeninfoSubjectKey = o.OAuth2TokeninfoSubjectKey
+		oauthConfig.TokenCookieName = o.OAuth2TokenCookieName
+		oauthConfig.ConnectionTimeout = o.OAuthTokeninfoTimeout
+		oauthConfig.MaxIdleConnectionsPerHost = o.IdleConnectionsPerHost
+		oauthConfig.Tracer = tracer
+
+		if err := initGrant(oauthConfig, &o); err != nil {
+			log.Errorf("Error while initializing oauth grant filter: %v.", err)
```
szuecs, line 1284: Error while initializing .. -> Failed to initialize ..

oasalonen: Good point. Fixed.

skipper.go
oasalonen: Fixed.

skipper.go
AlexanderYastrebov: As far as I can tell default filters add filters to routes and grant preprocessor adds a route. So if grant preprocessor is in the list _before_ default filters preprocessor they would be added to grant route as well which I guess is desirable. 
https://github.com/zalando/skipper/blob/11e4b519fd7b85b2bf0ea39c1bcac504a5544303/routing/datasource.go#L535-L537

filters/auth/grantconfig.go
AlexanderYastrebov: Basically everywhere it is not a pointer (e.g. filter structs) they will be fully copied on each call as
> Methods in Go are just syntactic sugar for a function which passes the receiver as the first formal parameter.

skipper.go
szuecs: Do we create an automated call back route for the redirect?
I am not sure if implicit routes like this were a good choice.

oasalonen: > Do we create an automated call back route for the redirect?

Yes that's what it looks like, and I completely missed that in the original PR, assuming that the callback should be explicit. Either the implicit route should be removed, or the documentation needs to be updated to reflect the fact that no explicit callback route is necessary. I'll defer to your judgement here on the best approach.

filters/auth/grantconfig.go
oasalonen: 👍  ok I updated `grant`, `grantCallback`, `grantPrep`, `cookie`, and their specs to use receiver pointers in functions instead.

config/mapflags.go
```diff
@@ -0,0 +1,69 @@
...
+
+const formatErrorString = "invalid map key-value pair, expected format key=value but got: '%v'"
+
+func newMapFlags() *mapFlags {
+	return &mapFlags{
+		values: make(map[string]string),
+	}
+}
+
+func (m mapFlags) String() string {
```
AlexanderYastrebov, line 22: should be pointer receiver as elsewhere

oasalonen: Fixed

filters/auth/grantcookie.go
aryszka: if refresh only after expiry, then there is a high chance that we miss the time window with the refresh when the token is still valid. Instead, we refresh on user activity. But to avoid refreshing on every request, we only refresh when the token is expire-expiry/10 old. This results 10% precision, which is tolerable from a human workflow's perspective. Just like it is tolerable to maximize the refresh rate in 1 minute.

Example, refreshed:

1. I login and get a token at 9:15 that expires at 9:45 (30 minutes expiration time)
2. I make some authenticated requests to a page between 9:15 and 9:18, no refresh
3. I make a request at 9:19, my token is still valid, and I get a refreshed token, and I will not need to login until 9:49, instead of 9:45
4. I can continue my workflow with step 2

Example, expired:

1. I login and get a token at 9:15 that expires at 9:45 (30 minutes expiration time)
2. I make some requests to a page between 9:15 and 9:17, but then I finish
3. I go back to the page at 9:46. My token is expired, and I have to login again
4. I understand this, because I didn't make any requests for 29 minutes, which I perceive as approximately the expiration time of 30 minutes
5. I can continue my workflow, but only from step 1

Does this make sense?

I think that deserves a code comment, though, I agree with that.

skipper.go
aryszka: I am checking this now, I just need to recollect the ideas behind my original approach :smile: 

The concern with the order of the DefaultFilters and the grant preprocessor seems to be valid.

@szuecs we should not have the grant filter on http routes, only on https routes. So we should generate the redirect route for the http routes, either. Do I misunderstand your concern?

aryszka: > So if grant preprocessor is in the list before default filters preprocessor they would be added to grant route as well which I guess is desirable.

I think we can argue both ways. The callback route is an implicitly generated route, not defined by the users/operator, so it may make sense that the default filters are not there. Imagine, when somebody uses a tokeninfo filter as a default filter, because in their setup it makes sense. But I see how we can argue the other way around just as well, e.g. because default filters should be just that: "default".

My suggestion is that to start, let's _not_ add the default filters to the callback route, and see in practice how far that brings us. If it turns out to be wrong, it's easier to fix that in a later pass.

WDYT?

filters/auth/grantcookie.go
```diff
@@ -0,0 +1,134 @@
...
+		request.Header.Del("Cookie")
+		for _, c := range new {
+			request.AddCookie(c)
+		}
+		return cookie, nil
+	}
+	return nil, http.ErrNoCookie
+}
+
+func CreateCookie(config OAuthConfig, host string, t *oauth2.Token) (*http.Cookie, error) {
```
aryszka, line 92: we don't need to export this function. If it is currently exported only because the grant_test.go uses `grant_test` as the package name, then we can just create another additional test file, using `grant` as the package name for those tests that directly call the `createCookie()` function.

filters/auth/grantcookie.go
aryszka: I checked: https://tools.ietf.org/html/rfc6749#section-1.5

I see that this approach is not required, and the refresh token can be used regardless of the validity of the access token. This means that we can just use the expiration of the token as is, and there is no need for the above flow.

filters/auth/grant.go
```diff
@@ -0,0 +1,197 @@
...
+
+	// oauth2.TokenSource implements the refresh functionality,
+	// we're hijacking it here.
+	tokenSource := f.config.GetConfig().TokenSource(ctx, token)
+	return tokenSource.Token()
+}
+
+func (f *grantFilter) refreshTokenIfRequired(c cookie) (*oauth2.Token, error) {
+	now := time.Now()
+	isAccessTokenExpired := c.isAccessTokenExpired()
```
aryszka, line 104: we could use the same single `now` when checking the expiration. Probably, we can just drop the `isAccessTokenExpired` helper function as it is simple enough.

aryszka: fine actually, seen it f.Request(). Fine actually.

skipper.go
szuecs: Discussed in chat:
   > One problem is that it limits the number of providers for one skipper deployment, because callback route is a single route with no    dispatch to another provider. We have also hard coded configuration right now to have only one provider possible anyways, so maybe this is fine right now. I think in the future we should think about how we can create less friction (check oidc filters are very different and build to use multiple providers), but have increased usability (this one is much better than oidc filters).

I would add default filters to the route as is, because you likely want to have lifo() for example set there, too. I don't think tokeninfo filters will be used in default filters.

aryszka:
LGTM

- not exporting the CreateCookie function can be fixed separately in my opinion.
- I am open for both approaches on the usage of the DefaultFilters

Thanks a lot @oasalonen @AlexanderYastrebov @szuecs !

docs/reference/filters.md
```diff
@@ -1084,6 +1084,99 @@ forwardToken("X-Tokeninfo-Forward")
...
+
+```
+grantCallback()
+```
+
+Skipper arguments:
+
+| Argument | Required? | Description |
+| -------- | --------- | ----------- |
+| `-oauth2-callback-path` | **yes** | path of the Skipper route containing the `grantCallback()` filter. Example: `-oauth2-callback-path=/oauth/callback` |
```
szuecs, line 1157: this is optional due to the PreProcessor route

docs/reference/filters.md
```diff
@@ -1084,6 +1084,99 @@ forwardToken("X-Tokeninfo-Forward")
...
+
+Skipper arguments:
+
+| Argument | Required? | Description |
+| -------- | --------- | ----------- |
+| `-enable-oauth2-grant-flow` | **yes** | toggle flag to enable the `oauthGrant()` filter. Must be set if you use the filter in routes. Example: `-enable-oauth2-grant-flow` |
+| `-oauth2-auth-url` | **yes** | URL of the OAuth2 provider's authorize endpoint. Example: `-oauth2-auth-url=https://identity.example.com/oauth2/authorize` |
+| `-oauth2-token-url` | **yes** | URL of the OAuth2 provider's token endpoint. Example: `-oauth2-token-url=https://identity.example.com/oauth2/token` |
+| `-oauth2-tokeninfo-url` | **yes** | URL of the OAuth2 provider's tokeninfo endpoint. Example: `-oauth2-tokeninfo-url=https://identity.example.com/oauth2/tokeninfo` |
+| `-oauth2-callback-path` | **yes** | path of the Skipper route containing the `grantCallback()` filter for accepting an authorization code and using it to get an access token. Example: `-oauth2-callback-path=/oauth/callback` |
```
aryszka, line 1129: docfix: i think the -oauth2-callback-path is optional. We use `/.well-known/oauth2-callback` by default.

szuecs, line 1129: this is optional due to the PreProcessor route

docs/reference/filters.md
```diff
@@ -1084,6 +1084,99 @@ forwardToken("X-Tokeninfo-Forward")
...
+
+```
+callback:
+    Path("/oauth/callback")
+    -> grantCallback()
+    -> <shunt>;
+
+all:
+    *
+    -> oauthGrant()
```
szuecs, line 1117: this route is added by the PreProcessor

docs/tutorials/auth.md
```diff
@@ -245,3 +245,155 @@ Access to path `/` would be granted to everyone in `example.org`, however path `
...
+It works as follows:
+
+1. A user makes a request to a route with `oauthGrant()`.
+1. The filter checks whether the request has a cookie called `oauth-grant`. If it does not, or
+   if the cookie and its tokens are invalid, it redirects the user to the OAuth2 provider's 
+   authorization endpoint.
+1. The user logs into the external OAuth2 provider, e.g. by providing a username and password.
+1. The provider redirects the user back to Skipper with an authorization code, using the 
+   callback URL parameter which was part of the previous redirect. The callback route must
+   have a `grantCallback()` filter defined.
```
aryszka, line 263: It might be useful to point out that this route is autogenerated.

docs/reference/filters.md
```diff
@@ -1084,6 +1084,99 @@ forwardToken("X-Tokeninfo-Forward")
...
+## grantCallback
+
+The filter accepts authorization codes as a result of an OAuth2 authorization code grant
+flow triggered by [oauthGrant](#oauthGrant). It uses the code to request access and
+refresh tokens from the OAuth2 provider's token endpoint.
+
+Examples:
+
+```
+grantCallback()
```
szuecs, line 1150: this is normally added in PreProcessor directive, which should be documented

docs/tutorials/auth.md
szuecs, line 263: Here we should also hint for the auto configured route

docs/tutorials/auth.md
```diff
@@ -245,3 +245,155 @@ Access to path `/` would be granted to everyone in `example.org`, however path `
...
+
+The `oauthGrant()` filter also supports token refreshing. Once the access token expires and
+the user makes another request, the filter automatically refreshes the token and sets the
+updated cookie in the response.
+
+### Instructions
+
+To use authorization grant flow, you need to:
+
+1. Configure OAuth2 credentials.
```
szuecs, line 286: Users should know the valid callback URL we automatically create, before they create credentials, because providers will ask for the callback URL, and if they do not know before, then they have to delete and recreate the credential.

docs/tutorials/auth.md
```diff
@@ -245,3 +245,155 @@ Access to path `/` would be granted to everyone in `example.org`, however path `
...
+The `-oauth2-callback-path` must match the path of the route configured with the
+`grantCallback()` filter (refer to the next section).
+
+You can configure the `oauthGrant()` filter further for your needs. See the
+[oauthGrant](../reference/filters.md#oauthGrant) filter reference for more details.
+
+#### Add filters to your routes
+
+You can protect any number of routes with the `oauthGrant()` filter. Unauthenticated users
+will be refused access and redirected to log in. You also need one `grantCallback()` filter
```
szuecs, line 344: hint missing that route with `grantCallback` is auto generated

docs/tutorials/auth.md
```diff
@@ -245,3 +245,155 @@ Access to path `/` would be granted to everyone in `example.org`, however path `
...
+    -> oauthGrant()
+    -> "http://localhost:9090";
+
+bar: 
+    Path("/bar")
+    -> oauthGrant()
+    -> "http://localhost:9090/";
+
+callback:
+    Path("/oauth/callback")
```
szuecs, line 359: Path will be `/.well-known/oauth2-callback` as far as I remember

filters/auth/grant.go
```diff
@@ -0,0 +1,197 @@
...
+	canRefresh := !f.config.DisableRefresh && c.RefreshToken != ""
+	shouldRefresh := isAccessTokenExpired || now.After(c.RefreshAfter)
```
szuecs, line 106: not sure if we need the `RefreshAfter` clause, because we can just use the RefreshToken in case AccessToken is expired, see also https://github.com/zalando/skipper/pull/1609#discussion_r531645346
Do I miss something?

filters/auth/grant.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil && c.isAccessTokenExpired() {
+		// Refresh failed and we no longer have a valid access token.
+		loginRedirect(ctx, f.config)
+		return
+	}
+
+	tokeninfo, err := f.config.TokeninfoClient.getTokeninfo(token.AccessToken, ctx)
+	if err != nil {
+		if err == errInvalidToken {
+			log.Errorf("Error while calling tokeninfo: %v.", err)
```
szuecs, line 164: Error while calling -> Failed to call

filters/auth/grant.go
```diff
@@ -0,0 +1,197 @@
...
+func (f *grantFilter) Response(ctx filters.FilterContext) {
+	container, ok := ctx.StateBag()[oidcClaimsCacheKey].(tokenContainer)
+	if !ok {
+		return
+	}
+
+	req := ctx.Request()
+	c, err := CreateCookie(f.config, req.Host, container.OAuth2Token)
+	if err != nil {
+		log.Errorf("Error while generating cookie: %v.", err)
```
szuecs, line 191: Error while generating -> Failed to generate

filters/auth/grantcallback.go
```diff
@@ -0,0 +1,92 @@
...
+		} else {
+			serverError(ctx)
+		}
+		return
+	}
+
+	redirectURI, _ := f.config.RedirectURLs(req)
+	token, err := f.exchangeAccessToken(code, redirectURI)
+	if err != nil {
+		log.Errorf("Error when exchanging access token: %v.", err)
```
szuecs, line 67: Error when exchanging -> Failed to exchange

filters/auth/grantcallback.go
```diff
@@ -0,0 +1,92 @@
...
+		serverError(ctx)
+		return
+	}
+
+	c, err := CreateCookie(f.config, req.Host, token)
+	if err != nil {
+		log.Errorf("Error while creating OAuth grant cookie: %v.", err)
```
szuecs, line 74: Error while creating -> Failed to create

filters/auth/grant.go
aryszka: after the linked considerations, this only depends on whether we see any value in refreshing the access token early on.

filters/auth/grantconfig.go
```diff
@@ -0,0 +1,280 @@
...
+			c.Tracer,
+		)
+		if err != nil {
+			return err
+		}
+		c.TokeninfoClient = client
+	}
+
+	if c.AuthClient == nil {
+		c.AuthClient = net.NewClient(net.Options{})
```
szuecs, line 171: We should pass the parameters, that we can here via Options.
For example Tracer, MaxIdleConnectionsPerHost and ConnectionTimeout

filters/auth/grant.go
```diff
@@ -0,0 +1,197 @@
...
+		return
+	}
+
+	rsp := ctx.Response()
+	rsp.Header.Add("Set-Cookie", c.String())
```
aryszka, line 196: In an earlier state of the PR, we only set the cookie when there was a successful login or refresh. Currently we set it in every case when there is a valid token. Is this on purpose, as in were there any problems detected with the previous approach?

filters/auth/grantconfig.go
```diff
@@ -0,0 +1,280 @@
...
+
+	return params
+}
+
+func (c *OAuthConfig) GetClientID() string {
+	if c.ClientID != "" {
+		return c.ClientID
+	}
+
+	if id, exists := c.SecretsProvider.GetSecret(c.ClientIDFile); exists {
```
szuecs, line 236: exists -> ok 
To use `ok` is the "Go way"

filters/auth/grantconfig.go
```diff
@@ -0,0 +1,280 @@
...
+
+	return ""
+}
+
+func (c *OAuthConfig) GetClientSecret() string {
+	if c.ClientSecret != "" {
+		return c.ClientSecret
+	}
+
+	if secret, exists := c.SecretsProvider.GetSecret(c.ClientSecretFile); exists {
```
szuecs, line 248: exists -> ok 
To use `ok` is the "Go way"

filters/auth/grantcookie.go
szuecs: As discussed with @aryszka we can just drop this func and its use

filters/auth/grantcookie.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"github.com/zalando/skipper/secrets"
+	"golang.org/x/oauth2"
+)
+
+type cookie struct {
+	AccessToken  string    `json:"access_token"`
+	RefreshToken string    `json:"refresh_token"`
+	Expiry       time.Time `json:"expiry,omitempty"`
+	RefreshAfter time.Time `json:"refresh_after,omitempty"`
```
szuecs, line 17: @aryszka RefreshAfter can also be deleted, right?

filters/auth/grantcookie.go
```diff
@@ -0,0 +1,134 @@
...
+
+func CreateCookie(config OAuthConfig, host string, t *oauth2.Token) (*http.Cookie, error) {
+	c := cookie{
+		AccessToken:  t.AccessToken,
+		RefreshToken: t.RefreshToken,
+		Expiry:       t.Expiry,
+	}
+
+	if !config.DisableRefresh {
+		c.RefreshAfter = refreshAfter(t.Expiry)
```
szuecs, line 100: Do we need this and the DisableRefresh config?
I think if we have no RefreshToken then we do not refresh, if we have one we refresh.
Any reason to disable this on purpose?

filters/auth/grantprep.go
```diff
@@ -0,0 +1,33 @@
+package auth
+
+import "github.com/zalando/skipper/eskip"
+
+const (
+	defaultCallbackRouteID = "__oauth2_grant_callback"
+	defaultCallbackPath    = "/.well-known/oauth2-callback"
```
szuecs, line 7: this should be documented in docs/..../.md

filters/auth/grantprep.go
```diff
@@ -0,0 +1,33 @@
...
+	defaultTokenCookieName = "oauth-grant"
```
szuecs, line 8: this should be documented in docs/..../.md

filters/auth/grantprep.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+type grantPrep struct {
+	config OAuthConfig
+}
+
+func (p *grantPrep) Do(r []*eskip.Route) []*eskip.Route {
```
szuecs, line 15: Add godoc:
```
Do adds automatically a global callback route to the routes
```

skipper.go
```diff
@@ -600,16 +600,65 @@ type Options struct {
...
+	// OAuth2ClientIDFile, the path of the file containing the OAuth2 client id of
+	// the current service, used to exchange the access code.
+	OAuth2ClientIDFile string
+
+	// OAuth2ClientSecretFile, the path of the file containing the secret associated
+	// with the ClientID, used to exchange the access code.
+	OAuth2ClientSecretFile string
+
+	// OAuth2CallbackPath contains the path where the OAuth2 callback requests with the
+	// authorization code should be redirected to.
```
szuecs, line 641: defaults to `/.well-known/oauth2-callback`

skipper.go
```diff
@@ -1200,6 +1258,33 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
+		oauthConfig.TokeninfoURL = o.OAuthTokeninfoURL
+		oauthConfig.SecretFile = o.OAuth2SecretFile
+		oauthConfig.ClientID = o.OAuth2ClientID
+		oauthConfig.ClientSecret = o.OAuth2ClientSecret
+		oauthConfig.ClientIDFile = o.OAuth2ClientIDFile
+		oauthConfig.ClientSecretFile = o.OAuth2ClientSecretFile
+		oauthConfig.CallbackPath = o.OAuth2CallbackPath
+		oauthConfig.AuthURLParameters = o.OAuth2AuthURLParameters
+		oauthConfig.SecretsProvider = sp
+		oauthConfig.Secrets = secrets.NewRegistry()
```
szuecs, line 1274: we can also use `o.SecretsRegistry` or do I miss something?

szuecs:
I added a couple of comments, thanks for doing the great work!

filters/auth/grant.go
aryszka: I think the solution is to change `==` to `!=`. We only need to log the unknown errors when calling tokeninfo.

filters/auth/grantcookie.go
aryszka: correct :+1:

filters/auth/grantcookie.go
aryszka: we can drop it. I don't recall the reason for this. @muaazsaleem by chance, do you remember if we added this on purpose in the preliminary prototype?

filters/auth/grantprep.go
aryszka: Is it necessary? the grantPrep is not an exported type. The right place to document what it does would be godoc for the `OAuthConfig.NewGrantPreprocessor` method.

skipper.go
aryszka: Looks legit.

Unrelated: just wondering the Registry.GetEncrypter() is not synchronized, shouldn't it be? See: https://github.com/zalando/skipper/blob/master/secrets/registry.go#L29

config/mapflags.go
szuecs: @AlexanderYastrebov we can let it like this but this is not true for String() it's normal to have value receiver, see also https://go-talks.appspot.com/github.com/szuecs/slides/go-class/go-class-kickstarter.slide#31 https://go-talks.appspot.com/github.com/szuecs/slides/go-class/go-class-kickstarter.slide#32

filters/auth/grant.go
szuecs: Ok then let's drop it!

filters/auth/grant.go
szuecs: or if not c.isAccessTokenExpired()

filters/auth/grantprep.go
szuecs: ok

skipper.go
szuecs: Yes, thanks will change it on Monday

filters/auth/grantcookie.go
muaazsaleem: I also don't see why we can't drop it.

filters/auth/grant.go
aryszka: i think the above condition should be inverted. Instead of `err == errInvalidToken`, it should be `err != errInvalidToken`, because errInvalidToken is a known case, and we don't need to log it, but when we don't know what the error is, e.g. a network failure when talking to tokeninfo, we want to log what the error was.

filters/auth/grantcookie.go
k-nut: Dropped in a5d702c

k-nut:
Ok, first of all hello from my side 👋 
I officially took this over from Olli today. I rebased on the latest master, and made the code changes that you pointed out. I'll hopefully be able to round up the documentation tomorrow so that we should be good to go here soon :)

skipper.go
szuecs: was merged

szuecs:
Thanks @k-nut !

docs/reference/filters.md
k-nut: I think you meant to refer to the `grantCallback` route that was also registered here. I removed it from the documentation as it is not directly related to `oauthGrant`

filters/auth/grant.go
k-nut: I couldn't find the version where the cookie was set conditionally. Do you have a commit-hash for me in which this was the case so that I can investigate?

k-nut:
Ok, I think all of the comments should be addressed now. It is getting a bit difficult to navigate this PR though. Please feel free to look over the (pretty atomic) commits that I made and let me know if I missed anything or misunderstood something.

docs/reference/filters.md
szuecs: yes

docs/tutorials/auth.md
```diff
@@ -245,3 +245,156 @@ Access to path `/` would be granted to everyone in `example.org`, however path `
...
+If the parameter is not given, it will be `/.well-known/oauth2-callback`
+
+```
+foo:
+    Path("/foo")
+    -> oauthGrant()
+    -> "http://localhost:9090";
+
+bar: 
+    Path("/bar")
```
szuecs, line 363: This should be either the default or `/oauth/callback` as defined above in the example how to [start skipper](https://github.com/zalando/skipper/pull/1609/files#diff-f3a465ccf6b9f785970a20701c0841b87572018455dd4e734f20d539d7b5d517R341).

filters/auth/grant.go
szuecs: I think it's fine like this, because we check that we have something in our statebag `ctx.StateBag()[oidcClaimsCacheKey].(tokenContainer)` that we set [here](https://github.com/zalando/skipper/pull/1609/files?short_path=f3a465c#diff-357ce891fbb16fce5b8c0e90e06a755b1bc39e7b354ae1ea9b4af027479bc89cR172). 
If we have nothing in Response we return early in the first 3 lines of the Response() function.

docs/tutorials/auth.md
k-nut: Wait, this isn't the callback route, it's just another route to showcase the `oauthGrant` filter. This part of the documentation had an additional route before to show how to register the `grantCallback` route but I removed that in commit e0926923f51de2306674a0b0b56fbd9f96997bef. I guess I could remove one of the two routes though? The second one doesn't really add any value...

szuecs: I agree, remove it

filters/auth/grantcookie.go
k-nut: You are right in that we only export his function so that we can use it in the test. It looks to me like this can't easily be changed though since the tests in `grant_test` require quite a bit of scaffolding (e.g. `grantQueryWithCookie`) and a lot of code would have to be moved as well. Or maybe I'm missing something here?
Looking at this made me realise though: The two new test files (`grantflowstate_test.go` and `grant_test.go`) are actually the only ones that have `auth_test` as a package. All others re-use `auth`. Would changing these tests to also live in the `auth` package be an alternative here?

szuecs: Yes, @k-nut you can create a separate test file with `auth` package and add your tests there that need access to `createCookie()`

docs/tutorials/auth.md
k-nut: Removed in 6825909

filters/auth/grant.go
aryszka: We should avoid setting the cookie unnecessarily. It's not a nice thing.

The commit-hash I am talking about: df776cf6d4c5a1f2af24784657ffee1c47ac1ff1

Though I don't think we should return to that hash at all, since the code has been improved and refactored a lot since then. I have a different suggestion, see further below. But first for reference: you can see in that hash, that the cookie is initially set on successful login in the callback route, then on refresh, the statebag key `oauth-grant-token` is used, but only when there was an actual refresh.

My suggestion now is to use the oidcClaimsCacheKey as is, unchanged. But in the grant filter, on the response, let's use a separate key, that would be only present if there was a refresh, and use that key for the cookie. We may have duplicate values in the statebag, but only when there was a refresh, so this is fine. WDYT?

filters/auth/grantcookie.go
k-nut: I added 1f8301226dfebda06e560f97e7ebcc8569ff0ad9 which introduces a couple of new helpers functions for the test to get test cookies (inspired by the one that we had already). Would this be acceptable?

k-nut: wait, that commit is broken. Fixing.

k-nut: Ok, this is commit 17b9590 now after amending

filters/auth/grant.go
szuecs: Let's do that

filters/auth/grantcookie.go
szuecs: yes that should be fine IIRC

filters/auth/grant.go
k-nut: I added a pretty naive implementation in commit f55f82e. Would that work or did I miss some context here?

szuecs:
:+1:

filters/auth/grant.go
aryszka: lgtm

aryszka:
:+1:

aryszka:
Thank you everybody for all the contributions here!

PR #1567, opened 2020-10-24
Adds copyStream buffer pool

AlexanderYastrebov:
Reduces memory allocation by reusing byte slices in `copyStream`.

**Profiling results**

```
bin/skipper -inline-routes='* -> inlineContent("Hello") -> <shunt>' -access-log-disabled -enable-profile
```

Baseline (`master`)
```
$ echo "GET http://localhost:9090/test" | vegeta attack -duration=10m -rate=0 --max-workers=500 | vegeta report
Requests      [total, rate, throughput]         16754506, 27924.03, 27923.66
Duration      [total, attack, wait]             10m0s, 10m0s, 8.097ms
Latencies     [min, mean, 50, 90, 95, 99, max]  89.763µs, 14.278ms, 9.62ms, 36.179ms, 43.932ms, 65.012ms, 228.741ms
Bytes In      [total, mean]                     83772530, 5.00
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:16754506
Error Set:
```
```
$ go tool pprof -top http://localhost:9911/debug/pprof/allocs
...
Showing nodes accounting for 194.44GB, 96.27% of 201.96GB total
Dropped 137 nodes (cum <= 1.01GB)
      flat  flat%   sum%        cum   cum%
  127.97GB 63.36% 63.36%      128GB 63.38%  github.com/zalando/skipper/proxy.copyStream
    9.04GB  4.48% 67.84%    10.13GB  5.01%  github.com/zalando/skipper/filters/builtin.(*inlineContent).Request
    8.16GB  4.04% 71.88%     8.43GB  4.17%  net/textproto.(*Reader).ReadMIMEHeader
       7GB  3.47% 75.35%        7GB  3.47%  net/http.Header.Clone
    5.95GB  2.95% 78.30%    23.89GB 11.83%  net/http.(*conn).readRequest
    5.46GB  2.70% 81.00%     5.46GB  2.70%  github.com/zalando/skipper/proxy.copyHeader
       5GB  2.47% 83.48%     5.12GB  2.53%  github.com/zalando/skipper/proxy.newContext
    4.02GB  1.99% 85.47%     6.21GB  3.08%  net/http.(*Request).WithContext (inline)
    4.01GB  1.99% 87.45%    15.52GB  7.69%  net/http.readRequest
[...]

PR #1616, opened 2020-11-20
Adds repeatContent filter

AlexanderYastrebov:
`repeatContent` filter generates response of specified size from repeated text

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -485,6 +485,21 @@ Example:
 * -> randomContent(42) -> <shunt>;
 ```
 
+## repeatContent
+
+Generate response of specified size from repeated text.
+
+Parameters:
+
+* text to repeat (string)
+* size of response in bytes (int)
+
+Example:
+
+```
+* -> repeatContent("I will not waste chalk. ", 1000) -> <shunt>;
+```
+
 ## latency
 
 Enable adding artificial latency
--- a/filters/diag/diag.go
+++ b/filters/diag/diag.go
@@ -12,6 +12,7 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"net/http"
+	"strconv"
 	"sync"
 	"time"
 
@@ -22,6 +23,7 @@ const defaultChunkSize = 512
 
 const (
 	RandomName           = "randomContent"
+	RepeatName           = "repeatContent"
 	LatencyName          = "latency"
 	ChunksName           = "chunks"
 	BandwidthName        = "bandwidth"
@@ -47,6 +49,16 @@ type random struct {
 	len  int64
 }
 
+type repeat struct {
+	bytes []byte
+	len   int64
+}
+
+type repeatReader struct {
+	bytes  []byte
+	offset int
+}
+
 type throttle struct {
 	typ       throttleType
 	chunkSize int
@@ -67,6 +79,15 @@ func kbps2bpms(kbps float64) float64 {
 // 	* -> randomContent(2048) -> <shunt>;
 //
 func NewRandom() filters.Spec { return &random{} }
+
+// NewRepeat creates a filter specification whose filter instances can be used
+// to respond to requests with a repeated text. It expects the text and
+// the byte length of the response body to be generated as arguments.
+// Eskip example:
+//
+// 	* -> repeatContent("x", 100) -> <shunt>;
+//
+func NewRepeat() filters.Spec { return &repeat{} }
 
 // NewLatency creates a filter specification whose filter instances can be used
 // to add additional latency to responses. It expects the latency in milliseconds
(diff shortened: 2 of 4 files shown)


docs/reference/filters.md
```diff
@@ -485,6 +485,21 @@ Example:
 * -> randomContent(42) -> <shunt>;
 ```
 
+## repeatContent
+
+Generate response of specified size from repeated text.
```
szuecs, line 490: Maybe a hint on memory usage would be great, because you can easily explode in memory `n=len(text) m=size: O(n*m)`

AlexanderYastrebov: The implementation does not allocate any buffers:
```
$ bin/skipper -inline-routes='* -> repeatContent("Hello", 100000000000) -> <shunt>' --write-timeout-server=10m
[APP]INFO[0000] route settings, reset, route: : * -> repeatContent("Hello", 100000000000) -> <shunt> 
127.0.0.1 - - [24/Nov/2020:00:31:30 +0100] "GET / HTTP/1.1" 200 100000000000 "-" "curl/7.58.0" 153980 localhost:9090 - -
```
```
$ curl localhost:9090 -o /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 93.1G    0 93.1G    0     0   619M      0 --:--:--  0:02:33 --:--:--  615M
```

AlexanderYastrebov: I think some background is missing on why I propose this filter.

While working on https://github.com/zalando/skipper/pull/1567 I have discovered there is no suitable way to generate response of a given size.

I have tried `randomContent` filter but it had low performance which was later improved by #1570 but is still subpar due to relatively slow random generator. 

Another option is to use fixed-sized files via `static` filter but it [allocates buffers](https://github.com/zalando/skipper/pull/1567#issuecomment-716702302) to copy file content and is not convenient as one has to prepare files. 

This implementation does not allocate any buffers but fills up a buffer provided by proxy's `copyStream` with repeated pattern using a fast copy-from-itself trick (like [bytes.Repeat](https://golang.org/pkg/bytes/#Repeat) does) - e.g. it copies `hello` into the buffer, then copies `hello` from the buffer into itself resulting in `hellohello` then again resulting in `hellohellohellohello` etc. So e.g. `repeatContent("x", 10000)` fills up default proxy's buffer of 8192 size just in log2(8192)=13 iterations (contrary to naive for-loop implementation which would require 8192 iterations).

I think this implementation is the fastest way to serve content of a given size in skipper. Serving file could be faster because of [`io.ReaderFrom`](https://golang.org/src/net/http/server.go#L581) which utilized `sendfile` if we would use `io.CopyBuffer` or support `io.ReaderFrom` in proxy's `copyStream` instead of hand-rolled loop.

AlexanderYastrebov: Here are some numbers for serving 8k response:
```
$ # randomContent
$ bin/skipper -inline-routes='* -> randomContent(8192) -> <shunt>' -access-log-disabled
$ echo "GET http://localhost:9090/" | vegeta attack -duration=1m -rate=0 --max-workers=500 | vegeta report 
Requests      [total, rate, throughput]         358698, 5978.31, 5970.25
Duration      [total, attack, wait]             1m0s, 1m0s, 80.97ms
Latencies     [min, mean, 50, 90, 95, 99, max]  217.34µs, 83.591ms, 88.377ms, 92.926ms, 94.426ms, 99.129ms, 276.118ms
Bytes In      [total, mean]                     2938454016, 8192.00
Status Codes  [code:count]                      200:358698  
Error Set:
```
```
$ # static
$ truncate -s 8k 8k.bin
$ bin/skipper -inline-routes='* -> static("/", "./") -> <shunt>' -access-log-disabled
$ echo "GET http://localhost:9090/8k.bin" | vegeta attack -duration=1m -rate=0 --max-workers=500 | vegeta report 
Requests      [total, rate, throughput]         1010975, 16849.50, 16847.79
Duration      [total, attack, wait]             1m0s, 1m0s, 6.089ms
Latencies     [min, mean, 50, 90, 95, 99, max]  114.431µs, 12.959ms, 9.043ms, 30.901ms, 38.343ms, 54.711ms, 122.876ms
Bytes In      [total, mean]                     8281907200, 8192.00
Status Codes  [code:count]                      200:1010975  
Error Set:
```
```
$ repeatContent
$ bin/skipper -inline-routes='* -> repeatContent("x", 8192) -> <shunt>' -access-log-disabled
$ echo "GET http://localhost:9090/" | vegeta attack -duration=1m -rate=0 --max-workers=500 | vegeta report 
Requests      [total, rate, throughput]         1155795, 19262.71, 19259.43
Duration      [total, attack, wait]             1m0s, 1m0s, 10.216ms
Latencies     [min, mean, 50, 90, 95, 99, max]  100.193µs, 11.104ms, 7.197ms, 27.356ms, 33.557ms, 45.277ms, 134.422ms
Bytes In      [total, mean]                     9468272640, 8192.00
Status Codes  [code:count]                      200:1155795  
Error Set:
```

szuecs: The memory size in skipper would be the size of the input string and the buffers would increase to number of unwritten bytes to the socket.
Ok great to see, had to check again that you work on the one byte slice and read in the wrapper io.LimitReader

szuecs:
:+1:

aryszka:
I think it would be more preferable, and more Skipper native if the repeatContent would be able to repeat any content like this:

```
repeatContent(9) -> inlineContent("foo")
```

or

```
repeatContent(9) -> static("/", /var/www")
```

I think the non-allocating behavior should be possible to achieve with the combination of `repeatContent` and `inlineContent`. This way the filter would be more composable and usable in more different scenarios.

AlexanderYastrebov:
The proposed implementation is straightforward and generates response of a given size, useful for throughput measurement as well as testing corner cases like buffer sizes etc. The text to repeat in fact is just for amusement, the filter is purely diagnostic.

What you suggest feels more higher level (repeats upstream response N times). It would be interesting to see the implementation as it probably would require buffering the upstream response to repeat. 

May we consider your suggestion not as an alternative but rather a complement implementation, i.e. can we have both this `repeatContent` and suggested `repeatResponse` (or `repeatBody`)? Also I am not against at all changing the name for this implementation in order not to take up the name if that is a concern.

filters/diag/diag.go
```diff
@@ -146,6 +167,59 @@ func (r *random) Request(ctx filters.FilterContext) {
...
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	return &repeat{[]byte(text), len}, nil
+}
+
+func (r *repeat) Request(ctx filters.FilterContext) {
+	ctx.Serve(&http.Response{
+		StatusCode: http.StatusOK,
+		Header:     http.Header{"Content-Length": []string{strconv.FormatInt(r.len, 10)}},
```
AlexanderYastrebov, line 203: I have added `Content-Length` which eliminates `Transfer-Encoding: chunked` and improves transfer speed significantly for large response:
```
$ bin/skipper -inline-routes='* -> repeatContent("x", 10000000000) -> <shunt>' -access-log-disabled
```
without content length:
```
$ curl localhost:9090 -o /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 9536M    0 9536M    0     0   729M      0 --:--:--  0:00:13 --:--:--  724M
```
with content length:
```
$ curl localhost:9090 -o /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 9536M  100 9536M    0     0  1945M      0  0:00:04  0:00:04 --:--:-- 1964M
```

aryszka:
> May we consider your suggestion not as an alternative but rather a complement implementation, i.e. can we have both this repeatContent and suggested repeatResponse (or repeatBody)? Also I am not against at all changing the name for this implementation in order not to take up the name if that is a concern.

Totally fair. We can use this cool name for the current implementation, too. :+1:

aryszka:
:+1:

szuecs:
:+1:

PR #1622, opened 2020-11-29
Fixes absorb(Silent) filter spec name

AlexanderYastrebov:
* Logs consumed bytes at regular intervals of one second

### DIFF
--- a/filters/diag/absorb.go
+++ b/filters/diag/absorb.go
@@ -2,7 +2,9 @@ package diag
 
 import (
 	"io"
+	"io/ioutil"
 	"net/http"
+	"time"
 
 	"github.com/zalando/skipper/filters"
 	"github.com/zalando/skipper/filters/flowid"
@@ -14,6 +16,8 @@ const AbsorbName = "absorb"
 
 // AbsorbSilentName contains the name of the absorbSilent filter.
 const AbsorbSilentName = "absorbSilent"
+
+const loggingInterval = time.Second
 
 type absorb struct {
 	logger logging.Logger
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

filters/diag/absorb.go
```diff
@@ -60,11 +61,18 @@ func NewAbsorbSilent() filters.Spec {
 	return withLogger(true, nil)
 }
 
-func (a absorb) Name() string                                            { return AbsorbName }
-func (a absorb) CreateFilter(args []interface{}) (filters.Filter, error) { return a, nil }
-func (a absorb) Response(filters.FilterContext)                          {}
+func (a *absorb) Name() string {
```
aryszka, line 64: why do we need pointers here? Not using pointers is a good indicator that the state of the filter doesn't change.

filters/diag/absorb.go
```diff
@@ -82,25 +90,12 @@ func (a absorb) Request(ctx filters.FilterContext) {
...
 
-	var count = 0
-	buf := make([]byte, 1<<12)
-	for {
-		n, err := req.Body.Read(buf)
-		count += n
-		if !a.silent {
-			a.logger.Infof("request %s, consumed bytes: %d", id, count)
-		}
+	count, err := io.Copy(ioutil.Discard, req.Body)
```
aryszka, line 93: this simplification would remove the visibility of slow streams. Let's put back the partial logs for the !silent absorb

filters/diag/absorb.go
AlexanderYastrebov: I am following this suggestion https://dave.cheney.net/2016/03/19/should-methods-be-declared-on-t-or-t

aryszka: TBH, not too convincing in this particular case but I am fine with it :smile:

filters/diag/absorb.go
AlexanderYastrebov: I've stumbled upon absorb name issue while testing megabyte-sized requests, for them such logging was excessive. The indication of the slow request could be above `received request to be absorbed: %s` and absent subsequent `request finished:`.

For the usecase where one needs to consume request fast this solution is swift while existing is slow.

aryszka: I see. Then maybe we would log the absorbed chunks only with a time threshold, like if now.Since(timeOfLastLog) > time.Second then log and set timeOfLastLog to now. There were some cases when we were investigating how the incoming body is being streamed.

For fast absorb, we still have the absorbSilent, where could use of course io.Copy

AlexanderYastrebov: I've updated PR as suggested

szuecs:
:+1:

aryszka:
:+1:

PR #1624, opened 2020-11-30
Fix/apimonitoring

szuecs:
TL;DR: Speedup 2x and allocs 1/2

Before this PR
```
% go test -bench=Benchmark_CreateFilter_FullConfigSingleApiNakadi -benchmem ./filters/apiusagemonitoring -count 5 G -vE 'ERRO|INF
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/filters/apiusagemonitoring
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          1257           1051305 ns/op          337476 B/op       3865 allocs/op
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          1215            990956 ns/op          337656 B/op       3865 allocs/op
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          1236            926331 ns/op          337245 B/op       3865 allocs/op
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          1222            941034 ns/op          337497 B/op       3865 allocs/op
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4           789           1430919 ns/op          337588 B/op       3865 allocs/op
```

cache based on map+sync.Mutex (only one regexp.Compile possible at a time)
```
% go test -bench=Benchmark_CreateFilter_FullConfigSingleApiNakadi -benchmem ./filters/apiusagemonitoring -count 5 G -vE 'ERRO|INF
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/filters/apiusagemonitoring
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2590            527939 ns/op           55714 B/op       1379 allocs/o
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2474            447375 ns/op           55688 B/op       1379 allocs/o
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2511            525478 ns/op           55711 B/op       1379 allocs/o
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2509            449144 ns/op           55751 B/op       1379 allocs/o
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2293            451283 ns/op           55655 B/op       1379 allocs/o
PASS
ok      github.com/zalando/skipper/filters/apiusagemonitoring   8.266s
```

cache based on sync.Map (could have concurrent regexp.Compile runs)
```
% go test -bench=Benchmark_CreateFilter_FullConfigSingleApiNakadi -benchmem ./filters/apiusagemonitoring -count 5 G -vE 'ERRO|INFO|level=info|level=error'
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/filters/apiusagemonitoring
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2534            446211 ns/op           55806 B/op       1381 allocs/op
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2520            447876 ns/op           55691 B/op       1381 allocs/op
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2373            447189 ns/op           55745 B/op       1381 allocs/op
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2473            438565 ns/op           55700 B/op       1381 allocs/op
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2491            449118 ns/op           55780 B/op       1381 allocs/op
PASS
ok      github.com/zalando/skipper/filters/apiusagemonitoring   6.833s
```

More clean implementation
```
% go test -bench=Benchmark_CreateFilter_FullConfigSingleApiNakadi -benchmem ./filters/apiusagemonitoring -count 5 G -vE 'ERRO|INFO|level=info|level=error'
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/filters/apiusagemonitoring
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2576            449664 ns/op           55642 B/op       1379 allocs/op
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2644            578603 ns/op           55726 B/op       1379 allocs/op
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2010            591767 ns/op           55861 B/op       1379 allocs/op
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2334            476716 ns/op           55667 B/op       1379 allocs/op
Benchmark_CreateFilter_FullConfigSingleApiNakadi-4          2384            464545 ns/op           55704 B/op       1379 allocs/op
PASS
ok      github.com/zalando/skipper/filters/apiusagemonitoring   8.197s

```

### DIFF
--- a/filters/apiusagemonitoring/spec.go
+++ b/filters/apiusagemonitoring/spec.go
@@ -25,6 +25,19 @@ var (
 	regCache = sync.Map{}
 )
 
+func loadOrCompileRegex(pattern string) (*regexp.Regexp, error) {
+	var err error
+	var reg *regexp.Regexp
+	regI, ok := regCache.Load(pattern)
+	if !ok {
+		reg, err = regexp.Compile(pattern)
+	} else {
+		reg = regI.(*regexp.Regexp)
+	}
+	regCache.Store(pattern, reg)
+	return reg, err
+}
+
 // NewApiUsageMonitoring creates a new instance of the API Monitoring filter
 // specification (its factory).
 func NewApiUsageMonitoring(
@@ -55,15 +68,7 @@ func NewApiUsageMonitoring(
 		}
 	}
 
-	var err error
-	var realmsTrackingMatcher *regexp.Regexp
-	realmsTrackingMatcherI, ok := regCache.Load(realmsTrackingPattern)
-	if !ok {
-		// compile realms regex
-		realmsTrackingMatcher, err = regexp.Compile(realmsTrackingPattern)
-	} else {
-		realmsTrackingMatcher = realmsTrackingMatcherI.(*regexp.Regexp)
-	}
+	realmsTrackingMatcher, err := loadOrCompileRegex(realmsTrackingPattern)
 	if err != nil {
 		log.Errorf(
 			"api-usage-monitoring-realmsTrackingPattern-tracking-pattern (global config) ignored: error compiling regular expression %q: %v",
@@ -71,7 +76,6 @@ func NewApiUsageMonitoring(
 		realmsTrackingMatcher = regexp.MustCompile("services")
 		log.Warn("defaulting to 'services' as api-usage-monitoring-realmsTrackingPattern-tracking-pattern (global config)")
 	}
-	regCache.Store(realmsTrackingPattern, realmsTrackingMatcher)
 
 	// Create the filter Spec
 	var unknownPathClientTracking *clientTrackingInfo = nil // client metrics feature is disabled
@@ -242,23 +246,13 @@ func (s *apiUsageMonitoringSpec) buildPathInfoListFromConfiguration(apis []*apiC
 			}
 			existingPathPattern[pathPattern] = info
 
-			// Compile the regular expression
-			var err error
-			var pathPatternMatcher *regexp.Regexp
-			pathPatternMatcherI, ok := regCache.Load(pathPattern)
-			if !ok {
-				pathPatternMatcher, err = regexp.Compile(pathPattern)
-				if err != nil {
-					regCache.Store(pathPattern, nil)
-					log.Errorf(
-						"args[%d].path_templates[%d] ignored: error compiling regular expression %q for path %q: %v",
-						apiIndex, templateIndex, pathPattern, info.PathTemplate, err)
-					continue
-				}
-				regCache.Store(pathPattern, pathPatternMatcher)
-
-			} else {
-				pathPatternMatcher = pathPatternMatcherI.(*regexp.Regexp)
+			pathPatternMatcher, err := loadOrCompileRegex(pathPattern)
+			if err != nil {
+				regCache.Store(pathPattern, nil)
+				log.Errorf(
+					"args[%d].path_templates[%d] ignored: error compiling regular expression %q for path %q: %v",
+					apiIndex, templateIndex, pathPattern, info.PathTemplate, err)
+				continue
 			}
 			if pathPatternMatcher == nil {
 				continue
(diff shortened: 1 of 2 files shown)


filters/apiusagemonitoring/spec.go
```diff
@@ -52,15 +54,24 @@ func NewApiUsageMonitoring(
...
+
+	var err error
+	var realmsTrackingMatcher *regexp.Regexp
+	realmsTrackingMatcherI, ok := regCache.Load(realmsTrackingPattern)
+	if !ok {
+		// compile realms regex
+		realmsTrackingMatcher, err = regexp.Compile(realmsTrackingPattern)
+	} else {
+		realmsTrackingMatcher = realmsTrackingMatcherI.(*regexp.Regexp)
+	}
```
tkrop, line 66: This strongly looks like a pattern that can be extracted into a common function `loadOrCreateRegex(pattern string, *regexp.Regexp alt) (*regexp.Regexp, Error)` for all three cases.

filters/apiusagemonitoring/spec.go
```diff
@@ -232,13 +243,27 @@ func (s *apiUsageMonitoringSpec) buildPathInfoListFromConfiguration(apis []*apiC
...
+						"args[%d].path_templates[%d] ignored: error compiling regular expression %q for path %q: %v",
+						apiIndex, templateIndex, pathPattern, info.PathTemplate, err)
+					continue
+				}
+				regCache.Store(pathPattern, pathPatternMatcher)
+
+			} else {
+				pathPatternMatcher = pathPatternMatcherI.(*regexp.Regexp)
+			}
+			if pathPatternMatcher == nil {
```
tkrop, line 263: I think this case cannot exist.

filters/apiusagemonitoring/spec_test.go
```diff
@@ -584,3 +584,62 @@ func Test_CreatePathPattern(t *testing.T) {
...
+		f, err := spec.CreateFilter(args)
+		if err != nil {
+			b.Fatalf("Failed to run CreateFilter: %v", err)
+		}
+		filter, ok := f.(*apiUsageMonitoringFilter)
+		if !ok || filter == nil {
+			b.Fatal("Failed to convert filter")
+		}
+	}
+}
```
tkrop, line 645: Not sure that this makes sense to add this as benchmark because it does not make the choice visible in the end.

tkrop (commented):
Is route and filter creation run in parallel at all? Anyhow, would also prefer a thread-safe implementation.

Looking on the benchmark results, makes me wondering whether the gain isn't even too low to go for this optimization. As I said in the chat, I would go for caching the whole filter instead of only the regular expressions, since the string concatenation process itself is very heavy weighted. Instead of caching many regular expressions, it would be much more promising to cache the whole filters.

I always hoped, that Skipper would provide such an optimization making the assumption that same filter arguments result in the same filter setup, but as quick fix I would suggest to implement the same logic in this filter only.

filters/apiusagemonitoring/spec.go
szuecs: makes sense!

filters/apiusagemonitoring/spec.go
szuecs: if we cached a nil object it can and you can do `nil.(*regexp.Regexp)`

filters/apiusagemonitoring/spec_test.go
szuecs: We can improve on every change the code path that is tested

filters/apiusagemonitoring/spec.go
aryszka: this load-compile-store logic appears 3 times, let's extract it to a single function

szuecs: I am on it

szuecs: pushed a fix

filters/apiusagemonitoring/spec.go
```diff
@@ -242,23 +246,13 @@ func (s *apiUsageMonitoringSpec) buildPathInfoListFromConfiguration(apis []*apiC
...
-						apiIndex, templateIndex, pathPattern, info.PathTemplate, err)
-					continue
-				}
-				regCache.Store(pathPattern, pathPatternMatcher)
-
-			} else {
-				pathPatternMatcher = pathPatternMatcherI.(*regexp.Regexp)
+			pathPatternMatcher, err := loadOrCompileRegex(pathPattern)
+			if err != nil {
+				regCache.Store(pathPattern, nil)
```
aryszka, line 251: this regCache.Store is unnecessary, because loadOrCompileRegex stores the nil for the pathPattern

szuecs: I think you have an outdated view, not?

szuecs: thanks, true, fixed

aryszka:
:+1:

tkrop:
:+1:

tkrop:
If you want to go for this solution. Looks fine for me and you are right of storing incorrect patterns with `nil`.

PR #1625, opened 2020-12-01
fix: filters should not hold reference to spec

szuecs:


aryszka:
:+1:

tkrop:
:+1:

tkrop:
What is the rational behind this rule?

There should not be a problem with this, since it is a reference to static singleton responsible for creating all filters.

aryszka:
I think it's not a rule per se, just was judged to be cleaner like this..?

szuecs:
@aryszka from my side yes, it's not a hard rule in case it makes sense to keep a reference, but the number of strings we duplicate here is limited and therefore I would rather have a little copy than a little dependency.

PR #1626, opened 2020-12-01
kubernetes dataclient - mark OriginMarker as deprecated

szuecs:
This code was for measuring a zalando internal SLO, which we do differently now. We mark the option as deprecated, log message and ignore the option


aryszka:
:+1:

szuecs:
:+1:

Issue #1628, opened 2020-12-03
Allow configuring the rate limit error response


herojan:
**Is your feature request related to a problem? Please describe.**
Our services are supposed to return json for all responses, and [problem json](https://opensource.zalando.com/restful-api-guidelines/#176) for error responses.  The rate limit filter returns a plaintext response, which means services need to handle that possibility specifically. 

<details>
  <summary>Example of a current ratelimit-exceeded response.</summary>
  
  ```
curl -i -H "Authorization: Bearer ${TOKEN}" "${URL}"
HTTP/2 429
date: Mon, 08 Apr 2019 15:03:21 GMT
content-type: text/plain; charset=utf-8
content-length: 18
retry-after: 55
server: Skipper
x-content-type-options: nosniff
x-rate-limit: 300

Too Many Requests
  ```
  
</details>

**Describe the solution you would like**
We would like to be able to configure the `clusterClientRatelimit` filter so that it returns a problem json response, with `Content-Type` of `application/problem+json`.

The solutions we had thought of was either a new config flag for skipper startup, indicating it should return json error responses, or a new param for the `clusterClientRatelimit` filter. After discussing internally we decided to go with a new param for `clusterClientRatelimit`, but we're not sure if it should be one or two extra params atm, one for `Content-Type`, one for response body.

Current solution:
**Add two new params, so now we would have e.g.**
```
clusterClientRatelimit("groupA", 10, "1h", "X-Forwarded-For", "application/problem+json", "{\"title\":\"Rate limit exceeded\",\"detail\":\"You have exceeded your ratelimit. See the retry-after and x-rate-limit headers for details.\",\"status\":429}")
```

**Describe alternatives you've considered (optional)**
We've thus far allowed teams to deal with it, but now that we have some time to work on it we'd like to solve this issue, since using skipper as an API Gateway should still allow the service to be compliant with the [API guidelines](https://opensource.zalando.com/restful-api-guidelines) 

**Would you like to work on it?**
Yes

AlexanderYastrebov:
It seems possible to achieve this with existing filters:
```
bin/skipper -inline-routes='* -> inlineContentIfStatus(429, "{\"title\":\"Rate limit exceeded\",\"detail\":\"You have exceeded your ratelimit. See the retry-after and x-rate-limit headers for details.\",\"status\":429}", "application/problem+json") -> ratelimit(1, "1m") -> "https://example.com"' -enable-ratelimits
```

herojan:
[...]

PR #1629, opened 2020-12-03
Allow specifying a custom content type and response body for rate limiting

herojan:
Initial attempt at closing https://github.com/zalando/skipper/issues/1628, not sure if it needs more testing and documentation

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -1354,11 +1354,17 @@ Parameters:
 * number of allowed requests per time period (int)
 * time period for requests being counted (time.Duration)
 * optional parameter to set the same client by header, in case the provided string contains `,`, it will combine all these headers (string)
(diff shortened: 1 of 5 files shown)


herojan:
Gonna close and reopen since I only changed docs so I don't think anything broke in that commit

szuecs:
@herojan any reason why not to support it in clusterRatelimit as well?

AlexanderYastrebov:
https://github.com/zalando/skipper/issues/1628#issuecomment-738301077

herojan:
> any reason why not to support it in clusterRatelimit as well?

No I just wasn't sure which filters to change, will I change clientRatelimit, ratelimit, clusterRatelimit, clusterClientRatelimit?

I could also try to make the `inlineContentIfStatus` filter (which @AlexanderYastrebov pointed out in the issue) work for responses from the rate limit filters, not sure how much work that would be atm though, I think each filter ignores the other filters atm and short circuits the chain.

PR #1632, opened 2020-12-04
Set status and headers in rate limit filter, instead of short circuiting

herojan:
Alternate fix for https://github.com/zalando/skipper/issues/1628.

Changes the rate limit filter so that it sets the status and headers instead of returning its own response immediately. This allows the inlineContentIfStatus filter to work with rate limits

### DIFF
--- a/filters/ratelimit/ratelimit.go
+++ b/filters/ratelimit/ratelimit.go
@@ -399,10 +399,9 @@ func (f *filter) Request(ctx filters.FilterContext) {
 	}
 
 	if !rateLimiter.AllowContext(ctx.Request().Context(), s) {
-		ctx.Serve(&http.Response{
-			StatusCode: http.StatusTooManyRequests,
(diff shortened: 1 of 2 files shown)


filters/ratelimit/ratelimit.go
```diff
@@ -399,10 +399,8 @@ func (f *filter) Request(ctx filters.FilterContext) {
 	}
 
 	if !rateLimiter.AllowContext(ctx.Request().Context(), s) {
-		ctx.Serve(&http.Response{
-			StatusCode: http.StatusTooManyRequests,
-			Header:     ratelimit.Headers(&f.settings, rateLimiter.RetryAfter(s)),
-		})
+		ctx.Response().StatusCode = http.StatusTooManyRequests
```
AlexanderYastrebov, line 402: This does not seem right. The whole point of `ctx.Serve` is to shunt route and do not call subsequent filters and backend.

PR #1633, opened 2020-12-04
support HTTP_PROXY, HTTPS_PROXY and NO_PROXY in http clients

szuecs:
feature: support HTTP_PROXY, HTTPS_PROXY and NO_PROXY in net.Client that is used in authclient for example

### DIFF
--- a/net/httpclient.go
+++ b/net/httpclient.go
@@ -133,6 +133,7 @@ func (c *Client) CloseIdleConnections() {
 // which can be nil to get the
 // https://godoc.org/github.com/opentracing/opentracing-go#NoopTracer.
 type Options struct {
+	Proxy func(req *http.Request) (*url.URL, error)
 	// DisableKeepAlives see https://golang.org/pkg/net/http/#Transport.DisableKeepAlives
 	DisableKeepAlives bool
 	// DisableCompression see https://golang.org/pkg/net/http/#Transport.DisableCompression
@@ -233,8 +234,12 @@ func NewTransport(options Options) *Transport {
 	if options.ExpectContinueTimeout == 0 {
 		options.ExpectContinueTimeout = options.Timeout
 	}
+	if options.Proxy == nil {
+		options.Proxy = http.ProxyFromEnvironment
+	}
 
 	htransport := &http.Transport{
+		Proxy:                  options.Proxy,
 		DisableKeepAlives:      options.DisableKeepAlives,
 		DisableCompression:     options.DisableCompression,
 		ForceAttemptHTTP2:      options.ForceAttemptHTTP2,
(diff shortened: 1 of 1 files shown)


net/httpclient.go
```diff
@@ -233,8 +234,12 @@ func NewTransport(options Options) *Transport {
 	if options.ExpectContinueTimeout == 0 {
 		options.ExpectContinueTimeout = options.Timeout
 	}
+	if options.Proxy == nil {
```
AlexanderYastrebov, line 237: This would not allow to create transport with `nil` proxy like it was before. 
Should `options.Proxy` be rather passes into `Transport` as is and configured on authClient side?
https://github.com/zalando/skipper/blob/9f8c62c12b9edb6d2f6147221a178383f8f3a4d6/filters/auth/authclient.go#L47-L54

net/httpclient.go
```diff
@@ -133,6 +133,7 @@ func (c *Client) CloseIdleConnections() {
 // which can be nil to get the
 // https://godoc.org/github.com/opentracing/opentracing-go#NoopTracer.
 type Options struct {
+	Proxy func(req *http.Request) (*url.URL, error)
```
AlexanderYastrebov, line 136: `// Proxy see https://golang.org/pkg/net/http/#Transport.Proxy`

net/httpclient.go
szuecs: I think it‘s great to act similar to the default Go and you can easily pass a noop func with that signature if you want.
As another thought of why it’s fine like this: If someone starts it with the environment variables set she will likely want this as default. We could also set default somewhere else, but I think right now it’s great like this, because everyone will just pass nothing and expect it to work.

net/httpclient.go
szuecs: So you mean that we drop the default nil case as written in the golang docs, but then everyone needs to do the inconvenient passing of the http.ProxyFromEnvironment (or how the name was). So for usability I would prefer it this way, because net.Client should be a better http Client configured with great defaults with a similar “interface”.

AlexanderYastrebov: Here I mean to add a comment like all other fields have to stay consistent.

net/httpclient.go
AlexanderYastrebov: Ok, not a big deal, the proxy could be still disabled by passing `func(*http.Request) (*url.URL, error) {return nil, nil}`
> // If Proxy is nil or returns a nil *URL, no proxy is used.

net/httpclient.go
szuecs: done, thanks!

szuecs:
:+1:

aryszka:
:+1:

PR #1643, opened 2020-12-09
Extends templating with request/response placeholders

AlexanderYastrebov:
Adds support for the following placeholders:
* request header
* request url query
* request cookie
* response header

See #1444

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -19,13 +19,27 @@ all: * -> filter1 -> filter2 -> "http://127.0.0.1:1234/";
 ## Template placeholders
 
 Several filters support template placeholders (`${var}`) in string parameters.
-Template placeholder is replaced by the value that is looked up in filter context path parameters.
-Missing template placeholder value interpretation depends on the filter.
+
+Template placeholder is replaced by the value that is looked up in the following sources:
+
+* request url path (`${request.path}`)
+* request url query (if starts with `request.query.` prefix, e.g `${request.query.q}` is replaced by `q` query parameter value)
+* request headers (if starts with `request.header.` prefix, e.g `${request.header.Content-Type}` is replaced by `Content-Type` request header value)
+* request cookies (if starts with `request.cookie.` prefix, e.g `${request.cookie.PHPSESSID}` is replaced by `PHPSESSID` request cookie value)
+* response headers (if starts with `response.header.` prefix, e.g `${response.header.Location}` is replaced by `Location` response header value)
+* filter context path parameters (e.g. `${id}` is replaced by `id` path parameter value)
+
+Missing value interpretation depends on the filter.
 
 Example route that rewrites path using template placeholder:
 
 ```
 u1: Path("/user/:id") -> setPath("/v2/user/${id}") -> <loopback>;
+```
+
+Example route that creates header from query parameter:
+```
+r: Path("/redirect") && QueryParam("to") -> status(303) -> setResponseHeader("Location", "${request.query.to}") -> <shunt>;
 ```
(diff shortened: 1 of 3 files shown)


docs/reference/filters.md
```diff
@@ -19,15 +19,31 @@ all: * -> filter1 -> filter2 -> "http://127.0.0.1:1234/";
 ## Template placeholders
 
 Several filters support template placeholders (`${var}`) in string parameters.
-Template placeholder is replaced by the value that is looked up in filter context path parameters.
-Missing template placeholder value interpretation depends on the filter.
+
+Template placeholder is replaced by the value that is looked up in the following sources:
+
+* request headers (if starts with `request.header.` prefix, e.g `${request.header.Content-Type}` is replaced by `Content-Type` request header value)
+* request url query (if starts with `request.url_query.` prefix, e.g `${request.url_query.q}` is replaced by `q` query parameter value)
```
aryszka, line 26: i would have named it just simply `request.query`

docs/reference/filters.md
```diff
@@ -19,15 +19,31 @@ all: * -> filter1 -> filter2 -> "http://127.0.0.1:1234/";
...
 
 Example route that rewrites path using template placeholder:
 
 ```
 u1: Path("/user/:id") -> setPath("/v2/user/${id}") -> <loopback>;
 ```
 
+Example route that rewrites query parameter into header:
+```
+q2h: Path("/oauth2/tokeninfo") && QueryParam("access_token") ->
```
aryszka, line 41: maybe we could use a more "innocent" example? Meaning where there is no access token in the URL, so that we don't encourage this practice.

aryszka:
lgtm, I only see some minor nit picking cases, fairly opinionated ones.

aryszka:
:+1:

docs/reference/filters.md
AlexanderYastrebov: A justification to have `url` in the name was to follow the go API where `query` is a field of `url` which is the field of `request`.

Here I've named it the same as it is named in [lua](https://github.com/zalando/skipper/blob/master/docs/reference/scripts.md#other-request-fields).
There it was named like that `url_<something>` because `url` was already defined as scalar string, so it would not be  possible to use `url.query` (in such case `url` has to be table and `query` would be a key).

So on the one hand it is the same naming as in lua (good) on the other it inherits lua's implementation history (bad).

AlexanderYastrebov:
@aryszka 
> some minor nit picking
not at all, let me fix

docs/reference/filters.md
aryszka: I see, and I missed that before. I leave it to your consideration :slightly_smiling_face:

docs/reference/filters.md
AlexanderYastrebov: This does not encourage but rather acknowledges that such [not recommended practice](https://tools.ietf.org/html/rfc6750#section-2.3) may exist. 

I agree it makes documentation reader to raise eyebrows which is not good so I will change it to something neutral.

docs/reference/filters.md
AlexanderYastrebov: I have changed it to `request.query`:
1. there is no need to follow go API naming
2. makes no sense to inherit lua implementation history
3. there is a `setQuery` filter (not a `setUrlQuery`)

I have also added `request.path` as a complement to the existing `setPath` filter.

szuecs:
:+1:

szuecs:
thanks great work!

aryszka:
:+1:

PR #1646, opened 2020-12-09
Match ingress to service using endpoints

### LINKED ISSUE #1645: Skipper creates shunt routes for valid deployments
**Describe the bug**
While upgrading our apigateway that uses skipper as a library from version `v0.11.144` to version `v0.11.204` we encoutered issues with skipper creating shunt routes that were not created with the older version.

Changes in `v0.11.174` seem to have introduced the issue. Skipper will wrongly try to match the targetPort instead of the containerPort when matching the ingress to the service.

**To Reproduce**
This `deployment.yaml` can be used to reproduce the problem:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: whoami-deployment
  labels:
    app: whoami
spec:
  replicas: 1
  selector:
    matchLabels:
      app: whoami
  template:
    metadata:
      labels:
        app: whoami
    spec:
      containers:
        - name: whoami
          image: traefik/whoami:v1.6.1
          ports:
            - containerPort: 80
              name: my-port
              protocol: TCP
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: whoami-deployment
  name: whoami-service
  namespace: default
spec:
  ports:
    - name: this-is-my-service-port-name
      port: 8080
      protocol: TCP
      targetPort: my-port
  selector:
    app: whoami
  sessionAffinity: None
  type: ClusterIP
---
kind: Ingress
metadata:
  annotations:
  labels:
    app: whoami
  name: whoami-ingress
  namespace: default
spec:
  rules:
    - host: whoami.test.domain.com
      http:
        paths:
          - backend:
              serviceName: whoami-service
              servicePort: 8080
```

**Expected behavior**
When applied on skipper `v0.11.144`, looking at skipper routes we can see the following:

```
$ curl localhost:9911/routes ; printf "\n\n"; curl -H "host:whoami.test.domain.com" localhost:8080
kube__healthz: Path("/kube-system/healthz") && SourceFromLast("10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128")
  -> enableAccessLog(4, 5)
  -> lifo(2000, 20000, "3s")
  -> status(200)
  -> <shunt>;

kube__redirect: Header("X-Forwarded-Proto", "http") && Weight(1000)
  -> enableAccessLog(4, 5)
  -> lifo(2000, 20000, "3s")
  -> redirectTo(308, "https:")
  -> <shunt>;

kube_default__whoami_ingress__whoami_test_domain_com____whoami_service: Host(/^whoami[.]test[.]domain[.]com$/)
  -> enableAccessLog(4, 5)
  -> lifo(2000, 20000, "3s")
  -> "http://10.96.237.94:8080";
```

```
Hostname: whoami-deployment-6786bf95fd-6hdhg
IP: 127.0.0.1
IP: 172.17.0.3
RemoteAddr: 192.168.64.28:34076
GET / HTTP/1.1
Host: whoami.test.domain.com
User-Agent: curl/7.64.1
Accept: */*
Accept-Encoding: gzip
```

**Observed behavior**
While skipper `v0.11.174` will return these routes:
```
$ curl localhost:9911/routes ; printf "\n\n"; curl -H "host:whoami.test.domain.com" localhost:8080
kube__healthz: Path("/kube-system/healthz") && SourceFromLast("10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128")
  -> enableAccessLog(4, 5)
  -> lifo(2000, 20000, "3s")
  -> status(200)
  -> <shunt>;

kube__redirect: Header("X-Forwarded-Proto", "http") && Weight(1000)
  -> enableAccessLog(4, 5)
  -> lifo(2000, 20000, "3s")
  -> redirectTo(308, "https:")
  -> <shunt>;

kube_default__whoami_ingress__whoami_test_domain_com____whoami_service: Host(/^whoami[.]test[.]domain[.]com$/)
  -> enableAccessLog(4, 5)
  -> lifo(2000, 20000, "3s")
  -> status(502)
  -> inlineContent("no endpoints")
  -> <shunt>;
```

```
no endpoints
```

---

We have some code changes ready that will change the matching behaviour of skipper to use the containerPort instead of the targetPort. The routes will then contain the IP and port of the endpoints resource:

```
$ curl localhost:9911/routes ; printf "\n\n"; curl -H "host:whoami.test.domain.com" localhost:8080
kube__healthz: Path("/kube-system/healthz") && SourceFromLast("10.0.0.0/8", "192.168.0.0/16", "172.16.0.0/12", "127.0.0.1/8", "fd00::/8", "::1/128")
  -> enableAccessLog(4, 5)
  -> lifo(2000, 20000, "3s")
  -> status(200)
  -> <shunt>;

kube__redirect: Header("X-Forwarded-Proto", "http") && Weight(1000)
  -> enableAccessLog(4, 5)
  -> lifo(2000, 20000, "3s")
  -> redirectTo(308, "https:")
  -> <shunt>;

kube_default__whoami_ingress__whoami_test_domain_com____whoami_service: Host(/^whoami[.]test[.]domain[.]com$/)
  -> enableAccessLog(4, 5)
  -> lifo(2000, 20000, "3s")
  -> "http://172.17.0.3:80";
```

```
Hostname: whoami-deployment-6786bf95fd-6hdhg
IP: 127.0.0.1
IP: 172.17.0.3
RemoteAddr: 172.17.0.1:46686
GET / HTTP/1.1
Host: whoami.test.domain.com
User-Agent: curl/7.64.1
Accept: */*
Accept-Encoding: gzip
```

We would like to get some more feedback on this. I will create a draft PR as well so you can take a look.

bl1nk:
This commit changes matching of ingresses to services to use the containerPort intead of the targetPort. Using the targetPort can lead to skipper being unable to route traffic to the target.

Two tests in TestConvertPathRule were failing but are not valid cases to test as Kubernetes will not allow you to create a service with two ports of the same name. They have been removed.

Fixes #1645


dataclients/kubernetes/testdata/ingress/ingress-data/ing-with-prule-named-port.yaml
```diff
@@ -41,6 +41,6 @@ subsets:
   - ip: 10.2.9.103
   - ip: 10.2.9.104
   ports:
-  - name: bar
+  - name: baz
```
szuecs, line 44: name does not matter as long targetPort is referenced by number to the ep port

dataclients/kubernetes/ingressdefinitions_test.go
```diff
@@ -38,16 +65,76 @@ func TestGetTargetPort(t *testing.T) {
...
 					},
-				}},
+				},
+			},
 			svcPort:     definitions.BackendPort{Value: 80},
 			expected:    nil,
 			errExpected: true,
 		},
+		{
+			name: "multiple service ports, by value",
```
szuecs, line 75: please use the new fixture test suite

dataclients/kubernetes/testdata/ingress/ingress-data/ing-with-prule-named-port.yaml
bl1nk: Isn't this a misconception? I understood it the following way: Kubernetes will reference the service port names, not the targert port names in the endpoint ports. 

So let's say we have the following resources on the cluster

```yaml
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: whoami-deployment
  name: whoami-service
  namespace: default
spec:
  clusterIP: 10.102.22.101
  ports:
  - name: this-is-my-service-port-name
    port: 8080
    protocol: TCP
    targetPort: my-port
  selector:
    app: whoami
  sessionAffinity: None
  type: ClusterIP
---
kind: Ingress
metadata:
  labels:
    app: whoami
  name: whoami-ingress
  namespace: default
spec:
  rules:
  - host: whoami.test.domain.com
    http:
      paths:
      - backend:
          serviceName: whoami-service
          servicePort: 8080
        pathType: ImplementationSpecific
```

Then the endpoint created by Kubernetes looks like this

```yaml
apiVersion: v1
kind: Endpoints
metadata:
  labels:
    app: whoami-deployment
  name: whoami-service
  namespace: default
subsets:
- addresses:
  - ip: 172.17.0.3
    nodeName: minikube
    targetRef:
      kind: Pod
      name: whoami-deployment-6786bf95fd-fnqnq
      namespace: default
  ports:
  - name: this-is-my-service-port-name
    port: 80
    protocol: TCP
```

Which indeed references the endpoint port by service port name `this-is-my-service-port-name` not the target port name `my-port`.

Unless I am missing something important I think my change is needed to reflect how Kubernetes would create the endpoints in the testcase.

dataclients/kubernetes/ingressdefinitions_test.go
bl1nk: I added tests covering these cases using the new fixture test suite in f5fda805e43059c699043f9a1d81e6b0f6ccc7cc

This test was testing `getServicePort()` specifically as it replaced `getTargetPort()`. Would you like me to remove these tests and only rely on the new tests instead?

szuecs: yes only the fixture tests

dataclients/kubernetes/testdata/ingress/ingress-data/ing-with-prule-named-port.yaml
szuecs: thanks reviewing this lead me to write https://github.com/zalando/skipper/issues/1645#issuecomment-742755534

dataclients/kubernetes/ingressdefinitions_test.go
bl1nk: Done. I put the tests into `dataclients/kubernetes/testdata/ingress/service-ports`. Is there any naming convention to follow? Happy to move the tests into `named-ports` or `ingress-data` if you see them fit one of these.

dataclients/kubernetes/testdata/ingress/ingress-data/ing-with-prule-named-port.yaml
bl1nk: Thanks for the writeup!

bl1nk:
Rebased onto master, Also fixed naming of the endpoints introduced in 4dcbf4bfffaf6d8f31bc7397dbd05181f4ff6857. Please take a look.

szuecs:
:+1:

aryszka:
Hi @bl1nk , thanks for the PR!

Generally looks good, but I think it still doesn't cover all the cases. This condition checks the endpoints.subsets.ports.name: https://github.com/zalando/skipper/pull/1646/files#diff-bcaf7ab9e470ce06d42cb44a2ac2cce9ef3d31e07e6a7050b81441edb9409141R77

But  in my observation, this name in the endpoints is not always present. I think we can fix this such, that if there is only one port in the endpoints, then use that, otherwise use the matching by name of the current PR.

I'll verify this now in a cluster with a couple of cases.

aryszka:
Just to confirm: I checked a few cases, extending this solution to return the single endpoint port when there is only one, should work.

aryszka:
would add this to the condition "...when there is only one": **"and the endpoint port has no name"**

szuecs:
@bl1nk please fix as @aryszka suggested, I confirmed it in sig-network and we tested it.

bl1nk:
Added the condition to return the endpoint of only one endpoint exists. Also added a test that verifies that an endpoint is returned if only one exists and it has no name. Please take a look again!

aryszka:
:+1:

dataclients/kubernetes/testdata/ingress/service-ports/multiple-by-value.yaml
```diff
@@ -0,0 +1,72 @@
...
+  name: myapp-service
+subsets:
+  - addresses:
+      - ip: 10.3.0.3
+        targetRef:
+          kind: Pod
+          name: myapp-deployment-6786bf95fd-fnqnq
+    ports:
+      - name: web
+        port: 80
```
szuecs, line 71: This does not look correct to me. It should be 8080, because it needs to be in sync with pod container spec ports

szuecs:
:+1:

dataclients/kubernetes/testdata/ingress/service-ports/multiple-by-value.yaml
bl1nk: I agree, fixed and pushed

szuecs:
I think there are some duplicate tests (named ports tests) and some specs that do not exist in Kubernetes. I will clean this up later.

szuecs:
@bl1nk that was fast :D thanks !

szuecs:
:+1:

aryszka:
:+1:

PR #1511, opened 2020-08-25
Drains tokeninfo response body to enable TCP connection reuse

AlexanderYastrebov:
AuthClient used by `oauthTokeninfo*` filters does not consume response body when token is invalid:
https://github.com/zalando/skipper/blob/e3372b500d401c8133367fdcb208f35e73ee4022/filters/auth/authclient.go#L110-L114

According to https://golang.org/pkg/net/http/#Client.Do

> If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) 
may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.

That is requests with invalid token prevent tokeninfo TCP connection reuse which leads to higher latencies.

Consider the following setup:
```
$ cat /tmp/tokeninfo.eskip 
tokeninfo: Path("/oauth2/tokeninfo")
-> status(400) -> inlineContent("{\"error\":\"invalid_request\",\"error_description\":\"Access Token not valid\"}")
-> <shunt>;

test: Path("/test")
-> oauthTokeninfoAllScope("foo", "bar")
-> status(204)
-> <shunt>;
```
and
```
bin/skipper -access-log-disabled -oauth2-tokeninfo-url="http://localhost:9090/oauth2/tokeninfo" -routes-file=/tmp/tokeninfo.eskip 
```

Benchmark results:
```
$ wrk --latency -H "Authorization: Bearer X" -c 4 -d 12 http://localhost:9090/test
Running 12s test @ http://localhost:9090/test
  2 threads and 4 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.39ms    1.37ms   9.61ms   80.53%
    Req/Sec     1.91k     1.93k    4.87k    67.50%
  Latency Distribution
     50%  528.00us
     75%    2.34ms
     90%    3.54ms
     99%    5.25ms
  45670 requests in 12.01s, 6.32MB read
  Non-2xx or 3xx responses: 45670
Requests/sec:   3803.71
Transfer/sec:    538.61KB
```

Consuming response body with:
```diff
@@ -110,6 +112,7 @@ func (ac *authClient) getTokeninfo(token string, ctx filters.FilterContext) (map
        defer rsp.Body.Close()
 
        if rsp.StatusCode != 200 {
+               io.Copy(ioutil.Discard, rsp.Body)
                return doc, errInvalidToken
        }
```

reduces latencies severalfold:
```
[...]

PR #1647, opened 2020-12-09
Add grantLogout filter to revoke OAuth2 tokens

k-nut:
This is part 2 of #1609.

Quoting the original commit message here:

> The new `grantLogout()` filter allows defining routes that call the
> OAuth2 provider's token revocation endpoint. It attempts to revoke both
> the access and refresh tokens.
> 
> According to RFC 7009, the provider does not need to support access
> token revocation. The filter supports this case by ignoring a 401 error
> with a `unsupported_token_type` error when revoking the access token.
> 
> The filter also instructs the client to delete the grant token cookie
> by setting a `Set-Cookie` header with an empty value and a `MaxAge` of
> `-1`.
> 
> In order to use the filter, Skipper must be provided the token
> revocation endpoint URL using `-oauth2-revoke-token-url`.


filters/auth/grantlogout.go
```diff
@@ -0,0 +1,198 @@
...
+type grantLogoutFilter struct {
+	config OAuthConfig
+}
+
+type revokeErrorResponse struct {
+	Error            string `json:"error"`
+	ErrorDescription string `json:"error_description"`
+}
+
+func (grantLogoutSpec) Name() string { return GrantLogoutName }
```
AlexanderYastrebov, line 45: https://dave.cheney.net/2016/03/19/should-methods-be-declared-on-t-or-t

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,198 @@
...
+	if clientID == "" {
+		return "", errors.New("failed to create token revoke auth header: no client ID")
+	}
+
+	clientSecret := f.config.GetClientSecret()
+	if clientSecret == "" {
+		return "", errors.New("failed to create token revoke auth header: no client secret")
+	}
+
+	credentials := fmt.Sprint(clientID, ":", clientSecret)
```
AlexanderYastrebov, line 62: Why fmt.Sprint?
`credentials := clientID + ":" + clientSecret`

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,198 @@
...
+			}
+		}
+
+		message := fmt.Sprintf(
+			"%s revocation failed: %d %s",
+			tokenType,
+			revokeResponse.StatusCode,
+			errorDescription,
+		)
+		return errors.New(message)
```
AlexanderYastrebov, line 141: https://golang.org/pkg/fmt/#Errorf

AlexanderYastrebov, line 141: This whole block feels complicated. Would it make sense to inline `getErrorResponse` and see if it could be simplified?

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,198 @@
...
+
+	return nil
+}
+
+// revokeTokens revokes the access and refresh tokens in the grant token cookie.
+func (f grantLogoutFilter) revokeTokens(cookie *cookie) error {
+	if cookie.AccessToken != "" {
+		if err := f.revokeTokenType(accessTokenType, cookie.AccessToken); err != nil {
+			return err
+		}
```
AlexanderYastrebov, line 152: Failure to revoke AccessToken prevents revoke of RefreshToken. Is that ok?

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,198 @@
...
+		unauthorized(
+			ctx,
+			"",
+			missingToken,
+			req.Host,
+			fmt.Sprintf("Token cookie %v has no tokens.", f.config.TokenCookieName))
+		return
+	}
+
+	err = f.revokeTokens(c)
```
AlexanderYastrebov, line 188: IMO makes sense to inline revokeTokens to make the flow straightforward

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	return nil
+}
+
+func (f grantLogoutFilter) Request(ctx filters.FilterContext) {
+	req := ctx.Request()
+
+	c, err := extractCookie(req, f.config)
+	if err != nil || c == nil {
```
AlexanderYastrebov, line 168: extractCookie does not return nil cookie, this looks defensive. 

In general we should strive for strict method contract "either valid value or error" so that error checking is always straightforward `if err != nil`. That said even if extractCookie would return nil cookie it should be rather fixed not to do that.

filters/auth/grantlogout.go
AlexanderYastrebov: After inlining probably could be combined with above `if c.AccessToken == "" && c.RefreshToken == "" `

filters/auth/grantlogout_test.go
```diff
@@ -0,0 +1,194 @@
...
+)
+
+type testRevokeError struct {
+	Error            string `json:"error"`
+	ErrorDescription string `json:"error_description"`
+}
+
+func newGrantLogoutTestServer() *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		revoke := func(w http.ResponseWriter, r *http.Request) {
```
AlexanderYastrebov, line 20: IMO no need for closure inside closure:
```go
if r.URL.Path != "/revoke" {
    w.WriteHeader(http.StatusNotFound)
    return
}
// here the business follows

File: filters/auth/grant_test.go [hunk 8]
```diff
@@ -21,10 +21,14 @@ import (
 const (
 	testToken                = auth.TestToken
 	testRefreshToken         = auth.TestRefreshToken
+	testAccessTokenExpiresIn = auth.TestAccessTokenExpiresIn
+	testClientID             = "some-id"
+	testClientSecret         = "some-secret"
 	testAccessCode           = "quxquuxquz"
 	testSecretFile           = "testdata/authsecret"
-	testAccessTokenExpiresIn = int(time.Hour / time.Second)
```
AlexanderYastrebov, line 26: IMO should be time.Duration - there are many conversions to time.Duration and only one place where int is required.

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		return err
+	}
+
+	auth, err := f.basicAuthHeader()
+	if err != nil {
+		return err
+	}
+
+	revokeRequest.Header.Add("Authorization", auth)
```
AlexanderYastrebov, line 112: https://golang.org/pkg/net/http/#Request.SetBasicAuth

filters/auth/grantlogout.go
AlexanderYastrebov: Or maybe refactor `getErrorResponse` into 
```go
func responseToError(response *http.Response) error 

...
if revokeResponse.StatusCode == 400 {
    return responseToError(revokeResponse)
}
```

filters/auth/grantlogout.go
k-nut: ✅  Done

filters/auth/grantlogout.go
k-nut: ✅  Done. This has been replaced with the `SetBasicAuth` method.

filters/auth/grantlogout.go
k-nut: ✅  Done

filters/auth/grantlogout.go
k-nut: ✅  Done

filters/auth/grantlogout_test.go
k-nut: ✅  Done. That does look a lot nicer :)

filters/auth/grant_test.go
k-nut: ✅  Done

filters/auth/grantlogout.go
k-nut: I thought about this some more and settled on pretty much your approach eventually. Let me know what you think.

filters/auth/grantlogout.go
k-nut: This is no longer the case with the new version. It always tries to revoke both and only then checks if the overall request was successful or not.

filters/auth/grantlogout.go
k-nut: This is inlined now. I did not combine the ifs though as they are returning different types of responses.

k-nut:
@AlexanderYastrebov thank you for all of your comments! I think they should all be addressed now. Feel free to give it another look.

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,202 @@
...
+	if clientID == "" {
+		return nil, errors.New("failed to create token revoke auth header: no client ID")
+	}
+
+	clientSecret := f.config.GetClientSecret()
+	if clientSecret == "" {
+		return nil, errors.New("failed to create token revoke auth header: no client secret")
+	}
+
+	return &basicAuthCredentials{
```
AlexanderYastrebov, line 68: Maybe just `return clientID, clientSecret, nil`?

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,202 @@
...
+		if err == nil {
+			return &errorResponse
+		}
+	}
+
+	return nil
+}
+
+func responseToError(response *http.Response, tokenType string) error {
+	errorResponse := getErrorResponse(response)
```
AlexanderYastrebov, line 89: `getErrorResponse` does not check response status. If response does not contain proper `revokeErrorResponse` json the next line would :boom:  

Maybe try to inline `getErrorResponse` to get straightforward flow. With that it could bail out early, e.g. on json parse error:
```go
err = json.Unmarshal(buf, &errorResponse)
if err != nil {
    return err
}
```

AlexanderYastrebov: I see that it gets here only on 400, though we can not rely that 400 would always return proper json.

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,202 @@
...
+	if errorResponse.Error == unsupportedTokenTypeError && tokenType == accessTokenType {
+		// Provider does not support revoking access tokens, which can happen according to RFC 7009.
+		// In that case this is not really an error.
+		return nil
+	}
+	return fmt.Errorf(
+		"%s revocation failed: %d %s",
+		tokenType,
+		response.StatusCode,
+		errorResponse.Error+": "+errorResponse.ErrorDescription,
```
AlexanderYastrebov, line 99: Format string and concatenation should not be mixed, should be changed to `"%s revocation failed: %d %s: %s"`

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	revokeRequest.SetBasicAuth(authCredentials.clientId, authCredentials.clientSecret)
+	revokeRequest.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+
+	revokeResponse, err := f.config.AuthClient.Do(revokeRequest)
+	if err != nil {
+		return err
+	}
+	defer revokeResponse.Body.Close()
```
AlexanderYastrebov, line 140: There is a catch here, closing is not enough, the body should be read-out. I guess there are quite some other places in the codebase but this one we can fix on the spot (read it and pass buffer to responseToError instead of response).

See https://github.com/zalando/skipper/pull/1511 as an example

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,202 @@
...
+
+	return &basicAuthCredentials{
+		clientId:     clientID,
+		clientSecret: clientSecret,
+	}, nil
+}
+
+func getErrorResponse(response *http.Response) *revokeErrorResponse {
+	buf, err := ioutil.ReadAll(response.Body)
+	if err == nil && buf != nil {
```
AlexanderYastrebov, line 76: `&& buf != nil` is redundant

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	if c.RefreshToken != "" {
+		refreshTokenRevokeError = f.revokeTokenType(refreshTokenType, c.RefreshToken)
+		if refreshTokenRevokeError != nil {
+			log.Error(err)
+		}
+	}
+
+	if refreshTokenRevokeError != nil || accessTokenRevokeError != nil {
```
aryszka, line 194: From a human user perspective, I started thinking about whether we should return the 'ugly' proxy defined error page for every case. Maybe in some of the error cases it is fine to forward the request to the backend and let the user see a nice page from the application. E.g: if we cannot revoke a token because it's already expired. This also depends on how often we expect such cases, because if it happens too rarely, then probably not worth it. WDYT?

filters/auth/grantlogout.go
k-nut: For some reason I considered this to not be go-y. But I like it better this way. Fixed.

filters/auth/grantlogout.go
k-nut: Fixed.

filters/auth/grantlogout.go
k-nut: This is fixed with the new flow where we made sure that the body always gets read out.

filters/auth/grantlogout.go
k-nut: That is good to know! Fixed.

filters/auth/grantlogout.go
k-nut: Removed.

filters/auth/grantlogout.go
k-nut: I'm thinking that this should happen only rarely. 
If we tried to revoke a token that was expired already, this part of the RFC would actually come into play:
> Note: invalid tokens do not cause an error response since the client
   cannot handle such an error in a reasonable way.  Moreover, the
   purpose of the revocation request, invalidating the particular token,
   is already achieved.


The spec also considers this option:
> If the server responds with HTTP status code 503, the client must
   assume the token still exists and may retry after a reasonable delay.
   The server may include a "Retry-After" header in the response to
   indicate how long the service is expected to be unavailable to the
   requesting client.

This is something that we don't do, hence this would trigger the ugly error on our side. 

So the options are
- 200 (token revoked or the token was invalid to begin with)
- 400 (sever could not revoke the token) 
  - this might not cause an error on our side though if the server replied that it cannot revoke an auth_token. In that case, we treat this like a 200
- 503 (temporary overload), not handled gracefully on our side

So all in all my understanding is that this part of the code would only be run if the token response endpoint was behaving unexpectedly. Or of course if something in our implementation did not match the spec which would cause the requests to fail but we should be able to fix this on our side then.

aryszka: understood, thanks for the explanation!

filters/auth/grantcookie_test.go
```diff
@@ -8,8 +8,9 @@ import (
 )
 
 const (
-	TestToken        = "foobarbaz"
-	TestRefreshToken = "refreshfoobarbaz"
+	TestToken                = "foobarbaz"
+	TestRefreshToken         = "refreshfoobarbaz"
+	TestAccessTokenExpiresIn = time.Hour
```
szuecs, line 13: I wonder if you need all these exported or can you re-declare them in auth_test if you require them there, too?

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,192 @@
+//
+// grantLogout filter
+//
+// Used for revoking access/refresh tokens generated by the oauthGrant() filter.
+//
```
szuecs, line 5: I think we do not need this doc. We have it in docs.go, which is the package documentation and for users in the docs/*/*.md

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/zalando/skipper/filters"
+)
+
+const (
+	GrantLogoutName           = "grantLogout"
+	revokeTokenKey            = "token"
+	revokeTokenTypeKey        = "token_type_hint"
+	refreshTokenType          = "refresh_token"
+	accessTokenType           = "access_token"
+	unsupportedTokenTypeError = "unsupported_token_type"
```
szuecs, line 28: Linter rule that is not yet enabled, but we should one day enforce it, says:
unsupportedTokenTypeError -> errUnsupportedTokenType

filters/auth/grantlogout.go
```diff
@@ -0,0 +1,192 @@
...
+		return err
+	}
+
+	clientId, clientSecret, err := f.getBasicAuthCredentials()
+	if err != nil {
+		return err
+	}
+
+	revokeRequest.SetBasicAuth(clientId, clientSecret)
+	revokeRequest.Header.Add("Content-Type", "application/x-www-form-urlencoded")
```
szuecs, line 119: Header.Add -> Header.Set

I guess it's empty in this case, but I would not like to have 2 Content-Type headers if there will be some day a default set

filters/auth/grantlogout_test.go
```diff
@@ -0,0 +1,216 @@
...
+			}
+			return
+		}
+	}
+
+	t.Fatalf("Cookie not found in response: '%s'", cookieName)
+}
+
+func TestGrantLogout(t *testing.T) {
+	t.Log("create a test provider")
```
szuecs, line 95: can you cleanup the logs please?
It's nice to have it while you start working on tests but for the project not. :)

filters/auth/grantcookie_test.go
k-nut: Re-declared them in auth_test and added a comment here that they should be kept in sync.

filters/auth/grantlogout.go
k-nut: Removed ✅

filters/auth/grantlogout.go
k-nut: Renamed ✅

filters/auth/grantlogout.go
k-nut: Ah, good call. Done ✅

filters/auth/grantlogout_test.go
k-nut: Removed the logs here and also in `grant_test.go` and `granclaimsquery_test.go`. Those two were added in the previous PR.

k-nut:
@szuecs thank you for your comments! They should be all addressed now. Feel free to give it another look!

szuecs:
:+1:

aryszka:
:+1:

PR #1648, opened 2020-12-09
cleanup log entries of normal operation in healthcheck postprocessor

aryszka:
These logs report about frequent normal operation, not necessary to print them.


szuecs:
:+1:

Issue #1537, opened 2020-10-04
Easier use of eskip.Route when using Skipper as code


peterklijn:
**Is your feature request related to a problem? Please describe.**
When using Skipper as a product, and defining routes via the eskip format, Skipper has good and clear documentation.
But when using Skipper as a framework, and defining routes in code, documentation is sparse and I often find myself browsing through the source code to understand how to set predicates and filters.

**Describe the solution you would like**
In addition to providing more documentation for Skipper as a framework, I think some helper functions could really improve the usability of Skipper as a framework. These helpers will not only more the code more concise and readable, they will also serve as documentation.

Take the following route as an example (inspired by the [sticky traffic predicate example](https://opensource.zalando.com/skipper/reference/predicates/#traffic)):

```golang
import (
	"github.com/zalando/skipper/eskip"
	"github.com/zalando/skipper/filters/cookie"
	"github.com/zalando/skipper/predicates/traffic"
	"github.com/zalando/skipper/routing"
)

routes := []*eskip.Route{
	{
		Id: "example__catalog_test_a",
		Filters: []*eskip.Filter{
			{
				Name: cookie.ResponseCookieFilterName,
				Args: []interface{}{"catalog-test", "A"},
			},
		},
		Predicates: []*eskip.Predicate{
			{
				Name: routing.PathName,
				Args: []interface{}{"/catalog"},
			},
			{
				Name: traffic.PredicateName,
				Args: []interface{}{.15, "catalog-test", "A"},
			},
		},
		Backend: "http://catalog-test-a",
	},
	{
		Id: "example__catalog_test_b",
		Filters: []*eskip.Filter{
			{
				Name: cookie.ResponseCookieFilterName,
				Args: []interface{}{"catalog-test", "B"},
			},
		},
		Predicates: []*eskip.Predicate{
			{
				Name: routing.PathName,
				Args: []interface{}{"/catalog"},
			},
			{
				Name: traffic.PredicateName,
				Args: []interface{}{.3, "catalog-test", "B"},
			},
		},
		Backend: "http://catalog-test-b",
	},
	{
		Id: "example__catalog",
		Filters: []*eskip.Filter{
			{
				Name: cookie.ResponseCookieFilterName,
				Args: []interface{}{"catalog-test", "default"},
			},
		},
		Predicates: []*eskip.Predicate{
			{
				Name: routing.PathName,
				Args: []interface{}{"/catalog"},
			},
		},
		Backend: "http://catalog",
	},
}
```

[...]

PR #1650, opened 2020-12-11
suppress health check noise

universam1:
This change suppresses logging of the Skipper health check path in normal conditions, reducing log noise.

The logging is enabled if the health check fails or debugging levels are enabled

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -13,6 +13,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/filters/accesslog"
 	"github.com/zalando/skipper/filters/builtin"
 	"github.com/zalando/skipper/predicates/source"
 )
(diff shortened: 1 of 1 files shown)


universam1:
@aryszka @szuecs as discussed in slack

szuecs:
:+1:

dataclients/kubernetes/kube.go
```diff
@@ -348,9 +349,19 @@ func shuntRoute(r *eskip.Route) {
...
 	if !healthy {
-		status = http.StatusServiceUnavailable
+		logFilters[0].Args = []interface{}{http.StatusServiceUnavailable}
+	}
+	// log if unhealthy or a debug loglevel
+	if healthy && !log.IsLevelEnabled(log.DebugLevel) {
+		logFilters = append(logFilters, &eskip.Filter{
+			Name: accesslog.DisableAccessLogName,
+			Args: []interface{}{200},
+		})
```
AlexanderYastrebov, line 364: An alternative could be
```go
var healthcheckFilters []*eskip.Filter
	if healthy {
		if log.IsLevelEnabled(log.DebugLevel) {
			healthcheckFilters, _ = eskip.ParseFilters(`status(200)`)
		} else {
			healthcheckFilters, _ = eskip.ParseFilters(`disableAccessLog(200) -> status(200)`)
		}
	} else {
		healthcheckFilters, _ = eskip.ParseFilters(`status(503)`)
	}
```
Also related to #1537

universam1: This passing is awesome to know!

My personal preference is to use static type compiler support whenever possible, not sure how expensive runtime parsing is but I’d expect a magnitude!?

AlexanderYastrebov: It is more about readabilty, skipper parses hundreds of routes on route table update anyway.

szuecs: We normally don’t parse filters like this in the dataclient if it’s not eskip or inline route

dataclients/kubernetes/kube.go
```diff
@@ -348,9 +349,19 @@ func shuntRoute(r *eskip.Route) {
 }
 
 func healthcheckRoute(healthy, reverseSourcePredicate bool) *eskip.Route {
-	status := http.StatusOK
+	logFilters := []*eskip.Filter{{
```
aryszka, line 352: i think we can also call `logFilters` as just `filters`, because their main gig is not the logging part but the status.

aryszka:
:+1:

dataclients/kubernetes/kube.go
szuecs: merged

PR #1656, opened 2020-12-16
update staticcheck and gosec

szuecs:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -924,7 +924,7 @@ func (c *Config) ToOptions() skipper.Options {
 	}
 
 	if c.Certificates != nil && len(c.Certificates) > 0 {
-		var tlsVersion uint16
+		var tlsVersion uint16 = tls.VersionTLS13
 
 		for _, s := range []string{c.TLSMinVersion, defaultMinTLSVersion} {
 			switch s {
--- a/config/config.go
+++ b/config/config.go
@@ -168,6 +168,9 @@ type Config struct {
 	ClientCertFile string            `yaml:"client-tls-cert"`
 	Certificates   []tls.Certificate `yaml:"-"`
 
+	// TLS version
+	TLSMinVersion string `yaml:"tls-min-version"`
+
 	// API Monitoring
 	ApiUsageMonitoringEnable                       bool   `yaml:"enable-api-usage-monitoring"`
 	ApiUsageMonitoringRealmKeys                    string `yaml:"api-usage-monitoring-realm-keys"`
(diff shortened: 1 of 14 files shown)


szuecs:
:+1:

config/config.go
```diff
@@ -260,6 +263,9 @@ const (
 	defaultApiUsageMonitoringDefaultClientTrackingPattern = ""
 	defaultApiUsageMonitoringRealmsTrackingPattern        = "services"
 
+	// TLS
+	defaultMinTLSVersion = "1.2"
```
aryszka, line 267: :heart:

config/config.go
```diff
@@ -912,8 +924,27 @@ func (c *Config) ToOptions() skipper.Options {
...
+				tlsVersion = tls.VersionTLS13
+			case "1.2", "12":
+				tlsVersion = tls.VersionTLS12
+			case "1.1", "11":
+				tlsVersion = tls.VersionTLS11
+			case "1.0", "10":
+				tlsVersion = tls.VersionTLS10
+			default:
+				log.Infof("No valid minimal TLS version confiured (set to '%s'), fallback to default: %s", c.TLSMinVersion, defaultMinTLSVersion)
+				continue
```
aryszka, line 941: this loop is a bit confusing like this. The term `continue` will apply to the `switch` and not to the `for`. Also the `break` below always breaks the loop, so `defaultMinTLSVersion` is never applied, only the initial value from the `tlsVersion` var initializer, which is 1.3. This also means that `skipper -help` would return the wrong default in the help text.

If I understand correctly, the intention is:

- if no min TLS version is set, then we use 1.3 (initial value of the `tlsVersion` var)
- if it is set but invalid, then we use 1.2 (from the `defaultMinTLSVersion` const)
- if it is set and valid, then we use whatever is set

I would simplify the intentions a bit. Let's refactor it such that:

- there is only one default
- the var is initialized with that default
- drop the `for` loop
- override the default only when a valid param is defined
- change `Infof` to `Warnf`, or, my preference: error and abort, but only when the user input is not empty

szuecs: continue is only valid in `for` not `switch`, so break will do the correct thing, see https://play.golang.org/p/mptms3QIAlv
In the end the var setting in the beginning to tls.VersionTLS13 is never used, but gosec is not happy if we have a zero value there, because it is similar tricked like you, so likely the code is complicated :)

aryszka: OK, sounds fair.

Only for consideration: shall it rather abort on invalid TLS user input?

aryszka:
:+1:

PR #1658, opened 2020-12-16
clone the RequestURI field with the request when splitting for loopback

### LINKED ISSUE #1657: teeLoopback() drops the net/http.Request.RequestURI field
**Describe the bug**
teeLoopback() drops the net/http.Request.RequestURI field by not copying it when cloning. Normally, this field should not be used, but rather the net/http.Request.URL.Path field should be used instead. But there can be filters and other extensions that rely on the field while routing, and therefore it should be copied.

**To Reproduce**

1. Create two routes, one using teeLoopback such that it's looped to the other. The other route should have a filter that relies on the RequestURI field
2. make any request to the first route and see that the filter using the RequestURI field in the tee route fails

**Expected behavior**

The RequestURI field should be populated.

**Observed behavior**

The RequestURI field is empty.

aryszka:
fixes #1657 

### DIFF
--- a/proxy/split_test.go
+++ b/proxy/split_test.go
@@ -1,0 +1,63 @@
+package proxy_test
+
+import (
+	"io"
+	"io/ioutil"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/filters"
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

aryszka:
:+1:

PR #1660, opened 2020-12-19
Uses separate secret reader for grant filters

AlexanderYastrebov:
Using shared secret reader is not safe in multitenant setup where users
can define their own routes and thus can access grant secret via `bearerinjector` filter.

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -1195,8 +1195,7 @@ Skipper arguments:
 | `-oauth2-client-secret-file` | conditional | path to the file containing the OAuth2 client secret. Required if you have not set `-oauth2-client-secret`. Example: `-oauth2-client-secret-file=/path/to/client_secret` |
 | `-oauth2-client-id` | conditional | OAuth2 client ID for authenticating with your OAuth2 provider. Required if you have not set `-oauth2-client-id-file`. Example: `-oauth2-client-id=myclientid` |
 | `-oauth2-client-secret` | conditional | OAuth2 client secret for authenticating with your OAuth2 provider. Required if you have not set `-oauth2-client-secret-file`. Example: `-oauth2-client-secret=myclientsecret` |
-| `-credentials-paths` | conditional | path to the directories containing the cookie encryption secret client ID, and client secret files. Required if you want Skipper to automatically update the secrets periodically. Example: `-credentials-paths=/path/to/secrets/,/path/to/othersecrets/` |
-| `-credentials-update-interval` | no | the time interval for updating credentials from files. Example: `-credentials-update-interval=30s` |
+| `-credentials-update-interval` | no | the time interval for updating client id and client secret from files. Example: `-credentials-update-interval=30s` |
 | `-oauth2-access-token-header-name` | no | the name of the request header where the user's bearer token should be set. Default: `Authorization`. Example: `-oauth2-access-token-header-name=X-Grant-Authorization` |
 | `-oauth2-auth-url-parameters` | no | any additional URL query parameters to set for the OAuth2 provider's authorize and token endpoint calls. Example: `-oauth2-auth-url-parameters=key1=foo,key2=bar` |
 | `-oauth2-callback-path` | no | path of the Skipper route containing the `grantCallback()` filter for accepting an authorization code and using it to get an access token. Example: `-oauth2-callback-path=/oauth/callback` |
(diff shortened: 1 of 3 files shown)


docs/reference/filters.md
```diff
@@ -1195,8 +1195,7 @@ Skipper arguments:
 | `-oauth2-client-secret-file` | conditional | path to the file containing the OAuth2 client secret. Required if you have not set `-oauth2-client-secret`. Example: `-oauth2-client-secret-file=/path/to/client_secret` |
 | `-oauth2-client-id` | conditional | OAuth2 client ID for authenticating with your OAuth2 provider. Required if you have not set `-oauth2-client-id-file`. Example: `-oauth2-client-id=myclientid` |
 | `-oauth2-client-secret` | conditional | OAuth2 client secret for authenticating with your OAuth2 provider. Required if you have not set `-oauth2-client-secret-file`. Example: `-oauth2-client-secret=myclientsecret` |
-| `-credentials-paths` | conditional | path to the directories containing the cookie encryption secret client ID, and client secret files. Required if you want Skipper to automatically update the secrets periodically. Example: `-credentials-paths=/path/to/secrets/,/path/to/othersecrets/` |
-| `-credentials-update-interval` | no | the time interval for updating credentials from files. Example: `-credentials-update-interval=30s` |
```
AlexanderYastrebov, line 1199: * `-credentials-paths` was not really used, files added to `SecretsProvider` are automatically refreshed.
* `-credentials-update-interval` is shared/reused with `bearerinjector`, could be introduced as a separate option for grant filter if needed.

filters/auth/grantconfig.go
```diff
@@ -40,27 +40,13 @@ type OAuthConfig struct {
 	// authorization code should be redirected to.
 	CallbackPath string
 
-	// ClientID, the OAuth2 client id of the current service, used to exchange
-	// the access code. Must be set if ClientIDFile is not provided.
-	ClientID string
+	// GetClientID, returns the OAuth2 client id of the current service, used to exchange
+	// the access code
+	GetClientID func() string
```
AlexanderYastrebov, line 45: `secrets` package does not provide a suitable `getter` interface.
`SecretsReader` (including `StaticSecret` implementation) require input argument which would be fake or empty string: https://github.com/zalando/skipper/blob/5801bbb44c13d3018490f4cfd1d9e3a50d897913/secrets/readers.go#L6-L14
`SecretSource` returns two-dimensional array: 
https://github.com/zalando/skipper/blob/5801bbb44c13d3018490f4cfd1d9e3a50d897913/secrets/encrypter.go#L19-L21
so I've decided to simply use generic getter function here.

AlexanderYastrebov:
FYI @k-nut

filters/auth/grantconfig.go
AlexanderYastrebov: I've decided to revert refactoring as it did not improve much

docs/tutorials/auth.md
```diff
@@ -311,12 +311,13 @@ secret as follows:
...
     -oauth2-client-secret-file=/path/to/client_secret \
     -oauth2-secret-file=/path/to/cookie_encryption_secret \
-    -credentials-paths=/path/to/ \
     -credentials-update-interval=30s
 ```
 
-You must set the `-credentials-paths` argument to the directory containing the secrets. You 
-can modify the secret update interval using the `-credentials-update-interval` argument. In
+Care must be taken when used in conjunction with `-credentials-paths` option because files
+from `-credentials-paths` are available to `bearerinjector` filter.
```
szuecs, line 318: Does this hold true?
Your current change has 2 non shared SecretsPaths instances that are not sharing the files.

AlexanderYastrebov: Yes, `SecretsPaths` are separate but they may scan the same folder - lets say one configures
1. ` -credentials-paths=/tmp/secrets`
2. `-oauth2-client-secret-file=/tmp/secrets/grant.secret`

I.e. grant secret is located in credentials path an therefore is available to `bearerinjector`.

AlexanderYastrebov: I have added demo example

szuecs: Ok yes thanks that makes sense if you use the same path. We should state that you should not use the same path for these.

AlexanderYastrebov: I have added
> That is `-credentials-paths=/path/to` in above example will expose grant files to `bearerinjector` filter.

szuecs:
@AlexanderYastrebov can you resolve the conflicts please?

aryszka:
lgtm

aryszka:
:+1:

szuecs:
:+1:

Issue #1516, opened 2020-08-28
FlowId filter inject tag in open tracing span


evangelion1204:
**Is your feature request related to a problem? Please describe.**
By tagging the spans it would enable people to do a search in their specific provider for a given flow id. Because Skipper is generating the id it would also be the perfect place to inject be. Especially helpful for people in 24/7 needing quickly map issues from skipper metrics to the actual tracing.

**Describe the solution you would like**
Enhancing the `flowid` filter to set the tag - an optional parameter for startup could be the name of the tag.

**Would you like to work on it?**
Yes, but no time

lmineiro:
An alternative that could make sense as well would be to add another flowId `[Generator](https://github.com/zalando/skipper/blob/master/filters/flowid/filter.go#L18)` that uses as FlowID the TraceID. that FlowID generator would then reuse the current tracing identifier as the FlowID instead of the other way around. This is actually the long term strategy for Zalando.

evangelion1204:
Sound reasonable - any plan on applying that change? Will that be a new filter or modify the existing one?

szuecs:
@AlexanderYastrebov can we do this with templating maybe?
I am thinking of a header to [tracingTag](https://opensource.zalando.com/skipper/reference/filters/#tracingtag) case.
[FlowID](https://opensource.zalando.com/skipper/reference/filters/#flowid) is created or reused from HTTP requests and if we would make `getHeader`, such that the following would work, it would create a lot more value, than the original request would add:
```
 getHeader("X-Flow-Id", $flowid) -> tracingTag("my-flow-id", $flowid)
```

AlexanderYastrebov:
@szuecs
If we enhance `tracingTag` to support template placeholders the solution would look like
```
* -> flowId("reuse") -> tracingTag("xflowid", "${request.header.X-Flow-Id}")
```

I have also noticed the following code
https://github.com/zalando/skipper/blob/5801bbb44c13d3018490f4cfd1d9e3a50d897913/proxy/proxy.go#L1454-L1456
but it seems there is a bug because `FlowIDTag` is defined as
https://github.com/zalando/skipper/blob/5801bbb44c13d3018490f4cfd1d9e3a50d897913/proxy/tracing.go#L11
I.e. it is supposed to work but overwrites(?) `component` tag if header is present.

[...]

PR #1662, opened 2020-12-21
Enhances `tracingTag` filter to support template placeholders

### LINKED ISSUE #704: Consider: filter to put a specific header to the opentracing span as a tag
Would it make sense to create a filter that takes a request header and puts it into the current span? E.g:

```
setTracingTag("X-Flow-ID", "http.flow_id")
```

would take the value of the X-Flow-ID header if any and set it as the http.flow_id tag in the ingress span.

AlexanderYastrebov:
Fixes #704
See also https://github.com/zalando/skipper/issues/1516#issuecomment-748941875

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -2179,9 +2179,17 @@ Syntax:
 tracingTag("<tag_name>", "<tag_value>")
 ```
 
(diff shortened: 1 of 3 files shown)


aryszka:
:+1:

szuecs:
:+1:

PR #1668, opened 2020-12-24
Decouples eskip.Template from filters.FilterContext

AlexanderYastrebov:
#1528 unified template placeholder resolution across filters but introduced coupling between `eskip.Template` and `filters.FilterContext`
This PR decouples them by introducing `TemplateContext` so that it is possible to use template placeholders with non-filter context.


aryszka:
lgtm

eskip/template.go
```diff
@@ -48,20 +55,10 @@ func (t *Template) Apply(get TemplateGetter) string {
...
-
-// ApplyResponseContext evaluates the template using a filter context, request and response attributes to resolve the
-// placeholders. Returns true if all placeholders resolved to non-empty values.
-func (t *Template) ApplyResponseContext(ctx filters.FilterContext) (string, bool) {
-	return t.apply(contextGetter(ctx, true))
-}
-
-func contextGetter(ctx filters.FilterContext, response bool) func(key string) string {
-	return func(key string) string {
+func (t *Template) ApplyContext(ctx TemplateContext) (string, bool) {
```
szuecs, line 60: The signature change would break for library users, right?

AlexanderYastrebov: It will though I am not sure if there are any as these are quite new methods introduces in #1528

szuecs: Can you add ApplyResponseContext to call ApplyContext and add a deprecation warning for it?
Then we can get rid of it in v1.

AlexanderYastrebov: IMO we can merge because
a) either there are no users of these methods since they are pretty new
b) or if there are users they would only need to change `Apply*Context` to `ApplyContext` as `filters.FilterContext` implements `TemplateContext`

Keeping  `Apply*Context` undermines the purpose of this PR which is to decouple template from filter context.

szuecs: I asked in slack if someone from the community disagrees with merging.
Can you change the minor version in VERSION file to make the change more visible?
If nobody has a strong opinion until tomorrow then I would say let's merge it.

AlexanderYastrebov: Thank you, I have bumped VERSION file.

szuecs: Seems everyone is fine with the change from the community

szuecs:
:+1:

aryszka:
:+1:

PR #1670, opened 2020-12-29
Closes upgraded connection on copy completion

### LINKED ISSUE #1669: Websocket connection is not closed when backend server is stopped
Hi all, first thank you for Skipper, we use it as "intermediate" ingress controller in one of our projects.

**Describe the bug**
I'll first give some context:
We use Skipper as ingress controller in a Kubernetes deployment. We have an application that makes uses of websocket connections.
The front-end of this application connects to the Java application using websockets. The front-end will show a message as soon as the websocket connection is broken. In our case we break the websocket connection by removing the pod containing the application. When not using Skipper, this is working fine and the application does notice that the websocket connection is broken. 
In addition, when we use nginx as ingress controller, we experience the same behavior. However, the browser is unable to notice that the connection is broken when using Skipper as ingress controller.

**To Reproduce**

1. have a kubernetes cluster ready, e.g. using minkube
2. deploy Skipper and the test application: https://gist.github.com/LEDfan/24bc8a648ebe9c2291eaf9447b7389bb
3. warning: the above link will make the default serviceaccount a cluster-admin *just for testing*
3. add an entry to `/etc/hosts` for `127.0.0.1	tabsets.local `
4. forward skipper port, e.g. using `sudo kubectl --kubeconfig=$HOME/.kube/config port-forward --namespace default `k get pods --selector "application=skipper-ingress" --output=jsonpath={.items..metadata.name}
` 80:9999`
5. go to `tabsets.local` in your browser
6. when you change the slider, some data will be queried over the websockets connection. The graph should be updated whenever the slider changes.
7. kill the `tabsets` pod

**Expected behavior**

The page becomes gray (see screenshot) indicating that the websocket connection is broken.

**Observed behavior**

Skipper keeps the connection alive. Therefore the application isn't aware that the connection is broken. You can still change the slider in the application, and the application will send data over the websocket connection..

Is this something you think you can solve? Would be really useful for us!

AlexanderYastrebov:
Proxy copies data via `io.Copy` between upgraded request and backend connections in
two goroutines and waits for both copy calls to complete:

```
wait {
        copy backend to request // 1
        copy request to backend // 2
}
```

When backend connection is closed, first copy call completes
but the second is still blocked on read from request connection until
it is closed (i.e. client disconnects).

This change waits for either copy to complete and closes both request and
backend connections and thus unblocks the second copy.

Fixes #1669

### DIFF
--- a/proxy/upgrade_test.go
+++ b/proxy/upgrade_test.go
@@ -14,7 +14,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
-	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -91,6 +91,7 @@ func TestServeHTTP(t *testing.T) {
 		route                   string
 		method                  string
 		backendClosesConnection bool
+		backendHangs            bool
 		noBackend               bool
 		backendStatusCode       int
 		expectedResponseBody    string
@@ -135,8 +136,21 @@ func TestServeHTTP(t *testing.T) {
 			expectedResponseBody:    "BACKEND ERROR",
 			backendClosesConnection: true,
 		},
+		{
+			msg:               "backend hangs",
+			route:             `route: Path("/ws") -> "%s";`,
+			method:            http.MethodGet,
+			backendStatusCode: http.StatusSwitchingProtocols,
+			backendHangs:      true,
+		},
 	} {
 		t.Run(ti.msg, func(t *testing.T) {
+			ti := ti // trick race detector
+			var clientConnClosed atomic.Value
+			clientAlive := func() bool {
+				return clientConnClosed.Load() == nil
+			}
+
 			backend := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 				w.WriteHeader(ti.backendStatusCode)
 				if ti.backendClosesConnection {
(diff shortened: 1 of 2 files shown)


AlexanderYastrebov:
There was this interesting commit https://github.com/zalando/skipper/commit/d66d0b4917f7ee6ce4bd54a975bd5400bf460cb7 within initial implementation that basically does the same - waits for either copy to complete and closes both connections via defer. We need to understand why was it changed to waitgroup.

proxy/upgrade_test.go
```diff
@@ -157,12 +170,18 @@ func TestServeHTTP(t *testing.T) {
...
+
 				for {
 					s, err := bufrw.ReadString('\n')
-					if err != nil {
+					if err != nil && clientAlive() {
 						t.Errorf("error reading string: %v", err)
 						return
 					}
+
+					if ti.backendHangs {
```
AlexanderYastrebov, line 181: This if causes race check to fail although `ti` is used the same way several lines above without a problem :thinking:

AlexanderYastrebov:
I have changed implementation to use channel as it was in d66d0b4

proxy/upgrade_test.go
AlexanderYastrebov: `ti := ti` seems to do the trick

aryszka:
lgtm

szuecs:
@AlexanderYastrebov can you check the failing tests please?
I retriggered them twice

AlexanderYastrebov:
@szuecs Those are two flaky tests unrelated to this change as far as I can tell:
```
--- FAIL: TestBreakerRate (0.13s)
    --- FAIL: TestBreakerRate/open,_fail_to_close (0.04s)
        breaker_test.go:173: wrong response status: 500, expected 503
```
and
```
--- FAIL: TestCheckLocalRateLimit (1.23s)
    ratelimit_test.go:164: unexpected code, expected 429, got 200
```
I will have a look at them (for the latter I have an idea already, it is related to #1671)

szuecs:
@AlexanderYastrebov I dopped the label to let you add it when tests are good.

AlexanderYastrebov:
@szuecs Ok, we need this https://github.com/zalando/skipper/pull/1682 to be merged first

szuecs:
:+1:

aryszka:
:+1:

Issue #1526, opened 2020-09-28
Kubernetes: use automatically an allowlist for cluster.local domain


szuecs:
For example given that you have 

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: myingress
spec:
  rules:
  - host: test-app.ingress.cluster.local
    http:
      paths:
      - backend:
          serviceName: test-app-service
          servicePort: main-port
```

I wan to have an allow list of IP CIDRs automatically enforced with `Source` or `SourceFromLast` predicates to separate internal routes from being called from public internet.

A default should add only private subnets to the allow list.

AlexanderYastrebov:
I am wondering what kind of a setup would allow call to reach `test-app.ingress.cluster.local` from internet?..
<details>
I assume for HTTPS terminating LB or HTTPS serving skipper there will be no certificate for `cluster.local`.
If there is no LB but just service that exposes HTTP on public IP the `Source(FromLast)` is easily defeated by supplying `X-Forwarded-For`.
Plain HTTP load balancer/proxy in front? Or what do I miss?
</details>

szuecs:
@AlexanderYastrebov we tested it two times already that ALB sets the ip of the client that connects to the ALB as last entry in xff header, so it can’t be defeated.
We could build something that automatically sets a Source/FromLast predicate with internal IP cidrs to routes with domain .local.

AlexanderYastrebov:
@szuecs I mean how do you issue request to HTTPS terminating ALB that has header "Host: test-app.ingress.cluster.local"? Or do you mean HTTP ALB?

szuecs:
You can send different SNI and host header, but it doesn’t matter in this case. The ALB doesn’t stop you from connecting tls with an unknown name, because it’s not mTLS.
You can perfectly use `curl -k https://<up> -H”Host: foo.cluster.local”` and without Source/FromLast predicate, skipper will happily forward the traffic to your internal endpoint.

AlexanderYastrebov:
Nice, I see. Initially I overlooked SNI vs Host header, then after reading about domain fronting I thought ALB would check host header against SNI but apparently it does not. Thank you for explanation.

jonathanbeber:
[...]

Issue #1614, opened 2020-11-19
"Secure" Source predicate which can't be manipulated by the client


mikkeloscar:
**Is your feature request related to a problem? Please describe.**
Currently skipper has two predicates for allowing clients access based on IP address/range: [`Source`](https://opensource.zalando.com/skipper/reference/predicates/#Source) and [`SourceFromLast`](https://opensource.zalando.com/skipper/reference/predicates/#sourcefromlast)

Both of them rely on first checking the `X-Forwarded-For (XFF)` header and falling back to the source IP of the request if `XFF` is not specified. This means that in some setups i.e. if skipper is getting traffic directly from the client (not through a proxy) then it's not possible to protect a route with `Source` or `SourceFromLast` because the client can modify the `XFF` header and work around any IP address limitation specified.

**Describe the solution you would like**
Skipper should support another predicate which only checks for the source IP such that it can't be faked by the client. I.e. `ClientIP` or `SourceIP`. (Alternative could also be a flag to the existing predicates such that it doesn't check XFF: `Source("10.2.0.0/16", "xff-false")` - something along those lines?)

**Would you like to work on it?**
Sure

szuecs:
I would rather create a new predicate called `RemoteIP(CIDRs)` that only considers http.Request.RemoteIP, because `Source()` is already var args and the change would break all predicates compatability.

aermakov-zalando:
I think we've discussed this before when talking about migrating to NLBs, but it might be a good idea to have a command-line switch for the behaviour of the `Source` predicate as well (or instead). For example, consider our current setup: we probably have more than one ingresses/routegroups relying on the Source filter to do IP-based filtering, which works because we use ELBs in front of Skipper. If we decide to switch to NLBs, however, ideally we'd want the predicates to keep working and doing the right thing in the new setup, because the alternative is a nightmarish migration for the users (which would also prevent us from switching back and forth instead of issues).

szuecs:
@aermakov-zalando it's a different problem how to migrate an NLB and not really scope of this PR nor this project.
Maybe we can add some flag to be able to add replace f1 with f2 for example to have a generic possibility to rewrite filters at our scale.

aermakov-zalando:
I don't see why this is out of scope. My point was that it might make more sense to give the users a higher-level predicate ("filter by source IP, I don't care how exactly it's determined") instead of very low-level ones that need to be changed if something is changed in the setup.

AlexanderYastrebov:
[...]

PR #1676, opened 2021-01-06
Add EastWest Range feature

jonathanbeber:
The current east-west communication feature has some problems as the
lack of support for cluster-only routes and the fact that these internal
routes would still be accessible from outside if changing the Host
header. More details of #1526

This commit adds the concept of EastWest Range feature, what makes
possible to define internal domains and custom predicates to Skipper.
With this information, Skipper's Kubernetes dataclient adds the given
predicates on every route identified as part of one internal domain.
Differently from the `-enable-kubernetes-east-west` and the
`-kubernetes-east-west-domain=.ingress.cluster.local` flags this feature
will not automatically create routes and both features shouldn't be used
in combination.

It requires changing the current way we convert routegroups to eskip
routes, splitting the hostnames in internal and external hosts. All the
hosts identified as internals require a separated route once it will
have their own `ClientIP` predicate.

Also, it adds two new flags to skipper configuration,
`-kubernetes-east-west-range-domains` and
`-kubernetes-east-west-range-predicates`. It also include the basic
documentation regarding their usage.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -120,20 +120,23 @@ type Config struct {
 	WaitFirstRouteLoad        bool                 `yaml:"wait-first-route-load"`
 
 	// Kubernetes:
-	KubernetesIngress           bool                `yaml:"kubernetes"`
-	KubernetesInCluster         bool                `yaml:"kubernetes-in-cluster"`
-	KubernetesURL               string              `yaml:"kubernetes-url"`
-	KubernetesHealthcheck       bool                `yaml:"kubernetes-healthcheck"`
-	KubernetesHTTPSRedirect     bool                `yaml:"kubernetes-https-redirect"`
-	KubernetesHTTPSRedirectCode int                 `yaml:"kubernetes-https-redirect-code"`
-	KubernetesIngressClass      string              `yaml:"kubernetes-ingress-class"`
-	KubernetesRouteGroupClass   string              `yaml:"kubernetes-routegroup-class"`
-	WhitelistedHealthCheckCIDR  string              `yaml:"whitelisted-healthcheck-cidr"`
-	KubernetesPathModeString    string              `yaml:"kubernetes-path-mode"`
-	KubernetesPathMode          kubernetes.PathMode `yaml:"-"`
-	KubernetesNamespace         string              `yaml:"kubernetes-namespace"`
-	KubernetesEnableEastWest    bool                `yaml:"enable-kubernetes-east-west"`
-	KubernetesEastWestDomain    string              `yaml:"kubernetes-east-west-domain"`
+	KubernetesIngress                       bool                `yaml:"kubernetes"`
+	KubernetesInCluster                     bool                `yaml:"kubernetes-in-cluster"`
+	KubernetesURL                           string              `yaml:"kubernetes-url"`
+	KubernetesHealthcheck                   bool                `yaml:"kubernetes-healthcheck"`
+	KubernetesHTTPSRedirect                 bool                `yaml:"kubernetes-https-redirect"`
+	KubernetesHTTPSRedirectCode             int                 `yaml:"kubernetes-https-redirect-code"`
+	KubernetesIngressClass                  string              `yaml:"kubernetes-ingress-class"`
+	KubernetesRouteGroupClass               string              `yaml:"kubernetes-routegroup-class"`
+	WhitelistedHealthCheckCIDR              string              `yaml:"whitelisted-healthcheck-cidr"`
+	KubernetesPathModeString                string              `yaml:"kubernetes-path-mode"`
+	KubernetesPathMode                      kubernetes.PathMode `yaml:"-"`
+	KubernetesNamespace                     string              `yaml:"kubernetes-namespace"`
+	KubernetesEnableEastWest                bool                `yaml:"enable-kubernetes-east-west"`
+	KubernetesEastWestDomain                string              `yaml:"kubernetes-east-west-domain"`
+	KubernetesEastWestRangeDomains          *listFlag           `yaml:"kubernetes-east-west-range-domains"`
+	KubernetesEastWestRangePredicatesString string              `yaml:"kubernetes-east-west-range-predicates"`
+	KubernetesEastWestRangePredicates       []*eskip.Predicate  `yaml:"-"`
 
 	// Default filters
 	DefaultFiltersDir string `yaml:"default-filters-dir"`
@@ -350,19 +353,21 @@ const (
 	waitFirstRouteLoadUsage        = "prevent starting the listener before the first batch of routes were loaded"
 
 	// Kubernetes:
-	kubernetesUsage                  = "enables skipper to generate routes for ingress resources in kubernetes cluster"
-	kubernetesInClusterUsage         = "specify if skipper is running inside kubernetes cluster"
-	kubernetesURLUsage               = "kubernetes API base URL for the ingress data client; requires kubectl proxy running; omit if kubernetes-in-cluster is set to true"
-	kubernetesHealthcheckUsage       = "automatic healthcheck route for internal IPs with path /kube-system/healthz; valid only with kubernetes"
-	kubernetesHTTPSRedirectUsage     = "automatic HTTP->HTTPS redirect route; valid only with kubernetes"
-	kubernetesHTTPSRedirectCodeUsage = "overrides the default redirect code (308) when used together with -kubernetes-https-redirect"
-	kubernetesIngressClassUsage      = "ingress class regular expression used to filter ingress resources for kubernetes"
-	kubernetesRouteGroupClassUsage   = "route group class regular expression used to filter route group resources for kubernetes"
-	whitelistedHealthCheckCIDRUsage  = "sets the iprange/CIDRS to be whitelisted during healthcheck"
-	kubernetesPathModeUsage          = "controls the default interpretation of Kubernetes ingress paths: <kubernetes-ingress|path-regexp|path-prefix>"
-	kubernetesNamespaceUsage         = "watch only this namespace for ingresses"
-	kubernetesEnableEastWestUsage    = "enables east-west communication, which automatically adds routes for Ingress objects with hostname <name>.<namespace>.skipper.cluster.local"
-	kubernetesEastWestDomainUsage    = "set the east-west domain, defaults to .skipper.cluster.local"
+	kubernetesUsage                        = "enables skipper to generate routes for ingress resources in kubernetes cluster"
+	kubernetesInClusterUsage               = "specify if skipper is running inside kubernetes cluster"
+	kubernetesURLUsage                     = "kubernetes API base URL for the ingress data client; requires kubectl proxy running; omit if kubernetes-in-cluster is set to true"
+	kubernetesHealthcheckUsage             = "automatic healthcheck route for internal IPs with path /kube-system/healthz; valid only with kubernetes"
+	kubernetesHTTPSRedirectUsage           = "automatic HTTP->HTTPS redirect route; valid only with kubernetes"
+	kubernetesHTTPSRedirectCodeUsage       = "overrides the default redirect code (308) when used together with -kubernetes-https-redirect"
+	kubernetesIngressClassUsage            = "ingress class regular expression used to filter ingress resources for kubernetes"
+	kubernetesRouteGroupClassUsage         = "route group class regular expression used to filter route group resources for kubernetes"
+	whitelistedHealthCheckCIDRUsage        = "sets the iprange/CIDRS to be whitelisted during healthcheck"
+	kubernetesPathModeUsage                = "controls the default interpretation of Kubernetes ingress paths: <kubernetes-ingress|path-regexp|path-prefix>"
+	kubernetesNamespaceUsage               = "watch only this namespace for ingresses"
+	kubernetesEnableEastWestUsage          = "enables east-west communication, which automatically adds routes for Ingress objects with hostname <name>.<namespace>.skipper.cluster.local"
+	kubernetesEastWestDomainUsage          = "set the east-west domain, defaults to .skipper.cluster.local"
+	kubernetesEastWestRangeDomainsUsage    = "set the the cluster internal domains for east west traffic. Identified routes to such domains will include the -kubernetes-east-west-range-predicates"
+	kubernetesEastWestRangePredicatesUsage = "set the predicates that will be appended to routes identified as to -kubernetes-east-west-range-domains"
 
 	// Auth:
 	oauth2GrantFlowEnableUsage           = "enables OAuth2 Grant Flow filter"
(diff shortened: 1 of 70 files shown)


config/config_test.go
```diff
@@ -64,6 +64,8 @@ func Test_NewConfig(t *testing.T) {
 				AppendFilters:                           &defaultFiltersFlags{},
 				PrependFilters:                          &defaultFiltersFlags{},
 				SourcePollTimeout:                       3000,
+				KubernetesEastWestRangeString:           commaListFlag(),
+				KubernetesEastWestRange:                 map[string]string{},
```
szuecs, line 68: Should be nil instead, if it doesn’t work, needs fix in code

dataclients/kubernetes/eastwest.go
```diff
@@ -43,3 +46,52 @@ func createEastWestRouteRG(name, ns, postfix string, r *eskip.Route) *eskip.Rout
 	ewr.Predicates = p
 	return ewr
 }
+
+func applyEastWestRange(eWRange map[string]string, host string, routes []*eskip.Route) {
+	for d, n := range eWRange {
```
szuecs, line 51: d is domain, n is cidr?
d seems fine, n is weird.

dataclients/kubernetes/eastwest.go
```diff
@@ -43,3 +46,52 @@ func createEastWestRouteRG(name, ns, postfix string, r *eskip.Route) *eskip.Rout
...
+		if strings.HasSuffix(host, d) {
+			applyEastWestRangePredicates(routes, n)
+		}
+	}
+}
+
+func applyEastWestRangePredicates(routes []*eskip.Route, n string) {
+	for _, route := range routes {
+		route.Predicates = append(route.Predicates, &eskip.Predicate{
+			Name: "SourceFromLast",
```
szuecs, line 61: That’s the wrong predicate. We only want to allow https://opensource.zalando.com/skipper/reference/predicates/#clientip without xff headers. Within cluster doesn’t pass the cloud loadbalancer

dataclients/kubernetes/eastwest.go
```diff
@@ -43,3 +46,52 @@ func createEastWestRouteRG(name, ns, postfix string, r *eskip.Route) *eskip.Rout
...
+			Args: []interface{}{n},
+		})
+	}
+}
+
+// ParseEastWestRange parses the multiple strings representations
+// of the internal domains and their CIDRs.
+func ParseEastWestRange(v []string) (map[string]string, error) {
+	eastWestRange := map[string]string{}
+	for _, i := range v {
```
szuecs, line 71: i seems to be a bad name. Maybe cidr, but i is normally the index var not a value which is not an int

dataclients/kubernetes/eastwest.go
```diff
@@ -43,3 +46,52 @@ func createEastWestRouteRG(name, ns, postfix string, r *eskip.Route) *eskip.Rout
...
+		}
+		if !strings.Contains(i, eastWestRangeDelimiter) {
+			return nil, fmt.Errorf("invalid east-west-range config: %s", i)
+		}
+		s := strings.Split(i, eastWestRangeDelimiter)
+		if len(s) != 2 {
+			return nil, fmt.Errorf("invalid east-west-range config: %s", i)
+		}
+		d, n := s[0], s[1]
+		_, net, err := net.ParseCIDR(n)
```
szuecs, line 84: net shadows lib, cidr would be better.
d and n are as I commented above.

dataclients/kubernetes/eastwest_test.go
```diff
@@ -0,0 +1,180 @@
...
+				"another.cluster.local": "10.7.0.0/16",
+			},
+		},
+	} {
+		t.Run(s.name, func(t *testing.T) {
+			r, e := ParseEastWestRange(s.values)
+			if e != nil {
+				if s.err != nil {
+					if e.Error() != s.err.Error() {
+						t.Error("invalid error received while parsing east-west range")
```
szuecs, line 156: t.Errorf instead logs please

szuecs: Same below multiple times

dataclients/kubernetes/routegroup.go
```diff
@@ -119,7 +120,10 @@ func toSymbol(p string) string {
 	return string(b)
 }
 
-func rgRouteID(namespace, name, subName string, index, subIndex int) string {
+func rgRouteID(namespace, name, subName string, index, subIndex int, internal bool) string {
+	if internal {
+		name += "_internal"
```
szuecs, line 125: Better prepend namespace

szuecs (changes requested):
Some nitpicks, but important ones.
<s>I think the tests should not all have the feature set, only the ones we add. </s>
The mergeHosts func deletion I would need to review more to judge good/bad, but as long tests pass we should be fine.
The configuration default should use nil instead of empty map.

dataclients/kubernetes/eastwest.go
jonathanbeber: should be network

jonathanbeber: will change.

dataclients/kubernetes/eastwest.go
jonathanbeber: good one. I will update the narrative too.

dataclients/kubernetes/eastwest.go
jonathanbeber: I was operating with the index in the very beggining and forgot to update. Thanks!

dataclients/kubernetes/eastwest.go
jonathanbeber: for the record, I'm not sure if by 
> without xff headers

you were referring to check whether the header is present, but in any case, that's not necessary since faking the `Remote-Addr` would result in a failed syn-ack.

dataclients/kubernetes/routegroup.go
jonathanbeber: thanks, I was not sure about that.

jonathanbeber:
Thank you very much, @szuecs. I addressed the points but two:

> I think the tests should not all have the feature set, only the ones we add.

I tried to add tests for the scenarios the new code impacts, also the ones I believed should be checked with this new scenario of multiple internal and multiple external domains. I tried to base myself on the current east-west tests cases too.
Also, I believe I probably created scenarios that are not completed required. Maybe you could, please, point me the ones you believe are not required and we could discuss them? WDYT?

> The mergeHosts func deletion I would need to review more to judge good/bad, but as long tests pass we should be fine.

It was used only during route groups conversion for the catch-all routes that were created later. Since now it creates the catch-all 
right after the `transformRouteGroup` I used the `ctx.hostRoutes` directly instead of merging it. As you said, tests are passing and we should be fine.

Regarding the wrong predicate, I left a comment on the specific line for a clarification of your comment.

dataclients/kubernetes/eastwest.go
mikkeloscar: @jonathanbeber I think sandor is referring to the fact that if you have an NLB (not ALB) in front of skipper then you can fake the xff header if `SourceFromLast` is used. This is why we created `ClientIP` which was done quite recently #1614 :)

mikkeloscar: That's also the reason `SourceFromLast` was mentioned in the internal narrative, because we didn't have anything else at the time and didn't yet consider NLB as an option.

jonathanbeber: Thanks, @mikkeloscar, then, indeed, using just `ClientIP` is considered enough, I'm marking it as resolved.

jonathanbeber: reopening it as after some test we found out that just `ClientIP` won't work for us: https://github.com/zalando/skipper/issues/1526#issuecomment-756903746

jonathanbeber: this means we will also have to support multiple CIDRs per domain (or at least two).

szuecs: Multiple cidr makes sense for migrations, too. Think about migration of subnetwork we already did. to increase the network size for clusters.

szuecs:
Adding tests are fine, but there are also modified tests (results modified), which means we don’t test the scenario without the eastwest range feature. So I would say duplicate some tests and add the eastwest range tests.

dataclients/kubernetes/eastwest.go
jonathanbeber: it's done now, it accepts predicates instead of CIDRs, what gives us a lot more flexibility.

config/config.go
```diff
@@ -688,13 +696,19 @@ func (c *Config) Parse() error {
 		return err
 	}
 
+	kubernetesEastWestRangePredicates, err := eskip.ParsePredicates(c.KubernetesEastWestRangePredicatesString)
+	if err != nil {
+		return fmt.Errorf("invalid east-west-range-predicates: %s", err.Error())
```
szuecs, line 701: Better use %w if you create a wrapped error:
```
fmt.Errorf("invalid east-west-range-predicates: %w", err)
```

szuecs (approved):

jonathanbeber:
:+1:

mikkeloscar (approved):

mikkeloscar:
:+1:

dataclients/kubernetes/ingress.go
```diff
@@ -47,6 +47,8 @@ type ingress struct {
 	pathMode                    PathMode
 	kubernetesEnableEastWest    bool
 	eastWestDomainRegexpPostfix string
+	eastWestRangeDomains        []string
+	EastWestRangePredicates     []*eskip.Predicate
```
szuecs, line 51: do we need to export this member?

dataclients/kubernetes/routegroup.go
```diff
@@ -513,49 +518,120 @@ func transformRouteGroup(ctx *routeGroupContext) ([]*eskip.Route, error) {
 	return explicitGroupRoutes(ctx)
 }
 
+func splitHosts(hosts []string, domains []string) (map[string]string, []string) {
+	i := map[string]string{}
```
szuecs, line 522: i is more an index var, so I would suggest `h` instead

dataclients/kubernetes/routegroup.go
```diff
@@ -513,49 +518,120 @@ func transformRouteGroup(ctx *routeGroupContext) ([]*eskip.Route, error) {
...
+	e := []string{}
```
szuecs, line 523: not sure about `e`, maybe `a` ?

dataclients/kubernetes/ingress.go
jonathanbeber: good catch

dataclients/kubernetes/routegroup.go
jonathanbeber: it's it from internal and e for external. Will fix it.

dataclients/kubernetes/routegroup.go
jonathanbeber: it's for internal and shouldn't be a map since we changed to receive the predicates. Will fix it.

jonathanbeber:
:+1:

docs/tutorials/operations.md
```diff
@@ -108,6 +108,73 @@ domain
 [known bug](https://github.com/zalando/skipper/issues/1024), if you
 combine it with custom routes.
 
+Alternatively, you can use Kubernetes East West Range feature. Use the
+flag `-kubernetes-east-west-range-domains` to define the cluster
+internal domains `-kubernetes-east-west-range-predicates` to define the
+[predicates](../../reference/predicates) that will be appended to every
```
szuecs, line 114: ../../reference/predicates -> ../../reference/predicates.md

docs/tutorials/operations.md
```diff
@@ -108,6 +108,73 @@ domain
 [known bug](https://github.com/zalando/skipper/issues/1024), if you
 combine it with custom routes.
 
+Alternatively, you can use Kubernetes East West Range feature. Use the
```
szuecs, line 111: Maybe we should add a subsection for all the "more features"

jonathanbeber: great idea, done!

szuecs:
:+1:

jonathanbeber:
:+1:

dataclients/kubernetes/routegroup.go
```diff
@@ -513,49 +518,120 @@ func transformRouteGroup(ctx *routeGroupContext) ([]*eskip.Route, error) {
...
+		// create it.
+		if len(externalHosts) != 0 || len(hosts) == 0 {
+			ctx := &routeGroupContext{
+				clusterState:          s,
+				defaultFilters:        df,
+				routeGroup:            rg,
+				hosts:                 externalHosts,
+				hostRx:                createHostRx(externalHosts...),
+				hostRoutes:            make(map[string][]*eskip.Route),
+				hasEastWestHost:       hasEastWestHost(r.options.KubernetesEastWestDomain, externalHosts),
```
AlexanderYastrebov, line 564: I think this `hasEastWestHost` uses deprecated option `KubernetesEastWestDomain` 
https://github.com/zalando/skipper/blob/62f282f0f25a7aa50635e2b1f6956c6000cfa35b/dataclients/kubernetes/kube.go#L152-L154
and maybe not needed here at all  (i.e. always false) as hosts were split (`splitHosts`) before and we know `externalHosts` do not contains east-west hosts...

dataclients/kubernetes/routegroup.go
```diff
@@ -513,49 +518,120 @@ func transformRouteGroup(ctx *routeGroupContext) ([]*eskip.Route, error) {
...
+			ri = append(ri, catchAll...)
+
+			rs = append(rs, ri...)
 		}
 
-		rs = append(rs, ri...)
-		mergeHostRoutes(hostRoutes, ctx.hostRoutes)
-	}
+		// Internal hosts
+		for _, host := range internalHosts {
```
AlexanderYastrebov, line 596: What is the reason here to loop over internal domains where for external domains we just combine them
```go
hostRx:                createHostRx(externalHosts...),
```
@jonathanbeber 
Why not to do the same here, i.e. get rid of the loop and use single `internalCtx`
```go
hostRx:                createHostRx(internalHosts...),
```

jonathanbeber: You're right, that was due to an [initial implementation](https://github.com/zalando/skipper/commit/b72f8d47cbda834395913171da2643dd466d0631#diff-3a9a80b20ebd6d5f62df3e5a8a080e2e022bb1ab9ec7a2089021f236f0f85b14R630) where we could define multiple IPs for different domains. The idea is that you could have `domain.internal` to the CIDR `10.0.0/8`, and `domain2.internal`  to the CIDR `192,168,0.0/24` for example. 

I will fix it.

dataclients/kubernetes/routegroup.go
jonathanbeber: As far as i understand we just deprecated but didn't remove it yet.

jonathanbeber: Also, I deprecated it after merging this PR, so we might want to revisit it soon.

dataclients/kubernetes/routegroup.go
AlexanderYastrebov: Thank you

jonathanbeber: #1758 should handle that.

PR #1671, opened 2020-12-29
Refactors flaky ratelimit tests

AlexanderYastrebov:
Signed-off-by: Alexander Yastrebov <alexander.yastrebov@zalando.de>

PR #1678, opened 2021-01-11
Fixes flaky ratelimit tests

AlexanderYastrebov:
Fixes three tests with the following structure:
```
0. setup ratelimit: N requests/time window
1. do N requests and expect 200 OK
2. do N requests and expect 429 Too Many Requests
3. sleep for time window
4. do N requests and expect 200 OK
```

Such test passes only if steps 1 and 2 are complete within time window.
If requests are slow (e.g. CI pipeline) then time window may pass before
steps are complete, limit expires and expectaions fail on step 2
(`expected 429, got 200`).

This change reduces N thus allowing lower request rate.

Followup on #1671


szuecs:
:+1:

szuecs:
@AlexanderYastrebov thanks so much!

aryszka:
:+1:

PR #1692, opened 2021-01-18
Deprecate kubernetes-enable-east-west feature

jonathanbeber:
With #1676 our approach to East-West traffic relies on Kubernetes
objects, Ingresses and RouteGroups, explicitly defining internal
hostnames instead of Skipper adding it automatically. This way, each
application can decide whether it accepts just internal traffic.

This commit deprecates the previous feature handling East-West traffic,
marking it as deprecated on the CLI help, raising a warning message when
used and updating the documentation. On the documentation, it marks the
old feature as deprecated with an `Attention` note, as well as, updates
any reference and tutorials to East-West traffic, suggesting the usage
of the East-West Range feature.


szuecs:
:+1:

aryszka:
:+1:

PR #1697, opened 2021-01-25
Opentracing: set http.status_code correctly

aermakov-zalando:
`Proxy.errorResponse()` sets `http.status_code` on the OpenTracing span _before_ it actually computes the correct value. This leads to very confusing traces where clients get a 404, but Skipper reports a 500 instead. Fix by computing the response status first.


szuecs:
:+1:

aermakov-zalando:
👍

aryszka:
:+1:

szuecs:
:+1:

PR #1700, opened 2021-01-26
Document default filters with regards to grantCallback route

k-nut:
After the oauth grant flow related filters where merged I realised that I could clean up our routes file and remove a manually added `grantCallback` route since it gets automatically added by Skipper. I then realised that this automatically added route does not apply the default filters. I think that is it relatively rare that users would want to apply default filters here but it should still be pointed out in the documentation.


szuecs:
:+1:

aryszka:
:+1:

PR #1703, opened 2021-01-26
Password support for Redis

### LINKED ISSUE #1702: Password support for Redis server

`Skipper` has `Redis` as a persistence option to enforce cluster rate limits. Current implementation doesn't support redis servers that use password.  

A new setting will be introduced `-swarm-redis-password` which will set the password for `Redis` ring.

Yes

c0nstantx:
This PR closes #1702

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"sort"
 	"strconv"
 	"strings"
@@ -207,7 +208,8 @@ type Config struct {
 	// swarm:
 	EnableSwarm bool `yaml:"enable-swarm"`
 	// redis based
-	SwarmRedisURLs         *listFlag     `yaml:"swarm-redis-urls"`
+	SwarmRedisURLs         *listFlag `yaml:"swarm-redis-urls"`
+	SwarmRedisPassword     string
 	SwarmRedisDialTimeout  time.Duration `yaml:"swarm-redis-dial-timeout"`
 	SwarmRedisReadTimeout  time.Duration `yaml:"swarm-redis-read-timeout"`
 	SwarmRedisWriteTimeout time.Duration `yaml:"swarm-redis-write-timeout"`
@@ -735,6 +737,7 @@ func (c *Config) Parse() error {
 		c.Certificates = certificates
 	}
 
+	c.parseEnv()
 	return nil
 }
 
@@ -913,6 +916,7 @@ func (c *Config) ToOptions() skipper.Options {
 		EnableSwarm: c.EnableSwarm,
 		// redis based
 		SwarmRedisURLs:         c.SwarmRedisURLs.values,
+		SwarmRedisPassword:     c.SwarmRedisPassword,
 		SwarmRedisDialTimeout:  c.SwarmRedisDialTimeout,
 		SwarmRedisReadTimeout:  c.SwarmRedisReadTimeout,
 		SwarmRedisWriteTimeout: c.SwarmRedisWriteTimeout,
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -996,3 +1002,11 @@ func (c *Config) parseHistogramBuckets() ([]float64, error) {
 	sort.Float64s(result)
 	return result, nil
 }
+
+func (c *Config) envOverrides() {
+	// Redis password
+	swarmRedisPassword := os.Getenv("SWARM_REDIS_PASSWORD")
```
AlexanderYastrebov, line 1008: I think the idea https://github.com/zalando/skipper/issues/1702#issuecomment-769816602 was to have it the other way around: if flag is not set use what is in env variable, i.e. explicit flag overrides environment.

My personal opinion is that we don't need the flag at all. Having it leaves a chance for some users to use it in order to later discover that it leaked and have to rotate passwords and asses the damage.

Different tools use different approaches, e.g. `redis-cli` allows flag while `psql` does not.

But maybe I just do not see the usecase for the flag.

c0nstantx: > I think the idea [#1702 (comment)](https://github.com/zalando/skipper/issues/1702#issuecomment-769816602) was to have it the other way around: if flag is not set use what is in env variable, i.e. explicit flag overrides environment.

You are right. I red it a bit hasty and got it the other way around.

> Different tools use different approaches, e.g. `redis-cli` allows flag while `psql` does not.

Usually `cli` clients have two options for password. Pass it interactively or as an option (I see [psql](https://www.postgresql.org/docs/current/app-psql.html) does the same), but never as an environment variable.

On the other hand, servers usually set password from a configuration file with some exceptions that also use `cli` flags like `redis-server --requirepass 1234` but the specific one seems that hides it from `ps`

So, if you ask me, since `skipper` is a server and not a client, I would go with the second approach. The issue with `skipper` though is that its configuration options are also `cli` flags.

Bottom line is that I would prefer to have it  **ONLY** in configuration file and not as flag nor environment variable.
That gives you the ability to run multiple instances of your server (I cannot find a use case for that) in the same environment with different passwords.

AlexanderYastrebov:
I think we should support env variable and not introduce the flag (unless there is a compelling usecase) to avoid its misuse. It is easier to introduce later then to remove it.

AlexanderYastrebov:
@c0nstantx Thank you for updating the PR. I think we still may support password from config file (i.e. to have env, config file but not the flag). You may cherry-pick https://github.com/zalando/skipper/commit/fbc0d7a583a080d6df3b5eb5de57b2a35ee23ca7 or I can make a followup PR once we land this one.

c0nstantx:
No problem. I will update it

config/config.go
```diff
@@ -208,6 +209,7 @@ type Config struct {
 	EnableSwarm bool `yaml:"enable-swarm"`
 	// redis based
 	SwarmRedisURLs         *listFlag     `yaml:"swarm-redis-urls"`
```
AlexanderYastrebov, line 211: SwarmRedisURLs usage string should mention env variable and config option like here https://github.com/zalando/skipper/commit/fbc0d7a583a080d6df3b5eb5de57b2a35ee23ca7#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R449

c0nstantx: Cool. Thanks

config/config.go
```diff
@@ -996,3 +1003,10 @@ func (c *Config) parseHistogramBuckets() ([]float64, error) {
 	sort.Float64s(result)
 	return result, nil
 }
+
+func (c *Config) parseEnv() {
+	// Set Redis password from environment variable if not set earlier (configuration file)
+	if c.SwarmRedisPassword == "" {
+		c.SwarmRedisPassword = os.Getenv("SWARM_REDIS_PASSWORD")
```
AlexanderYastrebov, line 1010: `redisPasswordEnv`

c0nstantx: This is what happens when you don't pay attention and think you can multitask :sweat_smile:

AlexanderYastrebov:
lgtm, we need maintainers to have a look

szuecs:
👍

aryszka:
:+1:

PR #1704, opened 2021-01-26
feature: dynamic ratelimit POC

szuecs:
RouteGroup:
```yaml
spec:
  backends:
  - name: my-backend
    serviceName: loadtest-backend
    servicePort: 80
    type: service
  defaultBackends:
  - backendName: my-backend
  hosts:
  - ratelimit-test.example.org
  routes:
  - filters:
    - clusterClientRatelimit("test", 10, "10s", "X-Test")
```

Backend:
```
% kubectl get deployments.apps loadtest-backend -o yaml | grep args: -A 11
      - args:
        - skipper
        - -serve-host-metrics
        - -disable-metrics-compat
        - -metrics-exp-decay-sample
        - -metrics-flavour=prometheus
        - -default-filters-prepend=enableAccessLog(4,5)
        - -inline-routes
        - 'h: Path("/healthz") -> inlineContent("OK") -> status(200) -> <shunt>; r:
          * -> setResponseHeader("X-Rate", "5 1m") -> latency(5) -> inlineContent("loadtest")
          -> <shunt>;'
```

test:
```
% cat test.sh
#!/bin/bash
for i in {01..10}
do
        curl -v https://ratelimit-test.example.org/ -H"X-Test: foo"
done

sleep 10
echo "################ 10s sleep #########################";

for i in {01..10}
do
        curl -v https://ratelimit-test.example.org/ -H"X-Test: foo"
done

% ./test.sh G -E '#####|HTTP/1.1' 
> GET / HTTP/1.1
< HTTP/1.1 200 OK
> GET / HTTP/1.1
< HTTP/1.1 200 OK
> GET / HTTP/1.1
< HTTP/1.1 200 OK
> GET / HTTP/1.1
< HTTP/1.1 200 OK
> GET / HTTP/1.1
< HTTP/1.1 200 OK
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests

################ 10s sleep #########################
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests
> GET / HTTP/1.1
< HTTP/1.1 429 Too Many Requests
```

### DIFF
--- a/filters/ratelimit/ratelimit.go
+++ b/filters/ratelimit/ratelimit.go
@@ -8,7 +8,9 @@ package ratelimit
 import (
 	"context"
 	"net/http"
+	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -23,8 +25,10 @@ type spec struct {
 }
 
 type filter struct {
-	settings ratelimit.Settings
-	provider RatelimitProvider
+	mu         sync.Mutex
+	settings   ratelimit.Settings
+	provider   RatelimitProvider
+	overwrites map[string]ratelimit.Settings
 }
 
 // RatelimitProvider returns a limit instance for provided Settings
@@ -259,7 +263,10 @@ func clusterClientRatelimitFilter(args []interface{}) (*filter, error) {
 		s.Lookuper = ratelimit.NewXForwardedForLookuper()
 	}
 
-	return &filter{settings: s}, nil
+	return &filter{
+		settings:   s,
+		overwrites: make(map[string]ratelimit.Settings),
+	}, nil
 }
 
 func getLookuper(s string) ratelimit.Lookuper {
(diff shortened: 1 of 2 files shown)


filters/ratelimit/ratelimit.go
```diff
@@ -23,8 +25,10 @@ type spec struct {
 }
 
 type filter struct {
-	settings ratelimit.Settings
-	provider RatelimitProvider
+	mu         sync.Mutex
+	settings   ratelimit.Settings
+	provider   RatelimitProvider
+	overwrites map[string]ratelimit.Settings
```
AlexanderYastrebov, line 31: This would be reset on route table update, wouldn't it?

AlexanderYastrebov: It is also instance local

szuecs: The POC assumes that the backend is constantly telling in the response path the "current" changed settings, so yes and yes, but it's just a POC. The question would be if this is a reasonable assumption that the backend responses with the decision like this or if the backend just response once with a change.

Issue #1705, opened 2021-01-26
Document clusterRatelimit memory requirements


AlexanderYastrebov:
**Is your feature request related to a problem? Please describe.**
The problem is how to configure Redis cluster used by [cluster ratelimit](https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#clusterclientratelimit) filters.

**Describe the solution you would like**
A formula to estimate Redis cluster size in terms of memory requirement and number of shards based on ratelimit configurations, number of skipper instances and total traffic.

**Would you like to work on it?**
Yes

Related to #1701

PR #1706, opened 2021-01-27
Reduces ratelimit key length and makes it fixed

AlexanderYastrebov:
* uses 128 bit hash function
* uses minimal prefix, hashes both group and text
* uses base64 encoding

An example key before and after:
```
ratelimit.hello.486ea46224d1bb4fb680f34f7c9ad96a8f24ec88be73ea8e5a6c65260e9cb8a7
R_F4DjTilcDIIVEHn_nAQsA
```
Key size reduced by 57 bytes.

Example of redis key with one hit stored before and after:
```
127.0.0.1:6379> zcard
ratelimit.hello.41a19288ea2042e9885dda76140d225b931ec56a7fafb41d932471995d7253db
(integer) 1
127.0.0.1:6379> memory usage
ratelimit.hello.41a19288ea2042e9885dda76140d225b931ec56a7fafb41d932471995d7253db
(integer) 169
127.0.0.1:6379> zcard Rwa0Jj-UpetSydKP8JahsNA
(integer) 1
127.0.0.1:6379> memory usage Rwa0Jj-UpetSydKP8JahsNA
(integer) 110
```
Also group names could be lengthy in a setup with mutiple ratelimits configured.

### DIFF
--- a/ratelimit/ratelimit.go
+++ b/ratelimit/ratelimit.go
@@ -3,8 +3,8 @@ package ratelimit
 import (
 	"bytes"
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
+	"crypto/md5" // #nosec
+	"encoding/base64"
 	"fmt"
 	"net/http"
 	"strconv"
@@ -456,7 +456,19 @@ func Headers(s *Settings, retryAfter int) http.Header {
 	}
 }
 
-func getHashedKey(clearText string) string {
-	h := sha256.Sum256([]byte(clearText))
-	return hex.EncodeToString(h[:])
+func getRatelimitKey(group, text string) string {
+	return buildKey("R", group, text)
+}
+
+func buildKey(prefix, group, text string) string {
+	h := md5.New() // #nosec
+	h.Write([]byte(group))
+	h.Write([]byte(text))
+	s := h.Sum(nil)
+
+	b := make([]byte, len(prefix)+base64.RawURLEncoding.EncodedLen(md5.Size))
+	copy(b, prefix)
+	base64.RawURLEncoding.Encode(b[len(prefix):], s)
+
+	return string(b)
 }
--- a/ratelimit/cluster.go
+++ b/ratelimit/cluster.go
@@ -1,10 +1,5 @@
 package ratelimit
 
-const (
-	swarmPrefix    = `ratelimit.`
-	swarmKeyFormat = swarmPrefix + "%s.%s"
-)
-
 // newClusterRateLimiter will return a limiter instance, that has a
 // cluster wide knowledge of ongoing requests. Settings are the normal
 // ratelimit settings, Swarmer is an instance satisfying the Swarmer
--- a/ratelimit/ratelimit_test.go
+++ b/ratelimit/ratelimit_test.go
@@ -335,3 +335,13 @@ func TestSettingsRatelimit(t *testing.T) {
 		}
 	})
 }
+
+func TestGetRatelimitKey(t *testing.T) {
+	// echo -n helloworld | md5sum | cut -d' ' -f1 | xxd -r -p | base64 | tr -d '=' | tr '+/' '-_'
+	const expected = "R" + "_F4DjTilcDIIVEHn_nAQsA"
+
+	key := getRatelimitKey("hello", "world")
+	if key != expected {
+		t.Errorf("expected %s, got %s", expected, key)
+	}
+}
--- a/ratelimit/redis.go
+++ b/ratelimit/redis.go
@@ -147,10 +147,6 @@ func newClusterRateLimiterRedis(s Settings, r *ring, group string) *clusterLimit
 	return rl
 }
 
-func (c *clusterLimitRedis) prefixKey(clearText string) string {
-	return fmt.Sprintf(swarmKeyFormat, c.group, clearText)
-}
-
 func (c *clusterLimitRedis) measureQuery(format, groupFormat string, fail *bool, start time.Time) {
 	result := "success"
 	if fail != nil && *fail {
@@ -208,9 +204,8 @@ func (c *clusterLimitRedis) startSpan(ctx context.Context, spanName string) func
 //
 // If a context is provided, it uses it for creating an OpenTracing span.
 func (c *clusterLimitRedis) AllowContext(ctx context.Context, clearText string) bool {
-	s := getHashedKey(clearText)
+	key := getRatelimitKey(c.group, clearText)
 	c.metrics.IncCounter(redisMetricsPrefix + "total")
-	key := c.prefixKey(s)
 
 	now := time.Now()
 	var queryFailure bool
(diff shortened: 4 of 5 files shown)


AlexanderYastrebov:
The group name is not encoded into key name anymore (e.g.#1600 used it to measure group key stats)

ratelimit/ratelimit.go
```diff
@@ -456,7 +457,13 @@ func Headers(s *Settings, retryAfter int) http.Header {
 	}
 }
 
-func getHashedKey(clearText string) string {
-	h := sha256.Sum256([]byte(clearText))
-	return hex.EncodeToString(h[:])
+func getKey(group, text string) string {
+	return getKeyImpl(group, text, md5.New()) // #nosec
```
AlexanderYastrebov, line 461: It should have minimal prefix in order to put ratelimit keys into separate namespace as there may be other key namespaces in the future.

szuecs:
Prefix is a have to such that we can understand data if we need to.
I am not sure if reducing security is great in this case. If you use authorization header to identify a client, then a malicious user might be able to find passwords more easily. Of course we have no salt, which would be even better in case of leak.

AlexanderYastrebov:
I have added a prefix.

> I am not sure if reducing security is great in this case. If you use authorization header to identify a client, then a malicious user might be able to find passwords more easily. Of course we have no salt, which would be even better in case of leak.

I changed hash function from 256 bit to 128 to reduce key length. The "security" of the keys has increased since now both group name and plain text are hashed so reverse hash search (e.g. rainbow tables) has less chance matching (e.g. dictionary keys become non-dictionary due to group prefix). 

For high cardinality keys (like JWT tokens, random uuids) I doubt it is practically much easier to crack md5 then sha256.
In fact md5 calculation is not much faster than sha256 (<2x on my machine):
```
$ openssl speed md5 sha256
...
The 'numbers' are in 1000s of bytes per second processed.
type             16 bytes     64 bytes    256 bytes   1024 bytes   8192 bytes  16384 bytes
md5             130430.01k   305034.47k   539497.13k   632668.64k   691205.46k   697117.35k
sha256           74478.77k   170761.37k   338456.58k   428023.81k   402284.54k   435901.78k
```
It also very much depends on the hardware, moreover there is more hardware available to bruteforce sha256 due to Bitcoin raise in recent years ([link](https://security.stackexchange.com/questions/95696/which-hash-algorithm-takes-longer-time-if-we-compare-between-md5-or-sha256)).

For "low" carnality keys (like number strings, ids) it may be possible to check whole keyspace, regardless of hash function or if there is salt or not.

AlexanderYastrebov:
I have added a prefix.

> I am not sure if reducing security is great in this case. If you use authorization header to identify a client, then a malicious user might be able to find passwords more easily. Of course we have no salt, which would be even better in case of leak.

I changed hash function from 256 bit to 128 to reduce key length. The "security" of the keys has increased since now both group name and plain text are hashed so reverse hash search (e.g. rainbow tables) has less chance matching (e.g. dictionary keys become non-dictionary due to group prefix). 

For high cardinality keys (like JWT tokens, random uuids) I doubt it is practically much easier to crack md5 then sha256.
In fact md5 calculation is not much faster than sha256 (<2x on my machine):
```
$ openssl speed md5 sha256
...
The 'numbers' are in 1000s of bytes per second processed.
type             16 bytes     64 bytes    256 bytes   1024 bytes   8192 bytes  16384 bytes
md5             130430.01k   305034.47k   539497.13k   632668.64k   691205.46k   697117.35k
sha256           74478.77k   170761.37k   338456.58k   428023.81k   402284.54k   435901.78k
```
It also very much depends on the hardware, moreover there is more hardware available to bruteforce sha256 due to Bitcoin raise in recent years ([link](https://security.stackexchange.com/questions/95696/which-hash-algorithm-takes-longer-time-if-we-compare-between-md5-or-sha256)).

For "low" carnality keys (like number strings, ids) it may be possible to check whole keyspace, regardless of hash function or if there is salt or not.

szuecs:
If we hash the group we can not check the group in the DBs. It's a tradeoff and I am not sure if the reduction of number of bytes makes sense in our case. I feel that we do here an early optimization that we don't need and I would like to not have everywhere /* nosec */ comments and GH checkers will also find this sooner or later....

AlexanderYastrebov:
> If we hash the group we can not check the group in the DBs. 

Do we have such need actually (besides #1600)?

> It's a tradeoff and I am not sure if the reduction of number of bytes makes sense in our case.

The reason for hashing group name is "group names could be lengthy in a setup with mutiple ratelimits configured".
In the scenario with many high cardinality keys (e.g. ip address, anonymous session ids), a few requests per key and per-route ratelimits (means each route needs its own group name) the key name size would dominate the total key size (name + hits).

> I feel that we do here an early optimization that we don't need

It may be so, I prefer to have a code to support the discussion hence this PR

> I would like to not have everywhere /* nosec */ comments

I dont like this markers either but the tool we use requires this because it does not understand the usage context of md5 here. The comment does not make md5 hash itself better or worse.

ratelimit/ratelimit.go
```diff
@@ -3,8 +3,8 @@ package ratelimit
 import (
 	"bytes"
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
+	"crypto/md5" // #nosec
```
aryszka, line 6: since we're here, do we even need a crypto level hash for this? Some time ago, I found a small analysis for another use case, and to me it seemed that FNV is quite OK for keys, when security is not in concern. (Cannot find the exact reference, but there are similar results published.)

szuecs: If we want to secure the data, because we likely store authorization headers in redis I am more in favor of a strong secure hash.

szuecs:
@AlexanderYastrebov we can check with pprof if we see high usage object creations in this code path, if we want to do the decision data driven. I think it’s not a problem right now and we did not invest too much into setting up a great load test, which we might want to do if we want to do optimizations in this code path.

ratelimit/ratelimit.go
aryszka: I think this is not the reason why crypto hashes need to be strong. We never compare these particular hashes in any real security scenario.

aryszka: ok, I think I see the reason

AlexanderYastrebov:
> we can check with pprof if we see high usage object creations in this code path, if we want to do the decision data driven.

I think I've failed to explain the objective of this PR :)
The goal is to make ratelimit key fixed so it is easier to reason about Redis cluster size and memory requirement (#1705)
At the moment the key includes group name which is provided by users and may be quite long. 

The memory requirement per key is*  `key size + c * N`, where `c` is some constant defined by `ZSET` implementation and `N` is a number of hits we store in `ZSET` for a configured time window. For high cardinality keys (say IP address or authorization header) we may end up storing a lot of keys with a few hits and the total key-value size would be dominated by key size, not value-size.

`*` theoretically, in [practice](https://redis.io/topics/memory-optimization) there are optimizations, fragmentation and much more


> since we're here, do we even need a crypto level hash for this? 

I've used cryptographic hash function that is known to provide a good distribution of the hashes and has a good [avalanche effect](https://en.wikipedia.org/wiki/Avalanche_effect). It may be that FNV would do, but for ratelimts we have to deal with large amount of keys so I didn't want to gamble key distribution (e.g. FNV is fine for consistent hash algorithm as we have only a few backends). I've picked md5 to get 128 bit key, see also some thoughts here https://github.com/zalando/skipper/pull/1706#issuecomment-770130361 regarding security.

szuecs:
@AlexanderYastrebov The key size is variable in the group key, but that's it. I think we can reasonably calculate with a fixed number for the group key and document that we used N for the fixed group size.

AlexanderYastrebov:
Alright, lets keep it in the sleeve for now.

PR #1707, opened 2021-01-29
Use file path on url to forward fastcgi requests

fbrousse:
The issue this PR tries to fix is the case where we want to forward the lookup path the the fastcgi backend to serve different files and not only index.php.

Last commit adds the return code 404 when file is not existing on backend :

```
::1 - - [03/Feb/2021:12:19:35 +0100] "HEAD /test5.php?name=foo HTTP/1.1" 404 10 "-" "curl/7.63.0" 8 localhost - -
::1 - - [03/Feb/2021:12:19:50 +0100] "GET /test2.php?name=foo HTTP/1.1" 200 28 "-" "curl/7.63.0" 7 localhost - -
```

Tested locally with docker-compose

### DIFF
--- a/proxy/fastcgi/fastcgi.go
+++ b/proxy/fastcgi/fastcgi.go
@@ -3,8 +3,10 @@ package fastcgi
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 
 	"github.com/yookoala/gofast"
 	"github.com/zalando/skipper/logging"
@@ -70,7 +72,19 @@ func (rt *RoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	resp.WriteTo(rr, errBuffer)
 
 	if errBuffer.Len() > 0 {
-		return nil, fmt.Errorf("gofast: error stream from application process %s", errBuffer.String())
+		if strings.Contains(errBuffer.String(), "Primary script unknown") {
+			body := "Not Found"
+			return &http.Response{
+				Status:        "404 Not Found",
+				StatusCode:    404,
+				Body:          ioutil.NopCloser(bytes.NewBufferString(body)),
+				ContentLength: int64(len(body)),
+				Request:       req,
+				Header:        make(http.Header),
+			}, nil
+		} else {
+			return nil, fmt.Errorf("gofast: error stream from application process %s", errBuffer.String())
+		}
 	}
 
 	return rr.Result(), nil
--- a/proxy/fastcgi/fastcgi.go
+++ b/proxy/fastcgi/fastcgi.go
@@ -75,7 +75,7 @@ func (rt *RoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 		if strings.Contains(errBuffer.String(), "Primary script unknown") {
 			body := http.StatusText(http.StatusNotFound)
 			return &http.Response{
-				Status:        "404 Not Found",
+				Status:        body,
 				StatusCode:    http.StatusNotFound,
 				Body:          ioutil.NopCloser(bytes.NewBufferString(body)),
 				ContentLength: int64(len(body)),
--- a/proxy/fastcgi/fastcgi.go
+++ b/proxy/fastcgi/fastcgi.go
@@ -73,7 +73,7 @@ func (rt *RoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 
 	if errBuffer.Len() > 0 {
 		if strings.Contains(errBuffer.String(), "Primary script unknown") {
-			body := "Not Found"
+			body := http.StatusText(http.StatusNotFound)
 			return &http.Response{
 				Status:        "404 Not Found",
 				StatusCode:    http.StatusNotFound,
--- a/proxy/fastcgi/fastcgi.go
+++ b/proxy/fastcgi/fastcgi.go
@@ -76,7 +76,7 @@ func (rt *RoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 			body := "Not Found"
 			return &http.Response{
 				Status:        "404 Not Found",
-				StatusCode:    404,
+				StatusCode:    http.StatusNotFound,
 				Body:          ioutil.NopCloser(bytes.NewBufferString(body)),
 				ContentLength: int64(len(body)),
 				Request:       req,
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -973,6 +973,9 @@ func (p *Proxy) getRoundTripper(ctx *context, req *http.Request) (http.RoundTrip
 		if sf, ok := ctx.StateBag()["fastCgiFilename"]; ok {
 			f = sf.(string)
 		}
+		if req.URL.Path[1:] != "" {
+			f = req.URL.Path[1:]
+		}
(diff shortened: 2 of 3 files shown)


fbrousse:
now returning 404 when file is not existing on backend :

```
::1 - - [03/Feb/2021:12:19:35 +0100] "HEAD /test5.php?name=foo HTTP/1.1" 404 10 "-" "curl/7.63.0" 8 localhost - -
::1 - - [03/Feb/2021:12:19:50 +0100] "GET /test2.php?name=foo HTTP/1.1" 200 28 "-" "curl/7.63.0" 7 localhost - -
```

ripx80:
Looks good to me, ready for the review.
Thank you @fbrousse for this good investigation and solution.

proxy/proxy.go
```diff
@@ -973,6 +980,9 @@ func (p *Proxy) getRoundTripper(ctx *context, req *http.Request) (http.RoundTrip
 		if sf, ok := ctx.StateBag()["fastCgiFilename"]; ok {
 			f = sf.(string)
 		}
+		if req.URL.Path[1:] != "" {
```
universam1, line 983: Might be a case for an else condition here as the explicit filename would take preference

szuecs: I think also statebag entry should have precedence, here the statebag decision would be overridden.

proxy/proxy.go
```diff
@@ -1253,6 +1263,7 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 	}
 
 	switch {
+	case code == 404:
```
AlexanderYastrebov, line 1266: What is the intent here?

proxy/proxy.go
AlexanderYastrebov:
```go
else if req.URL.Path != "/" {
```

proxy/proxy.go
szuecs: I think to ignore the error log, which is fine in this case.
Do you have another opinion?

proxy/proxy.go
```diff
@@ -958,7 +958,14 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 			p.tracing.setTag(ctx.proxySpan, HTTPStatusCodeTag, uint16(status))
 			return nil, &proxyError{err: err, code: status}
 		}
-
+		if req.URL.Scheme == "fastcgi" {
```
AlexanderYastrebov, line 961: It should be possible to handle this case in fastcgi roundtripper internally and respond with normal 404 instead of returning a error. This way there would be no need to this and proxy error special handling.

proxy/proxy.go
AlexanderYastrebov: I see this is coming from fastcgi special case when script is not found. I think it should be handled inside fastcgi roundtripper internally if it is not really a proxy error.

szuecs: Yeah great idea!

proxy/proxy.go
fbrousse: thanks ! done

fbrousse (commented):
Moved back the handling of the 404 return code to fastcgi.go instead of proxy.go as suggested.

proxy/proxy.go
fbrousse: done ! thanks.

proxy/proxy.go
fbrousse: yes, makes sense. Done.

AlexanderYastrebov:
@fbrousse Thank you for updating the PR, it looks much cleaner now.
Fast CGI is a contributed backend implementation and I think it makes sense to add tests to keep high level of standard, the ground work is already there, see fastcgi_test.go and proxy_test.go.

proxy/fastcgi/fastcgi.go
```diff
@@ -70,7 +72,19 @@ func (rt *RoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	resp.WriteTo(rr, errBuffer)
 
 	if errBuffer.Len() > 0 {
-		return nil, fmt.Errorf("gofast: error stream from application process %s", errBuffer.String())
+		if strings.Contains(errBuffer.String(), "Primary script unknown") {
+			body := "Not Found"
+			return &http.Response{
+				Status:        "404 Not Found",
+				StatusCode:    404,
```
universam1, line 79:
```suggestion
				StatusCode:    http.StatusNotFound,
```

proxy/fastcgi/fastcgi.go
```diff
@@ -70,7 +72,19 @@ func (rt *RoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	resp.WriteTo(rr, errBuffer)
 
 	if errBuffer.Len() > 0 {
-		return nil, fmt.Errorf("gofast: error stream from application process %s", errBuffer.String())
+		if strings.Contains(errBuffer.String(), "Primary script unknown") {
+			body := "Not Found"
```
universam1, line 76:
```suggestion
			body := http.StatusText(http.StatusNotFound)
```

proxy/fastcgi/fastcgi.go
```diff
@@ -70,7 +72,19 @@ func (rt *RoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	resp.WriteTo(rr, errBuffer)
 
 	if errBuffer.Len() > 0 {
-		return nil, fmt.Errorf("gofast: error stream from application process %s", errBuffer.String())
+		if strings.Contains(errBuffer.String(), "Primary script unknown") {
+			body := http.StatusText(http.StatusNotFound)
+			return &http.Response{
+				Status:        "404 Not Found",
```
szuecs, line 78: "404 Not Found" -> body

proxy/proxy.go
```diff
@@ -972,6 +972,8 @@ func (p *Proxy) getRoundTripper(ctx *context, req *http.Request) (http.RoundTrip
 		f := "index.php"
 		if sf, ok := ctx.StateBag()["fastCgiFilename"]; ok {
 			f = sf.(string)
+		} else if req.URL.Path != "/" {
```
szuecs, line 975: `len(req.URL.Path) > 1 && req.URL.Path != "/"`

szuecs:
We will wait for the tests @fbrousse , thanks!

proxy/proxy.go
AlexanderYastrebov: I could not create empty path via
```
$ bin/skipper -inline-routes='* -> setRequestHeader("X-Path", "${request.path}") -> logHeader() -> <shunt>'
```

proxy/fastcgi/fastcgi.go
fbrousse: done :) thanks.

proxy/proxy.go
fbrousse: done, Thanks.

fbrousse:
@szuecs adding unit test 👍

fbrousse:
@szuecs @AlexanderYastrebov any remarks before merging ?

szuecs:
:+1:

szuecs:
@fbrousse from my side it's fine. We have in the proxy.Proxy a defaultHTTPStatus setting which could be used in favor of the current hard coded 404, but I think we can improve this later.

aryszka:
:+1:

Issue #1711, opened 2021-02-09
Configurable consistent hash request key


AlexanderYastrebov:
**Is your feature request related to a problem? Please describe.**

Consistent [hash algorithm](https://github.com/zalando/skipper/blob/master/docs/reference/backends.md#load-balancer-backend) uses client IP address to consistently select backend endpoint to serve client request. Skipper should allow users to specify consistent hash request key per route created from request properties such as headers or url.

**Describe the solution you would like**

Allow Skipper users to configure consistent hash request key as a string with [template placeholders](https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#template-placeholders). This should provide necessary flexibility and extensibility due to future templating improvements. 

Example route:
```
products: Path("/products/:productName") 
  -> consistentHashKey("${productName}") 
  -> <consistentHash, "http://127.0.0.1:9998", "http://127.0.0.1:9997">;
```

**Describe alternatives you've considered (optional)**

None

**Additional context (optional)**

Internal discussion brought up several ways to configure request key template:
* configure with a filter that puts key in statebag and pass statebag to `LBContext`
* configure with a marker filter (noop filter, similar to `fadeIn` filter) and a route preprocessor 
* configure with a filter and pass in [stdlib context](https://golang.org/pkg/context/)
* extend eskip to support algorithm parameters

Related to #557

**Would you like to work on it?**

Yes

Issue #1620, opened 2020-11-25
Configurable ClientIP utility function


AlexanderYastrebov:
**Is your feature request related to a problem? Please describe.**
There are several modules where client IP address is required. 

Depending on the setup it can be reliably obtained from:
* `http.Request.RemoteAddr` attribute
* first `X-Forwarded-For` header (e.g. when behind a loadbalancer) 
* last `X-Forwarded-For` header

Different modules (predicates, rate limiters, load balancing) use different strategies to obtain client IP address. E.g. there are three predicates `Source`, `SourceFromLast` and `ClientIP` that implement three different ways to obtain client IP. 

Ratelimiters can obtain IP address from first `X-Forwarded-For` header (via `XForwardedForLookuper`) but not from last, there is also no way to obtain IP address from `http.Request.RemoteAddr` directly regardless of `X-Forwarded-For` (similar problem #1614 addresses).

ConsistentHash loadbalancer algorithm always uses first `X-Forwarded-For` header.

⚠️ It is important that all usecases require client IP address and the proper way to obtain it depends on the Skipper setup or environment

**Describe the solution you would like**
Define `ClientIP` utility function that uses implementation configured on startup via command flag, something like:
```diff
diff --git a/net/net.go b/net/net.go
index 5e2da71..7535b48 100644
--- a/net/net.go
+++ b/net/net.go
@@ -1,6 +1,7 @@
 package net
 
 import (
+       "fmt"
        "net"
        "net/http"
        "strings"
@@ -61,3 +62,32 @@ func RemoteHostFromLast(r *http.Request) net.IP {
 
        return parse(r.RemoteAddr)
 }
+
+// Returns remote address of the client obtained from http.Request.RemoteAddr
+func hostFromRemoteAddr(r *http.Request) net.IP {
+       h, _, _ := net.SplitHostPort(r.RemoteAddr)
+       return net.ParseIP(h)
+}
+
+var clientIPSource = hostFromRemoteAddr
+
+// Configures source used by ClientIP function to provide client IP address
+func ConfigureClientIPSource(src string) error {
+       switch src {
+       case "XFF":
+               clientIPSource = RemoteHost
+       case "XFFLast":
+               clientIPSource = RemoteHostFromLast
+       case "RemoteAddr":
+               clientIPSource = hostFromRemoteAddr
+       default:
+               return fmt.Errorf("Unsupported client IP source '%s'. Use one of XFF, XFFLast or RemoteAddr", src)
+       }
+       return nil
[...]

PR #1713, opened 2021-02-09
Adds request IP address template placeholders

AlexanderYastrebov:
Adds three new template placeholders that resolve into request IP address similar to existing `Source`, `SourceFromLast` and `ClientIP` predicates.

* could be handy to support default key in "Configurable consistent hash request key" (#1711)
* allows client rate limit based on last value of `X-Forwarded-For` (`XForwardedForLookuper` supports only first value)
* named after existing predicates to avoid confusion (could be extended if/when "Configurable ClientIP utility function" (#1620) is implemented)


szuecs:
:+1:

aryszka:
:+1:

Issue #1712, opened 2021-02-09
Improve consistent hash endpoint selection strategy


AlexanderYastrebov:
**Is your feature request related to a problem? Please describe.**

[Consistent hash](https://github.com/zalando/skipper/blob/master/docs/reference/backends.md#load-balancer-backend) load balancing algorithm uses simple endpoint selection strategy `endpoint = endpoints[hash(request) % len(endpoints)]` to select endpoint to serve client request. When endpoint list changes (e.g. due to backend re-scaling) the assignment of nearly every request may change.

**Describe the solution you would like**

Implement true consistent hashing algorithm that minimizes request-to-endpoint reassignment when endpoint list chages.

**Describe alternatives you've considered (optional)**

None

**Additional context (optional)**

Links:
* https://en.wikipedia.org/wiki/Consistent_hashing
* https://medium.com/better-programming/load-balancers-and-consistent-hashing-in-6-minutes-b5fc460aea4e
* https://github.com/dgryski/go-rendezvous (consider using implementation used by go-redis ring client)
* https://github.com/golang/groupcache/blob/master/consistenthash/consistenthash.go

Related to #557

**Would you like to work on it?**

Yes

cjbooms:
Worth investigating if the Bounded Load algorithm could be implemented when working on this ticket:
https://arxiv.org/abs/1608.01350
https://medium.com/vimeo-engineering-blog/improving-load-balancing-with-a-new-consistent-hashing-algorithm-9f1bd75709ed
This algorithm is still consistent hashing load balancing, but it provides excellent failover to neighbouring nodes

cjbooms:
This ticket can be closed if PR #1715 fixes the issue of mass request-to-endpoint reassignments when endpoint list changes 
I've created a separate ticket for the bounded load algorithm. 
https://github.com/zalando/skipper/issues/1769

PR #1715, opened 2021-02-09
Improves consistent hash endpoint selection strategy

AlexanderYastrebov:
See #1712

### DIFF
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -6,6 +6,7 @@ import (
 	"math"
 	"math/rand"
 	"net/url"
+	"sort"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


loadbalancer/algorithm.go
```diff
@@ -190,31 +191,51 @@ func (r *random) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
+func (ch consistentHash) Less(i, j int) bool { return ch[i].hash < ch[j].hash }
+func (ch consistentHash) Swap(i, j int)      { ch[i], ch[j] = ch[j], ch[i] }
 
 func newConsistentHash(endpoints []string) routing.LBAlgorithm {
-	return &consistentHash{}
+	ch := consistentHash(make([]endpointHash, len(endpoints)))
+	for i, ep := range endpoints {
+		ch[i] = endpointHash{i, hash(ep)}
+	}
+	sort.Sort(ch)
```
szuecs, line 211: This might benefit from `sort.Stable()`, wdyt?

loadbalancer/algorithm.go
```diff
@@ -190,31 +191,51 @@ func (r *random) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
+	h := fnv.New32a()
+	h.Write([]byte(s))
+	return h.Sum32()
+}
+
+// Returns index of endpoint with closest hash to key's hash
+func (ch consistentHash) search(key string) int {
+	h := hash(key)
+	i := sort.Search(ch.Len(), func(i int) bool { return ch[i].hash >= h })
+	if i == ch.Len() { // rollover
```
szuecs, line 225: could be simplified to:
`i := sort.Search(..) % ch.Len()`

loadbalancer/algorithm.go
AlexanderYastrebov: The actual order does not really matter as long key search leads to the same endpoint consistently. Also it is sorted by hash value which (pseudo-)random.

loadbalancer/algorithm.go
AlexanderYastrebov: Yes, this will also work, here I explicitly check for the special case in line with binary search [documentation](https://golang.org/pkg/sort/#Search):
> ... Search returns the first true index. If there is no such index, Search returns n.

> A common use of Search is to find the index i for a value x in a sorted, indexable data structure ...
> ... i is the index where it would be inserted.

loadbalancer/algorithm.go
AlexanderYastrebov: But I see a an edge case here: when two endpoints happen to have the same hash, the first will receive requests and the second will not.

AlexanderYastrebov: > But I see a an edge case here
 
Strictly speaking consistent hash *does not guarantee* neither even request distribution across endpoints nor that every endpoint receives requests at all. 

E.g. backend with two endpoints may end up with different hashes but a very narrow hash difference (e.g. 1) by a bad luck and most/all of the requests would go to the first (with lower hash value) endpoint as the request key hash would be either less than the hash of the first endpoint or greater than the hash of the second endpoint (due to low chance to happen right inbetween) and thus rollover to the first.

szuecs:
:+1:

loadbalancer/algorithm.go
szuecs: ok, seems fine

loadbalancer/algorithm.go
szuecs: As far as I understand `i` can't be `>= ch.Len()`, so the "rollover" can be done without comparison using modulo operation.
I am fine with the +3 lines, but for me it seems more natural to use modulo.

aryszka:
nice and simple :heart:

aryszka:
:+1:

PR #1718, opened 2021-02-13
Adds consistentHashKey filter

### LINKED ISSUE #1711: Configurable consistent hash request key

AlexanderYastrebov:
`consistentHashKey` filter configures request key used by `consistentHash` load balancing algorithm to select backend endpoint.

Fixes #1711

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -2296,3 +2296,27 @@ Example:
 ```
 endpointCreated("http://10.0.0.1:8080", "2020-12-18T15:30:00Z01:00")
 ```
+
+## consistentHashKey
+
+This filter sets the request key used by the [`consistentHash`](backends.md#load-balancer-backend) algorithm to select the backend endpoint.
+
+Parameters:
+
+* key (string)
+
+The key should contain [template placeholders](#template-placeholders), without placeholders the key
+is constant and therefore all requests would be made to the same endpoint.
+The algorithm will use the default key if any of the template placeholders can't be resolved.
+
+Examples:
+
+```
+pr: Path("/products/:productId")
+    -> consistentHashKey("${productId}")
+    -> <consistentHash, "http://127.0.0.1:9998", "http://127.0.0.1:9997">;
+```
+```
+consistentHashKey("${request.header.Authorization}")
+consistentHashKey("${request.source}") // same as the default key
+```
--- a/filters/consistenthash/key.go
+++ b/filters/consistenthash/key.go
@@ -1,0 +1,35 @@
+package consistenthash
+
+import (
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/loadbalancer"
+)
+
+type consistentHashKey struct {
+	template *eskip.Template
+}
+
+// NewConsistentHashKey creates a filter Spec, whose instances
+// set the request key used by the `consistentHash` algorithm to select backend endpoint
+func NewConsistentHashKey() filters.Spec { return &consistentHashKey{} }
+func (*consistentHashKey) Name() string  { return "consistentHashKey" }
+
+func (*consistentHashKey) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) != 1 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+	value, ok := args[0].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+	return &consistentHashKey{eskip.NewTemplate(value)}, nil
+}
+
+func (c *consistentHashKey) Request(ctx filters.FilterContext) {
+	if key, ok := c.template.ApplyContext(ctx); ok {
+		ctx.StateBag()[loadbalancer.ConsistentHashKey] = key
+	}
+}
+
+func (*consistentHashKey) Response(filters.FilterContext) {}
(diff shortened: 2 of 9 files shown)


routing/routing.go
```diff
@@ -182,15 +182,7 @@ type LBAlgorithm interface {
 type LBContext struct {
 	Request *http.Request
 	Route   *Route
-}
-
-// NewLBContext is used to create a new LBContext, to pass data to the
-// load balancer algorithms.
-func NewLBContext(r *http.Request, rt *Route) *LBContext {
```
szuecs, line 189: this is a breaking API change.
We should try to omit these if we can.
Maybe better to let it and mar as deprecated and point to the new usage.
We can get rid of it in v1

AlexanderYastrebov: Yes, but only if anyone uses this API. The fix is straightforward - create `LBContext` instead of calling this constructor. Extending `LBContext` in backward compatible way is possible while extending this constructor is not so hopefully this is a one-time change for the the users (if any) to accommodate to. We may also think about considering all APIs as experimental by default until they are stable enough to require no extension and marked explicitly as stable, otherwise we will continue to face such extension impediments.

> We can get rid of it in v1

IMO there is no timeline for v1 and there is a risk to end up with a fork like Python 2/3 so v1 may very-well never happen.


I can keep the constructor and create `LBContext` directly if we really want to stick to zero user (if any) disturbance policy, just thought we may discuss this.

szuecs: Let's ask @aryszka as a 3rd person, I am fine with following his judgement. :)

aryszka: The only problem is that we don't know and should not make assumptions about by whom and how it is used. Pragmatism is of course welcome, and sometimes even necessary, but in this particular case, it is very easy to do the right thing, which in my opinion is to keep the initializer and mark it deprecated.

For us it's also a learning: in Go we don't need to create useless initializers where all the fields of a structure are exported and the initializer has no valuable logic.

docs/reference/filters.md
```diff
@@ -2286,3 +2286,27 @@ Example:
...
+
+This filter sets the request key used by [`consistentHash`](backends.md#load-balancer-backend) algorithm to select backend endpoint.
+
+Parameters:
+
+* key (string)
+
+The key should contain [template placeholders](#template-placeholders), without placeholders the key
+is constant and therefore all request would be made to the same endpoint.
+If a template placeholder can't be resolved then filter does not set the key and the default key would be used.
```
aryszka, line 2300: to my ears, a 'the' is missing here: `...then the filter does not set...`

filters/consistenthash/key.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/loadbalancer"
+)
+
+type consistentHashKey struct {
+	template *eskip.Template
+}
+
+func NewConsistentHashKey() filters.Spec { return &consistentHashKey{} }
```
aryszka, line 13: `NewConsistentHashKey` missing godoc? The description from the reference could be adapated here.

docs/reference/filters.md
AlexanderYastrebov: Rephrased

proxy/proxy.go
```diff
@@ -481,7 +481,7 @@ func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHea
 		setRequestURLFromRequest(u, r)
 		setRequestURLForDynamicBackend(u, stateBag)
 	case eskip.LBBackend:
-		endpoint = setRequestURLForLoadBalancedBackend(u, rt, routing.NewLBContext(r, rt))
+		endpoint = setRequestURLForLoadBalancedBackend(u, rt, &routing.LBContext{Request: r, Route: rt, Params: stateBag})
```
AlexanderYastrebov, line 484: Here statebag is used as a parameter map for the `LBContext`. Its a kind of loose coupling between filters and algorithm, is that fine?

Other options I've considered did not look ok:
* stdlib request context needs request substitution in the filter context
* evaluating template within the algorithm requires it to be aware of path variables required by `eskip.TemplateContext`. It also requires route postprocessor

szuecs: For the stdlib case you would use context.WithValue() and you would overwrite somewhere ctx.Request().Context(), right?
So you don't need to overwrite Request, only the context or do I miss something?

For me the current approach seems also fine.

AlexanderYastrebov: https://golang.org/pkg/net/http/#Request.WithContext creates request copy that would need to replace request in filter context

aryszka: this looks fine to me

aryszka:
:+1:

szuecs:
👍

PR #1719, opened 2021-02-15
Always add http status code to initial span

### LINKED ISSUE #1145: opentracing reports 500 when a route was not found
**Describe the bug**
When the route lookup doesn't return a route, skipper returns 404 by default, or the default status code set from the startup options, but 500 is sent to opentracing instead.

**To Reproduce**
Add context, for example:
1. start skipper with no routing configured, but with opentracing enabled
2. make any http request to skipper
3. skipper should respond with 404
4. check the matching trace in the used opentracing provider

**Expected behavior**
the trace has recorded 404 as the http status tag

**Observed behavior**
the trace has recorded 500

rbarilani:
Try to always set `http.status_code` on the initial span too and not just on the `proxy` span to allow querying the main span by `http.status_code` regardless if the status code is coming from the downstream or by skipper itself (eg. not found route, filters serving the response).

~Should fix also #1145~

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1175,6 +1175,8 @@ func (p *Proxy) serveResponse(ctx *context) {
 		p.tracing.setTag(ctx.proxySpan, ClientRequestStateTag, ClientRequestCanceled)
 	}
 
+	p.tracing.setTag(ctx.initialSpan, HTTPStatusCodeTag, uint16(ctx.response.StatusCode))
+
 	ctx.responseWriter.WriteHeader(ctx.response.StatusCode)
 	ctx.responseWriter.Flush()
 	n, err := copyStream(ctx.responseWriter, ctx.response.Body)
@@ -1413,6 +1415,7 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx = newContext(lw, r, p)
 	ctx.startServe = time.Now()
 	ctx.tracer = p.tracing.tracer
+	ctx.initialSpan = span
 
 	defer func() {
 		if ctx.response != nil && ctx.response.Body != nil {
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1224,13 +1224,10 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 		code = perr.code
 	}
 
-	if span := ot.SpanFromContext(ctx.Request().Context()); span != nil {
-		p.tracing.setTag(span, ErrorTag, true)
-		p.tracing.setTag(span, HTTPStatusCodeTag, uint16(code))
-		if err == errRouteLookupFailed {
-			span.LogKV("event", "error", "message", errRouteLookup.Error())
-		}
-
+	p.tracing.setTag(ctx.initialSpan, ErrorTag, true)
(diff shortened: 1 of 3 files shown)


proxy/proxy.go
```diff
@@ -1413,6 +1415,7 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx = newContext(lw, r, p)
 	ctx.startServe = time.Now()
 	ctx.tracer = p.tracing.tracer
+	ctx.initialSpan = span
```
szuecs, line 1418: After p.do(ctx) call you can just do:
```
	p.tracing.setTag(ctx, HTTPStatusCodeTag, uint16(ctx.response.StatusCode))
```

rbarilani: @szuecs at that point the `ctx.response.StatusCode` can be `nil` throwing a panic.

proxy/proxy.go
```diff
@@ -1175,6 +1175,8 @@ func (p *Proxy) serveResponse(ctx *context) {
 		p.tracing.setTag(ctx.proxySpan, ClientRequestStateTag, ClientRequestCanceled)
 	}
 
+	p.tracing.setTag(ctx.initialSpan, HTTPStatusCodeTag, uint16(ctx.response.StatusCode))
```
AlexanderYastrebov, line 1178: This is a successful response branch of
https://github.com/zalando/skipper/blob/ff61a38328ee6160111dee40f935d0bd239b297c/proxy/proxy.go#L1426-L1432
but there is also a error branch (e.g. skipper failed to connect to backend). Should the  error branch also be covered?

rbarilani: @AlexanderYastrebov error should be already covered by this https://github.com/zalando/skipper/blob/master/proxy/proxy.go#L1225, inside `errorResponse`

AlexanderYastrebov: Could you rewrite it then to explicit `initialSpan` since you are introducing it?

rbarilani: @AlexanderYastrebov I am not 100% sure that they are interchangeable? Are they? `spanFromContext` is getting the currently "active span", it might be some other span eventually?

AlexanderYastrebov: From
https://github.com/zalando/skipper/blob/ff61a38328ee6160111dee40f935d0bd239b297c/proxy/proxy.go#L1411-L1413
https://github.com/zalando/skipper/blob/ff61a38328ee6160111dee40f935d0bd239b297c/proxy/context.go#L131-L151
https://github.com/opentracing/opentracing-go/blob/3088eee7e4d26010ac9a301e977fd2721dbebcbe/gocontext.go#L26-L32
I think 
https://github.com/zalando/skipper/blob/76fb0a7705c6238d6be1c3e441521bdb44752006/proxy/proxy.go#L1225-L1232
will get initial `span` and I think it is done like that because it was not passed around in the `ctx`.

Also from the use of initial span it looks like it is assumed to be non-nil so the nil check seems to be redundant.

rbarilani: @AlexanderYastrebov applied the requested change

szuecs:
:+1:

aryszka:
:+1:

PR #1723, opened 2021-02-17
add protocol arg to be used to switch of GRPC

szuecs:
add protocol arg to be used to switch of GRPC, a string arg was introduced instead of bool, because protocol might change in the future

### DIFF
--- a/tracing/tracers/lightstep/lightstep.go
+++ b/tracing/tracers/lightstep/lightstep.go
@@ -101,9 +101,7 @@ func parseOptions(opts []string) (lightstep.Options, error) {
 			cmdLine = parts[1]
 			logCmdLine = true
 		case "protocol":
-			if parts[1] != "grpc" {
-				useGRPC = false
-			}
+			useGRPC = parts[1] != "http"
 		case "log-events":
 			logEvents = true
 		case "max-buffered-spans":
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

tracing/tracers/lightstep/lightstep.go
```diff
@@ -99,6 +100,10 @@ func parseOptions(opts []string) (lightstep.Options, error) {
 		case "cmd-line":
 			cmdLine = parts[1]
 			logCmdLine = true
+		case "protocol":
+			if parts[1] != "grpc" {
+				useGRPC = false
```
vetinari, line 105: I'd make that explicit `http` or `grpc` with `grpc` as default. It's easier to figure out what you get instead of relying on the default in the tracer (that used to be `grpc` instead of `http` in earlier versions IIRC).

aryszka: I agree, let's make it explicit http, anticipating other variants, or subvariants of the current ones in the same option

szuecs: makes sense, but it has the drawback of changing defaults.

aryszka:
:+1:

szuecs:
:+1:

PR #1725, opened 2021-02-19
Uses debug log level for lifo error logs

AlexanderYastrebov:
to avoid flooding the logs when lifo filter overruns

### DIFF
--- a/filters/scheduler/lifo.go
+++ b/filters/scheduler/lifo.go
@@ -297,13 +297,13 @@ func request(q *scheduler.Queue, key string, ctx filters.FilterContext) {
 		// TODO: replace the log with metrics
 		switch err {
 		case jobqueue.ErrStackFull:
-			log.Errorf("Failed to get an entry on to the queue to process QueueFull: %v for host %s", err, ctx.Request().Host)
+			log.Debugf("Failed to get an entry on to the queue to process QueueFull: %v for host %s", err, ctx.Request().Host)
 			ctx.Serve(&http.Response{
 				StatusCode: http.StatusServiceUnavailable,
 				Status:     "Queue Full - https://opensource.zalando.com/skipper/operation/operation/#scheduler",
 			})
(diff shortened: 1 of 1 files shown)


szuecs:
Do you think metrics are enough?
@aryszka wdyt?

AlexanderYastrebov:
> Do you think metrics are enough?

From TODO comment
https://github.com/zalando/skipper/blob/7de16ef81048ba58da0771d340d995bdb4d1bf8f/filters/scheduler/lifo.go#L297
I assume there are some metrics missing

filters/scheduler/lifo.go
```diff
@@ -297,19 +297,19 @@ func request(q *scheduler.Queue, key string, ctx filters.FilterContext) {
...
 		case jobqueue.ErrTimeout:
-			log.Errorf("Failed to get an entry on to the queue to process Timeout: %v for host %s", err, ctx.Request().Host)
+			log.Debugf("Failed to get an entry on to the queue to process Timeout: %v for host %s", err, ctx.Request().Host)
 			ctx.Serve(&http.Response{
 				StatusCode: http.StatusBadGateway,
 				Status:     "Queue timeout - https://opensource.zalando.com/skipper/operation/operation/#scheduler",
 			})
 		default:
-			log.Errorf("Unknown error for route based LIFO: %v for host %s", err, ctx.Request().Host)
+			log.Debugf("Unknown error for route based LIFO: %v for host %s", err, ctx.Request().Host)
```
aryszka, line 312: I absolutely agree with turning the above two into debug, because they are considered normal operation of the lifo.
However, we can keep this last one as error, because this is something that we don't expect to happen, and we'd need to know the error message in production.

AlexanderYastrebov: Indeed, fixed.

szuecs:
👍

aryszka:
:+1:

PR #1731, opened 2021-02-22
Adds lua sleep function

AlexanderYastrebov:
Sleep function may be useful for prototyping but unfortunately Lua's
standard library does not provide it, see http://lua-users.org/wiki/SleepFunction


szuecs:
:+1:

aryszka:
:+1:

PR #1732, opened 2021-02-25
Update oauthGrant authorization header flag docs

mohabusama:
`oauth2-access-token-header-name` is needed by the `oauthGrant` filter to pass the bearer token in the request headers. The documented default value did not work.


szuecs:
:+1:

jonathanbeber:
:+1:

PR #1734, opened 2021-02-26
feat: add status code as arg to ratelimit and clusterRatelimit filters

### LINKED ISSUE #1728: rateLimit and clusterRateLimit filters should use HTTP 503 response status code

Rate limiting capabilities that are not based on a client quota (by IP, by header, etc.) such as the one activated by `rateLimit` and `clusterRateLimit` filters are frequently used to protect the backend to be overloaded by too many requests, stopping requests based on server maximum capacity.

The current HTTP status code returned by those rate limiters is `429`.

Considering HTTP semantics though, a [503](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) status code might be more appropriate given that 4xx error indicates a bad/malformed request sent by the client.  
In this case the client "hasn't done anything wrong", they didn't overflow the quota that was assigned to them, the maximum requests might be reached since some other client misbehaved, had more traffic to digest, etc.
The client didn't send too many requests in a given amount of time ("rate limiting"), it's the backend that is unavailable to receive more requests. 

As per [503 HTTP documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) a header like ` Retry-After` can be added to the response representing the estimated time for the recovery of the service as is used for [429](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429)'s responses.

The usage of a different status code would help also to distinguish in HTTP-based reports (logs, stats) the requests which were rate-limited since clients exceeded their quota from requests which were rate-limited since service unavailability.

Change or allow to change the response status code used by the rate limit filters to `503`.
To make this customization possible I would like to have a configuration value that enables to choose the response status code returned by the rate limit filters.
The value can be provided per filter for all routes (at skipper configuration level) and/or can be applied on a per route basis, eg.

```eskip
route_id: Path("/hello") -> clusterRateLimit(20, "1m", 503) -> https://mybackend.com
```

I have considered keeping the 429 status and use some other type of filtering on logs, open tracing traces to distinguish client quota-based rate-limited requests from "capacity-based" rate-limited requests but nothing simple and immediate as a different HTTP status code come out. 
We might think also to add optionally the information on the access logs but it sounds dirty.
We can filter on open tracing spans but I am not sure based on what especially when two rate limiters one client-based such as `clusterClientRateLimit` and `clusterRateLimit` are applied to the same route.

N/A

Yes

rbarilani:
Add the possibility to provide the status code as an argument for `ratelimit` and `clusterRatelimit` filters.
The value will be used as response status code for a rate limited request, default value if not provided is the standard `429`.

close #1728

### DIFF
--- a/filters/ratelimit/ratelimit.go
+++ b/filters/ratelimit/ratelimit.go
@@ -16,6 +16,8 @@ import (
 	"github.com/zalando/skipper/ratelimit"
 )
 
+const defaultStatusCode = http.StatusTooManyRequests
+
 type spec struct {
 	typ        ratelimit.RatelimitType
 	provider   RatelimitProvider
@@ -159,7 +161,7 @@ func (s *spec) Name() string {
 }
 
 func serviceRatelimitFilter(args []interface{}) (*filter, error) {
-	if len(args) < 2 {
+	if !(len(args) == 2 || len(args) == 3) {
 		return nil, filters.ErrInvalidFilterParameters
 	}
 
@@ -190,7 +192,7 @@ func serviceRatelimitFilter(args []interface{}) (*filter, error) {
 }
 
 func clusterRatelimitFilter(args []interface{}) (*filter, error) {
-	if len(args) < 3 {
+	if !(len(args) == 3 || len(args) == 4) {
 		return nil, filters.ErrInvalidFilterParameters
 	}
 
@@ -272,7 +274,7 @@ func clusterClientRatelimitFilter(args []interface{}) (*filter, error) {
 		s.Lookuper = ratelimit.NewXForwardedForLookuper()
 	}
 
-	return &filter{settings: s}, nil
+	return &filter{settings: s, statusCode: defaultStatusCode}, nil
 }
 
 func getLookuper(s string) ratelimit.Lookuper {
@@ -326,6 +328,7 @@ func clientRatelimitFilter(args []interface{}) (*filter, error) {
 			CleanInterval: 10 * timeWindow,
 			Lookuper:      lookuper,
 		},
+		statusCode: defaultStatusCode,
 	}, nil
 }
 
@@ -334,6 +337,7 @@ func disableFilter([]interface{}) (*filter, error) {
 		settings: ratelimit.Settings{
 			Type: ratelimit.DisableRatelimit,
 		},
+		statusCode: defaultStatusCode,
 	}, nil
 }
 
@@ -393,9 +397,9 @@ func getDurationArg(a interface{}) (time.Duration, error) {
 }
 
 func getStatusCodeArg(args []interface{}, index int) (int, error) {
-	// status code arg is optional so we return 0 but no error
+	// status code arg is optional so we return default status code but no error
 	if len(args) <= index {
-		return 0, nil
+		return defaultStatusCode, nil
 	}
 
 	statusCode, err := getIntArg(args[index])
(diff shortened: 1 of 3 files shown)


filters/ratelimit/ratelimit.go
```diff
@@ -379,6 +392,20 @@ func getDurationArg(a interface{}) (time.Duration, error) {
 	return time.Duration(i) * time.Second, err
 }
 
+func getStatusCodeArg(args []interface{}, index int) (int, error) {
+	// status code arg is optional so we return 0 but no error
```
AlexanderYastrebov, line 396: Simply return `http.StatusTooManyRequests` and don't bother checking zero in `Request`

rbarilani: @AlexanderYastrebov I can't since the argument is used just when creating `ratelimit` and `clusterRateLimit` filters, for the other filters the `statusCode` filter property will be equal to 0.

filters/ratelimit/ratelimit.go
```diff
@@ -158,7 +159,7 @@ func (s *spec) Name() string {
 }
 
 func serviceRatelimitFilter(args []interface{}) (*filter, error) {
-	if len(args) != 2 {
+	if len(args) < 2 {
```
aryszka, line 162: `len(args) < 2 && len(args) > 3`

filters/ratelimit/ratelimit.go
aryszka: to me it still seems possible that what @AlexanderYastrebov suggested could work. In general, how it should work:

- accept the optional parameter only where we use it
- don't accept more parameters

This should be possible, because the spec type has a typ field marking which concrete filter we're using, and each filter has it's own init function. Do I miss something?

rbarilani: > accept the optional parameter only where we use it 
The criteria is met with the current implementation. The optional param can be used just with "non-client rate limit" filters.
> don't accept more parameters 
I will update the code to have a more strict check as you suggested.

About
> Simply return `http.StatusTooManyRequests` and don't bother checking zero in `Request`.

I am using the filter struct to hold the status code which will be used for a limited response. Since we accept the optional parameter just on "non-client rate limit" filters for client-based rate limit filters the status code will be equal to 0. If you prefer, I can instead of checking zero in the `Request` method initialize the statusCode for each filter to the default (`http.StatusTooManyRequests`).

aryszka: > If you prefer, I can instead of checking zero in the Request method initialize the statusCode for each filter to the default (http.StatusTooManyRequests).

Yes, that's kind of also how I imagined it. Thx!

AlexanderYastrebov: @rbarilani Sorry, I've jumped in with a comment but then realized it is a draft so I've decided to look later.
What I've meant indeed is to have a status field initialized with default backward-compatible 429 and override it with optinal filter argument. Since status code does not change through the filter lifetime it is not necessary to do any runtime check in the `Request()` hot path.

rbarilani:
@aryszka @AlexanderYastrebov Hi folks, I have updated the implementation. If the implementation is fine I will proceed with documentation changes.

filters/ratelimit/ratelimit.go
```diff
@@ -379,6 +396,20 @@ func getDurationArg(a interface{}) (time.Duration, error) {
...
+	if len(args) <= index {
+		return defaultStatusCode, nil
+	}
+
+	statusCode, err := getIntArg(args[index])
+	if err != nil {
+		return 0, err
+	}
+
+	return statusCode, nil
```
AlexanderYastrebov, line 410:
```go
return getIntArg(args[index])
```

AlexanderYastrebov:
> If the implementation is fine I will proceed with documentation changes.

Not sure if we need to check status code belongs [100, 599], otherwise looks good to me.

filters/ratelimit/ratelimit.go
rbarilani: Hi @AlexanderYastrebov updated.

aryszka:
:+1:

Issue #1709, opened 2021-02-08
Wait for first sync for each configured dataclient


szuecs:
Right now if `-wait-first-route-load` is specified we wait for the first dataclient to return from `LoadAll()`.

The problem is this:

```
        o := skipper.Options{
	CustomDataClients: []routing.DataClient{
		dc1,
                dc2,
                ...
                dcN,
	},
        WaitFirstRouteLoad: true,
}

err = skipper.Run(o) // waits only for the first dc that did the LoadAll()
```

https://github.com/zalando/skipper/blob/master/routing/routing.go#L361 should get signaled which dataclients are expected to LoadAll and wait for all of them. A simple counter to wait for N LoadAll() might add a race condition, if we call LoadAll concurrently (I think we don't but this should be checked).

PR #1735, opened 2021-02-27
Support for watched remote eskip files

xavivars:


szuecs:
@xavivars from my side looks good to me. For creating a remoteFileServer for tests you can just use a simple server that serves file content.
We often use something like this as backend to test the proxy:
```
backend := httptest.NewServer(http.HandlerFunc(func(_ http.ResponseWriter, r *http.Request) {}))
```

Instead of `os.Create` I would prefer to use `os.OpenFile()` to set permissions to `0600`, such that you don't leak anything useful to other users running on the same host.

Instead of http.Get() you could use skipper's net.Client.Get(), but this is minor IMO. In case of a DNS switch to a new remote FileServer http.Get() might fail. Not 100% sure in this case, maybe it uses an empty Transport which would be fine.

Logging, I would not add a `verbose` member, but rely on logging settings.

xavivars:
I didn't know about `httptest`, and it's great! 

I've just changed the tests to use a test server, and I've also changed `os.Create` with `os.OpenFile`. 

I haven't changed the http client to avoid "creating" & "closing" a new http client there in code, but if you still think that's better, I can do that.

eskipfile/remote.go
```diff
@@ -0,0 +1,146 @@
...
+	remotePath      string
+	localPath       string
+	eskipFileClient *WatchClient
+	threshold       int
+	verbose         bool
+}
+
+// RemoteWatch creates a route configuration client with (remote) file watching. Watch doesn't follow file system nodes,
+// it always reads (or re-downloads) from the file identified by the initially provided file name.
+func RemoteWatch(rf string, th int, v bool) (routing.DataClient, error) {
```
aryszka, line 27: I think we should use here an Options structure instead of positional parameters, because it's fair to expect plenty more parameters in the future: timeout settings, tracing configuration, etc...

eskipfile/remote.go
```diff
@@ -0,0 +1,146 @@
...
+	if !isFileRemote(rf) {
+		return Watch(rf), nil
```
aryszka, line 29: instead of falling back to a local file in each case when it's not remote, how about doing the following:

- http:/https: -> use http(s)
- file: -> use local file
- no schema prefix -> use local file
- unknown schema prefix -> error

eskipfile/remote.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	return newRoutes, deletedRoutes, err
+}
+
+func isFileRemote(remotePath string) bool {
+	return strings.HasPrefix(remotePath, "http://") || strings.HasPrefix(remotePath, "https://")
+}
+
+func (client *remoteEskipFile) DownloadRemoteFile() error {
```
aryszka, line 114: i think the `DownloadRemoteFile` method is not used outside of this package, can we rename it to just lowercase `downloadRemoteFile`?

eskipfile/remote.go
```diff
@@ -0,0 +1,146 @@
...
+
+	if _, err = io.Copy(out, data); err != nil {
+		_ = out.Close()
+		return err
+	}
+
+	return out.Close()
+}
+
+func (client *remoteEskipFile) GetRemoteData() (io.ReadCloser, error) {
```
aryszka, line 135: i think the `GetRemoteData` method is not used outside of this package, can we rename it to just lowercase `getRemoteData`?

eskipfile/remote.go
```diff
@@ -0,0 +1,146 @@
...
+		remotePath: rf,
+		localPath:  tempFilename.Name(),
+		threshold:  th,
+		verbose:    v,
+	}
+
+	err = dataClient.DownloadRemoteFile()
+
+	if err != nil {
+		return nil, err
```
aryszka, line 48: I would defer the first download until the first call to LoadAll. The reason is that there can always be temporary hickups, and there's no reason for preventing Skipper to start up on the failure of the source.

What makes this a bit complex: is the default 404 OK in such cases? Our usual approach is that we don't report healthy from Skipper until the first successful route fetch.

eskipfile/remote.go
aryszka: also, I would move this decision logic outside of the `RemoteWatch` initializer, because this way it is a bit misleading

eskipfile/remote.go
aryszka: I see from the usage of the `preloaded` field that this is a conscious decision. Can you give some rationale for this approach?

eskipfile/remote.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// RemoteWatch creates a route configuration client with (remote) file watching. Watch doesn't follow file system nodes,
+// it always reads (or re-downloads) from the file identified by the initially provided file name.
+func RemoteWatch(rf string, th int, v bool) (routing.DataClient, error) {
+	if !isFileRemote(rf) {
+		return Watch(rf), nil
+	}
+
+	tempFilename, err := ioutil.TempFile("", "routes")
```
aryszka, line 32: I see some potential in the usage of the tempfile, e.g. it could be used to fetch from the remote less frequently or with varying frequency (like slowing down when there are no changes), than how often the routing mechanism calls the LoadUpdate method. However, in the current implementation, I don't see this yet. Which is fine. What I don't fully understand why do we use the tempfile here? Can you give a short explanation? I may miss something.

eskipfile/remote.go
xavivars: Good idea. Solved!

eskipfile/remote.go
xavivars: I actually was initially considering doing this logic in the `Watch` dataClient, and have a private `localWatch` and a private `remoteWatch` data clients, so: 
1. it's could be done so it's backwards compatible
2. a consolidated API for Watched files, regardless where they are.

But I didn't do it in case it felt too "disrupting". But if you like the approach, I'm happy to give another thought.

eskipfile/remote.go
xavivars: We're using a tempfile because this implementation relies 100% on the existing `Watch` data client, it's only a wrapper around it. When downloading the file, the data remains in memory, so we need it to dump it into the local file so Watch works properly.

Am I missing something?

eskipfile/remote.go
xavivars: I think all of this is a side effect of us trying to work around https://github.com/zalando/skipper/issues/1709 not working.

What we were doing (using skipper as a library) was to call force `DownloadRemoteFile` when the dataClient was created, and then when `Skipper.run` was being invoked, the file was already downloaded, and startup is faster (eskip files can be several MB, and download is not instantaneous).

But with #1709 being fixed now, I'm not sure all this logic is still needed.

eskipfile/remote.go
xavivars: We're using it from our `main.go`, so we "fail" (don't even do `skipper.run`) if the remote file is not downloadable: we prefer our cluster to keep "old" containers with a valid (even if maybe updated) route file in memory than to start with new but routeless instances.

eskipfile/remote.go
xavivars: Changed

eskipfile/remote.go
xavivars: I added a draft of how it would look like.

eskipfile/remote.go
szuecs: @xavivars we use the feature that Listener is only started after the first successful LoadAll() and health check from load balancer will keep us safe.
Additionally you can also test the endpoint with -status-checks

xavivars: Yes, the problem when I initially implemented this (more than one year ago) was that the feature you're talking wasn't working as we expected (due to the bug I reported & fixed few weeks ago). But agree, this shouldn't be needed anymore

aryszka (approved):

aryszka:
thx, for the updates, lgtm

PR #1738, opened 2021-03-05
remove URL tag in ingress span

szuecs:
removing URL tag in ingress span can save a lot of bytes for example with elastic search cases
Interesting observation: proxy span has no kv pairs from query, that's why I let it there.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -902,13 +902,13 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 	}
 	ctx.proxySpan = tracing.CreateSpan(spanName, req.Context(), p.tracing.tracer)
 
+	u := cloneURL(req.URL)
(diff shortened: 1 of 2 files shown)


proxy/proxy.go
```diff
@@ -1449,7 +1449,6 @@ func (p *Proxy) Close() error {
 func (p *Proxy) setCommonSpanInfo(u *url.URL, r *http.Request, s ot.Span) {
 	p.tracing.
 		setTag(s, ComponentTag, "skipper").
-		setTag(s, HTTPUrlTag, u.String()).
```
pitr, line 1452: you can replace with log instead

lmineiro: Replacing the tag with a log is not going to help reduce the size of the gRPC message

aryszka:
:+1:

szuecs:
:+1:

Issue #945, opened 2019-02-07
Integrate Go-fuzz for testing 


szuecs:
It would be good to have something used to fuzz skipper. https://github.com/dvyukov/go-fuzz could be something to think about

PR #1750, opened 2021-04-02
Add eskip fuzzer

AdamKorcz:
Closes https://github.com/zalando/skipper/issues/945

This PR adds the first fuzzer for skipper.

The fuzzer is implemented by way of go-fuzz.

It will also run continuously through OSS-fuzz, and I will be happy to set up an integration application over there, so it can run continuously. Once integrated, OSS-fuzz will send out notifications if any bugs are found.

### DIFF
--- a/eskip/fuzz.go
+++ b/eskip/fuzz.go
@@ -1,0 +1,6 @@
+package eskip
+
+func FuzzParse(data []byte) int {
+	_, _ = Parse(string(data))
+	return 1
+}


szuecs:
Thanks great for your help!
To understand better what it does:
You setup an exported func to make it callable by go-fuzz, which will be used and triggered by oss-fuzz. If we see panic or something other than 1 as return value we get a notification.
How do we get a notification?

szuecs:
👍

AdamKorcz:
> You setup an exported func to make it callable by go-fuzz, which will be used and triggered by oss-fuzz. If we see panic or something other than 1 as return value we get a notification.

Correct!

> How do we get a notification?

Emails are sent out with links to bug reports that contain a stack trace and a reproducer testcase.

szuecs:
Do you plan to create the oss-fuzz configuration or should we do it?
If you want to do it, please add e-mail addresses from https://github.com/zalando/skipper/blob/master/MAINTAINERS in the notification configuration.

AdamKorcz:
I will set up the configuration and build scripts on the OSS-fuzz side.

AdamKorcz:
@szuecs Do you happen to know who uses this project besides Zalando?
It increases the chances of integrating with OSS-fuzz if we can mention a few of its users in the process.

aryszka:
:+1:

szuecs:
I replied to your email.

PR #1755, opened 2021-04-15
Allow to reduce per Host and Route metrics domains

jonathanbeber:
The number of metrics for latency per host and per route include the
HTTP response status code and the HTTP method as dimensions. It
generated flexible metrics, allowing queries rich in details, like: "the
P99 of GET requests to the route abc that resulted in a 503", but, also,
generates a lot of metrics not always used.

For example, imagine a 32 long histogram metric bucket, a cluster with
30 hosts, 4 different reported status code and 3 different reported
methods. It might generate 11520 different metrics (32*30*4*3).

This commit adds the flags `-serve-method-metric` and
`-serve-status-code-metric`. It allow the disablement of the method
and/or status code labels from metrics. It implements the logic to
remove it from both codahale and prometheus metric flavours.

For compatibility reasons, yet another two flags are introduced:
`-serve-host-counter` and `-serve-route-counter`. These new counters
include all the dimensions, even if those are disabled for latency
buckets, for prometheus. That's due to prometheus [Histograms][0]
nature: "it also provides a sum of observations and an observation
count". So, users migrating from a full dimensions scenario to one
without them can still get counting metrics. The new counters have the
name `skipper_serve_route_count` and `skipper_serve_host_count`.

On the same example, removing the status code and method labels results
in 960 metrics, a ~90% reduction.

[0]: https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Histogram

Signed-off-by: Jonathan Juares Beber <jonathanbeber@gmail.com>


mikkeloscar:
:+1:

AlexanderYastrebov:
> For compatibility reasons, a new counter including all the dimensions,
even if both dimensions are disabled for latency buckets, is included
for prometheus.

Why is this additional counter is needed? Even when both method and status are disabled, there still will be counters for route and host, won't they?

lmineiro:
@jonathanbeber have you considered a multiple-value setting for `-serve-status-code-metric` such that skipper could still provide metrics per status class like 1XX, 2XX, 3XX, 4XX, 5XX? It would still reduce considerably the number of unique metrics while providing some reasonable insights given the very distinct semantics of each of the different HTTP status code classes

AlexanderYastrebov:
> For compatibility reasons, a new counter including all the dimensions,

I think I get it now, we still want counter for all dimensions even if some are turned-off

> A drawback of this implementation is that
when both method and status code dimensions are enabled the counter is
duplicated.

Then maybe additional counters should be included only when
```go
allDimensions := p.opts.EnableServeStatusCodeMetric && p.opts.EnableServeMethodMetric
...
if !allDimensions {
  p.serveRouteCounterM.WithLabelValues(fmt.Sprint(code), method, routeID).Inc()
}
```
to avoid duplication?

jonathanbeber:
@AlexanderYastrebov that's a good idea, I will implement it. Thanks.

jonathanbeber:
@AlexanderYastrebov well, now thinking better about that, the problem is the name of the metric. As Prometheus automatically creates the counter for histograms we would have a situation with names changing if you choose not to have all of them.

It would also be a problem for us during the migration phase. There will be a moment where we require the old counter with all the dimensions and the new one in parallel, to switch the queries and do not lose data. We could create (yet) another flag to enable and disable the general counter, but I think it's a cheap drawback to include another flag. We could also easily implement your idea after the migration, but I'm not sure how much I like it and how it could affect other skipper users migrating.

What's your thoughts on that?

AlexanderYastrebov:
> There will be a moment where we require the old counter with all the dimensions and the new one in parallel, to switch the queries and do not lose data.
> We could create (yet) another flag to enable and disable the general counter

Yeah, I have a feeling that we are working backwards (reducing) here. Maybe if that would be designed from ground up there would be two new flags to steer counters independently like `serve-method-metric-counter` and `serve-status-code-metric-counter`...

This PR will add counters for all dimensions for all users (i.e. duplicate metrics) by default if I get it right.

jonathanbeber:
> This PR will add counters for all dimensions for all users (i.e. duplicate metrics) by default if I get it right.
Correct, it will generate a new metric for all users: the counter with host/route, method and status code domain. They already have this metric if they enable `serveHostMetrics`/`serveRouteMetrics` once it comes with the histogram.

jonathanbeber:
@AlexanderYastrebov I just pushed two new flags, one for route counter and another for host counter. It removes the repetition (well, unless you choose to have it). Thanks for pointing it, I think it's better this way.

szuecs:
Please check if z-skipper case is fine, too.

jonathanbeber:
please keep in mind that z-skipper can still use the metrics the way it's today. It's even the default.

AlexanderYastrebov:
@jonathanbeber 
> I just pushed two new flags, one for route counter and another for host counter.

I like it better as well, also allows users to disable histograms but still have counters.

metrics/codahale.go
```diff
@@ -179,13 +177,24 @@ func (c *CodaHale) MeasureResponse(code int, method string, routeId string, star
...
+	keyServeHost := fmt.Sprintf("servehost.%s", hostForKey(host))
+
+	if c.options.EnableServeMethodMetric {
+		keyServeHost += fmt.Sprintf(".%s", method)
+		keyServeRoute += fmt.Sprintf(".%s", method)
+	}
+	if c.options.EnableServeStatusCodeMetric {
+		keyServeHost += fmt.Sprintf(".%d", code)
+		keyServeRoute += fmt.Sprintf(".%d", code)
+	}
```
AlexanderYastrebov, line 190: This will be executed even when `EnableServeRouteMetrics` or/and `EnableServeHostMetrics` are disabled.

AlexanderYastrebov: The same is for prometheus

jonathanbeber: yeah, although I know we might have a small performance/memory penalty for that, I thought the code would be more "direct". I was undecided, thanks for pointing it out, I just pushed a fix. It's hard to test, thought.

docs/operation/operation.md
```diff
@@ -327,6 +331,12 @@ This will make sure you will get stats for each "Host" header as "timers":
       }
     },
 
+Note you can reduce the domain of the metrics by removing the HTTP
```
mikkeloscar, line 334: nit: domain -> dimension?

metrics/metrics.go
```diff
@@ -99,10 +99,34 @@ type Options struct {
 	// for each route, additionally grouped by status and method.
 	EnableServeRouteMetrics bool
 
+	// If set, a counter for each route is generated, additionally
+	// grouped by status and method. It differs from the automatically
+	// generated counter from `EnableServeRouteMetrics` because it will
+	// always contain the status and method labels, idependently of the
```
aryszka, line 105: typo: independently

metrics/codahale.go
aryszka: as there are some frequent repeated string operations here, I would consider a microoptimization like this:

```
var keyServeRoute, keyServeHost string
hfk := hostForKey(host)
switch {
case c.options.EnableServeMethodMetric && c.options.EnableServeStatusCodeMetric:
    keyServeHost = fmt.Sprintf("servehost.%s.%s.%d", hfk, method, code)
    keyServeRoute = fmt.Sprintf("serveroute.%s.%s.%d", hfk, method, code)
case c.options.EnableServeMethodMetric:
    keyServeHost = fmt.Sprintf("servehost.%s.%s", hfk, method)
    keyServeRoute = fmt.Sprintf("serveroute.%s.%s", hfk, method)
case c.options.EnableServeStatusCodeMetric:
    keyServeHost = fmt.Sprintf("servehost.%s.%d", hfk, code)
    keyServeRoute = fmt.Sprintf("serveroute.%s.%d", hfk, code)
default:
    keyServeHost = fmt.Sprintf("servehost.%s", hfk)
    keyServeRoute = fmt.Sprintf("serveroute.%s", hfk)
}
```

though it's true that the current approach might be easier to read

jonathanbeber: I like this idea, @aryszka, I just had to change it slightly to include the routes ones:
```diff
- keyServeRoute = fmt.Sprintf("serveroute.%s.%s.%d", hfk, method, code)
+ keyServeRoute = fmt.Sprintf("serveroute.%s.%s.%d", routeId, method, code)
```

aryszka (approved):

aryszka:
:+1:

metrics/prometheus.go
```diff
@@ -384,12 +409,28 @@ func (p *Prometheus) MeasureServe(routeID, host, method string, code int, start
 	method = measuredMethod(method)
 	t := p.sinceS(start)
 
-	if p.opts.EnableServeRouteMetrics {
-		p.serveRouteM.WithLabelValues(fmt.Sprintf("%d", code), method, routeID).Observe(t)
+	if p.opts.EnableServeRouteMetrics || p.opts.EnableServeHostMetrics {
```
AlexanderYastrebov, line 412: I think this check may be avoided by utilizing nil slice trick like
```diff
-       if p.opts.EnableServeRouteMetrics || p.opts.EnableServeHostMetrics {
-               metrics := []string{}
-               if p.opts.EnableServeStatusCodeMetric {
-                       metrics = append(metrics, fmt.Sprint(code))
-               }
-               if p.opts.EnableServeMethodMetric {
-                       metrics = append(metrics, method)
-               }
-               if p.opts.EnableServeRouteMetrics {
-                       p.serveRouteM.WithLabelValues(append(metrics, routeID)...).Observe(t)
-               }
-               if p.opts.EnableServeHostMetrics {
-                       p.serveHostM.WithLabelValues(append(metrics, hostForKey(host))...).Observe(t)
-               }
+       var metrics []string
+       if p.opts.EnableServeStatusCodeMetric {
+               metrics = append(metrics, fmt.Sprint(code))
+       }
+       if p.opts.EnableServeMethodMetric {
+               metrics = append(metrics, method)
+       }
+
+       if p.opts.EnableServeRouteMetrics {
+               p.serveRouteM.WithLabelValues(append(metrics, routeID)...).Observe(t)
+       }
+       if p.opts.EnableServeHostMetrics {
+               p.serveHostM.WithLabelValues(append(metrics, hostForKey(host))...).Observe(t)
        }
```

metrics/codahale.go
```diff
@@ -178,14 +176,34 @@ func (c *CodaHale) MeasureResponse(code int, method string, routeId string, star
...
 
 func (c *CodaHale) MeasureServe(routeId, host, method string, code int, start time.Time) {
+	if !(c.options.EnableServeRouteMetrics || c.options.EnableServeHostMetrics) {
+		return
+	}
+
+	var keyServeRoute, keyServeHost string
 	method = measuredMethod(method)
+	hfk := hostForKey(host)
+	switch {
```
AlexanderYastrebov, line 186: Here it may be similar to prometheus using nil slice trick and strings.Join:
```diff
-       var keyServeRoute, keyServeHost string
-       method = measuredMethod(method)
-       hfk := hostForKey(host)
-       switch {
-       case c.options.EnableServeMethodMetric && c.options.EnableServeStatusCodeMetric:
-               keyServeHost = fmt.Sprintf("servehost.%s.%s.%d", hfk, method, code)
-               keyServeRoute = fmt.Sprintf("serveroute.%s.%s.%d", routeId, method, code)
-       case c.options.EnableServeMethodMetric:
-               keyServeHost = fmt.Sprintf("servehost.%s.%s", hfk, method)
-               keyServeRoute = fmt.Sprintf("serveroute.%s.%s", routeId, method)
-       case c.options.EnableServeStatusCodeMetric:
-               keyServeHost = fmt.Sprintf("servehost.%s.%d", hfk, code)
-               keyServeRoute = fmt.Sprintf("serveroute.%s.%d", routeId, code)
-       default:
-               keyServeHost = fmt.Sprintf("servehost.%s", hfk)
-               keyServeRoute = fmt.Sprintf("serveroute.%s", routeId)
+       var metrics []string
+       if c.options.EnableServeStatusCodeMetric {
+               metrics = append(metrics, fmt.Sprint(code))
+       }
+       if c.options.EnableServeMethodMetric {
+               metrics = append(metrics, measuredMethod(method))
        }
 
        if c.options.EnableServeRouteMetrics {
-               c.measureSince(keyServeRoute, start)
+               c.measureSince(strings.Join(append([]string{"serveroute", routeId}, metrics...), "."), start)
        }
 
        if c.options.EnableServeHostMetrics {
-               c.measureSince(keyServeHost, start)
+               c.measureSince(strings.Join(append([]string{"servehost", hostForKey(host)}, metrics...), "."), start)
        }
```

metrics/prometheus.go
jonathanbeber: if we do that it will be the situation it was before. [You mentioned](https://github.com/zalando/skipper/pull/1755#discussion_r614323607) that it would "be executed even when `EnableServeRouteMetrics` or/and `EnableServeHostMetrics` are disabled."

AlexanderYastrebov: Right, we definitely have a conflict between readability and peanut savings, will resolve :)

metrics/codahale.go
jonathanbeber: I think this way will be very similar to the way it was before. I'm fine with both solutions at this moment and I don't think any of them will have a huge increase in performance. I will let it for you and @aryszka decide.

aryszka:
:+1:

jonathanbeber:
:+1:

Issue #1756, opened 2021-04-16
There are security vulnerabilities in the go module dependencies


AClarkie:
**Describe the bug**
There are security vulnerabilities in dependencies

**To Reproduce**
Use any security scanning tool for example `snyk test` and view the output

**Expected behavior**

No security vulnerabilities in dependent packages. Could we upgrade the versions of `tidwall/gjson`, `go-yaml/yaml`, `yookoala/gofast`, `tidwall/match` and `miekg/dns` to later versions to resolve these as described below

**Observed behavior**

```
✗ Medium severity vulnerability found in github.com/tidwall/gjson
  Description: Regular Expression Denial of Service (ReDoS)
  Info: https://snyk.io/vuln/SNYK-GOLANG-GITHUBCOMTIDWALLGJSON-1055821
  Introduced through: github.com/zalando/skipper@0.13.52
  From: github.com/zalando/skipper@0.13.52 > github.com/zalando/skipper/filters/builtin@0.13.52 > github.com/zalando/skipper/filters/auth@0.13.52 > github.com/tidwall/gjson@1.4.0
  Fixed in: 1.6.5

✗ Medium severity vulnerability found in github.com/tidwall/gjson
  Description: Denial of Service (DoS)
  Info: https://snyk.io/vuln/SNYK-GOLANG-GITHUBCOMTIDWALLGJSON-1055822
  Introduced through: github.com/zalando/skipper@0.13.52
  From: github.com/zalando/skipper@0.13.52 > github.com/zalando/skipper/filters/builtin@0.13.52 > github.com/zalando/skipper/filters/auth@0.13.52 > github.com/tidwall/gjson@1.4.0
  Fixed in: 1.6.6

✗ Medium severity vulnerability found in github.com/tidwall/gjson
  Description: Denial of Service (DoS)
  Info: https://snyk.io/vuln/SNYK-GOLANG-GITHUBCOMTIDWALLGJSON-1056415
  Introduced through: github.com/zalando/skipper@0.13.52
  From: github.com/zalando/skipper@0.13.52 > github.com/zalando/skipper/filters/builtin@0.13.52 > github.com/zalando/skipper/filters/auth@0.13.52 > github.com/tidwall/gjson@1.4.0
  Fixed in: 1.6.6

✗ Medium severity vulnerability found in github.com/go-yaml/yaml
  Description: Denial of Service (DoS)
[...]

PR #1759, opened 2021-04-16
Update dependencies

mikkeloscar:
Update dependencies to address security vulnerabilities in modules as reported in #1756 

Also updates the import path for `gopkg.in/yaml.v2` (was `github.com/go-yaml/yaml` in some places which go modules could not understand when trying to update)


szuecs:
Can you please add also the tool to our CI pipeline, such that we see it in the future?

mikkeloscar:
We could add `snyk` like this: https://github.com/snyk/actions/tree/master/golang however it depends on a token so we need to register a user with snyk.io We can maybe use the teapot robot for this, WDYT? I could try to set it up next week, but would do it in a separate PR.

szuecs:
Sounds great to me, if you can try to do it.

szuecs:
👍

jonathanbeber:
:+1:

PR #1761, opened 2021-04-17
Supports whitespaces in oidcClaimsQuery strings

### LINKED ISSUE #1760: oidcClaimsQuery filter does not work properly with strings containing spaces
Filter with queries containing spaces do not apply properly.

Example:
`oidcClaimsQuery("/login:groups.#[==\"appX Tester\"]")` returns 200 regardless of any content of groups in id_token.

This is probably because of split function using spaces here:
https://github.com/zalando/skipper/blob/b1981c01bfd58b807dde5349b2d3110fe3ecc18d/filters/auth/oidc_introspection.go#L70

IMHO it's valid use case to use string literals with  spaces.

AlexanderYastrebov:
Fixes #1760

### DIFF
--- a/filters/auth/oidc_introspection.go
+++ b/filters/auth/oidc_introspection.go
@@ -67,7 +67,7 @@ func (spec *oidcIntrospectionSpec) CreateFilter(args []interface{}) (filters.Fil
 				return nil, fmt.Errorf("%v: malformatted filter arg %s", filters.ErrInvalidFilterParameters, arg)
 			}
 			pq := pathQuery{path: slice[0]}
-			for _, query := range strings.Split(slice[1], " ") {
+			for _, query := range splitQueries(slice[1]) {
 				if query == "" {
 					return nil, fmt.Errorf("%v: %s", errUnsupportedClaimSpecified, arg)
 				}
@@ -171,6 +171,18 @@ func (filter *oidcIntrospectionFilter) validateClaimsQuery(reqPath string, gotTo
 
 func (p pathQuery) String() string {
 	return fmt.Sprintf("path: '%s*', matching: %s", p.path, strings.Join(p.queries, " ,"))
+}
+
+// Splits space-delimited GJSON queries ignoring spaces within quoted strings
+func splitQueries(s string) (q []string) {
+	for _, p := range strings.Split(s, " ") {
+		if len(q) == 0 || strings.Count(q[len(q)-1], `"`)%2 == 0 {
+			q = append(q, p)
+		} else {
+			q[len(q)-1] = q[len(q)-1] + " " + p
+		}
+	}
+	return
 }
 
 func trimQuotes(s string) string {
--- a/filters/auth/oidc_introspection_test.go
+++ b/filters/auth/oidc_introspection_test.go
@@ -55,6 +55,20 @@ func TestCreateOIDCQueryClaimsFilter(t *testing.T) {
 			wantErr: assert.NoError,
 		},
 		{
+			name: "one path query with whitespace",
+			args: []interface{}{`/:[@this].#(sub=="white space")`},
+			want: &oidcIntrospectionFilter{
+				typ: checkOIDCQueryClaims,
+				paths: []pathQuery{
+					{
+						path:    "/",
+						queries: []string{`[@this].#(sub=="white space")`},
+					},
+				},
+			},
+			wantErr: assert.NoError,
+		},
+		{
 			name: "several path queries",
 			args: []interface{}{
 				"/some/path:[@this].#(sub==\"somesub\")",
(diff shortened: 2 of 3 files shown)


filters/auth/oidc_introspection.go
```diff
@@ -173,6 +173,18 @@ func (p pathQuery) String() string {
...
+	for _, p := range strings.Split(s, " ") {
+		if len(q) == 0 || strings.Count(q[len(q)-1], `"`)%2 == 0 {
+			q = append(q, p)
+		} else {
+			q[len(q)-1] = q[len(q)-1] + " " + p
+		}
+	}
+	return
+}
+
```
AlexanderYastrebov, line 187: It is better then simple `strings.Split(s, " ")` but will not work properly with escaped quote within string, e.g. `/asdf/:groups.#(%"wh\"ite space")`

AlexanderYastrebov: The split by whitespace could be avoided if GJSON would support OR conditions which it does not https://github.com/tidwall/gjson/issues/162
We could also allow having the same path multiple times in the filter list to emulate OR, unfortunately whitespace-separated queries for the same path were chosen and probably we should stick to it for backward compatibility.

Alternative implementation could use query validation but there is no such utility https://github.com/tidwall/gjson/issues/180

We could adjust quote count for the number of escaped quotes like
```go
last := q[len(q)-1]
unescapedQuotes := strings.Count(last, `"`) - strings.Count(last, `\"`)
```
but it seems GJSON does not handle quotes properly https://github.com/tidwall/gjson/issues/215 so maybe this simplistic approach is good enough until the need for the values with quotes arises.

AlexanderYastrebov: > We could adjust quote count for the number of escaped quotes

This is also an ad-hoc as it would fail on escaped backslash like e.g. in `/asdf/:groups.#(%"wh\\") groups.#(%"whatever")`

https://github.com/tidwall/gjson/issues/215 was fixed in the meantime

AlexanderYastrebov: In the meantime https://github.com/tidwall/gjson/issues/215 was fixed so it supports escaped sequences in the query now.

With that an alternative workaround may be to update library and use escape sequence for the whitespace like `oidcClaimsQuery("/login:groups.#[==\"appX\\u0020Tester\"]")` which could be tested in the playground https://gjson.dev/ with `groups.#[=="appX\u0020Tester"]` query and 
```json
{
  "groups": ["X", "appX Tester"]
}
```
document

aryszka: this looks good to me, and the workarounds using whitespace codes is also fine. I am thinking though the case about escaped quote signs. That would probably require iterating through the runes instead of using strings.Split. Let me check how that would look

aryszka: it would look something like this:

```Go
// Splits space-delimited GJSON queries ignoring spaces within quoted strings
// and applies trimming.
func splitQueries(s string) (q []string) {
	var (
		query           []rune
		quoted, escaped bool
	)

	for _, r := range []rune(s) {
		if escaped {
			escaped = false
			query = append(query, r)
			continue
		}

		if quoted {
			switch r {
			case '\\':
				escaped = true
			case '"':
				quoted = false
			}

			query = append(query, r)
			continue
		}

		if len(query) > 0 {
			if unicode.IsSpace(r) {
				q = append(q, string(query))
				query = nil
				continue
			}

			if r == '"' {
				quoted = true
			}

			query = append(query, r)
			continue
		}

		if unicode.IsSpace(r) {
			continue
		}

		if r == '"' {
			quoted = true
		}

		query = append(query, r)
	}

	if len(query) > 0 {
		q = append(q, string(query))
	}

	return q
}
```

aryszka:
:+1:

filters/auth/oidc_introspection.go
AlexanderYastrebov: It is interesting that space is permitted within array selector outside of the string literal, i.e. `groups.#[ == "X" ]` is a valid query that will stumble upon out split code both existing and proposed.

aryszka: valid point :+1:

szuecs:
What about https://pkg.go.dev/strings#FieldsFunc ?

szuecs:
👍

AlexanderYastrebov:
@szuecs 
> What about https://pkg.go.dev/strings#FieldsFunc ?

There core issue is how to identify split points which is not possible without proper query tokenizer support. 
This PR does best effort trading correctness for simplicity and assumes queries do not contain white spaces outside of quoted strings and quoted strings do not contain escaped quotes.

AlexanderYastrebov:
> trading correctness for simplicity

to be precise - it is not possible to split space-delimited queries correctly in general case as array selector may contain spaces outside of quoted strings https://github.com/zalando/skipper/pull/1761#discussion_r616792850

aryszka:
let's use this now

PR #1770, opened 2021-05-10
Inlines default config values and usage texts

AlexanderYastrebov:
This change inlines default config values and usage texts such that e.g.
```go
flag.DurationVar(&cfg.WriteTimeoutServer, "write-timeout-server", defaultWriteTimeoutServer, writeTimeoutServerUsage)
```
becomes
```go
flag.DurationVar(&cfg.WriteTimeoutServer, "write-timeout-server", 60*time.Second, "set WriteTimeout for http server connections")
```
which allows reader to see config key, its default value and usage at one place as well as enables grep and simpler referencing in github.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -232,236 +232,8 @@ type Config struct {
 }
 
 const (
-	// generic:
-	defaultAddress                         = ":9090"
-	defaultExpectedBytesPerRequest         = 50 * 1024 // 50kB
-	defaultEtcdPrefix                      = "/skipper"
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

AlexanderYastrebov:
This merge can wait, I would like to have @aryszka opinion aligned

aryszka:
thx! I absolutely agree with the change. I didn't anticipate it all at the time when there were only six flags :)

aryszka:
:+1:

PR #1776, opened 2021-05-20
Make RemoteWachOptions fields public

xavivars:
Making RemoteWatchOptions public

### DIFF
--- a/eskipfile/remote.go
+++ b/eskipfile/remote.go
@@ -49,15 +49,10 @@ func RemoteWatch(o *RemoteWatchOptions) (routing.DataClient, error) {
 		return nil, err
 	}
 
-	t := 0
-	if o.Threshold > 0 {
-		t = o.Threshold
-	}
-
 	dataClient := &remoteEskipFile{
 		remotePath: o.RemoteFile,
 		localPath:  tempFilename.Name(),
-		threshold:  t,
+		threshold:  o.Threshold,
 		verbose:    o.Verbose,
 	}
 
@@ -123,7 +118,7 @@ func (client *remoteEskipFile) LoadUpdate() ([]*eskip.Route, []string, error) {
 			log.Infof("New routes were loaded. New: %d; deleted: %d", len(newRoutes), len(deletedRoutes))
 
 			if client.threshold > 0 {
-				if len(newRoutes) > client.threshold || len(deletedRoutes) > client.threshold {
+				if len(newRoutes)+len(deletedRoutes) > client.threshold {
 					log.Warnf("Significant amount of routes was updated. New: %d; deleted: %d", len(newRoutes), len(deletedRoutes))
 				}
 			}
--- a/eskipfile/remote.go
+++ b/eskipfile/remote.go
@@ -23,17 +23,17 @@ type remoteEskipFile struct {
 }
 
 type RemoteWatchOptions struct {
-	remoteFile    string
-	threshold     int
-	verbose       bool
-	failOnStartup bool
+	RemoteFile    string
+	Threshold     int
+	Verbose       bool
+	FailOnStartup bool
 }
 
 // RemoteWatch creates a route configuration client with (remote) file watching. Watch doesn't follow file system nodes,
 // it always reads (or re-downloads) from the file identified by the initially provided file name.
 func RemoteWatch(o *RemoteWatchOptions) (routing.DataClient, error) {
-	if !isFileRemote(o.remoteFile) {
-		return Watch(o.remoteFile), nil
+	if !isFileRemote(o.RemoteFile) {
+		return Watch(o.RemoteFile), nil
 	}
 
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

eskipfile/remote.go
```diff
@@ -23,17 +23,17 @@ type remoteEskipFile struct {
...
 
 type RemoteWatchOptions struct {
-	remoteFile    string
-	threshold     int
-	verbose       bool
-	failOnStartup bool
+	RemoteFile    string
+	Threshold     int
+	Verbose       bool
+	FailOnStartup bool
```
AlexanderYastrebov, line 29: @xavivars Thank you for the improvement. Making this fields public makes them part of the module interface (and its not very useful when private) and therefore it would be nice to have a onliner doc comment for each field like e.g. https://github.com/zalando/skipper/blob/9c3be7ba13cb3520e07cd419dc971b2521951bda/proxy/proxy.go#L142-L148

xavivars: Resolved!

eskipfile/remote.go
```diff
@@ -23,17 +23,24 @@ type remoteEskipFile struct {
...
-	verbose       bool
-	failOnStartup bool
+	// URL of the route file
+	RemoteFile string
+
+	// Verbose mode for the dataClient
+	Verbose bool
+
+	// Amount of route changes that will trigger logs after route updates
+	Threshold int
```
AlexanderYastrebov, line 33: While we are on it, is this knob really necessary since it logs update sizes anyway?
https://github.com/zalando/skipper/blob/960ef409f56d38b13c2af2fb44f95751cc783d1e/eskipfile/remote.go#L110-L118

Also `if client.threshold >= 0` condition means it will warn-log even for default zero threshold which feels strange. I think the condition should be `> 0` but if we can get rid of unnecessary knob it would be the best. My concern is that once we make it public there is no way back.

Here https://github.com/zalando/skipper/blob/960ef409f56d38b13c2af2fb44f95751cc783d1e/eskipfile/remote.go#L91-L93
I think log level should be `Infof`

xavivars: I've modified the last to `Infof`, and also changed the condition to be `0 >' (with safer initialization).

The reason for the Warning to be there is because if suddenly there's a big jump on changed routes, we want to notice immediately (keep it mind this is downloading route files from a remote location, and you may even want to have certain monitoring connected to those warnings.

AlexanderYastrebov: > and you may even want to have certain monitoring connected to those warnings.

This is a good point, we may think (within separate scope) about exposing added/removed routes metric if that is interesting

AlexanderYastrebov:
:+1:

eskipfile/remote.go
```diff
@@ -42,14 +49,19 @@ func RemoteWatch(o *RemoteWatchOptions) (routing.DataClient, error) {
...
+		t = o.Threshold
+	}
+
 	dataClient := &remoteEskipFile{
-		remotePath: o.remoteFile,
+		remotePath: o.RemoteFile,
 		localPath:  tempFilename.Name(),
-		threshold:  o.threshold,
-		verbose:    o.verbose,
+		threshold:  t,
```
szuecs, line 60: Why not just `o.Threshold`?
It defaults to 0 and here we do not need to be that safe in case of negative numbers or do you expect someone to set -1?

eskipfile/remote.go
```diff
@@ -110,7 +122,7 @@ func (client *remoteEskipFile) LoadUpdate() ([]*eskip.Route, []string, error) {
 		if client.verbose {
 			log.Infof("New routes were loaded. New: %d; deleted: %d", len(newRoutes), len(deletedRoutes))
 
-			if client.threshold >= 0 {
+			if client.threshold > 0 {
```
szuecs, line 125: This means 0 and negative numbers  disable logging so the "safety guard makes less sense. :)

eskipfile/remote.go
```diff
@@ -110,7 +122,7 @@ func (client *remoteEskipFile) LoadUpdate() ([]*eskip.Route, []string, error) {
...
 				if len(newRoutes) > client.threshold || len(deletedRoutes) > client.threshold {
```
szuecs, line 126: What do you think about:
```
if len(newRoutes) + len(deletedRoutes) > client.threshold 
```

eskipfile/remote.go
xavivars: I was doing it just as a validation, but happy to remove it if you think it's not needed.

eskipfile/remote.go
xavivars: makes a lot of sense, I'll change it.

szuecs:
:+1:

AlexanderYastrebov:
:lgtm:

PR #1778, opened 2021-05-21
Predicates for Forwarded Header

xavivars:
Two predicates have been added:
* `ForwardedHeader`
* `ForwardedProtocol`.

They follow https://datatracker.ietf.org/doc/html/rfc7239

### DIFF
--- a/docs/reference/predicates.md
+++ b/docs/reference/predicates.md
@@ -128,6 +128,44 @@ Host(/^my-host-header\.example\.org$/)
 Host(/header\.example\.org$/)
 ```
 
+## Forwarded header predicates
+
+Uses standardized Forwarded header ([RFC 7239](https://tools.ietf.org/html/rfc7239))
+
+More info about the header: [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded)
+
+### ForwardedHost
+
+Regular expressions that the forwarded host header in the request must match.
+
+Parameters:
+
+* Host (regex)
+
+Examples:
+
+```
+ForwardedHost(/^my-host-header\.example\.org$/)
+ForwardedHost(/header\.example\.org$/)
+```
+
+### ForwardedProtocol
+
+Regular expressions that the forwarded host header in the request must match.
+
+Parameters:
+
+* Protocol (string)
+
+Only "http" and "https" values are allowed
+
+Examples:
+
+```
+ForwardedProtocol("http")
+ForwardedProtocol("https")
+```
+
 ## Weight (priority)
 
 By default, the weight (priority) of a route is determined by the number of defined predicates.
--- a/docs/reference/predicates.md
+++ b/docs/reference/predicates.md
@@ -151,7 +151,7 @@ ForwardedHost(/header\.example\.org$/)
 
 ### ForwardedProtocol
 
-Regular expressions that the forwarded host header in the request must match.
+Protocol the forwarded header in the request must match.
 
 Parameters:
 
--- a/predicates/forwarded/forwarded.go
+++ b/predicates/forwarded/forwarded.go
@@ -1,0 +1,156 @@
+/*
+Package forwarded implements a set of custom predicate to match routes
+based on the standardized Forwarded header.
+
+https://datatracker.ietf.org/doc/html/rfc7239
+https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded
+
+Examples:
+
+    // only match requests to "example.com"
+    example1: ForwardedHost("example.com") -> "http://example.org";
+
+    // only match requests to http
+    example2: ForwardedProtocol("http") -> "http://example.org";
+
+    // only match requests to https
+    example3: ForwardedProtocol("https") -> "http://example.org";
+*/
+package forwarded
+
+import (
+	"github.com/zalando/skipper/predicates"
+	"github.com/zalando/skipper/routing"
+	"net/http"
+	"regexp"
+	"strings"
+)
+
+const (
+	NameHost  = "ForwardedHost"
+	NameProto = "ForwardedProtocol"
+)
+
+const (
+	HTTP  string = "http"
+	HTTPS string = "https"
+)
+
+type hostPredicateSpec struct{}
+
+type protoPredicateSpec struct{}
+
+type hostPredicate struct {
+	host *regexp.Regexp
+}
+
+func (p *hostPredicateSpec) Create(args []interface{}) (routing.Predicate, error) {
+
+	if len(args) != 1 {
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+
+	value, ok := args[0].(string)
+	if !ok {
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+
+	if value == "" {
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+
+	re, err := regexp.Compile(value)
+	if err != nil {
+		return nil, err
+	}
+
+	return hostPredicate{host: re}, err
+}
+
+type protoPredicate struct {
+	proto string
+}
+
+func (p *protoPredicateSpec) Create(args []interface{}) (routing.Predicate, error) {
+
+	if len(args) != 1 {
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+
+	value, ok := args[0].(string)
+	if !ok {
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+
+	value = strings.ToLower(value)
+
+	switch value {
+	case HTTP, HTTPS:
+	default:
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+
+	return protoPredicate{proto: value}, nil
+}
+
+func NewForwardedHost() routing.PredicateSpec  { return &hostPredicateSpec{} }
+func NewForwardedProto() routing.PredicateSpec { return &protoPredicateSpec{} }
+
+func (p *hostPredicateSpec) Name() string {
+	return NameHost
+}
+
+func (p *protoPredicateSpec) Name() string {
+	return NameProto
+}
+
+func (p hostPredicate) Match(r *http.Request) bool {
+
+	fh := r.Header.Get("Forwarded")
+
+	if fh == "" {
+		return false
+	}
+
+	fw := parseForwarded(fh)
+
+	return p.host.MatchString(fw.host)
+}
+
+func (p protoPredicate) Match(r *http.Request) bool {
+
+	fh := r.Header.Get("Forwarded")
+
+	if fh == "" {
+		return false
+	}
+
+	fw := parseForwarded(fh)
+
+	return p.proto == fw.proto
+}
+
+type forwarded struct {
+	host  string
+	proto string
+}
+
+func parseForwarded(fh string) *forwarded {
+
+	f := &forwarded{}
+
+	for _, forwardedPair := range strings.Split(fh, ";") {
+		if tv := strings.SplitN(forwardedPair, "=", 2); len(tv) == 2 {
+			token, value := tv[0], tv[1]
+			token = strings.TrimSpace(token)
+			value = strings.TrimSpace(strings.Trim(value, `"`))
+			switch strings.ToLower(token) {
+			case "proto":
+				f.proto = value
+			case "host":
+				f.host = value
+			}
+		}
+	}
+	return f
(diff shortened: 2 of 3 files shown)


docs/reference/predicates.md
```diff
@@ -128,6 +128,44 @@ Host(/^my-host-header\.example\.org$/)
...
+Regular expressions that the forwarded host header in the request must match.
+
+Parameters:
+
+* Host (regex)
+
+Examples:
+
+```
+ForwardedHost(/^my-host-header\.example\.org$/)
```
AlexanderYastrebov, line 148: These predicates are equivalent to 
* `HeaderRegexp("Forwarded", "host=example\.com(;|$)")`
* `HeaderRegexp("Forwarded", "proto=http(;|$)")`
from what I can tell.

docs/reference/predicates.md
```diff
@@ -128,6 +128,44 @@ Host(/^my-host-header\.example\.org$/)
...
+
+### ForwardedProtocol
+
+Regular expressions that the forwarded host header in the request must match.
+
+Parameters:
+
+* Protocol (string)
+
+Only "http" and "https" values are allowed
```
AlexanderYastrebov, line 160: Looks like there is a contradiction between regexp and a set of predefined values

predicates/forwarded/forwarded.go
```diff
@@ -0,0 +1,156 @@
...
+
+const (
+	NameHost  = "ForwardedHost"
+	NameProto = "ForwardedProtocol"
+)
+
+const (
+	HTTP  string = "http"
+	HTTPS string = "https"
+)
```
AlexanderYastrebov, line 37: Is there a need for this public constants, maybe just inline them?

predicates/forwarded/forwarded.go
```diff
@@ -0,0 +1,156 @@
...
+			token = strings.TrimSpace(token)
+			value = strings.TrimSpace(strings.Trim(value, `"`))
+			switch strings.ToLower(token) {
+			case "proto":
+				f.proto = value
+			case "host":
+				f.host = value
+			}
+		}
+	}
```
AlexanderYastrebov, line 154: An alternative implementation could be to transform predicate argument into regexp and match the header value against it. Since regexp is used anyway we can avoid parsing the structure altogether.

predicates/forwarded/forwarded_test.go
```diff
@@ -0,0 +1,308 @@
...
+		isError bool
+	}{{
+		msg:     "Empty host should fail",
+		host:    "",
+		r:       request{},
+		matches: false,
+		isError: true,
+	}, {
+		msg:     "Empty Forwarded Header should not match",
+		host:    "example.com",
```
AlexanderYastrebov, line 29: IMO we should be more precise in the test examples as they also serve as a documentation. 
`example.com` regexp also matches `host=example.comma.org` as far as I can tell.

predicates/forwarded/forwarded_test.go
```diff
@@ -0,0 +1,308 @@
...
+		r:       request{},
+		matches: false,
+		isError: false,
+	}, {
+		msg:  "Same forwarded Header should match",
+		host: "example.com",
```
AlexanderYastrebov, line 35: See https://github.com/zalando/skipper/pull/1778/files#r636904660

predicates/forwarded/forwarded_test.go
```diff
@@ -0,0 +1,308 @@
...
+		r: request{
+			url: "https://myproxy.com/index.html",
+			headers: http.Header{
+				"Forwarded": []string{"for=192.0.2.60;proto=http;host=example.com;by=203.0.113.43"},
+			},
+		},
+		matches: true,
+		isError: false,
+	}, {
+		msg:  "Same forwarded Header should match",
```
AlexanderYastrebov, line 56: I think this one is redundant as there should be three "Same forwarded Header should match"  where host is at the start, middle and end of the header.

predicates/forwarded/forwarded_test.go
```diff
@@ -0,0 +1,308 @@
...
+			if hasError || tc.isError {
+				if !tc.isError {
+					t.Error("Predicate creation failed")
+				}
+
+				if !hasError {
+					t.Error("Predicate should have failed")
+				}
+
+				return
```
AlexanderYastrebov, line 128: I think if there is no need to continue the test a `t.Fatal` may be used.

predicates/forwarded/forwarded_test.go
```diff
@@ -0,0 +1,308 @@
...
+			}
+
+			r, err := newRequest(tc.r)
+			if !tc.isError && err != nil {
```
AlexanderYastrebov, line 132: I think `!tc.isError` is not needed here. Also if we can not create the request it makes sense to fail with Fatal.

predicates/forwarded/forwarded.go
xavivars: I'll inline them.

predicates/forwarded/forwarded_test.go
xavivars: Great point!

docs/reference/predicates.md
xavivars: Do you think we shouldn't add the predicates? or do you think we should just reuse the ones that exist (even if with a tiny layer on top?)

docs/reference/predicates.md
xavivars: probably too much copy-paste 😅

predicates/forwarded/forwarded_test.go
xavivars: This return is need (it would implcitly be 
```
if hasError && tc.isError {
    return
}
```

docs/reference/predicates.md
AlexanderYastrebov: > Do you think we shouldn't add the predicates? 

It is more of a note for myself, you and reviewers. I am not a fan of regexp hacking and prefer dedicated thing but new predicates also add maintenance/compatibility cost which we have to weight against the benefit.

> Or are you suggesting we should just reuse the ones that exist (even if with a tiny layer on top?)

I am not sure we could reuse them as they are legacy predicates implemented differently say from `Source`.

xavivars: That's much harder than I thoug... If "basic" predicates were exposed like any other predicate, it should be easy to do something like this:
```
func (p *hostPredicateSpec) Create(args []interface{}) (routing.Predicate, error) {
    spec := headerRegexp.New()
    return spec.Create("Forwarded", fmt.Sprintf("host=example\.com(;|$)", args[0])
}
```

predicates/forwarded/forwarded.go
xavivars: I was trying to think about this.... but building everythign into a single regexp may be actually harder:
- How do you combine the hole thing into a single regexp if what you get in the host is something like this?

`(^|www\.|subdomain)\.example\.org$`

The end regexp should not have ^ or $, but `host=` and `($|;)` instead, and that string manipulation could get tricky.

I'll leave it like it is now, we can always change the implementation given that all tests keep passing.

docs/reference/predicates.md
szuecs: This is an interesting case and why we should move the legacy predicates (I would not call them "basic").
IIRC there is no reason for non path related predicates to stay in the eskip package besides breaking compatibility.
Do we see a solution without a breaking change?
I am not really sure if someone had a use of single eskip package and use the legacy predicates from there as exposed in `eskip.Route`

xavivars: `legacy` is definitely a better way to describe them than `basic` :)

szuecs:
:+1:

predicates/forwarded/forwarded.go
```diff
@@ -0,0 +1,150 @@
...
+
+	value = strings.ToLower(value)
+
+	switch value {
+	case "http", "https":
+	default:
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+
+	return protoPredicate{proto: value}, nil
```
AlexanderYastrebov, line 87: minor: I think `ToLower` is redundant - the docs explicitly say it should be lowercase. 
If we change case people eventually would rely on it.
I would also move last return inside the switch:
```go
        switch value {
	case "http", "https":
        	return protoPredicate{proto: value}, nil
	default:
		return nil, predicates.ErrInvalidPredicateParameters
	}
```

predicates/forwarded/forwarded.go
```diff
@@ -0,0 +1,150 @@
...
+
+func parseForwarded(fh string) *forwarded {
+
+	f := &forwarded{}
+
+	for _, forwardedPair := range strings.Split(fh, ";") {
+		if tv := strings.SplitN(forwardedPair, "=", 2); len(tv) == 2 {
+			token, value := tv[0], tv[1]
+			token = strings.TrimSpace(token)
+			value = strings.TrimSpace(strings.Trim(value, `"`))
```
AlexanderYastrebov, line 140: I don't think we need to trim here, RFC https://datatracker.ietf.org/doc/html/rfc7239#section-4 does not allow arbitrary whitespaces.

predicates/forwarded/forwarded.go
```diff
@@ -0,0 +1,150 @@
...
+			switch strings.ToLower(token) {
```
AlexanderYastrebov, line 141: Not sure we need to lower, lest keep in mind this parsing would happen on every request

predicates/forwarded/forwarded_test.go
AlexanderYastrebov: I think it should be `^example\.com$` (note escaped dot) everywhere

predicates/forwarded/forwarded.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/zalando/skipper/routing"
+	"net/http"
+	"regexp"
+	"strings"
+)
+
+const (
+	NameHost  = "ForwardedHost"
+	NameProto = "ForwardedProtocol"
+)
```
AlexanderYastrebov, line 32: Should these be public? If not, lets inline them into `Name()`

predicates/forwarded/forwarded.go
AlexanderYastrebov: Agree, rewriting regexp does not seem right

predicates/forwarded/forwarded.go
szuecs: True and it should be handled by this:
```
func (h Header) Get(key string) string
    Get gets the first value associated with the given key. If there are no
    values associated with the key, Get returns "". It is case insensitive;
    textproto.CanonicalMIMEHeaderKey is used to canonicalize the provided key.
    To use non-canonical keys, access the map directly.
```

predicates/forwarded/forwarded.go
aryszka: Most of them are public. IIRC, the rationale for having them public was to be able to use them when creating predicates programmatically, e.g. in tests. I'd say that this is fine.

aryszka:
@xavivars thanks for the PR! Overall, it looks good to me, but I agree with @AlexanderYastrebov 's comments regarding the trimming and the lowercase conversion.

predicates/forwarded/forwarded_test.go
xavivars: Done!

predicates/forwarded/forwarded.go
xavivars: Not sure `func (h Header) Get(key string) string` handles this. The casing (lowercasing in this case) was for the parameter for the predicate, not for the actual value of the header.

xavivars: Removed the lowercase, and simplified the code as suggested.

predicates/forwarded/forwarded.go
xavivars: This is not removing whitespaces, but `"`. As you can check in https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6, value will usually be a quoted string.

xavivars: I changed the tests so they have the quotes, but also allowed them to be non-quoted, as tchar also allows for non-quoted strings.

predicates/forwarded/forwarded.go
xavivars: Fair point. I'll remove it.

aryszka:
:+1:

PR #1788, opened 2021-06-02
optimize Lua pool size

szuecs:
sets pool size dependent on memory available and keep a max of 1000 (TBD)

### DIFF
--- a/script/script.go
+++ b/script/script.go
@@ -11,6 +11,7 @@ import (
 	"net/url"
 	"os"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -25,18 +26,48 @@ import (
 	gjson "layeh.com/gopher-json"
 )
 
-// InitialPoolSize is the number of lua states created initially per route
-var InitialPoolSize int = 3
+const (
+	defaultPoolSize int = 10
+)
 
-// MaxPoolSize is the number of lua states stored per route - there may be more parallel
-// requests, but only this number is cached.
-var MaxPoolSize int = 10
+type luaScript struct {
+	// PoolSize is the number of lua states stored per route - there may be more parallel
+	// requests, but only this number is cached.
+	PoolSize int
 
-type luaScript struct{}
+	// number of filter instances
+	instances int
 
-// NewLuaScript creates a new filter spec for skipper
+	memoryAvailableBytes int
+}
+
+// NewLuaScript creates a new filter spec for skipper with default pool size
 func NewLuaScript() filters.Spec {
-	return &luaScript{}
+	return &luaScript{
+		PoolSize: defaultPoolSize,
+	}
+}
+
+// WithPoolSize creates a new filter spec for skipper with global pool size set
+func WithPoolSize(bytesAvailable int) filters.Spec {
+	memoryLimit := int(bytesAvailable / 10) // let's use only 1/10 for lua
+
+	// max 1GB
+	if memoryLimit > 1<<30 {
+		memoryLimit = 1 << 30
+	}
+
+	scriptSize := 1024 // assumption (I counted 248B for a request and response filter script, that generates 5B random data)
+	poolSize := memoryLimit / scriptSize
+	if poolSize > 1000 {
+		poolSize = 1000
+	}
+
+	log.Infof("Calculated poolsize: %d", poolSize)
+	return &luaScript{
+		PoolSize:             poolSize,
+		memoryAvailableBytes: memoryLimit,
+	}
 }
 
 // Name returns the name of the filter ("lua")
@@ -62,10 +93,15 @@ func (ls *luaScript) CreateFilter(config []interface{}) (filters.Filter, error) 
 		params = append(params, ps)
 	}
 
-	s := &script{source: src, routeParams: params}
-	if err := s.initScript(); err != nil {
+	s := &script{
+		source:      src,
+		routeParams: params,
+	}
+	if err := s.initScript(ls.PoolSize); err != nil {
 		return nil, err
 	}
+
+	ls.instances += 1
 	return s, nil
 }
 
@@ -75,6 +111,7 @@ func (s *script) getState() (*lua.LState, error) {
 		if L == nil {
 			return nil, errors.New("pool closed")
 		}
+		atomic.AddInt32(&s.idlePool, -1)
 		return L, nil
 	default:
 		return s.newState()
@@ -88,6 +125,7 @@ func (s *script) putState(L *lua.LState) {
 	}
 	select {
 	case s.pool <- L:
+		atomic.AddInt32(&s.idlePool, 1)
 	default: // pool full, close state
 		L.Close()
 	}
@@ -127,7 +165,7 @@ func sleep(L *lua.LState) int {
 	return 0
 }
 
-func (s *script) initScript() error {
+func (s *script) initScript(poolSize int) error {
 	// Compile
 	var reader io.Reader
 	var name string
@@ -176,18 +214,67 @@ func (s *script) initScript() error {
 	}
 
 	// Init state pool
-	s.pool = make(chan *lua.LState, MaxPoolSize) // FIXME make configurable
-	for i := 0; i < InitialPoolSize; i++ {
+	s.initPool(poolSize)
+	return nil
+}
+
+func (s *script) initPool(max int) error {
+	s.pool = make(chan *lua.LState, max)
+	for i := 0; i < max; i++ {
 		L, err := s.newState()
 		if err != nil {
 			return err
 		}
 		s.putState(L)
 	}
+	s.poolSize = max
+	return nil
+}
+
+// hopefully during the operation we do not run out of states at runtime
+func (s *script) resizePool(max int) error {
+	pool := make(chan *lua.LState, max)
+	idle := atomic.LoadInt32(&s.idlePool)
+	cnt := 0
+
+	// migrate half the idle pool
+	for i := int32(0); i < idle/2; i++ {
+		L, err := s.getState()
+		if err != nil {
+			break
+		}
+		s.putState(L)
+		cnt++
+	}
+	// switch pool
+	s.pool = pool
+
+	// migrate other half the idle pool
+	for i := idle / 2; i < idle; i++ {
+		L, err := s.getState()
+		if err != nil {
+			break
+		}
+		s.putState(L)
+		cnt++
+	}
+
+	// create rest
+	for i := cnt; i < max; i++ {
+		L, err := s.newState()
+		if err != nil {
+			return err
+		}
+		s.putState(L)
+	}
+
+	s.poolSize = max
 	return nil
 }
 
 type script struct {
+	idlePool                int32
+	poolSize                int
 	source                  string
 	routeParams             []string
 	pool                    chan *lua.LState
--- a/skipper.go
+++ b/skipper.go
@@ -53,6 +53,7 @@ import (
 	"github.com/zalando/skipper/ratelimit"
 	"github.com/zalando/skipper/routing"
 	"github.com/zalando/skipper/scheduler"
+	"github.com/zalando/skipper/script"
 	"github.com/zalando/skipper/secrets"
 	"github.com/zalando/skipper/swarm"
 	"github.com/zalando/skipper/tracing"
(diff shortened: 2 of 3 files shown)


script/script.go
```diff
@@ -25,18 +26,48 @@ import (
...
 func NewLuaScript() filters.Spec {
-	return &luaScript{}
+	return &luaScript{
+		PoolSize: defaultPoolSize,
+	}
+}
+
+// WithPoolSize creates a new filter spec for skipper with global pool size set
+func WithPoolSize(bytesAvailable int) filters.Spec {
+	memoryLimit := int(bytesAvailable / 10) // let's use only 1/10 for lua
```
aryszka, line 53: I would move this decision to the caller size. The function name and the argument is very misleading this way.

script/script.go
```diff
@@ -25,18 +26,48 @@ import (
...
+
+	// max 1GB
+	if memoryLimit > 1<<30 {
+		memoryLimit = 1 << 30
+	}
+
+	scriptSize := 1024 // assumption (I counted 248B for a request and response filter script, that generates 5B random data)
+	poolSize := memoryLimit / scriptSize
+	if poolSize > 1000 {
+		poolSize = 1000
```
aryszka, line 63: Why do we limit the pool size to 1000 here?

script/script.go
```diff
@@ -25,18 +26,48 @@ import (
...
+type luaScript struct {
+	// PoolSize is the number of lua states stored per route - there may be more parallel
+	// requests, but only this number is cached.
+	PoolSize int
 
-// MaxPoolSize is the number of lua states stored per route - there may be more parallel
-// requests, but only this number is cached.
-var MaxPoolSize int = 10
+	// number of filter instances
+	instances int
```
aryszka, line 39: what do we use this field for? As I see, we're only increment it currently. Let's remove it

script/script.go
```diff
@@ -25,18 +26,48 @@ import (
...
 
-type luaScript struct{}
+	memoryAvailableBytes int
```
aryszka, line 41: As I can see, we don't use this field. Let's remove it

script/script.go
```diff
@@ -176,18 +214,67 @@ func (s *script) initScript() error {
...
+		s.putState(L)
+	}
+	s.poolSize = max
+	return nil
+}
+
+// hopefully during the operation we do not run out of states at runtime
+func (s *script) resizePool(max int) error {
+	pool := make(chan *lua.LState, max)
+	idle := atomic.LoadInt32(&s.idlePool)
```
aryszka, line 237: instead of `atomic.LoadInt32(&s.idlePool)` just use `len(s.pool)`

script/script.go
```diff
@@ -88,6 +125,7 @@ func (s *script) putState(L *lua.LState) {
 	}
 	select {
 	case s.pool <- L:
+		atomic.AddInt32(&s.idlePool, 1)
```
aryszka, line 128: we don't need this `idlePool` field, we can just simply use `len(s.pool)`

script/script.go
```diff
@@ -176,18 +214,67 @@ func (s *script) initScript() error {
...
+		if err != nil {
+			break
+		}
+		s.putState(L)
+		cnt++
+	}
+	// switch pool
+	s.pool = pool
+
+	// migrate other half the idle pool
```
aryszka, line 252: I couldn't figure why we're migrating in two halves. What's the reason?

script/script.go
AlexanderYastrebov: +1, the caller also divides memory limit by 100

AlexanderYastrebov: +1, the caller also divides memory limit by 100

AlexanderYastrebov:
I think we need to better understand how well or not the pool actually behaves before making it more complex. 
E.g. I have added a log statement to see when new state is allocated (which presumably introduces latency)

```diff
diff --git a/script/script.go b/script/script.go
index 4b14682e..68a6d83c 100644
--- a/script/script.go
+++ b/script/script.go
@@ -77,6 +77,7 @@ func (s *script) getState() (*lua.LState, error) {
                }
                return L, nil
        default:
+               log.Info("newState")
                return s.newState()
        }
```
and run `bin/skipper -inline-routes='* -> lua("function request(c,p) sleep(1000) end") -> status(204) -> <shunt>' -access-log-disabled` - a lua script with 1s latency.
Then I put some load via `echo "GET http://localhost:9090/test" | vegeta attack -rate=100/s | vegeta report -every 1s` and indeed observe 1s latency:
```
Duration      [total, attack, wait]             56.981s, 55.98s, 1.001s
Latencies     [min, mean, 50, 90, 95, 99, max]  1s, 1.001s, 1.001s, 1.002s, 1.002s, 1.003s, 1.01s
Bytes In      [total, mean]                     0, 0.00
Status Codes  [code:count]                      204:5599  
Error Set:
```
What I do not understand though is that with 100 rps and 1s latency there should be 100 requests in parallel which with the default state pool size of 10 should create new states but I do not observe that. The pool is implemented as a go channel, maybe there is something interesting goes on when many goroutines read/write to it in parallel. 

The PR introduces some assumptions in form of constants and a bit of magic.
I also think it would be better to compare baseline against the changed version to see if it actually improves.

aryszka:
that's true. Right now this is WIP. We just discussed that we should start with converting the pool/filter instance setup to a shared global pool, where we can access the Lua states via a key, eg a hash of the code.

szuecs:
@AlexanderYastrebov I agree, it's just trying things as a wip and the pool should be bounded and different. We can even pre allocate a generic pool and load the code when script instance is created.
One thing to figure out is how to return to the generic pool. We would need a Close() on the filter and give it back removing the code of the filter. I think we have no Close() that we call if we delete a route for example.

script/script.go
szuecs: To be discussed, rest comments I changed all the code and I should not have opened the PR in this state, sorry for taking time to review :)

AlexanderYastrebov: >  I should not have opened the PR in this state, sorry for taking time to review

No problem at all

szuecs:
In my tests on Friday (code changes not pushed, I could observe more creates at >3500 rps, but my optimization is a slow down and the only thing I would do is to increase the pool size, maybe we should pass the poolsize per filter via create function and passed via cli arg and config.
If you have an instance in the pool it's crazy fast for an interpreted language that we hook in!

AlexanderYastrebov:
> maybe we should pass the pool size per filter

We have reserved parameter names starting with "lua-" [documented](https://github.com/zalando/skipper/blob/master/docs/reference/scripts.md#route-filters)
> Any parameter starting with "lua-" should not be used to pass values for the script - those will be used for configuring the filter.

so technically we may use it to set min and max pool size. Max pool size we may also be capped by startup config value (if/when needed).  The problem with "lua-" filter parameters is that once someone uses it - we would not be able to break it.

Previously I thought min pool size did not make much sense but now I think it does - it allows pool to grow to optimal size wrt request rate pattern on the given route. With this in mind maybe we do not need per-filter config but a global config to set max pool size per filter is enough.

> One thing to figure out is how to return to the generic pool.

We have an unsolved ticket for filter cleanup https://github.com/zalando/skipper/issues/202
We currently do not close states in the pool on route updates. Should not be a big deal unless someone uses `iolib` to [create temp files](https://github.com/yuin/gopher-lua/blob/f4c35e4016d9d8580b007ebaeb68ecd8e0b09f1c/_state.go#L1224-L1233) which are removed on state closing.  

> We just discussed that we should start with converting the pool/filter instance setup to a shared global pool on the spec, where we can access the Lua states via a key, eg a hash of the code.

Sound like global pool should be some kind of LRU cache or bounded priority queue but it is not immediately obvious how would it look like as it should support multiple values per key. What I like about current per-filter pool implementation is how simple it is.

szuecs:
After testing, I am all in with per filter pool. 
Right now it's fixed in size, but maybe we should let it grow. Grow the current pool seems not really possible right now because the size is the buffered channel size and I don't think it is easy to grow. Maybe we should use another data structure for this. Any ideas?

script/script.go
```diff
@@ -62,72 +130,92 @@ func (ls *luaScript) CreateFilter(config []interface{}) (filters.Filter, error)
...
+	L, err := s.globalPool.GetState()
+	if err != nil {
+		return nil, err
+	}
 
+	// put compiled script into the current LState
 	L.Push(L.NewFunctionFromProto(s.proto))
 
-	err := L.PCall(0, lua.MultRet, nil)
+	err = L.PCall(0, lua.MultRet, nil)
```
AlexanderYastrebov, line 172: We need to confirm somehow that it is ok to re-run possibly different pre-compiled scripts over the same lua state. The [docs](https://github.com/yuin/gopher-lua#sharing-lua-byte-code-between-lstates) only mention sharing the same code with many states (1:N) whereas this approach is M:N

script/script.go
```diff
@@ -25,18 +25,86 @@ import (
...
+	sp := &statePool{
+		pool: make(chan *lua.LState, max),
+	}
+	for i := 0; i < max; i++ {
+		state, err := newState()
+		if err != nil {
+			return nil, err
+		}
+		sp.pool <- state
+	}
```
AlexanderYastrebov, line 58: I think we may start with empty pool (or configurable min size) and allow it to grow organically. Assuming max size is passed as a flag this would allow us to set it to some quite high number while real amount of utilized states would be defined by traffic rate shape.

skipper.go
```diff
@@ -1357,6 +1363,14 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 		)
 	}
 
+	if bytes, err := getCgroupV1MemoryBytes(); err != nil {
+		o.CustomFilters = append(o.CustomFilters, script.NewLuaScript())
+	} else {
+		if int(bytes) > 2*(1<<30) {
+			o.CustomFilters = append(o.CustomFilters, script.WithPoolSize(10000))
+		}
+	}
```
AlexanderYastrebov, line 1372: Config option might be a better solution. Skipper operator knows its memory size upfront and we may provide guidance on setting max pool size based on available memory and benchmark numbers.

script/script.go
szuecs: The drawback would be that this could lead to latency spikes, because we run and rely on autoscaling.
Assumption is that this can lead to a latency spike (not 100% sure but in tests it felt a bit like this).

skipper.go
szuecs: yes, this was just the quick hack :)

script/script.go
```diff
@@ -25,18 +25,94 @@ import (
...
+const (
+	defaultPoolSize int = 1000
+	genericStateKey     = "generic"
+)
 
-// MaxPoolSize is the number of lua states stored per route - there may be more parallel
-// requests, but only this number is cached.
-var MaxPoolSize int = 10
+func newState() (*lua.LState, error) {
+	L := lua.NewState(lua.Options{
```
AlexanderYastrebov, line 34: I had this tuning in mind as a followup of #1792 
Maybe we should not change the defaults within this PR to have a fair measure of how global pool and max size improve(or not) the baseline.

script/script.go
szuecs: That's something we need to do before working too much more on this.
In my tests the results looked great.

script/script.go
szuecs: I agree, the current state is just to keep the discussions going on.

config/config.go
```diff
@@ -34,6 +34,7 @@ type Config struct {
 	ExpectedBytesPerRequest         int            `yaml:"expected-bytes-per-request"`
 	MaxTCPListenerConcurrency       int            `yaml:"max-tcp-listener-concurrency"`
 	MaxTCPListenerQueue             int            `yaml:"max-tcp-listener-queue"`
+	LuaPoolSize                     int            `yaml:"lua-pool-size"`
```
AlexanderYastrebov, line 37: LuaStatePoolSize?

script/script.go
```diff
@@ -62,72 +138,92 @@ func (ls *luaScript) CreateFilter(config []interface{}) (filters.Filter, error)
...
+// compile, save and test
+func (s *script) initScript() error {
+	if err := s.compileSource(); err != nil {
+		return err
 	}
-	log.Print(args...)
-	return 0
+
+	// Detect request and response functions
+	L, err := s.newState()
```
AlexanderYastrebov, line 195: For request/response function detection we need to use a new non-pooled state because the state from the pool may already have request/response functions defined

config/config.go

script/script.go
szuecs: I tested with 2 routes that have 2 different lua scripts and load tested, that this works with pooled states.
I think this is also part

szuecs:
How to block Lua VM?
To test growing pool size and how the limiting of number of Lua VMs work, right now vs. with the pre-allocated number of VMs.

PR #1790, opened 2021-06-04
Fix anchor link capitalisation in documentation

k-nut:
The headings are all converted to lowercase when they are turned into anchors so we also need to use lowercase when referencing them as internal skip links.

Signed-off-by: Knut Hühne <knut.huehne.external@zalando.de>

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -121,7 +121,7 @@ Path("/resource/:id") -> setRequestHeader("X-Resource-Id", "${id}") -> clusterCl
 
 ## appendRequestHeader
 
-Same as [setRequestHeader](#setRequestHeader),
+Same as [setRequestHeader](#setrequestheader),
 but appends the provided value to the already existing ones.
 
 ## dropRequestHeader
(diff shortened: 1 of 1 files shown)


k-nut:
I just noticed that this actually applies to a couple more links in this file. I'll amend my commit.

k-nut:
Ok, amended ✅

szuecs:
:+1:

aryszka:
:+1:

AlexanderYastrebov:
As a note camelCase anchors seem to work just fine in Chrome and Firefox, e.g. https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#setRequestHeader
though according to this discussion https://bugzilla.mozilla.org/show_bug.cgi?id=241063 url fragments should be case sensitive.
I guess browsers just implement case-independent comparison.

k-nut:
> As a note camelCase anchors seem to work just fine in Chrome and Firefox, e.g. https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#setRequestHeader
> though according to this discussion https://bugzilla.mozilla.org/show_bug.cgi?id=241063 url fragments should be case sensitive.
> I guess browsers just implement case-independent comparison.

Huh, interesting. It looks to me like this is actually GitHub doing some frontend magic. For the rendered documentation https://opensource.zalando.com/skipper/reference/filters/#setRequestHeader doesn't work whereas https://opensource.zalando.com/skipper/reference/filters/#setrequestheader does. TIL...

AlexanderYastrebov:
@k-nut Nice, I did not think about js magic. #1791 fixes the rest of the anchors.

PR #1795, opened 2021-06-17
add ratelimit type as Tag to Span

szuecs:
add ratelimit type as Tag to Span, such that we can see what type was used and search


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #1803, opened 2021-06-29
redis ring client hashing algorithms

szuecs:
expose hashing algorithm of redis ring client and have extra implementations for rendezvous with vnodes, jump and multi probe consistent hash, defaults to rendezvous without vnodes

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -435,7 +435,7 @@ func NewConfig() *Config {
 	// Swarm:
 	flag.BoolVar(&cfg.EnableSwarm, "enable-swarm", false, "enable swarm communication between nodes in a skipper fleet")
 	flag.Var(cfg.SwarmRedisURLs, "swarm-redis-urls", "Redis URLs as comma separated list, used for building a swarm, for example in redis based cluster ratelimits.\nUse "+redisPasswordEnv+" environment variable or 'swarm-redis-password' key in config file to set redis password")
-	flag.StringVar(&cfg.SwarmRedisHashAlgorithm, "swarm-redis-hash-algorithm", "", "sets hash algorithm to be used in redis ring client to find the shard <jump|mpchash|rendezvousVnodes>, defaults to library default rendezvous")
+	flag.StringVar(&cfg.SwarmRedisHashAlgorithm, "swarm-redis-hash-algorithm", "", "sets hash algorithm to be used in redis ring client to find the shard <jump|mpchash|rendezvous|rendezvousVnodes>, defaults to github.com/go-redis/redis default")
 	flag.DurationVar(&cfg.SwarmRedisDialTimeout, "swarm-redis-dial-timeout", net.DefaultDialTimeout, "set redis client dial timeout")
 	flag.DurationVar(&cfg.SwarmRedisReadTimeout, "swarm-redis-read-timeout", net.DefaultReadTimeout, "set redis socket read timeout")
 	flag.DurationVar(&cfg.SwarmRedisWriteTimeout, "swarm-redis-write-timeout", net.DefaultWriteTimeout, "set redis socket write timeout")
--- a/config/config.go
+++ b/config/config.go
@@ -212,14 +212,15 @@ type Config struct {
 	// swarm:
 	EnableSwarm bool `yaml:"enable-swarm"`
 	// redis based
-	SwarmRedisURLs         *listFlag     `yaml:"swarm-redis-urls"`
-	SwarmRedisPassword     string        `yaml:"swarm-redis-password"`
-	SwarmRedisDialTimeout  time.Duration `yaml:"swarm-redis-dial-timeout"`
-	SwarmRedisReadTimeout  time.Duration `yaml:"swarm-redis-read-timeout"`
-	SwarmRedisWriteTimeout time.Duration `yaml:"swarm-redis-write-timeout"`
-	SwarmRedisPoolTimeout  time.Duration `yaml:"swarm-redis-pool-timeout"`
-	SwarmRedisMinConns     int           `yaml:"swarm-redis-min-conns"`
-	SwarmRedisMaxConns     int           `yaml:"swarm-redis-max-conns"`
+	SwarmRedisURLs          *listFlag     `yaml:"swarm-redis-urls"`
+	SwarmRedisPassword      string        `yaml:"swarm-redis-password"`
+	SwarmRedisHashAlgorithm string        `yaml:"swarm-redis-hash-algorithm"`
+	SwarmRedisDialTimeout   time.Duration `yaml:"swarm-redis-dial-timeout"`
+	SwarmRedisReadTimeout   time.Duration `yaml:"swarm-redis-read-timeout"`
+	SwarmRedisWriteTimeout  time.Duration `yaml:"swarm-redis-write-timeout"`
+	SwarmRedisPoolTimeout   time.Duration `yaml:"swarm-redis-pool-timeout"`
+	SwarmRedisMinConns      int           `yaml:"swarm-redis-min-conns"`
+	SwarmRedisMaxConns      int           `yaml:"swarm-redis-max-conns"`
 	// swim based
 	SwarmKubernetesNamespace          string        `yaml:"swarm-namespace"`
 	SwarmKubernetesLabelSelectorKey   string        `yaml:"swarm-label-selector-key"`
@@ -434,6 +435,7 @@ func NewConfig() *Config {
 	// Swarm:
 	flag.BoolVar(&cfg.EnableSwarm, "enable-swarm", false, "enable swarm communication between nodes in a skipper fleet")
 	flag.Var(cfg.SwarmRedisURLs, "swarm-redis-urls", "Redis URLs as comma separated list, used for building a swarm, for example in redis based cluster ratelimits.\nUse "+redisPasswordEnv+" environment variable or 'swarm-redis-password' key in config file to set redis password")
+	flag.StringVar(&cfg.SwarmRedisHashAlgorithm, "swarm-redis-hash-algorithm", "", "sets hash algorithm to be used in redis ring client to find the shard <jump|mpchash|rendezvousVnodes>, defaults to library default rendezvous")
 	flag.DurationVar(&cfg.SwarmRedisDialTimeout, "swarm-redis-dial-timeout", net.DefaultDialTimeout, "set redis client dial timeout")
 	flag.DurationVar(&cfg.SwarmRedisReadTimeout, "swarm-redis-read-timeout", net.DefaultReadTimeout, "set redis socket read timeout")
 	flag.DurationVar(&cfg.SwarmRedisWriteTimeout, "swarm-redis-write-timeout", net.DefaultWriteTimeout, "set redis socket write timeout")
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -434,6 +435,7 @@ func NewConfig() *Config {
 	// Swarm:
 	flag.BoolVar(&cfg.EnableSwarm, "enable-swarm", false, "enable swarm communication between nodes in a skipper fleet")
 	flag.Var(cfg.SwarmRedisURLs, "swarm-redis-urls", "Redis URLs as comma separated list, used for building a swarm, for example in redis based cluster ratelimits.\nUse "+redisPasswordEnv+" environment variable or 'swarm-redis-password' key in config file to set redis password")
+	flag.StringVar(&cfg.SwarmRedisHashAlgorithm, "swarm-redis-hash-algorithm", "", "sets hash algorithm to be used in redis ring client to find the shard <jump|mpchash|rendezvousVnodes>, defaults to library default rendezvous")
```
AlexanderYastrebov, line 438: I think you may put `rendezvous` as a default value and then help message would mention it automatically so `, defaults to library default rendezvous` could be omitted.

AlexanderYastrebov: Also `rendezvous` should be be in the list of allowed values

AlexanderYastrebov: Although redis client library may change the default algorithm in the future.
Maybe we can have two options then: 
- when not set uses redis client library default whatever it is
- `rendezvous` uses rendezvous implementation (we would probably need to define a compatible wrapper)

szuecs: change as suggested

szuecs:
:+1:

net/redisclient.go
```diff
@@ -87,6 +101,96 @@ const (
...
+
+func NewJumpHash(shards []string) redis.ConsistentHash {
+	return &JumpHash{
+		hash:   fnv.New64(),
+		shards: shards,
+	}
+}
+
+func (j *JumpHash) Get(k string) string {
+	_, err := j.hash.Write([]byte(k))
```
AlexanderYastrebov, line 119: Is it safe to use the same hash instance from multiple threads?
I think we have to create a new hash instance on the fly here.

net/redisclient.go
```diff
@@ -87,6 +101,96 @@ const (
...
+	return mc.hash.Hash(k)
+}
+func siphash64seed(b []byte, s uint64) uint64 { return siphash.Hash(s, 0, b) }
+
+// rendezvous copied from github.com/go-redis/redis/v8@v8.3.3/ring.go
+type rendezvousWrapper struct {
+	*rendezvous.Rendezvous
+}
+
+func (w rendezvousWrapper) Get(key string) string {
```
AlexanderYastrebov, line 156: *rendezvousWrapper

net/redisclient.go
```diff
@@ -87,6 +101,96 @@ const (
...
+	return rendezvousWrapper{rendezvous.New(shards, xxhash.Sum64String)}
+}
+
+// rendezvous vnodes
+type rendezvousVnodes struct {
+	*rendezvous.Rendezvous
+	table map[string]string
+}
+
+func (w rendezvousVnodes) Get(key string) string {
```
AlexanderYastrebov, line 170: *rendezvousVnodes

net/redisclient.go
```diff
@@ -87,6 +101,96 @@ const (
 	defaultConnMetricsInterval = 60 * time.Second
 )
 
+// JumpHash
+// https://arxiv.org/pdf/1406.2294.pdf
+type JumpHash struct {
```
AlexanderYastrebov, line 106: Should this be private `jumpHash`?

net/redisclient.go
```diff
@@ -87,6 +101,96 @@ const (
...
+
+func (j *JumpHash) Get(k string) string {
+	_, err := j.hash.Write([]byte(k))
+	if err != nil {
+		log.Fatalf("Failed to write %s to hash: %v", k, err)
+	}
+
+	key := j.hash.Sum64()
+	h := jump.Hash(key, len(j.shards))
+	return j.shards[int(h)%len(j.shards)]
```
AlexanderYastrebov, line 126: https://github.com/dgryski/go-jump/blob/e1f439676b570800a863c37b78d1f868f51a85fc/jump.go#L10
> Hash consistently chooses a hash bucket number in the range [0, numBuckets) for the given key. numBuckets must be >= 1

so there is no need to use modulo operation.

But I think we have another problem here. 
Jump hash (as far as I can tell) is designed for the application where buckets/nodes are added sequentially and never removed from the middle.
When redis shard goes down, redis client recreates the hash with a list of available shards and thus breaks jump hash requirement of not removing the shard.

> This ability to add or remove buckets in any order can be valuable for cache servers where the servers are purely a performance improvement. But for data storage applications, where each bucket represents a different shard of the data, it is not acceptable for shards to simply disappear, because that shard is only place where the corresponding data is stored. Typically this is handled by either making the shards
redundant (with several replicas), or being able to quickly recover a replacement, or accepting lower availability for some data. Server death thus does not cause reallocation of data; only capacity changes do. This means that shards can be assigned numerical ids in increasing order as they are added, so that the active bucket ids always fill the range [0, num_buckets).

I.e. jump seems not applicable for our usecase because it will not provide the properties it have when conditions are not statisfied.

net/redisclient.go
```diff
@@ -87,6 +101,96 @@ const (
...
+// Multi-probe consistent hashing - mpchash
+// https://arxiv.org/pdf/1505.00062.pdf
+type multiprobe struct {
+	hash   *mpchash.Multi
+	shards []string
+}
+
+func NewMultiprobe(shards []string) redis.ConsistentHash {
+	return &multiprobe{
+		// 2 seeds and k=21
```
AlexanderYastrebov, line 138: 21 seems arbitrary, why not k=100?
Also
> We propose to store each node once but to hash keys K ways, returning the subsequent
node which is closest to a key hash. We call this multi-probe consistent hashing. This
requires O(n) space to store n nodes and O(K) time per lookup. Perhaps surprisingly,
using only K = 2 key hashes improves the peak-to-average load ratio from O(ln n) to O(1)
with high probability.

net/redisclient.go
```diff
@@ -87,6 +101,96 @@ const (
...
+		hash: mpchash.New(shards, siphash64seed, [2]uint64{1, 2}, 21),
+		// 2 seeds and k=41
+		//hash:   mpchash.New(shards, siphash64seed, [2]uint64{1, 2}, 41),
```
AlexanderYastrebov, line 141: Do we need this commented out block?

net/redisclient.go
```diff
@@ -87,6 +101,96 @@ const (
...
+	return v
+}
+
+func NewRendezvousVnodes(shards []string) redis.ConsistentHash {
+	N := 100
+	vshards := make([]string, N*len(shards))
+	table := make(map[string]string)
+	for i := 0; i < N; i++ {
+		for j, shard := range shards {
+			//vshard := fmt.Sprintf("%d%s", i, shard) // prefix
```
AlexanderYastrebov, line 185: Lets remove this commented out

net/redisclient.go
szuecs: does not need to be a pointer here

net/redisclient.go
szuecs: does not need to be a pointer here

net/redisclient.go
szuecs: It could be reused, but I was not sure about the others. I am happy to demote to private

net/redisclient.go
szuecs: There is never a condition that will remove shards from the ring in the middle permanently. It would be an incident if it would happen because of failing instance, so I think it's fine. I think we discussed this already in person. :)

net/redisclient.go
szuecs: It's 21 in the library itself in their tests/benchmarks, so I stick with that

net/redisclient.go
szuecs: no it was just for testing, did not help to increase k.

net/redisclient.go
szuecs: we should decide if we want to provide prefix or suffix vshards

net/redisclient.go
szuecs: true, will fix it

szuecs: done

net/redisclient.go
AlexanderYastrebov: The original jumphash library could be utilized if reuse is required

net/redisclient.go
AlexanderYastrebov: > in the middle permanently

Ok, but shard may go down for some time and that would shift key distribution.

Also jump hash assumes node list keeps the order on growth (i.e. existing shard keeps its index in the list). Redis client hash does not guarantee that shard list would contain newly added shard at the end and preserve the order.
Maybe it worth adding all assumptions as a comment.

The modulo is still redundant and could be simplified to `return j.shards[int(h)]`

Overall the only benefit of this algo over say rendevouz is that it "fast" but I think the performance gain is negligible compared to string hash calculation costs given there are expected only a handful of nodes.

net/redisclient.go
AlexanderYastrebov: Assuming xxhash has good https://en.wikipedia.org/wiki/Avalanche_effect it should not matter which one is used so lets remove the comment and stick with suffix

szuecs: removed

net/redisclient.go
szuecs: cleaned up and made a comment

net/redisclient.go
szuecs: made a comment

net/redisclient.go
szuecs: done

net/redisclient.go
szuecs: deleted the modulo operation

net/redisclient.go
AlexanderYastrebov: I think we need to create hash instance for each new key, i.e. inside `Get`

szuecs: yeah thanks, I had the fix, but reverted the file before pushing :D

szuecs: pushed the fix

AlexanderYastrebov:
:+1:

net/redisclient.go
```diff
@@ -87,6 +100,91 @@ const (
...
+	k := w.Lookup(key)
+	v, ok := w.table[k]
+	if !ok {
+		log.Printf("not found: %s in table for input: %s, so return %s", k, key, v)
+	}
+	return v
+}
+
+func NewRendezvousVnodes(shards []string) redis.ConsistentHash {
+	N := 100
```
aryszka, line 175: `const N = 100`?

net/redisclient.go
```diff
@@ -87,6 +100,91 @@ const (
...
+	vshards := make([]string, N*len(shards))
+	table := make(map[string]string)
+	for i := 0; i < N; i++ {
+		for j, shard := range shards {
+			vshard := fmt.Sprintf("%s%d", shard, i) // suffix
```
aryszka, line 180: vshard doesn't depend on j, let's move it out of the inner loop for clarity

aryszka:
:+1:

net/redisclient.go
AlexanderYastrebov: but depends on shard. Maybe swapping the loop would read better?..
```go
for _, shard := range shards {
  for i := 0; i < N; i++ {
    vshard := fmt.Sprintf("%s%d", shard, i)
  }
}

AlexanderYastrebov: Nevermind

szuecs: but the index in vshards needs j

net/redisclient.go
szuecs: changed to const

AlexanderYastrebov:
:+1:

net/redisclient.go
aryszka: true :)

aryszka:
:+1:

PR #1804, opened 2021-07-01
diagnostics filter jitter latency types uniform and normal

szuecs:


filters/diag/diag.go
```diff
@@ -417,3 +457,103 @@ func (t *throttle) Response(ctx filters.FilterContext) {
...
+		if len(args) != 4 {
+			return nil, errors.New("gaussian jitter filter requires 4 parameters")
+		}
+
+		s, ok := args[2].(string)
+		if !ok {
+			return nil, fmt.Errorf("failed to convert mu to string %v", args[2])
+		}
+
+		mu, err = strconv.ParseFloat(s, 32)
```
AlexanderYastrebov, line 508: I think we dont need to parse and can just check it is float64 like we do for numbers elsewhere (in any case `ParseFloat` should use 64 bits)

filters/diag/diag.go
```diff
@@ -417,3 +457,103 @@ func (t *throttle) Response(ctx filters.FilterContext) {
...
+	return &jitter{
+		typ:      j.typ,
+		delay:    delay,
+		interval: interval,
+		mu:       mu,
+		sigma:    sigma,
+	}, nil
+}
+
+func (j *jitter) Request(filters.FilterContext) {
```
AlexanderYastrebov, line 537: Should there be some tests?

filters/diag/diag.go
```diff
@@ -135,6 +153,28 @@ func NewBackendBandwidth() filters.Spec { return &throttle{typ: backendBandwidth
 //
 func NewBackendChunks() filters.Spec { return &throttle{typ: backendChunks} }
 
+// NewRandomJitter creates a jitter for requests with random
```
AlexanderYastrebov, line 156: Should it be "uniform" instead of "random" to be along with "normal" and "gaussian"?

filters/diag/diag.go
```diff
@@ -417,3 +457,103 @@ func (t *throttle) Response(ctx filters.FilterContext) {
...
+	case normalDist:
+		r := rand.NormFloat64()
+		f := r * float64(j.interval)
+		d += time.Duration(int64(f))
+	case gaussianDist:
+		r := rand.NormFloat64()
+		r = r*j.sigma + j.mu
+		f := r * float64(j.interval)
+		d += time.Duration(int64(f))
+	}
```
AlexanderYastrebov, line 554: Maybe lets just have one "normal" with sigma and mu (which is called `gaussianDist` here)
and let users to specify (sigma=1, mu=0) if they like.
https://en.wikipedia.org/wiki/Normal_distribution
https://stats.stackexchange.com/questions/55962/what-is-the-difference-between-a-normal-and-a-gaussian-distribution

filters/diag/diag.go
AlexanderYastrebov: We have `latency` and `backendLatency` that implement constant delay (IMO naming is not perfect - `latency` introduces delay on response path while `backendLatency` introduces delay on request path)

Should we add these as 
* `backendLatencyUniform`
* `backendLatencyNormal`
(and maybe add `latencyUniform` and `latencyNormal` later on) to stay consistent?

filters/diag/diag.go
```diff
@@ -135,6 +153,28 @@ func NewBackendBandwidth() filters.Spec { return &throttle{typ: backendBandwidth
 //
 func NewBackendChunks() filters.Spec { return &throttle{typ: backendChunks} }
 
+// NewRandomJitter creates a jitter for requests with random
+// distribution. Example delay around 1s with +/-120ms.
+//
+// 	* -> jitterRandom("1s", "120ms") -> "https://www.example.org";
```
AlexanderYastrebov, line 159: Should be added to `filters.md` as well

AlexanderYastrebov:
It is possible to implement the same in lua script but that may introduce extra latency besides requested due to lua overhead.

filters/diag/diag.go
szuecs: Yes, I just need to convert to float, no need to go through string :)
I am just writing tests, right now. If you have naming

filters/diag/diag.go
szuecs: that's why it's WIP :)

filters/diag/diag.go
szuecs: I am fine with that.

filters/diag/diag.go
szuecs: yeah docs missing

filters/diag/diag.go
szuecs: great idea!

szuecs: done

filters/diag/diag.go
szuecs: done

szuecs:
@AlexanderYastrebov yeah Lua could have been used, but I think jitter is useful for diagnostics filters.

filters/diag/diag.go
szuecs: done

filters/diag/diag.go
szuecs: done

filters/diag/diag.go
```diff
@@ -417,3 +451,89 @@ func (t *throttle) Response(ctx filters.FilterContext) {
...
+		return nil, fmt.Errorf("failed to parse duration delay %v: %w", args[0], err)
+	}
+
+	if interval, err = parseDuration(args[1]); err != nil {
+		return nil, fmt.Errorf("failed to parse duration interval %v: %w", args[1], err)
+	}
+
+	switch j.typ {
+	case normalDist:
+		switch len(args) {
```
AlexanderYastrebov, line 488: default is missing (e.g. len==3)

filters/diag/diag.go
AlexanderYastrebov: I've noticed `wip` label too late, sorry :)

filters/diag/diag.go
szuecs: so provide sigma as default?

filters/diag/diag.go
```diff
@@ -417,3 +451,89 @@ func (t *throttle) Response(ctx filters.FilterContext) {
...
+	case uniformDist:
+		/* #nosec */
+		r := rand.Float64() - 0.5 // +/- \epsilon
+		f := r * float64(j.interval)
+		d += time.Duration(int64(f))
+	case normalDist:
+		r := rand.NormFloat64()
+		r = r*j.sigma + j.mu
+		f := r * float64(j.interval)
+		d += time.Duration(int64(f))
```
AlexanderYastrebov, line 533: Lets check this, I think it should be
```go
d = rand.NormFloat64()*j.sigma + j.delay
```
i.e. mu is the mean or expectation of the distribution which is exactly the delay

szuecs: so you mean to drop j.mu, right?

AlexanderYastrebov: https://en.wikipedia.org/wiki/Standard_deviation
> A useful property of the standard deviation is that unlike the variance, it is expressed in the same unit as the data.

> If a data distribution is approximately normal then about 68 percent of the data values are within one standard deviation of the mean (mathematically, μ ± σ, where μ is the arithmetic mean), about 95 percent are within two standard deviations (μ ± 2σ), and about 99.7 percent lie within three standard deviations (μ ± 3σ). This is known as the 68-95-99.7 rule, or the empirical rule.

I think "normal" filter should have two  `time.Duration` parameters mean and sigma.

filters/diag/diag.go
szuecs: here it can only be 2 or 4, see https://github.com/zalando/skipper/pull/1804/files/2358921d50990ef3efa3da2240ad320e5b4fb1bf#diff-b091ac1989fb4f770f5e466781a6a12a59d432ec8f3ad1caaef8d8d2d245e950R474, but if we drop mu, then I can change it to 3

filters/diag/diag.go
szuecs: One problem is that I don't have +/-, only + in the normalDist case.

filters/diag/diag.go
```diff
@@ -417,3 +451,89 @@ func (t *throttle) Response(ctx filters.FilterContext) {
...
+
+func (j *jitter) Request(filters.FilterContext) {
+	d := j.delay
+
+	switch j.typ {
+	case uniformDist:
+		/* #nosec */
+		r := rand.Float64() - 0.5 // +/- \epsilon
+		f := r * float64(j.interval)
+		d += time.Duration(int64(f))
```
AlexanderYastrebov, line 528: for delay=100ms and interval=20ms
this would sleep 100ms +/- 10ms while docs say
> The uniformJitter filter creates a jitter latency in a uniform
distribution for requests. The first parameter is the mean and the
second is the window. In the example we would sleep for 100ms+/-20ms.

so either `interval`  should be doubled here or docs fixed.

szuecs: docs fixed  thanks :)

filters/diag/diag.go
szuecs: I used float64 because of https://golang.org/pkg/math/rand/#NormFloat64, but maybe it's better as you wrote.
Let me test.

docs/reference/filters.md
```diff
@@ -634,6 +634,32 @@ It logs with INFO level and a unique ID per request:
 The absorbSilent filter reads and discards the payload of the incoming requests. It only
 logs read errors other than EOF.
 
+## uniformJitter
+
+The uniformJitter filter creates a jitter latency in a uniform
+distribution for requests. The first parameter is the mean and the
+second is the window. In the example we would sleep for 100ms+/-10ms.
```
AlexanderYastrebov, line 641: I think users would tend to pick even values for "window". 
Would it be better if we call it "delta", something like:

> The `uniformJitter` filter creates uniformly distributed request latency on the interval `[mean-delta, mean+delta]`
> * -> uniformJitter("100ms, "10ms") -> "https://www.example.org";

so that user does not have to divide "window" to calculate max or min latency.

An alternative would be to specify min and max latency directly:
> The `uniformJitter` filter creates request latency uniformly distributed between min and max.
> * -> uniformJitter("90ms, "110ms") -> "https://www.example.org";

filters/diag/diag.go
AlexanderYastrebov: I think this was not resolved, will keep discussion open for now

filters/diag/diag.go
szuecs: I changed the code to have only 2 params for normalJitter() reusing the var from uniformJitter.

docs/reference/filters.md
szuecs: yeah true maybe better for the user. It was so nice in code :-D

szuecs: I changed as you suggested

filters/diag/diag.go
```diff
@@ -417,3 +445,58 @@ func (t *throttle) Response(ctx filters.FilterContext) {
...
+	}, nil
+}
+
+func (j *jitter) Request(filters.FilterContext) {
+	d := j.mean
+
+	switch j.typ {
+	case uniformDist:
+		/* #nosec */
+		r := rand.Float64() - 0.5   // +/- sizing
```
AlexanderYastrebov, line 490: >  It was so nice in code :-D

you can do 
```go
r := 2*rand.Float64() - 1
```
to get +/-1 but anyway there must be 2x somewhere :)

AlexanderYastrebov:
I did some tests, IMO looks legit.

### Uniform
```
bin/skipper -inline-routes='* -> jitterUniform("100ms", "20ms") -> status(204) -> <shunt>' -access-log-disabled
```
```
echo "GET http://localhost:9090/test" | vegeta attack -rate 500/s -duration 1m | vegeta report -type hist[0,80ms,90ms,100ms,110ms,120ms]
Bucket           #     %       Histogram
[0s,     80ms]   0     0.00%   
[80ms,   90ms]   6512  21.71%  ################
[90ms,   100ms]  7571  25.24%  ##################
[100ms,  110ms]  7433  24.78%  ##################
[110ms,  120ms]  7500  25.00%  ##################
[120ms,  +Inf]   984   3.28%   ##
```
Most of the requests uniformly distributed within `[80ms, 120ms]` interval as expected.

### Normal
```
bin/skipper -inline-routes='* -> jitterNormal("100ms", "20ms") -> status(204) -> <shunt>' -access-log-disabled
```
```
echo "GET http://localhost:9090/test" | vegeta attack -rate 500/s -duration 1m | vegeta report -type hist[0,40ms,60ms,80ms,100ms,120ms,140ms,160ms]
Bucket           #      %       Histogram
[0s,     40ms]   24     0.08%   
[40ms,   60ms]   564    1.88%   #
[60ms,   80ms]   3877   12.92%  #########
[80ms,   100ms]  9867   32.89%  ########################
[100ms,  120ms]  10441  34.80%  ##########################
[120ms,  140ms]  4405   14.68%  ###########
[140ms,  160ms]  770    2.57%   #
[160ms,  +Inf]   52     0.17%   
```
Matches https://en.wikipedia.org/wiki/68%E2%80%9395%E2%80%9399.7_rule

filters/diag/diag.go
szuecs: I would rather go for, but I am not sure if `jitter` is not the better name:
`responseLatencyUniform`
`responseLatencyNormal`

Maybe better `responseLatency` and deprecate `backendLatency`?

filters/diag/diag.go
szuecs: thanks fixed

filters/diag/diag.go
```diff
@@ -417,3 +445,58 @@ func (t *throttle) Response(ctx filters.FilterContext) {
...
+}
+
+func (j *jitter) Request(filters.FilterContext) {
+	d := j.mean
+
+	switch j.typ {
+	case uniformDist:
+		/* #nosec */
+		r := 2*rand.Float64() - 1   // +/- sizing
+		f := r * float64(2*j.delta) // resize to specified delta
```
AlexanderYastrebov, line 491: For
```go
r := 2*rand.Float64() - 1
```
we do not need to multiply delta, i.e. it should be 
```go
f := r * float64(j.delta)
```

szuecs: ok thanks, fixed

filters/diag/diag.go
szuecs: internally discussed -> `normalRequestLatency(), normalResponseLatency(), uniformRequestLatency(), uniformResponseLatency()`

AlexanderYastrebov:
:+1:

szuecs:
:+1:

aryszka:
:+1:

PR #1806, opened 2021-07-06
Prints deprecation warning

AlexanderYastrebov:
Prints deprecation warning when deprecated config parameter is set by flag or config yaml


config/config.go
```diff
@@ -488,10 +489,6 @@ func (c *Config) Parse() error {
 		return err
 	}
 
-	if c.KubernetesEnableEastWest {
-		log.Warn(`"kubernetes-enable-east-west" parameter is deprecated. Check the "kubernetes-east-west-range" feature`)
```
AlexanderYastrebov, line 492: Here the wrong name `kubernetes-enable-east-west` was used, should be `enable-kubernetes-east-west`

PR #1807, opened 2021-07-06
Prints deprecation warning

AlexanderYastrebov:
Prints deprecation warning when deprecated config parameter is set by flag or config yaml


config/config.go
```diff
@@ -488,10 +489,6 @@ func (c *Config) Parse() error {
 		return err
 	}
 
-	if c.KubernetesEnableEastWest {
-		log.Warn(`"kubernetes-enable-east-west" parameter is deprecated. Check the "kubernetes-east-west-range" feature`)
```
AlexanderYastrebov, line 492: Here the wrong name `kubernetes-enable-east-west` was used, should be `enable-kubernetes-east-west`

config/config.go
```diff
@@ -474,9 +474,10 @@ func (c *Config) Parse() error {
 		flag.Parse()
 	}
 
-	if c.ApiUsageMonitoringDefaultClientTrackingPattern != "" {
-		log.Warn(`"api-usage-monitoring-default-client-tracking-pattern" parameter is deprecated`)
-	}
+	c.deprecated("enable-prometheus-metrics", false)
+	c.deprecated("api-usage-monitoring-default-client-tracking-pattern", "")
+	c.deprecated("enable-kubernetes-east-west", false)
+	c.deprecated("kubernetes-east-west-domain", "")
```
AlexanderYastrebov, line 480: This is a simple but requires repetition of option name and a default value. Here is a bit more involved variant https://github.com/zalando/skipper/tree/deprecated-flag but I favor simplicity

szuecs:
:+1:

config/config_test.go
```diff
@@ -232,3 +233,51 @@ func TestMinTLSVersion(t *testing.T) {
...
+		log.SetOutput(o)
+		log.SetFormatter(f)
+	}()
+
+	formatter := &testFormatter{messages: make(map[string]log.Level)}
+
+	os.Args = []string{
+		"skipper",
+		"-config-file=testdata/deprecated.yaml",
+		"-enable-prometheus-metrics=false", // set to default value and does not produce deprecation warning
```
AlexanderYastrebov, line 261: It would be better if logic detects use of default values as well - i.e. any use of deprecated options

config/config.go
AlexanderYastrebov: I've changed the implementation to check if flag/config is set regardless of the value

aryszka:
:+1:

AlexanderYastrebov:
:+1:

PR #1809, opened 2021-07-07
Adds template placeholder support to `inlineContent` filter

AlexanderYastrebov:

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -446,11 +446,15 @@ Parameters:
 * content (string)
 * content type (string) - optional
 
-Example:
+content may contain [template placeholders](#template-placeholders).
+If a template placeholder can't be resolved then empty value is used for it.
+
+Examples:
 
 ```
 * -> inlineContent("<h1>Hello</h1>") -> <shunt>
 * -> inlineContent("[1,2,3]", "application/json") -> <shunt>
+* -> inlineContent("Your IP is ${request.clientIP}\n") -> <shunt>
 * -> status(418) -> inlineContent("Would you like a cup of tea?") -> <shunt>
 ```
 
--- a/filters/builtin/inlinecontent.go
+++ b/filters/builtin/inlinecontent.go
@@ -6,12 +6,13 @@ import (
 	"net/http"
 	"strconv"
 
+	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/filters"
 )
 
 type inlineContent struct {
-	text string
-	mime string
+	template *eskip.Template
+	mime     string
 }
 
 // Creates a filter spec for the inlineContent() filter.
@@ -24,7 +25,7 @@ type inlineContent struct {
 //
 //     * -> inlineContent("{\"foo\": 42}", "application/json") -> <shunt>
 //
-// It accepts two arguments: the content and the optional content type.
+// It accepts two arguments: the content template (see eskip.Template.ApplyContext) and the optional content type.
 // When the content type is not set, it tries to detect it using
 // http.DetectContentType.
 //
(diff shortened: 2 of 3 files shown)


filters/builtin/inlinecontent.go
```diff
@@ -36,51 +37,40 @@ func NewInlineContent() filters.Spec {
...
-		f.mime, err = stringArg(args[1])
-		if err != nil {
-			return nil, err
+		f.mime, ok = args[1].(string)
+		if !ok {
+			return nil, filters.ErrInvalidFilterParameters
 		}
 	} else {
-		f.mime = http.DetectContentType([]byte(f.text))
+		f.mime = http.DetectContentType([]byte(text))
```
aryszka, line 59: just nitpicking: can there be any special cases when the content type should be detected after the template has been resolved?

filters/builtin/inlinecontent_test.go
```diff
@@ -58,10 +59,10 @@ func TestInlineContentArgs(t *testing.T) {
 
 			c := f.(*inlineContent)
 
-			if c.text != test.expectedText {
+			if !reflect.DeepEqual(c.template, test.expectedTemplate) {
```
aryszka, line 62: not strictly for this PR, but I think we could just merge the two tests of TestInlineContentArgs and TestInlineContent into a single TestInlineContent one. On this line, we are comparing the template with DeepEqual opaque objects from another package, which in theory can cause various annoyances.

aryszka: ...but if just checked that the right text is returned when the filter is applied after we merged two groups of tests, then we don't need to do this.

filters/builtin/inlinecontent.go
AlexanderYastrebov: The main motivation is to allow request/path properties to be reflected back in the response.
If there is enough hint in the template the content type would be properly detected, e.g.:
```
$ bin/skipper -inline-routes='* -> inlineContent("<html>${request.query.content}</html>") -> <shunt>'
$ bin/skipper -inline-routes='* -> inlineContent("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<response>${request.query.content}</response>") -> <shunt>'
```

Of course we can imagine something like
```
* -> inlineContent("${request.query.content}") -> <shunt>

```
and
```
curl "localhost:9090/?content=hello+world"
curl -v "localhost:9090/?content=<html>hello</html>"
```
In such cases content would not be detected. 
For such cases is content type is known upfront and does not change users may explicitly set it.

We can make content type to support templates as well then client could pass the value as another request property.
Or we can move content detection into runtime after template evaluation.

I would keep it as is and leave for future improvement if we ever find a usecase for content type detection.

Let me add tests for content type detection with templates though.

filters/builtin/inlinecontent_test.go
AlexanderYastrebov: Good point, fixed

filters/builtin/inlinecontent.go
aryszka: Fair enough :+1:

aryszka:
:+1:

filters/builtin/inlinecontent.go
szuecs: I hope nobody will use the examples in a production service, because it's "html injection as a sevice" ;)

docs/reference/filters.md
```diff
@@ -446,11 +446,15 @@ Parameters:
...
-Example:
+content may contain [template placeholders](#template-placeholders).
+If a template placeholder can't be resolved then empty value is used for it.
+
+Examples:
 
 ```
 * -> inlineContent("<h1>Hello</h1>") -> <shunt>
 * -> inlineContent("[1,2,3]", "application/json") -> <shunt>
+* -> inlineContent("Your IP is ${request.clientIP}\n") -> <shunt>
```
szuecs, line 457: nice implementation of "whatsmyip" :)

docs/reference/filters.md
```diff
@@ -446,11 +446,15 @@ Parameters:
 * content (string)
 * content type (string) - optional
 
-Example:
+content may contain [template placeholders](#template-placeholders).
+If a template placeholder can't be resolved then empty value is used for it.
```
szuecs, line 450: Maybe you should add that using request data in the response is prone to injections and can lead to refletced XSS, HTML injection, ...

AlexanderYastrebov:
@szuecs Good point regrading content injection danger. This risk IMO lessens the value of having template support in inline content. If needed the same result might be achieved using lua:
```
$ bin/skipper -inline-routes='* -> lua("function request(c) c.serve({status_code=200, body=c.request.host .. \"\\n\"}) end") -> <shunt>'
```

Issue #1810, opened 2021-07-08
JWT Token validation without introspection endpoint


abinet:
In current project we are using skipper to cover authorization part for applications running on AKS. OpenID flow works fine, but because Azure AD does not provide token introspection endpoint, we are not able to validate tokens in service-2-service use case. To cover this gap we implemented a new filter jwtValidation. This filter extracts Authorization Bearer token, validates using public keys of Authorization server and stores info into same map as oauthOidcUserInfo does. So we can use oidcClaimsQuery for filtering based on claims extracted from the token.

Filter parameters are:
````
jwtValidation(authorization_provider_url, claims, upstream_headers)
````

Here is the example of usage:
````
zalando.org/skipper-filter: jwtValidation("https://login.microsoftonline.com/<tenant_id>/v2.0/",
      "sub", "X-Remote-User:sub") -> oidcClaimsQuery("/:groups.#[==\"group-a\"]","/:groups.#[==\"group-b\"]")
````

It would be great to review filter functionality with skipper team and contribute back.

aryszka:
Hi @abinet , thanks a lot. I agree with this proposal, the functionality makes a lot of sense. We also thought about it already, but we currently use a side car for this functionality internally and it was not a main priority to replace it with internal logic, but clearly, that would make the setup way simpler for everybody.

AlexanderYastrebov:
https://github.com/zalando/skipper/issues/1752 might be related

szuecs:
I added some comments.
Mainly: 
1. you do need a goroutine to refresh public keys from JWKs 
2. because of 1. you don't need the authclient at all
3. because of 1. you should not fetch public keys in Request() 
4. you should use a separate type for the query filter, similar to https://github.com/zalando/skipper/blob/master/filters/auth/grantclaimsquery.go

PR #1811, opened 2021-07-08
jwtValidation and forwardTokenField filters

abinet:
The filter extracts Authorization Bearer token, validates using public keys of Authorization server and stores info into same map as oauthOidcUserInfo does. So we can use oidcClaimsQuery for filtering based on claims extracted from the token.

Filter parameters are:

jwtValidation(authorization_provider_url, claims, upstream_headers)
Here is the example of usage:

zalando.org/skipper-filter: jwtValidation("https://login.microsoftonline.com/<tenant_id>/v2.0/",
      "sub", "X-Remote-User:sub") -> oidcClaimsQuery("/:groups.#[==\"group-a\"]","/:groups.#[==\"group-b\"]")

https://github.com/zalando/skipper/issues/1810


aryszka:
I like the name `jwtValidation` because it's simple, but it may not fit well with the existing conventions of the authnz filters. What are the thoughts of others on this?

szuecs:
I think naming us fine, but only reading the pr message, I think we miss an auto refresh of JKS. Keys (public keys). See tokenintrospection filters for issuer well known configuration integration.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,257 @@
...
+
+	cfg, err := getOpenIDConfig(issuerURL)
+	if err != nil {
+		return nil, err
+	}
+
+	var ac *authClient
+	var ok bool
+	if ac, ok = issuerAuthClient[issuerURL]; !ok {
+		ac, err = newAuthClient(cfg.JwksURI, tokenInfoSpanName, s.options.Timeout, s.options.MaxIdleConns, s.options.Tracer)
```
szuecs, line 85: The first argument `cfg.JwksURI` is wrong. JWKs URI is the one you can pull public keys from.
I would suggest to use a goroutine to refresh every 30m the public keys from the JWKs URI and below in `Request()` you just validate the token with the publickeys stored in the filter. You don't need the roundtrip to the JWKs uri in the hot path.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,257 @@
...
+				rsakey.N = new(big.Int).SetBytes(number)
+				rsakey.E = 65537
+				rsakeys[kid] = rsakey
+			}
+		} else {
+			log.Error("Not able to get public keys")
+			unauthorized(ctx, "", "Not able to get public keys", f.authClient.url.Hostname(), "")
+			return
+		}
+
```
szuecs, line 171: Please see https://github.com/zalando/skipper/pull/1811/files#r672359596 and check that you do not need https://github.com/zalando/skipper/pull/1811/files#diff-7632350376ed0aebec65031a9b5191d7ea917bfe9fbc8b0f87e0c7b865572304R146-R171

Starting from `body, err := f.authClient.getTokeninfo("", ctx)` until the comment, you do not need.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,257 @@
...
+	var container2 tokenContainer
+	container2.Claims = make(map[string]interface{})
+	for claim, value := range info.Claims.(jwt.MapClaims) {
+		container2.Claims[claim] = value
+	}
+	//container2.Claims = info.Claims.(jwt.MapClaims)
+
+	log.Errorf("Length %d", len(container2.Claims))
+	ctx.StateBag()[oidcClaimsCacheKey] = container2
```
szuecs, line 233: Please do not re-use oidcClaimsCacheKey, but better implement the functionality based on the oidc `oidcClaimsQuery()` filter. As an example that re-uses the code of `oidcClaimsQuery()` see https://github.com/zalando/skipper/blob/master/filters/auth/grantclaimsquery.go for the implementation of the query function.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+func JwtValidationWithOptions(
+	create func(time.Duration) filters.Spec,
+	o TokenintrospectionOptions,
+) filters.Spec {
+	s := create(o.Timeout)
+	ts, ok := s.(*tokenIntrospectionSpec)
+	if !ok {
+		return s
```
szuecs, line 55: Looks like pretty unsafe to me. A user wants to have X, but you get an error to create it so you return Y. This might be unexpected.

filters/auth/jwt_validation.go
abinet: This is clear in general. Will try to implement the change.

filters/auth/jwt_validation.go
abinet: This is clear in general. Will try to implement the change.

filters/auth/jwt_validation.go
abinet: Resolved this by providing following functionality: on filter creation I am getting oidc config with JWKs URI and scheduling the call to load keys from the uri every 30 min.

filters/auth/jwt_validation.go
abinet: @szuecs I need more clarification on this. grantClaimsQuery re-uses the functionality of oidcClaimsQuery, but I don't see  a place where oidcClaimsCacheKey gets overriden. So if I don't miss anything grantClaimsQuery should get a payload with oidcClaimsCacheKey as well. However I see tokeninfo filter is storing payload using  tokeninfoCacheKey = "tokeninfo"
https://github.com/zalando/skipper/blob/master/filters/auth/tokeninfo.go#L356 and tokenIntrospection is using tokenintrospectionCacheKey   = "tokenintrospection"
https://github.com/zalando/skipper/blob/master/filters/auth/tokenintrospection.go#L473 

How does this work?

AlexanderYastrebov: I think grant filter reuses statebag key
https://github.com/zalando/skipper/blob/39f672857320b501e86149bfc12962c47d4ab761/filters/auth/grant.go#L187-L194
and `grantClaimsQuery` simply creates an instance of `oidcClaimsQuery` under another name https://github.com/zalando/skipper/blob/39f672857320b501e86149bfc12962c47d4ab761/filters/auth/grantclaimsquery.go#L1-L6

We can do the same here if needed or make `oidcIntrospectionSpec` configurable with respect to the statebag key.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,301 @@
...
+	scheduler := gocron.NewScheduler(time.UTC)
+	scheduler.Every("30m").Do(func() {
+		err = fillPublicKeys(cfg.JwksURI)
+		if err != nil {
+			log.Errorf("Error getting public keys for url: %s", cfg.JwksURI)
+			return
+		}
+	})
+
+	scheduler.StartAsync()
```
AlexanderYastrebov, line 97: Lets also consider existing options referenced by https://github.com/golang-jwt/jwt/issues/73

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,301 @@
...
+
+		for _, header := range strings.Split(sargs[2], " ") {
+			sl := strings.SplitN(header, ":", 2)
+			if len(sl) != 2 || sl[0] == "" || sl[1] == "" {
+				return nil, fmt.Errorf("%w: malformatted filter for upstream headers %s", filters.ErrInvalidFilterParameters, sl)
+			}
+			f.upstreamHeaders[sl[0]] = sl[1]
+		}
+		log.Debugf("Upstream Headers: %v", f.upstreamHeaders)
+	}
```
AlexanderYastrebov, line 120: I think this filter should do bare minimum - validate token and maybe certain claims.

Token (or its fields) forwarding is an optional usecase and should be implemented by separate filter. Maybe separate filter is not even needed and https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#forwardtoken could be reused.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,301 @@
...
+
+	return f, nil
+}
+
+// String prints nicely the jwtValidationFilter configuration based on the
+// configuration and check used.
+func (f *jwtValidationFilter) String() string {
+	return fmt.Sprintf("%s(%s)", JwtValidationName, strings.Join(f.claims, ","))
+}
```
AlexanderYastrebov, line 129: I am not sure we have String() for other filters so lets remove if it is not used

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	m.Lock()
+	defer m.Unlock()
+
+	if doc["keys"] != nil {
+		for _, bodykey := range doc["keys"].([]interface{}) {
+			key := bodykey.(map[string]interface{})
+			kid := key["kid"].(string)
+			rsakeys[kid] = parsePublicKey(key["n"].(string))
```
AlexanderYastrebov, line 293: `kid` is not necessary unique across different issuers.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	if !info.Active() {
+		unauthorized(ctx, sub, inactiveToken, f.authClient.url.Hostname(), "")
+		return
+	}*/
+
+	sub := info.Claims.(jwt.MapClaims)["sub"].(string)
+
+	var allowed = f.validateAnyClaims(info)
```
AlexanderYastrebov, line 208: I see there is a family of `oauthTokenintrospection*` filters that validate any, all claims or key-values and I am wondering would we need to support the similar set of validations? If so I would prefer that we reuse validation code instead of copy-pasting.

Also would it be possible to completely remove claims validation here and rely on the `oidcClaimsQuery()` instead reducing this filter scope only to the JWT signature validation?

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,301 @@
...
+
+	jwtValidationFilter struct {
+		scheduler       *gocron.Scheduler
+		claims          []string
+		upstreamHeaders map[string]string
+	}
+)
+
+var m sync.Mutex
+var rsakeys map[string]*rsa.PublicKey = make(map[string]*rsa.PublicKey)
```
AlexanderYastrebov, line 39: I think for the filter to be usable we should support variety of algorithms (e.g. elliptic curves), not only RSA.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,301 @@
...
+	// adding upstream headers
+	f.setHeaders(ctx, info.Claims.(jwt.MapClaims))
+}
+
+func (f *jwtValidationFilter) Response(filters.FilterContext) {}
+
+// Close cleans-up the scheduler
+func (f *jwtValidationFilter) Close() {
+	f.scheduler.Clear()
+}
```
AlexanderYastrebov, line 235: This unfortunately would not work as expected due to https://github.com/zalando/skipper/issues/202

filters/auth/jwt_validation.go
abinet: This make sense from one side. From another side I was trying to keep the similarity to oidc flow where filter does check the claims

filters/auth/jwt_validation.go
abinet: Make sense however I am not familiar with different algorithms and I am afraid won't be able to implement the universal support here

filters/auth/jwt_validation.go
abinet: If I understand properly, you don't want to use scheduler anyways. So, this step will not be necessary then.

AlexanderYastrebov: This is just a note that cleanup unfortunately is not implemented yet though there are filters that implement Close method that is not used

filters/auth/jwt_validation.go
abinet: I see your point. However in this case whole PR will be useless for me because I won't be able to cover my use-case.

filters/auth/jwt_validation.go
abinet: Guys, still don't get what is the plan. Do I need to introduce new key, adjust oidcIntrospectionSpec to allow key's override and create a new xxxClaimsQuery re-using oidcIntrospectionSpec?

filters/auth/jwt_validation.go
abinet: Thanks for the hint. Will try to implement approach based on https://github.com/MicahParks/keyfunc

filters/auth/jwt_validation.go
AlexanderYastrebov: IMO we can reuse the key as grant filter does for now.

filters/auth/jwt_validation.go
AlexanderYastrebov: > do I understand properly that setContextRequestHeader filter is not working with maps

correct

> Maybe we can extend them to support map traversal

In such case `setContextRequestHeader` would have to know about statebag key name which would make it impossible to refactor later on. I would say statebag should be kept for internal communication within skipper, it is unfortunate `setContext*` are already there. See also related discussion https://github.com/zalando/skipper/pull/1528#discussion_r526505822 

I think we have the following options:
- use existing `forwardToken` and adjust backend to read json from the header value instead of plain string
- extend `forwardToken` somehow to produce non-json string response
- introduce separate filter to extract field in non-json format from the token payload

abinet: Let's try to proceed with third option - create separate filter which extracts fields/claims from statebag with key "oidcclaimscachekey". Any suggestion on naming (set|forward|apply)(Token|HeaderFromToken)(Claim|Field) ?

abinet: created a new filter forwardTokenPart and removed header functionality from jwtValidation

filters/auth/jwt_validation.go
abinet: so, I am not changing anything on this

abinet:
I suppose to have resolved all the issues mentioned by review. @szuecs @AlexanderYastrebov  can you have a look once you get time please.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	return f, nil
+}
+
+func getKeyFunction(url string) (jwks *keyfunc.JWKs, err error) {
+	if jwks, ok := jwksMap[url]; ok {
+		return jwks, nil
+	} else {
+		jwks, err := keyfunc.Get(url)
```
MicahParks, line 88: I saw in the PR comments that key rotation was desired. If that's the case, adding a [`keyfunc.Options`](https://pkg.go.dev/github.com/MicahParks/keyfunc#Options) argument to this function call with the `RefreshInterval`, `RefreshRateLimit`, and `RefreshUnknownKID` fields would be recommended.

filters/auth/forwardtokenpart.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"fmt"
+
+	"github.com/zalando/skipper/filters"
+	"golang.org/x/net/http/httpguts"
+)
+
+const (
+	ForwardTokenPartName = "forwardTokenPart"
```
AlexanderYastrebov, line 11: I know naming is hard but maybe it should be `forwardTokenField` as `part` feels a bit ambiguous?..
Also do we foresee the need to forward multiple fields e.g. comma-separated?

filters/auth/forwardtokenpart.go
```diff
@@ -0,0 +1,74 @@
...
+type (
+	forwardTokenPartSpec   struct{}
+	forwardTokenPartFilter struct {
+		HeaderName string
+		Key        string
+	}
+)
+
+// NewForwardTokenPart creates a filter to forward part of the result of token info or
+// token introspection or to the backend server.
```
AlexanderYastrebov, line 23: >...token introspection or <?> to the backend server.

Something seems to be missing at `<?>`. Also we would need to update filter docs similar to https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#forwardtoken

filters/auth/forwardtokenpart.go
```diff
@@ -0,0 +1,74 @@
...
+
+	remainingArgs := args[1:]
+	stringifiedRemainingArgs := make([]string, len(remainingArgs))
+	for i, v := range remainingArgs {
+		maskedKeyName, ok := v.(string)
+		if !ok {
+			return nil, filters.ErrInvalidFilterParameters
+		}
+		stringifiedRemainingArgs[i] = maskedKeyName
+	}
```
AlexanderYastrebov, line 53: The filter needs exactly two string arguments, please review this logic.

filters/auth/forwardtokenpart_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"testing"
+	"time"
+
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/net"
+	"github.com/zalando/skipper/proxy/proxytest"
+)
+
+func TestForwardTokenPart(t *testing.T) {
```
AlexanderYastrebov, line 20: Maybe we do not need full-fledged integration test that is hard to follow for a filter that simply gets and puts value from the state bag. 

Please have a look at e.g. https://github.com/zalando/skipper/blob/master/filters/ratelimit/backendratelimit_test.go and may be you would decide to rewrite it into a unit test.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func NewJwtValidation(timeout time.Duration) filters.Spec {
+	return NewJwtValidationWithOptions(TokenintrospectionOptions{
+		Timeout: timeout,
+		Tracer:  opentracing.NoopTracer{},
+	})
+}
+
+func JwtValidationWithOptions(
```
AlexanderYastrebov, line 43: Lets have a single constructor for simplicity

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,170 @@
...
+	create func(time.Duration) filters.Spec,
+	o TokenintrospectionOptions,
+) filters.Spec {
+	s := create(o.Timeout)
```
AlexanderYastrebov, line 47: I can not find this `create` function

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,170 @@
...
+	jwtValidationSpec struct {
+		options TokenintrospectionOptions
+	}
+
+	jwtValidationFilter struct {
+		jwksUri string
+	}
+)
+
+var jwksMap map[string]*keyfunc.JWKs = make(map[string]*keyfunc.JWKs)
```
AlexanderYastrebov, line 28: Please add a doc string describing what is the key in this map.

Also the map is global var and access is not guarded with a mutex, I guess this follows e.g.
https://github.com/zalando/skipper/blob/2059778e1924aaafb86accfc0b27e10245f13117/filters/auth/tokenintrospection.go#L72

Lets leave it as is for now, I would keep it scoped within the `jwtValidationSpec` and guard access with a mutex.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,170 @@
...
+		jwksUri: cfg.JwksURI,
+	}
+
+	return f, nil
+}
+
+func getKeyFunction(url string) (jwks *keyfunc.JWKs, err error) {
+	if jwks, ok := jwksMap[url]; ok {
+		return jwks, nil
+	} else {
```
AlexanderYastrebov, line 87: Lets remove `else` block - there is an early return in the `if`

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,170 @@
...
+		// Create the keyfunc options. Use an error handler that logs. Refresh the JWKs when a JWT signed by an unknown KID
+		// is found or at the specified interval. Rate limit these refreshes. Timeout the initial JWKs refresh request after
+		// 10 seconds. This timeout is also used to create the initial context.Context for keyfunc.Get.
+		refreshInterval := time.Hour
+		refreshRateLimit := time.Minute * 5
+		refreshTimeout := time.Second * 10
+		refreshUnknownKID := true
+		options := keyfunc.Options{
+			RefreshErrorHandler: func(err error) {
+				log.Printf("There was an error with the jwt.KeyFunc\nError:%s\n", err.Error())
```
AlexanderYastrebov, line 97: This should be ``Debugf`, `Warnf` or `Errorf`

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,170 @@
...
+		token, ok := getToken(r)
+		if !ok || token == "" {
+			unauthorized(ctx, "", missingToken, "", "")
+			return
+		}
+
+		jwks, err := getKeyFunction(f.jwksUri)
+		if err != nil {
+			log.Fatalf("Failed to get the JWKs from the given URL %s. Error:%s", f.jwksUri, err.Error())
+			unauthorized(ctx, "", "missing JWKs", "", "")
```
AlexanderYastrebov, line 129: Please add new `"missing JWKs"` reject reason similar to `missingToken`

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,170 @@
...
+			return
+		} else if parsedToken.Header["alg"] == nil {
+			log.Errorf("alg must be defined")
+			unauthorized(ctx, "", "alg must be defined", "", "")
+			return
+		}
+
+		info = *parsedToken
+	} else {
+		info = infoTemp.(jwt.Token)
```
AlexanderYastrebov, line 150: Something is wrong here: the `oidcClaimsCacheKey` value casted to `jwt.Token` but down the line `tokenContainer` is stored: `ctx.StateBag()[oidcClaimsCacheKey] = container`

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	sub := info.Claims.(jwt.MapClaims)["sub"].(string)
```
AlexanderYastrebov, line 153: Is it safe to do these casts?

AlexanderYastrebov: Should it check that sub is not missing?

filters/auth/jwt_validation_test.go
```diff
@@ -0,0 +1,210 @@
...
+		Issuer:    "test",
+		Subject:   "aaa",
+	}
+
+	token := jwt.NewWithClaims(jwt.SigningMethodRS256, claims)
+	token.Header["kid"] = kid
+
+	s, err := token.SignedString(privateKey)
+
+	fmt.Printf("Token: %v %v", s, err)
```
AlexanderYastrebov, line 45: To log something please pass `t *testing.T` and use `t.Logf()`

filters/auth/jwt_validation_test.go
```diff
@@ -0,0 +1,210 @@
...
+	err = parsedToken.Claims.Valid()
+
+	if err != nil {
+		t.Errorf("Failed token: %v", err)
+		return
+	}
+
+}
+
+func parsePublicKey(encodedKey string) *rsa.PublicKey {
```
AlexanderYastrebov, line 73: I think there is a method to get public key from private key https://pkg.go.dev/crypto/rsa#PrivateKey.Public

filters/auth/jwt_validation_test.go
```diff
@@ -0,0 +1,210 @@
...
+		authBaseURL: testAuthPath,
+		args:        []interface{}{},
+		hasAuth:     true,
+		auth:        createToken(),
+		expected:    http.StatusOK,
+	}} {
+		t.Run(ti.msg, func(t *testing.T) {
+			if ti.msg == "" {
+				t.Fatalf("unknown ti: %+v", ti)
+			}
```
AlexanderYastrebov, line 156: Please remove this, we don't need to check what we wrote ourselves two lines above.
I general I suggest to keep tests as concise as possible otherwise it is hard to follow and important stuff is obscured

filters/auth/jwt_validation_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+			f2 := f.(*jwtValidationFilter)
+			defer f2.Close()
+
+			fr := make(filters.Registry)
+			fr.Register(spec)
+			r := &eskip.Route{Filters: []*eskip.Filter{{Name: spec.Name(), Args: args}}, Backend: backend.URL}
+
+			proxy := proxytest.New(fr, r)
+			defer proxy.Close()
```
AlexanderYastrebov, line 179: Lets move proxy and route initialization out of the main test loop

filters/auth/jwt_validation_test.go
```diff
@@ -0,0 +1,210 @@
...
+		t.Run(ti.msg, func(t *testing.T) {
+			if ti.msg == "" {
+				t.Fatalf("unknown ti: %+v", ti)
+			}
+
+			var spec = NewJwtValidation(testAuthTimeout)
+
+			args := []interface{}{testOidcConfig.Issuer}
+			args = append(args, ti.args...)
+			f, err := spec.CreateFilter(args)
```
AlexanderYastrebov, line 162: We do not need to test filter creation here, I suggest to parse route and create proxy see e.g.
https://github.com/zalando/skipper/blob/2059778e1924aaafb86accfc0b27e10245f13117/proxy/backendratelimit_test.go#L95-L101

filters/auth/jwt_validation_test.go
```diff
@@ -0,0 +1,210 @@
...
+				return
+			}
+
+			req, err := http.NewRequest("GET", reqURL.String(), nil)
+			if err != nil {
+				t.Errorf("failed to create request %v", err)
+				return
+			}
+
+			if ti.hasAuth {
```
AlexanderYastrebov, line 193: Could just check `ti.auth != ""` so there is no need to have `ti.hasAuth` at all

filters/auth/jwt_validation_test.go
abinet: But we need to recreate proxy for each new test because of new spec and route or do I miss something?

AlexanderYastrebov: The spec and filter has the same args so we can have single proxy with a route for multiple tests. This test tests different tokens but not different filter definitions.

filters/auth/jwt_validation.go
abinet: It is here https://github.com/abinet/skipper/blob/master/skipper.go#L1235
And this is the reason why second constructor is being needed.

abinet:
@AlexanderYastrebov can you please review the latest changes.

docs/reference/filters.md
```diff
@@ -1192,6 +1192,21 @@ Read client-id and client-secret from environment variables
 secureOauthTokenintrospectionAllKV("issuerURL", "", "", "k1", "v1", "k2", "v2")
 ```
 
+## jwtValidation
+
+The filter parses bearer jwt token from Authorization header and validates the signature using public keys
+discovered via /.well-known/openid-configuratioon endpoint. Takes issuer url as single parameter.
```
AlexanderYastrebov, line 1198: configuratioon -> configuration

docs/reference/filters.md
```diff
@@ -1192,6 +1192,21 @@ Read client-id and client-secret from environment variables
...
+
+The filter parses bearer jwt token from Authorization header and validates the signature using public keys
+discovered via /.well-known/openid-configuratioon endpoint. Takes issuer url as single parameter. 
+The filter stores token palyoad into the state bag where it can be used by oidcClaimsQuery() or forwardTokenPart()
+ 
+
+Examples:
+
+```
+jwtValidation("issuerURL")
```
AlexanderYastrebov, line 1205: Lets provide some working example - it is not clear from the docs whether `/.well-known/openid-configuration` should be part of this url or not.

docs/reference/filters.md
```diff
@@ -1209,6 +1224,20 @@ forwardToken("X-Tokeninfo-Forward")
...
+The filter takes the comma-separated lists of header names and fields as its first and second arguments. The corresponding fields from the result of token info, token introspection or oidc user info are added as 
+corresponding headers when the request is passed to the backend. 
+
+If this filter is used when there is no token introspection, token info or oidc user info data
+then it does not have any effect.
+
+Examples:
+
+```
+forwardTokenField("X-Tokeninfo-Forward-Oid,X-Tokeninfo-Forward-Sub", "oid,sub")
```
AlexanderYastrebov, line 1238: Looking at the current implementation this would produce 
* `X-Tokeninfo-Forward-Oid: <oidc value>`
* `X-Tokeninfo-Forward-Sub: <sub value>`

Lets simplify it and only support single header and field name - if users need multiple headers they could use sequence of filters: `forwardTokenField("X-Tokeninfo-Forward-Oid", "oid") -> forwardTokenField("X-Tokeninfo-Forward-Sub", "sub")`.

filters/auth/forwardtokenfield.go
```diff
@@ -0,0 +1,98 @@
...
+func NewForwardTokenField() filters.Spec {
+	return &forwardTokenFieldSpec{}
+}
+
+func (s *forwardTokenFieldSpec) Name() string {
+	return ForwardTokenFieldName
+}
+
+func (*forwardTokenFieldSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) < 1 {
```
AlexanderYastrebov, line 35: Should be `if len(args) != 2`

filters/auth/forwardtokenfield_test.go
```diff
@@ -0,0 +1,61 @@
...
+	container.Claims["key2"] = "value2"
+
+	state[oidcClaimsCacheKey] = container
+
+	c := &filtertest.Context{FRequest: &http.Request{}, FStateBag: state}
+	c.FRequest.Header = make(http.Header)
+
+	f.Request(c)
+
+	if c.FRequest.Header["Header1"] == nil || c.FRequest.Header["Header1"][0] != "value1" {
```
AlexanderYastrebov, line 54: https://pkg.go.dev/net/http#Header.Get

filters/auth/forwardtokenfield_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	f, err := spec.CreateFilter([]interface{}{"Header1,Header2,Header3", "key1,key2,key3"})
+	if err != nil {
+		t.Fatal("failed to create filter")
+	}
+
+	var state = make(map[string]interface{})
+
+	var container tokenContainer
+	container.Claims = make(map[string]interface{})
```
AlexanderYastrebov, line 42: Can we use map literal here?
```go
container.Claims = map[string]interface{}{
  "key1": "value1",
  "key2": "value2",
}
```

filters/auth/jwt_validation.go
AlexanderYastrebov: I don't get it. Lets have single constructor with options, timeouts etc. Tests can introduce helper method to use whatever timeout or NoopTracer needed.

AlexanderYastrebov: I.e. have only `NewJwtValidationWithOptions`

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,171 @@
...
+func (s *jwtValidationSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	sargs, err := getStrings(args)
+	if err != nil {
+		return nil, err
+	}
+	if len(sargs) < 1 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	issuerURL := sargs[0]
```
AlexanderYastrebov, line 71: It should check there is exactly one argument
```go
if len(args) != 1 {
}
issuerURL, ok := args[0].(string)
if !ok {
}
```

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,171 @@
...
+		jwksUri: cfg.JwksURI,
+	}
+
+	return f, nil
+}
+
+func getKeyFunction(url string) (jwks *keyfunc.JWKs, err error) {
+	if jwks, ok := jwksMap[url]; ok {
+		return jwks, nil
+	}
```
AlexanderYastrebov, line 88: `getKeyFunction` is called from the concurrent `Request` so `jwksMap` should be guarded with mutex

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Create the keyfunc options. Use an error handler that logs. Refresh the JWKs when a JWT signed by an unknown KID
+	// is found or at the specified interval. Rate limit these refreshes. Timeout the initial JWKs refresh request after
+	// 10 seconds. This timeout is also used to create the initial context.Context for keyfunc.Get.
```
AlexanderYastrebov, line 92: This comment seems redundant, lets remove, reader should rather consult keyfunc library docs if needed

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,171 @@
...
+	refreshInterval := time.Hour
+	refreshRateLimit := time.Minute * 5
+	refreshTimeout := time.Second * 10
+	refreshUnknownKID := true
+	options := keyfunc.Options{
+		RefreshErrorHandler: func(err error) {
+			log.Errorf("There was an error with the jwt.KeyFunc\nError:%s\n", err.Error())
```
AlexanderYastrebov, line 99: It seems it would be called when key refresh fails - lets be specific and log that refresh failed as well as jwks url to simplify debugging.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,171 @@
...
+		},
+		RefreshInterval:   &refreshInterval,
+		RefreshRateLimit:  &refreshRateLimit,
+		RefreshTimeout:    &refreshTimeout,
+		RefreshUnknownKID: &refreshUnknownKID,
+	}
+
+	jwks, err = keyfunc.Get(url, options)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get the JWKs from the given URL %s Error:%s", url, err.Error())
```
AlexanderYastrebov, line 109: Please use %w to wrap the error, see https://pkg.go.dev/fmt#Errorf

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,171 @@
...
+	if !ok {
+		token, ok := getToken(r)
+		if !ok || token == "" {
+			unauthorized(ctx, "", missingToken, "", "")
+			return
+		}
+
+		jwks, err := getKeyFunction(f.jwksUri)
+		if err != nil {
+			log.Fatalf("Failed to get the JWKs from the given URL %s. Error:%s", f.jwksUri, err.Error())
```
AlexanderYastrebov, line 130: I am wondering should we rather pass debugInfo to the `unauthorized()` instead of logging error here. Also %v format specifier formats error without the need to call .Error()

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,171 @@
...
+			return
+		} else if !parsedToken.Valid {
+			log.Errorf("Invalid token")
+			unauthorized(ctx, "", invalidToken, "", "")
+			return
+		} else if parsedToken.Header["alg"] == nil {
+			log.Errorf("alg must be defined")
+			unauthorized(ctx, "", invalidToken, "", "")
+			return
+		}
```
AlexanderYastrebov, line 149: I think jwt lib verifies presence of `alg` so this branch is unreachable
https://github.com/golang-jwt/jwt/blob/2bd8ee77fc3a6b7116c5b5121cd7d07324423a85/parser.go#L139-L145

We should be careful about `none` algorithm though and disable it explicitly, see https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/

I think it is disabled by default https://github.com/golang-jwt/jwt#compliance but it is worth to verify/add a test.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,171 @@
...
+
+		info.Claims = claims
+	} else {
+		info = infoTemp.(tokenContainer)
+	}
+
+	sub, ok := info.Claims["sub"].(string)
```
AlexanderYastrebov, line 156: Why `sub` claim is required here?

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,171 @@
...
+	authorized(ctx, sub)
+
+	ctx.StateBag()[oidcClaimsCacheKey] = info
+}
+
+func (f *jwtValidationFilter) Response(filters.FilterContext) {}
+
+// Close cleans-up the scheduler
+func (f *jwtValidationFilter) Close() {
+}
```
AlexanderYastrebov, line 171: The comment and the method are misleading, lets remove

filters/auth/jwt_validation_test.go
```diff
@@ -0,0 +1,177 @@
...
+	parsedToken, err := jwt.Parse(s, func(token *jwt.Token) (interface{}, error) {
+		return &privateKey.PublicKey, nil
+	})
+
+	if err != nil {
+		t.Errorf("Failed to json decode: %v", err)
+		return
+	}
+
+	err = parsedToken.Claims.Valid()
```
AlexanderYastrebov, line 60: Should it check `parsedToken.Valid` instead?

filters/auth/jwt_validation_test.go
```diff
@@ -0,0 +1,177 @@
...
+			w.WriteHeader(489)
+			return
+		}
+
+		allKeys := map[string][]interface{}{}
+		allKeys["keys"] = append(allKeys["keys"], map[string]interface{}{"kid": kid, "n": base64.RawURLEncoding.EncodeToString(privateKey.PublicKey.N.Bytes()), "e": "AQAB"})
+		if r.URL.Path != testAuthPath {
+			w.WriteHeader(488)
+			return
+		}
```
AlexanderYastrebov, line 89: Do we need this?

filters/auth/jwt_validation_test.go
```diff
@@ -0,0 +1,177 @@
...
+	defer cli.Close()
+
+	backend := httptest.NewServer(http.HandlerFunc(func(http.ResponseWriter, *http.Request) {}))
+	defer backend.Close()
+
+	authServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != "GET" {
+			w.WriteHeader(489)
+			return
+		}
```
AlexanderYastrebov, line 82: Do we need this?

filters/auth/jwt_validation_test.go
```diff
@@ -0,0 +1,177 @@
...
+		if r.URL.Path != testAuthPath {
+			w.WriteHeader(488)
+			return
+		}
+
+		e := json.NewEncoder(w)
+		err2 := e.Encode(allKeys)
+		if err2 != nil && err2 != io.EOF {
+			t.Errorf("Failed to json encode: %v", err2)
+		}
```
AlexanderYastrebov, line 95:
```go
fmt.Fprintf(w, `{"keys":[{"kty":"RSA", "alg":"RS256", "kid": "%s", "n":"%s","e":"AQAB"}]}`, 
kid, base64.RawURLEncoding.EncodeToString(privateKey.PublicKey.N.Bytes()))
```
`AQAB` here is base64 of public exponent 65537 https://cs.opensource.google/go/go/+/refs/tags/go1.17:src/crypto/rsa/rsa.go;l=260

filters/auth/jwt_validation_test.go
```diff
@@ -0,0 +1,177 @@
...
+		auth:     "invalid-token",
+		expected: http.StatusUnauthorized,
+	}, {
+		msg:      "jwtValidation: valid token",
+		auth:     createToken(t),
+		expected: http.StatusOK,
+	}} {
+		t.Run(ti.msg, func(t *testing.T) {
+
+			req.Header.Set(authHeaderName, authHeaderPrefix+ti.auth)
```
AlexanderYastrebov, line 162: Lets create request a new here
```go
req, _ := http.NewRequest("GET", proxy.URL, nil)
```

Lets move prefix to the test case field such that we can have tests like:
```go
        {
		msg:      "invalid token type",
		auth:     "Wearer " +createToken(t) ,
		expected: http.StatusUnauthorized,
	},{
		msg:      "empty toke",
		auth:     "Bearer " ,
		expected: http.StatusUnauthorized,
	},
```

filters/auth/jwt_validation_test.go
```diff
@@ -0,0 +1,177 @@
...
+	r := &eskip.Route{Filters: []*eskip.Filter{{Name: spec.Name(), Args: args}}, Backend: backend.URL}
+
+	proxy := proxytest.New(fr, r)
+	defer proxy.Close()
+
+	reqURL, err := url.Parse(proxy.URL)
+	if err != nil {
+		t.Errorf("Failed to parse url %s: %v", proxy.URL, err)
+		return
+	}
```
AlexanderYastrebov, line 140: This is redundant, proxy url is valid

filters/auth/jwt_validation_test.go
AlexanderYastrebov: Lets also add a test for a token with `none` algorithm. The token should be valid according to JWT but the filter should refuse to accept it. See https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/

filters/auth/jwt_validation_test.go
```diff
@@ -0,0 +1,177 @@
...
+	err = parsedToken.Claims.Valid()
+
+	if err != nil {
+		t.Errorf("Failed token: %v", err)
+		return
+	}
+
+}
+
+func TestJWTValidation(t *testing.T) {
```
AlexanderYastrebov, line 69: This tests various tokens. There should be another test set that covers different key failure scenarios like permanent failure to load keys or key refresh failure

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,146 @@
...
+
+	authorized(ctx, sub)
+
+	ctx.StateBag()[oidcClaimsCacheKey] = info
+}
+
+func (f *jwtValidationFilter) Response(filters.FilterContext) {}
+
+func (f *jwtValidationFilter) Close() {
+}
```
AlexanderYastrebov, line 146: Lets remove this as it is never called

filters/auth/forwardtokenfield.go
```diff
@@ -0,0 +1,89 @@
...
+func (s *forwardTokenFieldSpec) Name() string {
+	return ForwardTokenFieldName
+}
+
+func (*forwardTokenFieldSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) != 2 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	headerName := args[0].(string)
```
AlexanderYastrebov, line 38:
```go
headerName, ok := args[0].(string)
if !ok {
}
```
the same is for the second arg.

The https://github.com/zalando/skipper/pull/1814 targets to eliminate such boilerplate

abinet: @AlexanderYastrebov but #1814 is not merged to master yet, how can I use it in my PR?

AlexanderYastrebov: It is just a note that hopefully in the future we'll get rid of the boilerplate, please use `headerName, ok := args[0].(string)` for now.

filters/auth/jwt_validation.go
abinet: we have to provide something in authorized() call, right?

abinet:
@AlexanderYastrebov please check my updates

filters/auth/forwardtokenfield.go
```diff
@@ -0,0 +1,95 @@
...
+
+	headerName, ok := args[0].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	valid := httpguts.ValidHeaderFieldName(headerName)
+	if !valid {
+		return nil, fmt.Errorf("header name %s in invalid", headerName)
+	}
```
AlexanderYastrebov, line 46: typo: _is_ invalid
minor: `if !httpguts.ValidHeaderFieldName(headerName) {`

filters/auth/forwardtokenfield.go
```diff
@@ -0,0 +1,95 @@
...
+
+	field, ok2 := args[1].(string)
```
AlexanderYastrebov, line 48: `ok` can be reused here

filters/auth/forwardtokenfield.go
```diff
@@ -0,0 +1,87 @@
...
+	if len(args) != 2 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	headerName, ok := args[0].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	valid := httpguts.ValidHeaderFieldName(headerName)
```
szuecs, line 43: why not using `http.CanonicalHeaderKey(headerName)`, so we don't need the error handling here?

filters/auth/forwardtokenfield_test.go
```diff
@@ -0,0 +1,83 @@
...
+		t.Error("wrong name")
+	}
+
+	f, err := spec.CreateFilter([]interface{}{"Header1", "claims.key1"})
+	if err != nil {
+		t.Fatal("failed to create filter")
+	}
+
+	var state = map[string]interface{}{
+		oidcClaimsCacheKey: tokenContainer{
```
szuecs, line 40: please create a table test to test all different cache keys

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,143 @@
...
+func (s *jwtValidationSpec) Name() string {
+	return JwtValidationName
+}
+
+func (s *jwtValidationSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	sargs, err := getStrings(args)
+	if err != nil {
+		return nil, err
+	}
+	if len(sargs) != 1 {
```
szuecs, line 48: this should be checked first to omit string creation

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,143 @@
...
+	defer m.Unlock()
+
+	if jwks, ok := jwksMap[url]; ok {
+		return jwks, nil
+	}
+
+	refreshInterval := time.Hour
+	refreshRateLimit := time.Minute * 5
+	refreshTimeout := time.Second * 10
+	refreshUnknownKID := true
```
szuecs, line 77: all these vars can be package global const instead to omit creating garbage

filters/auth/forwardtokenfield.go
abinet: According to `http.CanonicalHeaderKey(s string)` docs:
```
// If s contains a space or invalid header field bytes, it is
// returned without modifications.
```

So for me it means we still have to check result of `http.CanonicalHeaderKey(headerName)`

filters/auth/jwt_validation.go
abinet: @szuecs keyfunc.Options{} requires pointers not just values. That's why I have not declared all this settings as constants.

szuecs: Interesting you can't get a ptr from a const, I just checked it.

filters/auth/forwardtokenfield.go
szuecs: Good point

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,143 @@
...
+		RefreshRateLimit:  &refreshRateLimit,
+		RefreshTimeout:    &refreshTimeout,
+		RefreshUnknownKID: &refreshUnknownKID,
+	}
+
+	jwks, err = keyfunc.Get(url, options)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get the JWKs from the given URL %s Error:%w", url, err)
+	}
+	jwksMap[url] = jwks
```
szuecs, line 92: I think you only want to lock the line above. Right now you will block all requests, when you try to fetch keys from JWKs URLs, which could serve very slow and might lead to latency spikes.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	return f, nil
+}
+
+func getKeyFunction(url string) (jwks *keyfunc.JWKs, err error) {
+	m.Lock()
+	defer m.Unlock()
+
+	if jwks, ok := jwksMap[url]; ok {
```
szuecs, line 70: this also needs to be guarded of course if you drop the function lock/unlock

filters/auth/jwt_validation.go
szuecs: Thinking more about this problem, I think we should not automatically refresh from Request(), but have a separate goroutine that handles jwks key fetching, such that Request() will never block on waiting for a call to JWKs URIs.
A mutex to guard access to get the keys from a map is fine.

abinet: Wait wait. That is exactly what this code does. On every request it checks if we do have already existing function for the given jwksuri. If it is not the case the function will be created and stored in map. And all the refresh logic is being handled by function itself. So we only using mutex to prevent simultaneous acesses to the caching map.

MicahParks: The background refresh goroutine is launched during `keyfunc.Get`, if some certain options are given:

https://github.com/MicahParks/keyfunc/blob/4863ebcdd8cf9f667265e026039aa1b9ce3bd75a/get.go#L55

Make sure to call the [`JWKs.EndBackground`](https://pkg.go.dev/github.com/MicahParks/keyfunc#JWKs.EndBackground) method when you don't want this goroutine anymore.

filters/auth/jwt_validation.go
MicahParks: You could create a global un-exported variable that would serve the purpose of avoiding allocation/garbage collection.

filters/auth/jwt_validation.go
szuecs: Then we need more logic, because we would create a goroutine leak, because on every call of `getKeyFunction()` we have the static configuration that will spawn the background goroutine.
So I see the following change to be made:
1. on CreateFilter() you run once `getKeyFunction()` with the `refresh*` options to spawn the background refresher goroutine
2. the call to `getKeyFunction()` by filter `Request()` (hot path) has to have none of the options mentioned in https://github.com/MicahParks/keyfunc/blob/4863ebcdd8cf9f667265e026039aa1b9ce3bd75a/get.go#L46

abinet: @szuecs I am not quite sure if I understood you properly but check my latest changes - now the options and function are just created on filter creation and expected to be provided on request.

filters/auth/jwt_validation.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	f := &jwtValidationFilter{
+		jwksUri: cfg.JwksURI,
+	}
+
+	return f, nil
+}
+
+func getKeyFunction(url string, creationAllowed bool) (jwks *keyfunc.JWKs, err error) {
```
AlexanderYastrebov, line 76: I do not get it - why do we need `creationAllowed bool` if it is true in both `CreateFilter` and `Request` call sites.

If the goal is to ensure that keyFunc exists before `Request` then lets have two methods:
- `registerKeyFunc(jwksUri)` that would create the function, and put it into the map
- `getKeyFunc(jwksUri)` - this simply returns a value from the map, no need to check anything as the `jwksUri` is the same as was used  during creation

abinet: it was a typo, creationAllowed should be false for request. Fixed this in nex commit

AlexanderYastrebov: @abinet The point is that there are two actions `register` and `get` that should be separate functions - the goal is clarity.
E.g. `register` does not have to return keyFunk so
```go
_, err = getKeyFunction(cfg.JwksURI, true)
```
would be
```go
err := registerKeyFunction(cfg.JwksURI)
```
while `get` does not have to check the presence of the value as it always has to be there after `register` so
```go
jwks, err := getKeyFunction(f.jwksUri, false)
if err != nil {
...
}
```
becomes
```go
jwks := getKeyFunction(f.jwksUri)
```

filters/auth/forwardtokenfield.go
```diff
@@ -0,0 +1,87 @@
...
+
+type (
+	forwardTokenFieldSpec   struct{}
+	forwardTokenFieldFilter struct {
+		HeaderName string
+		Field      string
+	}
+)
+
+// NewForwardTokenPart creates a filter to forward fields from token info or
```
szuecs, line 23: wrong godoc: comment has to start with `NewForwardTokenField` (func name)

szuecs:
Other than the godoc is fine to merge IMO.
We could make the critical sections even smaller in `registerKeyFunction()`. We could split it into 2 there, because we only need to guard the map access. 
If you want to go further you could also change the Mutex to RWMutex and use RLock() and Lock(), because we rarely use  the write lock to update the map and we read much more often (every request), but that's optimization which we might not need.

abinet:
@szuecs resolved the doc issue and replaced with RWMutex. However I don't quite understand why pr /tests check have started to fail.

docs/reference/filters.md
```diff
@@ -1192,6 +1192,21 @@ Read client-id and client-secret from environment variables
 secureOauthTokenintrospectionAllKV("issuerURL", "", "", "k1", "v1", "k2", "v2")
 ```
 
+## jwtValidation
+
+The filter parses bearer jwt token from Authorization header and validates the signature using public keys
+discovered via /.well-known/openid-configuration endpoint. Takes issuer url as single parameter. 
+The filter stores token palyoad into the state bag where it can be used by oidcClaimsQuery() or forwardTokenPart()
```
AlexanderYastrebov, line 1199: typo: `payload` 
Or maybe it should be "claims"

AlexanderYastrebov:
@abinet Please update PR description to match the implementation

AlexanderYastrebov:
@abinet 
> However I don't quite understand why pr /tests check have started to fail.

Run `make fmt` to fix the formatting

abinet:
> 
> 
> @abinet
> 
> > However I don't quite understand why pr /tests check have started to fail.
> 
> Run `make fmt` to fix the formatting



> 
> 
> @abinet
> 
> > However I don't quite understand why pr /tests check have started to fail.
> 
> Run `make fmt` to fix the formatting

@AlexanderYastrebov  I've just merged all from zalando master. If I run make fmt it introduce the changes in files not modified in my PR. That means your master build should fail as well.

AlexanderYastrebov:
> However I don't quite understand why pr /tests check have started to fail.
> Run make fmt to fix the formatting

Yes, looks like golang 1.17 gofmt adds new comments https://tip.golang.org/doc/go1.17#gofmt
I will open a PR to fix

szuecs:
please rebase to master as we merged the fix for master, thanks

szuecs:
The last missing piece IMO is https://github.com/zalando/skipper/pull/1811#issuecomment-915403594.
To be more clear, I would like to have it like this:

```
func registerKeyFunction(url string) (err error) {
	m.RLock()
	if _, ok := jwksMap[url]; ok {
		m.Unlock()
		return nil
	}
	m.Unlock()
....
	jwks, err := keyfunc.Get(url, options)
	if err != nil {
		return fmt.Errorf("failed to get the JWKs from the given URL %s Error:%w", url, err)
	}
	m.Lock()
	jwksMap[url] = jwks
	m.Unlock()
	return nil
}
```
I want to have it because `keyfunc.Get(` can take long blocking all Requests with the write lock.
And we have now only small critical sections, of course we needed to write a bit more lines, but this is fine.

szuecs:
:+1:

aryszka:
:+1:

szuecs:
:+1:

szuecs:
@abinet Thanks great work!

PR #1813, opened 2021-07-08
Increases number of algorithm test iterations

AlexanderYastrebov:
Algorithm test performs `R` requests towards `N` backend
endpoints, counts requests per endpoint and verifies that expected
number of endpoints received at least one request.

For `random` and `powerOfRandomNChoices` algorithms it may happen
(by a bad luck) that some backend does not receive any requests which
fails the test.

The issue could be reproduced by
```
while go test -count=1 github.com/zalando/skipper/loadbalancer -run TestApply; do true; done
```
or by decresing `R`.

This change increases `R` from `100` to `1000` to reduce failure
chances.

### DIFF
--- a/loadbalancer/algorithm_test.go
+++ b/loadbalancer/algorithm_test.go
@@ -209,7 +209,8 @@ func TestSelectAlgorithm(t *testing.T) {
 }
 
 func TestApply(t *testing.T) {
-	N := 10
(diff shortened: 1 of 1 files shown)


aryszka:
:+1:

AlexanderYastrebov:
:+1:

Issue #814, opened 2018-10-06
refactor filter error handling


szuecs:
Instead of using filters.ErrInvalidFilterParameters we could add a function that returns an error to parameterize the error to log it centrally.

## Background

source: https://github.com/zalando/skipper/pull/668/files#r205731495

With the experience gained since the current approach was introduced, my ideal approach would be something like this:

*    return a typed error that contains details specific to the local code and reason
*    the type should help deciding in the caller code how to handle it or to just return it
*    if logging, log it as close to the root of the call tree as possible

PR #1814, opened 2021-07-10
experimental: Filter argument converter

AlexanderYastrebov:
Introduces filter argument converter that simplifies error handling and argument length checking and eliminates common `CreateFilter` boilerplate as well as unifies argument type conversion.
The implementation is based on the https://blog.golang.org/errors-are-values idea to postpone error handling until all expected arguments are processed. 
This enables:
* check arguments error only once
* richer error messages that describe all problems with arguments, not only the first encountered problem

The change touches several existing filters to demonstrate the usage.
Future extension may also include methods for common argument checks like `IntGreater(123)`, `IntBetween(1, 9)`, `StringNonEmpty()`, etc.

If the draft looks promising then many of the existing filter constructors could be simplified.

Improves on #814


filters/builtin/timeout.go
```diff
@@ -17,24 +17,8 @@ func NewBackendTimeout() filters.Spec {
...
-		}
-		tf.timeout = d
-	case time.Duration:
-		tf.timeout = v
-	default:
-		return nil, filters.ErrInvalidFilterParameters
-	}
-	return &tf, nil
+	a := filters.Args(args)
+	return &timeout{a.Duration()}, a.Err()
```
AlexanderYastrebov, line 21: Expects single valid duration argument as parsable string or decimal number of milliseconds in any other case (no arguments, several arguments, invalid or negative duration string) `a.Err()` is not null and contains detailed error message like:
```
$ bin/skipper -inline-routes='* -> backendTimeout() -> <shunt>' 2>&1 | grep error
[APP]time="2021-07-10T23:22:36+02:00" level=error msg="failed to process route (): failed to create backendTimeout filter: expects 1 argument"

$ bin/skipper -inline-routes='* -> backendTimeout("xyz") -> <shunt>' 2>&1 | grep error
[APP]time="2021-07-10T23:22:54+02:00" level=error msg="failed to process route (): failed to create backendTimeout filter: time: invalid duration \"xyz\""

$ bin/skipper -inline-routes='* -> backendTimeout("1s", "x") -> <shunt>' 2>&1 | grep error
[APP]time="2021-07-10T23:23:11+02:00" level=error msg="failed to process route (): failed to create backendTimeout filter: expects 1 argument"

$ bin/skipper -inline-routes='* -> backendTimeout("-10s") -> <shunt>' 2>&1 | grep error
[APP]time="2021-07-10T23:23:20+02:00" level=error msg="failed to process route (): failed to create backendTimeout filter: duration -10s is negative"

$ bin/skipper -inline-routes='* -> backendTimeout("x", "y") -> <shunt>' 2>&1 | grep error
[APP]time="2021-07-10T23:24:54+02:00" level=error msg="failed to process route (): failed to create backendTimeout filter: expects 1 argument, time: invalid duration \"x\""
```

Note that constructor always returns `&timeout{}` value even in case of error - this should not be a big problem as it is common to check error first and throw away the value if error is not nil.

filters/circuit/breaker.go
```diff
@@ -91,99 +68,30 @@ func (s *spec) Name() string {
...
 		settings: circuit.BreakerSettings{
 			Type:             circuit.ConsecutiveFailures,
-			Failures:         failures,
-			Timeout:          timeout,
-			HalfOpenRequests: halfOpenRequests,
-			IdleTTL:          idleTTL,
+			Failures:         a.Int(),
+			Timeout:          a.DurationOr(0),
+			HalfOpenRequests: a.IntOr(0),
+			IdleTTL:          a.DurationOr(0),
```
AlexanderYastrebov, line 78: Optional arguments: converted if present otherwise default value is returned

filters/diag/diag.go
```diff
@@ -182,15 +181,15 @@ func NewNormalResponseLatency() filters.Spec { return &jitter{typ: normalRespons
...
+	a := filters.Args(args)
+	len, err := a.Int64(), a.Err()
+	if err != nil {
+		return nil, err
 	}
-
-	if l, ok := args[0].(float64); ok {
-		return &random{rand: rand.New(rand.NewSource(time.Now().UnixNano())), len: int64(l)}, nil // #nosec
-	} else {
+	if len < 0 {
```
AlexanderYastrebov, line 189: Arguments could be assigned to new variables and then addition validation performed.

filters/diag/diag.go
```diff
@@ -214,30 +213,14 @@ func (r *random) Response(ctx filters.FilterContext) {}
...
-		len = v
-	default:
-		return nil, filters.ErrInvalidFilterParameters
+	a := filters.Args(args)
+	text, len, err := a.String(), a.Int64(), a.Err()
+	if err != nil {
+		return nil, err
 	}
-	if len < 0 {
+	if text == "" || len < 0 {
```
AlexanderYastrebov, line 221: The converter could be extended to provide common checks like:
```go
a := filters.Args(args)
return &repeat{[]byte(a.StringNonEmpty()), a.Int64GreaterEqual(0)}, a.Err()
```

filters/args.go
```diff
@@ -0,0 +1,181 @@
...
+			}
+		default:
+			a.err("%v is not a duration", x)
+			return
+		}
+		if d < 0 {
+			a.err("duration %v is negative", x)
+			return
+		}
+		return d
```
AlexanderYastrebov, line 139: The package should probably expose conversion methods to be used directly if needed

AlexanderYastrebov: Fixed

docs/reference/filters.md
```diff
@@ -1560,7 +1561,7 @@ Memory usage examples:
 Parameters:
 
 * number of allowed requests per time period (int)
-* time period for requests being counted (time.Duration)
+* time period for requests being counted (duration)
```
AlexanderYastrebov, line 1564: Needs checking, as it also accepts (undocumented) integer as number of _seconds_

docs/reference/filters.md
```diff
@@ -4,8 +4,9 @@ The parameters can be strings, regex or float64 / int
 
 * `string` is a string surrounded by double quotes (`"`)
 * `regex` is a regular expression, surrounded by `/`, e.g. `/^www\.example\.org(:\d+)?$/`
-* `int` / `float64` are usual (decimal) numbers like `401` or `1.23456`
-* `time` is a string in double quotes, parseable by [time.Duration](https://godoc.org/time#ParseDuration))
+* `int` an integer decimal number like `401` or `1.0`
+* `float64` a decimal number like `401` or `1.23456`
+* `duration` is a string in double quotes, parsable by [time.Duration](https://godoc.org/time#ParseDuration) or a decimal number of milliseconds
```
AlexanderYastrebov, line 9: It is nice to have single duration type, but we need to check that all implementations support decimal milliseconds. E.g. ratelimit family supports undocumented integer number of _seconds_

filters/args.go
```diff
@@ -0,0 +1,232 @@
...
+func (a *FilterArgs) DurationOr(defaultValue time.Duration) time.Duration {
+	if a.pos >= len(a.args) {
+		return defaultValue
+	}
+	return a.Duration()
+}
+
+func (a *FilterArgs) Err() error {
+	var errs []string
+	if a.pos != len(a.args) {
```
aryszka, line 202: :heart: 

This was the first thing I looking for, and ecce!

filters/circuit/breaker.go
aryszka: There can be more cases. The variadic number of arguments come to my mind. In those cases the last call could return a slice.

AlexanderYastrebov: >  last call could return a slice

There is Strings() method for that that return the rest of args converted to string slice.

docs/reference/filters.md
szuecs: and the change should not drop the godoc name, but maybe link it: [time.Duration](https://godoc.org/time#ParseDuration)

AlexanderYastrebov:
While exploring the filter implementations I've found filters that support duration string argument and number argument interpreted as number of _milliseconds_ (e.g. `backendTimeout`). There are also filters that support duration string and number argument but number is interpreted as number of _seconds_ (ratelimit filters, though undocumented).

Duration string explicitly configures time interval while the usage of implicit number of milliseconds/seconds is prone to mistakes (and there is no convenience to use number over string or `time.Duration` - it can be always converted to `time.Duration` if needed) therefore I think ultimately implicit number support in the duration arguments should be abandoned.

To stay backward compatible this PR should provide a family of duration converters that support pure duration arguments (string/`time.Duration`) and implicit numbers of seconds/milliseconds. It is also premature to speak about unified duration type and thus filter docs changes should be reverted.

docs/reference/filters.md
AlexanderYastrebov: The idea was to define `duration` type once at the top and reference it from filter args but it is too early, see https://github.com/zalando/skipper/pull/1814#issuecomment-895053332

docs/reference/filters.md
AlexanderYastrebov: See https://github.com/zalando/skipper/pull/1814#issuecomment-895053332

filters/diag/diag.go
AlexanderYastrebov: It is possible to add validating methods but their number could quickly go out of hands. It is also not possible to cover all validation usecases anyway because validation may require cross-argument check and additional logic (e.g. library call).
Therefore the good balance seems to be to leave type conversion to the converter and implement validation on the caller side.

filters/args.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+	}
+	for _, err := range a.errs {
+		errs = append(errs, err.Error())
+	}
+
+	if len(errs) == 0 {
+		return nil
+	} else {
+		return errors.New(strings.Join(errs, ", "))
```
szuecs, line 279: hmm that does not work well with errors.Is https://pkg.go.dev/errors, but I also have no better idea right now.
Maybe with https://pkg.go.dev/errors#As to have a generic FilterArgError?

szuecs: I asked in slack and got pointed to https://github.com/hashicorp/go-multierror

AlexanderYastrebov: Yes, but do we care about? I think the returned error should be treated as a single untyped error. The goal here is to have detailed message. The go-multierror variant would look like this https://github.com/zalando/skipper/tree/filter-args-go-multierror

szuecs: I am thinking about library users, that might want to check `error.Is`, so maybe my first idea with creating FilterArgError type that has the error string that you concatenated. Maybe it's also too complicated and not really useful.

szuecs:
:+1:

AlexanderYastrebov:
The converter could be useful for predicates as well so maybe it should be moved to `eskip` or neighbor package as it defines both `Filter` and `Predicate` interface that use untyped `Args`.

PR #1815, opened 2021-07-14
Allowed external names

aryszka:
create a feature with a toggle to let the operator control which external names are allowed and which route group network backends or explicit LB endpoints are allowed in the generated routing in the Kubernetes scenario.


AlexanderYastrebov:
:+1:

aryszka:
:+1:

Issue #1757, opened 2021-04-16
catch performance regressions between versions


aryszka:
@jonathanbeber showed me this project: https://github.com/knqyf263/cob

We could potentially use it, but of course other solutions, too, to catch small performance regressions between old and new versions of Skipper. We could refactor and extend the scripts in the [skptesting](https://github.com/zalando/skipper/tree/master/skptesting) package that are currently not part of the regular build.

aryszka:
we can use the test dataclient and the root package of Skipper to get the highest possible level of a Skipper process programmatically in the go test specs, and maintain realistic test scenarios [testdataclient](https://pkg.go.dev/github.com/zalando/skipper@v0.13.52/routing/testdataclient).

PR #1818, opened 2021-07-20
test cob - continuous benchmark with routing package

szuecs:
test cob https://github.com/zalando/skipper/issues/1757 in a concurrent pipeline

@aryszka @AlexanderYastrebov Do we want to have this?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,6 +104,9 @@ bench: build $(TEST_PLUGINS)
 	#
 	for p in $(PACKAGES); do GO111MODULE=$(GO111) go test -bench . $$p; done
 
+bench-compare:
+	for p in net pathmux proxy routing script; do echo "#### bench compare github.com/zalando/skipper/$$p ###"; cob --bench-args "test -bench . -benchmem github.com/zalando/skipper/$$p" 2>/dev/null; done
+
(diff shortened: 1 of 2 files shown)


szuecs:
The bench failed because it shows "slower than before" in some benchmarks. Seems not very stable results, as we did not do anything that could have caused this.

ruiaraujo:
@szuecs Cloud environments are terrible for benchmarking, to work correctly it needs some kind of correction factor applied with a known probe.

szuecs:
@ruiaraujo yes that's true

AlexanderYastrebov:
> Cloud environments are terrible for benchmarking, to work correctly it needs some kind of correction factor applied with a known probe.

The default threshold seems to be 20% which feels quite high already. 
Running the branch on the local machine I still get large differences up to 30%

For the purpose of catching performance regressions we may need to create special performance-regression benchmarks (or a separate package) and run only them via https://github.com/knqyf263/cob#run-only-those-benchmarks-matching-a-regular-expression

Sometimes I get results printed and sometimes not i.e. the library seems to produce unstable results.
I also do not like that library looks abandoned such that we had to fork it.

Issue #1816, opened 2021-07-14
Goroutine leak in the `teeLoopback`


ponimas:
**Describe the bug**
We have set up traffic shadowing with these 2 routes (simplified):

```
v1: Path("/some-path")
&& Traffic(0.5)
-> consecutiveBreaker(120)
-> tracingSpanName("span")
-> teeLoopback("tag")
-> "https://backend1";

v2: Tee("tag")
&& True()
-> "https://backend2";
```
The `backend1` is an unreliable service and it's being restarted periodically. At the exact moments of the restart, we observe goroutine leaks. 
![Screenshot_20210714_173429](https://user-images.githubusercontent.com/34179/125649986-98bd0d62-c701-45d4-9483-0fe21fdd78be.png)

We were able to get a goroutine stack dump from `/debug/pprof/goroutine` - [goroutine.zip](https://github.com/zalando/skipper/files/6817287/goroutine.zip)

```pprof
(pprof) top 22
Showing nodes accounting for 4300, 100% of 4302 total
Dropped 83 nodes (cum <= 21)
      flat  flat%   sum%        cum   cum%
      4300   100%   100%       4300   100%  runtime.gopark
         0     0%   100%       2130 49.51%  bufio.(*Writer).ReadFrom
         0     0%   100%       2130 49.51%  github.com/zalando/skipper/proxy.(*Proxy).do
         0     0%   100%       2130 49.51%  github.com/zalando/skipper/proxy.(*Proxy).makeBackendRequest
         0     0%   100%       2130 49.51%  github.com/zalando/skipper/proxy.(*context).Loopback
         0     0%   100%       2130 49.51%  io.(*LimitedReader).Read
         0     0%   100%       2130 49.51%  io.(*PipeReader).Read
         0     0%   100%       2130 49.51%  io.(*pipe).Read
         0     0%   100%       2130 49.51%  io.Copy (inline)
         0     0%   100%       2130 49.51%  io.copyBuffer
         0     0%   100%       2130 49.51%  net/http.(*Request).write
         0     0%   100%       2130 49.51%  net/http.(*Transport).RoundTrip
         0     0%   100%       2130 49.51%  net/http.(*Transport).roundTrip
         0     0%   100%       2130 49.51%  net/http.(*persistConn).mapRoundTripError
         0     0%   100%       2130 49.51%  net/http.(*persistConn).roundTrip
         0     0%   100%       2131 49.54%  net/http.(*persistConn).writeLoop
[...]

PR #1819, opened 2021-07-21
fix possible goroutine leak on breakers in case used in conjuction with teeLoopback filter

szuecs:
fix https://github.com/zalando/skipper/issues/1816
in case of an open breaker and route has a teeLoopback filter like `breaker -> teeLoopback()`, we did not consume the body such that the cloned request spawned in a goroutine never finished


aryszka:
:+1:

szuecs:
:+1:

PR #1820, opened 2021-07-27
add possibility to omit adding defaultfilters for specifi routes

szuecs:
add possibility to omit adding prepend/append defaultfilters in case there are routes with a predicate that we want to omit. For example Tee() is a good example when you might want to omit the processing in case you want to drop headers on ingress that are meant internal only


szuecs:
This was just a hack how we could do it and we talked about that people should use their own dataclient to make this happen.

PR #1822, opened 2021-07-29
route converter to be able to switch filters and predicates

szuecs:


config/config.go
```diff
@@ -121,6 +121,8 @@ type Config struct {
 	InlineRoutes              string               `yaml:"inline-routes"`
 	AppendFilters             *defaultFiltersFlags `yaml:"default-filters-append"`
 	PrependFilters            *defaultFiltersFlags `yaml:"default-filters-prepend"`
+	EditorRoute               *routeChangerConfig  `yaml:"edit-route"`
```
aryszka, line 124: Alternatively it could be: `EditRoute`

config/config.go
```diff
@@ -634,6 +640,8 @@ func (c *Config) ToOptions() skipper.Options {
 			Prepend: c.PrependFilters.filters,
 			Append:  c.AppendFilters.filters,
 		},
+		Clone:              eskip.NewClone(c.CloneRoute.Reg, c.CloneRoute.Repl),
```
aryszka, line 643: I think this also should be called CloneRoute and EditRoute

config/config_test.go
```diff
@@ -112,6 +112,8 @@ func Test_NewConfig(t *testing.T) {
 				EtcdTimeout:                             2 * time.Second,
 				AppendFilters:                           &defaultFiltersFlags{},
 				PrependFilters:                          &defaultFiltersFlags{},
+				CloneRoute:                         &routeChangerConfig{},
```
aryszka: is there some formatting problem, or just github displays it a bit off for me?

config/routechangerconfig.go
```diff
@@ -0,0 +1,38 @@
...
+	return fmt.Sprintf("/%s/%s/", rcc.Reg, rcc.Repl)
+}
+
+func (rcc *routeChangerConfig) Set(value string) error {
+	a := strings.Split(value, "/")
+	if len(a) != 4 {
+		return fmt.Errorf("unexpected size of string split: %d", len(a))
+	}
+	reg, repl := a[1], a[2]
+	rcc.Reg = regexp.MustCompile(reg)
```
aryszka, line 27: let's not use MustCompile here. We can use just compile and return an error. Then the flag package can display the error with the right context, and there is no value in showing the stacktrace here

szuecs: changed

config/config_test.go
szuecs: fixed

eskip/eskip.go
```diff
@@ -21,6 +21,144 @@ var (
...
+
+func doOneRoute(rx *regexp.Regexp, repl []byte, r *Route) bool {
+	if rx == nil {
+		return false
+	}
+	var changed bool
+
+	for i, p := range r.Predicates {
+		ps := p.String()
+		pss := rx.ReplaceAll([]byte(ps), repl)
```
aryszka, line 127: ReplaceAllString? https://pkg.go.dev/regexp?utm_source=godoc#Regexp.ReplaceAllString

eskip/eskip.go
```diff
@@ -21,6 +21,144 @@ var (
...
+		sps := string(pss)
+		if ps == sps {
+			continue
+		}
+
+		pp, err := ParsePredicates(sps)
+		if err != nil {
+			return false
```
aryszka, line 135: this error should not really happen at this point, unless we have a bug in the eskip package (which can happen). Shall we log this error with log.Error here? (Besides the logging question, continuing is good.)

eskip/eskip.go
```diff
@@ -21,6 +21,144 @@ var (
...
+		fs := f.String()
+		fss := rx.ReplaceAll([]byte(fs), repl)
+		sfs := string(fss)
+		if fs == sfs {
+			continue
+		}
+
+		ff, err := ParseFilters(sfs)
+		if err != nil {
+			return false
```
aryszka, line 152: same question as with the predicates

config/config.go
szuecs: renamed

config/config.go
szuecs: renamed

eskip/eskip.go
```diff
@@ -107,6 +245,15 @@ type Predicate struct {
 	Args []interface{} `json:"args"`
 }
 
+func (p *Predicate) String() string {
+	args, err := getStringArgs(len(p.Args), p.Args)
```
aryszka, line 249: getStringArgs is not good here, it used to convert from []interface{} to []string, but the values must be string. Use argsString instead, see: https://github.com/zalando/skipper/pull/1822/files

eskip/eskip.go
```diff
@@ -116,6 +263,15 @@ type Filter struct {
 	Args []interface{} `json:"args"`
 }
 
+func (f *Filter) String() string {
+	args, err := getStringArgs(len(f.Args), f.Args)
```
aryszka, line 267: use argsString

eskip/eskip.go
```diff
@@ -107,6 +245,15 @@ type Predicate struct {
 	Args []interface{} `json:"args"`
 }
 
+func (p *Predicate) String() string {
```
aryszka, line 248: this package has a predicateString function, we could use it inside this method

eskip/eskip.go
```diff
@@ -116,6 +263,15 @@ type Filter struct {
 	Args []interface{} `json:"args"`
 }
 
+func (f *Filter) String() string {
```
aryszka, line 266: this package has a filterString method of the route, we could extract its functionality and use it inside this method. See: https://github.com/zalando/skipper/blob/master/eskip/string.go#L122. The reason why that function should be used is that does the escaping right, so that the regexp replace would be executed on the string that user expects.

skipper.go
```diff
@@ -259,6 +259,15 @@ type Options struct {
 	// DefaultFilters will be applied to all routes automatically.
 	DefaultFilters *eskip.DefaultFilters
 
+	// Clone is a PreProcessor, that will be applied to all routes automatically. It
+	// will clone all matching routes and apply changes to the
+	// cloned routes.
+	Clone *eskip.Clone
```
aryszka: CloneRoute

skipper.go
```diff
@@ -259,6 +259,15 @@ type Options struct {
...
+
+	// Editor will be applied to all routes automatically and
+	// will apply changes to all matching routes.
+	Editor *eskip.Editor
```
aryszka: EditRoute

eskip/eskip.go
szuecs: done

eskip/eskip.go
szuecs: fixed

config/routechangerconfig.go
```diff
@@ -0,0 +1,39 @@
...
+	"strings"
+)
+
+type routeChangerConfig struct {
+	Reg  *regexp.Regexp
+	Repl string
+}
+
+func (rcc routeChangerConfig) String() string {
+	if rcc.Reg == nil {
```
aryszka, line 15: this if might not be necessary here. We can just return whatever the user wrote.

eskip/eskip.go
szuecs: fixed

eskip/eskip.go
```diff
@@ -21,6 +22,146 @@ var (
...
+func (c *Clone) Do(routes []*Route) []*Route {
+	if c.reg == nil {
+		return routes
+	}
+
+	result := make([]*Route, len(routes), 2*len(routes))
+	copy(result, routes)
+	for _, r := range routes {
+		rr := new(Route)
+		*rr = *r
```
aryszka, line 95: we should take the canonical form of the route, because there are the legacy predicates, that are not listed in the []*Predicate slice. We can do this:

```
*rr = *r
rr = Canonical(rr)
```

eskip/eskip.go
```diff
@@ -21,6 +22,146 @@ var (
...
+}
+
+func (e *Editor) Do(routes []*Route) []*Route {
+	if e.reg == nil {
+		return routes
+	}
+
+	for i, r := range routes {
+		rr := new(Route)
+		*rr = *r
```
aryszka, line 77: just like in the Clone, we should use the canonical form rr here, as well

skipper.go

skipper.go

eskip/eskip.go
szuecs: fixed

eskip/eskip.go
szuecs: fixed

config/routechangerconfig.go
szuecs: it's workaround for getting <NIL> string in some outputs

eskip/eskip.go
szuecs: fixed

eskip/eskip.go
szuecs: fixed

eskip/eskip.go
szuecs: fixed as discussed

eskip/eskip.go
szuecs: fixed as discussed

aryszka:
:+1:

config/config.go
szuecs: changed

config/config_test.go
szuecs: this was fixed

config/config.go
szuecs: I think fixed

config/routechangerconfig.go
szuecs: good catch

eskip/eskip.go
szuecs: done

szuecs:
:+1:

PR #1826, opened 2021-08-10
Unmutes testlogger if test verbose output is enabled

AlexanderYastrebov:

### DIFF
--- a/logging/loggingtest/logger.go
+++ b/logging/loggingtest/logger.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"strings"
+	"testing"
 	"time"
 )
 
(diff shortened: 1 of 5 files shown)


logging/loggingtest/logger.go
```diff
@@ -96,8 +97,7 @@ func New() *Logger {
 	mute := make(chan bool)
 	quit := make(chan struct{})
 
-	// start muted to reduce spam during failing tests
-	muted := true
+	muted := !testing.Verbose()
```
AlexanderYastrebov, line 100: Within #1825 I've observed the problem that was not apparent unless testlogger is unmuted.

logging/loggingtest/logger.go
```diff
@@ -96,8 +97,7 @@ func New() *Logger {
 	mute := make(chan bool)
 	quit := make(chan struct{})
 
-	// start muted to reduce spam during failing tests
```
AlexanderYastrebov, line 99: I think rich log output is exactly what is needed to debug the failing test. This PR proposes to unmute when verbose test output is enabled which requires re-running failing test with `go test -v`. Maybe we should simply unmute testlogger unconditionally.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #1827, opened 2021-08-13
Record metric if a lifo filter drops a request

wndhydrnt:
After the release of #1725, it is sometimes hard to correlate HTTP status codes the action of a lifo filter.

This change increases one of two counters each time a lifo queue is either full or an entry times out. It improves the visibility of the actions a lifo filter has taken without spamming the logs.

### DIFF
--- a/filters/scheduler/lifo.go
+++ b/filters/scheduler/lifo.go
@@ -294,7 +294,6 @@ func request(q *scheduler.Queue, key string, ctx filters.FilterContext) {
 
 	done, err := q.Wait()
 	if err != nil {
-		// TODO: replace the log with metrics
 		switch err {
 		case jobqueue.ErrStackFull:
 			log.Debugf("Failed to get an entry on to the queue to process QueueFull: %v for host %s", err, ctx.Request().Host)
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -57,13 +57,14 @@ type QueueStatus struct {
 // concurrency and queue size. Currently, they can be used from the lifo and lifoGroup
 // filters in the filters/scheduler package only.
 type Queue struct {
-	queue                      *jobqueue.Stack
-	config                     Config
-	metrics                    metrics.Metrics
-	activeRequestsMetricsKey   string
-	queuedRequestsMetricsKey   string
-	rejectedRequestsMetricsKey string
-	timedOutRequestsMetricsKey string
+	queue                    *jobqueue.Stack
+	config                   Config
+	metrics                  metrics.Metrics
+	activeRequestsMetricsKey string
+	errorFullMetricsKey      string
+	errorOtherMetricsKey     string
+	errorTimeoutMetricsKey   string
+	queuedRequestsMetricsKey string
 }
 
 // Options provides options for the registry.
@@ -135,9 +136,11 @@ func (q *Queue) Wait() (done func(), err error) {
 	if q.metrics != nil && err != nil {
 		switch err {
 		case jobqueue.ErrStackFull:
-			q.metrics.IncCounter(q.rejectedRequestsMetricsKey)
+			q.metrics.IncCounter(q.errorFullMetricsKey)
 		case jobqueue.ErrTimeout:
-			q.metrics.IncCounter(q.timedOutRequestsMetricsKey)
+			q.metrics.IncCounter(q.errorTimeoutMetricsKey)
+		default:
+			q.metrics.IncCounter(q.errorOtherMetricsKey)
 		}
 	}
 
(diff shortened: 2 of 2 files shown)


filters/scheduler/lifo.go
```diff
@@ -286,24 +287,25 @@ func (l *lifoGroupFilter) Response(ctx filters.FilterContext) {
...
 		return
 	}
 
 	done, err := q.Wait()
 	if err != nil {
-		// TODO: replace the log with metrics
 		switch err {
 		case jobqueue.ErrStackFull:
 			log.Debugf("Failed to get an entry on to the queue to process QueueFull: %v for host %s", err, ctx.Request().Host)
+			ctx.Metrics().IncCounter(fmt.Sprintf("lifo.%s.rejected", metricKey))
```
wndhydrnt, line 301: This implementation is basic and does not generate a huge diff. It is also possible to record these metrics in the `scheduler` package.
Benefits of doing that include the ability to create the metric keys once and binding the recording of the metrics to the option `EnableRouteLIFOMetrics`. The change would be large though.
Let me know if that is preferable.

AlexanderYastrebov: Hello, thank you for the contribution.
Indeed it looks like lifo queue already contains most of the infrastructure to collect metrics
https://github.com/zalando/skipper/blob/2059778e1924aaafb86accfc0b27e10245f13117/scheduler/scheduler.go#L190-L198
so it should be possible to collect these inside the `Wait`
https://github.com/zalando/skipper/blob/2059778e1924aaafb86accfc0b27e10245f13117/scheduler/scheduler.go#L126-L132
IMO that would be preferable because it would reside in one place, configurable by the option and metric key (queue name that is route id) would be uniform across metrics

wndhydrnt: Thanks for the feedback @AlexanderYastrebov !
I've pushed the changes.

scheduler/scheduler.go
```diff
@@ -194,6 +207,9 @@ func (r *Registry) newQueue(name string, c Config) *Queue {
 
 		q.activeRequestsMetricsKey = fmt.Sprintf("lifo.%s.active", name)
 		q.queuedRequestsMetricsKey = fmt.Sprintf("lifo.%s.queued", name)
+		q.rejectedRequestsMetricsKey = fmt.Sprintf("lifo.%s.rejected", name)
+		q.timedOutRequestsMetricsKey = fmt.Sprintf("lifo.%s.timed-out", name)
```
AlexanderYastrebov, line 211: Lets not use dash in the metric name here so it is single word and different metrics backed accidentally do not replace it with e.g. `_`

wndhydrnt: Removed the `-`.

scheduler/scheduler.go
```diff
@@ -128,7 +131,17 @@ type GroupedLIFOFilter interface {
...
-	return q.queue.Wait()
+	done, err = q.queue.Wait()
+	if q.metrics != nil && err != nil {
+		switch err {
+		case jobqueue.ErrStackFull:
+			q.metrics.IncCounter(q.rejectedRequestsMetricsKey)
+		case jobqueue.ErrTimeout:
+			q.metrics.IncCounter(q.timedOutRequestsMetricsKey)
+		}
+	}
```
AlexanderYastrebov, line 142: I think maybe we should name metrics after error types here like
- `lifo.%s.error.full`
- `lifo.%s.error.timeout`
- `lifo.%s.error.other` (missing default case)

Lets ask @aryszka and @szuecs 

As a note for prometheus metrics we could use single metric name (e.g. `lifo_errors`) and put error type (as well as queue name) as a label but I guess we are stuck with existing metrics interface that only supports metrics key for now.

wndhydrnt: Hey @AlexanderYastrebov 
I am fine with the schema for the keys, thanks for the suggestion.
However, I feel like changing how the metrics interface works is out of scope as my only goal is to record these metrics.

Should I wait for more feedback or implement the changes?

AlexanderYastrebov: @wndhydrnt Metrics interface refactoring is out of scope of course, its just a note.

The proposal does not measure other kind of errors besides full queue and timeout. Also my idea was to highlight in the key that these are error metrics compared to regular metrics like
https://github.com/zalando/skipper/blob/2059778e1924aaafb86accfc0b27e10245f13117/scheduler/scheduler.go#L195-L196

> Should I wait for more feedback or implement the changes?

You may do the changes if you agree. We would need another maintainer to have a look anyway.

wndhydrnt: Hello @AlexanderYastrebov 
I finally got around to implementing the changes. Besides your suggestions, the names of the new `*MetricsKey` variables have been changed because they got awfully long.
Please take a look.

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #1835, opened 2021-08-24
Updates go version in workflows

AlexanderYastrebov:
Also adds a link to semver documentation to clarify that version is not a prefix regexp.

The value `^1.15` is a so called ["caret" range](https://www.npmjs.com/package/semver#caret-ranges-123-025-004)
that matches 1.15 and any minor version above, e.g. 1.16 which could be
seen in the action logs:
```
Setup go stable version spec ^1.15
Found in cache @ /opt/hostedtoolcache/go/1.16.7/x64
Added go to the path
Successfully setup go version ^1.15
go version go1.16.7 linux/amd64
```

See also https://github.com/actions/setup-go/issues/101#issuecomment-814127065

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -7,18 +7,19 @@ jobs:
   tests:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-go@v2
-      with:
-        go-version: '^1.15'
-    - run: go version
-    - run: sudo apt-get install redis-server
-    - run: make deps
-    - run: make check-fmt
-    - run: make build
-    - run: make vet
-    - run: make staticcheck
-    - run: make check-race
-    - run: make cicheck
-    - run: make gosec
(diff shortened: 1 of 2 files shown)


.github/workflows/master.yaml
```diff
@@ -7,18 +7,19 @@ jobs:
...
-    - run: make staticcheck
-    - run: make check-race
-    - run: make cicheck
-    - run: make gosec
-    - run: make publish-coverage
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          # https://www.npmjs.com/package/semver#caret-ranges-123-025-004
+          go-version: '^1.15'
```
mikkeloscar, line 14: Since 1.15 is no longer a supported go version it may make sense to also update it here, to further avoid confusion? Now that you change this anyway :)

AlexanderYastrebov: Done, thanks

AlexanderYastrebov:
:+1:

PR #1839, opened 2021-08-26
Add port number as optional parameter when generating ingress Host predicate

### LINKED ISSUE #1281: Skipper doesn't strip port portion from host header when matching routes
**Describe the bug**
Skipper doesn't strip port portion from host header when matching routes

**To Reproduce**
Add context, for example:

```yaml
kind: Ingress
metadata:
  annotations:
  name: example
  namespace: default
spec:
  rules:
  - host: app.example.tld
    http:
      paths:
      - backend:
          serviceName: example
          servicePort: 80
```

2. http request with headers (truncate your secrets)

Works

```
curl -s https://app.example.tld:443
```

Doesn't work (I get 404)

```
curl -s -H "Host: app.example.tld:443" https://app.example.tld:443
```

**Expected behavior**

The route should match regardless whether the port specified in `Host` header, it is fully RFC valid (https://tools.ietf.org/html/rfc2616#section-14.23)

**Observed behavior**

When `Host` header contains a port portion the route no longer matched

girishc13:
As defined by the [RFC](https://tools.ietf.org/html/rfc2616#section-14.23)
Fixes #1281

### DIFF
--- a/dataclients/kubernetes/eastwest.go
+++ b/dataclients/kubernetes/eastwest.go
@@ -11,12 +11,12 @@ func eastWestRouteID(rid string) string {
 	return "kubeew" + rid[len(ingressRouteIDPrefix):]
 }
 
-func createEastWestRouteIng(eastWestDomainRegexpPostfix, name, ns string, r *eskip.Route) *eskip.Route {
+func createEastWestRouteIng(eastWestDomain, name, ns string, r *eskip.Route) *eskip.Route {
 	if strings.HasPrefix(r.Id, "kubeew") || ns == "" || name == "" {
 		return nil
 	}
 	ewR := *r
-	ewR.HostRegexps = []string{"^" + name + "[.]" + ns + eastWestDomainRegexpPostfix + "(:[0-9]+)?$"}
+	ewR.HostRegexps = []string{generateHostRegexp(name + "." + ns + "." + eastWestDomain)}
 	ewR.Id = eastWestRouteID(r.Id)
 	return &ewR
 }
--- a/dataclients/kubernetes/eastwest.go
+++ b/dataclients/kubernetes/eastwest.go
@@ -16,7 +16,7 @@ func createEastWestRouteIng(eastWestDomainRegexpPostfix, name, ns string, r *esk
 		return nil
 	}
 	ewR := *r
-	ewR.HostRegexps = []string{"^" + name + "[.]" + ns + eastWestDomainRegexpPostfix + "$"}
+	ewR.HostRegexps = []string{"^" + name + "[.]" + ns + eastWestDomainRegexpPostfix + "(:[0-9]+)?$"}
 	ewR.Id = eastWestRouteID(r.Id)
 	return &ewR
 }
--- a/dataclients/kubernetes/eastwest.go
+++ b/dataclients/kubernetes/eastwest.go
@@ -16,7 +16,7 @@ func createEastWestRouteIng(eastWestDomain, name, ns string, r *eskip.Route) *es
 		return nil
 	}
 	ewR := *r
-	ewR.HostRegexps = []string{generateHostRegexp(name + "." + ns + "." + eastWestDomain)}
+	ewR.HostRegexps = []string{createHostRx(name + "." + ns + "." + eastWestDomain)}
 	ewR.Id = eastWestRouteID(r.Id)
 	return &ewR
 }
--- a/dataclients/kubernetes/hosts.go
+++ b/dataclients/kubernetes/hosts.go
@@ -1,7 +1,6 @@
 package kubernetes
 
 import (
-	"fmt"
 	"net/url"
 	"regexp"
 	"strings"
@@ -9,25 +8,17 @@ import (
 	"github.com/zalando/skipper/eskip"
 )
 
-func rxDots(h string) string {
-	return strings.Replace(h, ".", "[.]", -1)
-}
-
-func appendOptionalPortRegex(host string) string {
-	return fmt.Sprintf("%s(:[0-9]+)?", host)
-}
-
-func createHostRx(h ...string) string {
-	if len(h) == 0 {
+func createHostRx(hosts ...string) string {
+	if len(hosts) == 0 {
 		return ""
 	}
 
-	hrx := make([]string, len(h))
-	for i := range h {
-		hrx[i] = appendOptionalPortRegex(rxDots(h[i]))
+	hrx := make([]string, len(hosts))
+	for i, host := range hosts {
+		hrx[i] = strings.Replace(host, ".", "[.]", -1) + "(:[0-9]+)?"
 	}
 
-	return fmt.Sprintf("^(%s)$", strings.Join(hrx, "|"))
+	return "^(" + strings.Join(hrx, "|") + ")$"
 }
 
 // hostCatchAllRoutes creates catch-all routes for those hosts that only have routes with
--- a/dataclients/kubernetes/hosts.go
+++ b/dataclients/kubernetes/hosts.go
@@ -13,6 +13,10 @@ func rxDots(h string) string {
 	return strings.Replace(h, ".", "[.]", -1)
 }
 
+func appendOptionalPortRegex(host string) string {
+	return fmt.Sprintf("%s(:[0-9]+)?", host)
+}
+
 func createHostRx(h ...string) string {
 	if len(h) == 0 {
 		return ""
@@ -20,7 +24,7 @@ func createHostRx(h ...string) string {
 
 	hrx := make([]string, len(h))
 	for i := range h {
-		hrx[i] = rxDots(h[i])
+		hrx[i] = appendOptionalPortRegex(rxDots(h[i]))
 	}
 
 	return fmt.Sprintf("^(%s)$", strings.Join(hrx, "|"))
--- a/dataclients/kubernetes/ingress.go
+++ b/dataclients/kubernetes/ingress.go
@@ -365,9 +365,6 @@ func (ing *ingress) addEndpointsRule(ic ingressContext, host string, prule *defi
 }
 
 func addExtraRoutes(ic ingressContext, ruleHost, path, eastWestDomain string, enableEastWest bool) {
-	// it is a regexp, would be better to have exact host, needs to be added in skipper
-	// this wrapping is temporary and escaping is not the right thing to do
-	// currently handled as mandatory
 	hosts := []string{createHostRx(ruleHost)}
 	// add extra routes from optional annotation
 	for extraIndex, r := range ic.extraRoutes {
--- a/dataclients/kubernetes/ingress.go
+++ b/dataclients/kubernetes/ingress.go
@@ -177,7 +177,7 @@ func convertPathRule(
 
 	var hostRegexp []string
 	if host != "" {
-		hostRegexp = []string{generateHostRegexp(host)}
+		hostRegexp = []string{createHostRx(host)}
 	}
 	svcPort := prule.Backend.ServicePort
 	svcName := prule.Backend.ServiceName
@@ -243,10 +243,6 @@ func convertPathRule(
 	setPath(pathMode, r, prule.Path)
 	setTraffic(r, svcName, prule.Backend.Traffic, prule.Backend.NoopCount)
 	return r, nil
-}
-
-func generateHostRegexp(host string) string {
-	return "^" + strings.Replace(host, ".", "[.]", -1) + "(:[0-9]+)?$"
 }
 
 func setTraffic(r *eskip.Route, svcName string, weight float64, noopCount int) {
@@ -489,7 +485,7 @@ func (ing *ingress) addSpecRule(ic ingressContext, ru *definitions.Rule) error {
 	// it is a regexp, would be better to have exact host, needs to be added in skipper
 	// this wrapping is temporary and escaping is not the right thing to do
 	// currently handled as mandatory
-	host := []string{generateHostRegexp(ru.Host)}
+	host := []string{createHostRx(ru.Host)}
 	// update Traffic field for each backend
 	computeBackendWeights(ic.backendWeights, ru)
 	for _, prule := range ru.Http.Paths {
--- a/dataclients/kubernetes/ingress.go
+++ b/dataclients/kubernetes/ingress.go
@@ -43,14 +43,14 @@ type ingressContext struct {
 }
 
 type ingress struct {
-	provideHTTPSRedirect        bool
-	httpsRedirectCode           int
-	pathMode                    PathMode
-	kubernetesEnableEastWest    bool
-	eastWestDomainRegexpPostfix string
-	eastWestRangeDomains        []string
-	eastWestRangePredicates     []*eskip.Predicate
-	allowedExternalNames        []*regexp.Regexp
+	provideHTTPSRedirect     bool
+	httpsRedirectCode        int
+	pathMode                 PathMode
+	kubernetesEnableEastWest bool
+	kubernetesEastWestDomain string
+	eastWestRangeDomains     []string
+	eastWestRangePredicates  []*eskip.Predicate
+	allowedExternalNames     []*regexp.Regexp
 }
 
 var nonWord = regexp.MustCompile(`\W`)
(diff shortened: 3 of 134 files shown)


dataclients/kubernetes/eastwest.go
```diff
@@ -16,7 +16,7 @@ func createEastWestRouteIng(eastWestDomainRegexpPostfix, name, ns string, r *esk
 		return nil
 	}
 	ewR := *r
-	ewR.HostRegexps = []string{"^" + name + "[.]" + ns + eastWestDomainRegexpPostfix + "$"}
+	ewR.HostRegexps = []string{"^" + name + "[.]" + ns + eastWestDomainRegexpPostfix + "(:[0-9]+)?$"}
```
szuecs, line 19: As far as I can see, there is no test covering this case.

dataclients/kubernetes/hosts.go
```diff
@@ -13,14 +13,18 @@ func rxDots(h string) string {
...
+}
+
 func createHostRx(h ...string) string {
 	if len(h) == 0 {
 		return ""
 	}
 
 	hrx := make([]string, len(h))
 	for i := range h {
-		hrx[i] = rxDots(h[i])
```
szuecs, line 23: rxDots() is not used anymore, please also cleanup this function definition in the PR, thanks

szuecs:
Thanks @girishc13 for the PR, in general looks good to me, some small things to change

dataclients/kubernetes/eastwest.go
girishc13: I added some basic unit tests for this method but I also found some tests in https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/kube_test.go#L3034 and the tests in the kubernetes testdata files.

dataclients/kubernetes/hosts.go
girishc13: The function is still used but it's wrapped as the argument for the new function.

szuecs: Oh my eyes :D

szuecs:
:+1:

dataclients/kubernetes/hosts.go
AlexanderYastrebov: We could move `generateHostRegexp` here from ingress.go and reuse

dataclients/kubernetes/eastwest.go
AlexanderYastrebov: Ideally we could pass `o.KubernetesEastWestDomain` instead of `eastWestDomainRegexpPostfix` and reuse `generateHostRegexp`:
```go
ewR.HostRegexps = []string{generateHostRegexp(name + "." + ns + "." + eastWestDomain)}
```
such that every place reuses the same utility function. It would be easier then to navigate and extend in the future.

Also here 
https://github.com/zalando/skipper/blob/3da61f0422ff1233f840514aec709072e396419f/dataclients/kubernetes/ingress.go#L394
the postfix is not adjusted with port capture.

girishc13: I tried to do this refactor but some tests related to the routeId generation are failing. I will need to go deeper.

dataclients/kubernetes/ingress.go
```diff
@@ -366,14 +361,14 @@ func (ing *ingress) addEndpointsRule(ic ingressContext, host string, prule *defi
 	}
 
 	if ing.kubernetesEnableEastWest {
-		ewRoute := createEastWestRouteIng(ing.eastWestDomainRegexpPostfix, meta.Name, meta.Namespace, endpointsRoute)
-		ewHost := fmt.Sprintf("%s.%s.%s", meta.Name, meta.Namespace, ing.eastWestDomainRegexpPostfix)
+		ewRoute := createEastWestRouteIng(ing.kubernetesEastWestDomain, meta.Name, meta.Namespace, endpointsRoute)
+		ewHost := fmt.Sprintf("%s.%s.%s", meta.Name, meta.Namespace, ing.kubernetesEastWestDomain)
```
AlexanderYastrebov, line 365: Lets check if this should be a regexp in fact

dataclients/kubernetes/ingress.go
```diff
@@ -394,8 +389,8 @@ func addExtraRoutes(ic ingressContext, hosts []string, host, path, eastWestDomai
 			log.Errorf("Failed to add route having %d path routes: %v", n, r)
 		}
 		if enableEastWest {
-			ewRoute := createEastWestRouteIng(eastWestDomainRegexpPostfix, name, ns, &route)
-			ewHost := fmt.Sprintf("%s.%s.%s", name, ns, eastWestDomainRegexpPostfix)
+			ewRoute := createEastWestRouteIng(eastWestDomain, name, ns, &route)
+			ewHost := fmt.Sprintf("%s.%s.%s", name, ns, eastWestDomain)
```
AlexanderYastrebov, line 393: Regexp?

dataclients/kubernetes/ingress.go
girishc13: AFAIK the `ewHost` is mostly used as the key for storing the host -> eskip.Route map. and the `route.HostRegexps` is used for the generating the Host predicate.

https://github.com/zalando/skipper/blob/5676cb745da768f18b685ce7e8e3f21c86f409e7/dataclients/kubernetes/ingress.go#L61

szuecs: ewHost is eastwest host (.skipper.cluster.local)

dataclients/kubernetes/hosts.go
AlexanderYastrebov: @girishc13 Lets reuse `generateHostRegexp` here so that we use a single way to create hostname regexp

dataclients/kubernetes/ingress.go
szuecs: @AlexanderYastrebov the tests are with files in dataclients/kubernetes/testdata/ if you have ideas what to test you can try to show that we have a bug right now.

AlexanderYastrebov: @szuecs I think this has to do with the deprecated `enable-kubernetes-east-west`
https://github.com/zalando/skipper/blob/5676cb745da768f18b685ce7e8e3f21c86f409e7/config/config.go#L376-L379

Here is the branch with two commits https://github.com/zalando/skipper/compare/east-west-and-range-compat
- first commit enables both deprecated and range options which generates duplicate routes
- second commit uses eastWestDomian instead of regexp suffix which applies east-west-range-predicates to the route generated by deprecated option

In general it looks like deprecated options and range options should not be used together.

For the subject discussion the `hostRoutes` map has a hostname as a key and therefore `ing.kubernetesEastWestDomain`, not regexp suffix should be used as proposed.

szuecs: @AlexanderYastrebov I am not if I understand now what you suggest to do. Can you clarify this with @girishc13 please?

AlexanderYastrebov: > what you suggest to do

Sorry, nothing, it should be host and it is host. Before this change it was regexp suffix which was wrong.

dataclients/kubernetes/hosts.go
szuecs: @girishc13 this seems to last and missing piece to do before we can merge.

girishc13: I took a quick look , there is one issue that's preventing me from directly using the `generateHostRegexp` method.
This method is using this statement for format the hosts:
```
return fmt.Sprintf("^(%s)$", strings.Join(hrx, "|"))
```
The `generateHostsRegexp` is formatting the host as
```
return "^" + strings.Replace(host, ".", "[.]", -1) + "(:[0-9]+)?$"
```

There is a mismatch in the capture group. I will take a deeper look but please feel free to provide suggestions.

girishc13: One idea is to extend the `generateHostRegexp` to accept two more arguments for begin expression and end expression like
```
generateHostRegexp(beginExp, host, endExp) {
   return beginExp + strings.Replace(host, ".", "[.]", -1) + "(:[0-9]+)?" + endExp
}
```

AlexanderYastrebov: There is already `createHostRx` that should do what we need, something like this maybe:
```diff
diff --git a/dataclients/kubernetes/eastwest.go b/dataclients/kubernetes/eastwest.go
index d20f814d..cecb0a37 100644
--- a/dataclients/kubernetes/eastwest.go
+++ b/dataclients/kubernetes/eastwest.go
@@ -16,7 +16,7 @@ func createEastWestRouteIng(eastWestDomain, name, ns string, r *eskip.Route) *es
 		return nil
 	}
 	ewR := *r
-	ewR.HostRegexps = []string{generateHostRegexp(name + "." + ns + "." + eastWestDomain)}
+	ewR.HostRegexps = []string{createHostRx(name + "." + ns + "." + eastWestDomain)}
 	ewR.Id = eastWestRouteID(r.Id)
 	return &ewR
 }
diff --git a/dataclients/kubernetes/hosts.go b/dataclients/kubernetes/hosts.go
index 147aa8c2..9a24e884 100644
--- a/dataclients/kubernetes/hosts.go
+++ b/dataclients/kubernetes/hosts.go
@@ -1,7 +1,6 @@
 package kubernetes
 
 import (
-	"fmt"
 	"net/url"
 	"regexp"
 	"strings"
@@ -9,25 +8,19 @@ import (
 	"github.com/zalando/skipper/eskip"
 )
 
-func rxDots(h string) string {
-	return strings.Replace(h, ".", "[.]", -1)
-}
-
-func appendOptionalPortRegex(host string) string {
-	return fmt.Sprintf("%s(:[0-9]+)?", host)
-}
-
-func createHostRx(h ...string) string {
-	if len(h) == 0 {
+func createHostRx(hosts ...string) string {
+	if len(hosts) == 0 {
 		return ""
 	}
-
-	hrx := make([]string, len(h))
-	for i := range h {
-		hrx[i] = appendOptionalPortRegex(rxDots(h[i]))
+	hrx := make([]string, len(hosts))
+	for i, host := range hosts {
+		hrx[i] = strings.Replace(host, ".", "[.]", -1) + "(:[0-9]+)?"
 	}
-
-	return fmt.Sprintf("^(%s)$", strings.Join(hrx, "|"))
+	// single host regexp could be made a bit cleaner
+	// 	if len(hrx) == 1 {
+	// 		return "^" + hrx[0] + "$"
+	// 	}
+	return "^(" + strings.Join(hrx, "|") + ")$"
 }
 
 // hostCatchAllRoutes creates catch-all routes for those hosts that only have routes with
diff --git a/dataclients/kubernetes/ingress.go b/dataclients/kubernetes/ingress.go
index ce8af574..7953671a 100644
--- a/dataclients/kubernetes/ingress.go
+++ b/dataclients/kubernetes/ingress.go
@@ -177,7 +177,7 @@ func convertPathRule(
 
 	var hostRegexp []string
 	if host != "" {
-		hostRegexp = []string{generateHostRegexp(host)}
+		hostRegexp = []string{createHostRx(host)}
 	}
 	svcPort := prule.Backend.ServicePort
 	svcName := prule.Backend.ServiceName
@@ -245,10 +245,6 @@ func convertPathRule(
 	return r, nil
 }
 
-func generateHostRegexp(host string) string {
-	return "^" + strings.Replace(host, ".", "[.]", -1) + "(:[0-9]+)?$"
-}
-
 func setTraffic(r *eskip.Route, svcName string, weight float64, noopCount int) {
 	// add traffic predicate if traffic weight is between 0.0 and 1.0
 	if 0.0 < weight && weight < 1.0 {
@@ -489,7 +485,7 @@ func (ing *ingress) addSpecRule(ic ingressContext, ru *definitions.Rule) error {
 	// it is a regexp, would be better to have exact host, needs to be added in skipper
 	// this wrapping is temporary and escaping is not the right thing to do
 	// currently handled as mandatory
-	host := []string{generateHostRegexp(ru.Host)}
+	host := []string{createHostRx(ru.Host)}
 	// update Traffic field for each backend
 	computeBackendWeights(ic.backendWeights, ru)
 	for _, prule := range ru.Http.Paths {
```

dataclients/kubernetes/ingress.go
```diff
@@ -718,9 +713,10 @@ func (ing *ingress) addCatchAllRoutes(host string, r *eskip.Route, redirect *red
 		HostRegexps: r.HostRegexps,
 		BackendType: eskip.ShuntBackend,
 	}
+	log.Infof("host: %s", host)
```
AlexanderYastrebov, line 716: Please remove

dataclients/kubernetes/ingress.go
```diff
@@ -490,11 +485,11 @@ func (ing *ingress) addSpecRule(ic ingressContext, ru *definitions.Rule) error {
 	// it is a regexp, would be better to have exact host, needs to be added in skipper
 	// this wrapping is temporary and escaping is not the right thing to do
 	// currently handled as mandatory
-	host := []string{"^" + strings.Replace(ru.Host, ".", "[.]", -1) + "$"}
+	host := []string{createHostRx(ru.Host)}
 	// update Traffic field for each backend
 	computeBackendWeights(ic.backendWeights, ru)
 	for _, prule := range ru.Http.Paths {
-		addExtraRoutes(ic, host, ru.Host, prule.Path, ing.eastWestDomainRegexpPostfix, ing.kubernetesEnableEastWest)
+		addExtraRoutes(ic, host, ru.Host, prule.Path, ing.kubernetesEastWestDomain, ing.kubernetesEnableEastWest)
```
AlexanderYastrebov, line 492: Since we change here, lets remove
https://github.com/zalando/skipper/blob/c518cb137eaaa8f4453f9904d13f6d5352527363/dataclients/kubernetes/ingress.go#L490-L493
pass only `ru.Host` and create regexp inside of `addExtraRoutes`
https://github.com/zalando/skipper/blob/c518cb137eaaa8f4453f9904d13f6d5352527363/dataclients/kubernetes/ingress.go#L378

dataclients/kubernetes/ingress.go
```diff
@@ -364,7 +364,11 @@ func (ing *ingress) addEndpointsRule(ic ingressContext, host string, prule *defi
 	return nil
 }
 
-func addExtraRoutes(ic ingressContext, hosts []string, host, path, eastWestDomain string, enableEastWest bool) {
+func addExtraRoutes(ic ingressContext, ruleHost, path, eastWestDomain string, enableEastWest bool) {
+	// it is a regexp, would be better to have exact host, needs to be added in skipper
+	// this wrapping is temporary and escaping is not the right thing to do
+	// currently handled as mandatory
```
AlexanderYastrebov, line 370: I think this comment is redundant now as it is clear that regexp is expected
```go
route.HostRegexps = hosts
```

AlexanderYastrebov:
Please fix DCO, otherwise looks good now.

AlexanderYastrebov:
:+1:

girishc13:
I had forgotten to sign my last commit.

AlexanderYastrebov:
:+1:

dataclients/kubernetes/testdata/ingress/eastwestrange/ing-with-1externalhost-1internalhost-1prule-1customroute-2ranges-1match.eskip
```diff
@@ -1,17 +1,17 @@
 kube_foo__qux__0__example_ingress_cluster_local_____:
-	Host("^example[.]ingress[.]cluster[.]local$") && PathRegexp("^/")
+	Host("^(example[.]ingress[.]cluster[.]local(:[0-9]+)?)$") && PathRegexp("^/")
```
szuecs, line 2: Why do we need nested capture groups here?
I think it should be `^example[.]ingress[.]cluster[.]local(:[0-9]+)?$"`, no?

szuecs:
Hm I think the regexp is now broken please see https://github.com/zalando/skipper/pull/1839/files#r712430340

dataclients/kubernetes/testdata/ingress/eastwestrange/ing-with-1externalhost-1internalhost-1prule-1customroute-2ranges-1match.eskip
AlexanderYastrebov: The outer brackets are added by the regexp generator function that supports a list of multiple hosts:
https://github.com/zalando/skipper/blob/ec6e5c7246d4f33b67a051e0fb80342af2061ae1/dataclients/kubernetes/hosts.go#L11-L22
It is possible to omit the outer brackets by checking the list length: 
```diff
+	 	if len(hrx) == 1 {
+	 		return "^" + hrx[0] + "$"
+	 	}
```

The values of capture groups are not used as far as I can tell so it does not really matter I think (we could also consider using non capturing groups for purity but that would look even more clumsy like `^(?:example[.]ingress[.]cluster[.]local(?::[0-9]+)?)$`). See also [subject regexp in the playground](https://regex101.com/r/Kljl0Y/1).

AlexanderYastrebov: The outer brackets are needed, without them in case of multiple hosts [a prefixed value will match](https://regex101.com/r/sa90ZK/1):
`^example[.]org(:[0-9]+)?|example[.]org(:[0-9]+)?$` matches `www.example.org`

AlexanderYastrebov: Without outer brackets [a prefixed value matches](https://regex101.com/r/sa90ZK/3):
`^example[.]org(:[0-9]+)?|example[.]com(:[0-9]+)?$` matches `www.example.com`.

Alternatively every host could be wrapped with `^...$` like [`^example[.]org(:[0-9]+)?$|^example[.]com(:[0-9]+)?$`](https://regex101.com/r/kUVIxl/1)

szuecs:
:+1:

Issue #1842, opened 2021-08-31
Enable remote eskip file functionality via a flag


marcinzaremba:
**Is your feature request related to a problem? Please describe.**
Currently Skipper has already [implemented](https://github.com/zalando/skipper/pull/1735) an ability to fetch eskip routes from a remote endpoint, however it is not possible to use this functionality when running Skipper as a standalone binary, namely there is no configuration flag that would enable it.

**Describe the solution you would like**
Add a flag / adopt existing flag to be able to use it

**Describe alternatives you've considered (optional)**
-

**Additional context (optional)**
-

**Would you like to work on it?**
Yes

xavivars:
It's great there's a flag for this now 😊

PR #1843, opened 2021-08-31
Initial implementation of the remote eskip file flag

marcinzaremba:
Issue: https://github.com/zalando/skipper/issues/1842


config/config.go
```diff
@@ -118,6 +118,8 @@ type Config struct {
 	InnkeeperPreRouteFilters  string               `yaml:"innkeeper-pre-route-filters"`
 	InnkeeperPostRouteFilters string               `yaml:"innkeeper-post-route-filters"`
 	RoutesFile                string               `yaml:"routes-file"`
+	RemoteRoutesFile          string               `yaml:"remote-routes-file"`
```
AlexanderYastrebov, line 121: `routes-url` maybe?

skipper.go
```diff
@@ -802,6 +811,20 @@ func createDataClients(o Options, auth innkeeper.Authentication) ([]routing.Data
 		}
 	}
 
+	if o.RemoteRoutesFile != "" {
+		for _, rf := range strings.Split(o.RemoteRoutesFile, ",") {
```
AlexanderYastrebov, line 815: Lets check if we can use `commaListFlag` instead

skipper.go
```diff
@@ -802,6 +811,20 @@ func createDataClients(o Options, auth innkeeper.Authentication) ([]routing.Data
...
+			client, err := eskipfile.RemoteWatch(&eskipfile.RemoteWatchOptions{
+				RemoteFile:    rf,
+				FailOnStartup: o.RemoteRoutesFailOnStartup,
```
AlexanderYastrebov, line 818: Do we need this configurable? Maybe just always fail like for routes from file?

config/config.go
AlexanderYastrebov: The point is that it is not necessary the file and we do not have to follow the internal client naming convention, the command line arguments are kind of user interface

skipper.go
marcinzaremba: I think this configurable makes sense from reliability perspective. I would say failing to load on startup does not really always imply a long term failure of the remote service (which can succeed just on a subsequent try) and it might be multiple opinions how to behave on such an occasion.

config/config.go
marcinzaremba: Fine by me.

skipper.go
```diff
@@ -802,6 +811,20 @@ func createDataClients(o Options, auth innkeeper.Authentication) ([]routing.Data
...
+			})
+			if err != nil {
+				log.Error("error while loading remote eskip file", err)
```
AlexanderYastrebov, line 821: lets include failed url - in case skipper is started with multiple arguments it would be easier to debug which url was not available

AlexanderYastrebov: Though maybe `err` would include it, still lets have a look from user perspective debugging the startup failure

skipper.go
marcinzaremba: Sure. I missed that piece. Thanks.

config/config.go
```diff
@@ -118,6 +118,8 @@ type Config struct {
 	InnkeeperPreRouteFilters  string               `yaml:"innkeeper-pre-route-filters"`
 	InnkeeperPostRouteFilters string               `yaml:"innkeeper-post-route-filters"`
 	RoutesFile                string               `yaml:"routes-file"`
+	RemoteRoutesURLs          *listFlag            `yaml:"remote-routes-urls"`
```
AlexanderYastrebov, line 121: nitpick: url can be local, do we need the `remote-` prefix?

skipper.go
```diff
@@ -802,6 +811,20 @@ func createDataClients(o Options, auth innkeeper.Authentication) ([]routing.Data
...
 	}
 
+	if len(o.RemoteRoutesURLs) > 0 {
+		for _, url := range o.RemoteRoutesURLs {
+			client, err := eskipfile.RemoteWatch(&eskipfile.RemoteWatchOptions{
+				RemoteFile:    url,
+				FailOnStartup: o.RemoteRoutesFailOnStartup,
+			})
+			if err != nil {
+				log.Errorf("error while loading remote eskip file from url %s: %s", url, err)
```
AlexanderYastrebov, line 821: lets remove `remote` and `file` from the message, e.g. `error while loading routes from url %s: %s`

config/config.go
marcinzaremba: My thinking was to group related flags with the same prefix (we introduce two) to avoid lonely flags that one can have trouble to connect to the specific feature. In this case `remote-` seems to be handy as that's the name of the feature. Do you have a better naming suggestion considering there are two flags?

config/config.go
```diff
@@ -350,6 +353,8 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.InnkeeperPreRouteFilters, "innkeeper-pre-route-filters", "", "filters to be prepended to each route loaded from Innkeeper")
 	flag.StringVar(&cfg.InnkeeperPostRouteFilters, "innkeeper-post-route-filters", "", "filters to be appended to each route loaded from Innkeeper")
 	flag.StringVar(&cfg.RoutesFile, "routes-file", "", "file containing route definitions")
+	flag.Var(cfg.RemoteRoutesURLs, "remote-routes-urls", "comma separated URLs to route definitions in eskip format")
+	flag.BoolVar(&cfg.RemoteRoutesFailOnStartup, "remote-routes-fail-on-startup", false, "fail fast when an initial attempt of fetching remote routes is unsuccessful")
```
AlexanderYastrebov, line 357: I think it should be `true` by default like for the files (fails if route file is malformed).

marcinzaremba: The thing is the error in question is not about parsing routes in this case, but availability of the remote endpoint.

marcinzaremba:
👍

skipper.go
```diff
@@ -250,6 +250,11 @@ type Options struct {
 	// command this option is used when starting it with the -routes-file flag.)
 	WatchRoutesFile string
 
+	// RouteURLs are URLs pointing to route definitions, in eskip format, with change watching enabled.
+	// Multiple may be given comma separated. (For the skipper
+	// command this option is used when starting it with the -remote-routes-url flag.)
```
AlexanderYastrebov, line 255: The flag name was changed. Maybe we do not need the part in the brackets at all.

AlexanderYastrebov:
:+1:

skipper.go
AlexanderYastrebov: It is also a slice now so `Multiple may be given comma separated.` is not relevant either

marcinzaremba: Good catch.

marcinzaremba: Fixed. Sorry for that. Simply missed that.

AlexanderYastrebov:
:+1:

marcinzaremba:
👍

szuecs:
:+1:

aryszka:
:+1:

PR #1846, opened 2021-09-06
Fix forwarded host when multiple entries

xavivars:
Caught corner case when multiple entries appear in the forwarded host predicate [see docs](https://www.rfc-editor.org/rfc/rfc7239#section-4)

> If the request is passing through several proxies, each proxy can add a set of parameters; it can also remove previously added "Forwarded" header fields.

> A proxy server that wants to add a new "Forwarded" header field value can either append it to the last existing "Forwarded" header field after a comma separator or add a new field at the end of the header block.  A proxy MAY remove all "Forwarded" header fields from a request.  It MUST, however, ensure that the correct header field is updated in case of multiple "Forwarded" header fields.

### DIFF
--- a/predicates/forwarded/forwarded.go
+++ b/predicates/forwarded/forwarded.go
@@ -131,7 +131,7 @@ func parseForwarded(fh string) *forwarded {
 	f := &forwarded{}
 
 	for _, forwardedFull := range strings.Split(fh, ",") {
-		for _, forwardedPair := range strings.Split(forwardedFull, ";") {
+		for _, forwardedPair := range strings.Split(strings.TrimSpace(forwardedFull), ";") {
 			if tv := strings.SplitN(forwardedPair, "=", 2); len(tv) == 2 {
 				token, value := tv[0], tv[1]
 				value = strings.Trim(value, `"`)
--- a/predicates/forwarded/forwarded.go
+++ b/predicates/forwarded/forwarded.go
@@ -130,17 +130,20 @@ func parseForwarded(fh string) *forwarded {
 
 	f := &forwarded{}
 
-	for _, forwardedPair := range strings.Split(fh, ";") {
-		if tv := strings.SplitN(forwardedPair, "=", 2); len(tv) == 2 {
-			token, value := tv[0], tv[1]
-			value = strings.Trim(value, `"`)
-			switch token {
-			case "proto":
-				f.proto = value
-			case "host":
-				f.host = value
+	for _, forwardedFull := range strings.Split(fh, ",") {
+		for _, forwardedPair := range strings.Split(forwardedFull, ";") {
+			if tv := strings.SplitN(forwardedPair, "=", 2); len(tv) == 2 {
+				token, value := tv[0], tv[1]
+				value = strings.Trim(value, `"`)
+				switch token {
+				case "proto":
+					f.proto = value
+				case "host":
+					f.host = value
+				}
 			}
 		}
 	}
+
 	return f
 }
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

szuecs:
I checked RFC and tested with playground: https://play.golang.org/p/Z-fzU2CYiRD

xavivars:
Yeah, I also checked the RFC, and while other "parts" of the header (for, by,...) can have different values when header is a multi-header format (comma separated), it says explicitly that host has to be the original Host header.

predicates/forwarded/forwarded.go
```diff
@@ -130,17 +130,20 @@ func parseForwarded(fh string) *forwarded {
...
+		for _, forwardedPair := range strings.Split(forwardedFull, ";") {
+			if tv := strings.SplitN(forwardedPair, "=", 2); len(tv) == 2 {
+				token, value := tv[0], tv[1]
+				value = strings.Trim(value, `"`)
+				switch token {
+				case "proto":
+					f.proto = value
+				case "host":
+					f.host = value
+				}
```
AlexanderYastrebov, line 143: This will only match the last value from the comma-separated list, i.e. the value added by the last proxy in the chain. Should we update docs https://github.com/zalando/skipper/blob/master/docs/reference/predicates.md#forwarded-header-predicates maybe?

There is a multi-valued example `Forwarded: for=192.0.2.43, for=198.51.100.17` in RFC that contains a space after the comma. I think current implementation does not consider space after the comma (e.g. `host=example.com, host=example.org;proto=https`)

xavivars: Good points:
- Docs have been updated
- Fixed and added a test for the whitespace issue.
- What would you suggest for your latest point? I know that's the case.... but not sure how to behave.... maybe just add that in the docs?

AlexanderYastrebov: I think there is no easy way to handle `host=example.com;proto=https, host=example.org` because we have separate `ForwardedHost` and `ForwardedProtocol` predicates.

I can imagine we could introduce `Forwarded("host=^.example\\.org$", "proto=https")` that would match only if there is a header value that matches _both_ host and proto field.

For the existing predicates and in this scope I think the best we can do is to document that predicates match _the last_  host/proto value in the header. Alternative is to refactor to match _first_ or _any_ value if that is desired.

I also suggest that we document on high-level (instead of technical jargon `predicate config would have f.proto = "https" ...`) by providing a header example and predicate combination like:

Example: `Forwarded: host=example.com;proto=https, host=example.org`
1. `ForwardedHost(/^example\.com$/)` - does not match
2. `ForwardedHost(/^example\.org$/)` - matches
3. `ForwardedHost(/^example\.org$/) && ForwardedProto("https")` - matches
4. `ForwardedHost(/^example\.com$/) && ForwardedProto("https")` - does not match

Maybe it is also worth adding tests for the `ForwardedHost && ForwardedProto` cases.
I would say cases 3 and 4 are surprising from the user perspective but could only be properly handled with a predicate that matches multiple fields simultaneously.

xavivars: Thanks a million for the review/suggestions!!!

 I'll make the suggested improvements and request a re-review later

xavivars: Done!

szuecs:
@xavivars please rebase master, thanks

szuecs:
:+1:

aryszka:
:+1:

PR #1847, opened 2021-09-06
Adds forwarded headers flags

AlexanderYastrebov:
Adds two new flags to control addition of `X-Forwarded-*` headers.
The addition happens before request routing so e.g. `SourceFromLast` predicate can rely on the header value.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -374,7 +374,7 @@ func NewConfig() *Config {
 		"X-Forwarded-For sets or appends with comma the remote IP of the request to the X-Forwarded-For header value\n"+
 		"X-Forwarded-Host sets X-Forwarded-Host value to the request host\n"+
 		"X-Forwarded-Proto=http or X-Forwarded-Proto=https sets X-Forwarded-Proto value")
-	flag.Var(cfg.ForwardedHeadersExcludeCIDRList, "forwarded-headers-exclude-cidrs", "disables addition of forwarded headers for the remote host IPs from the configured CIDRs")
+	flag.Var(cfg.ForwardedHeadersExcludeCIDRList, "forwarded-headers-exclude-cidrs", "disables addition of forwarded headers for the remote host IPs from the comma separated list of CIDRs")
 
 	// Kubernetes:
 	flag.BoolVar(&cfg.KubernetesIngress, "kubernetes", false, "enables skipper to generate routes for ingress resources in kubernetes cluster")
--- a/config/config.go
+++ b/config/config.go
@@ -798,7 +798,11 @@ func (c *Config) ToOptions() skipper.Options {
 
 	if c.ForwardedHeaders != (net.ForwardedHeaders{}) {
 		options.CustomHttpHandlerWrap = func(handler http.Handler) http.Handler {
-			return &net.ForwardedHeadersHandler{c.ForwardedHeaders, c.ForwardedHeadersExcludeCIDRs, handler}
+			return &net.ForwardedHeadersHandler{
+				Headers: c.ForwardedHeaders,
+				Exclude: c.ForwardedHeadersExcludeCIDRs,
+				Handler: handler,
+			}
 		}
 	}
 
--- a/config/config.go
+++ b/config/config.go
@@ -371,9 +371,8 @@ func NewConfig() *Config {
 
 	// Forwarded headers
 	flag.Var(cfg.ForwardedHeadersList, "forwarded-headers", "comma separated list of headers to add to the incoming request before routing\n"+
-		"X-Forwarded-For sets or appends request remote IP to the X-Forwarded-For header\n"+
-		"X-Forwarded-For-prepend sets or prepends request remote IP to the X-Forwarded-For header, overrides X-Forwarded-For\n"+
-		"X-Forwarded-Host sets X-Forwarded-Host to the request host\n"+
+		"X-Forwarded-For sets or appends with comma the remote IP of the request to the X-Forwarded-For header value\n"+
+		"X-Forwarded-Host sets X-Forwarded-Host value to the request host\n"+
 		"X-Forwarded-Proto=http or X-Forwarded-Proto=https sets X-Forwarded-Proto value")
 	flag.Var(cfg.ForwardedHeadersExcludeCIDRList, "forwarded-headers-exclude-cidrs", "disables addition of forwarded headers for the remote host IPs from the configured CIDRs")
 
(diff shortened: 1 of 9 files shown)


config/config.go
```diff
@@ -361,6 +369,14 @@ func NewConfig() *Config {
 	flag.Var(cfg.CloneRoute, "clone-route", "clone all matching routes and replace filters and predicates of all matched routes")
 	flag.BoolVar(&cfg.WaitFirstRouteLoad, "wait-first-route-load", false, "prevent starting the listener before the first batch of routes were loaded")
 
+	// Forwarded headers
+	flag.Var(cfg.ForwardedHeadersList, "forwarded-headers", "comma separated list of headers to add to the incoming request before routing\n"+
+		"X-Forwarded-For sets or appends request remote IP to the X-Forwarded-For header\n"+
+		"X-Forwarded-For-prepend sets or prepends request remote IP to the X-Forwarded-For header, overrides X-Forwarded-For\n"+
```
szuecs, line 375: AFAIK RFC misses the detail about how to add another entry to the list either prepend or append. The majority (Nginx, Apache, Google LB, ..) does prepend and ALB is the only one that does append, so I would rather switch to x-forwarded-for-append and default to prepend.
Do you have other information?
Context: I looked at this, when I did the Source() predicate investigation, that resulted in SourceFromLast() predicate.

config/config.go
```diff
@@ -776,6 +797,12 @@ func (c *Config) ToOptions() skipper.Options {
 		}
 	}
 
+	if c.ForwardedHeaders != (net.ForwardedHeaders{}) {
+		options.CustomHttpHandlerWrap = func(handler http.Handler) http.Handler {
```
szuecs, line 801: I wonder if we should allow a list here in order to allow custom skipper's to add a custom middleware chain here.

config/config.go
```diff
@@ -815,6 +842,33 @@ func (c *Config) parseHistogramBuckets() ([]float64, error) {
...
+		case "X-Forwarded-For":
+			c.ForwardedHeaders.For = true
+		case "X-Forwarded-For-prepend":
+			c.ForwardedHeaders.PrependFor = true
+		case "X-Forwarded-Host":
+			c.ForwardedHeaders.Host = true
+		case "X-Forwarded-Proto=http":
+			c.ForwardedHeaders.Proto = "http"
+		case "X-Forwarded-Proto=https":
+			c.ForwardedHeaders.Proto = "https"
```
szuecs, line 857: Should we also handle https://datatracker.ietf.org/doc/html/rfc7239#section-4 (more short https://en.wikipedia.org/wiki/X-Forwarded-For#Alternatives_and_variations)?
Since X-Forwarded-For was AFAIK never a standard, there is now RFC7239 that should change.
If not we might want to create an issue for this.

szuecs:
I think we also miss documentation and we might want to increase minor in https://github.com/zalando/skipper/blob/master/VERSION

config/config.go
AlexanderYastrebov: The link https://en.wikipedia.org/wiki/X-Forwarded-For#Format says it is appended
> X-Forwarded-For: client, proxy1, proxy2
> where the value is a comma+space separated list of IP addresses, the left-most being the original client, and each successive proxy that passed the request adding the IP address where it received the request from. 

Nginx appends:
https://www.nginx.com/resources/wiki/start/topics/examples/forwarded/
> NGINX even provides a $proxy_add_x_forwarded_for variable to automatically append $remote_addr to any incoming X-Forwarded-For headers.

Apache docs are unclear
https://httpd.apache.org/docs/2.4/mod/mod_proxy.html
but this answer states it appends
https://stackoverflow.com/questions/40411974/what-determines-when-mod-proxy-only-sometimes-add-internal-ip-addresses-to-x-for
and the code appends
https://github.com/apache/httpd/blob/9e02e2000cc05754a9c1c9a9aead7ce2cc8ea6ee/modules/proxy/proxy_util.c#L3992-L3996
https://github.com/apache/apr/blob/6128ac5307a958dccb1af4ebbb5cfa95607a8454/tables/apr_tables.c#L720-L722

HAProxy appends
http://cbonte.github.io/haproxy-dconv/2.5/configuration.html#4-option%20forwardfor
> Since this
header is always appended at the end of the existing header list, the server
must be configured to always use the last occurrence of this header only.

Note also that HAProxy has `[ except <network> ]` option that does the same as the proposed `forwarded-headers-exclude-cidrs`

In fact I can not find any case where XFF would be prepended :man_shrugging:

config/config.go
AlexanderYastrebov: We create options in this method
https://github.com/zalando/skipper/blob/1dc62b86f1d9d406de63f7fea392e6f0b8c24448/config/config.go#L559

It is the potential user that call `ToOptions()` should wrap if they want to keep the wrap intact

config/config.go
AlexanderYastrebov: Yes, I decided to keep `Forwarded` header out of scope to keep PR smaller, it could be added later on - I intentionally did not put `x-` prefix into the flag names such that we could add `Forwarded` or `Via` in the future

AlexanderYastrebov:
> I think we also miss documentation and we might want to increase minor in

I've tried to describe the functionality in the flag help.
The default flag values are noop and do not change the behavior so I do not think it deserves version bump.

config/config.go
szuecs: @AlexanderYastrebov yes they can, but in case `c.ForwardedHeaders != (net.ForwardedHeaders{})`, we overwrite it or our ForwardedHeaders handler does not work as configured.

szuecs:
@AlexanderYastrebov maybe you can add something for the operator user in docs/operation/operation.md, which will be rendered in https://opensource.zalando.com/skipper/operation/operation/

config/config.go
AlexanderYastrebov: We overwrite default value because we create `skipper.Options` anew

config/config.go
AlexanderYastrebov: Google LB appends two values client ip and LB ip as far as I can tell https://cloud.google.com/load-balancing/docs/https#x-forwarded-for_header

szuecs: @AlexanderYastrebov you posted `X-Forwarded-For: client, proxy1, proxy2`, client is first (prepend), if google LB or any other proxy appends itself, it's not append (not prepend either, but what can we do). Of course the XFF header can not be trusted in case the proxy in front does not write the connecting client into a specified field in the header, because it can be faked otherwise by the client connecting to the proxy in front of us. 
Following the current skipper naming I would prefer to have prepend/append in both (following default filters) or do default/append (following Source vs. SourceFromLast predicate).

config/config.go
szuecs: thanks, true

config/config.go
AlexanderYastrebov: > Following the current skipper naming I would prefer to have prepend/append in both (following default filters) or do default/append (following Source vs. SourceFromLast predicate).

I have removed `X-Forwarded-For-prepend` as I have not found any proxy implementation that _prepends_ request IP to the header value.

aryszka:
:+1:

PR #1848, opened 2021-09-07
Routes cache / adapter service 

marcinzaremba:
This PR introduces a new Skipper binary that serves a simple routes proxy/cache/server. It fetches routes from a data source regularly as a regular Skipper data client / routing does, however does as little processing on fetched routes as possible, keeps them always eskip-formatted in memory and thus is optimized for serving them fast. For the time being only Kubernetes data client is supported.
The rationale for the component is to avoid a potential data source overload in case Skipper is scaled horizontally to a significant amount of nodes, which can result in a significant RPS for data sources (where every instance polls regularly, which is usually couple of seconds). In effect a load on querying data sources should be proportional to the number of route server instances (and not to the number of Skipper instances, which can be even hundreds in some cases), which supposedly can scale better for Skipper purposes.


cmd/routesrv/main.go
```diff
@@ -0,0 +1,54 @@
...
+		return err
+	}
+
+	cache := newCache(dataclient, 3*time.Second)
+
+	http.HandleFunc("/routes", func(w http.ResponseWriter, r *http.Request) {
+		w.Write(cache.get())
+	})
+
+	http.ListenAndServe(":8080", nil)
```
AlexanderYastrebov, line 45: Should listen address be configurable?

cmd/routesrv/main.go
```diff
@@ -0,0 +1,54 @@
...
+		AllowedExternalNames:              o.KubernetesAllowedExternalNames,
+	})
+	if err != nil {
+		return err
+	}
+
+	cache := newCache(dataclient, 3*time.Second)
+
+	http.HandleFunc("/routes", func(w http.ResponseWriter, r *http.Request) {
+		w.Write(cache.get())
```
AlexanderYastrebov, line 42: Alternatively cache could implement `http.Handler`

cmd/routesrv/main.go
marcinzaremba: I am about to push changes with this implemented.

cmd/routesrv/main.go
marcinzaremba: I am about to push changes with this implemented.

cmd/routesrv/main.go
```diff
@@ -0,0 +1,262 @@
...
+	buf := &bytes.Buffer{}
+	eskip.Fprint(buf, eskip.PrettyPrintInfo{}, routes...)
+
+	e.mu.Lock()
+	oldData := e.data
+	e.data = buf.Bytes()
+	e.initialized = true
+	e.mu.Unlock()
+
+	return len(oldData), oldData == nil
```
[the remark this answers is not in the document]
AlexanderYastrebov: WDYT if we return (newData, oldData []byte) so that users can take length or detect the initialization by checking `oldData == nil` themselves. This also eliminates `initialized` field.

cmd/routesrv/main.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+func (e *eskipBytes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	span := tracing.CreateSpan("serve_routes", context.TODO(), e.tracer)
```
AlexanderYastrebov, line 59: `r.Context()` maybe?

cmd/routesrv/main.go
```diff
@@ -0,0 +1,262 @@
...
+		initializedTimestamp: promauto.NewGauge(prometheus.GaugeOpts{
+			Namespace: "routesrv",
+			Subsystem: "poller",
+			Name:      "initialized_timestamp",
+		}),
+		updatedTimestamp: promauto.NewGauge(prometheus.GaugeOpts{
+			Namespace: "routesrv",
+			Subsystem: "poller",
+			Name:      "updated_timestamp",
+		}),
```
AlexanderYastrebov, line 108: I am struggling to grasp what do we measure here

cmd/routesrv/main.go
```diff
@@ -0,0 +1,262 @@
...
+			if initialized {
+				log.Info("routes initialized")
+				span.SetTag("initialized", true)
+				p.metrics.initializedTimestamp.SetToCurrentTime()
+			} else {
+				log.Debug("routes updated")
+			}
+			p.metrics.updatedTimestamp.SetToCurrentTime()
+			span.SetTag("routes.received_len", routesLen)
+			span.SetTag("routes.stored_bytes", size)
```
AlexanderYastrebov, line 170: `routes.count` and `routes.bytes`?

cmd/routesrv/main.go
```diff
@@ -0,0 +1,262 @@
...
+// Reports healthy only when the bytes were initialized (set at least once).
+type eskipBytesStatus struct {
+	b *eskipBytes
+}
+
+var msgRoutesNotInitialized = []byte("routes were not initialized yet")
+
+func (s *eskipBytesStatus) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if _, initialized := s.b.bytes(); initialized {
+		w.WriteHeader(http.StatusOK)
```
AlexanderYastrebov, line 79: `http.StatusNoContent`?

cmd/routesrv/main.go
```diff
@@ -0,0 +1,262 @@
...
+
+	server := newServer(o.Address, b, s)
+
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		<-sigs
+		log.Info("shutting down")
+		close(poller.quit)
+		if err := server.Shutdown(context.Background()); err != nil {
```
AlexanderYastrebov, line 245: Do we need shutdown timeout here?

Also from https://pkg.go.dev/net/http#Server.Shutdown

> When Shutdown is called, Serve, ListenAndServe, and ListenAndServeTLS immediately return ErrServerClosed. Make sure the program doesn't exit and waits instead for Shutdown to return.

cmd/routesrv/options/options.go
```diff
@@ -0,0 +1,111 @@
...
+	// KubernetesNamespace is used to switch between monitoring ingresses in the cluster-scope or limit
+	// the ingresses to only those in the specified namespace. Defaults to "" which means monitor ingresses
+	// in the cluster-scope.
+	KubernetesNamespace string
+
+	// *DEPRECATED* KubernetesEnableEastWest enables cluster internal service to service communication, aka east-west traffic
+	KubernetesEnableEastWest bool
+
+	// *DEPRECATED* KubernetesEastWestDomain sets the cluster internal domain used to create additional routes in skipper, defaults to skipper.cluster.local
+	KubernetesEastWestDomain string
```
AlexanderYastrebov, line 71: Lets check if we need to pull this legacy stuff

AlexanderYastrebov:
I think there are no tests at the moment as it is WIP

cmd/routesrv/main.go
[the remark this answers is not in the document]
marcinzaremba: Makes sense.

cmd/routesrv/options/options.go
marcinzaremba: Considering it's a new binary and it was already marked `DEPRECATED` I guess we can drop it.

marcinzaremba: Checking code, if they are really deprecated and the same can be achieved with the suggested replacement (`...EastWestRange...`)

cmd/routesrv/main.go
marcinzaremba: Added prom native help messages to them.

marcinzaremba:
> I think there are no tests at the moment as it is WIP

Indeed. Once we are sure about the shape, some tests would be good.

cmd/routesrv/options/options.go
marcinzaremba: OK. Verified and got rid of 3 deprecated/redundant options.

cmd/routesrv/main.go
szuecs: For teardown, if we do the same as in skipper, it's fine. check https://github.com/zalando/skipper/blob/master/skipper.go#L1068-L1080

Shutdown() is safe to use if you are already faded out of the load balancer, so no new connections should be possible if you call Shutdown().
Because we get SIGTERM immediately we should add a gracePeriod, such that our pod is faded out of the endpoints list, that is part of the service on all nodes (iptables change on all nodes where skipper-ingress is running). So the current one is not safe, we need to do a sleep before Shutdown() is called.

szuecs:
For DCO check pass, please add to all git commit `-s`, thanks. You can do this via rebase on all commits, but maybe you squash into one before and then apply it on the single commit.

cmd/routesrv/main.go
marcinzaremba: Got it. I do not really wait for the shutdown to happen properly and this shutdown (if properly awaited) can take a while.

cmd/routesrv/main.go
```diff
@@ -0,0 +1,279 @@
...
+	e.mu.Unlock()
+
+	return e.data, oldData
+}
+
+func (e *eskipBytes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	span := tracing.CreateSpan("serve_routes", r.Context(), e.tracer)
+	defer span.Finish()
+
+	if data := e.bytes(); data != nil {
```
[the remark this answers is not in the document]
aryszka: i just noticed the earlier suggestion from @AlexanderYastrebov . But in Go, comparing slices to nil is not a good practice. The reason is that a lot of existing functions handle nil and []byte{} the same way, plus, handling nil and empty differently would make it necessary when writing new functions in our code to remember that we need to return empty instead of nil in this program.

Besides that, in this case we also carry here an important meaning: uninitialized. So, it's better to be explicit.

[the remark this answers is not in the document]
AlexanderYastrebov: > i just noticed the earlier suggestion

Well, maybe not the best one :) Lets make it explicit, I agree

cmd/routesrv/main.go
```diff
@@ -0,0 +1,279 @@
...
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGTERM)
+	wg.Add(1)
+	go func() {
+		<-sigs
+		shutdown(wg, poller, server, o.WaitForHealthcheckInterval)
+	}()
+
+	if err = server.ListenAndServe(); err != http.ErrServerClosed {
+		go shutdown(wg, poller, server, 0)
```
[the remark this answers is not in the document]
marcinzaremba: Correct.

logging/loggingtest/logger.go
```diff
@@ -209,3 +211,27 @@ func (tl *Logger) Info(a ...interface{})             { tl.log(a...) }
 func (tl *Logger) Infof(f string, a ...interface{})  { tl.logf(f, a...) }
 func (tl *Logger) Debug(a ...interface{})            { tl.log(a...) }
 func (tl *Logger) Debugf(f string, a ...interface{}) { tl.logf(f, a...) }
+
```
marcinzaremba, line 214: Not sure about this solution yet. Might be that I would change it to pass `Logger` explicitly to the package.

routesrv/routesrv_test.go
```diff
@@ -0,0 +1,78 @@
...
+var tl *loggingtest.Logger
+
+func TestMain(m *testing.M) {
+	flag.Parse()
+	tl = loggingtest.New()
+	logrus.AddHook(tl)
+	os.Exit(m.Run())
+}
+
+func newTestServer(t *testing.T) *routesrv.RouteServer {
```
marcinzaremba, line 26: TODO: Proper test cleanup and structure of helpers.

logging/loggingtest/logger.go
aryszka: the purpose of it is not clear to me. Where would it be used? What does the name 'Fire' mean?

marcinzaremba: This is an implementation of so-called `Hook` for logrus logging lib (https://github.com/sirupsen/logrus#hooks). It is utilized in test setup: https://github.com/zalando/skipper/pull/1848/files#diff-e856a3236a5ba5505ac2fea89a1b7a62b9db3282be1cb8a00eed0ea9de177d2bR26 in order to "forward" all the logs from the logrus to to the test logger to be able to assert on it. I tested it as an alternative to passing logger directly to the `routesrv.RouteServer` for test purposes (as it is implemented in the `routing` library for example). It works fine. I exercised this option to check how it behaves, because:
- i would like to avoid passing a specific logger type (`logrus.Logger`) as an explicit dependency to our new library,
- why not already existing Skipper logger interface as type? (https://github.com/zalando/skipper/blob/master/logging/logger.go#L9); because it does not account for structure logging and I would like not to resign from structure logging, because of that,
- so alternatively new logging interface would have needed to be developed (might be just private to this lib) or the current one extended; which IMHO is a bigger decision to discuss.

aryszka: perfect, thanks for the explanation

aryszka:
:+1:

routesrv/routesrv.go
```diff
@@ -0,0 +1,154 @@
...
+// ServeHTTP serves kept eskip-formatted routes under /routes
+// endpoint. Additionally it provides a simple health check under
+// /health and Prometheus-compatible metrics under /metrics.
+func (rs *RouteServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	rs.server.Handler.ServeHTTP(w, r)
+}
+
+func newShutdownFunc(rs *RouteServer) func(delay time.Duration) {
+	once := &sync.Once{}
+	rs.wg.Add(1)
```
AlexanderYastrebov, line 105: Hm, not sure why we need to tick wg here...

routesrv/routesrv.go
```diff
@@ -0,0 +1,154 @@
...
+	signal.Notify(sigs, syscall.SIGTERM)
+	go func() {
+		<-sigs
+		shutdown(opts.WaitForHealthcheckInterval)
+	}()
+
+	rs.StartUpdates()
+
+	if err = rs.server.ListenAndServe(); err != http.ErrServerClosed {
+		go shutdown(0)
```
szuecs, line 146: In order we want to run in production in Kubernetes, I think we need to change this to use `opt.ShutdownDelay` a `time.Duration` we set via flag.

routesrv/routesrv.go
AlexanderYastrebov: Ah, I see, we reuse it wait for shutdown completion

routesrv/routesrv.go
```diff
@@ -0,0 +1,154 @@
...
+// StartUpdates starts the data source polling process.
+func (rs *RouteServer) StartUpdates() {
+	rs.wg.Add(1)
+	go rs.poller.poll(rs.wg)
+}
+
+// StopUpdates stop the data source polling process.
+func (rs *RouteServer) StopUpdates() {
+	rs.poller.quit <- struct{}{}
+}
```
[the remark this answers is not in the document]
marcinzaremba: `wg` stayed here only for convenience when `Run` call is used: https://github.com/zalando/skipper/pull/1848/files#diff-903c01df2b8ad59771be10eabada676a9161af143e6ce387f9abe60070af5eb3R128 in order to wait for all the async activities with a single barrier. `StartUpdates` is not meant to be called multiple times, but you're right that now no one prevents doing it and neighboring `WaitGroup` can suggest so.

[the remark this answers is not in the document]
AlexanderYastrebov: > it should close the channel and wait on the waitgroup for all threads to complete

it can not since we reuse wg for waiting for shutdown completion. IMO we should not reuse but rather have a shutdown channel like 
https://github.com/zalando/skipper/blob/63b27681807f962a15372c41b43bb0ceea575068/skipper.go#L1080

(name `idleConnsCH` is a bit misleading though)

routesrv/routesrv.go
marcinzaremba: Indeed. It is a simple convenience that helps treating both shutdown scenarios the same way and spares additional syncing structures, however, I admit, is harder to read / express the purpose.

routesrv/routesrv_test.go
```diff
@@ -0,0 +1,223 @@
...
+	handler.set(newKubeAPI(t, loadKubeYAML(t, "testdata/lb-target-single.yaml")))
+	if err := tl.WaitForN(routesrv.LogRoutesUpdated, 2, waitTimeout*2); err != nil {
+		t.Error("routes not updated")
+	}
+	w2 := getRoutes(rs)
+
+	if bytes.Equal(w1.Body.Bytes(), w2.Body.Bytes()) {
+		t.Error("route contents were not updated")
+	}
+}
```
szuecs, line 223: great tests!
What about adding one that tests that we actually update in case of w1 is different from w2?

routesrv/testdata/lb-target-multi.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  namespace: namespace1
+  name: ingress1
+spec:
+  backend:
+    serviceName: service1
+    servicePort: port1
```
szuecs, line 9: maybe clarify in the file name that we test `default route`, which is what `spec.backend` defines AFAIR.

routesrv/testdata/lb-target-multi.eskip
```diff
@@ -0,0 +1,15 @@
+// default backend, target 1:
+kube_namespace1__ingress1______:
+  *
+  -> <roundRobin, "http://42.0.1.2:8080", "http://42.0.1.3:8080">;
```
szuecs, line 4: I think we should add another test that has no default backend.
The normal case for us at least is, that we have no default backend.

szuecs (changes requested):
Great work!
Some small things with applying godoc and a bit more of testing, which I think is a blocker, before merge.

routesrv/options.go
```diff
@@ -0,0 +1,111 @@
...
+	"regexp"
+	"time"
+
+	"github.com/zalando/skipper/dataclients/kubernetes"
+	"github.com/zalando/skipper/eskip"
+)
+
+// Options for initializing/running RouteServer
+type Options struct {
+	// Network address that routesrv should listen on.
```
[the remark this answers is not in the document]
marcinzaremba: This and all below was copied straight from the main `skipper.Options` as-is to avoid diverging the common options too much. Would we want to address godoc for both places in scope of this PR?

routesrv/routesrv.go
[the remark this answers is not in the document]
AlexanderYastrebov: > StartUpdates is not meant to be called multiple times

Then there is no need to tick wg here and pass it to the poller at all, the quit channel should be enough to control the poller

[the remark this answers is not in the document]
marcinzaremba: I think it would be beneficial if we divide the discussion to separate topics and conquer with a suited solution:
1. Do we want / is it beneficial to run multiple polling goroutines?
2. Do we want to enable them to be stopped and started as many times as needed or just start once and be able to stop once?

IMHO Ad 1. for the time being this is not needed, 2. I do not think this is needed as well.

Makefile
```diff
@@ -29,6 +29,9 @@ eskip: $(SOURCES) bindir
 webhook: $(SOURCES) bindir
 	GO111MODULE=$(GO111) go build -ldflags "-X main.version=$(VERSION) -X main.commit=$(COMMIT_HASH)" -o bin/webhook ./cmd/webhook/*.go
 
+routesrv: $(SOURCES) bindir
+	GO111MODULE=$(GO111) go build -ldflags "-X main.version=$(VERSION) -X main.commit=$(COMMIT_HASH)" -o bin/routesrv ./cmd/routesrv/*.go
```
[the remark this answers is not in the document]
marcinzaremba: For the time being we solve for Kubernetes, however there is nothing in code that blocks or is designed specifically that would block adding others IMHO. I would be agile here and wait if there is a need for more data sources. Currently we know that with Kubernetes one there can be scalability problems and this is where our focus is.

routesrv/polling.go
```diff
@@ -0,0 +1,111 @@
...
+	LogRoutesEmpty          = "received empty routes; ignoring"
+	LogRoutesInitialized    = "routes initialized"
+	LogRoutesUpdated        = "routes updated"
+)
+
+var (
+	pollingStarted = promauto.NewGauge(prometheus.GaugeOpts{
+		Namespace: "routesrv",
+		Name:      "polling_started_timestamp",
+		Help:      "UNIX time when the routes polling has started",
```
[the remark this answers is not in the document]
marcinzaremba: Exactly. UNIX timestamp in seconds like `date +%s`. I can improve the help message to be more specific here.

routesrv/eskipbytes.go
```diff
@@ -0,0 +1,74 @@
...
+
+	if data, initialized := e.bytes(); initialized {
+		w.Write(data)
+	} else {
+		w.WriteHeader(http.StatusNotFound)
+	}
+}
+
+// eskipBytesStatus serves as an HTTP health check for the referenced eskipBytes.
+// Reports healthy only when the bytes were initialized (set at least once).
```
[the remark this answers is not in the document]
marcinzaremba: Good point.

routesrv/routesrv.go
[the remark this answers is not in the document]
marcinzaremba: > Then there is no need to tick wg here and pass it to the poller at all, the quit channel should be enough to control the poller

You're right. For the current shape of the code yes, because there is no real action taken when `quit` "message" is received. The other signal would only be needed if we really wanted to wait for the poller to finish a potential cleanup / actions after the quit signal had been received. My thinking was rather for correctness to make sure the goroutine is really over and guard against potential bugs as well when some handling is added for handling quit signal in the poller. It might have been a premature optimization. I am not really attached to this solution, I rather treat is a matter of personal preference / style.

routesrv/eskipbytes.go
```diff
@@ -0,0 +1,74 @@
...
+type eskipBytesStatus struct {
+	b *eskipBytes
+}
+
+const msgRoutesNotInitialized = "routes were not initialized yet"
+
+func (s *eskipBytesStatus) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
marcinzaremba: Good point.

routesrv/routesrv.go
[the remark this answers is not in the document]
marcinzaremba: @szuecs @aryszka what do you think?

routesrv/routesrv.go
marcinzaremba: In my understanding there is no need for that. If the error received **is not** `http.ErrServerClosed`, it would suggest it is a startup error and thus the server was never started. I think it is just better to fail fast in this kind of scenario. Otherwise we would just wait for `opt.ShutdownDelay` without a need, thus delaying detection of a potential configuration failure for example.
All of this reasoning would be invalid, if other errors can be returned **after** the server was started. Maybe I am missing something, is there a possibility for other error than `http.ErrServerClosed` after the server is properly started?

routesrv/routesrv_test.go
marcinzaremba: Maybe I misunderstood your message but I think this is exactly this test:
1. We start the API server populated with resources (`testdata/lb-target-multi.yaml`) and polling.
2. We make sure the routes corresponding to the API resources were initialized.
3. We **change** what we serve with the API server (`testdata/lb-target-single.yaml`).
4. We make sure the new routes corresponding to the new API resources were updated.
5. We assert if we actually served something new (expecting the change, because of the change in the API server) by checking if `w1` is different from `w2`.

Maybe you can rephrase or clarify? I am also open for a quick call to do that.

routesrv/testdata/lb-target-multi.eskip
marcinzaremba: For this one and one below: My assumption was that the underlying data client is tested thoroughly already in other packages and the purpose of this test suite is to test `routesrv`. So I actually copied those two fixtures kinda randomly, because there were small enough and simply different. I might have chosen completely different ones, because I just needed some to feed to the test API server. Do you think we should test the portions of the Kubernetes data client behavior here as well?

routesrv/polling.go
```diff
@@ -0,0 +1,111 @@
...
+			span.LogKV(
+				"event", "error",
+				"message", msg,
+			)
+		case routesCount > 0:
+			routesBytes, initialized = p.b.formatAndSet(routes)
+			logger := log.WithFields(log.Fields{"count": routesCount, "bytes": routesBytes})
+			if initialized {
+				logger.Info(LogRoutesInitialized)
+				span.SetTag("routes.initialized", true)
```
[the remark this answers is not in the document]
marcinzaremba: Because only when we have some routes (`initialized`) our routes are considered valid, thus we are able to serve successfully to Skipper. Otherwise we can serve, however potentially with no use for Skipper. Specifically for Kubernetes data client no routes is not considered a valid state/update and is discarded as far as I understand the current implementation.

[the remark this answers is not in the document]
marcinzaremba: So we want to know / measure time from start to being setup to serve valid routes.

routesrv/eskipbytes.go
```diff
@@ -0,0 +1,74 @@
...
+
+	return e.data, e.initialized
+}
+
+// formatAndSet takes a slice of routes and stores them eskip-formatted
+// in a synchronized way. It returns a number of stored bytes and a boolean,
+// being true, when the stored bytes were set for the first time.
+func (e *eskipBytes) formatAndSet(routes []*eskip.Route) (int, bool) {
+	buf := &bytes.Buffer{}
+	eskip.Fprint(buf, eskip.PrettyPrintInfo{Pretty: false, IndentStr: ""}, routes...)
```
marcinzaremba, line 38: Better to pretty print for debugging purposes. Additional bytes do not really matter.

szuecs (approved):

szuecs:
:+1:

PR #1852, opened 2021-09-09
Closes backend recorder test server when done waiting

### LINKED ISSUE #1849: testing: close the underlying server of the backendtest.BackendRecorderHandler
the servers created here https://github.com/zalando/skipper/blob/master/proxy/backendtest/backendrecorder.go#L58 , should be closed when not used anymore

AlexanderYastrebov:
Fixes #1849

### DIFF
--- a/proxy/backendtest/backendrecorder.go
+++ b/proxy/backendtest/backendrecorder.go
@@ -19,12 +19,12 @@ type BackendRecorderHandler struct {
 	server   *httptest.Server
 	requests []RecordedRequest
 	mutex    sync.RWMutex
-	Done     <-chan time.Time
+	done     <-chan time.Time
 }
 
 func NewBackendRecorder(closeAfter time.Duration) *BackendRecorderHandler {
 	handler := &BackendRecorderHandler{
-		Done: time.After(closeAfter),
+		done: time.After(closeAfter),
 	}
 	server := httptest.NewServer(handler)
 	handler.server = server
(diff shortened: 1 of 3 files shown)


AlexanderYastrebov:
Build fails on golang1.17, requires https://github.com/zalando/skipper/pull/1853

proxy/backendtest/backendrecorder.go
```diff
@@ -60,6 +60,7 @@ func (rec *BackendRecorderHandler) GetURL() string {
 	return rec.server.URL
 }
 
-func (rec *BackendRecorderHandler) Close() {
+func (rec *BackendRecorderHandler) Done() {
```
szuecs, line 63: I think the func should not be renamed here, because Done() is used in stdlib waitgroup to signal that the goroutine has done the work. In this case we Close() our server if we get a signal via done channel.

szuecs:
@AlexanderYastrebov I think the changes are fine besides the rename.

proxy/backendtest/backendrecorder.go
AlexanderYastrebov: Before this change user of the Recorder would have to read (and block) on the done channel. This change wraps it into the method to also close the TestServer after wait time has elapsed. There is no need to have separate Close method.
I can rename `Done` to `Wait` if that feels better.

szuecs: Why `Done` to `Wait` and not to `Close`?

AlexanderYastrebov: > Why Done to Wait

Because users are supposed to call `recorder.Done()` at the point where they expect all the requests to complete in order to perform checks on the recorded data. If we call it `Close` there is a high chance that it would be used like
```go
recorder := NewBackendRecorder(10 * time.Millisecond)
defer recorder.Close()
```
which is wrong.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #1853, opened 2021-09-09
Adds golang1.17 build constraints

AlexanderYastrebov:
Golang 1.17 adds new syntax for build constraints, `gofmt` adds new
`//go:build` lines along with `// +build` during transition phase.
```
docker run --rm -v "$PWD":/usr/src/myapp -w /usr/src/myapp golang:1.17.0 sh -c "make fmt"
```
See https://tip.golang.org/doc/go1.17#gofmt

Note: the builds started to fail because we ask for golang version >= 1.6 (see https://github.com/zalando/skipper/pull/1835) and workflow started to build using 1.17

### DIFF
--- a/proxy/backendratelimit_test.go
+++ b/proxy/backendratelimit_test.go
@@ -1,3 +1,4 @@
+//go:build !race || redis
 // +build !race redis
 
 package proxy_test
--- a/ratelimit/cluster_test.go
+++ b/ratelimit/cluster_test.go
@@ -1,3 +1,4 @@
+//go:build !race || redis
 // +build !race redis
 
 package ratelimit
--- a/dataclients/routestring/example_test.go
+++ b/dataclients/routestring/example_test.go
@@ -1,3 +1,4 @@
+//go:build !race
 // +build !race
 
 package routestring_test
--- a/etcd/etcdtime_test.go
+++ b/etcd/etcdtime_test.go
@@ -1,3 +1,4 @@
+//go:build !race
 // +build !race
 
 package etcd
--- a/plugins_test.go
+++ b/plugins_test.go
@@ -1,3 +1,4 @@
+//go:build !race
 // +build !race
 
 package skipper
--- a/proxy/breaker_test.go
+++ b/proxy/breaker_test.go
@@ -1,3 +1,4 @@
+//go:build !race
 // +build !race
 
 package proxy_test
--- a/ratelimit/swim_test.go
+++ b/ratelimit/swim_test.go
@@ -1,3 +1,4 @@
+//go:build !race
 // +build !race
 
 package ratelimit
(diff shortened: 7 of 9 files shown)


proxy/backendratelimit_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build !race || redis
```
AlexanderYastrebov, line 1: Interestingly the new syntax was introduced to avoid confusion between && and ||. Now it is clear that it is OR and I am not sure that `// +build !race redis` is actually correct, maybe it should be `// +build !race,redis` which would translate into `//go:build !race && redis`

AlexanderYastrebov: Old syntax ref https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md#background-syntax

szuecs: I think `//go:build !race && redis` makes sense.

AlexanderYastrebov: I am wondering do we need `redis` tag at all?

szuecs: not everyone needs to have a local redis running

AlexanderYastrebov: The point is since it is `//go:build !race || redis` these tests were always executed during normal (not -race) tests regardless of redis tag

AlexanderYastrebov: > not everyone needs to have a local redis running

It requires redis binary as we start new redis server for each test
https://github.com/zalando/skipper/blob/6b089867410bf4b9f56b4686450cf95301ac02c7/net/redistest/redistest.go#L14

AlexanderYastrebov:
:+1:

proxy/backendratelimit_test.go
szuecs: I think what we want here is `//go:build !race && redis`

AlexanderYastrebov: We can merge like this since it is equivalent to the existing setup to unblock other builds and reconsider redis condition separately

ratelimit/cluster_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build !race || redis
```
szuecs, line 1: I guess this should also be &&

szuecs:
:+1:

ratelimit/cluster_test.go
AlexanderYastrebov: https://github.com/zalando/skipper/issues/1854

proxy/backendratelimit_test.go
AlexanderYastrebov: https://github.com/zalando/skipper/issues/1854

PR #1856, opened 2021-09-10
fix invalid escape sequence

danielseider:
fix invalid escape sequence

### DIFF
--- a/dataclients/kubernetes/deploy/apply/routegroups_crd.yaml
+++ b/dataclients/kubernetes/deploy/apply/routegroups_crd.yaml
@@ -107,7 +107,7 @@ spec:
               hosts:
                 description: List of hostnames for the RouteGroup.
                 items:
-                  pattern: "^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$"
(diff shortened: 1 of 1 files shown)


szuecs:
@danielseider thanks for reaching out. I think it might be an issue with evaluation. Do you use shell that could evaluate the string, such that you need to double escape it?

AlexanderYastrebov:
I think dot supposed to be escaped by backslash in regexp and backslash itself should be escaped in quoted yaml string
```
$ yamllint -d relaxed dataclients/kubernetes/deploy/apply/routegroups_crd.yaml | grep error
  110:61    error    syntax error: found unknown escape character '.'
```
so the fix seems correct.
See also http://blogs.perl.org/users/tinita/2018/03/strings-in-yaml---to-quote-or-not-to-quote.html

szuecs:
@danielseider can you do a `git ci --amend --signoff` and force push to fix the DCO check, please?

szuecs:
@danielseider can you squash it to one with rebase and make sure it's signed off?

https://github.com/zalando/skipper/pull/1856/commits shows 2 commits and I guess only one is signed off. DCO bot is a bit picky :( 

Thanks!

AlexanderYastrebov:
:+1:

szuecs:
👍

AlexanderYastrebov:
@danielseider Thank you.

PR #1861, opened 2021-09-21
Fixes testable example

### LINKED ISSUE #1860: eskip testable example is incorrectly passing
👋  Hello, 

I am investigating the golang issue: https://github.com/golang/go/issues/48362, where testable examples are causing false positives when the output comment block is not the last comment block example.

I ran a parser against the [top golang repos](https://github.com/kaxap/arl/blob/master/README-Go.md), including this repo, and I found this bug with one of your [testable examples](https://github.com/zalando/skipper/blob/11b7acad09fc7c0873b3c2cd4779f539a8aa9bcd/eskip/example_test.go#L24-L78).

The TODO comment block should be moved before the output comment block like so:

```
func Example() {
	code := `
		// Skipper - Eskip:
		// a routing table document, containing multiple route definitions
		// route definition to a jsx page renderer
		route0:
			PathRegexp(/\.html$/) && HeaderRegexp("Accept", "text/html") ->
			modPath(/\.html$/, ".jsx") ->
			requestHeader("X-Type", "page") ->
			"https://render.example.org";
		route1: Path("/some/path") -> "https://backend-0.example.org"; // a simple route
		// route definition with a shunt (no backend address)
		route2: Path("/some/other/path") -> static("/", "/var/www") -> <shunt>;
		// route definition directing requests to an api endpoint
		route3:
			Method("POST") && Path("/api") ->
			requestHeader("X-Type", "ajax-post") ->
			"https://api.example.org";
		// route definition with a loopback to route2 (no backend address)
		route4: Path("/some/alternative/path") -> setPath("/some/other/path") -> <loopback>;
        // route definition which forwards rest of requests (no backend address)
		route5: * -> requestHeader("X-Type", "page") -> <forward>;
		`

	routes, err := eskip.Parse(code)
	if err != nil {
		log.Println(err)
		return
	}

	format := "%v: [match] -> [%v filter(s) ->] <%v> \"%v\"\n"
	fmt.Println("Parsed routes:")
	for _, r := range routes {
		fmt.Printf(format, r.Id, len(r.Filters), r.BackendType, r.Backend)
	}
	// TODO: try to represent this compressed output in a nicer way. This is is now somewhat confusing
	// considering that it looks almost like eskip but it isn't.

	// output:
	// Parsed routes:
	// route0: [match] -> [2 filter(s) ->] <network> "https://render.example.org"
	// route1: [match] -> [0 filter(s) ->] <network> "https://backend-0.example.org"
	// route2: [match] -> [1 filter(s) ->] <shunt> ""
	// route3: [match] -> [1 filter(s) ->] <network> "https://api.example.org"
	// route4: [match] -> [1 filter(s) ->] <loopback> ""


}
```

When I made this change, the example failed:
```
$ go test .
2021/09/21 19:54:18 parse failed after token >, position 1015: syntax error
--- FAIL: Example (0.00s)
got:

want:
Parsed routes:
route0: [match] -> [2 filter(s) ->] <network> "https://render.example.org"
route1: [match] -> [0 filter(s) ->] <network> "https://backend-0.example.org"
route2: [match] -> [1 filter(s) ->] <shunt> ""
route3: [match] -> [1 filter(s) ->] <network> "https://api.example.org"
route4: [match] -> [1 filter(s) ->] <loopback> ""
FAIL
FAIL    github.com/zalando/skipper/eskip        0.016s
FAIL
```

AlexanderYastrebov:
The https://github.com/zalando/skipper/issues/1860 pointed out a bug
https://github.com/golang/go/issues/48362 that leads to the false
positive testable example caused by a trailing comment.

Moreover https://github.com/zalando/skipper/pull/918 changed the test
input without changing the output and that went unnoticed due to this bug.

This change removes trailing comment to fix the example.
It also removes the test input instead of adding a correct output because
the input syntax is incorrect therefore misleading.

Fixes #1860

### DIFF
--- a/eskip/example_test.go
+++ b/eskip/example_test.go
@@ -48,9 +48,6 @@ func Example() {
 
 		// route definition with a loopback to route2 (no backend address)
 		route4: Path("/some/alternative/path") -> setPath("/some/other/path") -> <loopback>;
-
-        // route definition which forwards rest of requests (no backend address)
-		route5: * -> requestHeader("X-Type", "page") -> <forward>;
 		`
 
 	routes, err := eskip.Parse(code)
@@ -72,9 +69,6 @@ func Example() {
 	// route2: [match] -> [1 filter(s) ->] <shunt> ""
 	// route3: [match] -> [1 filter(s) ->] <network> "https://api.example.org"
 	// route4: [match] -> [1 filter(s) ->] <loopback> ""
-
-	// TODO: try to represent this compressed output in a nicer way. This is is now somewhat confusing
-	// considering that it looks almost like eskip but it isn't.
 }
 
 func ExampleFilter() {


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #1862, opened 2021-09-23
Adds a flag to set `X-Forwarded-Port` header

AlexanderYastrebov:


config/config.go
```diff
@@ -372,7 +372,8 @@ func NewConfig() *Config {
 	flag.Var(cfg.ForwardedHeadersList, "forwarded-headers", "comma separated list of headers to add to the incoming request before routing\n"+
 		"X-Forwarded-For sets or appends with comma the remote IP of the request to the X-Forwarded-For header value\n"+
 		"X-Forwarded-Host sets X-Forwarded-Host value to the request host\n"+
-		"X-Forwarded-Proto=http or X-Forwarded-Proto=https sets X-Forwarded-Proto value")
+		"X-Forwarded-Port=<port> sets X-Forwarded-Port value\n"+
+		"X-Forwarded-Proto=<http|https> sets X-Forwarded-Proto value")
```
szuecs, line 376: these autoline breaks ...

config/config.go
```diff
@@ -846,14 +847,16 @@ func (c *Config) parseHistogramBuckets() ([]float64, error) {
...
-		case "X-Forwarded-For":
+		switch {
+		case header == "X-Forwarded-For":
 			c.ForwardedHeaders.For = true
-		case "X-Forwarded-Host":
+		case header == "X-Forwarded-Host":
 			c.ForwardedHeaders.Host = true
-		case "X-Forwarded-Proto=http":
+		case strings.HasPrefix(header, "X-Forwarded-Port="):
+			c.ForwardedHeaders.Port = strings.TrimPrefix(header, "X-Forwarded-Port=")
```
szuecs, line 856: hmm so it can be any kind of string?
Maybe better to check it's only a number of the valid port range (uint16)?

net/headers.go
```diff
@@ -14,6 +14,8 @@ type ForwardedHeaders struct {
 	PrependFor bool
 	// Sets X-Forwarded-Host to the request host
 	Host bool
+	// Sets X-Forwarded-Port value
+	Port string
```
szuecs, line 18: I think better type safe int or uint16

config/config.go
AlexanderYastrebov: I wrote that initially but then I think it does not worth it since there also will be a uint16->string conversion to the header value so I've though users are ably to provide correct value

szuecs: we could also just validate with strconv.Atoi if it returns error and if not we add the string to the port header. So no 2x conversion required.
For me the resasoning is also fine so I add my +1, decide yourself.

szuecs:
:+1:

aryszka:
:+1:

aryszka:
lgtm, but I am also wondering whether the X-Forwarded-Port, in the simplest cases could be just copied from the port that Skipper is listening on. I see, of course, how there are more complicated cases, too.

PR #1863, opened 2021-09-23
Fix/optional trailing dot kube

szuecs:
Kubernetes generated routes have optional port but not optional trailing dot, such that

`example.org.` in the host header would not match.
There is a longer discussion in https://lists.w3.org/Archives/Public/ietf-http-wg/2016JanMar/0447.html that turns out to show that proxies and some clients remove trailing dot in SNI and Host header which is good, but we don't fix bogus clients in this case.

A follow up is to create a filter that removes the trailing dot from the host header, such that in case preserHost(true) we would send the host header without trailing dot if the client sends a trailing dot

### DIFF
--- a/dataclients/kubernetes/hosts.go
+++ b/dataclients/kubernetes/hosts.go
@@ -18,7 +18,7 @@ func createHostRx(hosts ...string) string {
 		// trailing dots and port are not allowed in kube
 		// ingress spec, so we can append optional setting
 		// without check
-		hrx[i] = strings.Replace(host, ".", "[.]", -1) + "([.])?(:[0-9]+)?"
+		hrx[i] = strings.Replace(host, ".", "[.]", -1) + "[.]?(:[0-9]+)?"
 	}
 
 	return "^(" + strings.Join(hrx, "|") + ")$"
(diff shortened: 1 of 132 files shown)


dataclients/kubernetes/hosts.go
```diff
@@ -15,7 +15,10 @@ func createHostRx(hosts ...string) string {
 
 	hrx := make([]string, len(hosts))
 	for i, host := range hosts {
-		hrx[i] = strings.Replace(host, ".", "[.]", -1) + "(:[0-9]+)?"
+		// trailing dots and port are not allowed in kube
+		// ingress spec, so we can append optional setting
+		// without check
+		hrx[i] = strings.Replace(host, ".", "[.]", -1) + "([.])?(:[0-9]+)?"
```
AlexanderYastrebov, line 21: I think there is no need to have a capture group around the dot, i.e. `[.]?` should work

szuecs: fixed

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #1864, opened 2021-09-23
Feature/filter remove trailing dot

szuecs:
Follow up of #1863 
We are able to fix our behavior as a client. 

There is a longer discussion in https://lists.w3.org/Archives/Public/ietf-http-wg/2016JanMar/0447.html that turns out to show that proxies and some clients remove trailing dot in SNI and Host header which is good, but we don't fix bogus clients in this case.

### DIFF
--- a/rfc/patchhost.go
+++ b/rfc/patchhost.go
@@ -6,5 +6,6 @@ import "strings"
 // see also the discussion in
 // https://lists.w3.org/Archives/Public/ietf-http-wg/2016JanMar/0430.html.
 func PatchHost(host string) string {
+	host = strings.Replace(host, ".:", ":", -1)
 	return strings.TrimSuffix(host, ".")
 }
--- a/rfc/patchhost.go
+++ b/rfc/patchhost.go
@@ -1,0 +1,10 @@
+package rfc
+
+import "strings"
+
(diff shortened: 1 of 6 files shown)


rfc/patchhost.go
```diff
@@ -0,0 +1,10 @@
+package rfc
+
+import "strings"
+
+// PatchHost returns a host string without trailing dot. For details
+// see also the discussion in
+// https://lists.w3.org/Archives/Public/ietf-http-wg/2016JanMar/0430.html.
+func PatchHost(host string) string {
+	return strings.TrimSuffix(host, ".")
```
AlexanderYastrebov, line 9: https://pkg.go.dev/net/http#Request
> It may be of the form "host:port"

AlexanderYastrebov: Maybe?
```go
	host = strings.Replace(host, ".:", ":", -1)
	host = strings.TrimSuffix(host, ".")
```

szuecs: yes thanks there was also more wrong because of outgoinghost..., now it's fixed in all cases (tested with curl and nc)

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #1878, opened 2021-10-08
fix cve report from dependency

szuecs:
we were not using this vulnerable code, but GH report is always good to fix


szuecs:
:+1:

PR #1881, opened 2021-10-12
fix https redirects for fabric and custom route cases

szuecs:
fix ingress having predicate annotation or custom route annotation that is contrary to our https redirect routes


marcinzaremba:
👍

PR #1888, opened 2021-10-18
Fixing string escape issue

marcinzaremba:
Fixes https://github.com/zalando/skipper/issues/562
This is a breaking change.
It sets in stone how eskip parser handles escaping.


AlexanderYastrebov:
There is a somewhat related PR #1938

PR #1889, opened 2021-10-20
OT: enable group by client IP query

szuecs:
to be able to do group by client IP query to figure out if specific client produce an obstacle strip the port


szuecs:
:+1:

Issue #1477, opened 2020-07-08
kubernetes: Add support for new ingress resource field `pathType` added in Kubernetes v1.18


mikkeloscar:
Kubernetes v1.18 introduces changes to the Ingress resource: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.18.md#extending-ingress-with-and-replacing-a-deprecated-annotation-with-ingressclass

One of these changes is adding a `pathType` field to `ingress.spec.rules.http.paths.pathType` See relevant part of this [Ingress KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190125-ingress-api-group.md#paths-proposal)

Skipper already has some support like this via the `zalando.org/skipper-ingress-path-mode` annotation, but it would make sense to support the new `pathType` field directly in the Ingress spec.

szuecs:
SGTM

szuecs:
https://github.com/zalando/skipper/pull/1892 implements v1 ingress with the new pathType and new kind and released in https://github.com/zalando/skipper/releases/tag/v0.13.149

PR #1892, opened 2021-10-25
Feature/ingress v1

### LINKED ISSUE #1824: Kubernetes 1.22 api removed beta ingress
IngressesClusterURI for example is using beta and our tests should also be changed to return v1 networking objects.

Related #1477 #1478

according to https://www.civo.com/learn/migrating-your-ingresses-in-k3s-1-20 spec change is also in service that now has 
```
port:
  number:
  name:
```

see also https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122

szuecs:
fix #1824 and #1477

You can check what needs to be done in https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122
As an example, this shows what changed from our pov (TLS won't be implemented and external ref might also not a thing because we have no feature that we want to hook in via this):
```yaml
apiVersion: networking.k8s.io/v1  # of course version changed
kind: Ingress
metadata:
  namespace: foo
  name: qux
spec:
  defaultBackend:  # this was called "backend" and here also service object instead of serviceName and servicePort
    service:
      name: bar
      port:
        number: 1234   # yes we have a real number or a name as member of port :)
  rules:
  - host: example.ingress.cluster.local
    http:
      paths:
      - path: "/"
        pathType: Exact
        backend:
          service:      # service object instead of serviceName and servicePort
            name: bar
            port:
              name: baz
```

I decided to not mix the types we have for v1beta1, because it makes everything very hard to implement with type switches and interface{} usage and likely proxy types. I think a bit of function duplication in this case for a period of time until we abandon v1beta1 is fine.


aryszka:
I am ok with this approach of duplication, because this way it will be easier to clean up the beta related types and functions. However, this code may stick with us for a relatively long time with us, since we don't know yet when we will delete the old code. Would it make sense to mark the copied functions, the original ones, the new types and the old ones in a way, that could serve as a guidance in future during the deleting the beta related code, also with the consideration that potentially someone else will do it?

aryszka:
another idea, would it make sense to store the copied types and functions in separate go files?

dataclients/kubernetes/clusterclient.go
```diff
@@ -374,7 +414,16 @@ func (c *clusterClient) logMissingRouteGroupsOnce() {
 }
 
 func (c *clusterClient) fetchClusterState() (*clusterState, error) {
-	ingresses, err := c.loadIngresses()
+	var (
+		err         error
+		ingressesV1 []*definitions.IngressV1Item
+		ingresses   []*definitions.IngressItem
+	)
+	if c.ingressV1 {
```
aryszka, line 422: i am trying to think about a migration path. I thought we would support both variants for a while. How will it work? Wouldn't we end up with a chicken-and-egg situation when considering migrating the resources and skipper to v1?

AlexanderYastrebov:
> another idea, would it make sense to store the copied types and functions in separate go files?

I agree, maybe copy `ingress.go` to `ingressV1.go` and remove reused functions from v1 without touching beta. 
This should allow at least to diff the two files.

dataclients/kubernetes/clusterclient.go
szuecs: Ok this questions is a miss understanding of Kubernetes.
In Kubernetes you ask as client "please give me resource A in version vN". Kubernetes will translate v1 resources to v1beta1. In kubernetes 1.22 this won't happen anymore and since 1.18 (IIRC) you can already fetch v1 ingress, no matter who created the ingress in which version.

szuecs:
@aryszka I think it's a great idea to separate ingressv1beta1 and ingressv1 definitions. 
I would suggest that I move v1beta1 to a separate file and ingressv1, too.
Metadata is likely the only one that rest in the former file.
Like this we can easily clean up v1beta1, deleting 1 file and remove the v1 switching in functions.

szuecs:
I changed moved things to different files.
The only mixed file is now dataclients/kubernetes/ingress.go, the dataclients/kubernetes/definitions/ingressv1beta1.go can be just deleted similar to the tests whenever we delete the v1beta1 support.

szuecs:
:+1:

config/config.go
```diff
@@ -384,6 +385,7 @@ func NewConfig() *Config {
 	flag.BoolVar(&cfg.KubernetesHealthcheck, "kubernetes-healthcheck", true, "automatic healthcheck route for internal IPs with path /kube-system/healthz; valid only with kubernetes")
 	flag.BoolVar(&cfg.KubernetesHTTPSRedirect, "kubernetes-https-redirect", true, "automatic HTTP->HTTPS redirect route; valid only with kubernetes")
 	flag.IntVar(&cfg.KubernetesHTTPSRedirectCode, "kubernetes-https-redirect-code", 308, "overrides the default redirect code (308) when used together with -kubernetes-https-redirect")
+	flag.BoolVar(&cfg.KubernetesIngressV1, "kubernetes-ingress-v1", false, "enable kubernetes ingress version v1, defaults to version v1beta1")
```
AlexanderYastrebov, line 388: flag name does not match yaml property name

dataclients/kubernetes/definitions/ingressv1.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
+)
+
+type IngressV1List struct {
```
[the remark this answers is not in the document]
szuecs: But we have stripped down the definitions and metadata to only hold in memory what we really need.
I won't pull it into this project.

config/config.go
szuecs: do you have any suggestion?

dataclients/kubernetes/definitions/ingressv1.go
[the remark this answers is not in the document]
AlexanderYastrebov: Ok

dataclients/kubernetes/ingress.go
```diff
@@ -244,6 +265,99 @@ func convertPathRule(
 	return r, nil
 }
 
+func convertPathRuleV1(
```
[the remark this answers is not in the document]
szuecs: I separated version specific functions to 2 different files

config/config.go
AlexanderYastrebov: Now it is ok

dataclients/kubernetes/clusterclient.go
aryszka: got it, thx

aryszka:
:+1:

dataclients/kubernetes/ingressv1.go
```diff
@@ -0,0 +1,429 @@
...
+	"fmt"
+	"regexp"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/dataclients/kubernetes/definitions"
+	"github.com/zalando/skipper/eskip"
+)
+
+func setPathV1(m PathMode, r *eskip.Route, pathType, path string) {
```
AlexanderYastrebov, line 14: Should we check for empty path like in beta?
```go
	if p == "" {
		return
	}
```

szuecs: That's interesting if someone specifies a valid v1 pathType but no path.

dataclients/kubernetes/testdata/ingressV1/eastwestrange/ewring-without-internal-host-1prule.yaml
```diff
@@ -0,0 +1,48 @@
...
+kind: Ingress
+metadata:
+  name: qux
+  namespace: foo
+spec:
+  rules:
+  - host: www.example.org
+    http:
+      paths:
+      - path: "/"
```
AlexanderYastrebov, line 11: I think this definition is not valid
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122
> pathType is now required for each specified path.

AlexanderYastrebov: A lot of v1 definitions are not valid. I am wondering if we could either validate them automatically or convert beta versions to v1 using https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/#install-kubectl-convert-plugin maybe

AlexanderYastrebov: There are also not test fixtures for pathType `Prefix`, `Exact` and `ImplementationSpecific`

szuecs: fixed and also added separate tests for `Prefix` and `Exact` and both with `ImplementationSpecific` in one ingress definition.

dataclients/kubernetes/ingressv1.go
szuecs: Empty paths are not path rules. 
Example source file:
```
kind: Ingress
metadata:
  name: qux
  namespace: foo
spec:
  rules:
  - host: www.example.org
    http:
      paths:
      - path: ""
        pathType: ImplementationSpecific
        backend:
          service:
            name: qux
            port:
              name: baz
```

Example runtime ingress fetched:
```
kind: Ingress
metadata:
  name: qux
  namespace: foo
  resourceVersion: "1022182097"
  uid: de7c7d39-c22b-496e-ac7f-41471319c4f5
spec:
  rules:
  - host: www.example.org
    http:
      paths:
      - backend:
          service:
            name: qux
            port:
              name: baz
        pathType: ImplementationSpecific
status:
  loadBalancer: {}
```

szuecs: checked the code you are right it would change behavior if you set pathType to `Prefix` or `Exact`, thanks again for the review!

szuecs:
I built an internal tool to validate the change.
I validated that our 182 clusters are having quite equal routes with v1 or v1beta1 configured kubernetes dataclient (without apimonitoring integration, which would be more complex to get all resources).
There was two interesting findings:
1. two clusters have each one ingress that already uses v1 `pathType: Exact`
2. in a dev cluster we have one ingress class annotation, which we don't consider in v1, which should be fixed, because of simpler migrations for everyone.

szuecs:
I fixed the  ingress class annotation handling to be compatible with v1beta1 in case we run v1. I think it's the safest way to change from v1beta1 to v1.

szuecs:
@AlexanderYastrebov I cleaned up the test that was cleaned up by #1902

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #1897, opened 2021-11-04
feat: downgrade level of log message

tkrop:
Signed-off-by: tkrop <tronje.krop@zalando.de>

Downgrades the level of some log messages to warning, since they are not really errors but only hints about an potentially unexpected filter behavior.


tkrop:
:+1:

AlexanderYastrebov:
:+1:

Issue #1899, opened 2021-11-05
Healthcheck endpoint using routesrv


AlexanderYastrebov:
**Is your feature request related to a problem? Please describe.**

Currently there is no separate healthcheck facility that would signal Skipper shutdown correctly.
There is a healthcheck that observes SIGTERM but it is part of Kubernetes client and e.g. will be disabled when Skipper is configured to fetch routes from `routesrv` (introduced by #1848)

**Describe the solution you would like**

Add a healthcheck feature configured via flag(s).
The implementation could be extracted from Kubernetes client into a dedicated dataclient. 
The Kubernetes dataclient healthcheck feature should be probably kept in place for backward compatibility but the implementation could be shared.

When Skipper fetches routes from `routesrv` and `routesrv` uses Kubernetes client, Skipper should provide proposed healthcheck endpoint and `routesrv` should have `-kubernetes-healthcheck=false` to not shadow the healthcheck route.

**Would you like to work on it?**
Yes

AlexanderYastrebov:
> Add a healthcheck feature configured via flag(s).

#1900 proposes `Shutdown` predicate that could be used to define customized healthcheck (status, content, allowed IPs) using e.g. `-inline-routes`. 

> The Kubernetes dataclient healthcheck feature should be probably kept in place for backward compatibility but the implementation could be shared.

If #1900 is accepted then kubernetes client could be refactored to use it instead of listening for `SIGTERM`.

AlexanderYastrebov:
#1900 introduced `Shutdown` predicate that could be used to define healthcheck routes.

The [initial idea](https://github.com/zalando/skipper/issues/1899#issuecomment-966481031) was to disable k8s healthcheck via `-kubernetes-healthcheck=false` in `routesrv` component and use `-inline-routes` to define healthcheck using `Shutdown` in `ingress` component that [fetches routes](https://github.com/zalando-incubator/kubernetes-on-aws/blob/6969289a867fb2f230e7044dee9e8ea1a1f1d72e/cluster/manifests/skipper/deployment.yaml#L93-L96) from `routesrv`.

#1918 proposes to use `Shutdown` predicate for kubernetes dataclient healthcheck. This way healthcheck routes become static and could be provided by `routesrv` but evaluated in `ingress` component, i.e. there is no need for `-kubernetes-healthcheck=false` and `-inline-routes`.

PR #1900, opened 2021-11-11
Adds `Shutdown` predicate

AlexanderYastrebov:
`Shutdown` predicate evaluates to true if Skipper is shutting down after receiving
`SIGTERM` signal and can be used to implement customized healthcheck.

Updates #1899

### DIFF
--- a/docs/reference/predicates.md
+++ b/docs/reference/predicates.md
@@ -223,6 +223,15 @@ route1: Path("/test") -> "http://www.zalando.de";
 route2: Path("/test") && False() -> "http://www.github.com";
 ```
 
+## Shutdown
+
+Evaluates to true if Skipper is shutting down. Can be used to create customized healthcheck.
+
+```
+health_up: Path("/health") -> inlineContent("OK") -> <shunt>;
+health_down: Path("/health") && Shutdown() -> status(503) -> inlineContent("shutdown") -> <shunt>;
+```
+
 ## Method
 
 The HTTP method that the request must match. HTTP methods are one of
--- a/predicates/primitive/shutdown.go
+++ b/predicates/primitive/shutdown.go
@@ -1,0 +1,51 @@
+package primitive
+
+import (
+	"net/http"
+	"os"
+	"os/signal"
+	"sync/atomic"
+	"syscall"
+
+	"github.com/zalando/skipper/predicates"
+	"github.com/zalando/skipper/routing"
+
+	log "github.com/sirupsen/logrus"
+)
+
+type shutdown struct {
+	inShutdown int32
+}
+
+// NewShutdown provides a predicate spec to create predicates
+// that evaluate to true if Skipper is shutting down
+func NewShutdown() routing.PredicateSpec {
+	s, _ := newShutdown()
+	return s
+}
+
+func newShutdown() (routing.PredicateSpec, chan os.Signal) {
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGTERM)
+	s := &shutdown{}
+	go func() {
+		<-sigs
+		log.Infof("Got shutdown signal for %s predicates", s.Name())
+		atomic.StoreInt32(&s.inShutdown, 1)
+	}()
+	return s, sigs
+}
+
+func (*shutdown) Name() string { return predicates.ShutdownName }
+
+// Create returns a Predicate that evaluates to true if Skipper is shutting down
+func (s *shutdown) Create(args []interface{}) (routing.Predicate, error) {
+	if len(args) != 0 {
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+	return s, nil
+}
+
+func (s *shutdown) Match(*http.Request) bool {
+	return atomic.LoadInt32(&s.inShutdown) != 0
+}
--- a/predicates/predicates.go
+++ b/predicates/predicates.go
@@ -22,6 +22,7 @@ const (
 	WeightName                = "Weight"
 	TrueName                  = "True"
 	FalseName                 = "False"
+	ShutdownName              = "Shutdown"
 	MethodName                = "Method"
 	MethodsName               = "Methods"
 	HeaderName                = "Header"
--- a/predicates/primitive/shutdown_test.go
+++ b/predicates/primitive/shutdown_test.go
@@ -1,0 +1,28 @@
+package primitive
+
+import (
+	"net/http"
+	"syscall"
+	"testing"
+	"time"
+)
+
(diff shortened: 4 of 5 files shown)


predicates/primitive/shutdown.go
```diff
@@ -0,0 +1,51 @@
+package primitive
```
AlexanderYastrebov, line 1: Decided not to introduce a new package for now.

docs/reference/predicates.md
```diff
@@ -223,6 +223,15 @@ route1: Path("/test") -> "http://www.zalando.de";
 route2: Path("/test") && False() -> "http://www.github.com";
 ```
 
+## Shutdown
+
+Evaluates to true if Skipper is shutting down. Can be used to create customized healthcheck.
```
AlexanderYastrebov, line 228: Healthcheck is implemented using two routes and relies on route higher weight when `Shutdown()` is true.
Maybe there are other usecases where users would like to have route enabled only when not in shutdown. For that we could add a companion opposite predicate, e.g. `Running()`.

An alternative naming could be a pair of `Up()` and `Down() == Shutdown()` predicates.

docs/reference/predicates.md
```diff
@@ -223,6 +223,15 @@ route1: Path("/test") -> "http://www.zalando.de";
...
+
+```
+health_up: Path("/health") -> inlineContent("OK") -> <shunt>;
+health_down: Path("/health") && Shutdown() -> status(503) -> inlineContent("shutdown") -> <shunt>;
```
szuecs, line 232: I think it's a great idea, composable and usable in all kind of dataclients.

docs/reference/predicates.md
szuecs: I like the pair, even if we do not use it in kubernetes and therefore Up() and Down() might be better names, but I think Shutdown() and Running() are maybe more clear to the user that search for this predicate...

predicates/primitive/shutdown.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func newShutdown() (routing.PredicateSpec, chan os.Signal) {
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGTERM)
+	s := &shutdown{}
+	go func() {
+		<-sigs
+		log.Infof("Got shutdown signal for %s predicates", s.Name())
+		atomic.StoreInt32(&s.inShutdown, 1)
```
szuecs, line 34: I think for an int64 on 64bit platforms you do not need an atomic or guards, because it's an aligned access and there is only one read and one write operation that should be in an aligned access. From that pov you should also try to omit 32bit access, because it's slower than 64bit access (32bit AFAIK needs an additional shift operation).
I also don't see a problem if we would have here a race, that would create healthcheck true, but you already got a shutdown signal in the same frame of a couple of µs.

szuecs (commented):
fine for me, but also reasoned about the atomic and 32 bit op :)

docs/reference/predicates.md
AlexanderYastrebov: Ok, lets have it as is and add the opposite when there is a usecase for it.

predicates/primitive/shutdown.go
AlexanderYastrebov: http.Server uses int32 for this purpose. On x86 atmic.Store/Load are inlined into three instructions https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/runtime/internal/atomic/atomic_amd64.s;l=159-163  

> I also don't see a problem if we would have here a race

I just want clean code, the difference would be nanoseconds

AlexanderYastrebov: Also https://golang.org/ref/mem#tmp_1

predicates/primitive/shutdown.go
```diff
@@ -0,0 +1,51 @@
...
+func (s *shutdown) Create(args []interface{}) (routing.Predicate, error) {
+	if len(args) != 0 {
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+	return s, nil
+}
+
+func (s *shutdown) Match(*http.Request) bool {
+	return atomic.LoadInt32(&s.inShutdown) != 0
+}
```
AlexanderYastrebov, line 51: @szuecs 
```
$ go tool objdump bin/skipper | grep -F 'primitive.(*shutdown).Match' -A5
TEXT github.com/zalando/skipper/predicates/primitive.(*shutdown).Match(SB) /.../shutdown.go
  shutdown.go:50        0xf6d3c0                8b08                    MOVL 0(AX), CX          
  shutdown.go:50        0xf6d3c2                85c9                    TESTL CX, CX            
  shutdown.go:50        0xf6d3c4                0f95c0                  SETNE AL                
  shutdown.go:50        0xf6d3c7                c3                      RET      
```

predicates/primitive/shutdown.go
szuecs: :+1:

szuecs:
:+1:

predicates/primitive/shutdown.go
szuecs:
```
  % cat foo.go
package main

import (
        "fmt"
        "time"
)

var i int

func foo() bool {
        return i != 0 // line 11
}

func main() {
        foo()
        go func() {
                i = 1  // line 17
        }()
        for i := 0; i < 100; i++ {
                if foo() {  // line 20
                        println("not 0")
                }
                time.Sleep(1)
        }
        fmt.Println("end")
}
```

To show that all of it is atomic:
```
% go tool objdump foo | grep -E 'foo.go:(11|17|20)'
  foo.go:11             0x47e48b                48833d2d690d0000        CMPQ $0x0, main.i(SB)
  foo.go:20             0x47e493                74d5                    JE 0x47e46a
  foo.go:17             0x47e520                48c70595680d0001000000  MOVQ $0x1, main.i(SB)
```

Not sure if this is atomic enough, but I will ask in Go slack because now I am interested in this :)

szuecs: @AlexanderYastrebov there was an interesting answer that in x86 (the same seems true for amd64) sync.Load is implemented for int/int32/int64 https://go.godbolt.org/z/WsToKnxrs as a single MOVQ instruction so the result seems to be consistent if we use it or not and should not have any perf benefits if we drop it.
So let's go with your current code which is of course correct. :)

docs/reference/predicates.md
marcinzaremba: I really like the design as well. It seems to be elegant and pragmatic.

aryszka:
:+1:

PR #1903, opened 2021-11-19
Set OIDC cookie maxAge according to claims exp

AndreasSko:
Currently, Skipper hard-codes the maxAge of the OIDC cookie to one hour, even if the expiration time of the claim is longer than that. This results in early re-authentication, even though it might not really be necessary yet.

With this change, the exp field is attempted to be parsed. If it's successful, it will use that value as the maxAge for the OIDC cookie.

### DIFF
--- a/filters/auth/oidc.go
+++ b/filters/auth/oidc.go
@@ -418,7 +418,7 @@ func mergerCookies(cookies []http.Cookie) (cookie http.Cookie) {
 	return
 }
 
-func (f *tokenOidcFilter) doDownstreamRedirect(ctx filters.FilterContext, oidcState []byte, redirectUrl string) {
+func (f *tokenOidcFilter) doDownstreamRedirect(ctx filters.FilterContext, oidcState []byte, maxAge time.Duration, redirectUrl string) {
 	log.Debugf("Doing Downstream Redirect to :%s", redirectUrl)
 	r := &http.Response{
 		StatusCode: http.StatusTemporaryRedirect,
@@ -433,7 +433,7 @@ func (f *tokenOidcFilter) doDownstreamRedirect(ctx filters.FilterContext, oidcSt
 		Path:     "/",
 		Secure:   true,
 		HttpOnly: true,
-		MaxAge:   int(f.validity.Seconds()),
+		MaxAge:   int(maxAge.Seconds()),
 		Domain:   extractDomainFromHost(getHost(ctx.Request())),
 	})
 	for _, cookie := range oidcCookies {
@@ -638,7 +638,20 @@ func (f *tokenOidcFilter) callbackEndpoint(ctx filters.FilterContext) {
 		return
 	}
 
-	f.doDownstreamRedirect(ctx, encryptedData, oauthState.RedirectUrl)
+	f.doDownstreamRedirect(ctx, encryptedData, f.getMaxAge(claimsMap), oauthState.RedirectUrl)
+}
+
+func (f *tokenOidcFilter) getMaxAge(claimsMap map[string]interface{}) time.Duration {
+	maxAge := f.validity
+	if exp, ok := claimsMap["exp"].(float64); ok {
+		val := time.Until(time.Unix(int64(exp), 0))
+		if val > time.Minute {
+			maxAge = val
+			log.Debugf("Setting maxAge of OIDC cookie to %s", maxAge)
+		}
+	}
+
+	return maxAge
 }
 
 func (f *tokenOidcFilter) Request(ctx filters.FilterContext) {
--- a/filters/auth/oidc_test.go
+++ b/filters/auth/oidc_test.go
@@ -224,7 +224,7 @@ func createOIDCServer(cb, client, clientsecret string) *httptest.Server {
 					"sub":   testSub,
 					"aud":   validClient,
 					"iat":   time.Now().Add(-time.Minute).UTC().Unix(),
-					"exp":   time.Now().Add(time.Hour).UTC().Unix(),
+					"exp":   time.Now().Add(2 * time.Hour).UTC().Unix(),
 					"groups": []string{
 						"CD-Administrators",
 						"Purchasing-Department",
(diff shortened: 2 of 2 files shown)


filters/auth/oidc.go
```diff
@@ -763,6 +763,12 @@ func (f *tokenOidcFilter) tokenClaims(ctx filters.FilterContext, oauth2Token *oa
 		return nil, "", requestErrorf("claims do not contain sub")
 	}
 
+	exp, ok := tokenMap["exp"].(float64)
+	if ok {
+		f.validity = time.Until(time.Unix(int64(exp), 0))
```
universam1, line 768: I think we should check here also if the duration is valid, say if > time.Minute

szuecs: I agree otherwise we would DDoS the IAM endpoint if we have a 0 for example  due to some bug

szuecs:
lgtm, beside the comment above.

filters/auth/oidc.go
AndreasSko: Good point! Adjusted :slightly_smiling_face:

filters/auth/oidc.go
```diff
@@ -763,6 +763,15 @@ func (f *tokenOidcFilter) tokenClaims(ctx filters.FilterContext, oauth2Token *oa
...
 
+	exp, ok := tokenMap["exp"].(float64)
+	if ok {
+		val := time.Until(time.Unix(int64(exp), 0))
+		if val > time.Minute {
+			f.validity = val
+			log.Debugf("Setting validity of oidc filter to %s", f.validity)
+		}
+	}
+
```
AlexanderYastrebov, line 774: This does not seem as a correct approach. The filter instance serves multiple concurrent requests so it should be immutable in this case. 

The better approach seems to be to pass validity/max age `max age = claims[exp] or f.validity` to `doDownstreamRedirect` where cookies are created.

AlexanderYastrebov: Another thing to verify is whether `claims["exp"]` is always unmarshaled as `float64` otherwise we may need a bit more sophisticated time conversion logic.

AlexanderYastrebov: > is always unmarshaled as float64

I think existing test is enough

filters/auth/oidc_test.go
```diff
@@ -825,6 +825,12 @@ func TestOIDCSetup(t *testing.T) {
 			}
 			bs := string(b)
 			t.Logf("Got body: %s", bs)
+
+			// make sure cookies have correct maxAge according to exp
+			for _, c := range client.Jar.Cookies(reqURL) {
+				maxAge := time.Duration(c.MaxAge * int(time.Second)).Round(time.Hour)
```
AlexanderYastrebov, line 831: Lets round to a smaller value, e.g. 1 minute.

The cast to int does not feel safe/canonical, I think it should be `time.Duration(c.MaxAge) * time.Second` as per https://pkg.go.dev/time#Duration
> To convert an integer number of units to a Duration, multiply:

    seconds := 10
    fmt.Print(time.Duration(seconds)*time.Second) // prints 10s

AlexanderYastrebov: I think we should rather check that MaxAge is within `[2h - delta(e.g. 1 minute), 2h]` interval instead of rounding.

AndreasSko: Yeah, that makes sense. What do you think about:

```suggestion
				maxAge := time.Duration(c.MaxAge) * time.Second
				assert.InDelta(t, 2*time.Hour, maxAge, float64(time.Minute))
```
(though here I don't really like the conversion to float64)

AlexanderYastrebov: InDelta is not really correct as it must not be above 2h, maybe:
```go
	maxAge := time.Duration(10) * time.Second
	if maxAge < 2*time.Hour-time.Minute || maxAge > 2*time.Hour {

	}
```

AndreasSko:
@AlexanderYastrebov thank you for your review! I adjusted the PR accordingly. What do you think? :slightly_smiling_face:

filters/auth/oidc.go
```diff
@@ -638,7 +638,16 @@ func (f *tokenOidcFilter) callbackEndpoint(ctx filters.FilterContext) {
 		return
 	}
 
-	f.doDownstreamRedirect(ctx, encryptedData, oauthState.RedirectUrl)
+	maxAge := f.validity
+	if exp, ok := claimsMap["exp"].(float64); ok {
+		val := time.Until(time.Unix(int64(exp), 0))
+		if val > time.Minute {
+			maxAge = val
+			log.Debugf("Setting maxAge of OIDC cookie to %s", f.validity)
```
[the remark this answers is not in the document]
AlexanderYastrebov: Could we move this snippet into a `f.getMaxAge(claimsMap)` method so that whole code flow reads cleaner?

filters/auth/oidc_test.go
```diff
@@ -825,6 +825,13 @@ func TestOIDCSetup(t *testing.T) {
 			}
 			bs := string(b)
 			t.Logf("Got body: %s", bs)
+
+			// make sure cookies have correct maxAge according to exp
+			for _, c := range client.Jar.Cookies(reqURL) {
+				maxAge := time.Duration(c.MaxAge) * time.Second
+				assert.False(t, maxAge < 2*time.Hour-time.Minute || maxAge > 2*time.Hour,
```
AlexanderYastrebov, line 832: nit: I think it is better to revert condition like
```go
assert.True(t, maxAge > 2*time.Hour-time.Minute && maxAge < 2*time.Hour, "maxAge expected to be two hours")
```

AndreasSko:
@AlexanderYastrebov thank you again, you had some good points! :slightly_smiling_face: 
Adjusted it and put the logic into its own method `getMaxAge` (which is also unit tested now). Could you have another look again? :slightly_smiling_face:

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
@AndreasSko Thank you for the improvement
@szuecs Please take a look

szuecs:
:+1:

szuecs:
Thanks!

PR #1905, opened 2021-11-22
Adds `HeaderSHA256` predicate

AlexanderYastrebov:
`HeaderSHA256` predicate matches SHA-256 hash of the configured header value.

```
$ go test ./predicates/auth/ -run NONE -bench BenchmarkHeaderSHA256Match -benchmem
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/predicates/auth
BenchmarkHeaderSHA256Match-8     4190247               275.3 ns/op            32 B/op          1 allocs/op
```


docs/reference/predicates.md
```diff
@@ -363,6 +363,53 @@ JWTPayloadAllKVRegexp("iss", "^https://")
...
 
+### HeaderSHA256
+
+Matches if SHA-256 hash of the header value (secret) equals to any of the configured hash values.
+Several hash values could be used to match multiple secrets e.g. during secret rotation.
+
+Hash values only hide secrets from parties that have access to the source of Skipper routes.
+Authentication strength depends on the strength of the secret value so e.g.
+`HeaderSHA256("X-Secret", "2bb80d537b1da3e38bd30361aa855686bde0eacd7162fef6a25fe97bf527a25b")`
+is not stronger than just `Header("X-Secret", "secret")`.
```
AlexanderYastrebov, line 374: Add salt

szuecs:
What's the use case?
It sounds more like an `auth` filter to me. A predicate will slow down all route lookoups, that's why we do not validate `Jwt*` predicates. 
I think you want to do something independent of the chosen algorithm. What if SHA256 is broken and you need to use $someOtheHashFunction? I would suggest passing the chosen hash function as parameter to the filter.
On the other hand what's the difference between this and basic auth? Maybe better to make a better basic auth filter, that do not need a file but can have the salted,hashed secret as filter parameter?

AlexanderYastrebov:
> What's the use case?

Match routes with known header value without revealing the actual value to the parties that have access to the routes source (source code, ingress definitions, allowed to query `/routes` endpoint)

>  A predicate will slow down all route lookoups

The benchmark shows that it takes 340 ns to match header hash.

> I think you want to do something independent of the chosen algorithm. What if SHA256 is broken and you need to use $someOtheHashFunction? I would suggest passing the chosen hash function as parameter to the filter.

If that happens we will introduce `HeaderSomeOtheHashFunction` and discourage use of `HeaderSHA256`.

> On the other hand what's the difference between this and basic auth? Maybe better to make a better basic auth filter, that do not need a file but can have the salted,hashed secret as filter parameter?

This predicate allows users to implement Basic Authentication as well as shown in the docs without revealing the username:password pair.

docs/reference/predicates.md
```diff
@@ -363,6 +363,53 @@ JWTPayloadAllKVRegexp("iss", "^https://")
 JWTPayloadAnyKVRegexp("iss", "^https://")
 ```
 
+### HeaderSHA256
+
+Matches if SHA-256 hash of the header value (secret) equals to any of the configured hash values.
+Several hash values could be used to match multiple secrets e.g. during secret rotation.
```
AlexanderYastrebov, line 369: The same could be achieved using two routes with two different `HeaderSHA256` but it is more convenient to use single predicate with multiple secrets to avoid route duplication, e.g. in `zalando.org/skipper-predicate` ingress annotation.

docs/reference/predicates.md
AlexanderYastrebov: Added guide on how to generate secure header value. 
I think salt is not required if secret value is random, its length is at least 32 bytes (SHA-256 output size) and there is no need to hide the possible reuse of the same secret value in several routes.

In general proper salt value has its own requirements so the idea is to get rid of it if possible to minimize number of parameters that users need to specify and understand.

predicates/auth/headersha256.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (p *headerSha256Predicate) Match(r *http.Request) bool {
+	value := r.Header.Get(p.name)
+	if value == "" {
+		return false
+	}
+
+	s := sha256.Sum256([]byte(value))
+	valueHash := s[:]
```
szuecs, line 67: why do we need to copy here?
I guess `[][]byte` iterate so you compare `[]byte` with `[Size]byte`, but maybe we can create `[][Size]`, `Size` is defined by `sha256.Size` so should be possible.
Do I miss something?

AlexanderYastrebov: its not a copy, it creates slice from array returned by sha256.Sum256

AlexanderYastrebov: > but maybe we can create [][Size]

but then we would need to convert arrays to slices anyway to compare :) https://stackoverflow.com/questions/18561219/comparing-arrays-in-go-language

szuecs:
:+1:

predicates/auth/headersha256.go
szuecs: whatever you think is better. I think the preparation should always be the heavy lifting if possible and the hot path as few lines as possible, so I tend to the change, but if it's the same in benchmark results, then whatever you prefer seems fine

szuecs:
:+1:

PR #1906, opened 2021-11-23
Optimizes loadbalanced backend string

AlexanderYastrebov:
```
benchmark                      old ns/op     new ns/op     delta
BenchmarkLBBackendString-8     9714          1951          -79.92%
BenchmarkLBBackendString-8     8967          1960          -78.14%
BenchmarkLBBackendString-8     8975          1955          -78.22%
BenchmarkLBBackendString-8     9176          1944          -78.81%
BenchmarkLBBackendString-8     8938          1947          -78.22%
BenchmarkLBBackendString-8     8733          1953          -77.64%
BenchmarkLBBackendString-8     8690          1960          -77.45%
BenchmarkLBBackendString-8     8519          1960          -76.99%
BenchmarkLBBackendString-8     9161          1952          -78.69%
BenchmarkLBBackendString-8     9048          1814          -79.95%

benchmark                      old allocs     new allocs     delta
BenchmarkLBBackendString-8     114            12             -89.47%
BenchmarkLBBackendString-8     114            12             -89.47%
BenchmarkLBBackendString-8     114            12             -89.47%
BenchmarkLBBackendString-8     114            12             -89.47%
BenchmarkLBBackendString-8     114            12             -89.47%
BenchmarkLBBackendString-8     114            12             -89.47%
BenchmarkLBBackendString-8     114            12             -89.47%
BenchmarkLBBackendString-8     114            12             -89.47%
BenchmarkLBBackendString-8     114            12             -89.47%
BenchmarkLBBackendString-8     114            12             -89.47%

benchmark                      old bytes     new bytes     delta
BenchmarkLBBackendString-8     7956          4648          -41.58%
BenchmarkLBBackendString-8     7957          4648          -41.59%
BenchmarkLBBackendString-8     7956          4648          -41.58%
BenchmarkLBBackendString-8     7956          4648          -41.58%
BenchmarkLBBackendString-8     7957          4648          -41.59%
BenchmarkLBBackendString-8     7957          4648          -41.59%
BenchmarkLBBackendString-8     7957          4648          -41.59%
BenchmarkLBBackendString-8     7957          4648          -41.59%
BenchmarkLBBackendString-8     7956          4648          -41.58%
BenchmarkLBBackendString-8     7956          4648          -41.58%
```

Currently Skipper uses `Route.String()` in the hot-path https://github.com/zalando/skipper/blob/1e26fcd250309d05da5a7b086c784ba100cf07da/proxy/proxy.go#L915
[...]

PR #1907, opened 2021-11-23
Removes `skipper.route` tag from proxy span

AlexanderYastrebov:
`skipper.route` tag contains serialized route in eskip format.
The value could be quite large, especially when route has loadbalanced
backend with many endpoints.

For payload size efficiency reasons this tag was already excluded at Zalando,
see https://github.com/zalando-incubator/kubernetes-on-aws/pull/2173

During profiling it was also dicovered that route eskip serialization
performance is suboptimal and may take more than 7% of the total
`proxy.ServeHTTP` call runtime.
Screenshot from 2021-11-23 21-58-35

While it is possible to optimize route eskip serialization (see e.g.
https://github.com/zalando/skipper/pull/1906) this change simply drops
it from the request path.

Users may still rely on `skipper.route_id` tag to resolve route configuration and
`http.url` tag to get actual backend address.


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #1908, opened 2021-11-24
Adds predicate and filter name to error

AlexanderYastrebov:
```
$ bin/skipper -inline-routes='r1: QueryParam() -> <shunt>; r2: Unknown() -> <shunt>; r3: Path("/r3") -> unknown() -> <shunt>; r4: Path("/r4") -> setPath() -> <shunt>;'
...
[APP]ERRO[0000] failed to process route r1: failed to create predicate "QueryParam": invalid predicate parameters
[APP]ERRO[0000] failed to process route r2: predicate "Unknown" not found
[APP]ERRO[0000] failed to process route r3: filter "unknown" not found
[APP]ERRO[0000] failed to process route r4: failed to create filter "setPath": invalid filter parameters
```

### DIFF
--- a/routing/datasource_test.go
+++ b/routing/datasource_test.go
@@ -1,12 +1,16 @@
-package routing
+package routing_test
 
 import (
 	"testing"
 	"time"
 
 	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/filters/builtin"
 	"github.com/zalando/skipper/logging"
 	"github.com/zalando/skipper/logging/loggingtest"
+	"github.com/zalando/skipper/predicates/primitive"
+	"github.com/zalando/skipper/predicates/query"
+	"github.com/zalando/skipper/routing"
 	"github.com/zalando/skipper/routing/testdataclient"
 )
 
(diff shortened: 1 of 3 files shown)


routing/datasource_test.go
```diff
@@ -1,12 +1,16 @@
-package routing
+package routing_test
```
AlexanderYastrebov, line 1: To use exiting predicates test has to be in a different package. To test package private `processRouteDef` it is exported for tests similar to go stdlib, see `export_test.go`

szuecs: you can also add your new test in a separate `foo_test.go` file with the package change.

AlexanderYastrebov: There was already a test `TestErrorWhenUsingPrediateAsFilter` for error messages, to extend it I needed to have another predicate and filter. Separate `foo_test.go` with different package would require `export_test.go` anyway unless it would test different method.

szuecs: ok

szuecs:
:+1:

AlexanderYastrebov:
:+1:

Issue #386, opened 2017-08-02
provide possibility to set exact, non-regexp host predicates


aryszka:
it's a typical use case to match every route by the request host header, but the host values don't show a high variance, and so it is not very effective to execute a regular expression for every route to be matched. It could improve the lookup performance in many scenarios if there was an exact Host matching predicate, e.g:

```
HostExact("www.example.org")
```

On the other hand, it is also a typical use case to match different TLDs, e.g. country domains, and in our experience, the regexp matching of the host header is used primarily for this. Question: if we implemented an exact host predicate, what would be the best way to allow multiple hosts for the same route? How about:

```
HostExact("www.example.org", "www.example.net", "www.example.de", "www.example.fr")
```

?

@mo-gr, @lmineiro  do you think the above example would be a too strong precursor to "OR" predicates?

lmineiro:
The difference I anticipate with the proposal versus having the "OR" as part of the predicates matching is that we'd have to support expression grouping for more complex "predicate trains", like:

`HostExact("example.com") && Path("/foo") OR Path("/bar")`
vs
`HostExact("example.com") && (Path("/foo") OR Path("/bar"))`

This would become mandatory if the "OR" is introduced.

The variadic host matching for this new Predicate seems reasonable

mo-gr:
Maybe we should consider adding a custom predicate for our main use case? This would heavily reduce the amount of repetitive boilerplate and potential for stupid typos.
I'm thinking something like:

```
Zalando() -> ... -> "tailor"; // this matches all valid (www|m).zalando.TLD
ZalandoDesktop() -> ... -> "tailor"; // this matches all valid www.zalando.TLD
ZalandoMobile() -> ... -> "tailor"; // this matches all valid m.zalando.TLD
Zalando('de', 'ch', 'ch-fr') -> ... -> "tailor"; // this matches only (www|m).zalando.de and (www|m|fr|m-fr).zalando.ch
```

Internally, those could be implemented more efficient than a regular expression and it would cover all our use cases.

But apart from that, having a `HostExact` is probably a good idea. Having an `OR` is something, I would strongly object against.

aryszka:
the Zalando() predicates are definitely a z-skipper matter, but it's also useful insight for others and a good use case for the custom predicates, so we can give documentation hints.

grassator:
I think we should normalize the names of standard predicates as I think `Host` is the only one that takes Regexp. To maintain backwards compatibility I suggest adopting the following scheme:

[...]

PR #1911, opened 2021-11-29
Adds incoming request host normalization

AlexanderYastrebov:
Incoming request `Host` header may contain port number, a trailing dot or
have an uppercase value which are all valid according to various RFCs but are not common.

AWS Application Load Balancer (ALB) removes port number, trailing dot and
converts `Host` header to lower case so uncommon cases are never observed
by Skipper behind ALB.
AWS Network Load Balancer (NLB) passes `Host` header as is.

Several changes were introduced to account for uncommon port number and traling dot:
* `Host` predicate generated by kubernetes dataclient was changed to
match optional port and trailing dot by #1281 and #1863
* `rfcHost` filter was introduced to remove trailing dot #1864

Matching optional port and trailing dot makes `Host` regexp more complex and
adds a small performance penalty for the uncommon case.
Uppercase `Host` values are not supported at all.

This change introduces host normalization step that converts incoming request host
to lowercase and removes port and trailing dot if any before routing,
enabled by the new `-normalize-host` or by `-kubernetes` flags.

It allows `Host` regexp simplification and use of a prospective predicate that
compares strings instead of using regexp (see #386).

Updates #1281, #1863, #1864, #386

### DIFF
--- a/net/host_test.go
+++ b/net/host_test.go
@@ -1,0 +1,125 @@
+package net
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+)
+
+func TestHostPatch(t *testing.T) {
+	for config, cases := range map[HostPatch]map[string]string{
+		{}: {
+			"":                  "",
+			"example.org":       "example.org",
+			"example.org.":      "example.org.",
+			"example.org:8080":  "example.org:8080",
+			"example.org.:8080": "example.org.:8080",
+			"127.0.0.1":         "127.0.0.1",
+			"127.0.0.1:9090":    "127.0.0.1:9090",
+			"::1":               "::1",
+			"[::1]:9090":        "[::1]:9090",
+			"EXAMPLE.ORG":       "EXAMPLE.ORG",
+			"EXAMPLE.ORG:8080":  "EXAMPLE.ORG:8080",
+			"EXAMPLE.ORG.:8080": "EXAMPLE.ORG.:8080",
+		},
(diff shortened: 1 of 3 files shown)


net/host_test.go
```diff
@@ -0,0 +1,125 @@
...
+func BenchmarkHostPatchUncommon(b *testing.B) {
+	hp := HostPatch{RemovePort: true, RemoteTrailingDot: true, ToLower: true}
+	if hp.Apply("WWW.EXAMPLE.ORG.:8080") != "www.example.org" {
+		b.Fatal("expected www.example.org")
+	}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		hp.Apply("WWW.EXAMPLE.ORG.:8080")
+	}
+}
```
AlexanderYastrebov, line 125:
```
$ go test ./net/ -run NONE -bench BenchmarkHostPatch -benchtime=10s -benchmem
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/net
BenchmarkHostPatchCommon-8      500292288               23.72 ns/op            0 B/op          0 allocs/op
BenchmarkHostPatchUncommon-8    154102808               77.76 ns/op           16 B/op          1 allocs/op
PASS
ok      github.com/zalando/skipper/net  34.157s
```

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #1912, opened 2021-11-29
Adds `HostAny` predicate

### LINKED ISSUE #386: provide possibility to set exact, non-regexp host predicates

AlexanderYastrebov:
`HostAny` predicate evaluates to true if request host exactly matches any of the configured values.

Benchmark demonstrates performance of mathching single host across 10k routes:
```
$ go test ./routing/ -run NONE -bench BenchmarkPredicateHost -benchmem -benchtime=10s
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/routing
BenchmarkPredicateHostRegexpSingleWithDotAndPort-8          4429           2749788 ns/op              48 B/op          1 allocs/op
BenchmarkPredicateHostRegexpMultipleWithDotAndPort-8        4633           2619966 ns/op              48 B/op          1 allocs/op
BenchmarkPredicateHostRegexpSingleDomainOnly-8              8666           1342919 ns/op              48 B/op          1 allocs/op
BenchmarkPredicateHostRegexpMultipleDomainOnly-8            4980           2450593 ns/op              48 B/op          1 allocs/op
BenchmarkPredicateHostAnySingle-8                          36828            331182 ns/op              48 B/op          1 allocs/op
BenchmarkPredicateHostAnyMultiple-8                        28540            380094 ns/op              48 B/op          1 allocs/op
PASS
ok      github.com/zalando/skipper/routing      82.859s
```

Fixes #386

### DIFF
--- a/routing/export_test.go
+++ b/routing/export_test.go
@@ -1,5 +1,7 @@
 package routing
 
 var (
-	ProcessRouteDef = processRouteDef
+	ExportProcessRouteDef = processRouteDef
+	ExportNewMatcher      = newMatcher
+	ExportMatch           = (*matcher).match
 )
--- a/docs/reference/predicates.md
+++ b/docs/reference/predicates.md
@@ -128,6 +128,21 @@ Host(/^my-host-header\.example\.org$/)
 Host(/header\.example\.org$/)
 ```
 
+## HostAny
+
+Evaluates to true if request host exactly equals to any of the configured hostnames.
+
+Parameters:
+
+* hostnames (string)
+
+Examples:
+
+```
+HostAny("www.example.org", "www.example.com")
+HostAny("localhost:9090")
+```
+
 ## Forwarded header predicates
 
 Uses standardized Forwarded header ([RFC 7239](https://tools.ietf.org/html/rfc7239))
--- a/predicates/host/any.go
+++ b/predicates/host/any.go
@@ -1,0 +1,49 @@
+package host
+
+import (
+	"net/http"
+
+	"github.com/zalando/skipper/predicates"
+	"github.com/zalando/skipper/routing"
+)
+
+type anySpec struct{}
+
+type anyPredicate struct {
+	hosts []string
+}
+
+// NewAny creates a predicate specification, whose instances match request host.
+//
+// The HostAny predicate requires one or more string hostnames and matches if request host
+// exactly equals to any of the hostnames.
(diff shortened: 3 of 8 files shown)


routing/export_test.go
```diff
@@ -1,5 +1,7 @@
 package routing
 
 var (
-	ProcessRouteDef = processRouteDef
+	ExportProcessRouteDef = processRouteDef
+	ExportNewMatcher      = newMatcher
+	ExportMatch           = (*matcher).match
```
szuecs, line 6: I guess this change is fine

szuecs:
:+1:

routing/export_test.go
AlexanderYastrebov: This is need for tests that use `github.com/zalando/skipper/predicates/*`: if such test resides in the `routing` package it fails to import  `github.com/zalando/skipper/predicates/*` due to `imports github.com/zalando/skipper/routing: import cycle not allowed` error (because `github.com/zalando/skipper/predicates/*` imports `github.com/zalando/skipper/routing` which leads to an import cycle). If it resides in the `routing_test` packed it can not access non-exported symbols from `routing`.

The `export_test.go` is a trick from golang stdlib to export unexported symbols for the tests when test resides in another package. The convention seems to be to prefix it with `Export` and they are only visible in tests. For this PR benchmarks I need to use unexported `matcher` because I need to control the order of route list (datasource stores routes in map by id and therefore route list order of the derived matcher is not defined) and `HostAny` predicate from `github.com/zalando/skipper/predicates/host` that imports `github.com/zalando/skipper/routing`.

szuecs: Thanks for the clarification

AlexanderYastrebov:
:+1:

PR #1918, opened 2021-11-30
dataclients/kubernetes: use healthcheck based on Shutdown predicate

AlexanderYastrebov:
Updates #1899

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -5,16 +5,14 @@ import (
 	"net"
 	"net/http"
 	"os"
-	"os/signal"
 	"regexp"
 	"strings"
-	"syscall"
+	"text/template"
 	"time"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/filters"
-	"github.com/zalando/skipper/predicates"
 )
 
 const (
@@ -22,9 +20,7 @@ const (
 	defaultRouteGroupClass       = "skipper"
 	serviceHostEnvVar            = "KUBERNETES_SERVICE_HOST"
 	servicePortEnvVar            = "KUBERNETES_SERVICE_PORT"
-	healthcheckRouteID           = "kube__healthz"
 	httpRedirectRouteID          = "kube__redirect"
-	healthcheckPath              = "/kube-system/healthz"
 	defaultLoadBalancerAlgorithm = "roundRobin"
 	defaultEastWestDomain        = "skipper.cluster.local"
 )
@@ -192,13 +188,10 @@ type Client struct {
 	ingress                *ingress
 	routeGroups            *routeGroups
 	provideHealthcheck     bool
-	healthy                bool
 	provideHTTPSRedirect   bool
-	termReceived           bool
 	reverseSourcePredicate bool
 	httpsRedirectCode      int
 	current                map[string]*eskip.Route
-	sigs                   chan os.Signal
 	quit                   chan struct{}
 	defaultFiltersDir      string
 }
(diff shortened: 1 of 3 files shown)


dataclients/kubernetes/kube.go
```diff
@@ -470,7 +446,7 @@ func (c *Client) LoadUpdate() ([]*eskip.Route, []string, error) {
 		// TODO: use eskip.Eq()
 		if r, ok := next[id]; ok && r.String() != c.current[id].String() {
 			updatedRoutes = append(updatedRoutes, r)
-		} else if !ok && id != healthcheckRouteID && id != httpRedirectRouteID {
+		} else if !ok {
```
AlexanderYastrebov, line 449: Since healthcheck and globalredirect routes are all static we can move them into load `loadAndConvert()` and simplify update diff logic.

dataclients/kubernetes/kube.go
```diff
@@ -374,53 +369,42 @@ func shuntRoute(r *eskip.Route) {
...
-	if healthy && !log.IsLevelEnabled(log.DebugLevel) {
-		logFilters = append(logFilters, &eskip.Filter{
-			Name: filters.DisableAccessLogName,
-			Args: []interface{}{200},
-		})
-	}
+func healthcheckRoutes(reverseSourcePredicate bool) []*eskip.Route {
+	template := template.Must(template.New("healthcheck").Parse(`
+		kube__healthz_up:   Path("/kube-system/healthz") && {{.Source}}({{.SourceCIDRs}}) -> {{.LogFilter}} status(200) -> <shunt>;
+		kube__healthz_down: Path("/kube-system/healthz") && {{.Source}}({{.SourceCIDRs}}) && Shutdown() ->  status(503) -> <shunt>;
```
szuecs, line 375: Why you have no Logfilter in the shutdown route?

dataclients/kubernetes/update_test.go
```diff
@@ -25,9 +27,7 @@ func TestUpdateOnlyChangedRoutes(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if len(r) != 1 || r[0].Id != healthcheckRouteID {
-		t.Fatal("no healthcheck route received")
-	}
+	assert.EqualValues(t, healthcheckRoutes(false), r, "no healthcheck routes received")
```
szuecs, line 30: For one check add a dependency?
I would rather like to remove all the testify dependencies.

dataclients/kubernetes/kube.go
AlexanderYastrebov: I guess because we want to see healthcheck logs during shutdown.
This re-implements what we had before, the `disableAccessLog(200)` was introduced by #1650

szuecs: Right now, I can observe the shutdown health checks by ALBs or NLB control plane. I don't really care about the "up" message in healthchecks. For me the more interesting logs are the "down" messages.

dataclients/kubernetes/update_test.go
szuecs: fine for me if it prints the nested things as diff well.
I just don't like the helper library, because you can write a little more code that is understandable without it.

dataclients/kubernetes/kube.go
AlexanderYastrebov: Yes, this mutes access log for "up" unless debug level is enabled.

AlexanderYastrebov: I have renamed `LogFilter` to `DIsableAccessLog` to make it more clear

szuecs:
:+1:

AlexanderYastrebov:
:+1:

dataclients/kubernetes/kube.go
AlexanderYastrebov: Special handling of healthcheck route had caused problems in the past.
Previously healthcheck route was added to `updatedRoutes` after `loadAndConvert()` and until the #618 it happened on every poll cycle which caused constant routing table recreation https://github.com/zalando/skipper/issues/614#issuecomment-381357947

Static healthcheck and globalredirect routes means they do not change after initial `LoadAll()` and therefore will not end up neither in `updatedRoutes` nor in `deletedIDs`. They are also treated just like normal routes which simplifies the overall code flow.

PR #1920, opened 2021-12-02
ratelimit: refactor AllowContext

AlexanderYastrebov:
* moves metric collection outside of ratelimit logic
* uses single tracing span for allow check
* fails early on Redis error
* updates tests to verify result of each `Allow` call

### DIFF
--- a/ratelimit/redis.go
+++ b/ratelimit/redis.go
@@ -31,11 +31,8 @@ const (
 	allowMetricsFormatWithGroup      = redisMetricsPrefix + "query.allow.%s.%s"
 	retryAfterMetricsFormatWithGroup = redisMetricsPrefix + "query.retryafter.%s.%s"
 
-	allowAddSpanName           = "redis_allow_add_card"
-	allowExpireSpanName        = "redis_allow_expire"
-	allowCheckSpanName         = "redis_allow_check_card"
-	allowCheckRemRangeSpanName = "redis_allow_check_rem_range"
-	oldestScoreSpanName        = "redis_oldest_score"
+	allowSpanName       = "redis_allow"
+	oldestScoreSpanName = "redis_oldest_score"
 )
 
 // newClusterRateLimiterRedis creates a new clusterLimitRedis for given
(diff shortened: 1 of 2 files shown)


ratelimit/redis.go
```diff
@@ -120,78 +117,70 @@ func (c *clusterLimitRedis) startSpan(ctx context.Context, spanName string) func
...
 	clearBefore := now.Add(-c.window).UnixNano()
 
-	count, err := c.allowCheckCard(ctx, key, clearBefore)
+	// drop all elements of the set which occurred before one interval ago.
+	_, err := c.ringClient.ZRemRangeByScore(ctx, key, 0.0, float64(clearBefore))
 	if err != nil {
-		log.Errorf("Failed to get from redis cardinality: %v", err)
-		queryFailure = true
-		// we don't return here, as we still want to record the request with ZAdd, but we mark it as a
-		// failure for the metrics
```
AlexanderYastrebov, line 139: If Redis op failed the subsequent will likely fail as well so it makes not sense to complicate logic here and it is better just to return an error.

ratelimit/redis.go
```diff
@@ -120,78 +117,70 @@ func (c *clusterLimitRedis) startSpan(ctx context.Context, spanName string) func
...
+	now := time.Now()
+	finishSpan := c.startSpan(ctx, allowSpanName)
+
+	allow, err := c.allowContext(ctx, clearText)
+
+	finishSpan(err != nil)
+	queryFailure := err != nil
+	c.measureQuery(allowMetricsFormat, allowMetricsFormatWithGroup, &queryFailure, now)
+
+	if err != nil {
```
szuecs, line 130: not sure if this makes sense in details, but what if we would use `queryFailure` in all of the `err != nil` use

ratelimit/redis.go
```diff
@@ -120,78 +117,70 @@ func (c *clusterLimitRedis) startSpan(ctx context.Context, spanName string) func
...
-		queryFailure = true
-		// we don't return here, as we still want to record the request with ZAdd, but we mark it as a
-		// failure for the metrics
+		return false, err
+	}
+
+	// get cardinality
+	count, err := c.ringClient.ZCard(ctx, key)
+	if err != nil {
+		return false, err
```
szuecs, line 158: most of `return false, err` are changing the behavior of the function from fail to open to fail to closed. As far as I see without impact, because you change the decision to fail open on the caller function, but from reading point of view that can be miss leading, because you read a `allowContect` function with return `allow, error`.

AlexanderYastrebov: It has to return either true or false for the value along with an error. The fail open decision is explicitly made on the caller side and this is non-exported method. I could rewrite it to use return values like
```go
func (c *clusterLimitRedis) allowContext(ctx context.Context, clearText string) (allow bool, err error) {
...
	count, err := c.ringClient.ZCard(ctx, key)
	if err != nil {
		return
	}
...
```
but it does not really change much

ratelimit/redis.go
AlexanderYastrebov: Looks better indeed, fixed.

szuecs:
from my side lgtm.
@aryszka do you want to check if you agree with reducing the number of Spans?

ratelimit/redis.go
aryszka: This sounds reasonable, however, I don't know how ZRemRangeByScore behaves when a key is missing, e.g. when it's a new key or the selected redis instance has just restarted. If that happens, then we would never call ZAdd

ratelimit/redis.go
```diff
@@ -120,78 +117,70 @@ func (c *clusterLimitRedis) startSpan(ctx context.Context, spanName string) func
...
-	s := getHashedKey(clearText)
 	c.metrics.IncCounter(redisMetricsPrefix + "total")
+	now := time.Now()
+	finishSpan := c.startSpan(ctx, allowSpanName)
+
+	allow, err := c.allow(ctx, clearText)
+	failed := err != nil
+
+	finishSpan(failed)
+	c.measureQuery(allowMetricsFormat, allowMetricsFormatWithGroup, &failed, now)
```
aryszka, line 128: with this simplification (dropping the defer), we can also change the *bool argument to plain bool

aryszka:
> from my side lgtm.
> @aryszka do you want to check if you agree with reducing the number of Spans?

I am fine with reducing the spans, they haven't proved to be useful in my observation

ratelimit/redis.go
AlexanderYastrebov: https://redis.io/commands/zremrangebyscore returns the number of elements removed.
It seems to return 0 for non-existing zset, not an error:
```
127.0.0.1:6379> ZREMRANGEBYSCORE doesnotexist -inf 10
(integer) 0
```
Besides we should have failed tests that check denial if we never add a key.

aryszka: :+1:

aryszka:
:+1:

ratelimit/redis.go
AlexanderYastrebov: yes, I kept `measureQuery` as is for another round.

AlexanderYastrebov:
Redis test only checked last `Allow` result after a series of requests, I've changed it slightly to verify result of each `Allow` call.

aryszka:
:+1:

AlexanderYastrebov:
:+1:

PR #1921, opened 2021-12-02
proxy: remove http.remote_addr tracing tag

AlexanderYastrebov:
Removes `http.remote_addr` tag that contains request remote IP and port value.
There is still `http.remote_ip` tag that contains remote IP address without port.

### DIFF
--- a/proxy/tracing.go
+++ b/proxy/tracing.go
@@ -12,12 +12,10 @@ const (
 	HostnameTag           = "hostname"
 	HTTPHostTag           = "http.host"
 	HTTPMethodTag         = "http.method"
-	HTTPRemoteAddrTag     = "http.remote_addr"
 	HTTPRemoteIPTag       = "http.remote_ip"
 	HTTPPathTag           = "http.path"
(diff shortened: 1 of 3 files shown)


proxy/tracing.go
```diff
@@ -12,7 +12,7 @@ const (
 	HostnameTag           = "hostname"
 	HTTPHostTag           = "http.host"
 	HTTPMethodTag         = "http.method"
-	HTTPRemoteAddrTag     = "http.remote_addr"
+	HTTPRemoteAddrTag     = "http.remote_addr" // unused
```
szuecs, line 15: Maybe add a TODO: v1 cleanup or we just remove the const, IMO it's not really "public interface"

AlexanderYastrebov: yep, looks like they were made public unintentionally by #1066

AlexanderYastrebov: Removed this and another one unused since #1907

szuecs:
:+1:

AlexanderYastrebov:
:+1:

Issue #1059, opened 2019-05-20
implement xDS v2 APIs


szuecs:
If we implement xDS APIs, it would be possible to integrate with tools like istio. We could implement [xDS v2 APIs](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol) ([dev docs](https://github.com/envoyproxy/data-plane-api/blob/master/API_OVERVIEW.md)), which were [introduced](https://blog.envoyproxy.io/the-universal-data-plane-api-d15cec7a) by Matt Klein in the envoy proxy from lyft.

This would reverse the current dataclient idea, where skipper polls for route sources. 

Reason: zecke asked if it would makes sense to standardize xDS APIs more, which might makes sense in this context.
https://twitter.com/zecke42/status/1130258605884755969

Currently this issue would not have high priority, please add a comment if this would help you.

PR #1925, opened 2021-12-07
JSON umarshalling, take 2

marcinzaremba:
parsing json into eskip routes

@marcinzaremba @AlexanderYastrebov @szuecs , the current PR contains a json schema, but I propose to drop it for the following reasons:

- it doesn't provide too much value, because...
- validating semantic rules can only be done partially
- such validation may even be harmful, because in theory there can be use cases where processes could pass partially processed routes to each other
- makes symmetric parsing and serializing cumbersome
- maintaining this rule set in json format doesn't seem reasonable to me


szuecs:
👍

szuecs:
We discussed that we want to drop jsonSchema, because right now it might be too unstable (do we do changes we need to keep it in sync). If we observe that we do not change it within a long time like 6 months or 1 year we can create jsonSchema. I guess it would be interesting if we would provide a json API to push route changes to it.
https://github.com/zalando/skipper/issues/1059 is an issue that request to provide xDS API (envoy's "generic" LB API).

aryszka:
:+1:

PR #1927, opened 2021-12-07
Adds OIDC cookie domain tests

### LINKED ISSUE #1926: Enable hostname testing

Skipper tests use `httptest.NewServer` and connect to the address specified by test server URL property which has IP:port format. 
Sometimes it is required to make a request to the hostname to test request host handling. It is possible to set request host header like:
```go
	req, _ := http.NewRequest("GET", ts.URL, nil)
	req.Host = "my.test"
```

But sometimes tested components use internal http client that connects to the mocked backend and the only configurable property is the string backend address.
E.g. in case of https://github.com/zalando/skipper/pull/1923#discussion_r762988313 mock oidc server redirects client to the mocked redirect url. Since redirect url host is in form of IP:port it is not possible to verify OIDC cookie multi-level domain handling (#1909).

It would be great to have a set of test hostnames that resolve to localhost address and thus could be used for testing.

Introduce mock DNS resolver for tests that would resolve configured domains to the localhost address.
The alternative is to modify `/etc/hosts` like
```
127.0.0.1 h0.test
127.0.0.1 h1.test
```
but that is not convenient.

* https://github.com/foxcpp/go-mockdns
* https://github.com/miekg/dns

Yes

AlexanderYastrebov:
Adds utility to configure loopback names in tests and uses it to validate OIDC cookie domain.

Fixes #1926

### DIFF
--- a/net/dnstest/dnstest.go
+++ b/net/dnstest/dnstest.go
@@ -1,0 +1,109 @@
+package dnstest
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"testing"
+	"time"
+
+	"github.com/miekg/dns"
+)
+
+// Replaces net.DefaultResolver with a resolver that resolves
+// configured names to 127.0.0.1 and fails to resolve any other name.
+// Uses t.Cleanup to restore resolver after the test.
+func LoopbackNames(t *testing.T, first string, rest ...string) {
+	mux := dns.NewServeMux()
+	mux.HandleFunc(".", newDnsHandler(append(rest, first)))
+
+	s, err := startServer(mux)
+	if err != nil {
+		t.Fatal(err)
+		return
+	}
+	defaultResolver := net.DefaultResolver
+	net.DefaultResolver = serverResolver(s)
+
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

net/dnstest/dnstest.go
```diff
@@ -0,0 +1,109 @@
+package dnstest
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"testing"
+	"time"
+
+	"github.com/miekg/dns"
```
AlexanderYastrebov, line 10: I've also looked at https://github.com/foxcpp/go-mockdns which is also based on `github.com/miekg/dns` but decided not to pull-in intermediate dependency. I also think it has startup race which I try to avoid using `NotifyStartedFunc` (see `startServer`)

szuecs: That's great Miekg know what he does and interaction was always great.

AlexanderYastrebov:
:+1:

PR #1930, opened 2021-12-10
Propagate 403 status code from auth service in webhook filter

### LINKED ISSUE #1838: Webhook sends back 401 for all codes above 300
**Describe the bug**
I am using webhook filter to redirect the request to an external authorization service which returns 401 and 403 depending on the authorization policy. But the skipper filter handles this and always returns 401 to the client.

https://github.com/zalando/skipper/blob/master/filters/auth/webhook.go#L113

**To Reproduce**
1. Create a custom service for acl based authorization with appropriate return codes ex: 401 and 403
2. Add webhook filter in the ingress to forward the request to the external service for ACL based authorization
3. Invoke the resource API with valid token with insufficient permission for the operation on the resource
4. Test with invalid token -> external service returns 401, skipper filter returns 401
5. Test with valid token but insufficient permission -> external service returns 403, skipper returns 401

**Expected behavior**
Should return the same code as received from the external service.

**Observed behavior**
Always return Unauthorized (401) for all return codes >=300 from the external service.

noggan:
Handle status code 403 from auth service in the `webhook` filter, propagating the 403 status code to the caller.

Fixes #1838

### DIFF
--- a/filters/auth/webhook_test.go
+++ b/filters/auth/webhook_test.go
@@ -18,12 +18,13 @@ const headerToCopy = "X-Copy-Header"
 
 func TestWebhook(t *testing.T) {
 	for _, ti := range []struct {
-		msg         string
-		token       string
-		expected    int
-		authorized  bool
-		timeout     bool
-		copyHeaders bool
+		msg          string
+		token        string
+		expected     int
+		authorized   bool
+		timeout      bool
+		copyHeaders  bool
+		invalidScope bool
 	}{{
 		msg:         "invalid-token-should-be-unauthorized",
 		token:       "invalid-token",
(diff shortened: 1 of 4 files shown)


filters/auth/webhook_test.go
```diff
@@ -73,7 +80,12 @@ func TestWebhook(t *testing.T) {
 				tok = tok[len(authHeaderPrefix):]
 				switch tok {
 				case testToken:
-					w.WriteHeader(200)
+					if ti.invalidScope {
+						w.WriteHeader(http.StatusForbidden)
+						fmt.Fprintln(w, "Forbidden - Got token: "+tok)
+						return
+					}
+					w.WriteHeader(http.StatusOK)
```
AlexanderYastrebov, line 88: Lets add a `testInvalidScopeToken` and a case for it instead of `invalidScope` test case field.
This way we could add other types of failure tokens that result in different responses without extending test case struct.

Also please use `t.Lof` for logging purposes (there is another `fmt.Fprintln` below - it could be fixed as well)

AlexanderYastrebov: The `testWebHookInvalidScopeToken` could be scoped to the `TestWebhook` function for now.

AlexanderYastrebov: > Also please use t.Logf for logging purposes

Nevermind, just saw it writes response body :facepalm:

szuecs: I agree with using a different token without a struct change to check for invalid the status code should be enough.

noggan: Thanks for the input! Updated the PR.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #1933, opened 2021-12-10
Fixes multiflag yaml parsing

AlexanderYastrebov:

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -82,7 +82,7 @@ func Test_NewConfig(t *testing.T) {
 		},
 		{
 			name: "test only valid flag overwrite yaml file",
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

aryszka:
this just came to my mind. When we say "payload" in HTTP terms, that usually refers to the content body, while this is quite the opposite currently: it is only about the address line and the headers. I think we should rename 'payload' to something else. WDYT?

AlexanderYastrebov:
@aryszka The argument could be made other way around: request has URL, headers and body which altogether constitute payload.

aryszka:
in theory that completely sits. What I am talking about is what people usually think what payload is when talking about HTTP. I'll try to suggest a better name

AlexanderYastrebov:
:+1:

aryszka:
:+1:

PR #1936, opened 2021-12-16
add all options from stdlib to our httpclient 

szuecs:
add all options from stdlib to our httpclient net.Client and pass them as is

### DIFF
--- a/net/httpclient.go
+++ b/net/httpclient.go
@@ -133,6 +133,9 @@ func (c *Client) CloseIdleConnections() {
 // which can be nil to get the
 // https://godoc.org/github.com/opentracing/opentracing-go#NoopTracer.
 type Options struct {
+	// Transport see https://golang.org/pkg/net/http/#Transport
+	// In case Transport is not nil, the Transport arguments are used below.
+	Transport *http.Transport
 	// Proxy see https://golang.org/pkg/net/http/#Transport.Proxy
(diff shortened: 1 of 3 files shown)


net/httpclient.go
```diff
@@ -241,6 +267,11 @@ func NewTransport(options Options) *Transport {
 
 	htransport := &http.Transport{
```
AlexanderYastrebov, line 268: In retrospect it looks like it would be better to pass preconfigured `http.Transport` in the `Options` instead of duplication all its knobs...

szuecs: yes could be also a solution. Maybe I should create another NewClient and pass a http.Transport to support this and deprecate the other. wdyt?

AlexanderYastrebov: Or maybe lets have `*http.Transport` in the options such that when it is not nil it is used (otherwise new is created like before)?

AlexanderYastrebov: I.e. we can keep/deprecate existing options like `DisableKeepAlives` but if users need to configure e.g. DialContext they should provide a pre-configured transport

szuecs: makes sense!

szuecs: I changed the PR as you suggested

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #1938, opened 2021-12-21
Fixes eskip backslash escaping

AlexanderYastrebov:
The #1341 removed backslash escaping which leads to incorrect
serialization of strings with a backslash.

### DIFF
--- a/eskip/roundtrip_test.go
+++ b/eskip/roundtrip_test.go
@@ -1,0 +1,124 @@
+package eskip
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Test various strings are parsed to the same value after
+// serialization into eskip double-quoted string
+func TestRoundtripString(t *testing.T) {
+	for i, value := range []string{
+		`s`,
+		` `,
+		`/`,
+		`//`,
+		`\`,
+		`\\`,
+		`\x`,
+		`\\x`,
+		`/x`,
+		`//x`,
+		`"`,
+		`""`,
+		`\/`,
+		`\//`,
+		`/\`,
+		`/\\`,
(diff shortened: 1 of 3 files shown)


szuecs:
Do you think you can add a new full roundtrip test:
eskip text -> routes -> output routes -> parse succesfully without error

AlexanderYastrebov:
> Do you think you can add a new full roundtrip test

In this PR I focus on backslash escaping fix.

The roundtrip is not always possible, e.g. `PathRegexp("_\"_")` argument would be parsed to `_"_` but [serialized](https://github.com/zalando/skipper/blob/8444aeb54837f5ecfde1a6009d70a984ce920194/eskip/string.go#L92) to `PathRegexp(/_"_/)`. 
Similar `Path("\/")` argument would be parsed to `\/`(see #562) but serialized to `Path("\\/")`, i.e. both `Path("\/")` and `Path("\\/")` are valid and parse to the same argument `\/`.
Or due to #578 `PathRegexp(/[/]/)` argument would be parsed to `[/]` but serialized to `PathRegexp(/[\/]/)`, i.e. again both `PathRegexp(/[/]/)` and `PathRegexp(/[\/]/)` are valid and parse to the same argument `[/]`.

In the context of `routesrv` https://github.com/zalando/skipper/pull/1848 it is important that string arguments in memory have the same value in skipper as in routesrv, i.e. `string s -> eskip.Print -> eskip.Parse -> string t` chain leads to the `s == t`.

szuecs:
@AlexanderYastrebov makes sense, but then it also makes sense to add a test, which shows `string s -> eskip.Print -> eskip.Parse -> string t`, no?

AlexanderYastrebov:
Added print->parse roundtrip tests

szuecs:
:+1:

aryszka:
:+1:

PR #1940, opened 2021-12-23
remove logging in makeBackendRequest

szuecs:
remove logging in makeBackendRequest to log it once in one place instead, such that we have no double logging anymore

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1114,7 +1114,9 @@ func (p *Proxy) do(ctx *context) error {
 		backendStart := time.Now()
 		rsp, perr := p.makeBackendRequest(ctx, backendContext)
 		if perr != nil {
-			p.log.Errorf("Failed to do backend request: %v", perr)
+			if perr.code != 499 { // 499 is logged later with more context
+				p.log.Errorf("Failed to do backend request: %v", perr)
+			}
 			if done != nil {
 				done(false)
 			}
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -950,7 +950,7 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 				status = http.StatusServiceUnavailable
 			}
 			p.tracing.setTag(ctx.proxySpan, HTTPStatusCodeTag, uint16(status))
-			return nil, &proxyError{err: fmt.Errorf("net.Error during backend roundtrip to %s: timeout=%v temporary=%v: %w", req.URL.Host, nerr.Timeout(), nerr.Temporary(), err), code: status}
+			return nil, &proxyError{err: fmt.Errorf("net.Error during backend roundtrip to %s: timeout=%v temporary='%v': %w", req.URL.Host, nerr.Timeout(), nerr.Temporary(), err), code: status}
 		}
 
 		return nil, &proxyError{err: fmt.Errorf("unexpected error from Go stdlib net/http package during roundtrip: %w", err)}
@@ -1120,11 +1120,7 @@ func (p *Proxy) do(ctx *context) error {
 
 			p.metrics.IncErrorsBackend(ctx.route.Id)
 
-			if perr.code != 499 &&
-				retryable(ctx.Request()) &&
-				perr.DialError() &&
-				ctx.route.BackendType == eskip.LBBackend {
-
+			if retryable(ctx, perr) {
 				if ctx.proxySpan != nil {
 					ctx.proxySpan.Finish()
 					ctx.proxySpan = nil
@@ -1165,8 +1161,11 @@ func (p *Proxy) do(ctx *context) error {
 	return nil
 }
 
-func retryable(req *http.Request) bool {
-	return req != nil && (req.Body == nil || req.Body == http.NoBody)
+func retryable(ctx *context, perr *proxyError) bool {
+	req := ctx.Request()
+	return perr.code != 499 && perr.DialError() &&
+		ctx.route.BackendType == eskip.LBBackend &&
+		req != nil && (req.Body == nil || req.Body == http.NoBody)
 }
 
 func (p *Proxy) serveResponse(ctx *context) {
@@ -1270,10 +1269,10 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 		copyHeader(ctx.responseWriter.Header(), perr.additionalHeader)
 	}
 
-	msgPrefix := "error while proxying "
+	msgPrefix := "error while proxying"
 	logFunc := p.log.Errorf
 	if code == 499 {
-		msgPrefix = "client canceled "
+		msgPrefix = "client canceled"
 		logFunc = p.log.Infof
 	}
 	req := ctx.Request()
@@ -1282,9 +1281,9 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 	if i := strings.IndexRune(uri, '?'); i >= 0 {
 		uri = uri[:i]
 	}
-
 	logFunc(
-		msgPrefix+`after %v, route %s with backend %s %s%s, status code %d: %v, remote host: %s, request: "%s %s %s", user agent: "%s"`,
+		`%s after %v, route %s with backend %s %s%s, status code %d: %v, remote host: %s, request: "%s %s %s", user agent: "%s"`,
+		msgPrefix,
 		time.Since(ctx.startServe),
 		id,
 		backendType,
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -872,8 +872,7 @@ func (p *Proxy) makeUpgradeRequest(ctx *context, req *http.Request) error {
 func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Context) (*http.Response, *proxyError) {
 	req, endpoint, err := mapRequest(ctx, requestContext, p.flags.HopHeadersRemoval())
 	if err != nil {
-		p.log.Errorf("could not map backend request, caused by: %v", err)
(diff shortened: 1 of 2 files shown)


proxy/proxy.go
```diff
@@ -872,8 +872,7 @@ func (p *Proxy) makeUpgradeRequest(ctx *context, req *http.Request) error {
 func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Context) (*http.Response, *proxyError) {
 	req, endpoint, err := mapRequest(ctx, requestContext, p.flags.HopHeadersRemoval())
 	if err != nil {
-		p.log.Errorf("could not map backend request, caused by: %v", err)
-		return nil, &proxyError{err: err}
+		return nil, &proxyError{err: fmt.Errorf("could not map backend request, caused by: %w", err)}
```
AlexanderYastrebov, line 875: This kind of wrapping means we can not type assert on `proxyError.err` field anymore. Maybe its fine, I could not find any usage accept in `NetError()` which is not used so lets remove `NetError()` to reduce confusion.

proxy/proxy.go
```diff
@@ -946,12 +944,11 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 		ctx.proxySpan.LogKV("event", "error", "message", err.Error())
 
 		if perr, ok := err.(*proxyError); ok {
-			p.log.Errorf("Failed to do backend roundtrip to %s: %v", ctx.route.Backend, perr)
 			//p.lb.AddHealthcheck(ctx.route.Backend)
+			perr.err = fmt.Errorf("failed to do backend roundtrip to %s: %w", ctx.route.Backend, perr.err)
```
AlexanderYastrebov, line 948: Here I think we better use `req.URL` or something because `ctx.route.Backend` is only applicable to the network backend - i.e. it is empty for loadbalanced backend

szuecs: yeah seems that this is a leftover from the old days when we had no loadbalanced backend

proxy/proxy.go
szuecs: Maybe I should just use error.Is and not use type casting :)

AlexanderYastrebov: But I think we don't use it so lets remove, the type is private anyway

proxy/proxy.go
```diff
@@ -896,8 +895,7 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 
 	roundTripper, err := p.getRoundTripper(ctx, req)
 	if err != nil {
-		p.log.Errorf("Failed to get roundtripper: %v", err)
-		return nil, &proxyError{err: err, code: http.StatusBadGateway}
+		return nil, &proxyError{err: fmt.Errorf("Failed to get roundtripper: %w", err), code: http.StatusBadGateway}
```
AlexanderYastrebov, line 898: Lets fix `error strings should not be capitalized (ST1005)`

proxy/proxy.go
szuecs: https://go.dev/play/p/KA1Y0T8u3Dq

proxy/proxy.go
szuecs: I would go for req.URL.Host, because we don't need the specific URL and we would log basic auth credentials etc.

proxy/proxy.go
AlexanderYastrebov: I think we can omit `, caused by` in the message for clarity

proxy/proxy.go
```diff
@@ -960,11 +958,10 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 				status = http.StatusServiceUnavailable
 			}
 			p.tracing.setTag(ctx.proxySpan, HTTPStatusCodeTag, uint16(status))
-			return nil, &proxyError{err: err, code: status}
+			return nil, &proxyError{err: fmt.Errorf("net.Error during backend roundtrip to %s: timeout=%v temporary=%v: %w", ctx.route.Backend, nerr.Timeout(), nerr.Temporary(), err), code: status}
```
AlexanderYastrebov, line 961: Lets also use `req.URL.Host` here

proxy/proxy.go
```diff
@@ -371,7 +371,8 @@ func (e *proxyError) DialError() bool {
 }
 
 func (e *proxyError) NetError() net.Error {
```
AlexanderYastrebov, line 373: Lets remove `NetError()`, it is not used and `proxyError` is not exported

proxy/proxy.go

proxy/proxy.go

proxy/proxy.go
```diff
@@ -1125,6 +1114,9 @@ func (p *Proxy) do(ctx *context) error {
 		backendStart := time.Now()
 		rsp, perr := p.makeBackendRequest(ctx, backendContext)
 		if perr != nil {
+			if perr.code != 499 { // 499 is logged later with more context
+				p.log.Errorf("Failed to do backend request: %v", perr)
+			}
```
AlexanderYastrebov, line 1119: I think the proxy error is logged in `errorResponse` in all cases.
Why do we need this special case? Maybe we could fix the failing test somehow instead?

szuecs: Yes true but if we do not log it here we won't see retries in the logs, so if we only see this message and not the other it was a successful retry.
On the other hand we also try to retry on 499 if I understand our current code correctly, which might be also something to fix :)

szuecs: fixed

proxy/proxy.go
```diff
@@ -1131,7 +1120,11 @@ func (p *Proxy) do(ctx *context) error {
 
 			p.metrics.IncErrorsBackend(ctx.route.Id)
 
-			if retryable(ctx.Request()) && perr.DialError() && ctx.route.BackendType == eskip.LBBackend {
+			if perr.code != 499 &&
+				retryable(ctx.Request()) &&
+				perr.DialError() &&
+				ctx.route.BackendType == eskip.LBBackend {
```
AlexanderYastrebov, line 1126: Does it make sense to extract condition into a helper function like
```go
if retryable(ctx, perr) {
...
}
```
?

proxy/proxy.go
```diff
@@ -1279,55 +1270,35 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
...
+	}
+	req := ctx.Request()
+	remoteAddr := remoteHost(req)
+	uri := req.RequestURI
+	if i := strings.IndexRune(uri, '?'); i >= 0 {
+		uri = uri[:i]
 	}
 
+	logFunc(
+		msgPrefix+`after %v, route %s with backend %s %s%s, status code %d: %v, remote host: %s, request: "%s %s %s", user agent: "%s"`,
```
AlexanderYastrebov, line 1287: nit: we could name `msgPrefix` a `reason` and format it as any other argument like
```go
logf(
		`%s after %v, route %s with backend %s %s%s, status code %d: %v, remote host: %s, request: "%s %s %s", user agent: "%s"`,
		reason,
		time.Since(ctx.startServe),
		...)
```

AlexanderYastrebov:
:+1:

proxy/proxy.go
```diff
@@ -960,11 +950,10 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 				status = http.StatusServiceUnavailable
 			}
 			p.tracing.setTag(ctx.proxySpan, HTTPStatusCodeTag, uint16(status))
-			return nil, &proxyError{err: err, code: status}
+			return nil, &proxyError{err: fmt.Errorf("net.Error during backend roundtrip to %s: timeout=%v temporary=%v: %w", req.URL.Host, nerr.Timeout(), nerr.Temporary(), err), code: status}
```
aryszka, line 953: maybe it would be useful to put the temporary between single quotes like: `temporary='%v'`, so that it wouldn't get mixed up with the main `err` value's string

aryszka:
:+1:

proxy/proxy.go
szuecs: Temporary() returns bool so I think it's fine like this

proxy/proxy.go
szuecs: great idea, done

proxy/proxy.go
szuecs: changed

proxy/proxy.go
AlexanderYastrebov: Now the test for `retryable` should be updated which is good as it may test that 499 is not retried.

szuecs:
:+1:

proxy/proxy_test.go
```diff
@@ -323,38 +323,87 @@ func TestRetryable(t *testing.T) {
...
+
+			tt.req.URL.Host = u.Host
+
+			// find the call such that next should be failing
+			for {
+				rsp, _ := http.DefaultClient.Get(tt.req.URL.String())
+				if rsp.StatusCode == http.StatusOK {
+					break
+				}
+			}
```
AlexanderYastrebov, line 393: But if this hits unavailable `http://127.0.0.5:9` it will do retry and return OK just like it if hits working endpoint, won't it?

szuecs: that's a waiting loop until success, such that next will fail (below we do the real test)

szuecs: ah true, should do a POST request :D

szuecs: fixed

szuecs:
:+1:

aryszka:
:+1:

PR #1945, opened 2022-01-07
scheduler: ensure single lifo instance per route

AlexanderYastrebov:
Multiple `lifo` filter instances on the route share the same job queue but claim multiple slots per request.
This is not obvious when users configure `lifo` filter in the ingress manifest and another instance is added
by the `-default-filters-prepend` flag.

The change keeps only the last `lifo` filter instance on the route to enable default configuration override.

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -2237,8 +2237,7 @@ lifo(100, 150, "10s")
 The above configuration will set MaxConcurrency to 100, MaxQueueSize
 to 150 and Timeout to 10 seconds.
 
-When multiple lifo filters are set in a route, only one of them will be
-applied. It is undefined which one.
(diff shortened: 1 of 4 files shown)


szuecs:
@AlexanderYastrebov I think the same applies for lifoGroup() with the same group name. What do you think tackle this in the same PR or do you want to go on?

szuecs:
:+1:

AlexanderYastrebov:
> I think the same applies for lifoGroup() with the same group name. What do you think tackle this in the same PR or do you want to go on?

I am thinking to postpone it until we see a usecase when users need to re-define `lifoGroup` (like we have now for the `lifo` and `-default-filters-prepend`) in order not to impose any limitation without a real need.

AlexanderYastrebov:
:+1:

Issue #1851, opened 2021-09-08
Unknown error for route based LIFO: queue closed for host


AlexanderYastrebov:
**Describe the bug**

Users observe 500 response errors and `Unknown error for route based LIFO: queue closed for host` error in Skipper logs.

The error response is returned from 
https://github.com/zalando/skipper/blob/6b089867410bf4b9f56b4686450cf95301ac02c7/filters/scheduler/lifo.go#L295
https://github.com/zalando/skipper/blob/6b089867410bf4b9f56b4686450cf95301ac02c7/filters/scheduler/lifo.go#L310-L313
and originates in the [jobqueue](https://github.com/aryszka/jobqueue/blob/f6e428d125b3e03bce3c6a375e44a58361132f46/jobqueue.go#L240-L242).

It seems that queue got closed by scheduler post-processor on route update:
https://github.com/zalando/skipper/blob/6b089867410bf4b9f56b4686450cf95301ac02c7/scheduler/scheduler.go#L311-L318
while in-flight requests try to get slot from the queue (i.e. a race condition between ongoing requests and lifo scheduler post-processor).

`existingKeys` contains lifo filter names that are derived from route id:
https://github.com/zalando/skipper/blob/6b089867410bf4b9f56b4686450cf95301ac02c7/scheduler/scheduler.go#L248-L249

**To Reproduce**

1. Create a routes file and constantly update route id
```
$ for i in {1..1000}; do echo 'route_id'$i': Path("/") -> lifo(1000, 1000, "10s") -> status(204) -> <shunt>' > /tmp/lifo.eskip; sleep 3; done
```
2. Run Skipper
```
bin/skipper -routes-file=/tmp/lifo.eskip -access-log-disabled
```
3. Run vegeta
```
echo "GET http://localhost:9090/" | vegeta attack -rate 1000/s | vegeta report -every 1s
```
4. Observe occasional lifo errors in Skipper logs
```
[APP]INFO[0117] route settings, update, route: route_id20: Path("/") -> lifo(1000, 1000, "10s") -> status(204) -> <shunt> 
[APP]INFO[0117] route settings, update, deleted id: route_id19 
[APP]INFO[0117] route settings received                      
[APP]INFO[0117] route settings applied                       
[APP]ERRO[0117] Unknown error for route based LIFO: queue closed for host localhost:9090 
```
and 500s in vegeta report
```
Requests      [total, rate, throughput]  13006, 1000.02, 999.90
[...]

PR #1953, opened 2022-02-11
scheduler: delays closing of deleted lifo queues

AlexanderYastrebov:
Closing of a lifo queue on route deletion while in-flight requests try
to get a slot from it results in `Unknown error for route based LIFO: queue closed for host`
error and 500 response to the client.

To mitigate the problem this change puts deleted queues aside and closes them after one minute.
The proper fix would be to close queue when all route's in-flight requests complete, see #202.

Updates #1851

### DIFF
--- a/scheduler/export_test.go
+++ b/scheduler/export_test.go
@@ -1,0 +1,5 @@
+package scheduler
+
+var (
+	ExportQueueCloseDelay = &queueCloseDelay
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #1416, opened 2020-04-08
fix builds with new builder https://github.com/docker/buildx/issues/138

szuecs:
fix builds with new builder https://github.com/docker/buildx/issues/138
Signed-off-by: Sandor Szücs <sandor.szuecs@zalando.de>

PR #1954, opened 2022-02-14
Modify pipeline steps to build `arm` images

linki:
Following up on https://github.com/zalando/skipper/pull/1416.

### What does it do?

This builds skipper images for `arm` (v7, 32-bit) and `arm64` (v8, 64-bit).

The multi-stage build might look a bit strange to you but it delivers the correct result on the build system being used (CDP) which runs on `amd64` instances.

The first stage downloads an `amd64` image (via `--platform=linux/amd64` in combination with a multi-arch aware `alpine:3.13` image) within which it can run commands on the build architecture (`apk add` in our case but could be anything).

The second stage and final production image uses an `arm` base image (e.g. via `--platform=linux/arm64`) and merely copies in the desired artifacts (in our case that's the previously built `skipper` binary for the target architecture as well as the side effects of our additional commands. The latter will not always work like this depending on what you do but it works for our ca-certificates use case).

This allows building the image entirely on `amd64` hardware without the need for emulation while still resulting in the desired images, including its architecture related metadata.

### What are the alternatives?

Other options to build this are:
* run the build on `arm` hardware
* run it with `qemu`-emulation on `amd64` hardware like in https://github.com/zalando/skipper/pull/1416 or
* if possible drop any commands (e.g. `apk add`, `update-ca-certificates`) and merely copy in the binary

### Further remarks

In the future, we should provide multi-arch skipper images hosted on a multi-arch aware registry, such as Docker Hub, ECR or Zalando's new registry (PierOne is currently not capable of doing that). Multi-arch images allow for using the same image name and tag, e.g. `skipper:v1.2.3`, which gets resolved to the client's requested architecture when pulling. It allows to transparently use the same logical image on different hardware. This can be done in a separate PR.

I would like to ignore the warnings about non-compliant base images for now as the previously used base images where also non-compliant.


szuecs:
:+1:

AlexanderYastrebov:
:+1:

Issue #696, opened 2018-07-02
Brotli Support in `compress()` filter


grassator:
Brotli is [widely supported by the browsers](https://caniuse.com/#feat=brotli) and [seems to offer better compression than gzip with faster speed](https://blogs.akamai.com/2016/02/understanding-brotlis-potential.html).

There seem to be [official binding](https://github.com/google/brotli/tree/master/go) but no native go implementation from what I could find.

szuecs:
We could also implement an opensource filter plugin to implement brotli

grassator:
@szuecs Would you make it as a separate filter then?

aryszka:
wondering if the init phase of the brotli support requires any native calls. If it doesn't, then there's not much difference. In case we decide to build it as a plugin, then yes, it has to be a separate filter. Also, if it's a plugin, then we should extract the stream encoding logic of the current compress filter into its own support package (e.g. "streamencoding").

grassator:
@aryszka how would you imagine adding support without native calls?

szuecs:
@yes I would put it into https://github.com/skipper-plugins/filters

grassator:
@szuecs I'm not sure that a separate filter is a good idea. The problem is that the client sends a list of preferred encodings as a header: `Accept-Encoding: brotli,gzip,deflate`. If we implement `brotli()` as a separate filter, the order `compress()` and `brotli()` in the route will influence the resulting behavior which is brittle.

Alternative would be to have a default `compress()` filter implementation that does not include brotli and then have some kind of code generation / compile time flag that pulls in brotli native code.

We could also do `compressWithBrotliSupport()` but seems ugly as well...

szuecs:
You are right that it can be done withing `compress` filter or you could add filters to configure what compress will support via filters like `brotli()`.

szuecs:
https://github.com/andybalholm/brotli could be likely used to implement the reader and writer. I also created https://github.com/klauspost/compress/issues/443 , because this includes optimized drop in replacements for other compress algorithms, too.

PR #1956, opened 2022-02-15
Add brotli to skipper

fpozzobon:
Adding brotli compression / decompression to skipper

References: https://github.com/zalando/skipper/issues/696

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -62,6 +62,7 @@ type Config struct {
 	PredicatePlugins                *pluginFlag    `yaml:"predicate-plugin"`
 	DataclientPlugins               *pluginFlag    `yaml:"dataclient-plugin"`
 	MultiPlugins                    *pluginFlag    `yaml:"multi-plugin"`
+	CompressEncodings               *listFlag      `yaml:"compress-encodings"`
 
 	// logging, metrics, profiling, tracing:
 	EnablePrometheusMetrics             bool      `yaml:"enable-prometheus-metrics"`
@@ -281,6 +282,7 @@ func NewConfig() *Config {
 	cfg.RoutesURLs = commaListFlag()
 	cfg.ForwardedHeadersList = commaListFlag()
 	cfg.ForwardedHeadersExcludeCIDRList = commaListFlag()
+	cfg.CompressEncodings = commaListFlag("gzip", "br", "deflate")
 
 	flag.StringVar(&cfg.ConfigFile, "config-file", "", "if provided the flags will be loaded/overwritten by the values on the file (yaml)")
 
@@ -318,6 +320,7 @@ func NewConfig() *Config {
 	flag.Var(cfg.PredicatePlugins, "predicate-plugin", "set a custom predicate plugins to load, a comma separated list of name and arguments")
 	flag.Var(cfg.DataclientPlugins, "dataclient-plugin", "set a custom dataclient plugins to load, a comma separated list of name and arguments")
 	flag.Var(cfg.MultiPlugins, "multi-plugin", "set a custom multitype plugins to load, a comma separated list of name and arguments")
+	flag.Var(cfg.CompressEncodings, "compress-encodings", "set encodings supported for compression, when client has no preference, uses in priority in the order of this array")
 
 	// logging, metrics, tracing:
 	flag.BoolVar(&cfg.EnablePrometheusMetrics, "enable-prometheus-metrics", false, "*Deprecated*: use metrics-flavour. Switch to Prometheus metrics format to expose metrics")
@@ -672,6 +675,7 @@ func (c *Config) ToOptions() skipper.Options {
 		DataClientPlugins:               c.DataclientPlugins.values,
 		Plugins:                         c.MultiPlugins.values,
 		PluginDirs:                      []string{skipper.DefaultPluginDir},
+		CompressEncodings:               c.CompressEncodings.values,
 
 		// logging, metrics, profiling, tracing:
 		EnablePrometheusMetrics:             c.EnablePrometheusMetrics,
--- a/config/config.go
+++ b/config/config.go
@@ -282,7 +282,7 @@ func NewConfig() *Config {
 	cfg.RoutesURLs = commaListFlag()
 	cfg.ForwardedHeadersList = commaListFlag()
 	cfg.ForwardedHeadersExcludeCIDRList = commaListFlag()
-	cfg.CompressEncodings = commaListFlag("gzip", "br", "deflate")
+	cfg.CompressEncodings = commaListFlag("gzip", "deflate", "br")
 
 	flag.StringVar(&cfg.ConfigFile, "config-file", "", "if provided the flags will be loaded/overwritten by the values on the file (yaml)")
 
@@ -320,7 +320,7 @@ func NewConfig() *Config {
 	flag.Var(cfg.PredicatePlugins, "predicate-plugin", "set a custom predicate plugins to load, a comma separated list of name and arguments")
 	flag.Var(cfg.DataclientPlugins, "dataclient-plugin", "set a custom dataclient plugins to load, a comma separated list of name and arguments")
 	flag.Var(cfg.MultiPlugins, "multi-plugin", "set a custom multitype plugins to load, a comma separated list of name and arguments")
-	flag.Var(cfg.CompressEncodings, "compress-encodings", "set encodings supported for compression, when client has no preference, uses in priority in the order of this array")
+	flag.Var(cfg.CompressEncodings, "compress-encodings", "set encodings supported for compression, the order defines priority when Accept-Header has equal quality values, see RFC 7231 section 5.3.1")
 
 	// logging, metrics, tracing:
 	flag.BoolVar(&cfg.EnablePrometheusMetrics, "enable-prometheus-metrics", false, "*Deprecated*: use metrics-flavour. Switch to Prometheus metrics format to expose metrics")
--- a/docs/kubernetes/ingress-usage.md
+++ b/docs/kubernetes/ingress-usage.md
@@ -426,7 +426,7 @@ Add a HTTP header in the response path of your clients.
 
 ### Enable compression
 
-Compress responses with gzip/brotli (depending on accepted encoding).
+Compress responses with accepted encoding (more details [here](../reference/filters.md#compress)).
 
     compress() // compress all valid MIME types
     compress("text/html") // only compress HTML files
--- a/docs/kubernetes/ingress-usage.md
+++ b/docs/kubernetes/ingress-usage.md
@@ -424,13 +424,13 @@ Add a HTTP header in the response path of your clients.
 
     setResponseHeader("X-Foo", "bar")
 
-### Enable gzip
+### Enable compression
 
-Compress responses with gzip.
+Compress responses with gzip/brotli (depending on accepted encoding).
 
     compress() // compress all valid MIME types
     compress("text/html") // only compress HTML files
-    compress(9, "text/html") // control the level of compression, 1 = fastest, 9 = best compression, 0 = no compression
+    compress(11, "text/html") // control the level of compression, 1 = fastest, 11 = best compression (fallback to 9 for gzip), 0 = no compression
 
 ### Set the Path
 
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -372,9 +372,15 @@ no-compression, 1 means best-speed and 11 means best-compression. Example:
 ```
 
 The filter also checks the incoming request, if it accepts the supported encodings,
-explicitly stated in the Accept-Encoding header. The filter currently supports `gzip`, `brotli`
-and `deflate`. It does not assume that the client accepts any encoding if the
+explicitly stated in the Accept-Encoding header. 
+The filter currently supports by default `gzip`, `deflate` and `br` (can be overridden with flag `compress-encodings`). 
+It does not assume that the client accepts any encoding if the
 Accept-Encoding header is not set. It ignores * in the Accept-Encoding header.
+
+Supported encodings are prioritized on:
+- quality value provided by client
+- compress-encodings flag following order as provided if quality value is equal
+- `gzip`, `deflate`, `br` in this order otherwise
 
 When compressing the response, it updates the response header. It deletes the
 `Content-Length` value triggering the proxy to always return the response with chunked
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -392,7 +392,7 @@ The compression happens in a streaming way, using only a small internal buffer.
 ## decompress
 
 The filter, when executed on the response path, checks if the response entity is
-compressed by a supported algorithm. To decide, it checks the Content-Encoding
+compressed by a supported algorithm (`gzip`, `deflate`, `br`). To decide, it checks the Content-Encoding
 header.
 
 When compressing the response, it updates the response header. It deletes the
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -365,14 +365,14 @@ to be omitted:
 It is possible to control the compression level, by setting it as the first filter
 argument, in front of the MIME types. The default compression level is best-speed.
 The possible values are integers between 0 and 9 (inclusive), where 0 means
-no-compression, 1 means best-speed and 9 means best-compression. Example:
+no-compression, 1 means best-speed and 11 means best-compression. Example:
 
 ```
-* -> compress(9, "image/tiff") -> "https://www.example.org"
+* -> compress(11, "image/tiff") -> "https://www.example.org"
 ```
 
 The filter also checks the incoming request, if it accepts the supported encodings,
-explicitly stated in the Accept-Encoding header. The filter currently supports `gzip`
+explicitly stated in the Accept-Encoding header. The filter currently supports `gzip`, `brotli`
 and `deflate`. It does not assume that the client accepts any encoding if the
 Accept-Encoding header is not set. It ignores * in the Accept-Encoding header.
 
--- a/filters/builtin/compress.go
+++ b/filters/builtin/compress.go
@@ -28,8 +28,8 @@ type encoding struct {
 type encodings []*encoding
 
 type compress struct {
-	mime  []string
-	level int
+	mime     []string
+	level    int
 	encoding []string
 }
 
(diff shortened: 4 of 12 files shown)


filters/builtin/compress.go
```diff
@@ -268,9 +272,20 @@ func unsupported() {
 
 func newEncoder(enc string, level int) (encoder, error) {
 	switch enc {
+	case "br":
+		if level > brotli.BestCompression {
```
szuecs, line 276: this check you already did on filter creation (CreateFilter())

fpozzobon: removed :)

filters/builtin/compress.go
```diff
@@ -366,11 +390,20 @@ func (c *compress) Response(ctx filters.FilterContext) {
...
 	if enc == "" {
 		return
 	}
 
 	responseHeader(rsp, enc)
 	responseBody(rsp, enc, c.level)
 }
+
+func index(array []string, value string) int {
+	for k, v := range array {
```
szuecs, line 403: nit: k->i 
i is index
k is key

filters/builtin/compress.go
```diff
@@ -219,11 +232,12 @@ func acceptedEncoding(r *http.Request) string {
 		}
 
 		name := strings.ToLower(strings.TrimSpace(sp[0]))
-		if !stringsContain(supportedEncodings, name) {
+		prio := index(c.encoding, name)
```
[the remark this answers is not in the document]
AlexanderYastrebov: If we are looking to have a priority for the same quality value (or absence thereof) then I think we should look at stable sort https://pkg.go.dev/sort#Stable that preserves client preferred algorithm order from the header.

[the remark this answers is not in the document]
szuecs: @AlexanderYastrebov clients tell they can do X,Y,Z and the priority gives the owner of skipper a choice what algorithm should be preferred. For example you might want to use brotli as first choice and fallback to gzip. A Browser sends in some random order what they support, and then you get something you are lucky and have no control.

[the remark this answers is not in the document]
AlexanderYastrebov: > A Browser sends in some random order what they support, and then you get something you are lucky and have no control.

Looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding I think client sends its preferences using quality value for ordering (default quality value when absent is 1). The server serves the client so I think the logic should obey client preferences, i.e. use provided or default quality value without any additional server-side priority.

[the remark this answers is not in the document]
fpozzobon: @AlexanderYastrebov you are correct, the priority is kept on `quality value` if present but
checking on my Chrome browser, I can see the following:
`Accept-Encoding: gzip, deflate, br`

=> does it mean no preference or that `gzip` should be prioritised over `br`?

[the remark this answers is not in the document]
AlexanderYastrebov: I would say so, but I don't see order-dependent preference specified anywhere.
This answer https://stackoverflow.com/questions/3225136/http-what-is-the-preferred-accept-encoding-for-gzip-deflate also says first should be preferred but provides no reference.

If we would like to have a server side preference list then it would be interesting to address 
- why would brotli have priority over gzip (given that chrome does not do that and that it would be a change of behavior for the existing configurations)
- should the server-side priority list be configurable

E.g. Caddy implements configurable server side preference afaict from https://github.com/caddyserver/caddy/issues/3692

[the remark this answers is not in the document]
aryszka: I prefer the approach without the additional p field, and using stable sort to preserve the client specified order (which would be a fix of the current state in master)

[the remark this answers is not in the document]
AlexanderYastrebov: We've discussed this with @aryszka and came to conclusion that we should have configurable server-side priority for the algorithms in case quality values are equal (including default 1).

The default priority ordering should put brotly at the end (i.e. the priority should be `gzip > deflate > br`) to avoid unexpected algorithm change for the existing deployments.
Skipper operators then would be able to change the priority using a flag, something like `-compress-encoding-prefer=br,gzip,deflate` or e.g. exclude an algorithm like `-compress-encoding-prefer=br,gzip`.

skipper.go
```diff
@@ -1456,6 +1459,10 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 		)
 	}
 
+	if len(o.CompressEncodings) > 0 {
+		o.CustomFilters = append(o.CustomFilters, builtin.NewCompressWithEncodings(o.CompressEncodings))
```
AlexanderYastrebov, line 1463: I think we need to remove `NewCompress()` from built-in registry to avoid name clash

filters/builtin/compress.go
```diff
@@ -20,14 +21,16 @@ const bufferSize = 8192
...
 }
 
 type encodings []*encoding
 
 type compress struct {
-	mime  []string
-	level int
+	mime     []string
+	level    int
+	encoding []string
```
AlexanderYastrebov, line 33: I suggest to use a map here like
```go
map[string]int {
  "br": 1,
  "gzip": 2,
}
```
this should simplify priority lookup and would not require auxiliary `index` method

filters/builtin/compress.go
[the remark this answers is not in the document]
fpozzobon: thanks a lot for your feedbacks :)
I would suggest `compress-encodings` in place of `compress-encoding-prefer` (shorter)

config/config.go
```diff
@@ -318,6 +320,7 @@ func NewConfig() *Config {
 	flag.Var(cfg.PredicatePlugins, "predicate-plugin", "set a custom predicate plugins to load, a comma separated list of name and arguments")
 	flag.Var(cfg.DataclientPlugins, "dataclient-plugin", "set a custom dataclient plugins to load, a comma separated list of name and arguments")
 	flag.Var(cfg.MultiPlugins, "multi-plugin", "set a custom multitype plugins to load, a comma separated list of name and arguments")
+	flag.Var(cfg.CompressEncodings, "compress-encodings", "set encodings supported for compression, when client has no preference, uses in priority in the order of this array")
```
fpozzobon, line 323: @AlexanderYastrebov is the description clear on the expected behaviour?

filters/builtin/compress.go
```diff
@@ -123,28 +138,28 @@ func (e encodings) Swap(i, j int)      { e[i], e[j] = e[j], e[i] }
 //
 // The compression happens in a streaming way, using only a small internal buffer.
 //
-func NewCompress() filters.Spec { return &compress{} }
+func NewCompress() filters.Spec { return &compress{encoding: supportedEncodings} }
```
AlexanderYastrebov, line 141: Lets call `NewCompressWithEncodings()` here

filters/builtin/compress.go
```diff
@@ -123,28 +138,28 @@ func (e encodings) Swap(i, j int)      { e[i], e[j] = e[j], e[i] }
...
+
+func NewCompressWithEncodings(encoding []string) filters.Spec { return &compress{encoding: encoding} }
```
AlexanderYastrebov, line 143: Would be nice to check that argument contains only supported values

skipper.go
AlexanderYastrebov: I think we should not remove it from `builtin.MakeRegistry()` as it could potentially break library users (but I think we've done this for some other filters in the past that required new config), @szuecs @aryszka WDYT?

Maybe its fine to put it into custom filters, we would need to update a comment below (and/or add a log to `registry.Register`) to hint that `CustomFilters` may override built-in.

config/config.go
AlexanderYastrebov: Maybe something like
`set encodings supported for compression, the order defines priority when Accept-Header has equal quality values, see RFC 7231 section 5.3.1`

config/config.go
```diff
@@ -281,6 +282,7 @@ func NewConfig() *Config {
 	cfg.RoutesURLs = commaListFlag()
 	cfg.ForwardedHeadersList = commaListFlag()
 	cfg.ForwardedHeadersExcludeCIDRList = commaListFlag()
+	cfg.CompressEncodings = commaListFlag("gzip", "br", "deflate")
```
AlexanderYastrebov, line 285: I think the default should be `"gzip", "deflate", "br"` to keep backward compatibility

docs/reference/filters.md
```diff
@@ -365,14 +365,14 @@ to be omitted:
...
+no-compression, 1 means best-speed and 11 means best-compression. Example:
 
 ```
-* -> compress(9, "image/tiff") -> "https://www.example.org"
+* -> compress(11, "image/tiff") -> "https://www.example.org"
 ```
 
 The filter also checks the incoming request, if it accepts the supported encodings,
-explicitly stated in the Accept-Encoding header. The filter currently supports `gzip`
+explicitly stated in the Accept-Encoding header. The filter currently supports `gzip`, `brotli`
```
AlexanderYastrebov, line 375: We should mention config option here

docs/kubernetes/ingress-usage.md
```diff
@@ -424,13 +424,13 @@ Add a HTTP header in the response path of your clients.
 
     setResponseHeader("X-Foo", "bar")
 
-### Enable gzip
+### Enable compression
 
-Compress responses with gzip.
+Compress responses with gzip/brotli (depending on accepted encoding).
```
AlexanderYastrebov, line 429: Maybe we should just link to the filter docs here to not repeat the details

filters/builtin/compress.go
szuecs: no that's more complex than having a slice that has an order already

skipper.go
szuecs: We could also move NewCompress() to the else branch of the code here and this should not break skipper library users.

config/config.go
fpozzobon: sure I can change, it's only to enforce the allowed options here

filters/builtin/compress.go
```diff
@@ -104,47 +119,60 @@ func (e encodings) Swap(i, j int)      { e[i], e[j] = e[j], e[i] }
...
-// encoding and sets the Vary: Accept-Encoding header, if missing.
+// with chunked transfer encodingPriority, sets the Content-Encoding to the selected
+// encodingPriority and sets the Vary: Accept-Encoding header, if missing.
 //
 // The compression happens in a streaming way, using only a small internal buffer.
 //
-func NewCompress() filters.Spec { return &compress{} }
+func NewCompress() filters.Spec { return NewCompressWithEncodings(supportedEncodings) }
+
+func NewCompressWithEncodings(encoding []string) filters.Spec {
```
AlexanderYastrebov, line 147: Since we are introducing options, lets make it future proof by using `Options` struct and `NewCompressWithOptions`, otherwise we would need to invent a new constructor for each new option

filters/builtin/compress.go
```diff
@@ -37,8 +40,8 @@ type encoder interface {
 }
 
 var (
-	supportedEncodings  = []string{"gzip", "deflate"}
-	unsupportedEncoding = errors.New("unsupported encoding")
+	supportedEncodings  = []string{"gzip", "deflate", "br"}
+	unsupportedEncoding = errors.New("unsupported encodingPriority")
```
AlexanderYastrebov, line 44: I think `unsupported encoding` was fine

filters/builtin/compress.go
```diff
@@ -104,47 +119,60 @@ func (e encodings) Swap(i, j int)      { e[i], e[j] = e[j], e[i] }
...
 // The compression happens in a streaming way, using only a small internal buffer.
 //
-func NewCompress() filters.Spec { return &compress{} }
+func NewCompress() filters.Spec { return NewCompressWithEncodings(supportedEncodings) }
+
+func NewCompressWithEncodings(encoding []string) filters.Spec {
+	encodingMap := map[string]int{}
+	for i, v := range encoding {
+		if !stringsContain(supportedEncodings, v) {
+			continue // skipping unsupported encodingPriority
```
AlexanderYastrebov, line 151: I think it is better to return error here

fpozzobon: compromising with warning log, this case would happen only if developer add a new option for compress but forget to udpate `supportedEncodings`

filters/builtin/compress.go
fpozzobon: updated

filters/builtin/compress.go
fpozzobon: good point, updated

docs/reference/filters.md
fpozzobon: added

filters/builtin/compress.go
fpozzobon: it is indeed but as we need to loop through all the client supported encodings to find the "best match", the complexity is probably worth it? (implemented)

filters/builtin/compress.go
```diff
@@ -104,47 +119,60 @@ func (e encodings) Swap(i, j int)      { e[i], e[j] = e[j], e[i] }
...
+// with chunked transfer encodingPriority, sets the Content-Encoding to the selected
+// encodingPriority and sets the Vary: Accept-Encoding header, if missing.
 //
 // The compression happens in a streaming way, using only a small internal buffer.
 //
-func NewCompress() filters.Spec { return &compress{} }
+func NewCompress() filters.Spec { return NewCompressWithEncodings(supportedEncodings) }
+
+func NewCompressWithEncodings(encoding []string) filters.Spec {
+	encodingMap := map[string]int{}
```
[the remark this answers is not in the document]
fpozzobon: replaced with m, arg is now option

filters/filters.go
```diff
@@ -169,6 +170,9 @@ var ErrInvalidFilterParameters = errors.New("invalid filter parameters")
 
 // Registers a filter specification.
 func (r Registry) Register(s Spec) {
+	if _, ok := r[s.Name()]; ok {
+		log.Infof("Replacing registry: %s", s.Name())
```
AlexanderYastrebov, line 174: `name := s.Name()`
`Replacing %s filter specification`

filters/builtin/compress_test.go
```diff
@@ -270,14 +275,14 @@ func TestCompress(t *testing.T) {
...
-		"does not accept encoding",
+		"does not accept encodingPriority",
 		http.Header{},
 		3 * 8192,
 		nil,
 		"",
 		http.Header{},
 	}, {
-		"unknown encoding",
+		"unknown encodingPriority",
```
AlexanderYastrebov, line 285: Lets revert encoding->encodingPriority changes in strings and docs: the `encodingPriority` is an internal structure name while e.g. this test verifies that unknown encoding has no effect, i.e. its not related to "priority" in any way

fpozzobon: good catch, bad refactoring. done

filters/builtin/compress.go
```diff
@@ -104,47 +124,63 @@ func (e encodings) Swap(i, j int)      { e[i], e[j] = e[j], e[i] }
...
+	return NewCompressWithOptions(CompressOptions{supportedEncodings})
+}
+
+func NewCompressWithOptions(options CompressOptions) filters.Spec {
+	m := map[string]int{}
+	for i, v := range options.Encodings {
+		if !stringsContain(supportedEncodings, v) {
+			log.Warningf("Skipping unsupported encoding: %s", v)
+			continue
+		}
```
AlexanderYastrebov, line 160: Options are derived from user-supplied string flag, by returning a error and handling it in the startup routine we would prevent Skipper from starting with unsupported values e.g. due to typo and let users notice the problem right away instead of postponing it until they ever look into the logs.

filters/builtin/compress.go
AlexanderYastrebov: The suggestion was implemented. Map lookup is cleaner and does not require auxiliary `index` method that iterates slice to find a value.

filters/builtin/compress.go
fpozzobon: the options are already restricted in config.go `commaListFlag("gzip", "deflate", "br")`
=> the user will get an error if he tries to use a wrong option
=> the use case would be if a developer add a new option in `commaListFlag`

I understand your concern but given that this could occur only from developer mistake, maybe we can keep it simple here in place of introducing possible error returned from new?

AlexanderYastrebov: Ok, fair enough, I misinterpreted `commaListFlag("gzip", "deflate", "br")` as a default value while it is actually an allowlist.

AlexanderYastrebov: OTOH looking forward for new options we might not be so lucky to never have to return error and adding an error would be a breaking change for the `NewCompressWithOptions`

filters/builtin/compress_test.go
```diff
@@ -373,6 +387,42 @@ func TestCompress(t *testing.T) {
 		http.Header{
 			"Content-Encoding": []string{"deflate"},
 			"Vary":             []string{"Accept-Encoding"}},
+	}, {
+		"brotli",
```
AlexanderYastrebov, line 391: I think we do not test custom encoding preference order.
Maybe we could add `encodings []string` to the test case struct and when it is not empty then redefine compress filter using `NewCompressWithOptions` like we do in skipper.go.

This way we can test that e.g. `-compress-encodings=br,gzip` indeed picks `br` over `gzip` when `Accept-Encoding` has the Chrome/Firefox default value of `gzip, deflate, br`.

filters/builtin/compress.go
szuecs: We have to return `filters.Spec`, there is no error that we can return. Errors should be handled upfront before calling.

AlexanderYastrebov: > We have to return filters.Spec

Why? We can return `filters.Spec, error` from `NewCompressWithOptions` and let caller bail out on error. The existing `NewCompress` should continue to return just a spec as it uses a valid option value.

filters/builtin/compress_test.go
fpozzobon: good catch, added unit test

filters/builtin/compress.go
fpozzobon: even though there is 99% chance that it will never happen, ignoring error on NewCompress seems a bit of a trap for developer no? :)

szuecs: Hmm maybe you are right, but I think we have no case were we have 2 return values to instantiate a spec.
I wonder if I forgot why...

AlexanderYastrebov: >  ignoring error on NewCompress seems a bit of a trap

We can not change `NewCompress` signature to return error due to backwards compatibility but since it does not receive any argument we can make sure it passes safe defaults to `NewCompressWithOptions` that never cause an error.

fpozzobon: good point, updated following above suggestions.
NewCompress is indeed tested already in the unit test

aryszka:
:+1:

PR #1960, opened 2022-02-17
Use testcontainers to run `redis-server`

AlexanderYastrebov:
Mitigates build failures caused by package manager redirects during `redis-server` installation.
Note that change adds many dependencies but they are only used for tests and do not end up in the build.


AlexanderYastrebov:
If that works we can also use it for etcd
https://github.com/zalando/skipper/blob/100ef1b584304d1e17f33d07cf8952ded84252cb/etcd/etcdtest/mocketcd.go#L45-L47

net/redistest/redistest.go
```diff
@@ -2,57 +2,71 @@ package redistest
...
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+	defer cancel()
 
-	cmd := exec.CommandContext(ctx, "redis-server", args...) // #nosec
-	if err := cmd.Start(); err != nil {
+	container, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
+		ContainerRequest: testcontainers.ContainerRequest{
+			Image:        "redis:6-alpine",
+			Cmd:          args,
+			ExposedPorts: []string{"6379/tcp"},
```
szuecs, line 33: Does a test container gets it's own IP?

AlexanderYastrebov: It gets a mapped port on localhost that we obtain using .Endpoint below

szuecs: sounds like we could test sharding with this too, but we need to change the port for the next ones :)

szuecs:
Do you also want to add etcd?

AlexanderYastrebov:
> Do you also want to add etcd?

I thought about doing it in separate PR

delivery.yaml
```diff
@@ -8,8 +8,6 @@ pipeline:
   commands:
   - desc: build-push
     cmd: |
-      apt-get update  -o Acquire::http::AllowRedirect=false
-      apt-get install -o Acquire::http::AllowRedirect=false -y bzr redis-server jq
```
AlexanderYastrebov, line 12: I don't see `bzr` used anywhere but I think this would break a Makefile that depends on `jq`

aryszka:
:+1:

PR #1961, opened 2022-02-18
Reverts removal of `jq` install

AlexanderYastrebov:
`jq` installation was removed by #1960 but it is required by `packaging/Makefile`


delivery.yaml
```diff
@@ -8,6 +8,8 @@ pipeline:
   commands:
   - desc: build-push
     cmd: |
+      apt-get update
+      apt-get install -y jq
```
AlexanderYastrebov, line 12: It is not actually required for the build in the pipeline:
The `$$HOME/.docker/config.json` does not exist in the pipline image so the content is echoed directly without using jq
https://github.com/zalando/skipper/blob/1a1d1eb186e495b7e67f9b10d278b11652d44f89/packaging/Makefile#L61

AlexanderYastrebov:
Not needed for the build

AlexanderYastrebov:
I've forgot about this and did one more time https://github.com/zalando/skipper/pull/2120

Issue #1955, opened 2022-02-14
OIDC: Aggregated and Distributed Claims


universam1:
**Is your feature request related to a problem? Please describe.**

For OIDC (using filter `oauthOidcAnyClaims` ) with _Azure AD_ we are seeing users who are member of many groups to run into a [clipped groups token behaviour](https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens#groups-overage-claim) of Azure. 
The `groups` claim is replaced by `hasToken` and a pointer to the Graph API.
That breaks subsequent filter `oidcClaimsQuery` for group membership, as the expected claim of groups is empty

**Describe the solution you would like**
Skipper detects this flag `hasToken` and performs a lookup to the Graph API, retrieving the actual list of groups, storing in the statebag or cookie.
It could be a configurable option to perform this lookup.

**Additional context (optional)**
Would it make sense to implement this lookup to the Graph API, ~granted this looks like a Azure-ish behaviour~?

Which filter would be the proper place to extend?

**Would you like to work on it?**
Potentially, with support.

universam1:
It as actually a well defined standard behaviour: https://openid.net/specs/openid-connect-core-1_0.html#AggregatedDistributedClaims

webframp:
If it's part of the oidc standard then it seems potentially useful to implement to me. Interestingly both referenced example implementations look like they were able to be implemented using `golang.org/x/oauth2` without requiring any azure specific implementation library. That's a nice design goal I would think.

Although some nice libraries like https://github.com/manicminer/hamilton/ exist, it would not seem very nice to require that just to support this feature.

szuecs:
If it's part of the spec let's build it of course!
So I guess we should put the aggregated data into our container that we pass to stateBag which can queried via `oidcClaimsQuery` or how do you think to integrate it?

PR #1962, opened 2022-02-18
resolve Aggregated and Distributed Claims

universam1:
fixes: #1955

This solution is scoped to Azure behaviour, taking into account the specs from
https://openid.net/specs/openid-connect-core-1_0.html#AggregatedDistributedClaims

There are some Azure related API calls included but trying to support other providers, which is though unknown at this time.

it transforms a distributed claim

```json
{
    "_claim_names": {
        "groups": "src1"
    },
    "_claim_sources": {
        "src1": {
            "endpoint": "https://graph.windows.net/.../getMemberObjects"
        }
    }
}
```

into a full populated token, which is saved in `statebag` and in the `cookie` for follow up processing

```json
{
    "_claim_names": {
        "groups": "src1"
    },
    "_claim_sources": {
        "src1": {
            "endpoint": "https://graph.windows.net/.../getMemberObjects"
        }
    },
    "groups": [
        "group1",
        "group2",
        ...
    ]
}
```

### DIFF
--- a/filters/auth/oidc.go
+++ b/filters/auth/oidc.go
@@ -7,6 +7,7 @@ import (
 	"crypto/sha256"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -19,9 +20,11 @@ import (
 	"time"
 
 	"github.com/coreos/go-oidc"
+	"github.com/opentracing/opentracing-go"
 	log "github.com/sirupsen/logrus"
 	"github.com/tidwall/gjson"
 	"github.com/zalando/skipper/filters"
+	snet "github.com/zalando/skipper/net"
 	"github.com/zalando/skipper/secrets"
 	"golang.org/x/oauth2"
 )
@@ -38,7 +41,33 @@ const (
 	stateValidity       = 1 * time.Minute
 	oidcInfoHeader      = "Skipper-Oidc-Info"
 	cookieMaxSize       = 4093 // common cookie size limit http://browsercookielimits.squawky.net/
+
+	// Deprecated: The host of the Azure Active Directory (AAD) graph API
+	azureADGraphHost = "graph.windows.net"
 )
+
+var (
+	distributedClaimsClients = sync.Map{}
+	microsoftGraphHost       = "graph.microsoft.com" // global for testing
+)
+
+type distributedClaims struct {
+	ClaimNames   map[string]string      `json:"_claim_names"`
+	ClaimSources map[string]claimSource `json:"_claim_sources"`
+}
+
+type claimSource struct {
+	Endpoint    string `json:"endpoint"`
+	AccessToken string `json:"access_token,omitempty"`
+}
+
+type azureGraphGroups struct {
+	OdataNextLink string `json:"@odata.nextLink,omitempty"`
+	Value         []struct {
+		DisplayName string `json:"displayName"`
+		ID          string `json:"id"`
+	} `json:"value"`
+}
 
 // Filter parameter:
 //
@@ -57,11 +86,18 @@ const (
 	paramSubdomainsToRemove
 )
 
+type OidcOptions struct {
+	MaxIdleConns int
+	Timeout      time.Duration
+	Tracer       opentracing.Tracer
+}
+
 type (
 	tokenOidcSpec struct {
 		typ             roleCheckType
 		SecretsFile     string
 		secretsRegistry secrets.EncrypterCreator
+		options         OidcOptions
 	}
 
 	tokenOidcFilter struct {
@@ -79,6 +115,7 @@ type (
 		compressor         cookieCompression
 		upstreamHeaders    map[string]string
 		subdomainsToRemove int
+		oidcOptions        OidcOptions
 	}
 
 	tokenContainer struct {
@@ -98,21 +135,36 @@ type (
 	}
 )
 
-// NewOAuthOidcUserInfos creates filter spec which tests user info.
+// NewOAuthOidcUserInfosWithOptions creates filter spec which tests user info.
+func NewOAuthOidcUserInfosWithOptions(secretsFile string, secretsRegistry secrets.EncrypterCreator, o OidcOptions) filters.Spec {
+	return &tokenOidcSpec{typ: checkOIDCUserInfo, SecretsFile: secretsFile, secretsRegistry: secretsRegistry, options: o}
+}
+
+// Deprecated: use NewOAuthOidcUserInfosWithOptions instead.
 func NewOAuthOidcUserInfos(secretsFile string, secretsRegistry secrets.EncrypterCreator) filters.Spec {
-	return &tokenOidcSpec{typ: checkOIDCUserInfo, SecretsFile: secretsFile, secretsRegistry: secretsRegistry}
+	return NewOAuthOidcUserInfosWithOptions(secretsFile, secretsRegistry, OidcOptions{})
 }
 
-// NewOAuthOidcAnyClaims creates a filter spec which verifies that the token
+// NewOAuthOidcAnyClaimsWithOptions creates a filter spec which verifies that the token
 // has one of the claims specified
-func NewOAuthOidcAnyClaims(secretsFile string, secretsRegistry secrets.EncrypterCreator) filters.Spec {
-	return &tokenOidcSpec{typ: checkOIDCAnyClaims, SecretsFile: secretsFile, secretsRegistry: secretsRegistry}
+func NewOAuthOidcAnyClaimsWithOptions(secretsFile string, secretsRegistry secrets.EncrypterCreator, o OidcOptions) filters.Spec {
+	return &tokenOidcSpec{typ: checkOIDCAnyClaims, SecretsFile: secretsFile, secretsRegistry: secretsRegistry, options: o}
 }
 
-// NewOAuthOidcAllClaims creates a filter spec which verifies that the token
+// Deprecated: use NewOAuthOidcAnyClaimsWithOptions instead.
+func NewOAuthOidcAnyClaims(secretsFile string, secretsRegistry secrets.EncrypterCreator) filters.Spec {
+	return NewOAuthOidcAnyClaimsWithOptions(secretsFile, secretsRegistry, OidcOptions{})
+}
+
+// NewOAuthOidcAllClaimsWithOptions creates a filter spec which verifies that the token
 // has all the claims specified
+func NewOAuthOidcAllClaimsWithOptions(secretsFile string, secretsRegistry secrets.EncrypterCreator, o OidcOptions) filters.Spec {
+	return &tokenOidcSpec{typ: checkOIDCAllClaims, SecretsFile: secretsFile, secretsRegistry: secretsRegistry, options: o}
+}
+
+// Deprecated: use NewOAuthOidcAllClaimsWithOptions instead.
 func NewOAuthOidcAllClaims(secretsFile string, secretsRegistry secrets.EncrypterCreator) filters.Spec {
-	return &tokenOidcSpec{typ: checkOIDCAllClaims, SecretsFile: secretsFile, secretsRegistry: secretsRegistry}
+	return NewOAuthOidcAllClaimsWithOptions(secretsFile, secretsRegistry, OidcOptions{})
 }
 
 // CreateFilter creates an OpenID Connect authorization filter.
@@ -198,6 +250,7 @@ func (s *tokenOidcSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 		encrypter:          encrypter,
 		compressor:         newDeflatePoolCompressor(flate.BestCompression),
 		subdomainsToRemove: subdomainsToRemove,
+		oidcOptions:        s.options,
 	}
 
 	// user defined scopes
@@ -820,6 +873,10 @@ func (f *tokenOidcFilter) tokenClaims(ctx filters.FilterContext, oauth2Token *oa
 		return nil, "", requestErrorf("claims do not contain sub")
 	}
 
+	if err = f.handleDistributedClaims(r.Context(), idToken, oauth2Token, tokenMap); err != nil {
+		return nil, "", requestErrorf("failed to handle distributed claims: %v", err)
+	}
+
 	return tokenMap, sub, nil
 }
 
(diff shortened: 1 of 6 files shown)


universam1:
@szuecs marked a couple of open questions with TODO if you like to comment?

filters/auth/oidc.go
```diff
@@ -660,6 +690,122 @@ func (f *tokenOidcFilter) callbackEndpoint(ctx filters.FilterContext) {
...
+		return nil
+	}
+	claimMapSources := make(map[string]claimSource, len(distributedClaims.ClaimNames))
+	for claim, pointer := range distributedClaims.ClaimNames {
+		claimMapSources[claim] = distributedClaims.ClaimSources[pointer]
+	}
+
+	// TODO: what options should we implement?
+	// net.Options{
+	client := snet.NewClient(snet.Options{IdleConnTimeout: 2 * time.Second})
```
szuecs, line 733: client should be part of the filter or re-use authClient, if possible

filters/auth/oidc.go
```diff
@@ -38,8 +40,31 @@ const (
 	stateValidity       = 1 * time.Minute
 	oidcInfoHeader      = "Skipper-Oidc-Info"
 	cookieMaxSize       = 4093 // common cookie size limit http://browsercookielimits.squawky.net/
+
+	// Deprecated: The host of the Azure Active Directory (AAD) graph API
+	azureADGraphHost = "graph.windows.net"
```
szuecs, line 45: not sure if we need specific Azure/Microsoft related things.
What does it provide?
If it's required (looks like, but want to be sure), then please refactor code to have these separated in functions.

universam1: That's unfortunate that we need some customization for Azure. 
1. The endpoint returned by Azure should not be used according to docs https://docs.microsoft.com/en-us/graph/migrate-azure-ad-graph-overview and seeing that in other implementations 
https://github.com/grafana/grafana/blob/main/pkg/login/social/azuread_oauth.go#L183
https://github.com/hashicorp/vault-plugin-auth-jwt/blob/b724ae7ef01864b4f931814d75b875252076936e/provider_azure.go#L112-L120
2. The legacy endpoint does not provide useful information by just returning the GUID instead of group names, hence we have to use another query.

That's certainly the unpleasant part of this PR, happy for ideas how to account for such customizations.

filters/auth/oidc.go
universam1: By using the current access token for the (recursive) calls we'll have only a one time usage per URI + token AFAIK. I believe we would clash creds if we store the authClient per URI!?

szuecs: yeah or just add the client to the filter such that you have it already instantiated.

filters/auth/oidc.go
szuecs: Then let's have it, but create a func that patches the settings it needs so we do no pollute too much the main func.

ripx80: After some research we found out that not many services use this oidc function. 
For now we can implement a azure specific one and can let it open for the future to write generic code for more providers because the return json/struct is not strict specified in spec.

Now: azure specific func with no main pollution :-)
Future: use a generic provider with a interface

Is that ok for you? @szuecs and @universam1

universam1: Thanks for the research @ripx80

szuecs: I don't really get it. :) 
I think types and const are fine like this.

One issue is `snet.NewClient`, which I pointed out should be only instantiated once for every url target host, see https://github.com/zalando/skipper/pull/1967 and https://github.com/zalando/skipper/pull/1962/files#r810945168

and then there is the issue ` if uri.Host == azureADGraphHost { ... }`, which can be just refactored to a func instead of the if body.
I think `resolveDistributedClaim()` could be renamed to  `resolveDistributedClaimAzure()` and I would say good enough, if you add a test case, which might need some work to create the distributed claim targets.

universam1: Thanks @szuecs  for the detailed specs, that's certainly helpful and sounds reasonable.

Our question was, can we implement a generic, Azure agnostic implementation as well... It turns out that neither from the specs nor from other known IdPs we can discover a concrete implementation of distributed-claims response, therefore only Azure support is what we can offer for now. Maybe leaving for the future extension to support other provider that might become of interest?

szuecs: sgtm!

filters/auth/oidc.go
```diff
@@ -38,8 +40,31 @@ const (
...
+	ClaimSources map[string]claimSource `json:"_claim_sources"`
+}
+
+type claimSource struct {
+	Endpoint    string `json:"endpoint"`
+	AccessToken string `json:"access_token,omitempty"`
+}
+
+// The host of the Microsoft Graph API
+var microsoftGraphHost = "graph.microsoft.com" // global for testing purposes
```
universam1, line 59: not a particular nice solution but @szuecs no good idea how to intercept that API call with the unit test...

filters/auth/oidc.go
universam1: implemented

filters/auth/oidc.go
universam1: implemented

universam1:
@szuecs  ready for review, thanks

filters/auth/oidc.go
```diff
@@ -99,20 +126,20 @@ type (
...
+func NewOAuthOidcAnyClaims(secretsFile string, secretsRegistry secrets.EncrypterCreator, o TokenintrospectionOptions) filters.Spec {
+	return &tokenOidcSpec{typ: checkOIDCAnyClaims, SecretsFile: secretsFile, secretsRegistry: secretsRegistry, options: o}
 }
 
 // NewOAuthOidcAllClaims creates a filter spec which verifies that the token
 // has all the claims specified
-func NewOAuthOidcAllClaims(secretsFile string, secretsRegistry secrets.EncrypterCreator) filters.Spec {
-	return &tokenOidcSpec{typ: checkOIDCAllClaims, SecretsFile: secretsFile, secretsRegistry: secretsRegistry}
+func NewOAuthOidcAllClaims(secretsFile string, secretsRegistry secrets.EncrypterCreator, o TokenintrospectionOptions) filters.Spec {
+	return &tokenOidcSpec{typ: checkOIDCAllClaims, SecretsFile: secretsFile, secretsRegistry: secretsRegistry, options: o}
```
szuecs, line 142: This is breaking change to library Users. I think it would be better to create:
`NewOAuthOidc*WithOptions` instead and change to use it in skipper.go and mark the other as deprecated and move all code (tests) in this repository to use the new functions instead.

szuecs (changes requested):
some small refactorings to not do a breaking change for library users and omit breaking changes in the future

filters/auth/oidc.go
```diff
@@ -860,6 +900,120 @@ func (f *tokenOidcFilter) getTokenWithExchange(state *OauthState, ctx filters.Fi
...
+	idToken, err := f.verifier.Verify(ctx, oidcIDToken)
+	if err != nil {
+		return fmt.Errorf("failed to verify id token: %w", err)
+	}
+
+	// https://github.com/coreos/go-oidc/issues/171#issuecomment-1044286153
+	var distClaims distributedClaims
+	err = idToken.Claims(&distClaims)
+	if err != nil || distClaims.ClaimNames == nil || distClaims.ClaimSources == nil {
+		log.Debug("No distributed claims found: %w", err)
```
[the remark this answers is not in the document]
universam1: right thanks!

skipper.go
```diff
@@ -1317,9 +1317,9 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 		auth.TokenintrospectionWithOptions(auth.NewSecureOAuthTokenintrospectionAnyKV, tio),
 		auth.TokenintrospectionWithOptions(auth.NewSecureOAuthTokenintrospectionAllKV, tio),
 		auth.WebhookWithOptions(who),
-		auth.NewOAuthOidcUserInfos(o.OIDCSecretsFile, o.SecretsRegistry),
-		auth.NewOAuthOidcAnyClaims(o.OIDCSecretsFile, o.SecretsRegistry),
-		auth.NewOAuthOidcAllClaims(o.OIDCSecretsFile, o.SecretsRegistry),
+		auth.NewOAuthOidcUserInfos(o.OIDCSecretsFile, o.SecretsRegistry, tio),
+		auth.NewOAuthOidcAnyClaims(o.OIDCSecretsFile, o.SecretsRegistry, tio),
+		auth.NewOAuthOidcAllClaims(o.OIDCSecretsFile, o.SecretsRegistry, tio),
```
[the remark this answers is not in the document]
universam1: Q: when we define `tio` (aka  `OidcOptions`) wouldn't the condition always be true and never reach the legacy methods?

filters/auth/oidc.go
```diff
@@ -62,23 +87,25 @@ type (
 		typ             roleCheckType
 		SecretsFile     string
 		secretsRegistry secrets.EncrypterCreator
+		options         TokenintrospectionOptions
```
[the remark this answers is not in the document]
universam1: done

filters/auth/oidc.go
```diff
@@ -860,6 +900,120 @@ func (f *tokenOidcFilter) getTokenWithExchange(state *OauthState, ctx filters.Fi
...
+				return fmt.Errorf("oid claim not found in claims map")
+			}
+
+			v, err := f.handleDistributedClaimsAzure(uri, oauth2Token, oid)
+			if err != nil {
+				return fmt.Errorf("failed to get distributed Azure claim: %w", err)
+			}
+			claimsMap[claim] = v
+		default:
+			return fmt.Errorf("unsupported distributed claims endpoint: %s", uri.Host)
```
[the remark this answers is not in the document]
universam1: good idea, done

filters/auth/oidc.go
universam1: implemented

skipper.go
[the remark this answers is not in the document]
szuecs: I think it needs a change in https://github.com/zalando/skipper/blob/master/skipper.go#L1320-L1322 maybe wrap it similar to https://github.com/zalando/skipper/blob/master/skipper.go#L1318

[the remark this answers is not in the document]
szuecs: perfect

szuecs (approved):

filters/auth/oidc.go
```diff
@@ -198,6 +247,14 @@ func (s *tokenOidcSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 		encrypter:          encrypter,
 		compressor:         newDeflatePoolCompressor(flate.BestCompression),
 		subdomainsToRemove: subdomainsToRemove,
+		distributedClaimsClient: snet.NewClient(snet.Options{
```
AlexanderYastrebov, line 250: This will leak goroutines, we need to use static client or cache clients similar to https://github.com/zalando/skipper/pull/1967

filters/auth/oidc.go
```diff
@@ -860,6 +922,120 @@ func (f *tokenOidcFilter) getTokenWithExchange(state *OauthState, ctx filters.Fi
...
+// 	   }
+//   }
+// }
+//
+func (f *tokenOidcFilter) handleDistributedClaims(ctx context.Context, oidcIDToken string, oauth2Token *oauth2.Token,
+	claimsMap map[string]interface{}) error {
+	idToken, err := f.verifier.Verify(ctx, oidcIDToken)
+	if err != nil {
+		return fmt.Errorf("failed to verify id token: %w", err)
+	}
```
AlexanderYastrebov, line 948: Why do we validate token again? I think by this point token was already verified by `f.tokenClaims`, wasn't it?

filters/auth/oidc.go
```diff
@@ -860,6 +922,120 @@ func (f *tokenOidcFilter) getTokenWithExchange(state *OauthState, ctx filters.Fi
...
+
+	// https://github.com/coreos/go-oidc/issues/171#issuecomment-1044286153
+	var distClaims distributedClaims
+	err = idToken.Claims(&distClaims)
+	if err != nil || distClaims.ClaimNames == nil || distClaims.ClaimSources == nil {
+		log.Debugf("No distributed claims found: %v", err)
+		return nil
+	}
```
AlexanderYastrebov, line 956: It is better to handle error and invalid value cases separately. Also why error is swallowed here?

filters/auth/oidc.go
```diff
@@ -860,6 +922,120 @@ func (f *tokenOidcFilter) getTokenWithExchange(state *OauthState, ctx filters.Fi
...
+				return fmt.Errorf("oid claim not found in claims map")
+			}
+
+			v, err := f.handleDistributedClaimsAzure(uri, oauth2Token, oid)
+			if err != nil {
+				return fmt.Errorf("failed to get distributed Azure claim: %w", err)
+			}
+			claimsMap[claim] = v
+		default:
+			return fmt.Errorf("unsupported distributed claims endpoint '%s', please create an issue at https://github.com/zalando/skipper/issues/new/choose", uri.Host)
```
AlexanderYastrebov, line 981: Should this rather be a warning and ignoring claim that is not supported?
Also IMO link to the issues is redundant, we don't do it for other errors

filters/auth/oidc.go
```diff
@@ -618,6 +675,11 @@ func (f *tokenOidcFilter) callbackEndpoint(ctx filters.FilterContext) {
 		}
 	}
 
+	err = f.handleDistributedClaims(r.Context(), oidcIDToken, oauth2Token, claimsMap)
+	if err != nil {
+		log.Error(err)
+	}
+
```
AlexanderYastrebov, line 682: Lets move this call into `f.tokenClaims()` as this is a natural extension to it

filters/auth/oidc.go
```diff
@@ -860,6 +922,120 @@ func (f *tokenOidcFilter) getTokenWithExchange(state *OauthState, ctx filters.Fi
...
+		return nil
+	}
+
+	claimMapSources := make(map[string]claimSource, len(distClaims.ClaimNames))
+	for claim, pointer := range distClaims.ClaimNames {
+		claimMapSources[claim] = distClaims.ClaimSources[pointer]
+	}
+
+	for claim, source := range claimMapSources {
+		uri, err := url.Parse(source.Endpoint)
```
AlexanderYastrebov, line 964: Do we need this extra `claimMapSources`+loop? It also assumes `ClaimSources` always has a `pointer` key without checking. Maybe it should be like this:
```go
for claim, sourceName := range distClaims.ClaimNames {
  source, ok := distClaims.ClaimSources[sourceName]
  if !ok {
      return fmt.Errorf("invalid distributed claims: missing claim source for %s", claim)
  }
  uri, err := url.Parse(source.Endpoint)
  ...
}
```

filters/auth/oidc.go
```diff
@@ -860,6 +922,120 @@ func (f *tokenOidcFilter) getTokenWithExchange(state *OauthState, ctx filters.Fi
...
+		if err != nil {
+			return fmt.Errorf("failed to parse distributed claim endpoint: %w", err)
+		}
+		switch uri.Host {
+		case azureADGraphHost, microsoftGraphHost:
+			oid, ok := claimsMap["oid"].(string)
+			if !ok {
+				return fmt.Errorf("oid claim not found in claims map")
+			}
```
AlexanderYastrebov, line 973: Lets move this into `handleDistributedClaimsAzure` as the requirement for `oid` claim is azure specific

filters/auth/oidc.go
AlexanderYastrebov: Also I think its better to check `len(distClaims.ClaimNames) == 0` that would cover both nil and empty map cases.

filters/auth/oidc.go
universam1: @AlexanderYastrebov If so that would mean GC is not able to purge the filter instance once released - then other resources would be leaking as well

filters/auth/oidc.go
universam1: SGTM

filters/auth/oidc.go
universam1: SGTM

filters/auth/oidc.go
universam1: SGTM

filters/auth/oidc.go
universam1: SGTM

filters/auth/oidc.go
universam1: also obsolete now

filters/auth/oidc.go
AlexanderYastrebov: This goroutine https://github.com/zalando/skipper/blob/2c79795de8f376d34e5b4fecdcc58718cd90aeb6/net/httpclient.go#L282-L291 will never finish because we do not call `Close()` on the client so the number of goroutines will steadily grow over time due to route updates.
There is #202 to support filter resource cleanup but it is not fixed yet. The workaround for now is to cache clients per url like it is done in tokeninfo or webhook filters.

universam1: Thanks for the context, that is indeed a leak.

So far I can't see how the caching of the client per URL is applicable here, since every URL will most likely be a user-unique call, plus the credentials might clash when it is global...
Storing the snet.client as a global variable sounds like the only option?

szuecs: it's cached by host

universam1: Sorry guess I'm not getting it :) @szuecs  but that does not hold true at least for the tokeninfo where the URL including path is the key:

```go
	if ac, ok = tokeninfoAuthClient[s.options.URL]; !ok {
		ac, err = newAuthClient(s.options.URL, ...
```
https://github.com/zalando/skipper/blob/master/filters/auth/tokeninfo.go#L183-L184

But if you asking for this pattern we can do that here

szuecs: yes, because there is only one tokeninfo URL per skipper.
For OIDC I would do one per host, so parse URL string to URL and use the url.Host as a key to cache the client.

universam1: Ha good point, gotcha!
Implemented, please review

filters/auth/oidc.go
```diff
@@ -38,8 +41,33 @@ const (
 	stateValidity       = 1 * time.Minute
 	oidcInfoHeader      = "Skipper-Oidc-Info"
 	cookieMaxSize       = 4093 // common cookie size limit http://browsercookielimits.squawky.net/
+
+	// Deprecated: The host of the Azure Active Directory (AAD) graph API
+	azureADGraphHost = "graph.windows.net"
 )
 
+var distributedClaimsClients = sync.Map{}
```
universam1, line 49: Because of runtime host eval here I think we need to be thread safe

filters/auth/oidc.go
```diff
@@ -860,6 +916,137 @@ func (f *tokenOidcFilter) getTokenWithExchange(state *OauthState, ctx filters.Fi
...
+	client, ok := cli.(*snet.Client)
+	if !ok {
+		return nil, fmt.Errorf("invalid distributed claims client type")
+	}
+
+	res, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("unable to call API: %w", err)
+	}
+	body, err := ioutil.ReadAll(res.Body)
```
szuecs, line 1019: `ioutil` -> `io` since Go 1.16

filters/auth/oidc.go
```diff
@@ -860,6 +916,137 @@ func (f *tokenOidcFilter) getTokenWithExchange(state *OauthState, ctx filters.Fi
...
+		TLSHandshakeTimeout:     f.oidcOptions.Timeout,
+		MaxIdleConnsPerHost:     f.oidcOptions.MaxIdleConns,
+		Tracer:                  f.oidcOptions.Tracer,
+		OpentracingComponentTag: "skipper",
+		OpentracingSpanName:     "distributedClaims",
+	}))
+
+	client, ok := cli.(*snet.Client)
+	if !ok {
+		return nil, fmt.Errorf("invalid distributed claims client type")
```
szuecs, line 1012: `fmt.Errorf` -> `errors.New()` in case no formatstring needed

skipper.go
```diff
@@ -1298,6 +1298,12 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 		Tracer:       tracer,
 	}
 
+	oo := auth.OidcOptions{
+		Timeout:      o.OAuthTokenintrospectionTimeout,
```
szuecs, line 1302: This sounds a bit odd to re-use. I think we should add another option for this OpenID Connect config.

filters/auth/oidc.go
szuecs: The host is not known on filter creation as far as I understand, so seems to be right to use sync.Map or a Mutex, yes

skipper.go
universam1: SGTM

filters/auth/oidc.go
universam1: right thx

skipper.go
universam1: done please review @szuecs

szuecs:
:+1:

szuecs:
I think you @universam1 are done to merge here right?
@AlexanderYastrebov seems to be fine form my side

universam1:
> I think you @universam1 are done to merge here right? @AlexanderYastrebov seems to be fine form my side

Yes, thanks for asking - we are running this version in production clusters since some time - awaiting the release. Thanks

filters/auth/oidc.go
```diff
@@ -795,6 +848,10 @@ func (f *tokenOidcFilter) tokenClaims(ctx filters.FilterContext, oauth2Token *oa
 		return nil, "", requestErrorf("claims do not contain sub")
 	}
 
+	if err = f.handleDistributedClaims(r.Context(), idToken, oauth2Token, tokenMap); err != nil {
+		log.Error(err)
```
AlexanderYastrebov, line 852: Why don't we fail here on error? IMO in the auth related code it should be strict

filters/auth/oidc.go
```diff
@@ -860,6 +917,137 @@ func (f *tokenOidcFilter) getTokenWithExchange(state *OauthState, ctx filters.Fi
...
+	oauth2Token.SetAuthHeader(req)
+
+	cli, _ := distributedClaimsClients.LoadOrStore(url.Host, snet.NewClient(snet.Options{
+		ResponseHeaderTimeout:   f.oidcOptions.Timeout,
+		TLSHandshakeTimeout:     f.oidcOptions.Timeout,
+		MaxIdleConnsPerHost:     f.oidcOptions.MaxIdleConns,
+		Tracer:                  f.oidcOptions.Tracer,
+		OpentracingComponentTag: "skipper",
+		OpentracingSpanName:     "distributedClaims",
+	}))
```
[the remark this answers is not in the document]
szuecs: @AlexanderYastrebov yes true snet.NewClient should not be called everytime. 
What's your reasoning to store client per url?
disclaimer: I don't think it makes sense per url, because TCP connection pools will resolve DNS and connect to the resolved IPs, whatever url you have. So it makes sense to have per host, because of this mapping. If people will hit a limit of connections that creates latency they can increase MaxIdleConnsPerHost and call it a day.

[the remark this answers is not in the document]
universam1: Don't think storing on URL key would be a good idea as URLs are usually one-time use, which essentially results into a  leak.

filters/auth/oidc.go
universam1: ok

filters/auth/oidc.go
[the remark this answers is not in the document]
universam1: I split the LoadAndRestore into conditioned sections to prevent creation of a client instance.

I'd prefer to keep using the `sync.Map` here as it is faster and well tested:

> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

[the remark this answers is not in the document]
AlexanderYastrebov: > What's your reasoning to store client per url?

I think we do per URL elsewhere e.g. for tokeninfo clients. 
Lets store per host given your pros arguments.

> I split the LoadAndRestore into conditioned sections to prevent creation of a client instance.

Which introduces a race condition. When there is no mapping stored it could happen that two request create a client each, one will be stored and another will leak. Presence check and addition should be done while the map is locked for updates - that is `LoadOrStore` for (unfortunately it lacks capability of lazy value initialization https://github.com/golang/go/issues/44159)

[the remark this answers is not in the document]
AlexanderYastrebov: To avoid leaking the client I think we would need something like:
```go
	cli, ok := distributedClaimsClients.Load(url.Host)
	if !ok {
		newCli = initClient()
		cli, loaded = distributedClaimsClients.LoadOrStore(url.Host, newCli)
		if loaded {
			newCli.Close()
		}
	}
```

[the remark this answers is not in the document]
universam1: > To avoid leaking the client I think we would need something like:

True, that is the proper way, thanks.

> I think we do per URL elsewhere e.g. for tokeninfo clients.
Lets store per host given your pros arguments.

Thinking twice, why do we store the client as instances (per host) at the first place, as we do not have any host part stored with it? Logically, every client instance is going to be equal, wouldn't a single static client be just fine? @AlexanderYastrebov @szuecs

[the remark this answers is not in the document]
szuecs: In case of oidc it should be fine anyways. I am not sure anymore without reading the code why we have a client per url/host in case of tokeninfo/tokenintrospection , I think it's because we have low latency conditions (or it's about reloading configuration) and oidc is pretty much slow anyways (2-4 requests with redirects for one flow), so it doesn't really matter.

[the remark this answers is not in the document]
universam1: sorry the delay!
Thanks for the context @szuecs - then it should not hurt to keep it like it is.

@AlexanderYastrebov implemented your suggestion, please check

universam1:
@szuecs @AlexanderYastrebov Kindly asking for merging of this feature, we like to return to upstream for all of our clusters asap.
These changes have been tested extensively on some production clusters for weeks and we have positive results.

Thanks for considering a release!

szuecs:
:+1:

szuecs:
Thanks @universam1 !

PR #1965, opened 2022-02-22
clear invalid cookies

universam1:
If OIDC cookies are invalid, the client is redirected but the cookies are not cleared.

In case that the new cookie will produce smaller set of chunks, an invalid chunk is not overwritten and kept in the browser and Skipper will try to attach this junk to the new set.

This changes the behaviour in that regard that it clears relevant cookies in case of a new auth session.

### DIFF
--- a/filters/auth/oidc.go
+++ b/filters/auth/oidc.go
@@ -327,7 +327,7 @@ func (f *tokenOidcFilter) internalServerError(ctx filters.FilterContext) {
 // https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowSteps
 // 1. Client prepares an Authentication Request containing the desired request parameters.
 // 2. Client sends the request to the Authorization Server.
-func (f *tokenOidcFilter) doOauthRedirect(ctx filters.FilterContext, cookies ...*http.Cookie) {
+func (f *tokenOidcFilter) doOauthRedirect(ctx filters.FilterContext, cookies []*http.Cookie) {
 	nonce, err := f.encrypter.CreateNonce()
 	if err != nil {
 		log.Errorf("Failed to create nonce: %v.", err)
@@ -404,7 +404,7 @@ func getHost(request *http.Request) string {
 	}
 }
 
-func createOidcCookie(name string, value string, maxAge int, domain string) (cookie *http.Cookie) {
+func (f *tokenOidcFilter) createOidcCookie(ctx filters.FilterContext, name string, value string, maxAge int) (cookie *http.Cookie) {
 	return &http.Cookie{
 		Name:     name,
 		Value:    value,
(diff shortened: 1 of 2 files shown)


filters/auth/oidc.go
```diff
@@ -698,7 +701,20 @@ func (f *tokenOidcFilter) Request(ctx filters.FilterContext) {
...
+		for i, c := range cookies {
+			purgeCookies[i] = &http.Cookie{
+				Name:     c.Name,
+				Value:    "",
+				Path:     "/",
+				Domain:   extractDomainFromHost(ctx.Request().Host, 1),
+				MaxAge:   -1,
+				Secure:   true,
+				HttpOnly: true,
+			}
```
AlexanderYastrebov, line 715: Why not copy everything from `c` and set empty `Value` and `MaxAge`?
Or maybe even modify `cookies` values in-place and it to `f.doOauthRedirect`?..

universam1: The [request cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie) contains only `Name` and `Value`. In order to clear we need empty Value, Path, Domain, MaxAge and maybe Secure and HttpOnly. Sure we can reuse the cookie instance but only the Name is relevant the rest we have to set

universam1:
@szuecs think we are ready for review

icereed:
Found the bug. Since the original cookie was passed by-reference it was modified unintendedly.
By using now the pass-by-value approach (like on upstream) this is avoided.
The tests work now again locally.

filters/auth/oidc.go
```diff
@@ -698,7 +713,12 @@ func (f *tokenOidcFilter) Request(ctx filters.FilterContext) {
 			return
 		}
 		// 1. Client prepares an Authentication Request containing the desired request parameters.
-		f.doOauthRedirect(ctx)
+		// clear existing, invalid cookies
+		var purgeCookies = make([]*http.Cookie, len(cookies))
+		for i, c := range cookies {
+			purgeCookies[i] = deleteOidcCookie(c.Name, extractDomainFromHost(ctx.Request().Host, 1))
```
AlexanderYastrebov, line 719: `f.subdomainsToRemove` is not considered here

filters/auth/oidc.go
```diff
@@ -401,30 +404,46 @@ func getHost(request *http.Request) string {
 	}
 }
 
-func chunkCookie(cookie http.Cookie) (cookies []http.Cookie) {
+func createOidcCookie(name string, value string, maxAge int, domain string) (cookie *http.Cookie) {
```
AlexanderYastrebov, line 407: I suggest we make this a method of a filter and derive domain using `extractDomainFromHost(getHost(request), f.subdomainsToRemove))` instead of passing it

szuecs:
There are some open comments @universam1 .
Can you address them?

filters/auth/oidc.go
icereed: Fixed that in https://github.com/zalando/skipper/pull/1965/commits/39c16f047e62edf7b25f8690565e2ca46ddda2f4

filters/auth/oidc.go
icereed: Fixed that in https://github.com/zalando/skipper/pull/1965/commits/39c16f047e62edf7b25f8690565e2ca46ddda2f4

szuecs:
@icereed Alexander is in vacation so maybe it will take until next week to get an answer from him.

filters/auth/oidc.go
```diff
@@ -327,7 +327,7 @@ func (f *tokenOidcFilter) internalServerError(ctx filters.FilterContext) {
 // https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowSteps
 // 1. Client prepares an Authentication Request containing the desired request parameters.
 // 2. Client sends the request to the Authorization Server.
-func (f *tokenOidcFilter) doOauthRedirect(ctx filters.FilterContext) {
+func (f *tokenOidcFilter) doOauthRedirect(ctx filters.FilterContext, cookies ...*http.Cookie) {
```
AlexanderYastrebov, line 330: Why vararg and not a slice?
What is the point of using a slice of pointers instead of values?

icereed: Thanks for the hint regarding vararg. I replaced it with `[]*http.Cookie`.

The reason for using `[]*http.Cookie` instead of `[]http.Cookie` is the convention/example found in the golang `net/http` package:
* [`func (r *Request) Cookie(name string) (*Cookie, error)`](https://pkg.go.dev/net/http#Request.Cookie)
* [`func (r *Request) Cookies() []*Cookie`](https://pkg.go.dev/net/http#Request.Cookies)

Maybe it makes sense to follow that example to be closer to the `net/http` package? What do you think?

filters/auth/oidc.go
```diff
@@ -401,30 +404,46 @@ func getHost(request *http.Request) string {
...
+		MaxAge:   maxAge,
+		Domain:   extractDomainFromHost(getHost(ctx.Request()), f.subdomainsToRemove),
+	}
+}
+
+func (f *tokenOidcFilter) deleteOidcCookie(ctx filters.FilterContext, name string) (cookie *http.Cookie) {
+	return f.createOidcCookie(ctx, name, "", -1)
+}
+
+func chunkCookie(cookie http.Cookie) (cookies []*http.Cookie) {
```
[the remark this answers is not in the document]
icereed: Agree. I replaced all values with pointers to be closer to the golang `net/http` definition of [`func (r *Request) Cookies() []*Cookie#`](https://pkg.go.dev/net/http#Request.Cookies).

AlexanderYastrebov:
:+1:

szuecs:
:+1:

szuecs:
Thanks! Great work!

Issue #1968, opened 2022-02-24
Crashlooping on missing claim argument in oauthOidc*Claims filter


TBeijen:
**Describe the bug**
Skipper starts crashlooping when filter `oauthOidcAllClaims` is misconfigured by omitting the claims argument. This probably affects all `oauthOidc*` filters.

Running in Kubernetes, skipper version: `v0.13.167`

**To Reproduce**
Add a filter that's incomplete:

```
# Incomplete filter, missing the claims argument
zalando.org/skipper-filter: oauthOidcAllClaims("https://my-org.onelogin.com/oidc/2", "cliendId", "clientSecret", "https://myapp.mydomain.com/auth/provider/callback", "email name")

# Correct would be:
zalando.org/skipper-filter: oauthOidcAllClaims("https://my-org.onelogin.com/oidc/2", "cliendId", "clientSecret", "https://myapp.mydomain.com/auth/provider/callback", "email name", "email")
```

This results in panic errors in skipper like this:

```
panic: runtime error: index out of range [5] with length 5
goroutine 41 [running]:
github.com/zalando/skipper/filters/auth.(*tokenOidcSpec).CreateFilter(0xc0003d7050, {0xc000410000, 0xc0004125a0, 0x12})
	/workspace/filters/auth/oidc.go:212 +0x12de
github.com/zalando/skipper/routing.createFilter(0xc0004608c8, 0xc0004b2c00, 0xc0000dc590)
	/workspace/routing/datasource.go:225 +0x298
github.com/zalando/skipper/routing.createFilters(0xc0001f9a01, {0xc000406278, 0x1, 0x864226a12993958c}, 0x0)
	/workspace/routing/datasource.go:237 +0xb1
github.com/zalando/skipper/routing.processRouteDef(0xc0002b8dc0, 0x16, 0xc0004b4a20)
	/workspace/routing/datasource.go:460 +0x95
github.com/zalando/skipper/routing.processRouteDefs({0xc0003d7260, 0x0, 0xb2d05e00, {0xc0003b3440, 0x1, 0x1}, {0xc0002b8dc0, 0x16, 0x16}, 0x20, ...}, ...)
	/workspace/routing/datasource.go:497 +0x105
github.com/zalando/skipper/routing.receiveRouteMatcher({0xc0003d7260, 0x0, 0xb2d05e00, {0xc0003b3440, 0x1, 0x1}, {0xc0002b8dc0, 0x16, 0x16}, 0x20, ...}, ...)
[...]

PR #1969, opened 2022-02-24
fix: oidc*Claims filter paniced if no claims were used as input argument

szuecs:
fix https://github.com/zalando/skipper/issues/1968

### DIFF
--- a/filters/auth/oidc.go
+++ b/filters/auth/oidc.go
@@ -129,7 +129,7 @@ func (s *tokenOidcSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 	if err != nil {
 		return nil, err
 	}
-	if len(sargs) < paramClaims {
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

aryszka:
:+1:

AlexanderYastrebov:
This makes user-defined claims mandatory (which was true anyway otherwise CreateFilter panics) but we can alternatively make it optional by changing the condition https://github.com/zalando/skipper/blob/505a09d4f1c981971a79f1638378e1a2166edd71/filters/auth/oidc.go#L211-L214
to 
```go
if len(sargs) > paramClaims && len(sargs[paramClaims]) > 0
```
Maybe that was the intent initially, otherwise why have `if len(sargs[paramClaims]) > 0` check at all.

PR #1972, opened 2022-02-25
filters/auth: do not close shared auth clients

AlexanderYastrebov:
* removes auth filters `Close` methods: they are never called due to #202 and
  it makes no sense to close shared auth clients anyway
* removes mention of `Close` from the developement docs

Updates #202

### DIFF
--- a/docs/reference/development.md
+++ b/docs/reference/development.md
@@ -97,7 +97,7 @@ index 10d5769..da46fe0 100644
 
 Documentation for users should be done in `docs/`.
 
-```
+````
 diff --git a/docs/filters.md b/docs/filters.md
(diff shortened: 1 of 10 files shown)


docs/reference/development.md
```diff
@@ -97,7 +97,7 @@ index 10d5769..da46fe0 100644
 
 Documentation for users should be done in `docs/`.
 
-```
+````
```
AlexanderYastrebov, line 100: Use four ticks to avoid clash with three ticks in the diff

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #1973, opened 2022-02-25
debugging in VSCode

universam1:
It can be helpful to run Skipper in a debug session locally that enables one to inspect variables and do other debugging activities in order to analyze filter and token states.

adding docs how a VSCode launch configuration for a full OIDC roundtrip could look like

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ _test_plugins_fail/*.so
 opentracingplugin/build
 build/
 skptesting/lorem.html
+.vscode/*
--- a/docs/tutorials/development.md
+++ b/docs/tutorials/development.md
@@ -42,6 +42,45 @@ curl -s http://localhost:9999/foo
 2
 ```
 
+### Debugging Skipper
+
+It can be helpful to run Skipper in a debug session locally that enables one to inspect variables and do other debugging activities in order to analyze filter and token states.
+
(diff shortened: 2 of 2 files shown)


universam1:
@szuecs as discussed

webframp:
Nice one @universam1, now you just need to figure out how to do this with emacs 😉

szuecs:
:+1:

.gitignore
```diff
@@ -25,3 +25,4 @@ _test_plugins_fail/*.so
 opentracingplugin/build
 build/
 skptesting/lorem.html
+.vscode/*
```
aryszka, line 28: i am generally rather against including editor specific references, even in .gitignore. OTOH, I can see that there are already some items that are not related to the provided source code, like .idea. @szuecs @AlexanderYastrebov WDYT?

aryszka:
Great docs! Thanks!

My note about the .gitignore is really rather just a theoretical question.

aryszka:
:+1:

.gitignore
AlexanderYastrebov: I've configured to ignore `.idea/` globally using https://stackoverflow.com/a/7335487/828328
I think this makes more sense than adding it to any new cloned project's `.gitignore`

aryszka:
let's discuss the .gitignore separately, if it's even interesting for others. Merging this PR

PR #1974, opened 2022-03-05
Ingress v1 - TLS from Spec

### LINKED ISSUE #851: Poll TLS cert and keys from directories
This could enable to read secrets from Kubernetes secrets and terminate https in skipper at dynamic workloads, where certificates/keys are created in skipper would detect these new created ones.

### LINKED ISSUE #1780: Skipper mTLS support
Hi all,

does skipper support mTLS use-case? 

Nginx ingress controller use following annotations to configure client validation:

```
nginx.ingress.kubernetes.io/auth-tls-verify-client: "on"
nginx.ingress.kubernetes.io/auth-tls-secret: "default/mycerts"
```

Is it possible to get similar with skipper?

rickhlx:
Fixes #851 
Fixes #1780

This will allow us to define TLS certs at the Ingress resource and dynamically respond with the correct cert based on the request.


rickhlx:
@szuecs would love to receive some initial feedback on this proposal

routing/routing.go
```diff
@@ -374,6 +376,21 @@ func (r *Routing) startReceivingUpdates(o Options) {
 	}()
 }
 
+// ReturnCert matches an incoming TLS client to a TLS certificate
+//
+// TODO: Refactor 
+func (r *Routing) ReturnCert(helloInfo *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	rt := r.routeTable.Load().(*routeTable)
+	for _, r := range rt.validRoutes {
+		compiledRxs := make(map[string]*regexp.Regexp)
```
rickhlx, line 385: This logic could probably go in the [routing/matcher.go](https://github.com/zalando/skipper/blob/a3a29f87a6eaed3076091f556d416a67e29b25a8/routing/matcher.go) file

skipper.go
```diff
@@ -561,6 +561,9 @@ type Options struct {
 	// TLSMinVersion to set the minimal TLS version for all TLS configurations
 	TLSMinVersion uint16
 
+	// Read TLS certificates from route table
+	TLSFromRouteTable bool
```
rickhlx, line 565: Not married to this option name, let me know if you have suggestions.

dataclients/kubernetes/clusterclient.go
```diff
@@ -430,9 +444,22 @@ func (c *clusterClient) fetchClusterState() (*clusterState, error) {
...
+				for _, sec := range ing.Spec.IngressTLS {
+					secret, err := c.loadSecret(sec.SecretName, ing.Metadata.Namespace)
+					if err != nil {
+						log.Println(err)
+						return nil, err
+					}
+					secrets = append(secrets, secret)
+				}
+			}
+		}
```
AlexanderYastrebov, line 462: * this should better reside in the separate method e.g. `c.loadTLSSecrets(ingressesV1)`
* several ingresses may reuse the same secret so the code should load each `namespace+secret name` only once
* errors should be handled gracefully instead of early return, otherwise single malformed ingress renders whole cluster non-functioning
* specific `log.Error` or `log.Warn` should be used instead of `log.Println` like elsewhere

dataclients/kubernetes/clusterstate.go
```diff
@@ -16,6 +16,7 @@ type clusterState struct {
 	services        map[definitions.ResourceID]*service
 	endpoints       map[definitions.ResourceID]*endpoint
 	cachedEndpoints map[endpointID][]string
+	secrets         []*secret
```
AlexanderYastrebov, line 19: I think it should hold unique value per namespace+name similar to services and endpoints, e.g. `secrets map[definitions.ResourceID]*secret`

dataclients/kubernetes/ingress.go
```diff
@@ -351,6 +353,29 @@ func hasCatchAllRoutes(routes []*eskip.Route) bool {
...
+		for _, secret := range ic.state.secrets {
+			if secret.Meta.Name == secretName {
+				crt, err := b64.StdEncoding.DecodeString(secret.Data["tls.crt"])
+				if err != nil {
+					return err
+				}
+				key, err := b64.StdEncoding.DecodeString(secret.Data["tls.key"])
+				if err != nil {
+					return err
+				}
```
AlexanderYastrebov, line 367: * it should check the namespace as well there might be secrets with the same name in different namespaces. I think this part will be fixed if we change cluster state to use `secrets map[definitions.ResourceID]*secret` 
* it should check that secret is well-formed (here or maybe while reading the secret), i.e. check `secret.Type == "kubernetes.io/tls"` and presence of `tls.key` in the secret data (`StdEncoding.DecodeString` decodes empty string without error).

eskip/string.go
```diff
@@ -78,9 +80,17 @@ func argsString(args []interface{}) string {
...
 
+func tlsIsEmpty(t tls.Certificate) bool {
+	return reflect.DeepEqual(t, tls.Certificate{})
+}
+
 func (r *Route) predicateString() string {
 	var predicates []string
 
+	if !tlsIsEmpty(r.TLSCert) {
+		predicates = appendFmtEscape(predicates, `TLS(True)`, `"`)
```
AlexanderYastrebov, line 91: Why do we need this? Also I do not see new `TLS` predicate being introduced.

eskip/eskip.go
```diff
@@ -335,6 +336,9 @@ type Route struct {
 
 	// Namespace is deprecated and not used.
 	Namespace string
+
+	// TLS Certificate to use for terminating TLS on the route
+	TLSCert tls.Certificate
```
AlexanderYastrebov, line 341: On every routing table update cycle dataclient uses string route representation to figure out if route has changed. In the current implementation afaict the secret would be reloaded on every cluster state update and hence `tls.Certificate` will change but it is not reflected in the string representation of the route and therefore route will not be considered as updated.

AlexanderYastrebov:
The current design sets certificate to the route which creates a problem for route update when tls secret changes and also loops over all valid routes which is completely unnecessary (e.g. only a handful of routes out of thousands may require certificate); it also couples certificate lookup to the routing. The proposal does not addess explicitly the problem of multiple ingresses defining different secrets for the same host. It also reloads secrets and re-creates certificates on every ingress poll cycle (3 seconds by default) which might be ok for the initial implementation but be suboptimal performance-wise as secrets would change less often.

I propose we think about a new component e.g. `CertificateRegistry` that would implement
* `GetCertificate(helloInfo *tls.ClientHelloInfo) (*tls.Certificate, error)` to be used by server listener
* `RegisterCertificate(host string, *tls.Certificate) error` to be used by Kubernetes data client (not sure if we need host here at all as certificate should contain Common Name already and the default certificate lookup for "static" certificates uses https://pkg.go.dev/crypto/tls#ClientHelloInfo.SupportsCertificate afaict)

It is not clear what should be the strategy to cleanup of unused certificates and if we should decouple secret/certificate update from the ingress update at this point.

Also lets check how other ingress controllers implement the lookup logic.

AlexanderYastrebov:
> The proposal does not addess explicitly the problem of multiple ingresses defining different secrets for the same host.

There is a bit if chicken-and-egg problem - server listener has to choose the certificate before routing, i.e. it does not know yet which route the request will follow and hence can not determine the ingress and its tls config that defines route.

E.g. I do not see how and if it should be possible to define two ingresses that use different certificates for the same host and different paths:
```yaml
kind: Ingress
metadata:
  name: foo
spec:
  tls:
  - hosts:
      - https-example.example.com
    secretName: foo-secret
  rules:
  - host: https-example.example.com
    http:
      paths:
      - path: /foo
        pathType: Prefix
        backend:
          service:
            name: service1
            port:
              number: 80
---
kind: Ingress
metadata:
  name: bar
spec:
  tls:
  - hosts:
      - https-example.example.com
    secretName: bar-secret # different cert
  rules:
  - host: https-example.example.com
    http:
      paths:
      - path: /bar
        pathType: Prefix
        backend:
          service:
            name: service2
            port:
              number: 80
```
It is also not clear how certificate selection should work if one ingress defines the `spec.tls` but other does not while they use the same hostname:
```yaml
kind: Ingress
metadata:
  name: foo
spec:
  tls:
  - hosts:
      - https-example.example.com
    secretName: foo-secret
  rules:
  - host: https-example.example.com
    http:
      paths:
      - path: /foo
        pathType: Prefix
        backend:
          service:
            name: service1
            port:
              number: 80
---
kind: Ingress
metadata:
  name: bar
spec:
  # no tls defined for https-example.example.com
  rules:
  - host: https-example.example.com
    http:
      paths:
      - path: /bar
        pathType: Prefix
        backend:
          service:
            name: service2
            port:
              number: 80
```

It looks like the out best shot would be to have a global certificate registry that does not account for ingress/routing and let listener to pick first matching certificate. This way in example 1 different certs for the same host would be an undefined behaviour and in example 2 the second ingress would use certificate configured by the first.

eskip/string.go
rickhlx: Resolving this since not using routes to determine TLS certificates

eskip/eskip.go
rickhlx: Resolving this since not using routes to determine TLS certificates

skipper.go
rickhlx: Resolving this since not using routes to determine TLS certificates

routing/routing.go
rickhlx: Resolving this since not using routes to determine TLS certificates

dataclients/kubernetes/ingress.go
rickhlx: Added some validations and warnings before converting secret to certificate. Still unclear how we would handle two secrets with the same name in different namespaces. Do we require the secret to be in the same namespace as the Ingress?

rickhlx:
> It looks like the out best shot would be to have a global certificate registry that does not account for ingress/routing and let listener to pick first matching certificate

I like this idea. Terminating TLS should not care what the route is, should only care if a rule's host matches the tls host and add it to the registry. 

> It is not clear what should be the strategy to cleanup of unused certificates and if we should decouple secret/certificate update from the ingress update at this point.

Could it do something similar to how routing is updated? Get a list of current, updated and deleted TLS certs and send them over to the certificate registry?

rickhlx:
> Also lets check how other ingress controllers implement the lookup logic.

afaict the NGINX Ingress controller has a similar [store component
](https://github.com/kubernetes/ingress-nginx/tree/main/internal/ingress/controller/store) which implements an [sslStore tracker](https://github.com/kubernetes/ingress-nginx/blob/main/internal/ingress/controller/store/store.go#L221).

Certificates are [generated and stored](https://github.com/kubernetes/ingress-nginx/blob/main/internal/ingress/controller/store/backend_ssl.go#L36) in this tracker to later be [read](https://github.com/kubernetes/ingress-nginx/blob/e1eff7816005b3fad21a091e306132d11e0ca61e/internal/ingress/controller/controller.go#L1315). 
The [store](https://github.com/kubernetes/ingress-nginx/blob/e1eff7816005b3fad21a091e306132d11e0ca61e/internal/ingress/controller/store/store.go) has the logic to manage old/stale certificates.

dataclients/kubernetes/ingress.go
rickhlx: Added a validation to make sure ingress and secret are in the same namespace.

rickhlx:
@AlexanderYastrebov implemented your suggestion of a Certificate Registry. Included the following proposals:
* Generate a default TLS certificate as a fallback 
* Only update a certificate when it has changed
* Return the first TLS certificate that matches a configured host
  * As per the [Ingress - TLS](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls) docs:
  > `hosts` in the `tls` section need to explicitly match the `host` in the `rules` section.
* Initialize Certificate Registry, optionally enable TLS to use registry for `GetCertificate` using the flag `tls-certificate-registry`

If this something we could get your feedback on?

AlexanderYastrebov:
@rickhlx Hello, thank you, I'll have a look once I have some spare capacity

certregistry/certregistry.go
```diff
@@ -0,0 +1,98 @@
...
+// provided key. If the object already exists it will be updated or added otherwise. Returns 
+// true if key was changed.
+func (r *CertRegistry) SyncCert(key string, hosts []string, crt *tls.Certificate) bool {
+	cert := &tlsCertificate{
+		hosts: hosts, 
+		cert: crt,
+	}	
+
+	curr, err := r.getCertByKey(key)
+	if err == nil {
```
szuecs, line 71: If we get `err != nil`, we will just go to "adding cert" case

certregistry/certregistry.go
```diff
@@ -0,0 +1,98 @@
+package certregistry
```
szuecs, line 1: I belief that all of this code should be part of the secrets module instead, maybe as submodule.

certregistry/cert.go
```diff
@@ -0,0 +1,80 @@
+package certregistry
```
szuecs, line 1: I belief that all of this code should be part of the secrets module instead, maybe as submodule.

szuecs:
@rickhlx PR looks great, just a few small things to change.

certregistry/certregistry.go
rickhlx: Right, `if err != nil` it means the cert was not found with that key so we'll add it to the registry. I could use an `else` statement instead but it seemed redundant.

rickhlx: Improved the logic here so it's more clear what happens when we return `err` or not.

certregistry/certregistry.go
rickhlx: Moved to submodule of secrets.

rickhlx:
@szuecs thanks for the review! Implemented your suggestions and fixed the linting issues. Let me know how that looks. 😸

certregistry/certregistry.go
szuecs: I am not sure if I understand the logic here. :)
Why do we add a certificate that we don't found into our registry?
Do we need to refresh it for some reason?

szuecs: If we ignore it, we should not return error but bool to highlight "update required, as you know from map lookup for example:
```
if curr, ok := r.getCertByKey(key); ok {
    if !equalCert(curr, cert) {
	log.Debugf("updating certificate in registry - %s", key)
	r.addCertToRegistry(key, cert)
        return true
    } else {
        return false
    }
}

log.Debugf("adding new certificate to registry - %s", key)
r.addCertToRegistry(key, cert)
```

szuecs:
> > It looks like the out best shot would be to have a global certificate registry that does not account for ingress/routing and let listener to pick first matching certificate
> 
> I like this idea. Terminating TLS should not care what the route is, should only care if a rule's host matches the tls host and add it to the registry.

Pick first matching certificate is not a great idea!
Best would be to find the newest, best matching certificate as we implemented in https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/certs/matching.go
For example if you have 3 certificates `foo.example.org`, `*.example.org` and SAN with  `foo.example.org` and  `bar.example.org` :
a request to `foo.example.org` you want to have the newest certificate of `foo.example.org` or SAN presented and if none of these are valid anymore the wildcard cert. 

> 
> > It is not clear what should be the strategy to cleanup of unused certificates and if we should decouple secret/certificate update from the ingress update at this point.
> 
> Could it do something similar to how routing is updated? Get a list of current, updated and deleted TLS certs and send them over to the certificate registry?

Sounds good to me, but I think if not already done cleanup can be omitted for now. If it does not leak memory for every cycle but only real update then the memory footprint is too small and lifetime of a skipper process hopefully to short to make it hurt.

rickhlx:
> Best would be to find the newest, best matching certificate as we implemented in https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/certs/matching.go

In this case, we're not matching the certificate SAN, we're matching the Ingress TLS spec. Ideally we want to return the cert that matches a host in this list, regardless of if it's the correct certificate or not. 

For example, I can say I want to use the certificate with `*.foo.com` in the SAN but use `foo.bar.com`. The browser will complain but it should be acceptable to use the certificate. 

```yaml
kind: Ingress
metadata:
  name: foo
spec:
  tls:
  - hosts:
      - https-example.example.com <----- HERE
    secretName: foo-secret
  rules:
  - host: https-example.example.com
    http:
      paths:
      - path: /foo
        pathType: Prefix
        backend:
          service:
            name: service1
            port:
              number: 80
```

This is the first match we're comparing against. If we match the certificate host or SAN we could potentially be a certificate that was not configured for the ingress.

certregistry/certregistry.go
rickhlx: I see what you mean. Updated to return `bool` instead of `error`.

rickhlx:
@szuecs thanks for the feedback! I had already started working on a way to find the best certificate if more than one matches the ingress host. Please let me know what you think. Happy to use the map you suggested too.

secrets/certregistry/certregistry.go
```diff
@@ -85,12 +99,125 @@ func (r *CertRegistry) SyncCert(key string, hosts []string, crt *tls.Certificate
...
+	switch len(certList) {
+	case 0:
+		// no certs found so returning the default certificate
+		return r.defaultTLSCert, nil
+	case 1:
+		// only one certificate found, retuning it
+		return certList[0].cert, nil
+	default:
+		// more that one cert found, returning the best match
+		log.Warnf("found more than one certificate for %s, returning best one", hello.ServerName)
```
szuecs, line 120: no warning, it's expected to have it and this would flood on every connection the logs :)

szuecs:
@rickhlx looks great to me!
I guess the map change would be a small change. If it's a small change, please do it, because it will likely safe some µs and we have enough memory, so it should be good trade.

secrets/certregistry/certregistry.go
rickhlx: The reason I added the warning is this would only ever execute if there was two different ingresses using different secrets terminating the same exact host. IMO this is most likely a configuration mistake since it would be almost impossible to determine which Ingress to use.

rickhlx:
> I guess the map change would be a small change. If it's a small change, please do it, 

Not really a small change. The lookup map would need to be updated along with every other function that's using it to sync, search. The edge case where the `getBestMatchingCertificate` is executed is limited to what I think would be a configuration mistake. (two ingresses, different secrets, same host)

secrets/certregistry/certregistry.go
szuecs: Maybe to add: we have something that generates up to hundreds of ingress from another CRD, so I think it's fine to serve multiple ingress with the same host and while rotating certs you will have a split, some ingress might be updated already and others not yet, but skipper already got all ingress configuring the routes and TLS certs/secrets...

szuecs:
Then let's just use the slice as it is and drop the log.Warnf.

secrets/certregistry/certregistry.go
rickhlx: Thanks for the clarification! That actually makes a lot of sense.

szuecs:
:+1:

szuecs:
@rickhlx can you rebase on current master please, because there was a bugfix for 1.18 and I hope these errors would be fixed by that.

szuecs:
can you do:
```
git co master
git pull --rebase
git co tls-ingress
git rebase master
git push -f rickhlx tls-ingress
```

rickhlx:
@szuecs ran tests locally with `make cicheck` successfully.

rickhlx:
@szuecs I increased the timeout on `TestInitOrderandDefault` in the `metricsinit_test.go` file. I'm not 100% sure what the implications are but this seemed to resolve the flaky tests for me.

szuecs:
Thanks @rickhlx ,  I am currently fighting with caching and GH to get the button shown to approve the test run :D

szuecs:
:+1:

dataclients/kubernetes/clusterclient.go
```diff
@@ -400,6 +405,32 @@ func (c *clusterClient) loadServices() (map[definitions.ResourceID]*service, err
...
+			hasInvalidSecret = true
+			continue
+		}
+
+		result[secret.Meta.ToResourceID()] = secret
+	}
+
+	if hasInvalidSecret {
+		log.Errorf("Invalid secret resource detected.")
+	}
```
AlexanderYastrebov, line 429: IMO this does not help to debug the problem, we either need to log (maybe warn) each invalid secret namespace+name or silently ignore them

dataclients/kubernetes/ingress.go
```diff
@@ -351,11 +360,47 @@ func hasCatchAllRoutes(routes []*eskip.Route) bool {
 	return false
 }
 
+// addHostTLSCert adds a TLS certificate to the certificate registry when the referenced secret is found
+// and is a valid TLS secret.
+func addHostTLSCert(ic ingressContext, hosts []string, secretName string, ns string) error {
+	for _, secret := range ic.state.secrets {
```
AlexanderYastrebov, line 366: `secrets` is a map, it should lookup namespace+name instead of iterating

dataclients/kubernetes/ingressv1.go
```diff
@@ -311,6 +312,26 @@ func (ing *ingress) addSpecRuleV1(ic ingressContext, ru *definitions.RuleV1) err
...
+				err := addHostTLSCert(ic, ingtls.Hosts, ingtls.SecretName, ic.ingressV1.Metadata.Namespace)
+				if err != nil {
+					log.Errorf("failed using secret %s for %s TLS", ingtls.SecretName, htls)
+					return nil
+				}
+				log.Debugf("loaded tls certificate for %s", htls)
+				return nil
+			}
+		}
+	}
```
AlexanderYastrebov, line 330: What are these loops for? Why not just call `addHostTLSCert(ic, ingtls.Hosts, ingtls.SecretName, ic.ingressV1.Metadata.Namespace)`? If we want to be strict then we should call `addHostTLSCert` only with a list of hosts found in `rules` that are also in `tls` hosts.

dataclients/kubernetes/testdata/ingressV1/tls/tls-invalid-secret.yaml
```diff
@@ -0,0 +1,89 @@
...
+        protocol: TCP
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: myapp-secret
+  namespace: default
+type: Opaque
+data:
+  foo: bar
```
AlexanderYastrebov, line 89: Please add newline

dataclients/kubernetes/testdata/ingressV1/tls/tls-invalid-tls.yaml
```diff
@@ -0,0 +1,92 @@
...
+        protocol: TCP
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: myapp-secret
+  namespace: default
+type: kubernetes.io/tls
+data:
+  tls.crt: |
```
AlexanderYastrebov, line 89: minor: It is not clear out of the data how it is invalid. Maybe a simple `this is invalid` string would work equally well fro the test but would make it obvious that is it not base64 encoded PEM

secrets/certregistry/certregistry.go
```diff
@@ -0,0 +1,218 @@
...
+		defaultTLSCert: getFakeHostTLSCert(defaultHost),
+	}
+}
+
+func (r *CertRegistry) getCertByKey(key string) (*tlsCertificate, bool) {
+	r.mx.Lock()
+	defer r.mx.Unlock()
+
+	cert, ok := r.lookup[key]
+	if !ok || cert == nil {
```
AlexanderYastrebov, line 58: Why cert could be nil? Lets ensure we never put nil into the map

secrets/certregistry/certregistry.go
```diff
@@ -0,0 +1,218 @@
...
+	r.mx.Lock()
+	defer r.mx.Unlock()
+
+	r.lookup[key] = cert
+}
+
+// SyncCert takes a TLS certificate and list of hosts and saves them to the registry with the
+// provided key. If the object already exists it will be updated or added otherwise. Returns
+// true if key was changed.
+func (r *CertRegistry) SyncCert(key string, hosts []string, crt *tls.Certificate) bool {
```
AlexanderYastrebov, line 76: Why do we need to know if cert was changed? I don't see the result being used.
If we would check this then it should be done properly with mutex locked, otherwise there is a race condition between getting and adding the cert

AlexanderYastrebov: Why do we need key here at all? It looks like some magic value that turns out to be secret name+namespace. Maybe we should simply keep `map[string]*tls.Certificate` mappinf and use host a a key?

secrets/certregistry/certregistry.go
```diff
@@ -0,0 +1,218 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return cert, nil
+	}
+}
+
+// getBestMatchingCertificate uses a suffix search, best match operation, in order to find the best matching
+// certificate for a given hostname.
+func getBestMatchingCertificate(host string, certList []*tlsCertificate) (*tls.Certificate, error) {
```
AlexanderYastrebov, line 130: Is this logic defined somewhere in ingress spec?

secrets/certregistry/certregistry.go
```diff
@@ -0,0 +1,218 @@
...
+	hosts []string
+	cert  *tls.Certificate
+}
+
+// CertRegistry object holds TLS certificates to be used to terminate TLS connections
+// We ensure ensure syncronized access to them and hold a default certificate.
+type CertRegistry struct {
+	lookup         map[string]*tlsCertificate
+	mx             *sync.Mutex
+	defaultTLSCert *tls.Certificate
```
AlexanderYastrebov, line 38: Why do we need this fake cert?
Can we use existing `tls-cert` and `tls-key` flags for this?

dataclients/kubernetes/clusterclient.go
```diff
@@ -462,12 +493,18 @@ func (c *clusterClient) fetchClusterState() (*clusterState, error) {
 		return nil, err
 	}
 
+	secrets, err := c.loadSecrets()
```
AlexanderYastrebov, line 496: We should not load secrets if certificate registry is not used

secrets/certregistry/certregistry.go
AlexanderYastrebov: > Can we use existing
e.g. add a method to set default cert and call it from config setup

secrets/certregistry/certregistry.go
```diff
@@ -0,0 +1,218 @@
...
+	switch len(certList) {
+	case 0:
+		// no certs found so returning the default certificate
+		return r.defaultTLSCert, nil
+	case 1:
+		// only one certificate found, retuning it
+		return certList[0].cert, nil
+	default:
+		// more that one cert found, returning the best match
+		cert, err := getBestMatchingCertificate(hello.ServerName, certList)
```
AlexanderYastrebov, line 120: I think best selection should be done while reading ingresses so that we don't need to do this while looking up the certificate fro the request

AlexanderYastrebov:
@rickhlx

We've discussed with @szuecs this proposal and think it should be adjusted in the following way:
- certificate lookup from the request (hotpath) should be fast. To achieve that we propose to keep a simple 1-to-1 hostname-certificate mapping used for lookup
- in case there are multiple ingresses that define different certificates for the same host the implementation should pick the "best matching" certificate while processing ingresses (currently it does so while looking up certificate for the request, i.e. in the "hotpath") 
- we propose to start with simpler "best matching" algorithm that uses certificate `NotBefore` timestamp as a tie breaker (i.e. while processing ingresses and registering certificates the implementation picks "newest" observed certificate for the given hostname)

rickhlx:
@AlexanderYastrebov 

The only problem I see is when using the same certificate for multiple hosts. We'd be storing duplicate certificates under a different key name (host). But like you mentioned, if we don't mind using a bit more memory we'd lower latency introduced by the lookup.

 I like the idea! I'll go ahead and work on that on top of the comments from your feedback. Thank you!

dataclients/kubernetes/clusterclient.go
rickhlx: The secret is invalid because it's either `nil` or  it does not have metadata (name+namespace) so we cannot warn with that info. Updating so silently ignore.

dataclients/kubernetes/ingress.go
rickhlx: Using a ResourceID in map to find secrets now

dataclients/kubernetes/ingressv1.go
rickhlx: Simplified the lookup by getting a list of matching `hosts` in `tls` and `rules` and registering the matching hosts in the cert registry.

secrets/certregistry/certregistry.go
rickhlx: Now validating that cert is not `nil` before registering.

secrets/certregistry/certregistry.go
rickhlx: Updated to use the suggested register and lookup method with a simple `map[string]*tls.Certificate`

secrets/certregistry/certregistry.go
rickhlx: Simplified this logic to compare two certificates (if registering the same host) and register the newest certificate.

dataclients/kubernetes/clusterclient.go
rickhlx: Now making sure secrets are not loaded if cert registry is not enabled.

secrets/certregistry/certregistry.go
rickhlx: I propose we leave this logic at the certificate registry module. We'll make sure we load the best certificate (based on `NotBefore`) **if** a certificate is already registered under the same key (host). That way anything using the registry can just "sync" and the registry will how to handle registration.

This way we'll still remove it from the request but keep certificate logic at the cert registry module. Thoughts? @AlexanderYastrebov

secrets/certregistry/certregistry.go
rickhlx: I agree we should have a method to load a custom default certificate in the registry in case the user does not want an autogenerated one. However, requiring a TLS cert key/pair just to use Ingress TLS does not seem like a good experience. I think we should still generate a default cert in case `tls-cert` and `tis-key` are not loaded. That way users can still enable Ingress TLS but not have the key/pair prerequisite. 

This is also how `ingress-nginx` works [it generates a self-signed certificate](https://kubernetes.github.io/ingress-nginx/user-guide/tls/#default-ssl-certificate) if the `--default-ssl-certificate` secret is not specified.

rickhlx:
@AlexanderYastrebov @szuecs I have implemented the suggestions from your feedback, thank you for the review. I have some comments about requiring a TLS cert as a default certificate. If possible, could we choose to implement later along with removing certs from the registry?

secrets/certregistry/certregistry.go
rickhlx: > This is also how ingress-nginx works [it generates a self-signed certificate](https://kubernetes.github.io/ingress-nginx/user-guide/tls/#default-ssl-certificate) if the --default-ssl-certificate secret is not specified.

What do you think of adding a similar option to specify a secret to load as a default tls cert?

AlexanderYastrebov: > However, requiring a TLS cert key/pair just to use Ingress TLS does not seem like a good experience. I think we should still generate a default cert in case tls-cert and tis-key are not loaded.

IMO certificate generation is not the main Skipper business. We do not have to introduce any magic constants like defaultHost, default validity, key type etc and there always will be users that want something slightly different.

I think it is fine and versatile to defer default certificate creation to the user as they please and have no default certificate if `tls-cert` and `tis-key` are not set.

secrets/certregistry/certregistry.go
```diff
@@ -0,0 +1,117 @@
...
+		return r.defaultTLSCert, nil
+	}
+}
+
+// chooseBestCertificate compares two certificates and returns the newest certificate from
+// NotBefore date.
+func chooseBestCertificate(l *tls.Certificate, r *tls.Certificate) (*tls.Certificate, error) {
+	if l.Leaf == nil || r.Leaf == nil {
+		return nil, errNilLeafCertificate
+	}
```
AlexanderYastrebov, line 110: This looks redundant as we ensure cert.Leaf is always set.

secrets/certregistry/certregistry.go
```diff
@@ -0,0 +1,117 @@
...
+		curr, err := chooseBestCertificate(curr, cert)
+		if err != nil {
+			log.Warnf("choosing best certificate for %s failed, keeping current", host)
+			return
+		}
+		r.addCertToRegistry(host, curr)
+	} else {
+		log.Infof("adding certificate to registry - %s", host)
+		r.addCertToRegistry(host, cert)
+	}
```
AlexanderYastrebov, line 90: There is a race condition between `getCertByKey` and `addCertToRegistry` - it is not guaranteed if two goroutines try to add a cert for the same host that the "best" one would be added. The read-check-add sequence should be done while holding the mutex.

secrets/certregistry/certregistry.go
szuecs: I agree with Alexander.
We can then also move getFakeHostTLSCert() to test if we need it for test purposes. Less code is more good.

rickhlx: Agreed. I removed all the code related to default certificate generation. From the [crypto.tls config docs](https://pkg.go.dev/crypto/tls#Config) we see that: 
> GetCertificate returns a Certificate based on the given ClientHelloInfo. It will only be called if the client supplies SNI information or if Certificates is empty. If GetCertificate is nil or returns nil, then the certificate is retrieved from NameToCertificate. If NameToCertificate is nil, the best element of Certificates will be used.

So we're able to load tls certs if defined in `tls-cert` and `tis-key` as defaults if no cert is found in the registry.

secrets/certregistry/certregistry.go
rickhlx: Moved the lock to `SyncCert` and `GetCertFromHello`

secrets/certregistry/certregistry.go
```diff
@@ -0,0 +1,102 @@
...
+
+	return cert, true
+}
+
+func (r *CertRegistry) addCertToRegistry(key string, cert *tls.Certificate) error {
+	r.lookup[key] = cert
+
+	return nil
+}
```
AlexanderYastrebov, line 42: Lets inline these two methods for clarity

secrets/certregistry/certregistry.go
```diff
@@ -0,0 +1,102 @@
...
+	cert, found := r.getCertByKey(hello.ServerName)
+	if found {
+		return cert, nil
+	}
+	return nil, nil
+}
+
+// chooseBestCertificate compares two certificates and returns the newest certificate from
+// NotBefore date.
+func chooseBestCertificate(l *tls.Certificate, r *tls.Certificate) (*tls.Certificate, error) {
```
AlexanderYastrebov, line 96: This never returns error. Maybe lets inline it - later we could extract a method if this logic becomes more complex

secrets/certregistry/certregistry.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (r *CertRegistry) addCertToRegistry(key string, cert *tls.Certificate) error {
+	r.lookup[key] = cert
+
+	return nil
+}
+
+// SyncCert takes a host and TLS certificate and saves them to the registry with the
+// host as the key. If the cert already exists it will be updated or added otherwise.
+func (r *CertRegistry) SyncCert(host string, cert *tls.Certificate) {
```
AlexanderYastrebov, line 46: I think this method should return error instead of logging - this way we let caller to handle the error (i.e. log it).

AlexanderYastrebov: I would also rename the method and update the doc comment
```go
// Configures certificate for the host if no configuration exists or
// if certificate is valid (`NotBefore` field) after previously configured certificate.
func (r *CertRegistry) ConfigureCertificate(host string, cert *tls.Certificate) error
```

secrets/certregistry/certregistry.go
```diff
@@ -0,0 +1,102 @@
...
+		if curr.Leaf.Equal(cert.Leaf) {
+			return
+		}
+		log.Infof("updating existing certificate in registry - %s", host)
+		curr, err := chooseBestCertificate(curr, cert)
+		if err != nil {
+			log.Warnf("choosing best certificate for %s failed, keeping current", host)
+			return
+		}
+		r.addCertToRegistry(host, curr)
```
AlexanderYastrebov, line 73: We don't need to re-add `curr` certificate to the lookup if it is already there. If we inline `chooseBestCertificate` and `addCertToRegistry` it becomes straghtforward:
```go
if cert.Leaf.NotBefore.After(curr.Leaf.NotBefore) {
    r.lookup[key] = cert
}
```

When `curr` is the same as `cert`, `NotBefore` check is false so we can eliminate redundant equality check:
```go
if curr.Leaf.Equal(cert.Leaf) {
	return
}
```

AlexanderYastrebov:
@rickhlx Thank you for the persistence and commitment :+1:
I can only think of minor code changes for readability (see comments)  and we need to decide about flag name as it would be hard to change after.

config/config.go
```diff
@@ -485,6 +488,7 @@ func NewConfig() *Config {
 	flag.DurationVar(&cfg.ExpectContinueTimeoutBackend, "expect-continue-timeout-backend", 30*time.Second, "sets the HTTP expect continue timeout for backend connections")
 	flag.IntVar(&cfg.MaxIdleConnsBackend, "max-idle-connection-backend", 0, "sets the maximum idle connections for all backend connections")
 	flag.BoolVar(&cfg.DisableHTTPKeepalives, "disable-http-keepalives", false, "forces backend to always create a new connection")
+	flag.BoolVar(&cfg.TLSCertificateRegistry, "tls-certificate-registry", false, "uses the tls certificate registry to termiante tls connections")
```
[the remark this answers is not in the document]
rickhlx: Do we see this being used with `RouteGroup` resources? If so would it make more sense to name it `-kubernetes-enable-tls`?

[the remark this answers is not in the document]
AlexanderYastrebov: There are no plans but I can imagine tls support for routegroups as well. Either we add another flag to enable it or we use `-kubernetes-enable-tls` as you suggest.

rickhlx:
@AlexanderYastrebov thanks for the excellent communication and feedback! I implemented your suggestions and updated the flag with the proposed `-kubernetes-enable-tls`. Let me know if you want me to squash commits or do some cleanup. Thanks!

secrets/certregistry/certregistry.go
```diff
@@ -0,0 +1,72 @@
...
+
+	return &CertRegistry{
+		lookup: l,
+		mx:     &sync.Mutex{},
+	}
+}
+
+// Configures certificate for the host if no configuration exists or
+// if certificate is valid (`NotBefore` field) after previously configured certificate.
+func (r *CertRegistry) ConfigureCertificate(key string, cert *tls.Certificate) error {
```
AlexanderYastrebov, line 31: nit: lets rename `key` to `host`

dataclients/kubernetes/ingressdefinitions.go
```diff
@@ -172,3 +172,13 @@ type ClusterResourceList struct {
 	// Items, aka "resources".
 	Items []*ClusterResource `json:"resources"`
 }
+
+type secret struct {
+	Meta *definitions.Metadata `json:"Metadata"`
```
AlexanderYastrebov, line 177: nit: I think json field is lowercase and I would also name struct field exactly as a json field
```go
Metadata *definitions.Metadata `json:"metadata"`
```
but I guess it works either way

AlexanderYastrebov:
`make staticcheck` [fails](https://github.com/zalando/skipper/runs/5775592487?check_suite_focus=true) otherwise looks good.

> Let me know if you want me to squash commits

Not necessary as we have squash merges. I personally like to squash commits and describe the change in the commit message.

AlexanderYastrebov:
@rickhlx Thank you, great work!

AlexanderYastrebov:
:+1:

dataclients/kubernetes/ingress.go
```diff
@@ -421,3 +449,25 @@ func (ing *ingress) convert(state *clusterState, df defaultFilters) ([]*eskip.Ro
...
 	return routes, nil
 }
+
+func generateTLSCertFromSecret(secret *secret) (*tls.Certificate, error) {
+	if secret.Data[tlsSecretDataCrt] == "" || secret.Data[tlsSecretDataKey] == "" {
+		return nil, fmt.Errorf("secret must contain %s and %s in data field", tlsSecretDataCrt, tlsSecretDataKey)
+	}
+	crt, err := b64.StdEncoding.DecodeString(secret.Data[tlsSecretDataCrt])
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode secret data %s", tlsSecretDataCrt)
```
szuecs, line 459: `failed to decode secret data` -> `failed to decode crt data`

dataclients/kubernetes/ingress.go
```diff
@@ -421,3 +449,25 @@ func (ing *ingress) convert(state *clusterState, df defaultFilters) ([]*eskip.Ro
...
+	}
+	key, err := b64.StdEncoding.DecodeString(secret.Data[tlsSecretDataKey])
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode secret data %s", tlsSecretDataKey)
```
szuecs, line 463: `failed to decode secret data` -> `failed to decode key data` 

otherwise we can not decide if we have the first error case or this one

secrets/certregistry/certregistry.go
```diff
@@ -0,0 +1,72 @@
...
+// Configures certificate for the host if no configuration exists or
+// if certificate is valid (`NotBefore` field) after previously configured certificate.
+func (r *CertRegistry) ConfigureCertificate(host string, cert *tls.Certificate) error {
+	if cert == nil {
+		return fmt.Errorf("cannot configure nil certificate")
+	}
+	// loading parsed leaf certificate to certificate
+	leaf, err := x509.ParseCertificate(cert.Certificate[0])
+	if err != nil {
+		return fmt.Errorf("failed parsing leaf certificate")
```
szuecs, line 38: `"failed parsing leaf certificate"` -> `"failed parsing leaf certificate: %w", err`

secrets/certregistry/certregistry.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+// GetCertFromHello reads the SNI from a TLS client and returns the appropriate certificate.
+// If no certificate is found for the host it will return nil.
+func (r *CertRegistry) GetCertFromHello(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	r.mx.Lock()
+	defer r.mx.Unlock()
+
+	cert, found := r.lookup[hello.ServerName]
```
szuecs, line 67: This section because in the critical path could also be rewritten to:
```
r.mx.Lock()
cert, found := r.lookup[hello.ServerName]
r.mx.Unlock()
```

So we would omit holding the lock for the branch and defer adds no readability enhancement in this case, too.

szuecs:
I think 4 small changes and we are good to merge. Thanks for your great work @rickhlx and thanks @AlexanderYastrebov for the great review!

dataclients/kubernetes/ingress.go
rickhlx: This should would return `failed to decode secret data tls.crt` which is the specific secret data we're decoding.

dataclients/kubernetes/ingress.go
rickhlx: This should would return `failed to decode secret data tls.key`

rickhlx: Updated to make the message clearer: `failed to decode tls.key from secret myapp-secret`

dataclients/kubernetes/ingress.go
rickhlx: Updated to make the message clearer: `failed to decode tls.crt from secret myapp-secret`

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #1983, opened 2022-03-18
draft: Limits number of concurrent connections without queuing

AlexanderYastrebov:
TODO: metrics - unfortunately `netutil.LimitListener` does not expose number of active connections, it either needs to be extended upstream or vendored

### DIFF
--- a/skipper.go
+++ b/skipper.go
@@ -18,6 +18,8 @@ import (
 
 	stdlog "log"
 
+	"golang.org/x/net/netutil"
+
 	ot "github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


skipper.go
```diff
@@ -1024,7 +1026,11 @@ func listen(o *Options, mtr metrics.Metrics) (net.Listener, error) {
 	}
 
 	if !o.EnableTCPQueue {
-		return net.Listen("tcp", o.Address)
+		l, err := net.Listen("tcp", o.Address)
```
szuecs, line 1029: this does not look like not a TCP Queue :)
I think we want to have a "flavor": lifo vs fifo 
netutil.LimitListener is a fio queue

AlexanderYastrebov: Yes, `netutil.LimitListener` is not a queue - it limits number of active connections to N and never accepts more than that until some active connection is closed

PR #1984, opened 2022-03-18
Uses debug level for `http: URL query contains semicolon`

### LINKED ISSUE #1982: Verbose error log: URL query contains semicolon
**Describe the bug**
Skipper does not support parsing of URL queries that contain semicolon, but logs an error if clients send semicolons. Logged errors should be actionable, which in this context is not the case since this is something the client does. It's not an error from Skipper's perspective. 

**Expected behavior**
Log level of info would potentially (if a log message is actually required) make more sense in this case.

**Observed behavior**
with Skipper v0.13.185 
`time="2022-03-18T09:20:32Z" level=error msg="http: URL query contains semicolon, which is no longer a supported separator; parts of the query may be stripped when parsed; see golang.org/issue/25192\n"`

AlexanderYastrebov:
Fixes #1982

### DIFF
--- a/skipper.go
+++ b/skipper.go
@@ -811,9 +811,10 @@ type serverErrorLogWriter struct{}
 
 func (*serverErrorLogWriter) Write(p []byte) (int, error) {
 	m := string(p)
-	// https://github.com/golang/go/issues/26918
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

PR #1985, opened 2022-03-21
proxy: use debug level fo serveResponse logs

### LINKED ISSUE #1981: Verbose error log: context canceled
**Describe the bug**
Whenever a client cancels a request before the request can be processed, Skipper logs an error. Logged errors should be actionable, which in this context is not the case since this is something the client does. It's not an error from Skipper's perspective. 

**Expected behavior**
Log level of info would potentially make more sense in this case.

**Observed behavior**
with Skipper v0.13.185 
`time="2022-03-18T10:32:58Z" level=error msg="error while copying the response stream: context canceled"`

AlexanderYastrebov:
Existing log messages do no provide enough context about request and are only suitable for local debugging.
There are metrics and opentracing tags already in place to monitor problems in production setting.

Fixes #1981

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1190,7 +1190,7 @@ func (p *Proxy) serveResponse(ctx *context) {
 	if err := ctx.Request().Context().Err(); err != nil {
 		// deadline exceeded or canceled in stdlib, client closed request
 		// see https://github.com/zalando/skipper/pull/864
-		p.log.Infof("Client request: %v", err)
+		p.log.Debugf("Client request: %v", err)
 		ctx.response.StatusCode = 499
 		p.tracing.setTag(ctx.proxySpan, ClientRequestStateTag, ClientRequestCanceled)
 	}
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

proxy/proxy.go
```diff
@@ -1205,7 +1205,7 @@ func (p *Proxy) serveResponse(ctx *context) {
 	p.tracing.logStreamEvent(ctx.proxySpan, StreamBodyEvent, strconv.FormatInt(n, 10))
 	if err != nil {
 		p.metrics.IncErrorsStreaming(ctx.route.Id)
-		p.log.Errorf("error while copying the response stream: %v", err)
+		p.log.Debugf("error while copying the response stream: %v", err)
```
szuecs, line 1208: are you sure that we don't get this error in cases which we need?

AlexanderYastrebov: My point is that this message is useless for debugging in production because it lacks details of the request.
For monitoring purposes `streaming_errors` metric and/or opentracing could be used.

AlexanderYastrebov: > Then we should drop the log

This still could be useful for local single request debugging. Also there is a test that depends on this log line and I did not want to spend much time refactoring it

szuecs: :+1:

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #1987, opened 2022-03-22
golangci-lint integration

dzyanis:
I'd recommend to integrate [golangci-lint](https://golangci-lint.run/) and then enable linters one by one. There are many warnings


szuecs:
@dzyanis no thanks, the current linting setup is fine enough and I personally don't like golangci-lint too much.
It will create a lot of noise that we can't fix, because we want to have compatibility claim for our library users even if we are in v0.

PR #1990, opened 2022-03-28
Document fadeIn and Rolling Restart caveat

### LINKED ISSUE #1207: Slowly ramp up traffic to new backend-ips

otrosien:
Closes #1207


szuecs:
:+1:

PR #1991, opened 2022-03-30
Build with Go 1.18

mikkeloscar:
`staticcheck` has been updated to work with Go 1.18 so attempt to update the version used for building.

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           # https://www.npmjs.com/package/semver#caret-ranges-123-025-004
-          go-version: '^1.17'
+          go-version: '^1.18'
       - run: go version
       - run: make deps
       - run: make check-fmt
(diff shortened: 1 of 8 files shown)


AlexanderYastrebov:
I think static check needs to be updated. It is also possible to [install](https://staticcheck.io/docs/getting-started/) it using `go install` since 1.17

szuecs:
Yes `make deps` needs a change and
```
go install honnef.co/go/tools/cmd/staticcheck@latest
```
could be used or the versioned one

mikkeloscar:
Ah, didn't realize we install it manually when running on CDP. What do you think about not running `make deps` in CDP and just rely on `gosec` and `staticcheck` which are provided with the `cdp-runtime/go`?

For github action I will update `make deps` with new `staticcheck` version.

szuecs:
sgtm

mikkeloscar:
With `staticcheck` fixed it complained about us using the `net.Error.Temporary()` method which is now deprecated. Since we use it in several places I just ignored the warnings for now with a `lint:ignore` comment and created an issue to tackle it separately: https://github.com/zalando/skipper/issues/1992

mikkeloscar:
:+1:

mikkeloscar:
The DCO check seems to be stuck :( Not sure how to get it working.

szuecs:
@mikkeloscar I took the freedom to rebase and squash commits to one

mikkeloscar:
👍

AlexanderYastrebov:
:+1:

PR #1995, opened 2022-04-13
proxy: disable default net/http user agent

### LINKED ISSUE #1994: Preserve empty User-Agent header
**Describe the bug**

Requests sent through skipper with an empty `User-Agent` header have the value of `Go-http-client/1.1` added instead of preserving the empty value.

**To Reproduce**
Add context, for example:

1. Start skipper in proxy mode with `./bin/skipper -inline-routes='r1: * -> <roundRobin, "http://127.0.0.1:9001">' --address :9999`
2. Create a http server to print out the `User-Agent` header

```go
package main

import (
	"fmt"

	"github.com/gin-gonic/gin"
)

func main() {
	r := gin.Default()
	r.GET("/foo", func(c *gin.Context) {
		fmt.Println(c.Request.UserAgent())
		c.JSON(200, gin.H{
			"message": "pong",
		})
	})
	r.Run(":9001") // listen and serve on 0.0.0.0:9001
}
```

3. Start the http server

```sh
go run main.go
```

4. Send requests to skipper

```
curl http://localhost:9999/foo
curl http://localhost:9999/foo -H "User-Agent: "
```

Logs from the last step look like:

```
curl/7.64.1
[GIN] 2022/04/13 - 13:49:04 | 200 |      67.276µs |       127.0.0.1 | GET      "/foo"
Go-http-client/1.1
[GIN] 2022/04/13 - 13:49:38 | 200 |      76.356µs |       127.0.0.1 | GET      "/foo"
```

**Expected behavior**

Skipper preserves the empty `User-Agent` header.

**Observed behavior**

Skipper modifies `User-Agent` header to be `Go-http-client/1.1`.

AlexanderYastrebov:
Similar to https://github.com/golang/go/issues/15524
Fixes #1994

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -501,6 +501,10 @@ func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHea
 	} else {
 		rr.Header = cloneHeader(r.Header)
 	}
+	// Disable default net/http user agent when user agent is not specified
+	if _, ok := rr.Header["User-Agent"]; !ok {
+		rr.Header["User-Agent"] = []string{""}
+	}
 	rr.Host = host
 
 	// If there is basic auth configured in the URL we add them as headers
--- a/proxy/proxy_test.go
+++ b/proxy/proxy_test.go
@@ -2062,6 +2062,42 @@ func TestHopHeaderRemovalDisabled(t *testing.T) {
 	}
 }
 
+func TestUserAgent(t *testing.T) {
+	for _, tc := range []struct {
+		name      string
+		userAgent string
+	}{
+		{name: "no user agent"},
+		{name: "with user agent", userAgent: "test ua"},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			s := startTestServer(nil, 0, func(r *http.Request) {
+				if got := r.Header.Get("User-Agent"); tc.userAgent != got {
+					t.Errorf("user agent mismatch: expected %q, got %q", tc.userAgent, got)
+				}
+			})
+			defer s.Close()
+
+			r := httptest.NewRequest("GET", "http://example.com/foo", nil)
+			if tc.userAgent != "" {
+				r.Header.Set("User-Agent", tc.userAgent)
+			}
+
+			w := httptest.NewRecorder()
+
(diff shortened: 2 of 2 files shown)


proxy/proxy.go
```diff
@@ -501,6 +501,10 @@ func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHea
 	} else {
 		rr.Header = cloneHeader(r.Header)
 	}
+	// Disable default net/http user agent when user agent is not specified
+	if _, ok := rr.Header["User-Agent"]; !ok {
+		rr.Header["User-Agent"] = []string{""}
```
szuecs, line 506: I tested that this works, too and reads better:
```diff
diff --git proxy/proxy.go proxy/proxy.go
index e04f510d..91d5d5be 100644
--- proxy/proxy.go
+++ proxy/proxy.go
@@ -501,6 +501,10 @@ func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHea
        } else {
                rr.Header = cloneHeader(r.Header)
        }
+
+       if rr.Header.Get("User-Agent") == "" {
+               rr.Header.Set("User-Agent", "")
+       }
```

AlexanderYastrebov: I've implemented it this way because the fix for https://github.com/golang/go/issues/15524 as well as [request writer](https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/net/http/request.go;l=625-630) both check for the presence of the value treating headers as a map.
The `rr.Header["User-Agent"]` explicitly checks for absent value while `rr.Header.Get("User-Agent")` returns empty for both absent and empty stored value, i.e. ambiguous. 
Also both `Set` and `Get` transform header name into canonical form which is redundant since `User-Agent` is already a canonical form (not a big deal probably but why spend cycles on the hot path if we know it is useless).
We also treat Header type as a map in the `copyHeader` so it should not be a concern.

> and reads better

The fix uses undocumented feature of request writer that does not use default user agent (also undocumented) if the value is explicitly set to empty. Without knowing this your version reads weird anyway - "set to empty if it is empty"...

szuecs:
:+1:

PR #1996, opened 2022-04-14
Install `gover` via go install

AlexanderYastrebov:
Follow up on #1991

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,6 @@ clean:
 deps:
 	go env
 	./etcd/install.sh $(TEST_ETCD_VERSION)
-	mkdir -p .bin
 	@go install honnef.co/go/tools/cmd/staticcheck@latest
 	@go install github.com/securego/gosec/v2/cmd/gosec@latest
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -119,7 +119,6 @@ clean:
 deps:
 	go env
 	./etcd/install.sh $(TEST_ETCD_VERSION)
-	mkdir -p .bin
```
AlexanderYastrebov, line 122: Redundant as it is created by `./etcd/install.sh` (therefore clean target still contains `rm -rf .bin`)

Makefile
```diff
@@ -165,7 +164,7 @@ precommit: fmt build vet staticcheck check-race shortcheck
 			'{{if len .TestGoFiles}}"GO111MODULE=on go test -tags=redis -coverprofile={{.Dir}}/.coverprofile {{.ImportPath}}"{{end}}' \
 			$$p | xargs -i sh -c {}; \
 	done
-	go get github.com/modocache/gover
```
AlexanderYastrebov, line 168: https://go.dev/doc/go-get-install-deprecation

szuecs:
👍

PR #1999, opened 2022-04-20
Refactor/rm v1beta ingress

szuecs:
remove ingress v1beta1 and change all tests config accordingly

Missing: docs/ change to change all examples to v1 ingress

PR #2000, opened 2022-04-22
deprecate ingress v1beta1 

szuecs:
deprecate ingress v1beta1 and change all the docs, examples and deployment files

relates to https://github.com/zalando/skipper/pull/1999

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -88,7 +88,7 @@ type Options struct {
 	// in the cluster-scope.
 	KubernetesNamespace string
 
-	// KubernetesIngressV1 is used to switch between v1beta1 and v1. Kubernetes version 1.22 stopped support
+	// *DEPRECATED* KubernetesIngressV1 is used to switch between v1beta1 and v1. Kubernetes version 1.22 stopped support
 	// for v1beta1, so we have to provide a migration path and this will someday become the default.
 	KubernetesIngressV1 bool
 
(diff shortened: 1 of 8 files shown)


docs/kubernetes/ingress-controller.md
```diff
@@ -696,3 +705,81 @@ the communication work with TCP and UDP to the specified `swarm-port`:
...
+3. If you use Ingress v1beta1 and run Kubernetes cluster version that
+  does not support ingress v1, then you can't update skipper to
+  `>=0.14.0`, before you upgrade your Kubernetes cluster.
+4. If you use Ingress v1beta1 and run Kubernetes cluster version that
+  support ingress v1, then you need to allow skipper to access the new
+  APIs with a changed RBAC. See the guide below.
+
+
+If you are in case 4., you have to apply a change in your RBAC, please
+check the diff or the full rendered file.
```
AlexanderYastrebov, line 732: I think the real migration would take multiple steps:
1. Add new rbac for ingress v1
2. Update Skipper (or enable `-kubernetes-ingress-v1`)
3. Remove v1beta1 rbac

dataclients/kubernetes/kube.go
```diff
@@ -88,7 +88,7 @@ type Options struct {
 	// in the cluster-scope.
 	KubernetesNamespace string
 
-	// KubernetesIngressV1 is used to switch between v1beta1 and v1. Kubernetes version 1.22 stopped support
+	// *DEPRECATED* KubernetesIngressV1 is used to switch between v1beta1 and v1. Kubernetes version 1.22 stopped support
```
AlexanderYastrebov, line 91: minor: I think there is some kind of convention that e.g. staticcheck understands, maybe a `// DEPRECATED` comment on the same line.

docs/kubernetes/ingress-controller.md
AlexanderYastrebov: Ok, I see that the change adds v1 (not substitutes)

AlexanderYastrebov:
:+1:

docs/kubernetes/ingress-controller.md
szuecs: Some people might not need it, others do, so I guess it's fine, because I assume everyone tries to think about zero downtime migrations. The fail safe is kubernetes, because one crashing pod will stop the deployment from going further.

dataclients/kubernetes/kube.go
szuecs: I used it because we already use it like this, but great to know, maybe we should change this everywhere at once.

szuecs:
:+1:

PR #2005, opened 2022-05-16
docs: predicate should have fragment by name to reference easily

szuecs:


AlexanderYastrebov:
:+1:

PR #2006, opened 2022-05-17
docs: add a second header for Weight predicate

AlexanderYastrebov:
The second header adds another anchor that matches predicate name.


AlexanderYastrebov:
Closing in favor of #2005

PR #2009, opened 2022-05-25
feature: dynamic detect and resize redis instances

szuecs:
In case of kubernetes dataclient we can detect the number of redis shards available and trigger an update to the redis ringclient.
In order to use this feature we have to use a forked version of go-redis (see replace directive in go.mod) until it is merged and released https://github.com/go-redis/redis/pull/2093

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -169,6 +169,7 @@ type Config struct {
 	KubernetesAllowedExternalNames          regexpListFlag      `yaml:"kubernetes-allowed-external-names"`
 	KubernetesRedisServiceNamespace         string              `yaml:"kubernetes-redis-service-namespace"`
 	KubernetesRedisServiceName              string              `yaml:"kubernetes-redis-service-name"`
+	KubernetesRedisServicePort              int                 `yaml:"kubernetes-redis-service-port"`
 
 	// Default filters
 	DefaultFiltersDir string `yaml:"default-filters-dir"`
@@ -434,6 +435,7 @@ func NewConfig() *Config {
 	flag.Var(&cfg.KubernetesAllowedExternalNames, "kubernetes-allowed-external-name", "set zero or more regular expressions from which at least one should be matched by the external name services, route group network addresses and explicit endpoints domain names")
 	flag.StringVar(&cfg.KubernetesRedisServiceNamespace, "kubernetes-redis-service-namespace", "", "Sets namespace for redis to be used to lookup endpoints")
 	flag.StringVar(&cfg.KubernetesRedisServiceName, "kubernetes-redis-service-name", "", "Sets name for redis to be used to lookup endpoints")
+	flag.IntVar(&cfg.KubernetesRedisServicePort, "kubernetes-redis-service-port", 6379, "Sets the port for redis to be used to lookup endpoints")
 
 	// Auth:
 	flag.BoolVar(&cfg.EnableOAuth2GrantFlow, "enable-oauth2-grant-flow", false, "enables OAuth2 Grant Flow filter")
@@ -781,6 +783,7 @@ func (c *Config) ToOptions() skipper.Options {
 		KubernetesAllowedExternalNames:     c.KubernetesAllowedExternalNames,
 		KubernetesRedisServiceNamespace:    c.KubernetesRedisServiceNamespace,
 		KubernetesRedisServiceName:         c.KubernetesRedisServiceName,
+		KubernetesRedisServicePort:         c.KubernetesRedisServicePort,
 
 		// API Monitoring:
 		ApiUsageMonitoringEnable:                c.ApiUsageMonitoringEnable,
--- a/config/config.go
+++ b/config/config.go
@@ -167,8 +167,8 @@ type Config struct {
 	KubernetesEastWestRangePredicates       []*eskip.Predicate  `yaml:"-"`
 	KubernetesOnlyAllowedExternalNames      bool                `yaml:"kubernetes-only-allowed-external-names"`
 	KubernetesAllowedExternalNames          regexpListFlag      `yaml:"kubernetes-allowed-external-names"`
-	KubernetesRedisNamespace                string              `yaml:"kubernetes-redis-namespace"`
-	KubernetesRedisName                     string              `yaml:"kubernetes-redis-name"`
+	KubernetesRedisServiceNamespace         string              `yaml:"kubernetes-redis-service-namespace"`
+	KubernetesRedisServiceName              string              `yaml:"kubernetes-redis-service-name"`
 
 	// Default filters
 	DefaultFiltersDir string `yaml:"default-filters-dir"`
@@ -432,8 +432,8 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.KubernetesEastWestRangePredicatesString, "kubernetes-east-west-range-predicates", "", "set the predicates that will be appended to routes identified as to -kubernetes-east-west-range-domains")
 	flag.BoolVar(&cfg.KubernetesOnlyAllowedExternalNames, "kubernetes-only-allowed-external-names", false, "only accept external name services, route group network backends and route group explicit LB endpoints from an allow list defined by zero or more -kubernetes-allowed-external-name flags")
 	flag.Var(&cfg.KubernetesAllowedExternalNames, "kubernetes-allowed-external-name", "set zero or more regular expressions from which at least one should be matched by the external name services, route group network addresses and explicit endpoints domain names")
-	flag.StringVar(&cfg.KubernetesRedisNamespace, "kubernetes-redis-namespace", "", "Sets namespace for redis to be used to lookup endpoints")
-	flag.StringVar(&cfg.KubernetesRedisName, "kubernetes-redis-name", "", "Sets name for redis to be used to lookup endpoints")
+	flag.StringVar(&cfg.KubernetesRedisServiceNamespace, "kubernetes-redis-namespace", "", "Sets namespace for redis to be used to lookup endpoints")
+	flag.StringVar(&cfg.KubernetesRedisServiceName, "kubernetes-redis-name", "", "Sets name for redis to be used to lookup endpoints")
 
 	// Auth:
 	flag.BoolVar(&cfg.EnableOAuth2GrantFlow, "enable-oauth2-grant-flow", false, "enables OAuth2 Grant Flow filter")
@@ -779,8 +779,8 @@ func (c *Config) ToOptions() skipper.Options {
 		KubernetesEastWestRangePredicates:  c.KubernetesEastWestRangePredicates,
 		KubernetesOnlyAllowedExternalNames: c.KubernetesOnlyAllowedExternalNames,
 		KubernetesAllowedExternalNames:     c.KubernetesAllowedExternalNames,
-		KubernetesRedisNamespace:           c.KubernetesRedisNamespace,
-		KubernetesRedisName:                c.KubernetesRedisName,
+		KubernetesRedisNamespace:           c.KubernetesRedisServiceNamespace,
+		KubernetesRedisName:                c.KubernetesRedisServiceName,
 
 		// API Monitoring:
 		ApiUsageMonitoringEnable:                c.ApiUsageMonitoringEnable,
--- a/config/config.go
+++ b/config/config.go
@@ -432,8 +432,8 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.KubernetesEastWestRangePredicatesString, "kubernetes-east-west-range-predicates", "", "set the predicates that will be appended to routes identified as to -kubernetes-east-west-range-domains")
 	flag.BoolVar(&cfg.KubernetesOnlyAllowedExternalNames, "kubernetes-only-allowed-external-names", false, "only accept external name services, route group network backends and route group explicit LB endpoints from an allow list defined by zero or more -kubernetes-allowed-external-name flags")
 	flag.Var(&cfg.KubernetesAllowedExternalNames, "kubernetes-allowed-external-name", "set zero or more regular expressions from which at least one should be matched by the external name services, route group network addresses and explicit endpoints domain names")
-	flag.StringVar(&cfg.KubernetesRedisServiceNamespace, "kubernetes-redis-namespace", "", "Sets namespace for redis to be used to lookup endpoints")
-	flag.StringVar(&cfg.KubernetesRedisServiceName, "kubernetes-redis-name", "", "Sets name for redis to be used to lookup endpoints")
+	flag.StringVar(&cfg.KubernetesRedisServiceNamespace, "kubernetes-redis-service-namespace", "", "Sets namespace for redis to be used to lookup endpoints")
+	flag.StringVar(&cfg.KubernetesRedisServiceName, "kubernetes-redis-service-name", "", "Sets name for redis to be used to lookup endpoints")
 
 	// Auth:
 	flag.BoolVar(&cfg.EnableOAuth2GrantFlow, "enable-oauth2-grant-flow", false, "enables OAuth2 Grant Flow filter")
@@ -779,8 +779,8 @@ func (c *Config) ToOptions() skipper.Options {
 		KubernetesEastWestRangePredicates:  c.KubernetesEastWestRangePredicates,
 		KubernetesOnlyAllowedExternalNames: c.KubernetesOnlyAllowedExternalNames,
 		KubernetesAllowedExternalNames:     c.KubernetesAllowedExternalNames,
-		KubernetesRedisNamespace:           c.KubernetesRedisServiceNamespace,
-		KubernetesRedisName:                c.KubernetesRedisServiceName,
+		KubernetesRedisServiceNamespace:    c.KubernetesRedisServiceNamespace,
+		KubernetesRedisServiceName:         c.KubernetesRedisServiceName,
 
 		// API Monitoring:
 		ApiUsageMonitoringEnable:                c.ApiUsageMonitoringEnable,
--- a/config/config.go
+++ b/config/config.go
@@ -167,6 +167,8 @@ type Config struct {
 	KubernetesEastWestRangePredicates       []*eskip.Predicate  `yaml:"-"`
 	KubernetesOnlyAllowedExternalNames      bool                `yaml:"kubernetes-only-allowed-external-names"`
 	KubernetesAllowedExternalNames          regexpListFlag      `yaml:"kubernetes-allowed-external-names"`
+	KubernetesRedisNamespace                string              `yaml:"kubernetes-redis-namespace"`
+	KubernetesRedisName                     string              `yaml:"kubernetes-redis-name"`
 
 	// Default filters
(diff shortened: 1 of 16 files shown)


net/redisclient.go
```diff
@@ -25,6 +26,11 @@ import (
 type RedisOptions struct {
 	// Addrs are the list of redis shards
 	Addrs []string
+
+	// AddrUpdater is a func that is called with a channel that
+	// will be written to an updated redis address list
+	AddrUpdater func(quit chan struct{}, ch chan<- []string)
```
AlexanderYastrebov, line 32: Maybe it should be just a channel of updates - if not nil then redis client subscribes to it and updates address on each receive.

net/redisclient.go
```diff
@@ -222,11 +228,34 @@ func NewRedisRingClient(ro *RedisOptions) *RedisRingClient {
...
+		r.ch = ch
 	}
 
 	return r
 }
 
+func createAddressMap(addrs []string) map[string]string {
+	res := make(map[string]string)
+	for idx, addr := range addrs {
+		addr = strings.TrimLeft(addr, "TCP://")
```
AlexanderYastrebov, line 253: Should be `TrimPrefix` I guess

AlexanderYastrebov: Also I think this trim is only needed for dynamic resize so maybe it is better done at the caller

net/redisclient.go
```diff
@@ -270,9 +299,15 @@ func (r *RedisRingClient) StartSpan(operationName string, opts ...opentracing.St
 func (r *RedisRingClient) Close() {
 	if r != nil {
 		close(r.quit)
+		close(r.ch)
 	}
 }
 
+func (r *RedisRingClient) SetAddrs(ctx context.Context, addrs map[string]string) {
```
AlexanderYastrebov, line 306: Why context if it is not used?

skipper.go
```diff
@@ -1365,11 +1371,49 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
+						log.Infof("Redis updater called and found %d", len(a))
+
+						if old != len(a) && len(a) != 0 {
+							log.Infof("Redis updater updating %d != %d", old, len(a))
+							old = len(a)
+							w <- a
+						}
+					}
+				}
+			}
```
AlexanderYastrebov, line 1413: Lets extract to the helper method

skipper.go
```diff
@@ -1365,11 +1371,49 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 
 	var swarmer ratelimit.Swarmer
 	var redisOptions *skpnet.RedisOptions
+	log.Infof("enable swarm: %v", o.EnableSwarm)
 	if o.EnableSwarm {
 		if len(o.SwarmRedisURLs) > 0 {
```
AlexanderYastrebov, line 1376: For dynamic updates `SwarmRedisURLs` would be empty and the initial list should be discovered automatically I guess

szuecs:
waiting for https://github.com/go-redis/redis/pull/2093

net/redisclient.go
szuecs: changed the signature to `AddrUpdater func() []string` to used it here in a goroutine to poll kubernetes dataclient

net/redisclient.go
szuecs: TrimLeft -> TrimPrefix

I don't think we need to optimize it and it seems more well to cleanup the data we get, but I can also move the cleanup to skipper.go.

net/redisclient.go
szuecs: fixed

skipper.go
szuecs: refactored to a different implementation

skipper.go
szuecs: fixed

config/config.go
```diff
@@ -167,6 +167,8 @@ type Config struct {
 	KubernetesEastWestRangePredicates       []*eskip.Predicate  `yaml:"-"`
 	KubernetesOnlyAllowedExternalNames      bool                `yaml:"kubernetes-only-allowed-external-names"`
 	KubernetesAllowedExternalNames          regexpListFlag      `yaml:"kubernetes-allowed-external-names"`
+	KubernetesRedisNamespace                string              `yaml:"kubernetes-redis-namespace"`
+	KubernetesRedisName                     string              `yaml:"kubernetes-redis-name"`
```
AlexanderYastrebov, line 171: Should this be redis endpoints name (`Endpoints` name matches service name) ?

dataclients/kubernetes/clusterclient.go
```diff
@@ -438,7 +438,11 @@ func (c *clusterClient) loadEndpoints() (map[definitions.ResourceID]*endpoint, e
 	log.Debugf("all endpoints received: %d", len(endpoints.Items))
 	result := make(map[definitions.ResourceID]*endpoint)
 	for _, endpoint := range endpoints.Items {
-		result[endpoint.Meta.ToResourceID()] = endpoint
+		resID := endpoint.Meta.ToResourceID()
+		result[resID] = endpoint
+		if endpoint.Meta.Name == "skipper-ingress-redis" && endpoint.Meta.Namespace == "kube-system" {
+			log.Infof("clusterclient found addresses %d", len(endpoint.Subsets[0].Addresses))
+		}
```
AlexanderYastrebov, line 445: Why do we need this? Should it compare to configured `KubernetesRedisNamespace` and `KubernetesRedisName`?

dataclients/kubernetes/clusterstate.go
```diff
@@ -42,11 +42,15 @@ func (state *clusterState) getServiceRG(namespace, name string) (*service, error
 	return s, nil
 }
 
+func (state *clusterState) GetEndpointsByService(namespace, name, protocol string, servicePort *servicePort) []string {
```
AlexanderYastrebov, line 45: Why not rename `getEndpointsByService`  to `GetEndpointsByService`?

dataclients/kubernetes/kube.go
```diff
@@ -491,6 +494,20 @@ func (c *Client) fetchDefaultFilterConfigs() defaultFilters {
 	return filters
 }
 
+func (c *Client) GetEndpointAdresses(ns, name string) []string {
```
AlexanderYastrebov, line 497: Adresses -> Addresses

dataclients/kubernetes/kube.go
```diff
@@ -491,6 +494,20 @@ func (c *Client) fetchDefaultFilterConfigs() defaultFilters {
...
+	if c.state == nil {
+		log.Info("state is nil")
+		return nil
+	}
+
+	addrs := c.state.GetEndpointsByTarget(ns, name, "TCP", &definitions.BackendPort{
+		Value: 6379,
```
AlexanderYastrebov, line 504: Why is this hardcoded? Should it just get endpoints from the pre-configured namespace by the pre-configured name?

net/redisclient.go
```diff
@@ -226,11 +249,43 @@ func NewRedisRingClient(ro *RedisOptions) *RedisRingClient {
 		r.ring = redis.NewRing(ringOptions)
 		r.log = ro.Log
 		r.metricsPrefix = ro.MetricsPrefix
+
+		if ro.AddrUpdater != nil {
+			if ro.UpdateInterval == 0 {
+				ro.UpdateInterval = defaultUpdateInterval
+			}
+			go func() {
```
AlexanderYastrebov, line 257: Lets create a helper function for this

net/redisclient.go
```diff
@@ -226,11 +249,43 @@ func NewRedisRingClient(ro *RedisOptions) *RedisRingClient {
...
+				old := len(ringOptions.Addrs)
+				t := time.NewTicker(ro.UpdateInterval)
```
AlexanderYastrebov, line 259: `defer t.Close()` is missing

net/redisclient.go
```diff
@@ -226,11 +249,43 @@ func NewRedisRingClient(ro *RedisOptions) *RedisRingClient {
...
+				r.log.Info("Start goroutine to update redis instances every %s", ro.UpdateInterval)
+				for range t.C {
+					select {
+					case <-r.quit:
+						return
+					default:
+					}
+					addrs := ro.AddrUpdater()
+					if old != len(addrs) && len(addrs) != 0 {
```
AlexanderYastrebov, line 268: I think it is safe to call `r.SetAddrs` with the unchanged set of addresses so there is no need to check lengths. Also a single address can change so this check will skip an update.

net/redisclient.go
```diff
@@ -226,11 +249,43 @@ func NewRedisRingClient(ro *RedisOptions) *RedisRingClient {
...
 	}
 
 	return r
 }
 
+func createAddressMap(addrs []string) map[string]string {
+	res := make(map[string]string)
+	for idx, addr := range addrs {
+		addr = strings.TrimPrefix(addr, "TCP://")
+		res[fmt.Sprintf("redis%d", idx)] = addr
```
AlexanderYastrebov, line 284: Is it safe to use index as a key? What if there two addresses `[10.2.0.1, 10.2.0.2]` just change order to `[10.2.0.2, 10.2.0.1]`? Myybe we should use address as a key here `res[addr] = addr`

net/redisclient.go
```diff
@@ -226,11 +249,43 @@ func NewRedisRingClient(ro *RedisOptions) *RedisRingClient {
...
+		}
 	}
 
 	return r
 }
 
+func createAddressMap(addrs []string) map[string]string {
+	res := make(map[string]string)
+	for idx, addr := range addrs {
+		addr = strings.TrimPrefix(addr, "TCP://")
```
AlexanderYastrebov, line 283: I think this should be done inside of the `AddrUpdater` as this prefix is specific to the k8s endpoints.

config/config.go
szuecs: yes endpoints name matches service name and normal users have no idea about endpoints.

dataclients/kubernetes/clusterclient.go
szuecs: this is just my hack to see if it works at this place

dataclients/kubernetes/clusterstate.go
szuecs: yes true

dataclients/kubernetes/kube.go
szuecs: fixed

dataclients/kubernetes/kube.go
szuecs: We can also pass this in a func but then we would need to also expose this as config, and maybe this is too much

net/redisclient.go
szuecs: why?
I think it's good to copy everything to local func closure and it's a pattern that I follow to have background goroutines started in the NewX func.

net/redisclient.go
szuecs: fixed (it's `t.Stop()`)

net/redisclient.go
szuecs: no it would Lock unnecessary and create memory garbage, better filter early

net/redisclient.go
szuecs: Yes you are right we could do that.
fixed

net/redisclient.go
szuecs: yes but we loop here and not there..

szuecs: fixed

dataclients/kubernetes/clusterclient.go
szuecs: marked to remove before merge

config/config.go
AlexanderYastrebov: I mean maybe name it `kubernetes-redis-service-name`

net/redisclient.go
AlexanderYastrebov: > Also a single address can change so this check will skip an update.

What if single redis shard restarts - the len will be the same but address will change

dataclients/kubernetes/clusterstate.go
```diff
@@ -64,11 +64,15 @@ func (state *clusterState) getEndpointsByService(namespace, name, protocol strin
 	return targets
 }
 
+func (state *clusterState) GetEndpointsByTarget(namespace, name, protocol string, target *definitions.BackendPort) []string {
```
AlexanderYastrebov, line 67: I think `getEndpointsByTarget` also could be just made public

dataclients/kubernetes/kube.go
AlexanderYastrebov: Can you check if this is correct? I think it matches port by name which is not provided here and works because there is a single port in redis endpoints

net/redisclient.go
```diff
@@ -226,11 +248,43 @@ func NewRedisRingClient(ro *RedisOptions) *RedisRingClient {
...
+		if ro.AddrUpdater != nil {
+			if ro.UpdateInterval == 0 {
+				ro.UpdateInterval = defaultUpdateInterval
+			}
+			go func() {
+				old := len(ringOptions.Addrs)
+				t := time.NewTicker(ro.UpdateInterval)
+				defer t.Stop()
+				r.log.Info("Start goroutine to update redis instances every %s", ro.UpdateInterval)
+				for range t.C {
```
AlexanderYastrebov, line 261: I think this should `for { }` with a select from  t.C and r.quit, otherwise quit have to wait for another ticker cycle

net/redisclient.go
AlexanderYastrebov: > it would Lock unnecessary and create memory garbage

I dont think it is a big problem, one of the goals of https://github.com/go-redis/redis/pull/2093 IMO is to make `SetAddrs` idempotent to make it easy to use on the caller side (i.e. here).

dataclients/kubernetes/kube.go
szuecs: The code works, I tested the original one in our dev cluster.
Basically it's a copy from routegroup: https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/routegroup.go#L257-L262
The clusterstate is a cache with the cache key `endpointID` as we pass

net/redisclient.go
szuecs: fixed: I split this out

net/redisclient.go
szuecs: Regarding restart you are right I will check the whole list instead of only the number. 

I think we already see problems with linear scalability with number of CPUs in one node and all Locks in the hot path will have something to do with it (my rule of thumb). So in this case I think we can and should be smarter.

szuecs: fixed the restart case

dataclients/kubernetes/clusterstate.go
szuecs: done

net/redisclient.go
szuecs: yes true, done

dataclients/kubernetes/kube.go
AlexanderYastrebov: I want to say that you can put any number instead of `6379` and it will work

AlexanderYastrebov: I think I am wrong and it matches the port by value.
Maybe we can have a third config `KubernetesRedisServicePort` with the default value of 6379, otherwise it would not be possible to run Redis on another port.

net/redisclient.go
```diff
@@ -226,11 +248,67 @@ func NewRedisRingClient(ro *RedisOptions) *RedisRingClient {
 		r.ring = redis.NewRing(ringOptions)
 		r.log = ro.Log
 		r.metricsPrefix = ro.MetricsPrefix
+
+		if ro.AddrUpdater != nil {
+			if ro.UpdateInterval == 0 {
+				ro.UpdateInterval = defaultUpdateInterval
+			}
+			go r.startUpdater(context.Background())
```
AlexanderYastrebov, line 256: `context.Background()` argument is not used

net/redisclient.go
```diff
@@ -226,11 +248,67 @@ func NewRedisRingClient(ro *RedisOptions) *RedisRingClient {
 		r.ring = redis.NewRing(ringOptions)
 		r.log = ro.Log
 		r.metricsPrefix = ro.MetricsPrefix
+
+		if ro.AddrUpdater != nil {
+			if ro.UpdateInterval == 0 {
+				ro.UpdateInterval = defaultUpdateInterval
+			}
```
AlexanderYastrebov, line 255: minor: maybe move this into `startUpdater`

net/redisclient.go
AlexanderYastrebov: minor: maybe it should be `runUpdater`

net/redisclient.go
```diff
@@ -226,11 +248,67 @@ func NewRedisRingClient(ro *RedisOptions) *RedisRingClient {
...
+
+	for {
+		select {
+		case <-r.quit:
+			return
+		case <-time.After(r.options.UpdateInterval):
+		}
+
+		addrs := r.options.AddrUpdater()
+		if !hasAll(addrs, old) {
```
AlexanderYastrebov, line 299: Why not `reflect.DeepEqual`?

AlexanderYastrebov: Nevermind

net/redisclient.go
szuecs: I don't like to put it there because IMO options/defaults should be set in one place.

net/redisclient.go
szuecs: because []string vs map[string]struct{}

net/redisclient.go
szuecs: fixed

net/redisclient.go
```diff
@@ -25,6 +26,16 @@ import (
...
 	Addrs []string
+
+	// AddrUpdater is a func that is regularly called to update
+	// redis address list. This func should return a list of redis
+	// shards.
+	AddrUpdater func() []string
+
+	// UpdateInterval is the time.Duration that AddrUpdater is
+	// triggered and SetAddrs be used to update the redis shards
+	UpdateInterval time.Duration
```
AlexanderYastrebov, line 37: As an idea: alternatively we can have a channel for updates `chan []string` and update addresses when new values arrive. This way external updater function can either send updates at regular intervals or it may be a k8s watch/informer that would only send updates when endpoints change.

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f
 	github.com/dimfeld/httppath v0.0.0-20170720192232-ee938bf73598
 	github.com/ghodss/yaml v1.0.0
-	github.com/go-redis/redis/v8 v8.11.4
```
AlexanderYastrebov, line 19: Maybe it makes sense to update to v9 first

AlexanderYastrebov:
:+1:

szuecs:
:+1:

AlexanderYastrebov:
:+1:

Issue #1963, opened 2022-02-18
Go1.18 Rewrite string parsing with Cut()


szuecs:
Go 1.18: Rewrite string parsing with Cut()

For example tracer option parsing that use Split

bedakb:
I'd like to work on this one.

PR #2012, opened 2022-06-07
refactor: use strings.Cut for strings parsing

bedakb:
This PR uses Go 1.18 `strings.Cut` function and replaces usage of `strings.Split` and `strings.SplitN` where possible.

Resolves https://github.com/zalando/skipper/issues/1963

### DIFF
--- a/filters/auth/oidc_introspection.go
+++ b/filters/auth/oidc_introspection.go
@@ -64,12 +64,12 @@ func (spec *oidcIntrospectionSpec) CreateFilter(args []interface{}) (filters.Fil
 	switch filter.typ {
 	case checkOIDCQueryClaims:
 		for _, arg := range sargs {
-			slice := strings.SplitN(arg, ":", 2)
-			if len(slice) != 2 {
+			path, queries, found := strings.Cut(arg, ":")
+			if !found || path == "" {
 				return nil, fmt.Errorf("%v: malformatted filter arg %s", filters.ErrInvalidFilterParameters, arg)
 			}
-			pq := pathQuery{path: slice[0]}
-			for _, query := range splitQueries(slice[1]) {
+			pq := pathQuery{path: path}
+			for _, query := range splitQueries(queries) {
 				if query == "" {
 					return nil, fmt.Errorf("%v: %s", errUnsupportedClaimSpecified, arg)
 				}
--- a/filters/builtin/compress.go
+++ b/filters/builtin/compress.go
@@ -4,7 +4,6 @@ import (
 	"compress/flate"
 	"compress/gzip"
 	"errors"
-	"github.com/andybalholm/brotli"
 	"io"
 	"math"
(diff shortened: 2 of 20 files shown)


filters/auth/oidc_introspection.go
```diff
@@ -64,12 +64,12 @@ func (spec *oidcIntrospectionSpec) CreateFilter(args []interface{}) (filters.Fil
 	switch filter.typ {
 	case checkOIDCQueryClaims:
 		for _, arg := range sargs {
-			slice := strings.SplitN(arg, ":", 2)
-			if len(slice) != 2 {
+			path, queries, found := strings.Cut(arg, ":")
+			if !found || path == "" {
```
AlexanderYastrebov, line 68: The check for non-empty path might make sense but is out of scope for this refactoring. I suggest we keep existing logic here and elsewhere.

filters/builtin/compress.go
```diff
@@ -13,6 +12,8 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/andybalholm/brotli"
```
AlexanderYastrebov, line 15: While not a big deal, this does not seem to be related to the refactoring.

tracing/tracers/basic/basic.go
```diff
@@ -21,34 +21,34 @@ func InitTracer(opts []string) (opentracing.Tracer, error) {
...
-		switch parts[0] {
+		k, v, _ := strings.Cut(o, "=")
+		switch k {
 		case "drop-all-logs":
 			dropAllLogs = true
 
 		case "sample-modulo":
-			if len(parts) == 1 {
-				return nil, missingArg(parts[0])
+			if v == "" {
```
AlexanderYastrebov, line 30: Note: for equivalent refactoring it should check for `found` value as `v` could be empty, e.g. for `strings.Cut("foo=", "=")`.
Here and in most other places it is fine to check for empty `v` as empty value should not be permitted.

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
@bedakb Thank you for your contribution.

tracing/tracers/basic/basic.go
bedakb: I think `v == ""` must be kept regardless `found` is `true` or `false`. This particular example `strings.Cut("foo=", "=")` will return `"foo", "", true`.
Also It covers this case `strings.Cut("drop-all-logs", "=")` -> `"drop-all-logs", "", false`, so If we'd introduce a `found` check what should be do in case `found == false` ? Is doing a `continue` enough, so we skip to the next option ?

filters/builtin/compress.go
bedakb: ah sorry, that's just my IDE sorting the imports

szuecs:
:+1:

PR #2013, opened 2022-06-08
feature: add xff headers that are required by authelia

szuecs:
feature: add xff headers that are required by authelia, see also https://github.com/authelia/authelia/issues/3492

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -397,6 +397,8 @@ func NewConfig() *Config {
 	flag.Var(cfg.ForwardedHeadersList, "forwarded-headers", "comma separated list of headers to add to the incoming request before routing\n"+
 		"X-Forwarded-For sets or appends with comma the remote IP of the request to the X-Forwarded-For header value\n"+
 		"X-Forwarded-Host sets X-Forwarded-Host value to the request host\n"+
+		"X-Forwarded-Method sets X-Forwarded-Method value to the request method\n"+
+		"X-Forwarded-Uri sets X-Forwarded-Uri value to the requestURI\n"+
 		"X-Forwarded-Port=<port> sets X-Forwarded-Port value\n"+
 		"X-Forwarded-Proto=<http|https> sets X-Forwarded-Proto value")
 	flag.Var(cfg.ForwardedHeadersExcludeCIDRList, "forwarded-headers-exclude-cidrs", "disables addition of forwarded headers for the remote host IPs from the comma separated list of CIDRs")
@@ -964,6 +966,10 @@ func (c *Config) parseForwardedHeaders() error {
 			c.ForwardedHeaders.For = true
 		case header == "X-Forwarded-Host":
 			c.ForwardedHeaders.Host = true
+		case header == "X-Forwarded-Method":
+			c.ForwardedHeaders.Method = true
+		case header == "X-Forwarded-Uri":
+			c.ForwardedHeaders.Uri = true
 		case strings.HasPrefix(header, "X-Forwarded-Port="):
 			c.ForwardedHeaders.Port = strings.TrimPrefix(header, "X-Forwarded-Port=")
 		case header == "X-Forwarded-Proto=http":
--- a/net/headers.go
+++ b/net/headers.go
@@ -5,15 +5,19 @@ import (
 	"net/http"
 )
 
-// Sets non-standard X-Forwarded-* Headers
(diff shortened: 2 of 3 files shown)


MustafaSaber (approved):
👍

config/config.go
```diff
@@ -397,6 +398,9 @@ func NewConfig() *Config {
 	flag.Var(cfg.ForwardedHeadersList, "forwarded-headers", "comma separated list of headers to add to the incoming request before routing\n"+
 		"X-Forwarded-For sets or appends with comma the remote IP of the request to the X-Forwarded-For header value\n"+
 		"X-Forwarded-Host sets X-Forwarded-Host value to the request host\n"+
+		"X-Forwarded-Method sets X-Forwarded-Method value to the request method\n"+
+		"X-Forwarded-Uri sets X-Forwarded-Uri value to the requestURI\n"+
+		"X-Original-URL sets X-Original-URL to the absolute original request URL, requires X-Forwarded-Proto to be set\n"+
```
AlexanderYastrebov, line 403: Should it be `Uri` and `Url` or `URI` and `URL` maybe?

config/config.go
```diff
@@ -975,6 +985,10 @@ func (c *Config) parseForwardedHeaders() error {
 		}
 	}
 
+	if c.ForwardedHeaders.Proto == "" && c.ForwardedHeaders.OriginalURL {
+		return errors.New("invalid forwarded header configuration X-Original-URL requires X-Forwarded-Proto")
```
AlexanderYastrebov, line 989: Lets use `fmt.Errorf` like elsewhere

net/headers.go
```diff
@@ -42,12 +48,28 @@ func (h *ForwardedHeaders) Set(req *http.Request) {
...
+	}
+
 	if h.Port != "" {
 		req.Header.Set("X-Forwarded-Port", h.Port)
 	}
 
 	if h.Proto != "" {
 		req.Header.Set("X-Forwarded-Proto", h.Proto)
+
+		if h.OriginalURL {
```
AlexanderYastrebov, line 66: What would be the difference with `req.URL.String()`?

szuecs: h.OriginalURL is a bool or do you comment on the wrong line?

AlexanderYastrebov: I mean can we use `req.URL.String()` instead of `h.Proto+"://"+req.Host+":"+h.Port+req.RequestURI`?

config/config.go
szuecs: We use `http.CanonicalHeaderKey` in `copyHeader()`, so this would be a breaking change to change this behavior if someone implicitly relies on it.

config/config.go
AlexanderYastrebov: It is ok to use `fmt.Errorf` without arguments, see e.g. https://github.com/golang/go/issues/17173
Since both are equivalent it makes sense to stick to the one used elsewhere (`errors` package was not even imported before this chnage)

config/config.go
AlexanderYastrebov: Yeah, the actual header names would be canonical
```
X-Forwarded-Uri: /foo
X-Original-Url: https://127.0.0.1:9090/foo
```
so lets make flag names canonical as well fro consistency.

net/headers.go
AlexanderYastrebov: Also `req.Host` can contain `host:port` so the else branch may produce incorrect URL.

AlexanderYastrebov: The comment https://github.com/authelia/authelia/issues/3492#issuecomment-1150505829 suggests that 
> Probably best to ignore the X-Original-URL in that case

so maybe we just have `X-Forwarded-Uri` if it is enough and not bother reconstructing URL

config/config.go
szuecs: % grep -cr errors.New * | awk -F : 'BEGIN{i=0} {i=i+$2} END{print "count: "i}'
count: 325
% grep -cr fmt.Errorf * | awk -F : 'BEGIN{i=0} {i=i+$2} END{print "count: "i}'
count: 414

config/config.go
szuecs: X-Original-Url was dropped, so nothing to do

config/config.go
szuecs: X-Original-URL was dropped so nothing to do

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2019, opened 2022-06-16
Add instructions to build multi-arch container image

linki:
This adds build steps to build a new multi-arch container image for `skipper` (and tools) for `linux/amd64` and `linux/arm64`.

The heart of this change is the additional Makefile step called `docker.push.multiarch` which builds a multi-arch image using `buildx` by providing multiple platform arguments.

In order to do that I needed to change several parts of the Makefile but tried to keep everything backwards compatible. I would prefer to make the multi-arch build the single container image artifact and remove the separate Dockerfiles for `.arm` and `.arm64`. However, currently we don't have a public registry that supports multi-arch images where we publish skipper to, so I leave them in place for now.

The new multi-arch image is pushed to our private registry for now and since its base image only supports `amd64` and `arm64` at the moment we don't add the `arm/v7` variant because we don't need that at Zalando.

The final multi-arch image will live at `container-registry.zalando.net/teapot/skipper`.

There are some small changes though:
* The official image for skipper (`registry.opensource.zalan.do/teapot/skipper`) used to be build by the task that is now called `docker.build`. This task builds binaries in the local architecture into the current working directory and uses `docker build` to stick them into the Dockerfile. This is basically the upper half of the Makefile. However, the lower half of the Makefile consists of several more tasks to build binaries for all of the supported architectures. The official docker image is now build from these Makefile steps as well as using `buildx`. The resulting binaries and images are the same and one can think of removing the upper half of the Makefile now.
* All the binaries in the build process are now build by the tasks in the lower half of the Makefile (`build.linux.amd64` etc.). This made it very easy to build all the binaries for all architectures and then putting them all into a single multi-arch image. (Well, actually just amd64 and arm64 but you get the idea.)
* Therefore, the folder structure of the output `build/` folder has changed to be more consistent with the `--platform` variable of docker and looks like this now:
```console
[packaging] $ tree build
build
├── darwin
│   ├── amd64
│   │   ├── eskip
│   │   ├── routesrv
│   │   ├── skipper
│   │   └── webhook
│   └── arm64
│       ├── eskip
│       ├── routesrv
│       ├── skipper
│       └── webhook
├── linux
│   ├── amd64
│   │   ├── eskip
│   │   ├── routesrv
│   │   ├── skipper
│   │   └── webhook
│   ├── arm
│   │   └── v7
│   │       ├── eskip
│   │       ├── routesrv
│   │       ├── skipper
│   │       └── webhook
│   └── arm64
│       ├── eskip
│       ├── routesrv
│       ├── skipper
│       └── webhook
└── windows
    └── amd64
        ├── eskip
        ├── routesrv
        ├── skipper
        └── webhook

10 directories, 24 files
```
* I also added build steps for `darwin/arm64` for the new M1/2 macbooks.
* The naming of the *.tar.gz build artifacts have been slightly changed to be more consistent. They also contain _all_ the binaries and not just skipper with this change. 

If you want to check that all the binaries are build correctly, you can run `make build.package`.

Test images from a previous successful run can be found here:
* `registry.opensource.zalan.do/teapot/skipper-test:pr-2019-7` (single-arch amd64)
* `container-registry-test.zalando.net/teapot/skipper-test:pr-2019-7` (multi-arch)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ endif
 
 build: $(SOURCES) lib skipper eskip webhook routesrv
 
-build.linux.armv8:
+build.linux.arm64:
 	GOOS=linux GOARCH=arm64 CGO_ENABLED=0 go build -o bin/skipper -ldflags "-X main.version=$(VERSION) -X main.commit=$(COMMIT_HASH)" ./cmd/skipper
 
 build.linux.armv7:
@@ -47,7 +47,10 @@ build.linux.armv7:
 build.linux:
 	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o bin/skipper -ldflags "-X main.version=$(VERSION) -X main.commit=$(COMMIT_HASH)" ./cmd/skipper
 
-build.osx:
+build.darwin.arm64:
+	GOOS=darwin GOARCH=arm64 CGO_ENABLED=0 go build -o bin/skipper -ldflags "-X main.version=$(VERSION) -X main.commit=$(COMMIT_HASH)" ./cmd/skipper
+
+build.darwin:
 	GOOS=darwin GOARCH=amd64 CGO_ENABLED=0 go build -o bin/skipper -ldflags "-X main.version=$(VERSION) -X main.commit=$(COMMIT_HASH)" ./cmd/skipper
 
 build.windows:
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -34,10 +34,7 @@ pipeline:
       make deps cicheck staticcheck gosec
       git status
       git diff
-
-      cd packaging
-      make docker.build.amd64 && git status && git diff && make docker.push.amd64
-      make docker.push.multiarch
+      cd packaging && make docker.build.amd64 && git status && git diff && make docker.push.amd64 && make docker.push.multiarch
       if [[ $CDP_TARGET_BRANCH == master && ! $CDP_PULL_REQUEST_NUMBER ]]; then
         echo "Created docker image registry.opensource.zalan.do/teapot/skipper:${RELEASE_VERSION}"
         cdp-promote-image "${MULTIARCH_IMAGE}"
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -9,6 +9,7 @@ pipeline:
   - desc: build-push
     cmd: |
       IMAGE_REGISTRY="registry-write.opensource.zalan.do"
+      MULTIARCH_REGISTRY="container-registry-test.zalando.net"
       if [[ $CDP_TARGET_BRANCH == master && ! $CDP_PULL_REQUEST_NUMBER ]]; then
         LATEST_VERSION=$(git describe --tags --always | awk -F \- '{print $1}')
         CUR_PART=$(echo $LATEST_VERSION | awk -F . '{print $1"."$2}')
@@ -21,19 +22,26 @@ pipeline:
         IMAGE="${IMAGE_REGISTRY}/teapot/skipper:${RELEASE_VERSION}"
         ARM_IMAGE="${IMAGE_REGISTRY}/teapot/skipper-armv7:${RELEASE_VERSION}"
         ARM64_IMAGE="${IMAGE_REGISTRY}/teapot/skipper-arm64:${RELEASE_VERSION}"
+        MULTIARCH_IMAGE="${MULTIARCH_REGISTRY}/teapot/skipper:${RELEASE_VERSION}"
       else
         IMAGE="${IMAGE_REGISTRY}/teapot/skipper-test:${CDP_BUILD_VERSION}"
         ARM_IMAGE="${IMAGE_REGISTRY}/teapot/skipper-armv7-test:${CDP_BUILD_VERSION}"
         ARM64_IMAGE="${IMAGE_REGISTRY}/teapot/skipper-arm64-test:${CDP_BUILD_VERSION}"
+        MULTIARCH_IMAGE="${MULTIARCH_REGISTRY}/teapot/skipper-test:${CDP_BUILD_VERSION}"
       fi
-      export IMAGE ARM_IMAGE ARM64_IMAGE
+      export IMAGE ARM_IMAGE ARM64_IMAGE MULTIARCH_IMAGE
 
       make deps cicheck staticcheck gosec
       git status
       git diff
-      cd packaging && make docker.build.amd64 && git status && git diff && make docker.push.amd64
+
+      cd packaging
+      make docker.build.amd64 && git status && git diff && make docker.push.amd64
+      make docker.push.multiarch
       if [[ $CDP_TARGET_BRANCH == master && ! $CDP_PULL_REQUEST_NUMBER ]]; then
         echo "Created docker image registry.opensource.zalan.do/teapot/skipper:${RELEASE_VERSION}"
+        cdp-promote-image "${MULTIARCH_IMAGE}"
+        echo "Created multi-arch docker image container-registry.zalando.net/teapot/skipper:${RELEASE_VERSION}"
         echo "Creating docker image registry.opensource.zalan.do/teapot/skipper-arm64:${RELEASE_VERSION}"
         make docker.build.arm64 && git status && git diff && make docker.push.arm64
         echo "Creating docker image registry.opensource.zalan.do/teapot/skipper-armv7:${RELEASE_VERSION}"
@@ -62,6 +70,7 @@ pipeline:
         git gh-release --message-from-file "${tf}" "${files[@]}" "$RELEASE_VERSION"
       else
         echo "Created docker image registry.opensource.zalan.do/teapot/skipper-test:${CDP_BUILD_VERSION}"
+        echo "Created multi-arch docker image container-registry-test.zalando.net/teapot/skipper-test:${CDP_BUILD_VERSION}"
         echo "Not creating a release. No release version defined."
       fi
 - id: docs
--- a/packaging/Dockerfile
+++ b/packaging/Dockerfile
@@ -1,8 +1,14 @@
-FROM registry.opensource.zalan.do/library/alpine-3.13:latest
+ARG BASE_IMAGE=registry.opensource.zalan.do/library/alpine-3.13:latest
+FROM ${BASE_IMAGE}
 LABEL maintainer="Team Teapot @ Zalando SE <team-teapot@zalando.de>"
 RUN apk --no-cache add ca-certificates && update-ca-certificates
 RUN mkdir -p /usr/bin
-ADD skipper eskip webhook routesrv /usr/bin/
+ARG BUILD_FOLDER=build
+ARG TARGETPLATFORM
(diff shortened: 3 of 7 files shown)


Makefile
```diff
@@ -47,7 +47,10 @@ build.linux.armv7:
 build.linux:
 	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o bin/skipper -ldflags "-X main.version=$(VERSION) -X main.commit=$(COMMIT_HASH)" ./cmd/skipper
 
-build.osx:
+build.darwin.arm64:
```
linki, line 50: renamed `osx` to `darwin` to match `GOOS` and docker's `--platform` variable.

delivery.yaml
```diff
@@ -21,19 +22,26 @@ pipeline:
...
         ARM64_IMAGE="${IMAGE_REGISTRY}/teapot/skipper-arm64:${RELEASE_VERSION}"
+        MULTIARCH_IMAGE="${MULTIARCH_REGISTRY}/teapot/skipper:${RELEASE_VERSION}"
       else
         IMAGE="${IMAGE_REGISTRY}/teapot/skipper-test:${CDP_BUILD_VERSION}"
         ARM_IMAGE="${IMAGE_REGISTRY}/teapot/skipper-armv7-test:${CDP_BUILD_VERSION}"
         ARM64_IMAGE="${IMAGE_REGISTRY}/teapot/skipper-arm64-test:${CDP_BUILD_VERSION}"
+        MULTIARCH_IMAGE="${MULTIARCH_REGISTRY}/teapot/skipper-test:${CDP_BUILD_VERSION}"
       fi
-      export IMAGE ARM_IMAGE ARM64_IMAGE
+      export IMAGE ARM_IMAGE ARM64_IMAGE MULTIARCH_IMAGE
```
linki, line 32: I think we can collapse this into a single image at some point. But for now I would just do it like this.

delivery.yaml
```diff
@@ -21,19 +22,26 @@ pipeline:
...
 
       make deps cicheck staticcheck gosec
       git status
       git diff
-      cd packaging && make docker.build.amd64 && git status && git diff && make docker.push.amd64
+
+      cd packaging
+      make docker.build.amd64 && git status && git diff && make docker.push.amd64
+      make docker.push.multiarch
```
linki, line 40: This builds and pushes a test image.

packaging/Dockerfile
```diff
@@ -1,8 +1,14 @@
-FROM registry.opensource.zalan.do/library/alpine-3.13:latest
+ARG BASE_IMAGE=registry.opensource.zalan.do/library/alpine-3.13:latest
```
linki, line 1: Let's keep the public single-arch approved base image as default for now since this project is open source.

packaging/Dockerfile
```diff
@@ -1,8 +1,14 @@
...
 LABEL maintainer="Team Teapot @ Zalando SE <team-teapot@zalando.de>"
 RUN apk --no-cache add ca-certificates && update-ca-certificates
 RUN mkdir -p /usr/bin
-ADD skipper eskip webhook routesrv /usr/bin/
+ARG BUILD_FOLDER=build
+ARG TARGETPLATFORM
+ADD ${BUILD_FOLDER}/${TARGETPLATFORM}/skipper \
+    ${BUILD_FOLDER}/${TARGETPLATFORM}/eskip \
+    ${BUILD_FOLDER}/${TARGETPLATFORM}/webhook \
+    ${BUILD_FOLDER}/${TARGETPLATFORM}/routesrv /usr/bin/
```
linki, line 11: This is not perfectly backwards compatible because of `BUILD_FOLDER=build` instead of `BUILD_FOLDER=.`. However, the corresponding task in the Makefile provides that.

Changing the default to `.` would require the majority of `docker build` invocations to overwrite it. Having the default like this only requires one task to overwrite it.

packaging/Dockerfile.arm64
```diff
@@ -1,10 +1,10 @@
-FROM --platform=linux/amd64 alpine:3.13 AS build
+FROM --platform=linux/arm64 alpine:3.13
```
linki, line 1: `--platform` is not strictly required but it ensures that we use the correct architecture for the added arm binaries below.

packaging/Dockerfile.armv7
```diff
@@ -1,10 +1,10 @@
-FROM --platform=linux/amd64 alpine:3.13 AS build
+FROM --platform=linux/arm/v7 alpine:3.13
```
linki, line 1: `--platform` is not strictly required but it ensures that we use the correct architecture for the added `arm` binaries below.

packaging/Makefile
```diff
@@ -52,6 +56,17 @@ docker.push.arm64: docker.build.arm64
 docker.push.armv7: docker.build.armv7
 	docker push $(ARM_IMAGE)
 
+docker.push.multiarch: clean build.linux docker.build.enable
+	# Install qemu interpreter for arm64 (https://docs.docker.com/buildx/working-with-buildx/#build-multi-platform-images)
+	docker run --rm --privileged container-registry.zalando.net/teapot/tonistiigi-binfmt:qemu-v6.2.0-main-2 --install arm64
```
linki, line 61: Please see https://github.bus.zalan.do/automata/issues/issues/4491 for why this is required and for any progress which could make this unnecessary.

packaging/Makefile
```diff
@@ -52,6 +56,17 @@ docker.push.arm64: docker.build.arm64
...
+
+	# create a Buildkit builder with CDP specific configuration (https://cloud.docs.zalando.net/howtos/cdp-multiarch/)
+	docker buildx create --config /etc/cdp-buildkitd.toml --driver-opt network=host --bootstrap --use
+
+	# build multi-arch container image using a trusted multi-arch base image
+	docker buildx build --rm -t $(MULTIARCH_IMAGE) --platform linux/amd64,linux/arm64 --push \
+	  --build-arg BASE_IMAGE=container-registry.zalando.net/library/alpine-3.13:latest .
```
linki, line 68: Here we use a proper multi-arch base image that's approved for production. Since it's currently not public it's not the default yet and only used from here.

delivery.yaml
AlexanderYastrebov: The use of `&&` seems arbitrary - should we maybe `&&` all of the commands or have each on the single line for clarity?

linki:
Note, that I needed to use a third party image (mirrored to our registry) in order to initialize proper arm64 emulation (without it, it doesn't work). It's the very same approach that GitHub uses for GitHub Actions so it's not bad. However, there's active discussion on what causes the issues and what's the best approach going forward here: https://github.bus.zalan.do/automata/issues/issues/4491.

For now, I hope we can also make progress with the current approach. Since it only affects the private multi-arch image we can easily track where it's used.

packaging/Dockerfile.arm64
```diff
@@ -1,10 +1,10 @@
...
 LABEL maintainer="Team Teapot @ Zalando SE <team-teapot@zalando.de>"
 RUN apk --no-cache add ca-certificates && update-ca-certificates
-
-FROM --platform=linux/arm64 alpine:3.13
-COPY --from=build /etc/ssl/certs/ /etc/ssl/certs/
```
AlexanderYastrebov, line 6: I guess this was there to not `RUN apk --no-cache add ca-certificates && update-ca-certificates` on the result image.

packaging/Makefile
```diff
@@ -60,48 +75,65 @@ docker.build.enable:
...
+build.windows: $(addprefix build/windows/amd64/,$(BINARIES))
 
-build.linux.amd64:
+build/linux/amd64/%:
 	GO111MODULE=$(GO111) \
 	GOOS=linux \
 	GOARCH=amd64 \
 	CGO_ENABLED=$(CGO_ENABLED) \
-	go build -ldflags "-X main.version=$(VERSION) -X main.commit=$(COMMIT_HASH)" -o build/linux/skipper ../cmd/skipper/*.go
+	go build -ldflags "-X main.version=$(VERSION) -X main.commit=$(COMMIT_HASH)" -o build/linux/amd64/$(notdir $@) ../cmd/$(notdir $@)/*.go
```
AlexanderYastrebov, line 94: What `$(notdir $@)` is for? Or rather how did it work before - looks like only skipper binary was built...

delivery.yaml
szuecs: I think besides this here everything looks fine

packaging/Makefile
```diff
@@ -34,15 +35,18 @@ routesrv:
 	GO111MODULE=$(GO111) GOOS=$(GOOS) GOARCH=$(GOARCH) $(GOARM) CGO_ENABLED=$(CGO_ENABLED) go build -o routesrv ../cmd/routesrv/*.go
 
 clean:
-	rm -rf skipper eskip webhook routesrv build/
+	rm -rf $(BINARIES) build/
+
+docker.build: clean $(BINARIES)
```
AlexanderYastrebov, line 40: How does this relate to `docker-build` (with a dash) target?

packaging/Dockerfile.arm64
linki: It was. I added it recently to make a previous try work: https://github.com/zalando/skipper/pull/1954/files#diff-d644f844087e1cb2b3e229fef8aec05cfa86372250e0dfae56a9c87140c356f1.

This just restores the original version.

packaging/Makefile
linki: Correct! Only the skipper binary was built. You can double-check in the latest release tarballs: https://github.com/zalando/skipper/releases/download/v0.13.213/skipper-v0.13.213-linux-amd64.tar.gz

The docker image was built using other steps (further up) that build all binaries.

`notdir` is a handy way to extract the last part of the current Makefile's target, e.g. here it would be the plain binary name.

E.g. for the task `build/linux/amd64/skipper` `notdir` is `skipper` and `dir` would be `build/linux/amd64/` (https://www.gnu.org/software/make/manual/html_node/File-Name-Functions.html)

packaging/Makefile
linki: It came from this version: https://github.com/zalando/skipper/blob/41b621e7ea402c5f176456609e6c05d292bb20f8/packaging/Makefile#L40-L45 which is a bit inconsistent (see usage of build and buildx)

With this: `docker.build` is the "local" build (no architecture suffix, builds for local architecture) and uses the steps defined before.

Whereas `docker.build.amd64` etc. cross-compile for the target architectures using the steps at the bottom and buildx. `docker-build` and `docker-push` are wrappers to to invoke all three builds for the linux OS.

We could rename `docker.build` to `docker.build.local` or similar to avoid confusion.

delivery.yaml
linki: I separated the `cd packaging` from this particular command because _all_ subsequent commands expect to be run within that folder. That was hard to spot when it's lumped together with the first build.

I'm going with `&& \` all commands then because that's what you both agree on. But I wonder: putting each command on its own line should have the same affect since the pipeline will stop when one command fails.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2020, opened 2022-06-17
Make sure to import the built image to the local daemon

linki:
When using `buildx` the images stay in the remote build cache. So you either have to use `--push` to push them directly or use `--load` to import them to the local daemon and `docker push` from there.

Curiously, this worked fine for the `amd64` version but failed for the arm ones. I think the amd64 build is run by the local daemon (since it's the native platform) whereas the arm ones are used by the separate buildkit builder, hence it failed there. 

Let's do it correctly by loading the image to the local daemon first. Note, `--load` only works with single-arch images, which is the case here.


linki:
👍

AlexanderYastrebov:
:+1:

PR #2021, opened 2022-06-17
dataclients/kubernetes: preserve `RouteGroup` filters in shunt route

AlexanderYastrebov:
This change makes RouteGroup logic consistent with Ingress logic:

- Ingress logic prepends annotation and default filters after route is shunted
- RouteGroup adds user-configured filters before processing the backend and
  therefore `shuntRoute` removed user-defined filter.

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -361,16 +361,16 @@ func (c *Client) loadAndConvert() ([]*eskip.Route, error) {
 }
 
 func shuntRoute(r *eskip.Route) {
-	r.Filters = []*eskip.Filter{
-		{
+	r.Filters = append(r.Filters,
+		&eskip.Filter{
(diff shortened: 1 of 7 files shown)


AlexanderYastrebov:
An alternative is to ignore `Ingress` user filters when route is shunted, see #2024

AlexanderYastrebov:
I think it might make sense to preserve `RouteGroup` filters (those could e.g. `oauthTokeninfoAllScope`, `enableAccessLog` or `tracingSpanName`) in case endpoints are missing. It also feels more consistent - `shuntBackend` just changes the backend without doing anything else.

szuecs:
:-1:

szuecs:
As discussed internally I would like to rather change ingress to the routegroup way.

AlexanderYastrebov:
Closing in favour of https://github.com/zalando/skipper/pull/2024

PR #2022, opened 2022-06-17
Install arm (32-bit) emulator to build armv7 image

linki:
This must be installed if we want to build the armv7 image.


AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2023, opened 2022-06-17
Install all qemu interpreters before the build runs

linki:
Just moves the qemu interpreter installation before the build starts and out of the Makefile.

Before it only worked for armv7 because the corresponding Makefile steps runs after the one for multiarch. With this, the order doesn't matter and it's also cleaner to not have this in the Makefile but only in the (already CDP related) delivery.yaml.


linki:
👍

szuecs:
:+1:

PR #2028, opened 2022-06-20
fix: query

szuecs:
fix: Query predicate could be bypassed by prepared request, config is enabled by default and you can disable with -validate-query=false or


AlexanderYastrebov:
:+1:

AlexanderYastrebov:
:+1:

szuecs:
:+1:

Issue #2004, opened 2022-05-13
As a software operator, I want Skipper to shed traffic that I am unable to handle


andrewhowdencom:
**Is your feature request related to a problem? Please describe.**

When I am operating a service (particularly with skipper-ingress), that service will sometimes have a sudden and substantial increase in traffic. This can happen in a few situations:

1. An increase in traffic at the edge (such as abusive traffic)
2. An upstream service timing out, and retrying aggressively

The traffic frequently increases very large amounts; 100 → 200 or 300%. This increases the latency of the service (either through overloading that service directly, or through overloading the downstream services) after which the service eventually times out.

**Describe the solution you would like**

Skipper can parse the metadata associated with HTTP responses to determine the appropriate throughput of any given downstream system. By keeping a count of the:

- The HTTP Status code (e.g. HTTP 503)
- The HTTP Header (e.g. Retry-After, or similar)

Skipper can effectively understand whether the service is presently overloaded. Skipper can, in parallel, limit the throughput to any given "route" — in this case a pod.

Skipper can adjust the throughput toward a pod based on how overloaded the system is (as expected by the system returning a HTTP 503).

Related:

1. Netflix, Adaptive Concurrency: https://netflixtechblog.medium.com/performance-under-load-3e6fa9a60581
2. Envoy, Admission Control: https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/admission_control_filter

**Describe alternatives you've considered (optional)**

Alternative solutions include:

`backendRateLimit()` — This allows setting a limit per pod, but assumes that pods are stable and that requests are equivalent. Where the workload changes substantially (e.g. larger requests that are more computationally expensive), the effective capacity of any given pod goes down.

An adaptive solution would factor this in, and automatically reject traffic that could not be propagated. It would also work _blindly_ — that is, when a downstreams downstream was overloaded (as the HTTP 503 status would be propagated back).

**Additional context (optional)**
_Monitoring_
Should this be deployed, it will make the state of the system harder to reason about as the throughput can vary either per route or per route group. Given this, it should be possible to either dump (via SIGQUIT or similar) or export as a metric the current calculated throughput per route. 

This could be exported as a metric, but perhaps not by default as the cardinality is very high.

_Existing Systems_
While HTTP 503 can be associated with overload, it is also associated with other conditions that are more binary (such as pod ready). Given this, the design should probably match requests by both a HTTP status code and a header. For example, 

```
adaptiveRateLimit("503", "retry-after")
[...]

PR #2030, opened 2022-06-22
feature: admissionControl filter

szuecs:
load shedding filter admissionControl, part of https://github.com/zalando/skipper/issues/2004

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -372,8 +372,8 @@ no-compression, 1 means best-speed and 11 means best-compression. Example:
 ```
 
 The filter also checks the incoming request, if it accepts the supported encodings,
-explicitly stated in the Accept-Encoding header. 
-The filter currently supports by default `gzip`, `deflate` and `br` (can be overridden with flag `compress-encodings`). 
+explicitly stated in the Accept-Encoding header.
+The filter currently supports by default `gzip`, `deflate` and `br` (can be overridden with flag `compress-encodings`).
 It does not assume that the client accepts any encoding if the
 Accept-Encoding header is not set. It ignores * in the Accept-Encoding header.
 
@@ -1848,6 +1848,78 @@ Path("/cheap")     -> clusterLeakyBucketRatelimit("user-${request.cookie.Authori
 Path("/expensive") -> clusterLeakyBucketRatelimit("user-${request.cookie.Authorization}", 1, "1s", 5, 2) -> ...
 ```
 
+## shedder
+
+The basic idea of load shedding is to reduce errors by early stopping
+some of the ingress requests that create too much load and serving
+the maximum throughput the system can process at a point in time.
+
+There is a great talk by [Acacio Cruz from
+Google](https://www.youtube.com/watch?v=XNEIkivvaV4)
+that explains the basic principles.
+
+### admissionControl
+
+Implements an admission control filter, that rejects traffic by
+observed error rate and probability.
+
+The probability of rejection is calculated by the following equation:
+
+$$ P_{reject} = ( { n_{total} - { n_{success} \over threshold } \over n_{total} + 1} )^{ exponent } $$
+
+Examples:
+
+    admissionControl(metricSuffix, mode, d, windowSize, minRPS, successThreshold, maxRejectProbability, exponent)
+    admissionControl("myapp", "active", "1s", 5, 10, 0.95, 0.9, 0.5)
+
+Parameters:
+
+* metric suffix (string)
+* mode (enum)
+* d (time.Duration)
+* window size (int)
+* minRps (int)
+* success threshold (float64)
+* max reject probability (float64)
+* exponent (float64)
+
+Metric suffix is the chosen suffix key to expose reject counter,
+should be unique by filter instance
+
+Mode has 3 different possible values:
+
+* "active" will reject traffic
+* "inactive" will never reject traffic
+* "logInactive" will not reject traffic, but log to debug filter settings
+
+D the time duration of a single slot for required counters in our
+circular buffer of window size.
+
+Window size is the size of the circular buffer. It is used to snapshot
+counters to calculate total requests and number of success. It is
+within $[1, 100]$.
+
+MinRps is the minimum requests per second that have to pass this filter
+otherwise it will not reject traffic.
+
+Success threshold sets the lowest request success rate at which the
+filter will not reject requests. It is within $(0,1]$. A value of
+0.95 means an error rate of lower than 5% will not trigger
+rejects.
+
+Max reject probability sets the upper bound of reject probability. It
+is within (0,1]. A value of 0.95 means if backend errors with 100% it
+will only reject up to 95%.
+
+exponent is used to dictate the rejection probability. The
+calculation is done by $p = p^{exponent}$
+The exponent value is within $(0,\infty]$, to increase rejection
+probability you have to use values lower than 1:
+
+* 1: linear
+* 1/2: quadratic
+* 1/3: cubic
+
 ## lua
 
 See [the scripts page](scripts.md)
@@ -2557,7 +2629,7 @@ fadeIn("3m", 1.5)
 
 #### Warning on fadeIn and Rolling Restarts
 
-Traffic fade-in has the potential to skew the traffic to your backend pods in case of a rolling restart 
+Traffic fade-in has the potential to skew the traffic to your backend pods in case of a rolling restart
 (`kubectl rollout restart`), because it is very likely that the rolling restart is going faster than the
 fade-in duration. The image below shows an example of a rolling restart for a four-pod deployment (A, B, C, D)
 into (E, F, G, H), and the traffic share of each pod over time. While the ramp-up of the new pods is ongoing,
@@ -2565,7 +2637,7 @@ the remaining old pods will receive a largely increased traffic share (especiall
 example), as well as an over-propotional traffic share for the first pod in the rollout (E).
 
 To make rolling restarts safe, you need to slow them down by setting `spec.minReadySeconds` on the pod spec
-of your deployment or stackset, according to your fadeIn duration. 
+of your deployment or stackset, according to your fadeIn duration.
 
 ![Rolling Restart and Fade-In](../img/fadein_traffic_skew.png)
 
--- a/filters/shedder/admission.go
+++ b/filters/shedder/admission.go
@@ -97,43 +97,45 @@ type Options struct {
 
 type admissionControlPre struct{}
 
-// Do removes duplcate filters, because we can only handle one in a
+// Do removes duplicate filters, because we can only handle one in a
 // chain. The last one will override the others.
 func (spec *admissionControlPre) Do(routes []*eskip.Route) []*eskip.Route {
 	for _, r := range routes {
 		foundAt := -1
-		toDelete := make([]int, 0)
+		toDelete := make(map[int]struct{})
+
 		for i, f := range r.Filters {
 			if f.Name == filters.AdmissionControlName {
 				if foundAt != -1 {
-					toDelete = append(toDelete, foundAt)
+					toDelete[foundAt] = struct{}{}
 				}
 				foundAt = i
 			}
 		}
 
-		// loop from last to be safe with indexes
-		for i := len(toDelete) - 1; i >= 0; i-- {
-			// drop filter at position i
-			filters := append(r.Filters[i+1 : len(r.Filters)]) // maybe off by one if i is last
-			r.Filters = append(r.Filters[0:i], filters...)
+		if len(toDelete) == 0 {
+			continue
 		}
+
+		rf := make([]*eskip.Filter, 0, len(r.Filters)-len(toDelete))
+		for i, f := range r.Filters {
+			if _, ok := toDelete[i]; !ok {
+				rf = append(rf, f)
+			}
+		}
+		r.Filters = rf
 	}
 
 	return routes
 }
 
 type admissionControlPost struct {
-	mu      sync.Mutex // required for testing
 	filters map[string]*admissionControl
 }
 
 // Do implements routing.PostProcessor and makes it possible to close goroutines.
 func (spec *admissionControlPost) Do(routes []*routing.Route) []*routing.Route {
 	inUse := make(map[string]struct{})
-
-	spec.mu.Lock()
-	defer spec.mu.Unlock()
 
 	for _, r := range routes {
 
--- a/filters/shedder/admission.go
+++ b/filters/shedder/admission.go
@@ -1,0 +1,445 @@
+package shedder
+
+import (
+	"context"
+	"math"
+	"math/rand"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
+	log "github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/metrics"
+	"github.com/zalando/skipper/routing"
+	"go.uber.org/atomic"
+)
+
+func getIntArg(a interface{}) (int, error) {
+	if i, ok := a.(int); ok {
+		return i, nil
+	}
+
+	if f, ok := a.(float64); ok {
+		return int(f), nil
+	}
+
+	return 0, filters.ErrInvalidFilterParameters
+}
+
+func getDurationArg(a interface{}) (time.Duration, error) {
+	if s, ok := a.(string); ok {
+		return time.ParseDuration(s)
+	}
+	return 0, filters.ErrInvalidFilterParameters
+}
+
+func getFloat64Arg(a interface{}) (float64, error) {
+	if f, ok := a.(float64); ok {
+		return f, nil
+	}
+
+	return 0, filters.ErrInvalidFilterParameters
+}
+
+func getModeArg(a interface{}) (mode, error) {
+	s, ok := a.(string)
+	if !ok {
+		return 0, filters.ErrInvalidFilterParameters
+	}
+	switch s {
+	case "active":
+		return active, nil
+	case "inactive":
+		return inactive, nil
+	case "logInactive":
+		return logInactive, nil
+	}
+
+	return 0, filters.ErrInvalidFilterParameters
+}
+
+type mode int
+
+const (
+	inactive mode = iota + 1
+	logInactive
+	active
+)
+
+func (m mode) String() string {
+	switch m {
+	case active:
+		return "active"
+	case inactive:
+		return "inactive"
+	case logInactive:
+		return "logInactive"
+	}
+	return "unknown"
+}
+
+const (
+	counterPrefix            = "shedder.admission_control."
+	admissionControlSpanName = "admission_control"
+	admissionControlKey      = "shedder:admission_control"
+	admissionControlValue    = "reject"
+	minWindowSize            = 1
+	maxWindowSize            = 100
+)
+
+type Options struct {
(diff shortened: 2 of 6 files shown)


docs/reference/filters.md
```diff
@@ -1848,6 +1848,79 @@ Path("/cheap")     -> clusterLeakyBucketRatelimit("user-${request.cookie.Authori
...
+should be unique by filter instance
+
+Mode has 3 different possible values and defaults to "inactive":
+
+* "active" will reject traffic
+* "inactive" will never reject traffic (default)
+* "log" will not reject traffic, but log to debug filter settings
+
+D the time duration of a single slot for required counters in our
+circular buffer of window size and default to 1s.
```
AlexanderYastrebov, line 1895: For leaky bucket I decided not to use any defaults to make users understand what they are doing and because there is no one-size-fits all config for load schedding/ratelimit IMO.

docs/reference/filters.md
```diff
@@ -1848,6 +1848,79 @@ Path("/cheap")     -> clusterLeakyBucketRatelimit("user-${request.cookie.Authori
...
+0.95, which means an error rate of lower than 5% will not trigger
+rejects.
+
+Max reject probability sets the upper bound of reject probability. It
+is within (0,1] and defaults to 0.95, which means if backend errors
+with 100% it will only reject up to 95%.
+
+aggression is used to dictate the rejection probability. The
+calculation is done by $p = p^{1 \over aggression}$
+The aggression value is within $(0,\infty]$:
```
AlexanderYastrebov, line 1915: Aggression is similar to `fadeIn`'s `exponent`: `exponent = 1/aggression`.

filters/shedder/admission.go
```diff
@@ -0,0 +1,366 @@
...
+	return 0, filters.ErrInvalidFilterParameters
+}
+
+func getDurationArg(a interface{}) (time.Duration, error) {
+	if s, ok := a.(string); ok {
+		return time.ParseDuration(s)
+	}
+
+	i, err := getIntArg(a)
+	return time.Duration(i) * time.Millisecond, err
```
AlexanderYastrebov, line 37: I suggest we support duration string only, see https://github.com/zalando/skipper/pull/1814#issuecomment-895053332

filters/shedder/admission.go
```diff
@@ -0,0 +1,366 @@
...
+		return log, nil
+	}
+
+	return 0, filters.ErrInvalidFilterParameters
+}
+
+type mode int
+
+const (
+	inactive mode = iota
```
AlexanderYastrebov, line 68: Maybe `iota + 1` is better because then 0 (that may pop up by mistake somewhere as default int) is not a invalid mode

filters/shedder/admission.go
```diff
@@ -0,0 +1,366 @@
...
+	/* #nosec */
+	r := rand.Float64() // [0,1)
+
+	if ac.mode == log {
+		logrus.Infof("p: %0.2f, r: %0.2f", p, r)
+	}
+
+	if p > r {
+		ac.metrics.IncCounter(counterPrefix + ac.metricSuffix)
+		ext.Error.Set(span, true)
```
AlexanderYastrebov, line 331: I think this should be an error only if we serve StatusServiceUnavailable.

AlexanderYastrebov: Maybe we also should tag the span with the mode if it is not `active`

filters/shedder/admission.go
```diff
@@ -0,0 +1,366 @@
...
+func (ac *admissionControl) Request(ctx filters.FilterContext) {
+	span := ac.startSpan(ctx.Request().Context())
+	defer span.Finish()
+
+	p := ac.pReject() // [0, ac.maxRejectProbability] and -1 to disable
+	/* #nosec */
+	r := rand.Float64() // [0,1)
+
+	if ac.mode == log {
+		logrus.Infof("p: %0.2f, r: %0.2f", p, r)
```
AlexanderYastrebov, line 326: I think we alias logrus as `log` everywhere else.
Lets also add a `admissionControl: ` prefix to the log message

filters/shedder/admission.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+
+	if p > r {
+		ac.metrics.IncCounter(counterPrefix + ac.metricSuffix)
+		ext.Error.Set(span, true)
+
+		ctx.StateBag()["Admission-Control"] = "reject"
```
AlexanderYastrebov, line 333: minor: Maybe we should use more technically-looking key that does not look like a header, e.g. `admission:control` like e.g. these https://github.com/zalando/skipper/blob/c67b6443e6c20452cd6a816abce7b936cbf8bf1d/filters/filters.go#L12-L30

filters/shedder/admission.go
```diff
@@ -0,0 +1,366 @@
...
+		successThreshold:     threshold,
+		maxRejectProbability: maxRejectProbability,
+		aggression:           aggression,
+
+		totals:         make([]int64, windowSize),
+		success:        make([]int64, windowSize),
+		counter:        atomic.NewInt64(0),
+		successCounter: atomic.NewInt64(0),
+	}
+	go ac.tickWindows()
```
AlexanderYastrebov, line 259: Hm, this will leak goroutines.

filters/shedder/admission.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+}
+
+func (ac *admissionControl) count() (float64, float64) {
+	ac.mu.Lock()
+	defer ac.mu.Unlock()
+	return float64(sum(ac.totals)), float64(sum(ac.success))
+}
+
+func sum(a []int64) int64 {
```
AlexanderYastrebov, line 285: It could return `float64` right away

docs/reference/filters.md
```diff
@@ -1848,6 +1848,79 @@ Path("/cheap")     -> clusterLeakyBucketRatelimit("user-${request.cookie.Authori
...
+
+D the time duration of a single slot for required counters in our
+circular buffer of window size and default to 1s.
+
+Window size is the size of the circular buffer. It is used to snapshot
+counters to calculate total requests and number of success. It is
+within $[1, 100]$ and defaults to 10.
+
+Min requests is the minimum requests per $windowSize * d$ that have to pass this filter
+otherwise it will not reject traffic.
```
AlexanderYastrebov, line 1902: In the [envoy](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/admission_control_filter#overview) it is requests per second (rps), not per `$windowSize * d$`

filters/shedder/admission.go
```diff
@@ -0,0 +1,366 @@
...
+func (ac *admissionControl) pReject() float64 {
+	var rejectP float64
+
+	total, success := ac.count()
+	factor := float64(ac.d) / float64(time.Second)
+	if total < float64(ac.minRequests)*factor {
+		return -1
+	}
+
+	s := success / ac.successThreshold
```
AlexanderYastrebov, line 303: > Success threshold sets the lowest request success rate at which the
filter will not reject requests. It is within $(0,1]$ and defaults to
0.95, which means an error rate of lower than 5% will not trigger
rejects.

Maybe it would be cleaner if we calculate
```go
successRate := success / total
if successRate > ac.successThreshold {
    return -1
}
```

filters/shedder/admission.go
```diff
@@ -0,0 +1,366 @@
...
+	if ac.mode == log {
+		logrus.Infof("rejectP = (%0.2f - %0.2f) / (%0.2f + 1)  --- success: %0.2f and threshold: %0.2f", total, s, total, success, ac.successThreshold)
+	}
+	if total < s {
+		return -1
+	}
+	rejectP = (total - s) / (total + 1)
```
AlexanderYastrebov, line 310: Why do we subtract `s == success / ac.successThreshold` and not `success`?

AlexanderYastrebov: Ok, this is by the formula

filters/shedder/admission.go
```diff
@@ -0,0 +1,366 @@
...
+	return math.Max(rejectP, 0.0)
+}
+
+func (ac *admissionControl) Request(ctx filters.FilterContext) {
+	span := ac.startSpan(ctx.Request().Context())
+	defer span.Finish()
+
+	p := ac.pReject() // [0, ac.maxRejectProbability] and -1 to disable
+	/* #nosec */
+	r := rand.Float64() // [0,1)
```
AlexanderYastrebov, line 323: Lets change `pReject()` into `func (ac *admissionControl) shouldReject() bool` to keep math and HTTP separate?

filters/shedder/admission_test.go
```diff
@@ -0,0 +1,199 @@
...
+			}
+
+			failBackend := float64(0)
+			fail := float64(0)
+			ok := float64(0)
+			var N float64
+			// iterations to make sure we have enough traffic
+			for j := 0; j < 10; j++ {
+				for i := 0; i < 1+(2*ti.windowsize); i++ {
+					until := time.Now().Add(ti.d)
```
AlexanderYastrebov, line 159: Maybe get rid of the loops and have
```go
// run 20x of history size
until := time.After(20 * ti.d * ti.windowsize)

work:
for {
        select {
	case <-until:
		break work
	default:
	}
...
```

filters/shedder/admission_test.go
```diff
@@ -0,0 +1,199 @@
...
+	for _, ti := range []struct {
+		msg                        string
+		mode                       string
+		d                          time.Duration
+		windowsize                 int
+		minRequests                int
+		successThreshold           float64
+		maxrejectprobability       float64
+		aggression                 float64
+		N                          int     // iterations
```
AlexanderYastrebov, line 27: I think it is not used

filters/shedder/admission_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+			fr := make(filters.Registry)
+			fr.Register(spec)
+			r := &eskip.Route{Filters: []*eskip.Filter{{Name: spec.Name(), Args: args}}, Backend: backend.URL}
+
+			proxy := proxytest.New(fr, r)
+			reqURL, err := url.Parse(proxy.URL)
+			if err != nil {
+				t.Fatalf("Failed to parse url %s: %v", proxy.URL, err)
+			}
```
AlexanderYastrebov, line 143: Maybe this boilerplate could be moved into separate helper `func createAdmissionControlTestProxy(t, args, backend.URL)`?

docs/reference/filters.md
szuecs: I want to adjust defaults when we know there might be known good configuration for the bulk of services at best.
I don't think it's reasonable that the users that have no idea about the thing can judge if windowSize should be 5 or 50 and I want to make default a good choice and I tried to sort it such that simple things are in front and complicated things are last.
So it was a careful choice, that I would like to keep.

docs/reference/filters.md
szuecs: What is the suggestion?

filters/shedder/admission.go
szuecs: yeah , just copied it, will change it

filters/shedder/admission.go
szuecs: This was a chosen default to make it inactive by default

filters/shedder/admission.go
szuecs: That's to enable visibility before we make it "active", but I will add 
```
span.SetTag("mode", ac.mode.String())
```
such that it's observable that we don't block it, otherwise we can not understand with "inactive" mode in opentracing the behavior.

filters/shedder/admission.go
szuecs: fixed

filters/shedder/admission.go
szuecs: a sum() that gets int64 slice and returns float, hmm..
 no :)

docs/reference/filters.md
szuecs: Yes we can do better than envoy

filters/shedder/admission.go
```diff
@@ -0,0 +1,366 @@
...
+
+// calculates P_{reject} see https://opensource.zalando.com/skipper/reference/filters/#admissioncontrol
+func (ac *admissionControl) pReject() float64 {
+	var rejectP float64
+
+	total, success := ac.count()
+	factor := float64(ac.d) / float64(time.Second)
+	if total < float64(ac.minRequests)*factor {
+		return -1
+	}
```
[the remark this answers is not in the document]
szuecs: I had this as minRps before but why do we need to wait a second if we can observe it faster if we would like?

filters/shedder/admission.go
szuecs: s is reused in the calculation and check the envoy link to doc https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/admission_control_filter#overview

filters/shedder/admission.go
szuecs: I can add another func but mixing the math with rand handling seems not to make sense to me

szuecs: fixed

filters/shedder/admission_test.go
szuecs: I like the nested loops because you can see that the inner makes sense and the other is "have more requests than 100"

filters/shedder/admission_test.go
```diff
@@ -0,0 +1,199 @@
...
+			req, err := http.NewRequest("GET", reqURL.String(), nil)
+			if err != nil {
+				t.Error(err)
+				return
+			}
+
+			failBackend := float64(0)
+			fail := float64(0)
+			ok := float64(0)
+			var N float64
```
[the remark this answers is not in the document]
szuecs: fixed

filters/shedder/admission_test.go
szuecs: true, but maybe I should change the outer for loop to N<ti.N

I think I will do it.

filters/shedder/admission_test.go
```diff
@@ -0,0 +1,199 @@
...
+				}
+			}
+			t.Logf("ok: %0.2f, fail: %0.2f, failBackend: %0.2f", ok, fail, failBackend)
+
+			epsilon := 0.05 * N // maybe 5% instead of 0.1%
+			expectedFails := (N - failBackend) * ti.pExpectedAdmissionShedding
+
+			if expectedFails-epsilon > fail || fail > expectedFails+epsilon {
+				t.Errorf("Failed to get expected fails should be in: %0.2f < %0.2f < %0.2f", expectedFails-epsilon, fail, expectedFails+epsilon)
+			}
```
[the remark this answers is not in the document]
szuecs: I think it's more easy, than you suggestion and if you read it like a math equation it's simpler to follow:
```
A-e > fail > A+e
```

filters/shedder/admission_test.go
szuecs: IMO: as long as it fits on to one screen, you want to understand the test anyways, so read it within one page.

filters/shedder/admission.go
szuecs: Hmm, I thought you would over look this :D
fixed

docs/reference/filters.md
szuecs: I changed it to exponent, I like this also more!

szuecs: Please check if the change makes sense to you.

filters/shedder/admission.go
szuecs: I think I fixed it, so resolve. Discussion can be done offline

docs/reference/filters.md
szuecs: Agreed on first without defaults and later if we observe known good values, we can add defaults

filters/shedder/admission.go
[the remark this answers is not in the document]
szuecs: We agreed on discussion that it increases usability if it's independent of filter details so using RPS is better.

Because `time.Second / (ac.d * windowSize)` are known in CreateFilter and static for the filter life time, I pre calculate the value into `ac.averageRpsFactor`.

docs/reference/filters.md
```diff
@@ -1848,6 +1848,80 @@ Path("/cheap")     -> clusterLeakyBucketRatelimit("user-${request.cookie.Authori
...
 ```
 
+## shedder
+
+The basic idea of load shedding is to reduce errors by early stopping
+partial of the ingress requests that create too much load and serve
+the maximum throughput the system can do.
+
+There is a great talk by [Acacio Cruz from
+Google](https://www.youtube.com/watch?v=XNEIkivvaV4&feature=youtu.be)
```
bocytko, line 1858:
```suggestion
Google](https://www.youtube.com/watch?v=XNEIkivvaV4)
```

docs/reference/filters.md
```diff
@@ -1848,6 +1848,80 @@ Path("/cheap")     -> clusterLeakyBucketRatelimit("user-${request.cookie.Authori
 Path("/expensive") -> clusterLeakyBucketRatelimit("user-${request.cookie.Authorization}", 1, "1s", 5, 2) -> ...
 ```
 
+## shedder
+
+The basic idea of load shedding is to reduce errors by early stopping
+partial of the ingress requests that create too much load and serve
+the maximum throughput the system can do.
```
bocytko, line 1855:
```suggestion
The basic idea of load shedding is to reduce errors by early stopping
some of the ingress requests that create too much load and serving
the maximum throughput the system can process at a point in time.
```

docs/reference/filters.md
```diff
@@ -1848,6 +1848,80 @@ Path("/cheap")     -> clusterLeakyBucketRatelimit("user-${request.cookie.Authori
...
+    admissionControl(metricSuffix, mode, d, windowSize, minRPS, successThreshold, maxRejectProbability, exponent)
+    admissionControl("myapp", "active", "1s", 5, 10, 0.95, 0.9, 0.5)
+
+Parameters:
+
+* metric suffix (string)
+* mode (enum)
+* d (time.Duration)
+* window size (int)
+* min requests (int)
```
bocytko, line 1881: minRPS carries a slightly different meaning (per second)

docs/reference/filters.md
szuecs: thanks fixed

docs/reference/filters.md
szuecs: yes we discussed it elsewhere and I changed this behavior.

docs/reference/filters.md
szuecs: we discussed and the purpose of having rps is for the user as input to us.

filters/shedder/admission.go
szuecs: This is now fixed via PostProcessor handling

szuecs: This was fixed

filters/shedder/admission_test.go
```diff
@@ -0,0 +1,463 @@
...
+			result := <-ch
+			if result == nil {
+				t.Error("Failed to cleanup filters correctly, found more than one admissionControl filter in one route")
+			}
+		})
+	}
+}
+
+type validationPostProcessorNumberOfFilters struct {
+	ch chan []*routing.Route
```
AlexanderYastrebov, line 432: minor: Maybe it could receive `*testing.T` and call `t.Error` if admission filter counter is more than one

AlexanderYastrebov:
:+1:

filters/shedder/admission_test.go
szuecs: The problem is that t is not Thread safe. At least in my tests I got warnings or a panic, don't remember anymore.

szuecs:
👍

PR #2032, opened 2022-06-28
Log invalid query

MustafaSaber:


net/query.go
```diff
@@ -11,6 +13,7 @@
 
 func (q *ValidateQueryHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if _, err := url.ParseQuery(r.URL.RawQuery); err != nil {
+		log.Errorf("Invalid query: %s", r.URL.RawQuery)
```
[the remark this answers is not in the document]
szuecs: Don't log data, just log that it happened.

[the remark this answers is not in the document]
AlexanderYastrebov: Maybe it should log the error

config/config.go
```diff
@@ -142,8 +142,10 @@ type Config struct {
 	NormalizeHost bool          `yaml:"normalize-host"`
 	HostPatch     net.HostPatch `yaml:"-"`
 
-	ValidateQuery bool      `yaml:"validate-query"`
-	RefusePayload multiFlag `yaml:"refuse-payload"`
+	ValidateQuery         bool      `yaml:"validate-query"`
+	ValidateQueryLog      bool      `yaml:"validate-query-log"`
+	ValidateQueryLogLevel string    `yaml:"validate-query-log-level"`
```
szuecs, line 147: I think we should drop the idea if having a log level here. @AlexanderYastrebov I think we can't do this because the logging.Logger should be used and we can't have 2 level settings here.

AlexanderYastrebov: Why `logging.Logger` and not `github.com/sirupsen/logrus`?

net/query.go
```diff
@@ -3,14 +3,25 @@ package net
...
+	LogLevel    string
+	Handler     http.Handler
 }
 
 func (q *ValidateQueryHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if _, err := url.ParseQuery(r.URL.RawQuery); err != nil {
+
+		if q.LogsEnabled {
+			logLevel, _ := log.ParseLevel(q.LogLevel)
+			log.SetLevel(logLevel)
```
AlexanderYastrebov, line 21: This chnages global logger level which is wrong.
It should be something like
```go
switch q.level {
case "DEBUG":
  log.Debugf(...)
case "INFO":
  log.Infof(...)
...
}
```
or even
```go
log.StandardLogger().Logf(q.level, "invalid query: %v", err)
```

config/config.go
szuecs: Yeah but not as suggested before

mikkeloscar:
:+1:

PR #2034, opened 2022-06-30
Use arm64 emulator that CDP provides

linki:
CDP now works out of the box: https://github.bus.zalan.do/automata/cdp-ami/pull/28


linki:
If it fails on the `armv7` (32-bit) build, then just roll back or ask automata for armv7 support. I'm pretty sure it will work but just in case (It's only executed on master).

linki:
👍

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
Lets try

PR #2044, opened 2022-07-27
Feature/scheduler filter fifo

szuecs:
missing:
- [ ] metrics
- [ ] tests
- [ ] docs


filters/scheduler/fifo.go
```diff
@@ -0,0 +1,161 @@
...
+// queue for handling requests to limit concurrency of a route. The first
+// parameter is maxConcurrency the second maxQueueSize and the third
+// timeout.
+func (s *fifoSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	var f fifoFilter
+
+	// set defaults
+	f.config.MaxConcurrency = defaultMaxConcurreny
+	f.config.MaxQueueSize = defaultMaxQueueSize
+	f.config.Timeout = defaultTimeout
```
AlexanderYastrebov, line 41: Lets make all args mandatory as in the docs. There are only three and it forces user to think about the config.

scheduler/scheduler.go
```diff
@@ -69,17 +82,113 @@ type Queue struct {
...
+	}()
+
+	// limit concurrency
+	c, done := context.WithTimeout(ctx, timeout)
+	if err := fq.sem.Acquire(c, 1); err != nil {
+		cerr := c.Err()
+		if cerr == nil {
+			// should never happen
+			return done, err
+		}
```
AlexanderYastrebov, line 162: This seems redundant as Acquire returns context Err()

filters/scheduler/fifo.go
szuecs: I just changed the code analog to lifo(), but I can make it mandatory.

scheduler/scheduler.go
```diff
@@ -140,6 +267,49 @@ type GroupedLIFOFilter interface {
...
+// request needs to be rejected, an error will be returned.
+func (fq *FifoQueue) Wait(ctx context.Context) (func(), error) {
+	f, err := fq.queue.Wait(ctx)
+	if fq.metrics != nil {
+		switch err {
+		case ErrQueueFull:
+			fq.metrics.IncCounter(fq.errorFullMetricsKey)
+		case ErrQueueTimeout:
+			fq.metrics.IncCounter(fq.errorTimeoutMetricsKey)
+		case ErrClientCanceled:
```
AlexanderYastrebov, line 282: Is `fallthrough` missing? Or maybe add a comment if not...

scheduler/scheduler.go
```diff
@@ -69,17 +82,113 @@ type Queue struct {
...
+	return QueueStatus{
+		ActiveRequests: int(active),
+		QueuedRequests: int(queued),
+		Closed:         fq.closed,
+	}
+}
+
+// Reconfigure does nothing, because it creates a data race that would
+// require more mutex than we want to have
+func (fq *fifoQueue) Reconfigure(c Config) {
```
AlexanderYastrebov, line 129: fifoQueue is a non-exported type and probably should have non-exported methods

scheduler/scheduler.go
```diff
@@ -69,17 +82,113 @@ type Queue struct {
...
+}
+
+func (fq *fifoQueue) Status() QueueStatus {
+	fq.mu.RLock()
+	defer fq.mu.RUnlock()
+	maxConcurrency := fq.maxConcurrency
+	all := fq.counter.Load()
+	queued := float64(0)
+	if all > maxConcurrency {
+		queued = math.Max(float64(all-maxConcurrency), 0)
```
AlexanderYastrebov, line 117: Why do we need math.Max if we know `all > maxConcurrency` ?

scheduler/scheduler.go
```diff
@@ -69,17 +82,113 @@ type Queue struct {
...
+	maxConcurrency uint64
+	closed         bool
+}
+
+func (fq *fifoQueue) Status() QueueStatus {
+	fq.mu.RLock()
+	defer fq.mu.RUnlock()
+	maxConcurrency := fq.maxConcurrency
+	all := fq.counter.Load()
+	queued := float64(0)
```
AlexanderYastrebov, line 115: Why float64 and not int64?

scheduler/scheduler.go
```diff
@@ -69,17 +82,113 @@ type Queue struct {
...
+	if all > maxConcurrency {
+		queued = math.Max(float64(all-maxConcurrency), 0)
+	}
+	active := math.Min(float64(all), float64(maxConcurrency))
```
AlexanderYastrebov, line 119:
```go
if all > maxConcurrency {
  queued = all - maxConcurrency
  active = maxConcurrency
} else {
  queued = 0
  active = all
}
```

scheduler/scheduler.go
```diff
@@ -69,17 +82,113 @@ type Queue struct {
...
+	sem            *semaphore.Weighted
+	timeout        time.Duration
+	maxQueueSize   uint64
+	maxConcurrency uint64
+	closed         bool
+}
+
+func (fq *fifoQueue) Status() QueueStatus {
+	fq.mu.RLock()
+	defer fq.mu.RUnlock()
```
AlexanderYastrebov, line 112: I think this mutex can be released as soon as we snapshot fq.maxConcurrency and fq.closed

scheduler/scheduler.go
```diff
@@ -13,13 +16,23 @@ import (
 	"github.com/zalando/skipper/filters"
 	"github.com/zalando/skipper/metrics"
 	"github.com/zalando/skipper/routing"
+	"go.uber.org/atomic"
```
AlexanderYastrebov, line 19: Why not https://pkg.go.dev/sync/atomic ?

scheduler/scheduler.go
```diff
@@ -69,17 +82,113 @@ type Queue struct {
...
+	// queue full?
+	if fq.counter.Load() > maxConcurrency+maxQueueSize {
+		return func() {}, ErrQueueFull
+	}
+
+	// handle queue
+	fq.counter.Inc()
+	defer func() {
+		fq.counter.Dec()
+	}()
```
AlexanderYastrebov, line 153: I think here we can do
```go
n := fq.counter.Inc()
defer fq.counter.Dec()

if n > maxConcurrency+maxQueueSize {
  return nil, ErrQueueFull
}
```
without using Load()

scheduler/scheduler.go
szuecs: I saw negative values

scheduler/scheduler.go
szuecs: because of math.Max

scheduler/scheduler.go
szuecs: Because we have to wait for Go 1.19 https://github.com/zalando/skipper/issues/2064

scheduler/scheduler.go
szuecs: I will test this idea

szuecs: Seems to work, thanks changed

scheduler/scheduler.go
szuecs: true thanks, fixed

scheduler/scheduler.go
szuecs: thanks true, fixed

scheduler/scheduler.go
szuecs: changed because rest of your comments work

scheduler/scheduler.go
szuecs: true fixed, was following the lifo which is using and external dependency

scheduler/scheduler.go
szuecs: no this is visible in 499, not really a scheduler filter issue. I add a comment

scheduler/scheduler.go
szuecs: true fixed

filters/scheduler/fifo.go
```diff
@@ -0,0 +1,161 @@
...
+	ctx.StateBag()[fifoKey] = pending[:last]
+
+	if ok {
+		f.queue.Release()
+	}
+}
+
+const (
+	fifoKey string = "fifo"
+)
```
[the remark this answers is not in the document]
szuecs: done

filters/scheduler/fifo.go
```diff
@@ -0,0 +1,161 @@
...
+	last := len(pending) - 1
+	if last < 0 {
+		return
+	}
+	pending[last]()
+	ctx.StateBag()[fifoKey] = pending[:last]
+
+	if ok {
+		f.queue.Release()
+	}
```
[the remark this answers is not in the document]
szuecs: True much better idea!

filters/scheduler/fifo.go
```diff
@@ -0,0 +1,161 @@
...
+// if the bounded queue returns an error. Status code by Error:
+//
+// - 503 if queue full
+// - 502 if queue timeout
+// - 500 if error unknown
+func (f *fifoFilter) Request(ctx filters.FilterContext) {
+	c := ctx.Request().Context()
+	done, err := f.queue.Wait(c)
+	if err != nil {
+		done()
```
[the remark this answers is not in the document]
szuecs: for example if in wait(), context is created but Acquire fails with context.Canceled for example:
```
	c, done := context.WithTimeout(ctx, timeout)
	if err := fq.sem.Acquire(c, 1); err != nil {
		switch err {
		case context.DeadlineExceeded:
			return done, ErrQueueTimeout
		case context.Canceled:
			return done, ErrClientCanceled
		default:
			// does not exist yet in Go stdlib as of Go1.18.4
			return done, err
		}
	}
```

proxy/proxy.go
```diff
@@ -1055,6 +1055,12 @@ func (p *Proxy) do(ctx *context) error {
 	}
 
 	defer func() {
+		// TODO(sszuecs) do we need this? When does it happen?
+		// pendingFIFO, _ := ctx.StateBag()[scheduler.FIFOKey].([]func())
+		// for _, done := range pendingFIFO {
+		// 	done()
+		// }
+
```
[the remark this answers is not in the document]
szuecs: tested, that connect errors would block until fifo timeout if we don't have this, which would fail fast to 502.

filters/scheduler/fifo.go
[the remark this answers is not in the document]
AlexanderYastrebov: The usual convention is either valid value or error. The client than uses return value only when error is nil.

scheduler/scheduler.go
```diff
@@ -69,17 +81,114 @@ type Queue struct {
...
+		case context.Canceled:
+			return done, ErrClientCanceled
+		default:
+			// does not exist yet in Go stdlib as of Go1.18.4
+			return done, err
+		}
+	}
+	return func() {
+		fq.sem.Release(1)
+		done()
```
AlexanderYastrebov, line 172: Do we need to postpone context done() call? Why not call it just before return?

szuecs: fixed

filters/scheduler/fifo.go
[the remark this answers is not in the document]
szuecs: done

scheduler/scheduler.go
```diff
@@ -69,17 +81,114 @@ type Queue struct {
...
+	defer done()
+	if err := fq.sem.Acquire(c, 1); err != nil {
+		switch err {
+		case context.DeadlineExceeded:
+			return func() {}, ErrQueueTimeout
+		case context.Canceled:
+			return func() {}, ErrClientCanceled
+		default:
+			// does not exist yet in Go stdlib as of Go1.18.4
+			return func() {}, err
```
AlexanderYastrebov, line 168: I think it should `return nil, ErrQueueTimeout` and caller should not call done() when error is not nil

filters/scheduler/fifo.go
```diff
@@ -0,0 +1,153 @@
...
```
AlexanderYastrebov, line 100: I think Wait should not return valid `done` in case of error

AlexanderYastrebov:
:+1:

szuecs:
:+1:

Issue #2045, opened 2022-07-27
Use k8s server-side filtering for Ingresses


azarakovskiy:
**Is your feature request related to a problem? Please describe.**

_Check additional context to understand better :)_

Before we used a custom data client that doesn't get updated in runtime, and when on our dev environment Skipper app was idle under little traffic, and you could check in the metrics that it doesn't do much.

We are working to replace it with a K8S client that would use Ingress templates.

We enabled a K8S client, and the CPU usage went up a lot. It started to constantly do something even without any traffic. By looking in code, we concluded it's `func (c *clusterClient) fetchClusterState()` being called often. It loads all Ingresses/Services/Endpoints/Secrets, parses all of them to a data struct, and filters some of them out.

This sounds to me like a very CPU-intensive thing to do and it explains what Skipper is busy with -- loading it all again and again.

_Are we doing something wrong?_

**Describe the solution you would like**

Ingresses are filtered by `IngressClass` after all of them are loaded. Also, in K8S API almost every read endpoint supports `labelSelector`, including `.../ingresses`.

Since the Ingress class can be set to ingress templates, we can also set any labels, and then use server-side filtering. Is there any special reason why this is not done is Skipper? Why is everything loaded?

Services/Endpoints/Secrets would get the same filtering options, in that case, I guess. Details of how exactly it should work are not yet defined

**Describe alternatives you've considered (optional)**

**Additional context (optional)**

We are running a Skipper deployment in a k8s cluster A. It is a custom application that uses Skipper as a go dependency, and we have built some logic on top of it. It routes traffic to a couple of other clusters B & C, and also to A. 

Currently, the definition of all routes that we have in a system lives in a JSON file which is generated externally, committed to the repository from time to time, read at a start of a pod, and fed to a custom data client. We want to replace it with Ingress templates, and Skipper to consume them.

So we started to look at the K8S data client that is already in Skipper.

**Would you like to work on it?**
Yes

szuecs:
@azarakovskiy we poll all resources even if we filter ingressClass, because of simplicity. Keep also in mind the v1 ingress with v1 ingressClass is different than the annotation from the past.
Our current solution is to run a few `routesrv` and query these from skipper-ingress to offload traffic from kube-apiserver.
https://github.com/zalando/skipper/tree/master/cmd/routesrv

azarakovskiy:
@szuecs thanks, we had the similar option in mind too but decided to first verify if we could optimize the querying somehow.

[...]

PR #2048, opened 2022-08-02
Support K8S labelSelector filters for API requests

azarakovskiy:
This PR allows you to configure a kubernetes client to only load specific resources from the backend instead of loading everything and filtering on Skipper side.

This should reduce the load on Skipper and K8S if most of your resources shouldn't actually be loaded to the Skipper data client.

```
skipper.Options {
	KubernetesIngressLabelSelectors: map[string]string{"skipper-enabled": "true"},

	KubernetesEndpointsLabelSelectors: map[string]string{"skipper-enabled": "true"},

        // this one differs from the rest
	KubernetesSecretsLabelSelectors: map[string]string{"skipper-enabled": "true", "some-special": "value"},

        // this one is explicitly empty, so won't require labels and will be loaded
	KubernetesRouteGroupsLabelSelectors: map[string]string{},

        // this one is not defined, so it won't require labels and will be loaded
        // KubernetesServicesLabelSelectors: nil,
}
```
This will limit the list of resources that Skipper will try to load from K8S API.

Implementation for #2045.


szuecs:
Please add tests also in https://github.com/zalando/skipper/tree/master/dataclients/kubernetes/testdata/ingressV1/ingress-data .
You can override kube dataclient config in the `<mytest>.kube` file and if you need to enhance it you need to change `kubernetestest` package, there is an "apiserver" implementation that serves the json data.
The simple way to add a test is to create a `mytest.yaml` for what you serve in kubernetes and fetch with the dataclient and a `mytest.eskip` that is the corresponding route creation.

azarakovskiy:
> Please add tests also in https://github.com/zalando/skipper/tree/master/dataclients/kubernetes/testdata/ingressV1/ingress-data . You can override kube dataclient config in the `<mytest>.kube` file and if you need to enhance it you need to change `kubernetestest` package, there is an "apiserver" implementation that serves the json data. The simple way to add a test is to create a `mytest.yaml` for what you serve in kubernetes and fetch with the dataclient and a `mytest.eskip` that is the corresponding route creation.

Done. I had to hack it around a bit to filter out the byte slices but I tried to make it as readable as possible.
I figured out I don't like to refactor these tests to support this niche functionality since most of the tests won't need the added complexity. That was the reasoning behind hacking it into the existing way how it works and not refactoring everything. :)

dataclients/kubernetes/clusterclient.go
```diff
@@ -177,6 +192,33 @@ func newClusterClient(o Options, apiURL, ingCls, rgCls string, quit <-chan struc
...
+	var strs []string
+	for k, v := range selectors {
+		if v == "" {
+			strs = append(strs, k)
+		} else {
+			strs = append(strs, fmt.Sprintf(labelSelectorFmt, k, v))
+		}
+	}
+
+	return fmt.Sprintf(labelSelectorQueryFmt, url.QueryEscape(strings.Join(strs, ",")))
```
wimspaargaren, line 219: Since we already have `map[string]string` wouldn't it make more sense to just make an additional method something like:
`getJSONWithQueryParameter` and pass along the corresponding query params?

This would also make it as such that we don't need to gather everything here and query escape the values.

azarakovskiy: This would essentially mean that it would be built to a string with each request even though it never changes with the given setup, so can be just done once and appended to the URI.

dataclients/kubernetes/kube.go
```diff
@@ -133,6 +133,37 @@ type Options struct {
 	// skipper will load it. The default value for the RouteGroup class is 'skipper'.
 	RouteGroupClass string
 
+	// IngressLabelSelectors is a map of K8S labels to their values that must be present on a resource to be loaded
```
szuecs, line 136: Please no acronym (k8s) of kubernetes it should be consistent everywhere.

dataclients/kubernetes/kubernetestest/api.go
```diff
@@ -171,6 +172,73 @@ func (a *api) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+// Filters all resources that are already set in k8s namespace using the given selectors map.
+// All resources are initially set to `namespace` as slices of bytes and for most tests it's not needed to make it any more complex.
+// This helper function deserializes resources, finds a metadata with labels in them and check if they have all
+// requested labels. If they do, they are returned.
+func filterBySelectors(resources []byte, selectors map[string]string) []byte {
+	if selectors == nil || len(selectors) == 0 {
+		return resources
+	}
+
+	// temp struct to easier access labels field in the map of maps and avoid casting many times
```
szuecs, line 200: Just drop the comment

dataclients/kubernetes/kubernetestest/api.go
```diff
@@ -171,6 +172,73 @@ func (a *api) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		} `json:"items"`
+	}{}
+
+	// every resource but top level is deserialized because we need access to the indexed array
+	allItems := struct {
+		Items []interface{} `json:"items"`
+	}{}
+
+	if json.Unmarshal(resources, &labels) != nil || json.Unmarshal(resources, &allItems) != nil {
+		// if something goes wrong, ignore it and return everything, this should fail the test anyway
```
szuecs, line 215: Drop this comment

dataclients/kubernetes/kubernetestest/api.go
```diff
@@ -171,6 +172,73 @@ func (a *api) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			label, ok := item.Metadata.Labels[k]
+			allMatch = allMatch && ok && label == v
+		}
+		// if all found, return it as part of a filtered result
+		if allMatch {
+			filteredItems = append(filteredItems, allItems.Items[idx])
+		}
+	}
+
+	// encode it back to the original form with `items`
```
szuecs, line 233: Drop this comment

dataclients/kubernetes/path_test.go
```diff
@@ -79,10 +79,7 @@ func TestPathMatchingModes(t *testing.T) {
 	defer api.Close()
 
 	setIngressWithPath := func(p string, annotations ...string) {
-		i := testIngress(
-			"namespace1", "ingress1", "service1", "", "", "", "", "", "", definitions.BackendPort{Value: 8080}, 1.0,
-			testRule("www.example.org", testPathRule(p, "service1", definitions.BackendPort{Value: 8080})),
-		)
+		i := testIngress("namespace1", "ingress1", "service1", "", "", "", "", "", "", map[string]string{}, definitions.BackendPort{Value: 8080}, 1.0, testRule("www.example.org", testPathRule(p, "service1", definitions.BackendPort{Value: 8080})))
```
szuecs, line 82: If it's a noop then please don't change the code

dataclients/kubernetes/path_test.go
```diff
@@ -344,15 +341,9 @@ func TestIngressSpecificMode(t *testing.T) {
 	api := newTestAPI(t, s, &definitions.IngressList{})
 	defer api.Close()
 
-	ingressWithDefault := testIngress(
-		"namespace1", "ingress1", "service1", "", "", "", "", "", "", definitions.BackendPort{Value: 8080}, 1.0,
-		testRule("www.example.org", testPathRule("^/foo", "service1", definitions.BackendPort{Value: 8080})),
-	)
+	ingressWithDefault := testIngress("namespace1", "ingress1", "service1", "", "", "", "", "", "", map[string]string{}, definitions.BackendPort{Value: 8080}, 1.0, testRule("www.example.org", testPathRule("^/foo", "service1", definitions.BackendPort{Value: 8080})))
```
szuecs, line 344: Same here

dataclients/kubernetes/path_test.go
```diff
@@ -344,15 +341,9 @@ func TestIngressSpecificMode(t *testing.T) {
...
 
-	ingressWithCustom := testIngress(
-		"namespace1", "ingress1", "service1", "", "", "", "", "", "", definitions.BackendPort{Value: 8080}, 1.0,
-		testRule("www.example.org", testPathRule("/bar", "service1", definitions.BackendPort{Value: 8080})),
-	)
+	ingressWithCustom := testIngress("namespace1", "ingress1", "service1", "", "", "", "", "", "", map[string]string{}, definitions.BackendPort{Value: 8080}, 1.0, testRule("www.example.org", testPathRule("/bar", "service1", definitions.BackendPort{Value: 8080})))
```
szuecs, line 346: Same here

dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-label-selector-filter-defaults-to-ingress.eskip
```diff
@@ -0,0 +1,4 @@
+kube_foo__qux__www_example_org_____bar:
+  Host("^(www[.]example[.]org[.]?(:[0-9]+)?)$") &&
+  PathRegexp("^/")
```
szuecs, line 3: Predicates in one line

dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-label-selector-filter-defaults-to-ingress.eskip
```diff
@@ -0,0 +1,4 @@
...
+  -> setPath("/foo") -> <roundRobin, "http://10.2.9.103:8080", "http://10.2.9.104:8080">;
```
szuecs, line 4: Filters are multi line (one filter each line)

skipper.go
```diff
@@ -194,6 +194,37 @@ type Options struct {
 	// annotation, or which an empty annotation, will be loaded too.
 	KubernetesRouteGroupClass string
 
+	// IngressLabelSelectors is a map of K8S labels to their values that must be present on a resource to be loaded
```
szuecs, line 197: Please use proper Go doc. So start the doc by the name (prefix Kubernetes) in all below, too

dataclients/kubernetes/path_test.go
azarakovskiy: oops, I guess I accidentally formatted it in IDE 🥺

dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-label-selector-filter-defaults-to-ingress.eskip
azarakovskiy: Done!

I think I copied it from [another file](https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-setpath-filter-annotation.eskip) and just followed the same style as found there.

I can change it everywhere in this PR if you ask me to, that's easy :)

skipper.go
azarakovskiy: Copy-paste error. Fixed.

azarakovskiy:
UPD: I changed the default behavior to not default all labels to Ingress labels because I myself forgot about it when I started to use it and wasted some time :) https://github.com/zalando/skipper/pull/2048/commits/18cc5d3027a169c7bd7c275367a0fa5b8fb0d1b6

azarakovskiy:
@szuecs hey :) Could you please take another look, I've processed all your comments? If you feel somebody else needs to review it as well, could you also assign them? Thanks!

dataclients/kubernetes/kubernetestest/api.go
```diff
@@ -171,6 +172,69 @@ func (a *api) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	w.Write(b)
 }
 
+// Parses an optional parameter with `label selectors` into a map if present or, if not present, returns nil.
+func parseSelectors(r *http.Request) map[string]string {
+	rawSelectors := r.URL.Query()["labelSelector"]
```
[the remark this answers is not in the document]
szuecs: From my side it's only this to fix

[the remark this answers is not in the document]
azarakovskiy: ✔️ Done

Also caught a bug around that code and added a test for that 🥳

szuecs:
@AlexanderYastrebov from my side lgtm so please check

szuecs:
:+1:

szuecs (approved):

azarakovskiy:
@AlexanderYastrebov ? 🥺

AlexanderYastrebov:
I like the idea in general.
Should we consider supporting [label selector syntax](https://github.com/kubernetes/apimachinery/blob/31bc292891b6a196ad41946423c71bbe39c05d3c/pkg/labels/selector.go#L846-L884) instead of simplistic name-value set?

szuecs:
@AlexanderYastrebov The current version is very simple and apimachinery looks very bloated, so I would rather stick with this one.

dataclients/kubernetes/kube.go
```diff
@@ -133,6 +133,37 @@ type Options struct {
...
+	// IngressLabelSelectors is a map of kubernetes labels to their values that must be present on a resource to be loaded
+	// by the client. A label and its value on an Ingress must be match exactly to be loaded by Skipper.
+	// If the value is irrelevant for a given configuration, it can be left empty. The default
+	// value is no labels required.
+	// Examples:
+	//  Config [] will load all Ingresses.
+	// 	Config ["skipper-enabled": ""] will load only Ingresses with a label "skipper-enabled", no matter the value.
+	// 	Config ["skipper-enabled": "true"] will load only Ingresses with a label "skipper-enabled: true"
+	// 	Config ["skipper-enabled": "", "foo": "bar"] will load only Ingresses with both labels while label "foo" must have a value "bar".
+	IngressLabelSelectors map[string]string
```
AlexanderYastrebov, line 145: Should this (and others) rather be `IngressLabels` as "selector" is an expression that puts requirements on the label values https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ ?...

azarakovskiy:
> @AlexanderYastrebov The current version is very simple and apimachinery looks very bloated, so I would rather stick with this one.

I actually initially started to implement all the logic to use the [apimachinery's labels](https://github.com/kubernetes/apimachinery/blob/31bc292891b6a196ad41946423c71bbe39c05d3c/pkg/labels/selector.go#L846-L884) but it got very complicated very fast, and since I didn't need all that, I dropped the idea and kept it simple.

dataclients/kubernetes/kube.go
azarakovskiy: In the end, it ends up in a request parameter called `labelSelector` because it's just a specific kind of label selector with the rest not yet implemented.

I don't feel strongly about the naming here. However, if somebody ever decides to implement more from apimachinery to support more of that DSL, `IngressLabels` shorter names will become a tech debt.

szuecs: I agree please don't make it more complicated as needed.

AlexanderYastrebov: > However, if somebody ever decides to implement more from apimachinery to support more of that DSL, IngressLabels shorter names will become a tech debt.

That is what I had in mind: to name proposed configs `FooLabels` and if we ever decide to use DSL - add a new config `FooSelector`.

Lets keep it like that.

AlexanderYastrebov:
:+1:

PR #2052, opened 2022-08-08
[Proposal] Introduce AnySource predicate

mikkeloscar:
This proposes the introduction of a new Predicate: `AnySource` which is in family with the `Source`, `SourceFromLast` and `ClientIP` predicates, but works correctly no matter what load balancer may be in front of skipper.

Imagine the following setup where skipper is used as ingress in a Kubernetes cluster running in AWS, and [kube-ingress-aws-controller](https://github.com/zalando-incubator/kube-ingress-aws-controller) is used to manage Cloud Load balancers in front of skipper.

image

In this case you have a situation where one path to skipper is via a Network Load Balancer (NLB) and another path is via an Application Load Balancer (ALB). In this setup there is no existing predicate that can be used to ensure that only traffic from certain source IPs are matching a given Ingress configuration/route.
That is, if you use the `SourceFromLast` predicate you can correctly match on all traffic going through the ALB, because we can be sure that the ALB sets the last item of the `X-Forwarded-For` (XFF) header to the client IP of the client connecting to the ALB. However, with `SourceFromLast` we can't correctly match on traffic coming through the NLB because the XFF header could just be set by the client to any value.
similarly for traffic coming through the NLB we could correctly match it via `ClientIP` because the remote Address of the request will be the client connecting to the NLB, but for traffic coming through the ALB the remote Address will be the internal address of the load balancer, and thus we can't know what the real remote address was (unless we again look at XFF header).

`AnySource` solves this problem by implementing a more advanced logic which takes the load balancer addresses into account so it can distinguish traffic coming through ALB vs. NLB and apply the correct check respectively using the following logic:

```
if source_addr in allowed range:
  return true
if source_addr in known LB range AND XFF addr in allowed range:
  return true
else:
  return false
```

Configuration of the Load balancer ranges are done via the flag `-source-lb-cidrs` and is thus up to the operator of skipper, while the user of the `AnySource` predicate only have to define the relevant ranges for which she wants to allow access for a given route.

#### note

Right now it assumes the last item of XFF is trusted which is specific to how AWS ALBs work, to be more flexible we may need to make this configurable via another flag?

### TODO

* [ ] Align on the name of the predicate
* [x] Add documentation
* [x] Add test cases for IPV6.

### Alternative solution

As an alternative to adding a new Predicate, we could also add flags that change the behavior of the `Source` predicate only if set. That is, one flag `-source-lb-cidrs` to set an optional range of Load Balancer CIDRs and a flag `-source-xff-direction=first|last` to set which direction the XFF is read from. That way the behavior of `Source` depends on the skipper configuration and it can therefore be a detail on operators have to care about and not the users. The downside is that it requires a coordination between operators and users to introduce which is not needed if a new predicate is added (only users need to migrate in that case) but it seems like a nicer interface for users if you don't have to understand the complexity of the skipper setup to understand which predicate to use.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -138,6 +138,9 @@ type Config struct {
 	ForwardedHeadersExcludeCIDRList *listFlag            `yaml:"forwarded-headers-exclude-cidrs"`
 	ForwardedHeadersExcludeCIDRs    net.IPNets           `yaml:"-"`
 
+	SourceLBCIDRList *listFlag  `yaml:"source-lb-cidrs"`
+	SourceLBCIDRs    net.IPNets `yaml:"-"`
+
 	// host patch:
 	NormalizeHost bool          `yaml:"normalize-host"`
 	HostPatch     net.HostPatch `yaml:"-"`
@@ -292,6 +295,7 @@ func NewConfig() *Config {
 	cfg.RoutesURLs = commaListFlag()
 	cfg.ForwardedHeadersList = commaListFlag()
 	cfg.ForwardedHeadersExcludeCIDRList = commaListFlag()
+	cfg.SourceLBCIDRList = commaListFlag()
 	cfg.CompressEncodings = commaListFlag("gzip", "deflate", "br")
 
 	flag.StringVar(&cfg.ConfigFile, "config-file", "", "if provided the flags will be loaded/overwritten by the values on the file (yaml)")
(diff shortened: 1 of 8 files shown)


szuecs:
As far as I know SourceFromLast already works like you define in the description, because we set the xff similar to alb before we do the routing.

mikkeloscar:
Thanks @szuecs I did not remember this was implemented like this! Great, then it's already working as intended!

PR #2053, opened 2022-08-08
Update build to Go 1.19

mikkeloscar:
Update build to Go 1.19

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           # https://www.npmjs.com/package/semver#caret-ranges-123-025-004
-          go-version: '~1.18'
+          go-version: '^1.19'
       - run: go version
       - run: make deps
       - run: make check-fmt
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -8,7 +8,7 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           # https://www.npmjs.com/package/semver#caret-ranges-123-025-004
-          go-version: '~1.18'
+          go-version: '^1.19'
       - run: go version
       - run: make deps
       - run: make check-race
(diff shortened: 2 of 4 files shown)


proxy/upgrade_test.go
```diff
@@ -288,8 +288,8 @@ func TestServeHTTP(t *testing.T) {
 				return
 			}
 			if ti.backendClosesConnection {
-				if pong != "HTTP/1.1 400 Bad Request\r\n" {
-					t.Errorf("wrong bad response <%s>", pong)
+				if pong != `HTTP/1.1 200 OK\r\n` {
```
mikkeloscar, line 291: not 100% sure yet, why this changed.

szuecs: Hmm this looks wrong

mikkeloscar: I did a bisect of Go (from go1.18.5 to go1.19) and found this commit to be the cause of it: https://github.com/golang/go/commit/770e0e584a98dfd5e8d0d00558085c339fda0ed7

filters/builtin/compress.go
```diff
@@ -99,7 +99,7 @@ func (e encodings) Swap(i, j int) { e[i], e[j] = e[j], e[i] }
 //
 // Example:
 //
-// 	* -> compress() -> "https://www.example.org"
+//   - -> compress() -> "https://www.example.org"
```
AlexanderYastrebov, line 102: These changes look wrong, probably because example eskip is not a proper code block so godoc confuses it with item list

AlexanderYastrebov:
Likely needs this changed as well https://github.com/zalando/skipper/blob/608ac46dd277c0e0b6a6f002a7a3a703da64dbc0/delivery.yaml#L21

AlexanderYastrebov:
I've created https://github.com/zalando/skipper/pull/2105 to address the test failure problem

proxy/upgrade_test.go
mikkeloscar: Resolved in https://github.com/zalando/skipper/pull/2105

AlexanderYastrebov:
:+1:

mikkeloscar:
:+1:

PR #2054, opened 2022-08-09
Add custom weight for custom predicates

xavivars:
### tl;dr; The change itself

This PR adds the ability to provide extra weight for custom predicates (there were some conversations about it in https://github.com/zalando/skipper/pull/1338, even if no consensus).

Some details:
- It's an opt-in feature: by default, predicates do not have any custom weight associated with it (on top of the weight they already have "[because they exist](https://github.com/ruudk/skipper/blob/master/routing/matcher.go#L58)")
- It's backwards compatible: no effect to all existing predicates

I think the implementation is pretty straight forward, it doesn't impact the core types (Predicate and PredicateSpec remain the same), it's memory efficient (it doesn't store a new weight integer per predicate, and instead has the integer in the PredicateSpec itself) and it gives a ton of flexibility for those using Skipper as a library. 

### Very long context: 

This PR adds the ability to provide extra weight for custom predicates (there were some conversations about it in https://github.com/zalando/skipper/pull/1338, even if no consensus).

The reason why we would benefit of such feature is we have some custom predicates that evaluate the HTTP request and need to build some decision tree, and it's getting complex to manage weight independently of the predicate itself, as it forces use to re-evaluate queries as they are written to dynamically assign a value to the `Weight` predicate.

Some (simplified) example of the routes and custom predicates we have:

```
FEATURE_AND_VARIATION: Path("/") && InAbTest("experiment", "variation1") && FeatureEnabled("newFeature") -> setPath("/featureWithVariation") -> http://example.com
FEATURE_ONLY: Path("/") && FeatureEnabled("newFeature") -> setPath("/feature") -> http://example.com
VARIATION_ONLY: Path("/") && InAbTest("experiment", "variation1") -> setPath("/variation1") -> http://example.com
```
In this sample Eskip file, we're aiming to route to `FEATURE_AND_VARIATION` if both predicates match. But we can't just build composed routes all the time, because it wouldn't scale.

In that example, we want to give a higher weight to `FeatureEnabled` than to `InAbTest`, so we could do...

```
FEATURE_ONLY: Path("/") && FeatureEnabled("newFeature") && Weight(1)-> setPath("/feature") -> http://example.com
VARIATION_ONLY: Path("/") && InAbTest("experiment", "variation1") -> setPath("/variation1") -> http://example.com
```

But if there's something else in there (like another test running, or some other predicate), we still want `feature` to be the one, so by adding another route, we would need to update the weight of the first route to keep it working as we want.

```
FEATURE_ONLY: Path("/") && FeatureEnabled("newFeature") && Weight(2)-> setPath("/feature") -> http://example.com
VARIATION_ONLY: Path("/") && InAbTest("experiment", "variation1") -> setPath("/variation1") -> http://example.com
TWO_EXPERIMENTS_: Path("/") && InAbTest("experiment1", "variation1")  && InAbTest("experiment2", "variation1") -> setPath("/exp1exp2") -> http://example.com
```

An alternative (the one we're using) is to, for example, always add 1000 to the weight of a route including `FeatureEnabled`, 10 to the weight of a route including a `InAbTest`, etc. So when writing the routes, we keep a counter and depending on the predicates (and their types), we end up writing the accumulated weight:

```
FEATURE_AND_VARIATION: Path("/") && InAbTest("experiment", "variation1") && FeatureEnabled("newFeature") && Weight(1010) -> setPath("/featureWithVariation") -> http://example.com
FEATURE_ONLY: Path("/") && FeatureEnabled("newFeature")  && Weight(1000) -> setPath("/feature") -> http://example.com
VARIATION_ONLY: Path("/") && InAbTest("experiment", "variation1")  && Weight(10) -> setPath("/variation1") -> http://example.com
```

After (if) this PR is merged, we'd add those custom weights to the predicates themselves, so `FeatureEnabled` always adds a weight of 1000, and `InAbTest` always adds a weight of 10.

(as I said originally, this is a simplification: we already have more predicate types with different "built-in" weights)

### DIFF
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -394,7 +394,6 @@ func processPredicates(cpm map[string]PredicateSpec, defs []*eskip.Predicate) ([
 			weight += ws.Weight()
 		}
 
-
 		cps = append(cps, cp)
 	}
 
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -364,10 +364,14 @@ func processPredicates(cpm map[string]PredicateSpec, defs []*eskip.Predicate) ([
 	var weight int
 	for _, def := range defs {
 		if def.Name == predicates.WeightName {
+			var w int
 			var err error
-			if weight, err = parseWeightPredicateArgs(def.Args); err != nil {
+
+			if w, err = parseWeightPredicateArgs(def.Args); err != nil {
 				return nil, 0, err
 			}
+
+			weight += w
 
 			continue
 		}
@@ -385,6 +389,11 @@ func processPredicates(cpm map[string]PredicateSpec, defs []*eskip.Predicate) ([
 		if err != nil {
 			return nil, 0, fmt.Errorf("failed to create predicate %q: %w", spec.Name(), err)
 		}
+
+		if ws, ok := spec.(WeightedPredicateSpec); ok {
+			weight += ws.Weight()
+		}
+
 
 		cps = append(cps, cp)
 	}
--- a/routing/datasource_test.go
+++ b/routing/datasource_test.go
@@ -148,6 +148,9 @@ func TestProcessRouteDefWeight(t *testing.T) {
 			`Weight(20) -> <shunt>`,
 			20,
 		}, {
+			`Weight(20) && Weight(10)-> <shunt>`,
+			30,
+		}, {
 			`WeightedPredicate10() && Weight(20) -> <shunt>`,
 			30,
 		},
--- a/routing/datasource_test.go
+++ b/routing/datasource_test.go
@@ -284,9 +284,7 @@ func TestLogging(t *testing.T) {
 }
 
 type weightedPredicateSpec struct{}
-type weightedPredicate struct {
-	Weight int
-}
+type weightedPredicate struct{}
 
 func (w weightedPredicate) Match(request *http.Request) bool {
 	return true
@@ -297,9 +295,7 @@ func (w weightedPredicateSpec) Name() string {
 }
 
 func (w weightedPredicateSpec) Create([]interface{}) (routing.Predicate, error) {
-	return weightedPredicate{
-		Weight: w.Weight(),
-	}, nil
+	return weightedPredicate{}, nil
 }
 
 func (w weightedPredicateSpec) Weight() int {
--- a/routing/datasource_test.go
+++ b/routing/datasource_test.go
@@ -1,6 +1,7 @@
 package routing_test
 
 import (
+	"net/http"
 	"testing"
 	"time"
 
@@ -126,6 +127,65 @@ func TestProcessRouteDefErrors(t *testing.T) {
 	}
 }
 
+func TestProcessRouteDefWeight(t *testing.T) {
+
+	wps := weightedPredicateSpec{}
+
+	cpm := make(map[string]routing.PredicateSpec)
+	cpm[wps.Name()] = wps
+
+	for _, ti := range []struct {
+		route  string
+		weight int
+	}{
+		{
+			`Path("/foo") -> <shunt>`,
+			0,
+		}, {
+			`WeightedPredicate10() -> <shunt>`,
+			10,
+		}, {
+			`Weight(20) -> <shunt>`,
+			20,
+		}, {
+			`WeightedPredicate10() && Weight(20) -> <shunt>`,
+			30,
+		},
+	} {
+		func() {
+
+			dc, err := testdataclient.NewDoc(ti.route)
+			if err != nil {
+				t.Error(ti.route, err)
+
+				return
+			}
+
+			defs, err := dc.LoadAll()
+			if err != nil {
+				t.Error(ti.route, err)
+
+				return
+			}
+
+			r := defs[0]
+
+			_, weight, err := routing.ExportProcessPredicates(cpm, r.Predicates)
+			if err != nil {
+				t.Error(ti.route, err)
+
+				return
+			}
+
+			if weight != ti.weight {
+				t.Errorf("expected weight '%d'. Got: '%d' (%s)", ti.weight, weight, ti.route)
+
+				return
+			}
+		}()
+	}
+}
+
 func TestLogging(t *testing.T) {
 	if testing.Short() {
 		t.Skip()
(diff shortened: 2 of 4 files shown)


xavivars:
Will rebase on top of https://github.com/zalando/skipper/pull/2053 once merged, as failures are due to build process using 1.19. But would love to start discussion about the feature itself.

AlexanderYastrebov:
Hi, I get the idea of the change but struggle to understand the problem.

Could you maybe provide an example of a configuration that you think is hard to manage currently and _equivalent_ configuration if we assume this feature is present.

xavivars:
Hey @AlexanderYastrebov, sure. I tried in the longer explanation above, but I clearly didn't manage 😅 

This is a sample eskip file, using 3 custom predicates: InFeature, InExperiment, and InAudience. And we want the following priorities: InFeature > InExperiment > InAudience > Regular Predicates (QueryParam, Cookie, etc).

Currently, we could just add some weights whenever we add a predicate, like the priority described above: InFeature => 3, InExperiment => 2 and InAudience => 1.

```
Path("/") && InFeature() && Weight(3) -> "http://feature.backend.com"
Path("/") && InExperiment() && Weight(2) -> "http://experiment.backend.com"
Path("/") && InAudience() && Weight(1) -> "http://audience.backend.com"
```

And this works, great, we wouldn't really need this feature yet, but....

```
ROUTE_1: Path("/") && InFeature() && Weight(3) -> "http://feature.backend.com"
ROUTE_2: Path("/") && InExperiment() && Weight(2) && QueryParam("foo","bar") -> "http://experiment.backend.com"
ROUTE_3: Path("/") && InAudience() && Weight(1) -> "http://audience.backend.com"
ROUTE_4: Path("/") && QueryParam("foo","bar")  -> "http://default.backend.com"
```

Now, suddenly, ROUTE_2 and ROUTE_1 have the same weight, so this is not good enough. Same for ROUTE_3 and ROUTE_4. Let's make this a little bit more error proof: InFeature: 1,000,000, InExperiment 10,000 and InFeature 100

```
ROUTE_1: Path("/") && InFeature() && Weight(1000000) -> "http://feature.backend.com"
ROUTE_2: Path("/") && InExperiment() && Weight(10000) && QueryParam("foo","bar") -> "http://experiment.backend.com"
ROUTE_3: Path("/") && InAudience() && Weight(100) -> "http://audience.backend.com"
ROUTE_4: Path("/") && QueryParam("foo","bar")  -> "http://default.backend.com"
```

So it seems pretty easy. While it's not the best option possible (real priorities) using weights this way gets close enough. So the only thing we would need to do is every time we add InFeature to a route, we also add Weight(10^6), and that solves it, right? 

Wrong.

Let's imagine the following scenario (without weights)

```
ROUTE_1: Path("/") && InFeature() -> "http://feature.backend.com"
ROUTE_2: Path("/") && InFeature() && InExperiment() -> "http://combined.backend.com"
ROUTE_3: Path("/") && InExperiment() -> "http://experiment.backend.com"
```

In such scenario, we wouldn't be able to just add a weight predicate every time we add one of our custom predicates, because this wouldn't be valid:

```
ROUTE_1: Path("/") && InFeature() && Weight(1,000,000)-> "http://feature.backend.com"
ROUTE_2: Path("/") && InFeature() && Weight(1,000,000) && InExperiment() && Weight(10,000)-> "http://combined.backend.com" <----- This is not supported currently
ROUTE_3: Path("/") && InExperiment() && Weight(10,000) -> "http://experiment.backend.com"
```

Each route can only have one `Weight` predicate. 

So instead we would need to either
- keep route creation process simple, but create a (somewhat complex) post-processing step to parse and rewrite the routes to add a combined `Weight` predicate.
- make the route generation more complex to keep that "counter" updated _while writing the route_, so that we can write a single Weight predicate at the end.


Skipper already does the hard parts of both options here: does parsing of all incoming routes, and it also keeps a weight counter (using `Weight` predicate + amount of regular predicates). So this MR just connects the dots: allows custom predicates to also have a custom weight. In our case, that would significantly simplify some routes, improving readability. And as a _nice side effect_, it also enables having multiple Weight predicates in the same route for those who want to leverage that as a feature.

routing/datasource.go
```diff
@@ -364,11 +364,15 @@ func processPredicates(cpm map[string]PredicateSpec, defs []*eskip.Predicate) ([
...
 		if def.Name == predicates.WeightName {
+			var w int
 			var err error
-			if weight, err = parseWeightPredicateArgs(def.Args); err != nil {
+
+			if w, err = parseWeightPredicateArgs(def.Args); err != nil {
 				return nil, 0, err
 			}
 
+			weight += w
```
szuecs, line 374: This would sum all occurrences of `Weight(), so  `Weight(5) && Weight(15) = 20` and not `15` anymore

routing/datasource.go
```diff
@@ -386,6 +390,10 @@ func processPredicates(cpm map[string]PredicateSpec, defs []*eskip.Predicate) ([
 			return nil, 0, fmt.Errorf("failed to create predicate %q: %w", spec.Name(), err)
 		}
 
+		if ws, ok := spec.(WeightedPredicateSpec); ok {
+			weight += ws.Weight()
```
szuecs, line 394: This is enabling predicates to have their own weight

routing/datasource_test.go
```diff
@@ -126,6 +127,65 @@ func TestProcessRouteDefErrors(t *testing.T) {
...
+		weight int
+	}{
+		{
+			`Path("/foo") -> <shunt>`,
+			0,
+		}, {
+			`WeightedPredicate10() -> <shunt>`,
+			10,
+		}, {
+			`Weight(20) -> <shunt>`,
```
szuecs, line 148: Please add also a test with more than one `Weight` to make sure the count is like we want to have.

routing/datasource_test.go
```diff
@@ -222,3 +282,26 @@ func TestLogging(t *testing.T) {
 		)
 	})
 }
+
+type weightedPredicateSpec struct{}
+type weightedPredicate struct {
+	Weight int
```
szuecs, line 288: We don't need this

routing/datasource.go
xavivars: Correct. And that would technically be a backwards incompatible change. But being honest, I don't think anyone would expect that `PredicateA() && PredicateB()` worked differently than `PredicateB() && PredicateA()`...

routing/datasource_test.go
xavivars: Done.

routing/datasource.go
xavivars: Yes

routing/datasource.go
szuecs: I agree

szuecs:
@AlexanderYastrebov wdyt?
I think :+1: , what about increase minor version to v0.14, we are already again >x.y.200, so we could show this incompatible change in the release and highlight it.

AlexanderYastrebov:
https://github.com/zalando/skipper/pull/2054#issuecomment-1223784987

@xavivars Thank you for the explanation. Basically you want to have a weight associated with your custom predicate.
One way would be to only fix `Weight(X)` accumulation and use `MyPredicate1() && Weight(MyPredicate1Weight)`, `MyPredicate2() && Weight(MyPredicate2Weight)` pairs everywhere. You also propose to allow custom predicates to have their own associated weight to avoid carrying `Weight(X)` along with each custom predicate.

AlexanderYastrebov:
:+1:

xavivars:
https://github.com/zalando/skipper/pull/2054#issuecomment-1224066825

Thanks for summarizing :)

szuecs:
:+1:

PR #2060, opened 2022-08-22
Custom separator edit route

RomanZavodskikh:
Change -edit-route and -clone-route to support arbitrary pattern separator

### DIFF
--- a/config/routechangerconfig.go
+++ b/config/routechangerconfig.go
@@ -9,17 +9,22 @@ import (
 type routeChangerConfig struct {
 	Reg  *regexp.Regexp
 	Repl string
+	Sep  string
 }
 
 func (rcc routeChangerConfig) String() string {
 	if rcc.Reg == nil {
 		return ""
 	}
-	return fmt.Sprintf("/%s/%s/", rcc.Reg, rcc.Repl)
+	return fmt.Sprintf("%s%s%s%s%s", rcc.Sep, rcc.Reg, rcc.Sep, rcc.Repl, rcc.Sep)
 }
 
 func (rcc *routeChangerConfig) Set(value string) error {
-	a := strings.Split(value, "/")
+	if len(value) == 0 {
+		return fmt.Errorf("empty string as an argument is not allowed")
+	}
+	firstSym := value[:1]
+	a := strings.Split(value, firstSym)
 	if len(a) != 4 {
 		return fmt.Errorf("unexpected size of string split: %d", len(a))
 	}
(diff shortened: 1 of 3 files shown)


eskip/eskip.go
```diff
@@ -32,16 +32,18 @@ var (
 //        r0: Source("127.0.0.1/8", "10.0.0.0/8") -> inlineContent("OK") -> <shunt>
 //        # actual route
 //        edit_r0: ClientIP("127.0.0.1/8", "10.0.0.0/8") -> inlineContent("OK") -> <shunt>
-func NewEditor(reg *regexp.Regexp, repl string) *Editor {
+func NewEditor(reg *regexp.Regexp, repl string, sep string) *Editor {
 	return &Editor{
 		reg:  reg,
 		repl: repl,
+		sep:  sep,
```
AlexanderYastrebov, line 39: Why do we need to know separator here?

config/routechangerconfig.go
```diff
@@ -9,24 +9,27 @@ import (
...
 	if rcc.Reg == nil {
 		return ""
 	}
-	return fmt.Sprintf("/%s/%s/", rcc.Reg, rcc.Repl)
+	return fmt.Sprintf("%s%s%s%s%s", rcc.Sep, rcc.Reg, rcc.Sep, rcc.Repl, rcc.Sep)
 }
 
 func (rcc *routeChangerConfig) Set(value string) error {
-	a := strings.Split(value, "/")
+	firstSym := value[:1]
```
AlexanderYastrebov, line 23: We should check for empty `value` before, otherwise this would cause panic

eskip/eskip_test.go
```diff
@@ -620,6 +628,15 @@ func TestEditorPreProcessor(t *testing.T) {
...
 			want:   r1FilterChanged,
+		},
+		{
+			name: "test that regex with / works properly",
+			rep: &Editor{
+				reg:  regexp.MustCompile("/26"),
+				repl: "/24",
+			},
+			routes: r2,
+			want:   r2Changed,
```
AlexanderYastrebov, line 639: I am not sure we need this test - it tests that regexp replacement works but it does not test the change this PR introduces

AlexanderYastrebov:
It should be mentioned somewhere in the docs that first symbol of the value is used as a separator

eskip/eskip.go
RomanZavodskikh: I don't see any obvious reason for separator here too... However, won't we need it for some debug/logging/etc purposes?

AlexanderYastrebov: The Editor does not have to know that there is any separator involved at all - it should only know about match and replacement values

AlexanderYastrebov:
:+1:

szuecs:
:+1:

szuecs:
Thanks! Great first small change :)

Issue #2062, opened 2022-08-22
gosec v2.13.1 reports issue for use of http.ListenAndServe


mikkeloscar:
With `gosec v2.13.1` the following issues are reported:

```
make gosec
gosec -quiet -exclude="G101,G104,G304,G402" ./...
Results:


[/home/moscar/projects/teapot/skipper/skipper.go:1758] - G114 (CWE): Use of net/http serve function that has no support for setting timeouts (Confidence: HIGH, Severity: MEDIUM)
    1757: 		go func() {
  > 1758: 			if err := http.ListenAndServe(supportListener, mux); err != nil {
    1759: 				log.Errorf("Failed to start supportListener on %s: %v", supportListener, err)



[/home/moscar/projects/teapot/skipper/skipper.go:1734] - G114 (CWE): Use of net/http serve function that has no support for setting timeouts (Confidence: HIGH, Severity: MEDIUM)
    1733: 		log.Infof("debug listener on %v", o.DebugListener)
  > 1734: 		go func() { http.ListenAndServe(o.DebugListener, dbg) }()
    1735: 	}



Summary:
  Gosec  : dev
  Files  : 271
  Lines  : 48364
  Nosec  : 18
  Issues : 2

make: *** [Makefile:149: gosec] Error 1
```

The G114 check was introduced here: https://github.com/securego/gosec/pull/842

We should either exclude the check for now or fix the use of `http.ListenAndServe` to instead use a custom `http.Server` with timeouts as described here: https://blog.cloudflare.com/exposing-go-on-the-internet/

szuecs:
So gosec found the support listener and debug listener, which should not be deployed such that everyone from the internet can access it. I think we can safely add a `/* nosec*/` directive there to opt-out this check in the code part

PR #2018, opened 2022-06-15
gosec: configure http.Server timeouts to fix G112

AlexanderYastrebov:
Signed-off-by: Alexander Yastrebov <alexander.yastrebov@zalando.de>

PR #2063, opened 2022-08-22
ignore gosec g114 for support and debug listener

szuecs:
fix https://github.com/zalando/skipper/issues/2062

ignore gosec g114 for support and debug listener which should not be exposed to untrusted entities


AlexanderYastrebov:
I think I did a related fix within #2018
Also error checking is missing in one place and is incorrect in another (should check for ErrServerClosed), but I guess since these servers are never shutdown properly its not a big deal

AlexanderYastrebov:
:+1:

szuecs:
:+1:

szuecs:
If we want to be able to reload server to change timeout settings for example we can check it, but right now it's additional code that makes not much sense.

Issue #2065, opened 2022-08-23
Allow insecure traffic on separate port when TLS termination enabled


rickhlx:
**Is your feature request related to a problem? Please describe.**
When configuring TLS termination using Skipper Ingress TLS, all requests to Skipper need to be encrypted. It is not possible to have both `http` and `https` requests simultaneously. 

**Describe the solution you would like**
Allow users to specify separate ports for encrypted and unencrypted traffic when terminating TLS at Skipper. This could be done by allowing users to have a `-insecure-address` (default to `:9080`?) flag for when TLS is configured. 

**Describe alternatives you've considered (optional)**
* Having two instances of Skipper running, one for encrypted and another for non-encrypted. (Not a very good solution IMO)

**Would you like to work on it?**
Yes

szuecs:
I am fine with having this option and the question is how to achieve this with a non breaking change.
Right now the proposal would likely do a breaking change for people who run `-address=:9999` without TLS cert/key and for library users as well that won't set the new option.

rickhlx:
@szuecs We could ignore the `-insecure-address` flag when not using TLS cert/keys since traffic would already be unencrypted. This option would only be effective when TLS is enabled. Would that work?

PR #2066, opened 2022-08-24
feature: allow insecure connections when using tls

rickhlx:
Allow insecure connections on an alternate port when using TLS. This
will remove the requirement to encrypt all traffic when enabling TLS.

Fixes: #2065

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -30,6 +30,7 @@ type Config struct {
 
 	// generic:
 	Address                         string         `yaml:"address"`
+	InsecureAddress                 string         `yaml:"insecure-address"`
 	EnableTCPQueue                  bool           `yaml:"enable-tcp-queue"`
 	ExpectedBytesPerRequest         int            `yaml:"expected-bytes-per-request"`
 	MaxTCPListenerConcurrency       int            `yaml:"max-tcp-listener-concurrency"`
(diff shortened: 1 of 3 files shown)


skipper.go
```diff
@@ -1175,6 +1178,23 @@ func listenAndServeQuit(
 	log.Infof("proxy listener on %v", o.Address)
 
 	if srv.TLSConfig != nil {
+		if o.InsecureAddress != "" {
+			log.Infof("insecure listener on %v", o.InsecureAddress)
+
+			go func() {
+				o.Address = o.InsecureAddress
```
AlexanderYastrebov, line 1185: I do not like that we change `o.Address` here. Maybe we should change `listen` to receive address separately from options.

config/config.go
```diff
@@ -298,6 +299,7 @@ func NewConfig() *Config {
 
 	// generic:
 	flag.StringVar(&cfg.Address, "address", ":9090", "network address that skipper should listen on")
+	flag.StringVar(&cfg.InsecureAddress, "insecure-address", "", "insecure network address that skipper should listen on")
```
AlexanderYastrebov, line 302: `when TLS is enabled`

AlexanderYastrebov:
:+1:

szuecs:
:+1:

szuecs:
Thanks @rickhlx !

PR #2067, opened 2022-08-26
Fix: memory leak in admissionControl filter.

szuecs:
Fix: memory leak in admissionControl filter. In case routing changes triggers a lot of routing updates (f.e. every 3s) this would cause a leak of tickers and updates would be sent to the channel by Go runtime.

image

### DIFF
--- a/filters/shedder/admission.go
+++ b/filters/shedder/admission.go
@@ -331,6 +331,9 @@ func (ac *admissionControl) Close() {
 func (ac *admissionControl) tickWindows(d time.Duration) {
 	t := time.NewTicker(d)
 	i := 0
+	defer func() {
+		t.Stop()
+	}()
 	for range t.C {
 		select {
 		case <-ac.quit:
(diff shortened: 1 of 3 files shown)


filters/shedder/admission.go
```diff
@@ -331,6 +331,9 @@ func (ac *admissionControl) Close() {
 func (ac *admissionControl) tickWindows(d time.Duration) {
 	t := time.NewTicker(d)
 	i := 0
+	defer func() {
+		t.Stop()
+	}()
```
AlexanderYastrebov, line 336: Why not `defer t.Stop()`?

AlexanderYastrebov:
:+1:

filters/shedder/admission.go
szuecs: yeah true left over because I tried to close(t.C) but that did not work because of it's only receiver side

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2073, opened 2022-09-07
fix: fifo error metric counted "other" errors in case there is no error

szuecs:
fix: counter should reflect semaphore state, right now metrics show only 0 or 1 sometimes 2 as "active"

### DIFF
--- a/scheduler/export_test.go
+++ b/scheduler/export_test.go
@@ -3,3 +3,13 @@ package scheduler
 var (
 	ExportQueueCloseDelay = &queueCloseDelay
 )
+
+// GetFifoForeTest is only compiled and used in tests
+func (r *Registry) GetFifoForeTest(s string) (*FifoQueue, bool) {
+	id := queueId{
+		name:    s,
+		grouped: false,
+	}
+	q, ok := r.fifoQueues[id]
+	return q, ok
+}
--- a/scheduler/export_test.go
+++ b/scheduler/export_test.go
@@ -4,8 +4,8 @@ var (
 	ExportQueueCloseDelay = &queueCloseDelay
 )
 
-// GetFifoForeTest is only compiled and used in tests
-func (r *Registry) GetFifoForeTest(s string) (*FifoQueue, bool) {
+// GetFifoForTest is only compiled and used in tests
+func (r *Registry) GetFifoForTest(s string) (*FifoQueue, bool) {
 	id := queueId{
 		name:    s,
 		grouped: false,
--- a/scheduler/fifo_test.go
+++ b/scheduler/fifo_test.go
@@ -11,16 +11,20 @@ import (
 func TestFifo(t *testing.T) {
 	waitForStatus := func(t *testing.T, fq *FifoQueue, s QueueStatus) {
 		t.Helper()
+		if fq == nil {
+			t.Fatal("fq nil")
+		}
 		timeout := time.After(120 * time.Millisecond)
 		for {
 			time.Sleep(time.Millisecond)
-			if fq != nil && fq.Status() == s {
+			cur := fq.Status()
+			if cur == s {
 				return
 			}
 
 			select {
 			case <-timeout:
-				t.Fatal("failed to reach status")
+				t.Fatalf("failed to reach status, want %v, got: %v", s, cur)
 			default:
 			}
 		}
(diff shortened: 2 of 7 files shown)


scheduler/fifo_test.go
```diff
@@ -14,13 +14,14 @@ func TestFifo(t *testing.T) {
 		timeout := time.After(120 * time.Millisecond)
 		for {
 			time.Sleep(time.Millisecond)
-			if fq != nil && fq.Status() == s {
+			cur := fq.Status()
+			if fq != nil && cur == s {
```
AlexanderYastrebov, line 18: This calls `fq.Status()`  but then checks `fq != nil`

scheduler/scheduler.go
```diff
@@ -283,6 +285,8 @@ func (fq *FifoQueue) Wait(ctx context.Context) (func(), error) {
 			fq.metrics.IncCounter(fq.errorTimeoutMetricsKey)
 		case ErrClientCanceled:
 			// This case is handled in the proxy with status code 499
+		case nil:
+			// no error
```
AlexanderYastrebov, line 289: Maybe `if fq.metrics != nil && err != nil` to hint that this block only handles error case?

scheduler/scheduler.go
```diff
@@ -125,18 +126,18 @@ func (fq *fifoQueue) status() QueueStatus {
...
-// Reconfigure does nothing, because it creates a data race that would
-// require more mutex than we want to have
 func (fq *fifoQueue) reconfigure(c Config) {
 	fq.mu.Lock()
 	defer fq.mu.Unlock()
 	fq.maxConcurrency = uint64(c.MaxConcurrency)
 	fq.maxQueueSize = uint64(c.MaxQueueSize)
 	fq.timeout = c.Timeout
+	fq.sem = semaphore.NewWeighted(int64(c.MaxConcurrency))
+	fq.counter.Store(uint64(0))
```
AlexanderYastrebov, line 140: I think this is not correct - there might be in-flight requests that would call Dec() resulting in overflow.

szuecs: I will check what happens in this case and make sure it will be 0

szuecs: I found another bug with the counter while reconfigure we have a race condition in accessing the counter which can show up in high routing table update rate (test with 5ms flapping fifo config).

scheduler/scheduler.go
```diff
@@ -273,8 +282,14 @@ type GroupedLIFOFilter interface {
 // request needs to be rejected, an error will be returned and done
 // will be nil.
 func (fq *FifoQueue) Wait(ctx context.Context) (func(), error) {
+	if fq == nil {
+		return nil, errors.New("fq is nil")
+	}
+	if fq.queue == nil {
+		return nil, errors.New("queue is nil")
+	}
```
szuecs, line 290: should be cleaned up

scheduler/export_test.go
```diff
@@ -3,3 +3,13 @@ package scheduler
 var (
 	ExportQueueCloseDelay = &queueCloseDelay
 )
+
+// GetFifoForeTest is only compiled and used in tests
+func (r *Registry) GetFifoForeTest(s string) (*FifoQueue, bool) {
```
szuecs, line 8: GetFifoForeTest -> GetFifoForTest

scheduler/scheduler.go
```diff
@@ -125,39 +126,44 @@ func (fq *fifoQueue) status() QueueStatus {
...
 
-	// limit concurrency
+	// set timeout
 	c, done := context.WithTimeout(ctx, timeout)
 	defer done()
-	if err := fq.sem.Acquire(c, 1); err != nil {
+
+	// limit concurrency
+	if err := sem.Acquire(c, 1); err != nil {
+		fq.counter.Dec()
```
szuecs, line 166: use `cnt` instead

scheduler/scheduler.go
```diff
@@ -299,12 +314,17 @@ func (fq *FifoQueue) Config() Config {
...
+// reset the current state.
+func (fq *FifoQueue) Reconfigure(c Config) {
+	fq.config = c
+	fq.queue.reconfigure(c)
 }
 
 func (fq *FifoQueue) close() {
+	fq.queue.mu.Lock()
 	fq.queue.closed = true
+	fq.queue.mu.Unlock()
```
szuecs, line 327: this should delegate to fq.queue.close()

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2075, opened 2022-09-13
Tool/govulncheck should not block build

szuecs:
run govulncheck only on merge to master and do not block the build


AlexanderYastrebov:
:+1:

szuecs:
:+1:

Issue #2071, opened 2022-09-01
Replace matched regex in Response header


ioanacraciun:
**Is your feature request related to a problem? Please describe.**
Currently, as part of the Skipper Filters, only the `modRequestHeader` is provided. It replaces all matched regex expressions in the given request header. I would need to replace an expression in the response header, in my use case `Location`.

**Describe the solution you would like**
It would be useful to have a counterpart filter for `modRequestHeader`, except for response headers.

**Describe alternatives you've considered (optional)**

**Additional context (optional)**

**Would you like to work on it?**
No

szuecs:
Sound good

RomanZavodskikh:
Resolved in PR above.

PR #2076, opened 2022-09-14
Added modResponseHeader filter

RomanZavodskikh:

### DIFF
--- a/filters/builtin/mod_header.go
+++ b/filters/builtin/mod_header.go
@@ -73,3 +73,60 @@ func (f *modRequestHeader) Request(ctx filters.FilterContext) {
 }
 
 func (*modRequestHeader) Response(filters.FilterContext) {}
+
+type modResponseHeader struct {
+	headerName  string
+	rx          *regexp.Regexp
+	replacement string
+}
+
+// NewModResponseHeader returns a new filter Spec, whose instances execute
+// regexp.ReplaceAllString on the request host. Instances expect three
+// parameters: the header name, the expression to match and the replacement string.
+// Name: "modResponseHeader".
+func NewModResponseHeader() filters.Spec { return &modResponseHeader{} }
+
+func (spec *modResponseHeader) Name() string {
+	return filters.ModResponseHeaderName
+}
+
+//lint:ignore ST1016 "spec" makes sense here and we reuse the type for the filter
+func (spec *modResponseHeader) CreateFilter(config []interface{}) (filters.Filter, error) {
+	if len(config) != 3 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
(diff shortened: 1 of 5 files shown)


RomanZavodskikh:
Trying to fix issue https://github.com/zalando/skipper/issues/2071 with this PR.

filters/builtin/mod_header.go
```diff
@@ -73,3 +73,60 @@ func (f *modRequestHeader) Request(ctx filters.FilterContext) {
...
+}
+
+func (*modResponseHeader) Request(filters.FilterContext) {}
+
+func (f *modResponseHeader) Response(ctx filters.FilterContext) {
+	resp := ctx.Response()
+
+	if _, ok := resp.Header[http.CanonicalHeaderKey(f.headerName)]; !ok {
+		return
+	}
```
AlexanderYastrebov, line 129: There is a highlevel https://pkg.go.dev/net/http#Header.Get

AlexanderYastrebov:
I think the test is missing

RomanZavodskikh:
@AlexanderYastrebov , in this case 
```
google:
        *
        -> setPath("/search")
        -> setQuery("q", "godoc skipper")
        -> setResponseHeader("k1", "v489")
        -> modResponseHeader("k1", "^v489$", "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA")
        -> "https://www.google.com";
```
modResponseHeader does not modify `v489` value, but in this case
```
google:
        *
        -> setPath("/search")
        -> setQuery("q", "godoc skipper")
        -> setResponseHeader("k1", "v489")
        -> modResponseHeader("k1", "^v489$", "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA")
        -> "https://www.google.com";
```
it does. It is counterintuitive for me. Why response filters work from bottom to up, not vica versa? Is it the way it should be?

filters/builtin/mod_header.go
RomanZavodskikh: Yes, there is but it returns empty string in case of key is not present in header. I think it is better to leave header unpresent in this case than implicitly create a header with empty string value.

szuecs: I think the edge case in case there is a header with empty string as value and you do a modification like:
```
/.*/foo/
```
If there is no header it should stay no header and if the header is present with empty string it should set the value to "foo".

szuecs:
@RomanZavodskikh think of it as pipeline from client to backend and reverse. If you like you can think of filters as middleware chain.
Maybe counterintuitive first but it would be wrong to change the flow.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2078, opened 2022-09-19
fix: redis autoscaling introduced a concurrent map access

szuecs:
fix: redis autoscaling introduced a concurrent map access issue with kubernetes clusterstate, which was only used by one goroutine

### DIFF
--- a/dataclients/kubernetes/clusterstate.go
+++ b/dataclients/kubernetes/clusterstate.go
@@ -3,6 +3,7 @@ package kubernetes
 import (
 	"fmt"
 	"sort"
+	"sync"
 
 	log "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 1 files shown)


dataclients/kubernetes/clusterstate.go
```diff
@@ -49,6 +55,8 @@ func (state *clusterState) GetEndpointsByService(namespace, name, protocol strin
 		TargetPort: servicePort.TargetPort.String(),
 	}
 
+	state.mu.Lock()
+	defer state.mu.Unlock()
```
AlexanderYastrebov, line 59: Ideally we need to get rid of `state.cachedEndpoints` (or prepupulate it during cluster state construction) to make cluster state read-only

AlexanderYastrebov: https://github.com/zalando/skipper/issues/2079

szuecs:
:+1:

Issue #2079, opened 2022-09-19
Make cluster state read-only


AlexanderYastrebov:
Cluster state has `cachedEndpoints` that is updated in runtime and may cause `fatal error: concurrent map read and map write`
To mitigate the problem #2078 synchronizes access to it (somewhat indirectly).

The better way is to make cluster state immutable/read-only either by pre-populating `cachedEndpoints` or just getting rid of it and use `endpoints` on each lookup.

PR #2080, opened 2022-09-21
Made cachedEndpoints read-only to avoid write+read races

RomanZavodskikh:
Fixes https://github.com/zalando/skipper/issues/2079

### DIFF
--- a/dataclients/kubernetes/clusterstate_test.go
+++ b/dataclients/kubernetes/clusterstate_test.go
@@ -1,0 +1,87 @@
+package kubernetes
+
+import (
+	"strconv"
+	"testing"
+
+	"github.com/zalando/skipper/dataclients/kubernetes/definitions"
+)
+
+var dummy2 []string
+
+func benchmarkCachedEndpoints(b *testing.B, n int) {
+	endpoints := make(map[definitions.ResourceID]*endpoint)
+	for i := 0; i < n; i++ {
+		name := "foo-" + strconv.Itoa(i)
+		rid := newResourceID("default", name)
+		ep := &endpoint{
+			Meta: &definitions.Metadata{
+				Namespace: "default",
+				Name:      name,
+				Labels:    map[string]string{},
+			},
+			Subsets: []*subset{
+				{
+					Addresses: []*address{
+						{"192.168.0.1", "node1"},
+						{"192.168.0.2", "node2"},
+						{"192.168.0.3", "node3"},
+						{"192.168.0.4", "node4"},
+						{"192.168.0.5", "node5"},
+						{"192.168.0.6", "node6"},
+						{"192.168.0.7", "node7"},
+						{"192.168.0.8", "node8"},
+						{"192.168.0.9", "node9"},
+						{"192.168.0.10", "node10"},
+						{"192.168.0.11", "node11"},
+					},
+					Ports: []*port{
+						{"ssh", 22, "TCP"},
+						{"http", 80, "TCP"},
+					},
+				},
+			},
+		}
+		endpoints[rid] = ep
+	}
+
+	cs := &clusterState{
+		ingresses:       nil,
+		ingressesV1:     nil,
+		routeGroups:     nil,
+		services:        nil,
+		endpoints:       endpoints,
+		secrets:         nil,
+		cachedEndpoints: make(map[endpointID][]string),
+	}
+
+	b.ResetTimer()
+	dummy := []string{}
+	for i := 0; i < b.N; i++ {
+		dummy = cs.GetEndpointsByTarget("default", "foo-0", "TCP", &definitions.BackendPort{})
+	}
+	dummy2 = dummy
+}
+
+func BenchmarkCachedEndpoint(b *testing.B) {
+	for _, tt := range []struct {
+		name            string
+		endpointsNumber int
+	}{
+		{
+			name:            "1M Endpoints",
+			endpointsNumber: 1_000_000,
+		},
+		{
+			name:            "10K Endpoints",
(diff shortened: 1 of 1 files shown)


dataclients/kubernetes/clusterstate.go
```diff
@@ -3,15 +3,13 @@ package kubernetes
...
 	"sort"
-	"sync"
 
 	log "github.com/sirupsen/logrus"
 
 	"github.com/zalando/skipper/dataclients/kubernetes/definitions"
 )
 
 type clusterState struct {
-	mu              sync.Mutex
```
AlexanderYastrebov, line 14: I think cachedEndpoints should be removed as well

RomanZavodskikh:
https://github.com/zalando/skipper/issues/2079

RomanZavodskikh:
Need to write some performance tests to find out whether we need to remove cachedEndpoints.

dataclients/kubernetes/clusterstate_test.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"math/rand"
+	"testing"
+
+	"github.com/zalando/skipper/dataclients/kubernetes/definitions"
+)
+
+var letterRunes = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+
+func RandStringRunes(n int) string {
```
szuecs, line 12: does not need to be exposed, so lower case the first character of the func

dataclients/kubernetes/clusterstate_test.go
```diff
@@ -0,0 +1,60 @@
...
+	for i := range b {
+		b[i] = letterRunes[rand.Intn(len(letterRunes))]
+	}
+	return string(b)
+}
+
+func benchmarkCachedEndpoints(b *testing.B, n int) {
+	endpoints := make(map[definitions.ResourceID]*endpoint)
+	for i := 0; i < n; i++ {
+		name := randStringRunes(10)
```
AlexanderYastrebov, line 23: IMO this rand stuf is distrcating, we could simply use `fmt.Sprintf("foo-%d", i)` to generate distinct names

dataclients/kubernetes/clusterstate_test.go
```diff
@@ -0,0 +1,60 @@
...
+		if i == 0 {
+			name = "postgres"
+		}
+		rid := newResourceID("default", name)
+		ep := &endpoint{
+			Meta:    nil,
+			Subsets: nil,
```
AlexanderYastrebov, line 30: The functions we want to benchmark loop over  `Subsets` and `Ports` in it which is empty here

dataclients/kubernetes/clusterstate_test.go
```diff
@@ -0,0 +1,60 @@
...
+	return string(b)
+}
+
+func benchmarkCachedEndpoints(b *testing.B, n int) {
+	endpoints := make(map[definitions.ResourceID]*endpoint)
+	for i := 0; i < n; i++ {
+		name := randStringRunes(10)
+		if i == 0 {
+			name = "postgres"
+		}
```
AlexanderYastrebov, line 26: This is also redundant, target endpoint name could be `fmt.Sprintf("foo-%d", n / 2)` or even `"foo-0"`. Lets make it as simple as possible.

dataclients/kubernetes/clusterstate_test.go
```diff
@@ -0,0 +1,60 @@
...
+		ingressesV1: nil,
+		routeGroups: nil,
+		services:    nil,
+		endpoints:   endpoints,
+		secrets:     nil,
+	}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		cs.GetEndpointsByTarget("default", "postgres", "TCP", &definitions.BackendPort{})
```
AlexanderYastrebov, line 46: It could be a bit tricky to write proper benchmark due to compile optimizations, see
* https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go
* https://github.com/golang/go/issues/27400
* https://github.com/golang/go/issues/48768

Lets assign result to the global "sink" variable as suggested here https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go#compiler-optimisation

dataclients/kubernetes/clusterstate_test.go
RomanZavodskikh: Nice catch! It changes benchmark results significantly.

RomanZavodskikh:
New benchmark results:
```
benchmark                        old ns/op     new ns/op     delta
BenchmarkCachedEndpoints1-16     47.8          1901          +3876.99%
BenchmarkCachedEndpoints2-16     49.2          4664          +9371.97%
BenchmarkCachedEndpoints3-16     49.6          4866          +9702.58%

benchmark                        old allocs     new allocs     delta
BenchmarkCachedEndpoints1-16     0              17             +Inf%
BenchmarkCachedEndpoints2-16     0              17             +Inf%
BenchmarkCachedEndpoints3-16     0              17             +Inf%

benchmark                        old bytes     new bytes     delta
BenchmarkCachedEndpoints1-16     0             784           +Inf%
BenchmarkCachedEndpoints2-16     0             784           +Inf%
BenchmarkCachedEndpoints3-16     0             784           +Inf%
```

szuecs:
@RomanZavodskikh yeah sounds good to me.

AlexanderYastrebov:
> right now I lean towards merging only first commit 

Performance tests added by https://github.com/zalando/skipper/pull/2092

PR #2082, opened 2022-09-22
proxy: fix FIFO/LIFO cleanup

AlexanderYastrebov:
`*Proxy.do()` maybe called for the `<loopback>` backend which resulted in double call of pending `done` handlers.
In case `fifo` is configured on the main and loopback route double call leads to panic:

```sh
./bin/skipper -inline-routes='loop: Path("/loop") -> fifo(100, 100, "1s") -> inlineContent("ok") -> <shunt>; main: Path("/") -> fifo(100, 100, "1s") -> setPath("/loop") -> <loopback>;'

curl -v localhost:9090/

# in the Skipper logs
[APP]ERRO[0015] http: panic serving 127.0.0.1:44918: semaphore: released more than held
```

### DIFF
--- a/filters/scheduler/loopback_test.go
+++ b/filters/scheduler/loopback_test.go
@@ -1,0 +1,57 @@
+package scheduler_test
+
+import (
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

szuecs:
Tested also a couple of error cases locally and in a cluster

AlexanderYastrebov:
:+1:

PR #2083, opened 2022-09-23
Make use of default filters in `RouteSRV`

MustafaSaber:
In efforts to solve https://github.bus.zalan.do/pg9/issues/issues/33

# Summary

We need to decrease number of skipper instances pulling routes from kubeapiserver, right now if we have 100 skipper instances all of them will pull from kubeapiserver, which will put some unneeded pressure on it.

# Solution

The solution was to make a server to only pull routes from kubeapiserver and skipper will pull from that server.

**Old**
```
skipper -> apiserver
```
**New**
```
skipper -> routeSRV -> apiserver
```

RouteSRV was already implemented, but not deployed. In this PR we try to make sure there is no difference between `/routes` endpoints in skipper and routeSRV, while also moving all route handling to routeSRV instead of skipper.

- DefaultFilters
    -  - [x] Implementation
    -  - [x] Tests
    -  - [ ] Docs
- OAuth2
    -  - [x] Implementation
    -  - [x] Tests
    -  - [ ] Docs
- [ ] Go through all clusters and get the diff.
    

### DIFF
--- a/routesrv/options.go
+++ b/routesrv/options.go
@@ -7,7 +7,6 @@ import (
 	"github.com/zalando/skipper/dataclients/kubernetes"
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/filters"
-	"github.com/zalando/skipper/secrets"
 )
 
 // Options for initializing/running RouteServer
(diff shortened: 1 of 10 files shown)


routesrv/routesrv_test.go
```diff
@@ -256,3 +256,38 @@ func TestRoutesAreUpdated(t *testing.T) {
...
+	ks.Start()
+	defer ks.Close()
+	rs := newRouteServerWithOptions(t, routesrv.Options{
+		SourcePollTimeout: pollInterval,
+		KubernetesURL:     ks.URL,
+		DefaultFilters: &eskip.DefaultFilters{
+			Prepend: []*eskip.Filter{
+				{
+					Name: "enableAccessLog",
+					Args: []any{4, 5},
```
szuecs, line 272: ah nice use of any :)

szuecs:
lgtm

AlexanderYastrebov:
:+1:

routesrv/routesrv.go
```diff
@@ -79,13 +81,60 @@ func New(opts Options) (*RouteServer, error) {
...
+		oauthConfig.ClientSecretFile = opts.OAuth2ClientSecretFile
+		oauthConfig.CallbackPath = opts.OAuth2CallbackPath
+		oauthConfig.AuthURLParameters = opts.OAuth2AuthURLParameters
+		oauthConfig.SecretsProvider = grantSecrets
+		oauthConfig.Secrets = opts.SecretsRegistry
+		oauthConfig.AccessTokenHeaderName = opts.OAuth2AccessTokenHeaderName
+		oauthConfig.TokeninfoSubjectKey = opts.OAuth2TokeninfoSubjectKey
+		oauthConfig.TokenCookieName = opts.OAuth2TokenCookieName
+		oauthConfig.ConnectionTimeout = opts.OAuthTokeninfoTimeout
+		oauthConfig.Tracer = tracer
```
AlexanderYastrebov, line 120: Do we need all of this?
I think grant pre-processor only needs `OAuth2CallbackPath`

MustafaSaber: yeah we don't need all of this, I thought we need to call `Init` which is not true here.

AlexanderYastrebov:
:+1:

szuecs:
There is a missing test for append default filters, which should be added.

routesrv/options.go
```diff
@@ -114,10 +115,23 @@ type Options struct {
...
 
+	// DefaultFilters enables appending/prepending filters to all routes
+	DefaultFilters *eskip.DefaultFilters
+
 	// OriginMarker is *deprecated* and not used anymore. It will be deleted in v1.
 	OriginMarker bool
 
+	// List of custom filter specifications.
+	CustomFilters []filters.Spec
+
```
AlexanderYastrebov, line 126: Is this used anywhere?

AlexanderYastrebov:
:+1:

routesrv/options.go
MustafaSaber: no not yet

PR #2086, opened 2022-09-23
fix: add test with skipper, kubernetes and redis autoscaling updates

szuecs:
test: add test with skipper, kubernetes and redis autoscaling updates via kubernetes
test: enable changing redis update interval for testing purposes refactor: expose Endpoint and EndpointList
refactor: vegetaAttacker into httptest client
feature: redis can listen on any port add test conditions

### DIFF
--- a/dataclients/kubernetes/clusterstate.go
+++ b/dataclients/kubernetes/clusterstate.go
@@ -22,6 +22,8 @@ type clusterState struct {
 }
 
 func (state *clusterState) getService(namespace, name string) (*service, error) {
+	state.mu.Lock()
+	defer state.mu.Unlock()
 	s, ok := state.services[newResourceID(namespace, name)]
 	if !ok {
 		return nil, errServiceNotFound
@@ -36,6 +38,8 @@ func (state *clusterState) getService(namespace, name string) (*service, error) 
 }
 
 func (state *clusterState) getServiceRG(namespace, name string) (*service, error) {
+	state.mu.Lock()
+	defer state.mu.Unlock()
 	s, ok := state.services[newResourceID(namespace, name)]
 	if !ok {
 		return nil, fmt.Errorf("service not found: %s/%s", namespace, name)
@@ -51,6 +55,8 @@ func (state *clusterState) GetEndpointsByService(namespace, name, protocol strin
 		TargetPort: servicePort.TargetPort.String(),
 	}
 
+	state.mu.Lock()
+	defer state.mu.Unlock()
 	if cached, ok := state.cachedEndpoints[epID]; ok {
 		return cached
 	}
--- a/dataclients/kubernetes/clusterstate.go
+++ b/dataclients/kubernetes/clusterstate.go
@@ -22,8 +22,6 @@ type clusterState struct {
 }
 
 func (state *clusterState) getService(namespace, name string) (*service, error) {
-	state.mu.Lock()
-	defer state.mu.Unlock()
 	s, ok := state.services[newResourceID(namespace, name)]
 	if !ok {
 		return nil, errServiceNotFound
@@ -38,8 +36,6 @@ func (state *clusterState) getService(namespace, name string) (*service, error) 
 }
 
 func (state *clusterState) getServiceRG(namespace, name string) (*service, error) {
-	state.mu.Lock()
-	defer state.mu.Unlock()
 	s, ok := state.services[newResourceID(namespace, name)]
 	if !ok {
 		return nil, fmt.Errorf("service not found: %s/%s", namespace, name)
@@ -55,8 +51,6 @@ func (state *clusterState) GetEndpointsByService(namespace, name, protocol strin
 		TargetPort: servicePort.TargetPort.String(),
 	}
 
-	state.mu.Lock()
-	defer state.mu.Unlock()
 	if cached, ok := state.cachedEndpoints[epID]; ok {
 		return cached
 	}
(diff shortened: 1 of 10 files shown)


szuecs:
waits for https://github.com/zalando/skipper/pull/2080

dataclients/kubernetes/clusterstate.go
```diff
@@ -55,6 +51,26 @@ func (state *clusterState) GetEndpointsByService(namespace, name, protocol strin
...
+	targets := ep.targetsByServicePort(protocol, servicePort)
+	sort.Strings(targets)
+	state.cachedEndpoints[epID] = targets
+	return targets
+}
+
+func (state *clusterState) GetEndpointsByName(namespace, name, protocol string) []string {
+	epID := endpointID{
+		ResourceID: newResourceID(namespace, name),
+		Protocol:   protocol,
```
RomanZavodskikh, line 72: Why can we avoid initializing Port here?

RomanZavodskikh, line 72: Why can we just avoid initializing Port here?

dataclients/kubernetes/ingressdefinitions.go
```diff
@@ -138,6 +139,20 @@ func (ep endpoint) targetsByServiceTarget(protocol string, serviceTarget *defini
 	return nil
 }
 
+func (ep Endpoint) targets(protocol string) []string {
+	result := make([]string, 0)
+	for _, s := range ep.Subsets {
+		for _, p := range s.Ports {
+			for _, a := range s.Addresses {
+				result = append(result, formatEndpoint(a, p, protocol))
```
RomanZavodskikh, line 147: So, would we append to result every possible addr-port pair even if there is no opened port with corresponding number on corresponding address?

dataclients/kubernetes/kube.go
```diff
@@ -363,11 +364,14 @@ func mapRoutes(r []*eskip.Route) map[string]*eskip.Route {
 }
 
 func (c *Client) loadAndConvert() ([]*eskip.Route, error) {
+	c.mu.Lock()
```
RomanZavodskikh, line 367: Why not defer?

szuecs: Because the function does too much and I want to reduce holding the lock

dataclients/kubernetes/ingressdefinitions.go
szuecs: yes if we would have 10 redis pods with 2 ports we would have 20 endpoints, on the other hand I would expect that redis is only having one port in the service/endpoint and for testing this is more handy, because with testcontainers you get a random port and always localhost as ip, so we need to be able to have all the ports for testing purposes.

dataclients/kubernetes/clusterstate.go
szuecs: I don't see port here

szuecs: I don't see port here

RomanZavodskikh:
```
type endpointID struct {
	definitions.ResourceID
	TargetPort string
	Protocol   string
}
```
I see TargetPort here. However, I don't see it initialized here.

szuecs: This func will do Endpoint lookup by name and namespace which identifies the kubernetes object.
TargetPort makes no sense in case we want all ports and not one specific port. For tests we use random ports on 127.0.0.1 so we can only listen on multiple ports, so targetport makes no sense in this case.

dataclients/kubernetes/ingressdefinitions.go
```diff
@@ -74,13 +75,13 @@ func (s service) getTargetPortByValue(p int) (*definitions.BackendPort, bool) {
 	return nil, false
 }
 
-type endpoint struct {
+type Endpoint struct {
```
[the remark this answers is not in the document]
RomanZavodskikh: Discussed with @szuecs , he told me he can roll back this change.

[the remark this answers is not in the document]
szuecs: fixed

RomanZavodskikh:
:+1:

szuecs:
:+1:

dataclients/kubernetes/clusterstate.go
```diff
@@ -22,8 +22,6 @@ type clusterState struct {
 }
 
 func (state *clusterState) getService(namespace, name string) (*service, error) {
-	state.mu.Lock()
-	defer state.mu.Unlock()
```
AlexanderYastrebov, line 26: Why is this removed if we did not land https://github.com/zalando/skipper/pull/2080

Lets wait for it maybe?

dataclients/kubernetes/clusterstate.go
```diff
@@ -55,6 +51,26 @@ func (state *clusterState) GetEndpointsByService(namespace, name, protocol strin
...
+	}
+
+	ep, ok := state.endpoints[epID.ResourceID]
+	if !ok {
+		return nil
+	}
+
+	targets := ep.targetsByServicePort(protocol, servicePort)
+	sort.Strings(targets)
+	state.cachedEndpoints[epID] = targets
```
AlexanderYastrebov, line 65: This does not look correct, either mutex should be used or cache should not be updated

dataclients/kubernetes/clusterstate.go
szuecs: ok

dataclients/kubernetes/clusterstate.go
szuecs: fixed

szuecs:
:+1:

AlexanderYastrebov:
:+1:

Issue #2081, opened 2022-09-22
fadeIn filter support for consistentHash load balancer


herojan:
**Is your feature request related to a problem? Please describe.**
The consistentHash load balancing algorithm does not have support for the fadeIn filter. This means new pods immediately get as much traffic as old pods, which can cause overload. 

**Describe the solution you would like**
I would like to gradually add traffic to new nodes with the fadeIn filter. The logic would be the same as the existing fadeIn filter, except instead of choosing a random server when a new pod is fading in, it selects a new pod from the hash ring excluding all fading pods as options.

**Describe alternatives you've considered (optional)**
We have some other ways to potentially fix this, e.g. by filling our cache and warming up with a k8s startUp probe.

**Would you like to work on it?**
Yes, but it may not be important short term. Making this issue to see if the skipper team has any input on the idea

PR #2087, opened 2022-09-24
Implement fadeIn for consistent hash algorithm

herojan:
closes https://github.com/zalando/skipper/issues/2081

Leaves the same implementation for random and roundRobin

Logic for consistentHash when a chosen endpoint is fading and it is decided not to send traffic to it:
    - If all endpoints are fading, return original choice from consistent hash
    - Else run consistent hash again, but only using non-fading endpoints

### DIFF
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -202,7 +202,10 @@ type (
 		hash  uint64 // hash of endpoint
 	}
 	consistentHash struct {
-		hashRing []endpointHash // list of endpoints sorted by hash value
+		hashRing         []endpointHash // list of endpoints sorted by hash value
+		rand             *rand.Rand
+		notFadingIndexes []int
+		fadingWeights    []float64
 	}
 )
 
@@ -214,7 +217,10 @@ func (ch consistentHash) Swap(i, j int) {
 
 func newConsistentHashInternal(endpoints []string, hashesPerEndpoint int) routing.LBAlgorithm {
 	ch := consistentHash{
-		hashRing: make([]endpointHash, hashesPerEndpoint*len(endpoints)),
+		hashRing:         make([]endpointHash, hashesPerEndpoint*len(endpoints)),
+		rand:             rand.New(rand.NewSource(time.Now().UnixNano())),
+		notFadingIndexes: make([]int, 0, len(endpoints)),
+		fadingWeights:    make([]float64, 0, len(endpoints)),
 	}
 	for i, ep := range endpoints {
 		endpointStartIndex := hashesPerEndpoint * i
@@ -297,7 +303,11 @@ func (ch consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 		choice = ch.boundedLoadSearch(key, balanceFactor, ctx)
 	}
 
-	return ctx.Route.LBEndpoints[choice]
+	if ctx.Route.LBFadeInDuration <= 0 {
+		return ctx.Route.LBEndpoints[choice]
+	}
+
+	return withFadeIn(ch.rand, ctx, ch.notFadingIndexes, ch.fadingWeights, choice)
 }
 
 type powerOfRandomNChoices struct {
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -201,19 +201,25 @@ type (
 		index int    // index of endpoint in endpoint list
 		hash  uint64 // hash of endpoint
 	}
-	consistentHash []endpointHash // list of endpoints sorted by hash value
+	consistentHash struct {
+		hashRing []endpointHash // list of endpoints sorted by hash value
+	}
 )
 
-func (ch consistentHash) Len() int           { return len(ch) }
-func (ch consistentHash) Less(i, j int) bool { return ch[i].hash < ch[j].hash }
-func (ch consistentHash) Swap(i, j int)      { ch[i], ch[j] = ch[j], ch[i] }
+func (ch consistentHash) Len() int           { return len(ch.hashRing) }
+func (ch consistentHash) Less(i, j int) bool { return ch.hashRing[i].hash < ch.hashRing[j].hash }
+func (ch consistentHash) Swap(i, j int) {
+	ch.hashRing[i], ch.hashRing[j] = ch.hashRing[j], ch.hashRing[i]
+}
 
 func newConsistentHashInternal(endpoints []string, hashesPerEndpoint int) routing.LBAlgorithm {
-	ch := consistentHash(make([]endpointHash, hashesPerEndpoint*len(endpoints)))
+	ch := consistentHash{
+		hashRing: make([]endpointHash, hashesPerEndpoint*len(endpoints)),
+	}
 	for i, ep := range endpoints {
 		endpointStartIndex := hashesPerEndpoint * i
 		for j := 0; j < hashesPerEndpoint; j++ {
-			ch[endpointStartIndex+j] = endpointHash{i, hash(fmt.Sprintf("%s-%d", ep, j))}
+			ch.hashRing[endpointStartIndex+j] = endpointHash{i, hash(fmt.Sprintf("%s-%d", ep, j))}
 		}
 	}
 	sort.Sort(ch)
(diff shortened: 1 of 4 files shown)


cjbooms:
> when a chosen endpoint is fading

Is `fading` some ternary state on an endpoint:
FADING_DISABLED
FADING_ENABLED_ACCEPTING
FADING_ENABLED_REJECTING

herojan:
> > when a chosen endpoint is fading
> 
> Is `fading` some ternary state on an endpoint? FADING_DISABLED FADING_ENABLED_ACCEPTING FADING_ENABLED_REJECTING

No it's a percentage chance of directing traffic to a selected endpoint based on the fadeIn configuration and the age of the pod endpoint. This percentage increases the older the endpoint gets, until the pod age reaches the fadeIn duration config.

cjbooms:
> > > when a chosen endpoint is fading
> > 
> > 
> > Is `fading` some ternary state on an endpoint? FADING_DISABLED FADING_ENABLED_ACCEPTING FADING_ENABLED_REJECTING
> 
> An endpoint is either fading or not fading, and it's fading if its age is less than what is configured in the fadeIn filter.

But when fading, should an endpoint not receive some subset of the traffic? From your description it sounds like it won't get any traffic until the fadeIn period is over

herojan:
> > > > when a chosen endpoint is fading
> > > 
> > > 
> > > Is `fading` some ternary state on an endpoint? FADING_DISABLED FADING_ENABLED_ACCEPTING FADING_ENABLED_REJECTING
> > 
> > 
> > An endpoint is either fading or not fading, and it's fading if its age is less than what is configured in the fadeIn filter.
> 
> But when fading, should an endpoint not receive some subset of the traffic? From your description it sounds like it won't get any traffic until the fadeIn period is over

So the "fading" state is binary, and depends on endpoint age. But when a selected endpoint is fading, then a % chance of sending traffic to it is calculated based on its age and the fadeIn filter's configuration. By default the increase is linear, but you can also [configure it to be exponential](https://opensource.zalando.com/skipper/reference/filters/#fadein). So at age 0 there's 0% chance of sending the traffic to it, and when it gets to e.g. a configured 5m fadeIn, then it will be 100%. Otherwise it selects another endpoint based on the logic in the PR description.

cjbooms:
> > > > > when a chosen endpoint is fading
> > > > 
> > > > 
> > > > Is `fading` some ternary state on an endpoint? FADING_DISABLED FADING_ENABLED_ACCEPTING FADING_ENABLED_REJECTING
> > > 
> > > 
> > > An endpoint is either fading or not fading, and it's fading if its age is less than what is configured in the fadeIn filter.
> > 
> > 
> > But when fading, should an endpoint not receive some subset of the traffic? From your description it sounds like it won't get any traffic until the fadeIn period is over
> 
> So the "fading" state is binary, and depends on endpoint age. But when a selected endpoint is fading, then a % chance of sending traffic to it is calculated based on its age and the fadeIn filter's configuration. By default the increase is linear, but you can also [configure it to be exponential](https://opensource.zalando.com/skipper/reference/filters/#fadein). So at age 0 there's 0% chance of sending the traffic to it, and when it gets to e.g. a configured 5m fadeIn, then it will be 100%. Otherwise it selects another endpoint based on the logic in the PR description.

Sorry it is still not clear, possibly it's just because the term fading is being overloaded. I understand there are three separate states, excuse pseudo code:
```
val isFadeInFilterEnabled: Boolean // route configuration
val isEndpointInFadingPeriod: Boolean // fading state
val isFadingEndpointAcceptingRequest: Boolean // time-based percentage evaluation result
```

herojan:
If the fadeIn filter is on the route and the endpoint is younger than the fadeIn period, then it is considered fading. Otherwise it is not. Fading endpoints get a % of traffic which increases linearly (by default) with the endpoint's age until the fading period is done.

loadbalancer/algorithm.go
```diff
@@ -121,10 +117,29 @@ func withFadeIn(rnd *rand.Rand, ctx *routing.LBContext, wi []int, wf []float64,
...
 
-	return shiftToRemaining(rnd, ctx, wi, wf, now)
+	switch a := algo.(type) {
+	case *roundRobin:
+		return shiftToRemaining(a.rnd, ctx, notFadingIndexes, a.fadingWeights, now)
+	case *random:
+		return shiftToRemaining(a.rand, ctx, notFadingIndexes, a.fadingWeights, now)
+	case consistentHash:
+		// If all endpoints are fading, normal consistent hash result
+		if len(notFadingIndexes) == 0 {
```
szuecs, line 135: should not this be the default for all algorithms?

loadbalancer/algorithm.go
```diff
@@ -121,10 +117,29 @@ func withFadeIn(rnd *rand.Rand, ctx *routing.LBContext, wi []int, wf []float64,
...
+		}
 	}
 
-	return shiftToRemaining(rnd, ctx, wi, wf, now)
+	switch a := algo.(type) {
+	case *roundRobin:
+		return shiftToRemaining(a.rnd, ctx, notFadingIndexes, a.fadingWeights, now)
+	case *random:
+		return shiftToRemaining(a.rand, ctx, notFadingIndexes, a.fadingWeights, now)
+	case consistentHash:
```
szuecs, line 133: why this is no pointer but the others are?
I think all of these should have a similar type, so either ptr or value.

loadbalancer/algorithm.go
```diff
@@ -254,13 +279,17 @@ func computeLoadAverage(ctx *routing.LBContext) float64 {
 }
 
 // Returns index of endpoint with closest hash to key's hash, which is also below the target load
-func (ch consistentHash) boundedLoadSearch(key string, balanceFactor float64, ctx *routing.LBContext) int {
-	ringIndex := ch.searchRing(key)
+// endpointIsEligible function is used to skip endpoints we don't want, such as fading endpoints
```
szuecs, line 282: maybe rename `endpointIsEligible` to `skipEndpoint`

szuecs:
from my side lgtm, some comments, but I am also fine like this.
I guess the receiver should be changed from value to ptr receiver in order to waste less resources and to be similar to the others.
At least in the switch block looks really ugly to have ptr vs value cases.

loadbalancer/algorithm.go
herojan: Ah I just did it like this because the random and roundRobin algorithms implemented Apply as pointers and consistentHash didn't. I can change it so consistentHash uses pointers instead though :+1:

loadbalancer/algorithm.go
herojan: Grand that would be clearer :+1:

loadbalancer/algorithm.go
herojan: Currently, it looks like random and roundRobin randomly select an endpoint, but weighted by how old the endpoints are so that older endpoints are more likely to be selected. For fading with consistentHash, I was thinking to just do normal consistent hash without caring how old the endpoints were

szuecs: yes makes sense

szuecs:
:+1:

szuecs:
:+1:

AlexanderYastrebov:
@herojan Thank you for the contribution, I am still looking into the change as this part of Skipper is quite complex.

As far as I understand you'd like to keep consistent hash algo logic but skip fading endpoints.
If so, would it be possible to extend search comparator function instead such that it excludes fading endpoints?
https://github.com/zalando/skipper/blob/af67775fb49b52d045bed11d36e9411d5d93c6da/loadbalancer/algorithm.go#L231-L239

Something like
```go
func(i int) bool
{
    return ch[i].hash >= h && notFadingIn(i) // may require ref to the endpoints but its doable
}
```
instead of `func(i int) bool { return ch[i].hash >= h }`...

herojan:
> As far as I understand you'd like to keep consistent hash algo logic but skip fading endpoints.
> If so, would it be possible to extend search comparator function instead such that it excludes fading endpoints?

We do want to do that, but only when it is decided not to send traffic to a fading endpoint and we are doing the fallback endpoint search. For this, there is a change as you describe where we modify the `searchRing` function's comparator to include  the result of the `skipEndpoint` function. There are two functions which can be passed in as this `skipEndpoint` function, one which skips no endpoints (default) and one which skips fading endpoints, used when finding a fallback.

AlexanderYastrebov:
:+1:

PR #2088, opened 2022-09-27
Add an option to override the routing via services instead of endpoints

peterklijn:
PR relating to issue https://github.com/zalando/skipper/issues/2089

### DIFF
--- a/dataclients/kubernetes/ingressv1.go
+++ b/dataclients/kubernetes/ingressv1.go
@@ -40,7 +40,7 @@ func convertPathRuleV1(
 	prule *definitions.PathRuleV1,
 	pathMode PathMode,
 	allowedExternalNames []*regexp.Regexp,
-	routeTrafficUsingServices bool,
+	forceKubernetesService bool,
 ) (*eskip.Route, error) {
 
 	ns := metadata.Namespace
@@ -79,7 +79,7 @@ func convertPathRuleV1(
 		}
 	} else if svc.Spec.Type == "ExternalName" {
 		return externalNameRoute(ns, name, host, hostRegexp, svc, servicePort, allowedExternalNames)
-	} else if routeTrafficUsingServices {
+	} else if forceKubernetesService {
 		return &eskip.Route{
 			Id:          routeID(ns, name, host, prule.Path, svcName),
 			BackendType: eskip.NetworkBackend,
@@ -137,7 +137,6 @@ func convertPathRuleV1(
 
 func (ing *ingress) addEndpointsRuleV1(ic ingressContext, host string, prule *definitions.PathRuleV1) error {
 	meta := ic.ingressV1.Metadata
-
 	endpointsRoute, err := convertPathRuleV1(
 		ic.state,
 		meta,
@@ -145,7 +144,7 @@ func (ing *ingress) addEndpointsRuleV1(ic ingressContext, host string, prule *de
 		prule,
 		ic.pathMode,
 		ing.allowedExternalNames,
-		ing.routeTrafficUsingServices,
+		ing.forceKubernetesService,
 	)
 	if err != nil {
 		// if the service is not found the route should be removed
@@ -342,7 +341,7 @@ func (ing *ingress) addSpecIngressTLSV1(ic ingressContext, ingtls *definitions.T
 func (ing *ingress) convertDefaultBackendV1(
 	state *clusterState,
 	i *definitions.IngressV1Item,
-	routeTrafficUsingServices bool,
+	forceKubernetesService bool,
 ) (*eskip.Route, bool, error) {
 	// the usage of the default backend depends on what we want
 	// we can generate a hostname out of it based on shared rules
@@ -377,7 +376,7 @@ func (ing *ingress) convertDefaultBackendV1(
 	} else if svc.Spec.Type == "ExternalName" {
 		r, err := externalNameRoute(ns, name, "default", nil, svc, servicePort, ing.allowedExternalNames)
 		return r, err == nil, err
-	} else if routeTrafficUsingServices {
+	} else if forceKubernetesService {
 		return &eskip.Route{
 			Id:          routeID(ns, name, "", "", ""),
 			BackendType: eskip.NetworkBackend,
@@ -463,7 +462,7 @@ func (ing *ingress) ingressV1Route(
 	}
 
 	var route *eskip.Route
-	if r, ok, err := ing.convertDefaultBackendV1(state, i, ing.routeTrafficUsingServices); ok {
+	if r, ok, err := ing.convertDefaultBackendV1(state, i, ing.forceKubernetesService); ok {
 		route = r
 	} else if err != nil {
 		ic.logger.Errorf("error while converting default backend: %v", err)
--- a/dataclients/kubernetes/ingressv1.go
+++ b/dataclients/kubernetes/ingressv1.go
@@ -40,6 +40,7 @@ func convertPathRuleV1(
 	prule *definitions.PathRuleV1,
 	pathMode PathMode,
 	allowedExternalNames []*regexp.Regexp,
+	routeTrafficUsingServices bool,
 ) (*eskip.Route, error) {
 
 	ns := metadata.Namespace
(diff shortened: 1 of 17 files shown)


routesrv/options.go
```diff
@@ -100,6 +100,10 @@ type Options struct {
 	// used with external name services (type=ExternalName).
 	KubernetesAllowedExternalNames []*regexp.Regexp
 
+	// KubernetesForceService overrides the default Skipper functionality to route traffic using Kubernetes Endpoints,
+	// instead using Kubernetes Services.
+	KubernetesForceService bool
```
peterklijn, line 105: I chose to deviate from the `ForceKubernetesService` naming here, as this is not the `Options` struct relating to Kubernetes configuration, and the Kubernetes prefix seems more consistent with the other fields.

szuecs: Yes very good!
Can you add some docs into docs/kubernetes/ingress-controller.md to let people understand use case and how to configure?
I think code and tests are good to go.

szuecs:
:+1:

dataclients/kubernetes/ingressv1.go
```diff
@@ -78,6 +79,17 @@ func convertPathRuleV1(
...
+		r := &eskip.Route{
+			Id:          routeID(ns, name, host, prule.Path, svcName),
+			BackendType: eskip.NetworkBackend,
+			Backend:     serviceNameBackend(svcName, ns, servicePort),
+			HostRegexps: hostRegexp,
+		}
+
+		setPathV1(pathMode, r, prule.PathType, prule.Path)
+		setTraffic(r, svcName, prule.Backend.Traffic, prule.Backend.NoopCount)
+		return r, nil
```
AlexanderYastrebov, line 92: This looks the same as a special case `if len(eps) == 1` below. 
Wouldn't it be better to just have `eps = []string{serviceNameBackend(svcName, ns, servicePort)}` here?

docs/kubernetes/ingress-controller.md
```diff
@@ -163,6 +163,23 @@ TLS handshake error to the endpoint.  Other reasons are future support
...
+
+An example of scenario where you might need to use services is when you rely
+on Istio networking features to connect multiple clusters, as the IPs of
+Kubernetes endpoints will not resolve in all cases.
+
+If you find yourself in this category, you can override the default behaviour
+by setting the `KubernetesForceService` flag to `true` in the `Skipper.Options` struct.
+This will cause skipper to create routes with `BackendType=eskip.NetworkBackend` instead
+of `BackendType=eskip.LBBackend` and use the following address format:
+`http://<service name>.<namespace>.svc.cluster.local:<port>`.
```
AlexanderYastrebov, line 181: Maybe a link to the https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#services could be useful here

docs/kubernetes/ingress-controller.md
```diff
@@ -163,6 +163,23 @@ TLS handshake error to the endpoint.  Other reasons are future support
 of features like session affinity, different load balancer
 algorithms or distributed loadbalancing also known as service mesh.
 
+### Using services instead of endpoints
```
AlexanderYastrebov, line 166: minor: `Using Service instead of Endpoints` (capital letters)

docs/kubernetes/ingress-controller.md
AlexanderYastrebov: Also `cluster.local` is a default and might be different for some clusters. 

https://stackoverflow.com/questions/59559438/retrieve-the-full-name-of-a-service-in-kubernetes claims kubernetes would automatically set up the DNS search domains so maybe we could get away with `<service name>.<namespace>.svc` that would nto depend on cluster domain

@szuecs WDYT?

routesrv/options.go
AlexanderYastrebov: Maybe we could use `KubernetesUseServiceDomainName` to avoid negative connotation of "Force"?

docs/kubernetes/ingress-controller.md
```diff
@@ -163,6 +163,23 @@ TLS handshake error to the endpoint.  Other reasons are future support
...
+edge cases that require the use of [Kubernetes
+Services](http://kubernetes.io/docs/user-guide/services) instead.
+
+An example of scenario where you might need to use services is when you rely
+on Istio networking features to connect multiple clusters, as the IPs of
+Kubernetes endpoints will not resolve in all cases.
+
+If you find yourself in this category, you can override the default behaviour
+by setting the `KubernetesForceService` flag to `true` in the `Skipper.Options` struct.
+This will cause skipper to create routes with `BackendType=eskip.NetworkBackend` instead
```
AlexanderYastrebov, line 179: `Skipper` is a proper noun and should be written from capital

dataclients/kubernetes/ingressv1.go
AlexanderYastrebov: I guess we'd also need to have it for RouteGroups around here https://github.com/zalando/skipper/blob/d3e8dbd82527da74ac3fb4f4a9868a840478b81f/dataclients/kubernetes/routegroup.go#L241

AlexanderYastrebov:
@peterklijn Thank you for this contribution, it is a valuable addition to the Skipper.

dataclients/kubernetes/ingressv1.go
AlexanderYastrebov: @szuecs Do we need this for the `RouteGroup` or maybe users should explicitly use `network` backend (then this needs to be mentioned in the docs I guess)?

docs/kubernetes/ingress-controller.md
peterklijn: Let me know what you prefer @AlexanderYastrebov & @szuecs :)

dataclients/kubernetes/ingressv1.go
AlexanderYastrebov: > I don't fully understand your comments around RouteGroups

RouteGroup is a Skipper CRD and could also use Service domain name but on the other hand users have much more control over backend configuration so they may use `network` backend and Service domain name explicitly therefore I think we do not need to support this new config there.

AlexanderYastrebov: @szuecs Do we want this for [ingressv1beta1](https://github.com/zalando/skipper/blob/37245e82ade1d1aab359443e73d5fc0f966ade56/dataclients/kubernetes/ingressv1beta1.go#L103-L104) for consistency? I know that we plan to drop it https://github.com/zalando/skipper/pull/1999

peterklijn: > @szuecs Do we want this for [ingressv1beta1](https://github.com/zalando/skipper/blob/37245e82ade1d1aab359443e73d5fc0f966ade56/dataclients/kubernetes/ingressv1beta1.go#L103-L104) for consistency? I know that we plan to drop it #1999

I purposely left it out for two reasons:
- I can't imagine this feature is interesting for active users (why switch to services if the current setup works)
- I assumed the group of users that would like to start using ingresses today (but can't due to the default usage of Endpoints) are still rocking an old Kubernetes version, V1 is out for quite a while already right?

szuecs: I think:
- RouteGroup can use a service with DNS via network backend. In case there is a need we can add it
- v1beta1 I would not touch and we should clean this up after our internal migration (1year deprecation seems to be also fine for everyone else)

docs/kubernetes/ingress-controller.md
szuecs: Always use `<service name>.<namespace>.svc.cluster.local` the other depends too much on cluster internal DNS setup. You can also make sure that search domains are not propagated and honestly always use long names for generated stuff, it will always be less error prone.

dataclients/kubernetes/ingressv1.go
```diff
@@ -406,6 +413,14 @@ func (ing *ingress) convertDefaultBackendV1(
 	}, true, nil
 }
 
+func serviceNameBackend(svcName, svcNamespace string, servicePort *servicePort) string {
+	scheme := "https"
+	if n, _ := servicePort.TargetPort.Number(); n != 443 {
+		scheme = "http"
+	}
+	return fmt.Sprintf("%s://%s.%s.svc:%s", scheme, svcName, svcNamespace, servicePort.TargetPort)
```
szuecs, line 421: please add `.cluster.local` suffix

routesrv/options.go
szuecs: I prefer Force, because it's more clear that you override the default choice.

szuecs:
The .cluster.local should be fixed, everything else is fine

docs/kubernetes/ingress-controller.md
peterklijn: I undid the change, it's back to `<service name>.<namespace>.svc.cluster.local`

dataclients/kubernetes/ingressv1.go
peterklijn: Added it back 👍

peterklijn:
Sorry about the requesting review spam, I don't know why it keeps un-assigning you 🤷‍♂️

AlexanderYastrebov:
:+1:

szuecs:
:+1:

szuecs:
@peterklijn In general here you don't need to assign anyone. I do this to assign another maintainer when I think something is ready for review. We see all PRs, sometimes we just have a bit overload and can't reply to everyone in time.

PR #2093, opened 2022-10-11
add option to use Etag for /routes which serves eskip-formatted routes

nikhil88877:
relating to issue #2107

### DIFF
--- a/routesrv/eskipbytes.go
+++ b/routesrv/eskipbytes.go
@@ -2,8 +2,11 @@ package routesrv
 
 import (
 	"bytes"
+	"crypto/sha256"
+	"fmt"
 	"net/http"
 	"sync"
+	"time"
 
 	ot "github.com/opentracing/opentracing-go"
 	"github.com/zalando/skipper/eskip"
@@ -14,20 +17,22 @@ import (
 // provides synchronized r/w access to them. Additionally it can
 // serve as an HTTP handler exposing its content.
 type eskipBytes struct {
-	data        []byte
-	initialized bool
-	mu          sync.RWMutex
+	data         []byte
+	etag         string
+	lastModified time.Time
+	initialized  bool
+	mu           sync.RWMutex
 
 	tracer ot.Tracer
 }
 
 // bytes returns a slice to stored bytes, which are safe for reading,
 // and if there were already initialized.
-func (e *eskipBytes) bytes() ([]byte, bool) {
+func (e *eskipBytes) bytes() ([]byte, string, time.Time, bool) {
 	e.mu.RLock()
 	defer e.mu.RUnlock()
 
-	return e.data, e.initialized
+	return e.data, e.etag, e.lastModified, e.initialized
 }
 
 // formatAndSet takes a slice of routes and stores them eskip-formatted
@@ -39,7 +44,13 @@ func (e *eskipBytes) formatAndSet(routes []*eskip.Route) (int, bool) {
 
 	e.mu.Lock()
 	defer e.mu.Unlock()
-	e.data = buf.Bytes()
+	if updated := buf.Bytes(); !bytes.Equal(e.data, updated) {
+		e.lastModified = time.Now()
+		e.data = updated
+		h := sha256.New()
+		h.Write(e.data)
+		e.etag = fmt.Sprintf(`"%x"`, h.Sum(nil))
+	}
 	oldInitialized := e.initialized
 	e.initialized = true
 
@@ -50,8 +61,11 @@ func (e *eskipBytes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	span := tracing.CreateSpan("serve_routes", r.Context(), e.tracer)
 	defer span.Finish()
 
-	if data, initialized := e.bytes(); initialized {
-		w.Write(data)
+	data, etag, lastModified, initialized := e.bytes()
+	if initialized {
+		w.Header().Add("Etag", etag)
+		w.Header().Add("Content-Type", "text/plain; charset=utf-8")
+		http.ServeContent(w, r, "", lastModified, bytes.NewReader(data))
 	} else {
 		w.WriteHeader(http.StatusNotFound)
 	}
@@ -66,7 +80,7 @@ type eskipBytesStatus struct {
 const msgRoutesNotInitialized = "routes were not initialized yet"
 
 func (s *eskipBytesStatus) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if _, initialized := s.b.bytes(); initialized {
+	if _, _, _, initialized := s.b.bytes(); initialized {
 		w.WriteHeader(http.StatusNoContent)
 	} else {
 		http.Error(w, msgRoutesNotInitialized, http.StatusServiceUnavailable)
--- a/routesrv/polling.go
+++ b/routesrv/polling.go
@@ -3,6 +3,7 @@ package routesrv
 import (
 	"context"
 	"fmt"
+	"sort"
 	"sync"
 	"time"
 
@@ -98,6 +99,10 @@ func (p *poller) poll(wg *sync.WaitGroup) {
 				"message", msg,
 			)
 		case routesCount > 0:
+			// sort the routes, otherwise it will lead to different etag values for the same route list for different orders
+			sort.SliceStable(routes, func(i, j int) bool {
+				return routes[i].Id < routes[j].Id
+			})
 			routesBytes, initialized = p.b.formatAndSet(routes)
 			logger := log.WithFields(log.Fields{"count": routesCount, "bytes": routesBytes})
 			if initialized {
--- a/routesrv/routesrv_test.go
+++ b/routesrv/routesrv_test.go
@@ -102,6 +102,17 @@ func getRoutes(rs *routesrv.RouteServer) *httptest.ResponseRecorder {
 	return w
 }
 
+func getRoutesWithRequestHeadersSetting(rs *routesrv.RouteServer, header map[string]string) *httptest.ResponseRecorder {
+	w := httptest.NewRecorder()
+	r := httptest.NewRequest("GET", "/routes", nil)
+	for k, v := range header {
+		r.Header.Add(k, v)
+	}
+	rs.ServeHTTP(w, r)
+
+	return w
+}
+
 func wantHTTPCode(t *testing.T, w *httptest.ResponseRecorder, want int) {
 	got := w.Code
 	if got != want {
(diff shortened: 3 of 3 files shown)


szuecs:
@nikhil88877 please sign off commit: `git commit -sm "my message"` (-s is signoff), you can rebase sign-off or amend to do it, too.

routesrv/eskipbytes.go
```diff
@@ -15,8 +17,10 @@ import (
 // serve as an HTTP handler exposing its content.
 type eskipBytes struct {
 	data        []byte
+	hash        string
 	initialized bool
 	mu          sync.RWMutex
+	useEtag     bool
```
AlexanderYastrebov, line 23: Do we need this to be configurable? Maybe we should always calculate etag and leave it up to the client to use it if they like.

routesrv/eskipbytes.go
```diff
@@ -39,6 +43,11 @@ func (e *eskipBytes) formatAndSet(routes []*eskip.Route) (int, bool) {
 
 	e.mu.Lock()
 	defer e.mu.Unlock()
+	if e.useEtag && !bytes.Equal(e.data, buf.Bytes()) {
+			e.data = buf.Bytes()
```
AlexanderYastrebov, line 47: This looks redundant as we do it right outside of the if block

routesrv/eskipbytes.go
```diff
@@ -15,8 +17,10 @@ import (
 // serve as an HTTP handler exposing its content.
 type eskipBytes struct {
 	data        []byte
+	hash        string
```
AlexanderYastrebov, line 20: Why not call it `etag` ?

routesrv/eskipbytes.go
```diff
@@ -50,6 +59,10 @@ func (e *eskipBytes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	span := tracing.CreateSpan("serve_routes", r.Context(), e.tracer)
 	defer span.Finish()
 
+	if e.useEtag && e.hash != "" && r.Header.Get("Etag") == e.hash {
+		w.WriteHeader(http.StatusNotModified)
+		return
+	}
 	if data, initialized := e.bytes(); initialized {
 		w.Write(data)
```
AlexanderYastrebov, line 67: I think we'd need to set Etag header around here.
Also Etag header should follow certain format, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag

routesrv/eskipbytes_test.go
```diff
@@ -0,0 +1,231 @@
...
+						{
+							Name: filters.RedirectToName,
+							Args: []interface{}{float64(308), "https:"},
+						},
+					},
+					Shunt:       true,
+					BackendType: eskip.ShuntBackend,
+				},
+			},
+			Etag:      strconv.FormatUint(uint64(crc32.ChecksumIEEE([]byte(`kube__smth: Header("X-Smth", "smth") -> redirectTo(308, "https:") -> <shunt>;`))), 10),
```
AlexanderYastrebov, line 200: Use of CRC is an implementation detail, tests should not rely on any specific value but rather test subsequent request/response with the proper Etag value returned earlier

routesrv/eskipbytes.go
```diff
@@ -50,6 +59,10 @@ func (e *eskipBytes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	span := tracing.CreateSpan("serve_routes", r.Context(), e.tracer)
 	defer span.Finish()
 
+	if e.useEtag && e.hash != "" && r.Header.Get("Etag") == e.hash {
```
AlexanderYastrebov, line 62: I think this should check `If-None-Match` header, lets verify with the spec https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag

AlexanderYastrebov: Lets check if can utilize https://pkg.go.dev/net/http#ServeContent as it should implement proper support of caching headers

nikhil88877:
Hi @AlexanderYastrebov and @szuecs , 

Thanks a lot for reviewing the PR so fast. I apologize for not being able to reply sooner. It is actually an incomplete PR, hence it was in draft state. But thanks again for the feedback. 

I was running a few CPU benchmark tests on `ETag` versus not using them. And, we noticed that even when `eskip.Routes` list(or the ingresses) had not changed, `/routes` endpoint was still sending the routes list. On debugging it further, I found there is a hostRoutes map defined [here](https://github.com/zalando/skipper/blob/3c2f1081d6285e829352037cd62eb25ddc01fddc/dataclients/kubernetes/ingress.go#L47). If list of ingresses has more than one host defined, then the order of routes will not be guaranteed.

I can think of 2 different ways to tackle this issue.
1. Define another slice `hosts []string`, which will maintain the order of the hosts according to `IngressV1List` and use this slice to loop [here](https://github.com/zalando/skipper/blob/3c2f1081d6285e829352037cd62eb25ddc01fddc/dataclients/kubernetes/ingress.go#L423)
2. Or, we let the routes be defined as it is and in our formatAndSet(), we changed the `ETag` only when we see any change in the routes list based on ID(order will not matter). I'm not sure if the order of routes is important(if routes[0] take more precedence over routes[1]), in that case, we cannot choose this option.

If you guys can think of a better way, please do share your thoughts. Or if you feel, I should create a new issue and we take the discussion there, let me know. I feel before I complete this PR with more tests and your feedback, I should first tackle the route order issue otherwise, `ETag` feature will be incorrect.

PS: I'm very new to the skipper library and this was my first chance at deep diving into the skipper code. Please don't mind if my suggestion sounds silly or stupid.

szuecs:
Hi!
Order of routes don't matter.
I would sort route keys in the http handler that serves the routes. You point to dataclient kubernetes which is likely irrelevant to the problem. Another dataclient could also return not sorted routes so we should sort them in one place.

nikhil88877:
Hi @szuecs ,
I think there is some confusion or I'm missing something here. 
You are suggesting to sort the routing table here https://github.com/zalando/skipper/blob/master/routing/routing.go#L306
But I'm adding `Etag` functionality for `routesrv`, and HTTP handler that serves the request is defined here https://github.com/zalando/skipper/blob/master/routesrv/eskipbytes.go#L49

Based on this understanding, I added the `routes sort` logic after https://github.com/zalando/skipper/blob/master/routesrv/polling.go#L101
You can see the changes in this PR. If you don't agree, please share feedback

routesrv/eskipbytes.go
```diff
@@ -15,6 +19,7 @@ import (
 // serve as an HTTP handler exposing its content.
 type eskipBytes struct {
 	data        []byte
+	etag        string
```
AlexanderYastrebov, line 22: Lets keep track of modification timestamp and pass it to `http.ServeContent` - it would set `Last-Modified` header which is handy.

AlexanderYastrebov, line 22: Lets keep track of modification timestamp and pass it to `http.ServeContent` - it would set `Last-Modified` header which is handy.

routesrv/eskipbytes.go
```diff
@@ -51,7 +59,8 @@ func (e *eskipBytes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	defer span.Finish()
 
 	if data, initialized := e.bytes(); initialized {
-		w.Write(data)
+		w.Header().Add("Etag", e.etag)
+		http.ServeContent(w, r, "", time.Time{}, bytes.NewReader(data))
```
AlexanderYastrebov, line 63: https://pkg.go.dev/net/http#ServeContent
> If the response's Content-Type header is not set, ServeContent first tries to deduce the type from name's file extension and, if that fails, falls back to reading the first block of the content and passing it to DetectContentType.

Lets set content type explicitly to disable detection overhead

routesrv/eskipbytes.go
```diff
@@ -39,7 +44,10 @@ func (e *eskipBytes) formatAndSet(routes []*eskip.Route) (int, bool) {
 
 	e.mu.Lock()
 	defer e.mu.Unlock()
-	e.data = buf.Bytes()
+	if !bytes.Equal(e.data, buf.Bytes()) {
+		e.data = buf.Bytes()
+		e.etag = fmt.Sprintf(`"%s"`, strconv.FormatUint(uint64(crc32.ChecksumIEEE(e.data)), 10))
```
AlexanderYastrebov, line 49: I think 
```go
e.etag = fmt.Sprintf(`"%d"`, crc32.ChecksumIEEE(e.data))
```
should do

AlexanderYastrebov: BTW what is the reason to use crc32 over e.g. md5 or some other cryptographic hash?
I think md5 is quite popular for the purpose:
- https://cloud.google.com/storage/docs/hashes-etags
- https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html

routesrv/eskipbytes.go
```diff
@@ -39,7 +44,10 @@ func (e *eskipBytes) formatAndSet(routes []*eskip.Route) (int, bool) {
 
 	e.mu.Lock()
 	defer e.mu.Unlock()
-	e.data = buf.Bytes()
+	if !bytes.Equal(e.data, buf.Bytes()) {
```
AlexanderYastrebov, line 47:
```go
if updated := buf.Bytes(); !bytes.Equal(e.data, updated) {
  e.data = updated
  ...
}
```

routesrv/eskipbytes.go
```diff
@@ -51,7 +59,8 @@ func (e *eskipBytes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	defer span.Finish()
 
 	if data, initialized := e.bytes(); initialized {
-		w.Write(data)
+		w.Header().Add("Etag", e.etag)
```
AlexanderYastrebov, line 62: There is a race condition as etag could be modified after `e.bytes()` returns.
Lets drop `(e *eskipBytes) bytes()` and manually lock and unlock the mutex.

AlexanderYastrebov, line 62: There is a race condition as etag could be modified after `e.bytes()` returns.
Lets drop `(e *eskipBytes) bytes()` and manually lock and unlock the mutex.

AlexanderYastrebov: Or simply make `e.bytes()` return etag as well.

AlexanderYastrebov: Or simply make `e.bytes()` return etag as well.

routesrv/eskipbytes.go
nikhil88877: We did some internal benchmarking and found `crc32` to be the fastest.
Here is the benchmarking stats
`BenchmarkDiff-16             357       3171983 ns/op     1434948 B/op      36495 allocs/op // MD5
BenchmarkDiffCrc-16          416       2913816 ns/op     1434699 B/op      36494 allocs/op // IEEE
BenchmarkDiffCrc-16          423       2839012 ns/op     1434719 B/op      36494 allocs/op // Castagnoli
BenchmarkDiffCrc-16          100       10309766 ns/op    2872003 B/op      37901 allocs/op // Koopman
BenchmarkDiffAdler-16        410       2923837 ns/op     1434849 B/op      36494 allocs/op // Adler
BenchmarkDiffFnv-16          392       2902272 ns/op     1434974 B/op      36495 allocs/op // Fnv
BenchmarkDiffMaphash-16      378       2944291 ns/op     1474711 B/op      36509 allocs/op // Maphash `

Hence, Castagnoli or crc32 seems to be the fastest.
But I don't mind switching to `md5`. Let me know your preference

routesrv/eskipbytes.go
nikhil88877: Even though the 2nd solution(`e.bytes()` return the ETag) seems more elegant, I think we will have to go with the first one for the following reason:

1. the Second solution will make ETag generation lazy i.e. at the HTTP handler level and make ETag generation dependent on the client making the request. My worry is if the client decided to call `/routes` at a very high RPS, It can cause some performance issues.
2. With your other suggestion about using `lastModified` for the eskipBytes, I'll have to introduce manual mutex locking and unlocking anyways. So, I'll not gain much using lazy ETag generation.

AlexanderYastrebov: > the Second solution will make ETag generation lazy i.e. at the HTTP handler level and make ETag generation dependent on the client making the request.

I meant 
```go
func (e *eskipBytes) bytes() ([]byte, string, bool) {
	e.mu.RLock()
	defer e.mu.RUnlock()
	return e.data, e.etag, e.initialized
}
```
It could also be extended to return `lastModified` timestamp.

nikhil88877: Sorry, I misunderstood at first. But I did make the change based on your suggestion.

routesrv/eskipbytes.go
```diff
@@ -66,7 +78,7 @@ type eskipBytesStatus struct {
 const msgRoutesNotInitialized = "routes were not initialized yet"
 
 func (s *eskipBytesStatus) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if _, initialized := s.b.bytes(); initialized {
+	if s.b.initialized {
```
nikhil88877, line 81: @AlexanderYastrebov I made this change as I couldn't see why there can be a race condition here. Please suggest if I'm wrong here

routesrv/eskipbytes.go
nikhil88877: Good suggestion. 
**Fun fact**: Did you know `If-Modified-Since`  and `If-Unmodified-Since`  is used for the same purpose (makes requests conditional) but unlike `If-Unmodified-Since`, `If-Modified-Since` can only be used with a `GET`  or a `HEAD`

routesrv/eskipbytes.go
```diff
@@ -15,6 +18,8 @@ import (
 // serve as an HTTP handler exposing its content.
 type eskipBytes struct {
 	data        []byte
+	etag        string
+	lastModified time.Time
```
AlexanderYastrebov, line 22: Please run `make fmt`

routesrv/eskipbytes.go
AlexanderYastrebov: > I couldn't see why there can be a race condition here

The race condition exists when two or more threads access data and at least one modifies it.
The datatype does not really matter (although for certain machines and data types concurrent access and modification may be atomic).
Here `initialized` is read while being modified by another thread (holding the mutex).

We could do it like so
```go
if _, initialized, _, _ := e.bytes(); initialized {
}
```

routesrv/eskipbytes.go
```diff
@@ -14,20 +17,22 @@ import (
...
 
 // bytes returns a slice to stored bytes, which are safe for reading,
 // and if there were already initialized.
-func (e *eskipBytes) bytes() ([]byte, bool) {
+func (e *eskipBytes) bytes() ([]byte, bool, string, time.Time) {
 	e.mu.RLock()
 	defer e.mu.RUnlock()
 
-	return e.data, e.initialized
+	return e.data, e.initialized, e.etag, e.lastModified
```
AlexanderYastrebov, line 35: Lets move `initialized` to the end of the list

routesrv/eskipbytes.go
AlexanderYastrebov: Although the performance difference does not look dramatic, lets use crc32 for now.

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
@nikhil88877 Thank you for the contribution!

AlexanderYastrebov:
@nikhil88877 Maybe you would like to squash into one commit with a clean message as a last touch (otherwise we would squash before merge).

routesrv/eskipbytes.go
szuecs: Hmm crc32 is normally used to detect bit flips in network packet size of data. I think the problem can be that it will show the same output for different routes.
I would say let's use sha256 and call it a day. What we really want is a func that does it uniquely as possible, fast is not very important, it's counted in µs anyways, so who cares about this?
It's an optimization that does not change your hot data path.

nikhil88877: Generating ETag is tricky because there is a tradeoff between generating uniquely(correctness) vs (memory + CPU +  time) performance.

AlexanderYastrebov: Sure, lets use sha256.

routesrv/eskipbytes.go
nikhil88877: Being naive, I assumed reading is always an atomic operation. 
Thanks for the explanation.

AlexanderYastrebov:
:+1:

PR #2097, opened 2022-10-12
feature: ratelimit failure mode config filter

szuecs:
feature: ratelimit failure mode config filter ratelimitFailClosed() for all cluster*Ratelimit, and backendRatelimit filters


szuecs:
:+1:

filters/ratelimit/ratelimit.go
```diff
@@ -457,6 +460,12 @@ func getStatusCodeArg(args []interface{}, index int) (int, error) {
 
 // Request checks ratelimit using filter settings and serves `429 Too Many Requests` response if limit is reached
 func (f *filter) Request(ctx filters.FilterContext) {
+	e, ok := ctx.StateBag()[FailClosedKey]
+	if ok {
+		b, ok := e.(bool)
+		f.settings.FailClosed = ok && b
+	}
```
AlexanderYastrebov, line 467: I am not sure modifying `f.settings` is thread safe here.
Maybe we can utilize `sync.Once` or even a route `PreProprocessor` as we only need to configure it once. I guess this is the downside of having separate filter for the configuration.

docs/reference/filters.md
```diff
@@ -1882,6 +1884,26 @@ Path("/cheap")     -> clusterLeakyBucketRatelimit("user-${request.cookie.Authori
 Path("/expensive") -> clusterLeakyBucketRatelimit("user-${request.cookie.Authorization}", 1, "1s", 5, 2) -> ...
 ```
 
+### ratelimitFailClosed
```
AlexanderYastrebov, line 1887: Initially I was thinking about having a generic `ratelimitControl(config-key, config-value)` filter that we could re-use for other configs.
I guess a filter per specific config is also fine, moreover since in all ratelimit filters we fail open by default we can simplify and make this filter without arguments - users would add `ratelimitFailClosed()` or rely on the default.

filters/ratelimit/ratelimit.go
szuecs: done with PostProcessor

docs/reference/filters.md
szuecs: dropped all args

szuecs:
:+1:

filters/ratelimit/failure_mode.go
```diff
@@ -0,0 +1,76 @@
+package ratelimit
+
+import (
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/routing"
+)
+
+type FailureModeSpec struct{}
+type failureMode struct {
+	failClosed bool
```
AlexanderYastrebov, line 10: Is this used for anything?

filters/ratelimit/failure_mode.go
```diff
@@ -0,0 +1,76 @@
+package ratelimit
+
+import (
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/routing"
+)
+
+type FailureModeSpec struct{}
```
AlexanderYastrebov, line 8: Should it be named after filter, i.e. `RatelimitFailClosedSpec`?

filters/ratelimit/failure_mode.go
```diff
@@ -0,0 +1,76 @@
...
+				found = true
+				continue
+			}
+			// no config changes detected
+			if !found {
+				continue
+			}
+
+			switch f.Name {
+			// leacky bucket has no Settings
```
AlexanderYastrebov, line 40: leacky -> leaky

proxy/proxytest/proxytest.go
```diff
@@ -21,6 +21,10 @@ type TestProxy struct {
 	server  *httptest.Server
 }
 
+func WithParamsAndRoutingOptions(fr filters.Registry, proxyParams proxy.Params, o routing.Options, routes ...*eskip.Route) *TestProxy {
```
AlexanderYastrebov, line 24: Maybe just make `NewTestProxy` exported

skipper.go
```diff
@@ -1580,8 +1581,12 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 			o.ClusterRatelimitMaxGroupShards = 1
 		}
 
+		failureModeFilter := ratelimitfilters.NewFailureMode()
+		failureModeFilterPostProcessor, _ = failureModeFilter.(*ratelimitfilters.FailureModeSpec)
```
AlexanderYastrebov, line 1585: I think it is cleaner to have a separate constructor `ratelimitfilters.NewFailureModePostProcessor()` like it is done for other post-pocessors

filters/ratelimit/failure_mode.go
szuecs: done

proxy/proxytest/proxytest.go
szuecs: This will change much more and maybe we can get rid of all the different `With*` funcs in proxytest altogether in this case. . I think it's better as a separate PR.

skipper.go

filters/ratelimit/failure_mode.go
szuecs: done

filters/ratelimit/failure_mode.go
szuecs: no, cleaned up

proxy/proxytest/proxytest.go
AlexanderYastrebov: I meant rename `newTestProxy`  to `NewTestProxy`, we should have less of `WithWhatever` not more. But this not a big deal, lets have it like you propose.

AlexanderYastrebov:
:+1:

skipper.go
```diff
@@ -1728,6 +1732,7 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 			builtin.NewRouteCreationMetrics(mtr),
 			fadein.NewPostProcessor(),
 			admissionControlSpec.PostProcessor(),
+			failClosedRatelimitPostProcessor,
```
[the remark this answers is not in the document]
szuecs: I only add the postprocessor in case we have ratelimits, because otherwise it does not make sense to add.
I now check for nil before adding it to the PostProcessors

AlexanderYastrebov:
:+1:

PR #2108, opened 2022-10-18
Improve reliability of fadein test

herojan:
- Adds an array of keys which can be used to select each backend endpoint one by one for consistent hash, emulating round robin.
- Increases the tolerance, I saw failures locally for random and round robin too


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2109, opened 2022-10-18
feature: routesv HEAD response with X-Count header

szuecs:

### DIFF
--- a/routesrv/eskipbytes.go
+++ b/routesrv/eskipbytes.go
@@ -76,12 +76,7 @@ func (e *eskipBytes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		w.Header().Add("Content-Type", "text/plain; charset=utf-8")
 		w.Header().Add(routing.RoutesCountName, strconv.Itoa(e.count))
 
-		switch r.Method {
-		case http.MethodHead:
-			w.WriteHeader(http.StatusOK)
-		case http.MethodGet:
-			http.ServeContent(w, r, "", lastModified, bytes.NewReader(data))
-		}
+		http.ServeContent(w, r, "", lastModified, bytes.NewReader(data))
 	} else {
 		w.WriteHeader(http.StatusNotFound)
 	}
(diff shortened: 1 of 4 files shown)


routesrv/eskipbytes.go
```diff
@@ -61,11 +65,23 @@ func (e *eskipBytes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		w.Header().Add("Etag", etag)
 		w.Header().Add("Content-Type", "text/plain; charset=utf-8")
-		http.ServeContent(w, r, "", lastModified, bytes.NewReader(data))
+		w.Header().Add(routing.RoutesCountName, strconv.Itoa(e.count))
+
+		switch r.Method {
+		case http.MethodHead:
+			w.WriteHeader(http.StatusOK)
+		case http.MethodGet:
+			http.ServeContent(w, r, "", lastModified, bytes.NewReader(data))
```
AlexanderYastrebov, line 83: http.ServeContent supports HEAD method so I think we'd only need to add X-Count header

routesrv/eskipbytes.go
```diff
@@ -61,11 +65,23 @@ func (e *eskipBytes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+
 	data, etag, lastModified, initialized := e.bytes()
 	if initialized {
 		w.Header().Add("Etag", etag)
 		w.Header().Add("Content-Type", "text/plain; charset=utf-8")
-		http.ServeContent(w, r, "", lastModified, bytes.NewReader(data))
+		w.Header().Add(routing.RoutesCountName, strconv.Itoa(e.count))
```
AlexanderYastrebov, line 77: Here `e.count` is read without holding the mutex. Either `e.bytes()` should be extended to return count or we should drop it and lock/unclock mutex manually

szuecs: done

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2110, opened 2022-10-24
Support multiple values for -edit-route and -clone-route

### LINKED ISSUE #2106: Support multiple values for -edit-route and -clone-route

https://github.com/zalando/skipper/pull/1822 introduced `-edit-route` and `-clone-route`
https://github.com/zalando/skipper/pull/2060 extended it to support custom separator.

Skipper should support multiple values for `-edit-route` and `-clone-route` to enable complex route modifications.

Allow users to specify `-edit-route` and `-clone-route` multiple times, e.g.:
```sh
skipper -edit-route=/foo/bar/ -edit-route=/baz/qux/ ...
```

See https://github.com/zalando/skipper/blob/master/config/multiflag.go

Yes

RomanZavodskikh:
Fixes #2106


skipper.go
```diff
@@ -325,14 +325,14 @@ type Options struct {
 	// DisabledFilters is a list of filters unavailable for use
 	DisabledFilters []string
 
-	// CloneRoute is a PreProcessor, that will be applied to all routes automatically. It
-	// will clone all matching routes and apply changes to the
+	// CloneRoute is an array PreProcessor's, that will be applied to all routes
+	// automatically. They will clone all matching routes and apply changes to the
 	// cloned routes.
-	CloneRoute *eskip.Clone
+	CloneRoute *[]eskip.Clone
```
AlexanderYastrebov, line 331: I guess it should be `[]*eskip.Clone`

skipper.go
```diff
@@ -1770,11 +1770,17 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	}
 
 	if o.CloneRoute != nil {
-		ro.PreProcessors = append(ro.PreProcessors, o.CloneRoute)
+		for idx := range *o.CloneRoute {
+			cr := (*o.CloneRoute)[idx]
+			ro.PreProcessors = append(ro.PreProcessors, &cr)
+		}
```
AlexanderYastrebov, line 1776: A lot of `*` usually hints something wrong in the datatype design

config/routechangerconfig.go
```diff
@@ -6,17 +6,27 @@ import (
...
+	result := ""
+	for _, rcci := range rcc {
+		if rcci.Reg != nil {
+			result += fmt.Sprintf("%s%s%s%s%s", rcci.Sep, rcci.Reg,
+				rcci.Sep, rcci.Repl, rcci.Sep)
+		}
+		result += "\n"
+	}
+	if len(result) > 0 {
+		result = result[:len(result)-1] // to remove the last \n
```
AlexanderYastrebov, line 27: This is usually done
```go
for i, x := range xx {
  ...
  if i > 0 {
    result += separator
  }
}
```
or you may collect individual value into a slice and then https://pkg.go.dev/strings#Join

config/routechangerconfig.go
```diff
@@ -6,17 +6,27 @@ import (
...
 }
 
+type routeChangerConfig []routeChangerConfigItem
+
 func (rcc routeChangerConfig) String() string {
-	if rcc.Reg == nil {
-		return ""
+	result := ""
+	for _, rcci := range rcc {
+		if rcci.Reg != nil {
```
AlexanderYastrebov, line 20: Why Reg could be nil?

skipper.go
RomanZavodskikh: I do agree that `*[]eskip.Clone` is the worst choice I could make. :-)
However, due to Go slices internals ( https://go.dev/blog/slices-intro ) it seems to me I would try to make it `[]eskip.Clone` because it would decrease number of references levels.

config/routechangerconfig.go
RomanZavodskikh: I think it could not. However, this check is present in code before my changes so I saved it. Okay will remove.

skipper.go
RomanZavodskikh: Okay, I tried to make it `[]eskip.Clone` but due to `eskip.NewClone` returning `*Clone` it seems it won't help to decrease reference levels. Will keep it `[]*eskip.Clone`.

config/routechangerconfig.go
```diff
@@ -6,17 +6,25 @@ import (
...
 	Repl string
 	Sep  string
 }
 
+type routeChangerConfig []routeChangerConfigItem
+
 func (rcc routeChangerConfig) String() string {
-	if rcc.Reg == nil {
-		return ""
+	result := ""
```
AlexanderYastrebov, line 18: https://pkg.go.dev/strings#Builder may be used here like so https://go.dev/play/p/d4Ub_GXEmoP

skipper.go
```diff
@@ -1770,11 +1770,17 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	}
 
 	if o.CloneRoute != nil {
-		ro.PreProcessors = append(ro.PreProcessors, o.CloneRoute)
+		for idx := range o.CloneRoute {
+			cr := o.CloneRoute[idx]
+			ro.PreProcessors = append(ro.PreProcessors, cr)
+		}
```
AlexanderYastrebov, line 1776: Wouldn't this work?
```go
ro.PreProcessors = append(ro.PreProcessors, o.CloneRoute...)
```
if not it still could be simpler
```go
for _, cr := range o.CloneRoute {
	ro.PreProcessors = append(ro.PreProcessors, cr)
}
```

config/config.go
```diff
@@ -775,8 +775,8 @@ func (c *Config) ToOptions() skipper.Options {
 			Append:  c.AppendFilters.filters,
 		},
 		DisabledFilters:    c.DisabledFilters.values,
-		CloneRoute:         eskip.NewClone(c.CloneRoute.Reg, c.CloneRoute.Repl),
-		EditRoute:          eskip.NewEditor(c.EditRoute.Reg, c.EditRoute.Repl),
+		CloneRoute:         []*eskip.Clone{},
```
AlexanderYastrebov, line 778: No need to init with empty slice as `append` works on nil slices https://go.dev/tour/moretypes/15

config/config.go
```diff
@@ -887,6 +887,14 @@ func (c *Config) ToOptions() skipper.Options {
 
 		ClusterRatelimitMaxGroupShards: c.ClusterRatelimitMaxGroupShards,
 	}
+	for _, rcci := range *c.CloneRoute {
```
AlexanderYastrebov, line 890: Why do we need to deference pointer here?
Since `routeChangerConfig` became slice I think `c.CloneRoute` type should not be pointer type anymore. See how `multiFlag` is used

config/routechangerconfig.go
```diff
@@ -6,17 +6,25 @@ import (
...
 	Repl string
 	Sep  string
 }
 
+type routeChangerConfig []routeChangerConfigItem
+
 func (rcc routeChangerConfig) String() string {
-	if rcc.Reg == nil {
-		return ""
+	b := strings.Builder{}
```
AlexanderYastrebov, line 18: minor: this is equivalent but stdlib example suggests to use `var b strings.Builder`, see https://pkg.go.dev/strings#Builder

skipper.go
szuecs: The problem is that `routing.PreProcessor` is an interface type and `*eskip.Clone` is implementing the interface.
CloneRoute would have to be the interface type in case this ... thing would work as far as I understand.
If the second way works it's fine to use this one.

config/config.go
szuecs: https://www.youtube.com/watch?v=ynoY2xz-F8s is nice to watch to understand nil in Go. It's fundamental different than in other languages. You can even do something like this to work around nil ptr panic:

```
type t struct{
    s string
}

func (t *t) String() string {
    if t == nil {
        t := new(t)
    }
    return t.s
}
```

config/config.go
szuecs: Yes in Go you basically never need to dereference ptr yourself.

skipper.go
```diff
@@ -325,14 +325,14 @@ type Options struct {
 	// DisabledFilters is a list of filters unavailable for use
 	DisabledFilters []string
 
-	// CloneRoute is a PreProcessor, that will be applied to all routes automatically. It
-	// will clone all matching routes and apply changes to the
+	// CloneRoute is an array PreProcessor's, that will be applied to all routes
```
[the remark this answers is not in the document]
szuecs: It's not an array, but a slice. array != slice.
Array is fixed size and slice can grow.

config/routechangerconfig.go
```diff
@@ -28,11 +36,15 @@ func (rcc *routeChangerConfig) Set(value string) error {
...
 	reg, repl := a[1], a[2]
-	rcc.Reg, err = regexp.Compile(reg)
-	rcc.Repl = repl
-	rcc.Sep = firstSym
+
+	var rcci routeChangerConfigItem
+	rcci.Reg, err = regexp.Compile(reg)
+	rcci.Repl = repl
+	rcci.Sep = firstSym
+	*rcc = append(*rcc, rcci)
```
szuecs, line 47: This should work and reads nicer:
```
reg, err = regexp.Compile(reg)
if err != nil { return err }
rcci := &routeChangerConfigItem{
   Reg: reg,
   Repl: repl
   Sep: firstSym
}

rcc = append(rcc, rci)
```

config/routechangerconfig.go
```diff
@@ -6,17 +6,25 @@ import (
...
-	if rcc.Reg == nil {
-		return ""
+	b := strings.Builder{}
+	for i, rcci := range rcc {
+		if i > 0 {
+			b.WriteString("\n")
+		}
+
+		fmt.Fprintf(&b, "%s%s%s%s%s", rcci.Sep, rcci.Reg,
+			rcci.Sep, rcci.Repl, rcci.Sep)
```
szuecs, line 25: I would say implement Stringer interface in routeChangerConfigItem and use it

```
func (rcci routeChangerConfigItem) String() string {
    return rcci.Sep + rcci.Reg + rcci.Sep + rcci.Repl + rcci.Sep
}

// and then use
b.WriteString(rcci.String())
```

szuecs: alternatively you can also use Sprintf to return, but format string is more costly and we seem not to really need it other than using the regexp as string as we can do by calling String().

skipper.go
```diff
@@ -325,14 +325,14 @@ type Options struct {
...
+	// CloneRoute is a slice of PreProcessor's that will be applied to all routes
+	// automatically. They will clone all matching routes and apply changes to the
 	// cloned routes.
-	CloneRoute *eskip.Clone
+	CloneRoute []*eskip.Clone
 
 	// EditRoute will be applied to all routes automatically and
 	// will apply changes to all matching routes.
-	EditRoute *eskip.Editor
+	EditRoute []*eskip.Editor
```
szuecs, line 335: These changes are actually breaking changes.
@AlexanderYastrebov wdyt?

szuecs: We can have new option `EditRoutes` and just append EditRoute to EditRoutes

RomanZavodskikh: I don't see the case why this could break something.
In both cases I see:
```
./bin/skipper ... -edit-route ... -edit-route ...
```
and
```
edit-route: ... change1 ...
edit-route: ... change2 ...
```
in `config.yaml` both `edit-route` options are working fine.

In case when there is only one `edit-route` options (all use cases at the moment) this only `edit-route` option is still working good.

szuecs: yes binary users are not impaired, but library users will see a change from a single type to a slice type

RomanZavodskikh: @szuecs , how many library users do we have? Will they have problems only during compile time or during runtime too? It seems to me that if there are not too many users and there are only problems during compile time than it is better force them to recompile one more time but not to add additional workaround field in `skipper.Options` struct.

AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

MustafaSaber:
👍

PR #2112, opened 2022-10-24
lua: add flag to selectively enable modules and symbols

AlexanderYastrebov:

### DIFF
--- a/script/module.go
+++ b/script/module.go
@@ -1,0 +1,117 @@
+package script
+
+import (
+	"strings"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	lua "github.com/yuin/gopher-lua"
+)
+
+type luaModule struct {
+	name   string
+	loader lua.LGFunction
+
+	disabledSymbols []string
+}
+
+var standardModules = []luaModule{
+	// Load Package and Base first, see lua.LState.OpenLibs()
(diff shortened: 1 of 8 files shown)


script/module.go
```diff
@@ -0,0 +1,117 @@
...
+		for _, name := range m.disabledSymbols {
+			st.RawSetString(name, lua.LNil)
+		}
+	}
+}
+
+// withSymbols returns copy of module with selected symbols
+func (m luaModule) withSymbols(L *lua.LState, enabledSymbols []string) luaModule {
+	// gopher-lua does not have API to select enabled symbols,
+	// see https://github.com/yuin/gopher-lua/discussions/393
```
AlexanderYastrebov, line 54: https://github.com/yuin/gopher-lua/discussions/393

szuecs:
@AlexanderYastrebov should we have a default to only load a subset of modules and increase minor version to explain that this would break in case you use lua filter and you need to check the release/docs pages?
I think a subset would be something like http/json/base64 modules and maybe some string functions, which are safe to use.

AlexanderYastrebov:
This change is backward compatible. If we decide to do a breaking change then I would rather disable lua altogether by default

szuecs:
:+1:

AlexanderYastrebov:
:+1:

Issue #2003, opened 2022-05-09
panic on rolling out SWARM


universam1:
**Describe the bug**
Upon rolling out deployment changes enabling swarm mode, the new pods error out with

```
[APP]time="2022-05-09T12:46:57Z" level=error msg="SWARM: failed to parse the ip "
[APP]time="2022-05-09T12:46:57Z" level=error msg="SWARM: failed to parse the ip "
[APP]time="2022-05-09T12:46:57Z" level=error msg="SWARM: failed to parse the ip "
[APP]time="2022-05-09T12:46:57Z" level=error msg="SWARM: failed to parse the ip "
[APP]time="2022-05-09T12:46:57Z" level=error msg="SWARM: failed to parse the ip "
[APP]time="2022-05-09T12:46:57Z" level=error msg="SWARM: failed to parse the ip "
[APP]time="2022-05-09T12:46:57Z" level=info msg="SWARM: <nil> is going to join swarm of 9 nodes ([NodeInfo{name: skipper-ingress-5cc58d7889-c4zwf, 172.16.7.27:9990} NodeInfo{name: skipper-ingress-5cc58d7889-n58f4, 172.16.12.158:9990} NodeInfo{name: skipper-ingress-5cc58d7889-qbzt5, 172.16.15.209:9990} NodeInfo{name: skipper-ingress-5cc58d7889-qw6zk, 172.16.7.77:9990} NodeInfo{name: skipper-ingress-5cc58d7889-v828t, 172.16.23.212:9990} NodeInfo{name: skipper-ingress-6cfb658b7d-9n8vg, 172.16.7.133:9990} NodeInfo{name: skipper-ingress-6cfb658b7d-pjc6r, 172.16.20.170:9990} NodeInfo{name: skipper-ingress-6cfb658b7d-q98fx, 172.16.23.96:9990} NodeInfo{name: skipper-ingress-6cfb658b7d-vt5nl, 172.16.12.240:9990}])"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xe9e9cf]

goroutine 1 [running]:
github.com/zalando/skipper/swarm.Join({0x0, 0x400000, 0x12a05f200, 0x2706, 0xc00008d4a0, 0x0, 0x0, {0x0, 0x0}, 0x0}, ...)
    /workspace/swarm/swarm.go:209 +0x14f
[...]

PR #2116, opened 2022-10-26
Fix swarm.Join() when NodeInfo pointer is nil

demonCoder95:
Fixes issue of panicking swam by handling the case where `NodeInfo` pointer is `nil`.

https://github.com/zalando/skipper/issues/2003


szuecs:
:+1:

demonCoder95:
👍

Issue #1005, opened 2019-03-25
Skipper panics if swarm port occupied at start


roffe:
**Describe the bug**
Skipper panics on requests if the swarm port is occupied on runtime

**To Reproduce**
Add context, for example:
1. enable swim clustering
2. occupie port on the host with f.x `nc -lp 9990`
3. start skipper with swarm (swim) ratelimiting

**Expected behavior**
magic

**Observed behavior**
ALB throws 502 when req comes in to Skipper.
In skipper logs:

```
[APP]time="2019-03-25T15:17:52Z" level=error msg="SWARM: failed to create memberlist: Could not set up network transport: Failed to start TCP listener on \"10.0.117.154\" port 9990: listen tcp 10.0.117.154:9990: bind: address already in use"
[APP]time="2019-03-25T15:17:52Z" level=error msg="failed to init swarm with options &{swarm:0 MaxMessageBuffer:4194304 LeaveTimeout:5s SwarmPort:9990 KubernetesOptions:0xc0000b85f0 StaticSwarm:<nil> FakeSwarm:false FakeSwarmLocalNode: Debug:false}: Could not set up network transport: Failed to start TCP listener on \"10.0.117.154\" port 9990: listen tcp 10.0.117.154:9990: bind: address already in use"
[APP]time="2019-03-25T15:20:32Z" level=error msg="swarm is nil"
2019/03/25 15:20:32 http: panic serving 10.0.114.193:38580: runtime error: invalid memory address or nil pointer dereference
goroutine 427 [running]:
net/http.(*conn).serve.func1(0xc000308fa0)
	/usr/local/go/src/net/http/server.go:1769 +0xc9
panic(0xdd5320, 0x17ba690)
	/usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/zalando/skipper/swarm.(*Swarm).broadcast(0x0, 0xc0004a9c80, 0xc0004a9c00, 0xc00038c080)
	/workspace/swarm/swarm.go:362 +0x34
github.com/zalando/skipper/swarm.(*Swarm).ShareValue(0x0, 0xc00052c840, 0x20, 0xd7cde0, 0xc00038c080, 0x0, 0x1)
	/workspace/swarm/swarm.go:383 +0x7f
github.com/zalando/skipper/ratelimit.(*clusterLimitSwim).Allow(0xc00029c320, 0xc00038c070, 0xc, 0xc)
	/workspace/ratelimit/swim.go:96 +0x23e
[...]

PR #2118, opened 2022-10-27
Fix swarm panic when swarm is `nil`.

demonCoder95:
Fixes the panic in swarm `broadcast` function when the swarm pointer is `nil`.

Fixes: #1005 

Signed-off-by: noor <noormuhammadmalik95@gmail.com>


demonCoder95:
👍

szuecs:
:+1:

PR #2119, opened 2022-10-28
build: create git tag early

AlexanderYastrebov:
Build step pushes multiple docker images which may fail in the middle before git tag is created. Subsequent pipeline ends up reusing the tag from the previously failed pipeline and fails to push docker image with `tag invalid: tag already exists`.

Also adds `set -euo pipefail` to prevent masking of shell pipeline errors.

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -27,8 +27,11 @@ pipeline:
       docker buildx create --config /etc/cdp-buildkitd.toml --driver-opt network=host --bootstrap --use
   - desc: build-push
     cmd: |
+      set -euo pipefail
+
       IMAGE_REGISTRY="registry-write.opensource.zalan.do"
       MULTIARCH_REGISTRY="container-registry-test.zalando.net"
+
       if [[ $CDP_TARGET_BRANCH == master && ! $CDP_PULL_REQUEST_NUMBER ]]; then
         LATEST_VERSION=$(git describe --tags --always | awk -F \- '{print $1}')
         CUR_PART=$(echo $LATEST_VERSION | awk -F . '{print $1"."$2}')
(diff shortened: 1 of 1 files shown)


delivery.yaml
```diff
@@ -38,6 +40,10 @@ pipeline:
         if [ "$CUR_PART" != "$VERSION_PART" ]; then NEW_PATCH=0; fi
         RELEASE_VERSION=${VERSION_PART}.${NEW_PATCH}
         export VERSION="${RELEASE_VERSION}"
+
+        echo "Creating git tag: ${RELEASE_VERSION}"
+        git gh-tag "${RELEASE_VERSION}"
```
szuecs, line 45: This will create always a tag even if linters fail.

szuecs:
I think the PR is wrong @AlexanderYastrebov .

delivery.yaml
AlexanderYastrebov: This might be ok, at least it would not block subsequent pipelines like it does now but just "waste" a version number

AlexanderYastrebov: I can move it after `make deps cicheck staticcheck gosec` but before docker builds

szuecs: Yes please

AlexanderYastrebov: Done

delivery.yaml
```diff
@@ -51,21 +55,29 @@ pipeline:
...
+      cd packaging && make docker.build.amd64 && make docker.push.amd64 && make docker.push.multiarch
+
       if [[ $CDP_TARGET_BRANCH == master && ! $CDP_PULL_REQUEST_NUMBER ]]; then
         echo "Created docker image registry.opensource.zalan.do/teapot/skipper:${RELEASE_VERSION}"
+
         cdp-promote-image "${MULTIARCH_IMAGE}"
         echo "Created multi-arch docker image container-registry.zalando.net/teapot/skipper:${RELEASE_VERSION}"
+
         echo "Creating docker image registry.opensource.zalan.do/teapot/skipper-arm64:${RELEASE_VERSION}"
-        make docker.build.arm64 && git status && git diff && make docker.push.arm64
```
AlexanderYastrebov, line 62: Removed `git status && git diff` everywhere

szuecs: Why?

AlexanderYastrebov: Reverted

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2122, opened 2022-10-31
proxy: add request host to error log

AlexanderYastrebov:

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1314,7 +1314,7 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 		uri = uri[:i]
 	}
 	logFunc(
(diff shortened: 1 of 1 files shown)


szuecs:
We log routeID which should be easy to get the host and CodeQL says it's not secure to log content from users...

AlexanderYastrebov:
The change allows aggregating various errors by hostname. 

> We log routeID which should be easy to get the host

unless it is `_unknownroute_`

> CodeQL says it's not secure to log content from users

we log url path and user-agent just next to this

szuecs:
Yes but unknown route doesn't matter because it will end in 404.
Or do you want to check kubernetes user errors in access logs?

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2124, opened 2022-11-01
Feature / Adding url path components opentracing tag

mshafiee:
Currently, it is not possible to filter URLs that have variable in the middle of their path (e.g: Hash, product code, ...) In opentracing systems. The lack of this ability makes opentracing very ineffective in reporting, especially in preparing SLO reports. In this feature, we separate path components based on forward slashes and place each in a special tag.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1484,6 +1484,14 @@ func (p *Proxy) setCommonSpanInfo(u *url.URL, r *http.Request, s ot.Span) {
 		setTag(s, HostnameTag, p.hostname).
 		setTag(s, HTTPPathTag, u.Path).
 		setTag(s, HTTPHostTag, r.Host)
+	pathComponents := strings.Split(u.Path, "/")
+	for i, part := range pathComponents {
+		if len(part) < 1 {
+			continue
+		}
+		HTTPPathPartTag := fmt.Sprintf("%s_%d", HTTPPathPartPrefixTag, i)
+		p.tracing.setTag(s, HTTPPathPartTag, part)
+	}
 	if val := r.Header.Get("X-Flow-Id"); val != "" {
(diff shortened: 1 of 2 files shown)


AlexanderYastrebov:
Hello, thank you for the PR.
In general we would like to have minimal amount of tracing tags required due to performance and costs reasons.

> Currently, it is not possible to filter URLs that have variable in the middle of their path (e.g: Hash, product code, ...) 

Would something like this:
```sh
bin/skipper -inline-routes='Path("/products/:pid/whatever") -> tracingTag("http.path.pid", "${pid}") -> "http://backend.test"'
```
work for your usecase?

mshafiee:
The issue is that to calculate SLO, we need to measure metrics from the point closest to the user, which is naturally ingress. But the problem we have is that services like Lightstep do not have the ability to query the contents of the path field. They can only check if a value is equal or not. They do not have the contain function.
On the other hand, there are variable values ​​such as ProductID in many paths. Therefore, we have to make a change in the skipper that tags the components of the URL Path separately so that it is possible to query based on the Path.

It does not seem that such a feature can be implemented using [tracingTag](https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#tracingtag). It seems that the tracingTag feature is implemented to create fixed tags, not variable tags.

If there is concern about performance, we can use a command line flag to activate/deactivate this feature.

AlexanderYastrebov:
> It seems that the tracingTag feature is implemented to create fixed tags, not variable tags.

I am not sure I understand, the example from https://github.com/zalando/skipper/pull/2124#issuecomment-1298766869 would set tag value based on the actual value in the path.

Could you provide and example URL and resulting tag you'd like to end up with?

szuecs:
@mshafiee the current implementation will blow up the span and we can not afford to have this.

PR #2051, opened 2022-08-07
feat: added leakage verification 

alexandreliberato:
To avoid defects and resource waste we can use uber's goleak to detect goroutine leakages.

## Added
- uber-go/goleak library
- main_test files to test leakage in all test package tests (eskipfile, filters/serve and dataclients/kubernets)

Related to #1989

PR #2128, opened 2022-11-03
use goleak to autodetect leaks

szuecs:
replaces https://github.com/zalando/skipper/pull/2051/ by using https://github.com/AlexanderYastrebov/noleak

One problem I see is https://github.com/uber-go/goleak/issues/83 , so not sure if we want to merge this or better add only fixes for findings and drop the use of goleak. What do you think?

We agreed on that we use only package based tests and I decided to give https://github.com/AlexanderYastrebov/noleak a try since it's simpler than goleak and Uber engineers seem not to be responsive in this project.


dataclients/kubernetes/main_test.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"testing"
+
+	"go.uber.org/goleak"
+)
+
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
+}
```
szuecs, line 11: I am not completely sure what this does but https://github.com/uber-go/goleak#quick-start seems to suggest it will put it into all tests. I validated with a known broken test (proxy package that has test with go-redis that starts heartbeat which is not closed) and it does not seem to matter if we are in the same package `kubernetes` here or another package like `kubernetes_test`.

MustafaSaber: Then let's remove it if it doesn't matter?

MustafaSaber:
👍

dataclients/kubernetes/main_test.go
szuecs: I think this test injects the test for every test in the package

AlexanderYastrebov:
> not sure if we want to merge this or better add only fixes for findings and drop the use of goleak

Did it discover any leaks actually?

tracing/tracers/basic/basic.go
```diff
@@ -4,13 +4,25 @@ import (
...
 )
 
-func InitTracer(opts []string) (opentracing.Tracer, error) {
+type CloseableTracer interface {
+	opentracing.Tracer
+	Close()
+}
+
+type BasicTracer struct {
+	tracer opentracing.Tracer
```
AlexanderYastrebov, line 20: I think here you can embed `opentacing.Tracer` to avoid implementing method overrides
```go
type BasicTracer struct {
	opentracing.Tracer
	...
}
```

tracing/tracers/basic/basic.go
```diff
@@ -4,13 +4,25 @@ import (
...
 	opentracing "github.com/opentracing/opentracing-go"
 )
 
-func InitTracer(opts []string) (opentracing.Tracer, error) {
+type CloseableTracer interface {
+	opentracing.Tracer
+	Close()
+}
+
+type BasicTracer struct {
```
AlexanderYastrebov, line 19: Do we need this to be exported type?

tracing/tracers/basic/basic.go
```diff
@@ -4,13 +4,25 @@ import (
...
+	tracer opentracing.Tracer
+	quit   chan struct{}
+	once   sync.Once
+}
+
+func InitTracer(opts []string) (CloseableTracer, error) {
```
AlexanderYastrebov, line 25: This would be API breaking change

eskipfile/remote.go
```diff
@@ -135,6 +138,19 @@ func (client *remoteEskipFile) LoadUpdate() ([]*eskip.Route, []string, error) {
...
+		client.once.Do(func() {
+			if client.http != nil {
+				client.http.Close()
+			}
+			if client.eskipFileClient != nil {
+				client.eskipFileClient.Close()
+			}
+		})
+	}
+}
```
AlexanderYastrebov, line 152: Do we need nil checks here?
- nil *remoteEskipFile seems to be weird, I know go allows nil target object but I think by this check we may hide some kind of underlying problem
- client.http and client.eskipFileClient could not be nil afaict

eskipfile/remote_test.go
```diff
@@ -91,6 +91,13 @@ func TestLoadAll(t *testing.T) {
 		t.Run(test.title, func(t *testing.T) {
 			options := &RemoteWatchOptions{RemoteFile: s.URL, Threshold: 10, Verbose: true, FailOnStartup: true}
 			client, err := RemoteWatch(options)
+			defer func() {
+				c, ok := client.(*remoteEskipFile)
+				if ok {
```
AlexanderYastrebov, line 96: Do we need ok check here? I think its better for test to fail if the type ever changes

net/httpclient.go
```diff
@@ -70,10 +73,14 @@ func NewClient(o Options) *Client {
...
-	if c.sr != nil {
-		c.sr.Close()
-	}
+	c.once.Do(func() {
+		if c.tr != nil {
+			c.tr.Close()
+		}
+		if c.sr != nil {
+			c.sr.Close()
+		}
```
AlexanderYastrebov, line 82: Why do we need these nil checks?

net/httpclient.go
```diff
@@ -320,15 +329,25 @@ func WithBearerToken(t *Transport, bearerToken string) *Transport {
...
+	}
 }
 
 func (t *Transport) Close() {
-	if !t.closed {
-		t.closed = true
-		close(t.quit)
-	}
+	t.once.Do(func() {
+		if !t.closed {
```
AlexanderYastrebov, line 346: If there is no other way to close transport I think we do not need `t.closed` field at all

net/redisclient_test.go
```diff
@@ -156,7 +147,9 @@ func TestRedisClient(t *testing.T) {
 				}
 			}
 
-			go func() { <-ch }() // create client will block
+			if ch != nil {
+				go func() { <-ch }() // create client will block
+			}
```
AlexanderYastrebov, line 152: I think we could create a buffered channel `ch = make(chan struct{}, 1)` to make it cleaner and unblock first `AddrUpdater` call.

routing/datasource_test.go
```diff
@@ -16,6 +16,7 @@ import (
 )
 
 func TestNoMultipleTreePredicates(t *testing.T) {
+
```
AlexanderYastrebov, line 19: ?

AlexanderYastrebov:
> One problem I see is https://github.com/uber-go/goleak/issues/83 , so not sure if we want to merge this or better add only fixes for findings and drop the use of goleak. What do you think?

We should definitely include leak fixes!

There is no easy way to fix https://github.com/uber-go/goleak/issues/83 as `defer goleak.VerifyNone(t)` implementation looks at the snapshot of all goroutines at the test end so it would see goroutines started by other tests. 

There is also an older lib https://github.com/fortytw2/leaktest that looks at the diff between goroutine snapshots before test start and after test ends.

> Who knows how this works with Parallel tests.

Both libraries provide per-test checks (`defer goleak.VerifyNone(t)` and `defer leaktest.Check(t)()`) that would not work well with parallel tests.

`go-leak` provides `goleak.VerifyTestMain(m)` to check leaks after package tests (`leaktest` does not, see https://github.com/fortytw2/leaktest/issues/4) that should work for parallel tests.

I think we should use `goleak.VerifyTestMain(m)` per package and not pollute tests with `defer goleak.VerifyNone(t)`.
There is also `GODEBUG="tracebackancestors=N"` flag that simplifies leak debugging, see https://pkg.go.dev/runtime

PS: I was curious how it works and to understand re-implemented it here https://github.com/AlexanderYastrebov/noleak
It compares snapshots before and after test (like `leaktest`) and also supports `TestMain` (like `go-leak`)

eskipfile/remote_test.go
```diff
@@ -91,6 +91,13 @@ func TestLoadAll(t *testing.T) {
...
+					c.Close()
+					time.Sleep(time.Second)
```
AlexanderYastrebov, line 98: I think this sleep is redundant

eskipfile/remote_test.go
AlexanderYastrebov: I see now `RemoteWatch` API does not allow to close returned client :(

tracing/tracers/basic/basic.go
szuecs: fixed

eskipfile/remote.go
szuecs: I think if we cleanup (Close) the remoteEskipFile it should be nil, so every concurrent access would be nil ptr.
The client ones I will clean up, I agree.

eskipfile/remote_test.go
szuecs: I thought it fixes the goroutines cleanup, because Close() is normally async: close(quitCH) and then it has to schedule the goroutine in the `for{ select {} }` loop...

eskipfile/remote.go
AlexanderYastrebov: Hm, we use `client.once.Do` which should take care of concurrent access

net/httpclient.go
szuecs: yes otherwise:
```
% go test ./net                                                                 test/go-leak
DO NOT USE IN PRODUCTION
--- FAIL: TestClient (0.00s)
    --- FAIL: TestClient/All_defaults,_with_request_should_have_a_response (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xabe033]

goroutine 123 [running]:
panic({0xcb2160, 0x14543b0})
        /usr/share/go/src/runtime/panic.go:987 +0x3ba fp=0xc00001d7b8 sp=0xc00001d6f8 pc=0x4398da
testing.tRunner.func1.2({0xcb2160, 0x14543b0})
        /usr/share/go/src/testing/testing.go:1396 +0x24e fp=0xc00001d868 sp=0xc00001d7b8 pc=0x5203ae
testing.tRunner.func1()
        /usr/share/go/src/testing/testing.go:1399 +0x39f fp=0xc00001d9d0 sp=0xc00001d868 pc=0x51fe3f
runtime.deferCallSave(0xc00001daa0, 0xc00001df90?)
        /usr/share/go/src/runtime/panic.go:796 +0x88 fp=0xc00001d9e0 sp=0xc00001d9d0 pc=0x4394c8
runtime.runOpenDeferFrame(0xc000482190?, 0xc000482140)
        /usr/share/go/src/runtime/panic.go:769 +0x1a5 fp=0xc00001da28 sp=0xc00001d9e0 pc=0x4392e5
panic({0xcb2160, 0x14543b0})
        /usr/share/go/src/runtime/panic.go:884 +0x212 fp=0xc00001dae8 sp=0xc00001da28 pc=0x439732
runtime.panicmem(...)
        /usr/share/go/src/runtime/panic.go:260
runtime.sigpanic()
        /usr/share/go/src/runtime/signal_unix.go:835 +0x2f6 fp=0xc00001db38 sp=0xc00001dae8 pc=0x450716
github.com/zalando/skipper/net.(*Client).Close.func1()
        /home/sszuecs/go/src/github.com/zalando/skipper/net/httpclient.go:81 +0x33 fp=0xc00001db58 sp=0xc00001db38 pc=0xabe033
sync.(*Once).doSlow(0xdd1d90?, 0x21?)
        /usr/share/go/src/sync/once.go:74 +0xc2 fp=0xc00001dbb8 sp=0xc00001db58 pc=0x4783c2
sync.(*Once).Do(...)
        /usr/share/go/src/sync/once.go:65
github.com/zalando/skipper/net.(*Client).Close(0xc000134600?)
        /home/sszuecs/go/src/github.com/zalando/skipper/net/httpclient.go:79 +0x51 fp=0xc00001dbe8 sp=0xc00001dbb8 pc=0xabdfd1
github.com/zalando/skipper/net.TestClient.func1.4()
        /home/sszuecs/go/src/github.com/zalando/skipper/net/httpclient_test.go:164 +0x26 fp=0xc00001dc00 sp=0xc00001dbe8 pc=0xac8146
github.com/zalando/skipper/net.TestClient.func1(0xc0002031e0)
        /home/sszuecs/go/src/github.com/zalando/skipper/net/httpclient_test.go:199 +0xca4 fp=0xc00001df70 sp=0xc00001dc00 pc=0xac8084
testing.tRunner(0xc0002031e0, 0xc000012318)
```
which is a bit unclear how this can be true for `tr/sr` ...
These seem to be safe during initialization.

net/httpclient.go
szuecs: removed t.closed, thanks

net/httpclient.go
AlexanderYastrebov: Check out `shallowCopy`, I think it is missing `sr` copy. Maybe we can use pointer to `sync.Once` to use `tt := *t` and re-initialize it inside of `shallowCopy`.
TBH I do not thin that `WithX` API that requires `shallowCopy` is good

net/redisclient_test.go
szuecs: I tested the suggestion and it does not work

routing/datasource_test.go
szuecs: fixed

tracing/tracers/basic/basic.go
szuecs: I don't really like embedding, but it seems to work in this case.

tracing/tracers/basic/basic.go
szuecs: This is only basic tracer which is used only for tests. There is no value having this in any production workload, so I think it's fine to break API here.

eskipfile/remote.go
szuecs: but if `client` is nil how can it figure this out?

net/httpclient.go
szuecs: I think ptr to sync.Once is not a good idea. You might want to close the original after you got your replacement and then you closed also your replacement....

eskipfile/remote.go
AlexanderYastrebov: How client can be nil in the first place?

net/httpclient.go
AlexanderYastrebov: > and then you closed also your replacement

No, I mean
```go
func (t *Transport) shallowCopy() *Transport {
	tt := *t
	tt.once = &sync.Once{}
	return &tt
```
to automatically ensure all other fields are copied

szuecs: `sr` is not part of the Transport, but the Client so I think also the mentioned change won't be correct.

eskipfile/remote_test.go
AlexanderYastrebov: I still think `client.(*remoteEskipFile).Close()` would be better

net/httpclient.go
AlexanderYastrebov: Ah, ok, `tr` can not be nil but `sr` can.

The mentioned change is nice to ensure we do not forget to copy client fields we add later

szuecs: ok removed nil check for transport

eskipfile/remote_test.go
szuecs: why is this important?

eskipfile/remote.go
szuecs: I mean on tear down. You have 2 goroutines one is entering `once` and the other is blocked. After cleanup `client` is nil but you have a blocked goroutine that will be woken up to check `client.once.Do` and then you deref the nil ptr. This is just a wild guess, because of the test output.

eskipfile/remote_test.go
AlexanderYastrebov: To let if fail if client type changes in the future

net/redisclient_test.go
AlexanderYastrebov: Yeah, would need to consume first update produced by NewRedisRingClient:
```diff
diff --git a/net/redisclient_test.go b/net/redisclient_test.go
index e648a40f..72a8dc7a 100644
--- a/net/redisclient_test.go
+++ b/net/redisclient_test.go
@@ -149,14 +149,13 @@ func TestRedisClient(t *testing.T) {
                        if tt.options.AddrUpdater != nil {
                                orig = tt.options.AddrUpdater()
                                // test background update
-                               ch = make(chan struct{})
+                               ch = make(chan struct{}, 1)
                                tt.options.AddrUpdater = func() []string {
                                        ch <- struct{}{}
                                        return []string{redisAddr, redisAddr2}
                                }
                        }
 
-                       go func() { <-ch }() // create client will block
                        cli := NewRedisRingClient(tt.options)
                        defer func() {
                                if !cli.closed {
@@ -170,6 +169,8 @@ func TestRedisClient(t *testing.T) {
                        }
 
                        if tt.options.AddrUpdater != nil {
+                               <-ch // read update produced by NewRedisRingClient
+
                                if !cmp.Equal(orig, []string{redisAddr}) {
                                        t.Errorf("Failed to get addr: %v", cmp.Diff(orig, []string{redisAddr}))
                                }
```

Lets keep it like it is.

eskipfile/remote_test.go
szuecs: I get:
```
% go test ./eskipfile        
--- FAIL: TestLoadAll (0.00s)
    --- FAIL: TestLoadAll/Download_not_existing_remote_file_fails_in_NewRemoteEskipFile (0.00s)
panic: interface conversion: routing.DataClient is nil, not *eskipfile.remoteEskipFile [recovered]
        panic: interface conversion: routing.DataClient is nil, not *eskipfile.remoteEskipFile
..
github.com/zalando/skipper/eskipfile.TestLoadAll.func1(0xc000103520)
        /home/sszuecs/go/src/github.com/zalando/skipper/eskipfile/remote_test.go:94 +0x3d1 fp=0xc000409f70 sp=0xc000409ea0 pc=0xe0c151
testing.tRunner(0xc000103520, 0xc000118960)
```

szuecs: I have a fix: check `err == nil` before running the defer Close :)

AlexanderYastrebov: Need to move defer after err checking

eskipfile/remote.go
AlexanderYastrebov: I do not see where we set client to nil

tracing/tracers/basic/basic.go
AlexanderYastrebov: Ok, I am fine to have it like that

secrets/file.go
```diff
@@ -142,33 +142,39 @@ func (sp *SecretPaths) registerSecretFile(p string) error {
 
 // runRefresher refreshes all secrets, that are registered
 func (sp *SecretPaths) runRefresher() {
-	log.Infof("Run secrets path refresher every %s, but update once first", sp.refreshInterval)
-	var d time.Duration
+	sp.refresh()
```
AlexanderYastrebov, line 145: :+1:

szuecs: :)

eskipfile/remote_test.go
szuecs: What? See `==` vs `!=`, I know not the Go obvious to check `==`, but it's the simplest to do.

eskipfile/remote.go
szuecs: True we don't, but it also does not matter right now to do more checking than maybe needed. It's not the hot path and we can also make it "nice" later or is this for you a hard mandatory fix that we need to do now?

AlexanderYastrebov: This is a new method where we add != nil check for no reason. As a reader I am confused by this check - it hints that client can be nil while it can not. Why fix it later if we can just not add it? :shrug:

eskipfile/remote_test.go
AlexanderYastrebov: Ok :)

szuecs:
:+1:

eskipfile/remote.go
AlexanderYastrebov: I've removed nil check

AlexanderYastrebov:
:+1:

PR #2129, opened 2022-11-03
add context if ratelimit allows/denies to the span as Tag

szuecs:

### DIFF
--- a/ratelimit/redis.go
+++ b/ratelimit/redis.go
@@ -77,32 +77,28 @@ func (c *clusterLimitRedis) measureQuery(format, groupFormat string, fail *bool,
 	c.metrics.MeasureSince(key, start)
 }
 
-func (c *clusterLimitRedis) startSpan(ctx context.Context, spanName string) func(bool, bool) {
-	nop := func(bool, bool) {}
+func parentSpan(ctx context.Context) (opentracing.Span, bool) {
 	if ctx == nil {
-		return nop
+		return nil, false
 	}
 
 	parentSpan := opentracing.SpanFromContext(ctx)
 	if parentSpan == nil {
-		return nop
+		return nil, false
 	}
 
-	span := c.ringClient.StartSpan(spanName, opentracing.ChildOf(parentSpan.Context()))
-	ext.Component.Set(span, "skipper")
-	ext.SpanKind.Set(span, "client")
-	span.SetTag("ratelimit_type", c.typ)
-	span.SetTag("group", c.group)
-	span.SetTag("max_hits", c.maxHits)
-	span.SetTag("window", c.window.String())
+	return parentSpan, true
+}
 
-	return func(allowed, failed bool) {
-		if failed {
-			ext.Error.Set(span, true)
-		}
-		span.SetTag("allowed", allowed)
-
-		span.Finish()
+func (c *clusterLimitRedis) setCommonTags(span opentracing.Span) {
+	// TODO: https://github.com/zalando/skipper/issues/2136
+	if span != nil {
+		ext.Component.Set(span, "skipper")
+		ext.SpanKind.Set(span, "client")
+		span.SetTag("ratelimit_type", c.typ)
+		span.SetTag("group", c.group)
+		span.SetTag("max_hits", c.maxHits)
+		span.SetTag("window", c.window.String())
 	}
 }
 
@@ -122,16 +118,24 @@ func (c *clusterLimitRedis) startSpan(ctx context.Context, spanName string) func
 func (c *clusterLimitRedis) AllowContext(ctx context.Context, clearText string) bool {
 	c.metrics.IncCounter(redisMetricsPrefix + "total")
 	now := time.Now()
-	finishSpan := c.startSpan(ctx, allowSpanName)
+
+	var span opentracing.Span
+	if parentSpan, ok := parentSpan(ctx); ok {
+		span = c.ringClient.StartSpan(allowSpanName, opentracing.ChildOf(parentSpan.Context()))
+		defer span.Finish()
+	}
+	c.setCommonTags(span)
 
 	allow, err := c.allow(ctx, clearText)
 	failed := err != nil
-
 	if failed {
 		allow = !c.failClosed
+		setError(span)
+	}
+	if span != nil {
+		span.SetTag("allowed", allow)
 	}
 
-	finishSpan(allow, failed)
 	c.measureQuery(allowMetricsFormat, allowMetricsFormatWithGroup, &failed, now)
 
 	if allow {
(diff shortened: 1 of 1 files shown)


ratelimit/redis.go
```diff
@@ -96,10 +96,11 @@ func (c *clusterLimitRedis) startSpan(ctx context.Context, spanName string) func
 	span.SetTag("max_hits", c.maxHits)
 	span.SetTag("window", c.window.String())
 
-	return func(failed bool) {
+	return func(allowed, failed bool) {
 		if failed {
 			ext.Error.Set(span, true)
 		}
+		span.SetTag("allowed", allowed)
```
AlexanderYastrebov, line 103: Do we want to have it unconditionally or maybe only in case when request is rejected?

AlexanderYastrebov, line 103: Do we want to have it unconditionally or maybe only in case when request is rejected?

ratelimit/redis.go
```diff
@@ -96,10 +96,11 @@ func (c *clusterLimitRedis) startSpan(ctx context.Context, spanName string) func
 	span.SetTag("max_hits", c.maxHits)
 	span.SetTag("window", c.window.String())
 
-	return func(failed bool) {
+	return func(allowed, failed bool) {
```
AlexanderYastrebov, line 99: I think we can just set the tag manually in `AllowContext` and not set it in `oldest` at all

szuecs: We don't have access to span there and then we would need to unpack it there and make startSpan() useless. I don't really care if it's in `oldest` or not, does not hurt to have a true there.

ratelimit/redis.go
szuecs: Yes, I want to have it set like it is proposed.

szuecs: Yes, I want to have it set like it is proposed.

szuecs:
:+1:

ratelimit/redis.go
AlexanderYastrebov: > We don't have access to span

we may return it along with function

ratelimit/redis.go
```diff
@@ -222,28 +224,28 @@ func (c *clusterLimitRedis) oldest(ctx context.Context, clearText string) (time.
 	res, err := c.ringClient.ZRangeByScoreWithScoresFirst(ctx, key, 0.0, float64(now.UnixNano()), 0, 1)
 
 	if err != nil {
-		finishSpan(true)
+		finishSpan(true, true)
```
AlexanderYastrebov, line 227: It is really not clear what "allowed" means in the context of "oldest", that is why I suggest we dont set it in this call path.

ratelimit/redis.go
szuecs: I think I found a better way with an optional args list that is checked.

ratelimit/redis.go
szuecs: fixed

ratelimit/redis.go
```diff
@@ -121,17 +117,27 @@ func (c *clusterLimitRedis) startSpan(ctx context.Context, spanName string) func
 func (c *clusterLimitRedis) AllowContext(ctx context.Context, clearText string) bool {
 	c.metrics.IncCounter(redisMetricsPrefix + "total")
 	now := time.Now()
-	finishSpan := c.startSpan(ctx, allowSpanName)
+
+	var span opentracing.Span
+	parentSpan, ok := parentSpan(ctx)
+	if ok {
```
AlexanderYastrebov, line 123: minor: `if parentSpan, ok := parentSpan(ctx); ok ` to narrow `parentSpan` scope

ratelimit/redis.go
```diff
@@ -77,31 +77,27 @@ func (c *clusterLimitRedis) measureQuery(format, groupFormat string, fail *bool,
 	c.metrics.MeasureSince(key, start)
 }
 
-func (c *clusterLimitRedis) startSpan(ctx context.Context, spanName string) func(bool) {
-	nop := func(bool) {}
+func parentSpan(ctx context.Context) (opentracing.Span, bool) {
 	if ctx == nil {
```
AlexanderYastrebov, line 81: Can the `ctx` be ever nil actually?

ratelimit/redis.go
```diff
@@ -121,17 +117,27 @@ func (c *clusterLimitRedis) startSpan(ctx context.Context, spanName string) func
...
-	finishSpan(failed)
-	c.measureQuery(allowMetricsFormat, allowMetricsFormatWithGroup, &failed, now)
-
 	if failed {
 		allow = !c.failClosed
+		setError(span, true)
+	}
+	if span != nil {
+		span.SetTag("allowed", allow)
+		span.Finish()
```
AlexanderYastrebov, line 136: Maybe it is better to do `defer span.Finish()` at creation site

ratelimit/redis.go
```diff
@@ -77,31 +77,27 @@ func (c *clusterLimitRedis) measureQuery(format, groupFormat string, fail *bool,
...
+		return nil, false
 	}
 
 	parentSpan := opentracing.SpanFromContext(ctx)
 	if parentSpan == nil {
-		return nop
+		return nil, false
 	}
 
-	span := c.ringClient.StartSpan(spanName, opentracing.ChildOf(parentSpan.Context()))
```
AlexanderYastrebov, line 91: I think we can keep this and maybe even common part as caller never uses parentSpan

ratelimit/redis.go
szuecs: makes sense yes

ratelimit/redis.go
szuecs: in tests yes

ratelimit/redis.go
szuecs: yes then we also have the same pattern in both places

ratelimit/redis.go
szuecs: no, I don't want this. Who starts has to finish and this we can't do here anymore.

AlexanderYastrebov: Well, it is a member method of clusterLimitRedis named `startSpan` that starts it

ratelimit/redis.go
```diff
@@ -213,37 +218,49 @@ func (c *clusterLimitRedis) Delta(clearText string) time.Duration {
 	return d
 }
 
+func setError(span opentracing.Span, b bool) {
+	if span != nil {
+		ext.Error.Set(span, b)
+	}
+}
```
AlexanderYastrebov, line 225: I think this should only set `true` so we need to drop `b bool` and call it only where it is needed

szuecs: true

ratelimit/redis.go
```diff
@@ -77,31 +77,27 @@ func (c *clusterLimitRedis) measureQuery(format, groupFormat string, fail *bool,
 	c.metrics.MeasureSince(key, start)
 }
 
-func (c *clusterLimitRedis) startSpan(ctx context.Context, spanName string) func(bool) {
-	nop := func(bool) {}
+func parentSpan(ctx context.Context) (opentracing.Span, bool) {
```
AlexanderYastrebov, line 80: I guess we could check result != nil and drop bool

ratelimit/redis.go
szuecs: I think we can fix it later, because this will likely be a bigger refactoring. https://github.com/zalando/skipper/issues/2136

ratelimit/redis.go
```diff
@@ -77,31 +77,28 @@ func (c *clusterLimitRedis) measureQuery(format, groupFormat string, fail *bool,
...
-	return func(failed bool) {
-		if failed {
-			ext.Error.Set(span, true)
-		}
+	return parentSpan, true
+}
 
-		span.Finish()
+func (c *clusterLimitRedis) setCommonTags(span opentracing.Span) {
+	// TODO: https://github.com/zalando/skipper/issues/2136
```
AlexanderYastrebov, line 94: This comment should be at  parentSpan

ratelimit/redis.go
szuecs: done

ratelimit/redis.go
szuecs: Oh man, thanks, done

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2131, opened 2022-11-03
Add EditRoute flag to RouteSRV

MustafaSaber:
Follow up on https://github.bus.zalan.do/pg9/issues/issues/33

Right now CloneRoute testcase fail but we really care now about EditRoute. So to follow up on CloneRoute check here https://github.com/zalando/skipper/pull/2132

### DIFF
--- a/routesrv/polling.go
+++ b/routesrv/polling.go
@@ -75,22 +75,9 @@ func (p *poller) poll(wg *sync.WaitGroup) {
 		span := tracing.CreateSpan("poll_routes", context.TODO(), p.tracer)
 
 		routes, err := p.client.LoadAll()
-		if p.defaultFilters != nil {
-			routes = p.defaultFilters.Do(routes)
-		}
-		if p.oauth2Config != nil {
-			routes = p.oauth2Config.NewGrantPreprocessor().Do(routes)
-		}
-		if p.editRoute != nil {
-			for _, editor := range p.editRoute {
-				routes = editor.Do(routes)
-			}
-		}
-		if p.cloneRoute != nil {
-			for _, cloner := range p.cloneRoute {
-				routes = cloner.Do(routes)
-			}
-		}
+
+		routes = p.process(routes)
+
 		routesCount = len(routes)
 
 		switch {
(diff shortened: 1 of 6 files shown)


routesrv/polling.go
```diff
@@ -79,6 +81,16 @@ func (p *poller) poll(wg *sync.WaitGroup) {
 		if p.oauth2Config != nil {
 			routes = p.oauth2Config.NewGrantPreprocessor().Do(routes)
 		}
+		if p.editRoute != nil {
```
AlexanderYastrebov, line 84: Go allows range over nil slice https://go.dev/play/p/wJ-Rcpo2jnl

routesrv/polling.go
```diff
@@ -79,6 +81,16 @@ func (p *poller) poll(wg *sync.WaitGroup) {
...
+			for _, editor := range p.editRoute {
+				routes = editor.Do(routes)
+			}
+		}
+		if p.cloneRoute != nil {
+			for _, cloner := range p.cloneRoute {
+				routes = cloner.Do(routes)
+			}
+		}
```
AlexanderYastrebov, line 93: Route processing starts to grow, lets have a helper `process` or `prepare` method  and move all preprocessors as well as `sort.SliceStable` (from below) there.

AlexanderYastrebov, line 93: Route processing starts to grow, lets have a helper `process` or `prepare` method  and move all preprocessors as well as `sort.SliceStable` (from below) there.

szuecs: what?

routesrv/polling.go
```diff
@@ -127,3 +126,22 @@ func (p *poller) poll(wg *sync.WaitGroup) {
...
+	}
+	for _, editor := range p.editRoute {
+		routes = editor.Do(routes)
+	}
+
+	for _, cloner := range p.cloneRoute {
+		routes = cloner.Do(routes)
+	}
+
+	return routes
```
AlexanderYastrebov, line 146: lets also move https://github.com/zalando/skipper/blob/566fe8c4ad7f7d46da7a2bb9b8a1f6bb75125457/routesrv/polling.go#L102-L105 here (no need to check length, sort would be no-op for empty slice)

AlexanderYastrebov:
:+1:

MustafaSaber:
👍

Issue #2138, opened 2022-11-06
Add route server options to use k8s data client serverside filter functionality


nikhil88877:
**Is your feature request related to a problem? Please describe.**
Implementation of [issue](https://github.com/zalando/skipper/issues/2045) somehow missed adding options in `routesrv` that can take advantage of serverside k8s client filtering.

**Describe the solution you would like**
Add the k8s client labelSelectors options in `routesrv` 

**Describe alternatives you've considered (optional)**
NO

**Additional context (optional)**
NO

**Would you like to work on it?**
Yes

szuecs:
this was done

PR #2139, opened 2022-11-06
allow route server to filter k8s client serverside resources using label selectors

nikhil88877:
Implementation of this [issue](https://github.com/zalando/skipper/issues/2138) 

Signed-off-by: Nikhil Goel <nikhil8877@gmail.com>

### DIFF
--- a/routesrv/options.go
+++ b/routesrv/options.go
@@ -101,6 +101,28 @@ type Options struct {
 	// used with external name services (type=ExternalName).
 	KubernetesAllowedExternalNames []*regexp.Regexp
 
+	// KubernetesIngressLabelSelectors is a map of kubernetes labels to their values that must be present on a resource to be loaded
+	// by the client. A label and its value on an Ingress must be match exactly to be loaded by Skipper.
+	// If the value is irrelevant for a given configuration, it can be left empty. The default
+	// value is no labels required.
+	KubernetesIngressLabelSelectors map[string]string
+
+	// KubernetesServicesLabelSelectors is a map of kubernetes labels to their values that must be present on a resource to be loaded
+	// by the client.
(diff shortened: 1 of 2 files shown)


nikhil88877:
Tests for `labelSelectors` features are added extensively in the k8s data client label filter [implementation](https://github.com/zalando/skipper/pull/2048). I could not reason to add more tests on the `routesrv`.
Do let me know if you guys think otherwise.

AlexanderYastrebov:
I think we need to propagate flag values here https://github.com/zalando/skipper/blob/405acf62daf17b483a4658a0a84671979aa8a961/config/config.go#L636

nikhil88877:
> I think we need to propagate flag values here

I don't fully understand your comment(reasoning). But here is my understanding based on our setup,

We don't use the executable version of `routesrv`, which uses default config values. https://github.com/zalando/skipper/blob/405acf62daf17b483a4658a0a84671979aa8a961/cmd/routesrv/main.go#L1

Rather, we define our own custom `routesrv.Options`, where I added the required k8sLabelSelectors values based on how we define the k8s resources. This also allows `labelSelectors` to be optional and cater only when you need it.


Are you suggesting to define `routesrv.Options` using some default labelSelectors values here?
https://github.com/zalando/skipper/blob/405acf62daf17b483a4658a0a84671979aa8a961/config/config.go#L636

AlexanderYastrebov:
> I don't fully understand your comment(reasoning)

I was under impression that we have cli flags for these options (introduced by https://github.com/zalando/skipper/pull/2048) but apparently we do not, therefore we don't need `routesrv` flags (and we may introduce flags later).

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2142, opened 2022-11-08
filters/auth: add tokenInfo cache

AlexanderYastrebov:
When enabled caches tokeninfo for the duration of "expires_in" field value seconds or for the duration of the configured time-to-live (TTL) if it is not zero and less than "expires_in" value.

It evicts least recently used entry when number of cached entries reaches configured size boundary.

```
$ go test github.com/zalando/skipper/filters/auth -run=NONE -bench=BenchmarkTokeninfoCache -count=1
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/filters/auth
BenchmarkTokeninfoCache/tokens=1,cacheSize=1,p=0-8               1560343               646.0 ns/op           344 B/op          3 allocs/op
BenchmarkTokeninfoCache/tokens=2,cacheSize=2,p=0-8               1500441               793.3 ns/op           344 B/op          3 allocs/op
BenchmarkTokeninfoCache/tokens=100,cacheSize=100,p=0-8           1506963               811.9 ns/op           344 B/op          3 allocs/op
BenchmarkTokeninfoCache/tokens=100,cacheSize=100,p=10000-8       1802880               655.4 ns/op           365 B/op          3 allocs/op
BenchmarkTokeninfoCache/tokens=4,cacheSize=2,p=0-8                   426           2592878 ns/op              27 B/op          0 allocs/op
BenchmarkTokeninfoCache/tokens=100,cacheSize=10,p=0-8                416           2555909 ns/op              26 B/op          0 allocs/op
```


config/config.go
```diff
@@ -465,6 +467,8 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.Oauth2ClientSecretFile, "oauth2-client-secret-file", "", "sets the path of the file containing the OAuth2 client secret associated with the oauth2-client-id, used to exchange the access code")
 	flag.StringVar(&cfg.Oauth2CallbackPath, "oauth2-callback-path", "", "sets the path where the OAuth2 callback requests with the authorization code should be redirected to")
 	flag.DurationVar(&cfg.Oauth2TokeninfoTimeout, "oauth2-tokeninfo-timeout", 2*time.Second, "sets the default tokeninfo request timeout duration to 2000ms")
+	flag.IntVar(&cfg.Oauth2TokeninfoCacheSize, "oauth2-tokeninfo-cache-size", 0, "sets the size of the tokeninfo cache")
```
szuecs, line 470: docstring should tell the metric, I guess Bytes

config/config.go
```diff
@@ -465,6 +467,8 @@ func NewConfig() *Config {
...
+	flag.DurationVar(&cfg.Oauth2TokeninfoCacheTTL, "oauth2-tokeninfo-cache-ttl", 0, "limits the lifetime of a cached tokeninfo which otherwise equals to the tokeninfo 'expires_in' field value")
```
szuecs, line 471: As a user: Does 0 mean disabled or forever?

filters/auth/authclient.go
```diff
@@ -30,6 +30,12 @@ type authClient struct {
 	cli *net.Client
 }
 
+type tokeninfoClient interface {
+	getTokeninfo(token string, ctx filters.FilterContext) (map[string]any, error)
```
szuecs, line 34: If we have `any` here the PR should also change the definition of `func (ac *authClient) getTokeninfo(token string, ctx filters.FilterContext) (map[string]interface{}, error) ` I think.

filters/auth/tokeninfo.go
```diff
@@ -28,6 +28,15 @@ type TokeninfoOptions struct {
 	Timeout      time.Duration
 	MaxIdleConns int
 	Tracer       opentracing.Tracer
+
+	// CacheSize configures the size of the tokeninfo cache.
+	// The cache evicts least recently used items first.
+	CacheSize int
```
szuecs, line 34: docstring does not tell if this is number of items or Bytes or ..

filters/auth/tokeninfocache.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	return nil
+}
+
+func (c *tokeninfoCache) tryCache(token string, info map[string]any) {
+	expiresIn := expiresIn(info)
+	if expiresIn <= 0 {
+		return
+	}
+	if c.ttl > 0 && expiresIn > c.ttl {
```
szuecs, line 92: `if expiresIn <= 0`-> expiresIn is >0 and if we check against `expiresIn > c.ttl` we don't need `c.tt > 0`

szuecs:
In general great, a couple of comments inline.
One benchmark to compare highlevel the filter with and without cache would be great to see.

filters/auth/tokeninfocache.go
AlexanderYastrebov: `c.ttl` could be zero (i.e. it is an optional config to cap token's `expires_in`)

filters/auth/authclient.go
AlexanderYastrebov: I guess we need to create a ticket to change `interface{}` to `any` across the board https://stackoverflow.com/questions/71964357/what-is-the-any-type-in-go-1-18

AlexanderYastrebov: > Another maybe more important question would be if we can re-use it for other cases like tokenintrospection, which serves a similar purpose.

It should be possible but I think it makes sense to get experience using it with tokeninfo first.

config/config.go
AlexanderYastrebov: This set maximum number of cached tokens. I've rephrased flag help messages and docstrings

config/config.go
AlexanderYastrebov: Fixed

filters/auth/tokeninfo.go
AlexanderYastrebov: Fixed

filters/auth/tokeninfo_test.go
```diff
@@ -15,176 +17,174 @@ import (
...
+	case filters.OAuthTokeninfoAnyKVName:
+		return NewOAuthTokeninfoAnyKVWithOptions(options)
+	case filters.OAuthTokeninfoAllKVName:
+		return NewOAuthTokeninfoAllKVWithOptions(options)
+	default:
+		panic("unsupported auth type: " + authType)
+	}
+}
+
 func TestOAuth2Tokeninfo(t *testing.T) {
```
AlexanderYastrebov, line 35: Refactored this test into two tests: a tokeninfo logic test and invalid arguments test

AlexanderYastrebov:
@szuecs Lets maybe have another review pass.

filters/auth/tokeninfo_test.go
```diff
@@ -15,176 +17,174 @@ import (
...
-		hasAuth:     true,
-		auth:        testToken,
-		expected:    http.StatusForbidden,
+		msg:            "oauthTokeninfoAnyScope: valid token, one valid scope, cache enabled",
+		authType:       filters.OAuthTokeninfoAnyScopeName,
+		options:        TokeninfoOptions{CacheSize: 1},
+		args:           []interface{}{testScope},
+		auth:           testToken,
+		expected:       http.StatusOK,
+		expectRequests: 1,
```
szuecs, line 179: Please add a test case what happens on error cases with enabled Cache.
We don't want to cache invalid tokens (otherwise someone can send us huge token values and we run out of mmemory).

filters/auth/tokeninfo_test.go
```diff
@@ -245,21 +230,58 @@ func TestOAuth2Tokeninfo(t *testing.T) {
...
+			args:     []interface{}{"not-matching-scope"},
+		}, {
+			msg:      "allKV(): wrong number of arguments",
+			authType: filters.OAuthTokeninfoAllKVName,
+			args:     []interface{}{"not-matching-scope"},
+		},
+	} {
+		t.Run(ti.msg, func(t *testing.T) {
+			tio := TokeninfoOptions{
+				URL:     "https://authserver.test/info",
```
szuecs, line 277: in general I think we should use `.local` as DNS suffix, for example `https://authserver.test.local/info` to omit that one day `.test` is a valid public DNS domain, but I think it's also fine like this in this PR.

filters/auth/tokeninfocache.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+type (
+	tokeninfoCache struct {
+		client tokeninfoClient
+		size   int
+		ttl    time.Duration
+		now    func() time.Time
+
+		mu    sync.Mutex
```
szuecs, line 18: Did you check if RWMutex would be helpful?

filters/auth/tokeninfocache.go
szuecs: ok

filters/auth/tokeninfocache_test.go
```diff
@@ -0,0 +1,269 @@
...
+
+	// perform third request and check cached value
+	info, err = c.getTokeninfo(token, &filtertest.Context{FRequest: &http.Request{}})
+	require.NoError(t, err)
+	assert.Equal(t, int32(1), info["requestNumber"], "expected to cache response to the first auth request")
+}
+
+var infoSink atomic.Value
+
+func BenchmarkTokeninfoCache(b *testing.B) {
```
szuecs, line 195: can you share results of this benchmark in the PR comment?

filters/auth/tokeninfo_test.go
AlexanderYastrebov: `.test` is a reserved TLD https://www.rfc-editor.org/rfc/rfc2606.html

filters/auth/tokeninfocache.go
AlexanderYastrebov: I did not, somehow I have an impression that we should not prefer it to sync.Mutex but I can not find a link why.

filters/auth/tokeninfo_test.go
AlexanderYastrebov: Makes sense, added a test to ensure we don't cache invalid token.

filters/auth/tokeninfocache.go
AlexanderYastrebov: I am not sure if the mutex would be a bottleneck here, the benchmark with high number of goroutines does not look too off from the low number:
```
BenchmarkTokeninfoCache/tokens=100,cacheSize=100,p=0-8           1506963               811.9 ns/op           344 B/op          3 allocs/op
BenchmarkTokeninfoCache/tokens=100,cacheSize=100,p=10000-8       1802880               655.4 ns/op           365 B/op          3 allocs/op
```
Also cached() method checks expiration so if we use RWMutex it would need to re-acquire write lock somehow which would complicate things.

szuecs: true

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2143, opened 2022-11-09
Fill the authuser field of the access log

haveo:
The auth filter already stores the username of the identified user in the state bag. This PR forwards this value to the access log filter by storing it in the AccessEntry structure.

It is then used in the `authuser` field of the Common Log Format formatter. It can also be accessed by other formatters (such as the JSON formatter) via the `auth-user` field of the logrus Entry.

### DIFF
--- a/logging/access.go
+++ b/logging/access.go
@@ -15,7 +15,7 @@ import (
 
 const (
 	dateFormat      = "02/Jan/2006:15:04:05 -0700"
-	commonLogFormat = `%s - - [%s] "%s %s %s" %d %d`
+	commonLogFormat = `%s - %s [%s] "%s %s %s" %d %d`
 	// format:
 	// remote_host - - [date] "method uri protocol" status response_size "referer" "user_agent"
 	combinedLogFormat = commonLogFormat + ` "%s" "%s"`
@@ -44,6 +44,9 @@ type AccessEntry struct {
 
 	// The time that the request was received.
 	RequestTime time.Time
+
+	// The id of the authenticated user
+	AuthUser *string
 }
 
 // TODO: create individual instances from the access log and
@@ -82,7 +85,7 @@ func omitWhitespace(h string) string {
 
 func (f *accessLogFormatter) Format(e *logrus.Entry) ([]byte, error) {
 	keys := []string{
-		"host", "timestamp", "method", "uri", "proto",
+		"host", "auth-user", "timestamp", "method", "uri", "proto",
 		"status", "response-size", "referer", "user-agent",
 		"duration", "requested-host", "flow-id", "audit"}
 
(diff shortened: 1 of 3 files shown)


proxy/proxy.go
```diff
@@ -1413,12 +1414,18 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		statusCode := lw.GetCode()
 
 		if shouldLog(statusCode, accessLogEnabled) {
+			var authUser *string
+			if v, ok := ctx.stateBag[log.AuthUserKey]; ok {
+				s, _ := v.(string)
+				authUser = &s
+			}
```
AlexanderYastrebov, line 1421: Lets move this after we create `logging.AccessEntry`:
```go
if v, ok := ctx.stateBag[log.AuthUserKey]; ok {
  entry.AuthUser = v.(string)
}
```

proxy/proxy.go
```diff
@@ -27,6 +27,7 @@ import (
 	al "github.com/zalando/skipper/filters/accesslog"
 	circuitfilters "github.com/zalando/skipper/filters/circuit"
 	flowidFilter "github.com/zalando/skipper/filters/flowid"
+	log "github.com/zalando/skipper/filters/log"
```
AlexanderYastrebov, line 30: Please change import name to `filterslog or `flog`

logging/access.go
```diff
@@ -82,14 +85,16 @@ func omitWhitespace(h string) string {
...
 
 	values := make([]interface{}, len(keys))
 	for i, key := range keys {
 		if s, ok := e.Data[key].(string); ok {
 			values[i] = omitWhitespace(s)
+		} else if e.Data[key] == nil {
+			values[i] = "-"
 		} else {
 			values[i] = e.Data[key]
 		}
```
AlexanderYastrebov, line 100: I think this should not be changed at all - `omitWhitespace` changes empty value for `-`

logging/access.go
```diff
@@ -162,6 +167,10 @@ func LogAccess(entry *AccessEntry, additional map[string]interface{}) {
 		"audit":          auditHeader,
 	}
 
+	if entry.AuthUser != nil {
+		logData["auth-user"] = *entry.AuthUser
+	}
+
```
AlexanderYastrebov, line 173: lets do it like it is done for other fields.
```
authUser := ""
...
if entry.Request != nil {
...
authUser = entry.AuthUser
...
logData := logrus.Fields{
...
"auth-user": authUser,
```

logging/access.go
```diff
@@ -44,6 +44,9 @@ type AccessEntry struct {
 
 	// The time that the request was received.
 	RequestTime time.Time
+
+	// The id of the authenticated user
+	AuthUser *string
```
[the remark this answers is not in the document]
haveo: So you mean storing an empty string when the key is not present in the map? I basically added the indirection to avoid having to do that, but I can of course remove it if that's more in line with common practices here.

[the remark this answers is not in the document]
AlexanderYastrebov: Yes, empty string and `omitWhitespace` takes care to swap it for `"-"`

AlexanderYastrebov:
:+1:

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2147, opened 2022-11-10
fix: Change group name in case maxShards are set

szuecs:
fix: Change group name in case maxShards are set to not cause downtime on reconfiguration of maxShards which configures the vnode shards. This will relax the config to allow double amount of requests during the reconfiguration, which seems to be a better choice than being more strict by a factor, that depends on former and new value of maxShards.

### DIFF
--- a/filters/ratelimit/ratelimit.go
+++ b/filters/ratelimit/ratelimit.go
@@ -249,7 +249,7 @@ func clusterRatelimitFilter(maxShards int, args []interface{}) (*filter, error) 
 	if keyShards > 1 {
 		f.settings = ratelimit.Settings{
 			Type:       ratelimit.ClusterServiceRatelimit,
-			Group:      group + "." + strconv.Itoa(maxShards),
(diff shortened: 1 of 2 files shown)


filters/ratelimit/ratelimit.go
```diff
@@ -248,7 +249,7 @@ func clusterRatelimitFilter(maxShards int, args []interface{}) (*filter, error)
 	if keyShards > 1 {
 		f.settings = ratelimit.Settings{
 			Type:       ratelimit.ClusterServiceRatelimit,
-			Group:      group,
+			Group:      group + "." + strconv.Itoa(maxShards),
```
AlexanderYastrebov, line 252: Should it be `keyShards` for consistency with `MaxHits`?

szuecs: fixed

AlexanderYastrebov:
:+1:

PR #2148, opened 2022-11-16
Adding JSON equivalents to the {set,append}Context{Request,Response}Header functions.

ponimas:
This PR adds  JSON equivalents to all four of the `{set,append}Context{Request,Response}Header` functions.

At the moment context vales that are set as header or appended to headers  are [stringified](https://github.com/zalando/skipper/blob/master/filters/builtin/headerfilter.go#L252).
This doesn't work very well when the value in context is a map. [Tokeninfo's authmap](https://github.com/zalando/skipper/blob/master/filters/auth/tokeninfo.go#L359) is an example.


AlexanderYastrebov:
Hello, thank you for contribution. What is the usecase for this?

filters/builtin/headerfilter.go
```diff
@@ -242,14 +284,26 @@ func valueFromContext(
 	headerName,
 	contextKey string,
 	isRequest bool,
+	jsonify bool,
```
szuecs, line 287: isJSON

AlexanderYastrebov:
@ponimas No worries, thank you.

Issue #1992, opened 2022-03-30
net.Error.Temporary() is Deprecated in Go 1.18


mikkeloscar:
In Go 1.18 the `[net.Error.Temporary()](https://pkg.go.dev/net@master#Error)` method we use in a few places is deprecated. We should consider removing it.

For now we upgraded to 1.18 ignoring the deprecation

demonCoder95:
I believe given the package documentation, `Temporary` can be replaced by `Timeout` since most temporary errors are just timeout. This can be fixed by a simple refactor by replacing the method with the new one.

szuecs:
We should review our production logs for the relevant error strings and check if we see something or if you are right with your replacement suggestion.

demonCoder95:
That's a good idea! I can take a look to double-check before we move ahead with the PR.

gmikhaile:
Hey!
Is Timeout() appropriate replacement for Temporary()?

szuecs:
> > That's a good id ea! I can take a look to double-check before we move ahead with the PR.
> 
> Hey! Is Timeout() appropriate replacement for Temporary()?

no they are both different in net.Error interface:
```
/usr/share/go/src]% grep -rnE 'type Error interface' -A 10 net
net/net.go:393:type Error interface {
net/net.go-394- error
net/net.go-395- Timeout() bool // Is the error a timeout?
net/net.go-396-
net/net.go-397- // Deprecated: Temporary errors are not well-defined.
net/net.go-398- // Most "temporary" errors are timeouts, and the few exceptions are surprising.
net/net.go-399- // Do not use this method.
net/net.go-400- Temporary() bool
net/net.go-401-}
```

I think there is a Go issue to finally get rid of that, but because of compatibility nobody wants to delete it, I guess.

PR #2149, opened 2022-11-16
Refactor net.Error.Temporary

demonCoder95:
Provides a fix for https://github.com/zalando/skipper/issues/1992.


demonCoder95:
`Temporary` based tests were also dropped as no longer relevant.

proxy/proxy.go
```diff
@@ -966,8 +966,7 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 				status = http.StatusServiceUnavailable
 			}
 			p.tracing.setTag(ctx.proxySpan, HTTPStatusCodeTag, uint16(status))
-			//lint:ignore SA1019 Temporary is deprecated in Go 1.18, but keep it for now (https://github.com/zalando/skipper/issues/1992)
-			return nil, &proxyError{err: fmt.Errorf("net.Error during backend roundtrip to %s: timeout=%v temporary='%v': %w", req.URL.Host, nerr.Timeout(), nerr.Temporary(), err), code: status}
+			return nil, &proxyError{err: fmt.Errorf("net.Error during backend roundtrip to %s: timeout=%v : %w", req.URL.Host, nerr.Timeout(), err), code: status}
```
AlexanderYastrebov, line 969: Redundant whitespace before colon: `timeout=%v : %w"`

AlexanderYastrebov: Also maybe we can change `net.Error during backend roundtrip ...` to `%T during backend roundtrip ...` to get e.g. this output `*net.DNSError during backend roundtrip ...`

https://go.dev/play/p/QyCCO_4d5yL

demonCoder95: Nice idea with %T 🙂 Fixed in the next patchset!

proxy/proxy.go
```diff
@@ -966,8 +966,7 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 				status = http.StatusServiceUnavailable
 			}
 			p.tracing.setTag(ctx.proxySpan, HTTPStatusCodeTag, uint16(status))
-			//lint:ignore SA1019 Temporary is deprecated in Go 1.18, but keep it for now (https://github.com/zalando/skipper/issues/1992)
-			return nil, &proxyError{err: fmt.Errorf("net.Error during backend roundtrip to %s: timeout=%v temporary='%v': %w", req.URL.Host, nerr.Timeout(), nerr.Temporary(), err), code: status}
+			return nil, &proxyError{err: fmt.Errorf("%T during backend roundtrip to %s: timeout=%v: %w", nerr, req.URL.Host, nerr.Timeout(), err), code: status}
```
szuecs, line 969: Would `%T` show the interface? Not that we need a fixed `net.Error` here but this log message will of course change.

queuelistener/listener_test.go
```diff
@@ -353,28 +339,6 @@ func TestInterface(t *testing.T) {
 		}
 	})
 
-	t.Run("wrapped listener returns temporary error, logs and retries", func(t *testing.T) {
```
szuecs, line 356: Do we test timeout errors instead?

queuelistener/listener_test.go
```diff
@@ -21,22 +20,13 @@ type testConnection struct {
...
+	addr            net.Addr
+	conns           chan *testConnection
 }
 
-type testError struct{}
-
-var errTemporary testError
-
-func (err testError) Error() string   { return "test error" }
-func (err testError) Timeout() bool   { return false }
```
szuecs, line 37: Should not we test `Timeout()` ?

queuelistener/listener_test.go
```diff
@@ -982,27 +946,6 @@ func TestTeardown(t *testing.T) {
 }
 
 func TestMonitoring(t *testing.T) {
-	t.Run("logs the temporary errors", func(t *testing.T) {
```
szuecs, line 985: We should test timeouts

proxy/proxy.go
demonCoder95: No it won't. It'll just show `net.Error` as type.

demonCoder95:
I'll add unit tests for `Timeout()`.

queuelistener/listener_test.go
demonCoder95: Test added.

queuelistener/listener_test.go
demonCoder95: Yes. Test added in the latest patch.

queuelistener/listener_test.go
demonCoder95: Test added in the latest patch.

demonCoder95:
I refactored the temporary error tests for timeout error and it seems the tests were broken. Any ideas on how to fix that? Also, I'll take a look at some production logs, as suggested on the linked issue to check for appropriate error strings.

proxy/proxy.go
szuecs: ok

queuelistener/listener_test.go
```diff
@@ -21,21 +21,20 @@ type testConnection struct {
...
 type testError struct{}
 
-var errTemporary testError
+var errTimeout testError
 
-func (err testError) Error() string   { return "test error" }
-func (err testError) Timeout() bool   { return false }
-func (err testError) Temporary() bool { return true }
+func (err testError) Error() string { return "test error" }
+func (err testError) Timeout() bool { return false }
```
szuecs, line 37: Maybe the tests are waiting for Timeout() returning true?
I see that you wait for some log message but maybe it's the condition that is now different, because Temporary was "true" and Timeout is "false"

PR #2165, opened 2022-12-06
routing: detect and avoid loops

AlexanderYastrebov:
Proxy keeps track of traversed routes on each loopback and Matcher ignores already traversed routes.

Consider example routing table:
```
bin/skipper -inline-routes='r1: Path("/test") && Weight(1) -> appendRequestHeader("x-route", "r1") -> <loopback>; r2: Path("/test") && Weight(2) -> appendRequestHeader("x-route", "r2") -> <loopback>; r3: Path("/test") -> logHeader() -> inlineContent("ok") -> <shunt>;'
```
The request would traverse r2, then r1 and finally r3.

- [ ] Tests please

### DIFF
--- a/routing/loopback.go
+++ b/routing/loopback.go
@@ -1,0 +1,39 @@
+package routing
+
+import (
+	"context"
+	"net/http"
+)
+
+type (
+	routeChain struct {
+		head *Route
+		tail *routeChain
+	}
+
+	contextKey struct {
+		name string
+	}
+)
+
+var routeChainContextKey = &contextKey{"route-chain-context-key"}
+
+// WithRoute returns a shallow copy of request with route stored in its context.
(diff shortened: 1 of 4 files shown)


routing/loopback.go
```diff
@@ -0,0 +1,39 @@
+package routing
+
+import (
+	"context"
+	"net/http"
+)
+
+type (
+	routeChain struct {
```
RomanZavodskikh, line 9: Would you like to limit number of routes in the loop?

From my side, it looks like there should be such a limit not to let this routeChain OOM-kill skipper pod.

AlexanderYastrebov: Yeah, maybe we can re-purpose existing max loopbacks limit config for this.

PR #2166, opened 2022-12-10
proxy: use errors.Is instead of string match

AlexanderYastrebov:
See https://github.com/golang/go/issues/4373


szuecs:
👍

AlexanderYastrebov:
:+1:

PR #2167, opened 2022-12-11
Test/config

szuecs:
increase test coverage in config package by using config test pattern.

### DIFF
--- a/config/breakerflags_test.go
+++ b/config/breakerflags_test.go
@@ -42,10 +42,11 @@ func Test_breakerFlags_String(t *testing.T) {
 
 func Test_breakerFlags_Set(t *testing.T) {
 	tests := []struct {
-		name    string
-		args    string
-		wantErr bool
-		want    circuit.BreakerSettings
+		name      string
+		args      string
+		wantErr   bool
+		errString string
+		want      circuit.BreakerSettings
 	}{
 		{
 			name:    "test breaker settings",
@@ -73,9 +74,24 @@ func Test_breakerFlags_Set(t *testing.T) {
 			},
 		},
 		{
-			name:    "test breaker settings with wrong window",
-			args:    "type=consecutive,window=4s,host=example.com,timeout=3s,half-open-requests=3,idle-ttl=5s",
-			wantErr: true,
+			name:    "test breaker settings with failures",
+			args:    "type=consecutive,window=4,host=example.com,timeout=3s,half-open-requests=3,idle-ttl=5s,failures=2",
+			wantErr: false,
+			want: circuit.BreakerSettings{
+				Type:             circuit.ConsecutiveFailures,
+				Host:             "example.com",
+				Timeout:          3 * time.Second,
+				HalfOpenRequests: 3,
+				IdleTTL:          5 * time.Second,
+				Window:           4,
+				Failures:         2,
+			},
+		},
+		{
+			name:      "test breaker settings with wrong window",
+			args:      "type=consecutive,window=4s,host=example.com,timeout=3s,half-open-requests=3,idle-ttl=5s",
+			wantErr:   true,
+			errString: `strconv.Atoi: parsing "4s": invalid syntax`,
 		},
 		{
 			name:    "test breaker settings failurerate",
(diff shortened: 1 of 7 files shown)


config/breakerflags_test.go
```diff
@@ -124,6 +172,8 @@ func Test_breakerFlags_Set(t *testing.T) {
 				if cmp.Equal(b[0], tt.want) == false {
 					t.Errorf("breakerFlags.Set() got v, want v, %v", cmp.Diff(b[0], tt.want))
 				}
+			} else if tt.errString != err.Error() {
+				t.Errorf("Failed to get error string: %v", err)
```
AlexanderYastrebov, line 176: Maybe it makes sense to print tt.errString as well

config/config.go
```diff
@@ -574,30 +594,14 @@ func (c *Config) Parse() error {
...
-	}
-
-	c.ApplicationLogLevel = logLevel
-	c.KubernetesPathMode = kubernetesPathMode
-	c.KubernetesEastWestRangePredicates = kubernetesEastWestRangePredicates
-	c.HistogramMetricBuckets = histogramBuckets
+	c.ApplicationLogLevel, _ = log.ParseLevel(c.ApplicationLogLevelString)
+	c.KubernetesPathMode, _ = kubernetes.ParsePathMode(c.KubernetesPathModeString)
+	c.KubernetesEastWestRangePredicates, _ = eskip.ParsePredicates(c.KubernetesEastWestRangePredicatesString)
+	c.HistogramMetricBuckets, _ = c.parseHistogramBuckets()
```
AlexanderYastrebov, line 604: I think it was fine, maybe just move assignments (e.g. `c.ApplicationLogLevel = logLevel`) right after parsing error checks. This way we parse stuff twice but I guess its not a big deal.

config/listflag_test.go
```diff
@@ -107,4 +107,16 @@ func TestListFlag(t *testing.T) {
...
+		const input = `---
+foo: foo=bar
+bar: baz
+`
+		current := commaListFlag()
+		if err := yaml.Unmarshal([]byte(input), current); err == nil {
+			t.Errorf("Failed to get error from Unmarshal() for invalid input: %q", input)
+		}
+
+	})
```
AlexanderYastrebov, line 121: Maybe use `this is not yaml` as an input to make it obvious what is the tested problem

config/mapflags_test.go
```diff
@@ -62,6 +65,66 @@ func Test_mapFlags_Set(t *testing.T) {
...
+		t.Error("Failed to get nil error or lazy init if m is nil")
+	}
+}
+
+func Test_mapFlags_UnmarshalYAML(t *testing.T) {
+	tests := []struct {
+		name    string
+		yml     string
+		wantErr bool
+		want    map[string]string
```
AlexanderYastrebov, line 90: I find it easier to write two table tests: one for invalid input with expected error and one for valid inputs testing for no error

config/routechangerconfig_test.go
```diff
@@ -71,3 +99,22 @@ func Test_routeChangerConfigSeveralItems(t *testing.T) {
...
+func TestRouteChangerYaml(t *testing.T) {
+	m := routeChangerConfig{}
+	err := yaml.Unmarshal([]byte(`"#Source[(](.*)[)]#ClientIP($1)#"`), &m)
+	if err != nil {
+		t.Errorf("Failed to yaml unmarshal: %v", err)
+	}
+	if len(m) != 1 {
+		t.Errorf("Failed to get correct routechanger config: %d != 1", len(m))
+	}
+}
```
AlexanderYastrebov, line 112: I think I've used (misused?) `Test_NewConfig` to validate yaml parsing https://github.com/zalando/skipper/pull/1933

AlexanderYastrebov:
Look good to me, feel free to address minor comments if you like.

AlexanderYastrebov:
:+1:

config/breakerflags_test.go
szuecs: fixed

config/config.go
szuecs: The problem is that we can not check the validations because of `flag.Parse` and this is only happening once at startup so I think it's fine to do this twice.

config/listflag_test.go
szuecs: fixed

config/routechangerconfig_test.go
szuecs: I think the test is fine, but testing different cases to have more/all call paths covered requires sometimes more special tests.

config/mapflags_test.go
szuecs: I thought I need more tests, so that's why it's a table.

config/config.go
AlexanderYastrebov: Hm, I see not that you test `Validate`. Do we need it to be exported? There is also a chicken-and-egg problem here - we need to parse flags to get values we validate.

The global `flag.Parse()` is really unfortunate, maybe we could fix it using custom `flag.FlagSet` that we can stuff for testing, see https://eli.thegreenplace.net/2020/testing-flag-parsing-in-go-programs/

I did check if we can keep it backwards compatible for lib users.

szuecs: Validate->validate

AlexanderYastrebov:
[TestRedisClient](https://github.com/zalando/skipper/blob/d2fdca2868281a42f522c1b512f09322cadf7508/net/redisclient_test.go#L87) is flaky

szuecs:
:+1:

config/config.go
```diff
@@ -546,6 +546,26 @@ func NewConfig() *Config {
 	return cfg
 }
 
+func validate(c *Config) error {
```
AlexanderYastrebov, line 549: Could also be `func (c *Config) validate() error` I guess

AlexanderYastrebov:
:+1:

PR #2168, opened 2022-12-14
filters/auth: handle SecretProvider errors in OAuthConfig

AlexanderYastrebov:

### DIFF
--- a/filters/auth/grantconfig.go
+++ b/filters/auth/grantconfig.go
@@ -15,7 +15,6 @@ import (
 
 type OAuthConfig struct {
 	initialized bool
-	initErr     error
 	flowState   *flowState
 
 	// TokeninfoURL is the URL of the service to validate OAuth2 tokens.
@@ -54,15 +53,15 @@ type OAuthConfig struct {
 
 	// ClientIDFile, the path to the file containing the OAuth2 client id of
 	// the current service, used to exchange the access code. Must be set if
-	// ClientID is not provided.
+	// ClientID is not provided. Requires SecretsProvider and will be added to it.
 	ClientIDFile string
 
 	// ClientSecretFile, the path to the file containing the secret associated
 	// with the ClientID, used to exchange the access code. Must be set if
-	// ClientSecret is not provided.
+	// ClientSecret is not provided. Requires SecretsProvider and will be added to it.
 	ClientSecretFile string
 
-	// SecretsProvider is used to read client ID and secret values from the
+	// SecretsProvider is used to read ClientIDFile and ClientSecretFile from the
 	// file system. Supports secret rotation.
 	SecretsProvider secrets.SecretsProvider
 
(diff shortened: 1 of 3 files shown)


filters/auth/grantconfig.go
```diff
@@ -109,6 +108,8 @@ var (
 	ErrMissingSecretFile      = errors.New("missing secret file")
 	ErrMissingTokeninfoURL    = errors.New("missing tokeninfo URL")
 	ErrMissingProviderURLs    = errors.New("missing provider URLs")
+
+	errMissingSecretsProvider = errors.New("missing secrets provider")
```
szuecs, line 112: Why is this not public, which could be used and checked by skipper as library users?

filters/auth/grantconfig_test.go
```diff
@@ -0,0 +1,101 @@
...
+func TestOAuthConfigInit(t *testing.T) {
+	tempDir := t.TempDir()
+	missingIDFile := tempDir + "/missing-id"
+	missingSecretFile := tempDir + "/missing-secret"
+	existingFile := tempDir + "/existing"
+	require.NoError(t, os.WriteFile(existingFile, []byte("acontent"), 0644))
+
+	for i, tc := range []struct {
+		config        *auth.OAuthConfig
+		expectedError string
```
szuecs, line 23: this should be the error that we expect and not a string :)

filters/auth/grantconfig.go
AlexanderYastrebov: I think making those exported was a mistake.
These errors are returned only from `Init()`, how do you think it would be useful for users?

filters/auth/grantconfig_test.go
szuecs: It's called from `skipper` package and lib users can test the error if they like. I think this should be good for all cases when we return an error. Much better then the `net` stdlib that you can test only by string compare.

filters/auth/grantconfig.go
szuecs: Init is called by skipper.run() which is returned by skipper.Run() that is called by library users.

AlexanderYastrebov: Ok, but after this change Init does not always return exported errors but also os.Stat error when file is missing

szuecs: That's true but we can not do anything about it or wrap it, but this does not help.
I think client code could do some:
```
err := skipper.Run(opt)
switch err
case ErrSkipper1:
..
case ErrSkipper2,ErrSkipper3:
..
default:
//handle stdlib error
}
return
```

Of course in most cases they will just log.Fatalf and don't care about the types, but as library we should return typed errors as the best we can.

AlexanderYastrebov: I've made it exported, does not hurt as it won't be use anyway :)

szuecs:
👍

AlexanderYastrebov:
:+1:

PR #2170, opened 2022-12-20
proxy: fix goroutine leaks in tests

AlexanderYastrebov:

### DIFF
--- a/net/redisclient.go
+++ b/net/redisclient.go
@@ -359,7 +359,9 @@ func (r *RedisRingClient) Close() {
 	r.once.Do(func() {
 		r.closed = true
 		close(r.quit)
-		r.ring.Close()
+		if r.ring != nil {
+			r.ring.Close()
+		}
 	})
 }
 
--- a/proxy/fadeintesting_test.go
+++ b/proxy/fadeintesting_test.go
@@ -40,6 +40,7 @@ const (
 
 type fadeInDataClient struct {
 	reset, update chan []*eskip.Route
+	quit          chan struct{}
 }
 
 type fadeInBackendInstance struct {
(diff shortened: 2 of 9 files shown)


net/redisclient.go
```diff
@@ -359,7 +359,9 @@ func (r *RedisRingClient) Close() {
 	r.once.Do(func() {
 		r.closed = true
 		close(r.quit)
-		r.ring.Close()
+		if r.ring != nil {
```
szuecs, line 362: Ah re-introduced :)

proxy/fadeintesting_test.go
```diff
@@ -107,18 +109,33 @@ func createDataClient(r ...*eskip.Route) fadeInDataClient {
...
+		return r, nil
+	case <-c.quit:
+		return nil, nil
+	}
 }
 
 func (c fadeInDataClient) LoadUpdate() ([]*eskip.Route, []string, error) {
-	return <-c.update, nil, nil
+	select {
+	// hm, blocking dataclient?
```
szuecs, line 129: If you have 2 dataclients in use, what if only one fails such that 50% of your routes are not there. Do you want to startup normally or block and postpone/fail to start?
I think that's why we want to block in general.

szuecs:
:+1:

net/redisclient.go
AlexanderYastrebov: Yeah, I decided to plug nil check as quick fix but we should update ratelimit initialization such that it does not create redis client with nil ring

AlexanderYastrebov:
:+1:

proxy/fadeintesting_test.go
AlexanderYastrebov: Its a bit weird as we have polling cycle per dataclient but it works both ways

PR #2178, opened 2023-01-03
refactor: drop else, which is not needed

szuecs:

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -930,14 +930,12 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 
 	ctx.proxySpan.LogKV("http_roundtrip", EndEvent)
 	if err != nil {
-
 		if errors.Is(err, ErrBlocked) {
 			p.tracing.setTag(ctx.proxySpan, BlockTag, true)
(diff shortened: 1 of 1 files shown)


proxy/proxy.go
```diff
@@ -930,14 +930,12 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
...
 	if err != nil {
-
 		if errors.Is(err, ErrBlocked) {
 			p.tracing.setTag(ctx.proxySpan, BlockTag, true)
 			p.tracing.setTag(ctx.proxySpan, HTTPStatusCodeTag, uint16(http.StatusBadRequest))
 			return nil, &proxyError{err: err, code: http.StatusBadRequest}
-		} else {
-			p.tracing.setTag(ctx.proxySpan, ErrorTag, true)
 		}
+		p.tracing.setTag(ctx.proxySpan, ErrorTag, true)
```
AlexanderYastrebov, line 938: Maybe move it right after `if err != nil`

szuecs: no if we block we consider this not as error in processing, so no error tag

AlexanderYastrebov: Ok, I see

PR #2179, opened 2023-01-03
expose cookie validity as config - fix #2157

### LINKED ISSUE #2157: cookie lifetime
I use skipper to do OIDC pre-auth with the filter oauthOidcAllClaims. I noticed that the skipper cookie has a lifetime of 1 hour. For a Single-Page-Application the following situation happens: If a user let the site open for over 1 hour skipper tries to refresh the cookie and does a redirect to the auth-service. For normal apps/sites this works (http-requests). But for the XHR-Requests of javascript it doesn't work. In the end this leads to a failure on the Single-Page-Application. Only a full reload of the page helps to solve it. Are there some ideas how to fix that? Is it e.g. possible to set a longer lifetime for the skipper cookie? Or are there some parameters/filters to avoid such a situation?

I would prefer to set the cookie lifetime globally via flag.

szuecs:
expose cookie validity as config
fix https://github.com/zalando/skipper/issues/2157

TODO:
- [x] tests
- [x] exp integration (checked because done already)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -203,6 +203,7 @@ type Config struct {
 	Oauth2TokenCookieName           string        `yaml:"oauth2-token-cookie-name"`
 	WebhookTimeout                  time.Duration `yaml:"webhook-timeout"`
 	OidcSecretsFile                 string        `yaml:"oidc-secrets-file"`
+	OIDCCookieValidity              time.Duration `yaml:"oidc-cookie-validity"`
 	OidcDistributedClaimsTimeout    time.Duration `yaml:"oidc-distributed-claims-timeout"`
 	CredentialPaths                 *listFlag     `yaml:"credentials-paths"`
 	CredentialsUpdateInterval       time.Duration `yaml:"credentials-update-interval"`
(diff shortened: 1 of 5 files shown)


szuecs:
:+1:

config/config.go
```diff
@@ -203,6 +203,7 @@ type Config struct {
 	Oauth2TokenCookieName           string        `yaml:"oauth2-token-cookie-name"`
 	WebhookTimeout                  time.Duration `yaml:"webhook-timeout"`
 	OidcSecretsFile                 string        `yaml:"oidc-secrets-file"`
+	OIDCCookieValidity              time.Duration `yaml:"oidc-cookie-validity"`
```
AlexanderYastrebov, line 206: Shouldn't we add a flag for it?

szuecs: oops, yes

szuecs: fixed

MustafaSaber:
👍

szuecs:
:+1:

PR #2189, opened 2023-01-05
ensureutf8 on all strings that are passed to tracer

szuecs:


proxy/proxy.go
```diff
@@ -1489,7 +1489,7 @@ func (p *Proxy) setCommonSpanInfo(u *url.URL, r *http.Request, s ot.Span) {
 		setTag(s, HTTPPathTag, u.Path).
```
AlexanderYastrebov, line 1489: I guess this was missing

AlexanderYastrebov:
:+1:

proxy/proxy.go
szuecs: I tested that if you parse as we do with url.Parse all attributes are encoded correctly (path,query,hostname), the same for NewRequest(), so I don't know. I will try to remove the setTag encoder to test another assumption

szuecs:
:+1:

PR #2193, opened 2023-01-08
eskip: reuse lexer and parser

AlexanderYastrebov:
```
name               old time/op    new time/op    delta
ParsePredicates-8    4.80µs ± 4%    3.11µs ± 8%  -35.27%  (p=0.000 n=10+9)

name               old alloc/op   new alloc/op   delta
ParsePredicates-8    6.22kB ± 0%    0.74kB ± 0%  -88.03%  (p=0.000 n=10+10)

name               old allocs/op  new allocs/op  delta
ParsePredicates-8      37.0 ± 0%      35.0 ± 0%   -5.41%  (p=0.000 n=10+10)
```

### DIFF
--- a/eskip/eskip.go
+++ b/eskip/eskip.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"regexp"
 	"strings"
+	"sync"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/zalando/skipper/filters/flowid"
@@ -591,11 +592,28 @@ func newRouteDefinition(r *parsedRoute) (*Route, error) {
 	return rd, err
 }
 
+type eskipLexParser struct {
+	lexer  eskipLex
+	parser eskipParserImpl
+}
+
+var parserPool sync.Pool
+
 // executes the parser.
 func parse(code string) ([]*parsedRoute, error) {
-	l := newLexer(code)
-	eskipParse(l)
-	return l.routes, l.err
+	lp, ok := parserPool.Get().(*eskipLexParser)
+	if !ok {
+		lp = &eskipLexParser{}
+	} else {
+		*lp = eskipLexParser{}
+	}
+	defer parserPool.Put(lp)
+
+	lp.lexer.init(code)
+
+	lp.parser.Parse(&lp.lexer)
+
+	return lp.lexer.routes, lp.lexer.err
 }
 
 func partialRouteToRoute(format, p string) string {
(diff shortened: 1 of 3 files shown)


eskip/eskip.go
```diff
@@ -591,11 +592,28 @@ func newRouteDefinition(r *parsedRoute) (*Route, error) {
 	return rd, err
 }
 
+type eskipLexParser struct {
+	lexer  eskipLex
+	parser eskipParserImpl
+}
```
AlexanderYastrebov, line 598: Parser consists of 1+16 `eskipSymType` structs
https://github.com/zalando/skipper/blob/77f4c5e868917a90a8823d734e7186fb7a34140d/eskip/parser.go#L200-L204
which is quite large
https://github.com/zalando/skipper/blob/77f4c5e868917a90a8823d734e7186fb7a34140d/eskip/parser.go#L21-L44

A new lexer is always used with the parser so it makes sense to pool them together.

`eskipSymType` is in fact a union type, i.e. only one of the fields is used on each parse step.
I've experimented with implementing it as a 
```yacc
%union {
	val interface{}
}
```
and a set of converters like
```go
func (u *eskipSymType) token() string        { return u.val.(string) }
func (u *eskipSymType) stringvals() []string { return u.val.([]string) }
func (u *eskipSymType) route() *parsedRoute { return u.val.(*parsedRoute) }
...
```
similar to how it is done in [cockroachdb](https://github.com/cockroachdb/cockroach/blob/c21c90f93219b857858518d25a8bc061444d573c/pkg/sql/parser/sql.y#L205-L249).

This brings some improvement but requires quite heavy rewrite of parser.y grammar rules.

I have also couple of lexer refactorings that eliminate inefficiencies and reduce allocations but not as much as pooling lexer and parser. These could be done in subsequent PRs.

eskip/eskip.go
```diff
@@ -591,11 +592,28 @@ func newRouteDefinition(r *parsedRoute) (*Route, error) {
...
 // executes the parser.
 func parse(code string) ([]*parsedRoute, error) {
-	l := newLexer(code)
-	eskipParse(l)
-	return l.routes, l.err
+	lp, ok := parserPool.Get().(*eskipLexParser)
+	if !ok {
+		lp = &eskipLexParser{}
+	} else {
+		*lp = eskipLexParser{}
```
szuecs, line 608: Didn't we get the lp already? Why do we need to "reset" the value?

szuecs:
:+1:

AlexanderYastrebov:
:+1:

Issue #2197, opened 2023-01-13
filter swarm: panics and route is down


universam1:
**Describe the bug**

as requested https://github.com/zalando/skipper/issues/2196#issuecomment-1380455684 filing a bug for filter swarm

[v0.14.43](https://github.com/zalando/skipper/releases/tag/v0.14.43) causes routes to fail when runtime error happens with the swarm filter:

```
[APP]time="2023-01-12T14:07:06Z" level=error msg="cd clusterRatelimit failed to share value: cannot broadcast message, swarm is nil"

[APP]time="2023-01-12T14:17:20Z" level=error msg="error while proxying after 96.658µs, route kube_rg__redmine__cd_api_query__all__0_0 with backend lb ://, status code 500: panic caused by: runtime error: invalid memory address or nil pointer dereference, remote host: reducted, request: \"GET /issues.json HTTP/1.1\", host: api-reducted, user agent: \"curl/7.68.0\""
```

**Workaround**

Downgrade to 0.14.42, route is still served.

**Describe the bug**

A swarm error causes the routes to fail. Prior to v0.14.43 an error was thrown but the route was served. Now we have a service interruption.

**To Reproduce**
Add context, for example:
1. route or ingress definition 
```
    filters:
    - clusterRatelimit("dummy", 2, "1s")
```

**Expected behavior**

Throw an error or better no error

universam1:
Found this stacktrace log, thanks for the hint that this is only printed once:

```
[APP]time="2023-01-12T14:07:15Z" level=error msg="stacktrace of panic caused by: runtime error: invalid memory address or nil pointer dereference:
goroutine 167 [running]:
github.com/zalando/skipper/proxy.stack()
	/workspace/proxy/proxy.go:1021 +0x65
github.com/zalando/skipper/proxy.(*Proxy).do.func1.1()
	/workspace/proxy/proxy.go:1033 +0x38
golang.org/x/time/rate.(*Sometimes).Do(0xc0000d4e60, 0xc003d6af58)
	/go/pkg/mod/golang.org/x/time@v0.3.0/rate/sometimes.go:63 +0xf9
github.com/zalando/skipper/proxy.(*Proxy).do.func1()
	/workspace/proxy/proxy.go:1032 +0x95
panic({0xfef9c0, 0x1aadc00})
	/usr/local/go/src/runtime/panic.go:890 +0x262
[...]

PR #2198, opened 2023-01-16
swarm nil pointer dereference panic

universam1:
In case the Swarm is not operative, an error is logged.

However, that error is only logged and proceeded unverified.

This leads eventually into a panic of a nil pointer dereference.

fixes: #2197

Signed-off-by: Samuel Lang <gh@lang-sam.de>

### DIFF
--- a/ratelimit/swim.go
+++ b/ratelimit/swim.go
@@ -95,7 +95,8 @@ func (c *clusterLimitSwim) Allow(clearText string) bool {
 	_ = c.local.Allow(s) // update local rate limit
 
 	if err := c.swarm.ShareValue(key, t0); err != nil {
-		log.Errorf("%s clusterRatelimit failed to share value: %v", c.group, err)
+		log.Errorf("clusterRatelimit '%s' disabled, failed to share value: %v", c.group, err)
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

ratelimit/swim.go
```diff
@@ -95,7 +95,8 @@ func (c *clusterLimitSwim) Allow(clearText string) bool {
 	_ = c.local.Allow(s) // update local rate limit
 
 	if err := c.swarm.ShareValue(key, t0); err != nil {
-		log.Errorf("%s clusterRatelimit failed to share value: %v", c.group, err)
+		log.Errorf("clusterRatelimit '%s' disabled, failed to share value: %v", c.group, err)
+		return true // unsafe to continue otherwise
```
AlexanderYastrebov, line 99: For the ref: we may implement mode support sometime in the future https://github.com/zalando/skipper/pull/2097

AlexanderYastrebov:
Should we have a test that demonstrates panic and that fix is effective?

AlexanderYastrebov:
:+1:

PR #2199, opened 2023-01-16
Moved logging in admissionControl under the right if condition

RomanZavodskikh:
Text of the log message makes me think that such log message should not be printer not always but only in case where calculated RPS is less than required minimum.


szuecs:
:+1:

MustafaSaber:
👍

RomanZavodskikh:
:+1:

PR #2202, opened 2023-01-17
Create `-sort-predicates` flag for eskip

MustafaSaber:

### DIFF
--- a/cmd/eskip/args.go
+++ b/cmd/eskip/args.go
@@ -39,6 +39,7 @@ const (
 	appendFiltersFlag  = "append"
 	appendFileFlag     = "append-file"
 	prettyFlag         = "pretty"
+	sortPredicatesFlag = "sort-predicates"
 	indentStrFlag      = "indent"
 	jsonFlag           = "json"
 
@@ -75,6 +76,7 @@ var (
 	appendFiltersArg  string
 	appendFileArg     string
 	pretty            bool
+	sortPredicates    bool
 	indentStr         string
 	printJson         bool
 )
@@ -110,6 +112,7 @@ func initFlags() {
 	flags.StringVar(&appendFileArg, appendFileFlag, "", appendFileUsage)
 
 	flags.BoolVar(&pretty, prettyFlag, false, prettyUsage)
+	flags.BoolVar(&sortPredicates, sortPredicatesFlag, false, sortPredicateUsage)
 	flags.StringVar(&indentStr, indentStrFlag, "  ", indentStrUsage)
 	flags.BoolVar(&printJson, jsonFlag, false, jsonUsage)
 }
--- a/cmd/eskip/load.go
+++ b/cmd/eskip/load.go
@@ -126,7 +126,8 @@ func printCmd(a cmdArgs) error {
 			}
 		}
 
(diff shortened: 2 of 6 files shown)


cmd/eskip/load.go
```diff
@@ -126,7 +127,14 @@ func printCmd(a cmdArgs) error {
 			}
 		}
 
+		if sortPredicate {
+			for _, route := range lr.routes {
+				sort.Slice(route.Predicates, func(i, j int) bool { return i < j })
```
RomanZavodskikh, line 132: It seems to me such `sort.Slice` call does not actually mutate `route.Predicates` slice.
https://stackoverflow.com/questions/37695209/golang-sort-slice-ascending-or-descending 
As far as I see here to sort a slice with `sort.Slice` we need to use `i`, `j` variables as indexes to this slice. However, we compare here only indexes between each other without comparison *elements* of a slice.

RomanZavodskikh, line 132: Should there be some unit tests to check whether such sort actually works?

MustafaSaber: yes, that's in progress.

cmd/eskip/args.go
```diff
@@ -39,6 +39,7 @@ const (
 	appendFiltersFlag  = "append"
 	appendFileFlag     = "append-file"
 	prettyFlag         = "pretty"
+	sortPredicateFlag  = "sort-predicates"
```
AlexanderYastrebov, line 42: Lest make it plural `sortPredicatesFlag`

eskip/string.go
```diff
@@ -78,7 +80,7 @@ func argsString(args []interface{}) string {
 	return strings.Join(sargs, ", ")
 }
 
-func (r *Route) predicateString() string {
+func (r *Route) predicateString(prettyPrintInfo PrettyPrintInfo) string {
```
AlexanderYastrebov, line 83: We do not need whole PrettyPrintInfo here (it is also passed by value - creates a copy), lets just pass bool.

cmd/eskip/load.go
MustafaSaber: Implemented in `eskip/string_test.go`

eskip/string.go
```diff
@@ -107,9 +109,20 @@ func (r *Route) predicateString() string {
...
+		for _, p := range predicatesCopy {
+			if p.Name != "Any" {
+				predicates = appendFmt(predicates, "%s(%s)", p.Name, argsString(p.Args))
+			}
+		}
+	} else {
+		for _, p := range r.Predicates {
+			if p.Name != "Any" {
+				predicates = appendFmt(predicates, "%s(%s)", p.Name, argsString(p.Args))
+			}
```
AlexanderYastrebov, line 125: Lets have main logic loop outside of the condition something like:
```go
rp := r.Predicates
if sortPredicates {
    rp = make([]*Predicate, len(r.Predicates))
    copy(rp, r.Predicates)
    sort.Slice(rp, func(i, j int) bool { return rp[i].String() < rp[j].String() })
}

for _, p := range rp {
    if p.Name != "Any" {
        predicates = appendFmt(predicates, "%s(%s)", p.Name, argsString(p.Args))
    }
}
```

MustafaSaber:
👍

Issue #1909, opened 2021-11-26
Cookies for OIDC filter are being set for wrong domain


abinet:
We are experiencing an issue with skipper running in multiple environments dev and test and oidc cookie.

Skipper is creating two unique cookies per host like: 

host: dev.applications.example.com
skipperOauthOidc2c6815ff-a
skipperOauthOidc2c6815ff-b

host: test.applications.example.com
skipperOauthOidc84ee2c5e-a
skipperOauthOidc84ee2c5e-b

The problem is that skipper stores both cookie sets under parent host:
host: applications.example.com
skipperOauthOidc2c6815ff-a
skipperOauthOidc2c6815ff-b
skipperOauthOidc84ee2c5e-a
skipperOauthOidc84ee2c5e-b

And this leads to an 400 Bad Request "Request Header Or Cookie Too Large"

I suppose here is the root cause:
https://github.com/zalando/skipper/blame/master/filters/auth/oidc.go#L374
Not sure why it is needed to remove one domain level here.

szuecs:
@abinet we suppose that you have likely multiple applications that you want to have SSO for.
For example you have `a.example.org` and `b.example.org` and you want to use the same identity and cookie for both.
So if you would use:
`applications.dev.example.com`
`applications.test.example.com`
`applications.example.com`

it should also work.

abinet:
Thank you @szuecs  for the details and workaround. The subdomain removal make sense when you put a single skipper instance in front of multiple environments. However it does not work for cases like ours and changing the dns schema is quite complicated in our case as well. What would you say if we make this behavior configurable introducing either instance setting or filter parameter like  "extract-domain-level" with default value 2?

AlexanderYastrebov:
> What would you say if we make this behavior configurable introducing either instance setting or filter parameter like "extract-domain-level" with default value 2?

`oauthOidc*` filters already have quite bloated interface it would be challenging to cram a new parameter :(

AlexanderYastrebov:
We've have discussed the issue with @szuecs and came to the following points.

* it makes sense to have cookie domain level configurable in general as per https://github.com/zalando/skipper/issues/1909#issuecomment-984620970
* the configuration should be per filter, not global. Having it global makes it obscure/non-consistent and postpones the problem until when different domain levels are needed per route
[...]

PR #2203, opened 2023-01-18
filters/auth: add -oauth2-token-cookie-remove-subdomains flag

AlexanderYastrebov:
See related https://github.com/zalando/skipper/issues/1909


filters/auth/grantconfig.go
```diff
@@ -91,6 +92,11 @@ type OAuthConfig struct {
 	// encrypted access token after a successful token exchange.
 	TokenCookieName string
 
+	// TokenCookieRemoveSubdomains sets the number of subdomains to remove from
+	// the callback request hostname to obtain token cookie domain.
+	// Init converts default nil to 1.
+	TokenCookieRemoveSubdomains *int
```
AlexanderYastrebov, line 98: To stay backwards-compatible I had to use pointer here to tell zero and default value (one) apart.
The proper config initialization requires `Init()` call.

If we can sacrifice a bit of compatibility for library users then we could use just `int` and rely on the default skipper flag value of 1.

AlexanderYastrebov: There are not many options here https://stackoverflow.com/questions/68800319/how-to-differentiate-int-null-and-defaulted-to-zero-from-int-actually-equal-to-z

We could also create a wrapper type but that feels too much.

szuecs: If you need this 3 state: 0 value, 0, N, then ptr is the way to go

filters/auth/grant_test.go
```diff
@@ -214,14 +217,22 @@ func checkStatus(t *testing.T, rsp *http.Response, expectedStatus int) {
 	}
 }
 
-func checkRedirect(t *testing.T, rsp *http.Response, expectedURL string) {
+func checkRedirect(t *testing.T, rsp *http.Response, expectedLocationWithoutQuery string) {
```
szuecs, line 220: expectedURI maybe?

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2206, opened 2023-01-19
content length header predicate

adyach:
it should help to route requests based on content length header value

coauthored: @a1exsh

### DIFF
--- a/predicates/content/content_length_between.go
+++ b/predicates/content/content_length_between.go
@@ -14,7 +14,7 @@ type contentLengthBetweenPredicate struct {
 }
 
 // NewContentLengthBetween creates a predicate specification,
-// whose instances match content length header value in range from min to max inclusively.
+// whose instances match content length header value in range from min (inclusively) to max (exclusively).
 // example: ContentLengthBetween(0, 5000)
 func NewContentLengthBetween() routing.PredicateSpec { return &contentLengthBetweenSpec{} }
 
@@ -27,11 +27,23 @@ func (*contentLengthBetweenSpec) Create(args []interface{}) (routing.Predicate, 
 	if len(args) != 2 {
 		return nil, predicates.ErrInvalidPredicateParameters
 	}
-	minLength, _ := args[0].(int64)
-	maxLength, _ := args[1].(int64)
-	if minLength < 0 || maxLength < 0 || minLength > maxLength {
+
+	x, ok := args[0].(float64)
+	if !ok {
 		return nil, predicates.ErrInvalidPredicateParameters
 	}
+	minLength := int64(x)
+
+	x, ok = args[1].(float64)
+	if !ok {
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+	maxLength := int64(x)
+
+	if minLength < 0 || maxLength < 0 || minLength >= maxLength {
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+
 	return &contentLengthBetweenPredicate{
 		min: minLength,
 		max: maxLength,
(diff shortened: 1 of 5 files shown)


predicates/content/content_length_between.go
```diff
@@ -0,0 +1,47 @@
...
+	return predicates.ContentLengthBetweenName
+}
+
+// Create a predicate instance that evaluates content length header value range
+func (*contentLengthBetweenSpec) Create(args []interface{}) (routing.Predicate, error) {
+	if len(args) != 2 {
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+	minLength, _ := args[0].(int64)
+	maxLength, _ := args[1].(int64)
```
AlexanderYastrebov, line 31: Please check conversion success and return `ErrInvalidPredicateParameters` accordingly to prevent e.g. passing strings as arguments.

Also Skipper/eskip native number type is `float64` so you'd need to do something like:
```go
minLength := 0
if x, ok := args[0].(float64); !ok || float64(int(x)) != x {
    return nil, predicates.ErrInvalidPredicateParameters
} else {
    minLength = x
}
```
(See related https://github.com/zalando/skipper/pull/1814)

predicates/content/content_length_between.go
```diff
@@ -0,0 +1,47 @@
...
+		max: maxLength,
+	}, nil
+}
+
+func (p *contentLengthBetweenPredicate) Match(req *http.Request) bool {
+	length := req.ContentLength
+	if length >= p.min && length <= p.max {
+		return true
+	}
+	return false
```
AlexanderYastrebov, line 46:
```go
return length >= p.min && length <= p.max
```

predicates/content/content_length_between_test.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"fmt"
+	"net/http"
+	"testing"
+)
+
+func TestContentLengthMatch(t *testing.T) {
+	s := NewContentLengthBetween()
+	for _, tc := range []struct {
+		minLength int64
```
AlexanderYastrebov, line 12: I think `minLength` is not used in the test.

predicates/content/content_length_between_test.go
```diff
@@ -0,0 +1,59 @@
...
+		maxLength int64
```
AlexanderYastrebov, line 13: Better rename `maxLength` into `contentLength`

predicates/content/content_length_between.go
```diff
@@ -0,0 +1,47 @@
...
+type contentLengthBetweenSpec struct{}
+
+type contentLengthBetweenPredicate struct {
+	min int64
+	max int64
+}
+
+// NewContentLengthBetween creates a predicate specification,
+// whose instances match content length header value in range from min to max inclusively.
+// example: ContentLengthBetween(0, 5000)
```
AlexanderYastrebov, line 18: We need to think about UX here.
Usually interval knobs use left-closed, right-open intervals (in math notation `[left, right)`) for convenient splitting like:
```
r1: ContentLengthBetween(0, 5000) -> ... ;
r2: ContentLengthBetween(5000, 10000) -> ... ;
r3: ContentLengthBetween(10000, 100000) -> ... ;
r0: * -> ... // matches Content-Length >= 100000 
```

I think this is how https://github.com/zalando/skipper/blob/master/docs/reference/predicates.md#between works (although undocumented)

AlexanderYastrebov: Another question if we should also add `ContentLengthMin` and `ContentLengthMax` (or `ContentLengthBetween` with a single argument) but I do not have convincing arguments for it at the moment - it looks like proposed predicate covers all the possible cases.

predicates/content/content_length_between.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+type contentLengthBetweenSpec struct{}
+
+type contentLengthBetweenPredicate struct {
+	min int64
+	max int64
+}
+
+// NewContentLengthBetween creates a predicate specification,
```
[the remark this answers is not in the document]
AlexanderYastrebov: Maybe also note there that content length might be unknown (see https://pkg.go.dev/net/http#Request for details)

AlexanderYastrebov:
Please run `make fmt`

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
@adyach Thank you.

PR #2208, opened 2023-01-20
Sort all predicates

MustafaSaber:


eskip/string.go
```diff
@@ -99,8 +100,18 @@ func (r *Route) predicateString(sortPredicates bool) string {
...
-	for k, v := range r.Headers {
-		predicates = appendFmtEscape(predicates, `Header("%s", "%s")`, `"`, k, v)
+	headerKeys := make([]string, 0, len(r.Headers))
+
+	for k := range r.Headers {
+		headerKeys = append(headerKeys, k)
+	}
+
+	if sortHeaderPredicateValues {
+		sort.SliceStable(headerKeys, func(i, j int) bool { return r.Headers[headerKeys[i]] < r.Headers[headerKeys[j]] })
```
szuecs, line 110: please add line breaks around function body

eskip/string.go
```diff
@@ -80,7 +81,7 @@ func argsString(args []interface{}) string {
 	return strings.Join(sargs, ", ")
 }
 
-func (r *Route) predicateString(sortPredicates bool) string {
+func (r *Route) predicateString(sortPredicates, sortHeaderPredicateValues bool) string {
```
szuecs, line 84: why do we need 2 args or do we need to call this sometimes with `true, false` ?

AlexanderYastrebov: we can pass a pointer to PrettyPrintInfo

eskip/string.go
```diff
@@ -99,8 +100,18 @@ func (r *Route) predicateString(sortPredicates bool) string {
...
+
+	if sortHeaderPredicateValues {
+		sort.SliceStable(headerKeys, func(i, j int) bool { return r.Headers[headerKeys[i]] < r.Headers[headerKeys[j]] })
+	}
+
+	for _, key := range headerKeys {
+		predicates = appendFmtEscape(predicates, `Header("%s", "%s")`, `"`, key, r.Headers[key])
 	}
 
 	for k, rxs := range r.HeaderRegexps {
```
AlexanderYastrebov, line 117: HeaderRegexps is also a map so maybe we should sort it as well

eskip/string.go
AlexanderYastrebov: Here we sort by header value but we should sort by header name instead.
There could be different headers matching the same value, e.g. `Header("X-Foo", "avalue") && Header("X-Bar", "avalue")` but the configuration with the same header matching different values (e.g. `Header("X-Foo", "value1") && Header("X-Foo", "value2")`) is unlikely as it evaluates to false.

AlexanderYastrebov: Here we sort by header value but we should sort by header name instead.
There could be different headers matching the same value, e.g. `Header("X-Foo", "avalue") && Header("X-Bar", "avalue")` but the configuration with the same header matching different values (e.g. `Header("X-Foo", "value1") && Header("X-Foo", "value2")`) is unlikely as it evaluates to false.

eskip/string.go
```diff
@@ -99,8 +100,18 @@ func (r *Route) predicateString(sortPredicates bool) string {
...
 
-	for k, v := range r.Headers {
-		predicates = appendFmtEscape(predicates, `Header("%s", "%s")`, `"`, k, v)
+	headerKeys := make([]string, 0, len(r.Headers))
+
+	for k := range r.Headers {
+		headerKeys = append(headerKeys, k)
+	}
+
+	if sortHeaderPredicateValues {
```
AlexanderYastrebov, line 109: The list is small and sorting would be super fast, the main issue is allocations to copy names.

We should either drop the flag and always sort of create two branches such that when sorting is disabled we avoid copying header names.

AlexanderYastrebov: > We should either drop the flag and always sort of create two branches

Since we need them sorted only for a specific purpose lets have the flag and two branches

AlexanderYastrebov: > We should either drop the flag and always sort of create two branches

Since we need them sorted only for a specific purpose lets have the flag and two branches

eskip/string.go
```diff
@@ -80,7 +80,85 @@ func argsString(args []interface{}) string {
 	return strings.Join(sargs, ", ")
 }
 
-func (r *Route) predicateString(sortPredicates bool) string {
+func (r *Route) predicateStringSorted() string {
+	var predicates []string
+
+	cr := r.Copy()
```
AlexanderYastrebov, line 86: Copy transform legacy predicates into r.Predicates.

TBH I am not sure it is best to have two methods `predicateStringSorted` and `predicateString` - I would try to apply sorting like you proposed previously in place where needed inside the if block...

eskip/string.go
```diff
@@ -80,7 +80,85 @@ func argsString(args []interface{}) string {
...
+	headerKeys = make([]string, 0, len(cr.HeaderRegexps))
+
+	for k := range cr.HeaderRegexps {
+		headerKeys = append(headerKeys, k)
+	}
+
+	sort.SliceStable(headerKeys, func(i, j int) bool {
+		return headerKeys[i] > headerKeys[j]
+	},
+	)
```
AlexanderYastrebov, line 136: IMO it looks better when brace and bracket are together like
```go
sort.SliceStable(headerKeys, func(i, j int) bool {
    return headerKeys[i] > headerKeys[j]
})
```

eskip/string.go
MustafaSaber: then we will need to do this (copy and sort) for all maps/slices (HostRegex, PathRegex, Headers, HeadersRegex and Predicates)

> Copy transforms legacy predicates into r.Predicates

According to this it just make a deep copy, I'm I missing something?  https://github.com/zalando/skipper/blob/master/eskip/eskip.go#L371-L419

MustafaSaber: After some discussions we agreed it would be better to have one function so it's easier to maintain and more error prune.

Another idea was to sort the string after joining, but we will leave this for another day.

eskip/string.go
```diff
@@ -80,7 +80,83 @@ func argsString(args []interface{}) string {
...
+		if p.Name != "Any" {
+			predicates = appendFmt(predicates, "%s(%s)", p.Name, argsString(p.Args))
+		}
+	}
+
+	if len(predicates) == 0 {
+		predicates = append(predicates, "*")
+	}
+
+	return strings.Join(predicates, " && ")
```
MustafaSaber, line 156:
```suggestion
      if sortPredicates {
          sort.SliceStable(predicates, func(i, j int) bool {
		return cr.Predicates[i].String() < cr.Predicates[j].String()
	})
      }
      return strings.Join(predicates, " && ")
```

eskip/string.go
```diff
@@ -87,36 +87,94 @@ func (r *Route) predicateString(sortPredicates bool) string {
...
+	if sortPredicates {
+		headerKeys := make([]string, 0, len(r.HeaderRegexps))
+
+		for k := range r.HeaderRegexps {
+			headerKeys = append(headerKeys, k)
+		}
+
+		sort.SliceStable(headerKeys, func(i, j int) bool {
+			return headerKeys[i] > headerKeys[j]
+		})
```
AlexanderYastrebov, line 152: Lets add a `func sortedKeys[V](m map[string]V) []string` helper

eskip/string.go
```diff
@@ -87,36 +87,94 @@ func (r *Route) predicateString(sortPredicates bool) string {
...
+	if sortPredicates {
+
+		headerKeys := make([]string, 0, len(r.Headers))
+		for k := range r.Headers {
+			headerKeys = append(headerKeys, k)
+		}
+
+		sort.SliceStable(headerKeys, func(i, j int) bool {
+			return headerKeys[i] > headerKeys[j]
+		})
```
AlexanderYastrebov, line 131: https://github.com/zalando/skipper/pull/2208/files#r1086781102

eskip/string.go
```diff
@@ -87,36 +87,94 @@ func (r *Route) predicateString(sortPredicates bool) string {
...
 
-	for _, p := range r.PathRegexps {
+	copyPathRegexps := r.PathRegexps
+
+	if sortPredicates {
+		copyPathRegexps = make([]string, len(r.PathRegexps))
+		copy(copyPathRegexps, r.PathRegexps)
+		sort.SliceStable(copyPathRegexps, func(i, j int) bool {
+			return copyPathRegexps[i] > copyPathRegexps[j]
+		})
```
AlexanderYastrebov, line 111: Lets add `func sortedCopy(s []string) []string` helper

eskip/string.go
```diff
@@ -87,36 +87,94 @@ func (r *Route) predicateString(sortPredicates bool) string {
 		predicates = appendFmtEscape(predicates, `Path("%s")`, `"`, r.Path)
 	}
 
-	for _, h := range r.HostRegexps {
+	copyHostRegexps := r.HostRegexps
+
+	if sortPredicates {
+		copyHostRegexps = make([]string, len(r.HostRegexps))
+		copy(copyHostRegexps, r.HostRegexps)
+		sort.SliceStable(copyHostRegexps, func(i, j int) bool {
```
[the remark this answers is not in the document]
AlexanderYastrebov: https://github.com/zalando/skipper/pull/2208/files#r1086784002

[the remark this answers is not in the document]
MustafaSaber: I don't see where that link point specifically, but the only missing slice is predicate slice.

eskip/string.go
```diff
@@ -80,6 +80,22 @@ func argsString(args []interface{}) string {
 	return strings.Join(sargs, ", ")
 }
 
+func sortedKeys[V string | []string](m map[string]V) []string {
```
AlexanderYastrebov, line 83: `func sortedKeys[V any](m map[string]V) []string `

eskip/string.go
```diff
@@ -80,43 +80,91 @@ func argsString(args []interface{}) string {
...
+
 func (r *Route) predicateString(sortPredicates bool) string {
 	var predicates []string
 
 	if r.Path != "" {
 		predicates = appendFmtEscape(predicates, `Path("%s")`, `"`, r.Path)
 	}
 
-	for _, h := range r.HostRegexps {
+	copyHostRegexps := r.HostRegexps
```
AlexanderYastrebov, line 106: minor: maybe just `hostRegexps` as it is not always a copy

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
Feel free to address minor comments, otherwise its good to go.

eskip/string.go
MustafaSaber: we decided to go with `interface{}` instead of `any` since we override `any` in [eskip/parser.go](https://github.com/zalando/skipper/blob/master/eskip/parser.go#L47)

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2209, opened 2023-01-20
filters/auth: redirect grant callback request to the initial request domain

AlexanderYastrebov:
Normally callback host (`redirect_uri` auth parameter) matches initial request host because `redirect_uri` is created dynamically from the initial request and therefore this change is doing nothing.

This change is useful when `redirect_uri` is fixed via `-oauth2-auth-url-parameters`, initial request host is diffrent from the `redirect_uri` host but token cookie should be issued for the initial request domain.

Consider example configuration:

* `-oauth2-auth-url-parameters=redirect_uri=https://callback.example.org/a-callback`,
* `-oauth2-token-cookie-remove-subdomains=0`

Initial request to the `https://foo.example.org/test` triggers authorization flow, provider redirects user to the `https://callback.example.org/a-callback?...`, callback filter observes that initial request has diffrent host and redirects request to another callback at `https://foo.example.org/a-callback?...` which completes the flow and creates a token cookie for the `foo.example.org` domain.


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2213, opened 2023-01-26
filters/auth: make `oauthGrant` check token cookie is allowed for request domain

AlexanderYastrebov:
filters/auth: make oauthGrant check token cookie is allowed for request domain

This prevents reuse of leaked cookie in another domain

### DIFF
--- a/filters/auth/grantcookie.go
+++ b/filters/auth/grantcookie.go
@@ -3,7 +3,9 @@ package auth
 import (
 	"encoding/base64"
 	"encoding/json"
+	"net"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/zalando/skipper/secrets"
@@ -14,6 +16,7 @@ type cookie struct {
 	AccessToken  string    `json:"access_token"`
 	RefreshToken string    `json:"refresh_token"`
 	Expiry       time.Time `json:"expiry,omitempty"`
+	Domain       string    `json:"domain,omitempty"`
 }
 
 func decodeCookie(cookieHeader string, config *OAuthConfig) (c *cookie, err error) {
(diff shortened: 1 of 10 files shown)


szuecs:
:+1:

filters/auth/grantcookie.go
```diff
@@ -41,41 +44,48 @@ func (c *cookie) isAccessTokenExpired() bool {
...
-			request.AddCookie(c)
+		cookie, err := decodeCookie(c.Value, config)
+		if err == nil && cookie.allowedForHost(request.Host) {
+			request.Header.Del("Cookie")
+			for _, c := range cookies[:i] {
+				request.AddCookie(c)
+			}
+			for _, c := range cookies[i+1:] {
+				request.AddCookie(c)
+			}
```
szuecs, line 79: Tricky the two loops to spot that you drop element i

filters/auth/grantcookie.go
```diff
@@ -41,41 +44,48 @@ func (c *cookie) isAccessTokenExpired() bool {
...
 		}
-		new = append(new, c)
-	}
 
-	if cookie != nil {
-		request.Header.Del("Cookie")
-		for _, c := range new {
-			request.AddCookie(c)
+		cookie, err := decodeCookie(c.Value, config)
+		if err == nil && cookie.allowedForHost(request.Host) {
```
szuecs, line 72: Here we could also continue on err != nil
Which is maybe more obvious than the current code

szuecs (commented):
Feel free to merge looks good only small nits

filters/auth/grantcookie.go
AlexanderYastrebov: Thank you for pointing this!
I've refactored it into a single loop and also noticed we had no proper test for it so I've added one

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2217, opened 2023-02-03
Fix Makefile to pass CDP check for go1.20

RomanZavodskikh:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ bench: build $(TEST_PLUGINS)
 lint: build staticcheck
 
 clean:
-	go clean -i -cache -testcache ./...
+	go clean -i -cache -testcache
 	rm -rf .coverprofile-all .cover
 	rm -f ./_test_plugins/*.so
 	rm -f ./_test_plugins_fail/*.so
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -138,8 +138,9 @@ vet: $(SOURCES)
 # -ST1020 too many wrong comments on exported functions to fix right away
 # -ST1021 too many wrong comments on exported functions to fix right away
 # -ST1022 too many wrong comments on exported functions to fix right away
+# -SA1019 using non-cryptographic rand functions
 staticcheck: $(SOURCES)
-	staticcheck -checks "all,-ST1000,-ST1003,-ST1012,-ST1020,-ST1021" $(PACKAGES)
+	staticcheck -checks "all,-ST1000,-ST1003,-ST1012,-ST1020,-ST1021,-SA1019" $(PACKAGES)
```
mikkeloscar, line 143: I think it would be better to add this exception only at the place where it is. Otherwise we may miss introducing this in a new place in the future.

mikkeloscar: Or even better ofc. is to fix the cases where the deprecated function is used. :)

mikkeloscar:
:+1:

szuecs:
👍

PR #2218, opened 2023-02-03
RouteSRV: Basic validation

MustafaSaber:
Basic validation for routes to match skipper.

### DIFF
--- a/routesrv/options.go
+++ b/routesrv/options.go
@@ -166,4 +166,8 @@ type Options struct {
 	// EditRoute will be applied to all routes automatically and
 	// will apply changes to all matching routes.
 	EditRoute []*eskip.Editor
+
+	// DisabledFilters is a list of filters unavailable for use, routes with these filters will be ignored
+	// with a logged error message.
+	DisabledFilters []string
 }
--- a/routesrv/polling.go
+++ b/routesrv/polling.go
@@ -13,6 +13,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/filters/auth"
+	"github.com/zalando/skipper/predicates"
 	"github.com/zalando/skipper/routing"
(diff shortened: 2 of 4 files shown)


routesrv/polling.go
```diff
@@ -142,10 +144,57 @@ func (p *poller) process(routes []*eskip.Route) []*eskip.Route {
 		routes = cloner.Do(routes)
 	}
 
+	disabledFilters := len(p.disabledFilters) != 0
```
szuecs, line 147: maybe `hasDisabledFilters` to highlight bool?

routesrv/options.go
```diff
@@ -166,4 +166,7 @@ type Options struct {
 	// EditRoute will be applied to all routes automatically and
 	// will apply changes to all matching routes.
 	EditRoute []*eskip.Editor
+
+	// DisabledFilters is a list of filters unavailable for use
```
szuecs, line 170: What will happen if I apply a disabled filter if it's set here?
Do we ignore the filter or do we log an error and break update of the route?

These would be great to answer here.

routesrv/polling.go
```diff
@@ -142,10 +144,57 @@ func (p *poller) process(routes []*eskip.Route) []*eskip.Route {
...
+		}
+		if validRouteFilters {
+			validRoutes = append(validRoutes, r)
+		}
+	}
+
+	return validRoutes
+}
+
+func toBooleanMap[C comparable](values []C) map[C]bool {
```
szuecs, line 194: I would suggest to use struct{} instead of bool, which saves memory and you do not use the bool value anyways in line 179.
```
func toMap[C comparable](values []C) map[C]struct{} {
	h := make(map[C]struct{}, len(values))
	for _, value := range values {
		h[value] = struct{}{}
	}
	return h
}
```

AlexanderYastrebov:
I am not sure we need this actually. We can think of routesrv as another source of routes similar to string, file or k8s manifests - Skipper should take care of invalid/disabled filters.

PR #2222, opened 2023-02-08
since gosec is garbage we do not want have this crap

szuecs:
https://github.com/securego/gosec/issues/925


szuecs:
:+1:

RomanZavodskikh:
I don't believe it's so useless. :-)
However, even it is, you need to delete it from some other places too.
https://github.com/zalando/skipper/blob/master/delivery.yaml#L57
https://github.com/zalando/skipper/blob/master/.github/workflows/master.yaml#L25

PR #2229, opened 2023-02-13
Added the performance clusterLimitRedis.Allow operation

RomanZavodskikh:
Such performance tests are needed for us to find out how huge keys can Skipper ratelimiter use without significant performance degradation.

### DIFF
--- a/net/redistest/redistest.go
+++ b/net/redistest/redistest.go
@@ -10,11 +10,11 @@ import (
 	"github.com/testcontainers/testcontainers-go/wait"
 )
 
-func NewTestRedis(t *testing.T) (address string, done func()) {
+func NewTestRedis(t testing.TB) (address string, done func()) {
 	return NewTestRedisWithPassword(t, "")
 }
 
-func NewTestRedisWithPassword(t *testing.T, password string) (address string, done func()) {
+func NewTestRedisWithPassword(t testing.TB, password string) (address string, done func()) {
 	var args []string
 	if password != "" {
 		args = append(args, "--requirepass", password)
--- a/ratelimit/redis_test.go
+++ b/ratelimit/redis_test.go
@@ -2,6 +2,8 @@ package ratelimit
 
 import (
 	"context"
+	"fmt"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


szuecs:
:+1:

net/redistest/redistest.go
```diff
@@ -10,6 +10,62 @@ import (
 	"github.com/testcontainers/testcontainers-go/wait"
 )
 
+func NewBenchRedis(b *testing.B) (address string, done func()) {
+	return NewBenchRedisWithPassword(b, "")
+}
+
+func NewBenchRedisWithPassword(b *testing.B, password string) (address string, done func()) {
```
AlexanderYastrebov, line 17: Could you check which methods of testing.B we need and maybe we can change existing constructors to use [common TB](https://pkg.go.dev/testing#TB) interface?

ratelimit/redis_test.go
```diff
@@ -87,6 +89,37 @@ func Test_clusterLimitRedis_WithPass(t *testing.T) {
...
+func Benchmark_clusterLimitRedis_Allow(b *testing.B) {
+	redisAddr, done := redistest.NewBenchRedis(b)
+	defer done()
+
+	clusterClientlimit := Settings{
+		Type:       ClusterClientRatelimit,
+		Lookuper:   NewHeaderLookuper("X-Test"),
+		MaxHits:    10,
+		TimeWindow: time.Second,
+		Group:      "B",
```
AlexanderYastrebov, line 101: Ratelimiter implementation hashes the key and appends it to the group name:
https://github.com/zalando/skipper/blob/832dedc0262faf65c330af7b1edf7fc48f173206/ratelimit/redis.go#L139-L141
so we need to test how it performs for different group name lengths, not key lengths.

ratelimit/redis_test.go
```diff
@@ -87,6 +89,37 @@ func Test_clusterLimitRedis_WithPass(t *testing.T) {
...
+		b.Run(fmt.Sprintf("ratelimit with group name of 2^%d symbols", i), func(b *testing.B) {
+			ringClient := net.NewRedisRingClient(&net.RedisOptions{Addrs: []string{redisAddr}})
+			defer ringClient.Close()
+			c := newClusterRateLimiterRedis(
+				clusterClientlimit,
+				ringClient,
+				clusterClientlimit.Group,
+			)
+
+			groupName := strings.Repeat("a", 1<<i)
```
AlexanderYastrebov, line 114: This is not a group name but a key, see above.

ratelimit/redis_test.go
RomanZavodskikh: Yes, very good catch! :-)

AlexanderYastrebov:
Please add benchmark output to the commit message and PR body to see typical numbers

AlexanderYastrebov:
:+1:

ratelimit/redis_test.go
```diff
@@ -87,6 +89,37 @@ func Test_clusterLimitRedis_WithPass(t *testing.T) {
 	}
 }
 
+func Benchmark_clusterLimitRedis_Allow(b *testing.B) {
+	redisAddr, done := redistest.NewTestRedis(b)
+	defer done()
+
+	for i := 0; i < 21; i++ {
+		b.Run(fmt.Sprintf("ratelimit with group name of 2^%d symbols", i), func(b *testing.B) {
```
AlexanderYastrebov, line 97: minor: I think it is cleaner to format final decimal result (i.e. 1<<i) into benchmark name

RomanZavodskikh:
@AlexanderYastrebov , could you please reapprove? :-)

MustafaSaber:
👍

RomanZavodskikh:
:+1:

PR #2230, opened 2023-02-13
fix: admissionControl should count client hang-up

szuecs:
This introduces a possibility for filters to get called Response() even we got no response and the ctx.Response() will return nil. Filters like apiMonitoring depend a lot on the Response() values, so I decided not to fix it here.

### DIFF
--- a/filters/shedder/admission.go
+++ b/filters/shedder/admission.go
@@ -434,11 +434,9 @@ func (ac *admissionControl) Response(ctx filters.FilterContext) {
 	}
 
 	rsp := ctx.Response()
-	if rsp != nil {
-		code := rsp.StatusCode
-		if code < 499 {
-			ac.successCounter.Add(1)
-		}
+	code := rsp.StatusCode
+	if code < 499 {
+		ac.successCounter.Add(1)
 	}
 	ac.counter.Add(1)
 }
@@ -453,6 +451,3 @@ func (ac *admissionControl) startSpan(ctx context.Context) (span opentracing.Spa
 	}
 	return
 }
-
-// Opt-In for filters to get called Response(ctx) in case of errors
-func (ac *admissionControl) HandleErrorResponse() {}
--- a/filters/shedder/admission.go
+++ b/filters/shedder/admission.go
@@ -433,9 +433,12 @@ func (ac *admissionControl) Response(ctx filters.FilterContext) {
 		return
 	}
 
-	code := ctx.Response().StatusCode
-	if code < 499 {
-		ac.successCounter.Add(1)
+	rsp := ctx.Response()
+	if rsp != nil {
+		code := rsp.StatusCode
+		if code < 499 {
+			ac.successCounter.Add(1)
+		}
 	}
 	ac.counter.Add(1)
 }
@@ -450,3 +453,10 @@ func (ac *admissionControl) startSpan(ctx context.Context) (span opentracing.Spa
 	}
 	return
 }
+
+// Opt-In for filters to get called Response(ctx) in case of errors
+type errorHandlerFilter interface {
+	HandleErrorResponse()
+}
+
+func (ac *admissionControl) HandleErrorResponse() {}
--- a/filters/shedder/admission.go
+++ b/filters/shedder/admission.go
@@ -455,8 +455,4 @@ func (ac *admissionControl) startSpan(ctx context.Context) (span opentracing.Spa
 }
 
 // Opt-In for filters to get called Response(ctx) in case of errors
-type errorHandlerFilter interface {
-	HandleErrorResponse()
-}
-
 func (ac *admissionControl) HandleErrorResponse() {}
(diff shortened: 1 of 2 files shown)


filters/shedder/admission.go
```diff
@@ -450,3 +453,10 @@ func (ac *admissionControl) startSpan(ctx context.Context) (span opentracing.Spa
 	}
 	return
 }
+
+// Opt-In for filters to get called Response(ctx) in case of errors
+type errorHandlerFilter interface {
+	HandleErrorResponse()
+}
```
AlexanderYastrebov, line 460: I think we do not need it here, its enough to define method (maybe with a Opt-in comment)

szuecs: yeah as staticcheck wrote :)

proxy/proxy.go
```diff
@@ -1132,6 +1132,8 @@ func (p *Proxy) do(ctx *context) (err error) {
 		backendStart := time.Now()
 		rsp, perr := p.makeBackendRequest(ctx, backendContext)
 		if perr != nil {
+			p.handleFiltersOnError(ctx, processedFilters)
```
AlexanderYastrebov, line 1135: Would it be possible to create a response similar to what `p.errorResponse` does and set it as a response in the context?

filters/shedder/admission.go
```diff
@@ -433,9 +433,12 @@ func (ac *admissionControl) Response(ctx filters.FilterContext) {
...
 
-	code := ctx.Response().StatusCode
-	if code < 499 {
-		ac.successCounter.Add(1)
+	rsp := ctx.Response()
+	if rsp != nil {
+		code := rsp.StatusCode
+		if code < 499 {
+			ac.successCounter.Add(1)
+		}
```
AlexanderYastrebov, line 441: I think this would not work as we translate proxyError caused by client into 499 response in the p.errorResponse

proxy/proxy.go
```diff
@@ -1526,3 +1529,35 @@ func ensureUTF8(s string) string {
 	}
 	return fmt.Sprintf("invalid utf-8: %q", s)
 }
+
+// Opt-In for filters to get called Response(ctx) in case of errors
+type errorHandlerFilter interface {
+	HandleErrorResponse()
+}
+
+func (p *Proxy) handleFiltersOnError(ctx *context, filters []*routing.RouteFilter) {
```
AlexanderYastrebov, line 1538: Lets name it `applyFiltersOnError` similar to existing `applyFiltersToRequest` and `applyFiltersToResponse`

proxy/proxy.go
```diff
@@ -1526,3 +1529,35 @@ func ensureUTF8(s string) string {
...
+	filtersStart := time.Now()
+	filterTracing := p.tracing.startFilterTracing("response_filters", ctx)
+	defer filterTracing.finish()
+
+	last := len(filters) - 1
+	for i := range filters {
+		fi := filters[last-i]
```
AlexanderYastrebov, line 1545: Not sure why it was done like that before, I'd do:
```go
for i:=len(filters)-1; i>=0; i-- {
  fi := filters[i]
}
```

proxy/proxy.go
AlexanderYastrebov: Something like
```go
if perr != nil {
    ctx.response = makeErrorResponse(perr)
    p.handleFiltersOnError(ctx, processedFilters)
    ...
```
where `makeErrorResponse` replicates data/logic from `p.errorResponse`+`p.sendError` and `p.sendError` then would just send `ctx.response`.

filters/shedder/admission.go
szuecs: Right now it works because rsp is nil , but requires change if I change to create a fakeResponse as you suggest in another comment.

proxy/proxy.go
szuecs: I can change both to match it, but I think we should have one style.

proxy/proxy.go
szuecs: @AlexanderYastrebov that looks like it can't work. `p.errorResponse` and `p.sendError` require access to ctx and `p`. I think then we would need to do:
```
if perr != nil {
    p.makeErrorResponse(ctx, perr)
    p.handleFiltersOnError(ctx, processedFilters)
```

proxy/proxy.go
szuecs: done

proxy/proxy.go
szuecs: done

proxy/proxy.go
szuecs: Do you mean that all changes/writes to ctx.responseWriter in `p.errorResponse`+`p.sendError` shouldbe replicated to create a response?
Looking into the methods show up more questions than answers to your comment, so please clarify. :)

PR #2232, opened 2023-02-14
refactor: proxy error response

szuecs:
refactor: proxy error response to always have a response defined to simplify the logic

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1273,29 +1273,31 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 		msgPrefix = "client canceled"
 		logFunc = p.log.Infof
 	}
-	req := ctx.Request()
-	remoteAddr := remoteHost(req)
-	uri := req.RequestURI
-	if i := strings.IndexRune(uri, '?'); i >= 0 {
-		uri = uri[:i]
+	if id != unknownRouteID {
+		req := ctx.Request()
+		remoteAddr := remoteHost(req)
+		uri := req.RequestURI
+		if i := strings.IndexRune(uri, '?'); i >= 0 {
+			uri = uri[:i]
+		}
+		logFunc(
+			`%s after %v, route %s with backend %s %s%s, status code %d: %v, remote host: %s, request: "%s %s %s", host: %s, user agent: "%s"`,
+			msgPrefix,
+			time.Since(ctx.startServe),
+			id,
+			backendType,
+			backend,
+			flowIdLog,
+			ctx.response.StatusCode,
+			err,
+			remoteAddr,
+			req.Method,
+			uri,
+			req.Proto,
+			req.Host,
+			req.UserAgent(),
+		)
 	}
-	logFunc(
-		`%s after %v, route %s with backend %s %s%s, status code %d: %v, remote host: %s, request: "%s %s %s", host: %s, user agent: "%s"`,
-		msgPrefix,
-		time.Since(ctx.startServe),
-		id,
-		backendType,
-		backend,
-		flowIdLog,
-		ctx.response.StatusCode,
-		err,
-		remoteAddr,
-		req.Method,
-		uri,
-		req.Proto,
-		req.Host,
-		req.UserAgent(),
-	)
 
 	copyHeader(ctx.responseWriter.Header(), ctx.response.Header)
 	ctx.responseWriter.WriteHeader(ctx.response.StatusCode)
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -761,9 +761,8 @@ func (p *Proxy) applyFiltersToResponse(filters []*routing.RouteFilter, ctx *cont
 	filterTracing := p.tracing.startFilterTracing("response_filters", ctx)
 	defer filterTracing.finish()
 
-	last := len(filters) - 1
-	for i := range filters {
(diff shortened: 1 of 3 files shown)


proxy/proxy.go
```diff
@@ -1033,11 +1011,17 @@ func (p *Proxy) do(ctx *context) (err error) {
...
+			perr := &proxyError{
+				err: fmt.Errorf("panic caused by: %v", r),
+			}
+			p.makeErrorResponse(ctx, perr)
+			err = perr
 		}
 	}()
 
 	if ctx.executionCounter > p.maxLoops {
+		// TODO(sszuecs): think about setting status code to 465 (AWS redirect loop) or similar
```
AlexanderYastrebov, line 1023: I think it is 463, lets maybe add a link https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-troubleshooting.html#http-463-issues

proxy/proxy.go
```diff
@@ -1132,6 +1124,8 @@ func (p *Proxy) do(ctx *context) (err error) {
 		backendStart := time.Now()
 		rsp, perr := p.makeBackendRequest(ctx, backendContext)
 		if perr != nil {
+			p.makeErrorResponse(ctx, perr)
```
AlexanderYastrebov, line 1127: Lets move this close to the return in the else branch

proxy/proxy.go
```diff
@@ -1151,6 +1145,7 @@ func (p *Proxy) do(ctx *context) (err error) {
 				rsp, perr2 = p.makeBackendRequest(ctx, backendContext)
 				if perr2 != nil {
 					p.log.Errorf("Failed to retry backend request: %v", perr2)
+					p.makeErrorResponse(ctx, perr2)
```
AlexanderYastrebov, line 1148: Lets move this close to the return perr2

proxy/proxy.go
szuecs: yes maybe that. I remember 465, but there is only ERR_TOO_MANY_REDIRECTS and nothing about the status code.
It could be that it always adds to XFF and entry, which would be 463.
https://aws.amazon.com/premiumsupport/knowledge-center/redirect-http-https-elb/ has no status code

proxy/proxy.go
szuecs: done

proxy/proxy.go
szuecs: done

proxy/proxy.go
szuecs: With this weak data I will not add wrong docs to it.

proxy/proxy_test.go
```diff
@@ -2009,29 +2009,30 @@ func TestAccessLogOnFailedRequest(t *testing.T) {
...
 	}
 
 	expected := fmt.Sprintf(`"GET / HTTP/1.1" %d %d "-" "Go-http-client/1.1"`, http.StatusBadGateway, len(http.StatusText(http.StatusBadGateway))+1)
 	if !strings.Contains(output, expected) || !strings.Contains(output, proxyURL.Host) {
-		t.Error("failed to log access", output, expected)
-		t.Log(output)
-		t.Log(expected)
+		t.Errorf("Failed to get accesslog '%v' '%v'", output, expected)
+		t.Logf("out: '%v'", output)
+		t.Logf("expect: '%v'", expected)
```
AlexanderYastrebov, line 2035: This is redundant, values would be logged by the above Errof call

proxy/proxy_test.go
```diff
@@ -1990,16 +1990,16 @@ func TestEnableAccessLogWithFilter(t *testing.T) {
...
-		AccessLogOutput: &buf})
+		AccessLogOutput: buf})
 
 	s := httptest.NewServer(http.HandlerFunc(func(http.ResponseWriter, *http.Request) {}))
 	s.Close()
 
 	p, err := newTestProxy(fmt.Sprintf(`* -> "%s"`, s.URL), 0)
 	if err != nil {
-		t.Error(err)
+		t.Errorf("Failed to create test proxy: %v", err)
```
AlexanderYastrebov, line 2002: Maybe just t.Fatal(err)

proxy/utils.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"bytes"
+	"sync"
+)
+
+type LockedBuffer struct {
+	mu  sync.Mutex
+	buf *bytes.Buffer
+}
```
AlexanderYastrebov, line 11: I'd use embedding like :)
```go
type SyncBuffer struct {
	bytes.Buffer
	mu  sync.Mutex
}
```

AlexanderYastrebov:
Great to see this working!

AlexanderYastrebov:
:+1:

proxy/utils.go
szuecs: you can also embed sync.Mutex, but there are some conditions which are odd, so normally I try to not use embedding

proxy/proxy_test.go
szuecs: yes, I changed it to have a cmp.Diff to spot diffs maybe easier

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2237, opened 2023-02-20
RouteSRV: Expose redis shards endpoints

MustafaSaber:


config/config.go
```diff
@@ -660,6 +663,29 @@ func (c *Config) ToRouteSrvOptions() routesrv.Options {
 		// Auth:
 		EnableOAuth2GrantFlow: c.EnableOAuth2GrantFlow,
 		OAuth2CallbackPath:    c.Oauth2CallbackPath,
+
+		// swarm:
```
szuecs, line 667: we don't need the swarm options (timeouts and access to redis are only interesting in the data plane)

net/redisclient.go
```diff
@@ -197,6 +199,19 @@ func NewRendezvousVnodes(shards []string) redis.ConsistentHash {
 	return rendezvousVnodes{rendezvous.New(vshards, xxhash.Sum64String), table}
 }
 
+func (rrc RedisRingClient) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
szuecs, line 202: TH GH action should be solved by using a pointer receiver instead of the value receiver.

config/config.go
MustafaSaber: Can you check now? Removed redisClient connection.

routesrv/routesrv.go
```diff
@@ -24,6 +25,36 @@ type RouteServer struct {
 	wg     *sync.WaitGroup
 }
 
+type RedisHandler struct {
+	AddrUpdater func() []byte
+}
+
+func (rh *RedisHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "GET" && r.Method != "HEAD" {
```
szuecs, line 33: I think GET is the only valid method in this case

routesrv/routesrv.go
```diff
@@ -91,6 +115,22 @@ func New(opts Options) (*RouteServer, error) {
 		oauthConfig.CallbackPath = opts.OAuth2CallbackPath
 	}
 
+	var rh *RedisHandler
+	// in case we have kubernetes dataclient and we can detect redis instances, we patch redisOptions
+	if opts.KubernetesRedisServiceNamespace != "" && opts.KubernetesRedisServiceName != "" {
+		log.Infof("Use endpoints %s/%s to fetch updated redis shards", opts.KubernetesRedisServiceNamespace, opts.KubernetesRedisServiceName)
+		rh = &RedisHandler{}
+		rh.AddrUpdater = getRedisAddresses(opts.KubernetesRedisServiceNamespace, opts.KubernetesRedisServiceName, dataclient)
+		handler.Handle("/swarm", rh)
```
szuecs, line 124: Let's add another line with:
```
dataclient.LoadAll()
```
to make sure that we have an initial clusterstate and drop the comment in line 43

routesrv/routesrv.go
```diff
@@ -24,6 +25,36 @@ type RouteServer struct {
...
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+	w.Header().Add("Content-Type", "text/plain; charset=utf-8")
+	w.Write(rh.AddrUpdater())
+}
+
+func getRedisAddresses(namespace, name string, kdc *kubernetes.Client) func() []byte {
+	return func() []byte {
+		// TODO(sszuecs): make sure kubernetes dataclient is already initialized and
```
szuecs, line 43: the comment can be dropped in case the other comment was done.

routesrv/routesrv.go
MustafaSaber: Thanks was struggling with the testcase

routesrv/routesrv.go
```diff
@@ -91,6 +84,26 @@ func New(opts Options) (*RouteServer, error) {
...
+	// in case we have kubernetes dataclient and we can detect redis instances, we patch redisOptions
+	if opts.KubernetesRedisServiceNamespace != "" && opts.KubernetesRedisServiceName != "" {
+		log.Infof("Use endpoints %s/%s to fetch updated redis shards", opts.KubernetesRedisServiceNamespace, opts.KubernetesRedisServiceName)
+		rh = &RedisHandler{}
+		_, err := dataclient.LoadAll()
+		if err != nil {
+			return nil, err
+		}
+		rh.AddrUpdater = getRedisAddresses(opts.KubernetesRedisServiceNamespace, opts.KubernetesRedisServiceName, dataclient)
+		handler.Handle("/swarm", rh)
```
szuecs, line 97: @AlexanderYastrebov what do you think about /swarm as path endpoint?

AlexanderYastrebov: I would call it redis, not sure we need to abstract here

routesrv/routesrv_test.go
```diff
@@ -196,6 +213,30 @@ func TestFetchedRoutesAreServedInEskipFormat(t *testing.T) {
...
+		KubernetesRedisServiceNamespace: "namespace1",
+		KubernetesRedisServiceName:      "service1",
+	})
+
+	w := getRedisURLs(rs)
+
+	want := parseRedisIP(t, "testdata/redis-ip.txt")
+	got := w.Body.Bytes()
+
+	if !bytes.Equal(got, want) {
```
szuecs, line 233: I don't know but if this test is flaky we need to sort the string values by line to compare

routesrv/routesrv.go
szuecs: maybe /swarm/redis/endpoints or /swarm/redis/shards ?
Like this we can add `/swarm/swim/*` later

routesrv/routesrv_test.go
MustafaSaber: I can sort, but I don't think it's flaky. I need to check how the IPs are returned from k8s dataclient first.

szuecs: just run something like:
```
go test -run MYTest -count 10 ./routesrv
```

szuecs:
:+1:

MustafaSaber:
Will move forward with this for now and come back later if we need to change anything.

MustafaSaber:
👍

PR #2239, opened 2023-02-22
feature: apply filters on response that want to be processed

szuecs:
feature: apply filters on response that want to be processed
fix: admissionControl 499 cases should be counted

- [x] tests
- [x] docs

### DIFF
--- a/docs/tutorials/development.md
+++ b/docs/tutorials/development.md
@@ -166,7 +166,7 @@ func (f *myFilter) Response(ctx filters.FilterContext) {
      // change data in ctx.Response() for example
 }
 
-func (f *myFilter) Close() {
+func (f *myFilter) Close() error {
      // cleanup your filter
 }
 ```
--- a/docs/tutorials/development.md
+++ b/docs/tutorials/development.md
@@ -137,6 +137,86 @@ func (f *myFilter) Response(ctx filters.FilterContext) {
 
 Find a detailed example at [how to develop a filter](../reference/development.md#how-to-develop-a-filter).
 
+### Filters with cleanup
+
+Sometimes your filter needs to cleanup resources on shutdown. In Go
+functions that do this have often the name `Close()`.
+There is the `filters.FilterCloser` interface that if you comply with
+it, the routing.Route will make sure your filters are closed in case
+of `routing.Routing` was closed.
+
+```go
+type myFilter struct{}
+
+func NewMyFilter() filters.Spec {
+	return &myFilter{}
+}
+
+func (spec *myFilter) Name() string { return "myFilter" }
+
+func (spec *myFilter) CreateFilter(config []interface{}) (filters.Filter, error) {
+     return NewMyFilter(), nil
+}
+
+func (f *myFilter) Request(ctx filters.FilterContext) {
+     // change data in ctx.Request() for example
+}
+
+func (f *myFilter) Response(ctx filters.FilterContext) {
+     // change data in ctx.Response() for example
+}
+
+func (f *myFilter) Close() {
+     // cleanup your filter
+}
+```
+
+### Filters with error handling
+
+Sometimes you want to have a filter that wants to get called
+`Response()` even if the proxy will not send a response from the
+backend, for example you want to count error status codes, like
+the [admissionControl](../reference/filters.md#admissioncontrol)
+filter.
+In this case you need to comply with the following proxy interface:
+
+```go
+// errorHandlerFilter is an opt-in for filters to get called
+// Response(ctx) in case of errors.
+type errorHandlerFilter interface {
+	// HandleErrorResponse returns true in case a filter wants to get called
+	HandleErrorResponse() bool
+}
+```
+
+Example:
+```go
+type myFilter struct{}
+
+func NewMyFilter() filters.Spec {
+	return &myFilter{}
+}
+
+func (spec *myFilter) Name() string { return "myFilter" }
+
+func (spec *myFilter) CreateFilter(config []interface{}) (filters.Filter, error) {
+     return NewMyFilter(), nil
+}
+
+func (f *myFilter) Request(ctx filters.FilterContext) {
+     // change data in ctx.Request() for example
+}
+
+func (f *myFilter) Response(ctx filters.FilterContext) {
+     // change data in ctx.Response() for example
+}
+
+func (f *myFilter) HandleErrorResponse() bool() {
+     return true
+}
+```
+
+
 ## Predicates
 
 Predicates allow to match a condition, that can be based on arbitrary
--- a/filters/filters.go
+++ b/filters/filters.go
@@ -2,6 +2,7 @@ package filters
 
 import (
 	"errors"
+	"io"
 	"net/http"
 	"time"
 
@@ -157,7 +158,7 @@ type Filter interface {
 // goroutines.
 type FilterCloser interface {
 	Filter
-	Close() error
+	io.Closer
 }
 
 // Spec objects are specifications for filters. When initializing the routes,
--- a/filters/filters.go
+++ b/filters/filters.go
@@ -157,7 +157,7 @@ type Filter interface {
 // goroutines.
 type FilterCloser interface {
 	Filter
-	Close()
+	Close() error
 }
 
 // Spec objects are specifications for filters. When initializing the routes,
--- a/filters/filters.go
+++ b/filters/filters.go
@@ -151,6 +151,15 @@ type Filter interface {
 	Response(FilterContext)
 }
 
+// FilterCloser are Filters that need to cleanup resources after
+// filter teminiation. For example Filters, that create a goroutine
+// for some reason need to cleanup their goroutine or they would leak
+// goroutines.
+type FilterCloser interface {
+	Filter
+	Close()
+}
+
 // Spec objects are specifications for filters. When initializing the routes,
 // the Filter instances are created using the Spec objects found in the
 // registry.
--- a/filters/filters.go
+++ b/filters/filters.go
@@ -152,7 +152,7 @@ type Filter interface {
 }
 
 // FilterCloser are Filters that need to cleanup resources after
-// filter teminiation. For example Filters, that create a goroutine
+// filter termination. For example Filters, that create a goroutine
 // for some reason need to cleanup their goroutine or they would leak
 // goroutines.
 type FilterCloser interface {
--- a/proxy/failing_filter_test.go
+++ b/proxy/failing_filter_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/zalando/skipper/net/httptest"
 	"github.com/zalando/skipper/proxy"
 	"github.com/zalando/skipper/proxy/proxytest"
+	"github.com/zalando/skipper/routing"
 )
 
 func TestResponseFilterOnProxyError(t *testing.T) {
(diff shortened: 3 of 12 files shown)


proxy/proxy.go
```diff
@@ -1272,6 +1279,9 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 	if ctx.response.StatusCode == 499 {
 		msgPrefix = "client canceled"
 		logFunc = p.log.Infof
+		if p.accessLogDisabled {
+			logFunc = p.log.Debugf
+		}
```
AlexanderYastrebov, line 1284: Why this change in this scope?

szuecs: Because I think we did not caught it the last time and I saw it during tests.

filters/filters.go
```diff
@@ -151,6 +151,15 @@ type Filter interface {
...
 }
 
+// FilterCloser are Filters that need to cleanup resources after
+// filter teminiation. For example Filters, that create a goroutine
+// for some reason need to cleanup their goroutine or they would leak
+// goroutines.
+type FilterCloser interface {
+	Filter
+	Close()
+}
```
AlexanderYastrebov, line 161: We implement goroutine cleanup via admissionControlPost

Maybe in your test you could utilize it and e.g. update routing table to an empty value just before test ends such that post-processor closes all instances.

Alternatively we may call cleanup manually on the post-postprocessor like it is done here https://github.com/zalando/skipper/blob/e57538ee7028f79baa0a285db268c4e7c3cf346e/filters/auth/main_test.go#L11-L17

AlexanderYastrebov: This is related to the long known but hard problem of filter resource cleanup https://github.com/zalando/skipper/issues/202

szuecs: How can you access the PostProcessor from the test?

proxy/failing_filter_test.go
```diff
@@ -0,0 +1,292 @@
...
+	}))
+	defer backend.Close()
+
+	var routes = fmt.Sprintf(`
+                main: * -> admissionControl("mygroup", "active", "1s", 5, 0, 0.99, 0.9, 1.0) -> "%s";
+        `, backend.URL)
+	r := eskip.MustParse(routes)
+
+	fr := make(filters.Registry)
+	fr.Register(shedder.NewAdmissionControl(shedder.Options{}))
```
AlexanderYastrebov, line 51: I think we need to create and configure postProcessor for it.

szuecs: Yes true that might help

szuecs: I tested it and it does not help. PostProcessor is only working on routing updates.

filters/filters.go
szuecs: I found a way to fix all the things. It did not require a TestMain hack.

proxy/proxy.go
```diff
@@ -1551,3 +1561,36 @@ func (p *Proxy) makeErrorResponse(ctx *context, perr *proxyError) {
...
+
+func (p *Proxy) applyFiltersOnError(ctx *context, filters []*routing.RouteFilter) {
+	filtersStart := time.Now()
+	filterTracing := p.tracing.startFilterTracing("response_filters", ctx)
+	defer filterTracing.finish()
+
+	for i := len(filters) - 1; i >= 0; i-- {
+		fi := filters[i]
+
+		if _, ok := fi.Filter.(errorHandlerFilter); !ok {
```
AlexanderYastrebov, line 1580: I think here we want
```go
if ehf, ok := fi.Filter.(errorHandlerFilter); !ok || !ehf.HandleErrorResponse() {
  continue
}
```

routing/datasource.go
```diff
@@ -585,6 +586,23 @@ func receiveRouteMatcher(o Options, out chan<- *routeTable, quit <-chan struct{}
...
+			}
+			for _, r := range routes {
+				for _, f := range r.Filters {
+					fc, ok := f.Filter.(filters.FilterCloser)
+					if ok {
+						fc.Close()
+					}
+				}
+			}
+
```
AlexanderYastrebov, line 605: I guess we can just keep a pointer to the routes:
```diff
diff --git a/routing/datasource.go b/routing/datasource.go
index 9b679df4..c2ab4ace 100644
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -520,6 +520,7 @@ func processRouteDefs(o Options, fr filters.Registry, defs []*eskip.Route) (rout
 
 type routeTable struct {
        m             *matcher
+       routes        []*Route
        validRoutes   []*eskip.Route
        invalidRoutes []*eskip.Route
        created       time.Time
@@ -574,6 +575,7 @@ func receiveRouteMatcher(o Options, out chan<- *routeTable, quit <-chan struct{}
 
                        rt = &routeTable{
                                m:             m,
+                               routes:        routes,
                                validRoutes:   validRoutes,
                                invalidRoutes: invalidRoutes,
                                created:       time.Now().UTC(),
@@ -585,6 +587,13 @@ func receiveRouteMatcher(o Options, out chan<- *routeTable, quit <-chan struct{}
                        updatesRelay = updates
                        outRelay = nil
                case <-quit:
+                       for _, route := range rt.routes {
+                               for _, f := range route.Filters {
+                                       if fc, ok := f.Filter.(filters.FilterCloser); ok {
+                                               fc.Close()
+                                       }
+                               }
+                       }
                        return
                }
        }
```

AlexanderYastrebov: I think we can do that in the startReceivingUpdates

```go
--- a/routing/routing.go
+++ b/routing/routing.go
@@ -371,6 +371,10 @@ func (r *Routing) startReceivingUpdates(o Options) {
                                }
                                r.log.Info("route settings applied")
                        case <-r.quit:
+                               rt := r.routeTable.Load().(*routeTable)
+                               if rt != nil {
+                                       // ...
+                               }
                                return
                        }
                }
```

proxy/proxy.go
szuecs: done

routing/datasource.go
szuecs: The problem with routeTable is that we do not have all routes here, only access to leafRoutes as far as I tried last week.

szuecs: I guess you suggest to add the routes to the routeTable to store the link to all the routes, which would be good enough to fix the problem, yes.

AlexanderYastrebov: yeah, keep the link and close filters from startReceivingUpdates

filters/filters.go
```diff
@@ -151,6 +151,15 @@ type Filter interface {
 	Response(FilterContext)
 }
 
+// FilterCloser are Filters that need to cleanup resources after
+// filter teminiation. For example Filters, that create a goroutine
+// for some reason need to cleanup their goroutine or they would leak
+// goroutines.
+type FilterCloser interface {
+	Filter
+	Close()
```
AlexanderYastrebov, line 160: Maybe it should be `Close() error` (i.e. io.Closer) before its too late - we can always ignore error but it would be hard to introduce it later.

routing/datasource.go
szuecs: Thanks for the input I put it into Routing.Close, because there it makes most sense and then we do not have the problem of nil ptr.
Let's see what the tests say.

routing/routing.go
```diff
@@ -422,6 +422,11 @@ func (r *Routing) Get() *RouteLookup {
 
 // Close closes routing, stops receiving routes.
 func (r *Routing) Close() {
+	var rt *routeTable
+	rt, ok := r.routeTable.Load().(*routeTable)
+	if ok {
+		rt.Close()
+	}
```
AlexanderYastrebov, line 429: I think it is better to put it into startReceivingUpdates into `case <-r.quit:` - this way we know there will be no updates anymore

routing/datasource.go
```diff
@@ -520,11 +521,27 @@ func processRouteDefs(o Options, fr filters.Registry, defs []*eskip.Route) (rout
...
+	once          sync.Once
+	routes        []*Route // only used for closing
 	validRoutes   []*eskip.Route
 	invalidRoutes []*eskip.Route
 	created       time.Time
 }
 
+// Close routeTable will cleanup all underlying resources, that could
+// leak goroutines.
+func (rt *routeTable) Close() {
```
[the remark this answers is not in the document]
szuecs: yes but it looks so ugly in case you have syntax highlighting, because `close` is an internal func like println. I agree will make it lower case, but I wait to finish the tests

routing/routing.go
szuecs: It's not possible because routeTable is nil, just check the 2-3 lines above the `<-quit` case, there we set `rt = nil`

AlexanderYastrebov: I mean in the startReceivingUpdates, not in the receiveRouteMatcher

szuecs: changed, let's see what tests are saying. One problem could be that it's async closing by another goroutine and goroutine leak detection might fail

docs/tutorials/development.md
```diff
@@ -137,6 +137,86 @@ func (f *myFilter) Response(ctx filters.FilterContext) {
 
 Find a detailed example at [how to develop a filter](../reference/development.md#how-to-develop-a-filter).
 
+### Filters with cleanup
```
AlexanderYastrebov, line 140: I am not sure we should document it here at this time as this might create confusion for the users.
This change only closes filters on the routing shutdown, i.e. once in the Skipper lifetime while filters could be recreated on each route update. Users may get confused and think Close will be called for each their filter instance that is not used anymore.

The reason it works for admission control is because there is a postprocesor that closes unused filters (which technically is still incorrect as there might be in-flight requests during route update)

I.e. it does not solve https://github.com/zalando/skipper/issues/202 but only makes a small step towards.

filters/filters.go
AlexanderYastrebov: With returned error you can decide to ignore it or handle. Without error there is no way to handle an error if any.

AlexanderYastrebov:
:+1:

filters/filters.go
szuecs: fixed

docs/tutorials/development.md
szuecs: good point about the incorrect closing. Maybe we should introduce a closing registry which will wait for timeouts like ResponseHeaderTimeout long before it closes the filters and we move in the PostProcessor the "Close" call to something like closingRegistry.Add(filterInstance). wdyt?

docs/tutorials/development.md
```diff
@@ -137,6 +137,86 @@ func (f *myFilter) Response(ctx filters.FilterContext) {
...
+
+func (f *myFilter) Request(ctx filters.FilterContext) {
+     // change data in ctx.Request() for example
+}
+
+func (f *myFilter) Response(ctx filters.FilterContext) {
+     // change data in ctx.Response() for example
+}
+
+func (f *myFilter) Close() {
```
AlexanderYastrebov, line 169: Should return error

filters/filters.go
szuecs: fixed

docs/tutorials/development.md
szuecs: a sync.Atomic as gauge per route?

docs/tutorials/development.md
szuecs: fixed

filters/filters.go
```diff
@@ -151,6 +151,15 @@ type Filter interface {
 	Response(FilterContext)
 }
 
+// FilterCloser are Filters that need to cleanup resources after
+// filter teminiation. For example Filters, that create a goroutine
```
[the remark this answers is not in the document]
szuecs: fixed

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2242, opened 2023-02-24
Pull redis endpoints from remote source

MustafaSaber:


net/redisclient.go
```diff
@@ -197,6 +203,31 @@ func NewRendezvousVnodes(shards []string) redis.ConsistentHash {
 	return rendezvousVnodes{rendezvous.New(vshards, xxhash.Sum64String), table}
 }
 
+func (r *RedisRingClient) updateIPsFromURL() func() []string {
+	return func() []string {
+		// TODO(sszuecs): make sure kubernetes dataclient is already initialized and
+		// has polled the data once or kdc.GetEndpointAdresses should be blocking
+		// call to kubernetes API
```
szuecs, line 210: you can drop the comment

net/redisclient.go
```diff
@@ -197,6 +203,31 @@ func NewRendezvousVnodes(shards []string) redis.ConsistentHash {
 	return rendezvousVnodes{rendezvous.New(vshards, xxhash.Sum64String), table}
 }
 
+func (r *RedisRingClient) updateIPsFromURL() func() []string {
```
szuecs, line 206: Why this is part of `net` package?

net/redisclient.go
```diff
@@ -249,11 +280,15 @@ func NewRedisRingClient(ro *RedisOptions) *RedisRingClient {
 		r.log = ro.Log
 		r.metricsPrefix = ro.MetricsPrefix
 
-		if ro.AddrUpdater != nil {
+		if ro.AddrUpdater != nil || ro.RedisAddress != "" {
```
szuecs, line 283: I think you should provide an AddrUpdater func that is able to do the remote fetch for you.

net/redisclient.go
szuecs: see also https://github.com/zalando/skipper/blob/437633129a60549e16d8cf93098b6301d06f5e28/skipper.go#L1279-L1291 , similar would be to add a func in skipper.go.

skipper.go
```diff
@@ -1563,6 +1587,9 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 			} else {
 				log.Errorf("Failed to find kubernetes dataclient, but redis shards should be get by kubernetes svc %s/%s", o.KubernetesRedisServiceNamespace, o.KubernetesRedisServiceName)
 			}
+		} else if redisOptions != nil && o.SwarmRedisRemote != "" {
+			// in case we can pull redis endpoints from URL.
```
szuecs, line 1591: just drop the comment

szuecs:
I miss a test. Maybe you can create a test similar to [skipper_test.go](https://github.com/zalando/skipper/blob/master/skipper_test.go#L387)

routesrv/testdata/redis-ip.json
```diff
@@ -0,0 +1 @@
+{"Endpoints":[{"Url":"10.2.5.7:6379"},{"Url":"10.2.55.7:6379"}]}
```
MustafaSaber, line 1: No newline to match json response

MustafaSaber: Remove newline with https://github.com/zalando/skipper/blob/ec12fa30dcf422a9bcbfce42e033d6ba7a7befa7/routesrv/routesrv_test.go#L95-L103

skipper.go
```diff
@@ -1292,6 +1295,38 @@ func getRedisUpdaterFunc(namespace, name string, kdc *kubernetes.Client) func()
 	}
 }
 
+func updateEndpointsFromURL(address string) func() []string {
+	return func() []string {
+		/* #nosec */
+		resp, err := http.Get(address)
+		if err != nil {
+			log.Errorf("failed to connect to redis endpoint %v, due to: %v", address, err)
```
szuecs, line 1303: in all error cases you can return nil to exit early or you likely end up in panic

skipper.go
```diff
@@ -1292,6 +1295,38 @@ func getRedisUpdaterFunc(namespace, name string, kdc *kubernetes.Client) func()
 	}
 }
 
+func updateEndpointsFromURL(address string) func() []string {
+	return func() []string {
+		/* #nosec */
+		resp, err := http.Get(address)
```
szuecs, line 1301: likely you should start the function closure by assigning `address` to a local var like:
```
address := address
resp, err := http.Get(address)
... 
```

api/types.go
```diff
@@ -0,0 +1,11 @@
+package api
```
szuecs, line 1: I think our conclusion was to add the types into `skipper` and `routesrv` packages, because we want to be client/server independent of each other.

skipper.go
MustafaSaber: tested and still fails, I saw `nosec` in couple of places

docs/tutorials/ratelimit.md
```diff
@@ -96,9 +96,19 @@ and your dataclient in use.
 
 This solution is independent of the dataclient being used.  You have
 to run N number of [Redis](https://redis.io) instances, where N is >
-0.  Specify `-swarm-redis-urls`, multiple instances can be separated
+0.
+
+Can be achieved with multiple ways.
```
szuecs, line 101: This sounds a bit disconnected. Maybe something like:
``` 
There are 2 different configurations to assign redis instances as a skipper redis swarm. 

1) static: ...
2) automatic: ...
```

routesrv/redishandler.go
```diff
@@ -12,25 +13,40 @@ type RedisHandler struct {
...
-	w.Header().Add("Content-Type", "text/plain; charset=utf-8")
+	w.Header().Add("Content-Type", "application/json")
 	w.Write(rh.AddrUpdater())
 }
 
 func getRedisAddresses(namespace, name string, kdc *kubernetes.Client) func() []byte {
 	return func() []byte {
-		data := make([]byte, 0)
+		endpoints := make([]RedisEndpoint, 0)
+		toreturn := RedisEndpoints{Endpoints: endpoints}
```
szuecs, line 36: toreturn -> redisEndpoints or eps

skipper.go
```diff
@@ -1292,6 +1294,49 @@ func getRedisUpdaterFunc(namespace, name string, kdc *kubernetes.Client) func()
...
+
+		target := &RedisEndpoints{}
+
+		err = json.Unmarshal(body, target)
+		if err != nil {
+			log.Errorf("Failed to decode body to json %v", err)
+			return nil
+		}
+
+		a := make([]string, 0)
```
szuecs, line 1329: Since we know the len, we can set the capacity:
`make([]string, 0, len(target.Endpoints))`

skipper_test.go
```diff
@@ -384,6 +385,268 @@ func Example_ratelimitRegistryBinding() {
...
+	go run(o, sigs, nil)
+
+	// wait for proxy being ready
+	ready = false
+	for !ready {
+		rsp, _ := http.DefaultClient.Get("http://localhost:9090/kube-system/healthz")
+		if rsp != nil && rsp.StatusCode == 200 {
+			ready = true
+		}
+		time.Sleep(100 * time.Millisecond)
```
szuecs, line 622: this should be in else block or you wait even if you are ready

skipper_test.go
```diff
@@ -384,6 +385,268 @@ func Example_ratelimitRegistryBinding() {
...
+	}
+	countOK, ok := va.CountStatus(http.StatusOK)
+	if countOK == 0 {
+		t.Fatalf("Some requests should have passed: %d %v", countOK, ok)
+	}
+
+	countLimited, ok := va.CountStatus(http.StatusTooManyRequests)
+	if !ok || countLimited < countOK {
+		t.Fatalf("count TooMany should be higher than OKs: %d < %d: %v", countLimited, countOK, ok)
+	}
```
szuecs, line 645: you have `clusterRatelimit("foo", 1, "1s")`  so you should check that you allow 5*1rps (+/-1 rps) and rest  StatusTooManyRequests

skipper_test.go
```diff
@@ -384,6 +385,268 @@ func Example_ratelimitRegistryBinding() {
...
+	rate := 10
+	sec := 5
+	va := httptest.NewVegetaAttacker("http://localhost:9090/test", rate, time.Second, time.Second)
+	va.Attack(io.Discard, time.Duration(sec)*time.Second, "mytest")
+	t.Logf("Success [0..1]: %0.2f", va.Success())
+
+	if successRate := va.Success(); successRate < 0.1 && successRate > 0.5 {
+		t.Fatalf("Test should have a success rate between %0.2f < %0.2f < %0.2f", 0.1, successRate, 0.5)
+	}
+	if reqCount := va.TotalRequests(); reqCount < uint64(rate*sec) {
```
szuecs, line 634: I guess `rate*sec`  should also have +/-1 (think epsilon)

skipper_test.go
```diff
@@ -384,6 +385,266 @@ func Example_ratelimitRegistryBinding() {
...
+		time.Sleep(100 * time.Millisecond)
+	}
+
+	rate := 10
+	sec := 5
+	va := httptest.NewVegetaAttacker("http://localhost:9090/test", rate, time.Second, time.Second)
+	va.Attack(io.Discard, time.Duration(sec)*time.Second, "mytest")
+	t.Logf("Success [0..1]: %0.2f", va.Success())
+
+	if successRate := va.Success(); successRate < 0.1 || successRate > 0.5 {
```
szuecs, line 629: here we check success rate is between 10% and 50%: interval notation: successRate in (0.1, 0.5).
total requests 10req*5s=50
filter allows 1RPS, so 5 allowed requests in 5s
Expected successRate: 5/50=1/10=0.1
So why do we test (0.1, 0.5) as success?
I would rather say we should test for 0.1 +/- \epsilon where you can tune \epsilon such that the test is not flaky, but shows that the test makes sense. So maybe \epsilon=0.05 for example

docs/tutorials/ratelimit.md
```diff
@@ -96,9 +96,21 @@ and your dataclient in use.
...
+
+There are 2 different configurations to assign redis instances as a skipper redis swarm.
+
+- Static:
+    - Specify `-swarm-redis-urls`, multiple instances can be separated
+    by `,`, for example: `-swarm-redis-urls=redis1:6379,redis2:6379`. For
+    running skipper in Kubernetes with this. Use this if you don't need to scale your redis instances. 
+
+- Automatic:
+    - Specify `-kubernetes-redis-service-namespace=default` and `-kubernetes-redis-service-name=redis`, this will create k8s dataclient which will take care of collecting redis instances endpoints. This will allow you to scale redis instances as much as you want.
```
szuecs, line 109: k8s -> kubernetes, no acronyms please

skipper.go
szuecs: if we don't fail then not

MustafaSaber:
Ran `for i in {1..15}; do go test -timeout 30s -run ^TestConcurrentKubernetesClusterStateAccessWithRemoteRedis$; done` to verify it's not flaky

skipper.go
```diff
@@ -1294,6 +1296,49 @@ func getRedisUpdaterFunc(namespace, name string, kdc *kubernetes.Client) func()
...
+		}
+
+		a := make([]string, 0, len(target.Endpoints))
+		for _, endpoint := range target.Endpoints {
+			a = append(a, endpoint.Url)
+		}
+
+		log.Infof("Redis updater called and found %d redis endpoints", len(a))
+
+		return a
```
szuecs, line 1338: in case target.Endpoints=nil because of an error to reach the remote endpoints, we need to check that we don't change anything in net.RedisClient or wherever we use the func returned from updateEndpointsFromURL(). IIRC this is passed to redisclient UpdateFunc or whatever the name was.
What we need to omit is that we turn to 0 redis instances in case of an error.

szuecs:
@MustafaSaber you are getting close to it. Only one validation that is important at runtime.

net/redisclient.go
```diff
@@ -299,7 +299,7 @@ func (r *RedisRingClient) startUpdater(ctx context.Context) {
 		}
 
 		addrs := r.options.AddrUpdater()
-		if !hasAll(addrs, old) {
+		if addrs != nil && !hasAll(addrs, old) {
```
MustafaSaber, line 302: Chose `addrs != nil` instead of `len(addrs) != 0` because function may return empty list of redis for other reasons, So in this case those are 2 different cases.

skipper.go
szuecs: perfect, thanks!

szuecs:
:+1:

szuecs:
@AlexanderYastrebov do you want to check this PR, too?

docs/tutorials/ratelimit.md
```diff
@@ -96,9 +96,21 @@ and your dataclient in use.
...
-by `,`, for example: `-swarm-redis-urls=redis1:6379,redis2:6379`. For
-running skipper in Kubernetes with this, see also [Running with
+0.
+
+There are 2 different configurations to assign redis instances as a skipper redis swarm.
+
+- Static:
+    - Specify `-swarm-redis-urls`, multiple instances can be separated
+    by `,`, for example: `-swarm-redis-urls=redis1:6379,redis2:6379`. For
+    running skipper in Kubernetes with this. Use this if you don't need to scale your redis instances.
```
AlexanderYastrebov, line 106: > For running skipper in Kubernetes with this.

This sentence feels a bit out of place

docs/tutorials/ratelimit.md
```diff
@@ -96,9 +96,21 @@ and your dataclient in use.
 
 This solution is independent of the dataclient being used.  You have
 to run N number of [Redis](https://redis.io) instances, where N is >
-0.  Specify `-swarm-redis-urls`, multiple instances can be separated
-by `,`, for example: `-swarm-redis-urls=redis1:6379,redis2:6379`. For
-running skipper in Kubernetes with this, see also [Running with
+0.
+
+There are 2 different configurations to assign redis instances as a skipper redis swarm.
```
AlexanderYastrebov, line 101: Or is it three?
- static
- via k8s service selector
- via http endpoint

routesrv/redishandler.go
```diff
@@ -12,25 +13,40 @@ type RedisHandler struct {
 	AddrUpdater func() []byte
 }
 
+type RedisEndpoint struct {
+	URL string `json:"url"`
```
AlexanderYastrebov, line 17: Is it url really or maybe it should be `address`?

routesrv/routesrv_test.go
```diff
@@ -227,7 +228,7 @@ func TestRedisIPs(t *testing.T) {
 
 	w := getRedisURLs(rs)
 
-	want := parseRedisIP(t, "testdata/redis-ip.txt")
+	want := parseRedisIP(t, "testdata/redis-ip.json")
 	got := w.Body.Bytes()
 
 	if !bytes.Equal(got, want) {
```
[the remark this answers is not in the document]
AlexanderYastrebov: Also I think we should check status code first.

AlexanderYastrebov:
> do you want to check this PR, too?

Posted couple of comments, my main concern is error handling

skipper.go
```diff
@@ -1294,6 +1296,49 @@ func getRedisUpdaterFunc(namespace, name string, kdc *kubernetes.Client) func()
...
+			log.Errorf("failed to read to redis response %v", err)
+			return nil
+		}
+
+		target := &RedisEndpoints{}
+
+		err = json.Unmarshal(body, target)
+		if err != nil {
+			log.Errorf("Failed to decode body to json %v", err)
+			return nil
```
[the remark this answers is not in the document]
MustafaSaber: it doesn't reset `SetAddrs` function will ignore it len is 0 https://github.com/zalando/skipper/blob/693a042dcfdeb771930fa86eac2826f9d7ee9ae3/net/redisclient.go#L369-L374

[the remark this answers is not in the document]
szuecs: @AlexanderYastrebov that's only true in case you want to support scale down to 0, which is not really working anywhere. So actually it's fine to not change the interface.

routesrv/redishandler.go
```diff
@@ -12,25 +13,40 @@ type RedisHandler struct {
...
 		w.WriteHeader(http.StatusMethodNotAllowed)
 		return
 	}
-	w.Header().Add("Content-Type", "text/plain; charset=utf-8")
+	w.Header().Add("Content-Type", "application/json")
 	w.Write(rh.AddrUpdater())
 }
 
 func getRedisAddresses(namespace, name string, kdc *kubernetes.Client) func() []byte {
 	return func() []byte {
```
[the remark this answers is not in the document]
AlexanderYastrebov: In general if there is an error the service should respond with a proper status code. Then we can monitor it on the client side if needed. Currently we signal error by an empty list which is not obvious. It also does not allow us to classify errors unlike status codes.

routesrv/redishandler.go
```diff
@@ -12,25 +13,40 @@ type RedisHandler struct {
...
-		data := make([]byte, 0)
+		endpoints := make([]RedisEndpoint, 0)
+		RedisEndpoints := RedisEndpoints{Endpoints: endpoints}
```
AlexanderYastrebov, line 36: Maybe there is not need to shadow type name here, just
```go
result := RedisEndpoints{}
```

routesrv/redishandler.go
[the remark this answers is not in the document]
szuecs: Yes true, we could also just have an error counter and monitor it like that.
Then 5xx without error counter increase means some underlying (Go stdlib/runimte) things had an error.

MustafaSaber:
Any other thoughts @szuecs @AlexanderYastrebov ?

AlexanderYastrebov:
> Any other thoughts

I'd still prefer that we serve 500 error in case we can not obtain Redis addresses but I'll approve so we can move on.

MustafaSaber:
👍

PR #2244, opened 2023-02-27
filters/auth/grant: configurable tokeninfo keys

AlexanderYastrebov:

### DIFF
--- a/filters/auth/grant.go
+++ b/filters/auth/grant.go
@@ -126,29 +126,43 @@ func (f *grantFilter) refreshTokenIfRequired(c *cookie, ctx filters.FilterContex
 	}
 }
 
-func (f *grantFilter) setAccessTokenHeader(req *http.Request, token string) {
+func (f *grantFilter) setupToken(token *oauth2.Token, tokeninfo map[string]interface{}, ctx filters.FilterContext) error {
 	if f.config.AccessTokenHeaderName != "" {
-		req.Header.Set(f.config.AccessTokenHeaderName, authHeaderPrefix+token)
+		ctx.Request().Header.Set(f.config.AccessTokenHeaderName, authHeaderPrefix+token.AccessToken)
 	}
-}
 
-func (f *grantFilter) createTokenContainer(token *oauth2.Token, tokeninfo map[string]interface{}) (tokenContainer, error) {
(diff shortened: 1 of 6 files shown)


filters/auth/grantconfig.go
```diff
@@ -84,6 +84,9 @@ type OAuthConfig struct {
 	// on the request to a header with this name.
 	AccessTokenHeaderName string
 
+	// GrantTokeninfoKeys, optional. When not empty, keys not in this list are removed from the tokeninfo map.
+	GrantTokeninfoKeys []string
```
AlexanderYastrebov, line 88: Maybe `map[string]struct{}`?

szuecs: Pass it as []string and convert it to map[string]struct{} in CreateFilter ?

filters/auth/grant.go
```diff
@@ -126,29 +126,43 @@ func (f *grantFilter) refreshTokenIfRequired(c *cookie, ctx filters.FilterContex
...
+	}
+
+	// By piggy-backing on the OIDC token container,
+	// we gain downstream compatibility with the oidcClaimsQuery filter.
+	ctx.StateBag()[oidcClaimsCacheKey] = tokenContainer{
 		OAuth2Token: token,
 		Subject:     subject,
 		Claims:      tokeninfo,
-	}, nil
+	}
```
AlexanderYastrebov, line 159: @szuecs 
I just moved this piece around without changing. Token container is stored in statebag to make https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#grantclaimsquery work which is an alias for oidcClaimsQuery. I agree it was not the best idea but it slipped through the review back then.

I can fix it in the followup PR - we'd need to make `oidcClaimsQuery` parameterized on key value. It could also break some configurations if users use `oidcClaimsQuery` instead of `grantClaimsQuery` after `oauthGrant`.

filters/auth/grantconfig.go
AlexanderYastrebov: Added a lookup map to the config

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2245, opened 2023-03-02
feature: expose number of known redis shards

szuecs:
feature: expose number of known redis shards

### DIFF
--- a/net/redisclient.go
+++ b/net/redisclient.go
@@ -299,10 +299,10 @@ func (r *RedisRingClient) startUpdater(ctx context.Context) {
 		}
 
 		addrs := r.options.AddrUpdater()
-		r.metrics.UpdateGauge(r.metricsPrefix+"shards", float64(len(addrs)))
 		if !hasAll(addrs, old) {
 			r.log.Infof("Redis updater updating old(%d) != new(%d)", len(old), len(addrs))
 			r.SetAddrs(ctx, addrs)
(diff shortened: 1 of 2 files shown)


net/redisclient.go
```diff
@@ -299,6 +299,7 @@ func (r *RedisRingClient) startUpdater(ctx context.Context) {
 		}
 
 		addrs := r.options.AddrUpdater()
+		r.metrics.UpdateGauge(r.metricsPrefix+"shards", float64(len(addrs)))
```
AlexanderYastrebov, line 302: Maybe we can add `r.metrics.UpdateGauge(r.metricsPrefix+"shards", float64(r.ring.Len()))` to the `StartMetricsCollection` instead?

AlexanderYastrebov: OTOH here it would be an instant update. Lets maybe have `r.metrics.UpdateGauge(r.metricsPrefix+"shards", float64(r.ring.Len()))` right after  `r.SetAddrs(ctx, addrs)`

szuecs: But if old and addrs is both 0 from the start the gauge will not change.
I thought it might make sense to have this independent of the internals, but I think you are right to put it 4lines down and ask the ring for the length.

AlexanderYastrebov:
:+1:

MustafaSaber:
👍

PR #2246, opened 2023-03-03
filters/auth/grant: support client id and secret file placeholders

AlexanderYastrebov:
This change enables `{host}` placeholder in the client id and secret filenames.

E.g. for the request to `foo.example.org` when flag values are `-oauth2-client-id-file=/var/run/secrets/{host}-client-id` and `-oauth2-client-secret-file=/var/run/secrets/{host}-client-secret` the client id and secret files would be `/var/run/secrets/foo.example.org-client-id` and `/var/run/secrets/foo.example.org-client-secret` respectively.

### DIFF
--- a/filters/auth/grant_test.go
+++ b/filters/auth/grant_test.go
@@ -1,12 +1,16 @@
 package auth_test
 
 import (
+	"context"
 	"crypto/tls"
 	"encoding/json"
 	"net"
 	"net/http"
+	"net/http/cookiejar"
 	"net/http/httptest"
 	"net/url"
+	"os"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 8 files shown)


filters/auth/grant_test.go
```diff
@@ -35,6 +40,42 @@ const (
...
+
+func (rt *loggingRoundTripper) RoundTrip(req *http.Request) (resp *http.Response, err error) {
+	rt.t.Logf("\n%v", rt.requestString(req))
+
+	resp, err = rt.RoundTripper.RoundTrip(req)
+
+	if err == nil {
+		rt.t.Logf("\n%v", rt.responseString(resp))
+	} else {
+		rt.t.Logf("response err: %v", err)
```
szuecs, line 56: nice

szuecs:
:+1:

AlexanderYastrebov:
:+1:

Issue #2175, opened 2023-01-02
refactor: config to FlagSet for testability


szuecs:
context: https://github.com/zalando/skipper/pull/2167 and https://github.com/zalando/skipper/pull/2167#discussion_r1060135429 

[This blogpost](https://eli.thegreenplace.net/2020/testing-flag-parsing-in-go-programs/) wrote that we could use flag.FlagSet or invoke the cli itself to properly test flags

The change should keep the interface to not break any users (library or binary users)

MustafaSaber:
I think this is done right @RomanZavodskikh ?

RomanZavodskikh:
Yes.

PR #2248, opened 2023-03-14
Use flags.FlagSet to increase testability

RomanZavodskikh:
Such change makes skipper and routesrv configuration more incapsulated, os.Args slice is not used directly in tests any more.

Updates #2175

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -27,6 +27,7 @@ import (
 
 type Config struct {
 	ConfigFile string
+	flags      *flag.FlagSet
 
 	// generic:
 	Address                         string         `yaml:"address"`
@@ -300,6 +301,7 @@ func NewConfig() *Config {
 	cfg.LuaModules = commaListFlag()
 	cfg.Oauth2GrantTokeninfoKeys = commaListFlag()
 
+	flag := flag.NewFlagSet("", flag.ExitOnError)
 	flag.StringVar(&cfg.ConfigFile, "config-file", "", "if provided the flags will be loaded/overwritten by the values on the file (yaml)")
 
 	// generic:
@@ -536,6 +538,7 @@ func NewConfig() *Config {
 
 	flag.Var(cfg.LuaModules, "lua-modules", "comma separated list of lua filter modules. Use <module>.<symbol> to selectively enable module symbols, for example: package,base._G,base.print,json")
 
+	cfg.flags = flag
 	return cfg
 }
 
@@ -560,11 +563,19 @@ func validate(c *Config) error {
 }
 
 func (c *Config) Parse() error {
-	flag.Parse()
+	return c.ParseArgs(os.Args[0], os.Args[1:])
+}
+
+func (c *Config) ParseArgs(progname string, args []string) error {
+	c.flags.Init(progname, flag.ExitOnError)
+	err := c.flags.Parse(args)
+	if err != nil {
+		return err
+	}
 
 	// check if arguments were correctly parsed.
-	if len(flag.Args()) != 0 {
-		return fmt.Errorf("invalid arguments: %s", flag.Args())
+	if len(c.flags.Args()) != 0 {
+		return fmt.Errorf("invalid arguments: %s", c.flags.Args())
 	}
 
 	configKeys := make(map[string]interface{})
@@ -581,10 +592,13 @@ func (c *Config) Parse() error {
 
 		_ = yaml.Unmarshal(yamlFile, configKeys)
 
-		flag.Parse()
+		err = c.flags.Parse(args)
+		if err != nil {
+			return err
+		}
 	}
 
-	checkDeprecated(configKeys,
+	c.checkDeprecated(configKeys,
 		"enable-prometheus-metrics",
 		"api-usage-monitoring-default-client-tracking-pattern",
 		"enable-kubernetes-east-west",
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -559,12 +562,12 @@ func validate(c *Config) error {
 	return c.parseForwardedHeaders()
 }
 
-func (c *Config) Parse() error {
-	flag.Parse()
+func (c *Config) Parse(args []string) error {
+	c.Flags.Parse(args)
```
AlexanderYastrebov, line 566: Lets check error here

config/config.go
```diff
@@ -300,242 +301,244 @@ func NewConfig() *Config {
 	cfg.LuaModules = commaListFlag()
 	cfg.Oauth2GrantTokeninfoKeys = commaListFlag()
 
-	flag.StringVar(&cfg.ConfigFile, "config-file", "", "if provided the flags will be loaded/overwritten by the values on the file (yaml)")
+	flags := flag.NewFlagSet(progname, flag.ContinueOnError)
```
AlexanderYastrebov, line 304: Global flags use `ExitOnError`

config/config.go
```diff
@@ -27,6 +27,7 @@ import (
 
 type Config struct {
 	ConfigFile string
+	Flags      *flag.FlagSet
```
AlexanderYastrebov, line 30: Do we need this to be exported field?

cmd/skipper/main.go
```diff
@@ -29,8 +30,8 @@ var (
 )
 
 func main() {
-	cfg := config.NewConfig()
-	if err := cfg.Parse(); err != nil {
+	cfg := config.NewConfig("skipper")
```
AlexanderYastrebov, line 33: Why not `os.Args[0]`?

config/config.go
AlexanderYastrebov, line 304: I think we can use 
```go
flag := flag.NewFlagSet(progname, flag.ContinueOnError)
```
It will mask package name but will keep following flag definitions intact

config/config.go
```diff
@@ -277,7 +278,7 @@ const (
 	redisPasswordEnv = "SWARM_REDIS_PASSWORD"
 )
 
-func NewConfig() *Config {
+func NewConfig(progname string) *Config {
```
AlexanderYastrebov, line 281: This would be an API change.
Lets use empty name and then call https://pkg.go.dev/flag#FlagSet.Init later in Parse (before `c.Flags.Parse(args)`)

config/config.go
```diff
@@ -559,12 +562,12 @@ func validate(c *Config) error {
 	return c.parseForwardedHeaders()
 }
 
-func (c *Config) Parse() error {
-	flag.Parse()
+func (c *Config) Parse(args []string) error {
```
AlexanderYastrebov, line 565: This would be API change.

Lets add a new method 
```go
func (c *Config) ParseArgs(name, args []string) error 
```
and call change existing one to the 
```go
func (c *Config) Parse() error {
    return c.ParseArgs(os.Args[0], os.Args[1:])
}
```

config/config.go
MustafaSaber: If it will only change in the function scope then I think it's ok but we meed to make sure it works

config/config.go
RomanZavodskikh: No, we don't.

cmd/skipper/main.go
RomanZavodskikh: I thought maybe it would be better to explicitly tell that that's the "skipper" or "routesrv" command regardless of how user names binary on his/her machine. But okay, I think it's more convenient to use `os.Args[0]`.

config/config_test.go
```diff
@@ -50,12 +45,11 @@ func TestEnvOverrides_SwarmRedisPassword(t *testing.T) {
...
 	} {
 		t.Run(tt.name, func(t *testing.T) {
-			// Run test in non-concurrent way to avoid collisions of other test cases that parse config file
-			os.Args = tt.args
 			if tt.env != "" {
 				t.Setenv(redisPasswordEnv, tt.env)
 			}
-			err := cfg.Parse()
+			cfg := NewConfig()
+			err := cfg.ParseArgs(tt.args[0], tt.args[1:])
```
MustafaSaber, line 52:
```suggestion
			err := cfg.Parse(tt.args)
```

RomanZavodskikh: Adding arguments to `cfg.Parse` function would cause API change which we would like to avoid to maintain backcompatibility.

config/config.go
AlexanderYastrebov: Nevertheless lets return possible error here. We may also consider using `flag.ContinueOnError`

config/config.go
```diff
@@ -560,11 +563,16 @@ func validate(c *Config) error {
 }
 
 func (c *Config) Parse() error {
-	flag.Parse()
+	return c.ParseArgs(os.Args[0], os.Args[1:])
+}
+
+func (c *Config) ParseArgs(progname string, args []string) error {
+	c.flags.Init(progname, flag.ExitOnError)
+	c.flags.Parse(args)
```
AlexanderYastrebov, line 571:
```go
err := c.flags.Parse(args)
if err != nil {
  return err
}
```
or at least hint that we ignore it 
```go
_ = c.flags.Parse(args)
```

config/config.go
```diff
@@ -581,10 +589,10 @@ func (c *Config) Parse() error {
 
 		_ = yaml.Unmarshal(yamlFile, configKeys)
 
-		flag.Parse()
+		c.flags.Parse(args)
```
AlexanderYastrebov, line 592: https://github.com/zalando/skipper/pull/2248/files#r1140031366

config/config.go
RomanZavodskikh: Okay, I missed this comment somehow: https://github.com/zalando/skipper/pull/2248#discussion_r1137497631
Let's returned error for the possible future use case when `ContinueOnError` would be used.

AlexanderYastrebov:
@RomanZavodskikh Lets wait for @szuecs to have a look

szuecs:
:+1:

Issue #1858, opened 2021-09-17
Credentials folder path should detect addition and removal of files


AlexanderYastrebov:
**Is your feature request related to a problem? Please describe.**

Users of [`bearerinjector`](https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#bearerinjector) may configure **folders** and **files** to look secrets at via `-credentials-paths` option. 
When **folder** is configured Skipper lists all the files and re-reads their content periodically (configured via `-credentials-update-interval`).
Newly added files are not detected automatically and require Skipper restart.
Removal of pre-existing file from the folder produces periodic error log due to refresh failure.

**Describe the solution you would like**

For every configured **folder** Skipper should re-read folder content to detect newly added files.
Files removed from the **folder** should not produce errors (to be discussed).
Note that configured **file** paths should continue to produce periodic error log on file removal.
The removal of configured **folder** itself should produce periodic error log as well.

The idea is that for dynamically added/removed files users configure **folder** path and for expected static files users configure **file** path and they should see the error if expected file has been gone.

**Would you like to work on it?**
Yes

szuecs:
What if a single instance has a missing secret file which is required to run?
Right now we would return an error and after change we would write error logs.
Maybe it's better to crash if file is not there or start shutdown procedure. 
Maybe it's also just ok to log, but it would contradict the self healing approach if we would just stop in case files are not there.

PR #2249, opened 2023-03-15
secrets: discover files added or removed to the directory path

AlexanderYastrebov:
This change tracks registered paths instead of secret files and re-reads directory contents on each refresh cycle to discover added or removed files.

- [x] requires https://github.com/zalando/skipper/pull/2252

Fixes https://github.com/zalando/skipper/issues/1858

### DIFF
--- a/secrets/file.go
+++ b/secrets/file.go
@@ -1,15 +1,15 @@
 package secrets
 
 import (
+	"bytes"
 	"errors"
 	"os"
 	"path/filepath"
-	"reflect"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	log "github.com/sirupsen/logrus"
-	"golang.org/x/sync/syncmap"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


szuecs:
How can you be sure that tests are working as expected if you change tests and code at the same time in the same commit?

One change we have in the pr is: The symlink was resolved. Now we would just follow the symlink wherever it points to.

AlexanderYastrebov:
> How can you be sure that tests are working as expected

Well, I wrote, read and run them as well as tested skipper locally :)
I would like to refactor tests to look like scenarios. I can split this PR in two: refactor tests and then implement added/removed file discovery.

> One change we have in the pr is: The symlink was resolved. Now we would just follow the symlink wherever it points to.

There is no change, symlink and regular files were treated the same way previously and I only combined two switch cases:
https://github.com/zalando/skipper/blob/2bcde1d5444987897b2ff20f9bb2f90bebe850bf/secrets/file.go#L100-L109

AlexanderYastrebov:
> I can split this PR in two: refactor tests and then implement added/removed file discovery.

I've created https://github.com/zalando/skipper/pull/2252 that refactors tests - it should better demonstrate subsequent changes done by this PR.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2250, opened 2023-03-16
webhook: remove kubernetes dependency

AlexanderYastrebov:


VERSION
```diff
@@ -1 +1 @@
-v0.15
+v0.16
```
RomanZavodskikh, line 1: Why do we need to up? Because of `type Admitter interface` change?

AlexanderYastrebov: Yeah, I thought we should bump minor version due to dependency change

cmd/webhook/admission/definitions.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type GroupVersionResource struct {
+	Group    string `json:"group"`
+	Version  string `json:"version"`
+	Resource string `json:"resource"`
+}
+
+type TypeMeta struct {
+	Kind       string `json:"kind,omitempty"`
```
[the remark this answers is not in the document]
AlexanderYastrebov: No, those comments for k8s codegenerator which we do not use

cmd/webhook/admission/definitions.go
```diff
@@ -0,0 +1,57 @@
...
+	APIVersion string `json:"apiVersion,omitempty"`
+}
+
+type ObjectMeta struct {
+	Name         string `json:"name,omitempty"`
+	GenerateName string `json:"generateName,omitempty"`
```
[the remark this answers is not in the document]
AlexanderYastrebov: Yeah, I only declared fields that we need

RomanZavodskikh:
:+1:

cmd/webhook/admission/definitions.go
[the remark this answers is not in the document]
MustafaSaber: Shouldn't this put an overhead of maintaining these fields later?

VERSION
szuecs: Do you think it's relevant for library users?
Normally I would increase minor version to notify library users.

AlexanderYastrebov: Technically it breaks API (changes Admitter interface).
I do not think anyone uses this package, maybe we should also unexport its symbols

AlexanderYastrebov: > maybe we should also unexport its symbols

I've made symbols unexported

szuecs:
:+1:

PR #2254, opened 2023-03-22
secrets: do not log subdirectory read errors

AlexanderYastrebov:
Updates https://github.com/zalando/skipper/issues/1858

### DIFF
--- a/secrets/file.go
+++ b/secrets/file.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


secrets/file.go
```diff
@@ -167,7 +168,7 @@ func addPath(secrets secretMap, path string) {
 		}
 		for _, match := range matches {
 			data, err := readSecretFile(match)
-			if err != nil {
+			if err != nil && !errors.Is(err, syscall.EISDIR) {
```
AlexanderYastrebov, line 171: @szuecs 
I've played with https://pkg.go.dev/os#Readlink and https://pkg.go.dev/path/filepath#EvalSymlinks but that becomes a mess quickly so I think it is easier to just mute `Failed to read path /etc/skipper/hostname-credentials/..data: read /etc/skipper/hostname-credentials/..data: is a directory` errors.

See also https://stackoverflow.com/questions/18062026/resolve-symlinks-in-go

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2257, opened 2023-03-27
skipper: fix handling of empty value in listFlag

AlexanderYastrebov:
#0

### DIFF
--- a/config/listflag.go
+++ b/config/listflag.go
@@ -34,8 +34,13 @@ func (lf *listFlag) Set(value string) error {
 		return nil
 	}
 
-	lf.value = value
-	lf.values = strings.Split(value, lf.sep)
+	if value == "" {
+		lf.value = ""
+		lf.values = nil
+	} else {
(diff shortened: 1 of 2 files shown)


MustafaSaber:
👍

config/listflag.go
```diff
@@ -34,8 +34,13 @@ func (lf *listFlag) Set(value string) error {
 		return nil
 	}
 
-	lf.value = value
-	lf.values = strings.Split(value, lf.sep)
+	if value == "" {
```
szuecs, line 37: this case is all defaults, and you should better test for only the `!=` case and put the code from the else branch here.

AlexanderYastrebov: Do you suggest to just reverse the condition or omit else branch?
It should reset previous value so we'll have else branch anyway.

szuecs: If it needs to reset, it's fine like that

szuecs:
:+1:

Issue #2262, opened 2023-03-31
Allow only lua files, no inline scripts


szuecs:
**Is your feature request related to a problem? Please describe.**

We want to be able to not allow lua inline scripts but allow only lua files bundled with the ingress component

**Describe the solution you would like**

Allow to choose if you want only lua files or script or nothing or all by flag : `-lua-sources=""|"file"|"inline"|"file,inline">`

PR #2264, opened 2023-04-05
feature: lua config to selective enable inline,file,none

szuecs:
feature that fixes https://github.com/zalando/skipper/issues/2262

config option -lua-sources allows to set:
- `-lua-sources=none` to disable lua sources
- `-lua-sources=file` to only allow file sources
- `-lua-sources=inline` to only allow inline scripts
- `-lua-sources=inline -lua-sources=file` to allow both, which is also the default as before.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -301,6 +301,7 @@ func NewConfig() *Config {
 	cfg.ForwardedHeadersExcludeCIDRList = commaListFlag()
 	cfg.CompressEncodings = commaListFlag("gzip", "deflate", "br")
 	cfg.LuaModules = commaListFlag()
+	cfg.LuaSources = commaListFlag()
 	cfg.Oauth2GrantTokeninfoKeys = commaListFlag()
 
 	flag := flag.NewFlagSet("", flag.ExitOnError)
--- a/config/config.go
+++ b/config/config.go
@@ -269,7 +269,7 @@ type Config struct {
 	ClusterRatelimitMaxGroupShards int `yaml:"cluster-ratelimit-max-group-shards"`
 
 	LuaModules *listFlag `yaml:"lua-modules"`
-	LuaSources string    `yaml:"lua-sources"`
+	LuaSources *listFlag `yaml:"lua-sources"`
 }
 
 const (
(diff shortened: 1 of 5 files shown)


script/script.go
```diff
@@ -51,16 +62,29 @@ func NewLuaScript() filters.Spec {
...
 // NewLuaScriptWithOptions creates a new filter spec with options
 func NewLuaScriptWithOptions(opts LuaOptions) (filters.Spec, error) {
-	return &luaScript{opts.Modules}, nil
+	sources := opts.Sources
+	if len(sources) == 0 {
+		// backwards compatible default, allow all
+		sources = append(sources, "file", "inline")
+	} else if contains("none", opts.Sources) {
+		sources = nil
+	}
```
AlexanderYastrebov, line 71: I think we do not need this special casing as we check if list contains specific values

config/config.go
```diff
@@ -269,6 +269,7 @@ type Config struct {
 	ClusterRatelimitMaxGroupShards int `yaml:"cluster-ratelimit-max-group-shards"`
 
 	LuaModules *listFlag `yaml:"lua-modules"`
+	LuaSources string    `yaml:"lua-sources"`
```
AlexanderYastrebov, line 272: Should be listFlag I think

script/script.go
szuecs: it's better for testing errors

config/config.go
```diff
@@ -539,6 +540,7 @@ func NewConfig() *Config {
 	flag.IntVar(&cfg.ClusterRatelimitMaxGroupShards, "cluster-ratelimit-max-group-shards", 1, "sets the maximum number of group shards for the clusterRatelimit filter")
 
 	flag.Var(cfg.LuaModules, "lua-modules", "comma separated list of lua filter modules. Use <module>.<symbol> to selectively enable module symbols, for example: package,base._G,base.print,json")
+	flag.StringVar(&cfg.LuaSources, "lua-sources", "", `comma separated list of lua input types for the lua() filter. Valid sources <""|"file"|"inline"|"file,inline"|"none">. Use "file" to only allow lua file references in lua filter. It defaults to "", that allows is the same as "file,inline". "none" disable lua filters.`)
```
AlexanderYastrebov, line 543: Maybe `Default "" is the same as "file,inline". Use "none" to disable lua filters.`

config/config.go
AlexanderYastrebov: Lets use listFlag as it supports proper splitting and handling of empty value

config/config.go
szuecs: fixed

skipper.go
```diff
@@ -1706,8 +1715,12 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 		o.CustomFilters = append(o.CustomFilters, compress)
 	}
 
-	if len(o.LuaModules) > 0 {
-		lua, err := script.NewLuaScriptWithOptions(script.LuaOptions{Modules: o.LuaModules})
+	if len(o.LuaModules) > 0 && len(o.LuaSources) > 0 {
```
AlexanderYastrebov, line 1718: Here it should be OR condition. We have default config for lua in builtin filters and this one overrides it

config/config.go

skipper.go
szuecs: the condition I dropped, because that's not correct anymore, because default empty list and disable is []string{"none"}

skipper.go
```diff
@@ -1707,7 +1717,10 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	}
 
 	if len(o.LuaModules) > 0 {
```
AlexanderYastrebov, line 1719: I thin I should be `if len(o.LuaModules) > 0 || len(o.LuaSources) > 0)`

szuecs: @AlexanderYastrebov  This will change the result for library users, because `len(o.LuaSources) == 0` for the default.

szuecs: I dropped the condition as we talked

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2266, opened 2023-04-12
fix: fifo() filter does not account for client timeouts

szuecs:
fix: fifo() filter does not account for client timeouts leading to longer downtime than required

### DIFF
--- a/filters/scheduler/fifo.go
+++ b/filters/scheduler/fifo.go
@@ -151,3 +151,8 @@ func (f *fifoFilter) Response(ctx filters.FilterContext) {
 	pending[last]()
 	ctx.StateBag()[fifoKey] = pending[:last]
 }
+
+// HandleErrorResponse is to opt-in for filters to get called
+// Response(ctx) in case of errors via proxy. It has to return true to
+// opt-in.
+func (f *fifoFilter) HandleErrorResponse() bool { return true }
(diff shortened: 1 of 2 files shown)


filters/scheduler/fifo.go
```diff
@@ -151,3 +151,8 @@ func (f *fifoFilter) Response(ctx filters.FilterContext) {
 	pending[last]()
 	ctx.StateBag()[fifoKey] = pending[:last]
 }
+
+// HandleErrorResponse is to opt-in for filters to get called
+// Response(ctx) in case of errors via proxy. It has to return true to
+// opt-in.
+func (f *fifoFilter) HandleErrorResponse() bool { return true }
```
RomanZavodskikh, line 158: Where this function is called from?

RomanZavodskikh: https://github.com/zalando/skipper/pull/2239

RomanZavodskikh:
:+1:

filters/scheduler/fifo.go
szuecs: proxy/proxy.go:1580:            if ehf, ok := fi.Filter.(errorHandlerFilter); !ok || !ehf.HandleErrorResponse() {

szuecs:
Hmm I can't reproduce the error in a test, without commenting https://github.com/zalando/skipper/blob/master/proxy/proxy.go#L1035-L1038 , which handles the issue I thought could have happened....

PR #2267, opened 2023-04-13
add openssf scorecard and auto update GH action

szuecs:
add openssf scorecard and auto update GH action
doc fix: cii best practices is called openssf best practices nowadays

Following https://github.com/ossf/scorecard-action#installation


MustafaSaber:
👍

PR #2269, opened 2023-04-14
Provide TraceID in the logs

RomanZavodskikh:

### DIFF
--- a/filters/filters.go
+++ b/filters/filters.go
@@ -117,6 +117,16 @@ type FilterContext interface {
 
 	// Performs a new route lookup and executes the matched route if any
 	Loopback()
+
+	Logger() FilterContextLogger
+}
+
+// FilterContextLogger is the logger which logs messages with additional context information.
+type FilterContextLogger interface {
+	Debugf(format string, args ...interface{})
+	Infof(format string, args ...interface{})
+	Warnf(format string, args ...interface{})
+	Errorf(format string, args ...interface{})
 }
 
 // Metrics provides possibility to use custom metrics from filter implementations. The custom metrics will
--- a/filters/filtertest/filtertest.go
+++ b/filters/filtertest/filtertest.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/opentracing/opentracing-go"
 	"github.com/zalando/skipper/filters"
+
+	log "github.com/sirupsen/logrus"
 )
 
 // Noop filter, used to verify the filter name and the args in the route.
@@ -58,6 +60,10 @@ func (fc *Context) Tracer() opentracing.Tracer {
 }
 func (fc *Context) ParentSpan() opentracing.Span {
 	return opentracing.StartSpan("test_span")
+}
+
+func (fc *Context) Logger() filters.FilterContextLogger {
+	return log.StandardLogger()
 }
 
 func (fc *Context) Serve(resp *http.Response) {
--- a/proxy/context.go
+++ b/proxy/context.go
@@ -13,6 +13,9 @@ import (
 	"github.com/zalando/skipper/filters"
 	"github.com/zalando/skipper/metrics"
 	"github.com/zalando/skipper/routing"
+	"github.com/zalando/skipper/tracing"
+
+	log "github.com/sirupsen/logrus"
 )
 
 const unknownHost = "_unknownhost_"
@@ -205,6 +208,15 @@ func (c *context) SetOutgoingHost(h string)            { c.outgoingHost = h }
 func (c *context) Metrics() filters.Metrics            { return c.metrics }
 func (c *context) Tracer() opentracing.Tracer          { return c.tracer }
 func (c *context) ParentSpan() opentracing.Span        { return c.parentSpan }
+
+func (c *context) Logger() filters.FilterContextLogger {
+	traceId := tracing.GetTraceID(opentracing.SpanFromContext(c.request.Context()))
+	if traceId == "" {
+		return log.StandardLogger()
+	} else {
+		return log.WithFields(log.Fields{"trace_id": traceId})
+	}
+}
 
 func (c *context) Serve(r *http.Response) {
 	r.Request = c.Request()
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -61,6 +61,11 @@ import (
 	"github.com/zalando/skipper/tracing/tracers/instana"
 	"github.com/zalando/skipper/tracing/tracers/jaeger"
 	"github.com/zalando/skipper/tracing/tracers/lightstep"
+
+	originstana "github.com/instana/go-sensor"
+	origlightstep "github.com/lightstep/lightstep-tracer-go"
+	origbasic "github.com/opentracing/basictracer-go"
+	origjaeger "github.com/uber/jaeger-client-go"
 )
 
 // InitTracer initializes an opentracing tracer. The first option item is the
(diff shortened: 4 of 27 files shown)


AlexanderYastrebov:
Please describe the problem it solves. Is this open tracing relevant only? Do we need to log it for each log statement or only for selected logs?
See also https://github.com/sirupsen/logrus#fields

RomanZavodskikh:
@AlexanderYastrebov this is a draft PR for me to create a test image and use this image to check how will it work (and would it work at all) to provide traceID inside all logs. I just picked up some logs but hope this change to have an effect on all possible logs.

proxy/context.go
```diff
@@ -206,6 +210,38 @@ func (c *context) Metrics() filters.Metrics            { return c.metrics }
 func (c *context) Tracer() opentracing.Tracer          { return c.tracer }
 func (c *context) ParentSpan() opentracing.Span        { return c.parentSpan }
 
+func (c *context) Logger() *log.Entry {
+	traceId := c.getTraceID()
+	return log.WithFields(log.Fields{"trace_id": traceId})
+}
+
+func (c *context) getTraceID() string {
```
AlexanderYastrebov, line 218: Lets do the following:
- return empty value from `getTraceID` if we can not obtain trace id
- return log.WithFields only if traceId is not empty, otherwise fallback to the default log

proxy/context.go
```diff
@@ -206,6 +210,38 @@ func (c *context) Metrics() filters.Metrics            { return c.metrics }
...
+
+	spanContext := span.Context()
+	if spanContext == nil {
+		return "(nil span context)"
+	}
+
+	switch spanContextType := spanContext.(type) {
+	case lightstep.SpanContext:
+		_ = spanContextType
+		return fmt.Sprintf("0x%x", spanContext.(lightstep.SpanContext).TraceID)
```
AlexanderYastrebov, line 239: you can use spanContextType.TraceID

proxy/context.go
```diff
@@ -206,6 +210,38 @@ func (c *context) Metrics() filters.Metrics            { return c.metrics }
...
+	return log.WithFields(log.Fields{"trace_id": traceId})
+}
+
+func (c *context) getTraceID() string {
+	if c == nil {
+		return "(nil context)"
+	}
+	if c.request == nil {
+		return "(nil request)"
+	}
```
AlexanderYastrebov, line 224: Is it possible for context and request to be nil actually?

proxy/context.go
AlexanderYastrebov: See https://go.dev/tour/methods/16

script/script_test.go
```diff
@@ -42,6 +44,38 @@ func (l *luaContext) ParentSpan() opentracing.Span          { return nil }
...
+	}
+
+	switch spanContextType := spanContext.(type) {
+	case lightstep.SpanContext:
+		_ = spanContextType
+		return fmt.Sprintf("0x%x", spanContext.(lightstep.SpanContext).TraceID)
+	}
+
+	return "(unknown span context type)"
+}
```
AlexanderYastrebov, line 77: It looks like we can create a helper (maybe in tracing package) to extract traceId from http.Request and reuse it in all filterContext implementations.

proxy/context.go
RomanZavodskikh: I've seen such case in `make cicheck`. Will look deeped whether it's applicable for real life.

RomanZavodskikh: Okay, context and request both cannot be nil both for `proxy.context` and `script.luaContext`. However, `request == nil` is possible for `filtertest.Context` so I will keep this check only there.

script/script_test.go
```diff
@@ -42,6 +44,38 @@ func (l *luaContext) ParentSpan() opentracing.Span          { return nil }
...
+	if spanContext == nil {
+		return "(nil span context)"
+	}
+
+	switch spanContextType := spanContext.(type) {
+	case lightstep.SpanContext:
+		return fmt.Sprintf("0x%x", spanContextType.TraceID)
+	}
+
+	return "(unknown span context type)"
```
RomanZavodskikh, line 76: Please do not return this string. Better use helper.

filters/filters.go
```diff
@@ -117,6 +117,19 @@ type FilterContext interface {
...
+
+// FilterContextLogger is the logger object which can log messages from filter
+// including context information from this filter.
+type FilterContextLogger interface {
+	Debugf(format string, args ...interface{})
+	Infof(format string, args ...interface{})
+	Warnf(format string, args ...interface{})
+	Errorf(format string, args ...interface{})
+	Println(args ...interface{})
+	Warningf(format string, args ...interface{})
```
AlexanderYastrebov, line 132: Lets remove these two. Warningf is a synonym for Warnf and we should use Debugf or Infof instead of Printf

filters/filtertest/filtertest.go
```diff
@@ -60,6 +63,15 @@ func (fc *Context) ParentSpan() opentracing.Span {
 	return opentracing.StartSpan("test_span")
 }
 
+func (fc *Context) Logger() filters.FilterContextLogger {
+	traceId := tracing.GetTraceID(fc.FRequest)
+	if traceId == "" {
+		return log.WithFields(log.Fields{})
```
AlexanderYastrebov, line 69: I think we can `return log.StandardLogger()` here to avoid unnecessary empty fields allocation

script/script_test.go
```diff
@@ -42,6 +43,15 @@ func (l *luaContext) ParentSpan() opentracing.Span          { return nil }
 func (l *luaContext) Split() (filters.FilterContext, error) { return nil, nil }
 func (l *luaContext) Loopback()                             {}
 
+func (l *luaContext) Logger() filters.FilterContextLogger {
+	traceId := tracing.GetTraceID(l.request)
+	if traceId == "" {
+		return log.WithFields(log.Fields{})
```
AlexanderYastrebov, line 49: `return log.StandardLogger()`

tracing/tracing.go
```diff
@@ -146,3 +149,24 @@ func LogKV(k, v string, ctx context.Context) {
...
+	}
+
+	spanContext := span.Context()
+	if spanContext == nil {
+		return ""
+	}
+
+	switch spanContextType := spanContext.(type) {
+	case origlightstep.SpanContext:
+		return fmt.Sprintf("0x%x", spanContextType.TraceID)
```
AlexanderYastrebov, line 168: Please add a comment why we format it as a hex with the `0x` prefix

RomanZavodskikh: Decided to remove 0x prefix to span_id to be easy copy-pastable.

filters/filters.go
```diff
@@ -117,6 +117,17 @@ type FilterContext interface {
 
 	// Performs a new route lookup and executes the matched route if any
 	Loopback()
+
+	Logger() FilterContextLogger
+}
+
+// FilterContextLogger is the logger object which can log messages from filter
+// including context information from this filter.
```
AlexanderYastrebov, line 125: `// FilterContextLogger is the logger which logs messages with additional context information`

tracing/tracing.go
```diff
@@ -146,3 +152,30 @@ func LogKV(k, v string, ctx context.Context) {
...
+		return ""
+	}
+
+	switch spanContextType := spanContext.(type) {
+	case origbasic.SpanContext:
+		return fmt.Sprintf("%x", spanContextType.TraceID)
+	case originstana.SpanContext:
+		return fmt.Sprintf("%x", spanContextType.TraceID)
+	case origjaeger.SpanContext:
+		return fmt.Sprintf("%x", spanContextType.TraceID().String())
```
AlexanderYastrebov, line 175: `spanContextType.TraceID().String()` already returns hexadecimal string so the format specifier should be `%s`

tracing/tracing.go
```diff
@@ -146,3 +152,30 @@ func LogKV(k, v string, ctx context.Context) {
 		span.LogKV(k, v)
 	}
 }
+
+// GetTraceID retrieves TraceID from HTTP request, for example to search for this trace
+// in the UI of your tracing solution and to get more context about it
+func GetTraceID(req *http.Request) string {
```
AlexanderYastrebov, line 158: Lets pass it `ot.Span` instead of request to not create a dependency on net/http

script/script_test.go
```diff
@@ -42,6 +43,15 @@ func (l *luaContext) ParentSpan() opentracing.Span          { return nil }
...
 func (l *luaContext) Loopback()                             {}
 
+func (l *luaContext) Logger() filters.FilterContextLogger {
+	traceId := tracing.GetTraceID(l.request)
+	if traceId == "" {
+		return log.StandardLogger()
+	} else {
+		return log.WithFields(log.Fields{"trace_id": traceId})
+	}
+}
```
AlexanderYastrebov, line 53: Lets get rid of luaContext altogether, see https://github.com/zalando/skipper/pull/2273

filters/filtertest/filtertest.go
```diff
@@ -60,6 +63,15 @@ func (fc *Context) ParentSpan() opentracing.Span {
 	return opentracing.StartSpan("test_span")
 }
 
+func (fc *Context) Logger() filters.FilterContextLogger {
```
AlexanderYastrebov, line 66: This is a test context. I think we should just return log.StandardLogger() for simplicity

script/script_test.go
RomanZavodskikh: Looks good to me. :-)

AlexanderYastrebov:
:+1:

PR #2270, opened 2023-04-14
fix: apiUsageMonitoring to be able to observe client timeout

szuecs:
fix: apiUsageMonitoring to be able to observe client timeout error case and record 499 status codes

fix this: https://github.com/zalando/skipper/issues/1238#issuecomment-727878056

### DIFF
--- a/filters/apiusagemonitoring/filter.go
+++ b/filters/apiusagemonitoring/filter.go
@@ -46,6 +46,11 @@ type apiUsageMonitoringStateBag struct {
 	begin time.Time
 }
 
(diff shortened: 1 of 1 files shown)


AlexanderYastrebov:
Would be nice to have a test (e.g. that it increments 502 for unreachable backend) but  I see there are only filter unit tests and no testproxy tests.

AlexanderYastrebov:
:+1:

szuecs:
Yes, exactly why I did only the change without test...

szuecs:
:+1:

PR #2271, opened 2023-04-14
feature: read timeout filter and write timeout filter

szuecs:
feature: read timeout filter and write timeout filter

### DIFF
--- a/filters/builtin/builtin.go
+++ b/filters/builtin/builtin.go
@@ -156,6 +156,8 @@ func Filters() []filters.Spec {
 		NewHeaderToQuery(),
 		NewQueryToHeader(),
 		NewBackendTimeout(),
+		NewReadTimeout(),
+		NewWriteTimeout(),
 		NewSetDynamicBackendHostFromHeader(),
 		NewSetDynamicBackendSchemeFromHeader(),
 		NewSetDynamicBackendUrlFromHeader(),
--- a/filters/builtin/timeout.go
+++ b/filters/builtin/timeout.go
@@ -6,22 +6,56 @@ import (
 	"github.com/zalando/skipper/filters"
 )
 
+type timeoutType int
+
+const (
+	backendTimeout timeoutType = iota + 1
+	readTimeout
+	writeTimeout
+)
+
 type timeout struct {
+	typ     timeoutType
 	timeout time.Duration
 }
 
 func NewBackendTimeout() filters.Spec {
-	return &timeout{}
+	return &timeout{
+		typ: backendTimeout,
+	}
 }
 
-func (*timeout) Name() string { return filters.BackendTimeoutName }
+func NewReadTimeout() filters.Spec {
+	return &timeout{
+		typ: readTimeout,
+	}
+}
 
-func (*timeout) CreateFilter(args []interface{}) (filters.Filter, error) {
+func NewWriteTimeout() filters.Spec {
+	return &timeout{
+		typ: writeTimeout,
+	}
+}
+
+func (t *timeout) Name() string {
+	switch t.typ {
+	case backendTimeout:
+		return filters.BackendTimeoutName
+	case readTimeout:
+		return filters.ReadTimeoutName
+	case writeTimeout:
+		return filters.WriteTimeoutName
+	}
+	return "unknownFilter"
+}
+
+func (t *timeout) CreateFilter(args []interface{}) (filters.Filter, error) {
 	if len(args) != 1 {
 		return nil, filters.ErrInvalidFilterParameters
 	}
 
 	var tf timeout
+	tf.typ = t.typ
 	switch v := args[0].(type) {
 	case string:
 		d, err := time.ParseDuration(v)
@@ -31,15 +65,34 @@ func (*timeout) CreateFilter(args []interface{}) (filters.Filter, error) {
 		tf.timeout = d
 	case time.Duration:
 		tf.timeout = v
+	case int:
+		tf.timeout = time.Duration(v) * time.Millisecond
 	default:
 		return nil, filters.ErrInvalidFilterParameters
 	}
 	return &tf, nil
 }
 
+// Request allows overwrite of timeout settings.
+//
+// Type backend timeout sets the timeout for the backend roundtrip.
+//
+// Type read timeout sets the timeout to read the request including the body.
+// It uses http.ResponseController to SetReadDeadline().
+//
+// Type write timeout allows to set a timeout for writing the response.
+// It uses http.ResponseController to SetWriteDeadline().
+//
+// All these timeouts are set at specific points in proxy.Proxy.
 func (t *timeout) Request(ctx filters.FilterContext) {
-	// allows overwrite
-	ctx.StateBag()[filters.BackendTimeout] = t.timeout
+	switch t.typ {
+	case backendTimeout:
+		ctx.StateBag()[filters.BackendTimeout] = t.timeout
+	case readTimeout:
+		ctx.StateBag()[filters.ReadTimeout] = t.timeout
+	case writeTimeout:
+		ctx.StateBag()[filters.WriteTimeout] = t.timeout
+	}
 }
 
-func (t *timeout) Response(filters.FilterContext) {}
+func (*timeout) Response(filters.FilterContext) {}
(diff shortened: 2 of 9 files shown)


szuecs:
@AlexanderYastrebov any idea why readTimeout would not work?
minimal Example that does not work as I would expect: https://go.dev/play/p/g-JHpG2Vwem

filters/builtin/timeout.go
```diff
@@ -3,25 +3,60 @@ package builtin
...
 	"time"
 
+	log "github.com/sirupsen/logrus"
 	"github.com/zalando/skipper/filters"
 )
 
+type timeoutType int
+
+const (
+	requestTimeout timeoutType = iota + 1
```
AlexanderYastrebov, line 13: Why not call it `backendTimeout`?

filters/filters.go
```diff
@@ -31,13 +31,17 @@ const (
...
+// FilterContext object providing state and information that is unique to a request.
 type FilterContext interface {
 	// The response writer object belonging to the incoming request. Used by
 	// filters that handle the requests themselves.
 	// Deprecated: use Response() or Serve()
 	ResponseWriter() http.ResponseWriter
 
+	// ResponseController returns the http.ResponseController
+	// attached to the current request processing.
+	ResponseController() *http.ResponseController
```
AlexanderYastrebov, line 43: We deprecated ResponseWriter() but adding ResponseController() that exposes low-level http details.
Alternative could be to implement it like we implemented `backendTimeout` - store timeouts in the statebag and apply in proxy before doing backed request.

proxy/context.go
```diff
@@ -133,13 +136,14 @@ func newContext(
...
 	c := &context{
-		responseWriter: w,
-		request:        r,
-		stateBag:       make(map[string]interface{}),
-		outgoingHost:   r.Host,
-		metrics:        &filterMetrics{impl: p.metrics},
-		proxy:          p,
-		routeLookup:    p.routing.Get(),
+		responseWriter:     w,
+		responseController: http.NewResponseController(w),
```
AlexanderYastrebov, line 140: It willnot be used for most of the requests so we can make ResponseController() return http.NewResponseController(w) on the fly

proxy/context.go
```diff
@@ -20,10 +20,13 @@ const unknownHost = "_unknownhost_"
 type flushedResponseWriter interface {
 	http.ResponseWriter
 	http.Flusher
+	SetReadDeadline(time.Time) error
+	SetWriteDeadline(time.Time) error
```
AlexanderYastrebov, line 24: Why do we need these?

filters/builtin/builtin.go
```diff
@@ -156,6 +156,8 @@ func Filters() []filters.Spec {
 		NewHeaderToQuery(),
 		NewQueryToHeader(),
 		NewBackendTimeout(),
+		NewReadTimeout(),
+		NewWriteTimeout(),
```
AlexanderYastrebov, line 160: We should document these new filters

filters/filters.go
szuecs: The problem is that these (read and write timeout) are not exposed.

filters/builtin/builtin.go
szuecs: Yes, will do whenever writeTimeout works as I expect :)

filters/filters.go
AlexanderYastrebov: Yeah, I mean create ResponseController in proxy.go around here instead of exposing it from the context:
https://github.com/zalando/skipper/blob/3c60b4ee71d0b18a36f889b26904279e0aa1e33c/proxy/proxy.go#L1124-L1127

proxy/context.go
szuecs: Do you mean `func (*context) ResponseController() *http.ResponseController`?

logging/loggingwriter.go
```diff
@@ -47,6 +48,14 @@ func (lw *LoggingWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
 	return nil, nil, fmt.Errorf("could not hijack connection")
 }
 
+func (lw *LoggingWriter) SetReadDeadline(deadline time.Time) error {
+	return http.NewResponseController(lw.writer).SetReadDeadline(deadline)
+}
+
+func (lw *LoggingWriter) SetWriteDeadline(deadline time.Time) error {
+	return http.NewResponseController(lw.writer).SetWriteDeadline(deadline)
+}
```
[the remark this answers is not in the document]
szuecs: do you have an idea about the signature of Unwrap() ?
The linked resource has nothing that would help.

proxy/context.go
szuecs: to set the corresponding deadlines

AlexanderYastrebov: I think we would not need it if we implement proper Unwrap method

logging/loggingwriter.go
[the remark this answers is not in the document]
AlexanderYastrebov: https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/net/http/responsecontroller.go;l=41-43
```go
type rwUnwrapper interface {
	Unwrap() ResponseWriter
}
```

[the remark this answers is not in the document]
szuecs: done

proxy/context.go
szuecs: done

proxy/context.go
szuecs: done

filters/filters.go
szuecs: yes, did that an reused the time.Now value from just there.

filters/builtin/timeout.go
szuecs: done

filters/builtin/builtin.go
szuecs: done

proxy/proxy.go
```diff
@@ -1127,6 +1127,12 @@ func (p *Proxy) do(ctx *context) (err error) {
 		}
 
 		backendStart := time.Now()
+		if d, ok := ctx.StateBag()[filters.ReadTimeout].(time.Duration); ok {
+			e := ctx.ResponseController().SetReadDeadline(backendStart.Add(d))
```
AlexanderYastrebov, line 1131: Why not use conventional `err` name?

filters/builtin/timeout.go
```diff
@@ -6,22 +6,56 @@ import (
...
+type timeoutType int
+
+const (
+	backendTimeout timeoutType = iota + 1
+	readTimeout
+	writeTimeout
+)
+
 type timeout struct {
+	typ     timeoutType
```
AlexanderYastrebov, line 18: I guess we could use string filter name here which would not require a new type and would simplify Name():
```
func (t *timeout) Name() string {
  return string(t.name)
}
```

filters/builtin/timeout.go
```diff
@@ -31,15 +65,34 @@ func (*timeout) CreateFilter(args []interface{}) (filters.Filter, error) {
 		tf.timeout = d
 	case time.Duration:
 		tf.timeout = v
+	case int:
+		tf.timeout = time.Duration(v) * time.Millisecond
```
AlexanderYastrebov, line 69: Lets not support untyped integer and require either duration string (eskip) or time.Duration (for lib users).
See reasoning here https://github.com/zalando/skipper/pull/1814#issuecomment-895053332

szuecs:
:+1:

proxy/proxy.go
szuecs: Because we don't want to override the response error

filters/builtin/timeout.go
szuecs: Then untyped will be `ns`, which I think is bad. There was no discussion or agreement on your cited comment...
latency() is also having int -> value in ms. I don't see the problem why we should change this.

AlexanderYastrebov: > Then untyped will be ns, which I think is bad

Why do you think so? I propose we do not support integer at all, only duration strings like "1s" or time.Duration value for lib users to eliminate any room for millisecond vs second mistake

proxy/proxy.go
AlexanderYastrebov: But it will be if-block scoped variable

szuecs: Oh, oh no it depends and `vet` will not allow this. :)

filters/builtin/timeout.go
szuecs: fixed

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2272, opened 2023-04-18
fix: tee() filter client has unexpected slow connection pool

szuecs:
fix: tee() filter client has unexpected slow connection pool

### DIFF
--- a/filters/tee/tee.go
+++ b/filters/tee/tee.go
@@ -10,6 +10,7 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/net"
 )
 
 const (
@@ -46,7 +47,7 @@ const (
 )
 
 type tee struct {
-	client            *http.Client
+	client            *net.Client
 	typ               teeType
 	host              string
 	scheme            string
@@ -171,6 +172,11 @@ func (r *tee) Request(fc filters.FilterContext) {
 	}()
 }
 
+func (r *tee) Close() error {
+	r.client.Close()
+	return nil
+}
+
 // copies requests changes URL and Host in request.
 // If 2nd and 3rd params are given path is also modified by applying regexp
 // Returns the cloned request and the tee body to be used on the main request.
@@ -218,15 +224,12 @@ func cloneRequest(t *tee, req *http.Request) (*http.Request, io.ReadCloser, erro
 // If only one parameter is given shadow backend is used as it is specified
 // If second and third parameters are also set, then path is modified
 func (spec *teeSpec) CreateFilter(config []interface{}) (filters.Filter, error) {
-	client := &http.Client{Timeout: spec.options.Timeout}
-
+	var checkRedirect func(req *http.Request, via []*http.Request) error
 	if spec.options.NoFollow {
-		client.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+		checkRedirect = func(req *http.Request, via []*http.Request) error {
 			return http.ErrUseLastResponse
 		}
 	}
-
-	tee := tee{client: client}
 
 	if len(config) == 0 {
 		return nil, filters.ErrInvalidFilterParameters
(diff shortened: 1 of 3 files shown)


filters/tee/tee.go
```diff
@@ -171,6 +172,11 @@ func (r *tee) Request(fc filters.FilterContext) {
 	}()
 }
 
+func (r *tee) Close() error {
+	r.client.Close()
+	return nil
+}
```
AlexanderYastrebov, line 178: Note that this is not called so we'll leak http clients. 
For the auth filters we mitigate this by keeping a global client cache of clients per url.

AlexanderYastrebov:
Could you provide a bit more details what and how this change will be fixing?

filters/tee/tee.go
szuecs: True this requires a fix before merge

szuecs: fixed

szuecs:
@AlexanderYastrebov is the PR description enough or do I need to elaborate?

szuecs:
TODO as discussed: change the PR not to change `builtin` analog to lua config

filters/tee/tee.go
```diff
@@ -236,43 +295,71 @@ func (spec *teeSpec) CreateFilter(config []interface{}) (filters.Filter, error)
...
 		return nil, err
 	}
 
-	if len(config) == 1 {
-		tee.typ = asBackend
-		return &tee, nil
+	var client *net.Client
+	teeClients.mu.Lock()
+	if client, ok = teeClients.store[u.Host]; !ok {
+		client = net.NewClient(net.Options{
```
AlexanderYastrebov, line 306: Need to put client into teeClients.store

AlexanderYastrebov: Also would need to cleanup clients in main_test.go similar to what we do in filters/auth package

filters/tee/tee.go
```diff
@@ -236,43 +295,71 @@ func (spec *teeSpec) CreateFilter(config []interface{}) (filters.Filter, error)
...
-	if len(config) == 3 {
+	switch len(config) {
+	case 1:
+		tee.typ = asBackend
+		return &tee, nil
+	case 3:
+		// modpath
 		expr, ok := config[1].(string)
 		if !ok {
+			client.Close()
```
AlexanderYastrebov, line 335: I think we should not close it after we put it into global lookup.
Maybe it is even fine to leave it there even if we return error

szuecs: yeah that was also a feeling that I have. I think we should follow that as we are 2 having that :)

filters/tee/tee.go
szuecs: We might also want to make sure that if there is a closed client in the pool we do not use it.
The problem is that if we call `tee.Close()` we close the client but do not remove it from the map.
wdyt?

szuecs:
:+1:

szuecs:
:+1:

filters/tee/tee.go
```diff
@@ -263,16 +364,21 @@ func (spec *teeSpec) CreateFilter(config []interface{}) (filters.Filter, error)
...
 		tee.rx = rx
 		tee.replacement = replacement
 
 		return &tee, nil
+	default:
+		teeClients.mu.Lock()
+		delete(teeClients.store, u.Host)
+		teeClients.mu.Unlock()
+		client.Close()
+		return nil, filters.ErrInvalidFilterParameters
```
AlexanderYastrebov, line 380: I think this is not correct as it may close a client used by another filter instance

szuecs: fixed

filters/tee/tee.go
```diff
@@ -21,7 +25,12 @@ const (
 	NoFollowName = filters.TeenfName
 )
 
-const defaultTeeTimeout = time.Second
+const (
+	defaultTeeTimeout          = time.Second
+	defaultMaxIdleConns        = 100
+	defaultMaxIdleConnsPerHost = 50
```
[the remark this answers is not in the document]
szuecs: fixed

filters/tee/tee.go
```diff
@@ -263,16 +341,17 @@ func (spec *teeSpec) CreateFilter(config []interface{}) (filters.Filter, error)
 		rx, err := regexp.Compile(expr)
 
 		if err != nil {
+			client.Close()
```
AlexanderYastrebov, line 344: This is not needed as we do not close clients

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2274, opened 2023-04-20
Adds request rawquery template placeholder

chahn:
Adds new template placeholder `${request.rawquery}` providing the raw query string (e.g. `q=foo&r=bar`) of the request.


docs/reference/filters.md
```diff
@@ -25,6 +25,7 @@ Template placeholder is replaced by the value that is looked up in the following
 * request method (`${request.method}`)
 * request host (`${request.host}`)
 * request url path (`${request.path}`)
+* request url rawquery (`${request.rawquery}`)
```
AlexanderYastrebov, line 28: I am wondering if we should name it just `${request.query}`

eskip/template_test.go
```diff
@@ -253,6 +253,16 @@ func TestTemplateApplyContext(t *testing.T) {
 		"hello ",
 		false, // path is empty
 
+	}, {
+		"request rawquery",
+		"hello ${request.rawquery}",
+		&filtertest.Context{
+			FRequest: &http.Request{
+				URL: parseUrl("http://example.com/path?q-Q=foo&P_p=bar"),
```
AlexanderYastrebov, line 261: Please add a test case for:
- url without query `http://example.com/foo`
- url with a query without value `http://example.com/foo?bar`

AlexanderYastrebov: Maybe also a case for URL-encoded value

AlexanderYastrebov:
Please run `git commit --amend -s` to fix DCO

docs/reference/filters.md
chahn: I even thought about it but `${request.query.q}` would return a value only and `${request.query}`would return key(s) and value(s) of the query. For me this seemed to be a little inconsistency or even confusing.
What do you think? I am also happy with `${request.query}` if you would prefer?

eskip/template_test.go
chahn: I extended the test case `request rawquery` accordingly and added an additional test case `request rawquery is empty`.
Please have a look.

docs/reference/filters.md
AlexanderYastrebov: Ok, lets name it raw query - this would be somewhat consistent with lua properties https://github.com/zalando/skipper/blob/master/docs/reference/scripts.md#other-request-fields

But lets name camelcase `rawQuery` to match existing `sourceFromLast` and `clientIP`

AlexanderYastrebov: Lets also describe that it is `encoded request URL query values, without '?'`

chahn: Thanks for your suggestions @AlexanderYastrebov.
I changed it to `${request.rawQuery}` and tried to describe it as precisely as possible in the documentation.

AlexanderYastrebov: You need to fix tests as well

chahn: Sorry @AlexanderYastrebov you're right.
I currently fixed the tests.

chahn: And I fixed the test case `request_rawquery_is_empty` (expected `true` by mistake, instead of `false`).

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2286, opened 2023-05-03
fix: use pinned dependencies in gh actions

szuecs:
fix: use best practice from openssf and use pinned dependencies in gh actions

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -37,7 +37,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5
       with:
         # We must fetch at least the immediate parents so that if this is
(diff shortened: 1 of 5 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -37,15 +37,15 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5
```
AlexanderYastrebov, line 40: How did you obtain these hashes?

AlexanderYastrebov:
:+1:

.github/workflows/codeql-analysis.yml
szuecs: https://github.com/actions/checkout/releases/tag/v2.7.0 and then click the small hash to get https://github.com/actions/checkout/commit/ee0669bd1cc54295c223e0bb666b733df41de1c5 , then copy the long sha

szuecs:
:+1:

Issue #2268, opened 2023-04-14
RouteGroup traffic switching should take care about internal routing details


szuecs:
**Is your feature request related to a problem? Please describe.**

As of today, you can specify `weight` in the [backendReference object](https://opensource.zalando.com/skipper/kubernetes/routegroup-crd/#backend-reference), but this can lead to wrong routing behavior, because of [route matching](https://opensource.zalando.com/skipper/tutorials/basics/#route-matching).
Example:

```
apiVersion: zalando.org/v1
kind: RouteGroup
metadata:
spec:
  defaultBackends:
  - backendName: v1
    weight: 90
  - backendName: v2
    weight: 10
  routes:
  - pathSubtree: /
  - pathSubtree: /
    predicates:
    - Cookie("foo", "^")
``` 
This routegroup will be translated to eskip in skipper:
```
v1_r1: PathSubtree("/") && Traffic(0.9) -> ... -> <backend1>
v2_r1: PathSubtree("/") -> ... -> <backend2>
v1_r2: PathSubtree("/") && Cookie("foo", "^")  && Traffic(0.9) -> ... -> <backend1>
v2_r2: PathSubtree("/") && Cookie("foo", "^")  -> ... -> <backend2>
```
or
```
v1_r1: PathSubtree("/") -> ..
v2_r1: PathSubtree("/") && Traffic(0.1) -> ..
v1_r2: PathSubtree("/") && Cookie("foo", "^")  -> ..
v2_r2: PathSubtree("/") && Cookie("foo", "^")  && Traffic(0.1) -> ..
```
In the first case v1_r1 and v2_r2 have the same weights, such that traffic split is not 90/10 but closer to 50/50 (or even 45/55)
In the second case v2_r1 and v1_r2 have the same weights, such that traffic split is not persisted based on cookie at 10% chance. 

**Describe the solution you would like**

Skipper should internally understand the route weights and balance the weights such that the intended behavior would happen.
To resolve the routing issue, the routes for the example above should be: 
```
v1_r1: PathSubtree("/") && Traffic(0.9) -> ..
v2_r1: PathSubtree("/") -> ..
v1_r2: PathSubtree("/") && Cookie("foo", "^")  && True() && Traffic(0.9) -> ..
v2_r2: PathSubtree("/") && Cookie("foo", "^") && True() -> ..
```
or
```
v1_r1: PathSubtree("/") -> ..
v2_r1: PathSubtree("/") && Traffic(0.1) -> ..
[...]

PR #2287, opened 2023-05-03
predicates: add TrafficSegment predicate

AlexanderYastrebov:
This change adds routing context to pass information between multiple predicates and uses it to implement new TrafficSegment predicate.

### DIFF
--- a/docs/reference/predicates.md
+++ b/docs/reference/predicates.md
@@ -716,6 +716,32 @@ catalog:
     "https://catalog";
 ```
 
+## TrafficSegment
+
+TrafficSegment predicate requires two number arguments $min$ and $max$
+from an interval $[0, 1]$ (from zero included to one included) and $min <= max$.
+
+Let $r$ be one-per-request uniform random number value from $[0, 1)$.
+TrafficSegment matches if $r$ belongs to an interval from $[min, max)$.
+Upper interval boundary $max$ is excluded to simplify definition of
+adjacent intervals - the upper boundary of the first interval
+then equals lower boundary of the next and so on, e.g. $[0, 0.25)$ and $[0.25, 1)$.
+
+This predicate has weight of -1 and therefore does not affect route weight.
+
+Parameters:
+
+* min (decimal) from an interval [0, 1]
+* max (decimal) from an interval [0, 1], min <= max
+
+Example of routes splitting traffic in 50%+30%+20% proportion:
+
+```
+r50: Path("/test") && TrafficSegment(0.0, 0.5) -> <shunt>;
+r30: Path("/test") && TrafficSegment(0.5, 0.8) -> <shunt>;
+r20: Path("/test") && TrafficSegment(0.8, 1.0) -> <shunt>;
+```
+
 ## ContentLengthBetween
 
 The ContentLengthBetween predicate matches a route when a request content length header value is between min and max provided values.
--- a/predicates/traffic/segment.go
+++ b/predicates/traffic/segment.go
@@ -1,0 +1,77 @@
+package traffic
+
+import (
+	"math/rand"
+	"net/http"
+
+	"github.com/zalando/skipper/predicates"
+	"github.com/zalando/skipper/routing"
+)
+
+type (
+	segmentSpec      struct{}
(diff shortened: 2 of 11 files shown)


docs/reference/predicates.md
```diff
@@ -716,6 +716,32 @@ catalog:
...
+## TrafficSegment
+
+TrafficSegment predicate requires two number arguments _min_ and _max_
+from an interval [0, 1] (from zero included to one included) and _min_ <= _max_.
+
+Let _r_ be one-per-request uniform random number value from [0, 1).
+TrafficSegment matches if _r_ belongs to an interval from [_min_, _max_).
+Upper interval boundary _max_ is excluded to simplify definition of
+adjacent intervals - the upper boundary of the first interval
+then equals lower boundary of the next and so on, e.g. [0, 0.25) and [0.25, 1).
```
szuecs, line 728: We can use math quotes similar to admissionControl filter doc: `$[1, 100]$` renders $[1, 100]$.
I think we can use it also for $[min, max)$ for example

docs/reference/predicates.md
```diff
@@ -716,6 +716,32 @@ catalog:
...
+
+This predicate has weight of -1 and therefore does not affect route weight.
+
+Parameters:
+
+* min (decimal) from an interval [0, 1]
+* max (decimal) from an interval [0, 1], min <= max
+
+Example of routes splitting traffic in 50%+30%+20% proportion:
```
szuecs, line 737: Maybe also tell about the advantage compared to `Traffic`, because the new one does not need to use `True` hacks in order to have the routes ordered correctly in case we have multi color traffic switching.

AlexanderYastrebov: `Traffic` and `TrafficSegment` are low-level plumbing. If we accept `TrafficSegment` next step I plan is to re-implement Ingress and RouteGroup backend weight using it and there we can describe why we change it and its benefits.

routing/context.go
```diff
@@ -0,0 +1,36 @@
...
+	m, _ := ctx.Value(routingContextKey).(*sync.Map)
+
+	// https://github.com/golang/go/issues/44159#issuecomment-780774977
+	val, ok := m.Load(key)
+	if !ok {
+		val = defaultValue()
+		val, _ = m.LoadOrStore(key, val)
+	}
+	return val.(V)
+}
```
AlexanderYastrebov, line 36: In principle it should be possible to implement multi-threaded route lookup (because predicates do not mutate request) so I've implemented routing context having this in mind.

szuecs:
:+1:

szuecs:
:+1:

docs/reference/predicates.md
AlexanderYastrebov: Fixed

szuecs:
:+1:

predicates/traffic/segment.go
```diff
@@ -0,0 +1,77 @@
...
+//	r30: Path("/test") && TrafficSegment(0.5, 0.8) -> <shunt>;
+//	r20: Path("/test") && TrafficSegment(0.8, 1.0) -> <shunt>;
+func (*segmentSpec) Create(args []any) (routing.Predicate, error) {
+	if len(args) != 2 {
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+
+	p, ok := &segmentPredicate{}, false
+
+	if p.min, ok = args[0].(float64); !ok || p.min < 0 || p.min > 1 {
```
RomanZavodskikh, line 52: How can we stop user for providing invalid input (segments that do not sum up to 1)?
For example like this:
```
r55: Path("/test") && TrafficSegment(0.0, 0.55) -> <shunt>;
r40: Path("/test") && TrafficSegment(0.55, 0.95) -> <shunt>;
```

AlexanderYastrebov: We do not. This predicate is a ["plumbing"](https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain) and meant to be a building block of higher level algorithm that must ensure there no gaps or overlaps between segments.

PR #2291, opened 2023-05-05
fix: openssf card scan told to remove top level permissions

szuecs:
fix: openssf card scan told to remove top level permissions, as every job has a set of permissions we should be able to drop top-level permissions to 0


AlexanderYastrebov:
:+1:

PR #2292, opened 2023-05-05
dataclients/kubernetes: use fixtures to test Ingress backend weights

AlexanderYastrebov:
This change converts tests of computeBackendWeightsV1 into equivalent Ingress fixtures which should simplify refactoring of the weight calculation algorithm.


AlexanderYastrebov:
:+1:

Issue #2111, opened 2022-10-24
Support Etag for remote file data client


szuecs:
Follow up after supporting ETag on routesrv https://github.com/zalando/skipper/issues/2107 is to make the dataclient aware of the ETag.

nikhil88877:
Heyy, I got caught up with other work and could not follow up on this. I'll create PR this week 🙏

PR #2294, opened 2023-05-08
Optimize routeSRV network consumption.

MustafaSaber:

### DIFF
--- a/routesrv/eskipbytes.go
+++ b/routesrv/eskipbytes.go
@@ -72,6 +72,7 @@ func (e *eskipBytes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if initialized {
 		w.Header().Add("Etag", etag)
 		w.Header().Add("Content-Type", "text/plain; charset=utf-8")
+		w.Header().Add("Content-Encoding", "gzip")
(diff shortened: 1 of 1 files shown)


routesrv/eskipbytes.go
```diff
@@ -72,6 +72,7 @@ func (e *eskipBytes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if initialized {
 		w.Header().Add("Etag", etag)
 		w.Header().Add("Content-Type", "text/plain; charset=utf-8")
+		w.Header().Add("Content-Encoding", "gzip")
```
AlexanderYastrebov, line 75: It is not enough to add a header to serve gzip - the response body should be compressed as well, see https://stackoverflow.com/questions/14073393/serving-gzipped-content-for-go

In general the compression trades network consumption for cpu usage on both server and client sides. We should justify this change somehow, e.g. by showing how much it will compress.

szuecs: Maybe add the compress filter in the route chain to serve the content to re-use skipper's feature set. It's of course a little more work.

MustafaSaber:
Closing as we decided to complete Etag implementation (https://github.com/zalando/skipper/issues/2111) first and see how it will perform.

PR #2300, opened 2023-05-08
Duplicate permissions block

MustafaSaber:
Follow up https://github.com/zalando/skipper/pull/2291

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,5 +1,4 @@
 name: docs-ci
-permissions: {}
 
 on:
   workflow_dispatch:


.github/workflows/docs.yaml
```diff
@@ -1,5 +1,4 @@
 name: docs-ci
-permissions: {}
```
AlexanderYastrebov, line 2: So this will end up having
```yaml
permissions:
  contents: write
```
from below.

Do we want it like that or do we want empty permissions here?

AlexanderYastrebov:
:+1:

.github/workflows/docs.yaml
MustafaSaber: We want `write` permissions not the empty one.

MustafaSaber:
👍

PR #2304, opened 2023-05-09
Support `Etag` from DataClient prespective

MustafaSaber:
Fixes https://github.com/zalando/skipper/issues/2111

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -115,23 +115,24 @@ type Config struct {
 	SuppressRouteUpdateLogs             bool      `yaml:"suppress-route-update-logs"`
 
 	// route sources:
-	EtcdUrls           string               `yaml:"etcd-urls"`
-	EtcdPrefix         string               `yaml:"etcd-prefix"`
-	EtcdTimeout        time.Duration        `yaml:"etcd-timeout"`
-	EtcdInsecure       bool                 `yaml:"etcd-insecure"`
-	EtcdOAuthToken     string               `yaml:"etcd-oauth-token"`
-	EtcdUsername       string               `yaml:"etcd-username"`
-	EtcdPassword       string               `yaml:"etcd-password"`
-	RoutesFile         string               `yaml:"routes-file"`
-	RoutesURLs         *listFlag            `yaml:"routes-urls"`
-	InlineRoutes       string               `yaml:"inline-routes"`
-	AppendFilters      *defaultFiltersFlags `yaml:"default-filters-append"`
-	PrependFilters     *defaultFiltersFlags `yaml:"default-filters-prepend"`
-	DisabledFilters    *listFlag            `yaml:"disabled-filters"`
-	EditRoute          routeChangerConfig   `yaml:"edit-route"`
-	CloneRoute         routeChangerConfig   `yaml:"clone-route"`
-	SourcePollTimeout  int64                `yaml:"source-poll-timeout"`
-	WaitFirstRouteLoad bool                 `yaml:"wait-first-route-load"`
+	EtcdUrls            string               `yaml:"etcd-urls"`
+	EtcdPrefix          string               `yaml:"etcd-prefix"`
+	EtcdTimeout         time.Duration        `yaml:"etcd-timeout"`
+	EtcdInsecure        bool                 `yaml:"etcd-insecure"`
+	EtcdOAuthToken      string               `yaml:"etcd-oauth-token"`
+	EtcdUsername        string               `yaml:"etcd-username"`
+	EtcdPassword        string               `yaml:"etcd-password"`
+	RoutesFile          string               `yaml:"routes-file"`
+	RoutesURLs          *listFlag            `yaml:"routes-urls"`
+	EnableRoutesCaching bool                 `yaml:"enable-routes-cache"`
+	InlineRoutes        string               `yaml:"inline-routes"`
+	AppendFilters       *defaultFiltersFlags `yaml:"default-filters-append"`
+	PrependFilters      *defaultFiltersFlags `yaml:"default-filters-prepend"`
+	DisabledFilters     *listFlag            `yaml:"disabled-filters"`
+	EditRoute           routeChangerConfig   `yaml:"edit-route"`
+	CloneRoute          routeChangerConfig   `yaml:"clone-route"`
+	SourcePollTimeout   int64                `yaml:"source-poll-timeout"`
+	WaitFirstRouteLoad  bool                 `yaml:"wait-first-route-load"`
 
 	// Forwarded headers
 	ForwardedHeadersList            *listFlag            `yaml:"forwarded-headers"`
@@ -400,6 +401,7 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.EtcdPassword, "etcd-password", "", "optional password for basic authentication with etcd")
 	flag.StringVar(&cfg.RoutesFile, "routes-file", "", "file containing route definitions")
 	flag.Var(cfg.RoutesURLs, "routes-urls", "comma separated URLs to route definitions in eskip format")
+	flag.BoolVar(&cfg.EnableRoutesCaching, "enable-routes-cache", false, "use lasted fetched routes if not modified.")
 	flag.StringVar(&cfg.InlineRoutes, "inline-routes", "", "inline routes in eskip format")
 	flag.Int64Var(&cfg.SourcePollTimeout, "source-poll-timeout", int64(3000), "polling timeout of the routing data sources, in milliseconds")
 	flag.Var(cfg.AppendFilters, "default-filters-append", "set of default filters to apply to append to all filters of all routes")
@@ -788,16 +790,17 @@ func (c *Config) ToOptions() skipper.Options {
 		SuppressRouteUpdateLogs:             c.SuppressRouteUpdateLogs,
 
 		// route sources:
-		EtcdUrls:        eus,
-		EtcdPrefix:      c.EtcdPrefix,
-		EtcdWaitTimeout: c.EtcdTimeout,
-		EtcdInsecure:    c.EtcdInsecure,
-		EtcdOAuthToken:  c.EtcdOAuthToken,
-		EtcdUsername:    c.EtcdUsername,
-		EtcdPassword:    c.EtcdPassword,
-		WatchRoutesFile: c.RoutesFile,
-		RoutesURLs:      c.RoutesURLs.values,
-		InlineRoutes:    c.InlineRoutes,
+		EtcdUrls:            eus,
+		EtcdPrefix:          c.EtcdPrefix,
+		EtcdWaitTimeout:     c.EtcdTimeout,
+		EtcdInsecure:        c.EtcdInsecure,
+		EtcdOAuthToken:      c.EtcdOAuthToken,
+		EtcdUsername:        c.EtcdUsername,
+		EtcdPassword:        c.EtcdPassword,
+		WatchRoutesFile:     c.RoutesFile,
+		RoutesURLs:          c.RoutesURLs.values,
+		EnableRoutesCaching: c.EnableRoutesCaching,
+		InlineRoutes:        c.InlineRoutes,
 		DefaultFilters: &eskip.DefaultFilters{
 			Prepend: c.PrependFilters.filters,
 			Append:  c.AppendFilters.filters,
--- a/config/config.go
+++ b/config/config.go
@@ -115,24 +115,23 @@ type Config struct {
 	SuppressRouteUpdateLogs             bool      `yaml:"suppress-route-update-logs"`
 
 	// route sources:
-	EtcdUrls            string               `yaml:"etcd-urls"`
-	EtcdPrefix          string               `yaml:"etcd-prefix"`
-	EtcdTimeout         time.Duration        `yaml:"etcd-timeout"`
-	EtcdInsecure        bool                 `yaml:"etcd-insecure"`
-	EtcdOAuthToken      string               `yaml:"etcd-oauth-token"`
-	EtcdUsername        string               `yaml:"etcd-username"`
-	EtcdPassword        string               `yaml:"etcd-password"`
-	RoutesFile          string               `yaml:"routes-file"`
-	RoutesURLs          *listFlag            `yaml:"routes-urls"`
-	EnableRoutesCaching bool                 `yaml:"enable-routes-cache"`
-	InlineRoutes        string               `yaml:"inline-routes"`
-	AppendFilters       *defaultFiltersFlags `yaml:"default-filters-append"`
-	PrependFilters      *defaultFiltersFlags `yaml:"default-filters-prepend"`
-	DisabledFilters     *listFlag            `yaml:"disabled-filters"`
-	EditRoute           routeChangerConfig   `yaml:"edit-route"`
-	CloneRoute          routeChangerConfig   `yaml:"clone-route"`
-	SourcePollTimeout   int64                `yaml:"source-poll-timeout"`
-	WaitFirstRouteLoad  bool                 `yaml:"wait-first-route-load"`
+	EtcdUrls           string               `yaml:"etcd-urls"`
+	EtcdPrefix         string               `yaml:"etcd-prefix"`
+	EtcdTimeout        time.Duration        `yaml:"etcd-timeout"`
+	EtcdInsecure       bool                 `yaml:"etcd-insecure"`
+	EtcdOAuthToken     string               `yaml:"etcd-oauth-token"`
+	EtcdUsername       string               `yaml:"etcd-username"`
+	EtcdPassword       string               `yaml:"etcd-password"`
+	RoutesFile         string               `yaml:"routes-file"`
+	RoutesURLs         *listFlag            `yaml:"routes-urls"`
+	InlineRoutes       string               `yaml:"inline-routes"`
+	AppendFilters      *defaultFiltersFlags `yaml:"default-filters-append"`
+	PrependFilters     *defaultFiltersFlags `yaml:"default-filters-prepend"`
+	DisabledFilters    *listFlag            `yaml:"disabled-filters"`
+	EditRoute          routeChangerConfig   `yaml:"edit-route"`
+	CloneRoute         routeChangerConfig   `yaml:"clone-route"`
+	SourcePollTimeout  int64                `yaml:"source-poll-timeout"`
+	WaitFirstRouteLoad bool                 `yaml:"wait-first-route-load"`
 
 	// Forwarded headers
 	ForwardedHeadersList            *listFlag            `yaml:"forwarded-headers"`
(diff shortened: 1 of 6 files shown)


routesrv/eskipbytes.go
```diff
@@ -61,6 +61,13 @@ func (e *eskipBytes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	ifNoneMatch := r.Header.Get("If-None-Match")
+	if ifNoneMatch != "" && ifNoneMatch == e.etag {
```
szuecs, line 65: if you don't need a var it's temporary, so you can scope it to the block:
```
if ifNoneMatch := r.Header.Get("If-None-Match"); ifNoneMatch != "" && ifNoneMatch == e.etag {
...
}
```

skipper.go
```diff
@@ -287,6 +287,9 @@ type Options struct {
 	// RouteURLs are URLs pointing to route definitions, in eskip format, with change watching enabled.
 	RoutesURLs []string
 
+	// EnableRoutesCaching allow skipper to use latest pulled routes if not modified.
+	EnableRoutesCaching bool
```
szuecs, line 291: I think we do not need this option. Why would you want to not use it?

eskipfile/remote_test.go
```diff
@@ -82,18 +84,54 @@ func TestLoadAll(t *testing.T) {
...
 		t.Run(test.title, func(t *testing.T) {
-			options := &RemoteWatchOptions{RemoteFile: s.URL, Threshold: 10, Verbose: true, FailOnStartup: true}
+			options := &RemoteWatchOptions{RemoteFile: s.URL, Threshold: 10, Verbose: true, FailOnStartup: true, EnableRoutesCaching: test.cached}
 			client, err := RemoteWatch(options)
 			if err == nil {
 				defer client.(*remoteEskipFile).Close()
 			}
 
+			if test.cached {
+				f, err := os.OpenFile(client.(*remoteEskipFile).localPath, os.O_APPEND|os.O_WRONLY, os.ModeAppend)
```
szuecs, line 118: The one that open(2) has also to do close(2).
So after error check `defer f.Close()`

And for readability, please use `fd` instead of `f` for file descriptors.

skipper.go
szuecs: the workaround is revert to the version before and fix the bug

eskipfile/remote.go
```diff
@@ -172,13 +181,24 @@ func (client *remoteEskipFile) DownloadRemoteFile() error {
...
+	client.isCached = false
+
 	resp, err := client.http.Get(client.remotePath)
 	if err != nil {
 		return nil, err
 	}
+
 	if resp.StatusCode != 200 {
+		if resp.StatusCode == 304 {
+			client.isCached = true
```
szuecs, line 194: another field another data race ;)
I think it's better to return a pre-defined error like:
```
ErrContentNotChanged = errors.New("content in cache did not change")
```
and then return this and check this where you need.

net/httpclient.go
```diff
@@ -75,6 +76,10 @@ func NewClient(o Options) *Client {
 	return c
 }
 
+func (c *Client) SetEtag(etag string) {
```
szuecs, line 79: This assumes we only use one URL per client, but you could also have clients that query API x and API y that both support ETag. If you introduce a setter, maybe also a getter and pass hostname?

eskipfile/remote.go
MustafaSaber: you mean error instead of field right?

config/config.go
```diff
@@ -400,6 +401,7 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.EtcdPassword, "etcd-password", "", "optional password for basic authentication with etcd")
 	flag.StringVar(&cfg.RoutesFile, "routes-file", "", "file containing route definitions")
 	flag.Var(cfg.RoutesURLs, "routes-urls", "comma separated URLs to route definitions in eskip format")
+	flag.BoolVar(&cfg.EnableRoutesCaching, "enable-routes-cache", false, "use lasted fetched routes if not modified.")
```
szuecs, line 404: We should not have this. 😊

eskipfile/remote.go
```diff
@@ -172,13 +182,22 @@ func (client *remoteEskipFile) DownloadRemoteFile() error {
...
 	resp, err := client.http.Get(client.remotePath)
 	if err != nil {
 		return nil, err
 	}
+
 	if resp.StatusCode != 200 {
+		if resp.StatusCode == 304 {
+			return nil, ErrContentNotChanged
+		}
 		return nil, errors.New("download file failed")
```
szuecs, line 195: Maybe add status code in this error

net/httpclient.go
szuecs: My thinking was wrong here, because we have one remote for one dataclient.

eskipfile/remote.go
```diff
@@ -172,13 +179,29 @@ func (client *remoteEskipFile) DownloadRemoteFile() error {
 }
 
 func (client *remoteEskipFile) getRemoteData() (io.ReadCloser, error) {
-	resp, err := client.http.Get(client.remotePath)
+	req, err := http.NewRequest("GET", client.remotePath, nil)
+
+	if err != nil {
+		return nil, errors.New("failed to create request")
```
AlexanderYastrebov, line 185: Either wrap the error
```go
return nil, fmt.Errorf("failed to create request: %w", err)
```
or just `return nil, err`

eskipfile/remote.go
```diff
@@ -172,13 +179,29 @@ func (client *remoteEskipFile) DownloadRemoteFile() error {
...
+
+	resp, err := client.http.Do(req)
 	if err != nil {
 		return nil, err
 	}
+
 	if resp.StatusCode != 200 {
+		if resp.StatusCode == 304 {
+			return nil, ErrContentNotChanged
+		}
```
AlexanderYastrebov, line 200: Lets move this outside of and before `if resp.StatusCode != 200`

eskipfile/remote.go
```diff
@@ -154,6 +158,9 @@ func isFileRemote(remotePath string) bool {
 func (client *remoteEskipFile) DownloadRemoteFile() error {
 	data, err := client.getRemoteData()
 	if err != nil {
+		if errors.Is(err, ErrContentNotChanged) {
```
AlexanderYastrebov, line 161: You may also return `changed` boolean flag instead of using sentinel error value

MustafaSaber: Please check the first commit. It started as having a boolean value in `remoteEskipFile` and setting it inside the condition, the Error approach was better. However, if we want to return boolean flag we will need to change function schema

routesrv/eskipbytes.go
```diff
@@ -61,10 +61,19 @@ func (e *eskipBytes) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	e.mu.RLock()
+	etag := e.etag
+	e.mu.RUnlock()
+
+	if ifNoneMatch := r.Header.Get("If-None-Match"); ifNoneMatch != "" && ifNoneMatch == etag {
+		// Client has a matching ETag, return a 304 Not Modified response
+		w.WriteHeader(http.StatusNotModified)
+		return
+	}
+
```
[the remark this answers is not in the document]
MustafaSaber: Great observation, thanks!

eskipfile/remote.go
```diff
@@ -172,13 +182,31 @@ func (client *remoteEskipFile) DownloadRemoteFile() error {
...
+		req.Header.Set("If-None-Match", client.etag)
+	}
+
+	resp, err := client.http.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode == 304 {
+		client.onCachedSometimes.Do(func() { log.Info("remote routes are cached") })
```
AlexanderYastrebov, line 201: Why do we want to log this?

MustafaSaber: I thought we may need to log every now and then that we use cached version.

MustafaSaber: Would you prefer if I remove it?

eskipfile/remote.go
```diff
@@ -15,6 +16,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+var ErrContentNotChanged = errors.New("content in cache did not change, 304 reponse status code")
```
AlexanderYastrebov, line 19: Let make it unexported as it is not used outside of this package

eskipfile/remote.go
```diff
@@ -172,13 +179,30 @@ func (client *remoteEskipFile) DownloadRemoteFile() error {
...
 
 func (client *remoteEskipFile) getRemoteData() (io.ReadCloser, error) {
-	resp, err := client.http.Get(client.remotePath)
+	req, err := http.NewRequest("GET", client.remotePath, nil)
+
 	if err != nil {
 		return nil, err
 	}
+
+	if len(client.etag) != 0 {
```
AlexanderYastrebov, line 188: nit: `if client.etag != ""`

eskipfile/remote_test.go
```diff
@@ -60,6 +61,7 @@ func TestLoadAll(t *testing.T) {
 		routeStatusCode int
 		expected        []*eskip.Route
 		fail            bool
+		cached          bool
```
AlexanderYastrebov, line 64: How do we use this field?

MustafaSaber: it's a leftover

eskipfile/remote.go
MustafaSaber: shouldn't the len be more generic?

eskipfile/remote_test.go
```diff
@@ -94,6 +114,27 @@ func TestLoadAll(t *testing.T) {
...
+				if err != nil {
+					t.Error(err)
+					return
+				}
+				err = fd.Sync()
+				if err != nil {
+					t.Error(err)
+					return
+				}
+			}
```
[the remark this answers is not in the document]
AlexanderYastrebov: Lets have a separate test fro the caching behavior where client makes several requests and test server checks etag and returns 200 or 304 accordingly.

eskipfile/remote_test.go
```diff
@@ -195,3 +196,120 @@ func createTestServer(c string, sc int) *httptest.Server {
...
+		BackendType: eskip.NetworkBackend,
+		Shunt:       false,
+		Backend:     "https://example.com/",
+	}}
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("ETag", "test-etag")
+		if noneMatch := r.Header.Get("If-None-Match"); noneMatch == "test-etag" {
+			print("\n============= I'm matched here =============\n")
+			// w.Header().Set("ETag", "test-etag")
```
szuecs, line 219: print and commented line to cleanup

eskipfile/remote.go
```diff
@@ -172,13 +179,30 @@ func (client *remoteEskipFile) DownloadRemoteFile() error {
...
+	if client.etag != "" {
+		req.Header.Set("If-None-Match", client.etag)
+	}
+
+	resp, err := client.http.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode == 304 {
```
AlexanderYastrebov, line 197: We need to close the resp.Body here because we do not return it

AlexanderYastrebov: Maybe we should also check that we actually set etag like
```go
if client.etag != "" && resp.StatusCode == 304 {
...
}
```

eskipfile/remote.go
AlexanderYastrebov: Lets add status code to the error via fmt.Errorf(...)

eskipfile/remote_test.go
AlexanderYastrebov: You may use t.Logf() instaed

eskipfile/remote_test.go
```diff
@@ -195,3 +196,120 @@ func createTestServer(c string, sc int) *httptest.Server {
...
+		Filters: []*eskip.Filter{{
+			Name: "setPath",
+			Args: []interface{}{
+				"/homepage/",
+			},
+		}},
+		BackendType: eskip.NetworkBackend,
+		Shunt:       false,
+		Backend:     "https://example.com/",
+	}}
```
AlexanderYastrebov, line 213: Please consider eskip.MustParse(`foo: * -> ...`)

eskipfile/remote_test.go
```diff
@@ -195,3 +196,120 @@ func createTestServer(c string, sc int) *httptest.Server {
 		io.WriteString(w, c)
 	}))
 }
+
+func TestRoutesCaching(t *testing.T) {
+	var expected = []*eskip.Route{{
```
AlexanderYastrebov, line 201:
```go
expected := ...
```

eskipfile/remote_test.go
```diff
@@ -195,3 +196,120 @@ func createTestServer(c string, sc int) *httptest.Server {
...
+			w.WriteHeader(http.StatusNotModified)
+			return
+		}
+		io.WriteString(w, fmt.Sprintf("VALID: %v;", routeBody))
+	}))
+	defer server.Close()
+
+	options := &RemoteWatchOptions{RemoteFile: server.URL, Threshold: 10, Verbose: true, FailOnStartup: true}
+	client, err := RemoteWatch(options) // first call
+	if err == nil {
```
AlexanderYastrebov, line 229: You may use `require.NoError(t, err)` or classic 
```go
if err != nil {
  t.Fatal(err)
  return
}
```

eskipfile/remote_test.go
```diff
@@ -195,3 +196,120 @@ func createTestServer(c string, sc int) *httptest.Server {
...
+				"/homepage/",
+			},
+		}},
+		BackendType: eskip.NetworkBackend,
+		Shunt:       false,
+		Backend:     "https://example.com/",
+	}}
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("ETag", "test-etag")
```
AlexanderYastrebov, line 216: This should be only set when we respond with 200

eskipfile/remote_test.go
```diff
@@ -195,3 +196,120 @@ func createTestServer(c string, sc int) *httptest.Server {
...
+	if err == nil {
+		defer client.(*remoteEskipFile).Close()
+	}
+
+	_, err = client.(*remoteEskipFile).getRemoteData() // cached, to make sure it returns the right status code
+
+	if !errors.Is(err, errContentNotChanged) {
+		t.Error(err)
+		return
+	}
```
AlexanderYastrebov, line 238: This is testing internals

eskipfile/remote_test.go
```diff
@@ -195,3 +196,120 @@ func createTestServer(c string, sc int) *httptest.Server {
...
+
+	if !eskip.EqLists(r, expected) {
+		t.Errorf("invalid cached routes received\n%s", cmp.Diff(r, expected))
+		return
+	}
+
+}
+
+func TestRoutesCachingWrongEtag(t *testing.T) {
+	var flushCache = false
```
AlexanderYastrebov, line 255: flushCache := false

AlexanderYastrebov: This variable is accessed by the test and test server so there is a race condition and the test likely will fail when launched with -race flag

AlexanderYastrebov: Consider using https://pkg.go.dev/sync/atomic#Bool

eskipfile/remote_test.go
```diff
@@ -69,107 +67,100 @@ func TestLoadAll(t *testing.T) {
...
 
-			if err == nil && test.fail {
-				t.Error("failed to fail")
-				return
-			} else if err != nil && !test.fail {
-				t.Error(err)
-				return
-			} else if test.fail {
+			if test.fail {
+				assert.Condition(t, func() bool { return err != nil }, "expected error")
```
AlexanderYastrebov, line 84: [asser](https://pkg.go.dev/github.com/stretchr/testify/assert#Error)

eskipfile/remote_test.go
```diff
@@ -69,107 +67,100 @@ func TestLoadAll(t *testing.T) {
...
+		contentUpdated: fmt.Sprintf("DIFFERENT_ID: %v;\nVALID: %v;", routeBody, routeBody),
+		expectedToFail: false,
 	},
 	} {
 		t.Run(test.title, func(t *testing.T) {
-			testValidServer := createTestServer(test.validRouteContent, 200)
-			defer testValidServer.Close()
+			ch := make(chan string, 1)
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				routeString := <-ch
```
AlexanderYastrebov, line 122: This will work but you have to send response content before each request.
Alternatively you can use e.g. https://pkg.go.dev/sync/atomic#Value and update it when needed

eskipfile/remote_test.go
```diff
@@ -189,9 +180,111 @@ func TestHTTPTimeout(t *testing.T) {
...
+
+}
+
+func TestRoutesCachingWrongEtag(t *testing.T) {
+	alterante := atomic.Int32{}
+	count200s := atomic.Int32{}
+	count304s := atomic.Int32{}
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if noneMatch := r.Header.Get("If-None-Match"); (alterante.Load()%2 != 0 && noneMatch == "test-etag") || (alterante.Load()%2 == 0 && noneMatch == "different-etag") {
```
AlexanderYastrebov, line 227:
```go
extectedEtag = "test-etag"
if alterante.Load()%2 == 0 {
    extectedEtag = "different-etag"
}

if r.Header.Get("If-None-Match") == extectedEtag {
...
}

File: eskipfile/remote_test.go [hunk 28]
```diff
@@ -189,9 +180,111 @@ func TestHTTPTimeout(t *testing.T) {
...
+	r, err := client.LoadAll()
+
+	require.NoError(t, err)
+
+	t.Logf("uncached responses received: %d", count200s.Load())
+	assert.Equal(t, int32(2), count200s.Load())
+	t.Logf("cached responses received: %d", count304s.Load())
+	assert.Equal(t, int32(0), count304s.Load())
+
+	expected := getExpectedRoutes(&alterante)
```
AlexanderYastrebov, line 259: Please inline getExpectedRoutes. It also should not depend on the alterante - at this point you know how many requests were done and should check expected result explicitly.

eskipfile/remote_test.go
MustafaSaber: I prefer to go with the channels for now to force setting the body.

szuecs: Since we do not have goroutines, we could also just change the content itself, without an issue.
Do I miss anything?

eskipfile/remote_test.go
szuecs: extectedEtag -> expectedTag
alterante -> alternate

eskipfile/remote_test.go
AlexanderYastrebov: test server serves in a separate goroutine

eskipfile/remote_test.go
MustafaSaber: why the ETag change?

eskipfile/remote.go
```diff
@@ -154,31 +161,42 @@ func isFileRemote(remotePath string) bool {
...
+
 	if resp.StatusCode != 200 {
-		return nil, errors.New("download file failed")
+		return nil, fmt.Errorf("failed to download remote file %s, status code: %d", client.remotePath, resp.StatusCode)
 	}
 
-	return resp.Body, nil
+	client.etag = resp.Header.Get("ETag")
+
+	return io.ReadAll(resp.Body)
```
AlexanderYastrebov, line 201: This reads everything into memory just to dump into the file a moment later. Previuosly it was correctly returning io.ReadCloser

AlexanderYastrebov:
:+1:

MustafaSaber:
👍

Issue #2307, opened 2023-05-10
Configurable backend response decompression


AlexanderYastrebov:
**Is your feature request related to a problem? Please describe.**

Skipper proxy component [uses http.Transport](https://github.com/zalando/skipper/blob/d85fb509049cc2a8d580941653b98f8af21f0dbf/proxy/proxy.go#L633-L647) to perform backend requests. When Skipper client does not specify that it accepts gzip response encoding, the http.Transport [adds `Accept-Encoding: gzip` request header](https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/net/http/transport.go;l=2567-2590) and if backend serves compressed response the Transport [will transparently decompress it](https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/net/http/transport.go;l=2209-2216) before serving back to the client. This behavior can be configured by the [`DisableCompression` http.Transport](https://pkg.go.dev/net/http#Transport) option:
```go
	// DisableCompression, if true, prevents the Transport from
	// requesting compression with an "Accept-Encoding: gzip"
	// request header when the Request contains no existing
	// Accept-Encoding value. If the Transport requests gzip on
	// its own and gets a gzipped response, it's transparently
	// decoded in the Response.Body. However, if the user
	// explicitly requested gzip it is not automatically
	// uncompressed.
	DisableCompression bool
```

E.g. consider the following two routes:
```
bin/skipper -inline-routes='
backend: Path("/backend") -> logHeader("request") -> inlineContent("backend response\n") -> <shunt>;
test: Path("/test") -> setPath("/backend") -> "http://localhost:9090"'
```

Logs of GET request to the `/test`:
```
~$ curl -v localhost:9090/test 
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 9090 (#0)
> GET /test HTTP/1.1
> Host: localhost:9090
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Length: 17
< Content-Type: text/plain; charset=utf-8
< Date: Wed, 10 May 2023 13:38:49 GMT
< Server: Skipper
< 
backend response
* Connection #0 to host localhost left intact
```
show that request to the "backend" receives `Accept-Encoding: gzip` header added by the transport:
```
[APP]INFO[0003] GET /backend HTTP/1.1
Host: localhost:9090
User-Agent: curl/7.58.0
Accept: */*
[...]

PR #2308, opened 2023-05-10
proxy: add uncompressed response counter

AlexanderYastrebov:
* Adds previously missing tests for proxy metrics
* Adds GetBody helper to the proxytest client

For https://github.com/zalando/skipper/issues/2307

### DIFF
--- a/proxy/metrics_test.go
+++ b/proxy/metrics_test.go
@@ -1,0 +1,65 @@
+package proxy_test
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/zalando/skipper/filters/builtin"
+	"github.com/zalando/skipper/metrics"
+	"github.com/zalando/skipper/metrics/metricstest"
+	"github.com/zalando/skipper/proxy/proxytest"
+	"github.com/zalando/skipper/routing"
+	"github.com/zalando/skipper/routing/testdataclient"
+)
+
+func TestMetricsUncompressed(t *testing.T) {
+	dm := metrics.Default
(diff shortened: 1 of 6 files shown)


proxy/proxy.go
```diff
@@ -915,6 +915,9 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 		return nil, &proxyError{err: fmt.Errorf("unexpected error from Go stdlib net/http package during roundtrip: %w", err)}
 	}
 	p.tracing.setTag(ctx.proxySpan, HTTPStatusCodeTag, uint16(response.StatusCode))
+	if response.Uncompressed {
+		p.metrics.IncCounter("experimental.uncompressed")
+	}
```
AlexanderYastrebov, line 920: We may decide to permanently disable transport compression.
In this case this metric will always be zero and useless.
So I used experimental prefix to hint that we may remove soon.

proxy/metrics_test.go
```diff
@@ -0,0 +1,65 @@
...
+	defer p.Close()
+
+	err := dc.UpdateDoc(fmt.Sprintf(`
+		test: Path("/test") -> setPath("/backend") -> "%s";
+		backend: Path("/backend") && Header("Accept-Encoding", "gzip") -> compress() -> inlineContent("backend response") -> <shunt>;
+	`, p.URL), nil)
+	require.NoError(t, err)
+
+	// wait for route update
+	time.Sleep(10 * time.Millisecond)
```
szuecs, line 45: I think you can also set `WaitFirstRouteLoad: true`

szuecs: That option is in skipper.Options, but you can likely use `./routing/routing.go:134:       SignalFirstLoad bool` and pass a channel or so. Maybe not in proxytest.

PR #2315, opened 2023-05-10
fix: http.ResponseController is available since Go-1.20

szuecs:
fix: http.ResponseController is available since Go-1.20

see also: https://github.com/zalando/skipper/actions/runs/4940453717/jobs/8832346991#step:5:157


szuecs:
:+1:

szuecs:
Seems not relevant...

PR #2332, opened 2023-05-15
tool: osv vulnerability scanner

szuecs:
tool: osv vulnerability scanner
https://github.com/zalando/skipper/security/code-scanning/110 suggest to add this and https://osv.dev/ looks like it makes sense to use as it is OpenSSF related project


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2333, opened 2023-05-15
Feature: http retry

szuecs:
Feature: http retry in httpclient


szuecs:
Features that would be interesting 
https://brooker.co.za/blog/2022/02/28/retries.html
https://aws.amazon.com/builders-library/timeouts-retries-and-backoff-with-jitter/

net/httpclient.go
```diff
@@ -125,9 +166,41 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
...
+	}
 	return c.client.Do(req)
 }
 
+func (c *Client) Retry(req *http.Request) (*http.Response, error) {
+	if req.Body == nil || req.Body == http.NoBody {
+		return c.Do(req)
+	}
+
+	if rc, err := req.GetBody(); err == nil {
```
szuecs, line 182: this seems to work in all cases I try, but this only works for 3 different types of body creations https://github.com/golang/go/blob/master/src/net/http/request.go#L877-L881 as far as I understand.

szuecs: GetBody() has a problem: 	// Next line panics on TestClientRetryBodyHalfReader

bjhaid:
From reading the code it is unclear if it tries a new backend. If it doesn't, is it desirable for the new `retry` filter to attempt a different backend from the previously failed one?

szuecs:
@bjhaid yes in case of loadbalancer backend it would retry by "running" again the algorithm, so in case of roundRobin it would try the current "next". I guess it's fine to run the same algorithm again, because if you think about a single backend error. For example a broken node issue that makes a backend instance to a half broke state, that can respond to "/ready" but not handle most of the requests anymore. I would guess a "network" backend target would exactly do the same: It will  run some lb algorithm in some load balancer in front of the application.

Do you see an issue with this behavior?

bjhaid:
> The retry executes https://github.com/zalando/skipper/pull/2333/files#diff-1d74870266949f1927a500e5f0002617239d3e3cd9bcceee6be014738e5e0f2bR1272 makeBackendRequest again, which contains the [call](https://github.com/zalando/skipper/blob/585758e246ad9a6c94cabc60edc8673a6a921a50/proxy/proxy.go#L922) to mapRequest() which will execute [selectEndpoint](https://github.com/zalando/skipper/blob/585758e246ad9a6c94cabc60edc8673a6a921a50/proxy/proxy.go#L568) which will [execute the lb algorithm](https://github.com/zalando/skipper/blob/585758e246ad9a6c94cabc60edc8673a6a921a50/proxy/proxy.go#L548)

> Do you see an issue with this behavior?

Thanks for the clarification! It wasn't obvious to me from reading the changeset. This behavior makes sense to me. I am also interested in seeing this ship thanks for reviving it ❤️

net/httpclient.go
```diff
@@ -125,9 +132,41 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 			req.Header.Set("Authorization", "Bearer "+string(b))
 		}
 	}
+	if req.Body != nil && req.Body != http.NoBody && req.ContentLength > 0 {
+		retryBuffer := skpio.NewCopyBodyStream(int(req.ContentLength), &bytes.Buffer{}, req.Body)
+		c.retryBuffers.Store(req, retryBuffer)
```
AlexanderYastrebov, line 137: Hm, I do not see a corresponding Delete. Looks like this will leak buffers.

AlexanderYastrebov: Do we need this in the client? Why not let users use CopyBodyStream like we do in proxy?

io/copy_stream.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (cb *CopyBodyStream) Len() int {
+	return cb.buf.Len()
+}
+
+func (cb *CopyBodyStream) Read(p []byte) (n int, err error) {
+	n, err = cb.input.Read(p)
+	if cb.left > 0 && n > 0 {
+		m := min(n, cb.left)
+		cb.buf.Write(p[:m])
```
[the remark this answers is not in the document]
szuecs: in-memory buffer write error -> oom, no?
I am fine to add log.Errorf if you think it's useful

proxy/proxy.go
```diff
@@ -1263,6 +1289,23 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 
 		if rsp.StatusCode >= http.StatusInternalServerError {
 			p.metrics.MeasureBackend5xx(backendStart)
+
+			if retryConfig != nil {
+				ctx.Logger().Infof("execute retry filter")
```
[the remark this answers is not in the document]
szuecs: I will drop the logs, IMO better than creating logs you never look at :)

net/httpclient.go
szuecs: The proxy code is about one request, but here it's about a client that could hundreds of requests and we need to get the right body for the retry. If we want to support a retry with body, we need to store it somewhere.

szuecs: I saw we can use LoadAndDelete() instead of Load(), but this would make only a single retry possible, but maybe that's good enough for now.

szuecs: > Do we need this in the client? Why not let users use CopyBodyStream like we do in proxy?

I thought it makes sense to provide it for simplicity as a user.

io/copy_stream.go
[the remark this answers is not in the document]
szuecs: fixed

io/copy_stream.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"bytes"
+	"io"
+)
+
+type bodyBuffer struct{ *bytes.Buffer }
+
+func (buf *bodyBuffer) Close() error {
+	return nil
+}
```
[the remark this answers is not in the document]
szuecs: done

io/copy_stream_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"io"
+	"testing"
+)
+
+type tbuf struct{ *bytes.Buffer }
+
+func (tb *tbuf) Read(p []byte) (int, error) {
+	return tb.Buffer.Read(p)
+}
+func (tb *tbuf) Close() error {
```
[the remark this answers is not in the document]
szuecs: done

bjhaid:
@szuecs is there anything I can do to help make progress on this PR.... I will love to see this ship.

szuecs:
@bjhaid sorry, I was in vacation and I just came back reading backlog of changes and emails.
Let me see what is missing to make it happen.

bjhaid:
> A great way to help would be to tell what you expect from retry() and what kind of configuration you would like to pass to this filter.

something like `retry(int N)` where `N` is the number of times it is retried(capped at some reasonable value) with a default like 3.

Issue #2301, opened 2023-05-08
Log traceId in proxy logger


RomanZavodskikh:
**Is your feature request related to a problem? Please describe.**
Log traceId in proxy logger in every log line it's known.
This change would increase the visibility of skipper because it would be possible to see a connections between log lines and traces inside Opentracing/Lightstep/any other tracing solution.

**Describe the solution you would like**
Take a look at the log lines nearby and inject traceId in every place it could be theoretically found out.
https://github.com/zalando/skipper/blob/master/proxy/proxy.go
To do so, we need to expose `GetTraceID` method somehow https://github.com/zalando/skipper/blob/v0.16.29/tracing/tracing.go#L157

**Would you like to work on it?**
Yes

szuecs:
@RomanZavodskikh I guess we can close this now.

PR #2347, opened 2023-05-25
Log traceId in proxy logger

RomanZavodskikh:
Updates #2301


szuecs:
:+1:

filters/filters.go
```diff
@@ -124,7 +124,7 @@ type FilterContext interface {
 	// Performs a new route lookup and executes the matched route if any
 	Loopback()
 
-	Logger() FilterContextLogger
+	Logger(...FilterContextLogger) FilterContextLogger
```
AlexanderYastrebov, line 127: This is not the best API design, it is not clear what happens if someone passes 0, 1, or many loggers.

proxy/context.go
```diff
@@ -284,7 +291,7 @@ func (c *context) Split() (filters.FilterContext, error) {
 	u.Host = originalRequest.Host
 	cr, body, err := cloneRequestForSplit(u, originalRequest)
 	if err != nil {
-		c.proxy.log.Errorf("context: failed to clone request: %v", err)
+		c.Logger(c.proxy.log).Errorf("context: failed to clone request: %v", err)
```
AlexanderYastrebov, line 294: This was we couple proxy logger with context logger.
I think we should add logger implementation that logs trace id and use it instead here:
https://github.com/zalando/skipper/blob/29c3ea1378faea74f7c19920ae2aeb341dd273c7/proxy/proxy.go#L715

AlexanderYastrebov: I see the problem now, we need to resolve trace id in runtime.
Maybe we can use `ctx.Logger()` instead of `c.proxy.log` where applicable?

RomanZavodskikh: I tried this, see https://github.com/zalando/skipper/commit/76edb7929c0d4234538becb5c6ff0022d915b7cd

The reason why I haven't use such solution are tests failures, see https://github.com/zalando/skipper/actions/runs/5078553903/jobs/9123211440

The reason is that `make cicheck` reads information from `c.proxy.log` to test logging.

AlexanderYastrebov: Then maybe we should fix failing tests to not rely on proxy log?

filters/filters.go
RomanZavodskikh: Okay, found some way not to change this API.

proxy/context.go
RomanZavodskikh: > Then maybe we should fix failing tests to not rely on proxy log?
When I have a look on failing tests, it's not obvious for me how to fix it because tests pass its own logger instance.
https://github.com/zalando/skipper/blob/v0.16.68/proxy/proxy_test.go#L162

However, as `proxy.context` has `proxy` field, I can get proxy log through it, without additional args.
Changed PR to do it.

proxy/context.go
```diff
@@ -218,11 +218,15 @@ func (c *context) ParentSpan() opentracing.Span        { return c.parentSpan }
 func (c *context) Logger() filters.FilterContextLogger {
 	if c.logger == nil {
```
szuecs, line 219: maybe better:
```
if c.logger != nil {
    return c.logger
}
...rest here...
```

proxy/context.go
```diff
@@ -218,11 +218,15 @@ func (c *context) ParentSpan() opentracing.Span        { return c.parentSpan }
...
-		if traceId != "" {
-			c.logger = log.WithFields(log.Fields{"trace_id": traceId})
+		if c.proxy != nil && c.proxy.log != nil {
+			c.logger = c.proxy.log
 		} else {
 			c.logger = log.StandardLogger()
 		}
+
+		if traceId != "" {
+			c.logger = c.logger.(*log.Logger).WithFields(log.Fields{"trace_id": traceId})
```
RomanZavodskikh, line 228: @AlexanderYastrebov , what do you think, can I just do type assertion here and assume it will never fail? Or should I better do type switch here instead?

szuecs: `*log.Logger` is `*logrus.Logger`, I would say no

szuecs: Why not add this to filters.FilterContextLogger and then no cast?

RomanZavodskikh: If you mean adding `WithFields(logrus.Fields)` method to `filters.FilterContextLogger` context, it makes no sense to do this, because this method has argument type specific from logrus. Given that, `filters.FilterContextLogger` interface could not abstract logging libraries other than logrus.

proxy/context.go
AlexanderYastrebov: This would essentially make ctx.Logger() return proxy.log. I am not sure whether it is a good idea.
Is it hard to fix proxy tests that rely on proxy log? How many are there?

RomanZavodskikh: 6 tests. Is it hard?.. Maybe there is some easy way, but I don't see.

```
 func TestBackendTimeoutInTheMiddleOfServiceResponse(t *testing.T) {
 func TestClientTimeout(t *testing.T) {
 func TestClientCancellation(t *testing.T) {
 func TestClientTimeoutBeforeStreaming(t *testing.T) {
 func TestFilterPanic(t *testing.T) {
 func TestFilterPanicPrintStackRate(t *testing.T) {
```

RomanZavodskikh: > This would essentially make ctx.Logger() return proxy.log

Actually I was thinking about it, and I still don't see any really convenient way to bind `loggingtest.Logger` (which is used by unit tests to count stats about logs and assigned to `c.proxy.log`) and `ctx.Logger()` which is created inside `proxy/proxy.go:ServeHttp` function.

Another solution which I could propose is to return proxy.log only if `ctx.Logger()` called inside unit tests.

This is what this code intended for:
```go
if c.proxy != nil && c.proxy.log != nil {
		switch c.proxy.log.(type) {
		case *loggingtest.Logger:
			c.logger = c.proxy.log
		}
	}
```

proxy/context.go
RomanZavodskikh: Discussed after standup, let's check two options:
1) Hijack test to read logrus output OR
2) Implement `withFields` in `logging.Logger`

Let's try what works in more beautiful way.

proxy/backendtimeout_test.go
```diff
@@ -69,6 +73,13 @@ func TestFastService(t *testing.T) {
 }
 
 func TestBackendTimeoutInTheMiddleOfServiceResponse(t *testing.T) {
+	var logBuf bytes.Buffer
+	log.SetOutput(&logBuf)
+	defer log.SetOutput(os.Stderr)
```
AlexanderYastrebov, line 78: Lets use https://pkg.go.dev/testing#T.Cleanup

proxy/backendtimeout_test.go
```diff
@@ -112,8 +123,9 @@ func TestBackendTimeoutInTheMiddleOfServiceResponse(t *testing.T) {
 		t.Errorf("expected partial content, got %s", content)
 	}
 
+	time.Sleep(100 * time.Millisecond)
```
AlexanderYastrebov, line 126: Since we need to wait for the logs in all cases, lets make `containsEveryLineCount` signature the same as `WaitFor`, i.e. require duration argument.

proxy/proxy_test.go
```diff
@@ -133,6 +135,34 @@ func (srw *syncResponseWriter) Len() int {
...
+		}
+
+		if found == count {
+			return false
+		}
+
+		found++
+		s = s[i+len(substr):]
+	}
+}
```
AlexanderYastrebov, line 153: https://pkg.go.dev/strings#Count

proxy/proxy_test.go
```diff
@@ -133,6 +135,34 @@ func (srw *syncResponseWriter) Len() int {
...
+
+func containsEveryLineCount(s, substr string, count int) bool {
+	l := strings.Split(substr, "\n")
+	for _, li := range l {
+		if !containsCount(s, li, count) {
```
AlexanderYastrebov, line 158: Are you sure we want count per line and not just a total count in s?

proxy/backendtimeout_test.go
```diff
@@ -69,6 +73,13 @@ func TestFastService(t *testing.T) {
 }
 
 func TestBackendTimeoutInTheMiddleOfServiceResponse(t *testing.T) {
+	var logBuf bytes.Buffer
```
AlexanderYastrebov, line 76: Maybe we can create a `testLog` type that would encapsulate buffer, take care of output restoration and have `WaitFor` method. It would be use like:
```go

tl := NewTestLog(t) // uses *testing.T to restore logrus output via T.Cleanup
...
if err = tl.WaitFor(msg, 100*time.Millisecond); err != nil {
...
}
```

proxy/proxy_test.go
RomanZavodskikh: Dropped this function.

proxy/context.go
RomanZavodskikh: Discussed with Alexander. It's better not to change production code to make tests happy. It must be vice versa.

proxy/backendtimeout_test.go
RomanZavodskikh: I think it should be used only when the amount of code here will be much larger.
```go
testLog := NewTestLog()
defer testLog.Close()
```

This code is simple and straightforward. using `T.Cleanup` will make it look not so straightforward:
```go
testLog := NewTestLog()
t.Cleanup(func() {
    testLog.Close()
})
```

proxy/proxy_test.go
```diff
@@ -73,6 +75,87 @@ type listener struct {
 	lastConn chan net.Conn
 }
 
+type blockableBuffer struct {
+	inner *bytes.Buffer
```
AlexanderYastrebov, line 79: * do not use pointer, name it buf
* move mutex field before buffer

proxy/proxy_test.go
```diff
@@ -73,6 +75,87 @@ type listener struct {
...
+	w.m.Unlock()
+}
+
+func (w *blockableBuffer) RLock() {
+	w.m.RLock()
+}
+
+func (w *blockableBuffer) RUnlock() {
+	w.m.RUnlock()
+}
```
AlexanderYastrebov, line 111: Does this need to be exported?

proxy/proxy_test.go
```diff
@@ -73,6 +75,87 @@ type listener struct {
 	lastConn chan net.Conn
 }
 
+type blockableBuffer struct {
+	inner *bytes.Buffer
+	m     sync.RWMutex
```
AlexanderYastrebov, line 80: Lets use plain mutex here.
Check the docs for RWMutex to better understand the difference to mutex.

proxy/proxy_test.go
```diff
@@ -73,6 +75,87 @@ type listener struct {
...
+type testLog struct {
+	buf      *blockableBuffer
+	oldOut   io.Writer
+	oldLevel log.Level
+}
+
+func NewTestLog() *testLog {
+	oldOut := os.Stderr
+	oldLevel := log.GetLevel()
+	buf := &blockableBuffer{inner: &bytes.Buffer{}, m: sync.RWMutex{}}
```
AlexanderYastrebov, line 122: you can omit `m: sync.RWMutex{}` and the innre as well if you change it to the value instead of the pointer

proxy/proxy_test.go
```diff
@@ -73,6 +75,87 @@ type listener struct {
...
+}
+
+type testLog struct {
+	buf      *blockableBuffer
+	oldOut   io.Writer
+	oldLevel log.Level
+}
+
+func NewTestLog() *testLog {
+	oldOut := os.Stderr
```
AlexanderYastrebov, line 120: It is better to capture the actual value and later restore it instead of assuming that it is equal to default stderr.

proxy/proxy_test.go
```diff
@@ -73,6 +75,87 @@ type listener struct {
...
+	timeout := time.After(to)
+	ticker := time.NewTicker(10 * time.Millisecond)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-timeout:
+			return fmt.Errorf("timeout waiting for log entry: %s", exp)
+		case <-ticker.C:
+			if l.Count(exp) >= n {
```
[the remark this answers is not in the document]
AlexanderYastrebov: It is fine, WaitForN waits for at least n occurrences

proxy/proxy_test.go
RomanZavodskikh: Removed it at all. :-)

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2348, opened 2023-05-25
test: add test to make sure that you can use lb algorithm with service

szuecs:
test: add test to make sure that you can use lb algorithm with service


MustafaSaber:
👍

PR #2360, opened 2023-05-31
Log traceId in proxy logger

RomanZavodskikh:
Updates #2301

### DIFF
--- a/filters/diag/absorb.go
+++ b/filters/diag/absorb.go
@@ -28,7 +28,7 @@ type absorb struct {
 
 func withLogger(silent bool, l logging.Logger) filters.Spec {
 	if l == nil {
-		l = &logging.DefaultLog{}
+		l = logging.New()
 	}
 
 	id, err := flowid.NewStandardGenerator(flowid.MinLength)
--- a/logging/logger.go
+++ b/logging/logger.go
@@ -1,9 +1,14 @@
 package logging
 
-import "github.com/sirupsen/logrus"
+import (
+	"github.com/sirupsen/logrus"
+)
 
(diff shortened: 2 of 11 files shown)


logging/logger.go
```diff
@@ -31,13 +36,32 @@ type Logger interface {
...
+}
+func (dl *DefaultLog) WithFields(fields map[string]interface{}) Logger {
+	for k, v := range fields {
+		dl.fields[k] = v
+	}
+	return dl
+}
+func New() *DefaultLog {
+	return &DefaultLog{logger: *logrus.New()}
+}
```
AlexanderYastrebov, line 67: I think it should be something like:
```go
type DefaultLog struct {
	logger logrus.Logger
}

func (dl *DefaultLog) Error(a ...interface{}) { dl.logger.Error(a...) }
// other logger methods omitted for brevity

func New() Logger {
	return &DefaultLog{logger: logrus.StandardLogger()}
}

func (dl *DefaultLog) WithFields(fields map[string]interface{}) Logger {
	return &DefaultLog{logger: dl.logger.WithFields(logrus.Fields(fields))}
}
```
i.e. there is no need to store `fields map[string]interface{}` and call WithFields on each logging call.

filters/diag/absorb.go
```diff
@@ -28,7 +28,7 @@ type absorb struct {
 
 func withLogger(silent bool, l logging.Logger) filters.Spec {
 	if l == nil {
-		l = &logging.DefaultLog{}
+		l = logging.New()
```
AlexanderYastrebov, line 31: Maybe we should refactor this filter to use `ctx.Logger()`

PR #2361, opened 2023-06-01
feature: add TTL support for routes in etcd

chahn:
Since we process a very large number of routes, we want to automatically clean up deprecated and thus expired routes.
Therefore, we use the TTL of etcd to automatically expire and remove old routes stored in etcd.

Currently routes with expired TTL are removed from etcd but stay active in Skipper until a restart of Skipper is performed.
Expiring routes from etcd provoke a parsing error when `LoadUpdate()` is executed.

This patch would allow Skipper to support both the already supported etcd `delete` action and additionally the `expire` action to ensure that routes are removed.

The tests are extended accordingly to verify the correct behavior of expiring routes.
`TestReceiveExpire` adds a route with the minimum TTL of 1 second. This route must expire and be removed.

### DIFF
--- a/etcd/etcd_test.go
+++ b/etcd/etcd_test.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/etcd/etcdtest"
@@ -307,6 +308,53 @@ func TestReceiveInsert(t *testing.T) {
 
 	if len(ds) != 0 {
 		t.Error("unexpected delete")
+	}
+}
+
+func TestReceiveExpire(t *testing.T) {
+	if testing.Short() {
+		t.Skip()
+	}
+	if err := etcdtest.DeleteAll(); err != nil {
+		t.Fatal(err)
+		return
+	}
+
+	c, err := New(Options{etcdtest.Urls, "/skippertest", 0, false, "", "", ""})
+	if err != nil {
+		t.Fatal(err)
+		return
+	}
+
+	_, err = c.LoadAll()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Will expire after a TTL of 1 second
+	etcdtest.PutDataToTTL("/skippertest", "pdp", `Path("/pdp") -> "https://expire.example.org"`, 1)
+
+	// Wait until the route is expired with a timeout of 5 seconds
+	timeout := time.After(5 * time.Second)
+	for {
+		select {
+		case <-timeout:
+			t.Error("Timeout: route should expire after 1 second but did not expire within 5 seconds")
+			return
+		default:
+			rs, es, err := c.LoadUpdate()
+			if err != nil {
+				t.Fatal(err)
+				return
+			}
+
+			if checkDeleted(es, "pdp") {
+				if len(rs) != 0 {
+					t.Fatal("unexpected upsert")
+				}
+				return
+			}
+		}
 	}
 }
 
--- a/etcd/etcdtest/mocketcd.go
+++ b/etcd/etcdtest/mocketcd.go
@@ -16,6 +16,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 )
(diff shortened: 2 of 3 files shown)


etcd/etcd_test.go
```diff
@@ -310,6 +311,50 @@ func TestReceiveInsert(t *testing.T) {
...
+
+	// Will expire after a TTL of 1 second
+	etcdtest.PutDataToTTL("/skippertest", "pdp", `Path("/pdp") -> "https://expire.example.org"`, 1)
+
+	// Wait until the route is expired with a timeout of 5 seconds
+	timeout := time.After(5 * time.Second)
+	for {
+		select {
+		case <-timeout:
+			t.Errorf("Timeout: route should expire after 1 second but did not expire within 5 seconds")
```
AlexanderYastrebov, line 339: minor: Maybe t.Fatalf() is more appropriate here

etcd/etcd_test.go
```diff
@@ -310,6 +311,50 @@ func TestReceiveInsert(t *testing.T) {
...
+			return
+		default:
+			rs, es, err := c.LoadUpdate()
+			if err != nil {
+				t.Error(err)
```
AlexanderYastrebov, line 344: t.Fatalf()

etcd/etcd_test.go
```diff
@@ -310,6 +311,50 @@ func TestReceiveInsert(t *testing.T) {
...
+		return
+	}
+
+	c, err := New(Options{etcdtest.Urls, "/skippertest", 0, false, "", "", ""})
+	if err != nil {
+		t.Error(err)
+		return
+	}
+
+	c.LoadAll()
```
AlexanderYastrebov, line 329: Error is not checked here

etcd/etcd_test.go
AlexanderYastrebov: I mean use t.Fatalf() whenever no error is expected and the check is not related to the change and use t.Errorf() to check errors related to the change.

etcd/etcdtest/mocketcd.go
```diff
@@ -169,8 +170,16 @@ func PutData(key, data string) error {
...
 // Saves a route in etcd with the specified prefix.
 func PutDataTo(prefix, key, data string) error {
+	return PutDataToTTL("/skippertest", key, data, -1)
+}
+
+// Saves a route with TTL in etcd with the specified prefix.
+func PutDataToTTL(prefix, key, data string, ttl int) error {
 	v := make(url.Values)
 	v.Add("value", data)
+	if ttl >= 0 {
```
AlexanderYastrebov, line 180: I think it is better to check that it is strictly greater than zero. This way we avoid "magic" constants

etcd/etcdtest/mocketcd.go
```diff
@@ -169,8 +170,16 @@ func PutData(key, data string) error {
 
 // Saves a route in etcd with the specified prefix.
 func PutDataTo(prefix, key, data string) error {
+	return PutDataToTTL("/skippertest", key, data, -1)
```
AlexanderYastrebov, line 173: `"/skippertest"` -> `prefix`

chahn: I fixed this. Thanks a lot.

etcd/etcdtest/mocketcd.go
chahn: Makes sense. I changed this to strictly greater than zero.

etcd/etcd_test.go
chahn: Error is checked now.

etcd/etcd_test.go
chahn: Thanks for the suggestion.
I now try to use t.Error in the context of the change and t.Fatal otherwise, e.g. while preparing the test.

etcd/etcd_test.go
chahn: Thanks for the suggestion.
I now try to use t.Error in the context of the change and t.Fatal otherwise, e.g. while preparing the test.

chahn:
Thank you @AlexanderYastrebov for your feedback.
I have tried to implement all your suggestions and recommendations in the latest commit.

etcd/etcdtest/mocketcd.go
```diff
@@ -169,8 +170,16 @@ func PutData(key, data string) error {
 
 // Saves a route in etcd with the specified prefix.
 func PutDataTo(prefix, key, data string) error {
+	return PutDataToTTL(prefix, key, data, -1)
```
AlexanderYastrebov, line 173: Should pass zero ttl here

chahn: Thanks!
TTL is zero now in the latest commit.
```
...
return PutDataToTTL(prefix, key, data, 0)
```

AlexanderYastrebov:
:+1:

szuecs:
:+1:

szuecs:
@chahn thanks!

chahn:
Thanks a lot for your support and the merge @AlexanderYastrebov and @szuecs

PR #2362, opened 2023-06-01
crd: update RouteGroup CRD

AlexanderYastrebov:
This change updates RouteGroup CRD from https://github.com/szuecs/routegroup-client

It adds a link to the origin to make it clear that CRD is the copy.

E.g. #1856 fixed hostname pattern directly and the problem in origin was rediscovered again and fixed by https://github.com/szuecs/routegroup-client/issues/10


MustafaSaber:
👍

mikkeloscar:
:+1:

Issue #2346, opened 2023-05-25
admission control like behavior for backend lb endpoints


szuecs:
**Is your feature request related to a problem? Please describe.**

Skipper proxy/loadbalancer should detect and drain load balancer endpoints (members) automatically if we detect single or low number of faulty/suffering endpoints. 

During an incident and more regular since a year we observe issues with faulty EC2 nodes. Faulty nodes in the sense of long period (some minutes) of latency and errors. Possibly also Accept-Q is saturated leading to timeouts in establishing TCP connections.

**Describe the solution you would like**

Similar to admissionControl filter I want to have a probability based load shedding that drains load balancer members.

**Describe alternatives you've considered (optional)**

I checked that the Linux Kernel can not be configured to send RST packets if the accept-Q is full. It will just drop packets. Maybe an alternative could be done with BPF/XDP. 

**Additional context (optional)**

backend roundtrip is executing the call https://github.com/zalando/skipper/blob/master/proxy/proxy.go#L872
we need to understand which error case is an issue. Maybe we need to count errors per lb member (in our code it's called `endpoint` similar to the kubernetes object) and if we have single outstanding endpoint that one we want to shedd.

LB intregration starts from here: https://github.com/zalando/skipper/blob/master/proxy/proxy.go#L820

algorithm of load shedding can be used similar to admissionControl filter with probability based load shedding
https://github.com/zalando/skipper/blob/master/filters/shedder/admission.go#L415

These were trials from longer time ago to filter out endpoints from the list https://github.com/zalando/skipper/blob/master/proxy/proxy.go#L898 https://github.com/zalando/skipper/blob/master/proxy/proxy.go#L903 , likely these are outdated and should be removed when working on this issue.

PR #2363, opened 2023-06-01
Make loadbalancer avoid sending requests to unhealthy nodes

RomanZavodskikh:
Updates #2346

### DIFF
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -77,7 +77,13 @@ func goodEndpoint(rnd *rand.Rand, ep *routing.LBEndpoint, ctx *routing.LBContext
 	)
 	goodFadeIn := f == 1 || rnd.Float64() < f
 
-	return goodFadeIn
+	failedReqs, totalReqs := (int64)(0), (int64)(0)
+	if ep.Metrics != nil {
+		failedReqs, totalReqs = ep.Metrics.GetFailedRequests()
+	}
+	goodHealthyEp := totalReqs <= 10 || float64(failedReqs)/float64(totalReqs) < 0.95 || rnd.Float64() < 0.05
+
+	return goodFadeIn && goodHealthyEp
 }
 
 type roundRobin struct {
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -67,87 +67,23 @@ func fadeIn(now time.Time, duration time.Duration, exponent float64, detected ti
 	return math.Pow(float64(rel)/float64(duration), exponent)
 }
 
-func shiftWeighted(rnd *rand.Rand, ctx *routing.LBContext, w []float64, now time.Time) routing.LBEndpoint {
-	var sum float64
-	weightSums := w
-	rt := ctx.Route
-	ep := ctx.Route.LBEndpoints
-	for _, epi := range ep {
-		wi := fadeIn(now, rt.LBFadeInDuration, rt.LBFadeInExponent, epi.Detected)
-		sum += wi
-		weightSums = append(weightSums, sum)
-	}
-
-	choice := ep[len(weightSums)-1]
-	r := rnd.Float64() * sum
-	for i := range weightSums {
-		if weightSums[i] > r {
-			choice = ep[i]
-			break
-		}
-	}
-
-	return choice
-}
-
-func shiftToRemaining(rnd *rand.Rand, ctx *routing.LBContext, wi []int, wf []float64, now time.Time) routing.LBEndpoint {
-	notFadingIndexes := wi
-	ep := ctx.Route.LBEndpoints
-
-	// if all endpoints are fading, the simplest approach is to use the oldest,
-	// this departs from the desired curve, but guarantees monotonic fade-in. From
-	// the perspective of the oldest endpoint, this is temporarily the same as if
-	// there was no fade-in.
-	if len(notFadingIndexes) == 0 {
-		return shiftWeighted(rnd, ctx, wf, now)
-	}
-
-	// otherwise equally distribute between the old endpoints
-	return ep[notFadingIndexes[rnd.Intn(len(notFadingIndexes))]]
-}
-
-func withFadeIn(rnd *rand.Rand, ctx *routing.LBContext, notFadingIndexes []int, choice int, algo routing.LBAlgorithm) routing.LBEndpoint {
-	ep := ctx.Route.LBEndpoints
+func goodEndpoint(rnd *rand.Rand, ep *routing.LBEndpoint, ctx *routing.LBContext) bool {
 	now := time.Now()
 	f := fadeIn(
 		now,
 		ctx.Route.LBFadeInDuration,
 		ctx.Route.LBFadeInExponent,
-		ctx.Route.LBEndpoints[choice].Detected,
+		ep.Detected,
 	)
+	goodFadeIn := f == 1 || rnd.Float64() < f
 
-	if rnd.Float64() < f {
-		return ep[choice]
-	}
-	for i := 0; i < len(ep); i++ {
-		if _, fadingIn := fadeInState(now, ctx.Route.LBFadeInDuration, ep[i].Detected); !fadingIn {
-			notFadingIndexes = append(notFadingIndexes, i)
-		}
-	}
-
-	switch a := algo.(type) {
-	case *roundRobin:
-		return shiftToRemaining(a.rnd, ctx, notFadingIndexes, a.fadingWeights, now)
-	case *random:
-		return shiftToRemaining(a.rand, ctx, notFadingIndexes, a.fadingWeights, now)
-	case *consistentHash:
-		// If all endpoints are fading, normal consistent hash result
-		if len(notFadingIndexes) == 0 {
-			return ep[choice]
-		}
-		// otherwise calculate consistent hash again using endpoints which are not fading
-		return ep[a.chooseConsistentHashEndpoint(ctx, skipFadingEndpoints(notFadingIndexes))]
-	default:
-		return ep[choice]
-	}
+	return goodFadeIn
 }
 
 type roundRobin struct {
-	mx               sync.Mutex
-	index            int
-	rnd              *rand.Rand
-	notFadingIndexes []int
-	fadingWeights    []float64
+	mx    sync.Mutex
+	index int
+	rnd   *rand.Rand
 }
 
 func newRoundRobin(endpoints []string) routing.LBAlgorithm {
@@ -155,10 +91,6 @@ func newRoundRobin(endpoints []string) routing.LBAlgorithm {
 	return &roundRobin{
 		index: rnd.Intn(len(endpoints)),
 		rnd:   rnd,
-
-		// preallocating frequently used slice
-		notFadingIndexes: make([]int, 0, len(endpoints)),
-		fadingWeights:    make([]float64, 0, len(endpoints)),
 	}
 }
 
@@ -170,19 +102,24 @@ func (r *roundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 
 	r.mx.Lock()
 	defer r.mx.Unlock()
-	r.index = (r.index + 1) % len(ctx.Route.LBEndpoints)
 
-	if ctx.Route.LBFadeInDuration <= 0 {
-		return ctx.Route.LBEndpoints[r.index]
+	endpoints := []routing.LBEndpoint{}
+	for _, e := range ctx.Route.LBEndpoints {
+		e := e
+		if goodEndpoint(r.rnd, &e, ctx) {
+			endpoints = append(endpoints, e)
+		}
 	}
+	if len(endpoints) == 0 {
+		endpoints = ctx.Route.LBEndpoints
+	}
+	r.index = (r.index + 1) % len(endpoints)
 
-	return withFadeIn(r.rnd, ctx, r.notFadingIndexes, r.index, r)
+	return endpoints[r.index]
 }
 
 type random struct {
-	rand             *rand.Rand
-	notFadingIndexes []int
-	fadingWeights    []float64
+	rand *rand.Rand
 }
 
 func newRandom(endpoints []string) routing.LBAlgorithm {
@@ -190,10 +127,6 @@ func newRandom(endpoints []string) routing.LBAlgorithm {
 	// #nosec
 	return &random{
 		rand: rand.New(rand.NewSource(t)),
-
-		// preallocating frequently used slice
-		notFadingIndexes: make([]int, 0, len(endpoints)),
-		fadingWeights:    make([]float64, 0, len(endpoints)),
 	}
 }
 
@@ -203,12 +136,19 @@ func (r *random) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 		return ctx.Route.LBEndpoints[0]
 	}
 
-	i := r.rand.Intn(len(ctx.Route.LBEndpoints))
-	if ctx.Route.LBFadeInDuration <= 0 {
-		return ctx.Route.LBEndpoints[i]
+	endpoints := []routing.LBEndpoint{}
+	for _, e := range ctx.Route.LBEndpoints {
+		e := e
+		if goodEndpoint(r.rand, &e, ctx) {
+			endpoints = append(endpoints, e)
+		}
 	}
+	if len(endpoints) == 0 {
+		endpoints = ctx.Route.LBEndpoints
+	}
+	i := r.rand.Intn(len(endpoints))
 
-	return withFadeIn(r.rand, ctx, r.notFadingIndexes, i, r)
+	return endpoints[i]
 }
 
 type (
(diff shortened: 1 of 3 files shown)


proxy/proxy.go
```diff
@@ -870,6 +870,11 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 	req = injectClientTrace(req, ctx.proxySpan)
 
 	response, err := roundTripper.RoundTrip(req)
+	if response.StatusCode >= 400 {
```
szuecs, line 873: this we can't do on >=400. We need to check the errors that return timeouts and connection errors. 
Everything that is a valid response we do accept as "success".

routing/routing.go
```diff
@@ -160,11 +166,44 @@ func (m *LBMetrics) DecInflightRequest() {
...
 func (m *LBMetrics) GetInflightRequests() int {
 	return int(atomic.LoadInt64(&m.inflightRequests))
 }
 
+// AddFailedRequest increments the counters related to failed requests and total requests from the proxy to a given backend.
+func (m *LBMetrics) AddFailedRequest() {
+	m.m.Lock()
+	defer m.m.Unlock()
+
+	m.failedRequests++
```
szuecs, line 179: why not `atomic.AddInt64(&m.failedRequests, 1)`

routing/routing.go
```diff
@@ -160,11 +166,44 @@ func (m *LBMetrics) DecInflightRequest() {
...
+	m.totalRequests++
```
szuecs, line 180: why not 
```
atomic.AddInt64(&m.failedRequests, 1)
atomic.AddInt64(&m.totalRequests, 1)
```

routing/routing.go
```diff
@@ -160,11 +166,44 @@ func (m *LBMetrics) DecInflightRequest() {
...
+}
+
+func (m *LBMetrics) AddSucceededRequest() {
+	m.m.Lock()
+	defer m.m.Unlock()
+
+	m.totalRequests++
```
szuecs, line 187: why not 
```
atomic.AddInt64(&m.totalRequests, 1)
```

routing/routing.go
RomanZavodskikh: Because in this section of code `m.m` is locked, so we can rely on no other goroutine using those fields.
So both usual and atomic instructions are correct.
However, atomic ones are slower, because they add `mfence` instructions, like here https://stackoverflow.com/questions/51846894/what-is-the-performance-of-stdatomic-vs-non-atomic-variables

routing/routing.go
RomanZavodskikh: Because in this section of code `m.m` is locked, so we can rely on no other goroutine using those fields.
So both usual and atomic instructions are correct.
However, atomic ones are slower, because they add `mfence` instructions, like here https://stackoverflow.com/questions/51846894/what-is-the-performance-of-stdatomic-vs-non-atomic-variables

loadbalancer/algorithm.go
```diff
@@ -208,7 +226,9 @@ func (r *random) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 		return ctx.Route.LBEndpoints[i]
 	}
 
-	return withFadeIn(r.rand, ctx, r.notFadingIndexes, i, r)
+	i = withFadeIn(r.rand, ctx, r.notFadingIndexes, i, r)
+	i = withHealthCheck(r.rand, ctx, r.notFadingIndexes, i, r)
```
lucastt, line 230: I wonder if it would not be better to actually have a new LB algorithm instead of modify random. Since you avoid unhealthy node it is not exactly random anymore.

loadbalancer/algorithm.go
```diff
@@ -133,12 +132,30 @@ func withFadeIn(rnd *rand.Rand, ctx *routing.LBContext, notFadingIndexes []int,
...
+	default:
+		return choice
+	}
+}
+
+func withHealthCheck(rnd *rand.Rand, ctx *routing.LBContext, notFadingIndexes []int, choice int, algo routing.LBAlgorithm) int {
+	now := time.Now()
+	failedReqs, totalReqs := ctx.Route.LBEndpoints[choice].Metrics.GetFailedRequests()
+
+	if totalReqs == 0 || float64(failedReqs)/float64(totalReqs) < 0.95 {
```
lucastt, line 148: This behavior is ok, but looks more like a CB than a admission control behavior. I might be reading this wrong but what I understand here is: Once a backend reaches 95% error rate we just stop sending requests to it until there is no active requests to this backend, then we try again and if the backend shows apparent recover we keep sending requests to it, if it reaches 95% error rate again we stop and so on.

First I don't have big problems with this behavior but from the task description what I would expect is load being drained from this backend until we reach a point where the backend reach a equilibrium and is able to serve the requests sent to it. This is much less aggressive behavior too. And is a admission control like behavior.

Have you thought about the pros and cons of this two approaches? If yes, would you mind describing then a bit for me to understand the reasoning?

From a quick look my thinking is: in case of network partition the CB behavior would be better because it acts much faster, but in case of overload admission control seems to fare better since it will not just ditch the load of a overloaded backend to other backends that might be struggling or not. Basicaly in the overload case CB behavior could spread the failure to other backends.

If there is a case that this behaviour could be a choice for the user it might make sense to make it configurable through some flag or parameter.

loadbalancer/algorithm.go
lucastt: This same thing could be applied to round robin also. Instead of a whole LB algorithm this could be just an option to both algorithms like and configurable through some flag or parameter.

proxy/proxy.go
lucastt: I agree with the fact error code >=400 is not what we want because 4xx is a valid response, its client side error. But 5xx even being a valid response means something is up with the backend, 5xx are always server side errors. So maybe something like `err != nil || esponse.StatusCode >= 500` would be better?

routing/routing.go
lucastt: I'm not sure if this post is the same case we are seeing here. They are explicitly comparing a normal operation vs an atomic one, but the performance constraints also involve the mutex lock. Atomic instructions are busy waiting, mutex free up CPU for other threads. I can't tell one or the other is more efficient for our use case without benchmarks, but our use case seems different from the question in your link. Or I might be imagining things, either way its a valid discussion. As food for thought: https://stackoverflow.com/questions/15056237/which-is-more-efficient-basic-mutex-lock-or-atomic-integer

loadbalancer/algorithm.go
RomanZavodskikh: Thanks for feedback Lucas! The very first thing to do from my PoV is to split withFadeIn and withHealthCheck parts to make them not so coupled together.
But from the first glance I would like AC (not CB) to be here, because I would like some requests to go to unhealthy node just to find out the moment when it will be healthy again.

routing/routing.go
szuecs: I swapped the lines, because you want to add total before failed, such that you do not count failed more than total (could lead to failed is bigger than total).
In general I am sure that lock will be slower as @lucastt summarized. If you like to check it, create a micro benchmark. Do not trust random posts by people, even if this would be super-hero-popular-10x-coder, it can be wrong.

routing/routing.go
szuecs: I guess we discuss it in the comment above https://github.com/zalando/skipper/pull/2363#discussion_r1213458051

loadbalancer/algorithm.go
szuecs: We need to have a strategy pattern somewhere.
I think what we want is:
- probability based drops (or better re-schedule), similar to admissionControl reject calculation to allow small amount of traffic to broken endpoints to observe PHC (passive health check)
- exponential backoff

loadbalancer/algorithm.go
szuecs: We need to have a strategy pattern somewhere.
I think what we want is:
- probability based drops (or better re-schedule), similar to admissionControl reject calculation to allow small amount of traffic to broken endpoints to observe PHC (passive health check)
- exponential backoff

routing/routing.go
szuecs: Make sure you increase first total requests and after that the failed ones.

loadbalancer/algorithm.go
```diff
@@ -67,98 +67,36 @@ func fadeIn(now time.Time, duration time.Duration, exponent float64, detected ti
...
+func goodEndpoint(rnd *rand.Rand, ep *routing.LBEndpoint, ctx *routing.LBContext) bool {
 	now := time.Now()
 	f := fadeIn(
 		now,
 		ctx.Route.LBFadeInDuration,
 		ctx.Route.LBFadeInExponent,
-		ctx.Route.LBEndpoints[choice].Detected,
+		ep.Detected,
 	)
+	goodFadeIn := rnd.Float64() < f
```
RomanZavodskikh, line 78: `f == 1 || rnd.Float64() < f` not to call rnd without need?

loadbalancer/algorithm.go
```diff
@@ -170,30 +108,31 @@ func (r *roundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 
 	r.mx.Lock()
 	defer r.mx.Unlock()
-	r.index = (r.index + 1) % len(ctx.Route.LBEndpoints)
 
-	if ctx.Route.LBFadeInDuration <= 0 {
-		return ctx.Route.LBEndpoints[r.index]
+	endpoints := []routing.LBEndpoint{}
```
RomanZavodskikh, line 112: Can I avoid repetitions (this looks the same as in random and almost the same as in consistent-hast)?

loadbalancer/algorithm.go
```diff
@@ -67,98 +67,36 @@ func fadeIn(now time.Time, duration time.Duration, exponent float64, detected ti
...
-	}
-	for i := 0; i < len(ep); i++ {
-		if _, fadingIn := fadeInState(now, ctx.Route.LBFadeInDuration, ep[i].Detected); !fadingIn {
-			notFadingIndexes = append(notFadingIndexes, i)
-		}
+	failedReqs, totalReqs := (int64)(0), (int64)(0)
+	if ep.Metrics != nil {
+		failedReqs, totalReqs = ep.Metrics.GetFailedRequests()
 	}
+	goodHealthyEp := totalReqs <= 10 || float64(failedReqs)/float64(totalReqs) < 0.95 || rnd.Float64() < 0.05
```
RomanZavodskikh, line 84: Such constants look like good baseline values to me. How can I make them configurable in a good way?

loadbalancer/algorithm.go
```diff
@@ -67,98 +67,36 @@ func fadeIn(now time.Time, duration time.Duration, exponent float64, detected ti
...
+	goodFadeIn := f == 1 || rnd.Float64() < f
 
-	if rnd.Float64() < f {
-		return ep[choice]
-	}
-	for i := 0; i < len(ep); i++ {
-		if _, fadingIn := fadeInState(now, ctx.Route.LBFadeInDuration, ep[i].Detected); !fadingIn {
-			notFadingIndexes = append(notFadingIndexes, i)
-		}
+	failedReqs, totalReqs := (int64)(0), (int64)(0)
```
szuecs, line 80: why not use the defaults:
```
var failedReqs, totalReqs int64
```

loadbalancer/algorithm.go
szuecs: I guess we would have to store them in a new type and you have no "pure" func, but a "method".

loadbalancer/algorithm.go
```diff
@@ -170,30 +108,31 @@ func (r *roundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
-	if ctx.Route.LBFadeInDuration <= 0 {
-		return ctx.Route.LBEndpoints[r.index]
+	endpoints := []routing.LBEndpoint{}
+	for _, e := range ctx.Route.LBEndpoints {
+		e := e
+		if goodEndpoint(r.rnd, &e, ctx) {
+			endpoints = append(endpoints, e)
+		}
+	}
+	if len(endpoints) == 0 {
```
lucastt, line 119: What will happen if a very small number of endpoints is healthy? Shouldn't we have some kind of guardrail for this? In this case there is two possibilities:

1. This algorithm sheds some load;
2. This algorithm does not mind if skipper overload some healthy backends;

The first option is kinda breaking the separation of concerns of these features, I don't think an LB algorithm should shed load, it seems a bit out of place and this is not what the author chose to do, which I agree. 

But there is some concern that if very few endpoints are healthy this would make the system collapse. So maybe this algorithm should only be available to routes with some form of load shedding strategy configured. I don't have a well formed opinion about it, what do you think?

Well, at least we should document this risk very well.

loadbalancer/fadein_test.go
```diff
@@ -146,33 +146,13 @@ func testFadeIn(
...
 			// Print CSV-like output for, where row number represents time and
-			// column represents endpoint.
+			// column represents endpoint. You can vizualize it using
+			// ./skptesting/run_fadein_test.sh from the skipper repo root.
 			t.Log("CSV " + fmt.Sprintf("%d,", i) + strings.Join(showStats, ","))
 		}
 	})
 }
 
-// For each endpoint, return a hash key which will make the consistent hash algorithm select it.
```
lucastt, line 155: Why is this test not valid anymore?

loadbalancer/fadein_test.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	fadeInDuration    = 100 * time.Millisecond
+	fadeInDuration    = 500 * time.Millisecond
```
lucastt, line 15: This is a rather specific change, why is 500ms better than 100ms here? Is this for visualization reasons? Because to me it should work independently of the `fadeInDuration` here.

RomanZavodskikh: No, visualization works fine regardless of `fadeInDuration`. Such changes is needed for tests like `TestFadeIn/consistent-hash,_5` (in `loadbalancer/fadein_test.go`) because in the beginning of it there are some hosts are being "warmed up" and therefore they receive not so many requests (ones or tens) and so there're good changes that by some random reason the number of requests will go down by some little (ones or tens) amount.

Basically, this test checks that number of requests sent to every host is monotonic (decreases for old hosts, increases for fading-in hosts), so basically for fading-in hosts the condition looks like "for every period of time the number of requests in the next period of time should be no less that 0.6*number of requests in the current period". When the number of requests is small, some random reason (in the Go scheduler or random generator, for example) may easily make this condition false even for rightly implemented load balancer.

That's why I increase fadeInDuration: to increase the time window during which reqs are collected, so increase the number of requests in every specific moment of time, so decrease the influence of random events here.

loadbalancer/fadein_test.go
```diff
@@ -76,15 +76,15 @@ func testFadeIn(
...
 		var stats []string
 		stop := time.After(fadeInDuration)
-		// Emulate the load balancer loop, sending requests to it with iterative hash keys
-		// of every endpoint over and over again till fadeIn period is over.
+		// Emulate the load balancer loop, sending requests to it with random hash keys
+		// over and over again till fadeIn period is over.
 		func() {
 			for {
-				ctx.Params[ConsistentHashKey] = hashKeys[len(stats)%len(hashKeys)]
+				ctx.Params[ConsistentHashKey] = strconv.Itoa(rnd.Intn(100000))
```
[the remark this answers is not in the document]
RomanZavodskikh: The old version of test code creates the limited amount of keys: one key for one host, so basically in case of 6 hosts (3 "old" ones: `a, b, c` and 3 "fading-in": `d, e, f`) this code will send request with the following key values in the loop:
```
h_a
h_b
h_c
h_d
h_e
h_f
h_a
h_b
... (~100k reqs for single test)
```

As in the beginning of this test, nodes `d, e, f` process very few amount of requests (like here https://github.com/zalando/skipper/pull/2363#discussion_r1223171116), instead such requests are redirected to the corresponding host in the consistent hash circle (https://www.toptal.com/big-data/consistent-hashing). 
For example, requests to **d** may be redirected to **e**, **e** - to **f**, **f** - to **a**.
In the old version this corresponding is constant, because the same six keys are used, so there is high chance that host **d** will receive not so many requests.

So, the solution I propose here is to use different keys for the forementioned redirection to be not constant, but random instead.

loadbalancer/fadein_test.go
RomanZavodskikh: After random keys are used here
https://github.com/zalando/skipper/pull/2363#discussion_r1223200218
this function is not called nowhere.

loadbalancer/fadein_test.go
RomanZavodskikh: This PR tends to be too huge, moved this part to another PR: https://github.com/zalando/skipper/pull/2380

loadbalancer/algorithm.go
```diff
@@ -170,30 +108,31 @@ func (r *roundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 
 	r.mx.Lock()
 	defer r.mx.Unlock()
-	r.index = (r.index + 1) % len(ctx.Route.LBEndpoints)
 
-	if ctx.Route.LBFadeInDuration <= 0 {
-		return ctx.Route.LBEndpoints[r.index]
+	endpoints := []routing.LBEndpoint{}
+	for _, e := range ctx.Route.LBEndpoints {
```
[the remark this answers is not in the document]
RomanZavodskikh: The first solution I see is to create goroutine updating such slice for every route. So it mease "1 goroutine / 1 route". That could lead to creation of ~10k goroutines, which is ~10 times bigger we have now.

RomanZavodskikh:
Done in the scope of https://github.com/zalando/skipper/pull/2888 .

PR #2364, opened 2023-06-01
Add Open Policy Agent based filters

mjungsbluth:
This is an initial PR to embed Open Policy Agent as a library and expose it via Skipper filters. 

The initial two filters allow the following two use cases:
- API Authorisation like offered by Envoy
- Serving Requests via an OPA policy without forwarding requests

It works without assumptions on a specific control plane the same way OSS Open Policy Agent would work

This PR is currently flagged as draft to discuss several items. The code itself is tested and has also been validated by building it as a plugin and running some use cases against it. 

Discussion points:
- Dependencies: This adds quite a few new dependencies in `go.mod`
- Filters for partial evaluation and for inline Rego are not part of this PR and can be added later
- Logging: OPA is quite chatty, this is exposed via the same logrus logger that Skipper uses with some context attributes to highlight that it is coming from OPA
- Metrics: OPA metrics are currently not exposed as Skipper metrics
- Tracing: This PR adds basic tracing for the policy decision and it adds the bundle name + decision IDs as tags. Errors that happen during evaluation are also added to both the Span and logged. The latter part should be discussed as this is only needed as far as I can see if no tracing is enabled. 

Many of the above points probably do not need to be addressed with the first PR but it would be helpful to capture future direction in the discussion. 

Open after review:
- [x] Document tracing in operation docs
- [x] Document metrics in operational docs
- [ ] Create issue to create filters for partial evaluation and for inline Rego that are not part of this PR and can be added later


go.mod
```diff
@@ -16,19 +16,23 @@ require (
...
 	github.com/miekg/dns v1.1.54
 	github.com/oklog/ulid v1.3.1
+	github.com/open-policy-agent/opa v0.52.0
+	github.com/open-policy-agent/opa-envoy-plugin v0.52.0-envoy
 	github.com/opentracing/basictracer-go v1.1.0
 	github.com/opentracing/opentracing-go v1.2.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.15.1
-	github.com/rcrowley/go-metrics v0.0.0-20181016184325-3113b8401b8a
+	github.com/rcrowley/go-metrics v0.0.0-20201227073835-cf1acfcdf475
```
mjungsbluth, line 35: This change breaks the CodaHale tests even though they can be individually run

szuecs: please rebase on top of master in case https://github.com/zalando/skipper/pull/2373 was merged. This will fix the metrics test issues

szuecs:
To the Discussion points:

>    Dependencies: This adds quite a few new dependencies in go.mod

That's fine because it's a major feature that we want to support.

>    Filters for partial evaluation and for inline Rego are not part of this PR and can be added later

That sounds like we should create an "enhancement" issue.

>    Logging: OPA is quite chatty, this is exposed via the same logrus logger that Skipper uses with some context attributes to highlight that it is coming from OPA

I think we want to see only errors with interaction with OPA not the decision if it's possible.

>    Metrics: OPA metrics are currently not exposed as Skipper metrics

I think latency and decisions (allow/deny counters) would be great to have. 

>    Tracing: This PR adds basic tracing for the policy decision and it adds the bundle name + decision IDs as tags. Errors that happen during evaluation are also added to both the Span and logged. The latter part should be discussed as this is only needed as far as I can see if no tracing is enabled.

sounds good would be nice to copy a picture from this to the PR.

go.mod
mjungsbluth: Done and that fixed the issue

mjungsbluth:
> That sounds like we should create an "enhancement" issue.

Yes, will do this once this PR is close to finalisation. 

> I think we want to see only errors with interaction with OPA not the decision if it's possible.

It currently logs decisions on debug. It logs errors (that also result in a 500) for requests as well as adding it to the trace. This can still flood the logs in case sth fundamentally goes wrong (i.e. if you break the policy rules with structural problems). I am not sure how is is done in other parts. 

> Re: tracing. sounds good would be nice to copy a picture from this to the PR or better put it into docs/operations/operations.md in our operators guide.

Yes, that makes sense. We will need to capture the traces first to do that.

config/config.go
```diff
@@ -270,6 +270,10 @@ type Config struct {
 
 	LuaModules *listFlag `yaml:"lua-modules"`
 	LuaSources *listFlag `yaml:"lua-sources"`
+
+	// Open Policy Agent
```
szuecs, line 274: Please drop all comments that are obvious, thanks!

config/config.go
```diff
@@ -478,6 +482,10 @@ func NewConfig() *Config {
 	flag.Var(cfg.CredentialPaths, "credentials-paths", "directories or files to watch for credentials to use by bearerinjector filter")
 	flag.DurationVar(&cfg.CredentialsUpdateInterval, "credentials-update-interval", 10*time.Minute, "sets the interval to update secrets")
 
+	// Open Policy Agent
```
szuecs, line 485: I think OPA should be part of Auth, so drop comment and move up the lines.

docs/reference/filters.md
```diff
@@ -1695,6 +1695,212 @@ oidcClaimsQuery("/:name%\"*One\"", "/path:groups.#[%\"*-Test-Users\"] groups.#[=
 
 As of now there is no negative/deny rule possible. The first matching path is evaluated against the defined query/queries and if positive, permitted.
 
+## Open Policy Agent
```
szuecs, line 1698: I really like these docs, but I would put it into docs/tutorials/auth.md. The filter reference should explain the filter itself from the filter user pov not how to setup skipper. A link from reference to the tutorial will help to find the setup, in case someone found the filter reference and wants to understand how to set skipper up to do auth with OPA. I hope this makes sense to you.

filters/openpolicyagent/authorizewithregopolicy/authorizewithregopolicy.go
```diff
@@ -0,0 +1,163 @@
...
+
+func NewAuthorizeWithRegoPolicySpec(factory openpolicyagent.OpenPolicyAgentFactory, opts ...func(*openpolicyagent.OpenPolicyAgentInstanceConfig) error) Spec {
+	return Spec{
+		factory:    factory,
+		configOpts: opts,
+	}
+}
+
+func (s Spec) Name() string {
+	return "authorizeWithRegoPolicy"
```
szuecs, line 28: use a const in filters/filters.go (all filter names are declared there)

filters/openpolicyagent/internal/envoy/policytype.go
```diff
@@ -0,0 +1,21 @@
...
+import "errors"
+
+type PolicyType string
+
+const (
+	IngressPolicyType PolicyType = "ingress"
+
+	EgressPolicyType PolicyType = "egress"
+
+	AppPolicyType PolicyType = "app"
```
szuecs, line 12: enum you do normally like this:
```
type PolicyType int

const (
	Ingress PolicyType = iota + 1
	Egress
        App
)
```

filters/openpolicyagent/internal/envoy/envoyplugin.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if err := cfg.ParseQuery(); err != nil {
+		return nil, err
+	}
+
+	return &cfg, nil
+}
+
+func (p *Plugin) Reconfigure(ctx context.Context, config interface{}) {
+
```
szuecs, line 46: missing or not needed? Maybe doc why it's not implemented.

filters/openpolicyagent/common.go
```diff
@@ -0,0 +1,7 @@
+package openpolicyagent
+
+const (
+	OpenPolicyAgentDecisionKey        = "open-policy-agent:decision"
+	OpenPolicyAgentDecisionBodyKey    = "open-policy-agent:decision-body"
+	OpenPolicyAgentDecisionHeadersKey = "open-policy-agent:decision-headers"
```
szuecs, line 6: I would move all const to filters/openpolicyagent/openpolicyagent.go and drop this file

filters/openpolicyagent/authorizewithregopolicy/authorizewithregopolicy_test.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+		delete(actualValueSet, val)
+	}
+
+	return len(actualValueSet) == 0
+}
+
+func isHeadersAbsent(unwantedHeaders http.Header, headers http.Header) bool {
+	for headerName := range unwantedHeaders {
+		if _, headerFound := headers[headerName]; headerFound {
```
szuecs, line 225: In Go you would always use `ok` instead of for example `headerFound`

filters/openpolicyagent/authorizewithregopolicy/authorizewithregopolicy.go
```diff
@@ -0,0 +1,163 @@
...
+
+	if len(sargs) < 1 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	if len(sargs) > 2 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	bundleName := sargs[paramBundleName]
```
szuecs, line 52: I would say keep it simple without const and use plain `sargs[0]` and `sargs[1]`.

filters/openpolicyagent/authorizewithregopolicy/authorizewithregopolicy.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	f.opa.Logger().WithFields(map[string]interface{}{
+		"query":               f.opa.EnvoyPluginConfig().ParsedQuery.String(),
+		"dry-run":             f.opa.EnvoyPluginConfig().DryRun,
+		"decision":            result.Decision,
+		"err":                 err,
+		"txn":                 result.TxnID,
+		"metrics":             result.Metrics.All(),
+		"total_decision_time": time.Since(start),
```
szuecs, line 108: I think for a debug log that nobody enables a timer is a bit expensive. I would prefer to expose a metric with the same value, but on the other hand we also measure filter runtime in the proxy so I would question the time measurement here.

filters/openpolicyagent/authorizewithregopolicy/authorizewithregopolicy.go
```diff
@@ -0,0 +1,163 @@
...
+	}).Debug("Authorizing request with decision.")
+
+	if !f.opa.EnvoyPluginConfig().DryRun {
```
szuecs, line 111: The "good" code path on the left
```
if f.opa.EnvoyPluginConfig().DryRun {
   return
}
...
```

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -0,0 +1,318 @@
...
+	// This allows to save memory on bundles that are shared
+	// between different policies (i.e. global team memberships)
+	// This not possible due to some limitations in OPA
+	// See https://github.com/open-policy-agent/opa/issues/5707
+	bundleStorage storage.Store
+}
+
+func NewOpenPolicyAgentFactory() OpenPolicyAgentFactory {
+	return OpenPolicyAgentFactory{
+		//BundleStorage: inmem.New(),
```
szuecs, line 40: ?

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -0,0 +1,318 @@
...
+		cfg.configTemplate, err = os.ReadFile(configTemplate())
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return &cfg, nil
+}
+
+func configTemplate() string {
```
szuecs, line 116: Either this
```
func configTemplate() string {
	if s, ok = os.LookupEnv("OPA_CONFIG_TEMPLATE_FILE"); ok {
		return s
	}
	return "opaconfig.yaml"
}
```
or this
```
func configTemplate() string {
	if s, ok = os.LookupEnv("OPA_CONFIG_TEMPLATE_FILE"); !ok {
		return "opaconfig.yaml"
	} else {
		return s
	}
}
```

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -0,0 +1,318 @@
...
+
+func (cfg *OpenPolicyAgentInstanceConfig) GetEnvoyContextExtensions() map[string]string {
+	return cfg.envoyContextExtensions
+}
+
+func NewOpenPolicyAgentConfig(opts ...func(*OpenPolicyAgentInstanceConfig) error) (*OpenPolicyAgentInstanceConfig, error) {
+	cfg := OpenPolicyAgentInstanceConfig{}
+
+	cfg.policyType = envoy.IngressPolicyType
+	if val, found := os.LookupEnv("OPA_ENVOY_POLICY_TYPE"); found {
```
szuecs, line 91: in Go: `ok` instead of `found`

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -0,0 +1,318 @@
...
+	if configTemplate, found = os.LookupEnv("OPA_CONFIG_TEMPLATE_FILE"); !found {
+		configTemplate = "opaconfig.yaml"
+	}
+
+	return configTemplate
+}
+
+func (factory *OpenPolicyAgentFactory) BundleStorage() storage.Store {
+	if factory.bundleStorage == nil {
+		return inmem.New()
```
szuecs, line 128: not sure how often this will be used. Does it makes sense to store it in `factory.bundleStorage`?

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	ctx := context.Background()
+	if err = engine.Start(ctx); err != nil {
+		return nil, err
+	}
+
+	if err := engine.waitPluginsReady(
+		100*time.Millisecond,
+		time.Second*time.Duration(30)); err != nil {
```
szuecs, line 155: this reads a bit wild. Can you unfold the func please and maybe reuse `err`?
so 
```
err = engine.waitPluginsReady(100*time.Millisecond, time.Second*time.Duration(30))
if err != nil {...}
```

filters/openpolicyagent/response.go
```diff
@@ -0,0 +1,75 @@
...
+		}).Info("Rejecting request because of an invalid decision")
+	}
+
+	fc.Serve(&resp)
+}
+
+func (opa *OpenPolicyAgentInstance) ServeResponse(fc filters.FilterContext, span opentracing.Span, result *envoyauth.EvalResult) {
+	resp := http.Response{}
+
+	status, err := result.GetResponseHTTPStatus()
```
szuecs, line 48: drop intermediate and use resp.StatusCode will reduce memory footprint:
```
var err error
resp.StatusCode , err =  result.GetResponseHTTPStatus()
if er != nil {...}
```

The same pattern applies below with `resp.Header`

config/config.go
mjungsbluth: You can also argue that it is not: one of the two filters that is introduced is for serving requests, not authz. 

Your call, I am happy with either and agree that many people would file it under auth

docs/reference/filters.md
mjungsbluth: Totally, will split

filters/openpolicyagent/authorizewithregopolicy/authorizewithregopolicy.go
mjungsbluth: Yes, I would also prefer to remove the debug logs. You can also still enable console decision logging via OPA so there is little benefit anyways. This currently just mimics what the ops Envoy plugin logs.

As for the metrics: Yes that is sth that I also had in mind. We can just measure the evaluation time which is a bit less than the proxy timing and handling the response around and at least gives insights into the latency introduced by policy evaluation itself. 

So I'd propose:
- Remove the debug logging
- Add a metric for eval times per filter instance (identified by "bundle name"), that way we can split the eval latencies per bundle as eval times between two bundles are not really comparable.

wdyt?

filters/openpolicyagent/evaluation.go
```diff
@@ -0,0 +1,88 @@
...
+		span.SetTag("decision_id", result.DecisionID)
+	}
+
+	if err != nil {
+		opa.Logger().WithFields(map[string]interface{}{"err": err}).Error("Unable to generate decision ID.")
+		return nil, func() error { return nil }, err
+	}
+
+	var input map[string]interface{}
+	stop := func() error {
```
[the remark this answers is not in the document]
mjungsbluth: I was actually pretty unhappy with this piece (this is also similar in the OPA Envoy plugin), that will clean it up a lot...

filters/openpolicyagent/internal/envoy/envoyplugin.go
mjungsbluth: Part of the interface. Will document why the empty implementation is ok.

filters/openpolicyagent/internal/envoy/policytype.go
mjungsbluth: Agreed. This is however part of how OPA integrates with Envoy. I.e. these exact strings are passed and also respected / required by Styra.

Let me check if we can remove these strings alltogether, I think they are just passed through, so we could also pass them as command line arg.

filters/openpolicyagent/openpolicyagent.go
mjungsbluth: Yes, the issue that is required to put this back in is anyways linked a few lines up. Will remove

filters/openpolicyagent/openpolicyagent.go
mjungsbluth: Yes, that is where I want it as that would also share memory. There is an issue raised in OPA to support  this but it does not work today (See the comment on the factory). 

Factually this is called once per filter instance but agree that this is not clear from the code. Let me rework this a bit that the code makes sense today.

filters/openpolicyagent/authorizewithregopolicy/authorizewithregopolicy.go
szuecs: Sounds like a great proposal. I guess the number of different "bundle name" will not be a huge number (something like 1-2 by application on average).

filters/openpolicyagent/openpolicyagent.go
szuecs: Beware that in some clusters we have routing changes  every 10s. If we need to survive datastructures from one routing tree to the other we need to use a registry and hand-over the object instances by PostProcessors.
Examples are scheduler filters "fifo", "lifo" and shedder "admissionControl".
Right now I am not sure if we need this here. I am just reading the change for the first time.

filters/openpolicyagent/serveresponsewithregopolicy/serveresponsewithregopolicy.go
```diff
@@ -0,0 +1,119 @@
...
+
+	if len(sargs) < 1 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	if len(sargs) > 2 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	bundleName := sargs[paramBundleName]
```
szuecs, line 51: The same here as in the other filter, do not use const but explicit 0 and 1 to index sargs.

filters/openpolicyagent/serveresponsewithregopolicy/serveresponsewithregopolicy.go
```diff
@@ -0,0 +1,119 @@
...
+
+	configOptions := s.configOpts
```
szuecs, line 53: maybe `options` (in other places we use `opt`, but maybe that's to close to `opa`), `configOpts` and `configOptions` seem to be too long as names.

filters/openpolicyagent/serveresponsewithregopolicy/serveresponsewithregopolicy.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	f.opa.Logger().WithFields(map[string]interface{}{
+		"query":               f.opa.EnvoyPluginConfig().ParsedQuery.String(),
+		"dry-run":             f.opa.EnvoyPluginConfig().DryRun,
+		"decision":            result.Decision,
+		"err":                 err,
+		"txn":                 result.TxnID,
+		"metrics":             result.Metrics.All(),
+		"total_decision_time": time.Since(start),
```
szuecs, line 107: same here as in the other filter

filters/openpolicyagent/serveresponsewithregopolicy/serveresponsewithregopolicy.go
```diff
@@ -0,0 +1,119 @@
...
+	}).Debug("Authorizing request with decision.")
+
+	if !f.opa.EnvoyPluginConfig().DryRun {
```
szuecs, line 110: if f.opa.EnvoyPluginConfig().DryRun {
   return
}
if err != nil {
...
} else {
..
}
```

File: config/config.go (hunk 3 above)
Comment (mjungsbluth) [reply config/config.go]: Moved it to auth

File: filters/openpolicyagent/internal/envoy/policytype.go (hunk 6 above)
Comment (mjungsbluth) [reply filters/openpolicyagent/internal/envoy/policytype.go]: I was able to remove this entire path and make it Skipper external config

File: filters/openpolicyagent/internal/envoy/envoyplugin.go (hunk 7 above)
Comment (mjungsbluth) [reply filters/openpolicyagent/internal/envoy/envoyplugin.go]: This funnily enough is actually a bug that also exists in the OPA Envoy plugin. It should actually never materialise but it would still be surprising behaviour

**Comment by szuecs** [comment]
@mjungsbluth from my first read of the change looks all more or less good. I have to understand more and re-read multiple times. So if you fixed all the comments, please tell, then I will have a second round.

File: filters/openpolicyagent/serveresponsewithregopolicy/serveresponsewithregopolicy.go (hunk 21 above)
Comment (mjungsbluth) [reply filters/openpolicyagent/serveresponsewithregopolicy/serveresponsewithregopolicy.go:48]: will just change it to `opts`, then it is consistent

File: filters/openpolicyagent/authorizewithregopolicy/authorizewithregopolicy.go (hunk 11 above)
Comment (mjungsbluth) [reply filters/openpolicyagent/authorizewithregopolicy/authorizewithregopolicy.go]: Typically you'd have one "bundle" per application or even coarser with a normal discovery setup (the bundle name would identify the discovery bundle that then calculates the list of actual bundles). So my guess is that you can end up with 10s per Skipper deployment

File: filters/openpolicyagent/openpolicyagent.go (hunk 16 above)
Comment (mjungsbluth) [reply filters/openpolicyagent/openpolicyagent.go]: Interesting. Right now it would download the bundles again on a new route. It is completely stateless but 10s is only one order of magnitude more than the bootstrap.

File: config/config.go [hunk 24]
```diff
@@ -477,6 +481,8 @@ func NewConfig() *Config {
 	flag.DurationVar(&cfg.OidcDistributedClaimsTimeout, "oidc-distributed-claims-timeout", 2*time.Second, "sets the default OIDC distributed claims request timeout duration to 2000ms")
 	flag.Var(cfg.CredentialPaths, "credentials-paths", "directories or files to watch for credentials to use by bearerinjector filter")
 	flag.DurationVar(&cfg.CredentialsUpdateInterval, "credentials-update-interval", 10*time.Minute, "sets the interval to update secrets")
+	flag.BoolVar(&cfg.EnableOpenPolicyAgent, "enable-open-policy-agent", false, "enables Open Policy Agent filters")
+	flag.StringVar(&cfg.OpenPolicyAgentConfigTemplate, "open-policy-agent-config-template", "", "file containing a template for an Open Policy Agent configuration file that is interpolated for each OPA filter instance")
```
szuecs, line 486: Why don't we have also OpenPolicyAgentEnvoyMetadata as flag added here?

docs/reference/filters.md
```diff
@@ -1695,6 +1695,130 @@ oidcClaimsQuery("/:name%\"*One\"", "/path:groups.#[%\"*-Test-Users\"] groups.#[=
...
+```
+
+Example (passing context):
+```
+serveResponseWithRegoPolicy("my-app-id", "com.mydomain.xxx.myprop: myvalue")
+```
+
+#### Data Flows
+
+## Diagram
```
szuecs, line 1796: `##` seems to be the wrong number of hashmarks.

docs/tutorials/auth.md
```diff
@@ -415,3 +415,90 @@ scope by doing the following:
...
+discovery:
+  name: discovery
+  prefix: "/applications/{{ .bundlename }}"
+```
+
+The variable `.bundlename` is the first argument in the following filters and can be in any format that OPA can understand, so for example application IDs from a registry, uuids, ...
+
+### Envoy Structures
+
+While Envoy is an alternative OSS product similar to Skipper, it has already defined structures for how external authorization should be done and also how  authorization decisions can influence the Envoy response. Open Policy Agent already has direct support for this and also commercial control planes support this. On top of this [examples and documentation](https://www.openpolicyagent.org/docs/latest/envoy-primer/) already exist.
```
szuecs, line 447: `how  authorization` -> `how authorization`

szuecs: Reading that everyone uses this data structure, maybe we should not name it "envoy", but have one reference that we use the same as envoy as everyone else. We don't really need to "pollute" the names with envoy all over the place?!

docs/tutorials/auth.md
```diff
@@ -415,3 +415,90 @@ scope by doing the following:
...
+    dry-run: false
+decision_logs:
+  console: true
+```
+
+Start Skipper with 
+
+```
+skipper --enable-open-policy-agent --open-policy-agent-config-template opaconfig.yaml \
+  --inline-routes 'notfound: * -> authorizeWithRegoPolicy("<playground-bundle-id>") -> inlineContent("<h1>Authorized Hello</h1>") -> <shunt>'
```
szuecs, line 496: I tried to run this example and I think something is wrong here.

```
% ./bin/skipper --enable-open-policy-agent --open-policy-agent-config-template opaconfig.yaml --inline-routes 'notfound: * -> authorizeWithRegoPolicy("9VuBTBib62") -> inlineContent("<h1>Authorized Hello</h1>") -> <shunt>'
..
[APP]INFO[0000] route settings, reset, route: notfound: * -> authorizeWithRegoPolicy("9VuBTBib62") -> inlineContent("<h1>Authorized Hello</h1>") -> <shunt>
[APP]ERRO[0000] failed to process route notfound: failed to create filter "authorizeWithRegoPolicy": open : no such file or directory
```

filters/openpolicyagent/authorizewithregopolicy/authorizewithregopolicy_test.go
```diff
@@ -0,0 +1,230 @@
...
+			assert.Equal(t, ti.expectedStatus, rsp.StatusCode, "HTTP status does not match")
+
+			assert.True(t, isHeadersPresent(ti.expectedHeaders, rsp.Header), "HTTP Headers do not match")
+
+			defer rsp.Body.Close()
+			body, err := io.ReadAll(rsp.Body)
+			if err != nil {
+				t.Error(err)
+			}
+			assert.Equal(t, ti.expectedBody, string(body), "HTTP Headers do not match")
```
szuecs, line 184: The message `"HTTP Headers do not match"` does not match the test case (body is checked here)

filters/openpolicyagent/authorizewithregopolicy/authorizewithregopolicy_test.go
```diff
@@ -0,0 +1,230 @@
...
+			}
+
+			assert.Equal(t, ti.expectedStatus, rsp.StatusCode, "HTTP status does not match")
+
+			assert.True(t, isHeadersPresent(ti.expectedHeaders, rsp.Header), "HTTP Headers do not match")
+
+			defer rsp.Body.Close()
+			body, err := io.ReadAll(rsp.Body)
+			if err != nil {
+				t.Error(err)
```
szuecs, line 182: In most cases you can use `t.Fatal` (fail quickly) instead of continue running the test with `t.Error`.

config/config.go
```diff
@@ -270,6 +270,10 @@ type Config struct {
 
 	LuaModules *listFlag `yaml:"lua-modules"`
 	LuaSources *listFlag `yaml:"lua-sources"`
+
+	EnableOpenPolicyAgent         bool   `yaml:"enable-open-policy-agent"`
+	OpenPolicyAgentConfigTemplate string `yaml:"open-policy-agent-config-template"`
+	OpenPolicyAgentEnvoyMetadata  string `yaml:"open-policy-agent-envoy-metadata"`
```
szuecs, line 276: I am not sure if we should leak implementation details (`envoy`) here

filters/openpolicyagent/evaluation.go
```diff
@@ -0,0 +1,76 @@
...
+		stopeval()
+		err := opa.logDecision(ctx, input, result, err)
+		if err != nil {
+			opa.Logger().WithFields(map[string]interface{}{"err": err}).Error("Unable to log decision to control plane.")
+		}
+	}()
+
+	if ctx.Err() != nil {
+		err = errors.Wrap(ctx.Err(), "check request timed out before query execution")
+		return nil, err
```
szuecs, line 43:
```
err = errors.Wrap(ctx.Err(), "check request timed out before query execution")
return nil, err
```
to
```
err = errors.Wrap(ctx.Err(), "check request timed out before query execution")
return nil, fmt.Errorf("check request timed out before query execution: %w", ctx.Err())
```

filters/openpolicyagent/evaluation.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/open-policy-agent/opa/server"
+	"github.com/open-policy-agent/opa/tracing"
+	"github.com/opentracing/opentracing-go"
+	"github.com/pkg/errors"
+)
+
+func (opa *OpenPolicyAgentInstance) Eval(ctx context.Context, req *ext_authz_v3.CheckRequest) (*envoyauth.EvalResult, error) {
+	var err error
```
szuecs, line 19: this you don't need anymore, because you use `:=` 2 lines below to have `err` defined

filters/openpolicyagent/evaluation.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	inputValue, err := ast.InterfaceToValue(input)
+	if err != nil {
+		return nil, err
+	}
+
+	err = envoyauth.Eval(ctx, opa, inputValue, result, rego.DistributedTracingOpts(tracing.Options{opa}))
```
szuecs, line 57: I see where it comes from but it reads a bit weird to do `envoyauth` and not `opaauth`. I guess we can rename the import, not sure if it makes sense to everyone to do my proposed change, but personally I would wonder why we do something envoy as it is also an http proxy and we do not want to run 2 proxies.

filters/openpolicyagent/internal/envoy/skipperadapter.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"net/http"
+	"strings"
+
+	ext_authz_v3_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	ext_authz_v3 "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
+)
+
+func AdaptToEnvoyExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metadata, contextExtensions map[string]string) *ext_authz_v3.CheckRequest {
```
szuecs, line 11: maybe rename to `AdaptExtAuthRequest`, because otherwise we duplicate `envoy` while reading the caller code.

filters/openpolicyagent/internal/envoy/skipperadapter.go
```diff
@@ -0,0 +1,34 @@
...
+
+	headers := make(map[string]string)
+	for h, vv := range req.Header {
+		// This makes headers in the input compatible with what Envoy does, i.e. allows to use policy fragments designed for envoy
+		// See: https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/header_casing#http-1-1-header-casing
+		headers[strings.ToLower(h)] = strings.Join(vv, ", ")
```
szuecs, line 17: Is it mandatory to copy all headers for something to work or is this nice to have?
For me this creates a lot of data by maybe even untrusted clients that could use it to create GC pressure.

I guess we need it because, but feels bad:
```
% go doc github.com/envoyproxy/go-control-plane/envoy/service/auth/v3.AttributeContext_HttpRequest.Headers
package authv3 // import "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"

type AttributeContext_HttpRequest struct {
    // The HTTP request headers. If multiple headers share the same key, they must
    // be merged according to the HTTP spec. All header keys must be lower-cased,
    // because HTTP header keys are case-insensitive.
    Headers map[string]string

    // ... other fields elided ...
}
```

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -0,0 +1,325 @@
...
+}
+
+func configTemplate() string {
+	if s, ok := os.LookupEnv("OPA_CONFIG_TEMPLATE_FILE"); ok {
+		return s
+	}
+	return "opaconfig.yaml"
+}
+
+func (factory *OpenPolicyAgentFactory) NewOpenPolicyAgentEnvoyInstance(bundleName string, config OpenPolicyAgentInstanceConfig, filterName string) (*OpenPolicyAgentInstance, error) {
```
szuecs, line 138: NewOpenPolicyAgentEnvoyInstance returns OpenPolicyAgentInstanceConfig, so drop also Envoy in this func makes sense

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -0,0 +1,325 @@
...
+
+func (opa *OpenPolicyAgentInstance) ParsedQuery() ast.Body {
+	return opa.EnvoyPluginConfig().ParsedQuery
+}
+
+func (opa *OpenPolicyAgentInstance) Store() storage.Store            { return opa.manager.Store }
+func (opa *OpenPolicyAgentInstance) Compiler() *ast.Compiler         { return opa.manager.GetCompiler() }
+func (opa *OpenPolicyAgentInstance) Runtime() *ast.Term              { return opa.manager.Info }
+func (opa *OpenPolicyAgentInstance) Logger() logging.Logger          { return opa.manager.Logger() }
+func (opa *OpenPolicyAgentInstance) PreparedQueryDoOnce() *sync.Once { return opa.preparedQueryDoOnce }
```
szuecs, line 315: I wonder who is calling this method. I don't find it in the PR code.

filters/openpolicyagent/response.go
```diff
@@ -0,0 +1,76 @@
...
+	"io"
+
+	"net/http"
+
+	"github.com/open-policy-agent/opa-envoy-plugin/envoyauth"
+	"github.com/opentracing/opentracing-go"
+	"github.com/zalando/skipper/filters"
+)
+
+func (opa *OpenPolicyAgentInstance) RejectInvalidDecisionError(fc filters.FilterContext, span opentracing.Span, result *envoyauth.EvalResult, err error) {
```
szuecs, line 14: You log errors in the span, but the span is not marked as error via `span.SetTag("error", true)`, which is likely how people want to filter their streams to observe reject rates.

filters/openpolicyagent/authorizewithregopolicy/authorizewithregopolicy.go
szuecs: Seems to be good to go

filters/openpolicyagent/openpolicyagent.go
szuecs: Sounds like we want to have a registry. If we would have 40k routes in a cluster and do this every 10s, it will be a bit huge.

szuecs:
@mjungsbluth thanks great changes, reads much better the second time. I added a couple of more comments.

config/config.go
mjungsbluth: That was an oversight.

config/config.go
mjungsbluth: Your call, we can strip `envoy`. the format however is specific and needs to follow the structure of https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#config-core-v3-metadata

docs/reference/filters.md
mjungsbluth: Removed it...

docs/tutorials/auth.md
mjungsbluth: Agreed. The text was for a different audience, will shorten it to one sentence that references the existing docs.

filters/openpolicyagent/openpolicyagent.go
mjungsbluth: These methods implement the envoyauth.EvalContext interface that is passed to the Eval method. I can also create an extra struct or am open to other ideas to make it more readable

filters/openpolicyagent/evaluation.go
mjungsbluth: We could also use `opaenvoyplugin.Eval`. `opaauth` might be confusing as well as this is not from the core Go library but then you still can just navigate to the actual target.

docs/tutorials/auth.md
mjungsbluth: Yes, I broke it apparently with the last change to extract the policy type. When you try running it again, also try enabling debug logging. I re-routed INFO logging by OPA to DEBUG as it is very chatty.

filters/openpolicyagent/internal/envoy/skipperadapter.go
mjungsbluth: To support the http spec fully and support repeated headers in policies, I fear we need it.

filters/openpolicyagent/response.go
mjungsbluth: Good catch! We still have to run Skipper with enabled tracing to capture how practical the traces are.

mjungsbluth:
@szuecs could you have another look? I feel there are only a few remaining items. The ops documentation for tracing and metrics is also added and tracing has been validated using lightstep.

PR #2365, opened 2023-06-01
fix: enable only lb healthchecker if not nil

szuecs:
active health checks would be done if we would add members to check which we do not do, but we also do not need to start the loop, in case we have no lb created


AlexanderYastrebov:
:+1:

MustafaSaber:
👍

PR #2370, opened 2023-06-05
test: use and clean up temporary etcd data directory

chahn:
etcd stores its data in the `data-dir`. If this directory is not explicitly specified as a parameter, etcd creates a directory `default.etcd` in the directory where the etcd process is started.

When the Skipper etcd tests begin, the directory `default.etcd` is created by etcd for storing its data but it is not deleted after the etcd process is terminated.
During subsequent tests in the same environment, Skipper would receive routes from previous tests because of the preserved data in etcd and by calling `etcdGetUpdates()` (using `etcdIndex` and `waitIndex`).
This causes the tests to behave unexpectedly.

This patch creates a temporary `data-dir` before starting etcd and makes sure that it is deleted after the tests.
This way the tests are always running with an empty etcd in a defined state.

### DIFF
--- a/etcd/etcdtest/mocketcd.go
+++ b/etcd/etcdtest/mocketcd.go
@@ -24,6 +24,7 @@ import (
 var Urls []string
 
 var etcd *exec.Cmd
+var etcdDataDir string
 
 func makeLocalUrls(ports ...int) []string {
 	urls := make([]string, len(ports))
@@ -55,6 +56,12 @@ func StartProjectRoot(projectRoot string) error {
 	Urls = makeLocalUrls(randPort(), randPort())
 	clientUrlsString := strings.Join(Urls, ",")
 
+	dir, err := os.MkdirTemp("", "etcdtest")
+	if err != nil {
+		return err
+	}
+	etcdDataDir = dir
+
 	var binary string
 	if projectRoot != "" {
 		binary = filepath.Join(projectRoot, ".bin/etcd")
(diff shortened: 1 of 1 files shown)


etcd/etcdtest/mocketcd.go
```diff
@@ -38,6 +39,15 @@ func randPort() int {
...
 }
 
+// Create a temporary directory for etcd's data-dir.
+func makeTempDir() (string, error) {
+	dir, err := os.MkdirTemp("", "etcdtest")
+	if err != nil {
+		return "", err
+	}
+	return dir, nil
+}
```
AlexanderYastrebov, line 49: This is for single use - lets just inline it.

etcd/etcdtest/mocketcd.go
```diff
@@ -117,6 +135,12 @@ func Stop() error {
 		return nil
 	}
 
+	// Ensure that etcd's data-dir (etcdDataDir) is deleted even if killing the process fails.
+	defer func() {
+		os.RemoveAll(etcdDataDir)
+		etcdDataDir = "" // Clear the dataDir variable.
```
AlexanderYastrebov, line 141: IMO this and other comments are redundant as it is quite clear what is going on.

AlexanderYastrebov:
Thank you, it looks good.

AlexanderYastrebov:
:+1:

etcd/etcdtest/mocketcd.go
chahn: I stripped the comments as supposed.

etcd/etcdtest/mocketcd.go
chahn: It is inline now with the latest commit.
Thanks for the feedback @AlexanderYastrebov

MustafaSaber:
👍

chahn:
Thanks for merging and especially for your support @AlexanderYastrebov and @MustafaSaber

PR #2373, opened 2023-06-06
update: github.com/rcrowley/go-metrics - codahale

szuecs:
update: github.com/rcrowley/go-metrics - codahale
test: fix metrics tests to use separate created metrics instead relying on implicitly defined runtime metrics, which are dropped by sync.Once package global in https://github.com/rcrowley/go-metrics/commit/615496d9b7277a58d61fdeb8f9a3d4185fcf1bb2


szuecs:
:+1:

PR #2374, opened 2023-06-06
optimize: no goroutine on metric updates

szuecs:
optimize: no goroutine on metric updates, because it creates only overhead for the scheduler and GC


szuecs:
:+1:

metrics/codahale.go
```diff
@@ -121,8 +121,7 @@ func (c *CodaHale) IncFloatCounterBy(key string, value float64) {
 }
 
 func (c *CodaHale) measureSince(key string, start time.Time) {
-	d := time.Since(start)
-	go c.updateTimer(key, d)
+	c.updateTimer(key, time.Since(start))
```
AlexanderYastrebov, line 124: I see both StandardRegistry.GetOrRegister and StandardTimer.Update use sync.Mutex.
Ideally we should benchmark if this change makes things worse or maybe leave it like it is now

metrics/codahale_test.go
```diff
@@ -516,3 +516,26 @@ func TestCodaHaleServeMetrics(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkMeasureSince(b *testing.B) {
+	c := NewCodaHale(Options{})
+	now := time.Now()
+
+	b.Run("measureSince", func(b *testing.B) {
```
AlexanderYastrebov, line 524: Why do we need to use sub benchmark here instead of just having the loop directly?

szuecs: I can drop it if you like but I don't see that this is bad.

AlexanderYastrebov: It is not wrong, just not clear why

PR #2375, opened 2023-06-06
predicates/traffic: add TrafficSegment route test

AlexanderYastrebov:
TrafficSegment predicate has a weight of -1 and thus does not affect final route weight.

This test verifies that requests with a specific header do not match routes with TrafficSegment despite the same number of predicates.


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2381, opened 2023-06-08
Refactor load balancer, decouple it from fadeIn

RomanZavodskikh:
Updates #2346

+ Added sync.Mutex to routing.Route
+ Created HealthyEndpoint* fields in routing.Routing
+ Refactor load balancer, decouple it from fadeIn
+ Updated the code for CI tests to work
+ Fixed the searchRing function for consistentHash

### DIFF
--- a/filters/builtin/creationmetrics_test.go
+++ b/filters/builtin/creationmetrics_test.go
@@ -75,17 +75,17 @@ func TestRouteCreationMetrics_reportRouteCreationTimes(t *testing.T) {
 func TestRouteCreationMetrics_startTimes(t *testing.T) {
 	for _, tt := range []struct {
 		name     string
-		route    routing.Route
+		route    *routing.Route
 		expected map[string]time.Time
 	}{
 		{
 			name:     "no start time provided",
-			route:    routing.Route{},
+			route:    &routing.Route{},
 			expected: map[string]time.Time{},
 		},
 		{
 			name: "start time from origin marker",
-			route: routing.Route{Filters: []*routing.RouteFilter{
+			route: &routing.Route{Filters: []*routing.RouteFilter{
 				{Filter: &OriginMarker{Origin: "origin", Id: "config0", Created: time0}},
 				{Filter: &OriginMarker{Origin: "origin", Id: "config1", Created: time1}},
 			}},
@@ -94,9 +94,9 @@ func TestRouteCreationMetrics_startTimes(t *testing.T) {
 	} {
 		t.Run(tt.name, func(t *testing.T) {
 			metrics := NewRouteCreationMetrics(&metricstest.MockMetrics{})
-			assert.Equal(t, tt.expected, metrics.startTimes(&tt.route))
+			assert.Equal(t, tt.expected, metrics.startTimes(tt.route))
 			//should be cached
-			assert.Empty(t, metrics.startTimes(&tt.route))
+			assert.Empty(t, metrics.startTimes(tt.route))
 		})
 	}
 }
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -244,15 +244,16 @@ func newPowerOfRandomNChoices([]string) routing.LBAlgorithm {
 
 // Apply implements routing.LBAlgorithm with power of random N choices algorithm.
 func (p *powerOfRandomNChoices) Apply(ctx *routing.LBContext) routing.LBEndpoint {
-	ne := len(ctx.Route.LBEndpoints)
+	endpoints := ctx.Route.GetHealthyEndpoints()
+	ne := len(endpoints)
 
 	p.mx.Lock()
 	defer p.mx.Unlock()
 
-	best := ctx.Route.LBEndpoints[p.rnd.Intn(ne)]
+	best := endpoints[p.rnd.Intn(ne)]
 
 	for i := 1; i < p.numberOfChoices; i++ {
-		ce := ctx.Route.LBEndpoints[p.rnd.Intn(ne)]
+		ce := endpoints[p.rnd.Intn(ne)]
 
 		if p.getScore(ce) > p.getScore(best) {
 			best = ce
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -3,13 +3,10 @@ package loadbalancer
 import (
 	"errors"
 	"fmt"
-	"math"
 	"math/rand"
 	"net/url"
 	"sort"
 	"sync"
-	"sync/atomic"
-	"time"
 
 	"github.com/cespare/xxhash/v2"
 	log "github.com/sirupsen/logrus"
@@ -54,97 +51,8 @@ var (
 	defaultAlgorithm = newRoundRobin
 )
 
-func fadeInState(now time.Time, duration time.Duration, detected time.Time) (time.Duration, bool) {
-	rel := now.Sub(detected)
-	return rel, rel > 0 && rel < duration
-}
-
-func fadeIn(now time.Time, duration time.Duration, exponent float64, detected time.Time) float64 {
-	rel, fadingIn := fadeInState(now, duration, detected)
-	if !fadingIn {
-		return 1
-	}
-
-	return math.Pow(float64(rel)/float64(duration), exponent)
-}
-
-func shiftWeighted(rnd *rand.Rand, ctx *routing.LBContext, now time.Time) routing.LBEndpoint {
-	var sum float64
-	rt := ctx.Route
-	ep := ctx.Route.LBEndpoints
-	for _, epi := range ep {
-		wi := fadeIn(now, rt.LBFadeInDuration, rt.LBFadeInExponent, epi.Detected)
-		sum += wi
-	}
-
-	choice := ep[len(ep)-1]
-	r := rnd.Float64() * sum
-	var upto float64
-	for i, epi := range ep {
-		upto += fadeIn(now, rt.LBFadeInDuration, rt.LBFadeInExponent, epi.Detected)
-		if upto > r {
-			choice = ep[i]
-			break
-		}
-	}
-
-	return choice
-}
-
-func shiftToRemaining(rnd *rand.Rand, ctx *routing.LBContext, wi []int, now time.Time) routing.LBEndpoint {
-	notFadingIndexes := wi
-	ep := ctx.Route.LBEndpoints
-
-	// if all endpoints are fading, the simplest approach is to use the oldest,
-	// this departs from the desired curve, but guarantees monotonic fade-in. From
-	// the perspective of the oldest endpoint, this is temporarily the same as if
-	// there was no fade-in.
-	if len(notFadingIndexes) == 0 {
-		return shiftWeighted(rnd, ctx, now)
-	}
-
-	// otherwise equally distribute between the old endpoints
-	return ep[notFadingIndexes[rnd.Intn(len(notFadingIndexes))]]
-}
-
-func withFadeIn(rnd *rand.Rand, ctx *routing.LBContext, choice int, algo routing.LBAlgorithm) routing.LBEndpoint {
-	ep := ctx.Route.LBEndpoints
-	now := time.Now()
-	f := fadeIn(
-		now,
-		ctx.Route.LBFadeInDuration,
-		ctx.Route.LBFadeInExponent,
-		ctx.Route.LBEndpoints[choice].Detected,
-	)
-
-	if rnd.Float64() < f {
-		return ep[choice]
-	}
-	notFadingIndexes := make([]int, 0, len(ep))
-	for i := 0; i < len(ep); i++ {
-		if _, fadingIn := fadeInState(now, ctx.Route.LBFadeInDuration, ep[i].Detected); !fadingIn {
-			notFadingIndexes = append(notFadingIndexes, i)
-		}
-	}
-
-	switch a := algo.(type) {
-	case *roundRobin:
-		return shiftToRemaining(a.rnd, ctx, notFadingIndexes, now)
-	case *random:
-		return shiftToRemaining(a.rnd, ctx, notFadingIndexes, now)
-	case *consistentHash:
-		// If all endpoints are fading, normal consistent hash result
-		if len(notFadingIndexes) == 0 {
-			return ep[choice]
-		}
-		// otherwise calculate consistent hash again using endpoints which are not fading
-		return ep[a.chooseConsistentHashEndpoint(ctx, skipFadingEndpoints(notFadingIndexes))]
-	default:
-		return ep[choice]
-	}
-}
-
 type roundRobin struct {
+	mx    sync.Mutex
 	index int64
 	rnd   *rand.Rand
 }
@@ -163,13 +71,13 @@ func (r *roundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 		return ctx.Route.LBEndpoints[0]
 	}
 
-	index := int(atomic.AddInt64(&r.index, 1) % int64(len(ctx.Route.LBEndpoints)))
+	endpoints := ctx.Route.GetHealthyEndpoints()
 
-	if ctx.Route.LBFadeInDuration <= 0 {
-		return ctx.Route.LBEndpoints[index]
-	}
+	r.mx.Lock()
+	defer r.mx.Unlock()
 
-	return withFadeIn(r.rnd, ctx, index, r)
+	r.index = (r.index + 1) % int64(len(endpoints))
+	return endpoints[r.index]
 }
 
 type random struct {
@@ -189,12 +97,10 @@ func (r *random) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 		return ctx.Route.LBEndpoints[0]
 	}
 
-	i := r.rnd.Intn(len(ctx.Route.LBEndpoints))
-	if ctx.Route.LBFadeInDuration <= 0 {
-		return ctx.Route.LBEndpoints[i]
-	}
+	endpoints := ctx.Route.GetHealthyEndpoints()
+	i := r.rnd.Intn(len(endpoints))
 
-	return withFadeIn(r.rnd, ctx, i, r)
+	return endpoints[i]
 }
 
 type (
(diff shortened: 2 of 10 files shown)


loadbalancer/algorithm.go
```diff
@@ -170,30 +102,31 @@ func (r *roundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
-	if ctx.Route.LBFadeInDuration <= 0 {
-		return ctx.Route.LBEndpoints[r.index]
+	endpoints := []routing.LBEndpoint{}
+	for _, e := range ctx.Route.LBEndpoints {
+		e := e
+		if goodEndpoint(r.rnd, &e, ctx) {
+			endpoints = append(endpoints, e)
+		}
+	}
+	if len(endpoints) == 0 {
```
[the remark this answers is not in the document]
szuecs: Some load balancers (ALB, NLB) choose to send traffic to all endpoints

RomanZavodskikh:
Added some code trying to avoid races.

routing/datasource.go
```diff
@@ -526,6 +527,42 @@ type routeTable struct {
 	validRoutes   []*eskip.Route
 	invalidRoutes []*eskip.Route
 	created       time.Time
+	closeCh       chan struct{}
```
szuecs, line 530: normally that's called `quit`

routing/datasource.go
```diff
@@ -526,6 +527,42 @@ type routeTable struct {
...
+}
+
+func (rt *routeTable) goodEndpointerGoroutine() {
+	timer := time.NewTicker(1 * time.Millisecond)
```
szuecs, line 534: that's a bit too often, maybe every second is ok

routing/routing.go
```diff
@@ -202,6 +203,7 @@ func NewLBContext(r *http.Request, rt *Route) *LBContext {
 
 // Route object with preprocessed filter instances.
 type Route struct {
+	mx sync.Mutex
```
szuecs, line 206: mx -> mu

routing/routing.go
```diff
@@ -361,6 +377,12 @@ func (r *Routing) startReceivingUpdates(o Options) {
 		for {
 			select {
 			case rt := <-c:
+				var rtToClose *routeTable
+				rtToClose, ok := r.routeTable.Load().(*routeTable)
+				if ok {
+					rtToClose.close()
```
szuecs, line 383: pretty sure this leads to a lot of 404. If you need to close it then after the Store below.

szuecs:
@RomanZavodskikh I think the only place to start a goroutine is the startReceivingUpdates() func in routing.go.

RomanZavodskikh:
Decided not to close `routeTable` object in the startReceivingUpdates goroutine. Instead I decided to create quit channel for `goodEndpointerGoroutine` instead.

loadbalancer/algorithm.go
```diff
@@ -217,9 +123,9 @@ type (
 		hash  uint64 // hash of endpoint
 	}
 	consistentHash struct {
-		hashRing         []endpointHash // list of endpoints sorted by hash value
-		rand             *rand.Rand
-		notFadingIndexes []int
+		hashRing []endpointHash // list of endpoints sorted by hash value
+		rand     *rand.Rand
+		allHosts map[string]struct{}
```
RomanZavodskikh, line 128: I can safely remove this value.

RomanZavodskikh:
Made `consistentHash.Apply` work in O(1) time.

loadbalancer/algorithm.go
```diff
@@ -53,162 +52,69 @@ var (
...
-
-		// preallocating frequently used slice
-		notFadingIndexes: make([]int, 0, len(endpoints)),
-		fadingWeights:    make([]float64, 0, len(endpoints)),
 	}
 }
 
 // Apply implements routing.LBAlgorithm with a roundrobin algorithm.
 func (r *roundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
+	ctx.Route.Lock()
```
RomanZavodskikh, line 71: Maybe I can remove this Lock after first if block?

loadbalancer/algorithm_test.go
```diff
@@ -313,6 +314,16 @@ func TestConsistentHashBoundedLoadSearch(t *testing.T) {
...
 		},
 	}})[0]
+	route.GoodHosts = map[string]struct{}{}
+	for i := range endpoints {
+		route.LBEndpoints = append(route.LBEndpoints, routing.LBEndpoint{
+			Host:     endpoints[i],
+			Detected: time.Now().Add(-3 * time.Minute),
+			Metrics:  &routing.LBMetrics{},
+		})
+		route.GoodHosts[endpoints[i][7:]] = struct{}{}
```
RomanZavodskikh, line 324: [7:] here needed to remove "http://" from the beginning.

loadbalancer/fadein_test.go
```diff
@@ -85,6 +98,7 @@ func testFadeIn(
 		func() {
 			for {
 				ctx.Params[ConsistentHashKey] = strconv.Itoa(rnd.Intn(100000))
+				tick <- time.Now()
```
szuecs, line 101: do we use this value?

routing/datasource.go
```diff
@@ -528,6 +531,69 @@ type routeTable struct {
...
+}
+
+func GoodEndpointerGoroutine(routes []*Route, quit <-chan struct{}, tick <-chan time.Time) {
+	rnd := rand.New(rand.NewSource(time.Now().UnixNano())) // #nosec
+
+	for {
+		select {
+		case <-quit:
+			return
+		case <-tick:
```
szuecs, line 568: Here we do not use time.Now() value that we sent via this chan.

routing/routing.go
```diff
@@ -369,8 +386,17 @@ func (r *Routing) startReceivingUpdates(o Options) {
 						r.firstLoadSignaled = true
 					}
 				}
+
+				close(goodEndpointerQuit)
+				timer.Stop()
+				goodEndpointerQuit = make(chan struct{})
+				timer = time.NewTicker(500 * time.Millisecond)
+				go GoodEndpointerGoroutine(rt.routes, goodEndpointerQuit, timer.C)
```
szuecs, line 394: this code block feels if we could make it better, but I have no idea how. On the other hand it's only run every time on routing tree change, so maybe it's fine. I am a bit worried about high traffic while updating this here.
@AlexanderYastrebov wdyt?

loadbalancer/algorithm.go
```diff
@@ -168,32 +72,30 @@ func (r *roundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 		return ctx.Route.LBEndpoints[0]
 	}
 
+	ctx.Route.Lock()
+	defer ctx.Route.Unlock()
```
szuecs, line 76: Here I am a bit worried about the 2-phase lock at high traffic routes.
What do you think @AlexanderYastrebov ?

szuecs:
The code looks good to me some worries, but it's more feeling than anything else, so data will convince me.

loadbalancer/algorithm.go
AlexanderYastrebov: Here we lock only to check whether ctx.Route.GoodEndpoints is empty.
Maybe we should juts have 
```go
func (r *Route) GetHealthyEndpoints() []LBEndpoint
```
that would take care of proper locking internally.
It may also guarantee to never return an empty list.

loadbalancer/algorithm.go
```diff
@@ -168,32 +72,30 @@ func (r *roundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
+
 	r.mx.Lock()
 	defer r.mx.Unlock()
-	r.index = (r.index + 1) % len(ctx.Route.LBEndpoints)
 
-	if ctx.Route.LBFadeInDuration <= 0 {
-		return ctx.Route.LBEndpoints[r.index]
+	endpoints := ctx.Route.GoodEndpoints
```
AlexanderYastrebov, line 81: With this refactoring the list of endpoints is variable between calls to Apply.
We need to make sure that algorithms continue to behave as expected (esp. consitent hash that uses pre-calculated datastructure based on the full list of endpoints).

loadbalancer/algorithm.go
```diff
@@ -203,12 +105,16 @@ func (r *random) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
-	if ctx.Route.LBFadeInDuration <= 0 {
-		return ctx.Route.LBEndpoints[i]
+	ctx.Route.Lock()
+	defer ctx.Route.Unlock()
+
+	endpoints := ctx.Route.GoodEndpoints
+	if len(endpoints) == 0 {
+		endpoints = ctx.Route.LBEndpoints
 	}
+	i := r.rand.Intn(len(endpoints))
```
AlexanderYastrebov, line 115: Here we miss locking as rand.Intn is not thread safe.

loadbalancer/algorithm.go
```diff
@@ -203,12 +105,16 @@ func (r *random) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 		return ctx.Route.LBEndpoints[0]
 	}
 
-	i := r.rand.Intn(len(ctx.Route.LBEndpoints))
-	if ctx.Route.LBFadeInDuration <= 0 {
-		return ctx.Route.LBEndpoints[i]
+	ctx.Route.Lock()
```
AlexanderYastrebov, line 108: https://github.com/zalando/skipper/pull/2381/files#r1235283494

loadbalancer/algorithm.go
```diff
@@ -217,9 +123,8 @@ type (
 		hash  uint64 // hash of endpoint
 	}
 	consistentHash struct {
-		hashRing         []endpointHash // list of endpoints sorted by hash value
-		rand             *rand.Rand
-		notFadingIndexes []int
+		hashRing []endpointHash // list of endpoints sorted by hash value
+		rand     *rand.Rand
```
AlexanderYastrebov, line 127: https://github.com/zalando/skipper/pull/2381/files#r1235311364

loadbalancer/algorithm.go
```diff
@@ -309,13 +213,14 @@ func (ch *consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
-
-	if ctx.Route.LBFadeInDuration <= 0 {
-		return ctx.Route.LBEndpoints[choice]
-	}
+	ctx.Route.Lock()
+	defer ctx.Route.Unlock()
 
-	return withFadeIn(ch.rand, ctx, ch.notFadingIndexes, choice, ch)
+	choice := ch.chooseConsistentHashEndpoint(ctx, func(i int) bool {
+		_, ok := ctx.Route.GoodHosts[ctx.Route.LBEndpoints[i].Host]
```
AlexanderYastrebov, line 220: Here is it not clear why we use Host instead of treating endpoint string as opaque value.

routing/datasource.go
```diff
@@ -528,6 +531,69 @@ type routeTable struct {
 	created       time.Time
 }
 
+func fadeInState(now time.Time, duration time.Duration, detected time.Time) (time.Duration, bool) {
```
AlexanderYastrebov, line 534: Maybe put these functions into a separate file

routing/datasource.go
```diff
@@ -528,6 +531,69 @@ type routeTable struct {
...
+	f := fadeIn(
+		now,
+		route.LBFadeInDuration,
+		route.LBFadeInExponent,
+		ep.Detected,
+	)
+	goodFadeIn := f == 1 || rnd.Float64() < f
+
+	return goodFadeIn
+}
```
AlexanderYastrebov, line 559: This should also work for routes that do not use fadeIn.
Note that pre-existing code contains check like this:
```go
if ctx.Route.LBFadeInDuration <= 0 {
  return ctx.Route.LBEndpoints[i]
}
```

routing/routing.go
AlexanderYastrebov, line 394: Do we need to run a goroutine for each table update or maybe we can have one routine running for the whole lifetime of the datasource?

routing/routing.go
```diff
@@ -369,8 +386,17 @@ func (r *Routing) startReceivingUpdates(o Options) {
 						r.firstLoadSignaled = true
 					}
 				}
+
+				close(goodEndpointerQuit)
+				timer.Stop()
+				goodEndpointerQuit = make(chan struct{})
+				timer = time.NewTicker(500 * time.Millisecond)
```
AlexanderYastrebov, line 393: Why do we need to pass timer to GoodEndpointerGoroutine, i.e. why not create it inside?

AlexanderYastrebov: Why is GoodEndpointerGoroutine exported?

routing/routing.go
AlexanderYastrebov: If we create a single long-running routine it could also reuse r.quit channel.

routing/datasource.go
```diff
@@ -528,6 +531,69 @@ type routeTable struct {
...
+		case <-quit:
+			return
+		case <-tick:
+			for _, ri := range routes {
+				if ri.Route.BackendType != eskip.LBBackend {
+					continue
+				}
+
+				if len(ri.Route.LBEndpoints) == 0 {
+					log.Errorf("failed to post-process LB route: %s, no endpoints defined", ri.Id)
```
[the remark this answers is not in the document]
AlexanderYastrebov: Check if there is an invariant (established outside) that for loadbalanced backend list of endpoints is not empty.

routing/datasource.go
```diff
@@ -528,6 +531,69 @@ type routeTable struct {
...
+					continue
+				}
+
+				newGoodEndpoints := []LBEndpoint{}
```
AlexanderYastrebov, line 579: Makes sense to allocate enough memory right away
`make([]LBEndpoint, len(ri.LBEndpoints))`

routing/datasource.go
```diff
@@ -528,6 +531,69 @@ type routeTable struct {
...
+				for i := range ri.LBEndpoints {
+					if goodEndpoint(rnd, &ri.LBEndpoints[i], ri) {
+						newGoodEndpoints = append(newGoodEndpoints, ri.LBEndpoints[i])
+						newGoodHosts[ri.LBEndpoints[i].Host] = struct{}{}
+					}
+				}
+
+				ri.Lock()
+				ri.GoodEndpoints = newGoodEndpoints
+				ri.GoodHosts = newGoodHosts
```
AlexanderYastrebov, line 590: Lets check if we really need it.

AlexanderYastrebov: In any case we do not need it for any algorithm other than consistentHash

szuecs: Maybe the intention is to have this for https://github.com/zalando/skipper/issues/2346

loadbalancer/algorithm.go
RomanZavodskikh: https://github.com/zalando/skipper/pull/2412
I make `rand.Intn` thread safe in this PR.

routing/datasource.go
RomanZavodskikh: `fadeIn` returns 1 when endpoint already passed "fading-in" state.

routing/datasource.go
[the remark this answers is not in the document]
szuecs: I think it's fine to log this error, because in case Kubernetes dataclient sees "0" endpoints, it will rewrite the route to a `<shunt>` route.

routing/datasource.go
szuecs: or ` make([]LBEndpoint, 0, len(ri.LBEndpoints))`  in case you use `append()` (read cap vs len)

routing/datasource.go
RomanZavodskikh: That's an intention to make consistentHash.Apply to work in O(1). I fill up this map(set) here, but lookup in this map is O(1) regardless of its size.

szuecs: Not reviewed again the code but make sure the data structure is long living and only created at routing tree create/update

filters/builtin/creationmetrics_test.go
```diff
@@ -75,17 +75,17 @@ func TestRouteCreationMetrics_reportRouteCreationTimes(t *testing.T) {
 func TestRouteCreationMetrics_startTimes(t *testing.T) {
 	for _, tt := range []struct {
 		name     string
-		route    routing.Route
+		route    *routing.Route
```
RomanZavodskikh, line 78: Maybe pointer to mutex instead of mutex?

loadbalancer/algorithm.go
```diff
@@ -163,13 +73,18 @@ func (r *roundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 		return ctx.Route.LBEndpoints[0]
 	}
 
-	index := int(atomic.AddInt64(&r.index, 1) % int64(len(ctx.Route.LBEndpoints)))
+	endpoints := ctx.Route.GetHealthyEndpoints()
 
-	if ctx.Route.LBFadeInDuration <= 0 {
-		return ctx.Route.LBEndpoints[index]
+	r.mx.Lock()
+	if len(endpoints) > int(r.oldEpsLen) {
```
RomanZavodskikh, line 79: This code won't work in case of fadeIn.

RomanZavodskikh, line 79: Here should be the simplest option.

loadbalancer/algorithm.go
```diff
@@ -293,13 +209,13 @@ func (ch *consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
...
-		return ctx.Route.LBEndpoints[choice]
-	}
+	endpointsSet := ctx.Route.GetHealthyEndpointsSet()
 
-	return withFadeIn(ch.rnd, ctx, choice, ch)
+	choice := ch.chooseConsistentHashEndpoint(ctx, func(i int) bool {
+		_, ok := endpointsSet[ctx.Route.LBEndpoints[i]]
+		return !ok
+	})
+	return ctx.Route.LBEndpoints[choice]
```
RomanZavodskikh, line 218: return ch.Choose(endpointsSet)

routing/routing.go
```diff
@@ -246,6 +254,30 @@ type Route struct {
 	LBFadeInExponent float64
 }
 
+func (r *Route) GetHealthyEndpoints() []LBEndpoint {
+	r.mx.Lock()
+	defer r.mx.Unlock()
+
+	if len(r.HealthyEndpoints) == 0 {
```
RomanZavodskikh, line 261: Let's initialize in constructor/updateHealthyEndpoints.

routing/datasource.go
```diff
@@ -528,6 +530,67 @@ type routeTable struct {
...
+				if ri.Route.BackendType != eskip.LBBackend {
+					continue
+				}
+
+				if len(ri.Route.LBEndpoints) == 0 {
+					log.Errorf("failed to post-process LB route: %s, no endpoints defined", ri.Id)
+					continue
+				}
+
+				newHealthyEndpoints := make([]LBEndpoint, 0, len(ri.LBEndpoints))
```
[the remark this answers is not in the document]
RomanZavodskikh: The slice like this is already defined as []LBEndpoint in the skipper code. I think it makes sense to comply to this decision and **not to** make this variable []*LBEndpoint.
https://github.com/zalando/skipper/blob/v0.17.8/routing/routing.go#L229

routing/datasource.go
```diff
@@ -528,6 +530,67 @@ type routeTable struct {
...
+				newHealthyEndpointsSet := make(map[LBEndpoint]struct{}, len(ri.LBEndpoints))
```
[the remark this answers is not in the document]
szuecs: as map key you need a ptr

RomanZavodskikh:
Will do this work here https://github.com/zalando/skipper/pull/2634 .
Here unfortunately I've taken too much work I should've split it before creating this huge PR. So I did in https://github.com/zalando/skipper/pull/2634 and other PRs.

PR #2384, opened 2023-06-09
dataclients/kubernetes: refactor Ingress and RouteGroup backend weight calculation

AlexanderYastrebov:
* use the same algorithm to calculate arguments for Traffic() predicate for both Ingress and RouteGroup
* add GetBackendTrafficCalculator factory method to support alternative implementations

For #2268

### DIFF
--- a/dataclients/kubernetes/ingress.go
+++ b/dataclients/kubernetes/ingress.go
@@ -13,7 +13,6 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/zalando/skipper/dataclients/kubernetes/definitions"
 	"github.com/zalando/skipper/eskip"
-	"github.com/zalando/skipper/predicates"
 	"github.com/zalando/skipper/secrets/certregistry"
 )
 
@@ -46,6 +45,7 @@ type ingressContext struct {
 	hostRoutes          map[string][]*eskip.Route
 	defaultFilters      defaultFilters
 	certificateRegistry *certregistry.CertRegistry
+	calculateTraffic    func([]*weightedIngressBackend) map[string]backendTraffic
 }
 
 type ingress struct {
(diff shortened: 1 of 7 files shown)


dataclients/kubernetes/ingress.go
```diff
@@ -192,7 +175,7 @@ func applyAnnotationPredicates(m PathMode, r *eskip.Route, annotation string) er
 	return nil
 }
 
-func addExtraRoutes(ic ingressContext, ruleHost, path, pathType, eastWestDomain string, enableEastWest bool) {
+func addExtraRoutes(ic *ingressContext, ruleHost, path, pathType, eastWestDomain string, enableEastWest bool) {
```
AlexanderYastrebov, line 178: Use pointer here and elsewhere to avoid needless copying

dataclients/kubernetes/traffic.go
```diff
@@ -0,0 +1,122 @@
...
+
+// backendTraffic specifies whether a given backend is allowed to receive any traffic and
+// modifies route to receive the desired traffic portion
+type backendTraffic interface {
+	allowed() bool
+	apply(*eskip.Route)
+}
+
+// GetBackendTrafficCalculator returns a function that calculates backendTraffic for each backend using specified algorithm
+func GetBackendTrafficCalculator[T definitions.WeightedBackend](algorithm string) func(b []T) map[string]backendTraffic {
```
AlexanderYastrebov, line 17: This factory function is generic to support both ingress and routegroup weighted backends and avoid copying slices (see https://go.dev/doc/faq#convert_slice_of_interface and https://stackoverflow.com/questions/12753805/type-converting-slices-of-interfaces)

Alternatively it could use RouteGroup's `*BackendReference` but:
* it is defined for routegroup but would be used by ingress as well
* `BackendReference.Weight` int type needs to be changed to float64 along with additional validation that user-provided values are whole numbers

szuecs:
:+1:

dataclients/kubernetes/traffic.go
```diff
@@ -0,0 +1,122 @@
...
+	var lastWithWeight int
+	for i, w := range weights {
+		if w > 0 {
+			lastWithWeight = i
+		}
+	}
+
+	bt := make(map[string]backendTraffic)
+
+	for i, bi := range b {
```
MustafaSaber, line 88: I can't map `bi` name to anything

RomanZavodskikh: Probably it means "Backend Index". it is used just 2 lines below:
```
bt[bi.GetName()] = ct
```

AlexanderYastrebov: bi == item of b

MustafaSaber:
👍

PR #2396, opened 2023-06-14
predicates/traffic: add TrafficSegment min equals max tests

AlexanderYastrebov:
* use test request with exactly the same random value as min and max arguments
* add tests for common edge cases

For #2268

### DIFF
--- a/predicates/traffic/segment_test.go
+++ b/predicates/traffic/segment_test.go
@@ -89,16 +89,21 @@ func TestTrafficSegmentMatch(t *testing.T) {
 }
 
 func TestTrafficSegmentMinEqualsMax(t *testing.T) {
-	pp := eskip.MustParsePredicates(`TrafficSegment(0.5, 0.5)`)
-	require.Len(t, pp, 1)
+	for _, minMax := range []float64{
+		0.0,
+		1.0 / 10.0, // can not be represented exactly as float64
+		0.5,
+		2.0 / 3.0, // can not be represented exactly as float64
+		1.0,
+	} {
+		t.Run(fmt.Sprintf("minMax=%v", minMax), func(t *testing.T) {
+			spec := traffic.NewSegment()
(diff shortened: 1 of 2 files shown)


predicates/traffic/segment_test.go
```diff
@@ -89,16 +89,21 @@ func TestTrafficSegmentMatch(t *testing.T) {
...
 func TestTrafficSegmentMinEqualsMax(t *testing.T) {
-	pp := eskip.MustParsePredicates(`TrafficSegment(0.5, 0.5)`)
-	require.Len(t, pp, 1)
-
-	spec := traffic.NewSegment()
-	p, err := spec.Create(pp[0].Args)
-	require.NoError(t, err)
+	for _, minMax := range []float64{
+		0.0,
+		1.0 / 10.0, // can not be represented exactly as float64
```
RomanZavodskikh, line 94: minor: what's the problem?
```
rzavodskikh@ZALANDO-39125:/tmp$ cat main.go 
package main

import "fmt"

func main() {
	fmt.Printf("%+v\n", 1.0 / 10.0 == 0.1)
}
rzavodskikh@ZALANDO-39125:/tmp$ go build main.go 
rzavodskikh@ZALANDO-39125:/tmp$ ./main 
true
```

MustafaSaber:
👍

predicates/traffic/segment_test.go
AlexanderYastrebov: https://go.dev/play/p/Hm0ja99z5mr
```go
package main

import "fmt"

func main() {
	fmt.Printf("%0.40f", 1.0/10.0)
}
```
```
0.1000000000000000055511151231257827021182
Program exited.
```

AlexanderYastrebov: See also https://en.wikipedia.org/wiki/Floating-point_arithmetic#Accuracy_problems

PR #2408, opened 2023-06-19
optimize: change max from 1GB to 4GB to accept tcp conns

szuecs:
optimize: change max from 1GB to 4GB to accept tcp connections in the tcp listener
otherwise we throttle performance for large nodes


szuecs:
:+1:

Issue #2349, opened 2023-05-26
Add filter `dropRequestCookie`


oporkka:
**Is your feature request related to a problem? Please describe.**

We want to be in control of cookies in our Skipper level, and prevent certain cookies to be propagated to backends. 

**Describe the solution you would like**

A cookie like `dropRequestCookie` would help us here. It would go through the cookies in the `Cookie` header, and drop the cookie with the matching name. It would be kind of opposite to `requestCookie`, which adds a cookie with a certain name and value to the request

**Describe alternatives you've considered (optional)**

Use `requestCookie` to set the desired cookie value to an empty string, when propagating the request to the backend. It will avoid big cookie data values to be propagated, although the cookie name will be still propagated.

**Would you like to work on it?**
Yes, possibly, but little time

PR #2410, opened 2023-06-19
feature: dropRequestCookie filter

szuecs:
feature: dropRequestCookie filter

closes https://github.com/zalando/skipper/issues/2349


filters/cookie/cookie_test.go
```diff
@@ -382,3 +383,101 @@ func TestSetCookie(t *testing.T) {
 		}
 	}
 }
+
+func TestDropCookie(t *testing.T) {
```
RomanZavodskikh, line 387: If this filter could have only one arg, that's okay. If not, we need to test filter with two args as well.

szuecs: Yes right now it's only one arg allowed

filters/cookie/cookie.go
```diff
@@ -76,6 +77,76 @@ type filter struct {
...
+
+	lines := req.Header["Cookie"]
+	if len(lines) == 0 {
+		return
+	}
+
+	var result []string
+
+	// modified version of readCookies
+	// https://github.com/golang/go/blob/master/src/net/http/cookie.go#L277
```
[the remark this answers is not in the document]
szuecs: Given that the user that creates a route is the one that could "send" malicious input to the filter to in worst case delete more than one cookie, I don't see the point of copying so much internal Go stdlib code.

filters/cookie/cookie.go
```diff
@@ -76,6 +77,76 @@ type filter struct {
...
+			}
+
+			result = append(result, part)
+		}
+	}
+
+	if len(result) == 0 {
+		delete(req.Header, "Cookie")
+	} else {
+		req.Header["Cookie"] = result
```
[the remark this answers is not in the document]
AlexanderYastrebov: https://pkg.go.dev/net/http#Request.AddCookie
> AddCookie does not attach more than one Cookie header field. That means all cookies, if any, are written into the same line, separated by semicolon.

filters/cookie/cookie_test.go
```diff
@@ -382,3 +383,101 @@ func TestSetCookie(t *testing.T) {
...
+			},
+			wantCookies: nil,
+			wantErr:     false,
+		},
+		{
+			name: "test two cookies and one cookie with match",
+			arg:  "foo",
+			cookies: http.Header{
+				"Cookie": []string{
+					"foo=foo1",
```
AlexanderYastrebov, line 429: We should add more test cases that contain multiple cookies per value

AlexanderYastrebov: I think there could also be several cookies with the same name

RomanZavodskikh:
:+1:

filters/cookie/cookie.go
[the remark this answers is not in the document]
szuecs: yes I see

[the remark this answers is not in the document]
AlexanderYastrebov: > We have extractCookie function for OAuth grant flow

I mean this
https://github.com/zalando/skipper/blob/e88336207452706b8fe92e3086a6b93b327dfbd5/filters/auth/grantcookie.go#L57-L83

docs/reference/filters.md
```diff
@@ -1696,6 +1696,20 @@ oidcClaimsQuery("/:name%\"*One\"", "/path:groups.#[%\"*-Test-Users\"] groups.#[=
 As of now there is no negative/deny rule possible. The first matching path is evaluated against the defined query/queries and if positive, permitted.
 
 ## Cookie Handling
+### dropRequestCookie
```
AlexanderYastrebov, line 1699: Also add dropResponseCookie

filters/cookie/cookie_test.go
szuecs: done

filters/cookie/cookie_test.go
```diff
@@ -382,3 +383,342 @@ func TestSetCookie(t *testing.T) {
...
+		// {
+		// 	name: "test one multivalue cookie with 2 names and no match",
+		// 	arg:  "foo",
+		// 	cookies: http.Header{
+		// 		"Set-Cookie": []string{
+		// 			"bar=foo1;qux=baz",
+		// 		},
+		// 	},
+		// 	wantCookies: map[string][]string{"bar": {"foo1"}, "qux": {"baz"}},
+		// },
```
szuecs, line 627: @AlexanderYastrebov any idea why this would not work?
I get only cookie "bar" no "qux"...

docs/reference/filters.md
szuecs: done

filters/cookie/cookie_test.go
AlexanderYastrebov: Set-Cookie header format differs from Cookie header format - it does not allow multiple cookies per line but uses semicolon as attribute separator.
See
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie

The dropResponseCookie test should be changed to use multiline values like
```go
"Set-Cookie": []string{
	"bar=foo1",
	"qux=baz",
},
```

szuecs: Oh, good that you checked it

szuecs:
:+1:

PR #2414, opened 2023-06-21
add option to set request timeout for request

chandu188:
The current package has timeouts specific to each stage of an http request E.g: `ResponseHeaderTimeout` `TLSHandshakeTimeout`. This PR adds an option to set a timeout that would be applicable to the whole request.

### DIFF
--- a/net/httpclient.go
+++ b/net/httpclient.go
@@ -1,6 +1,7 @@
 package net
 
 import (
+	"context"
 	"crypto/tls"
 	"fmt"
 	"io"
@@ -212,13 +213,14 @@ type Options struct {
 // Transport wraps an http.Transport and adds support for tracing and
 // bearerToken injection.
 type Transport struct {
-	once          sync.Once
-	quit          chan struct{}
-	tr            *http.Transport
-	tracer        opentracing.Tracer
-	spanName      string
-	componentName string
-	bearerToken   string
+	once           sync.Once
+	quit           chan struct{}
+	tr             *http.Transport
+	tracer         opentracing.Tracer
+	spanName       string
+	componentName  string
+	bearerToken    string
+	requestTimeout time.Duration
 }
 
 // NewTransport creates a wrapped http.Transport, with regular DNS
@@ -332,6 +334,12 @@ func WithBearerToken(t *Transport, bearerToken string) *Transport {
 	return tt
 }
 
+func WithRequestTimeout(t *Transport, timeout time.Duration) *Transport {
+	tt := t.shallowCopy()
+	tt.requestTimeout = timeout
+	return tt
+}
+
 func (t *Transport) shallowCopy() *Transport {
 	return &Transport{
 		once:          sync.Once{},
@@ -367,6 +375,11 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	}
 	if t.bearerToken != "" {
 		req.Header.Set("Authorization", "Bearer "+t.bearerToken)
+	}
+	if t.requestTimeout != 0 {
+		reqWithTimeoutCtx, cancelReqContext := context.WithTimeout(req.Context(), t.requestTimeout)
+		req = req.WithContext(reqWithTimeoutCtx)
+		defer cancelReqContext()
 	}
 	rsp, err := t.tr.RoundTrip(req)
 	if span != nil {
--- a/net/httpclient_test.go
+++ b/net/httpclient_test.go
@@ -247,11 +247,16 @@ func TestTransport(t *testing.T) {
 			wantErr:     false,
 		},
 		{
-			name:        "context deadline error should occur when the request is timedout",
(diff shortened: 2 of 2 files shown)


net/httpclient_test.go
```diff
@@ -245,13 +246,24 @@ func TestTransport(t *testing.T) {
 			req:         httptest.NewRequest("GET", "http://example.com/", nil),
 			wantErr:     false,
 		},
+		{
+			name:        "context deadline error should occur when the request is timedout",
+			bearerToken: string(testToken),
+			req:         httptest.NewRequest("GET", "http://example.com/", nil),
+			wantErr:     true,
+			reqTimeout:  10 * time.Millisecond,
```
szuecs, line 254: please also add a test that has `wantErr: false`

net/httpclient_test.go
```diff
@@ -245,13 +246,29 @@ func TestTransport(t *testing.T) {
...
 	} {
 		t.Run(tt.name, func(t *testing.T) {
 			s := startTestServer(func(r *http.Request) {
 				if r.Method != tt.req.Method {
 					t.Errorf("wrong request method got: %s, want: %s", r.Method, tt.req.Method)
 				}
 
+				if tt.reqTimeout != 0 && tt.wantErr {
+					time.Sleep(tt.reqTimeout + 10*time.Millisecond)
+				}
```
AlexanderYastrebov, line 270: Lets have two independent values: tt.reqTimeout for client and tt.serverDelay.
Then we can have two test cases:
* reqTimeout > serverDelay
* reqTimeout < serverDelay

net/httpclient_test.go
```diff
@@ -245,13 +246,29 @@ func TestTransport(t *testing.T) {
...
+			name:       "context deadline error should occur when the request is timedout",
+			req:        httptest.NewRequest("GET", "http://example.com/", nil),
+			wantErr:    true,
+			reqTimeout: 10 * time.Millisecond,
+		},
+		{
+			name:       "no context deadline error should occur when the request is processed within the requested timeout",
+			req:        httptest.NewRequest("GET", "http://example.com/", nil),
+			wantErr:    false,
+			reqTimeout: 5 * time.Millisecond,
```
szuecs, line 259: will be likely too flaky in GH actions.

net/httpclient.go
```diff
@@ -368,6 +376,11 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	if t.bearerToken != "" {
 		req.Header.Set("Authorization", "Bearer "+t.bearerToken)
 	}
+	if t.requestTimeout != 0 {
+		reqWithTimeoutCtx, cancelReqContext := context.WithTimeout(req.Context(), t.requestTimeout)
+		req = req.WithContext(reqWithTimeoutCtx)
+		defer cancelReqContext()
```
AlexanderYastrebov, line 382: This cancels context before response has been consumed.

https://pkg.go.dev/net/http#Request.WithContext
> For outgoing client request, the context controls the entire lifetime of a request and its response: obtaining a connection, sending the request, and reading the response headers and body.

chandu188: @AlexanderYastrebov thank you for pointing that out, but I'm not sure if we have any ways to cover only request timeout then, I see another method https://pkg.go.dev/net/http#NewRequestWithContext to create a request with context and notice that the behavior of the context is the same there as well, so I'm not sure how to proceed. Do you have any suggestions?

AlexanderYastrebov: I do not see an easy way to use request context inside of the client. But client user may always create request with context and cancel it after consuming the response.

chandu188: @AlexanderYastrebov Yeah seems like. @szuecs do you have any suggestions?

szuecs: I guess you would need to have a `net.NewRequest` that you would then need to use to apply it with a default.
```
type struct O {
defaultDuration time.Duration
}

// func NewRequestWithContext(ctx context.Context, method, url string, body io.Reader) (*Request, error)
func (o *O) NewRequest(ctx context.Context, method, url string, body io.Reader)  (*http.Request, error) {
   return http.NewRequestWithContext( context.WithTimeout(ctx, o.defaultDuration), method, body)
}
```

Maybe `O` is the Client ? But this would be also a bit weird to do `client.NewRequest()`.

AlexanderYastrebov: Do we really need to complicate client when user can easily create and close request with context on their own?

szuecs: @AlexanderYastrebov No, I think you are right. The use needs to do anyways the call and then it's likely better to have a const/var in their code base that they reference to get their "default" request timeout.

szuecs:
I think as discussed we see that this contribution was a nice try, but it does not really help the user not to have to interact with creating a request with context, such that the helper would not help.
Therefore closing the PR. Thanks for your investment!

PR #1873, opened 2021-10-05
feature: wire all kubernetes dataclients features

szuecs:
feature: wire all kubernetes dataclients features that are available, also deprecated one because these might be in use

Signed-off-by: Sandor Szücs <sandor.szuecs@zalando.de>

PR #2415, opened 2023-06-21
routesrv: reuse skipper.Options

AlexanderYastrebov:
Reuse skipper.Options datastructure to avoid duplicate work of adding the same flags/options to both skipper and routesrv.

Followup on https://github.com/zalando/skipper/pull/1873


AlexanderYastrebov:
For whatever reason github did not update this PR after forced push, see followoup https://github.com/zalando/skipper/pull/2416

PR #2416, opened 2023-06-21
routesrv: reuse skipper.Options 

AlexanderYastrebov:
Reuse skipper.Options datastructure to avoid duplicate work of adding
the same flags/options to both skipper and routesrv.

Followup on https://github.com/zalando/skipper/pull/1873

- [x] requires https://github.com/zalando/skipper/pull/2419

Signed-off-by: Alexander Yastrebov <alexander.yastrebov@zalando.de>

PR #2419, opened 2023-06-22
skipper: enable skipper_test package

AlexanderYastrebov:
* removes unexported testOptions
* exports run with shutdown channels for tests


redis_test.go
```diff
@@ -0,0 +1,556 @@
...
+	"github.com/zalando/skipper/routesrv"
+	"github.com/zalando/skipper/routing"
+	"github.com/zalando/skipper/routing/testdataclient"
+	"github.com/zalando/skipper/scheduler"
+	"github.com/zalando/skipper/tracing/tracingtest"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestConcurrentKubernetesClusterStateAccessWithRemoteRedis(t *testing.T) {
```
AlexanderYastrebov, line 35: This test uses routesrv.
Within https://github.com/zalando/skipper/pull/2416 I plan to reuse skipper.Options for both routesrv and skipper. To avoid `import cycle not allowed in test` error this test is moved to `skipper_test` package.

I've also moved the second test because it is very similar although does not use routesrv.

MustafaSaber:
👍

AlexanderYastrebov:
:+1:

PR #2421, opened 2023-06-22
skipper: decouple tlsConfig from KubernetesEnableTLS option

AlexanderYastrebov:
Currently CertRegistry is created when KubernetesEnableTLS is set but this may change in the future so this change refactors tlsConfig to check CertRegistry instead of KubernetesEnableTLS option.


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2423, opened 2023-06-23
predicates/traffic: fix flaky TrafficSegment tests

AlexanderYastrebov:
Test failure could be reproduced via:
```
go test ./predicates/traffic/ -count=100 -run=TestTrafficSegmentSplit
```

* use delta of expected value instead of total number of test requests
* use predefined number of test requests for all tests

Based on the biggest difference seen for 1000 test requests:
```
Max difference between 500 and 462 allowed is 25, but difference was 38
```
the delta should be 38/500=0.076

To reduce flakiness:

* increase total number of samples to 10'000
* then increase delta to 6% based on the failed test case output

### DIFF
--- a/predicates/traffic/segment_test.go
+++ b/predicates/traffic/segment_test.go
@@ -51,8 +51,12 @@ func requestWithR(r float64) *http.Request {
 	return req
 }
 
-func doN(t *testing.T, n int, client *proxytest.TestClient, request func() *http.Request) map[int]int {
-	codes := make(map[int]int)
+// doN performs a number of requests and returns the number of responses for each status code and
+// a total number of requests performed.
+// Results use float64 type to simplify fractional comparisons.
+func doN(t *testing.T, client *proxytest.TestClient, request func() *http.Request) (map[int]float64, float64) {
+	const n = 10_000
+	codes := make(map[int]float64)
 	for i := 0; i < n; i++ {
 		rsp, err := client.Do(request())
 		require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


predicates/traffic/segment_test.go
```diff
@@ -51,20 +51,30 @@ func requestWithR(r float64) *http.Request {
 	return req
 }
 
-func doN(t *testing.T, n int, client *proxytest.TestClient, request func() *http.Request) map[int]int {
-	codes := make(map[int]int)
+// doN performs a number of requests and returns the number of responses for each status code and
+// a total number of requests performed.
+// Results use float64 type to simplify fractional comparisons.
+func doN(t *testing.T, client *proxytest.TestClient, request func() *http.Request) (map[int]float64, float64) {
```
szuecs, line 57: Where is the N?
Maybe better to call it `do` ?

szuecs:
:+1:

predicates/traffic/segment_test.go
AlexanderYastrebov: it is to highlight that it does multiple requests

AlexanderYastrebov:
:+1:

predicates/traffic/segment_test.go
```diff
@@ -129,18 +139,13 @@ func TestTrafficSegmentSplit(t *testing.T) {
...
-		N     = 1_000
-		delta = 0.05 * N
-	)
-
-	codes := getN(t, N, p.Client(), p.URL+"/test")
+	codes, n := getN(t, p.Client(), p.URL+"/test")
 
 	t.Logf("Response codes: %v", codes)
 
-	assert.InDelta(t, N*0.5, codes[200], delta)
```
MustafaSaber, line 141: Why moving to a function?

Issue #2280, opened 2023-04-23
Docs: Filter example contains error


bocytko:
**Describe the bug**

Taking the example filter code from https://opensource.zalando.com/skipper/tutorials/development/#filters results in an error:

```
cannot use NewMyFilter() (value of type filters.Spec) as filters.Filter value in return statement: filters.Spec does not implement filters.Filter (missing method Request)
```

**To Reproduce**
Add context, for example:
1. Setup new filter using [example from docs](https://opensource.zalando.com/skipper/tutorials/development/#filters).

**Expected behavior**

```diff
@@ -11,7 +11,7 @@ func NewMyFilter() filters.Spec {
 func (spec *myFilter) Name() string { return "myFilter" }
 
 func (spec *myFilter) CreateFilter(config []interface{}) (filters.Filter, error) {
-       return NewMyFilter(), nil
+       return &myFilter{}, nil
 }
 
 func (f *myFilter) Request(ctx filters.FilterContext) {
```

There is another example under https://opensource.zalando.com/skipper/reference/plugins/#example-filter-plugin which is complete and compiles. Maybe worth deduplicating the documentation here.

**Observed behavior**
Compiler error when using example from documentation.

szuecs:
I think a better fix should be

func NewMyFilter() *myFilter {
    return &myFilter{}
}

Maybe we need to expose the type by changing it to MyFilter, but this should work.
I guess somehow unboxing doesn't work as expected in case we have interface A implemented by T which also implements B, but as A can't be converted to B automatically.

bocytko:
@szuecs you're the expert. I just provided an example that compiled and worked when tested.

AlexanderYastrebov:
Fixed by #2425

PR #2425, opened 2023-06-27
fix: new filter creation examples

alexandreliberato:
Fix documentation, see: https://github.com/zalando/skipper/issues/2280

### DIFF
--- a/docs/tutorials/development.md
+++ b/docs/tutorials/development.md
@@ -116,15 +116,11 @@ The simplest filter possible is, if `filters.Spec` and
 ```go
 type myFilter struct{}
 
-func NewMyFilter() filters.Spec {
+func NewMyFilter() *myFilter {
 	return &myFilter{}
 }
 
 func (spec *myFilter) Name() string { return "myFilter" }
-
-func (spec *myFilter) CreateFilter(config []interface{}) (filters.Filter, error) {
-     return NewMyFilter(), nil
-}
 
 func (f *myFilter) Request(ctx filters.FilterContext) {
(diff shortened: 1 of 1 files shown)


docs/tutorials/development.md
```diff
@@ -116,16 +116,12 @@ The simplest filter possible is, if `filters.Spec` and
...
 type myFilter struct{}
 
-func NewMyFilter() filters.Spec {
+func NewMyFilter() *myFilter {
 	return &myFilter{}
 }
 
 func (spec *myFilter) Name() string { return "myFilter" }
 
-func (spec *myFilter) CreateFilter(config []interface{}) (filters.Filter, error) {
```
AlexanderYastrebov, line 125: This method is a part of filters.Spec interface so it must be implemented
https://github.com/zalando/skipper/blob/399b08b3eba8f01b8ff4efd5ecccbb14fb8bd2e7/filters/filters.go#L180-L190

alexandreliberato: It is everything correct now?

AlexanderYastrebov: What is the point of this change? The title says it fixes something but it does not.

alexandreliberato: It's about this issue: https://github.com/zalando/skipper/issues/2280
If I understood correctly it was necessary to add this code to the documentation. That's right? 
If Im wrong about that, may you make it more clear to me? Thanks in advance.

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2431, opened 2023-06-29
Get rid of kingpin

MustafaSaber:
Kingpin is only used in webhook admission, it makes sense to remove it and use flag package like we do everywhere.

This will stay in draft mode sense we don't have a test case for webhook admission flags, so another PR will be created to test the flags first, then we will follow up on this.

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"flag"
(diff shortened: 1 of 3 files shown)


MustafaSaber:
I can't find an easy way to test flags here.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2432, opened 2023-06-30
filters/diag: add histogram latency filters

AlexanderYastrebov:
Add filters that add latency based on the configured histogram. See https://stats.stackexchange.com/questions/464006/generating-samples-from-a-histogram

Example:
```
$ bin/skipper -inline-routes='* -> histogramRequestLatency("5ms", 20, "10ms", 30, "15ms", 50, "20ms") -> status(204) -> <shunt>' -access-log-disabled
```

Latency distribution:
```
~$ echo "GET http://localhost:9090" | vegeta attack -rate=100/s -duration=20s | vegeta report -type hist[0ms,6ms,11ms,16ms,21ms,26ms]
Bucket         #    %       Histogram
[0s,    6ms]   8    0.40%
[6ms,   11ms]  372  18.60%  #############
[11ms,  16ms]  581  29.05%  #####################
[16ms,  21ms]  964  48.20%  ####################################
[21ms,  26ms]  75   3.75%   ##
[26ms,  +Inf]  0    0.00%
```
(note that buckets are shifted by 1ms to account for intrinsic latency)


filters/diag/histogram.go
```diff
@@ -0,0 +1,123 @@
...
+		if err != nil || max <= 0 {
+			return nil, filters.ErrInvalidFilterParameters
+		}
+		if i > 0 && max <= f.buckets[i/2-1] {
+			return nil, filters.ErrInvalidFilterParameters
+		}
+		f.buckets[i/2] = max
+	}
+
+	if f.sum > 0 {
```
MustafaSaber, line 86: shouldn't this return an error instead of going around it?

szuecs: We use relative weights in other cases (f.e. `Traffic()`), so should be fine

filters/diag/histogram.go
```diff
@@ -0,0 +1,123 @@
...
+		if upto >= r {
+			break
+		}
+	}
+	max := f.buckets[i]
+	var min time.Duration
+	if i > 0 {
+		min = f.buckets[i-1]
+	}
+	return min + time.Duration(rand.Int63n(int64(max-min))) // #nosec
```
szuecs, line 122: How does `min + x` match the example:
```
// Eskip example:
//
//	r: * -> histogramRequestLatency(50, "5ms", 0, "10ms", 30, "15ms", 20, "20ms") -> "https://www.example.org";
//
// The example above adds a latency
// * between implicit 0ms and 5ms to 50% of the requests   <--
```

AlexanderYastrebov: Previous loop selects `i`-th bucket according to its weight. This bucket left boundary `min` is `bucket[i-1]` (or zero) and right boundary `max` is `bucket[i]`. Then filter selects a random value from `[min, max)` interval as `rand(min, max) = min + rand(0, max-min)`

filters/diag/histogram.go
AlexanderYastrebov: When sum is zero filter becomes a noop. Maybe we should disallow all zero weights and indeed return an error here.

AlexanderYastrebov: Fixed

filters/diag/histogram.go
szuecs: Where is the implicit 0ms?

AlexanderYastrebov: The example is not wrong, please see docs for explanation:
```
The example above adds a latency
* between implicit 0ms and 5ms to 50% of the requests
* between 5ms and 10ms to 0% of the requests
* between 10ms and 15ms to 30% of the requests
* and between 15ms and 20ms to 20% of the requests.
```

AlexanderYastrebov: If it makes it easier to understand we may change filter to always require `"0ms"` (or any zero duration e.g. "0h") as a first argument.

AlexanderYastrebov: Or I we can actually swap order of weights and duration and require first and last arguments to be durations, e.g:
`histogramRequestLatency("5ms", 10, "10ms", 0, "20ms", 90, "30ms")` would add 5-10ms latency to 10% of requests and 20-30ms to 90%

szuecs: Order of args does not matter.

I think what needs to be clear is min. Either min is the upper bound of the leftmost bucket or the example or code is wrong.
Maybe min is not the right name for it. That's just what is not easy to understand and maybe you should define a min as absolute min?!

AlexanderYastrebov: Thank you.

> I think your last comment would make most sense as user input.

I've updated the code accordingly.

szuecs:
:+1:

MustafaSaber:
👍

Issue #2435, opened 2023-07-02
Fix flaky TestTeeEndToEndBody2TeeRoutesAndClosing


AlexanderYastrebov:
```
--- FAIL: TestTeeEndToEndBody2TeeRoutesAndClosing (0.00s)
    tee_test.go:265: shadow handler got a request, but should not
panic: close of closed channel [recovered]
	panic: close of closed channel

goroutine 50 [running]:
testing.tRunner.func1.2({0x128aa60, 0x14e1d80})
	/usr/local/go/src/testing/testing.go:1526 +0x24e
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1529 +0x39f
panic({0x128aa60, 0x14e1d80})
	/usr/local/go/src/runtime/panic.go:884 +0x213
runtime.Goexit()
	/usr/local/go/src/runtime/panic.go:522 +0x177
testing.(*common).FailNow(0xc00028cb60)
	/usr/local/go/src/testing/testing.go:980 +0x4c
testing.(*common).Fatal(0xc00028cb60, {0xc00014db88?, 0x13afa38?, 0xc0002927f0?})
	/usr/local/go/src/testing/testing.go:1057 +0x58
github.com/zalando/skipper/filters/tee.TestTeeEndToEndBody2TeeRoutesAndClosing(0xc00028cb60)
	/workspace/filters/tee/tee_test.go:265 +0xe8f
testing.tRunner(0xc00028cb60, 0x140b158)
	/usr/local/go/src/testing/testing.go:1576 +0x10b
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1629 +0x3ea
FAIL	github.com/zalando/skipper/filters/tee	0.014s
FAIL
/bin/sh: 1: break: Illegal number: -1
```

https://github.com/zalando/skipper/pull/2427#issuecomment-1610233234

Related https://github.com/zalando/skipper/pull/2272

AlexanderYastrebov:
https://github.com/zalando/skipper/actions/runs/6036110251/job/16377774962?pr=2549

```
--- FAIL: TestTeeEndToEndBody2TeeRoutesAndClosing (0.01s)
    tee_test.go:268: shadow handler got a request, but should not
panic: close of closed channel [recovered]
	panic: close of closed channela

goroutine 50 [running]:
testing.tRunner.func1.2({0x17ecde0, 0x1aa2790})
[...]

PR #2443, opened 2023-07-04
Feature/admission control multi layer support

szuecs:
feature: admission control multi layer support

fix: inode admission controllers should not count shedded responses by leaf admission controllers 
test: create test that check the leaf admission controllers will not interfere with inode admission controllers in the call tree 
test: fix flaky test (works on my machine)

### DIFF
--- a/filters/shedder/admission_test.go
+++ b/filters/shedder/admission_test.go
@@ -2,25 +2,30 @@ package shedder
 
 import (
 	"fmt"
+	"io"
 	"math/rand"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"sort"
+	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/filters"
 	"github.com/zalando/skipper/filters/builtin"
+	"github.com/zalando/skipper/metrics"
+	"github.com/zalando/skipper/metrics/metricstest"
+	"github.com/zalando/skipper/net"
 	"github.com/zalando/skipper/proxy/proxytest"
 	"github.com/zalando/skipper/routing"
 	"github.com/zalando/skipper/routing/testdataclient"
+	"golang.org/x/time/rate"
 )
 
 func TestAdmissionControl(t *testing.T) {
-	// fixed rand to have non flaky tests
-	rand := rand.New(rand.NewSource(5))
-
 	for _, ti := range []struct {
 		msg                        string
 		mode                       string
@@ -92,7 +97,7 @@ func TestAdmissionControl(t *testing.T) {
 		exponent:                   1.0,
 		N:                          20,
 		pBackendErr:                0.2,
-		pExpectedAdmissionShedding: 0.1,
+		pExpectedAdmissionShedding: 0.15,
 	}, {
 		msg:                        "medium error rate and bigger than threshold will block some traffic",
 		mode:                       "active",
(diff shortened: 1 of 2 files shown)


szuecs:
Also got https://github.com/zalando/skipper/issues/2435

filters/shedder/admission_test.go
```diff
@@ -208,6 +240,127 @@ func TestAdmissionControl(t *testing.T) {
...
+			w.WriteHeader(http.StatusOK)
+		}
+	}))
+	defer backend.Close()
+
+	spec := NewAdmissionControl(Options{}).(*AdmissionControlSpec)
+
+	argsLeaf := make([]interface{}, 0, 6)
+	argsLeaf = append(argsLeaf, "testmetric-leaf", "active", "5ms", 5, 5, 0.9, 0.95, 1.0)
+	fLeaf, err := spec.CreateFilter(argsLeaf)
```
AlexanderYastrebov, line 266: I think filter creation in this tests obscures the main test logic. Do we need it here?

filters/shedder/admission_test.go
```diff
@@ -208,6 +240,127 @@ func TestAdmissionControl(t *testing.T) {
...
+	defer proxyLeaf.Close()
+
+	r := &eskip.Route{Filters: []*eskip.Filter{{Name: spec.Name(), Args: args}}, Backend: proxyLeaf.URL}
+	proxy := proxytest.New(fr, r)
+	defer proxy.Close()
+
+	reqURL, err := url.Parse(proxy.URL)
+	if err != nil {
+		t.Fatalf("Failed to parse url %s: %v", proxy.URL, err)
+	}
```
AlexanderYastrebov, line 296: Proxy url is always valid, no need to parse it

filters/shedder/admission_test.go
```diff
@@ -208,6 +240,127 @@ func TestAdmissionControl(t *testing.T) {
...
+
+	rt := net.NewTransport(net.Options{})
+	defer rt.Close()
```
AlexanderYastrebov, line 299: Why not `proxy.Client()` which is https://pkg.go.dev/net/http/httptest#Server.Client

filters/shedder/admission_test.go
```diff
@@ -117,10 +122,37 @@ func TestAdmissionControl(t *testing.T) {
...
+		mode:                       "inactive",
+		d:                          10 * time.Millisecond,
+		windowsize:                 5,
+		minRequests:                10,
+		successThreshold:           0.9,
+		maxrejectprobability:       0.95,
+		exponent:                   1.0,
+		N:                          20,
+		pBackendErr:                0.8,
+		pExpectedAdmissionShedding: 0.0,
```
AlexanderYastrebov, line 136: I just trust these numbers make sense :)

filters/shedder/admission_test.go
```diff
@@ -208,6 +240,127 @@ func TestAdmissionControl(t *testing.T) {
...
+
+	N := 1000
+	var wg sync.WaitGroup
+	wg.Add(N)
+	lim := rate.NewLimiter(50, 50)
+	now := time.Now()
+	for i := 0; i < N; i++ {
+		r := lim.Reserve()
+		for !r.OK() {
+			time.Sleep(10 * time.Microsecond)
```
AlexanderYastrebov, line 315: This makes tests run longer. We need to think whether it is necessary and maybe we can mock the clock like in https://github.com/zalando/skipper/pull/2432 and https://github.com/zalando/skipper/pull/2453

szuecs: I need to have something that jitters the funcs, because otherwise you get errors from the runtime.

filters/shedder/admission_test.go
szuecs: I basically copied the last test that has `pExpectedAdmissionShedding: 0.91,` and for "inactive" I expect  `pExpectedAdmissionShedding: 0`

filters/shedder/admission_test.go
szuecs: In this case I want to have a well configured connection pool. In cases which require a significant amount of requests to make sense we should rely on our configuration.

filters/shedder/admission_test.go
szuecs: Strictly we don't need it but creating invalid filters is easy to do with so many args, so I think it's better to bail out early.

szuecs:
:+1:

Issue #2140, opened 2022-11-07
Tokeninfo connection error response


AlexanderYastrebov:
When Skipper has connection errors to tokeninfo, the response status code should be 500 and not 401

AlexanderYastrebov:
Or at least operator should be able to tell apart invalid token from token info connection errors.
AuthClient returns `errInvalidToken` on any error
https://github.com/zalando/skipper/blob/405acf62daf17b483a4658a0a84671979aa8a961/filters/auth/authclient.go#L116-L125
so filter never logs `Error while calling tokeninfo`
https://github.com/zalando/skipper/blob/405acf62daf17b483a4658a0a84671979aa8a961/filters/auth/tokeninfo.go#L321-L331

szuecs:
@AlexanderYastrebov not sure if we should leak this to the client. Maybe better to have a generic error message to the client and fix the issue by logging the reason.

AlexanderYastrebov:
@szuecs Agree, the client gets response from `unauthorized(ctx, "", reason, f.authClient.url.Hostname(), "") `, lets keep it like that for all errors. In the authClient we may return different error for 5xx token info response and have it logged.

PR #2454, opened 2023-07-12
auth: log if http status code is 5xx

MustafaSaber:
fixes https://github.com/zalando/skipper/issues/2140

### DIFF
--- a/filters/auth/authclient.go
+++ b/filters/auth/authclient.go
@@ -91,6 +91,12 @@ func (ac *authClient) getTokenintrospect(token string, ctx filters.FilterContext
 	}
 	defer rsp.Body.Close()
 
+	// check 5xx status codes
+	if rsp.StatusCode >= 500 && rsp.StatusCode < 600 {
+		io.Copy(io.Discard, rsp.Body)
+		return nil, fmt.Errorf("token introspection failed with status code: %d", rsp.StatusCode)
+	}
+
 	if rsp.StatusCode != 200 {
 		io.Copy(io.Discard, rsp.Body)
 		return nil, errInvalidToken
@@ -124,6 +130,12 @@ func (ac *authClient) getTokeninfo(token string, ctx filters.FilterContext) (map
 		return doc, err
 	}
 	defer rsp.Body.Close()
+
+	// check 5xx status codes
+	if rsp.StatusCode >= 500 && rsp.StatusCode < 600 {
+		io.Copy(io.Discard, rsp.Body)
+		return nil, fmt.Errorf("token info failed with status code: %d", rsp.StatusCode)
+	}
 
 	if rsp.StatusCode != 200 {
 		io.Copy(io.Discard, rsp.Body)
--- a/filters/auth/authclient.go
+++ b/filters/auth/authclient.go
@@ -93,7 +93,7 @@ func (ac *authClient) getTokenintrospect(token string, ctx filters.FilterContext
 
 	if rsp.StatusCode != 200 {
 		io.Copy(io.Discard, rsp.Body)
-		if rsp.StatusCode >= 500 && rsp.StatusCode < 600 {
+		if rsp.StatusCode >= 500 && rsp.StatusCode <= 599 {
 			return nil, fmt.Errorf("token introspection failed with status code: %d", rsp.StatusCode)
 		}
 		return nil, errInvalidToken
@@ -130,7 +130,7 @@ func (ac *authClient) getTokeninfo(token string, ctx filters.FilterContext) (map
 
 	if rsp.StatusCode != 200 {
 		io.Copy(io.Discard, rsp.Body)
-		if rsp.StatusCode >= 500 && rsp.StatusCode < 600 {
+		if rsp.StatusCode >= 500 && rsp.StatusCode <= 599 {
 			return nil, fmt.Errorf("token info failed with status code: %d", rsp.StatusCode)
 		}
 		return doc, errInvalidToken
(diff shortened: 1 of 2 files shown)


MustafaSaber:
Still thinking of a way to test this

filters/auth/authclient.go
```diff
@@ -91,6 +91,12 @@ func (ac *authClient) getTokenintrospect(token string, ctx filters.FilterContext
 	}
 	defer rsp.Body.Close()
 
+	// check 5xx status codes
+	if rsp.StatusCode >= 500 && rsp.StatusCode < 600 {
+		io.Copy(io.Discard, rsp.Body)
+		return nil, fmt.Errorf("token introspection failed with status code: %d", rsp.StatusCode)
+	}
```
AlexanderYastrebov, line 98: I think it is better to move this check into `if rsp.StatusCode != 200`

AlexanderYastrebov: > it is better

So that we have generic check that it is not 200 and then return specific error for 5xx.
I would also change `rsp.StatusCode < 600` to `rsp.StatusCode <= 599` - to highlight we are checking 5xx range.

AlexanderYastrebov:
> Still thinking of a way to test this

I think you can take this as a blueprint and verify response code and body
https://github.com/zalando/skipper/blob/3cffdd6444ca5c1383e052cdf61eda14d2f6c13e/filters/auth/tokeninfo_test.go#L189-L257

filters/auth/tokeninfo_test.go
```diff
@@ -397,6 +398,51 @@ func TestOAuth2TokenTimeout(t *testing.T) {
...
+	r := &eskip.Route{Filters: []*eskip.Filter{{Name: spec.Name(), Args: args}}, Backend: backend.URL}
+
+	proxy := proxytest.New(fr, r)
+	defer proxy.Close()
+
+	reqURL, err := url.Parse(proxy.URL)
+	require.NoError(t, err, "Failed to parse url %s: %v", proxy.URL, err)
+
+	req, err := http.NewRequest("GET", reqURL.String(), nil)
+	require.NoError(t, err)
```
AlexanderYastrebov, line 434: Just use `proxy.URL` directly as we know it is valid

filters/auth/authclient.go
```diff
@@ -93,6 +93,9 @@ func (ac *authClient) getTokenintrospect(token string, ctx filters.FilterContext
 
 	if rsp.StatusCode != 200 {
 		io.Copy(io.Discard, rsp.Body)
+		if rsp.StatusCode >= 500 && rsp.StatusCode <= 599 {
```
szuecs, line 96: Don't you think that you want to know if status code is >=600? I mean for sure I would like to see if this happens because it's a bug somewhere likely in stdlib.
Maybe also 499 or 0 would be great to check (maybe switch block and return different errors to differentiate them)

filters/auth/authclient.go
```diff
@@ -127,6 +130,9 @@ func (ac *authClient) getTokeninfo(token string, ctx filters.FilterContext) (map
...
```
szuecs, line 133: Same as above

szuecs (changes requested):

filters/auth/tokeninfo_test.go
```diff
@@ -397,6 +398,48 @@ func TestOAuth2TokenTimeout(t *testing.T) {
...
+		Timeout: testAuthTimeout,
+	}
+	args := []interface{}{testScope}
+
+	t.Logf("ti.options: %#v", opts)
+
+	spec := newOAuthTokeninfoSpec(filters.OAuthTokeninfoAnyScopeName, opts)
+
+	_, err := spec.CreateFilter(args)
+	require.NoError(t, err, "error creating filter")
```
AlexanderYastrebov, line 421: No need to create filter, we know args are valid

filters/auth/tokeninfo_test.go
```diff
@@ -397,6 +398,48 @@ func TestOAuth2TokenTimeout(t *testing.T) {
...
+
+	fr := make(filters.Registry)
+	fr.Register(spec)
+	r := &eskip.Route{Filters: []*eskip.Filter{{Name: spec.Name(), Args: args}}, Backend: backend.URL}
```
AlexanderYastrebov, line 425: Could use
```go
r := eskip.MustParse(`* -> ...`)
```

filters/auth/tokeninfo_test.go
```diff
@@ -397,6 +398,48 @@ func TestOAuth2TokenTimeout(t *testing.T) {
...
+
+	proxy := proxytest.New(fr, r)
+	defer proxy.Close()
+
+	req, err := http.NewRequest("GET", proxy.URL, nil)
+	require.NoError(t, err)
+
+	req.Header.Set(authHeaderName, authHeaderPrefix+testToken)
+
+	rsp, err := http.DefaultClient.Do(req)
```
AlexanderYastrebov, line 435: Please use proxy.Client(), see https://pkg.go.dev/net/http/httptest#Server.Client

filters/auth/authclient.go
```diff
@@ -93,6 +93,9 @@ func (ac *authClient) getTokenintrospect(token string, ctx filters.FilterContext
 
 	if rsp.StatusCode != 200 {
 		io.Copy(io.Discard, rsp.Body)
+		if rsp.StatusCode != 403 && rsp.StatusCode != 401 {
+			return nil, fmt.Errorf("token introspection failed with status code: %d", rsp.StatusCode)
```
AlexanderYastrebov, line 97: Lets do just "failed with status code: %d" - the caller knows and logs what it calls (token introspection or tokeninfo)

filters/auth/tokeninfo_test.go
MustafaSaber: `MustParse` returns a list of routes

AlexanderYastrebov: proxytest.New supports varargs `proxytest.New(fr, r...)`

AlexanderYastrebov:
:+1:

MustafaSaber:
👍

PR #1866, opened 2021-09-23
Implements HTTPS graceful shutdown

AlexanderYastrebov:
Fixes #1865

Signed-off-by: Alexander Yastrebov <alexander.yastrebov@zalando.de>

PR #2455, opened 2023-07-12
skipper: use queuelistener for HTTPS

AlexanderYastrebov:
PR #1866 implemented HTTPS graceful shutdown but did not use queuelistener.

### DIFF
--- a/skipper.go
+++ b/skipper.go
@@ -1135,9 +1135,6 @@ func (o *Options) tlsConfig(cr *certregistry.CertRegistry) (*tls.Config, error) 
 }
 
 func listen(o *Options, address string, mtr metrics.Metrics) (net.Listener, error) {
-	if address == "" {
-		address = ":http"
-	}
 
 	if !o.EnableTCPQueue {
 		return net.Listen("tcp", address)
@@ -1203,9 +1200,19 @@ func listenAndServeQuit(
 	if err != nil {
 		return err
 	}
+	serveTLS := tlsConfig != nil
+
+	address := o.Address
+	if address == "" {
+		if serveTLS {
+			address = ":https"
+		} else {
+			address = ":http"
+		}
+	}
 
 	srv := &http.Server{
-		Addr:              o.Address,
+		Addr:              address,
 		TLSConfig:         tlsConfig,
 		Handler:           proxy,
 		ReadTimeout:       o.ReadTimeoutServer,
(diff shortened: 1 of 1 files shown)


skipper.go
```diff
@@ -1203,9 +1200,19 @@ func listenAndServeQuit(
...
+	serveTLS := tlsConfig != nil
+
+	address := o.Address
+	if address == "" {
+		if serveTLS {
+			address = ":https"
+		} else {
+			address = ":http"
+		}
+	}
```
AlexanderYastrebov, line 1212: Moved out empty HTTP address handling from `listen` and added empty HTTPS address handling as [ListenAndServeTLS does](https://github.com/golang/go/blob/6244b1946bc2101b01955468f1be502dbadd6807/src/net/http/server.go#L3272-L3274).

skipper.go
```diff
@@ -1203,9 +1200,19 @@ func listenAndServeQuit(
...
 
 	srv := &http.Server{
-		Addr:              o.Address,
+		Addr:              address,
```
AlexanderYastrebov, line 1215: `Addr` can be dropped as it is not used by  `Serve` and `ServeTLS`, only by `ListenAndServe` and `ListenAndServeTLS` which we do not use.

skipper.go
```diff
@@ -1223,6 +1230,13 @@ func listenAndServeQuit(
 		}
 	}
 
+	log.Infof("Listen on %v", address)
+
+	l, err := listen(o, address, mtr)
+	if err != nil {
+		return err
+	}
```
AlexanderYastrebov, line 1238: Now here is the single place where main listener is created (ignoring insecure listener) which opens up a way to wrap it e.g. to track connections.

skipper.go
```diff
@@ -1266,16 +1278,12 @@ func listenAndServeQuit(
...
 
-		if err := srv.ListenAndServeTLS("", ""); err != http.ErrServerClosed {
-			log.Errorf("ListenAndServeTLS failed: %v", err)
+		if err := srv.ServeTLS(l, "", ""); err != http.ErrServerClosed {
+			log.Errorf("ServeTLS failed: %v", err)
 			return err
 		}
 	} else {
 		log.Infof("TLS settings not found, defaulting to HTTP")
-		l, err := listen(o, o.Address, mtr)
```
MustafaSaber, line 1275: Will this break anything?

AlexanderYastrebov: listen() creates queuelistener if config is set, in this PR I move it out of "not HTTPS" branch to also use for HTTPS setup.

AlexanderYastrebov: I.e. for HTTP case nothing changes, for HTTPS queue listener will be used when config is specified.

You may also check stdlib for how srv.ListenAndServeTLS is different from srv.ServeTLS

MustafaSaber: So it will also default, I didn't look close enough, thanks!

AlexanderYastrebov:
:+1:

PR #2456, opened 2023-07-13
Enable PATCH method for being measured

javespi:
[PATCH](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH) method is missed when they are measured; so it's tagged as `_unknownmethod_`

Example of dashboard in Grafana showing these metrics:
image


metrics/utils.go
```diff
@@ -31,6 +31,7 @@ func measuredMethod(m string) string {
 		"HEAD",
 		"POST",
 		"PUT",
+		"PATCH",
```
AlexanderYastrebov, line 34: Thank you for the fix. Could please also add TRACE and CONNECT methods such that we end up having all methods from here https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

AlexanderYastrebov:
Thank you :+1:

AlexanderYastrebov:
:+1:

szuecs:
:+1:

MustafaSaber:
Thanks for your contribution!

javespi:
You are welcome!

PR #2458, opened 2023-07-13
Temporary skip tests using net/redistest

AlexanderYastrebov:
Redis testcontainer fails to start due to https://github.com/testcontainers/testcontainers-go/issues/1359 caused by https://github.com/golang/go/issues/61076 and we can not pin cdp-runtime/go to a working patch version before 1.20.6

### DIFF
--- a/net/redistest/redistest.go
+++ b/net/redistest/redistest.go
@@ -15,6 +15,11 @@ func NewTestRedis(t testing.TB) (address string, done func()) {
 }
 
 func NewTestRedisWithPassword(t testing.TB, password string) (address string, done func()) {
+
+	if fixDeadline, _ := time.Parse("2006-01-02", "2023-08-01"); time.Now().Before(fixDeadline) {
+		t.Skip("https://github.com/testcontainers/testcontainers-go/issues/1359")
+	}
+
 	var args []string
 	if password != "" {
 		args = append(args, "--requirepass", password)


AlexanderYastrebov:
:+1:

net/redistest/redistest.go
```diff
@@ -15,6 +15,9 @@ func NewTestRedis(t testing.TB) (address string, done func()) {
 }
 
 func NewTestRedisWithPassword(t testing.TB, password string) (address string, done func()) {
+
+	t.Skip("https://github.com/testcontainers/testcontainers-go/issues/1359")
```
RomanZavodskikh, line 19: Can we set some deadline after which this skip won't work?
Something like this:
```
if time.Now() < time.Parse("2023-Sep-01", "2014-Feb-04") {
    t.Skip("https://github.com/testcontainers/testcontainers-go/issues/1359")
}
```

AlexanderYastrebov: Good idea, fixed.

net/redistest/redistest.go
```diff
@@ -15,6 +15,11 @@ func NewTestRedis(t testing.TB) (address string, done func()) {
 }
 
 func NewTestRedisWithPassword(t testing.TB, password string) (address string, done func()) {
+
+	if fixDeadline, _ := time.Parse("2006-01-02", "2023-08-01"); time.Now().Before(fixDeadline) {
```
AlexanderYastrebov, line 19: Could use `time.DateOnly` but its only available since go1.20

RomanZavodskikh:
:+1:

net/redistest/redistest.go
MustafaSaber: I don't get the relevance of 2006 date

MustafaSaber, line 19: I don't get the relevance of 2006 date

AlexanderYastrebov: Check out this https://pkg.go.dev/time#Parse

AlexanderYastrebov:
Merging as testcontainers fix may take a while

AlexanderYastrebov:
:+1:

PR #1562, opened 2020-10-22
Adds Content-Length header for error response

AlexanderYastrebov:
`Content-Length` header disables `Transfer-Encoding: chunked` and allows
client to fail earlier in case ServeHTTP hangs for some reason.

Consider malformed server that hangs:
```go
package main

import (
	"fmt"
	"log"
	"net/http"
	"time"
)

func main() {
	http.HandleFunc("/error", func(w http.ResponseWriter, r *http.Request) {

		code := http.StatusBadGateway
		text := http.StatusText(code)

		if r.URL.Query().Get("cl") != "" {
			w.Header().Set("Content-Length", fmt.Sprintf("%d", 1+len(text)))
		}
		http.Error(w, text, code)

		f := w.(http.Flusher)
		f.Flush()

		time.Sleep(10 * time.Hour)
	})

	fmt.Printf("Starting server at port 8080\n")
	if err := http.ListenAndServe(":8080", nil); err != nil {
		log.Fatal(err)
	}
}
```
and
```
$ go run error.go 
Starting server at port 8080
```
so
```
$ curl -v "localhost:8080/error"
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /error HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 502 Bad Gateway
< Content-Type: text/plain; charset=utf-8
< X-Content-Type-Options: nosniff
< Date: Thu, 22 Oct 2020 14:10:17 GMT
< Transfer-Encoding: chunked
< 
Bad Gateway

```
hangs, and
```
$ curl -v "localhost:8080/error?cl=1"
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /error?cl=1 HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 502 Bad Gateway
< Content-Length: 12
< Content-Type: text/plain; charset=utf-8
< X-Content-Type-Options: nosniff
< Date: Thu, 22 Oct 2020 14:10:47 GMT
< 
Bad Gateway
* Connection #0 to host localhost left intact
```
does not.

PR #2459, opened 2023-07-14
proxy: set zero content length for default response

AlexanderYastrebov:
Content-Length header prevents http.ResponseWriter from writing chunked response.

It turns out that reading empty chunked response is slow(er). In particular this change speeds up TrafficSegment tests that perform 10k requests to routes that send no body:
```
r50: Path("/test") && TrafficSegment(0.0, 0.5) -> status(200) -> <shunt>;
r30: Path("/test") && TrafficSegment(0.5, 0.8) -> status(201) -> <shunt>;
r20: Path("/test") && TrafficSegment(0.8, 1.0) -> status(202) -> <shunt>;
```

Test runtime before the change:
```
$ go test ./predicates/traffic/ -count=1 -run=TestTrafficSegment
ok      github.com/zalando/skipper/predicates/traffic   9.357s
```
and after:
```
ok      github.com/zalando/skipper/predicates/traffic   4.260s
```

Related #1562

### DIFF
--- a/proxy/defaultresponse_test.go
+++ b/proxy/defaultresponse_test.go
@@ -1,0 +1,26 @@
+package proxy_test
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/proxy/proxytest"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDefaultResponse(t *testing.T) {
+	p := proxytest.Config{
+		Routes: eskip.MustParse(`* -> <shunt>`),
+	}.Create()
+	defer p.Close()
+
+	rsp, body, err := p.Client().GetBody(p.URL)
+	require.NoError(t, err)
+
+	assert.Equal(t, http.StatusNotFound, rsp.StatusCode)
+	assert.Len(t, body, 0)
+	assert.Equal(t, "0", rsp.Header.Get("Content-Length"))
+}
--- a/proxy/context.go
+++ b/proxy/context.go
@@ -1,7 +1,6 @@
 package proxy
 
 import (
-	"bytes"
 	stdlibcontext "context"
 	"errors"
 	"io"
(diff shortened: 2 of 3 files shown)


proxy/defaultresponse_test.go
```diff
@@ -0,0 +1,26 @@
...
+		Routes: eskip.MustParse(`* -> <shunt>`),
+	}.Create()
+	defer p.Close()
+
+	rsp, body, err := p.Client().GetBody(p.URL)
+	require.NoError(t, err)
+
+	assert.Equal(t, http.StatusNotFound, rsp.StatusCode)
+	assert.Len(t, body, 0)
+	assert.Equal(t, "0", rsp.Header.Get("Content-Length"))
```
AlexanderYastrebov, line 25: This is an indirect test. It is hard to test that response is not chunked because http client handles chunked encoding transparently. To test it we would need to create net connection and send/receive raw HTTP request and response.

MustafaSaber:
👍

proxy/defaultresponse_test.go
szuecs: I don't mind, but sounds simple enough to do:
```
GET / HTTP/1.1\r\n
Host: test\r\n
\r\n
```

szuecs:
@AlexanderYastrebov I think it does not comply with https://www.rfc-editor.org/rfc/rfc9110.html#name-content-length , see for example by status code "MUST NOT".

AlexanderYastrebov:
@szuecs Could you quote the part you mean?

szuecs:
> A server MUST NOT send a Content-Length header field in any response with a status code of [1xx (Informational)](https://www.rfc-editor.org/rfc/rfc9110.html#status.1xx) or [204 (No Content)](https://www.rfc-editor.org/rfc/rfc9110.html#status.204). A server MUST NOT send a Content-Length header field in any [2xx (Successful)](https://www.rfc-editor.org/rfc/rfc9110.html#status.2xx) response to a CONNECT request ([Section 9.3.6](https://www.rfc-editor.org/rfc/rfc9110.html#CONNECT)).

AlexanderYastrebov:
Maybe it is not worth the effort.

szuecs:
@AlexanderYastrebov maybe the cases are also fine. Maybe you are right 😉

PR #2457, opened 2023-07-13
filters/auth: allow insecure grant flow

AlexanderYastrebov:
Add a flag to allow insecure grant flow:
* issue token cookie without secure attribute
* use http schem for callback url

PR #2460, opened 2023-07-14
Fixed the searchRing function for consistentHash

RomanZavodskikh:
This fix is needed to comply the requirements of input for binary search (sort.Search).


RomanZavodskikh:
Attaching two pair graphs to show that this change really makes sense. Here are results of 'TestFadeIn_consistent-hash,_4', the first one before fix, the second if after. This test introduces three "old" nodes and one "fading in" one. As we can see, fix makes load between three "old" nodes more even.
TestFadeIn_consistent-hash,_4
TestFadeIn_consistent-hash,_4

RomanZavodskikh:
Something similar in 'TestFadeIn/consistent-hash,_5'. Three "old" nodes, three "fading in" ones. Before fix, one of three "old" nodes get less load, after fix all three get the same load.
TestFadeIn_consistent-hash,_5
TestFadeIn_consistent-hash,_5

RomanZavodskikh:
CI failed because of https://github.com/zalando/skipper/pull/2457#issuecomment-1634899799

AlexanderYastrebov:
consistentHash fadeIn support was introduced by https://github.com/zalando/skipper/pull/2087
@herojan FYI

AlexanderYastrebov:
@RomanZavodskikh Please update commit message (and PR description) with the description of the problem and why this is the proper fix.

AlexanderYastrebov:
> Attaching two pair graphs to show that this change really makes sense.

It is not obvious how to interpret the graphs tbh. We should think how to make it easier to eyeball sanity check, maybe color-code datapoints according to endpoint.

AlexanderYastrebov:
@RomanZavodskikh Would it be possible to craft a specific test that fails before the fix and passes after?

szuecs:
Great work!

szuecs:
:+1:

herojan:
Nice find, makes sense!

loadbalancer/algorithm.go
```diff
@@ -241,10 +241,13 @@ func hash(s string) uint64 {
 // Returns index in hash ring with the closest hash to key's hash
 func (ch *consistentHash) searchRing(key string, skipEndpoint func(int) bool) int {
 	h := hash(key)
-	i := sort.Search(ch.Len(), func(i int) bool { return ch.hashRing[i].hash >= h && !skipEndpoint(ch.hashRing[i].index) })
+	i := sort.Search(ch.Len(), func(i int) bool { return ch.hashRing[i].hash >= h })
 	if i == ch.Len() { // rollover
 		i = 0
 	}
+	for skipEndpoint(ch.hashRing[i].index) {
```
herojan, line 248: The only potential issue I see is that this function is no longer safe if all endpoints are fading, since it would be an infinite loop. I believe this currently can't happen because it's handled at https://github.com/zalando/skipper/blob/24e762b2c17c4d1c94da1cd24dbcc205557028d4/loadbalancer/algorithm.go#L137, but it's not that obvious since it relies on the caller handling this case. So perhaps make the function return the result of the binary search if it does a full loop of the hashring?

RomanZavodskikh: As for me, the current code seems logical to me. However, not to miss this case next time I've added `TestApplyEndsWhenAllEndpointsAreFading` to check that (LBAlgorithm).Apply finishes for any LBAlgorithm.

loadbalancer/fadein_test.go
```diff
@@ -190,6 +191,151 @@ func TestFadeIn(t *testing.T) {
...
+}
+
+func testFadeInLoadBetweenOldEps(
+	t *testing.T,
+	name string,
+	algorithm func([]string) routing.LBAlgorithm,
+	endpointAges ...time.Duration,
+) {
+	t.Run(name, func(t *testing.T) {
+		old := 2 * fadeInDurationHuge
```
RomanZavodskikh, line 209: This code looks kinda repetitive. Maybe I have some way to use it only once.

loadbalancer/fadein_test.go
```diff
@@ -42,41 +43,46 @@ func checkMonotony(direction, prev, next int) bool {
 	}
 }
 
+func initializeEndpoints(algorithm func([]string) routing.LBAlgorithm, endpointAges []time.Duration, fadeInDuration time.Duration) (*routing.LBContext, []string) {
```
AlexanderYastrebov, line 46: algorithm looks unused

loadbalancer/fadein_test.go
```diff
@@ -190,6 +196,100 @@ func TestFadeIn(t *testing.T) {
...
+		ctx, eps := initializeEndpoints(algorithm, endpointAges, fadeInDurationHuge)
+
+		a := algorithm(eps)
+		rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
+		t.Log("test start", time.Now())
+		var stats []string
+		stop := time.After(3 * time.Second)
+		// Emulate the load balancer loop, sending requests to it with random hash keys
+		// over and over again till fadeIn period is over.
+		func() {
```
AlexanderYastrebov, line 225: Why do we need this func() wrapper?

loadbalancer/fadein_test.go
```diff
@@ -190,6 +196,100 @@ func TestFadeIn(t *testing.T) {
...
+					return
+				default:
+				}
+			}
+		}()
+
+		nReqs := map[string]int{}
+		for _, ep := range stats {
+			nReqs[ep]++
+		}
```
AlexanderYastrebov, line 241: Why not count hosts directly in the loop (i.e. get rid of stats)?

loadbalancer/fadein_test.go
RomanZavodskikh: Looks like it's the one of the way to make break outer loop inside switch statement, like this:
https://stackoverflow.com/questions/11104085/does-a-break-statement-break-from-a-switch-select

loadbalancer/fadein_test.go
RomanZavodskikh: It was. :-)

loadbalancer/fadein_test.go
```diff
@@ -42,59 +43,64 @@ func checkMonotony(direction, prev, next int) bool {
...
-			for {
-				ctx.Params[ConsistentHashKey] = strconv.Itoa(rnd.Intn(100000))
-				ep := a.Apply(ctx)
-				stats = append(stats, ep.Host)
-				select {
-				case <-stop:
-					return
-				default:
-				}
+	outer:
```
AlexanderYastrebov, line 93: I see now why func was used - to break out of outer loop.

loadbalancer/fadein_test.go
```diff
@@ -42,59 +43,64 @@ func checkMonotony(direction, prev, next int) bool {
...
+func initializeEndpoints(endpointAges []time.Duration, fadeInDuration time.Duration) (*routing.LBContext, []string) {
+	var detectionTimes []time.Time
+	now := time.Now()
+	for _, ea := range endpointAges {
+		detectionTimes = append(detectionTimes, now.Add(-ea))
+	}
+
+	var eps []string
+	for i := range endpointAges {
+		eps = append(eps, string('a'+rune(i)))
```
RomanZavodskikh, line 55: Maybe change to 'http://foo-i.test'.

loadbalancer/fadein_test.go
RomanZavodskikh, line 93: @RomanZavodskikh , you can revert this to make diff smaller.

loadbalancer/fadein_test.go
```diff
@@ -190,6 +196,95 @@ func TestFadeIn(t *testing.T) {
...
+) {
+	t.Run(name, func(t *testing.T) {
+		old := 2 * fadeInDurationHuge
+		new := time.Duration(0)
+		ctx, eps := initializeEndpoints(endpointAges, fadeInDurationHuge)
+
+		a := algorithm(eps)
+		rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
+		nReqs := map[string]int{}
+		stop := time.After(3 * time.Second)
```
RomanZavodskikh, line 221: You had an idea than "3 * time.Second" is less that "24 * time.Hour".
So, it makes sense to move both constants to the beginning of this file and explicitly tell that
one constant << another constant there.

RomanZavodskikh, line 221: BTW Make a comment about this "3 * time.Second << 24 * time.Hour" assumption here.

loadbalancer/fadein_test.go
```diff
@@ -190,6 +196,95 @@ func TestFadeIn(t *testing.T) {
...
+
+		max := 0
+		for _, v := range nReqs {
+			if v > max {
+				max = v
+			}
+		}
+
+		for idx, ep := range eps {
+			if endpointAges[idx] == old && !closeToMax(nReqs[ep], max) {
```
RomanZavodskikh, line 246: Consider rewriting this to ...
```
if old {
...
} else {
// new
...

}
```

RomanZavodskikh, line 246: Consider using testify library here. It has InRange function.

loadbalancer/fadein_test.go
```diff
@@ -190,6 +196,95 @@ func TestFadeIn(t *testing.T) {
...
+		a := algorithm(eps)
+		rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
+		nReqs := map[string]int{}
+		stop := time.After(3 * time.Second)
+
+		t.Log("test start", time.Now())
+		// Emulate the load balancer loop, sending requests to it with random hash keys
+		// over and over again till fadeIn period is over.
+	outer:
+		for {
```
RomanZavodskikh, line 227: consider making huge number of request (for example 10000 reqs) instead of time-based test.

loadbalancer/fadein_test.go
```diff
@@ -190,6 +196,95 @@ func TestFadeIn(t *testing.T) {
...
+
+func closeToMax(arg int, max int) bool {
+	return arg > 90*max/100 && arg <= max
+}
+
+func testFadeInLoadBetweenOldEps(
+	t *testing.T,
+	name string,
+	algorithm func([]string) routing.LBAlgorithm,
+	endpointAges ...time.Duration,
```
RomanZavodskikh, line 211: consider using nOld, nNew instead of endpointAges.

loadbalancer/fadein_test.go
```diff
@@ -190,6 +196,95 @@ func TestFadeIn(t *testing.T) {
...
+	nEndpoints int,
+) {
+	// Initialize every endpoint with zero: every endpoint is new
+	endpointAges := make([]time.Duration, nEndpoints)
+
+	ctx, eps := initializeEndpoints(endpointAges, fadeInDurationHuge)
+
+	a := algorithm(eps)
+	rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
+	ctx.Params[ConsistentHashKey] = strconv.Itoa(rnd.Intn(100000))
```
RomanZavodskikh, line 276: As this parameter is used one time, you can use constant here (or remove at all).

loadbalancer/fadein_test.go
```diff
@@ -190,6 +196,95 @@ func TestFadeIn(t *testing.T) {
...
+	a.Apply(ctx)
```
RomanZavodskikh, line 277: consider using channel and separate goroutine here to explicitly state timeout here.

loadbalancer/fadein_test.go
RomanZavodskikh: I think it is better to leave this code as it is at the moment, because rewriting it in the manner from previous comment will make one more level of indentation, which I don't like.

loadbalancer/fadein_test.go
```diff
@@ -9,13 +9,17 @@ import (
...
 
 const (
-	fadeInDuration    = 500 * time.Millisecond
-	bucketCount       = 20
-	monotonyTolerance = 0.4 // we need to use a high tolerance for CI testing
+	fadeInDuration     = 500 * time.Millisecond
+	fadeInDurationHuge = 24 * time.Hour // we need this to be sure we're at the very beginning of fading in
+	applyTimeout       = 2 * time.Millisecond
+	numberOfReqs       = 100000
+	bucketCount        = 20
```
AlexanderYastrebov, line 21: Please put these into the only test where they're used.

loadbalancer/fadein_test.go
```diff
@@ -178,16 +198,109 @@ func TestFadeIn(t *testing.T) {
...
+		t.Log("test start", time.Now())
+		// Emulate the load balancer loop, sending requests to it with random hash keys
+		// over and over again till fadeIn period is over.
+		for i := 0; i < numberOfReqs; i++ {
+			ctx.Params[ConsistentHashKey] = strconv.Itoa(rnd.Intn(100000))
+			ep := a.Apply(ctx)
+			nReqs[ep.Host]++
+		}
+
+		oldEpsExpectedLoad := numberOfReqs / nOld
```
AlexanderYastrebov, line 245: maybe `expectedReqsPerOldEndpoint`

loadbalancer/fadein_test.go
```diff
@@ -178,16 +198,109 @@ func TestFadeIn(t *testing.T) {
...
+		for idx, ep := range eps {
+			if endpointAges[idx] == old {
+				assert.InEpsilon(t, oldEpsExpectedLoad, nReqs[ep], 0.2)
+			}
+			if endpointAges[idx] == new {
+				assert.InDelta(t, 0, nReqs[ep], float64(oldEpsExpectedLoad)*0.05)
```
AlexanderYastrebov, line 251: Shouldn't this just be close to zero? I.e. why do we use oldEpsExpectedLoad for _new_ endpoints?

RomanZavodskikh: I need somehow to define *absolute* error (because relative doesn't work for zero). Decided to create one more constant `acceptableErrorNearZero`.

loadbalancer/fadein_test.go
```diff
@@ -42,41 +44,52 @@ func checkMonotony(direction, prev, next int) bool {
...
+	now := time.Now()
+	for _, ea := range endpointAges {
+		detectionTimes = append(detectionTimes, now.Add(-ea))
+	}
+
+	var eps []string
+	for i := range endpointAges {
+		name := ""
+		if endpointAges[i] == 0 {
+			name = fmt.Sprintf("new-%d.test", i)
```
AlexanderYastrebov, line 58: Could also use `endpointAges[i].String()` to form the name

AlexanderYastrebov: eps = append(eps, fmt.Sprintf("ep-%s.test", endpointAges[i]))

AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

Issue #1865, opened 2021-09-23
Graceful shutdown is not working in HTTPS mode


AlexanderYastrebov:
**Describe the bug**

#988 implemented graceful shutdown but it is not working in HTTPS mode because of an early return:
https://github.com/zalando/skipper/blob/27f48492393ba36ae3ef4d6000f0e0a454a161af/skipper.go#L1053

**To Reproduce**

1. Run Skipper in HTTPS mode
```
$ bin/skipper -tls-cert=fixtures/test.crt -tls-key=fixtures/test.key -inline-routes '* -> status(204) -> <shunt>'
```
2. Send SIGTERM
```
$ kill -SIGTERM $(pgrep skipper)
```
3. Observe immediate shutdown

**Expected behavior**

Graceful shutdown like in HTTP mode, e.g. the logs:
```
Got shutdown signal, wait 45s for health check
```

PR #988, opened 2019-03-12
Fix/graceful shutdown

szuecs:
In combination with kube-ingress-aws-controller we seem to have again a problem on graceful shutdown with recent Kubernetes versions. I remember testing this before and it was reliably working, but somehow we introduced a bug on tear down, which this PR fixes. I added a test to test clean graceful shutdown with ongoing requests.

Probably fixes #964, too.

Manually tested in Kubernetes with vegeta and 50 req/s to a cluster with 3 skipper-ingress in a 3 AZ cluster with ALB infront created by kube-ingress-aws-controller.
Now even `kubectl -n kube-system delete pods -l application=skipper-ingress` is safe to run.

Manually tested on localhost

Backend 1m latency:
```
% ./bin/skipper -inline-routes '* -> latency("1m") -> inlineContent("Hello from 127.0.0.1:9090") -> <shunt>' -support-listener :9111
[APP]INFO[0000] Expose metrics in codahale format
[APP]INFO[0000] support listener on :9111
[APP]INFO[0000] proxy listener on :9090
[APP]INFO[0000] TLS settings not found, defaulting to HTTP
[APP]INFO[0000] route settings, reset, route: : * -> latency("1m") -> inlineContent("Hello from 127.0.0.1:9090") -> <shunt>
[APP]INFO[0000] route settings received
[APP]INFO[0000] route settings applied

127.0.0.1 - - [12/Mar/2019:09:51:35 +0100] "GET / HTTP/1.1" 200 25 "-" "curl/7.49.0" 60000 127.0.0.1:9090 - -
```

lb-skipper in front:
```
./bin/skipper -address=:9999 -serve-host-metrics -route-backend-metrics -experimental-upgrade -inline-routes='* -> <roundRobin, "http://127.0.0.1:9998", "http://127.0.0.1:9090">'
```

client started before SIGINT to lb-skipper:
```
curl -v http://127.0.0.1:9999/                                  suess-new
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 9999 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:9999
> User-Agent: curl/7.49.0
> Accept: */*
>
< HTTP/1.1 200 OK
< Content-Length: 25
< Content-Type: text/plain; charset=utf-8
< Date: Tue, 12 Mar 2019 08:59:53 GMT
< Server: Skipper
< Connection: close
<
* Closing connection 0
Hello from 127.0.0.1:9090
```

[...]

PR #2462, opened 2023-07-15
skipper: fix shutdown test

AlexanderYastrebov:
Make test request in parallel to ensure another request is refused not only after shutdown started but also while shutdown is still in progress (i.e. while test request is still in progress).

Verify shutdown completes after test request is finished.

Rename tests to have the same prefix.

Updates #988 #1865

### DIFF
--- a/skipper_test.go
+++ b/skipper_test.go
@@ -283,17 +283,22 @@ func TestHTTPServer(t *testing.T) {
 	}
 }
 
-func TestHTTPServerShutdown(t *testing.T) {
+func TestServerShutdownHTTP(t *testing.T) {
 	o := &Options{}
 	testServerShutdown(t, o, "http")
 }
 
-func TestHTTPSServerShutdown(t *testing.T) {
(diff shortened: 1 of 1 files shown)


skipper_test.go
```diff
@@ -329,22 +335,39 @@ func testServerShutdown(t *testing.T, o *Options, scheme string) {
 
 	time.Sleep(shutdownDelay / 2)
 
-	t.Logf("ongoing request passing in before shutdown")
-	r, err := waitConnGet(testUrl)
-	require.NoError(t, err)
-	require.Equal(t, 200, r.StatusCode)
```
AlexanderYastrebov, line 335: Here was the problem: request takes 3s so at this points it is 3.5 seconds after shutdown started. The subsequent request surely fails because servers has completed shutdown after this.

MustafaSaber:
👍

MustafaSaber (commented):
I like how logs describe the flow

AlexanderYastrebov:
:+1:

PR #2469, opened 2023-07-17
packaging: fix Dockerfile to support dependabot updates

AlexanderYastrebov:
Dependabot does not update ARG instruction and updates only the first FROM instruction, see https://github.com/dependabot/dependabot-core/issues/2057
This change uses workaround described here https://github.com/dependabot/dependabot-core/issues/2057#issuecomment-1351660410


MustafaSaber:
👍

AlexanderYastrebov:
:+1:

PR #2475, opened 2023-07-18
skipper: do not log nil error after shutdown

AlexanderYastrebov:
This disables `level=fatal msg="<nil>"` message after succesfull shutdown.


AlexanderYastrebov:
:+1:

MustafaSaber:
👍

Issue #1618, opened 2020-11-24
validation webhook did not caught miss spelled filter


szuecs:
**Describe the bug**

RouteGroup created
```yaml
spec:
  backends:
  - address: https://app.example.org/
    name: app
    type: network
  defaultBackends:
  - backendName: app
  hosts:
  - my-host.example.org
  routes:
  - filters:
    - bearerInjector("app/app-secret") # <- filter does not exist (camelcase here)
    pathSubtree: /
```

**Expected behavior**

This should not be an allowed change

**Observed behavior**

It was submitted without error

AlexanderYastrebov:
I guess the proper validation should
1. parse eskip and check there is a single filter (doable)
2. check filter name (seems doable, needs filter registry)
3. check args (seems tricky as it would need to actually create filter which might be impossible (e.g. `static` filter checks directory permissions) )

somewhere along the https://github.com/zalando/skipper/blob/807227afddc22b17d6291458c0944d0a251d5ae8/dataclients/kubernetes/definitions/routegroups.go#L397-L407

The same is for predicates. Maybe 1. and 2. could be a good start.

AlexanderYastrebov:
Linking the comment https://github.com/zalando/skipper/issues/1537#issuecomment-744544223 from @aryszka touching route validation

AlexanderYastrebov:
Related issue https://github.com/zalando/skipper/issues/1333

AlexanderYastrebov:
Related issue https://github.com/zalando/skipper/issues/1333

MustafaSaber:
I see 2 problems with point 2 in this comment https://github.com/zalando/skipper/issues/1618#issuecomment-733952158

1. We will need to export filter registry somehow from Skipper and will break admission webhook interface and we won't be able to send it to anyway.
     1. https://github.com/zalando/skipper/blob/master/cmd/webhook/main.go#L58 ->
     2. https://github.com/zalando/skipper/blob/master/cmd/webhook/admission/admission.go#L106 -> 
     3. https://github.com/zalando/skipper/blob/master/cmd/webhook/admission/admission.go#L72 -> 
     4. https://github.com/zalando/skipper/blob/master/cmd/webhook/admission/admission.go#L87 (where validation happens)
     How can we move filterRegistry from `i` to `iv` if it's created in skipper? 

2. We don't have predicate registry so checking predicate name only also not valid

MustafaSaber:
We need the routing tree so some ideas:

[...]

PR #2478, opened 2023-07-21
[experimental] Move admission webhook into skipper for better validation

MustafaSaber:
This moves admission webhook binary into skipper to later initiate filters/predicates and make more strict validations.

Efforts to solve this https://github.com/zalando/skipper/issues/1618


skipper.go
```diff
@@ -1959,6 +1976,16 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	routing := routing.New(ro)
 	defer routing.Close()
 
+	if o.EnableValidationWebhook {
```
RomanZavodskikh, line 1979: Here I see the code starting some alternative mode of skipper binary. So I would expect something like `routesrv.Run()` call nearby. Could you please kindly show me where it is? Is it possible to put it near this line of code?

MustafaSaber: In the next line you can see 'webhook.Run()'

RomanZavodskikh: Yes, I see `webhook.Run()`, but I would expect to see `routesrv.Run()` too. Both of them, basically.
From my side it makes some sense, because both "webhook" and "routesrv" are some "custom mode" of running skipper binary.

MustafaSaber: why should they run together? Every mode is a separate one and everyone run it's own required place and own required options

MustafaSaber:
This was done by other set of PRs from @vlktna

PR #2479, opened 2023-07-24
predicates: add Config predicate

AlexanderYastrebov:
E.g. it could be used to let skipper instance know in which zone it is deployed:
```yaml
containers:
  - env:
      - name: TOPOLOGY_ZONE
        valueFrom:
          fieldRef:
            fieldPath: metadata.annotations['topology.kubernetes.io/zone']
    args:
      - skipper
      - -config-predicate-values=zone=$(TOPOLOGY_ZONE)
      - ...
```

such that it only enables specific route for that zone:
```
za: Path("/foo") && Config("zone", "a") -> "http://backend.zone-a.test";
zb: Path("/foo") && Config("zone", "b") -> "http://backend.zone-b.test";
zc: Path("/foo") && Config("zone", "c") -> "http://backend.zone-c.test";
```

- [ ] tests please

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -66,6 +66,7 @@ type Config struct {
 	DataclientPlugins               *pluginFlag    `yaml:"dataclient-plugin"`
 	MultiPlugins                    *pluginFlag    `yaml:"multi-plugin"`
 	CompressEncodings               *listFlag      `yaml:"compress-encodings"`
+	ConfigPredicateValues           mapFlags       `yaml:"config-predicate-values"`
 
 	// logging, metrics, profiling, tracing:
 	EnablePrometheusMetrics             bool      `yaml:"enable-prometheus-metrics"`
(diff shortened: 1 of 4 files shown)


szuecs (commented):
Why do we need this as predicate?

AlexanderYastrebov:
Zone awareness is one example but predicate is generic and covers all kind of usecases where operator wants to customize the same set of routes based on the particular instance properties.

It is an experiment and I was not going to push it forward until we see some other usecases.

szuecs:
Ok, right now I don't see a case that we want to have a predicate for this. AZ aware routing should be done differently by leveraging endpointslices. If someone has a non-kubernetes related use case, we can add this or if someone has another idea of Config() usage.

AlexanderYastrebov:
I'll close for now.

Issue #2488, opened 2023-07-31
Remove use of github.com/pkg/errors


AlexanderYastrebov:
See https://github.com/pkg/errors/issues/245

We can probably replace it by https://pkg.go.dev/errors#Join introduced in 1.20

AlexanderYastrebov:
After https://github.com/zalando/skipper/pull/2492 we still have it as indirect
```
$ go mod why -m github.com/pkg/errors
# github.com/pkg/errors
github.com/zalando/skipper/tracing/tracers/jaeger
github.com/uber/jaeger-client-go/config
github.com/pkg/errors
```
Related to https://github.com/zalando/skipper/issues/2153 and https://github.com/zalando/skipper/issues/2104

AlexanderYastrebov:
If jaeger dependency is removed then 
```
$ go mod why -m github.com/pkg/errors
# github.com/pkg/errors
github.com/zalando/skipper/net/redistest
github.com/testcontainers/testcontainers-go
github.com/docker/docker/api/types/registry
github.com/pkg/errors
```

AlexanderYastrebov:
There are other dependencies, namely OPA, using github.com/pkg/errors:
```
$ go mod graph | grep github.com/pkg/errors
github.com/zalando/skipper github.com/pkg/errors@v0.9.1
github.com/containerd/containerd@v1.7.4 github.com/pkg/errors@v0.9.1
github.com/dgraph-io/badger/v3@v3.2103.5 github.com/pkg/errors@v0.9.1
github.com/dgraph-io/ristretto@v0.1.1 github.com/pkg/errors@v0.9.1
github.com/open-policy-agent/opa@v0.56.0 github.com/pkg/errors@v0.9.1
github.com/open-policy-agent/opa-envoy-plugin@v0.56.0-envoy github.com/pkg/errors@v0.9.1
github.com/opencontainers/image-spec@v1.1.0-rc4 github.com/pkg/errors@v0.9.1
github.com/testcontainers/testcontainers-go@v0.23.0 github.com/pkg/errors@v0.9.1
```

PR #2484, opened 2023-07-26
Create `validator` object to give more validation flexibility

MustafaSaber:
This will help with https://github.com/zalando/skipper/pull/2478 to try improving validation and how to get around some `DataClients` tests

Also this run `eskip.ParseFilters` by default so that's a step forward.

Related to https://github.com/zalando/skipper/issues/1618

PR #2492, opened 2023-08-01
Fix wrapping errors

MustafaSaber:
Updates #2488
For #2484

`err = errors.Wrap(err, nerr.Error())` also seems wrong as [if error is `nil` it returns `nil`](https://github.com/pkg/errors/blob/5dd12d0cfe7f152f80558d591504ce685299311e/errors.go#L184-L187) 

Running `go mod tidy` won't remove `github.com/pkg/errors` as `cron` predicate uses https://pkg.go.dev/github.com/sarslanhan/cronmask which uses it (fixed with sarslanhan/cronmask#3) and now `github.com/uber/jaeger-client-go/config`, maybe will create a PR there to remove it.


dataclients/kubernetes/definitions/ingressv1.go
```diff
@@ -101,15 +101,10 @@ func ValidateIngressesV1(ingressList IngressV1List) error {
 			name := i.Metadata.Name
 			namespace := i.Metadata.Namespace
 			nerr = fmt.Errorf("%s/%s: %w", name, namespace, nerr)
-			err = errors.Wrap(err, nerr.Error())
+			err = errorsJoin(err, nerr)
```
AlexanderYastrebov, line 104: Lets update `// ValidateIngresses is a no-op` comment as its definitely not a no-op

dataclients/kubernetes/definitions/routegroups.go
```diff
@@ -463,15 +464,7 @@ func ValidateRouteGroups(rl *RouteGroupList) error {
...
-		}
-	}
-
-	if err != nil {
-		return err
+		err = errorsJoin(err, ValidateRouteGroup(i))
 	}
-
-	return nil
+	return err
```
AlexanderYastrebov, line 469: This should be like
```go
var errs []error
for _, i := range rl.Items {
    errs = append(errs, ValidateRouteGroup(i))
}
return errorsJoin(errs...)
```

dataclients/kubernetes/definitions/definitions.go
```diff
@@ -43,3 +44,16 @@ type WeightedBackend interface {
 	GetName() string
 	GetWeight() float64
 }
+
+func errorsJoin(errs ...error) error {
```
AlexanderYastrebov, line 48: So its only used from the methods that are never called by the Skipper.
We should add a test for it.

AlexanderYastrebov:
> Running go mod tidy won't remove github.com/pkg/errors as cron predicate uses

Changed `fixes #2488` to `Updates #2488`

dataclients/kubernetes/definitions/definitions.go
MustafaSaber: Actually they aren't called anywhere

AlexanderYastrebov: Makes sense to note that we are adding this because we can not use https://pkg.go.dev/errors#Join before 1.21 release.
If we stick to "\n" delimiter then we can add a TODO to use stdlib errors.Join

AlexanderYastrebov:
> Running go mod tidy won't remove github.com/pkg/errors as cron predicate uses

Also created https://github.com/sarslanhan/cronmask/issues/2

MustafaSaber:
Created https://github.com/sarslanhan/cronmask/pull/3

dataclients/kubernetes/definitions/definitions.go
AlexanderYastrebov: PR message (or better commit message) should note that original `err = errors.Wrap(err, nerr.Error())` is wrong because it returns nil when err is nil.

dataclients/kubernetes/definitions/routegroups.go
MustafaSaber: both should work actually.

AlexanderYastrebov: Yes, but then we decide to change `errors.New(strings.Join(errVals, "\n"))` to `fmt.Errorf("multiple errors: %s", strings.Join(errVals, "\n"))` and message becomes screwed up.
The helper as well as errors.Join is designed to handle multiple errors, not two so lets use it.

dataclients/kubernetes/definitions/ingressv1.go
MustafaSaber: It's actually no-op

MustafaSaber: `ValidateIngressV1` always return `nil`.

dataclients/kubernetes/definitions/testdata/errorwrapdata/routegroups.json
```diff
@@ -0,0 +1,41 @@
+{
+    "items": [
```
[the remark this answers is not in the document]
AlexanderYastrebov: In yaml you can also put a comment telling why routegroup is not valid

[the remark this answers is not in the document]
MustafaSaber: But json is easier for validation webhook

dataclients/kubernetes/definitions/definitions_test.go
```diff
@@ -1,11 +1,27 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/zalando/skipper/dataclients/kubernetes/definitions"
 	"github.com/zalando/skipper/dataclients/kubernetes/kubernetestest"
 )
 
 func TestRouteGroupValidation(t *testing.T) {
 	kubernetestest.FixturesToTest(t, "testdata/validation")
 }
+
+func TestRouteGroupsValidationErrorWrapping(t *testing.T) {
```
AlexanderYastrebov, line 17: This is testing ValidateRouteGroups that is not covered by TestRouteGroupValidation so it should be `TestValidateRouteGroups`. Error wrapping is a detail.

dataclients/kubernetes/definitions/ingressv1.go
AlexanderYastrebov: This also should collect errors in a slice first.

AlexanderYastrebov: Please fix the comment to `// ValidateIngressesV1 validates an IngressV1List` or simply remove it

dataclients/kubernetes/definitions/testdata/errorwrapdata/routegroups.yaml
```diff
@@ -0,0 +1,21 @@
...
+    kind: RouteGroup
+    metadata:
+        name: rg1
+    spec:
+        hosts:
+        - test.example.com
+        routes:
+        - filters:
+            - inlineContent("/foo")
+        - path: /foo
```
AlexanderYastrebov, line 21: Please add a newline. Makes sense to add a comment what's wrong with these routegroups.

dataclients/kubernetes/definitions/routegroups.go
```diff
@@ -121,7 +122,7 @@ type skipperBackendParser struct {
 
 type BackendReference struct {
 	// BackendName references the skipperBackend by name
-	BackendName string `json:"backendName"`
+	BackendName string `json:"backendName" yaml:"backendName"`
```
MustafaSaber, line 125: Interesting why is this the only field that was making a problem

dataclients/kubernetes/definitions/testdata/errorwrapdata/routegroups.yaml
MustafaSaber: This should be resolved.

dataclients/kubernetes/definitions/ingressv1.go
MustafaSaber: This should be resolved

dataclients/kubernetes/definitions/routegroups.go
AlexanderYastrebov: K8s decided not to support yaml tags as that is a hassle to have a tag per each serialization format. Instead they decided to convert yaml to json first and then use json deserializer, see https://pkg.go.dev/sigs.k8s.io/yaml#section-readme for more details.

For us this means that definitions.ParseRouteGroupsYAML (again not used anywhere) is broken. OTOH since we know it is broken it means noone uses it so we can remove.

AlexanderYastrebov: To load your testcase without adding yaml tags you can use ghodss/yaml that we import as a dependency.
I am also not sure how existing yaml test fixtures work without yaml tag and `(sb *SkipperBackend) UnmarshalYAML`, please check.

MustafaSaber: > not sure how existing yaml test fixtures work without yaml tag

Because we load json
https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/clusterclient.go#L355-L359

MustafaSaber: I think currently we have 2 directions

- Use json in the test
- Try to create the test under `testFixtures` stuff to also use k8s dataclient
- Fix Yaml marshalling

I would stick with first option

AlexanderYastrebov: > Because we load json

but fixtures are yamls :)

AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

PR #2493, opened 2023-08-03
webhook: validate skipper annotations in ingress

MustafaSaber:
This should introduce the custom `ValidationWebhook` for skipper ingress annotations (e.g `zalando.org/skipper-routes`, `zalando.org/skipper-filter` and `zalando.org/skipper-predicate`), currently we skip the annotation if something is wrong from skipper side at https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/ingress.go#L227-L271

This is also a followup to https://github.com/zalando/skipper/issues/1618


dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -0,0 +1,68 @@
...
+	skipperfilterAnnotationKey    = "zalando.org/skipper-filter"
+	skipperpredicateAnnotationKey = "zalando.org/skipper-predicate"
+	skipperRoutesAnnotationKey    = "zalando.org/skipper-routes"
+)
+
+type IngressV1Validator struct{}
+
+var defaultIngressV1Validator = IngressV1Validator{}
+
+func ValidateIngressV1(item *IngressV1Item) error {
```
MustafaSaber, line 19: No tests added in skipper for now since we don't validate there, will try to add some.

cmd/webhook/admission/ingressadmission.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (IngressAdmitter) name() string {
+	return "ingress"
+}
+
+func (IngressAdmitter) admit(req *admissionRequest) (*admissionResponse, error) {
+	ingressItem := definitions.IngressV1Item{}
+	err := json.Unmarshal(req.Object, &ingressItem)
+	if err != nil {
+		emsg := fmt.Sprintf("could not parse Ingress, %v", err)
```
szuecs, line 22: Message starts with upper case and use `: %v` to highlight error message

cmd/webhook/admission/ingressadmission.go
```diff
@@ -0,0 +1,50 @@
...
+			Allowed: false,
+			Result: &status{
+				Message: emsg,
+			},
+		}, nil
+	}
+
+	err = definitions.ValidateIngressV1(&ingressItem)
+	if err != nil {
+		emsg := fmt.Sprintf("Ingress validation failed, %v", err)
```
szuecs, line 35: Message use `: %v` to highlight error message

cmd/webhook/admission/rgadmission.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (RouteGroupAdmitter) name() string {
+	return "routegroup"
+}
+
+func (RouteGroupAdmitter) admit(req *admissionRequest) (*admissionResponse, error) {
+	rgItem := definitions.RouteGroupItem{}
+	err := json.Unmarshal(req.Object, &rgItem)
+	if err != nil {
+		emsg := fmt.Sprintf("could not parse RouteGroup, %v", err)
```
szuecs, line 22: same as https://github.com/zalando/skipper/pull/2493/files#r1285988263

cmd/webhook/admission/rgadmission.go
```diff
@@ -0,0 +1,50 @@
...
+			Allowed: false,
+			Result: &status{
+				Message: emsg,
+			},
+		}, nil
+	}
+
+	err = definitions.ValidateRouteGroup(&rgItem)
+	if err != nil {
+		emsg := fmt.Sprintf("could not validate RouteGroup, %v", err)
```
szuecs, line 35: same as https://github.com/zalando/skipper/pull/2493/files#r1285988683

cmd/webhook/admission/testdata/ingress/invalid-filters-and-predicates.json
```diff
@@ -0,0 +1,40 @@
...
+    "request": {
+        "uid": "req-uid",
+        "name": "req1",
+        "namespace": "n1",
+        "object": {
+            "metadata": {
+                "name": "ing1",
+                "namespace": "ing1",
+                "annotations": {
+                    "zalando.org/skipper-filter": "status&(10) -> inlineContent(\"This should work\")",
```
szuecs, line 11: Is `status&(10)` by intention with `&`? If so why?

cmd/webhook/admission/testdata/ingress/invalid-filters.json
```diff
@@ -0,0 +1,39 @@
...
+    "request": {
+        "uid": "req-uid",
+        "name": "req1",
+        "namespace": "n1",
+        "object": {
+            "metadata": {
+                "name": "ing1",
+                "namespace": "ing1",
+                "annotations": {
+                    "zalando.org/skipper-filter": "status&(10) -> inlineContent(\"This should't work\")"
```
szuecs, line 11: same as https://github.com/zalando/skipper/pull/2493/files#r1285992635

cmd/webhook/admission/testdata/ingress/valid-ingress-with-annotations.json
```diff
@@ -0,0 +1,41 @@
...
+        "name": "req1",
+        "namespace": "n1",
+        "object": {
+            "metadata": {
+                "name": "ing1",
+                "namespace": "ing1",
+                "annotations": {
+                    "zalando.org/skipper-filter": "status(10) -> inlineContent(\"This should work\")",
+                    "zalando.org/skipper-predicate": "Header(\"test\", \"test\") && Path(\"/login\")",
+                    "zalando.org/skipper-routes": "r1: Header(\"test2\", \"test2\") && Path(\"/login\") -> status(200) -> \"shunt\""
```
szuecs, line 13: Why is `"shunt"` not wrong?

cmd/webhook/admission/testdata/ingress/invalid-filters-and-predicates.json
AlexanderYastrebov: I think we should use something more explicit as a wring eskip, e.g. `this-is-invalid(filter)`

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -0,0 +1,68 @@
...
+
+const (
+	skipperfilterAnnotationKey    = "zalando.org/skipper-filter"
+	skipperpredicateAnnotationKey = "zalando.org/skipper-predicate"
+	skipperRoutesAnnotationKey    = "zalando.org/skipper-routes"
+)
+
+type IngressV1Validator struct{}
+
+var defaultIngressV1Validator = IngressV1Validator{}
```
szuecs, line 17: hmm why do we use a package global here?

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -0,0 +1,68 @@
...
+	var errs []error
+
+	errs = append(errs, defaultIngressV1Validator.validateFilterAnnotation(item.Metadata.Annotations))
+	errs = append(errs, defaultIngressV1Validator.validatePredicateAnnotation(item.Metadata.Annotations))
+	errs = append(errs, defaultIngressV1Validator.validateRoutesAnnotation(item.Metadata.Annotations))
+
+	return errorsJoin(errs...)
+}
+
+func (IngressV1Validator) validateFilterAnnotation(annotations map[string]string) error {
```
szuecs, line 46: If you don't need to be a "method" you can change it to be a "pure" func.

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -0,0 +1,68 @@
...
+	if filters, ok := annotations[skipperfilterAnnotationKey]; ok {
+		_, err := eskip.ParseFilters(filters)
+		return err
+	}
+	return nil
+}
+
+func (IngressV1Validator) validatePredicateAnnotation(annotations map[string]string) error {
```
szuecs, line 54: same as https://github.com/zalando/skipper/pull/2493/files#r1286008810

cmd/webhook/admission/testdata/ingress/valid-ingress-with-annotations.json
MustafaSaber: Why should it? It's a valid eskip, right now that's what we do

dataclients/kubernetes/definitions/ingressvalidator.go
MustafaSaber: Because later we will need to use this in webhook & ingress definitions and maybe custmoize with filter registry and so on

dataclients/kubernetes/definitions/ingressvalidator.go
MustafaSaber: A method will be needed once we add filter/predicate registry

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -0,0 +1,58 @@
+package definitions
+
+import (
+	"fmt"
+
+	"github.com/zalando/skipper/eskip"
+)
+
+const (
+	skipperfilterAnnotationKey    = "zalando.org/skipper-filter"
```
MustafaSaber, line 10: Should we export those variables from https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/ingress.go#L23-L25 instead of duplication

cmd/webhook/admission/ingressadmission.go
```diff
@@ -0,0 +1,50 @@
...
+func (IngressAdmitter) name() string {
+	return "ingress"
+}
+
+func (IngressAdmitter) admit(req *admissionRequest) (*admissionResponse, error) {
+	ingressItem := definitions.IngressV1Item{}
+	err := json.Unmarshal(req.Object, &ingressItem)
+	if err != nil {
+		emsg := fmt.Sprintf("Could not parse Ingress: %v", err)
+		log.Error(emsg)
```
AlexanderYastrebov, line 22: Maybe we don't need this as user will see the error.

cmd/webhook/admission/ingressadmission.go
```diff
@@ -0,0 +1,50 @@
...
+		return &admissionResponse{
+			UID:     req.UID,
+			Allowed: false,
+			Result: &status{
+				Message: emsg,
+			},
+		}, nil
+	}
+
+	ingressValidatore := definitions.IngressV1Validator{}
```
AlexanderYastrebov, line 32: typo?

AlexanderYastrebov, line 32: Lets make a field of IngressAdmitter

cmd/webhook/admission/rgadmission.go
```diff
@@ -0,0 +1,50 @@
+package admission
```
AlexanderYastrebov, line 1: Let name the file `routegroup.go`

cmd/webhook/admission/testdata/ingress/valid-ingress-with-annotations.json
AlexanderYastrebov: Lets put a valid value here e.g. `http://foo.test` to reduce confusion, see also https://www.rfc-editor.org/rfc/rfc6761

> Why is "shunt" not wrong?

We'd need to tighten backend validation, lets tackle it in a separate PR.

cmd/webhook/admission/testdata/ingress/valid-ingress-with-annotations.json
```diff
@@ -0,0 +1,41 @@
...
+    "request": {
+        "uid": "req-uid",
+        "name": "req1",
+        "namespace": "n1",
+        "object": {
+            "metadata": {
+                "name": "ing1",
+                "namespace": "ing1",
+                "annotations": {
+                    "zalando.org/skipper-filter": "status(10) -> inlineContent(\"This should work\")",
```
AlexanderYastrebov, line 11: Make it status(200)

dataclients/kubernetes/definitions/ingressvalidator.go
AlexanderYastrebov: Maybe export it and move to this package (definitions).

cmd/webhook/admission/testdata/ingress/valid-ingress-with-annotations.json
szuecs: Yes we can do this but we should catch also these kind of missing URL, in case we assume `network` backend we should have a `scheme` and `host` (or IP). "shunt" might match as a network backend today, but this is not "valid" in a sense from user pov, so we should not have this in tests, that we want not to test "wrong backend".

cmd/webhook/admission/ingressadmission.go
```diff
@@ -0,0 +1,50 @@
+package admission
```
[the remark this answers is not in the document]
szuecs: I wonder if we should rename admission to validation, because we do not plan to change documents but to validate them.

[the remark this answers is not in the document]
AlexanderYastrebov: There are exported symbols. If we rename I propose we unexport everything to allow active development.

cmd/webhook/admission/routegroup.go
```diff
@@ -0,0 +1,51 @@
...
+	rgItem := definitions.RouteGroupItem{}
+	err := json.Unmarshal(req.Object, &rgItem)
+	if err != nil {
+		return &admissionResponse{
+			UID:     req.UID,
+			Allowed: false,
+			Result: &status{
+				Message: err.Error(),
+			},
+		}, err
```
MustafaSaber, line 33: We never returned error here with response, so metrics for rejected request was wrong.

See https://github.com/zalando/skipper/blob/master/cmd/webhook/admission/admission.go#L161-L167

dataclients/kubernetes/clusterclient.go
```diff
@@ -358,10 +358,11 @@ func (c *clusterClient) LoadRouteGroups() ([]*definitions.RouteGroupItem, error)
 		return nil, err
 	}
 
+	routeGroupValidator := &definitions.RouteGroupValidator{}
 	rgs := make([]*definitions.RouteGroupItem, 0, len(rgl.Items))
 	for _, i := range rgl.Items {
 		// Validate RouteGroup item.
-		if err := definitions.ValidateRouteGroup(i); err != nil {
+		if err := routeGroupValidator.Validate(i); err != nil {
```
MustafaSaber, line 365: This will break api, should I increase version?

cmd/webhook/admission/routegroup.go
AlexanderYastrebov: Generally we either should return either nil *admissionResponse or nil error. but never both nil or both set. On the caller side it is always
```go
foo, err := bar()
if err != nil {
  // handle error and return
}
// handle foo
```

AlexanderYastrebov: In this case and below I think it is valid to return nil error since admissionResponse would not allow and would contain error message.

dataclients/kubernetes/clusterclient.go
AlexanderYastrebov: I think we should make routeGroupValidator a field of clusterClient

dataclients/kubernetes/definitions/ingressv1.go
```diff
@@ -3,10 +3,17 @@ package definitions
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 	"strconv"
 )
 
+const (
+	SkipperfilterAnnotationKey    = "zalando.org/skipper-filter"
```
AlexanderYastrebov, line 10: We should highlight that these are Ingress annotations and I think we do not need to have "Key" suffix, so:
```
IngressFilterAnnotation
IngressPredicateAnnotation
IngressRoutesAnnotation
```

dataclients/kubernetes/clusterclient.go
AlexanderYastrebov: I just bumped minor version in https://github.com/zalando/skipper/pull/2505 so may you don't have to.

cmd/webhook/admission/routegroup.go
MustafaSaber: I don't need the response here yes thanks for the hint, it's also already handled in the Handler

AlexanderYastrebov:
:+1:

MustafaSaber:
👍

PR #2495, opened 2023-08-04
filters/cookie: allow expiration

### LINKED ISSUE #2494: No way to expire a cookie

There is no way to delete a cookie via `responseCookie("foo", "deleted", 0)` because 0 means no max-age attribute [net/http/cookie.go](https://github.com/golang/go/blob/74caf475e3b31a463637bfe3cbdb317674f0930f/src/net/http/cookie.go#L32-L35):
```go
	// MaxAge=0 means no 'Max-Age' attribute specified.
	// MaxAge<0 means delete cookie now, equivalently 'Max-Age: 0'
	// MaxAge>0 means Max-Age attribute present and given in seconds
	MaxAge   int
```
and eskip does not support negative numbers https://github.com/zalando/skipper/issues/194.

Note that `dropResponseCookie` filter removes header value and does not tell client to remove the cookie.
Note also that `responseCookie` derives cookie domain from the request host
https://github.com/zalando/skipper/blob/92e0d11af311949faf68d55635ec302c66d38b1c/filters/cookie/cookie.go#L281-L287
by removing one subdomain unconditionally:
https://github.com/zalando/skipper/blob/92e0d11af311949faf68d55635ec302c66d38b1c/filters/cookie/cookie.go#L300-L312

There should be a way to delete a cookie (at least created via `responseCookie` due to the way how domain is set)

Its possible to workaround the problem by
`appendResponseHeader("Set-Cookie", "language-preference=del; Path=/; Domain=foo.test; Max-Age=0; HttpOnly; Secure")`

In k8s ingress setting this could be tricky for multi-host RouteGroups/Ingresses because one can not use `Domain=${request.host}` due to the way how cookie domain is derived and also because `request.host` may contain port number (depends on the configuration, see related #1911)

https://stackoverflow.com/questions/5285940/correct-way-to-delete-cookies-server-side

Yes

AlexanderYastrebov:
Fixes #2494

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -1765,10 +1765,11 @@ requestCookie("test-session", "abc")
 
 ### responseCookie
 
-Appends cookies to responses in the "Set-Cookie" header. The response cookie
-accepts an optional argument to control the max-age property of the cookie,
-of type `int`, in seconds. The response cookie accepts an optional fourth
-argument, "change-only", to control if the cookie should be set on every
+Appends a cookie to the response via "Set-Cookie" header.
+It derives cookie domain by removing one subdomain from the request hostname domain.
+The filter accepts an optional argument to set the `Max-Age` attribute of the cookie, of type `int`, in seconds.
+Use zero to expire the cookie immediately.
+An optional fourth argument, "change-only", controls if the cookie should be set on every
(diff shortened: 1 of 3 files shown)


MustafaSaber:
👍

szuecs:
:+1:

PR #2499, opened 2023-08-06
Fix goroutine leaks in tests

AlexanderYastrebov:
Adds leak detector to more packages and fixes detected leaks.


metrics/codahale.go
```diff
@@ -234,6 +237,23 @@ func (c *CodaHale) IncErrorsStreaming(routeId string) {
 	}
 }
 
+func (c *CodaHale) Close() {
+	close(c.quit)
```
szuecs, line 241: Maybe better to also wrap it into a `once.Do` as we do in other cases, because it will panic to call it twice.

metrics/metrics.go
```diff
@@ -76,6 +76,7 @@ type Metrics interface {
 	IncErrorsStreaming(routeId string)
 	RegisterHandler(path string, handler *http.ServeMux)
 	UpdateGauge(key string, value float64)
+	Close()
```
szuecs, line 79: I suggest to add a comment into commit and release message and increase minor version. I guess nobody implements this, so not really breaking but who knows. :)

metrics/codahale.go
AlexanderYastrebov: yes, but OTOH it would uncover incorrect closing.

AlexanderYastrebov: Could also return error if its already closed.

metrics/metrics.go
AlexanderYastrebov: Will move metrics change into a separate PR.

AlexanderYastrebov:
:+1:

MustafaSaber:
👍

metrics/metrics.go
AlexanderYastrebov: #2505

Issue #2386, opened 2023-06-09
Flaky TestAdmissionControl


AlexanderYastrebov:
`TestAdmissionControl` fails quite often:
```
--- FAIL: TestAdmissionControl (7.01s)
    --- FAIL: TestAdmissionControl/small_error_rate_and_bigger_than_threshold_will_block_some_traffic (1.00s)
        admission_test.go:187: ok: 1312.00, fail: 278.00, failBackend: 357.00
        admission_test.go:193: Failed to get expected fails should be in: 61.65 < 278.00 < 256.35
```
E.g. https://github.com/zalando/skipper/actions/runs/5223927424/jobs/9431473370?pr=2385

szuecs:
Yeah I wonder why

AlexanderYastrebov:
Related #2319

AlexanderYastrebov:
```
--- FAIL: TestAdmissionControl (9.03s)
    --- FAIL: TestAdmissionControl/medium_error_rate_and_bigger_than_threshold_will_block_some_traffic (1.00s)
        admission_test.go:225: ok: 424.00, fail: 465.00, failBackend: 355.00
        admission_test.go:231: Failed to get expected fails should be in: 484.54 < 465.00 < 608.94
```

szuecs:
again https://github.com/zalando/skipper/actions/runs/10349009756/job/28642495361

szuecs:
```
time="2025-09-30T09:20:25Z" level=info msg="admissionControl: p: 0.74, r: 0.09"
time="2025-09-30T09:20:25Z" level=info msg="admissionControl: total < s = false, rejectP = (24.00 - 5.56) / (24.00 + 1)  --- success: 5.00 and threshold: 0.90"
time="2025-09-30T09:20:25Z" level=info msg="admissionControl: p: 0.74, r: 0.09"
time="2025-09-30T09:20:25Z" level=info msg="admissionControl: total < s = false, rejectP = (24.00 - 5.56) / (24.00 + 1)  --- success: 5.00 and threshold: 0.90"
time="2025-09-30T09:20:25Z" level=info msg="admissionControl: p: 0.74, r: 0.75"
time="2025-09-30T09:20:25Z" level=info msg="admissionControl: total < s = false, rejectP = (27.00 - 6.67) / (27.00 + 1)  --- success: 6.00 and threshold: 0.90"
time="2025-09-30T09:20:25Z" level=info msg="admissionControl: p: 0.73, r: 0.20"
[...]

PR #2506, opened 2023-08-07
queuelistener: fix testListener goroutine leak

AlexanderYastrebov:
Queuelistener continously calls Accept on the testListener which keeps a buffer of last accepted testConnections in the conns channel.

The test code races to receive from the same channel so Accept goroutine may occasionally block receiving from the empty channel which would be detected by goroutine leak detector introduced by the #2499.

The problem could be reproduced e.g. by running:
```
GODEBUG=tracebackancestors=10 go test ./queuelistener/ -run=TestTeardown/connections_accepted_from_the_wrapped_listener_closed_after_tear_down -count=1000

noleak: 1 active

goroutine 2655 [chan receive]:
github.com/zalando/skipper/queuelistener.(*testListener).Accept(0xc0000e7830)
...
```

This change avoids blocking Accept on potentially empty channel.

### DIFF
--- a/queuelistener/listener_test.go
+++ b/queuelistener/listener_test.go
@@ -80,8 +80,13 @@ func (l *testListener) Accept() (net.Conn, error) {
 		select {
 		case l.conns <- c:
 		default:
-			// drop one if cannot store the latest
-			<-l.conns
+			// Drop one if cannot store the latest.
+			// The test might have received a connection in the meantime so do not block.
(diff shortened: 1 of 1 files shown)


AlexanderYastrebov:
check-race failed due to https://github.com/zalando/skipper/issues/2386

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2510, opened 2023-08-11
🚑️ OIDC: use Azure-AD unique identifier

universam1:
We are facing a security vulnerability by the use of `DisplayName` as the unique identifier in Azure-AD.

Since this field can be duplicated it can be exploited to gain privileges.

https://morgansimonsen.com/2016/06/28/azure-ad-allows-duplicate-group-names/

Switching to the guaranteed unique identifier `onPremisesSamAccountName` available in the Microsoft Graph API

### DIFF
--- a/filters/auth/oidc.go
+++ b/filters/auth/oidc.go
@@ -65,8 +65,8 @@ type claimSource struct {
 type azureGraphGroups struct {
 	OdataNextLink string `json:"@odata.nextLink,omitempty"`
 	Value         []struct {
-		DisplayName string `json:"displayName"`
-		ID          string `json:"id"`
+		OnPremisesSamAccountName string `json:"onPremisesSamAccountName"`
+		ID                       string `json:"id"`
 	} `json:"value"`
 }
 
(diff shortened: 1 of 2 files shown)


universam1:
@szuecs Do you see a chance this to be merged soon?

szuecs:
:+1:

MustafaSaber:
👍

filters/auth/oidc.go
```diff
@@ -1075,7 +1075,9 @@ func (f *tokenOidcFilter) resolveDistributedClaimAzure(url *url.URL, oauth2Token
 		return nil, fmt.Errorf("unabled to decode response: %w", err)
 	}
 	for _, v := range target.Value {
-		values = append(values, v.DisplayName)
+		if v.OnPremisesSamAccountName != "" {
```
AlexanderYastrebov, line 1078: > Switching to the guaranteed unique identifier onPremisesSamAccountName available in the Microsoft Graph API

Could it be empty?

AlexanderYastrebov: And should we have a test to ensure behaviour when it is empty?

universam1: Yes it can be empty and extended a test to reflect it.

AlexanderYastrebov: `onPremisesSamAccountName` does not look like a standard attribute so we have to rely on you testing that this works correctly.

I guess `id` is also not stable/reliable to use for the purpose, is it?

universam1: The onPremisesSamAccountName is the equivalent of the samAccountName issued with the standard group claim. 

But it might be that someone rather would like the GUID / ID to be issued, it is configurable in Azure app registration. 

If you want @AlexanderYastrebov the field selector could be extended to be configurable. If so, which way, using a flag or another filter parameter?

AlexanderYastrebov: We do not use azure-based auth and have no experience with it. Lets have it like that and see if we get any feedback on the change from other users.

PR #2511, opened 2023-08-11
doc: release should contain the multiarch build and remove outdated arm64/armv7 builds

szuecs:
doc: release should contain the multiarch build and remove outdated arm64/armv7 builds

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -43,13 +43,9 @@ pipeline:
         export VERSION="${RELEASE_VERSION}"
 
         IMAGE="${IMAGE_REGISTRY}/teapot/skipper:${RELEASE_VERSION}"
-        ARM_IMAGE="${IMAGE_REGISTRY}/teapot/skipper-armv7:${RELEASE_VERSION}"
-        ARM64_IMAGE="${IMAGE_REGISTRY}/teapot/skipper-arm64:${RELEASE_VERSION}"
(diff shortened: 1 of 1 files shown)


delivery.yaml
```diff
@@ -74,12 +70,6 @@ pipeline:
         cdp-promote-image "${MULTIARCH_IMAGE}"
         echo "Created multi-arch docker image container-registry.zalando.net/teapot/skipper:${RELEASE_VERSION}"
 
-        echo "Creating docker image registry.opensource.zalan.do/teapot/skipper-arm64:${RELEASE_VERSION}"
-        make docker.build.arm64 && git status && git diff && make docker.push.arm64
-
-        echo "Creating docker image registry.opensource.zalan.do/teapot/skipper-armv7:${RELEASE_VERSION}"
-        make docker.build.armv7 && git status && git diff && make docker.push.armv7
-
```
AlexanderYastrebov, line 82: Should we clean up Makefiles and Dockerfiles?

szuecs: I think multiarch is right now only arm64 and am64, so cleaning up armv7 (RPIs <v4) would not make sense.
I think it's fine to have these targets, but maybe we can also re-think it

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2518, opened 2023-08-16
OPA: authorization based on request body

mjungsbluth:
This PR adds the ability to use the request's body in authorisation decisions with Open Policy Agent. 

It supports the OPA Envoy plugin behaviour and only parses the body up to a configurable maximum size and otherwise tees the request body. This is to avoid parsing large bodies and create memory and performance issues inside Skipper. 

Disabling body parsing can be configured via the Open Policy Agent configuration file. It supports the same content types that the upstream OPA plugin supports: `application/json` and `multipart/form-data`. 

To make negative impact on latency or memory consumption measurable, two new filter names are introduced that capture the intent to use the body. The implementation is also smart in the sense that it will check if the policy actually uses the body on from the input object before inspecting the body stream.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -501,7 +501,7 @@ func NewConfig() *Config {
 	flag.BoolVar(&cfg.EnableOpenPolicyAgent, "enable-open-policy-agent", false, "enables Open Policy Agent filters")
 	flag.StringVar(&cfg.OpenPolicyAgentConfigTemplate, "open-policy-agent-config-template", "", "file containing a template for an Open Policy Agent configuration file that is interpolated for each OPA filter instance")
 	flag.StringVar(&cfg.OpenPolicyAgentEnvoyMetadata, "open-policy-agent-envoy-metadata", "", "JSON file containing meta-data passed as input for compatibility with Envoy policies in the format")
-	flag.DurationVar(&cfg.OpenPolicyAgentCleanerInterval, "open-policy-agent-cleaner-interval", openpolicyagent.DefaultCleanerInterval, "Duration in seconds to wait before cleaning up unused opa instances")
+	flag.DurationVar(&cfg.OpenPolicyAgentCleanerInterval, "open-policy-agent-cleaner-interval", openpolicyagent.DefaultCleanIdlePeriod, "Duration in seconds to wait before cleaning up unused opa instances")
 	flag.DurationVar(&cfg.OpenPolicyAgentStartupTimeout, "open-policy-agent-startup-timeout", openpolicyagent.DefaultOpaStartupTimeout, "Maximum duration in seconds to wait for the open policy agent to start up")
 	flag.Int64Var(&cfg.OpenPolicyAgentMaxRequestBodySize, "open-policy-agent-max-request-body-size", openpolicyagent.DefaultMaxRequestBodySize, "Maximum number of bytes from a http request body that are passed as input to the policy")
 	flag.Int64Var(&cfg.OpenPolicyAgentMaxMemoryBodyParsing, "open-policy-agent-max-memory-body-parsing", openpolicyagent.DefaultMaxMemoryBodyParsing, "Total number of bytes used to parse http request bodies across all requests. Once the limit is met, requests will be rejected.")
--- a/config/config.go
+++ b/config/config.go
@@ -278,11 +278,13 @@ type Config struct {
 	LuaModules *listFlag `yaml:"lua-modules"`
 	LuaSources *listFlag `yaml:"lua-sources"`
 
-	EnableOpenPolicyAgent          bool          `yaml:"enable-open-policy-agent"`
-	OpenPolicyAgentConfigTemplate  string        `yaml:"open-policy-agent-config-template"`
-	OpenPolicyAgentEnvoyMetadata   string        `yaml:"open-policy-agent-envoy-metadata"`
-	OpenPolicyAgentCleanerInterval time.Duration `yaml:"open-policy-agent-cleaner-interval"`
-	OpenPolicyAgentStartupTimeout  time.Duration `yaml:"open-policy-agent-startup-timeout"`
+	EnableOpenPolicyAgent               bool          `yaml:"enable-open-policy-agent"`
+	OpenPolicyAgentConfigTemplate       string        `yaml:"open-policy-agent-config-template"`
+	OpenPolicyAgentEnvoyMetadata        string        `yaml:"open-policy-agent-envoy-metadata"`
+	OpenPolicyAgentCleanerInterval      time.Duration `yaml:"open-policy-agent-cleaner-interval"`
+	OpenPolicyAgentStartupTimeout       time.Duration `yaml:"open-policy-agent-startup-timeout"`
+	OpenPolicyAgentMaxRequestBodySize   int64         `yaml:"open-policy-agent-max-request-body-size"`
+	OpenPolicyAgentMaxMemoryBodyParsing int64         `yaml:"open-policy-agent-max-memory-body-parsing"`
 }
 
 const (
@@ -501,6 +503,8 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.OpenPolicyAgentEnvoyMetadata, "open-policy-agent-envoy-metadata", "", "JSON file containing meta-data passed as input for compatibility with Envoy policies in the format")
 	flag.DurationVar(&cfg.OpenPolicyAgentCleanerInterval, "open-policy-agent-cleaner-interval", openpolicyagent.DefaultCleanerInterval, "Duration in seconds to wait before cleaning up unused opa instances")
 	flag.DurationVar(&cfg.OpenPolicyAgentStartupTimeout, "open-policy-agent-startup-timeout", openpolicyagent.DefaultOpaStartupTimeout, "Maximum duration in seconds to wait for the open policy agent to start up")
+	flag.Int64Var(&cfg.OpenPolicyAgentMaxRequestBodySize, "open-policy-agent-max-request-body-size", openpolicyagent.DefaultMaxRequestBodySize, "Maximum number of bytes from a http request body that are passed as input to the policy")
+	flag.Int64Var(&cfg.OpenPolicyAgentMaxMemoryBodyParsing, "open-policy-agent-max-memory-body-parsing", openpolicyagent.DefaultMaxMemoryBodyParsing, "Total number of bytes used to parse http request bodies across all requests. Once the limit is met, requests will be rejected.")
 
 	// TLS client certs
 	flag.StringVar(&cfg.ClientKeyFile, "client-tls-key", "", "TLS Key file for backend connections, multiple keys may be given comma separated - the order must match the certs")
@@ -901,11 +905,13 @@ func (c *Config) ToOptions() skipper.Options {
 		LuaModules: c.LuaModules.values,
 		LuaSources: c.LuaSources.values,
 
-		EnableOpenPolicyAgent:          c.EnableOpenPolicyAgent,
-		OpenPolicyAgentConfigTemplate:  c.OpenPolicyAgentConfigTemplate,
-		OpenPolicyAgentEnvoyMetadata:   c.OpenPolicyAgentEnvoyMetadata,
-		OpenPolicyAgentCleanerInterval: c.OpenPolicyAgentCleanerInterval,
-		OpenPolicyAgentStartupTimeout:  c.OpenPolicyAgentStartupTimeout,
+		EnableOpenPolicyAgent:               c.EnableOpenPolicyAgent,
+		OpenPolicyAgentConfigTemplate:       c.OpenPolicyAgentConfigTemplate,
+		OpenPolicyAgentEnvoyMetadata:        c.OpenPolicyAgentEnvoyMetadata,
+		OpenPolicyAgentCleanerInterval:      c.OpenPolicyAgentCleanerInterval,
+		OpenPolicyAgentStartupTimeout:       c.OpenPolicyAgentStartupTimeout,
+		OpenPolicyAgentMaxRequestBodySize:   c.OpenPolicyAgentMaxRequestBodySize,
+		OpenPolicyAgentMaxMemoryBodyParsing: c.OpenPolicyAgentMaxMemoryBodyParsing,
 	}
 	for _, rcci := range c.CloneRoute {
 		eskipClone := eskip.NewClone(rcci.Reg, rcci.Repl)
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	log "github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/filters/openpolicyagent"
 	"github.com/zalando/skipper/net"
 	"github.com/zalando/skipper/proxy"
 	"gopkg.in/yaml.v2"
@@ -160,8 +161,10 @@ func defaultConfig() *Config {
 		ValidateQueryLog:                        true,
 		LuaModules:                              commaListFlag(),
 		LuaSources:                              commaListFlag(),
-		OpenPolicyAgentCleanerInterval:          10 * time.Second,
+		OpenPolicyAgentCleanerInterval:          openpolicyagent.DefaultCleanIdlePeriod,
 		OpenPolicyAgentStartupTimeout:           30 * time.Second,
+		OpenPolicyAgentMaxRequestBodySize:       openpolicyagent.DefaultMaxRequestBodySize,
+		OpenPolicyAgentMaxMemoryBodyParsing:     openpolicyagent.DefaultMaxMemoryBodyParsing,
 	}
 }
 
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -1879,6 +1879,20 @@ Headers both to the upstream and the downstream service can be manipulated the s
 
 This allows both to add and remove unwanted headers in allow/deny cases.
 
+#### opaAuthorizeRequestWithBody
+
+Requests can also be authorized based on the request body the same way that is supported with the [Open Policy Agent Envoy plugin](https://www.openpolicyagent.org/docs/latest/envoy-primer/#example-input), look for the input attribute `parsed_body` in the upstream documentation.
+
+This filter has the same parameters that the `opaAuthorizeRequest` filter has.
+
+A request's body is parsed up to a maximum size with a default of 1MB that can be configured via the `-open-policy-agent-max-request-body-size` command line argument. To avoid OOM errors due to too many concurrent authorized body requests, another flag `-open-policy-agent-max-memory-body-parsing` controls how much memory can be used across all requests with a default of 100MB. If in-flight requests that use body authorization exceed that limit, incoming requests that use the body will be rejected with an internal server error. The number of concurrent requests is 
+
+$$ n_{max-memory-body-parsing} \over min(avg(n_{request-content-length}), n_{max-request-body-size}) $$
+
+so if requests on average have 100KB and the maximum memory is set to 100MB, on average 1024 authorized requests can be processed concurrently. 
+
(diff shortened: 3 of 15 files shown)


filters/openpolicyagent/openpolicyagent.go
```diff
@@ -37,6 +40,7 @@ const (
 	defaultReuseDuration       = 30 * time.Second
 	defaultShutdownGracePeriod = 30 * time.Second
 	DefaultCleanIdlePeriod     = 10 * time.Second
+	DefaultMaxBodySize         = 128 * 1024 * 1024
```
AlexanderYastrebov, line 43: This feels to much. What is the rule of thumb to estimate this?

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -470,6 +485,42 @@ func (opa *OpenPolicyAgentInstance) MetricsKey(key string) string {
...
+	var rawBody []byte
+	if body != nil && !opa.EnvoyPluginConfig().SkipRequestBodyParse && opa.maxBodyBytes > 0 {
+		bufferedReader := bufio.NewReaderSize(body, opa.maxBodyBytes)
+		wrapper := &bufferedReadCloser{
+			closer: req.Body,
+			reader: bufferedReader,
+		}
+
+		var err error
+		rawBody, err = bufferedReader.Peek(opa.maxBodyBytes)
```
szuecs, line 512: That's likely blocking call to read(2) and can be used as DoS vector.
Can you please make sure it's not blocking, but streaming in case you need to work on the request body?

filters/openpolicyagent/openpolicyagent.go
AlexanderYastrebov: I suggest we have no default constant, have flag value zero and configure actual value in the deployment manifest.

szuecs: We could also make it lower case, private, and change it later. I agree that right now 128MB seems to be a bit big for a body.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -470,6 +485,42 @@ func (opa *OpenPolicyAgentInstance) MetricsKey(key string) string {
...
+
+func (r *bufferedReadCloser) Close() error {
+	return r.closer.Close()
+}
+
+func (opa *OpenPolicyAgentInstance) ExtractHttpBodyOptionally(req *http.Request) (io.ReadCloser, []byte, error) {
+	body := req.Body
+	var rawBody []byte
+	if body != nil && !opa.EnvoyPluginConfig().SkipRequestBodyParse && opa.maxBodyBytes > 0 {
+		bufferedReader := bufio.NewReaderSize(body, opa.maxBodyBytes)
```
AlexanderYastrebov, line 505: This will always allocate maxBodyBytes while actual body could be much smaller

filters/openpolicyagent/openpolicyagent.go
adetunjiakintundeakinde: Would 10 MB be fine?

mjungsbluth: I have to admit I got carried away and did not see it when reviewing my changes. 128KB is what I intended to actually use. This is based on internal use cases that we looked at where 60-100KB of JSON are valid cases so far. 
I think the max header size is currently 1MB which we could also default to and would already be generous. 10MB is already strange for an authz use case IMO.

filters/openpolicyagent/openpolicyagent.go
mjungsbluth: Correct, it saved re-implementing a smarter buffered reader which I can also do. My assumption was that the buffer is typically in the range of hundreds of KBs so should not hurt too much. Would you recommend re-doing this with a fixed size buffer (that is in low two digits KBs)?

filters/openpolicyagent/openpolicyagent.go
szuecs: I would suggest to create another filter for the body inspection.
Then for the implementation it would be a streaming filter that works on the body and as soon as it is clear that the body should be denied stop by returning an error, similar to blockContent filter.
For details how to do streaming filters: https://github.com/zalando/skipper/pull/2428 , I also had some more diret pointers in chat.

adetunjiakintundeakinde: @szuecs I have some questions
With streaming do you mean that
1. We read the body partially like the blockContent filter does?
2. With streaming is the expectation to still hit the backend before the request filter finishes the processing?

mjungsbluth: Thanks for the pointers @szuecs, seems doable after reading the linked PR, just a few clarifying questions (and slightly separate topics)

(a) Separate Filter
Is the motivation to not touch the body unless needed or is there some other thing that is missing? 
It should be possible to be a bit smarter in the filter and only parse the body only if it is really needed by the policy (needs some exploration with partial evaluation) and it fits from a size perspective. From a UX perspective I think ending up with different filters for each use case could be confusing but it is also a bit hard to predict which other cases we'll see.  This would also be a good motivator to include a benchmark...

(b) Streaming body
Given that the mentioned PR is not yet merged, this essentially would result in a special purpose wrapper, do you think this is ok and we try to unify later or should we wait for the PR to be merged?
Secondly, the policy can influence the returned body / headers. If the deny happens during the body processing, I am not yet sure where we would implement that. I assume that the error returned during `Read()` is somehow propagated to the `Response()` method of the filter or is there a better way for this?

szuecs: (a) if it's possible to know at parse time that you need to touch the body, then it would be great to only touch the body in case you need.
(b) yes it would already send the request headers to the backend and start streaming, but your filter will be able to stop the body streaming, because it gets the bytes before these will be written into the wire. You can pass errors to the Response of the filter via filtercontext's stateBag (map[string]interface{}) .

mjungsbluth: Thanks, then I'll proceed with making the changes:
(a) will try to get this info, if that is not feasible or slow, we can default to the separate filter
(b) will switch to a streaming processing then with the information you provided...

mjungsbluth: When starting to implement (b), I encountered an issue that I missed: The decision can influence the headers that are sent to the downstream service and we have an internal use case that will rely on that. 
I assume that the actual `Read` for passing the request is actually done by the `http.Client` that makes the outgoing request. This client actually copies the headers before starting to read the body which means that they can not longer be influenced. 
I am now a bit unsure how to proceed. (a) seems possible after initial testing and the read buffer can be optimised so that parsing the body does not always consume the max body size but this would still block in the `Request()` method of the filter. I will reach out via internal chat to discuss options...

mjungsbluth:
The PR is updated with the following changes:
- There are two separate filter aliases that enable body parsing but otherwise re-use the existing filter code
- body parsing is only done if not allowed by the control plane and the policy actually depends on the body for decision making
- The body is read in chunks (by default 8kb) and stops at either content length or the maximum configured buffer size (1MB = may header size)

Happy to change things esp. if there are concerns with the default sizes.

config/config.go
```diff
@@ -497,6 +498,7 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.OpenPolicyAgentConfigTemplate, "open-policy-agent-config-template", "", "file containing a template for an Open Policy Agent configuration file that is interpolated for each OPA filter instance")
 	flag.StringVar(&cfg.OpenPolicyAgentEnvoyMetadata, "open-policy-agent-envoy-metadata", "", "JSON file containing meta-data passed as input for compatibility with Envoy policies in the format")
 	flag.DurationVar(&cfg.OpenPolicyAgentCleanerInterval, "open-policy-agent-cleaner-interval", openpolicyagent.DefaultCleanIdlePeriod, "JSON file containing meta-data passed as input for compatibility with Envoy policies in the format")
+	flag.Uint64Var(&cfg.OpenPolicyAgentMaxBodySize, "open-policy-agent-max-body-size", http.DefaultMaxHeaderBytes, "Maximum number of bytes from the body that are passed as input to the policy")
```
szuecs, line 501: Is this reasonable to depend on stdlib max header bytes?
I guess we should have our own MaxBodySize in openpolicyagent package.

config/config_test.go
```diff
@@ -161,6 +161,7 @@ func defaultConfig() *Config {
 		LuaModules:                              commaListFlag(),
 		LuaSources:                              commaListFlag(),
 		OpenPolicyAgentCleanerInterval:          10 * time.Second,
+		OpenPolicyAgentMaxBodySize:              1048576,
```
szuecs, line 164: openpolicyagent.MaxBodySize ?

docs/reference/filters.md
```diff
@@ -1796,7 +1796,17 @@ The difference is that if the decision in (3) is equivalent to false, the respon
...
-This allows both to add and remove unwanted headers in allow/deny cases. 
+This allows both to add and remove unwanted headers in allow/deny cases.
+
+#### opaAuthorizeRequestWithBody
+
+Requests can also be authorized based on the request body the same way that is supported with the [Open Policy Agent Envoy plugin](https://www.openpolicyagent.org/docs/latest/envoy-primer/#example-input), look for the input attribute `parsed_body` in the upstream documentation.
+
+This filter has the same parameters that the `opaAuthorizeRequest` filter has.
+
+The body is parsed up to a maximum size that can be configured via the `-open-policy-agent-max-body-size` command line argument.
```
szuecs, line 1807: Tell about the default size here, too

docs/reference/filters.md
```diff
@@ -1841,6 +1851,16 @@ For this filter, the data flow looks like this independent of an allow/deny deci
 
 ```
 
+#### opaServeResponseWithReqBody
+
+If you want to serve requests directly from an Open Policy Agent policy that uses the request body, this can be done by using the `input.parsed_body` attribute the same way that is supported with the [Open Policy Agent Envoy plugin](https://www.openpolicyagent.org/docs/latest/envoy-primer/#example-input).
+
+This filter has the same parameters that the `opaServeResponse` filter has.
+
+The body is parsed up to a maximum size that can be configured via the `-open-policy-agent-max-body-size` command line argument.
```
szuecs, line 1860: tell about the default size, too

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -470,6 +502,100 @@ func (opa *OpenPolicyAgentInstance) MetricsKey(key string) string {
...
+		}
+	})
+	return err
+}
+
+func (opa *OpenPolicyAgentInstance) ExtractHttpBodyOptionally(req *http.Request) (io.ReadCloser, []byte, error) {
+	body := req.Body
+
+	if body != nil && !opa.EnvoyPluginConfig().SkipRequestBodyParse &&
+		opa.maxBodyBytes > 0 && req.ContentLength <= int64(opa.maxBodyBytes) {
```
szuecs, line 580: `opa.maxBodyBytes > 0` seems to be included in `req.ContentLength <= int64(opa.maxBodyBytes)` and 0 case should be handled by `body != nil`

szuecs (commented):
Can you please add a test case for chaining the request filter?
I expect that people will use chaining and 2 days ago we found https://github.com/zalando/skipper/issues/2605.

config/config.go
mjungsbluth: Will change

mjungsbluth:
> Can you please add a test case for chaining the request filter?
> I expect that people will use chaining and 2 days ago we found #2605.

Interesting 🤔. Yes will do that…

mjungsbluth:
The PR has been updated, the chaining luckily worked without any changes and the other comments have been addressed as well.

szuecs:
:+1:

szuecs:
@AlexanderYastrebov please review, thanks

mjungsbluth:
I will add one more memory setting to keep the total size of all concurrent bodies in check and avoid an OOM.

mjungsbluth:
Added a new memory setting that is global for all in-flight requests that depend on body authz which is enforced using a semaphore. Exceeding that limit will start failing requests with a 5xx status. Happy to chat on names and if we should block instead of failing ...

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -573,27 +590,45 @@ func (m *bufferedBodyReader) Close() error {
 	return err
 }
 
-func (opa *OpenPolicyAgentInstance) ExtractHttpBodyOptionally(req *http.Request) (io.ReadCloser, []byte, error) {
+func bodyUpperBound(contentLength, maxBodyBytes int64) int64 {
+	if contentLength <= 0 {
+		return maxBodyBytes
```
szuecs, line 595: Do you have a test that has no body in the request?

skipper.go
```diff
@@ -909,11 +909,12 @@ type Options struct {
...
-	EnableOpenPolicyAgent          bool
-	OpenPolicyAgentConfigTemplate  string
-	OpenPolicyAgentEnvoyMetadata   string
-	OpenPolicyAgentCleanerInterval time.Duration
-	OpenPolicyAgentMaxBodySize     uint64
+	EnableOpenPolicyAgent             bool
+	OpenPolicyAgentConfigTemplate     string
+	OpenPolicyAgentEnvoyMetadata      string
+	OpenPolicyAgentCleanerInterval    time.Duration
+	OpenPolicyAgentMaxRequestBodySize int64
```
szuecs, line 916: Maybe:
OpenPolicyAgentMaxRequestBodySize -> OpenPolicyAgentMaxConcurrentRequestsForBodyFilter 
I wonder a bit about what this semaphore has to do with the body size other than the filter does body inspection.

docs/reference/filters.md
```diff
@@ -1804,7 +1804,7 @@ Requests can also be authorized based on the request body the same way that is s
 
 This filter has the same parameters that the `opaAuthorizeRequest` filter has.
 
-The body is parsed up to a maximum size with a default of 1MB that can be configured via the `-open-policy-agent-max-body-size` command line argument.
+The body is parsed up to a maximum size with a default of 1MB that can be configured via the `-open-policy-agent-max-request-body-size` command line argument. To avoid OOM errors due to too many authorized body requests, another flag `-open-policy-agent-max-total-body-size` controls how much memory can be used across all requests with a default of 100MB. If  in-flight requests that use body authorization exceed that limit, incoming requests that use the body will be rejected with an internal server error.
```
szuecs, line 1807: Maybe also explain the equation that is used to limit max concurrent users.
Is it `100MB/1MB = 100`  or something else?

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -100,19 +110,23 @@ func WithCleanInterval(interval time.Duration) func(*OpenPolicyAgentRegistry) er
...
+		maxRequestBodyBytes: DefaultMaxTotalBodySize,
+		bodyReadBufferSize:  defaultBodyBufferSize,
 	}
 
 	for _, opt := range opts {
 		opt(registry)
 	}
 
+	if registry.totalBodySizeSem == nil {
+		registry.totalBodySizeSem = semaphore.NewWeighted(DefaultMaxTotalBodySize)
```
szuecs, line 127: this reads a little weird :)
Why the weighted semaphore is created with DefaultMaxTotalBodySize ?
DefaultMaxTotalBodySize sounds like 100MB or 1MB, but that's not the input of the semaphore...

filters/openpolicyagent/openpolicyagent.go
mjungsbluth: Actually no. Will add one...

skipper.go
mjungsbluth: Maybe: 
OpenPolicyAgentMaxPerRequestBodySize
and OpenPolicyAgentMaxConcurrentRequestBodyMemory

but also open to other suggestions. One effectively controls how much memory each request can consume and the second is a shared constrained resource (the total buffer that can be used to parse bodies) which is protected by the semaphore.

docs/reference/filters.md
mjungsbluth: Will add this. It is effectively depending on the actual content length. So if you have an average of 80kb bodies, you can have 100MB/80KB=1280 concurrent requests before the filters would start to consume a lot more memory than granted.

filters/openpolicyagent/openpolicyagent.go
mjungsbluth: The semaphore guards the bytes of the memory allocated for body parsing. This means that Aquire and Release can just use the number of bytes that the filter will consume (content length or max) to make sure the overall.

I am aware that semaphores are typically used for few digit countable resources but I have seen this pattern work well in C++ applications in the past and in this case no assumption on the actual number of concurrent requests need to be made, the max allocated memory is the only fixed value...

szuecs: Yeah makes sense

skipper.go
szuecs: then let's not change it, I think I just overlooked the semaphore use

szuecs:
> Added a new memory setting that is global for all in-flight requests that depend on body authz which is enforced using a semaphore. Exceeding that limit will start failing requests with a 5xx status. Happy to chat on names and if we should block instead of failing ...

Fail fast is preferable.

docs/reference/filters.md
mjungsbluth: Amended the docs.

filters/openpolicyagent/openpolicyagent.go
mjungsbluth: Added a few tests against unknown and nil bodies.

docs/reference/filters.md
```diff
@@ -1857,7 +1857,7 @@ If you want to serve requests directly from an Open Policy Agent policy that use
...
+A request's body is parsed up to a maximum size with a default of 1MB that can be configured via the `-open-policy-agent-max-request-body-size` command line argument. To avoid OOM errors due to too many concurrent authorized body requests, another flag `-open-policy-agent-max-memory-body-parsing` controls how much memory can be used across all requests with a default of 100MB. If  in-flight requests that use body authorization exceed that limit, incoming requests that use the body will be rejected with an internal server error. The number of concurrent requests is <max-memory-body-parsing> / min(avg(<request content-length>), <max-request-body-size>), so if requests on average have 100KB and the maximum memory is set to 100MB, on average 1024 authorized requests can be processed concurrently.
```
szuecs, line 1860: You can also use Math syntax like in https://github.com/zalando/skipper/blob/master/docs/reference/filters.md?plain=1#L2289 
I think this would be better readable.

szuecs:
lgtm, only the doc could be nicer :)

docs/reference/filters.md
mjungsbluth: Hah, did not know that. Looks indeed nicer, will change

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
```diff
@@ -137,23 +160,106 @@ func TestAuthorizeRequestFilter(t *testing.T) {
...
+			requestMethod:     "GET",
 			contextExtensions: "",
 			expectedStatus:    http.StatusInternalServerError,
 			expectedBody:      "",
 			expectedHeaders:   make(http.Header),
 			backendHeaders:    make(http.Header),
 			removeHeaders:     make(http.Header),
 		},
+		{
+			msg:             "Allow With Body",
```
szuecs, line 176: what about tests that exceeds the max body size?

mjungsbluth: @szuecs Wdyt, is that test enough or should we replicate it here as well?

szuecs: ok thanks!

szuecs:
:+1:

AlexanderYastrebov:
@mjungsbluth This needs a rebase

mjungsbluth:
@AlexanderYastrebov rebased onto current tip...

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2522, opened 2023-08-18
gosec: fix G601 rule

AlexanderYastrebov:
gosec started to produce false positives
"G601 (CWE-118): Implicit memory aliasing in for loop." for slices of struct pointers, see https://github.com/securego/gosec/issues/966

This changes fixes G601 rule wwarnings instead of muting it.

### DIFF
--- a/cmd/eskip/load.go
+++ b/cmd/eskip/load.go
@@ -34,7 +34,8 @@ var invalidRouteExpression = errors.New("one or more invalid route expressions")
 // parse errors if any.
 func mapRouteInfo(allInfo []*eskip.RouteInfo) loadResult {
 	lr := loadResult{make([]*eskip.Route, len(allInfo)), make(map[string]error)}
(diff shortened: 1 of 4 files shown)


AlexanderYastrebov:
See also
* https://husni.dev/beware-of-implicit-memory-aliasing-in-go-foor-loop/
* https://github.com/golang/go/issues/60078
* https://go.googlesource.com/proposal/+/master/design/60078-loopvar.md

RomanZavodskikh:
Can we somehow make gosec (and CI tests) fail in case somebody will violate this recommendation again? At the moment it seems that G601 is still a warning that could be easily ignored.

AlexanderYastrebov:
@RomanZavodskikh `make gosec` fails properly so its ok.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2524, opened 2023-08-18
Eskip check fails on empty URL on network backend.

### LINKED ISSUE #2523: eskip check don't fail when backend is empty string
**Describe the bug**
`eskip check routes.eskip` don't fail if there's a route with backend is an empty string but `skipper -routes-file routes.eskip ` fails with the message `failed to process route foo: parse "": empty url `

**To Reproduce**
Using image base `registry.opensource.zalan.do/teapot/skipper:v0.16.67`  executing `eskip check routes.eskip` there's no failure.

Executing `skipper -routes-file test.eskip`
> [APP]INFO[0000] Expose metrics in codahale format            
> [APP]INFO[0000] enable swarm: false                          
> [APP]INFO[0000] Replacing tee filter specification           
> [APP]INFO[0000] Replacing teenf filter specification         
> [APP]INFO[0000] Replacing lua filter specification           
> [APP]INFO[0000] support listener on :9911                    
> [APP]INFO[0000] Dataclients are updated once, first load complete 
> [APP]INFO[0000] Listen on :9090                              
> [APP]INFO[0000] route settings, reset, route: foo: Path("/foo") -> setPath("/") -> "" 
> [APP]INFO[0000] route settings, reset, route: bar: Path("/bar") -> setPath("/") -> "https://bar.example.org/" 
> [APP]INFO[0000] route settings received                      
> **[APP]ERRO[0000] failed to process route foo: parse "": empty url** 
> [APP]INFO[0000] route settings applied                       
> [APP]INFO[0000] TLS settings not found, defaulting to HTTP 

`routes.eskip` file content 

> foo: Path("/foo") -> setPath("/") -> ""
> bar: Path("/bar") -> setPath("/") -> "https://bar.example.org/" 

**Expected behavior**
Consistent behaviour between `eskip` and `skipper`, if `skipper` fails loading routes with empty backend then `eskip` should also fail

**Observed behavior**

xavivars:
This change to `eskip check` validates no routes have empty backend. 

"empty url" error falls between the "_pure syntactical_" checks that `eskip check` runs at validation time and the "_route-specific setup-dependant_" checks that skipper does at runtime (making sure, for example, that all predicates / filters exist, etc).

While this is not a fatal error, and loading the skip files in skipper works (only the specific route gets discarded), we think having this validation in `eskip check` is a better approach, because this is an error that is known in advance (no URL can be empty) and we could not find a reason we'd ever want to, on purpose, push a non-working route to production.

 Fixes #2523

### DIFF
--- a/cmd/eskip/load.go
+++ b/cmd/eskip/load.go
@@ -103,7 +103,24 @@ func checkCmd(a cmdArgs) error {
 		return err
 	}
 
+	err = checkEmptyBackends(routes)
+
+	if err != nil {
+		return err
+	}
+
 	return checkRepeatedRouteIds(routes)
+}
+
+func checkEmptyBackends(routes []*eskip.Route) error {
+
+	for _, route := range routes {
+		if route.BackendType == eskip.NetworkBackend && route.Backend == "" {
+			return errors.New("Route has empty backend: " + route.Id)
+		}
+	}
+
+	return nil
 }
 
 // command executed for print.
--- a/cmd/eskip/load.go
+++ b/cmd/eskip/load.go
@@ -93,7 +93,7 @@ func checkRepeatedRouteIds(routes []*eskip.Route) error {
 func checkEmptyBackends(routes []*eskip.Route) error {
 	for _, route := range routes {
 		if route.BackendType == eskip.NetworkBackend && route.Backend == "" {
-			return fmt.Errorf("route %s has empty backend", route.Id)
+			return fmt.Errorf("route has empty backend: %s", route.Id)
 		}
 	}
 
--- a/cmd/eskip/load.go
+++ b/cmd/eskip/load.go
@@ -83,7 +83,7 @@ func checkRepeatedRouteIds(routes []*eskip.Route) error {
 	ids := map[string]bool{}
 	for _, route := range routes {
 		if ids[route.Id] {
-			return errors.New("Repeating route with id " + route.Id)
+			return fmt.Errorf("repeating route with id %s", route.Id)
 		}
 		ids[route.Id] = true
 	}
--- a/cmd/eskip/load.go
+++ b/cmd/eskip/load.go
@@ -90,6 +90,16 @@ func checkRepeatedRouteIds(routes []*eskip.Route) error {
 	return nil
 }
 
+func checkEmptyBackends(routes []*eskip.Route) error {
+	for _, route := range routes {
+		if route.BackendType == eskip.NetworkBackend && route.Backend == "" {
+			return fmt.Errorf("route %s has empty backend", route.Id)
+		}
+	}
+
+	return nil
+}
+
 // load and parse routes, ignore parse errors.
 func loadRoutesUnchecked(in *medium) []*eskip.Route {
 	lr, _ := loadRoutes(in)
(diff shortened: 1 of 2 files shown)


cmd/eskip/load.go
```diff
@@ -103,9 +103,26 @@ func checkCmd(a cmdArgs) error {
 		return err
 	}
 
+	err = checkEmptyBackends(routes)
+
```
AlexanderYastrebov, line 107: No need to have a blank line inbetween

cmd/eskip/load.go
```diff
@@ -103,9 +103,26 @@ func checkCmd(a cmdArgs) error {
...
+	}
+
 	return checkRepeatedRouteIds(routes)
 }
 
+func checkEmptyBackends(routes []*eskip.Route) error {
+
+	for _, route := range routes {
+		if route.BackendType == eskip.NetworkBackend && route.Backend == "" {
+			return errors.New("Route has empty backend: " + route.Id)
```
AlexanderYastrebov, line 119: fmt.Errorf would be cleaner

cmd/eskip/load.go
```diff
@@ -103,9 +103,26 @@ func checkCmd(a cmdArgs) error {
...
+
+	if err != nil {
+		return err
+	}
+
 	return checkRepeatedRouteIds(routes)
 }
 
+func checkEmptyBackends(routes []*eskip.Route) error {
+
```
AlexanderYastrebov, line 116: Please drop blank lines

cmd/eskip/load.go
AlexanderYastrebov: Error message should start from lower case (I know you've copied this from checkRepeatedRouteIds)

cmd/eskip/load_test.go
```diff
@@ -142,6 +142,20 @@ func TestCheckRepeatingRouteIds(t *testing.T) {
 	}
 }
 
+func TestCheckEmptyBackend(t *testing.T) {
+	const name = "testFile"
+	err := withFile(name, `foo: Method("POST") -> "https://www.example1.org";empty: Method("POST") -> "";`, func(_ *os.File) {
+		err := checkCmd(cmdArgs{in: &medium{typ: file, path: name}})
+		if err == nil {
+			t.Error("Expected an error for empty backend")
+		}
```
AlexanderYastrebov, line 151: Lets check specific error value here.

AlexanderYastrebov:
Thank you. 
I think we can have it like you propose (after comments are addressed).

In general I think we should make backed validation more strict during eskip parsing - then k8s dataclient and webhook would benefit from it as well. @MustafaSaber is working on validation topic, see related https://github.com/zalando/skipper/issues/1618

cmd/eskip/load.go
xavivars: Do you want me to also fix the ones in `checkRepeatedRouteIds`? To me, it makes little sense to have different ways of formatting errors few lines apart...

cmd/eskip/load_test.go
xavivars: Do you mean something like `if err == nil || !errors.Is(err, expectedError) ` ?

AlexanderYastrebov: No, just something like `if err == nil || err.Error() != "route has empty backend: empty"`

cmd/eskip/load.go
AlexanderYastrebov: Yes, please fix another case

xavivars:
Thanks for the quick review!

I think I've addressed all your comments now.

And I agree: stricter backend validation would be great.

AlexanderYastrebov:
Thank you.

szuecs:
:+1:

PR #2525, opened 2023-08-18
add Tracer to skipper options (overrides OpenTracing)

gpradeepkrishna:
This PR enables applications that extends skipper to pre-initialize `Tracer` instance and feed it directly to Skipper overriding `OpenTracing`. The need for this arises due to the inherent behaviour of skipper making the `Tracer` available only via request context. Thus preventing the extending application to reuse the tracer instance.

### DIFF
--- a/skipper.go
+++ b/skipper.go
@@ -667,9 +667,9 @@ type Options struct {
 	// for a route when it's available (e.g. for RouteGroups)
 	OpenTracingBackendNameTag bool
 
-	// Tracer allows pre-created tracer to be passed on to skipper. Providing the tracer
-	// overrides options provided under OpenTracing property.
-	Tracer ot.Tracer
+	// OpenTracingTracer allows pre-created tracer to be passed on to skipper. Providing the
+	// tracer instance overrides options provided under OpenTracing property.
+	OpenTracingTracer ot.Tracer
 
 	// PluginDir defines the directory to load plugins from, DEPRECATED, use PluginDirs
 	PluginDir string
@@ -1149,9 +1149,9 @@ func (o *Options) tlsConfig(cr *certregistry.CertRegistry) (*tls.Config, error) 
 	return config, nil
 }
 
-func (o *Options) tracerInstance() (ot.Tracer, error) {
-	if o.Tracer != nil {
-		return o.Tracer, nil
+func (o *Options) openTracingTracerInstance() (ot.Tracer, error) {
+	if o.OpenTracingTracer != nil {
+		return o.OpenTracingTracer, nil
 	}
 
 	if len(o.OpenTracing) > 0 {
@@ -1493,7 +1493,7 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 
 	o.PluginDirs = append(o.PluginDirs, o.PluginDir)
 
-	tracer, err := o.tracerInstance()
+	tracer, err := o.openTracingTracerInstance()
 	if err != nil {
 		return err
 	}
--- a/skipper.go
+++ b/skipper.go
@@ -667,6 +667,10 @@ type Options struct {
 	// for a route when it's available (e.g. for RouteGroups)
 	OpenTracingBackendNameTag bool
 
+	// Tracer allows pre-created tracer to be passed on to skipper. Providing the tracer
+	// overrides options provided under OpenTracing property.
+	Tracer ot.Tracer
+
 	// PluginDir defines the directory to load plugins from, DEPRECATED, use PluginDirs
 	PluginDir string
 	// PluginDirs defines the directories to load plugins from
@@ -1468,20 +1472,13 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 
 	o.PluginDirs = append(o.PluginDirs, o.PluginDir)
 
-	var tracer ot.Tracer
-	if len(o.OpenTracing) > 0 {
-		tracer, err = tracing.InitTracer(o.OpenTracing)
-		if err != nil {
-			return err
-		}
-	} else {
-		// always have a tracer available, so filter authors can rely on the
-		// existence of a tracer
-		tracer, _ = tracing.LoadTracingPlugin(o.PluginDirs, []string{"noop"})
+	tracer, err := initializeTracer(o)
+	if err != nil {
+		return err
 	}
 
 	// tee filters override if we have a tracer
-	if len(o.OpenTracing) > 0 {
+	if tracer != nil {
 		o.CustomFilters = append(o.CustomFilters,
 			// tee()
 			teefilters.WithOptions(teefilters.Options{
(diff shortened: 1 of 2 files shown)


skipper.go
```diff
@@ -1468,20 +1472,13 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
-		// existence of a tracer
-		tracer, _ = tracing.LoadTracingPlugin(o.PluginDirs, []string{"noop"})
+	tracer, err := initializeTracer(o)
+	if err != nil {
+		return err
 	}
 
 	// tee filters override if we have a tracer
-	if len(o.OpenTracing) > 0 {
+	if tracer != nil {
```
AlexanderYastrebov, line 1481: I think we can drop this condition since the above code ensures there is always a tracer available.

skipper.go
```diff
@@ -2028,6 +2025,20 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	return listenAndServeQuit(o.CustomHttpHandlerWrap(proxy), &o, sig, idleConnsCH, mtr, cr)
 }
 
+func initializeTracer(o Options) (ot.Tracer, error) {
```
AlexanderYastrebov, line 2028: Lets make it a method similar to `filterRegistry()`

skipper_test.go
```diff
@@ -544,3 +544,32 @@ func TestDataClients(t *testing.T) {
 
 	sigs <- syscall.SIGTERM
 }
+
+func TestTracerInstanceTakesPrecedenceOverTracingOptions(t *testing.T) {
```
AlexanderYastrebov, line 548: Not sure if we need full integration test here, maybe lets make it similar to TestOptionsFilterRegistry

skipper.go
gpradeepkrishna: Agreed!

skipper.go
gpradeepkrishna: Makes sense, thanks

skipper_test.go
gpradeepkrishna: Will amend this as well.

skipper_test.go
```diff
@@ -109,6 +109,30 @@ func TestOptionsFilterRegistry(t *testing.T) {
...
+	tracer := &tracingtest.Tracer{}
+	// run skipper proxy that we want to test
+	o := Options{
+		Tracer:      tracer,
+		OpenTracing: []string{"noop"},
+	}
+
+	tr, err := o.tracerInstance()
+	assert.Nil(t, err)
+	assert.NotNil(t, tr)
```
AlexanderYastrebov, line 122:
```go
assert.NoError(t, err)
assert.Same(t, tracer, tr)
```
maybe also rename `tr` to `result` or `instance`

skipper_test.go
```diff
@@ -109,6 +109,30 @@ func TestOptionsFilterRegistry(t *testing.T) {
...
+}
+
+func TestOptionsTracerInstanceFallbacksToOpenTracingWhenTracerIsNil(t *testing.T) {
+	// run skipper proxy that we want to test
+	o := Options{
+		OpenTracing: []string{"noop"},
+	}
+
+	tr, err := o.tracerInstance()
+	assert.Nil(t, err)
```
AlexanderYastrebov, line 132: assert.NoError(t, err)

skipper_test.go
```diff
@@ -109,6 +109,28 @@ func TestOptionsFilterRegistry(t *testing.T) {
...
+		Tracer:      tracer,
+		OpenTracing: []string{"noop"},
+	}
+
+	tr, err := o.tracerInstance()
+	assert.Nil(t, err)
+	assert.Same(t, tr, tracer)
+}
+
+func TestOptionsTracerInstanceFallbacksToOpenTracingWhenTracerIsNil(t *testing.T) {
```
AlexanderYastrebov, line 124: Maybe just name it `TestOptionsOpenTracing`.

Lets also add a test (`TestOptionsDefaultTracer`) for the fallback tracer from empty `Options{}`

skipper_test.go
```diff
@@ -109,6 +109,28 @@ func TestOptionsFilterRegistry(t *testing.T) {
...
+func TestOptionsTracerInstanceOverridesOpenTracing(t *testing.T) {
+	tracer := &tracingtest.Tracer{}
+	o := Options{
+		Tracer:      tracer,
+		OpenTracing: []string{"noop"},
+	}
+
+	tr, err := o.tracerInstance()
+	assert.NoError(t, err)
+	assert.Same(t, tr, tracer)
```
AlexanderYastrebov, line 121: expected shoudl be the second argument, i.e. `assert.Same(t, tracer, tr)` - that's why I suggested to rename one ore the other

skipper_test.go
gpradeepkrishna: Looks like we overlooked a case where tracer can still be nil. This is when `o.Tracer = nil`, `o.OpenTracing = nil` & `pluginDirs = nil`.

AlexanderYastrebov: Hm, ok, then I guess we can return (nil, nil), have a test for it and restore nil check for tee filter override

gpradeepkrishna: alternatively we can also send back a noop tracer. WDYT?

skipper.go
```diff
@@ -1145,6 +1149,24 @@ func (o *Options) tlsConfig(cr *certregistry.CertRegistry) (*tls.Config, error)
...
+	if o.Tracer != nil {
+		return o.Tracer, nil
+	}
+
+	if len(o.OpenTracing) > 0 {
+		return tracing.InitTracer(o.OpenTracing)
+	} else {
+		// always have a tracer available, so filter authors can rely on the
+		// existence of a tracer
+		tracer, _ := tracing.LoadTracingPlugin(o.PluginDirs, []string{"noop"})
```
AlexanderYastrebov, line 1162: The default option is to use `noop` so we do not need to create ot.NoopTracer{} so lets not swallow the error but simply `return tracing.LoadTracingPlugin(o.PluginDirs, []string{"noop"})`

skipper_test.go
AlexanderYastrebov: I think we should change `LoadTracingPlugin` to return error if no plugin was loaded.

skipper.go
AlexanderYastrebov: Lets do this:
```go
if len(o.OpenTracing) > 0 {
  return tracing.InitTracer(o.OpenTracing)
} else {
  // always have a tracer available, so filter authors can rely on the
  // existence of a tracer
  tracer, err := tracing.LoadTracingPlugin(o.PluginDirs, []string{"noop"})
  if err != nil {
    return err
  } else if tracer == nil {
    // LoadTracingPlugin unfortunately may return nil tracer
    return fmt.Errorf("failed to load tracing plugin")
  }
  return tracer, nil
```

skipper_test.go
AlexanderYastrebov: I propose https://github.com/zalando/skipper/pull/2525#discussion_r1299939623

AlexanderYastrebov:
:+1:

skipper.go
```diff
@@ -667,6 +667,10 @@ type Options struct {
 	// for a route when it's available (e.g. for RouteGroups)
 	OpenTracingBackendNameTag bool
 
+	// Tracer allows pre-created tracer to be passed on to skipper. Providing the tracer
+	// overrides options provided under OpenTracing property.
+	Tracer ot.Tracer
```
szuecs, line 672: @lucastt fyi, let's talk tomorrow about it

szuecs (commented):
Let's talk tomorrow.
One question would be for a OT to OTel migration, how we would allow to set this. Should we have both OT and OTel tracer allowed to pass or should we only allow OTel and use the compatibility layer.

skipper.go
lucastt: Same as the other comment, try to be specific with the naming `openTracingTracer` or something like that. Something that specifies that this is OpenTracing, not OpenTelemetry, so when we migrate to OpenTelemetry we support you getting one instance or the other.

skipper.go
```diff
@@ -1145,6 +1149,27 @@ func (o *Options) tlsConfig(cr *certregistry.CertRegistry) (*tls.Config, error)
 	return config, nil
 }
 
+func (o *Options) tracerInstance() (ot.Tracer, error) {
```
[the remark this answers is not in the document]
gpradeepkrishna: Agreed and changed

AlexanderYastrebov:
:+1:

lucastt:
👍

PR #2526, opened 2023-08-18
routesrv: log route updates only when changed

AlexanderYastrebov:
* log and change route update metric only when routes changed
* speedup tests from 20s to 2.5s by reducing poll interval

### DIFF
--- a/routesrv/polling.go
+++ b/routesrv/polling.go
@@ -63,12 +63,6 @@ type poller struct {
 func (p *poller) poll(wg *sync.WaitGroup) {
 	defer wg.Done()
 
-	var (
-		routesCount, routesBytes int
-		initialized              bool
-		msg                      string
-	)
-
 	log.WithField("timeout", p.timeout).Info(LogPollingStarted)
 	ticker := time.NewTicker(p.timeout)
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

AlexanderYastrebov:
:+1:

routesrv/polling.go
```diff
@@ -98,19 +92,20 @@ func (p *poller) poll(wg *sync.WaitGroup) {
...
-			routesBytes, initialized = p.b.formatAndSet(routes)
+			routesBytes, initialized, updated := p.b.formatAndSet(routes)
 			logger := log.WithFields(log.Fields{"count": routesCount, "bytes": routesBytes})
 			if initialized {
 				logger.Info(LogRoutesInitialized)
 				span.SetTag("routes.initialized", true)
 				routesInitialized.SetToCurrentTime()
-			} else {
+			}
+			if updated {
```
MustafaSaber, line 105: should we set SpanTag for `routes.updated`?

AlexanderYastrebov: Makes sense, thank you.
Added in the #2531

PR #2531, opened 2023-08-21
routesrv: fix flaky TestESkipBytesHandlerWithOldLastModified

AlexanderYastrebov:
* fix flaky test by mocking clock instead of using sleep.
* add tracing tag on route update

Follows up on #2526


RomanZavodskikh:
:+1:

AlexanderYastrebov:
:+1:

PR #2532, opened 2023-08-21
Allow setting additional headers for response when do redirects

trkohler:
During one of the incidents at Zalando we encountered a need to [set cache header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) when we do [permanent redirects.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308)

Would it be possible to propagate custom headers into redirect function?

### DIFF
--- a/filters/builtin/redirect.go
+++ b/filters/builtin/redirect.go
@@ -30,6 +30,12 @@ const (
 	redToLower
 )
 
+var sensitiveHeaders = map[string]bool{
+	"Authrorization": true,
+	"Cookie":         true,
+	"Set-Cookie":     true,
+}
+
 // Filter to return
 type redirect struct {
 	typ      redirectType
@@ -160,7 +166,9 @@ func redirectWithType(ctx filters.FilterContext, code int, location *url.URL, ty
 	if headers != nil {
 		headersResponse = make(http.Header)
 		for k, v := range *headers {
-			headersResponse.Set(k, v)
+			if headerIsAllowed(k) {
+				headersResponse.Set(k, v)
+			}
 		}
 	}
 	headersResponse.Set("Location", getLocation(ctx, location, typ))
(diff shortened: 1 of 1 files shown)


szuecs:
I guess we should be able to deny list certain headers. For example `Authorization` header we do not want to leak in a chain of calls, for example.

AlexanderYastrebov:
Why not simply add `setResponseHeader("foo", "bar")` before redirect filter (you can also use https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#template-placeholders)?

Could you please outline the configuration where you'd like to have response header?

trkohler:
> Could you please outline the configuration where you'd like to have response headers with redirect?

```
func (s *filter) Request(ctx filters.FilterContext) {
	req := ctx.Request()
	if strings.HasSuffix(req.URL.Path, "/") || req.URL.Path == "" {
		return
	}

	uc := *req.URL
	uc.Path += "/"

	status := http.StatusMovedPermanently
	if s.flags.is(temporaryRedirect) {
		status = http.StatusFound
	}

	builtin.Redirect(ctx, status, &uc)
}
```

we use `builtin.Redirect` right now. We can overwrite it to our own or ask you about a possibility to extend it to cover additional header setting (that's what I am trying to do).

We need to set `Cache-Control` header on permanent redirect because in case of error in routing, browser would cache redirect forever if cache control is not set. During incident, we had thousands of product page requests affected and we want to adjust all permanent redirects to be covered with `Cache-Control`.

AlexanderYastrebov:
Why not 
```go
builtin.Redirect(ctx, status, &uc)
ctx.Response().Header.Add("Foo", "Bar")
```

trkohler:
thank you @AlexanderYastrebov it worked out. I'm closing PR

PR #2534, opened 2023-08-22
routesrv: fix flaky TestESkipBytesHandlerWithStaleEtag and others

AlexanderYastrebov:
Due to missing rs.StopUpdates() finished tests continue to write to the shared test log which could be observed as a lot of "failed to fetch routes" messages.
If k8s api server of the current test happens to start on the same port as the server from the previous test then current test would react on the log messages produced by the route server of the previous test and may fail sporadically.

The better fix would be to get rid of the shared test logger but this requires larger refactoring and could be done later.


AlexanderYastrebov:
:+1:

PR #2535, opened 2023-08-22
Introduce endpointregistry component

RomanZavodskikh:
This component is needed to create an easy way for loadbalancer, fadein, etc. to find out required information about endpoints to increase readability of those parts and enable us to provide new features there, for example, passive health checks for endpoints.

### DIFF
--- a/filters/fadein/fadein.go
+++ b/filters/fadein/fadein.go
@@ -201,7 +201,6 @@ func NewPostProcessor() routing.PostProcessor {
 }
 
 func (p *postProcessor) Do(r []*routing.Route) []*routing.Route {
-	const configErrFmt = "Error while processing endpoint fade-in settings: %s, %s, %v."
 	now := time.Now()
 
 	for _, ri := range r {
@@ -229,13 +228,7 @@ func (p *postProcessor) Do(r []*routing.Route) []*routing.Route {
 		for i := range ri.LBEndpoints {
 			ep := &ri.LBEndpoints[i]
 
-			s, h, err := normalizeSchemeHost(ep.Scheme, ep.Host)
-			if err != nil {
-				log.Errorf(configErrFmt, ep.Scheme, ep.Host, err)
-				continue
-			}
-
-			key := endpointKey(s, h)
+			key := endpointKey(ep.Scheme, ep.Host)
 			detected := p.detected[key].when
 			if detected.IsZero() || endpointsCreated[key].After(detected) {
 				detected = now
--- a/filters/fadein/fadein_test.go
+++ b/filters/fadein/fadein_test.go
@@ -265,8 +265,8 @@ func TestPostProcessor(t *testing.T) {
 
 		rt, _ := createRouting(t, routes)
 		r := route(rt, "/")
-		if r == nil || len(r.LBEndpoints) == 0 || !r.LBEndpoints[0].Detected.IsZero() {
-			t.Fatal("failed to ignore invalid LB endpoint")
+		if r != nil {
+			t.Fatal("created invalid LB endpoint")
 		}
 	})
 
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"math"
 	"math/rand"
+	"net"
 	"net/url"
 	"sort"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -14,7 +16,7 @@ import (
 	"github.com/cespare/xxhash/v2"
 	log "github.com/sirupsen/logrus"
 	"github.com/zalando/skipper/eskip"
-	"github.com/zalando/skipper/net"
+	snet "github.com/zalando/skipper/net"
 	"github.com/zalando/skipper/routing"
 )
 
@@ -308,7 +310,7 @@ func (ch *consistentHash) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 func (ch *consistentHash) chooseConsistentHashEndpoint(ctx *routing.LBContext, skipEndpoint func(int) bool) int {
 	key, ok := ctx.Params[ConsistentHashKey].(string)
 	if !ok {
-		key = net.RemoteHost(ctx.Request).String()
+		key = snet.RemoteHost(ctx.Request).String()
 	}
 	balanceFactor, ok := ctx.Params[ConsistentHashBalanceFactor].(float64)
 	var choice int
@@ -431,9 +433,14 @@ func parseEndpoints(r *routing.Route) error {
 			return err
 		}
 
+		scheme, host, err := normalizeSchemeHost(eu.Scheme, eu.Host)
+		if err != nil {
+			return err
+		}
+
 		r.LBEndpoints[i] = routing.LBEndpoint{
-			Scheme:  eu.Scheme,
-			Host:    eu.Host,
+			Scheme:  scheme,
+			Host:    host,
 			Metrics: &routing.LBMetrics{},
 		}
 	}
@@ -454,6 +461,33 @@ func setAlgorithm(r *routing.Route) error {
 
 	r.LBAlgorithm = initialize(r.Route.LBEndpoints)
 	return nil
+}
+
+func normalizeSchemeHost(s, h string) (string, string, error) {
+	// endpoint address cannot contain path, the rest is not case sensitive
+	s, h = strings.ToLower(s), strings.ToLower(h)
+
+	hh, p, err := net.SplitHostPort(h)
+	if err != nil {
+		// what is the actual right way of doing this, considering IPv6 addresses, too?
+		if !strings.Contains(err.Error(), "missing port") {
+			return "", "", err
+		}
+
+		p = ""
+	} else {
+		h = hh
+	}
+
+	switch {
+	case p == "" && s == "http":
+		p = "80"
+	case p == "" && s == "https":
+		p = "443"
+	}
+
+	h = net.JoinHostPort(h, p)
+	return s, h, nil
 }
 
 // Do implements routing.PostProcessor
--- a/loadbalancer/algorithm_test.go
+++ b/loadbalancer/algorithm_test.go
@@ -43,7 +43,7 @@ func TestSelectAlgorithm(t *testing.T) {
 
 		if len(rr[0].LBEndpoints) != 1 ||
 			rr[0].LBEndpoints[0].Scheme != "https" ||
-			rr[0].LBEndpoints[0].Host != "www.example.org" ||
+			rr[0].LBEndpoints[0].Host != "www.example.org:443" ||
 			rr[0].LBEndpoints[0].Metrics == nil {
 			t.Fatal("failed to set the endpoints")
 		}
@@ -70,7 +70,7 @@ func TestSelectAlgorithm(t *testing.T) {
 
 		if len(rr[0].LBEndpoints) != 1 ||
 			rr[0].LBEndpoints[0].Scheme != "https" ||
-			rr[0].LBEndpoints[0].Host != "www.example.org" ||
+			rr[0].LBEndpoints[0].Host != "www.example.org:443" ||
 			rr[0].LBEndpoints[0].Metrics == nil {
 			t.Fatal("failed to set the endpoints")
 		}
@@ -97,7 +97,7 @@ func TestSelectAlgorithm(t *testing.T) {
 
 		if len(rr[0].LBEndpoints) != 1 ||
 			rr[0].LBEndpoints[0].Scheme != "https" ||
-			rr[0].LBEndpoints[0].Host != "www.example.org" ||
+			rr[0].LBEndpoints[0].Host != "www.example.org:443" ||
 			rr[0].LBEndpoints[0].Metrics == nil {
 			t.Fatal("failed to set the endpoints")
 		}
@@ -124,7 +124,7 @@ func TestSelectAlgorithm(t *testing.T) {
 
 		if len(rr[0].LBEndpoints) != 1 ||
 			rr[0].LBEndpoints[0].Scheme != "https" ||
-			rr[0].LBEndpoints[0].Host != "www.example.org" ||
+			rr[0].LBEndpoints[0].Host != "www.example.org:443" ||
 			rr[0].LBEndpoints[0].Metrics == nil {
 			t.Fatal("failed to set the endpoints")
 		}
@@ -151,7 +151,7 @@ func TestSelectAlgorithm(t *testing.T) {
 
 		if len(rr[0].LBEndpoints) != 1 ||
 			rr[0].LBEndpoints[0].Scheme != "https" ||
-			rr[0].LBEndpoints[0].Host != "www.example.org" ||
+			rr[0].LBEndpoints[0].Host != "www.example.org:443" ||
 			rr[0].LBEndpoints[0].Metrics == nil {
 			t.Fatal("failed to set the endpoints")
 		}
@@ -258,8 +258,9 @@ func TestApply(t *testing.T) {
 			rt := p.Do([]*routing.Route{r})
 
 			lbctx := &routing.LBContext{
-				Request: req,
-				Route:   rt[0],
+				Request:  req,
+				Route:    rt[0],
+				Registry: routing.NewEndpointRegistry(routing.RegistryOptions{}),
 			}
 
 			h := make(map[string]int)
@@ -314,7 +315,12 @@ func TestConsistentHashBoundedLoadSearch(t *testing.T) {
 		},
 	}})[0]
 	ch := route.LBAlgorithm.(*consistentHash)
-	ctx := &routing.LBContext{Request: r, Route: route, Params: map[string]interface{}{ConsistentHashBalanceFactor: 1.25}}
+	ctx := &routing.LBContext{
+		Request:  r,
+		Route:    route,
+		Params:   map[string]interface{}{ConsistentHashBalanceFactor: 1.25},
+		Registry: routing.NewEndpointRegistry(routing.RegistryOptions{}),
+	}
 	noLoad := ch.Apply(ctx)
 	nonBounded := ch.Apply(&routing.LBContext{Request: r, Route: route, Params: map[string]interface{}{}})
 
@@ -322,21 +328,21 @@ func TestConsistentHashBoundedLoadSearch(t *testing.T) {
 		t.Error("When no endpoints are overloaded, the chosen endpoint should be the same as standard consistentHash")
 	}
 	// now we know that noLoad is the endpoint which should be requested for somekey if load is not an issue.
-	addInflightRequests(noLoad, 20)
+	addInflightRequests(ctx.Registry, noLoad, 20)
 	failover1 := ch.Apply(ctx)
 	if failover1 == nonBounded {
 		t.Error("When the selected endpoint is overloaded, the chosen endpoint should be different to standard consistentHash")
 	}
 
 	// now if 2 endpoints are overloaded, the request should go to the final endpoint
-	addInflightRequests(failover1, 20)
+	addInflightRequests(ctx.Registry, failover1, 20)
 	failover2 := ch.Apply(ctx)
 	if failover2 == nonBounded || failover2 == failover1 {
 		t.Error("Only the final endpoint had load below the average * balanceFactor, so it should have been selected.")
 	}
 
 	// now all will have same load, should select the original endpoint again
-	addInflightRequests(failover2, 20)
+	addInflightRequests(ctx.Registry, failover2, 20)
 	allLoaded := ch.Apply(ctx)
 	if allLoaded != nonBounded {
 		t.Error("When all endpoints have the same load, the consistentHash endpoint should be chosen again.")
@@ -386,7 +392,12 @@ func TestConsistentHashBoundedLoadDistribution(t *testing.T) {
 	}})[0]
 	ch := route.LBAlgorithm.(*consistentHash)
 	balanceFactor := 1.25
-	ctx := &routing.LBContext{Request: r, Route: route, Params: map[string]interface{}{ConsistentHashBalanceFactor: balanceFactor}}
+	ctx := &routing.LBContext{
+		Request:  r,
+		Route:    route,
+		Params:   map[string]interface{}{ConsistentHashBalanceFactor: balanceFactor},
+		Registry: routing.NewEndpointRegistry(routing.RegistryOptions{}),
+	}
 
 	for i := 0; i < 100; i++ {
 		ep := ch.Apply(ctx)
@@ -399,6 +410,7 @@ func TestConsistentHashBoundedLoadDistribution(t *testing.T) {
 			t.Errorf("Expected in-flight requests for each endpoint to be less than %d. In-flight request counts: %d, %d, %d", limit, ifr0, ifr1, ifr2)
 		}
 		ep.Metrics.IncInflightRequest()
+		ctx.Registry.IncInflightRequest(ep.Host)
 	}
 }
 
@@ -417,9 +429,10 @@ func TestConsistentHashKeyDistribution(t *testing.T) {
 	}
 }
 
-func addInflightRequests(endpoint routing.LBEndpoint, count int) {
+func addInflightRequests(registry *routing.EndpointRegistry, endpoint routing.LBEndpoint, count int) {
 	for i := 0; i < count; i++ {
 		endpoint.Metrics.IncInflightRequest()
+		registry.IncInflightRequest(endpoint.Host)
 	}
 }
 
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -244,6 +244,11 @@ type Params struct {
 	// It allows to add additional logic (for example tracing) by providing a wrapper function
 	// which accepts original skipper http.RoundTripper as an argument and returns a wrapped roundtripper
 	CustomHttpRoundTripperWrap func(http.RoundTripper) http.RoundTripper
+
+	// Registry provides key-value API which uses "host:port" string as a key
+	// and returns some metadata about endpoint. Information about the metadata
+	// returned from the registry could be found in routing.Metrics interface.
+	EndpointRegistry *routing.EndpointRegistry
 }
 
 type (
@@ -323,6 +328,7 @@ type Proxy struct {
 	maxLoops                 int
 	defaultHTTPStatus        int
 	routing                  *routing.Routing
+	registry                 *routing.EndpointRegistry
 	roundTripper             http.RoundTripper
 	priorityRoutes           []PriorityRoute
 	flags                    Flags
@@ -471,7 +477,7 @@ func setRequestURLForLoadBalancedBackend(u *url.URL, rt *routing.Route, lbctx *r
 
 // creates an outgoing http request to be forwarded to the route endpoint
 // based on the augmented incoming request
-func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHeaders bool) (*http.Request, *routing.LBEndpoint, error) {
+func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHeaders bool, registry *routing.EndpointRegistry) (*http.Request, *routing.LBEndpoint, error) {
 	var endpoint *routing.LBEndpoint
 	r := ctx.request
 	rt := ctx.route
@@ -484,7 +490,7 @@ func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHea
 		setRequestURLFromRequest(u, r)
 		setRequestURLForDynamicBackend(u, stateBag)
 	case eskip.LBBackend:
-		endpoint = setRequestURLForLoadBalancedBackend(u, rt, &routing.LBContext{Request: r, Route: rt, Params: stateBag})
+		endpoint = setRequestURLForLoadBalancedBackend(u, rt, &routing.LBContext{Request: r, Route: rt, Params: stateBag, Registry: registry})
 	default:
 		u.Scheme = rt.Scheme
 		u.Host = rt.Host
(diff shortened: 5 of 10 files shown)


endpointregistry/epregistry.go
```diff
@@ -0,0 +1,62 @@
...
+	"time"
+)
+
+type entry struct {
+	detected         time.Time
+	inflightRequests int64
+}
+
+var data map[string]entry = map[string]entry{}
+var m sync.Mutex
```
AlexanderYastrebov, line 14: Lets define a type and have methods on it. This way we can throw it away or have several of them (e.g. for tests, or to run multiple proxies and dataclients). In general try to avoid global variables.

endpointregistry/epregistry.go
```diff
@@ -0,0 +1,62 @@
...
+	defer m.Unlock()
+
+	data = map[string]entry{}
+}
+
+func SetDetectedTime(key string, detected time.Time) {
+	m.Lock()
+	defer m.Unlock()
+
+	e := data[key]
```
AlexanderYastrebov, line 27: What happens if the key already exists?

endpointregistry/epregistry.go
```diff
@@ -0,0 +1,62 @@
...
+	e.detected = detected
+	data[key] = e
+}
+
+func GetDetectedTime(key string) time.Time {
+	m.Lock()
+	defer m.Unlock()
+
+	return data[key].detected
```
AlexanderYastrebov, line 36: What if the key does not exist?

endpointregistry/epregistry.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func GetInflightRequests(key string) int {
```
AlexanderYastrebov, line 39: Why int?

filters/fadein/fadein.go
```diff
@@ -236,12 +237,12 @@ func (p *postProcessor) Do(r []*routing.Route) []*routing.Route {
...
 
 			key := endpointKey(s, h)
-			detected := p.detected[key].when
+			detected := endpointregistry.GetDetectedTime(h)
 			if detected.IsZero() || endpointsCreated[key].After(detected) {
 				detected = now
 			}
 
-			ep.Detected = detected
+			endpointregistry.SetDetectedTime(h, detected)
```
AlexanderYastrebov, line 245: Looks like you decided to keep existing fadeIn machinery for now, didn't you?

We can also remove stale endpoints around  `delete(p.detected, key)`

endpointregistry/epregistry.go
```diff
@@ -0,0 +1,62 @@
+package endpointregistry
```
AlexanderYastrebov, line 1: Lets name `endpoints` maybe?

filters/fadein/fadein_test.go
```diff
@@ -234,6 +235,9 @@ func TestPostProcessor(t *testing.T) {
 			baz: Path("/baz") -> <"http://10.0.1.1:8080", "http://10.0.1.2:8080">
 		`
 
+		// This Clear should be in the beginning of every test for tests not to interact
+		// between each other.
+		endpointregistry.Clear()
```
AlexanderYastrebov, line 240: https://github.com/zalando/skipper/pull/2535/files#r1301908469

filters/fadein/fadein.go
```diff
@@ -236,12 +237,12 @@ func (p *postProcessor) Do(r []*routing.Route) []*routing.Route {
 			}
 
 			key := endpointKey(s, h)
-			detected := p.detected[key].when
+			detected := endpointregistry.GetDetectedTime(h)
```
AlexanderYastrebov, line 240: We need to pay attention/define what is the endpoint key. Here you use hoist and fadeIn uses scheme+host

filters/fadein/fadein.go
AlexanderYastrebov: I think we need a dedicated post processor to track endpoints instead of plugging here

RomanZavodskikh: Yeah, in the scope of the current PR I am going only to introduce endpointregistry. When this component will be ready, I will continue to work on #2381 as I will have a way to implement in a clean way.

endpointregistry/epregistry.go
szuecs: @AlexanderYastrebov do you have partial inflight requests that you would count with fractions of int?

filters/fadein/fadein.go
szuecs: We get scheme+host from kubernetes endpoints IIRC, that's why we use it for redis, too. Maybe we can do this along with redis.

endpointregistry/epregistry.go
AlexanderYastrebov: I mean why we cast int64 to int here

endpoints/epregistry.go
```diff
@@ -0,0 +1,70 @@
+package endpoints
```
AlexanderYastrebov, line 1: rename to registry.go

endpoints/epregistry.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"sync"
+	"time"
+)
+
+type entry struct {
```
AlexanderYastrebov, line 8: Note: maybe we should consider something like this
```go
type Metrics interface {
   InflightRequests() int64
   DetectedTime() time.Time
}

func (r *Registry) getMetrics(key string) Metrics
```

filters/fadein/fadein.go
```diff
@@ -236,12 +236,12 @@ func (p *postProcessor) Do(r []*routing.Route) []*routing.Route {
 			}
 
 			key := endpointKey(s, h)
-			detected := p.detected[key].when
+			detected := ri.Registry.GetDetectedTime(h)
```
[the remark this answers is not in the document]
AlexanderYastrebov: Lets document what registry expects as a key

RomanZavodskikh:
It's better to create test for endpoint registry (thought it may look very simple).

filters/fadein/fadein.go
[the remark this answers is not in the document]
RomanZavodskikh: https://github.com/zalando/skipper/pull/2535/files#diff-15c9e8a131637d0caed5f79ff829ce894fea1803e4c3edbd67fe4bfea5919f42R39

endpoints/registry.go
```diff
@@ -0,0 +1,85 @@
...
+
+type Metrics interface {
+	DetectedTime() time.Time
+	InflightRequests() int64
+}
+
+type entry struct {
+	detected         time.Time
+	inflightRequests int64
+}
```
AlexanderYastrebov, line 16: you may add
```go
var _ Metrics = &entry{}
```
to hint reader that it implements interface

endpoints/registry.go
```diff
@@ -0,0 +1,85 @@
...
+	return &Registry{data: map[string]*entry{}}
+}
+
+func (r *Registry) GetMetrics(key string) Metrics {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	e := r.getOrInitEntryLocked(key)
+	copy := &entry{}
+	*copy = *e
```
AlexanderYastrebov, line 59: No need to copy as there is no way to mutate value via Metrics interface

endpoints/registry_test.go
```diff
@@ -0,0 +1,46 @@
...
+	r := NewRegistry(RegistryOptions{})
+
+	mBefore := r.GetMetrics("some key")
+	r.IncInflightRequest("some key")
+	mAfter := r.GetMetrics("some key")
+
+	assert.Equal(t, int64(0), mBefore.InflightRequests())
+	assert.Equal(t, int64(1), mAfter.InflightRequests())
+
+	ts, _ := time.Parse("2006-01-02", "2023-08-29")
```
AlexanderYastrebov, line 28: Use predefined format constant for time.Parse, see https://pkg.go.dev/time#pkg-constants

loadbalancer/algorithm_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"testing"
 
+	endpoints_pkg "github.com/zalando/skipper/endpoints"
```
AlexanderYastrebov, line 9: Why not `endpoints`? You can rename local variable if needed

proxy/proxy.go
```diff
@@ -244,6 +246,8 @@ type Params struct {
 	// It allows to add additional logic (for example tracing) by providing a wrapper function
 	// which accepts original skipper http.RoundTripper as an argument and returns a wrapped roundtripper
 	CustomHttpRoundTripperWrap func(http.RoundTripper) http.RoundTripper
+
+	Registry *endpoints.Registry
```
AlexanderYastrebov, line 250: Godoc please

proxy/proxy.go
```diff
@@ -829,6 +867,14 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 	if endpoint != nil {
 		endpoint.Metrics.IncInflightRequest()
 		defer endpoint.Metrics.DecInflightRequest()
+
+		_, h, err := normalizeSchemeHost(endpoint.Scheme, endpoint.Host)
```
AlexanderYastrebov, line 871: This should not happen on the request path. Either during route creation or not at all.

proxy/upgrade_test.go
```diff
@@ -359,8 +360,9 @@ func TestAuditLogging(t *testing.T) {
 			defer dc.Close()
 
 			rt := routing.New(routing.Options{
-				DataClients: []routing.DataClient{dc},
-				Log:         tl,
+				DataClients:    []routing.DataClient{dc},
+				PostProcessors: []routing.PostProcessor{routing.NewEndpointsRegistry(endpoints.NewRegistry(endpoints.RegistryOptions{}))},
```
AlexanderYastrebov, line 364: Why do we need registry in this test?

endpoints/registry.go
RomanZavodskikh: But there is a way to mutate: if user of `endpoints.Registry` will use `endpoint.Registry.SetDetectedTime` with the same key, it will get the same pointer and will overwrite it, for the user called `endpoint.Registry.GetMetrics` as well.

proxy/upgrade_test.go
RomanZavodskikh: Was added while experimenting with registry. Of course, it's not needed here, removed.

routing/registry.go
```diff
@@ -0,0 +1,142 @@
...
+	for _, route := range routes {
+		for i := range route.LBEndpoints {
+			epi := &route.LBEndpoints[i]
+			s, h, err := normalizeSchemeHost(epi.Scheme, epi.Host)
+			if err != nil {
+				log.Errorf(configErrFmt, epi.Scheme, epi.Host, err)
+				continue
+			}
+
+			epi.Scheme = s
```
RomanZavodskikh, line 81: It's a good idea to normalize scheme+host here. However, it would be better to normalize them at the very creation of LBEndpoint structure.

routing/endpoint_registry.go
```diff
@@ -0,0 +1,147 @@
...
+	case p == "" && s == "https":
+		p = "443"
+	}
+
+	h = net.JoinHostPort(h, p)
+	return s, h, nil
+}
+
+func (r *EndpointRegistry) Do(routes []*Route) []*Route {
+	const configErrFmt = "Error while processing endpoint fade-in settings: %s, %s, %v."
```
AlexanderYastrebov, line 71: Please inline this - it is easier to read format string along with the values

routing/endpoint_registry.go
```diff
@@ -0,0 +1,147 @@
...
+
+	data map[string]*entry
+}
+
+var _ PostProcessor = &EndpointRegistry{}
+
+type RegistryOptions struct {
+}
+
+func normalizeSchemeHost(s, h string) (string, string, error) {
```
AlexanderYastrebov, line 43: Please move unexported methods after exported

proxy/proxy.go
```diff
@@ -244,6 +244,11 @@ type Params struct {
 	// It allows to add additional logic (for example tracing) by providing a wrapper function
 	// which accepts original skipper http.RoundTripper as an argument and returns a wrapped roundtripper
 	CustomHttpRoundTripperWrap func(http.RoundTripper) http.RoundTripper
+
+	// Registry provides key-value API which uses host from normalizeSchemeHost as a key
+	// and returns some metadata about endpoint which could be used to perform better load balancing,
+	// fadeIn, etc.
```
AlexanderYastrebov, line 250: I guess you mean that key is "host:port" string.
Also omit "which could be used to perform better load balancing, fadeIn, etc.".
User can see details about available data in the Metrics type docs.

routing/endpoint_registry.go
```diff
@@ -0,0 +1,147 @@
...
+		for i := range route.LBEndpoints {
+			epi := &route.LBEndpoints[i]
+			s, h, err := normalizeSchemeHost(epi.Scheme, epi.Host)
+			if err != nil {
+				log.Errorf(configErrFmt, epi.Scheme, epi.Host, err)
+				continue
+			}
+
+			epi.Scheme = s
+			epi.Host = h
```
AlexanderYastrebov, line 82: So here your idea is to make endpoint host to be always in "host:port" format. Maybe we should make it earlier during _pre_-processing of routes.

AlexanderYastrebov: Or in the (p *algorithmProvider) Do

RomanZavodskikh: The fact that confuses me about normalizing scheme and host in the **pre-processing** stage is that `routing.PreProcessor` preprocesses `eskip.Route` instead of `routing.Route`:
```
type PreProcessor interface {
	Do([]*eskip.Route) []*eskip.Route
}
```
instead of 
```
type PostProcessor interface {
	Do([]*Route) []*Route
}
```
I am not sure whether it is possible to persist normalized fields in `eskip.Route` in the right way. But I do agree it would be nice to do this as soon as possible.

RomanZavodskikh: Additionally, I do remember I tried to do this in the `(p *algorithmProvider) Do` method, but such validation makes some test in filters/fadein fail, for example https://github.com/zalando/skipper/blob/v0.17.32/filters/fadein/fadein_test.go#L261. So I decided not to normalize fields in the `algorithmProvider` postprocessor, but I am doing it just right after this postprocessor, see https://github.com/zalando/skipper/pull/2535/files#diff-9b0b16d06f27b640eb77f6ba0dee84df6a660528e5cccd81839e0330c4b92dd2R1889. It seems to me this place is soon enough for (almost) every other component to see normalized fields and late enough to be easily backward compatible.

routing/endpoint_registry.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+	}
+
+	for _, route := range routes {
+		for _, epi := range route.LBEndpoints {
+			metrics := r.GetMetrics(epi.Host)
+			if metrics.DetectedTime().IsZero() {
```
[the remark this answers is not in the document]
AlexanderYastrebov: There are no concurrent post-processors and if endpoint does not exist yet it is not in the active routes so I do not think there is a race somewhere. Or is there?

filters/fadein/fadein.go
```diff
@@ -229,13 +228,7 @@ func (p *postProcessor) Do(r []*routing.Route) []*routing.Route {
 		for i := range ri.LBEndpoints {
 			ep := &ri.LBEndpoints[i]
 
-			s, h, err := normalizeSchemeHost(ep.Scheme, ep.Host)
```
AlexanderYastrebov, line 232: Should we remove normalizeSchemeHost since it is not used here anymore?

loadbalancer/algorithm.go
```diff
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"math"
 	"math/rand"
+	stdlib_net "net"
```
AlexanderYastrebov, line 8: Maybe lest use "net" and alias skipper/net to "snet"

proxy/proxy.go
```diff
@@ -244,6 +244,11 @@ type Params struct {
 	// It allows to add additional logic (for example tracing) by providing a wrapper function
 	// which accepts original skipper http.RoundTripper as an argument and returns a wrapped roundtripper
 	CustomHttpRoundTripperWrap func(http.RoundTripper) http.RoundTripper
+
+	// Registry provides key-value API which uses "host:port" string as a key
+	// and returns some metadata about endpoint. Information about the metadata
+	// returned from the registry could be found in routing.metrics interface.
```
AlexanderYastrebov, line 250: Is it `routing.metrics`?

routing/endpoint_registry.go
```diff
@@ -0,0 +1,103 @@
+package routing
```
AlexanderYastrebov, line 1: The convention is not to use underscore in the filename.

routing/endpoint_registry.go
```diff
@@ -0,0 +1,103 @@
...
+
+import (
+	"sync"
+	"time"
+)
+
+// metrics describe the data about endpoint that could be
+// used to perform better load balancing, fadeIn, etc.
+type metrics interface {
```
AlexanderYastrebov, line 10: So we do not export it?

routing/endpoint_registry.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (r *EndpointRegistry) GetMetrics(key string) metrics {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	e := r.getOrInitEntryLocked(key)
+	copy := &entry{}
+	*copy = *e
+	return copy
```
AlexanderYastrebov, line 64: I think you can just `return *e`

filters/fadein/fadein.go
RomanZavodskikh: No, we should keep it, because it's still used in the `func (endpointCreated( CreateFilter ...`.
This function parses filter args from dataclient, so basically it parses user input which should be normalized as well.

proxy/proxy.go
RomanZavodskikh: At the moment its methods are used only in the `routing` Package, so I decided to keep it private until some other package will use some of those methods.

proxy/proxy.go
```diff
@@ -829,6 +840,9 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 	if endpoint != nil {
 		endpoint.Metrics.IncInflightRequest()
 		defer endpoint.Metrics.DecInflightRequest()
+
+		p.registry.IncInflightRequest(endpoint.Host)
+		defer p.registry.DecInflightRequest(endpoint.Host)
```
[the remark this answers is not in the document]
RomanZavodskikh: I have created a benchmark which calls `metrics.IncInflightRequests` concurrently. Depending on the number of concurrent goroutines this operation could take up to 200ns in the worst case. It does seem negligible to me.
image

routing/endpoint_registry.go
RomanZavodskikh: It is possible, yeah, but does not make sense from the performance point of view. You can see the version with non-pointer receiever on the left and pointer reciever on the right. This benchstat shows that it is better to leave this getter as it is.
```
                                       │ ./epregistry.txt │           ./epregistry_pr.txt           │
                                       │      sec/op      │    sec/op     vs base                   │
GetInflightRequests/1_goroutines-16         147.95n ±  3%   17.52n ±  7%  -88.16% (n=18+10)
GetInflightRequests/2_goroutines-16         181.65n ±  3%   61.22n ± 23%  -66.30% (p=0.000 n=18+10)
GetInflightRequests/3_goroutines-16         213.20n ±  5%   78.99n ±  6%  -62.95% (p=0.000 n=18+10)
GetInflightRequests/4_goroutines-16          224.0n ±  2%   112.9n ± 27%  -49.59% (p=0.000 n=18+10)
GetInflightRequests/5_goroutines-16          237.2n ±  2%   129.5n ± 13%  -45.42% (p=0.000 n=18+10)
GetInflightRequests/6_goroutines-16          241.2n ±  5%   153.0n ± 13%  -36.55% (p=0.000 n=18+10)
GetInflightRequests/7_goroutines-16          255.6n ±  2%   157.6n ± 19%  -38.35% (p=0.000 n=18+10)
GetInflightRequests/8_goroutines-16          264.7n ± 12%   161.9n ± 14%  -38.82% (p=0.000 n=18+10)
GetInflightRequests/12_goroutines-16         297.6n ±  4%   182.0n ± 19%  -38.84% (p=0.000 n=17+10)
GetInflightRequests/16_goroutines-16         312.6n ±  8%   186.0n ±  3%  -40.52% (p=0.000 n=17+10)
GetInflightRequests/24_goroutines-16         335.4n ±  8%   165.3n ± 12%  -50.72% (p=0.000 n=17+10)
GetInflightRequests/32_goroutines-16         337.0n ±  9%   137.7n ± 23%  -59.15% (p=0.000 n=17+10)
GetInflightRequests/48_goroutines-16         308.7n ±  6%   114.8n ± 11%  -62.81% (p=0.000 n=17+10)
GetInflightRequests/64_goroutines-16         299.5n ± 14%   110.4n ±  8%  -63.14% (p=0.000 n=17+10)
GetInflightRequests/128_goroutines-16        299.7n ±  8%   110.4n ±  4%  -63.15% (p=0.000 n=17+10)
GetInflightRequests/256_goroutines-16        304.6n ±  8%   114.0n ±  7%  -62.56% (p=0.000 n=17+10)
GetInflightRequests/512_goroutines-16        303.5n ±  4%   113.0n ±  6%  -62.75% (p=0.000 n=17+10)
GetInflightRequests/768_goroutines-16        314.4n ±  5%   121.4n ±  7%  -61.40% (p=0.000 n=17+10)
GetInflightRequests/1024_goroutines-16       286.0n ± 11%   120.4n ±  9%  -57.92% (p=0.000 n=17+10)
GetInflightRequests/1536_goroutines-16       273.1n ± 12%   116.8n ±  8%  -57.23% (p=0.000 n=17+10)
GetInflightRequests/2048_goroutines-16       247.8n ± 11%   116.3n ±  6%  -53.07% (p=0.000 n=17+9)
GetInflightRequests/4096_goroutines-16       218.7n ±  5%   104.9n ±  8%  -52.03% (p=0.000 n=17+9)
geomean                                      263.3n         112.7n        -57.18%
```

AlexanderYastrebov:
:+1:

routing/endpoint_registry.go
AlexanderYastrebov: I see, you return pointer to a copy.
It makes sense to check allocations in the bechmark as well (via b.ReportAllocs())

RomanZavodskikh: https://pkg.go.dev/testing#B.ReportAllocs
> It is equivalent to setting -test.benchmem

I have used this setting as well, and that is what I see in benchstat. Pointer receiver seems the best choice from any point of view.

```
                                       │ ./epregistry.txt │          ./epregistry_pr.txt           │
                                       │       B/op       │   B/op     vs base                     │
GetInflightRequests/1_goroutines-16            32.00 ± 0%   0.00 ± 0%  -100.00% (n=18+17)
GetInflightRequests/2_goroutines-16            32.00 ± 0%   0.00 ± 0%  -100.00% (n=18+17)
GetInflightRequests/3_goroutines-16            32.00 ± 0%   0.00 ± 0%  -100.00% (n=18+17)
GetInflightRequests/4_goroutines-16            32.00 ± 0%   0.00 ± 0%  -100.00% (n=18+17)
GetInflightRequests/5_goroutines-16            32.00 ± 0%   0.00 ± 0%  -100.00% (n=18+17)
GetInflightRequests/6_goroutines-16            32.00 ± 0%   0.00 ± 0%  -100.00% (n=18+17)
GetInflightRequests/7_goroutines-16            32.00 ± 0%   0.00 ± 0%  -100.00% (n=18+17)
GetInflightRequests/8_goroutines-16            32.00 ± 0%   0.00 ± 0%  -100.00% (n=18+17)
GetInflightRequests/12_goroutines-16           32.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=17)
GetInflightRequests/16_goroutines-16           32.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=17)
GetInflightRequests/24_goroutines-16           32.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=17)
GetInflightRequests/32_goroutines-16           32.00 ± 0%   0.00 ± 0%  -100.00% (n=17+16)
GetInflightRequests/48_goroutines-16           32.00 ± 0%   0.00 ± 0%  -100.00% (n=17+16)
GetInflightRequests/64_goroutines-16           32.00 ± 0%   0.00 ± 0%  -100.00% (n=17+16)
GetInflightRequests/128_goroutines-16          32.00 ± 0%   0.00 ± 0%  -100.00% (n=17+16)
GetInflightRequests/256_goroutines-16          32.00 ± 0%   0.00 ± 0%  -100.00% (n=17+16)
GetInflightRequests/512_goroutines-16          32.00 ± 0%   0.00 ± 0%  -100.00% (n=17+16)
GetInflightRequests/768_goroutines-16          32.00 ± 0%   0.00 ± 0%  -100.00% (n=17+16)
GetInflightRequests/1024_goroutines-16         32.00 ± 0%   0.00 ± 0%  -100.00% (n=17+16)
GetInflightRequests/1536_goroutines-16         32.00 ± 0%   0.00 ± 0%  -100.00% (n=17+16)
GetInflightRequests/2048_goroutines-16         32.00 ± 0%   0.00 ± 0%  -100.00% (n=17+16)
GetInflightRequests/4096_goroutines-16         32.00 ± 0%   0.00 ± 0%  -100.00% (n=17+16)
geomean                                        32.00                   ?                       ¹ ²
¹ summaries must be >0 to compute geomean
² ratios must be >0 to compute geomean

                                       │ ./epregistry.txt │           ./epregistry_pr.txt           │
                                       │    allocs/op     │ allocs/op   vs base                     │
GetInflightRequests/1_goroutines-16            1.000 ± 0%   0.000 ± 0%  -100.00% (n=18+17)
GetInflightRequests/2_goroutines-16            1.000 ± 0%   0.000 ± 0%  -100.00% (n=18+17)
GetInflightRequests/3_goroutines-16            1.000 ± 0%   0.000 ± 0%  -100.00% (n=18+17)
GetInflightRequests/4_goroutines-16            1.000 ± 0%   0.000 ± 0%  -100.00% (n=18+17)
GetInflightRequests/5_goroutines-16            1.000 ± 0%   0.000 ± 0%  -100.00% (n=18+17)
GetInflightRequests/6_goroutines-16            1.000 ± 0%   0.000 ± 0%  -100.00% (n=18+17)
GetInflightRequests/7_goroutines-16            1.000 ± 0%   0.000 ± 0%  -100.00% (n=18+17)
GetInflightRequests/8_goroutines-16            1.000 ± 0%   0.000 ± 0%  -100.00% (n=18+17)
GetInflightRequests/12_goroutines-16           1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=17)
GetInflightRequests/16_goroutines-16           1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=17)
GetInflightRequests/24_goroutines-16           1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=17)
GetInflightRequests/32_goroutines-16           1.000 ± 0%   0.000 ± 0%  -100.00% (n=17+16)
GetInflightRequests/48_goroutines-16           1.000 ± 0%   0.000 ± 0%  -100.00% (n=17+16)
GetInflightRequests/64_goroutines-16           1.000 ± 0%   0.000 ± 0%  -100.00% (n=17+16)
GetInflightRequests/128_goroutines-16          1.000 ± 0%   0.000 ± 0%  -100.00% (n=17+16)
GetInflightRequests/256_goroutines-16          1.000 ± 0%   0.000 ± 0%  -100.00% (n=17+16)
GetInflightRequests/512_goroutines-16          1.000 ± 0%   0.000 ± 0%  -100.00% (n=17+16)
GetInflightRequests/768_goroutines-16          1.000 ± 0%   0.000 ± 0%  -100.00% (n=17+16)
GetInflightRequests/1024_goroutines-16         1.000 ± 0%   0.000 ± 0%  -100.00% (n=17+16)
GetInflightRequests/1536_goroutines-16         1.000 ± 0%   0.000 ± 0%  -100.00% (n=17+16)
GetInflightRequests/2048_goroutines-16         1.000 ± 0%   0.000 ± 0%  -100.00% (n=17+16)
GetInflightRequests/4096_goroutines-16         1.000 ± 0%   0.000 ± 0%  -100.00% (n=17+16)
geomean                                        1.000                    ?                       ¹ ²
¹ summaries must be >0 to compute geomean
² ratios must be >0 to compute geomean
```

loadbalancer/algorithm_test.go
```diff
@@ -43,7 +43,7 @@ func TestSelectAlgorithm(t *testing.T) {
 
 		if len(rr[0].LBEndpoints) != 1 ||
 			rr[0].LBEndpoints[0].Scheme != "https" ||
-			rr[0].LBEndpoints[0].Host != "www.example.org" ||
+			rr[0].LBEndpoints[0].Host != "www.example.org:443" ||
```
szuecs, line 46: I guess this is fine, because it does not change the outgoing request nor change the metrics

routing/endpointregistry.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+	return routes
+}
+
+func NewEndpointRegistry(o RegistryOptions) *EndpointRegistry {
+	return &EndpointRegistry{data: map[string]*entry{}}
+}
+
+func (r *EndpointRegistry) GetMetrics(key string) metrics {
```
szuecs, line 57: Did you hit this in your benchmark with a lot of entries (maybe 10k) in the registry?

routing/endpointregistry_test.go
```diff
@@ -0,0 +1,74 @@
...
+		r := NewEndpointRegistry(RegistryOptions{})
+		wg := sync.WaitGroup{}
+		for i := 0; i < goroutines; i++ {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				for n := 0; n < b.N/goroutines; n++ {
+					r.IncInflightRequest(key)
+				}
+			}()
```
szuecs, line 63: I am not sure if the number of goroutines will matter here a lot. It could be that it runs one after the other without a lot of contention.

routing/endpointregistry.go
RomanZavodskikh: I added the code which adds a lot of entries in the benchmark, the results are essentially the same.

AlexanderYastrebov:
:+1:

szuecs:
:+1:

RomanZavodskikh:
:+1:

PR #2536, opened 2023-08-22
Remove unused Dockerfile

AlexanderYastrebov:
It was introduced by #1571 and is not used.
Skipper images are built from Dockerfiles located in the /packaging folder. This should also reduce the number of dependabot updates.


RomanZavodskikh:
:+1:

AlexanderYastrebov:
:+1:

PR #2540, opened 2023-08-22
fix: flaky test webhook

szuecs:
fix: flaky test webhook


AlexanderYastrebov:
:+1:

AlexanderYastrebov:
I think the real issue is that auth server sleep amount is very close to the webhook timeout:
```go
if ti.timeout {
	time.Sleep(time.Second + time.Millisecond)
}
```
I think we can reduce this to ~100ms and set webhook timeout to ~50ms - this way we have a large margin between sleep and timeout AND speedup the test.

Issue #2433, opened 2023-07-02
Changes relevant to go1.21 release


AlexanderYastrebov:
- [ ] https://github.com/zalando/skipper/pull/2228
- [ ] Use global random generator instead of rand.New() where seed does not matter. It is automatically seeded since go1.20 https://github.com/golang/go/issues/54880 and becomes lock-free in go 1.21 https://github.com/golang/go/issues/49892

AlexanderYastrebov:
Release notes https://go.dev/doc/go1.21

szuecs:
I don't like the "use go test -fullpath" idea. I tested it to see the output but I think it's better to have project scoped file names.

szuecs:
They mention the use of math/rand/v2 should be used because if some vendored package seeds the global random src.

Closing because all parts are done

PR #2543, opened 2023-08-23
fix: NewResponseController was added in Go1.20

szuecs:
fix: https://pkg.go.dev/net/http#NewResponseController was added in Go1.20 so we should also tell dependencies that this is a requirement

Updates https://github.com/zalando/skipper/issues/2433


szuecs:
:+1:

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2546, opened 2023-08-29
packaging: remove base image tag

AlexanderYastrebov:
When hash is specified the tag is ignored, see https://github.com/moby/moby/issues/37866

The tag apparently confused dependabot and prevents it from updating the image:
```
updater | 2023/08/29 00:21:53 INFO <job_714994438> Checking if library/alpine-3 latest needs updating
updater | 2023/08/29 00:21:53 INFO <job_714994438> Latest version is latest
  proxy | 2023/08/29 00:21:53 [018] HEAD https://registry.opensource.zalan.do:443/v2/library/alpine-3/manifests/latest
  proxy | 2023/08/29 00:21:53 [018] 200 https://registry.opensource.zalan.do:443/v2/library/alpine-3/manifests/latest
updater | 2023/08/29 00:21:53 INFO <job_714994438> Pull request already exists for library/alpine-3 with latest version latest
```

PR #2549, opened 2023-08-31
packaging: use latest label

AlexanderYastrebov:
Dependabot fails to update registry.opensource.zalan.do/library/alpine-3 base image hash, see https://github.com/dependabot/dependabot-core/issues/7387

This change
* removes image hash and re-introduces latest label. For multiarch and ghcr.io builds base image is specified via BASE_IMAGE build argument and also uses latest label.
* uses COPY instead of ADD following https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy
* prints alpine release version during build
* removes redundant commands
* does not touch Dockerfile.arm64 and Dockerfile.armv7 - they are almost identical to Dockerfile and we may unify and use a single Dockerfile for all builds later.

Followup on #2546

### DIFF
--- a/packaging/Dockerfile
+++ b/packaging/Dockerfile
@@ -1,16 +1,20 @@
 ARG BASE_IMAGE=default
-FROM registry.opensource.zalan.do/library/alpine-3@sha256:2213d4d74c39af5313b631cbde2630b4007755b280f0f6b98867f66103b76113 AS default
-FROM ${BASE_IMAGE}
-LABEL maintainer="Team Gateway&Proxy @ Zalando SE <team-gwproxy@zalando.de>"
-RUN apk --no-cache add ca-certificates && update-ca-certificates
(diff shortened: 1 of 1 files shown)


szuecs:
I think we had this in the past, but OpenSSF does not like `latest` , because there is no need that "latest" is ever updated and all the security tooling can not see that it updates AFAIK.

AlexanderYastrebov:
Ok we have a choice: please OpenSSF or update base image manually

szuecs:
@AlexanderYastrebov I think we should do both. We should make sure that we can revert the PR that is the "fix". Right now you refactor and fix, so I would suggest to do 2 PRs.

AlexanderYastrebov:
Ok, created https://github.com/zalando/skipper/pull/2554

AlexanderYastrebov:
Closing in favour of #2563

PR #2553, opened 2023-08-31
build: remove cicheck

### LINKED ISSUE #1854: Reconsider use of `redis` build tag
With in the https://github.com/zalando/skipper/pull/1853#discussion_r705322502 it was discovered that `redis` tag condition is not used properly and that redis-dependent tests were always executed.
This is not a bad thing to run as many tests as possible so maybe we should simply eliminate redis build tag altogether.

The idea behind the tag was to exclude tests that depend on the `redis-server` binary. The alternative could be to run all tests by default and introduce a make target to disable redis tests explicitly

AlexanderYastrebov:
Make cicheck is equvalent to shortcheck due to wrong redis build tag use, see https://github.com/zalando/skipper/issues/1854

This change removes redis build tag, build comments and replaces cicheck with the shortcheck target.

Fixes https://github.com/zalando/skipper/issues/1854


delivery.yaml
```diff
@@ -50,7 +50,7 @@ pipeline:
       fi
       export IMAGE MULTIARCH_IMAGE
 
-      make deps cicheck staticcheck gosec
+      make deps check-fmt vet staticcheck gosec shortcheck
```
AlexanderYastrebov, line 53: Move validators before tests to fail early

.github/workflows/master.yaml
```diff
@@ -20,11 +20,9 @@ jobs:
       - run: go version
       - run: make deps
       - run: make check-fmt
-      - run: make build
       - run: make vet
       - run: make staticcheck
       - run: make check-race
-      - run: make cicheck
```
AlexanderYastrebov, line 27: There is make check-race before

.github/workflows/master.yaml
```diff
@@ -20,11 +20,9 @@ jobs:
       - run: go version
       - run: make deps
       - run: make check-fmt
-      - run: make build
```
AlexanderYastrebov, line 23: check-race requires build target

.github/workflows/master.yaml
RomanZavodskikh: So, this step is valid, but redundant, right?

AlexanderYastrebov: yes

circuit/registry_test.go
```diff
@@ -1,5 +1,3 @@
-/* +build !race */
```
AlexanderYastrebov, line 1: This likely never worked as expected https://pkg.go.dev/go/build#hdr-Build_Constraints

RomanZavodskikh:
:+1:

AlexanderYastrebov:
:+1:

PR #2554, opened 2023-08-31
packaging: use latest label in Dockerfile

AlexanderYastrebov:
Dependabot fails to update registry.opensource.zalan.do/library/alpine-3 base image hash, see https://github.com/dependabot/dependabot-core/issues/7387

This change removes image hash and re-introduces latest label. For multiarch and ghcr.io builds base image is specified via BASE_IMAGE build argument and also uses latest label.

Followup on #2546


szuecs:
:+1:

AlexanderYastrebov:
https://github.com/zalando/skipper/issues/2557 flaked

AlexanderYastrebov:
:+1:

PR #2560, opened 2023-08-31
secrets: speedup TestCipherRefreshing

AlexanderYastrebov:
#0


AlexanderYastrebov:
It flakes this way

AlexanderYastrebov:
It has scrypt.Key inside :)

PR #2563, opened 2023-09-01
packaging: update Dockerfile

AlexanderYastrebov:
* uses COPY instead of ADD following https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy
* prints alpine release version during build
* removes redundant commands
* does not touch Dockerfile.arm64 and Dockerfile.armv7 - they are almost identical to Dockerfile and we may unify and use a single Dockerfile for all builds later.

Followup on #2546


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #427, opened 2017-09-02
add community to readme

szuecs:
to share our open proposal google drive, we should have a section about it in our readme

PR #2565, opened 2023-09-01
Feature: Kubernetes endpointslices to scale-out

szuecs:
Feature: Kubernetes endpointslices to scale out services to more than 1000 pods 
Related risk: [#427](https://github.bus.zalan.do/risk-register/zalando/issues/427)


dataclients/kubernetes/clusterclient.go
```diff
@@ -486,9 +580,23 @@ func (c *clusterClient) fetchClusterState() (*clusterState, error) {
...
+	// 	// TODO(sszuecs): ignore because of tests not having endpointslices
+	// 	return nil, err
+	// }
+
+	// TODO(sszuecs) deprecated
+	endpoints, err2 := c.loadEndpoints()
+	// if err != nil {
+	// 	return nil, err
+	// }
+	if endpoints == nil && endpointSlices == nil {
```
szuecs, line 594: any idea how to solve this good?
Basically it's enough to fetch one of these (endpointslices or endpoints) if for example permissions are not set correctly or we run in a cluster that has no endpointslices.

AlexanderYastrebov: Lets do endpoint-slices flag. if false fetch and use endpoints otherwise endpointslices

dataclients/kubernetes/ingressdefinitions.go
```diff
@@ -65,6 +65,145 @@ func (s service) getTargetPortByValue(p int) (*definitions.BackendPort, bool) {
...
+	}
+	for _, ep := range eps.Endpoints {
+		result = append(result, formatEndpointString(ep.Address, backendProtocol, port))
+	}
+
+	return result
+}
+
+type endpointSliceList struct {
+	Meta  *definitions.ListMetadata
```
AlexanderYastrebov, line 145: Should it be tagged as "`json:",inline"`"?

Maybe we can just use existing metadata here and simply add continue and remaining fields without introducing ListMetadata type?

dataclients/kubernetes/clusterclient.go
```diff
@@ -449,6 +457,92 @@ func (c *clusterClient) loadEndpoints() (map[definitions.ResourceID]*endpoint, e
...
+// be checked. We read all endpointslices and create de-duplicated
+// business objects skipperEndpointSlice instead of raw Kubernetes
+// objects, because we need just a clean list of load balancer
+// members. The returned map will return the full list of ready
+// non-terminating endpoints that should be in the load balancer of a
+// given service, check endpointSlice.ToResourceID().
+func (c *clusterClient) loadEndpointSlices() (map[definitions.ResourceID]*skipperEndpointSlice, error) {
+	var endpointSlices endpointSliceList
+
+	if err := c.getJSON(c.endpointSlicesURI+c.endpointSlicesLabelSelectors, &endpointSlices); err != nil {
```
AlexanderYastrebov, line 472: I think this should be a loop while .metadata.continue field is not empty.
It also should set default fetch limit.

AlexanderYastrebov: I think it also makes sense to fetch all pages of k8s objects first and then process/transform them in to return result in a separate loop.

dataclients/kubernetes/definitions/definitions.go
```diff
@@ -9,6 +9,16 @@ import (
 
 var errInvalidMetadata = errors.New("invalid metadata")
 
+type ListMetadata struct {
+	*Metadata
```
AlexanderYastrebov, line 13: I think ListMetadata does not embed standard metadata

dataclients/kubernetes/ingressdefinitions.go
AlexanderYastrebov: I see now that continue field is part of metadata. I think it should be tagged with "`json:"metadata"`"

dataclients/kubernetes/ingressdefinitions.go
```diff
@@ -65,6 +65,145 @@ func (s service) getTargetPortByValue(p int) (*definitions.BackendPort, bool) {
...
+	Items []*endpointSlice `json:"items"`
+}
+
+// see https://kubernetes.io/docs/reference/kubernetes-api/service-resources/endpoint-slice-v1/#EndpointSlice
+type endpointSlice struct {
+	Meta        *definitions.Metadata     `json:"Metadata"`
```
AlexanderYastrebov, line 151: Lowercase "metadata"

dataclients/kubernetes/ingressdefinitions.go
```diff
@@ -65,6 +65,145 @@ func (s service) getTargetPortByValue(p int) (*definitions.BackendPort, bool) {
 	return nil, false
 }
 
+// There are [1..N] Kubernetes endpointslices created for a single Kubernetes service.
+// Kubernetes endpointslices of a given service can have duplicates with different states.
+// Therefore Kubernetes endpointslices need to be de-duplicated before usage.
+// The business object skipperEndpointSlice is a de-duplicated endpoint list that concats all endpointslices of a given service into one slice of skipperEndpointSlice.
+type skipperEndpointSlice struct {
```
AlexanderYastrebov, line 72: Lets move this into separate enpointslices.go file

dataclients/kubernetes/ingressdefinitions.go
```diff
@@ -65,6 +65,145 @@ func (s service) getTargetPortByValue(p int) (*definitions.BackendPort, bool) {
...
+	Ports     []*endpointSlicePort
+}
+
+// Conditions have to be evaluated before creation
+type skipperEndpoint struct {
+	Address string
+	Zone    string
+}
+
+func (eps skipperEndpointSlice) getPort(protocol, pName string, pValue int) int {
```
AlexanderYastrebov, line 84: Lets use pointer receiver following https://dave.cheney.net/2016/03/19/should-methods-be-declared-on-t-or-t

dataclients/kubernetes/kubernetestest/api.go
```diff
@@ -281,6 +284,10 @@ func initNamespace(kinds map[string][]interface{}) (ns namespace, err error) {
 		return
 	}
 
+	if err = itemsJSON(&ns.endpointslices, kinds["EndpointSlice"]); err != nil {
```
AlexanderYastrebov, line 287: Should we model pagination here?

dataclients/kubernetes/ingressdefinitions.go
szuecs: We don't need "continue". It was a miss understanding by myself but I clarified with sig-network.
The "continue" only exists in Lists because it is about number of objects in lists (for example GET all pods and if this listing is more than 1000 objects). One endpointslice holds a list of endpoints but it not a list request but a GET of a single object.

dataclients/kubernetes/kubernetestest/api.go
szuecs: It would be nice to have but it's not the problem we need to solve, see also the comment above

dataclients/kubernetes/ingressdefinitions.go
szuecs: this will addressed in the next push

dataclients/kubernetes/definitions/definitions.go
szuecs: this will addressed in the next push

dataclients/kubernetes/ingressdefinitions.go
szuecs: this will addressed in the next push

dataclients/kubernetes/ingressdefinitions.go
szuecs: endpoints are also not in a separated file. I will split out both endpoints.go and endpointslices.go

dataclients/kubernetes/kubernetestest/api.go
szuecs: https://github.com/zalando/skipper/issues/2574

dataclients/kubernetes/ingressdefinitions.go
szuecs: But we use value receiver in all the other functions, too (service, endpoints). Do you mean I should change all?
To have only one different from the other makes no sense IMO.

AlexanderYastrebov: Up to you

dataclients/kubernetes/ingressdefinitions.go
szuecs: done

dataclients/kubernetes/ingressdefinitions.go
szuecs: done

dataclients/kubernetes/clusterclient.go
szuecs: https://github.com/zalando/skipper/issues/2574

dataclients/kubernetes/clusterclient.go
szuecs: done

dataclients/kubernetes/clusterclient.go
```diff
@@ -449,6 +459,88 @@ func (c *clusterClient) loadEndpoints() (map[definitions.ResourceID]*endpoint, e
...
+			resEps := make(map[string]*skipperEndpoint)
+			terminatingEps := make(map[string]struct{})
+
+			for _, ep := range epSlices[i].Endpoints {
+				// if conditions are nil then we need to treat is as ready
+				if ep.Conditions == nil {
+					resEps[ep.Addresses[0]] = &skipperEndpoint{
+						// Addresses [1..100] of the same AddressType, as kube-proxy we use the first
+						// see also https://github.com/kubernetes/kubernetes/issues/106267
+						Address: ep.Addresses[0],
```
AlexanderYastrebov, line 506: Lets intriduce local variable `address := ep.Addresses[0]` with the comment and use it everywhere

dataclients/kubernetes/clusterclient.go
```diff
@@ -449,6 +459,88 @@ func (c *clusterClient) loadEndpoints() (map[definitions.ResourceID]*endpoint, e
...
+						Zone:    ep.Zone,
+					}
+					continue
+				}
+
+				if ep.isTerminating() {
+					terminatingEps[ep.Addresses[0]] = struct{}{}
+					// if we had this one with a non termiating condition, we should delete it, because of eventual consistency it is actually terminating
```
AlexanderYastrebov, line 514: termiating -> terminating

dataclients/kubernetes/clusterclient.go
```diff
@@ -449,6 +459,88 @@ func (c *clusterClient) loadEndpoints() (map[definitions.ResourceID]*endpoint, e
...
+			for _, ep := range epSlices[i].Endpoints {
+				// if conditions are nil then we need to treat is as ready
+				if ep.Conditions == nil {
+					resEps[ep.Addresses[0]] = &skipperEndpoint{
+						// Addresses [1..100] of the same AddressType, as kube-proxy we use the first
+						// see also https://github.com/kubernetes/kubernetes/issues/106267
+						Address: ep.Addresses[0],
+						Zone:    ep.Zone,
+					}
+					continue
```
AlexanderYastrebov, line 509: Lets use `else if` instead of continue

dataclients/kubernetes/clusterclient.go
```diff
@@ -449,6 +459,88 @@ func (c *clusterClient) loadEndpoints() (map[definitions.ResourceID]*endpoint, e
...
+				if ep.isTerminating() {
+					terminatingEps[ep.Addresses[0]] = struct{}{}
+					// if we had this one with a non termiating condition, we should delete it, because of eventual consistency it is actually terminating
+					delete(resEps, ep.Addresses[0])
+					continue
+				}
+				// already known terminating?
+				if _, ok := terminatingEps[ep.Addresses[0]]; ok {
+					continue
+				}
```
AlexanderYastrebov, line 521: Does this mean it was observed termination but now it is not (because previous ep.isTerminating() was false)?

AlexanderYastrebov: Should we log a warning maybe?

dataclients/kubernetes/clusterclient.go
```diff
@@ -449,6 +459,88 @@ func (c *clusterClient) loadEndpoints() (map[definitions.ResourceID]*endpoint, e
...
+			continue
+		}
+
+		result[resID] = &skipperEndpointSlice{
+			Meta: epSlices[0].Meta,
+		}
+
+		for i := range epSlices {
+			resEps := make(map[string]*skipperEndpoint)
+			terminatingEps := make(map[string]struct{})
```
AlexanderYastrebov, line 498: Is the scope of these correct? They will be reset on each new slice...

dataclients/kubernetes/endpointslices.go
```diff
@@ -0,0 +1,154 @@
...
+	var result []string
+	for _, ep := range eps.Endpoints {
+		result = append(result, formatEndpointString(ep.Address, backendProtocol, port))
+	}
+
+	return result
+}
+
+func (eps *skipperEndpointSlice) targets(protocol, backendProtocol string) []string {
+	result := make([]string, 0)
```
AlexanderYastrebov, line 77: We can pre-allocate capacity of len(eps.Endpoints) here

dataclients/kubernetes/kube.go
```diff
@@ -564,7 +573,7 @@ func (c *Client) GetEndpointAddresses(ns, name string) []string {
 	if c.state == nil {
 		return nil
 	}
-	return c.state.GetEndpointsByName(ns, name, "TCP")
```
AlexanderYastrebov, line 567: This "TCP" looks suspicious.
We use GetEndpointAddresses only for redis updater and then trim "TCP://" prefix.
For endpoint slices now I see you need "TCP" too lookup port and "http" to format url.
Lets rename `backendProtocol` to `scheme` here and elsewhere.
I also think this change may break Redis updater because result will now have "http://" prefix, please check.

config/config.go
```diff
@@ -787,6 +789,7 @@ func (c *Config) ToOptions() skipper.Options {
 		WhitelistedHealthCheckCIDR:             whitelistCIDRS,
 		KubernetesPathMode:                     c.KubernetesPathMode,
 		KubernetesNamespace:                    c.KubernetesNamespace,
+		KubernetesEnableEndpointslices:         c.KubernetesEnableEndpointSlices,
```
AlexanderYastrebov, line 792: We also need to update KubernetesDataClientOptions() to have this flag in routesrv.

dataclients/kubernetes/clusterclient.go
szuecs: done

dataclients/kubernetes/clusterclient.go
szuecs: done

dataclients/kubernetes/clusterclient.go
szuecs: We do not want to have endpoints with `Terminating: true` in the list of load balancer members.
I don't expect a terminating pods to become `Terminating: false` again.
Endpointslices can have duplicate entries with different states, so if you have 2 slices and one endpoint in both (think also eventual consistency), it should be enough to know that one is terminating to drop it out of the member list.

dataclients/kubernetes/clusterclient.go
szuecs: I guess no, I think there is a missing test coverage here, too!

dataclients/kubernetes/endpointslices.go
szuecs: done

dataclients/kubernetes/kube.go
szuecs: The redis updater call path was also changed to trim http://. I think I want to re-do this in the future. "TCP" comes from the Kubernetes service (you can basically have their the values "TCP" or "UDP" or "STCP" and in our case it's always "TCP", because we do not support http3 nor quic as client).
On the other hand the backendProtocol (or scheme) is used to add the scheme to endpoints which is in general useful (and we might be able to get this from "appProtocol" in endpointslices) but for simplicity and as we support only http I think it's ok to have it hard coded right now.

Will rename backendProtocol->scheme

szuecs: done

config/config.go
szuecs: this was already done https://github.com/zalando/skipper/pull/2565/files#diff-9b0b16d06f27b640eb77f6ba0dee84df6a660528e5cccd81839e0330c4b92dd2R923

dataclients/kubernetes/clusterclient.go
szuecs: done

dataclients/kubernetes/clusterclient.go
szuecs: fixed

dataclients/kubernetes/clusterclient.go
szuecs: I had to change the order of the logic based on testing input.

szuecs:
Rebased and squashed most of the commits. Not sure if the refactoring commits are better to split out, but I can also rebase it into a single commit for the release notes.
wdyt @AlexanderYastrebov what would fit better the release notes workflow?

ThorbjoernG:
It's a pretty big change that touches some key features. I would be in favor of breaking it up into smaller changes and rolling out iterative.
changing the VERSION sounds reasonable.

szuecs:
Splitting is not really possible. Test coverage is ~90% and deployment will be validated before rollout.
See also https://github.com/zalando-incubator/kubernetes-on-aws/pull/6434 which will deploy the new feature as shadow stack and we can validate that we get the same results before rolling out to production.

dataclients/kubernetes/endpoints.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+type endpointID struct {
+	definitions.ResourceID
+	TargetPort string
+	Protocol   string
+}
+
+type endpoint struct {
+	Meta    *definitions.Metadata `json:"Metadata"`
```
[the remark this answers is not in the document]
szuecs: Great spot! Oops, we had it since some time: https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/ingressdefinitions.go#L41

fixed

szuecs:
CodeQL says: 
```
RequestError [HttpError]: API rate limit exceeded for installation ID 5260298. If you reach out to GitHub Support for help, please include the request ID 6463:243E:62D4D2:663DDE:65005276.
```

AlexanderYastrebov:
Rebased on top of master

szuecs:
Test runs comparing 200 clusters showed that the change seems to be creating the same routes.

szuecs:
:+1:

szuecs:
:+1:

AlexanderYastrebov:
:+1:

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2575, opened 2023-09-04
proxy: fix flaky TestAuditLogging

### LINKED ISSUE #2572: Flaky TestAuditLogging
```
--- FAIL: TestAuditLogging (0.00s)
    --- FAIL: TestAuditLogging/on (0.00s)
        upgrade_test.go:427: failed to enable audit log
```

AlexanderYastrebov:
The TestAuditLogging could flake because it checks audit log after receiving response message from the connection which may happen before write to audit log by io.MultiWriter has finished.

It could be reproduced via:
```
go test ./proxy -run=TestAuditLogging -count=10000 -failfast
```

This change writes to auditLog before writing to the connection which also would show failed message in the audit log in case write to connection fails.

Fixes #2572

### DIFF
--- a/proxy/upgrade.go
+++ b/proxy/upgrade.go
@@ -161,7 +161,7 @@ func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
 	done := make(chan struct{}, 2)
 
 	if p.useAuditLog {
-		copyAsync("backend->request+audit", backendConn, io.MultiWriter(requestHijackedConn, p.auditLogOut), done)
+		copyAsync("backend->request+audit", backendConn, io.MultiWriter(p.auditLogOut, requestHijackedConn), done)
 	} else {
 		copyAsync("backend->request", backendConn, requestHijackedConn, done)
 	}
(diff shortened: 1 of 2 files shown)


proxy/upgrade_test.go
```diff
@@ -344,10 +344,9 @@ func TestAuditLogging(t *testing.T) {
 		return func(t *testing.T) {
 			wss := httptest.NewServer(websocket.Handler(func(ws *websocket.Conn) {
 				if _, err := io.Copy(ws, ws); err != nil {
-					t.Fatal(err)
+					t.Error(err)
```
AlexanderYastrebov, line 347: Avoid "using t.Fatal in a goroutine"

proxy/upgrade.go
```diff
@@ -161,7 +161,7 @@ func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
 	done := make(chan struct{}, 2)
 
 	if p.useAuditLog {
-		copyAsync("backend->request+audit", backendConn, io.MultiWriter(requestHijackedConn, p.auditLogOut), done)
+		copyAsync("backend->request+audit", backendConn, io.MultiWriter(p.auditLogOut, requestHijackedConn), done)
```
szuecs, line 164: Does this help or is this unrelated?

AlexanderYastrebov: It is the fix - this way audit log is always written first. I think it also makes sense to have it in the audit log even if subsequent write to connection fails.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2578, opened 2023-09-07
[WIP] Add otel tracing implementation support

lucastt:
The idea of this PR is to add a new tracing implementation to skipper run options that would support OpenTelemetry (Otel), instead of the deprecated OpenTracing (Ot).

The initial plan is just to support Otel tracer creation and wrap it into a bridge tracer that implements Ot Tracer interface and pass this bridge tracer along to other modules.

With next PRs we would start to have Otel tracers as first class citizens in our structs and interfaces enabling a skipper user to use full Otel support in all functionalities.

PS:. This PR is in a draft state and may be changed a lot. This is only created here to facilitate discussion on the topic and not as a final proposition.


proxy/proxy.go
```diff
@@ -237,7 +245,7 @@ type Params struct {
 
 	// OpenTracing contains parameters related to OpenTracing instrumentation. For default values
 	// check OpenTracingParams
-	OpenTracing *OpenTracingParams
+	OpenTracing *TracingParams
```
lucastt, line 248: I'm not very happy with the name of this parameter. But I think it's a bit weird to have two `TracingParams`, one for OT and other for OTel. I'm open to suggestions.

proxy/proxy.go
```diff
@@ -118,9 +119,12 @@ const (
 	OptionsHopHeadersRemoval = Options(HopHeadersRemoval)
 )
 
-type OpenTracingParams struct {
-	// Tracer holds the tracer enabled for this proxy instance
-	Tracer ot.Tracer
+type TracingParams struct {
+	// OtTracer holds the tracer enabled for this proxy instance
+	OtTracer ot.Tracer
```
szuecs, line 124: That's not backwards compatible change

proxy/proxy.go
```diff
@@ -118,9 +119,12 @@ const (
 	OptionsHopHeadersRemoval = Options(HopHeadersRemoval)
 )
 
-type OpenTracingParams struct {
-	// Tracer holds the tracer enabled for this proxy instance
-	Tracer ot.Tracer
+type TracingParams struct {
```
szuecs, line 122: That's not backwards compatible change

proxy/proxy.go
szuecs: We need to have a backwards compatible change, so better not to change the original type names.

proxy/tracing.go
```diff
@@ -67,40 +74,66 @@ func newProxyTracing(p *OpenTracingParams) *proxyTracing {
...
+	if t.useOtel && span.Otel == nil {
+		return
+	}
+
+	if !t.useOtel && span.Ot == nil {
+		return
+	}
+
+	if t.useOtel {
+		span.Otel.AddEvent(
```
szuecs, line 102: you can also implement the logic in `(*tracing.SpanWrapper).LogKV(k ,v string)`

proxy/tracing.go
```diff
@@ -67,40 +74,66 @@ func newProxyTracing(p *OpenTracingParams) *proxyTracing {
...
+	}
+
 	if !t.excludeTags[key] {
-		if s, ok := value.(string); ok {
-			span.SetTag(key, ensureUTF8(s))
+		if t.useOtel {
+			span.Otel.SetAttributes(attribute.String(key, ensureUTF8(value)))
 		} else {
-			span.SetTag(key, value)
+			span.Ot.SetTag(key, ensureUTF8(value))
```
szuecs, line 129: you can also implement `setTag()` in `(*tracing.SpanWrapper).setTag(k,v string)`

proxy/tracing.go
```diff
@@ -112,26 +145,61 @@ func (t *proxyTracing) startFilterTracing(operation string, ctx *context) *filte
...
+	span := &tracing.SpanWrapper{}
+
+	if t.useOtel {
+		var reqCtx goCtx.Context
+		reqCtx, span.Otel = t.otelTracer.Start(ctx.request.Context(), operation)
+		ctx.request.WithContext(reqCtx)
+		return &filterTracing{span, t.logFilterLifecycleEvents, t.useOtel}
+	}
+
+	span.Ot = tracing.CreateSpan(operation, ctx.request.Context(), t.otTracer)
```
szuecs, line 157: you can also implement `(*tracing.SpanWrapper).startFilterTracing(operation string, ctx *context)`

proxy/tracing.go
```diff
@@ -112,26 +145,61 @@ func (t *proxyTracing) startFilterTracing(operation string, ctx *context) *filte
...
-	if t != nil {
-		t.span.Finish()
+	if t == nil {
+		return
+	}
+
+	if t.useOtel {
+		t.span.Otel.End()
+	} else {
+		t.span.Ot.Finish()
```
szuecs, line 171: implement this if..else in `(*tracing.SpanWrapper).end()` or `.finish()`

proxy/tracing.go
```diff
@@ -112,26 +145,61 @@ func (t *proxyTracing) startFilterTracing(operation string, ctx *context) *filte
...
 
 func (t *filterTracing) logStart(filterName string) {
-	if t != nil && t.logEvents {
-		t.span.LogKV(filterName, StartEvent)
+	if t == nil || !t.logEvents {
+		return
+	}
+
+	if t.useOtel {
+		t.span.Otel.AddEvent(
```
szuecs, line 181: same here if..else you can do in the wrapper

proxy/tracing.go
```diff
@@ -112,26 +145,61 @@ func (t *proxyTracing) startFilterTracing(operation string, ctx *context) *filte
...
 
 func (t *filterTracing) logEnd(filterName string) {
-	if t != nil && t.logEvents {
-		t.span.LogKV(filterName, EndEvent)
+	if t == nil || !t.logEvents {
+		return
+	}
+
+	if t.useOtel {
+		t.span.Otel.AddEvent(
```
szuecs, line 197: here also

tracing/spanWrapper.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	ot "github.com/opentracing/opentracing-go"
+	opentelemetry "go.opentelemetry.io/otel/trace"
+)
+
+type SpanWrapper struct {
+	Ot      ot.Span
+	Otel    opentelemetry.Span
+	useOtel bool
+}
```
szuecs, line 12: Implement all methods here to dispatch which implementation to use. This should not leak to all the other packages.

tracing/tracers/otel/otel.go
```diff
@@ -0,0 +1,299 @@
...
+		otlptracehttp.WithEndpoint(net.JoinHostPort(tb.opt.host, strconv.Itoa(tb.opt.port))),
+	}
+
+	if tb.opt.insecure {
+		clientOptions = append(clientOptions, otlptracehttp.WithInsecure())
+	}
+
+	// OpenTelemetry only supports gzip or no compression:
+	// https://github.com/open-telemetry/opentelemetry-go/blob/3c476ce1816ae6f38758e90cc36d8b77ebcc223b/exporters/otlp/otlptrace/internal/otlpconfig/optiontypes.go#L29
+	if tb.opt.compressionMethod == "gzip" {
```
szuecs, line 219: We may want to return error in case wrong compression method give

tracing/tracers/otel/otel.go
```diff
@@ -0,0 +1,299 @@
...
+		otlptracegrpc.WithEndpoint(net.JoinHostPort(tb.opt.host, strconv.Itoa(tb.opt.port))),
+	}
+
+	if tb.opt.insecure {
+		clientOptions = append(clientOptions, otlptracegrpc.WithInsecure())
+	}
+
+	// OpenTelemetry only supports gzip or no compression:
+	// https://github.com/open-telemetry/opentelemetry-go/blob/exporters/otlp/otlptrace/v1.17.0/exporters/otlp/otlptrace/otlptracegrpc/options.go#L91
+	if tb.opt.compressionMethod == "gzip" {
```
szuecs, line 191: We may want to return error in case wrong compression method give

tracing/tracers/otel/otel.go
```diff
@@ -0,0 +1,299 @@
...
+	} else {
+		err = tb.setupGRPCExporter(ctx)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	tags := []attribute.KeyValue{
+		semconv.ServiceNameKey.String(tb.opt.ServiceName),
+		semconv.ServiceVersionKey.String(tb.opt.ServiceVersion), // ?
```
szuecs, line 247: what's the question?
Sounds good to have the version mandatory if you ask me

tracing/tracing.go
```diff
@@ -72,7 +74,7 @@ import (
 // tracer implementation name.
 func InitTracer(opts []string) (tracer ot.Tracer, err error) {
 	if len(opts) == 0 {
-		return nil, errors.New("opentracing: the implementation parameter is mandatory")
+		return nil, errors.New("tracing: the implementation parameter is mandatory")
```
szuecs, line 77: maybe better not to change this

tracing/tracing.go
```diff
@@ -93,6 +95,21 @@ func InitTracer(opts []string) (tracer ot.Tracer, err error) {
 	}
 }
 
+// TryOtelTracerWithBridge tries to initialize a OpenTelemetry tracer and a
+// OpenTelemetry bridge tracer. If the user did not specify otel as tracer
+// implementation, it will just return nil for both tracers and no error.
+func TryOtelTracerWithBridge(ctx context.Context, opts []string) (opentelemetry.Tracer, ot.Tracer, error) {
+	if len(opts) == 0 {
+		return nil, nil, errors.New("tracing: the implementation parameter is mandatory")
```
szuecs, line 103: maybe better to `tracing` -> `otel.tracing` or so to differentiate the log to the code. It's always good to be able to grep the log in source code and get only one line that shows up and not more than one.

tracing/tracing.go
```diff
@@ -145,6 +162,17 @@ func CreateSpan(name string, ctx context.Context, openTracer ot.Tracer) ot.Span
...
 }
 
+// CreateSpan creates a started span from an optional given parent from context
+func CreateOtelSpan(name string, ctx context.Context, tracer opentelemetry.Tracer) opentelemetry.Span {
+	parentSpan := opentelemetry.SpanFromContext(ctx)
+	if parentSpan == nil {
+		_, span := tracer.Start(ctx, name)
+		return span
+	}
+	tracer.
```
szuecs, line 172: that does not build :)

tracing/tracers/otel/otel.go
lucastt: Oh, that was a comment I forgot to delete.

filters/filters.go
```diff
@@ -107,12 +109,18 @@ type FilterContext interface {
 	// Allow filters to collect metrics other than the default metrics (Filter Request, Filter Response methods)
 	Metrics() Metrics
 
-	// Allow filters to add Tags, Baggage to the trace or set the ComponentName.
+	// Allow filters to add Tags, Baggage to open tracing trace or set the ComponentName.
 	Tracer() opentracing.Tracer
```
RomanZavodskikh, line 113: Is it okay to still keep FilterContext.{Tracer,ParentSpan} methods returning "deprecated" opentracing types exposed?

go.mod
```diff
@@ -46,6 +46,13 @@ require (
 	github.com/uber/jaeger-lib v2.4.1+incompatible
 	github.com/yookoala/gofast v0.7.0
 	github.com/yuin/gopher-lua v1.1.0
+	go.opentelemetry.io/otel v1.17.0
```
RomanZavodskikh, line 49: Is it okay to have v1.17.0 and v1.16.0 mixed here?

proxy/tracing.go
```diff
@@ -112,26 +110,34 @@ func (t *proxyTracing) startFilterTracing(operation string, ctx *context) *filte
...
+	ctx.request = ctx.request.WithContext(reqCtx)
 	ctx.parentSpan = span
 
 	return &filterTracing{span, t.logFilterLifecycleEvents}
 }
 
 func (t *filterTracing) finish() {
-	if t != nil {
-		t.span.Finish()
+	if t == nil {
```
RomanZavodskikh, line 122: minor: why did you decide to change those functions at all? It looks it's still better to leave the as it is:

+ the code before and after have the same results
+ the size of this PR will be decreased -> it will be slightly easier to read

tracing/tracing.go
```diff
@@ -93,6 +95,28 @@ func InitTracer(opts []string) (tracer ot.Tracer, err error) {
...
+// InitTracer function. Independently of wich tracer is created the function will return a
+// properly configured *TracerWrapper instance.
+func InitTracerWrapper(ctx context.Context, opts []string) (*TracerWrapper, error) {
+	var err error
+	tracer := &TracerWrapper{}
+	if len(opts) == 0 {
+		return nil, errors.New("tracerWrapper: tracing implementation was not specified")
+	}
+
+	if opts[0] != "otel" {
```
RomanZavodskikh, line 110: Should we return some error if `opts[0] == "someDefinitelyWrongString"`?

tracing/tracers/otel/otel.go
```diff
@@ -0,0 +1,299 @@
...
+	return t, otTracer, nil
+}
+
+func parseOptions(opts []string) (options, error) {
+	config := options{}
+	globalTags := make(map[string]string)
+
+	for _, o := range opts {
+		key, val, _ := strings.Cut(o, "=")
+		switch key {
```
RomanZavodskikh, line 83: Should we return and error when `key == "someDefinitelyWrongKey"`?

lucastt: I also agree that this behavior is desirable, but all other tracers just ignore keys that don't exist and fails only when a required value is not provided, so I just followed the same behavior for otel tracer.

proxy/tracing.go
lucastt: Because if I don't change it will expect a open tracing tracer interface and we no longer have access to this tracer in this part of the code, also the idea is to move away from open tracing so keep this interface here would not be productive for how we want this package to evolve.

go.mod
lucastt: Good catch, its fixed now. :)

filters/filters.go
lucastt: Yeah, the idea is to keep supporting both for the next year because currently all custom filter implementations that use tracing are still using open telemetry, we don't want to break backward compatibility and leave some time for users change their code base. I think something that we could improve is add a warning to these functions when used, so skipper users can see warnings in the logs that this functionality will be deprecated soon.

tracing/tracing.go
lucastt: I changed this, now everything is initialized in the same place and we only have one place that can produce this error.

filters/auth/authclient.go
```diff
@@ -36,9 +38,9 @@ type tokeninfoClient interface {
 
 var _ tokeninfoClient = &authClient{}
 
-func newAuthClient(baseURL, spanName string, timeout time.Duration, maxIdleConns int, tracer opentracing.Tracer) (*authClient, error) {
+func newAuthClient(baseURL, spanName string, timeout time.Duration, maxIdleConns int, tracer trace.Tracer) (*authClient, error) {
 	if tracer == nil {
-		tracer = opentracing.NoopTracer{}
+		tracer = &tracing.TracerWrapper{Ot: opentracing.NoopTracer{}}
```
szuecs, line 43: why not https://pkg.go.dev/go.opentelemetry.io/otel/trace@v1.24.0/noop#Tracer ?

filters/openpolicyagent/evaluation.go
```diff
@@ -12,10 +12,12 @@ import (
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/server"
 	"github.com/open-policy-agent/opa/tracing"
-	"github.com/opentracing/opentracing-go"
+	skipperTacing "github.com/zalando/skipper/tracing"
```
szuecs, line 15: skipperTacing -> skipperTracing

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest.go
```diff
@@ -124,7 +124,7 @@ func (f *opaAuthorizeRequestFilter) Request(fc filters.FilterContext) {
 	authzreq := envoy.AdaptToExtAuthRequest(req, f.opa.InstanceConfig().GetEnvoyMetadata(), f.envoyContextExtensions, rawBodyBytes)
 
 	start := time.Now()
-	result, err := f.opa.Eval(ctx, authzreq)
+	result, err := f.opa.Eval(ctx, fc.Tracer(), authzreq)
```
szuecs, line 127: better to pass `span`, then to pass the tracer. You do not need to get span from context within `opa.Eval`, or do I miss something?

filters/openpolicyagent/opaserveresponse/opaserveresponse.go
```diff
@@ -117,7 +117,7 @@ func (f *opaServeResponseFilter) Request(fc filters.FilterContext) {
 	authzreq := envoy.AdaptToExtAuthRequest(fc.Request(), f.opa.InstanceConfig().GetEnvoyMetadata(), f.envoyContextExtensions, rawBodyBytes)
 
 	start := time.Now()
-	result, err := f.opa.Eval(ctx, authzreq)
+	result, err := f.opa.Eval(ctx, fc.Tracer(), authzreq)
```
szuecs, line 120: same as https://github.com/zalando/skipper/pull/2578/files#r1530875775

filters/shedder/admission.go
```diff
@@ -195,7 +197,7 @@ type admissionControl struct {
 func NewAdmissionControl(o Options) filters.Spec {
 	tracer := o.Tracer
 	if tracer == nil {
-		tracer = &opentracing.NoopTracer{}
+		tracer = &tracing.TracerWrapper{Ot: &opentracing.NoopTracer{}}
```
szuecs, line 200: similar to https://github.com/zalando/skipper/pull/2578/files#r1530867916

net/httpclient.go
```diff
@@ -227,8 +235,12 @@ type Transport struct {
 // not leak a goroutine.
 func NewTransport(options Options) *Transport {
 	// set default tracer
-	if options.Tracer == nil {
-		options.Tracer = &opentracing.NoopTracer{}
+	if options.Tracer == nil && options.OtelTracer == nil {
+		options.OtelTracer = &tracing.TracerWrapper{Ot: &opentracing.NoopTracer{}}
```
szuecs, line 239: why not https://pkg.go.dev/go.opentelemetry.io/otel/trace@v1.24.0/noop#Tracer ?

net/httpclient.go
```diff
@@ -359,86 +371,82 @@ func (t *Transport) CloseIdleConnections() {
...
-	ext.HTTPMethod.Set(span, req.Method)
-	ext.SpanKind.Set(span, "client")
+	span.SetAttributes(attribute.String(tracing.ComponentTag, t.componentName))
+	span.SetAttributes(semconv.HTTPURL(req.URL.String()))
+	span.SetAttributes(semconv.HTTPMethod(req.Method))
+	span.SetAttributes(attribute.String(tracing.SpanKindTag, "client"))
 
-	_ = t.tracer.Inject(span.Context(), opentracing.HTTPHeaders, opentracing.HTTPHeadersCarrier(req.Header))
+	// Isso aqui acho que era melhor extrair da struct e receber a struct pra tentar fazer a conversao dentro do pacote
+	// Vai ficar mais limpo e assim tira a preocupacao de conversao de quem usa o wrapper.
```
szuecs, line 406: :)

predicates/traffic/segment_test.go
```diff
@@ -14,6 +15,7 @@ import (
 
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/filters/builtin"
+
```
szuecs, line 18: Pleas drop newlines

proxy/fadein_test.go
```diff
@@ -7,14 +7,14 @@ The following tests address the fade-in behavior expectations as observed from t
...
 Available data:
-- fade-in configuration: this contains the fade-in duration and optionally the fade-in degree.
-- detection time: this is the time when Skipper detected an endpoint that it didn't know about.
-- created time, optional: this is external information about the created of an endpoint, when available. E.g.
-  it can be the time when a Kubernetes pod became ready. It can be used as a fallback value for the detection
-  time in cases when Skipper (re)starts. When present, it renders any detection time past this value as invalid.
-- synced detection time, optional: it's the detection time of an endpoint shared across multiple Skipper
-  instances, stored in Redis or shared as a Skipper swarm datagram. It can be used to continue the fade-in when
-  Skipper gets (re)started. It has a precedence over the external created time in these cases.
+  - fade-in configuration: this contains the fade-in duration and optionally the fade-in degree.
```
szuecs, line 10: Please drop out this file in the PR

proxy/fadeintesting_test.go
```diff
@@ -534,10 +534,12 @@ func endpointStartTest(
 
 		p := startProxy(t, b)
 		defer p.close()
+
```
szuecs, line 537: Please drop out this file in the PR

proxy/proxy.go
```diff
@@ -844,7 +882,6 @@ func (p *Proxy) applyFiltersToRequest(f []*routing.RouteFilter, ctx *context) []
 		ctx.setMetricsPrefix(fi.Name)
 
 		fi.Request(ctx)
-
```
szuecs, line 847: please add the newline back

proxy/proxy.go
```diff
@@ -960,42 +999,40 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
...
+		p.tracing.setTag(ctx.proxySpan, ErrorTag, "true")
 
 		// Check if the request has been cancelled or timed out
 		// The roundtrip error `err` may be different:
 		// - for `Canceled` it could be either the same `context canceled` or `unexpected EOF` (net.OpError)
 		// - for `DeadlineExceeded` it is net.Error(timeout=true, temporary=true) wrapping this `context deadline exceeded`
-		if cerr := req.Context().Err(); cerr != nil {
-			ctx.proxySpan.LogKV("event", "error", "message", ensureUTF8(cerr.Error()))
+		if cerr := reqContext.Err(); cerr != nil {
+			p.tracing.logErrorEvent(ctx.proxySpan, cerr.Error())
```
szuecs, line 1027: are you sure you don't need `ensureUTF8`?
I think we should wrap `cerr.Error()` here

proxy/proxy.go
```diff
@@ -960,42 +999,40 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
...
 			if cerr == stdlibcontext.Canceled {
 				return nil, &proxyError{err: cerr, code: 499}
 			} else if cerr == stdlibcontext.DeadlineExceeded {
 				return nil, &proxyError{err: cerr, code: http.StatusGatewayTimeout}
 			}
 		}
 
-		ctx.proxySpan.LogKV("event", "error", "message", ensureUTF8(err.Error()))
+		p.tracing.logErrorEvent(ctx.proxySpan, err.Error())
```
szuecs, line 1035: are you sure you don't need `ensureUTF8`?
I think we should wrap `cerr.Error()` here

proxy/tracing.go
```diff
@@ -112,26 +121,33 @@ func (t *proxyTracing) startFilterTracing(operation string, ctx *context) *filte
...
+
+	_, span := t.Start(ctx.request.Context(), operation)
 	ctx.parentSpan = span
 
 	return &filterTracing{span, t.logFilterLifecycleEvents}
 }
 
 func (t *filterTracing) finish() {
-	if t != nil {
-		t.span.Finish()
```
szuecs, line 123: why not let the `!= nil` and replace the block with `t.span.End()` ?

proxy/tracing.go
```diff
@@ -112,26 +121,33 @@ func (t *proxyTracing) startFilterTracing(operation string, ctx *context) *filte
...
+	if t == nil {
+		return
 	}
+
+	t.span.End()
 }
 
 func (t *filterTracing) logStart(filterName string) {
-	if t != nil && t.logEvents {
```
szuecs, line 128: same as https://github.com/zalando/skipper/pull/2578/files#r1530916986 I would try to let all logic the same but replace the content of the block as necessary

proxy/tracing.go
```diff
@@ -112,26 +121,33 @@ func (t *proxyTracing) startFilterTracing(operation string, ctx *context) *filte
...
 	}
+
+	t.span.AddEvent(filterName, trace.WithAttributes(attribute.String(filterName, StartEvent)))
 }
 
 func (t *filterTracing) logEnd(filterName string) {
-	if t != nil && t.logEvents {
-		t.span.LogKV(filterName, EndEvent)
+	if t == nil || !t.logEvents {
+		return
```
szuecs, line 149: same as https://github.com/zalando/skipper/pull/2578/files#r1530917988

proxytest_output.txt
```diff
@@ -0,0 +1,4699 @@
+=== RUN   TestSlowService
```
szuecs, line 1: I guess this file you can delete from the PR

routesrv/routesrv.go
```diff
@@ -61,6 +61,7 @@ func New(opts skipper.Options) (*RouteServer, error) {
 	if len(opentracingOpts) == 0 {
 		opentracingOpts = []string{"noop"}
 	}
+
```
szuecs, line 64: please drop the whitespace change

skipper.go
```diff
@@ -1167,9 +1173,13 @@ func (o *Options) tlsConfig(cr *certregistry.CertRegistry) (*tls.Config, error)
 	return config, nil
 }
 
-func (o *Options) openTracingTracerInstance() (ot.Tracer, error) {
+func (o *Options) tracerInstance() (trace.Tracer, error) {
```
szuecs, line 1176: I don't see that we use this function in skipper.go. As far as I understand we should use this function to set "tracer" in run() or not?

szuecs: If we only use this func in test we can likely move it to a _test.go file

tracing/tracers/otel/otel.go
```diff
@@ -0,0 +1,320 @@
...
+		case "custom-headers":
+			config.customHeaders = make(map[string]string)
+			err := json.Unmarshal([]byte(val), &config.customHeaders)
+			if err != nil {
+				return options{}, fmt.Errorf("failed to parse %s as key-value custom headers: %w", val, err)
```
szuecs, line 101: the "val" should always be formatted by %q instead of %s

tracing/tracingtest/oteltesttracer.go
```diff
@@ -0,0 +1,268 @@
...
+	return &OtelSpan{
+		Trace:       t.TraceContent,
+		StartTime:   time.Now(),
+		tracer:      t,
+		spanContext: sc.WithTraceID(traceID),
+		Attributes:  make(map[string]interface{}),
+	}
+}
+
+// Create, start, and return a new Span with the given `operationName` if
```
szuecs, line 159: proper godoc requires to start with func name: `Start ...`

tracing/tracingtest/oteltesttracer.go
```diff
@@ -0,0 +1,268 @@
...
+			if ok {
+				t.TraceContent = tc
+			}
+		}
+	}
+
+	return trace.ContextWithSpan(context.WithValue(c, traceContentContextKey, t.TraceContent), s), s
+}
+
+// Returns all Ended spans that were created by this tracer.
```
szuecs, line 199: proper godoc requires to start with func name: `RecordedSpans ...`

tracing/tracingtest/oteltesttracer.go
```diff
@@ -0,0 +1,268 @@
...
+	return t.recordedSpans
+}
+
+func (t *OtelTracer) recordSpan(span *OtelSpan) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	t.recordedSpans = append(t.recordedSpans, span)
+}
+
+// Sets the end timestamp and finalizes Span state.
```
szuecs, line 212: `End ..`

tracing/tracingtest/oteltesttracer.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+}
+
+// SpanContext() yields the SpanContext for this Span. Note that the return
+// value of Context() is still valid after a call to Span.End().
+func (s *OtelSpan) SpanContext() trace.SpanContext {
+	return s.spanContext
+}
+
+// Sets or changes the operation name.
```
szuecs, line 226: `SetName ...`

tracing/tracingtest/oteltesttracer.go
```diff
@@ -0,0 +1,268 @@
...
+func (s *OtelSpan) SetName(operationName string) {
+	s.OperationName = operationName
+}
+
+// Adds a tag/attribute to the span.
```
szuecs, line 231: `SetAttributes ..`

tracing/tracingtest/oteltesttracer.go
```diff
@@ -0,0 +1,268 @@
...
+func (s *OtelSpan) SetAttributes(kv ...attribute.KeyValue) {
+	for _, attr := range kv {
+		s.Attributes[string(attr.Key)] = attr.Value.AsInterface()
+	}
+}
+
+// Add an Event to the span
```
szuecs, line 238: `AddEvent ...`

tracing/tracingtest/oteltesttracer.go
```diff
@@ -0,0 +1,268 @@
...
+func (s *OtelSpan) AddEvent(k string, opts ...trace.EventOption) {
+	ec := trace.NewEventConfig(opts...)
+	s.Events = append(s.Events, sdk.Event{
+		Name:       k,
+		Attributes: ec.Attributes(),
+		Time:       time.Now(),
+	})
+}
+
+// Returns wether this is a recording Span or not. For this implementation
```
szuecs, line 248: `IsRecording ...`

tracing/tracingtest/oteltesttracer.go
```diff
@@ -0,0 +1,268 @@
...
+// this is always true.
+func (s *OtelSpan) IsRecording() bool {
+	// Is there any moment that this is false for Opentelemetry?
+	return true
+}
+
+// Record an error into the span
```
szuecs, line 255: RecordError ...`

tracing/tracingtest/oteltesttracer.go
```diff
@@ -0,0 +1,268 @@
...
+func (s *OtelSpan) RecordError(err error, options ...trace.EventOption) {
+	s.AddEvent("error", trace.WithAttributes(attribute.String("message", err.Error())))
+}
+
+// Not implemented
```
szuecs, line 260: drop comment

tracing/tracingtest/oteltesttracer.go
```diff
@@ -0,0 +1,268 @@
...
+func (s *OtelSpan) SetStatus(code codes.Code, description string) {
+	panic("SetStatus is not implemented")
+}
+
+// Not implemented
```
szuecs, line 265: drop comment

tracing/wrappers.go
```diff
@@ -0,0 +1,168 @@
...
+	otelembedded "go.opentelemetry.io/otel/trace/embedded"
+)
+
+type otSpanContextKey struct{}
+
+var wireContextKey otSpanContextKey
+
+// TracerWrapper is a wrapper around OpenTracing tracer that implements
+// Open OpenTelemetry tracer interface, translating all calls to OpenTelemetry
+// API to OpenTracing API effectivelly enabling Skipper to operate with
```
szuecs, line 23: effectively

tracing/wrappers.go
```diff
@@ -0,0 +1,168 @@
...
+	"go.opentelemetry.io/otel/trace"
+	otelembedded "go.opentelemetry.io/otel/trace/embedded"
+)
+
+type otSpanContextKey struct{}
+
+var wireContextKey otSpanContextKey
+
+// TracerWrapper is a wrapper around OpenTracing tracer that implements
+// Open OpenTelemetry tracer interface, translating all calls to OpenTelemetry
```
szuecs, line 22: drop "Open"

tracing/wrappers.go
```diff
@@ -0,0 +1,168 @@
...
+// API to OpenTracing API effectivelly enabling Skipper to operate with
+// both OpenTelemetry and OpenTracing keeping retrocompatibility.
+type TracerWrapper struct {
+	otelembedded.Tracer
+	Ot ot.Tracer
+}
+
+// SpanWrapper is a wrapper around OpenTracing span that implements
+// Open OpenTelemetry Span interface, translating all calls to OpenTelemetry
```
szuecs, line 31: drop "Open"

tracing/wrappers.go
```diff
@@ -0,0 +1,168 @@
...
+// API to OpenTracing API effectivelly enabling Skipper to operate with
```
szuecs, line 32: effectively

tracing/wrappers.go
```diff
@@ -0,0 +1,168 @@
...
+// github.com/opentracing/opentracing-go/ext/tags.go
+const (
+	ComponentTag = "component"
+	SpanKindTag  = "span.kind"
+	ErrorTag     = "error"
+)
+
+// Start Creates a SpanWrapper adds it to Go stdlib context and finaly returns both the span and the context.
+// currently TracerWrapper.Start does not convert any Open telemetry start option to Open tracing start
+// options, the parameter 'opts' is mostly ignored.
```
szuecs, line 52: Drop "mostly" as we ignore it here and only have it for compatibility reasons

proxytest_output.txt
lucastt: Oops, sorry about that, I must have added it to the commit without noticing it.

filters/auth/authclient.go
lucastt: Yeah, I replaced this in a few places, I must have missed this one. Thanks :)

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest.go
lucastt: Yeah, sure we could just pass the span down to `Eval(...)` and get the span from the context in the Request function, if this is better.

net/httpclient.go
lucastt: 😄 🇧🇷

proxy/proxy.go
lucastt: I remember thinking about that, but I don't remember my conclusion, its been a while... 🤔 

I'll just wrap it for now.

proxy/tracing.go
lucastt: This is because this function went through a few changes, and before I had two different behaviors here because the logic to decide between Otel and Ot was leaking to this package. But now there is no reason for that, its no problem to have the block inside the `!= nil`.

skipper.go
lucastt: Hmm yeah, this should be used in the `run()` function, it actually seems to me that we are missing the definition of tracer in the options. in the `run()` function and even if it was correct there we would be just duplicating code, since all these checks are done in `tracerInstance(...)`anyway.

lucastt:
👍

filters/tee/tee.go
```diff
@@ -46,8 +45,8 @@ type Options struct {
 	// Timeout specifies a time limit for requests made by tee filter.
 	Timeout time.Duration
 
-	// Tracer is the opentracing tracer to use in the client
-	Tracer opentracing.Tracer
+	// Tracer is the open telemetry tracer to use in the client
+	Tracer trace.Tracer
```
MustafaSaber, line 49: sometimes I see `Tracer` and others `OtelTracer`, is this intended?

lucastt: Yeah, in some places I used `OtelTracer` to differentiate from `Tracer` and `OtTracer`, the former is only OT if we need to be backward compatible and the later is kept in places we need access to both tracers and I wanted to make it clear that it is one or the other. Marjority of cases it should be just `tracer` or `Tracer`.

lucastt:
Regarding how I tested these changes locally the strategy I used was unity tests + stdouttrace [1], which is a tracer that implements `otel.Tracer` interface and drops the span in JSON form in stdout or any other file.

I didn't test it with directly with specific collectors or vendors like lightstep, I was planning to run skipper on a custom cluster and test it with lightstep collectors there before deploying.

stdouttrace tests all behaviors supported by Otel SDK (OTLP) what is not well tested is the config part, the arguments we would pass in skipper parameters that would configure the communication between skipper tracer and a tracer collector or tracing commercial solution.

[1]: https://pkg.go.dev/go.opentelemetry.io/otel/exporters/stdout/stdouttrace

cc: @MustafaSaber

filters/tee/tee.go
lucastt: Also, when we discontinue OpenTracing,  which will probably take a while, we can have all of them as just Tracer/tracer.

scheduler/proxy_test.go
```diff
@@ -83,7 +83,7 @@ func TestFifoChanges(t *testing.T) {
 		t.Helper()
 
 		client := net.NewClient(net.Options{
-			ResponseHeaderTimeout: 2 * time.Second,
+			ResponseHeaderTimeout: 3 * time.Second,
```
MustafaSaber, line 86: why was this changed?

skipper.go
```diff
@@ -1521,24 +1524,26 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
-			NoFollow: false,
-		}),
-		// teenf()
-		teefilters.WithOptions(teefilters.Options{
-			NoFollow: true,
-			Tracer:   tracer,
-		}),
-	)
+	// tee filters override if we have a tracer
+	if len(o.OpenTracing) > 0 {
```
MustafaSaber, line 1533: does the override only apply for opentracing?

scheduler/proxy_test.go
lucastt: It was flaky and was failing most of the time in my local environment, increasing the timeout solved the problem.

skipper.go
lucastt: This option has a very unfortunate name, but it contains both OpenTracing and OpenTelemetry tracers. I could not change the name to something with more meaning than `Options.OpenTracing` to not break backward compatibility with library users, which is not great but we don't have much option. :/

tracing/tracers/otel/otel.go
```diff
@@ -0,0 +1,320 @@
...
+			config.exportTimeout = v
+		case "tag":
+			if val != "" {
+				tag, tagVal, found := strings.Cut(val, "=")
+				if !found {
+					return options{}, fmt.Errorf("missing value for tag %q", val)
+				}
+				globalTags[tag] = tagVal
+			}
+		case "otlp-exporter-endpoint":
```
szuecs, line 143: From testing in the cluster: maybe an URL?

net/httpclient.go
```diff
@@ -359,86 +372,80 @@ func (t *Transport) CloseIdleConnections() {
...
-func injectClientTrace(req *http.Request, span opentracing.Span) *http.Request {
+func injectClientTrace(req *http.Request, span trace.Span) *http.Request {
 	trace := &httptrace.ClientTrace{
 		DNSStart: func(httptrace.DNSStartInfo) {
-			span.LogKV("DNS", "start")
+			span.AddEvent("DNS", trace.WithAttributes(attribute.String("DNS", "start")))
 		},
 		DNSDone: func(httptrace.DNSDoneInfo) {
-			span.LogKV("DNS", "end")
+			span.AddEvent("DNS", trace.WithAttributes(attribute.String("DNS", "stop")))
```
AlexanderYastrebov, line 417: Why end changed to stop?

proxy/context.go
```diff
@@ -287,17 +289,19 @@ func (c *context) Split() (filters.FilterContext, error) {
...
+
 	originalRequest.Body = body
 	cc.request = cr
 	return cc, nil
 }
 
 func (c *context) Loopback() {
-	loopSpan := c.Tracer().StartSpan(c.proxy.tracing.initialOperationName, opentracing.ChildOf(c.ParentSpan().Context()))
-	defer loopSpan.Finish()
+	_, loopSpan := c.Tracer().Start(c.request.Context(), c.proxy.tracing.initialOperationName)
```
AlexanderYastrebov, line 302: What return value is ignored here and why?

proxy/proxy.go
```diff
@@ -120,9 +121,13 @@ const (
 )
 
 type OpenTracingParams struct {
-	// Tracer holds the tracer enabled for this proxy instance
+	// OtTracer holds the open telemetry tracer enabled for this proxy instance
+	// DEPRECATED, use TracerWrapper.
 	Tracer ot.Tracer
 
+	// OtelTracer holds the opentracing tracer enabled for this proxy instance
+	OtelTracer trace.Tracer
```
AlexanderYastrebov, line 129: Here and elsewhere we keep old and new tracer fields. 
Why not use only OTEL tracer and provide users an adapter for opentracing tracer?

proxy/proxy.go
```diff
@@ -827,6 +846,25 @@ func WithParams(p Params) *Proxy {
 	}
 }
 
+type otelTransport struct {
+	tr     *http.Transport
+	tracer trace.Tracer
+}
```
AlexanderYastrebov, line 852: Why do we need this? Somehow we created proxy span without it before.

proxy/proxy.go
```diff
@@ -1173,14 +1211,14 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 		ctx.ensureDefaultResponse()
 	} else if ctx.route.BackendType == eskip.LoopBackend {
 		loopCTX := ctx.clone()
-		loopSpan := tracing.CreateSpan("loopback", ctx.request.Context(), p.tracing.tracer)
+		_, loopSpan := p.tracing.Start(ctx.request.Context(), "loopback")
```
AlexanderYastrebov, line 1214: What return value is ignored here and why?

proxy/proxy.go
```diff
@@ -1310,7 +1354,7 @@ func (p *Proxy) serveResponse(ctx *context) {
 		p.tracing.setTag(ctx.proxySpan, ClientRequestStateTag, ClientRequestCanceled)
 	}
 
-	p.tracing.setTag(ctx.initialSpan, HTTPStatusCodeTag, uint16(ctx.response.StatusCode))
+	p.tracing.setTag(ctx.initialSpan, HTTPStatusCodeTag, fmt.Sprint(uint16(ctx.response.StatusCode)))
```
AlexanderYastrebov, line 1357: We used attributes elsewhere but here it is "tag", why is it so?

proxy/tracing.go
```diff
@@ -76,31 +81,35 @@ func newProxyTracing(p *OpenTracingParams) *proxyTracing {
...
 		return t
 	}
 
 	if !t.excludeTags[key] {
-		if s, ok := value.(string); ok {
-			span.SetTag(key, ensureUTF8(s))
-		} else {
-			span.SetTag(key, value)
-		}
+		span.SetAttributes(attribute.String(key, value))
```
AlexanderYastrebov, line 106: ensureUTF8 was lost with this change.

tracing/tracingtest/oteltesttracer.go
```diff
@@ -0,0 +1,266 @@
...
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/propagation"
+	sdk "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/trace"
+	otelembedded "go.opentelemetry.io/otel/trace/embedded"
+)
+
+// OtelTracer is an implementation of opentelemetry.OtelTracer for testing. It records
+// the defined spans during a series of operations.
+type OtelTracer struct {
```
AlexanderYastrebov, line 24: I am wondering  if we should take a chance and get rid of tracing and tracing/tracingtest packages and just use otel directly (and https://pkg.go.dev/go.opentelemetry.io/otel/oteltest).

AlexanderYastrebov: In any case why do we implement own test tracer instead of using https://pkg.go.dev/go.opentelemetry.io/otel/oteltest ?

AlexanderYastrebov (commented):
Thank you for this big effort. I did a first review pass and left my comments.

proxy/proxy.go
szuecs: Actually we do this from skipper.Run() point of view @lucastt , so we likely can delete proxy.Tracer

proxy/proxy.go
szuecs: This is an interesting question and I think it's how OTel is different from OT. Please confirm @lucastt

proxy/proxy.go
lucastt: Yeah, it is possible, I just considered this as something that people using skipper as a library could be using so I wanted to keep backward compatibility. Under the hood the old tracer is converted into a tracerWrapper instance that convert calls from OTEL to OT. 

So yeah, if this part does not need to be backward compatible it would be much better to have a single tracer instance.

proxy/proxy.go
lucastt: Yeah, maybe this is not necessary. Because the goal of this struct is to inject the span into the request, but we already do this in `net/httpclient.go:406` so the only case this struct would be necessary is if we make requests using this Roundtripper directly, without using skipper http client.

net/httpclient.go
lucastt: Oh, this is a mistake. Good catch!

proxy/context.go
lucastt: This is an stdlib context. This context doesn't seem to be used in the function.

One thing that could be done is update the request context so it carries the span information forward if any other function or package wants to use it. 

But in theory this was not done in the past either, `Ot.ContextWithSpan(...)` was never called here so the context was never updated, the only difference is that `Otel.Tracer.Start(...)` always returns a context, which was not true for OT implementation.

proxy/proxy.go
lucastt: Same reason as here: https://github.com/zalando/skipper/pull/2578/files/a25b1840dc0decb0dfda88d2ca0163452eca73cd#r1542959056

tracing/tracers/otel/otel.go
lucastt: Yeah, I'll try to test with a local OTLP collector to make sure no parameter is missing. But one of the things we need to change is this parameter should not be split by port and address only because we need the URL path to route it to the proper OTLP endpoint in most collectors.

tracing/tracingtest/oteltesttracer.go
lucastt: Good catch. This is a mistake, I didn't know oteltest package, this seems to be a better solution for tracing tests for now on. But maybe we could address this in a different PR? It seems to me this will bring a lot of new changes to an already very big PR. WDYT?

Issue #1546, opened 2020-10-09
log error information in rate limiting spans when they are marked as error


aryszka:

AlexanderYastrebov:
I think the problem is that tracing data becomes too large in case of high load and if redis starts to error. We already reduced tracing data size over several rounds in the past.
 
I think for ratelimits we should rather implement sampled logging like in https://github.com/zalando/skipper/pull/2194 or https://github.com/zalando/skipper/issues/2186

PR #2581, opened 2023-09-07
Add sampled error logging for redis errors

nabeelbukhari:
We observed that error message is not logged as part of span. This PR introduces sampled logging for redis errors.

Fixes https://github.com/zalando/skipper/issues/1546

### DIFF
--- a/ratelimit/redis.go
+++ b/ratelimit/redis.go
@@ -12,6 +12,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/zalando/skipper/metrics"
 	"github.com/zalando/skipper/net"
+	"golang.org/x/time/rate"
 )
 
 // clusterLimitRedis stores all data required for the cluster ratelimit.
@@ -23,6 +24,7 @@ type clusterLimitRedis struct {
 	window     time.Duration
 	ringClient *net.RedisRingClient
 	metrics    metrics.Metrics
+	sometimes  rate.Sometimes
 }
 
 const (
@@ -52,6 +54,7 @@ func newClusterRateLimiterRedis(s Settings, r *net.RedisRingClient, group string
 		window:     s.TimeWindow,
 		ringClient: r,
 		metrics:    metrics.Default,
+		sometimes:  rate.Sometimes{First: 3, Interval: 1 * time.Second},
 	}
 
 	return rl
@@ -121,6 +124,7 @@ func (c *clusterLimitRedis) Allow(ctx context.Context, clearText string) bool {
 	if failed {
 		allow = !c.failClosed
 		setError(span)
+		c.logError("Failed to determine if operation is allowed: %v", err)
 	}
 	if span != nil {
 		span.SetTag("allowed", allow)
@@ -194,7 +198,7 @@ func (c *clusterLimitRedis) Delta(clearText string) time.Duration {
 	now := time.Now()
 	d, err := c.deltaFrom(context.Background(), clearText, now)
 	if err != nil {
-		log.Errorf("Failed to redis get the duration until the next call is allowed: %v", err)
+		c.logError("Failed to get the duration until the next call is allowed: %v", err)
 
 		// Earlier, we returned duration since time=0 in these error cases. It is more graceful to the
 		// client applications to return 0.
@@ -208,6 +212,12 @@ func setError(span opentracing.Span) {
 	if span != nil {
 		ext.Error.Set(span, true)
 	}
+}
+
+func (c *clusterLimitRedis) logError(format string, err error) {
+	c.sometimes.Do(func() {
+		log.Errorf(format, err)
+	})
 }
 
 func (c *clusterLimitRedis) oldest(ctx context.Context, clearText string) (time.Time, error) {
@@ -257,7 +267,7 @@ func (c *clusterLimitRedis) oldest(ctx context.Context, clearText string) (time.
 func (c *clusterLimitRedis) Oldest(clearText string) time.Time {
 	t, err := c.oldest(context.Background(), clearText)
 	if err != nil {
-		log.Errorf("Failed to get from redis the oldest known request time: %v", err)
+		c.logError("Failed to get the oldest known request time: %v", err)
 		return time.Time{}
 	}
 
@@ -285,7 +295,7 @@ func (c *clusterLimitRedis) RetryAfterContext(ctx context.Context, clearText str
 
 	retr, err := c.deltaFrom(ctx, clearText, now)
 	if err != nil {
-		log.Errorf("Failed to get from redis the duration to wait with the next request: %v", err)
+		c.logError("Failed to get the duration to wait until the next request: %v", err)
 		queryFailure = true
 		return minWait
 	}


ratelimit/redis.go
```diff
@@ -210,6 +218,12 @@ func setError(span opentracing.Span) {
 	}
 }
 
+func logError(span opentracing.Span, err error) {
+	if span != nil {
+		ext.LogError(span, err)
+	}
+}
```
AlexanderYastrebov, line 225: Do we want sampled span logging or sampled logging (`log.Errorf("whatever: v", err)`)?

ratelimit/redis.go
```diff
@@ -121,6 +124,11 @@ func (c *clusterLimitRedis) Allow(ctx context.Context, clearText string) bool {
 	if failed {
 		allow = !c.failClosed
 		setError(span)
+		if err != nil {
```
AlexanderYastrebov, line 127: err is not nil here since this is how `failed` is defined

ratelimit/redis.go
AlexanderYastrebov: @szuecs

nabeelbukhari: I prefer sampled span logging because we have existing tracing support for redis.

AlexanderYastrebov:
@nabeelbukhari Please `git commit -s --amend` to sign-off the commit and make DCO check happy.

szuecs:
@nabeelbukhari why do we need this change?

ratelimit/redis.go
szuecs: Why do you think sampled span logs bring value?
If these are sampled you likely won't find any of the sampled logs.

nabeelbukhari: @suzes I want to have visibility on what went wrong if redis call fails. Currently we only have light step traces to find if call failed.

szuecs: But then sampling makes no sense.

AlexanderYastrebov: I think the problem with sampling span logs is that it would be hard to find it - that is why OI proposed to use regular logs - those we'll see all at once.

ratelimit/redis.go
```diff
@@ -52,6 +54,7 @@ func newClusterRateLimiterRedis(s Settings, r *net.RedisRingClient, group string
 		window:     s.TimeWindow,
 		ringClient: r,
 		metrics:    metrics.Default,
+		sometimes:  rate.Sometimes{First: 3, Interval: 50 * time.Millisecond},
```
AlexanderYastrebov, line 57: What is the reasoning behind 50ms? Why not 1s?

ratelimit/redis.go
```diff
@@ -210,6 +216,12 @@ func setError(span opentracing.Span) {
 	}
 }
 
+func logError(span opentracing.Span, err error) {
+	if span != nil {
```
AlexanderYastrebov, line 220: We do not need nil check here - we always have a span

ratelimit/redis.go
```diff
@@ -226,6 +238,9 @@ func (c *clusterLimitRedis) oldest(ctx context.Context, clearText string) (time.
 
 	if err != nil {
 		setError(span)
+		c.sometimes.Do(func() {
```
AlexanderYastrebov, line 241: Lets move c.sometimes.Do inside of `logError`

ratelimit/redis.go
AlexanderYastrebov: We've clarified with @nabeelbukhari that it is possible to filter error spans that also have logs.

ratelimit/redis.go
nabeelbukhari: This interval defines how frequently we add span with errors. I though to use 50 ms instead of second because the error may be occuring for a short time and we can miss it if we set to once a sec. I am flexible with changing it to few hundred ms as well.

ratelimit/redis.go
AlexanderYastrebov: But then we need to log errors instead.

szuecs: I think it's more valuable to have errors in logs. If we see it's too many we can wrap it to stop too much logs. I think we should keep things simple and right now we create complexity for one narrow use.

ratelimit/redis.go
```diff
@@ -210,6 +219,12 @@ func setError(span opentracing.Span) {
 	}
 }
 
+func logError(c *clusterLimitRedis, err error) {
```
AlexanderYastrebov, line 222: Plaese make it  `func (c *clusterLimitRedis) logError(err error)`

ratelimit/redis.go
nabeelbukhari: @szuecs @AlexanderYastrebov I changed logic and now we do logs instead of tagging to span. Please took another look.

ratelimit/redis.go
AlexanderYastrebov: I do not think we should care too much for short error spikes - these would be visisble as error spans. For prolonged (several seconds) failures we would see error logs and we do not want to flood logs. I would say 1s is a good round number to set as a rate limit.

ratelimit/redis.go
```diff
@@ -147,12 +151,14 @@ func (c *clusterLimitRedis) allow(ctx context.Context, clearText string) (bool,
 	// drop all elements of the set which occurred before one interval ago.
 	_, err := c.ringClient.ZRemRangeByScore(ctx, key, 0.0, float64(clearBefore))
 	if err != nil {
+		c.logError(err)
```
AlexanderYastrebov, line 154: I think logError in this method are redundadnt - we log error in the Allow that calls it.

ratelimit/redis.go
```diff
@@ -242,6 +258,7 @@ func (c *clusterLimitRedis) oldest(ctx context.Context, clearText string) (time.
 	oldest, err := strconv.ParseInt(s, 10, 64)
 	if err != nil {
 		setError(span)
+		c.logError(err)
```
AlexanderYastrebov, line 261: This is redundant as we already log error in Oldest(

ratelimit/redis.go
```diff
@@ -121,6 +124,7 @@ func (c *clusterLimitRedis) Allow(ctx context.Context, clearText string) bool {
 	if failed {
 		allow = !c.failClosed
 		setError(span)
+		c.logError("Failed to redis to determine if operation is allowed", err)
```
AlexanderYastrebov, line 127: Please rephrase into simple sentence and add %v format specifier: `Failed to determine if operation is allowed: %v` (also elsewhere)

ratelimit/redis.go
```diff
@@ -194,7 +198,7 @@ func (c *clusterLimitRedis) Delta(clearText string) time.Duration {
 	now := time.Now()
 	d, err := c.deltaFrom(context.Background(), clearText, now)
 	if err != nil {
-		log.Errorf("Failed to redis get the duration until the next call is allowed: %v", err)
+		c.logError("Failed to redis get the duration until the next call is allowed", err)
```
AlexanderYastrebov, line 201: `Failed to get the duration until the next call is allowed: %v`

ratelimit/redis.go
```diff
@@ -257,7 +267,7 @@ func (c *clusterLimitRedis) oldest(ctx context.Context, clearText string) (time.
 func (c *clusterLimitRedis) Oldest(clearText string) time.Time {
 	t, err := c.oldest(context.Background(), clearText)
 	if err != nil {
-		log.Errorf("Failed to get from redis the oldest known request time: %v", err)
+		c.logError("Failed to get from redis the oldest known request time:", err)
```
AlexanderYastrebov, line 270: `Failed to get the oldest known request time: %v`

ratelimit/redis.go
```diff
@@ -285,7 +295,7 @@ func (c *clusterLimitRedis) RetryAfterContext(ctx context.Context, clearText str
 
 	retr, err := c.deltaFrom(ctx, clearText, now)
 	if err != nil {
-		log.Errorf("Failed to get from redis the duration to wait with the next request: %v", err)
+		c.logError("Failed to get from redis the duration to wait with the next request:", err)
```
AlexanderYastrebov, line 298: `Failed to get the duration to wait until the next request: %v`

ratelimit/redis.go
```diff
@@ -210,6 +214,12 @@ func setError(span opentracing.Span) {
 	}
 }
 
+func (c *clusterLimitRedis) logError(description string, err error) {
```
AlexanderYastrebov, line 217: description -> format

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
:+1:

nabeelbukhari:
👍

PR #2590, opened 2023-09-12
dataclients/kubernetes: add token file flag

AlexanderYastrebov:
Add flag to supply token file when running outside of cluster.
This is useful to test skipper/routesrv outside of the cluster without using kubectl proxy.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -151,6 +151,7 @@ type Config struct {
 	KubernetesIngress                       bool                               `yaml:"kubernetes"`
 	KubernetesInCluster                     bool                               `yaml:"kubernetes-in-cluster"`
 	KubernetesURL                           string                             `yaml:"kubernetes-url"`
+	KubernetesTokenFile                     string                             `yaml:"kubernetes-token-file"`
 	KubernetesHealthcheck                   bool                               `yaml:"kubernetes-healthcheck"`
 	KubernetesHTTPSRedirect                 bool                               `yaml:"kubernetes-https-redirect"`
 	KubernetesHTTPSRedirectCode             int                                `yaml:"kubernetes-https-redirect-code"`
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -437,7 +438,8 @@ func NewConfig() *Config {
 	// Kubernetes:
 	flag.BoolVar(&cfg.KubernetesIngress, "kubernetes", false, "enables skipper to generate routes for ingress resources in kubernetes cluster. Enables -normalize-host")
 	flag.BoolVar(&cfg.KubernetesInCluster, "kubernetes-in-cluster", false, "specify if skipper is running inside kubernetes cluster")
-	flag.StringVar(&cfg.KubernetesURL, "kubernetes-url", "", "kubernetes API base URL for the ingress data client; requires kubectl proxy running; omit if kubernetes-in-cluster is set to true")
+	flag.StringVar(&cfg.KubernetesURL, "kubernetes-url", "", "kubernetes API base URL for the ingress data client, omit if kubernetes-in-cluster is set to true")
```
RomanZavodskikh, line 441: Can we rely on user of skipper to "omit if ... is set to true"? It seems to me it's better to enforce it somehow.

dataclients/kubernetes/clusterclient.go
```diff
@@ -168,17 +169,22 @@ func newClusterClient(o Options, apiURL, ingCls, rgCls string, quit <-chan struc
...
-			return nil, err
+		c.tokenFile = serviceAccountDir + serviceAccountTokenKey
+	} else if o.TokenFile != "" {
+		c.tokenProvider = secrets.NewSecretPaths(time.Minute)
+		c.tokenFile = o.TokenFile
+	}
+
+	if c.tokenProvider != nil {
+		if err := c.tokenProvider.Add(c.tokenFile); err != nil {
+			return nil, fmt.Errorf("failed to add secret %s: %v", c.tokenFile, err)
```
RomanZavodskikh, line 180: Is it okay to change letter case here? Users may grep by "Failed to Add" and lost this log message.

AlexanderYastrebov: I just make is all consistent. Return error instead of logging and make it a proper sentence - skipper will not start if error happens. Also we do not guarantee log format, at least here (maybe we should be more careful about proxy logs).

config/config.go
AlexanderYastrebov: Updated usages

dataclients/kubernetes/clusterclient.go
szuecs: Let's have an issue and discuss pros and cons and what would be the exact rule we want to apply.
Normally errors should be found somewhere in the log, but I guess here we have some info/debug log and an error with different case-sensitive characters. I am not sure why these have to be in-sync, but maybe there's some great idea here.

AlexanderYastrebov:
:+1:

PR #2591, opened 2023-09-13
routesrv: log route changes for debugging

AlexanderYastrebov:
Log route changes for debugging

### DIFF
--- a/routesrv/polling.go
+++ b/routesrv/polling.go
@@ -68,6 +68,7 @@ func (p *poller) poll(wg *sync.WaitGroup) {
 	defer ticker.Stop()
 	pollingStarted.SetToCurrentTime()
 
+	var lastRoutesById map[string]string
 	for {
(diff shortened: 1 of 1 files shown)


routesrv/polling.go
```diff
@@ -109,6 +110,12 @@ func (p *poller) poll(wg *sync.WaitGroup) {
 			}
 			span.SetTag("routes.count", routesCount)
 			span.SetTag("routes.bytes", routesBytes)
+
+			if updated && log.GetLevel() == log.DebugLevel {
```
RomanZavodskikh, line 114: Maybe `>= log.DebugLevel` here? There is at least one more verbose leve: `TraceLevel`

https://pkg.go.dev/github.com/sirupsen/logrus#pkg-variables

AlexanderYastrebov: Indeed, fixed.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2592, opened 2023-09-13
Clean up removed endpoints in endpointregistry

RomanZavodskikh:
Clean up removed endpoints in endpointregistry

### DIFF
--- a/routing/endpointregistry.go
+++ b/routing/endpointregistry.go
@@ -3,7 +3,11 @@ package routing
 import (
 	"sync"
 	"time"
+
+	"github.com/zalando/skipper/eskip"
 )
+
+const lastSeenTimeout = 1 * time.Minute
 
 // Metrics describe the data about endpoint that could be
 // used to perform better load balancing, fadeIn, etc.
@@ -28,6 +32,8 @@ func (e *entry) InflightRequests() int64 {
 }
 
 type EndpointRegistry struct {
+	lastSeen map[string]time.Time
+
 	mu sync.Mutex
 
 	data map[string]*entry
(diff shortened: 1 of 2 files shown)


routing/endpointregistry.go
```diff
@@ -38,20 +48,81 @@ var _ PostProcessor = &EndpointRegistry{}
 type RegistryOptions struct {
 }
 
+func normalizeSchemeHost(s, h string) (string, string, error) {
```
AlexanderYastrebov, line 51: We should stop copying this - this would be third instance I think

routing/endpointregistry.go
```diff
@@ -38,20 +48,81 @@ var _ PostProcessor = &EndpointRegistry{}
...
+			r.mu.Unlock()
+		}
+	}
+
 	return routes
 }
 
 func NewEndpointRegistry(o RegistryOptions) *EndpointRegistry {
-	return &EndpointRegistry{data: map[string]*entry{}}
+	return &EndpointRegistry{data: map[string]*entry{}, lastSeen: map[string]time.Time{}}
```
AlexanderYastrebov, line 125: Please reformat into multiple lines

routing/endpointregistry.go
```diff
@@ -1,10 +1,18 @@
...
+	"net/url"
+	"strings"
 	"sync"
 	"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/eskip"
 )
 
+const entryTTL = 1 * time.Minute
```
AlexanderYastrebov, line 14: `lastSeenTimeout`?

routing/endpointregistry.go
RomanZavodskikh: It's used for NetworkBackend only, so it's more appropriate to remove it for now and think how to use it in the next PR.

routing/endpointregistry_test.go
```diff
@@ -48,6 +49,29 @@ func TestSetAndGetAnotherKey(t *testing.T) {
...
+	route := &Route{LBEndpoints: []LBEndpoint{
+		{Host: "existing endpoint"},
+	}}
+	route.BackendType = eskip.LBBackend
+
+	r := NewEndpointRegistry(RegistryOptions{})
+
+	r.IncInflightRequest("existing endpoint")
+	r.IncInflightRequest("removed endpoint")
+	r.lastSeen["removed endpoint"] = time.Now().Add(-lastSeenTimeout)
```
AlexanderYastrebov, line 62: This tests internals and would break after refactoring.
Ideally we should test behavior, e.g.:
1. r.Do(<route with endpoint1 and endpoint2>)
2. e.g. check detected time of endpoint1 and endpoint2
3. wait (this could be done by moving "now" function and moving clock forward)
4. r.Do(<route with endpoint1 only>)
5. check that endpoint2 was removed e.g. by checking its detected time

PR #2593, opened 2023-09-13
dataclients/kubernetes: add failing testcase for named service target port

AlexanderYastrebov:
For https://github.com/zalando/skipper/pull/2565


dataclients/kubernetes/testdata/ingressV1/service-ports-endpointslices/foo.yaml
```diff
@@ -0,0 +1,87 @@
...
+  clusterIPs:
+    - 10.5.69.228
+  internalTrafficPolicy: Cluster
+  ipFamilies:
+    - IPv4
+  ipFamilyPolicy: SingleStack
+  ports:
+    - port: 80
+      protocol: TCP
+      targetPort: http-grafana
```
AlexanderYastrebov, line 38: Here service uses named target port which could only be resolved by reading pod definition.

Endpoints logic has a special condition for a single port:
https://github.com/zalando/skipper/blob/52ef4ef61928da148983bf9aca40cc893d99baa6/dataclients/kubernetes/ingressdefinitions.go#L91-L99

AlexanderYastrebov:
Pushed to the https://github.com/zalando/skipper/pull/2565

Issue #2595, opened 2023-09-14
route health state as Prometheus metrics


universam1:
**Is your feature request related to a problem? Please describe.**
As discussed in [slack](https://gophers.slack.com/archives/C82Q5JNH5/p1691523249802909?thread_ts=1691075117.833439&cid=C82Q5JNH5) having the counters of `X-Count` and `X-Timestamp` exposed as Prometheus metrics would help to alert on a Skipper replica being out of sync.

**Describe the solution you would like**
Prometheus metric exposing those counters, for customer using Prometheus ecosystem for alerting.

**Describe alternatives you've considered (optional)**
Having to invent a new monitoring concept parsing the route health endpoints `http://{}:9911/routes`

**Would you like to work on it?**
Yes - with advice

universam1:
FYI @szuecs

AlexanderYastrebov:
Skipper /routes handler uses "routing" object. I guess we'd need to wire metrics in and update counters around here
https://github.com/zalando/skipper/blob/4b2d6eb2161f726f1b6648de45b5e354c4896056/routing/routing.go#L364-L373

universam1:
closed via #2597

PR #2597, opened 2023-09-14
adding Prometheus metric for Routes health

universam1:
ref: https://github.com/zalando/skipper/issues/2595

adding Prometheus metrics for Routes health of `routes-total` and `routes-last-update` gauges that will allow to defines alerts if Skipper instances drift.

### DIFF
--- a/routing/routing.go
+++ b/routing/routing.go
@@ -12,6 +12,7 @@ import (
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/filters"
 	"github.com/zalando/skipper/logging"
+	"github.com/zalando/skipper/metrics"
 	"github.com/zalando/skipper/predicates"
 )
 
@@ -122,6 +123,10 @@ type Options struct {
 
 	// SuppressLogs indicates whether to log only a summary of the route changes.
 	SuppressLogs bool
+
+	// Metrics is used to collect monitoring data about the routes health, including
+	// total number of routes applied and UNIX time of the last routes update.
+	Metrics metrics.Metrics
 
 	// PreProcessors contains custom eskip.Route pre-processors.
 	PreProcessors []PreProcessor
@@ -268,6 +273,7 @@ type Routing struct {
 	firstLoad         chan struct{}
 	firstLoadSignaled bool
 	quit              chan struct{}
+	metrics           metrics.Metrics
 }
 
 // New initializes a routing instance, and starts listening for route
@@ -278,6 +284,7 @@ func New(o Options) *Routing {
 	}
 
 	r := &Routing{log: o.Log, firstLoad: make(chan struct{}), quit: make(chan struct{})}
+	r.metrics = o.Metrics
 	if !o.SignalFirstLoad {
 		close(r.firstLoad)
 		r.firstLoadSignaled = true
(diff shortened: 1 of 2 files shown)


webframp:
We definitely don't expect this to be merged as is but wanted to get a PR up to make sure we're on the right track with how to wire the counters in.

metrics/codahale.go
```diff
@@ -111,6 +111,10 @@ func (c *CodaHale) UpdateGauge(key string, v float64) {
 	c.getGauge(key).Update(v)
 }
 
+func (c *CodaHale) UpdateRoutesGauge(v int) {
+	// not implemented
+}
```
AlexanderYastrebov, line 116: There is `UpdateGauge` which supports custom key. Check how it is used elsewhere, I think we'd prefer it unless there is a solid reason not to.

proxy/proxy.go
```diff
@@ -705,8 +705,8 @@ func WithParams(p Params) *Proxy {
 	if p.EndpointRegistry == nil {
 		p.EndpointRegistry = routing.NewEndpointRegistry(routing.RegistryOptions{})
 	}
-
 	hostname := os.Getenv("HOSTNAME")
+	p.Routing.Metrics = m
```
AlexanderYastrebov, line 709: This does not look right - it modifies proxy parameters.

It should be somewhere in skipper.go like
```go
routing := routing.New(ro)
defer routing.Close()

routing.Metrics = mtr
```

metrics/codahale.go
universam1: great point, this makes the implementation even simplier!

proxy/proxy.go
universam1: great point, this makes the implementation even simplier!

universam1:
The metrics, combined with a diff threshold seem quite useful:
image

routing/routing.go
```diff
@@ -268,6 +269,7 @@ type Routing struct {
 	firstLoad         chan struct{}
 	firstLoadSignaled bool
 	quit              chan struct{}
+	Metrics           metrics.Metrics
```
AlexanderYastrebov, line 272: Please make field unexported like others

routing/routing.go
```diff
@@ -371,6 +373,8 @@ func (r *Routing) startReceivingUpdates(o Options) {
 					}
 				}
 				r.log.Info("route settings applied")
+				r.Metrics.UpdateGauge("routes-total", float64(len(rt.validRoutes)))
+				r.Metrics.UpdateGauge("routes-last-update", float64(rt.created.Unix()))
```
AlexanderYastrebov, line 377: Looking at other places I think we have a kind of convention for metric key names: one or multiple parts joined with a dot.

Lets name these  `routes.total` and `routes.updated_timestamp`, somewhat similar to:
https://github.com/zalando/skipper/blob/96567ac366a3fb104cdec59409fa810de7e7d591/routesrv/polling.go#L40-L44

routing/routing.go
universam1: Which way else, via param?

AlexanderYastrebov: Ah, yeah, we would need to add it to the routing.Options

routing/routing.go
universam1: makes sense, applied

routing/routing.go
universam1: applied!

routing/routing.go
```diff
@@ -371,6 +378,8 @@ func (r *Routing) startReceivingUpdates(o Options) {
 					}
 				}
 				r.log.Info("route settings applied")
+				r.metrics.UpdateGauge("routes.total", float64(len(rt.validRoutes)))
+				r.metrics.UpdateGauge("routes.updated_timestamp", float64(rt.created.Unix()))
```
AlexanderYastrebov, line 382: I think we may need to wrap this into nil check - existing codebases do not supply metrics instance

universam1: done

AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

universam1:
Thank you @AlexanderYastrebov @szuecs @RomanZavodskikh for quick release!

Issue #2316, opened 2023-05-10
Allow tracingTag filter to use response header values


lukas-c-wilhelm:
**Is your feature request related to a problem? Please describe.**
An application might provide a successful response that is nevertheless degraded in quality (e.g. using a static fallback value for some property). A measurement of error rate of an application might want to consider such responses as errors even if they are technically successful. If this monitoring should be based on spans, the span has to receive an error tag based on a response property.

**Describe the solution you would like**
The `tracingTag` filter should be able to use response header values.

**Describe alternatives you've considered (optional)**
n/a

**Additional context (optional)**
n/a

**Would you like to work on it?**
Yes

AlexanderYastrebov:
`tracingTag` currently is only applied n the request path. We may add a companion filter `responseTracingTag` to do the same on the response path.

>  the span has to receive an error tag based on a response property.

If we want to set tag based on the response code then we'd need to add another filter that implements this.

PR #2607, opened 2023-09-20
allow tagFilter to use response data

lukas-c-wilhelm:
Fixes https://github.com/zalando/skipper/issues/2316

This change introduces a new filter `tracingTagOnResponse` that works just like the tracingTag filter, but is applied only after ther request has been processed. This allows using properties of the response (a header value) in the tag.

The use case for this is that users might want to consider an operation as failed even if it technically succeeds, e.g. because fallbacks were used. With this change, a response header can be sent, e.g. `"X-Fallback": "true"`, and then the filter `tracingTagOnResponse("error", "${response.header.X-Fallback}")` would result in the span being marked as error.

Another use cases could be that metrics should be captured by use case, but the use case is not apparent from the request and instead only defined by the result of the request processing.


AlexanderYastrebov:
Looks like it attempts to fix https://github.com/zalando/skipper/issues/2316

filters/tracing/tag.go
```diff
@@ -55,5 +63,3 @@ func (f tagFilter) Request(ctx filters.FilterContext) {
 		span.SetTag(f.tagName, v)
```
AlexanderYastrebov, line 63: With this chnage this will be executed twice on request and response path.
E.g. `tracingTag("foo", "${request.header.X-Foo}") -> setRequestHeader("X-Foo", "bar")` will set tag to `bar` I think.

lukas-c-wilhelm: Thanks for the hint!. I noticed it when reviewing. I'm currently changing it to just add a filter.

filters/tracing/tagonresponse.go
```diff
@@ -0,0 +1,59 @@
...
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	return tagOnResponseFilter{
+		tagName:  tagName,
+		tagValue: eskip.NewTemplate(tagValue),
+	}, nil
+}
+
+func (f tagOnResponseFilter) Request(filters.FilterContext) {}
```
szuecs, line 47: let's use ptr receiver

szuecs:
You also need to register the spec into the builtins https://github.com/zalando/skipper/blob/master/filters/builtin/builtin.go#L210 to make it available for use in skipper.

filters/tracing/tagonresponse.go
lukas-c-wilhelm: Done. This is my first piece of go code, so I hope I did it correctly. I checked other filters and copied from there.

lukas-c-wilhelm:
@szuecs Thanks for your review. I addressed your comments.

filters/tracing/tagonresponse.go
```diff
@@ -0,0 +1,59 @@
...
+	return &tagOnResponseFilter{
+		tagName:  tagName,
+		tagValue: eskip.NewTemplate(tagValue),
+	}, nil
+}
+
+func (f *tagOnResponseFilter) Request(filters.FilterContext) {}
+
+func (f *tagOnResponseFilter) Response(ctx filters.FilterContext) {
+	req := ctx.Request()
```
szuecs, line 50: You can also drop the intermediate variable.
The return value is never nil, so no check required anyways.

szuecs:
Only one tiny nit and it's ready to merge, thanks

filters/tracing/tagonresponse.go
lukas-c-wilhelm: I'll apply the same changes to `tag.go` as well.

AlexanderYastrebov:
Looks ok, couple of ideas.

Since the filter matches 99% with existing you could also add a flag like here https://github.com/zalando/skipper/blob/master/filters/diag/histogram.go

```
$ grep -o -E '".*esponse.*"' filters/filters.go
"modResponseHeader"
"setResponseHeader"
"appendResponseHeader"
"dropResponseHeader"
"setContextResponseHeader"
"appendContextResponseHeader"
"copyResponseHeader"
"uniformResponseLatency"
"normalResponseLatency"
"histogramResponseLatency"
"dropResponseCookie"
"responseCookie"
"opaServeResponse"

$ grep -o -i -E '".*from.*"' filters/filters.go | sort
"setDynamicBackendHostFromHeader"
"setDynamicBackendSchemeFromHeader"
"setDynamicBackendUrlFromHeader"
```

Should we name it `tracingTagFromResponse` maybe?

szuecs:
@AlexanderYastrebov yeah makes sense to have "typ" like in https://github.com/zalando/skipper/blob/master/filters/diag/histogram.go#L39 and check in Request/Response https://github.com/zalando/skipper/blob/master/filters/diag/histogram.go#L104.

lukas-c-wilhelm:
@AlexanderYastrebov @szuecs I'll try to implement it like that.

szuecs:
:+1:

filters/tracing/tag.go
```diff
@@ -38,22 +46,36 @@ func (s tagSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
...
+	if f.tagFromResponse {
+		span := opentracing.SpanFromContext(ctx.Request().Context())
+		if span == nil {
+			return
+		}
+
+		if v, ok := f.tagValue.ApplyContext(ctx); ok {
+			span.SetTag(f.tagName, v)
+		}
+	}
```
AlexanderYastrebov, line 80: Could you move this to a helper `func (f *tagFilter) setTag(ctx filters.FilterContext)` please?

filters/tracing/tag_test.go
```diff
@@ -31,7 +31,14 @@ func TestTracingTagNil(t *testing.T) {
...
+	if (&tagSpec{
+		typ: filters.TracingTagName,
+	}).Name() != filters.TracingTagName {
+		t.Error("Wrong tag spec name")
+	}
+	if (&tagSpec{
+		typ: filters.TracingTagFromResponseName,
+	}).Name() != filters.TracingTagFromResponseName {
 		t.Error("Wrong tag spec name")
 	}
```
AlexanderYastrebov, line 43: This should test
```go
if NewTag().Name() != filters.TracingTagName {
  t.Error("Wrong tag spec name")
}

if NewTagFromResponse().Name() != filters.TracingTagFromResponseName {
  t.Error("Wrong tag spec name")
}
```

filters/tracing/tag_test.go
```diff
@@ -55,18 +62,21 @@ func TestTracingTag(t *testing.T) {
 
 	for _, ti := range []struct {
 		name     string
+		filter   filters.Spec
```
AlexanderYastrebov, line 65: Please name it `spec`, not `filter`

filters/tracing/tag_test.go
```diff
@@ -91,19 +115,47 @@ func TestTracingTag(t *testing.T) {
...
 			if got := span.Tag("test_tag"); got != ti.expected {
 				t.Errorf("unexpected tag value '%v' != '%v'", got, ti.expected)
 			}
-
-			f.Response(ti.context)
 		})
 	}
 }
+
+func TestTagFilterIgnoresResponse(t *testing.T) {
```
AlexanderYastrebov, line 133: I think this could be a test case in the TestTracingTag
```go
	{
		"tag ignores response",
		NewTag(),
		"${response.header.X-Fallback}",
		&filtertest.Context{
			FRequest: &http.Request{
				Header: http.Header{
					"X-Fallback": []string{"true"},
				},
			},
		},
		"",
	}
```
and likely we need the symmetric that NewTagFromResponse ignores request.

filters/builtin/builtin.go
```diff
@@ -208,6 +208,7 @@ func Filters() []filters.Spec {
 		tracing.NewBaggageToTagFilter(),
 		tracing.NewTag(),
 		tracing.NewStateBagToTag(),
+		tracing.NewTagFromResponse(),
```
AlexanderYastrebov, line 211: Please move this one line up to have it next to tracing.NewTag()

filters/tracing/tag_test.go
lukas-c-wilhelm: Not sure about this. Wouldn't the header in the request only match `"${request.header.X-Fallback}"`, not the provided template for the **response**? So in any case the result would be `nil`, right?

My reasoning was that we need to have a case where a header is only present in the response, not in the request. 

Will add a test for the symmetrical case, though.

AlexanderYastrebov: Sorry, I meant to use FResponse of course, i.e. filter is NewTag but template is `"${response.header.X-Fallback}"` and `X-Fallback` is a response header

AlexanderYastrebov: We can have `request` and `response` fields in the test cases instead of context, create context at the start of the test and set response after f.Request(ctx) but before f.Response(ctx):

```go
ctx := &filtertest.Context{
  FRequest: ti.request,
}
...

f.Request(ctx)

ctx.FResponse = ti.response

f.Response(ctx)
```

lukas-c-wilhelm: This works for a test that show `tracingTag` only runs on the request. It does not work for the symmetric case to test that `tracingTagFromResponse` only runs on the response. To prove this, the request in the context has to be changed (header removed) to prove that `tracingTagFromResponse` cannot make use of headers that are only present at request time. It is a bit artificial anyway.

Would you prefer to just have one test according to the method you described, or separate tests as they are now? Or a different change?

lukas-c-wilhelm: Alternatively, I could change the test code to reset the context to blank after the request call. But this would also be a bit strange, because in a real scenario the request context would be available also at response time, if I understand it correctly.

AlexanderYastrebov: Lets have a single additional test case then.

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
@lukas-c-wilhelm Thank you for your contribution.

filters/tracing/tag_test.go
szuecs: In the additional test you can also have a real backend httptest.Server to create the response header to test it.

szuecs:
:+1:

PR #2612, opened 2023-09-22
dataclients/kubernetes: RouteGroup catchall interferes with Ingress

AlexanderYastrebov:
Dataclient creates a catchall route for RouteGroup that interferes with an Ingress route and leads to sporadic 404 responses.

Dataclient should probably create catchall routes after processing all Ingresses and RouteGroups.

Alternatively we may consider disabling catchall routes completely since default response code when there is no route is 404. This should also reduce routing table size.

### DIFF
--- a/dataclients/kubernetes/mixed_test.go
+++ b/dataclients/kubernetes/mixed_test.go
@@ -1,0 +1,13 @@
+package kubernetes_test
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


AlexanderYastrebov:
> Alternatively we may consider disabling catchall routes completely

See https://github.com/zalando/skipper/pull/2656

PR #2613, opened 2023-09-22
Check the absence of inflight requests before removing endpoint from endpointregistry

RomanZavodskikh:
This check is needed for endpoints with latency more than lastSeenTimeout. They should still be kept in endpointregistry even after this timeout until all pending DecInflightOperations are done.


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2620, opened 2023-09-25
secrets: handle empty secrets

AlexanderYastrebov:
Logs error on add and treat empty file as removed on update.

In particular it errors instead of using empty value when secret path comes from process substitution
```
$ bin/routesrv -kubernetes-token-file=<(echo foobar)
ERRO[0000] Failed to read file /dev/fd/63: empty file
FATA[0000] failed to get secret /dev/fd/63
```

which is a symlink to a named pipe
```
$ ls -l <(echo foobar)
lr-x------ 1 jdoe jdoe 64 Sep 25 13:36 /dev/fd/63 -> 'pipe:[7925874]'
```

### DIFF
--- a/secrets/file.go
+++ b/secrets/file.go
@@ -20,6 +20,7 @@ const (
 var (
 	ErrWrongFileType    = errors.New("file type not supported")
 	ErrFailedToReadFile = errors.New("failed to read file")
+	errEmptyFile        = errors.New("empty file")
 )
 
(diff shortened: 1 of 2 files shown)


RomanZavodskikh:
:+1:

AlexanderYastrebov:
pipeline flaked due to #2621

szuecs:
:+1:

PR #2628, opened 2023-09-26
Check endpointregistry Detected field in CI tests

RomanZavodskikh:
Check endpointregistry Detected field in CI tests

### DIFF
--- a/filters/fadein/fadein_test.go
+++ b/filters/fadein/fadein_test.go
@@ -183,7 +183,7 @@ func TestCreateEndpointCreated(t *testing.T) {
 }
 
 func TestPostProcessor(t *testing.T) {
-	createRouting := func(t *testing.T, routes string) (*routing.Routing, func(string)) {
+	createRouting := func(t *testing.T, routes string, endpointRegistry *routing.EndpointRegistry) (*routing.Routing, func(string)) {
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

filters/fadein/fadein_test.go
```diff
@@ -183,12 +183,16 @@ func TestCreateEndpointCreated(t *testing.T) {
...
+	createRouting := func(t *testing.T, routes string, endpointRegistry *routing.EndpointRegistry) (*routing.Routing, func(string)) {
 		dc, err := testdataclient.NewDoc(routes)
 		if err != nil {
 			t.Fatal(err)
 		}
 
+		if endpointRegistry == nil {
+			endpointRegistry = routing.NewEndpointRegistry(routing.RegistryOptions{})
+		}
+
```
AlexanderYastrebov, line 195: Do we need this? Is there a call to createRouting that does not pass registry? Its better to pass it always or check that nil also works

RomanZavodskikh: nil doesn't work (and it must not) because it is `routing.PostProcessor` and `Do` method is called periodically. So nil will panic.
```
Its better to pass it always
```
This option is the best I think.

PR #2631, opened 2023-09-26
feature: routesrv add prometheus metrics

szuecs:
feature: routesrv add prometheus metrics
feature: routesrv enable profiling
test: add more routesrv coverage to reach close to 80%

Example metrics:
```
# HELP routesrv_backend_combined_duration_seconds Duration in seconds of a proxy backend combined.
# TYPE routesrv_backend_combined_duration_seconds histogram
routesrv_backend_combined_duration_seconds_bucket{le="0.005"} 10
routesrv_backend_combined_duration_seconds_bucket{le="0.01"} 10
routesrv_backend_combined_duration_seconds_bucket{le="0.025"} 10
routesrv_backend_combined_duration_seconds_bucket{le="0.05"} 10
routesrv_backend_combined_duration_seconds_bucket{le="0.1"} 10
routesrv_backend_combined_duration_seconds_bucket{le="0.25"} 10
routesrv_backend_combined_duration_seconds_bucket{le="0.5"} 10
routesrv_backend_combined_duration_seconds_bucket{le="1"} 10
routesrv_backend_combined_duration_seconds_bucket{le="2.5"} 10
routesrv_backend_combined_duration_seconds_bucket{le="5"} 10
routesrv_backend_combined_duration_seconds_bucket{le="10"} 10
routesrv_backend_combined_duration_seconds_bucket{le="+Inf"} 10
routesrv_backend_combined_duration_seconds_sum 0.0038043880000000006
routesrv_backend_combined_duration_seconds_count 10
# HELP routesrv_backend_duration_seconds Duration in seconds of a proxy backend.
# TYPE routesrv_backend_duration_seconds histogram
routesrv_backend_duration_seconds_bucket{host="",route="routersv",le="0.005"} 10
routesrv_backend_duration_seconds_bucket{host="",route="routersv",le="0.01"} 10
routesrv_backend_duration_seconds_bucket{host="",route="routersv",le="0.025"} 10
routesrv_backend_duration_seconds_bucket{host="",route="routersv",le="0.05"} 10
routesrv_backend_duration_seconds_bucket{host="",route="routersv",le="0.1"} 10
routesrv_backend_duration_seconds_bucket{host="",route="routersv",le="0.25"} 10
routesrv_backend_duration_seconds_bucket{host="",route="routersv",le="0.5"} 10
routesrv_backend_duration_seconds_bucket{host="",route="routersv",le="1"} 10
routesrv_backend_duration_seconds_bucket{host="",route="routersv",le="2.5"} 10
routesrv_backend_duration_seconds_bucket{host="",route="routersv",le="5"} 10
routesrv_backend_duration_seconds_bucket{host="",route="routersv",le="10"} 10
routesrv_backend_duration_seconds_bucket{host="",route="routersv",le="+Inf"} 10
routesrv_backend_duration_seconds_sum{host="",route="routersv"} 0.0038043880000000006
routesrv_backend_duration_seconds_count{host="",route="routersv"} 10
# HELP routesrv_custom_gauges Gauges number of custom metrics.
# TYPE routesrv_custom_gauges gauge
routesrv_custom_gauges{key="polling_started_timestamp"} 1.6983456727187953e+09
routesrv_custom_gauges{key="routes.byte"} 93714
routesrv_custom_gauges{key="routes.initialized_timestamp"} 1.698345673637034e+09
routesrv_custom_gauges{key="routes.total"} 266
routesrv_custom_gauges{key="routes.updated_timestamp"} 1.698345673637066e+09
# HELP routesrv_custom_total Total number of custom metrics.
# TYPE routesrv_custom_total counter
routesrv_custom_total{key="200"} 10
```


routesrv/polling.go
```diff
@@ -56,8 +37,9 @@ type poller struct {
 	editRoute      []*eskip.Editor
 	cloneRoute     []*eskip.Clone
 
-	// tracer
-	tracer ot.Tracer
+	// visibility
+	tracer  ot.Tracer
+	metrics metrics.Metrics
```
AlexanderYastrebov, line 42: Why do you decide to use Metrics instead of barebone prometheus?
For stock prometheus metrics we can use labels.

routesrv/polling.go
```diff
@@ -101,12 +86,15 @@ func (p *poller) poll(wg *sync.WaitGroup) {
 			if initialized {
 				logger.Info(LogRoutesInitialized)
 				span.SetTag("routes.initialized", true)
-				routesInitialized.SetToCurrentTime()
+				p.metrics.UpdateGauge("routes_initialized_timestamp", (float64(time.Now().UnixNano()) / 1e9))
```
AlexanderYastrebov, line 89: I guess we could have a helper like
```go
func (p *poller) setGaugeToCurrentTime(name string) {
    p.metrics.UpdateGauge(name, (float64(time.Now().UnixNano()) / 1e9))
}
```
to make it a bit cleaner.

routesrv/routesrv.go
```diff
@@ -42,12 +64,24 @@ func New(opts skipper.Options) (*RouteServer, error) {
...
+	mux := http.NewServeMux()
+	mux.Handle("/health", bs)
+	mux.Handle("/routes", b)
+	mux.Handle("/metrics", metricsHandler)
+	mux.Handle("/metrics/", metricsHandler)
+
+	if opts.EnableProfile {
+		mux.Handle("/debug/pprof", metricsHandler)
+		mux.Handle("/debug/pprof/", metricsHandler)
+	}
```
AlexanderYastrebov, line 84: We should serve these from another port (support listener) since we allow user access to /routes

routesrv/polling.go
```diff
@@ -90,6 +72,9 @@ func (p *poller) poll(wg *sync.WaitGroup) {
 		case routesCount == 0:
 			log.Error(LogRoutesEmpty)
 
+			p.metrics.UpdateGauge("routes_count", 0)
+			p.metrics.UpdateGauge("routes_byte", 0)
```
AlexanderYastrebov, line 76: I think we should not set metrics to zero here as we do not reset routes.

AlexanderYastrebov: Maybe `p.metrics.IncCounter("routes_empty")`

routesrv/routesrv.go
szuecs: yes, true what do you think about splitting similar to skipper itself?
/routes and /health would be the main listener and /metrics and /debug on supportListener.

routesrv/eskipbytes.go
```diff
@@ -11,10 +11,35 @@ import (
...
+func (w *responseWriterInterceptor) WriteHeader(statusCode int) {
+	w.statusCode = statusCode
+	w.ResponseWriter.WriteHeader(statusCode)
+}
+
+func (w *responseWriterInterceptor) Header() http.Header {
+	return w.ResponseWriter.Header()
+}
+
+func (w *responseWriterInterceptor) Write(p []byte) (int, error) {
```
AlexanderYastrebov, line 34: Do we use bytesWritten at all?

routesrv/eskipbytes.go
```diff
@@ -11,10 +11,35 @@ import (
 
 	ot "github.com/opentracing/opentracing-go"
 	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/metrics"
 	"github.com/zalando/skipper/routing"
 	"github.com/zalando/skipper/tracing"
 )
 
+type responseWriterInterceptor struct {
```
AlexanderYastrebov, line 19: This will break all usecases for Flusher, Pusher, CloseNotifier, and Hijacker.

routesrv/polling.go
szuecs: Mainly reuse our code. Our code also uses labels, so I don't see the benefit of defining here a large chunk of implementation specific metrics.

routesrv/polling.go
szuecs: Yes good idea!

routesrv/eskipbytes.go
AlexanderYastrebov: I guess it is not important for our usecase. Its a pity that the only way to capture status is to wrap ResponseWriter.

Here are some interesting related links:
* https://stackoverflow.com/questions/53272536/how-do-i-get-response-statuscode-in-golang-middleware
* https://github.com/kubernetes/kubernetes/blob/a9593d634c6a053848413e600dadbf974627515f/staging/src/k8s.io/apiserver/pkg/endpoints/responsewriter/wrapper.go#L37-L96
* https://github.com/golang/go/issues/54136#issuecomment-1259894379

routesrv/polling.go
szuecs: done

routesrv/eskipbytes.go
AlexanderYastrebov, line 34: https://pkg.go.dev/net/http#ResponseWriter

// If WriteHeader has not yet been called, Write calls
// WriteHeader(http.StatusOK) before writing the data.

I guess we need to if status is 0 (i.e. WriteHeader was not called) and use 200 instead.

routesrv/routesrv.go
AlexanderYastrebov: Yeah, I think we should have it the same as in skipper for simplicity

routesrv/eskipbytes.go
szuecs: True forget to expose it

routesrv/polling.go
```diff
@@ -90,6 +72,9 @@ func (p *poller) poll(wg *sync.WaitGroup) {
 		case routesCount == 0:
 			log.Error(LogRoutesEmpty)
 
+			p.metrics.UpdateGauge("routes_count", 0)
```
AlexanderYastrebov, line 75: Lets check if we should use the same metrics names as in https://github.com/zalando/skipper/pull/2597

routesrv/eskipbytes.go
szuecs: Kubernetes code looks like Kubernetes code :)
The SO fixes just half of the problem.

routesrv/eskipbytes.go
szuecs: this we do now

routesrv/routesrv.go
szuecs: this was done

routesrv/polling.go
szuecs: done

webframp:
Also commented in Slack. 

These look nice. Seems to still be complementary to what @universam1 and I worked on in https://github.com/zalando/skipper/pull/2597 even if we need to update a few of queries or dashboards to use these new names.

How do you see keys on `routesrv_custom_gauges` being used? Mostly wondering if you had any useful example promql queries that you've used in testing so far.

I guess it's just adding more keys than what we had started with, since there's now start, initialized and updated.

universam1:
Great progress!
I like the new buckets for histograms which might become handy at some point.
Recently the discussion came up if we got metrics for endpointslice length for skipper instance sanity validation, would that fit here?

szuecs:
> Also commented in Slack.

Me too, but let's have the discussion here, then we are transparent to everyone.

> 
> These look nice. Seems to still be complementary to what @universam1 and I worked on in #2597 even if we need to update a few of queries or dashboards to use these new names.

Yes that's what @AlexanderYastrebov pointed out and why I reached out to you.

> 
> How do you see keys on `routesrv_custom_gauges` being used? Mostly wondering if you had any useful example promql queries that you've used in testing so far.

The problem of Prometheus is that you need to set metric names and define the label keys at `init()` time and metrics.Metrics can not know this ahead of time.
The workaround is to use "key" values as the identifiers of "what" the metric is about.
Because you created the PR that exposed the data in routing package and maybe also rely on the current Prometheus , I wanted to ask for a buy-in that the change is fine.

> 
> I guess it's just adding more keys than what we had started with, since there's now start, initialized and updated.

Yes exactly, but also change the current routesrv metrics, because today routesrv uses plain Prometheus metrics https://github.com/zalando/skipper/blob/master/routesrv/polling.go#L29-L45

I would of course document the metric keys in docs/operation/operation.md , because it is will not be obvious to everyone.

szuecs:
> Great progress! I like the new buckets for histograms which might become handy at some point. Recently the discussion came up if we got metrics for endpointslice length for skipper instance sanity validation, would that fit here?

endpointslice length: Do you mean per slice or the total number of endpoints inside the slices for a given service?

universam1:
> endpointslice length: Do you mean per slice or the total number of endpoints inside the slices for a given service?

Guess lenght partitioned by service via label would be easy to correlate to Kube API. 
Would provide sanity check and discover update delays

szuecs:
> > endpointslice length: Do you mean per slice or the total number of endpoints inside the slices for a given service?
> 
> Guess lenght partitioned by service via label would be easy to correlate to Kube API. Would provide sanity check and discover update delays

so basically what you want to know is number of lb members by service, right?
The endpointslices can have duplicates and different states of these duplicate endpoints.

PR #2633, opened 2023-09-27
OPA: Improve opa startup delay handling

wisinghe:
This PR adds the following modifications

- Time to wait for opa plugins to report `READY` is made configurable with flag `open-policy-agent-startup-timeout`
- If all plugins do not report `READY` within the given timeout, started opa instance is closed and an error is returned
- Tests added for different startup errors


config/config.go
```diff
@@ -497,6 +498,7 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.OpenPolicyAgentConfigTemplate, "open-policy-agent-config-template", "", "file containing a template for an Open Policy Agent configuration file that is interpolated for each OPA filter instance")
 	flag.StringVar(&cfg.OpenPolicyAgentEnvoyMetadata, "open-policy-agent-envoy-metadata", "", "JSON file containing meta-data passed as input for compatibility with Envoy policies in the format")
 	flag.DurationVar(&cfg.OpenPolicyAgentCleanerInterval, "open-policy-agent-cleaner-interval", openpolicyagent.DefaultCleanIdlePeriod, "JSON file containing meta-data passed as input for compatibility with Envoy policies in the format")
+	flag.DurationVar(&cfg.OpenPolicyAgentStartupTimeout, "open-policy-agent-startup-timeout", openpolicyagent.DefaultOpaStartupTimeout, "Duration to wait for the open policy agent to start up")
```
Pushpalanka, line 501: Nitpick: Would be convenient to indicate the time unit in the description.

Pushpalanka (commented):
Added some nitpick comments. Feel free to ignore as well.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -292,9 +301,9 @@ func (registry *OpenPolicyAgentRegistry) newOpenPolicyAgentInstance(bundleName s
 		return nil, err
 	}
 
-	err = engine.waitPluginsReady(100*time.Millisecond, 30*time.Second)
+	err = engine.waitPluginsReady(100*time.Millisecond, config.startupTimeout)
```
mjungsbluth, line 304: You will need to make this a default on config in case this is used without using the `WithStartupTimeout` option. All the options are optional an can be passed but it should work without making them mandatory

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -409,9 +417,22 @@ func (opa *OpenPolicyAgentInstance) waitPluginsReady(checkInterval, timeout time
...
+		for pluginName, status := range opa.manager.PluginStatus() {
+			if status != nil && status.State != plugins.StateOK {
+				opa.Logger().WithFields(map[string]interface{}{
+					"plugin_name":   pluginName,
+					"plugin_state":  status.State,
+					"error_message": status.Message,
+				}).Error("Open policy agent plugin did not start in %v", timeout)
+			}
+		}
+		err = fmt.Errorf("one or more open policy agent plugins failed to start in %v with error: %v", timeout, err.Error())
```
mjungsbluth, line 432:
```suggestion
		err = fmt.Errorf("one or more open policy agent plugins failed to start in %v with error: %w", timeout, err)
```
This allows error chaining.

filters/openpolicyagent/openpolicyagent.go
wisinghe: Added in line 162

go.sum
```diff
@@ -13,37 +13,155 @@ cloud.google.com/go v0.56.0/go.mod h1:jr7tqZxxKOVYizybht9+26Z/gUq7tiRzu+ACVAMbKV
 cloud.google.com/go v0.57.0/go.mod h1:oXiQ6Rzq3RAkkY7N6t3TcE6jE+CIBBbA36lwQ1JyzZs=
 cloud.google.com/go v0.62.0/go.mod h1:jmCYTdRCQuc1PHIIJ/maLInMho30T/Y0M4hTdTShOYc=
 cloud.google.com/go v0.65.0/go.mod h1:O5N8zS7uWy9vkA9vayVHs65eM1ubvY4h553ofrNHObY=
+cloud.google.com/go v0.110.2/go.mod h1:k04UEeEtb6ZBRTv3dZz4CeJC3jKGxyhl0sAiVVquxiw=
```
szuecs, line 16: I think you won't need the changes in go.sum.

config/config.go
```diff
@@ -496,7 +497,8 @@ func NewConfig() *Config {
 	flag.BoolVar(&cfg.EnableOpenPolicyAgent, "enable-open-policy-agent", false, "enables Open Policy Agent filters")
 	flag.StringVar(&cfg.OpenPolicyAgentConfigTemplate, "open-policy-agent-config-template", "", "file containing a template for an Open Policy Agent configuration file that is interpolated for each OPA filter instance")
 	flag.StringVar(&cfg.OpenPolicyAgentEnvoyMetadata, "open-policy-agent-envoy-metadata", "", "JSON file containing meta-data passed as input for compatibility with Envoy policies in the format")
-	flag.DurationVar(&cfg.OpenPolicyAgentCleanerInterval, "open-policy-agent-cleaner-interval", openpolicyagent.DefaultCleanIdlePeriod, "JSON file containing meta-data passed as input for compatibility with Envoy policies in the format")
+	flag.DurationVar(&cfg.OpenPolicyAgentCleanerInterval, "open-policy-agent-cleaner-interval", openpolicyagent.DefaultCleanIdlePeriod, "Duration in seconds to wait before cleaning up unused opa instances")
```
AlexanderYastrebov, line 500: Can we rename `DefaultCleanIdlePeriod` to `DefaultCleanerInterval` for consistency?

szuecs:
@wisinghe 2 things:
- How do you make sure that skipper does not start it's Listener in case of OPA times out?
- I don't see that there is a test that covers a timeout, please add one.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -292,9 +303,9 @@ func (registry *OpenPolicyAgentRegistry) newOpenPolicyAgentInstance(bundleName s
 		return nil, err
 	}
 
-	err = engine.waitPluginsReady(100*time.Millisecond, 30*time.Second)
+	err = engine.waitPluginsReady(100*time.Millisecond, config.startupTimeout)
 	if err != nil {
-		engine.Logger().WithFields(map[string]interface{}{"err": err}).Error("Failed to wait for plugins activation.")
+		engine.Close(ctx)
```
AlexanderYastrebov, line 308: I am wondering if we should use https://pkg.go.dev/context#WithTimeout for `engine.Start(ctx)` instead of `context.Background()`

wisinghe:
> @wisinghe 2 things:
> 
> * How do you make sure that skipper does not start it's Listener in case of OPA times out?
> * I don't see that there is a test that covers a timeout, please add one.

@szuecs regarding the 1st point, do you mean the skipper itself should not start (listening on 9090 by default) in case the opa plugin fails with timeout?

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -287,14 +297,12 @@ func (registry *OpenPolicyAgentRegistry) newOpenPolicyAgentInstance(bundleName s
...
-	}
+	ctx, cancel := context.WithTimeout(context.Background(), config.startupTimeout)
+	defer cancel()
 
-	err = engine.waitPluginsReady(100*time.Millisecond, 30*time.Second)
-	if err != nil {
-		engine.Logger().WithFields(map[string]interface{}{"err": err}).Error("Failed to wait for plugins activation.")
+	if err = engine.Start(ctx, config.startupTimeout); err != nil {
+		engine.Close(ctx)
+		engine = &OpenPolicyAgentInstance{}
```
AlexanderYastrebov, line 305: This looks redundant.

filters/openpolicyagent/openpolicyagent.go
AlexanderYastrebov: Note that if context has expired, `engine.Close(ctx)` would likely fail.
Do we even need to Close if it failed to start?

wisinghe: Within `engine.Start`, there's a faster function that starts the engine officially. However we need some extra opa plugins to be ready in order for the started engine to be useful. That extra wait for the plugins is likely to exceed the deadline for various error scenarios. Even when the deadline is met, engine is started with non ready plugins therefore had to be closed explicitly. But let me try to make it a bit better

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -409,9 +407,48 @@ func (opa *OpenPolicyAgentInstance) waitPluginsReady(checkInterval, timeout time
...
+		return fmt.Errorf("one or more open policy agent plugins failed to start in %v with error: %w", timeout, err)
+	}
+	return nil
+}
+
+func (opa *OpenPolicyAgentInstance) Close(ctx context.Context) {
+	opa.once.Do(func() {
+		opa.manager.Stop(ctx)
+	})
+	opa.stopped = true
```
AlexanderYastrebov, line 432: Should be inside once.Do

wisinghe:
> @wisinghe 2 things:
> 
> * How do you make sure that skipper does not start it's Listener in case of OPA times out?
> * I don't see that there is a test that covers a timeout, please add one.

Introduced a new struct field `stopped` to `OpenPolicyAgentInstance` to keep the track of the instance's state. In case of a timeout, we call an `Stop()` function which will update the field as well stopping the started plugins. Also added a test to confirm this.

Regarding the 2nd point, added tests (for example https://github.com/zalando/skipper/pull/2633/files#diff-fa1cae8cd3077210b2f862d9eca9c7afbf3c502d3a5b6731ff5083609f69a982R228) create a time out scenario by providing incorrect configuration. Incorrect config leads to plugin not reporting ready until the timeout and therefore closing the engine + returning error.

AlexanderYastrebov:
Please resolve addressed comments and let us know when it is ready for the review.

wisinghe:
@AlexanderYastrebov @szuecs Thanks for the reviews. Addressed/resolved the comments so far. Kindly have another look and let me know if there are further concerns

AlexanderYastrebov:
:+1:

wisinghe:
👍

szuecs:
:+1:

PR #2427, opened 2023-06-27
dataclient/kubernetes: reduce amount of logs

AlexanderYastrebov:
Log Ingress/RouteGroup processing messages at most once per minute unless log level is debug.

PR #2637, opened 2023-09-29
dataclients/kubernetes: improve logging

AlexanderYastrebov:
* log number of received routegroups
* log distinct messages once per ingress/routegroup
* log number or all routes loaded and mapped

### DIFF
--- a/dataclients/kubernetes/clusterclient_test.go
+++ b/dataclients/kubernetes/clusterclient_test.go
@@ -6,8 +6,11 @@ import (
 	"os"
 	"strings"
 	"testing"
+	"time"
 
 	log "github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/zalando/skipper/dataclients/kubernetes"
 	"github.com/zalando/skipper/dataclients/kubernetes/kubernetestest"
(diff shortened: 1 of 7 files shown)


dataclients/kubernetes/clusterclient_test.go
```diff
@@ -251,3 +254,63 @@ spec:
 		})
 	}
 }
+
+func TestLoggingInterval(t *testing.T) {
```
AlexanderYastrebov, line 258: Moved from logger_test.go

dataclients/kubernetes/logger_test.go
```diff
@@ -1,76 +1,82 @@
...
+		l.Debugf("test message")
+		l.Infof("test message")
+		l.Errorf("test message")
+
+		assert.Equal(t, 0, strings.Count(out.String(), `level=debug`))
+		assert.Equal(t, 0, strings.Count(out.String(), `level=info`))
+		assert.Equal(t, 1, strings.Count(out.String(), `level=error`))
+	})
+
+	t.Run("test logs once per level", func(t *testing.T) {
```
MustafaSaber, line 52: Maybe another test case for multiple logs but we only count 1? this should yield also to 1 in debug level and not 1 in other levels right?

AlexanderYastrebov: Not sure I get it. Previous test checks we log a single message and this checks that we log once

dataclients/kubernetes/logger.go
```diff
@@ -1,32 +1,62 @@
...
-		l.logger.Errorf(format, args...)
+		l.once(log.ErrorLevel, format, args...)
+	}
+}
+
+func (l *logger) once(level log.Level, format string, args ...any) {
+	if !l.logger.Logger.IsLevelEnabled(level) {
+		return
+	}
+	l.mu.Lock()
```
RomanZavodskikh, line 49: Is it okay to create high lock contention on the logger (from the quick look it seems it must be used very frequently)?

dataclients/kubernetes/logger_test.go
MustafaSaber: Yeah my bad, I misunderstood the loop

dataclients/kubernetes/logger.go
RomanZavodskikh: Mustafa pointed out we have some loggingInterval constant https://github.com/zalando/skipper/blob/v0.18.14/dataclients/kubernetes/logger_test.go#L42 which defines how often logger will produce output.
Let's imagine l.Errorf is callel 1_000_000 times in the single loggingInterval. In this case, will this mutex be locked 1_000_000 times? Or only once?

AlexanderYastrebov: Dataclient is actually single threaded so there is no lock contention, I added mutex because underlying logger is thread-safe so I wanted to keep this property.

We ratelimit logs to log every minute, see https://github.com/zalando/skipper/pull/2427

This change eliminates duplicate messages within single update cycle.

RomanZavodskikh:
:+1:

szuecs:
:+1:

Issue #2394, opened 2023-06-12
Better handling for duplicate routes in skipper


MustafaSaber:
### What needs to be done and why is it needed?

We go through all routes and see if `routeID` and body of routes match we log a warning [code](https://github.com/zalando/skipper/blob/47eef54a73200270ccf78211e9b05f09c9919e94/dataclients/kubernetes/kube.go#L354-L371)
So if we have 2 catchall routes for same hostnames but different RG it will also log, which could be misleading

### What needs to be changed?
**For discussion**

Option 1: Only log when RG/Ingress that has duplicate hostname.
Option 2: Add ingress/rg name to routeID (If we add namespace+name it will make route ids different but we would still have duplicate per-host catchall routes.)
Option 3: We should change [main routegroup loop](https://github.com/zalando/skipper/blob/47eef54a73200270ccf78211e9b05f09c9919e94/dataclients/kubernetes/routegroup.go#L560-L663) to only add catchall hostname routes once per hostname. 

The same issue may also exist for ingress, we need to check that e.g. by adding a test case with two ingresses for the same hostname.

- [ ] Add test cases for 2 different ingress/routegroups that won't log

AlexanderYastrebov:
Also Skipper datasource neither detects duplicate routes from the same dataclient nor between multiple dataclients and overwrites existing route with the latest update which could be shown via:
```diff
diff --git a/routing/datasource.go b/routing/datasource.go
index cf006f6f..74cf4ad3 100644
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -120,7 +120,7 @@ func receiveFromClient(c DataClient, o Options, out chan<- *incomingData, quit <
 
 // applies incoming route definitions to key/route map, where
 // the keys are the route ids.
-func applyIncoming(defs routeDefs, d *incomingData) routeDefs {
+func applyIncoming(defs routeDefs, d *incomingData, l logging.Logger) routeDefs {
        if d.typ == incomingReset || defs == nil {
                defs = make(routeDefs)
        }
@@ -133,6 +133,9 @@ func applyIncoming(defs routeDefs, d *incomingData) routeDefs {
 
        if d.typ == incomingReset || d.typ == incomingUpdate {
                for _, def := range d.upsertedRoutes {
+                       if _, ok := defs[def.Id]; ok {
+                               l.Infof("overwriting %q route definition (same dataclient)", def.Id)
+                       }
                        defs[def.Id] = def
                }
        }
[...]

PR #2646, opened 2023-10-04
dataclients/kubernetes: do not create catchall routes

AlexanderYastrebov:
Catchall routes are created for Ingresses and RouteGroups that do not define predicates matching all requests for the host.

Initially catchall routes were introduced by #425 for #408. Later it was discovered that multiple Ingresses for the same host may create conflicting catchall routes which was fixed by #436.

Addition of RouteGroups #1180 introduced catchall routes for them as well but did not take into account that and Ingress and RouteGroup may use the same host which results in a catchall route created e.g. for RouetGroup that interferes with routes created for Ingress.

This change removes catchall route logic because it simply shunts the request and responds with the default response status which matches the behaviour when Skipper can not look up route at all.

Updates #408
Updates #2394


AlexanderYastrebov:
While #408 intended to return 502 for route lookup failures and 404 for catchall routes it looks like this was overlooked in the implementation https://github.com/zalando/skipper/pull/425#discussion_r135405470 which made catchall routes return the same default code as when route lookup failed and thus made catchall routes redundant.

AlexanderYastrebov:
See #2656

PR #2634, opened 2023-09-27
Decouple fadeIn from loadbalancer

RomanZavodskikh:
Decouple fadeIn from loadbalancer

PR #2648, opened 2023-10-05
Added LBEndpoint field in routing.LBContext struct

RomanZavodskikh:
The routing.Route field of routing.LBContext struct will be eventually removed because loadbalancer in general does not require the full information about route, only about endpoints to balance load between.

You can think about this PR as preliminary one for the https://github.com/zalando/skipper/pull/2634

### DIFF
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -73,7 +73,7 @@ func fadeIn(now time.Time, duration time.Duration, exponent float64, detected ti
 func shiftWeighted(rnd *rand.Rand, ctx *routing.LBContext, now time.Time) routing.LBEndpoint {
 	var sum float64
 	rt := ctx.Route
-	ep := ctx.Route.LBEndpoints
+	ep := ctx.LBEndpoints
 	for _, epi := range ep {
 		detected := ctx.Registry.GetMetrics(epi.Host).DetectedTime()
 		wi := fadeIn(now, rt.LBFadeInDuration, rt.LBFadeInExponent, detected)
@@ -97,7 +97,7 @@ func shiftWeighted(rnd *rand.Rand, ctx *routing.LBContext, now time.Time) routin
 
 func shiftToRemaining(rnd *rand.Rand, ctx *routing.LBContext, wi []int, now time.Time) routing.LBEndpoint {
 	notFadingIndexes := wi
-	ep := ctx.Route.LBEndpoints
+	ep := ctx.LBEndpoints
 
 	// if all endpoints are fading, the simplest approach is to use the oldest,
 	// this departs from the desired curve, but guarantees monotonic fade-in. From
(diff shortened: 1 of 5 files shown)


proxy/proxy.go
```diff
@@ -469,6 +469,8 @@ func setRequestURLForDynamicBackend(u *url.URL, stateBag map[string]interface{})
 }
 
 func setRequestURLForLoadBalancedBackend(u *url.URL, rt *routing.Route, lbctx *routing.LBContext) *routing.LBEndpoint {
+	lbctx.LBEndpoints = rt.LBEndpoints
```
AlexanderYastrebov, line 472: Lets do this either at the place where context is created or maybe better move context creation here.

routing/routing.go
```diff
@@ -190,10 +190,11 @@ type LBAlgorithm interface {
...
 type LBContext struct {
-	Request  *http.Request
-	Route    *Route
-	Params   map[string]interface{}
-	Registry *EndpointRegistry
+	Request     *http.Request
+	Route       *Route
+	Params      map[string]interface{}
+	Registry    *EndpointRegistry
+	LBEndpoints []LBEndpoint
```
AlexanderYastrebov, line 197: Lets move this between Route and Params

proxy/proxy.go
RomanZavodskikh: > Lets do this either at the place where context is 
I like this idea, I've done that.

> or maybe better move context creation here.
I like this way less, because I need to redeclare a lot variables from `func mapRequest(...` in `func setRequestURLForLoadBalancedBackend(...`
```
index 80b41df3..dddf641f 100644
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -468,7 +468,13 @@ func setRequestURLForDynamicBackend(u *url.URL, stateBag map[string]interface{})
        }
 }
 
-func setRequestURLForLoadBalancedBackend(u *url.URL, rt *routing.Route, lbctx *routing.LBContext) *routing.LBEndpoint {
+func setRequestURLForLoadBalancedBackend(ctx *context, registry *routing.EndpointRegistry) *routing.LBEndpoint {
+       r := ctx.request
+       rt := ctx.route
+       stateBag := ctx.StateBag()
+       u := r.URL
+       lbctx := &routing.LBContext{Request: r, Route: rt, LBEndpoints: rt.LBEndpoints, Params: stateBag, Registry: registry}
+
        e := rt.LBAlgorithm.Apply(lbctx)
        u.Scheme = e.Scheme
        u.Host = e.Host
@@ -490,7 +496,7 @@ func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHea
        case eskip.LBBackend:
-               endpoint = setRequestURLForLoadBalancedBackend(u, rt, &routing.LBContext{Request: r, Route: rt, LBEndpoints: rt.LBEndpoints, Params: stateBag, Registry: registry})
+               endpoint = setRequestURLForLoadBalancedBackend(ctx, registry)
        default:
                u.Scheme = rt.Scheme
                u.Host = rt.Host

```

loadbalancer/algorithm.go
```diff
@@ -73,7 +73,7 @@ func fadeIn(now time.Time, duration time.Duration, exponent float64, detected ti
 func shiftWeighted(rnd *rand.Rand, ctx *routing.LBContext, now time.Time) routing.LBEndpoint {
 	var sum float64
 	rt := ctx.Route
-	ep := ctx.Route.LBEndpoints
```
AlexanderYastrebov, line 76: Please check that we do not accidentally use `ctx.Route.LBEndpoints` somewhere else in algorithms.

AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

proxy/proxy.go
AlexanderYastrebov: This is actually better because it nicely contains all endpoint selection logic and creates a place to extend in the future.
Also  
```
	u.Scheme = e.Scheme
	u.Host = e.Host
```
can be moved out of setRequestURLForLoadBalancedBackend as its not related to endpoint selection,
setRequestURLForLoadBalancedBackend could be renamed to selectLoadBalancedEndpoint (or something alike)

AlexanderYastrebov: https://github.com/zalando/skipper/pull/2652

PR #2649, opened 2023-10-05
test: add shadow traffic tests

szuecs:
test: add shadow traffic tests

### DIFF
--- a/proxy/shadow_test.go
+++ b/proxy/shadow_test.go
@@ -1,0 +1,398 @@
+package proxy
+
+import (
+	"fmt"
+	"io"
+	"math/rand"
+	"net/http"
+	stdlibhttptest "net/http/httptest"
+	"os"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/filters/scheduler"
+	"github.com/zalando/skipper/filters/tee"
+	"github.com/zalando/skipper/metrics/metricstest"
+	"github.com/zalando/skipper/net/httptest"
+	"github.com/zalando/skipper/predicates/primitive"
+	teepred "github.com/zalando/skipper/predicates/tee"
+	"github.com/zalando/skipper/predicates/traffic"
+	"github.com/zalando/skipper/routing"
+	"github.com/zalando/skipper/routing/testdataclient"
+	sched "github.com/zalando/skipper/scheduler"
+)
+
+func TestShadowSingle(t *testing.T) {
+	s := stdlibhttptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Add("Backend", "main")
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte("OK"))
+	}))
+	defer s.Close()
+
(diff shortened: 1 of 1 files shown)


proxy/shadow_test.go
```diff
@@ -0,0 +1,405 @@
...
+
+	ts := stdlibhttptest.NewServer(pr)
+	defer ts.Close()
+
+	reqURL, err := url.Parse(ts.URL)
+	if err != nil {
+		t.Fatalf("Failed to parse url %s: %v", ts.URL, err)
+	}
+
+	client := http.DefaultClient
```
AlexanderYastrebov, line 98: Why not use ts.Client() here?

proxy/shadow_test.go
```diff
@@ -0,0 +1,405 @@
...
+	})
+	defer pr.Close()
+
+	ts := stdlibhttptest.NewServer(pr)
+	defer ts.Close()
+
+	reqURL, err := url.Parse(ts.URL)
+	if err != nil {
+		t.Fatalf("Failed to parse url %s: %v", ts.URL, err)
+	}
```
AlexanderYastrebov, line 96: I think we should just use `ts.URL` in the request there is no need to parse it

proxy/shadow_test.go
```diff
@@ -0,0 +1,405 @@
...
+	fr.Register(scheduler.NewFifo())
+
+	doc := fmt.Sprintf(`
+main: PathSubtree("/")
+-> "%s";
+split: PathSubtree("/") && Traffic(0.5)
+-> teeLoopback("test")
+-> "%s";
+shadow: PathSubtree("/") && Tee("test") && True()
+-> "%s";
```
AlexanderYastrebov, line 61: Maybe if we indent inside of string block it would look nicer?

proxy/shadow_test.go
szuecs: done

proxy/shadow_test.go
szuecs: done

proxy/shadow_test.go
szuecs: done

szuecs:
:+1:

proxy/shadow_test.go
```diff
@@ -0,0 +1,398 @@
...
+					return fmt.Errorf("fifo full %d", statusFifoFull)
+				}
+				if statusFifoTimeout == 0 {
+					return fmt.Errorf("fifo timeout %d", statusFifoTimeout)
+				}
+				if statusFifoErr != 0 {
+					return fmt.Errorf("fifo err %d", statusFifoErr)
+				}
+				return nil
+			},
```
AlexanderYastrebov, line 232: This is quite hard to read when placed in the table test. Alternatively you could have a test helper method that setups proxy and runs load and then use it in a series of nested t.Runs like
```go
t.Run("foo", func(t *testing.T) {
	va := runTest(`
	  main: PathSubtree("/")
            -> fifo(2, 5, "200ms")
            -> "$backendURL";
          split: PathSubtree("/") && Traffic(1.0)
            -> fifo(2, 5, "200ms")
            -> teeLoopback("test")
            -> "$backendURL";
          shadow: PathSubtree("/") && Tee("test") && True()
            -> fifo(2, 5, "200ms")
            -> "$shadowURL";
	`)
	...
})
```

AlexanderYastrebov:
:+1:

PR #2652, opened 2023-10-05
proxy: refactor loadbalanced endpoint selection

AlexanderYastrebov:
Rename setRequestURLForLoadBalancedBackend to selectEndpoint and move loadbalancing context creation inside

Followup on #2648

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -468,10 +468,18 @@ func setRequestURLForDynamicBackend(u *url.URL, stateBag map[string]interface{})
 	}
 }
 
-func setRequestURLForLoadBalancedBackend(u *url.URL, rt *routing.Route, lbctx *routing.LBContext) *routing.LBEndpoint {
+func selectEndpoint(ctx *context, registry *routing.EndpointRegistry) *routing.LBEndpoint {
+	rt := ctx.route
+
+	lbctx := &routing.LBContext{
+		Request:     ctx.request,
+		Route:       rt,
+		LBEndpoints: rt.LBEndpoints,
+		Params:      ctx.StateBag(),
+		Registry:    registry,
+	}
 	e := rt.LBAlgorithm.Apply(lbctx)
-	u.Scheme = e.Scheme
-	u.Host = e.Host
+
 	return &e
 }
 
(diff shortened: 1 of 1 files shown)


proxy/proxy.go
```diff
@@ -468,10 +468,18 @@ func setRequestURLForDynamicBackend(u *url.URL, stateBag map[string]interface{})
 	}
 }
 
-func setRequestURLForLoadBalancedBackend(u *url.URL, rt *routing.Route, lbctx *routing.LBContext) *routing.LBEndpoint {
+func selectEndpoint(ctx *context, registry *routing.EndpointRegistry) *routing.LBEndpoint {
+	rt := ctx.route
+
```
AlexanderYastrebov, line 473: Here we can put fadeIn and passive healthchecks

proxy/proxy.go
```diff
@@ -468,10 +468,18 @@ func setRequestURLForDynamicBackend(u *url.URL, stateBag map[string]interface{})
 	}
 }
 
-func setRequestURLForLoadBalancedBackend(u *url.URL, rt *routing.Route, lbctx *routing.LBContext) *routing.LBEndpoint {
+func selectEndpoint(ctx *context, registry *routing.EndpointRegistry) *routing.LBEndpoint {
+	rt := ctx.route
```
RomanZavodskikh, line 472: minor: I don't understand why do we need rt variable here and why cannot we inline `ctx.route`?

proxy/proxy.go
```diff
@@ -490,7 +498,9 @@ func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHea
 		setRequestURLFromRequest(u, r)
 		setRequestURLForDynamicBackend(u, stateBag)
 	case eskip.LBBackend:
-		endpoint = setRequestURLForLoadBalancedBackend(u, rt, &routing.LBContext{Request: r, Route: rt, LBEndpoints: rt.LBEndpoints, Params: stateBag, Registry: registry})
+		endpoint = selectEndpoint(ctx, registry)
+		u.Scheme = endpoint.Scheme
```
RomanZavodskikh, line 502: minor: maybe use fallthrough here?

RomanZavodskikh:
:+1:

proxy/proxy.go
AlexanderYastrebov: rt.Apply

proxy/proxy.go
AlexanderYastrebov: no need to overcomplicate to save 1 line

AlexanderYastrebov: also it uses endpoint fields, not route fields

AlexanderYastrebov:
:+1:

PR #2667, opened 2023-10-09
filters/ratelimit: apply ratelimitFailClosed regardless of the position in the route

AlexanderYastrebov:
ratelimitFailClosed is a route-wide configuration filter so it should apply to all ratelimit filters regardless of its position. 

PR consists of two commits: first refactors tests to simplify and allow routes with multiple filters and second makes ratelimitFailClosed apply regardless of the position in the route.

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -1721,7 +1721,7 @@ As of now there is no negative/deny rule possible. The first matching path is ev
 
 ### Open Policy Agent
 
-To get started with [Open Policy Agent](https://www.openpolicyagent.org/), also have a look at the [tutorial](../tutorials/auth.md#open-policy-agent). This section is only a reference for the implemented filters. 
+To get started with [Open Policy Agent](https://www.openpolicyagent.org/), also have a look at the [tutorial](../tutorials/auth.md#open-policy-agent). This section is only a reference for the implemented filters.
 
 #### opaAuthorizeRequest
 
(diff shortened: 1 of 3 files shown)


docs/reference/filters.md
```diff
@@ -2255,16 +2255,17 @@ Path("/expensive") -> clusterLeakyBucketRatelimit("user-${request.cookie.Authori
 
 ### ratelimitFailClosed
 
-This filter changes the failure mode for rate limit filters. If the
-filter is present, infrastructure issues will lead to rate limit.
+This filter changes the failure mode for all rate limit filters of the route.
+By default rate limit filters fail open on infrastructure errors (e.g. redis is down).
```
gerkenv, line 2259:
```suggestion
By default rate limit filters fail open on infrastructure errors (e.g. redis is down), i.e request passes through.
```

szuecs:
@AlexanderYastrebov I think my intention was to have it as it was done. Not position independent for the route, because it would be weird from reading eskip. Don't you think so?

AlexanderYastrebov:
@szuecs Yeah, I understand, but is there a usecase to apply it only for the rest of the filter chain? If e.g. redis fails it likely fails for all filters in the chain.
I think by making it position-independent along with clarified docs we would make it less error-prone.

> it would be weird from reading eskip

We already have position independent filters, e.g. tracingTag, fadeIn, consistentHashKey, etc

szuecs:
:+1:

szuecs:
@AlexanderYastrebov should we formulate this out as expectation for the user, how a "config" filter will work?
Maybe it's the time to write it down that "config" filters will change all filters of a route.
wdyt?

AlexanderYastrebov:
:+1:

PR #2669, opened 2023-10-10
filters/auth: document forwardToken key sorting

AlexanderYastrebov:
Add a note that forwardToken JSON value has sorted keys and can be used as a ratelimit key.


szuecs:
:+1:

MustafaSaber:
👍

AlexanderYastrebov:
This was intended for internal usecase which was solved differently by means of
```
Path("/foo") 
-> setContextRequestHeader("X-Uid", "auth-user")
-> clusterClientRatelimit("foo", 10, "1m", "X-Uid")
-> dropRequestHeader("X-Uid")
```

Decided not to proceed with documenting this as tokeninfo array fields may have undefined and non-stable order.

PR #2645, opened 2023-10-04
Get inflight requests number from endpointregistry instead of endpoint metrics

RomanZavodskikh:
Get inflight requests number from endpointregistry instead of endpoint metrics

PR #2673, opened 2023-10-12
Do not use endpointregistry in the hotpath

RomanZavodskikh:
The performance of endpointregistry under circumstances of high concurenncy is the very discussable topic, it does not seem to work perfectly at the moment.


szuecs:
:+1:

AlexanderYastrebov:
:+1:

bocytko:
Could you also reference the commit & version that introduced this change?

szuecs:
@bocytko it was https://github.com/zalando/skipper/pull/2645

Issue #2156, opened 2022-11-24
Allow masking selected query parameters in access logs


AlexanderYastrebov:
**Is your feature request related to a problem? Please describe.**

Query parameters may contain sensitive data.
Currently it is only possible to disable access logs per-route via `disableAccessLog()` filter or disable query logging globally via `-access-log-strip-query` flag.
Users would like to have access logs but avoid logging sensitive data.

**Describe the solution you would like**

A filter (e.g. `maskAccessLogQuery("foo", "bar")`) that configures a list of query parameter names that should be masked in access logs.

**Additional context (optional)**

Masking individual parameters is an extension of the https://github.com/zalando/skipper/issues/705 idea.
The proposed filter could be used to achieve that when configured without arguments (e.g. `maskAccessLogQuery()` or maybe separate `disableAccessLogQuery()`)

See also https://github.com/zalando/skipper/issues/551

**Would you like to work on it?**
Yes

AlexanderYastrebov:
Masking could be implemented by hashing the actual value such that masked values still could be grouped.

PR #2674, opened 2023-10-12
New filter: maskAccessLogQuery

remychantenay:
## Summary
Adding a new `maskAccessLogQuery` filter to mask/obfuscate values of specific – sensitive – query parameters.

Towards #2156 

## Testing
```
# skip
hello: Path("/hello") -> maskAccessLogQuery("key_2") -> "https://www.example.org"

curl "localhost:9090/hello?key_1=foo&key_2=bar"
127.0.0.1 - - [18/Oct/2023:08:10:33 +0200] "GET /hello?key_1=foo&key_2=5234164152756840025 HTTP/1.1" 404 1256 "-" "curl/8.1.2" 555 localhost:9090 - -
```

### DIFF
--- a/logging/access.go
+++ b/logging/access.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"net"
 	"net/http"
-	"strconv"
 	"strings"
 	"time"
 
@@ -23,6 +22,10 @@ const (
 	combinedLogFormat = commonLogFormat + ` "%s" "%s"`
 	// We add the duration in ms, a requested host and a flow id and audit log
 	accessLogFormat = combinedLogFormat + " %d %s %s %s\n"
+
+	// KeyMaskedQueryParams represents the key used to store and retrieve masked query parameters
+	// from the additional data.
+	KeyMaskedQueryParams = "maskedQueryParams"
 )
 
 type accessLogFormatter struct {
@@ -111,13 +114,32 @@ func stripQueryString(u string) string {
 	}
 }
 
+// maskQueryParams masks (i.e., hashing) specific query parameters in the provided request's URI.
+// Returns the obfuscated URI.
+func maskQueryParams(req *http.Request, maskedQueryParams []string) string {
+	strippedURI := stripQueryString(req.RequestURI)
+
+	params := req.URL.Query()
+	for i := range maskedQueryParams {
+		val := params.Get(maskedQueryParams[i])
+		if val == "" {
+			continue
+		}
+
+		hashed := hash(val)
+		params.Set(maskedQueryParams[i], fmt.Sprintf("%d", hashed))
+	}
+
+	return fmt.Sprintf("%s?%s", strippedURI, params.Encode())
+}
+
 func hash(val string) uint64 {
 	return xxhash.Sum64String(val)
 }
 
 // Logs an access event in Apache combined log format (with a minor customization with the duration).
 // Additional allows to provide extra data that may be also logged, depending on the specific log format.
-func LogAccess(entry *AccessEntry, additional map[string]interface{}, maskedQueryParams []string) {
+func LogAccess(entry *AccessEntry, additional map[string]interface{}) {
 	if accessLog == nil || entry == nil {
 		return
 	}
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -37,6 +37,7 @@ require (
 	github.com/sarslanhan/cronmask v0.0.0-20230801193303-54e29300a091
 	github.com/sirupsen/logrus v1.9.3
 	github.com/sony/gobreaker v0.5.0
+	github.com/spaolacci/murmur3 v1.1.0
```
szuecs, line 40: please use github.com/cespare/xxhash/v2 instead. This one we already import, use and tested to be very good for sharding. From [benchmarks](https://xxhash.com/) it also looks like xxhash is 10x faster.

filters/accesslog/control.go
```diff
@@ -87,3 +91,32 @@ func (*enableAccessLog) Name() string { return filters.EnableAccessLogName }
...
+//
+//	maskAccessLogQuery("key_1", "key_2") to mask the value of provided keys in the access log.
+func NewMaskAccessLogQuery() filters.Spec {
+	return &maskAccessLogQuery{}
+}
+
+func (*maskAccessLogQuery) Name() string { return filters.MaskAccessLogQueryName }
+
+func (al *maskAccessLogQuery) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) <= 0 {
```
AlexanderYastrebov, line 108: len can't be negative

filters/accesslog/control.go
```diff
@@ -87,3 +91,32 @@ func (*enableAccessLog) Name() string { return filters.EnableAccessLogName }
...
+	keys := make([]string, len(args))
+	for i := range args {
+		if key, ok := args[i].(string); ok && key != "" {
+			keys[i] = key
+		} else {
+			return nil, filters.ErrInvalidFilterParameters
+		}
+	}
+
+	return &AccessLogFilter{Enable: true, MaskedQueryParams: keys}, nil
```
AlexanderYastrebov, line 121: This will override any preceding filter confgiration.

Maybe we should think about supporting a usecase where we could specify [default masked values](https://github.com/zalando-incubator/kubernetes-on-aws/blob/c78b2da5a3d0e9b59517f1d9317502369c347aae/cluster/config-defaults.yaml#L95)
```
disableAccessLog(2,3,404,429) -> accessLogMaskQueryParams("access_token") -> fifo(2000,20,"1s")
```
and then user may specify additional values.
OTOH proposed solution is simple and in line with what already exists.

logging/access.go
```diff
@@ -109,9 +113,19 @@ func stripQueryString(u string) string {
...
+	h32.Write([]byte(val))
+	hashedVal := h32.Sum32()
+
+	return fmt.Sprint(hashedVal)
+}
+
 // Logs an access event in Apache combined log format (with a minor customization with the duration).
 // Additional allows to provide extra data that may be also logged, depending on the specific log format.
-func LogAccess(entry *AccessEntry, additional map[string]interface{}) {
+func LogAccess(entry *AccessEntry, additional map[string]interface{}, maskedQueryParams []string) {
```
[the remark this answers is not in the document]
AlexanderYastrebov: There is an exported statebag key AccessLogAdditionalDataKey to provide additional data
and additional data can override predefined fields https://github.com/zalando/skipper/blob/9b570970ea935053830703fcd947b0aa76896625/logging/access.go#L152-L171

so another option is to create a filter that would mask request uri and put it into `ctx.StateBag()[al.AccessLogAdditionalDataKey]["uri"]`.

filters/accesslog/control.go
remychantenay: Of course, done 👍

logging/access.go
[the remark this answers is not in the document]
remychantenay: Thank you for the feedback 👍

I updated the code and went with @szuecs's suggestion. Hope that's fine.

filters/accesslog/control.go
remychantenay: Hey @AlexanderYastrebov 

Yeah, I wanted to keep it simple. In hindsight, I am not sure my thinking is right.

If the current implementation is a no-go, I will try and spend a bit more time and see if I can come up with something when I get back from vacay (from tomorrow till the beginning of next month.)

logging/access.go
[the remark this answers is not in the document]
AlexanderYastrebov: Since we rely on additional data anyways lets use `"uri"` key - this way masking logic would belong to the filter and not to the access log package (e.g. there could be several ways to mask uri). The filter would need only to populate additional data in the statebag.

You can export `const UriAdditionalDataKey = "uri"` if you want and use it everywhere instead of `"uri"` .

PR #2675, opened 2023-10-13
feature: limit concurrency

szuecs:


proxy/proxy.go
```diff
@@ -1484,6 +1484,11 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			ctx.Logger().Errorf("Failed to set write deadline: %v", e)
 		}
 	}
+	if sbf, ok := ctx.StateBag()[filters.LimitConcurrency]; ok {
+		if f, ok := sbf.(func()); ok {
+			defer f()
+		}
+	}
```
AlexanderYastrebov, line 1491: Why do we need to defer it here instead of just decrementing on Response?

filters/scheduler/limitconcurrency.go
```diff
@@ -0,0 +1,71 @@
...
+	"net/http"
+	"sync/atomic"
+
+	"github.com/zalando/skipper/filters"
+)
+
+type (
+	limitConcurrencySpec   struct{}
+	limitConcurrencyFilter struct {
+		concurrentRequests *atomic.Int64
```
AlexanderYastrebov, line 14: There is no need to use pointer here

filters/scheduler/limitconcurrency.go
```diff
@@ -0,0 +1,71 @@
...
+func (s *limitConcurrencySpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) != 1 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	cc, err := intArg(args[0])
+	if err != nil {
+		return nil, err
+	}
+	if cc < 1 {
```
AlexanderYastrebov, line 39: I think we can allow zero for completeness here.

filters/scheduler/limitconcurrency.go
```diff
@@ -0,0 +1,71 @@
...
+	}, nil
+}
+
+// Request is the filter.Filter interface implementation. Request will
+// increase the number of concurrent requests and respond to the caller,
+// if the bounded queue returns an error. Status code by Error:
+//
+// - 503 if full
+func (f *limitConcurrencyFilter) Request(ctx filters.FilterContext) {
+	if f.maxConcurrency > f.concurrentRequests.Load() {
```
AlexanderYastrebov, line 55: This condition looks wrong

AlexanderYastrebov: I think this https://pkg.go.dev/golang.org/x/sync/semaphore#Weighted.TryAcquire might be more suitable here.

proxy/proxy.go
szuecs: Because if the backend is fully asynchronous we need to decrement after copystream (after the full body was sent to client )

szuecs:
close this in favor of https://github.com/zalando/skipper/pull/2685

PR #2676, opened 2023-10-13
feature: filter fifoWithBody

szuecs:
feature: filter fifoWithBody that works similar to fifo(), but release deferred until body streaming to client was finished

### DIFF
--- a/filters/scheduler/fifo.go
+++ b/filters/scheduler/fifo.go
@@ -150,3 +150,9 @@ func (f *fifoFilter) Response(ctx filters.FilterContext) {
 	pending[last]()
 	ctx.StateBag()[fifoKey] = pending[:last]
 }
+
+// HandleErrorResponse is to opt-in for filters to get called
+// Response(ctx) in case of errors via proxy. It has to return true to opt-in.
+func (f *fifoFilter) HandleErrorResponse() bool {
+	return true
+}
--- a/filters/scheduler/fifo.go
+++ b/filters/scheduler/fifo.go
@@ -15,20 +15,44 @@ const (
 	fifoKey string = "fifo"
 )
 
+type fifoType int
+
+const (
+	fifo fifoType = iota + 1
+	fifoWithBody
+)
+
 type (
-	fifoSpec   struct{}
+	fifoSpec struct {
+		typ fifoType
+	}
 	fifoFilter struct {
 		config scheduler.Config
 		queue  *scheduler.FifoQueue
+		typ    fifoType
 	}
 )
 
 func NewFifo() filters.Spec {
-	return &fifoSpec{}
+	return &fifoSpec{
+		typ: fifo,
+	}
 }
 
-func (*fifoSpec) Name() string {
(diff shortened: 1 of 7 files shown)


szuecs:
@AlexanderYastrebov maybe better than https://github.com/zalando/skipper/pull/2675 wdyt?

filters/scheduler/fifo.go
```diff
@@ -139,14 +164,26 @@ func (f *fifoFilter) Request(ctx filters.FilterContext) {
...
 func (f *fifoFilter) Response(ctx filters.FilterContext) {
-	pending, ok := ctx.StateBag()[fifoKey].([]func())
-	if !ok {
-		return
+	g := f.createResponse(ctx)
+	switch f.typ {
+	case fifo:
+		g()
+	case fifoWithBody:
+		ctx.StateBag()[filters.FifoWithBody] = g
```
AlexanderYastrebov, line 172: Its not clear how multiple fifoWithBody would chain and how they would interplay with regular fifo.
I think you'd need to have the same logic stored under different keys and then run chain of hooks in two places in proxy depending on the key at proxy.do and proxy.ServeHTTP:


```go
...
func (f *fifoFilter) createResponse(ctx filters.FilterContext, typ string) {
		pending, ok := ctx.StateBag()[typ].([]func())
		if !ok {
			return
		}
		last := len(pending) - 1
		if last < 0 {
			return
		}
		pending[last]()
		ctx.StateBag()[fifoKey] = pending[:last]
}

...
// existing in proxy.do
pendingFIFO, _ := stateBag[scheduler.FIFOKey].([]func())
for _, done := range pendingFIFO {
	done()
}
...
// in proxy.ServeHTTP
pendingFIFO, _ := stateBag[scheduler.FIFOWithBodyKey].([]func())
for _, done := range pendingFIFO {
	done()
}
```

filters/scheduler/fifo.go
```diff
@@ -15,20 +15,44 @@ const (
 	fifoKey string = "fifo"
 )
 
+type fifoType int
+
+const (
+	fifo fifoType = iota + 1
+	fifoWithBody
+)
```
AlexanderYastrebov, line 23: You can use filter name as spec.typ and have withBody bool in filter similar to https://github.com/zalando/skipper/blob/2a202c47a7b0f80091f4492374b5c83c868ea9cb/filters/diag/histogram.go#L56-L57

szuecs: I like the enum style better than string.

filters/scheduler/fifo.go
szuecs: Yes that's a good point. I also thought about that I need to do something like that.
What do you like more to support the copystream part? this one or the other?

AlexanderYastrebov: I think we can remove existing pendingFIFO from proxy.go see #2680

szuecs:
close to create a PR to cw2023freeze

PR #1086, opened 2019-06-07
fix stuck lifo() -> done() notifications

aryszka:
make sure each lifo done() is called and handle done() for multiple lifo() filters

Signed-off-by: Arpad Ryszka <arpad.ryszka@gmail.com>

PR #2680, opened 2023-10-16
filters/scheduler: handle errors

AlexanderYastrebov:
Make lifo, lifoGroup and fifo filters backed error aware and remove cleanup from the proxy.

Follow up on #1054, #1086 and #2239
Updates #1238


szuecs:
:+1:

Issue #2579, opened 2023-09-07
scheduler filter include streaming part


szuecs:
**Is your feature request related to a problem? Please describe.**

As owner of a streaming application that only support N concurrent users I want to be able to limit the traffic to N.
Right now if you do have a backend that does async response where the streaming part can be long, you get much more than N requests, because the scheduler filters response part is triggered before the response body streaming is happening.  

**Describe the solution you would like**

Make it possible to limit streaming concurrency, too.
Basically decrease the concurrency counter after body streaming was done.

**Describe alternatives you've considered (optional)**

We tried `lifo` which does not work as expected and a workaround seems to be `backendRatelimit` which has a lower error margin than lifo for an application which streams the response.

**Additional context (optional)**
internal finding

AshrafAhmed9:
Looks like this got half done. #2685 added `fifoWithBody`, which releases the slot from the proxy after the body is streamed instead of in the filter. `lifo` never got that — `lifoFilter.Response()` still calls `response(scheduler.LIFOKey, ctx)` straight away, so the reservation is gone before anything is written. Which is the case in the original report, since it mentions lifo not holding the limit.

Want me to add a `lifoWithBody` mirroring the fifo one?

One thing I'd want your call on first. The proxy hook is keyed on the filter name:

```go
// fifoWithBody() filter
if sbf, ok := ctx.StateBag()[filters.FifoWithBodyName]; ok {
```

So either I add a second lookup next to it for lifo, or turn it into one list that any scheduler filter registers into. The second is cleaner but it touches the existing fifo path, so I'd default to the first unless you'd rather not have that block copied a third time later.

Also, `lifoGroup` as well or just plain `lifo` to start?

For tests I'd hold the response body open and assert inflight stays at the configured concurrency while it streams — that should fail on lifo as it is today, so the test shows the gap before it fixes it.

szuecs:
Yeah if you implement, do both lifo and lifoGroup WithBody alongside the fifo.

PR #2685, opened 2023-10-16
feature: filter fifoWithBody

szuecs:
feature: filter fifoWithBody that works similar to fifo(), but release deferred until body streaming to client was finished

fix: https://github.com/zalando/skipper/issues/2579
fix: we did not check context canceled before, see also https://github.com/golang/go/issues/63615

### DIFF
--- a/filters/scheduler/fifo_test.go
+++ b/filters/scheduler/fifo_test.go
@@ -403,8 +403,8 @@ func TestFifo(t *testing.T) {
 			wantOkRate:  1.0,
 		},
 		{
-			name:          "fifoWithbody simple ok",
-			filter:        `fifoWithbody(3, 5, "1s")`,
+			name:          "fifoWithBody simple ok",
+			filter:        `fifoWithBody(3, 5, "1s")`,
 			freq:          20,
 			per:           100 * time.Millisecond,
 			backendTime:   1 * time.Millisecond,
@@ -421,8 +421,8 @@ func TestFifo(t *testing.T) {
 			wantOkRate:    0,
 		},
 		{
-			name:          "fifoWithbody simple client canceled",
-			filter:        `fifoWithbody(3, 5, "1s")`,
+			name:          "fifoWithBody simple client canceled",
+			filter:        `fifoWithBody(3, 5, "1s")`,
 			freq:          20,
 			per:           100 * time.Millisecond,
 			backendTime:   1 * time.Millisecond,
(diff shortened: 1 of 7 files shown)


filters/scheduler/fifo_test.go
```diff
@@ -155,14 +424,99 @@ func TestFifo(t *testing.T) {
...
+			backendTime:   1 * time.Millisecond,
+			clientTimeout: time.Second,
+			wantConfig: scheduler.Config{
+				MaxConcurrency: 3,
+				MaxQueueSize:   5,
+				Timeout:        time.Second,
+			},
+			wantParseErr: false,
+			wantOkRate:   1.0,
+			epsilon:      1,
```
AlexanderYastrebov, line 445: I think it is better to split filter creation test and behaviour tests. 
I would also propose to define filter via eskip and use eskip.MustParse in the behaviour test.

I started to do a refactoring, you can see the idea here https://github.com/zalando/skipper/compare/filters/scheduler/refactor-fifo-test

filters/scheduler/fifo_test.go
```diff
@@ -244,18 +617,20 @@ func TestFifo(t *testing.T) {
...
-				fmtStr = `aroute: * -> fifo(%v) -> "%s"`
+				fmtStr = `aroute: * -> %s(%v) -> "%s"`
 			case 2:
-				fmtStr = `aroute: * -> fifo(%v, %v) -> "%s"`
+				fmtStr = `aroute: * -> %s(%v, %v) -> "%s"`
 			case 3:
-				fmtStr = `aroute: * -> fifo(%v, %v, "%v") -> "%s"`
+				fmtStr = `aroute: * -> %s(%v, %v, "%v") -> "%s"`
 			default:
 				t.Fatalf("Test not possible %d >3", len(tt.args))
```
AlexanderYastrebov, line 628: https://github.com/zalando/skipper/pull/2685/files#r1365380057

scheduler/scheduler.go
```diff
@@ -179,6 +193,18 @@ func (fq *fifoQueue) wait(ctx context.Context) (func(), error) {
...
+		// see also: https://github.com/golang/go/issues/63615
+		// The behavior can change as in the proposed change, so the code path on error would apply.
+		// If the proposed change was merge this code path below can likely be cleaned up.
+		select {
+		case <-c.Done():
+			// We timed out in Acquire, so it never increased semaphore by one
+			cnt.Add(-1)
+			return nil, ErrQueueTimeout
+		default:
+		}
```
AlexanderYastrebov, line 207: I think we don't need this part as we checked that context is not expired before Acquire.
If we have it then we should use ctx.Err() instead of select to make it all consistent.

filters/scheduler/fifo_test.go
AlexanderYastrebov: I've created https://github.com/zalando/skipper/pull/2693

scheduler/scheduler.go
szuecs: I was thinking about the case when request times out when we block in the Acquire() call.
But I think we tested it and observed that this would return error, so yes I will cleanup

szuecs: fixed

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2681, opened 2023-10-16
drop: gosec from blockin build in CI pipeline

szuecs:
drop: gosec from blockin build in CI pipeline

PR #2689, opened 2023-10-17
fix: master build by dropping gosec from GH action

szuecs:
fix: master build by dropping gosec from GH action, because of recent gosec changes

Follow up on #2681


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2691, opened 2023-10-18
Prevent multiple items per yaml item

MustafaSaber:
Related to https://github.com/zalando/skipper/issues/1618.

### DIFF
--- a/dataclients/kubernetes/definitions/routegroupvalidator.go
+++ b/dataclients/kubernetes/definitions/routegroupvalidator.go
@@ -1,10 +1,17 @@
 package definitions
 
 import (
+	"errors"
+
 	"github.com/zalando/skipper/eskip"
 )
 
 type RouteGroupValidator struct{}
+
+var (
+	errMultipleFilters    = errors.New("multiple filters per yaml/json array item is not supported")
+	errMultiplePredicates = errors.New("multiple predicates per yaml/json array item is not supported")
+)
 
 var defaultRouteGroupValidator = &RouteGroupValidator{}
 
@@ -56,7 +63,10 @@ func (rgv *RouteGroupValidator) filtersValidation(item *RouteGroupItem) error {
 	var errs []error
 	for _, r := range item.Spec.Routes {
 		for _, f := range r.Filters {
-			_, err := eskip.ParseFilters(f)
+			filters, err := eskip.ParseFilters(f)
+			if len(filters) > 1 {
+				errs = append(errs, errMultipleFilters)
+			}
 			errs = append(errs, err)
 		}
 	}
@@ -68,7 +78,10 @@ func (rgv *RouteGroupValidator) predicatesValidation(item *RouteGroupItem) error
 	var errs []error
 	for _, r := range item.Spec.Routes {
 		for _, p := range r.Predicates {
-			_, err := eskip.ParsePredicates(p)
+			predicates, err := eskip.ParsePredicates(p)
+			if len(predicates) > 1 {
+				errs = append(errs, errMultiplePredicates)
+			}
 			errs = append(errs, err)
 		}
 	}
--- a/dataclients/kubernetes/definitions/routegroupvalidator.go
+++ b/dataclients/kubernetes/definitions/routegroupvalidator.go
@@ -10,8 +10,8 @@ import (
 type RouteGroupValidator struct{}
 
 var (
-	errMultipleFilters    = errors.New("single filter expected")
-	errMultiplePredicates = errors.New("single predicate expected")
+	errSingleFilterExpected    = errors.New("single filter expected")
+	errSinglePredicateExpected = errors.New("single predicate expected")
 )
 
 var defaultRouteGroupValidator = &RouteGroupValidator{}
@@ -66,7 +66,7 @@ func (rgv *RouteGroupValidator) filtersValidation(item *RouteGroupItem) error {
 		for _, f := range r.Filters {
 			filters, err := eskip.ParseFilters(f)
 			if len(filters) != 1 && err == nil {
-				errs = append(errs, fmt.Errorf("%w at \"%s\"", errMultipleFilters, f))
+				errs = append(errs, fmt.Errorf("%w at \"%s\"", errSingleFilterExpected, f))
 			}
 			errs = append(errs, err)
 		}
@@ -81,7 +81,7 @@ func (rgv *RouteGroupValidator) predicatesValidation(item *RouteGroupItem) error
 		for _, p := range r.Predicates {
 			predicates, err := eskip.ParsePredicates(p)
 			if len(predicates) != 1 && err == nil {
-				errs = append(errs, fmt.Errorf("%w at \"%s\"", errMultiplePredicates, p))
+				errs = append(errs, fmt.Errorf("%w at \"%s\"", errSinglePredicateExpected, p))
 			}
 			errs = append(errs, err)
 		}
--- a/dataclients/kubernetes/definitions/routegroupvalidator.go
+++ b/dataclients/kubernetes/definitions/routegroupvalidator.go
@@ -65,10 +65,11 @@ func (rgv *RouteGroupValidator) filtersValidation(item *RouteGroupItem) error {
 	for _, r := range item.Spec.Routes {
 		for _, f := range r.Filters {
 			filters, err := eskip.ParseFilters(f)
-			if len(filters) != 1 && err == nil {
+			if err != nil {
+				errs = append(errs, err)
+			} else if len(filters) != 1 {
 				errs = append(errs, fmt.Errorf("%w at \"%s\"", errSingleFilterExpected, f))
 			}
-			errs = append(errs, err)
 		}
 	}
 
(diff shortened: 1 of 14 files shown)


dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -56,7 +63,10 @@ func (rgv *RouteGroupValidator) filtersValidation(item *RouteGroupItem) error {
 	var errs []error
 	for _, r := range item.Spec.Routes {
 		for _, f := range r.Filters {
-			_, err := eskip.ParseFilters(f)
+			filters, err := eskip.ParseFilters(f)
+			if len(filters) > 1 {
+				errs = append(errs, errMultipleFilters)
+			}
```
AlexanderYastrebov, line 69: I think we can check `len(filters) == 1` and make error message like "single filter expected".
This would cover both chain and empty string case (not sure if ParseFilters fails on empty string or not). 

I also think we should not mention json/yaml in the error message as it is ambiguous and also routegroup could be created programmatically.

dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -56,7 +63,10 @@ func (rgv *RouteGroupValidator) filtersValidation(item *RouteGroupItem) error {
 	var errs []error
 	for _, r := range item.Spec.Routes {
 		for _, f := range r.Filters {
-			_, err := eskip.ParseFilters(f)
+			filters, err := eskip.ParseFilters(f)
+			if len(filters) > 1 {
```
AlexanderYastrebov, line 67: I think we should check number of filters only when err is nil.

MustafaSaber: Why? I think we should send both errors to decrease roundtrips of fixes

dataclients/kubernetes/definitions/routegroupvalidator.go
MustafaSaber: for empty string we already validate with https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/definitions/routegroupvalidator.go#L134-L140

MustafaSaber: Since we check empty first `> 1` makes more sense because if the filter was invalid, `ParseFilters` will return empty filter which will then push 2 errors (parsing & single)

dataclients/kubernetes/definitions/routegroupvalidator.go
AlexanderYastrebov: Lets add a test case for empty value, maybe eskip.Parse already handles it

MustafaSaber: there is one already

* https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/definitions/testdata/validation/route-with-empty-filter.yaml
* https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/definitions/testdata/validation/route-with-empty-predicate.yaml

MustafaSaber: without validation `ParseFilter/predicate` return no error for empty string

dataclients/kubernetes/definitions/routegroupvalidator.go
AlexanderYastrebov: Then maybe we should move empty check here to have filter validation in one place.

Note that value might also contain only whitespaces `"  "` which would produce zero filters but will not be empty - currently we do not catch this case. Here parse would correctly handle empty value like that

dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -1,11 +1,19 @@
...
+	"fmt"
+
 	"github.com/zalando/skipper/eskip"
 )
 
 type RouteGroupValidator struct{}
 
+var (
+	errMultipleFilters    = errors.New("single filter expected")
+	errMultiplePredicates = errors.New("single predicate expected")
```
AlexanderYastrebov, line 14: Since we use this for both no filter and multiple filters I think it should be named like `errSingleFilterExpected`

dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -56,7 +64,10 @@ func (rgv *RouteGroupValidator) filtersValidation(item *RouteGroupItem) error {
 	var errs []error
 	for _, r := range item.Spec.Routes {
 		for _, f := range r.Filters {
-			_, err := eskip.ParseFilters(f)
+			filters, err := eskip.ParseFilters(f)
+			if len(filters) != 1 && err == nil {
```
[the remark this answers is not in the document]
MustafaSaber: but I want both, I don't want one or either. If we have 2 filters and one of them failed parsing then I want 2 errors to appear in validation one for parsing and one for number of filters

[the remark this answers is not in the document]
AlexanderYastrebov: >  If we have 2 filters and one of them failed parsing

In such case I think eskip.ParseFilters(f) will just return an error (you can check).
It is an established go pattern to check error first and if it is not nil ignore the result (which is often empty or malformed so cant be relied on anyways).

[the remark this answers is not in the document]
RomanZavodskikh: > In such case I think eskip.ParseFilters(f) will just return an error (you can check).

No, it won't.

```
rzavodskikh@ZALANDO-39125:/tmp/tmp$ cat main.go 
package main

import (
	"fmt"

)

func main() {
	filters, err := eskip.ParseFilters("status(201) -> inlineContent(\"hi\")")
	fmt.Printf("filters: %+v, err: %+v\n", filters, err)
}
rzavodskikh@ZALANDO-39125:/tmp/tmp$ go build main.go
rzavodskikh@ZALANDO-39125:/tmp/tmp$ ./main 
filters: [status(201) inlineContent("hi")], err: <nil>
rzavodskikh@ZALANDO-39125:/tmp/tmp$ 

```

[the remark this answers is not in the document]
RomanZavodskikh: > In such case I think eskip.ParseFilters(f) will just return an error (you can check).
Yes, it will. Moreover, filters variable would be empty in this case, even if first filter is correct. So, I do agree it's better not to return two error in this case, one would be enough.

```
rzavodskikh@ZALANDO-39125:/tmp/tmp$ cat main.go 
package main

import (
	"fmt"

)

func main() {
	filters, err := eskip.ParseFilters("status(201) -> inlineContent(\"hi\"")
	fmt.Printf("filters: %+v, err: %+v\n", filters, err)
}
rzavodskikh@ZALANDO-39125:/tmp/tmp$ go build main.go
rzavodskikh@ZALANDO-39125:/tmp/tmp$ ./main 
filters: [], err: parse failed after token ->, last route id: , position 41: syntax error
rzavodskikh@ZALANDO-39125:/tmp/tmp$ 
```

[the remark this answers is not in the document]
MustafaSaber: I don't get the problem between the comment and what I wrote.
`ParseFilters` either return an error or list of filters

In case of list of filters I want to make sure there is only one filter, and in case of error I want to return the error with other errors. I don't want to only return this specific error because it will cause more roundtrips of fixes.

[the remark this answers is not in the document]
RomanZavodskikh: Discussed this part of code on a call, I showed one more time that ParseFilters may return:
+ non-empty filters slice XOR
+ not nil error
but it never returns them at the same call.

So, from my pov it does not make sense try to put two errors at the `errs` slice here, so the code of Alexander at the beginning from this discussion makes the most sense to me here.

RomanZavodskikh:
:+1:

MustafaSaber:
👍

dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -68,8 +80,12 @@ func (rgv *RouteGroupValidator) predicatesValidation(item *RouteGroupItem) error
 	var errs []error
 	for _, r := range item.Spec.Routes {
 		for _, p := range r.Predicates {
-			_, err := eskip.ParsePredicates(p)
-			errs = append(errs, err)
+			predicates, err := eskip.ParsePredicates(p)
+			if err != nil {
+				errs = append(errs, err)
+			} else if len(predicates) != 1 {
+				errs = append(errs, fmt.Errorf("%w at \"%s\"", errSinglePredicateExpected, p))
```
szuecs, line 87: %s -> %q

dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -56,8 +64,12 @@ func (rgv *RouteGroupValidator) filtersValidation(item *RouteGroupItem) error {
 	var errs []error
 	for _, r := range item.Spec.Routes {
 		for _, f := range r.Filters {
-			_, err := eskip.ParseFilters(f)
-			errs = append(errs, err)
+			filters, err := eskip.ParseFilters(f)
+			if err != nil {
+				errs = append(errs, err)
+			} else if len(filters) != 1 {
+				errs = append(errs, fmt.Errorf("%w at \"%s\"", errSingleFilterExpected, f))
```
szuecs, line 71: %s -> %q

dataclients/kubernetes/definitions/routegroupvalidator.go
MustafaSaber: https://github.com/zalando/skipper/pull/2700

dataclients/kubernetes/definitions/routegroupvalidator.go
MustafaSaber: https://github.com/zalando/skipper/pull/2700

PR #2706, opened 2023-10-30
feature: add "version" label to all prometheus metrics

szuecs:
feature: add "version" label to all prometheus metrics


szuecs:
:+1:

MustafaSaber:
👍

metrics/prometheus.go
```diff
@@ -54,6 +56,9 @@ type Prometheus struct {
 
 // NewPrometheus returns a new Prometheus metric backend.
 func NewPrometheus(opts Options) *Prometheus {
+	if version == "" {
+		version = "<unknown>"
+	}
```
AlexanderYastrebov, line 61: Why not pass version in the Options and set it via already available main.version?

szuecs: This discussion we already had.
The reason is that with how we do this now in the linker is the best option to not trick ourselves.
It's simple and works at build time.

AlexanderYastrebov: I struggle to see how setting an obscure global variable via linker (which also must have the same value as an existing variable) is better then passing it via Options. Library users must update their Makefiles

AlexanderYastrebov: It should be also possible to add version label from pod labels during scarping

AlexanderYastrebov: > It should be also possible to add version label from pod labels during scarping

See https://github.com/zalando-incubator/kubernetes-on-aws/pull/6579

szuecs: The question is what is more ugly.
IMO linker variable is not really obscure, but the labeling of an independently configured metrics store actually is. :)

Issue #2684, opened 2023-10-16
endpointslices Serving vs Ready condition possible bug 


szuecs:
After a discussion with a colleague @dtrsan , I think we have a possible bug that could be triggered in some backend pod error condition.

Particular the definition of isReady() in
https://github.com/zalando/skipper/blob/30eada3fb855b8df19bccadaaf7926970df1376e/dataclients/kubernetes/endpointslices.go#L151-L157

Maybe there are these cases that should return isReady=false instead of true right now:
1) 
Serving: true
Ready: false
Terminating: nil
2)
Serving: nil
Ready: false
Terminating: nil

Maybe these both conditions would happen in case a pod is oom killed, but maybe these also never happen in Kubernetes. It seems for me that these two are examples could be a valid change.

AlexanderYastrebov:
I.e. you propose to change from `(ready || serving)` to `(ready && serving)`

szuecs:
I would propose to drop the serving condition, because for us it makes no issue if we would drop an endpoint if it's serving, but not ready. If a pod is `ready=true` in theory at start it could be `serving=false`. 
So right now I don't see the issue of dropping serving in the condition. I think `serving=true` is set for networkpolicy providers such that they do not stop packets until `serving=false`, but that's only an educated guess from my participation in sig-network.

szuecs:
this was done by https://github.com/zalando/skipper/pull/2709

PR #2709, opened 2023-10-31
fix: endpointslices should not consider Serving condition

szuecs:
see also https://github.com/zalando/skipper/issues/2684

added: test case for `ready: false` and `nil` cases


MustafaSaber:
👍

RomanZavodskikh:
:+1:

szuecs:
:+1:

PR #2714, opened 2023-11-03
Feature: tracing set route id to ingress spans

szuecs:
feature: add routeID to traces for shunt and loopback routes

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1087,7 +1087,13 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 		ctx.ensureDefaultResponse()
 	} else if ctx.route.BackendType == eskip.LoopBackend {
 		loopCTX := ctx.clone()
-		loopSpan := loopCTX.Tracer().StartSpan(p.tracing.initialOperationName, ot.ChildOf(ctx.ParentSpan().Context()))
+		loopSpan := tracing.CreateSpan("loopback", ctx.request.Context(), p.tracing.tracer)
+		p.tracing.
+			setTag(loopSpan, SpanKindTag, SpanKindServer).
+			setTag(loopSpan, SkipperRouteIDTag, ctx.route.Id)
+		p.setCommonSpanInfo(ctx.Request().URL, ctx.Request(), loopSpan)
+
+		defer loopSpan.Finish()
 
 		if err := p.do(loopCTX, loopSpan); err != nil {
 			// in case of error we have to copy the response in this recursion unwinding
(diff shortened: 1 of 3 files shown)


proxy/proxy.go
```diff
@@ -1024,7 +1024,7 @@ func stack() []byte {
 	}
 }
 
-func (p *Proxy) do(ctx *context) (err error) {
+func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
```
RomanZavodskikh, line 1027: Can we have `parentSpan` as a field of `context` structure?

szuecs: Why do you think it's better than a local?

RomanZavodskikh: There are a lot of arguments in `proxy.context` struct[1]. It seems this structure was created to avoid overloading of `*Proxy.go` signature:
```go
func (p *Proxy) do(responseWriter flushedResponseWriter, request *http.Request, response *http.Response, <~10-20 more arguments>...)
```
The change proposed in the current PR introduces the new argument to ~20 arguments of `*Proxy.go` function, so the most straightforward way I see is to add `parentSpan` to the `type context struct`, where all other arguments are located in at the moment, this was we will maximize the readability of `*Proxy.go` signature.

[1] https://github.com/zalando/skipper/blob/v0.18.42/proxy/context.go#L29

RomanZavodskikh: However, if `parentSpan` is somewhat special in comparison of `type context struct` fields, then the signature you propose at the moment may be the best decision.

szuecs: Maybe I can explain why I chose to use a parameter instead of changing proxy.context member:
1. parentSpan changes, but most of context members are assigned once and never change or are there to track changes and chose to use explicitly a copy of unmodified data (`original*` or `initialSpan` for example).
2. if we add parentSpan as member we likely need a mutex to guard access because there is also a bunch of filters that work on this data. The flakyness of testing it will also increase I would guess. I think parentSpan is a good parameter of this func instead of being a member.

RomanZavodskikh: If I get it right, here[1] you clone the `context`, but the only thing you change after this clone is `loopSpan`. And that's the particular reason you would like `loopSpan` to be the separate argument. Okay, it makes some sense to me. Did I get it right?

[1] https://github.com/zalando/skipper/pull/2714/files#diff-1d74870266949f1927a500e5f0002617239d3e3cd9bcceee6be014738e5e0f2bR1090

RomanZavodskikh:
:+1:

RomanZavodskikh:
:+1:

szuecs:
:+1:

MustafaSaber:
Do we need to update something in the docs here https://opensource.zalando.com/skipper/operation/operation/#ingress-span?

PR #2719, opened 2023-11-06
Validate backend path

MustafaSaber:
Make sure that network backends doesn't include paths.

### DIFF
--- a/cmd/webhook/admission/admission_test.go
+++ b/cmd/webhook/admission/admission_test.go
@@ -133,6 +133,11 @@ func TestRouteGroupAdmitter(t *testing.T) {
 			inputFile: "rg-with-multiple-predicates.json",
 			message:   `single predicate expected at \"Method(\\\"GET\\\") && Path(\\\"/\\\")\"\nsingle predicate expected at \" \"`,
 		},
+		{
+			name:      "routegroup with invalid backends",
+			inputFile: "rg-with-invalid-backend-path.json",
+			message:   `backend address \"http://example.com/foo\" contains path\nbackend address \"http://example.com/foo/bar\" contains path\nbackend address \"http://example.com/foo/\" contains path\nbackend address \"/foo\" contains path\nbackend address \"http://example.com/\" contains path\nbackend address \"example.com/\" contains path\nbackend address \"example.com/foo\" contains path`,
+		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
 			expectedResponse := responseAllowedFmt
--- a/cmd/webhook/admission/admission_test.go
+++ b/cmd/webhook/admission/admission_test.go
@@ -136,7 +136,7 @@ func TestRouteGroupAdmitter(t *testing.T) {
 		{
 			name:      "routegroup with invalid backends",
 			inputFile: "rg-with-invalid-backend-path.json",
-			message:   `backend address \"http://example.com/foo\" contains path\nbackend address \"http://example.com/foo/bar\" contains path\nbackend address \"http://example.com/foo/\" contains path\nbackend address \"/foo\" contains path\nbackend address \"http://example.com/\" contains path\nbackend address \"example.com/\" contains path\nbackend address \"example.com/foo\" contains path`,
+			message:   `backend address \"http://example.com/foo\" contains path\nbackend address \"http://example.com/foo/bar\" contains path\nbackend address \"http://example.com/foo/\" contains path\nbackend address \"/foo\" contains path\nbackend address \"http://example.com/\" contains path\nbackend address \"example.com/\" contains path\nbackend address \"example.com/foo\" contains path\nbackend address \"http://example.com?foo=bar\" contains path`,
 		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
--- a/cmd/webhook/admission/admission_test.go
+++ b/cmd/webhook/admission/admission_test.go
@@ -136,7 +136,7 @@ func TestRouteGroupAdmitter(t *testing.T) {
 		{
 			name:      "routegroup with invalid backends",
 			inputFile: "rg-with-invalid-backend-path.json",
-			message:   `backend address \"http://example.com/foo\" contains path or query\nbackend address \"http://example.com/foo/bar\" contains path or query\nbackend address \"http://example.com/foo/\" contains path or query\nbackend address \"/foo\" contains path or query\nbackend address \"http://example.com/\" contains path or query\nbackend address \"example.com/\" contains path or query\nbackend address \"example.com/foo\" contains path or query\nbackend address \"http://example.com?foo=bar\" contains path or query`,
+			message:   `backend address \"http://example.com/foo\" contains path, query or missing scheme\nbackend address \"http://example.com/foo/bar\" contains path, query or missing scheme\nbackend address \"http://example.com/foo/\" contains path, query or missing scheme\nbackend address \"/foo\" contains path, query or missing scheme\nbackend address \"http://example.com/\" contains path, query or missing scheme\nbackend address \"example.com/\" contains path, query or missing scheme\nbackend address \"example.com/foo\" contains path, query or missing scheme\nbackend address \"http://example.com?foo=bar\" contains path, query or missing scheme\nbackend address \"example.com\" contains path, query or missing scheme`,
 		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
--- a/cmd/webhook/admission/admission_test.go
+++ b/cmd/webhook/admission/admission_test.go
@@ -136,7 +136,7 @@ func TestRouteGroupAdmitter(t *testing.T) {
 		{
 			name:      "routegroup with invalid backends",
 			inputFile: "rg-with-invalid-backend-path.json",
-			message:   `backend address \"http://example.com/foo\" contains path\nbackend address \"http://example.com/foo/bar\" contains path\nbackend address \"http://example.com/foo/\" contains path\nbackend address \"/foo\" contains path\nbackend address \"http://example.com/\" contains path\nbackend address \"example.com/\" contains path\nbackend address \"example.com/foo\" contains path\nbackend address \"http://example.com?foo=bar\" contains path`,
+			message:   `backend address \"http://example.com/foo\" contains path or query\nbackend address \"http://example.com/foo/bar\" contains path or query\nbackend address \"http://example.com/foo/\" contains path or query\nbackend address \"/foo\" contains path or query\nbackend address \"http://example.com/\" contains path or query\nbackend address \"example.com/\" contains path or query\nbackend address \"example.com/foo\" contains path or query\nbackend address \"http://example.com?foo=bar\" contains path or query`,
 		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
--- a/cmd/webhook/admission/testdata/rg/rg-with-invalid-backend-path.json
+++ b/cmd/webhook/admission/testdata/rg/rg-with-invalid-backend-path.json
@@ -21,7 +21,7 @@
             "type": "shunt"
           },
           {
-            "name": "invalid-backend1",
+            "name": "valid-backend2",
             "type": "network",
             "address": "http://example.com"
           },
--- a/cmd/webhook/admission/testdata/rg/rg-with-invalid-backend-path.json
+++ b/cmd/webhook/admission/testdata/rg/rg-with-invalid-backend-path.json
@@ -64,6 +64,11 @@
             "name" : "invalid-backend8",
             "type" : "network",
             "address" : "example.com/foo"
+          },
+          {
+            "name" : "invalid-backend9",
+            "type" : "network",
+            "address" : "http://example.com?foo=bar"
           }
         ],
         "defaultBackends": [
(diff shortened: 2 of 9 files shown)


dataclients/kubernetes/definitions/testdata/validation/route-with-invalid-backend-with-path.log
```diff
@@ -0,0 +1 @@
+backend address \"http://example.com/foo\" contains path\nbackend address \"http://example.com/foo/bar\" contains path\nbackend address \"/foo\" contains path\nbackend address \"/foo/bar\" contains path
```
MustafaSaber, line 1: Tests fails because regexp doesn't work well with escape character "\" 

example https://go.dev/play/p/mbhk8lJI8xa

dataclients/kubernetes/definitions/testdata/errorwrapdata/errors.log
```diff
@@ -4,3 +4,7 @@ single predicate expected at "Path(\"/foo\") && Method(\"GET\")"
 single predicate expected at ""
 single filter expected at "inlineContent(\"/foo\") -> status(200)"
 single filter expected at " "
+backend address "http://example.com/foo" contains path
+backend address "http://example.com/foo/bar" contains path
+backend address "http://example.com/foo/" contains path
```
RomanZavodskikh, line 9: Should we also add the following failures?
```
http://example.com
http://example.com/
https://example.com/foo
example.com/
example.com/foo
```

dataclients/kubernetes/definitions/testdata/validation/route-with-invalid-backend-with-path.log
RomanZavodskikh, line 1: minor: add newline here for :stop_sign: at the end of this file to disappear.

dataclients/kubernetes/definitions/testdata/validation/route-with-multiple-filters-one-item.log
```diff
@@ -1 +1 @@
-single filter expected
+single filter expected at "foo(42) -> bar(24)"
```
RomanZavodskikh, line 1: minor: add newline here for :stop_sign: at the end of this file to disappear.

RomanZavodskikh, line 1: I don't quite understand, what Skipper user should do when he/she wants to have 2+ filters for single route? It seems to me it's valid use case.

dataclients/kubernetes/definitions/testdata/errorwrapdata/errors.log
MustafaSaber: why `http://example.com` is a failure?

I think `https://example.com/foo` is already covered by `http://example.com/foo`

Will add some of them later, I still didn't make tests work.

dataclients/kubernetes/definitions/testdata/validation/route-with-invalid-backend-with-path.log
MustafaSaber: It's like this because I still didn't make tests work.

dataclients/kubernetes/definitions/testdata/validation/route-with-multiple-filters-one-item.log
MustafaSaber: It's like this because I still didn't make tests work.

dataclients/kubernetes/definitions/testdata/errorwrapdata/errors.log
RomanZavodskikh: > why http://example.com is a failure?

I just was thinking that explicit https configuration is not allowed as well.

dataclients/kubernetes/definitions/testdata/validation/route-with-multiple-filters-one-item.log
MustafaSaber: This is 2 filters in one yaml item
```yaml
filters:
  - foo(42) -> bar(24)
```

We don't want this, and also this is a test case from another [PR](https://github.com/zalando/skipper/pull/2691) that was already merged.

We want people to use

```yaml
filters:
  - foo(42)
  - bar(24)
```

dataclients/kubernetes/definitions/testdata/errorwrapdata/errors.log
MustafaSaber: All we care about in this PR if the address has path or not. Maybe we need to add validation later for https but I don't see why it's not allowed, maybe we talk to east-west route without certificate.

dataclients/kubernetes/definitions/testdata/validation/route-with-invalid-backend-with-path.log
MustafaSaber: After some tests made it work https://go.dev/play/p/yZUl2NCiP_n.

dataclients/kubernetes/definitions/testdata/errorwrapdata/errors.log
szuecs: What should work is: `http://user:pass@example.com`

dataclients/kubernetes/definitions/testdata/validation/route-with-multiple-filters-one-item.log
```diff
@@ -1 +1 @@
-single filter expected
+single filter expected at \\\"foo\(42\) -> bar\(24\)\\\"
```
MustafaSaber, line 1: Not related to the change but makes this test case more robust.

dataclients/kubernetes/definitions/testdata/validation/route-with-multiple-predicates-one-item.log
```diff
@@ -1 +1 @@
-single predicate expected
+single predicate expected at \\\"Foo\(\\\\\\"X-Bar\\\\\\", \\\\\\"42\\\\\\"\) && Bar\(\\\\\\"X-Foo\\\\\\", \\\\\\"24\\\\\\"\)\\\"
```
MustafaSaber, line 1: Not related to the change but makes this test case more robust.

cmd/webhook/admission/testdata/rg/rg-with-invalid-backend-path.json
```diff
@@ -0,0 +1,93 @@
...
+      "spec": {
+        "backends": [
+          {
+            "name": "valid-backend1",
+            "type": "shunt"
+          },
+          {
+            "name": "invalid-backend1",
+            "type": "network",
+            "address": "http://example.com"
```
szuecs, line 26: what is here invalid?

dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -91,6 +93,19 @@ func (rgv *RouteGroupValidator) predicatesValidation(item *RouteGroupItem) error
 	return errorsJoin(errs...)
 }
 
+func (rgv *RouteGroupValidator) validateBackends(item *RouteGroupItem) error {
+	var errs []error
+	for _, backend := range item.Spec.Backends {
+		address, err := url.Parse(backend.Address)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to parse backend address %q: %w", backend.Address, err))
+		} else if address.Path != "" {
```
szuecs, line 102: Maybe also check for query?

szuecs:
:+1:

dataclients/kubernetes/definitions/routegroupvalidator.go
MustafaSaber: I was checking if we can use queryparam without at least one path so `example.com?foo=bar` vs `example.com/?foo=bar`. With url.Parse https://go.dev/play/p/iRU9__MGc8S you can achieve both so it makes sense to also detect this case.

dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -99,7 +99,7 @@ func (rgv *RouteGroupValidator) validateBackends(item *RouteGroupItem) error {
 		address, err := url.Parse(backend.Address)
 		if err != nil {
 			errs = append(errs, fmt.Errorf("failed to parse backend address %q: %w", backend.Address, err))
-		} else if address.Path != "" {
+		} else if address.Path != "" || address.RawQuery != "" {
 			errs = append(errs, fmt.Errorf("backend address %q contains path", backend.Address))
```
szuecs, line 103: `path` -> `path or query`

szuecs:
:+1:

cmd/webhook/admission/admission_test.go
```diff
@@ -133,6 +133,11 @@ func TestRouteGroupAdmitter(t *testing.T) {
 			inputFile: "rg-with-multiple-predicates.json",
 			message:   `single predicate expected at \"Method(\\\"GET\\\") && Path(\\\"/\\\")\"\nsingle predicate expected at \" \"`,
 		},
+		{
+			name:      "routegroup with invalid backends",
+			inputFile: "rg-with-invalid-backend-path.json",
+			message:   `backend address \"http://example.com/foo\" contains path or query\nbackend address \"http://example.com/foo/bar\" contains path or query\nbackend address \"http://example.com/foo/\" contains path or query\nbackend address \"/foo\" contains path or query\nbackend address \"http://example.com/\" contains path or query\nbackend address \"example.com/\" contains path or query\nbackend address \"example.com/foo\" contains path or query\nbackend address \"http://example.com?foo=bar\" contains path or query`,
+		},
```
AlexanderYastrebov, line 140: Do we need an essentially copy-pasted test case for the webhook given that it relies on the dataclients Validator that has its own test?

dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -37,6 +38,7 @@ func (rgv *RouteGroupValidator) Validate(item *RouteGroupItem) error {
 	var errs []error
 	errs = append(errs, rgv.filtersValidation(item))
 	errs = append(errs, rgv.predicatesValidation(item))
+	errs = append(errs, rgv.validateBackends(item))
```
AlexanderYastrebov, line 41: Note how naming is not consistent here. IMO filtersValidation shoudl be validateFilters and predicatesValidation - validatePredicates.

dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -91,6 +93,19 @@ func (rgv *RouteGroupValidator) predicatesValidation(item *RouteGroupItem) error
...
 
+func (rgv *RouteGroupValidator) validateBackends(item *RouteGroupItem) error {
+	var errs []error
+	for _, backend := range item.Spec.Backends {
+		address, err := url.Parse(backend.Address)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to parse backend address %q: %w", backend.Address, err))
+		} else if address.Path != "" || address.RawQuery != "" {
+			errs = append(errs, fmt.Errorf("backend address %q contains path or query", backend.Address))
+		}
```
AlexanderYastrebov, line 104: Should we also check that schema is specified?

MustafaSaber: I don't know if we should do, maybe if we have schema other than (http|https)?  https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/hosts.go#L76-L83

cmd/webhook/admission/admission_test.go
MustafaSaber: If both validators have the same options then no, but maybe webhook has more strict way of validation later. Also this tests webhook response, so the error maybe the same but the response is different.

dataclients/kubernetes/definitions/routegroupvalidator.go
AlexanderYastrebov: Skipper also supports `fastcgi://`. I think checking that its not empty is enough for now, i.e. do not allow `example.com` (should be `https://example.com`)

cmd/webhook/admission/testdata/rg/rg-with-invalid-backend-path.json
```diff
@@ -0,0 +1,98 @@
...
+          },
+          {
+            "name": "invalid-backend5",
+            "type": "network",
+            "address": "/foo"
+          },
+          {
+            "name": "valid-backend2",
+            "type": "network",
+            "address": "http://user:pass@example.com"
```
RomanZavodskikh, line 51: Does "http://user@example.com" make sense as valid address? If it makes sense, should we add it as an example of valid backend as well?

MustafaSaber: we already covered `http://user:pass@example.com` which implies that `http://user@example.com` is a valid one I don't think we need to add extra test cases there. It won't hurt but since it's already passes [`url.Parse`](https://go.dev/play/p/lflJeQc4nYx) and there is no explicit condition to make it fail then no need to cover.

RomanZavodskikh:
:+1:

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2724, opened 2023-11-08
fix: master build was broken because test was not removed

szuecs:
fix: master build was broken since 15513a52119cda20e515a27d37d38bd57bd8dee4 because test was not removed


MustafaSaber:
👍

szuecs:
:+1:

PR #2730, opened 2023-11-10
update: gh action as pinned dependency as OpenSSF scorecard suggests

szuecs:
update: gh action as pinned dependency as OpenSSF scorecard suggests


szuecs:
:+1:

RomanZavodskikh:
:+1:

PR #2736, opened 2023-11-14
build: use CDP runtime to build docs

AlexanderYastrebov:
Use CDP runtime instead of deprecated overlay to build doc site.

Tested locally via:
```
docker run -it -p 8000:8000 --rm -v $PWD:/workspace pierone.stups.zalan.do/cdp-runtime/mkdocs-v9 sh -c 'pipx inject mkdocs-material markdown-include && serve-docs'
```

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -71,7 +71,6 @@ markdown_extensions:
     generic: true
 
 extra_javascript:
-  - javascripts/mathjax.js
   - https://polyfill.io/v3/polyfill.min.js?features=es6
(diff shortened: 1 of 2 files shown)


mkdocs.yml
```diff
@@ -71,7 +71,6 @@ markdown_extensions:
     generic: true
 
 extra_javascript:
-  - javascripts/mathjax.js
```
AlexanderYastrebov, line 74: There is no mathjax.js but it is delivered from CDN link below

MustafaSaber:
👍

RomanZavodskikh:
:+1:

szuecs:
:+1:

PR #2738, opened 2023-11-15
routesrv: accept gzip encoding

AlexanderYastrebov:
Support gzip encoding to reduce response size.

### DIFF
--- a/routesrv/eskipbytes.go
+++ b/routesrv/eskipbytes.go
@@ -2,10 +2,12 @@ package routesrv
 
 import (
 	"bytes"
+	"compress/gzip"
 	"crypto/sha256"
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -50,12 +52,15 @@ var (
 // provides synchronized r/w access to them. Additionally it can
 // serve as an HTTP handler exposing its content.
 type eskipBytes struct {
+	mu           sync.RWMutex
 	data         []byte
-	etag         string
+	hash         string
 	lastModified time.Time
 	initialized  bool
 	count        int
-	mu           sync.RWMutex
+
+	zw    *gzip.Writer
+	zdata []byte
 
 	tracer  ot.Tracer
 	metrics metrics.Metrics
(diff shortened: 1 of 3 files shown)


szuecs:
please create the PR on top of cw2023freeze

routesrv/eskipbytes.go
```diff
@@ -50,12 +52,15 @@ var (
 // provides synchronized r/w access to them. Additionally it can
 // serve as an HTTP handler exposing its content.
 type eskipBytes struct {
+	mu           sync.RWMutex
 	data         []byte
-	etag         string
```
RomanZavodskikh, line 54: Why we don't need etag more?

MustafaSaber: I think it was just renamed to `hash` but I think `etag` is more expressive, on another note if we will rename here it makes sense to rename in https://github.com/zalando/skipper/blob/master/eskipfile/remote.go#L31

routesrv/eskipbytes.go
```diff
@@ -112,17 +137,24 @@ func (e *eskipBytes) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
-		w.Header().Add("Etag", etag)
-		w.Header().Add("Content-Type", "text/plain; charset=utf-8")
-		w.Header().Add(routing.RoutesCountName, strconv.Itoa(count))
-
-		http.ServeContent(w, r, "", lastModified, bytes.NewReader(data))
+		w.Header().Set("Content-Type", "text/plain; charset=utf-8")
+		w.Header().Set(routing.RoutesCountName, strconv.Itoa(count))
+
+		if strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") && len(zdata) > 0 {
+			w.Header().Set("Etag", `"`+hash+`+gzip"`)
```
MustafaSaber, line 151: maybe use fmt.Sprintf?

routesrv/eskipbytes.go
AlexanderYastrebov: gzip-encoded content must have different Etag from plain text content. I derive gzip Etag by adding `+gzip` suffix (Apache's mod_deflate adds `-gzip` but I saw `+gzip` from `curl -v --compressed example.com` and I like it more with plus sign).
Since Etag value must be quoted I store "hash" and quote the value before setting Etag header therefore I renamed field member from etag to hash. I.e. `hash == "cafebabe"` translates into `Etag: "cafebabe"` or `Etag: "cafebabe+gzip"` header values depending on the Accept-Encoding.

routesrv/eskipbytes.go
AlexanderYastrebov: yeah, I though about it but it scatches me to use printf to add quotes

AlexanderYastrebov:
```go
w.Header().Set("Etag", `"`+hash+`"`)
w.Header().Set("Etag", `"`+hash+`+gzip"`)
// vs
w.Header().Set("Etag", fmt.Sprintf(`"%s"`, hash))
w.Header().Set("Etag", fmt.Sprintf(`"%s+gzip"`, hash))

```

routesrv/eskipbytes.go
AlexanderYastrebov: > if we will rename here it makes sense to rename in

Not, there is it a Etag value (quoted hash)

routesrv/eskipbytes.go
MustafaSaber: I don't know if `%q` makes it better
```
fmt.Sprintf("%q", hash)
fmt.Sprintf("%q", hash+"+gzip")
```

szuecs: maybe: `fmt.Sprintf("%qgzip", hash)`

szuecs: ah no this does not work https://go.dev/play/p/xy_341u63TE.
I think @AlexanderYastrebov already tried the best one, in case he wants to have a string like:
`"...gzip"` and not `"..."gzip`

szuecs:
:+1:

MustafaSaber:
👍

Issue #1952, opened 2022-02-08
Add support to refer K8S secrets with skipper filter 


tamer-abdulghani:
**Backgroud**
We are trying to deploy ingress objects that uses Skipper. Those ingresses have Microsoft OAuth configuration which need a service principle (clientId/clientSecret). Currently we are storing these secrets in GitHub secrets and override the ingress objects during helm deployment to K8S. 
To avoid storing these client-id/client-secret in GitHub, we are thinking to store them directly in K8S secrets. However, then we must need to have a way to refer those secrets in ingresses objects which is currently not available in Skipper afaik.

**Proposal**
Add support to refer K8S secrets in ingresses objects that uses skipper ingress class.

**My current ingress object looks like:**
as you see, I'm overriding `client-id/client-secret` during helm deployment. But it would be much better and more secure to refer them somehow to K8S secrets. 

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: "{{ .Chart.Name }}-s-ingress"
  annotations:
    kubernetes.io/ingress.class: skipper
    zalando.org/skipper-filter: | 
      oauthOidcUserInfo(https://login.microsoftonline.com/TENAND_ID/v2.0, 
          "{{ .Values.ingress.clientId }}", 
          "{{ .Values.ingress.clientSecret }}", 
          https://{{ .Values.ingress.host }}/oauth/callback, 
          "email profile", 
          "",
          "", 
          "X-Remote-User:claims.email",
          "0"
          )
          -> oidcClaimsQuery("/:{{- include "groupFilter"  . -}}",
            "/:@_:email%\[*@mydomain.com\](mailto:*@mydomain.com\)")
          -> dropRequestHeader("Cookie")
  labels:
    app: "{{ .Release.Name }}"
spec:
  rules:
    - host: "{{ .Values.ingress.host }}"
      http:
        paths:
        - path: "/{{  .Values.ingress.path }}/*"
          pathType: ImplementationSpecific
          backend:
            service:
              name: "{{ .Chart.Name }}-svc"
              port:
                number: 8080
```

szuecs:
@tamer-abdulghani , I would not do it via Kubernetes secrets API, but use the `secrets` module that works with the filesystem. Secrets can be mounted to the pod, so you can configure skipper to update these in memory secrets. If you want to propose a filter change it would be great to see.

AlexanderYastrebov:
Some thoughts for the context.

[...]

PR #2740, opened 2023-11-15
filters/auth: add setRequestHeaderFromSecret filter

AlexanderYastrebov:
Add a filter to set request header value from secret with optional prefix and suffix.

It is similar to `bearerinjector` which is equivalent to `setRequestHeaderFromSecret("Authorization", "/tokens/my-token", "Bearer ")`

For #1952

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -2957,31 +2957,54 @@ the -rfc-patch-path flag. See
 [URI standards interpretation](../operation/operation.md#uri-standards-interpretation).
 
 ## Egress
-### bearerinjector
 
-This filter injects `Bearer` tokens into `Authorization` headers read
-from file providing the token as content. This is only for use cases
-using skipper as sidecar to inject tokens for the application on the
+### setRequestHeaderFromSecret
+
+This filter sets request header to the secret value with optional prefix and suffix.
+This is only for use cases using skipper as sidecar to inject tokens for the application on the
 [**egress**](egress.md) path, if it's used in the **ingress** path you likely
 create a security issue for your application.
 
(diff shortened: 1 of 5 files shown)


filters/auth/secretheader_test.go
```diff
@@ -0,0 +1,102 @@
+package auth_test
```
AlexanderYastrebov, line 1: Tests cover 100% of the code

docs/reference/filters.md
```diff
@@ -2957,31 +2957,54 @@ the -rfc-patch-path flag. See
...
-`/tmp/secrets/write-token` and `/tmp/secrets/read-token`.
+`/tmp/secrets/get-token`, `/tmp/secrets/post-token` and `/tmp/secrets/put-token`.
+
+### bearerinjector
+
+This filter injects `Bearer` tokens into `Authorization` headers read
+from file providing the token as content.
+
+It is a special form of `setRequestHeaderFromSecret` with `"Authorization"` header name,
+`"Bearer "` prefix and empty suffix.
```
szuecs, line 2999: If so we could replace the implementation of bearerinjector by a configuration of setRequestHeaderFromSecret, so less code and better tested.
Wdyt?

szuecs:
:+1:

docs/reference/filters.md
AlexanderYastrebov: I thought about it but then we would need to limit number of arguments as bearerinjector only requires one so I decided ti does not worth it.

szuecs: ok

MustafaSaber:
👍

Issue #2282, opened 2023-04-24
Filters: logBody("request", "response")


bocytko:
**Is your feature request related to a problem? Please describe.**
I'd like to use skipper as proxy for APIs and log the request & response pairs, so that I can later mock the APIs.

**Describe the solution you would like**
Skipper already provides a filter [logHeader](https://opensource.zalando.com/skipper/reference/filters/#logheader) to log request headers. Similarly, there could be a filter provided `logBody` that logs the request body/response (e.g. up to a certain size) (or a lua function that provides access to the request/response body).

**Describe alternatives you've considered (optional)**
- Running skipper as library and writing the filter myself.
- Tee'ing the response to another application that will do the logging.

**Additional context (optional)**
- [nginx](https://gist.github.com/morhekil/1ff0e902ed4de2adcb7a) can do this ([alternative](https://www.hardill.me.uk/wordpress/2018/03/14/logging-requests-and-response-with-nginx/))

**Would you like to work on it?**
No

AlexanderYastrebov:
I think there is [`auditLog`](https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#auditlog) filter for this purpose (but we have [it disabled](https://github.com/zalando-incubator/kubernetes-on-aws/blob/df67dd437c48355955bb96e5306bbdd2800d6681/cluster/manifests/skipper/deployment.yaml#L158) in our deployment).

bocytko:
I see. I used it and it seems the default when running skipper as library was 0 bytes. Will try again soon and close the issue if it works well.

bocytko:
Skipper with `auditLog()` configured:


Simulated backend application:
```sh
$ ./bin/skipper -inline-routes='all: * -> inlineContent("<h1>response</h1>") -> <shunt>' -address=:9092
```

Request proxy:
```sh
$ ./bin/skipper -inline-routes='all: * -> auditLog() -> "http://localhost:9092"' -max-audit-body=1024
```

Request:
```sh
$ curl -v GET localhost:9090 -d 'a=b'
```

Logs:
```sh
$ ./bin/skipper -inline-routes='all: * -> auditLog() -> "http://localhost:9092"' -max-audit-body=1024
[APP]INFO[0000] Expose metrics in codahale format            
[APP]INFO[0000] enable swarm: false                          
[APP]INFO[0000] Replacing lua filter specification           
[APP]INFO[0000] support listener on :9911                    
[...]

PR #2741, opened 2023-11-17
Filter: logBody("request", "response")

trkohler:
This filter logs body for the request if body has proper content type. When body is json / xml, it applies formatting. 

should resolve: https://github.com/zalando/skipper/issues/2282

### DIFF
--- a/filters/diag/logbody.go
+++ b/filters/diag/logbody.go
@@ -1,0 +1,179 @@
+package diag
+
+import (
+	"bytes"
+	"encoding/json"
+	"encoding/xml"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/zalando/skipper/filters"
+)
+
+type logBody struct {
+	request  bool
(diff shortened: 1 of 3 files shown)


trkohler:
closing because streaming approach is preferred and it's being worked on in https://github.com/zalando/skipper/pull/2428

szuecs:
That's also pretty nice to have formatting as option maybe. I did only plain string log. Do you have any use case in mind for the pretty print?
We could do this as option to logBody(). Right now we have only "request" and "response" as options, but we could also add something like "pretty".

trkohler:
I've noticed that in example there (https://github.com/zalando/skipper/issues/2282#issuecomment-1521440418) json was used and I thought that in the most cases users would be interested to log requests/response bodies which have some structure, rather than octet streams / gifs / etc. And because of that, I thought it would be a convenient to support formatting out of the box.

Issue #2043, opened 2022-07-26
document redis autoscaling


szuecs:
https://github.com/zalando/skipper/pull/2009 added redis based swarm autoscaling possibility
We lack in documentation for this feature:
- [ ] kubernetes dataclient docs https://opensource.zalando.com/skipper/kubernetes/ingress-controller/#redis-based
- [ ] swarm redis tutorial https://opensource.zalando.com/skipper/tutorials/ratelimit/#redis-based-cluster-ratelimits

trkohler:
I have briefly went through docs and haven't found mention on how skipper handles redis swarm, so it must mean the issue is still relevant?

PR #2744, opened 2023-11-20
Redis Swarm Configuration documented

trkohler:
should resolve: https://github.com/zalando/skipper/issues/2043

my notes & understanding on how it works based on what I read https://github.com/zalando/skipper/pull/2009/files

`kubernetes-redis-service-namespace` - sets namespace for redis to be used to lookup endpoints
`kubernetes-redis-service-name` - sets name for redis to be used to lookup endpoints. The only required parameter to redis swarm to be working. 
`kubernetes-redis-service-port` - sets port for redis to be used to lookup endpoints, default value is 6379

skipper is getting all endpoints using kubernetes service for redis cluster
then it updates the list of those addresses each 10 seconds
updates works in background with the separate goroutine
this enables autodiscovery for new redis instances


szuecs:
@trkohler if you create an image it's of course great for the reader.

Maybe also tell something about the routesrv case, because if you have a non-kubernetes skipper data-plane and use routesrv, you need to set these options to routesrv and set `-swarm-redis-remote` to the skipper data-plane to fetch the redis endpoints from routesrv API.

MustafaSaber:
https://opensource.zalando.com/skipper/tutorials/ratelimit/#redis-based-cluster-ratelimits

szuecs:
:+1:

docs/tutorials/ratelimit.md
```diff
@@ -113,6 +113,16 @@ There are 2 different configurations to assign redis instances as a skipper redi
 see also [Running with
 Redis based Cluster Ratelimits](../kubernetes/ingress-controller.md#redis-based)
 
+#### Redis Swarm Configuration
+
+When working with redis swarm, use kubernetes service selector. Configure it with `-kubernetes-redis-service-namespace` and `-kubernetes-redis-service-name` flags.
```
AlexanderYastrebov, line 118: I think we should capitalize Redis and Kubernetes

AlexanderYastrebov:
:+1:

Issue #1470, opened 2020-07-07
fix logHeader test


szuecs:
The current test is dumping log and does a string Equal.
The logHeader loops around req.Header map and is not stable sorted.
This creates test flakyness.

I would like to omit output sorting in the filter because it means we read 2 times the map.

I tried to use proxytest with Log in a first hack to have better control, but I was not able to do this as fast as we need the filter change.

szuecs:
@AlexanderYastrebov never saw this happening, so I guess it can be closed

PR #2747, opened 2023-11-22
test: increase test coverage for logHeader and test truncate authorization header

szuecs:
test: increase test coverage for logHeader and test truncate authorization header

### DIFF
--- a/filters/diag/logheader_test.go
+++ b/filters/diag/logheader_test.go
@@ -12,6 +12,25 @@ import (
 	"github.com/zalando/skipper/filters/filtertest"
 )
 
+func TestCreateFilterLogHeader(t *testing.T) {
+	lgh := logHeader{}
+	f, err := lgh.CreateFilter([]interface{}{"request", "response"})
+	if err != nil {
+		t.Fatal(err)
+	}
+	lgh = f.(logHeader)
+	if !(lgh.request && lgh.response) {
+		t.Errorf("Failed to set members: %v %v", lgh.request, lgh.response)
+	}
+}
+func TestCreateFilterLogHeaderWrongInput(t *testing.T) {
+	lgh := logHeader{}
+	_, err := lgh.CreateFilter([]interface{}{5})
+	if err == nil {
+		t.Fatal("Failed to get expected error 5 is no string")
+	}
+}
+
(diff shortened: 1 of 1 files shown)


filters/diag/logheader_test.go
```diff
@@ -152,3 +171,109 @@ func TestLogHeaderRequestResponse(t *testing.T) {
...
+	outputVerify.WriteString("Response for ")
+	outputVerify.WriteString(req.Method)
+	outputVerify.WriteString(" ")
+	outputVerify.WriteString(req.URL.Path)
+	outputVerify.WriteString(" ")
+	outputVerify.WriteString(req.Proto)
+	outputVerify.WriteString("\r\n")
+	outputVerify.WriteString(resp.Status)
+	outputVerify.WriteString("\r\n")
+	for k, v := range resp.Header {
```
AlexanderYastrebov, line 252: This will flake

AlexanderYastrebov: I think we have https://github.com/zalando/skipper/issues/1470 for this.
I guess we can sort headers by name before logging.

szuecs: This will only flake if there are more than one header. That's why I delete it to only have one

szuecs: You can also see bell that the following existing test has the same pattern and never appeared on our flaky test list.

filters/diag/logheader_test.go
```diff
@@ -152,3 +171,109 @@ func TestLogHeaderRequestResponse(t *testing.T) {
...
+
+	loggerVerify = bytes.NewBuffer(nil)
+	log.SetOutput(loggerVerify)
+	log.Println(outputVerify.String())
+
+	loggerTest = bytes.NewBuffer(nil)
+	log.SetOutput(loggerTest)
+
+	lh.Response(ctx)
+	if loggerTest.Len() == 0 || !bytes.Equal(loggerTest.Bytes(), loggerVerify.Bytes()) {
```
AlexanderYastrebov, line 271: I think there is no need to check length

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2748, opened 2023-11-26
Conditional rate limiting

ponimas:
This PR adds ability for the conditional rate limiting.
This extension allows you to rate limit clients based on existence or absence of certain headers (with or without specified values) without affecting the clients who do not fit the conditions.
##### An example:
```
-> clusterClientRatelimit("A", 30, "10s", "!X-Customer")
-> clusterClientRatelimit("B", 10, "10s", "~X-Customer,True-Client-IP")
```
The first ratelimit will be applied only to customers with the existing `X-Customer` header, the second one to the customers with the missing `X-Customer` header.
##### Syntax:
- If a header is prepended with `!`, the rate limiter will be applied only if this header exists in the request.

- If a header is prepended with `!` **and** also has a value separated by `=`, the rate limiter will be applied only if this header exists in the request and has this exact value.

- If a header  is prepended with `~`, the rate limiter will be applied only if this header does not exist in the request.

- If a header is prepended with `~` **and** also has a value separated by `=`, the rate limiter will be applied only if this header with this exact value does not exist in the request.

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -2011,12 +2011,16 @@ Parameters:
 
 * number of allowed requests per time period (int)
 * time period for requests being counted (time.Duration)
-* optional parameter to set the same client by header, in case the provided string contains `,`, it will combine all these headers (string)
+* optional parameter to set the same client by header, in case the provided string contains `,`, it will combine all these headers (string).
+  Header names can be prefixed by symbols '!', indicating that the rate limit applies only if this header exists in the request, and '~',
+  indicating that the rate limit applies only if this header is missing. In both cases, it is possible to provide a value for
+  fine-grained control in the form `!HeaderName=Value`.
 
 ```
 clientRatelimit(3, "1m")
 clientRatelimit(3, "1m", "Authorization")
 clientRatelimit(3, "1m", "X-Foo,Authorization,X-Bar")
+clientRatelimit(3, "1m", "X-Foo,Authorization,!X-Bar=foo")
 ```
 
 See also the [ratelimit docs](https://godoc.org/github.com/zalando/skipper/ratelimit).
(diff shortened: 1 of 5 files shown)


AlexanderYastrebov:
Why do we need this when the same can and should be achieved using predicates?

```
Header("X-Customer", ''JohnDoe") && True() -> clusterClientRatelimit("A", 30, "10s", "X-Customer") -> ...
HeaderRegexp("X-Customer", ''.+") -> clusterClientRatelimit("A", 30, "10s", "X-Customer") -> ...
* -> clusterClientRatelimit("B", 10, "10s", "True-Client-IP")
```

ponimas:
@szuecs but it was rebased :)

szuecs:
> @szuecs but it was rebased :)

just saw "merge" in the email notification

AlexanderYastrebov:
> your example increases the number of routes to maintain.

By how much? How would a set of routes that: have ratelimit `X` for `X-Customer=John Doe`, ratelimit `Y` for any other `X-Customer` value and ratelimit value for requests without `X-Customer` look like with your proposal?

I have several concerns with this approach:
1. The proposal can be achieved with already existing means - predicates. They are well-specified and allow many kinds of conditions over request, not only headers.
2. `clusterClientRatelimit` filter has already quite complex but at least well-understood contract. Allowing micro syntax into its header parameter increases complexity and makes implementation and usage more error-prone. We already have https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#template-placeholders but we refrain from adding support of it to `clusterClientRatelimit` to avoid complexity increase as one can always construct an auxiliary ratelimit header like:
```
// Ratelimit per resource
Path("/resource/:id") -> setRequestHeader("X-Resource-Id", "${id}") -> clusterClientRatelimit("resource", 10, "1m", "X-Resource-Id") -> "https://backend.example.org";
```

3. micro syntax is specified ad-hoc to solve one current problem. E.g. it does not address other than header properties of requests or matching a set values. For arbitrary complex expressions we'd better use a proper language (see https://github.com/zalando/skipper/blob/master/docs/reference/scripts.md). We already have one micro-syntax https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#template-placeholders and the proposal has nothing in common with it.

Maybe with more realistic example it could be shown how this proposal makes solution simpler than one built with existing means.

If we still decide to move forward, I suggest that we move this micro syntax into a separate new filter that evaluates it into a header value and then ratelimit using this header (similar to the above example). I also think we should rely on or borrow or extend existing https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#template-placeholders instead of inventing a new thing.

AlexanderYastrebov:
I think this was closed because cw2023freeze branch was removed.

ponimas:
@AlexanderYastrebov 
> By how much? 

In you example it will be trippled
It will be a single route
```
Header("X-Customer", "JohnDoe") && True() -> clusterClientRatelimit("A", 30, "10s", "X-Customer") -> ...
HeaderRegexp("X-Customer", ".+") -> clusterClientRatelimit("A", 30, "10s", "X-Customer") -> ...
* -> clusterClientRatelimit("B", 10, "10s", "True-Client-IP")
```

```
*
 ->  clusterClientRatelimit("A", 30, "10s", "!X-Customer=JohnDoe")
 ->  clusterClientRatelimit("A", 30, "10s", "~X-Customer=JohnDoe,!X-Customer")
 -> clusterClientRatelimit("B", 10, "10s", "~X-Customer,True-Client-IP")
 -> ...
```

PR #2752, opened 2023-11-27
Revert version metric

AlexanderYastrebov:
Version label could be added using Prometheus re-labeling, see https://github.com/zalando-incubator/kubernetes-on-aws/pull/6579


szuecs:
:+1:

PR #2757, opened 2023-11-28
validator: use ingress validator with dataclient

MustafaSaber:
Relates to https://github.com/zalando/skipper/issues/1618

IngressValidator was introduced in https://github.com/zalando/skipper/pull/2493 while we still validate inside skipper with some functions that will only drop the annotation but Ingress will still pass and work for example https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/ingress.go#L257C1-L269C2 which will drop the annotation only if parsing fail.

This PR should fix this and at least drop the Ingress resource if validation for skipper annotations fail.

### DIFF
--- a/dataclients/kubernetes/ingress.go
+++ b/dataclients/kubernetes/ingress.go
@@ -255,15 +255,11 @@ func annotationPredicate(m *definitions.Metadata) string {
 }
 
 // parse routes annotation
-func extraRoutes(m *definitions.Metadata, logger *logger) []*eskip.Route {
+func extraRoutes(m *definitions.Metadata) []*eskip.Route {
 	var extraRoutes []*eskip.Route
 	annotationRoutes := m.Annotations[definitions.IngressRoutesAnnotation]
 	if annotationRoutes != "" {
-		var err error
-		extraRoutes, err = eskip.Parse(annotationRoutes)
-		if err != nil {
-			logger.Errorf("Failed to parse routes from %s, skipping: %v", definitions.IngressRoutesAnnotation, err)
-		}
+		extraRoutes, _ = eskip.Parse(annotationRoutes) // We ignore the error here because it should be handled by the validator object
 	}
 	return extraRoutes
 }
--- a/dataclients/kubernetes/ingressv1.go
+++ b/dataclients/kubernetes/ingressv1.go
@@ -432,7 +432,7 @@ func (ing *ingress) ingressV1Route(
 		logger:              logger,
 		annotationFilters:   annotationFilter(i.Metadata, logger),
 		annotationPredicate: annotationPredicate(i.Metadata),
-		extraRoutes:         extraRoutes(i.Metadata, logger),
+		extraRoutes:         extraRoutes(i.Metadata),
 		backendWeights:      backendWeights(i.Metadata, logger),
 		pathMode:            pathMode(i.Metadata, ing.pathMode, logger),
 		redirect:            redirect,
--- a/dataclients/kubernetes/clusterclient.go
+++ b/dataclients/kubernetes/clusterclient.go
@@ -79,6 +79,7 @@ type clusterClient struct {
 
 	loggedMissingRouteGroups bool
 	routeGroupValidator      *definitions.RouteGroupValidator
+	ingressValidator         *definitions.IngressV1Validator
 }
 
 var (
@@ -173,6 +174,7 @@ func newClusterClient(o Options, apiURL, ingCls, rgCls string, quit <-chan struc
 		apiURL:                       apiURL,
 		certificateRegistry:          o.CertificateRegistry,
 		routeGroupValidator:          &definitions.RouteGroupValidator{},
+		ingressValidator:             &definitions.IngressV1Validator{},
 		enableEndpointSlices:         o.KubernetesEnableEndpointslices,
 	}
 
@@ -369,7 +371,16 @@ func (c *clusterClient) loadIngressesV1() ([]*definitions.IngressV1Item, error) 
 
 	sortByMetadata(fItems, func(i int) *definitions.Metadata { return fItems[i].Metadata })
 
-	return fItems, nil
+	validatedItems := make([]*definitions.IngressV1Item, 0, len(fItems))
+	for _, i := range fItems {
+		if err := c.ingressValidator.Validate(i); err != nil {
+			log.Errorf("[ingress] %v", err)
+			continue
+		}
+		validatedItems = append(validatedItems, i)
+	}
+
+	return validatedItems, nil
 }
 
 func (c *clusterClient) LoadRouteGroups() ([]*definitions.RouteGroupItem, error) {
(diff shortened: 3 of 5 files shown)


dataclients/kubernetes/ingressv1.go
```diff
@@ -410,6 +410,12 @@ func (ing *ingress) ingressV1Route(
 	}
 	logger := newLogger("Ingress", i.Metadata.Namespace, i.Metadata.Name, loggingEnabled)
 
+	ingressValidator := &definitions.IngressV1Validator{}
+	if err := ingressValidator.Validate(i); err != nil {
+		logger.Errorf("Ingress validation failed: %v", err)
+		return nil, nil // Drop the resource without failing LoadAll
```
MustafaSaber, line 416: If we returned and error here LoadAll will fail which doesn't make sense, so either move on with this or think of another solution.

MustafaSaber:
We can also start ignoring errors in https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/ingress.go#L226 & https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/ingress.go#L258 and then we can remove loggers

dataclients/kubernetes/ingress.go
```diff
@@ -255,15 +255,11 @@ func annotationPredicate(m *definitions.Metadata) string {
...
+func extraRoutes(m *definitions.Metadata) []*eskip.Route {
 	var extraRoutes []*eskip.Route
 	annotationRoutes := m.Annotations[definitions.IngressRoutesAnnotation]
 	if annotationRoutes != "" {
-		var err error
-		extraRoutes, err = eskip.Parse(annotationRoutes)
-		if err != nil {
-			logger.Errorf("Failed to parse routes from %s, skipping: %v", definitions.IngressRoutesAnnotation, err)
-		}
+		extraRoutes, _ = eskip.Parse(annotationRoutes) // We ignore the error here because it should be handled by the validator object
```
RomanZavodskikh, line 262: Do we need to do the same change in `func annotationFilter`?

MustafaSaber: Not yet, because in annotationFilter we also deal with `zalando.org/ratelimit` which we don't validate yet.

dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-invalid-routes-annotation.log
```diff
@@ -0,0 +1,2 @@
+Ingress validation failed: invalid \\\"zalando\.org\/skipper-routes\\\" annotation: invalid predicate count arg
```
szuecs, line 1: The log sounds weird for this test

szuecs: Please change also the file names to reflect _what_ is invalid

dataclients/kubernetes/ingress.go
MustafaSaber: 1. deployments without validator webhook will still work because we depend on ingressValidator object which is already part of dataclient.
2. Currently it should migrate safely as we only remove logging from this function which is swapped by logging from Validate function.

dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-invalid-routes-annotation.yaml
```diff
@@ -0,0 +1,74 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  namespace: foo
+  name: qux
+  annotations:
+    zalando.org/skipper-routes: |
+      r1: Header("test") && Path("/login") -> status(200) -> "http://foo.test"
```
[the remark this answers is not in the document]
MustafaSaber: why is it wrong? `Path` here is a valid predicate

dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-invalid-routes-annotation.log
szuecs: Ah I guess Header() needs 2, right?
Make it obvious please!

dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-invalid-routes-annotation.yaml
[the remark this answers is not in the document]
szuecs: You have Path/PathSubtree also in ingress (depends of the configuration what is used for ` pathType: ImplementationSpecific`, but with some configuration this could be not allowed to have 2 tree based predicates

dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-invalid-routes-annotation.log
MustafaSaber: > For example why do we have 2 predicates here, we don't check if we can do more than 1 predicate so why having 2?

That's a route, we don't allow multiple filters/predicates per yaml/json array item, but we allow multiples with routes. I can remove `Path` if it will make the error more clear, I just wanted to make one right and one wrong.

dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-invalid-routes-annotation.yaml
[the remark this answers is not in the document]
MustafaSaber: but it's still a valid ingress, do we have something that will prevent user from using `Path` here?

dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-invalid-routes-annotation.log
szuecs: Yes please! If you want to have more complex tests, add more tests and name them that it's obvious. :)
The filters/predicates items are only for structured data, here we have a plain eskip route string, so no number of predicates or filters to validate.

dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-invalid-routes-annotation.yaml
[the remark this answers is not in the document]
MustafaSaber: So there is 2 validations here, this route was going to pass ingressValidation which only parse the route, the route itself in this case is valid. Another function `addSpecRuleV1` which is called after ingressValidator would fail because what you just described (thanks for the example!).

AlexanderYastrebov:
:+1:

szuecs:
:+1:

MustafaSaber:
This is safe to merge according to our reports.

AlexanderYastrebov:
:+1:

MustafaSaber:
👍

PR #2759, opened 2023-11-30
Return endpointregistry to the load balancer

RomanZavodskikh:
This PR is intended to return endpointregistry component in the hotpath back, however, without introducing lock contention.
At the moment I create this PR to have a test image to perform load tests.

### DIFF
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -75,7 +75,7 @@ func shiftWeighted(rnd *rand.Rand, ctx *routing.LBContext, now time.Time) routin
 	rt := ctx.Route
 	ep := ctx.LBEndpoints
 	for _, epi := range ep {
-		wi := fadeIn(now, rt.LBFadeInDuration, rt.LBFadeInExponent, epi.Detected)
+		wi := fadeIn(now, rt.LBFadeInDuration, rt.LBFadeInExponent, epi.EndpointRegistryMetrics.DetectedTime())
 		sum += wi
 	}
 
@@ -83,7 +83,7 @@ func shiftWeighted(rnd *rand.Rand, ctx *routing.LBContext, now time.Time) routin
 	r := rnd.Float64() * sum
 	var upto float64
 	for i, epi := range ep {
-		upto += fadeIn(now, rt.LBFadeInDuration, rt.LBFadeInExponent, epi.Detected)
+		upto += fadeIn(now, rt.LBFadeInDuration, rt.LBFadeInExponent, epi.EndpointRegistryMetrics.DetectedTime())
 		if upto > r {
 			choice = ep[i]
 			break
@@ -116,7 +116,7 @@ func withFadeIn(rnd *rand.Rand, ctx *routing.LBContext, choice int, algo routing
 		now,
 		ctx.Route.LBFadeInDuration,
 		ctx.Route.LBFadeInExponent,
-		ctx.LBEndpoints[choice].Detected,
+		ctx.LBEndpoints[choice].EndpointRegistryMetrics.DetectedTime(),
 	)
 
 	if rnd.Float64() < f {
@@ -124,7 +124,7 @@ func withFadeIn(rnd *rand.Rand, ctx *routing.LBContext, choice int, algo routing
 	}
 	notFadingIndexes := make([]int, 0, len(ep))
 	for i := 0; i < len(ep); i++ {
-		if _, fadingIn := fadeInState(now, ctx.Route.LBFadeInDuration, ep[i].Detected); !fadingIn {
+		if _, fadingIn := fadeInState(now, ctx.Route.LBFadeInDuration, ep[i].EndpointRegistryMetrics.DetectedTime()); !fadingIn {
 			notFadingIndexes = append(notFadingIndexes, i)
 		}
 	}
@@ -263,7 +263,7 @@ func computeLoadAverage(ctx *routing.LBContext) float64 {
 	sum := 1.0 // add 1 to include the request that just arrived
 	endpoints := ctx.LBEndpoints
 	for _, v := range endpoints {
-		sum += float64(v.Metrics.GetInflightRequests())
+		sum += float64(v.EndpointRegistryMetrics.InflightRequests())
 	}
 	return sum / float64(len(endpoints))
 }
@@ -280,10 +280,10 @@ func (ch *consistentHash) boundedLoadSearch(key string, balanceFactor float64, c
 		if skipEndpoint(endpointIndex) {
 			continue
 		}
-		load := ctx.LBEndpoints[endpointIndex].Metrics.GetInflightRequests()
+		load := ctx.LBEndpoints[endpointIndex].EndpointRegistryMetrics.InflightRequests()
 		// We know there must be an endpoint whose load <= average load.
 		// Since targetLoad >= average load (balancerFactor >= 1), there must also be an endpoint with load <= targetLoad.
-		if load <= int(targetLoad) {
+		if float64(load) <= targetLoad {
 			break
 		}
 		ringIndex = (ringIndex + 1) % ch.Len()
@@ -375,7 +375,7 @@ func (p *powerOfRandomNChoices) Apply(ctx *routing.LBContext) routing.LBEndpoint
 // getScore returns negative value of inflightrequests count.
 func (p *powerOfRandomNChoices) getScore(e routing.LBEndpoint) int64 {
 	// endpoints with higher inflight request should have lower score
-	return -int64(e.Metrics.GetInflightRequests())
+	return -e.EndpointRegistryMetrics.InflightRequests()
 }
 
 type (
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -848,6 +848,10 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 	if endpoint != nil {
 		endpoint.Metrics.IncInflightRequest()
 		defer endpoint.Metrics.DecInflightRequest()
+
+		entry := p.registry.GetMetrics(endpoint.Host)
+		entry.IncInflightRequest()
+		defer entry.DecInflightRequest()
 	}
 
 	if p.experimentalUpgrade && isUpgradeRequest(req) {
--- a/routing/endpointregistry.go
+++ b/routing/endpointregistry.go
@@ -86,13 +86,14 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
 
 	for _, route := range routes {
 		if route.BackendType == eskip.LBBackend {
-			for _, epi := range route.LBEndpoints {
-				metrics := r.GetMetrics(epi.Host)
-				if metrics.DetectedTime().IsZero() {
-					metrics.SetDetected(now)
+			for i := range route.LBEndpoints {
+				epi := &route.LBEndpoints[i]
+				epi.EndpointRegistryMetrics = r.GetMetrics(epi.Host)
+				if epi.EndpointRegistryMetrics.DetectedTime().IsZero() {
+					epi.EndpointRegistryMetrics.SetDetected(now)
 				}
 
-				metrics.SetLastSeen(now)
+				epi.EndpointRegistryMetrics.SetLastSeen(now)
 			}
 		}
 	}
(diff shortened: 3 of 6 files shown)


routing/endpointregistry.go
```diff
@@ -18,6 +19,7 @@ type Metrics interface {
 
 type entry struct {
 	detected         time.Time
+	lastSeen         time.Time
 	inflightRequests int64
```
RomanZavodskikh, line 23: I was thinking about turning `inflightRequests` into `atomic.Int64`, but decided not to do this, because of `func (e *EndpointRegistry) GetMetrics(key string)` methods which returns a safe-to-read copy of entry. I expect "usual reads" to be much faster that `func (x *atomic.Int64) Load()`.

RomanZavodskikh:
Removing draft status and "do not merge" label because I performed the same loadtest that recreate the bug mentioned in this comment, and those tests are fine for this version of skipper-ingress.

https://github.com/zalando/skipper/pull/2645#discussion_r1360481696

routing/endpointregistry.go
```diff
@@ -83,57 +81,35 @@ func NewEndpointRegistry(o RegistryOptions) *EndpointRegistry {
...
 }
 
 func (r *EndpointRegistry) SetDetectedTime(key string, detected time.Time) {
-	r.mu.Lock()
-	defer r.mu.Unlock()
-
-	e := r.getOrInitEntryLocked(key)
-	e.detected = detected
+	e, _ := r.data.LoadOrStore(key, &entry{})
+	e.(*entry).detected = detected
```
AlexanderYastrebov, line 99: It is not safe to read and write this field concurrently. I think we use it only in post-processor but nothing tells client that it should not be used elsewhere.

Maybe we can make it non-exported or introduce a constructor like

 ```go
 func (r *EndpointRegistry) GetOrCreateMetrics(key string) (m Metrics, created bool) 
 ```

loadbalancer/algorithm.go
```diff
@@ -280,10 +284,10 @@ func (ch *consistentHash) boundedLoadSearch(key string, balanceFactor float64, c
 		if skipEndpoint(endpointIndex) {
 			continue
 		}
-		load := ctx.LBEndpoints[endpointIndex].Metrics.GetInflightRequests()
+		load := ctx.Registry.GetMetrics(ctx.LBEndpoints[endpointIndex].Host).InflightRequests()
 		// We know there must be an endpoint whose load <= average load.
 		// Since targetLoad >= average load (balancerFactor >= 1), there must also be an endpoint with load <= targetLoad.
-		if load <= int(targetLoad) {
+		if load <= int64(targetLoad) {
```
AlexanderYastrebov, line 290: I think load should be converted to float64 instead

routing/endpointregistry.go
RomanZavodskikh: > Maybe we can make it non-exported

I think I can make it non-exported (after changing some test code a little bit), it's not a problem.

> or introduce a constructor like

I don't understand how creating additional return value in getter can help us avoid get+set of the same timestamp races.

routing/endpointregistry.go
```diff
@@ -32,19 +34,18 @@ func (e *entry) InflightRequests() int64 {
...
-	data map[string]*entry
+	// map[string]*entry
+	data sync.Map
 }
 
 var _ PostProcessor = &EndpointRegistry{}
 
 type RegistryOptions struct {
 	LastSeenTimeout time.Duration
+	Detected        map[string]time.Time
```
RomanZavodskikh, line 48: This additional option is needed to make `func (r *EndpointRegistry) setDetectedTime(key string, detected time.Time)` non-exported and still have tests working.

routing/endpointregistry.go
AlexanderYastrebov: The idea was that GetOrCreateMetrics would set detected time and created would hint if metrics were created or reused similar to LoadOrStore

AlexanderYastrebov: BTW don't we need to allow SetDetectedTime for the endpointCreated filter?

routing/endpointregistry.go
```diff
@@ -82,58 +81,43 @@ func NewEndpointRegistry(o RegistryOptions) *EndpointRegistry {
...
 
 func (r *EndpointRegistry) GetMetrics(key string) Metrics {
-	r.mu.Lock()
-	defer r.mu.Unlock()
-
-	e := r.getOrInitEntryLocked(key)
+	e, _ := r.data.LoadOrStore(key, &entry{})
 	copy := &entry{}
-	*copy = *e
+	*copy = *e.(*entry)
```
AlexanderYastrebov, line 101: This is not thread-safe. I think we should create a test that calls GetMetrics and IncInflightRequest in parallel and it should not fail under race detector.

routing/endpointregistry.go
RomanZavodskikh: > BTW don't we need to allow SetDetectedTime for the endpointCreated filter?

https://opensource.zalando.com/skipper/reference/filters/#endpointcreated
> This filter is typically automatically appended

AFAIR endpointCreated is needed to store the information about endpoint detection timestamp and needed only for us as Skipper developers, not for Skipper users. As endpointregistry becomes available, I don't see any point for us as Skipper developers to have this filter, because endpointregistry stores the information itself.
Should we still give users an options to use `endpointCreated`?

routing/endpointregistry.go
RomanZavodskikh: https://github.com/zalando/skipper/pull/2759/files#diff-92e82ef5e7102bf8a84e0f19a069277acf1b50b758a92fd35d62521d45378b2aR96

routing/endpointregistry.go
RomanZavodskikh: I decided to keep `setDetectedTime` non-exported, it's easy to make export it if required.

RomanZavodskikh:
BTW I can check that endpointregistry.go is 100% covered with tests, do feel it makes sense to add such check inside Github Actions?
```
$ go test -count=1 ./routing/ -coverprofile profile.cov && go tool cover -func=profile.cov | grep endpointregistry.go | awk ' { print $3 } '
ok      github.com/zalando/skipper/routing      0.295s  coverage: 91.2% of statements
100.0%
100.0%
100.0%
100.0%
100.0%
100.0%
100.0%
100.0%
100.0%
$
```

Makefile
```diff
@@ -182,3 +182,8 @@ cover: .coverprofile-all ## coverage test and show it in your browser
 .PHONY: show-cover
 show-cover: .coverprofile-all
 	go tool cover -html .coverprofile-all
+
+.PHONY: check-coverage
+check-coverage: ## check 100% test coverage of the selected packages
+	go test -count=1 ./routing/ -coverprofile profile.cov
+	for p in $$(go tool cover -func=profile.cov | grep endpointregistry.go | awk ' { print $$3 } '); do if [ "a$$p" != "a100.0%" ]; then exit 1; fi; done
```
szuecs, line 189: It's a bit too specific Makefile target to my taste

RomanZavodskikh: Okay, will create separate PR for such change to limit the scope of the current PR.

routing/endpointregistry.go
```diff
@@ -82,58 +88,49 @@ func NewEndpointRegistry(o RegistryOptions) *EndpointRegistry {
...
+	// no need to copy lastSeen, because it is needed only for endpointregistry itself
 	return copy
 }
 
-func (r *EndpointRegistry) SetDetectedTime(key string, detected time.Time) {
-	r.mu.Lock()
-	defer r.mu.Unlock()
+func (r *EndpointRegistry) setDetectedTime(key string, detected time.Time) {
+	e, _ := r.data.LoadOrStore(key, newEntry())
+	e.(*entry).detected.Store(detected)
```
RomanZavodskikh, line 120: I think I will change this command to compare-and-swap, it seems to fit better here.

routing/endpointregistry.go
AlexanderYastrebov: Why do we need a copy? I think we should return pointer with safe accessor methods

loadbalancer/algorithm.go
```diff
@@ -112,19 +114,21 @@ func shiftToRemaining(rnd *rand.Rand, ctx *routing.LBContext, wi []int, now time
...
+		detected,
 	)
 
 	if rnd.Float64() < f {
 		return ep[choice]
 	}
 	notFadingIndexes := make([]int, 0, len(ep))
 	for i := 0; i < len(ep); i++ {
-		if _, fadingIn := fadeInState(now, ctx.Route.LBFadeInDuration, ep[i].Detected); !fadingIn {
+		detected := ctx.Registry.GetMetrics(ep[i].Host).DetectedTime()
```
AlexanderYastrebov, line 130: Detected time never changes but we read it a lot on each request. I am wondering if we can store it in registry put set actual value on the `ep` with a postprocessor.

proxy/proxy.go
```diff
@@ -848,6 +848,9 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 	if endpoint != nil {
 		endpoint.Metrics.IncInflightRequest()
 		defer endpoint.Metrics.DecInflightRequest()
+
+		p.registry.IncInflightRequest(endpoint.Host)
+		defer p.registry.DecInflightRequest(endpoint.Host)
```
AlexanderYastrebov, line 853: Can it happen that we increment and decrement on different instances of metrics here?

AlexanderYastrebov:
Lets refactor registry first to return pointer to entry (without copying) and make entry values getters/setters thread safe.
Then we can change LBEndpoint to use Metrics instead of *LBMetrics and populate ref in the post-processor to avoid `ctx.Registry.GetMetrics(ep[i].Host)` in runtime altogether.

RomanZavodskikh:
> Lets refactor registry first to return pointer to entry (without copying) and make entry values getters/setters thread safe.

https://github.com/zalando/skipper/pull/2795

szuecs:
:+1:

routing/routing.go
```diff
@@ -179,6 +179,11 @@ type LBEndpoint struct {
 	// Detected represents the time when skipper instances first detected a new LB endpoint. This detection
 	// time is used for the fade-in feature of the round-robin and random LB algorithms.
 	Detected time.Time
+
+	// There is ongoing effort to move all dynamic data about endpoints to the EndpointRegistry instead of the
+	// LBEndpoint struct. This will allow us to remove Metrics and Detected fields above and replace them with
+	// EndpointRegistryMetrics which has host-wide metrics instead of route-wide metrics.
+	EndpointRegistryMetrics Metrics
```
AlexanderYastrebov, line 186: I am wondering if we should return `*LBMetrics` from endpoint registry instead of adding new filed here.

proxy/proxy.go
```diff
@@ -848,6 +848,10 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 	if endpoint != nil {
 		endpoint.Metrics.IncInflightRequest()
 		defer endpoint.Metrics.DecInflightRequest()
+
+		entry := p.registry.GetMetrics(endpoint.Host)
+		entry.IncInflightRequest()
+		defer entry.DecInflightRequest()
```
AlexanderYastrebov, line 854: We should also count inflight requests for non-loadbalanced backends. k8s dataclient uses network backend when service has a single endpoint https://github.com/zalando/skipper/blob/07dee9e2c9876937147cb6bc9e467544ba31ce21/dataclients/kubernetes/ingressv1.go#L137-L149

routing/endpointregistry.go
```diff
@@ -86,13 +86,14 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
 
 	for _, route := range routes {
 		if route.BackendType == eskip.LBBackend {
-			for _, epi := range route.LBEndpoints {
-				metrics := r.GetMetrics(epi.Host)
-				if metrics.DetectedTime().IsZero() {
-					metrics.SetDetected(now)
+			for i := range route.LBEndpoints {
```
AlexanderYastrebov, line 89: It should update timestamps for network backends as well, see https://github.com/zalando/skipper/pull/2759/files#r1441546084

RomanZavodskikh:
As this PR changed a lot since the creation, I'll open new one soon.

RomanZavodskikh:
Please feel free to write feedback about this PR here:
https://github.com/zalando/skipper/pull/2823

proxy/proxy.go
RomanZavodskikh: From my side, it looks like it's better to return `hostname:port` from `mapRequest` and don't think what backend type it is at all.

routing/routing.go
RomanZavodskikh: Yes, I think it's possible to make `LBMetrics` methods to call endpointregistry `Metrics` methods.

PR #2760, opened 2023-11-30
feature: tarpit() filter

szuecs:
feature: tarpit() filter to let connection to the client open until it closes the connection

### DIFF
--- a/filters/diag/tarpit.go
+++ b/filters/diag/tarpit.go
@@ -1,0 +1,59 @@
+package diag
+
+import (
+	"net/http"
+	"time"
+
+	"github.com/zalando/skipper/filters"
+)
+
+type tarpitSpec struct{}
+
+type tarpit struct {
+	d time.Duration
+}
+
+func NewTarpit() filters.Spec {
+	return &tarpitSpec{}
+}
+
+func (t *tarpitSpec) Name() string {
+	return filters.TarpitName
+}
+
+func (t *tarpitSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) != 1 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+	s, ok := args[0].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	d, err := time.ParseDuration(s)
+	if err != nil {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	return &tarpit{d: d}, nil
+}
+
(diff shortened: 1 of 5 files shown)


filters/diag/tarpit.go
```diff
@@ -0,0 +1,60 @@
...
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	d, err := time.ParseDuration(s)
+	if err != nil {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	t.d = d
```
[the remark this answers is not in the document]
szuecs: Haha yeah good point 👍

filters/diag/tarpit.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"net/http"
+	"time"
+
+	"github.com/zalando/skipper/filters"
+)
+
+type tarpitSpec struct {
+	d time.Duration
```
AlexanderYastrebov, line 11: I think this field is unused

szuecs: ah yes it's used as setting the default but never possible to use the default.

szuecs: fixed

AlexanderYastrebov:
I think this is equivalent in behavior with
```
$ bin/skipper -inline-routes='* -> chunks(1, "1s") -> dropResponseHeader("Content-Length") -> repeatContent("X", 1000000000) -> status(200) -> <shunt>'
...
~$ nc localhost 9090
GET / HTTP/1.1        
Host: localhost:9090

HTTP/1.1 200 OK
Server: Skipper
Date: Fri, 01 Dec 2023 09:32:40 GMT
Transfer-Encoding: chunked

1
X
1
X
1
X
1
X
1
X
1
X
1
X
1
X
^C
```
-- it responds with `Transfer-Encoding: chunked` and sends chunks of 1 byte.

If we do not drop Content-Length then it just slowly sends response:
```
$ bin/skipper -inline-routes='* -> chunks(1, "1s") -> repeatContent("X", 10000) -> status(200) -> <shunt>'
...
~$ nc localhost 9090
GET / HTTP/1.1 
Host: localhost:9090

HTTP/1.1 200 OK
Content-Length: 10000
Server: Skipper
Date: Fri, 01 Dec 2023 09:33:42 GMT

XXXXXX^C
```

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2775, opened 2023-12-04
filters/tracing: improve stateBagToTag 

AlexanderYastrebov:
* reduce allocations when tag value is string
* check statebag value first to bail out early if it is absent
* restrict max number of arguments to two

```
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/filters/tracing
StateBagToTag_StringValue-8   313.1n ± 2%   124.3n ± 1%  -60.29% (p=0.000 n=10)

                            │    B/op    │   B/op     vs base                 │
StateBagToTag_StringValue-8   19.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=10)

                            │   HEAD~1   │                HEAD                 │
                            │ allocs/op  │ allocs/op   vs base                 │
StateBagToTag_StringValue-8   2.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)
```

For https://github.com/zalando-incubator/kubernetes-on-aws/pull/6604

### DIFF
--- a/filters/tracing/statebagtotag.go
+++ b/filters/tracing/statebagtotag.go
@@ -25,7 +25,7 @@ func (stateBagToTagSpec) Name() string {
 }
 
 func (stateBagToTagSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
-	if len(args) < 1 {
+	if len(args) < 1 || len(args) > 2 {
 		return nil, filters.ErrInvalidFilterParameters
 	}
 
@@ -43,7 +43,7 @@ func (stateBagToTagSpec) CreateFilter(args []interface{}) (filters.Filter, error
 		tagName = tagNameArg
 	}
 
-	return stateBagToTagFilter{
+	return &stateBagToTagFilter{
 		stateBagItemName: stateBagItemName,
 		tagName:          tagName,
 	}, nil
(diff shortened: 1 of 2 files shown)


filters/tracing/statebagtotag.go
```diff
@@ -53,16 +53,22 @@ func NewStateBagToTag() filters.Spec {
...
 	span := opentracing.SpanFromContext(ctx.Request().Context())
 	if span == nil {
 		return
 	}
-	value, ok := ctx.StateBag()[f.stateBagItemName]
-	if !ok {
-		return
+
+	if _, ok := value.(string); ok {
+		span.SetTag(f.tagName, value)
```
AlexanderYastrebov, line 68: See https://github.com/golang/go/issues/64541 why not
```go
if s, ok := value.(string); ok {
    span.SetTag(f.tagName, s)
}
```

szuecs: I guess we have similar cases all over the place.

filters/tracing/statebagtotag.go
```diff
@@ -53,16 +53,22 @@ func NewStateBagToTag() filters.Spec {
...
+	} else {
+		span.SetTag(f.tagName, fmt.Sprint(value))
 	}
-	span.SetTag(f.tagName, fmt.Sprint(value))
 }
 
-func (stateBagToTagFilter) Response(ctx filters.FilterContext) {}
+func (f *stateBagToTagFilter) Response(ctx filters.FilterContext) {}
```
szuecs, line 74: You don't need to assign it to a name, or any linter that yells?

AlexanderYastrebov: I remember there was a linter warning in the past so I changed that but maybe its not needed anymore

AlexanderYastrebov: Removed the name

filters/tracing/statebagtotag.go
AlexanderYastrebov: I checked all span.SetTag callsites and its fine. There may be other places that receive `interface{}` like SetTag of course but I think this is a very specific pattern here - optimizing for string case.

We could simply use `span.SetTag(f.tagName, value)` without type assertion but that is not guaranteed to work for types other than string, bool and ints:
```go
// Adds a tag to the span.
	//
	// If there is a pre-existing tag set for `key`, it is overwritten.
	//
	// Tag values can be numeric types, strings, or bools. The behavior of
	// other tag value types is undefined at the OpenTracing level. If a
	// tracing system does not know how to handle a particular value type, it
	// may ignore the tag, but shall not panic.
	//
	// Returns a reference to this Span for chaining.
	SetTag(key string, value interface{}) Span
```
so I decided to keep fmt.Sprint()

MustafaSaber:
👍

AlexanderYastrebov:
:+1:

PR #2776, opened 2023-12-05
skipper: remove old fuzz targets.

sepehrdaddev:
this will remove the old fuzz targets in favor of the ones that are in the `fuzz/` directory,

this is a continuation of https://github.com/zalando/skipper/pull/2745


sepehrdaddev:
:+1:

PR #2784, opened 2023-12-08
add clusterfuzzlite PR fuzzing.

sepehrdaddev:
add [clusterfuzzlite](https://google.github.io/clusterfuzzlite/) PR fuzzing.

it uses the same code as oss-fuzz but it will trigger fuzzing on every PR and it runs for 300 seconds

### DIFF
--- a/.clusterfuzzlite/Dockerfile
+++ b/.clusterfuzzlite/Dockerfile
@@ -1,0 +1,6 @@
+FROM gcr.io/oss-fuzz-base/base-builder-go
+
+COPY . $SRC/skipper
+COPY ./.clusterfuzzlite/build.sh $SRC/
+
+WORKDIR $SRC/skipper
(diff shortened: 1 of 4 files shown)


sepehrdaddev:
:+1:

AlexanderYastrebov:
Won't this add 10 minutes to PR build?

szuecs:
@sepehrdaddev that's too long maybe 2m?
It blocks the build

sepehrdaddev:
@szuecs I'm not sure which build you are referring to the one that builds the docker images and etc run in parallel to this (in fact all the checks are running in parallel since this is a separate workflow)

as for the timing, it take about 5 mins to build the fuzzers, so I cannot change that but the remaining 10 mins I can change.

sepehrdaddev:
I have updated the fuzz-seconds to be 300 seconds so its a bit faster.

sepehrdaddev:
:+1:

sepehrdaddev:
:+1:

PR #2788, opened 2023-12-09
predicates/auth: fix array conversion

AlexanderYastrebov:
* fix array conversion that was necessary due to https://github.com/golang/go/issues/46505
* preallocate slice


RomanZavodskikh:
:+1:

AlexanderYastrebov:
:+1:

PR #2796, opened 2023-12-15
dataclients/kubernetes: partially create RouteGroup routes on missing service

AlexanderYastrebov:
Ignore routes that reference missing service.

### DIFF
--- a/dataclients/kubernetes/clusterclient_test.go
+++ b/dataclients/kubernetes/clusterclient_test.go
@@ -266,7 +266,7 @@ func TestLoggingInterval(t *testing.T) {
 	defer log.SetOutput(os.Stderr)
 
 	countMessages := func() int {
-		return strings.Count(out.String(), "Error transforming external hosts")
+		return strings.Count(out.String(), "Ignoring route 0: service not found: default/myapp")
 	}
 
 	a, err := kubernetestest.NewAPI(kubernetestest.TestAPIOptions{}, manifest)
--- a/dataclients/kubernetes/routegroup.go
+++ b/dataclients/kubernetes/routegroup.go
@@ -407,9 +407,13 @@ func transformExplicitGroupRoute(ctx *routeContext) (*eskip.Route, error) {
 // explicitGroupRoutes creates routes for those route groups that have the
 // `route` field explicitly defined.
 func explicitGroupRoutes(ctx *routeGroupContext) ([]*eskip.Route, error) {
-	var routes []*eskip.Route
+	var result []*eskip.Route
 	rg := ctx.routeGroup
+
+nextRoute:
 	for routeIndex, rgr := range rg.Spec.Routes {
+		var routes []*eskip.Route
+
 		if len(rgr.Methods) == 0 {
 			rgr.Methods = []string{""}
 		}
@@ -437,7 +441,8 @@ func explicitGroupRoutes(ctx *routeGroupContext) ([]*eskip.Route, error) {
 					backend:    be,
 				})
 				if err != nil {
-					return nil, err
+					ctx.logger.Errorf("Ignoring route %d: %v", routeIndex, err)
+					continue nextRoute
 				}
 
 				backendTraffic[bref.BackendName].apply(r)
@@ -447,9 +452,11 @@ func explicitGroupRoutes(ctx *routeGroupContext) ([]*eskip.Route, error) {
 				routes = appendHTTPSRedirect(ctx, routes, r)
 			}
 		}
+
+		result = append(result, routes...)
 	}
 
-	return routes, nil
+	return result, nil
 }
 
 func transformRouteGroup(ctx *routeGroupContext) ([]*eskip.Route, error) {
--- a/dataclients/kubernetes/testdata/routegroups/convert/missing-service-1-of-2.yaml
+++ b/dataclients/kubernetes/testdata/routegroups/convert/missing-service-1-of-2.yaml
@@ -1,0 +1,55 @@
+apiVersion: zalando.org/v1
+kind: RouteGroup
+metadata:
+  name: myapp
+spec:
+  hosts:
+    - example.org
+  backends:
+    - name: myapp-1
+      type: service
+      serviceName: myapp-1
+      servicePort: 80
+    - name: myapp-2
+      type: service
+      serviceName: myapp-2
+      servicePort: 80
+  defaultBackends:
+    - backendName: myapp-2
+  routes:
+    - path: /app1
+      backends:
+        - backendname: myapp-1
+    - path: /app2
+      backends:
+        - backendname: myapp-2
+    - path: /split
+      backends:
+        - backendname: myapp-1
+        - backendname: myapp-2
+    - path: /default
+
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: myapp-1
+spec:
+  ports:
+    - port: 80
+      protocol: TCP
+      targetPort: 80
+  selector:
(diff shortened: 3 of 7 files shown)


AlexanderYastrebov:
Current implementation produces no routes and therefore this test case fails:
```
--- FAIL: TestRouteGroupConvert (0.08s)
    --- FAIL: TestRouteGroupConvert/missing-service-1-of-2 (0.00s)
        fixtures.go:297: Failed to convert the resources to the right routes.
        fixtures.go:298: routes: 0, expected: 2
        fixtures.go:299: got:
        fixtures.go:300: expected:
            kube_rg____example_org__catchall__0_0: Host("^(example[.]org[.]?(:[0-9]+)?)$") -> <shunt>;
            kube_rg__default__myapp__all__0_0: Host("^(example[.]org[.]?(:[0-9]+)?)$") && Path("/app1") -> <roundRobin, "http://10.2.4.16:80", "http://10.2.4.8:80">;
        fixtures.go:301: diff
              string(
            - 	"kube_rg____example_org__catchall__0_0: Host(\"^(example[.]org[.]?(:[0-9]+)?)$\")\n-> <shunt>;\n\nkube_rg__default__myapp__all__0_0: Host(\"^(example[.]org[.]?(:[0-9]+)?)$\") && Path(\"/app1\")\n-> <roundRobin, \"http://10.2.4.16:80\", \"http://10.2.4.8:80\">;",
            + 	"",
              )
            :
        fixtures.go:199: Captured logs:
        fixtures.go:200: time="2023-12-15T13:45:41Z" level=error msg="Error transforming external hosts: service not found: default/myapp-2" kind=RouteGroup name=myapp ns=
```

dataclients/kubernetes/routegroup.go
```diff
@@ -437,7 +441,8 @@ func explicitGroupRoutes(ctx *routeGroupContext) ([]*eskip.Route, error) {
 					backend:    be,
 				})
 				if err != nil {
-					return nil, err
+					ctx.logger.Errorf("Ignoring route %d: %v", routeIndex, err)
+					continue nextRoute
```
MustafaSaber, line 445: Nice to know!

dataclients/kubernetes/clusterclient_test.go
```diff
@@ -266,7 +266,7 @@ func TestLoggingInterval(t *testing.T) {
 	defer log.SetOutput(os.Stderr)
 
 	countMessages := func() int {
-		return strings.Count(out.String(), "Error transforming external hosts")
+		return strings.Count(out.String(), "Ignoring route 0: service not found: default/myapp")
```
MustafaSaber, line 269: I think we need to refactor this counting thing

AlexanderYastrebov: Why? It uses route index. We can do i+1 but I think this will be more confusing. Or do you have another idea?

szuecs: @AlexanderYastrebov do we break if we have endpoints but no service?
If so: Why? 

Test case looks like this and I would expect it to work (likely it breaks because of service port lookup, which might be the right thing):
```
apiVersion: zalando.org/v1
kind: RouteGroup
metadata:
  name: myapp
spec:
  hosts:
  - example.org
  backends:
  - name: myapp
    type: service
    serviceName: myapp
    servicePort: 80
  defaultBackends:
  - backendName: myapp
  routes:
  - path: /app
---
apiVersion: v1
kind: Endpoints
metadata:
  name: myapp
subsets:
- addresses:
  - ip: 10.2.4.8
  - ip: 10.2.4.16
  ports:
  - port: 80
```

MustafaSaber: I was talking about the whole test function, I remember I had to change the manifest before for validation, the more we tighten validation the more the need to change in this test case https://github.com/zalando/skipper/commit/92e0d11af311949faf68d55635ec302c66d38b1c

dataclients/kubernetes/routegroup.go
szuecs: In case of /split this will drop routes that might work before

dataclients/kubernetes/testdata/routegroups/convert/missing-service-1-of-2.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - path: /app1
+      backends:
+        - backendname: myapp-1
+    - path: /app2
+      backends:
+        - backendname: myapp-2
+    - path: /split
+      backends:
+        - backendname: myapp-1
+        - backendname: myapp-2
```
szuecs, line 29: I would expect this to have a route for /split and myapp-2.

dataclients/kubernetes/routegroup.go
AlexanderYastrebov: No, before it would ignore the whole route group if any of the services is missing.

dataclients/kubernetes/testdata/routegroups/convert/missing-service-1-of-2.yaml
AlexanderYastrebov: Proposed change creates routes for services that exist but does not create routes if any of the backends is missing like before. The traffic split will not be correct if service is missing.

szuecs: Yes I see and it's an enhancement, but if you think about an actual deployment with a wrong config for the new version deployment (for example missing service.yaml for myapp-1), then this could lead in traffic switching cases to unexpected routing drops

dataclients/kubernetes/routegroup.go
szuecs: I mean /split with only my-app2 and then you add my-app1 with a bogus config as an example.

dataclients/kubernetes/clusterclient_test.go
AlexanderYastrebov: > do we break if we have endpoints but no service?
> If so: Why?

The code checks service existence and it uses it to resolve port, see https://github.com/zalando/skipper/blob/ab3a18d1d4441585d069621a9adadf418fcc2cf0/dataclients/kubernetes/routegroup.go#L171-L193

AlexanderYastrebov: The scope of this change is to fix the bug we observe. I think we should handle test improvement or reconsider endpoint port handling in a separate effort (if at all).

dataclients/kubernetes/testdata/routegroups/convert/missing-service-1-of-2.yaml
AlexanderYastrebov: It is not worse than before and I want to fix the case we have. We may consider creating routes for the /split case separately

szuecs:
:+1:

dataclients/kubernetes/testdata/routegroups/convert/missing-service-1-of-2.yaml
AlexanderYastrebov: We would need to refactor traffic calculation. Currently we discover missing services on the go but we'd need to first check all services, exclude missing and only then recalculate traffic. Besides there might be other reasons why it could not create a route for the routegroup (not allowed external backend or malformed predicate/filter) so I think its safe to just throw away routegroup route if there is any kind of error with it instead of trying to guess/construct half-working configuration.

AlexanderYastrebov:
:+1:

PR #2814, opened 2024-01-01
Dynamically secure RouteGroups using Kubernetes TLS Secrets

### LINKED ISSUE #2811: Secure RouteGroups with Kubernetes TLS Secrets
Skipper currently has the ability to dynamically terminate Ingress using Kubernetes TLS secrets.

https://github.com/zalando/skipper/pull/1974

We would like to add this functionality to RouteGroups as well.

* Update the RouteGroup CRD to add a TLS secret to be used to terminate the connection.
* Update Skipper to terminate TLS on RouteGroups

The only alternative is to use Ingress which is limited.

Yes

rickhlx:
Closes #2811 

* Updating RouteGroup CRD from https://github.com/szuecs/routegroup-client/pull/16
* Refactor to use single function to add a cert to registry on both RouteGroups and Ingress
* Both internal and external RouteGroups can terminate TLS with Kubernetes secret


dataclients/kubernetes/routegroup.go
```diff
@@ -478,7 +480,34 @@ func splitHosts(hosts []string, domains []string) ([]string, []string) {
...
+// and adds the TLS secret to the registry if a match is found.
+func (r *routeGroups) addRouteGroupTLS(ctx *routeGroupContext, tls *definitions.RouteTLSSpec) {
+	// Host in the tls section need to explicitly match the host in the RouteGroup
+	hostlist := compareStringList(tls.Hosts, ctx.routeGroup.Spec.UniqueHosts())
+	if len(hostlist) == 0 {
+		ctx.logger.Infof("No matching tls hosts found")
+		return
+	}
+
+	// Skip adding certs to registry since no certs defined
```
rickhlx, line 493: Ingress config has `secretName` as optional, to indicate TLS is terminated with the default cert SNI.

/cc @AlexanderYastrebov @szuecs

dataclients/kubernetes/routegroup.go
```diff
@@ -478,7 +480,34 @@ func splitHosts(hosts []string, domains []string) ([]string, []string) {
 	return internalHosts, externalHosts
 }
 
-func (r *routeGroups) convert(s *clusterState, df defaultFilters, loggingEnabled bool) ([]*eskip.Route, error) {
+// addRouteGroupTLS compares the RouteGroup host list and the RouteGroup TLS host list
+// and adds the TLS secret to the registry if a match is found.
+func (r *routeGroups) addRouteGroupTLS(ctx *routeGroupContext, tls *definitions.RouteTLSSpec) {
+	// Host in the tls section need to explicitly match the host in the RouteGroup
+	hostlist := compareStringList(tls.Hosts, ctx.routeGroup.Spec.UniqueHosts())
```
szuecs, line 487: I think what we want here is to check if `tls.Hosts` is a subset of `ctx.routeGroup.Spec.UniqueHosts()`. 
`compareStringList` returns all items from `tls.Hosts` that are also part of `ctx.routeGroup.Spec.UniqueHosts()`.
So it reduces the hostlist.

Did you do this by intention and I miss something?

dataclients/kubernetes/routegroup.go
szuecs: sounds good

dataclients/kubernetes/routegroup.go
```diff
@@ -478,7 +480,34 @@ func splitHosts(hosts []string, domains []string) ([]string, []string) {
...
+	if len(hostlist) == 0 {
+		ctx.logger.Infof("No matching tls hosts found")
```
szuecs, line 489: maybe this is a miss configuration and we should log the values of `tls.Hosts` and `ctx.routeGroup.Spec.UniqueHosts()´ to help people debugging the miss configuration.

dataclients/kubernetes/routegroup.go
```diff
@@ -478,7 +480,34 @@ func splitHosts(hosts []string, domains []string) ([]string, []string) {
...
+		return
+	}
+
+	// Skip adding certs to registry since no certs defined
+	if tls.SecretName == "" {
+		ctx.logger.Infof("No tls secret defined for hosts - %s", tls.Hosts)
```
szuecs, line 495: So everyone that relies on the default cert will get every time this log line.
Maybe `Infof` -> `Debugf` or something like https://pkg.go.dev/golang.org/x/time/rate#Sometimes as we do in https://github.com/zalando/skipper/blob/master/proxy/proxy.go#L743?

dataclients/kubernetes/routegroup.go
```diff
@@ -478,7 +480,34 @@ func splitHosts(hosts []string, domains []string) ([]string, []string) {
...
+		return
+	}
+
+	// Secrets should always reside in the same namespace as the RouteGroup
+	secretID := &definitions.ResourceID{Name: tls.SecretName, Namespace: ctx.routeGroup.Metadata.Namespace}
```
szuecs, line 500: I would use a plain value instead of a pointer here:
```
secretID := definitions.ResourceID{Name: tls.SecretName, Namespace: ctx.routeGroup.Metadata.Namespace}secret, ok := ctx.state.secrets[secretID]
```

The rest of the code won't change.

dataclients/kubernetes/testdata/ingressV1/tls/tls-no-secret.yaml
```diff
@@ -0,0 +1,91 @@
+apiVersion: apps/v1
+kind: Deployment
```
szuecs, line 2: We don't read Deployment so you can also drop these yaml objects in tests.

dataclients/kubernetes/routegroup.go
rickhlx: We only want to add certs which have a match from from tls.Hosts and the routeGroup hosts, so reducing the list is desired. If there are no matches we log and return, but if there are matches, we use the reduced list to add the matching tls certs to the registry. 

We could log the non-matching hosts in tls.hosts to help debugging a miss configuration. What do you think?

dataclients/kubernetes/routegroup.go
rickhlx: Went with the Debug log level here. Let me know if you want to do the alternative.

dataclients/kubernetes/testdata/ingressV1/tls/tls-no-secret.yaml
rickhlx: Also removed from existing Ingress test fixtures.

dataclients/kubernetes/routegroup.go
szuecs: Yes logging makes sense.
I wonder in case of an update that is applied wrong, what we want to have as behavior. For example a timeline like this:
1. rg foo has tls.hosts foo.example, bar.example and there is a referenced secret that has only these two names foo.example, bar.example as hosts in SAN
2. rg was updated to add baz.example and remove bar.example in tls.hosts

The current behavior would drop bar.example.

Should this better stop updating it, such that bar.example is not replaced by baz.example ?!

I am thinking of migrations, but can not come up with an idea how this could break less with either case, so from my side we can also just log it.

@AlexanderYastrebov wdyt?

szuecs:
This is the error:
```
 go vet ./...
# github.com/zalando/skipper/dataclients/kubernetes
Error: dataclients/kubernetes/ingressv1.go:318:23: call of addTLSCertToRegistry copies lock value: github.com/zalando/skipper/secrets/certregistry.CertRegistry contains sync.Mutex
Error: dataclients/kubernetes/kube.go:601:30: addTLSCertToRegistry passes lock by value: github.com/zalando/skipper/secrets/certregistry.CertRegistry contains sync.Mutex
Error: dataclients/kubernetes/routegroup.go:515:23: call of addTLSCertToRegistry copies lock value: github.com/zalando/skipper/secrets/certregistry.CertRegistry contains sync.Mutex
make: *** [Makefile:127: vet] Error 1
Error: Process completed with exit code 2.
```

dataclients/kubernetes/ingressv1.go
```diff
@@ -296,12 +296,26 @@ func (ing *ingress) addSpecIngressTLSV1(ic *ingressContext, ingtls *definitions.
...
 	// Secrets should always reside in same namespace as the Ingress
-	secretID := &definitions.ResourceID{Name: ingtls.SecretName, Namespace: ic.ingressV1.Metadata.Namespace}
-	addHostTLSCert(ic, hostlist, secretID)
+	secretID := definitions.ResourceID{Name: ingtls.SecretName, Namespace: ic.ingressV1.Metadata.Namespace}
+	secret, ok := ic.state.secrets[secretID]
+	if !ok {
+		ic.logger.Errorf("Failed to find secret %s in namespace %s", secretID.Name, secretID.Namespace)
+		return
+	}
+	addTLSCertToRegistry(*ic.certificateRegistry, ic.logger, hostlist, secret)
```
szuecs, line 318: I think you just want to pass `ic.certificateRegistry` as pointer here

dataclients/kubernetes/kube.go
```diff
@@ -595,3 +595,19 @@ func compareStringList(a, b []string) []string {
 	}
 	return c
 }
+
+// addTLSCertToRegistry adds a TLS certificate to the certificate registry per host using the provided
+// Kubernetes TLS secret
+func addTLSCertToRegistry(cr certregistry.CertRegistry, logger *logger, hosts []string, secret *secret) {
```
szuecs, line 601: should be: `*certregistry.CertRegistry`

dataclients/kubernetes/kube.go
```diff
@@ -595,3 +595,19 @@ func compareStringList(a, b []string) []string {
 	}
 	return c
 }
+
+// addTLSCertToRegistry adds a TLS certificate to the certificate registry per host using the provided
+// Kubernetes TLS secret
+func addTLSCertToRegistry(cr *certregistry.CertRegistry, logger *logger, hosts []string, secret *secret) {
+	cert, err := generateTLSCertFromSecret(secret)
+	if err != nil {
+		logger.Errorf("Failed to generate TLS certificate from secret: %v", err)
```
AlexanderYastrebov, line 604: Maybe pass secretId and log it as well

dataclients/kubernetes/ingressv1.go
```diff
@@ -296,12 +296,26 @@ func (ing *ingress) addSpecIngressTLSV1(ic *ingressContext, ingtls *definitions.
 	// Hosts in the tls section need to explicitly match the host in the rules section.
 	hostlist := compareStringList(ingtls.Hosts, definitions.GetHostsFromIngressRulesV1(ic.ingressV1))
 	if len(hostlist) == 0 {
-		ic.logger.Infof("No matching tls hosts found")
+		ic.logger.Infof("No matching tls hosts found - tls hosts: %s, ingress hosts: %s", ingtls.Hosts, definitions.GetHostsFromIngressRulesV1(ic.ingressV1))
```
AlexanderYastrebov, line 299: Please introduce local variable for `definitions.GetHostsFromIngressRulesV1(ic.ingressV1)`

dataclients/kubernetes/routegroup.go
AlexanderYastrebov: Maybe this should be a part of validator and invalid routegroup should be rejected. We already use validator in webhook and dataclient.

dataclients/kubernetes/testdata/ingressV1/tls/tls-host-mismatch.log
```diff
@@ -0,0 +1,2 @@
+level=info msg="Hosts in TLS and Ingress don't match: tls hosts: \[example.org bar.org\], ingress hosts: \[example.org\]" kind=Ingress name=myapp-ingress ns=default
+level=info msg="adding certificate to registry - example.org"
```
AlexanderYastrebov, line 2: Please add newline at the end.

dataclients/kubernetes/testdata/ingressV1/tls/tls-host-mismatch.yaml
```diff
@@ -0,0 +1,69 @@
...
...TjBwR2l1NXVNRVFKaHd4S3N6T0JGWGtyRG8KQndJREFRQUJvMU13VVRBZEJnTlZIUTRFRmdRVUVXTTh0WjRsVng0MlFMeE8xSE03dGRCZUMvOHdId1lEVlIwagpCQmd3Rm9BVUVXTTh0WjRsVng0MlFMeE8xSE03dGRCZUMvOHdEd1lEVlIwVEFRSC9CQVV3QXdFQi96QU5CZ2txCmhraUc5dzBCQVFzRkFBT0NBZ0VBd3UwN2doaHRyUkNMZ0JuNENHbE5vVkxLRkx2SjQ3T21GUUc1eVp5MTEvdzkKaC9oblJMeDVCbk8wb1lZMmw4M0Z3OUozUWVIaThDTk84Ujc4NTRmRk56WGMxSEFZa1RWc1VZbE9wOW4wTTYwSgpKb3MyUFA5TldNTUpCcFg3Q3JQYUZRRjJoU09hb1NqaXZ1dVoxRnQrZVZEY3FWMjM0VFkrK25hYzBRc053RTYrCnpMai91TGFEM0xWUXh6Y2RuNWMrYVpSNjV0K3I0Q0RsWW1MMGVJY2RTeXF5UUtUWDlMaE1lVXQ3RUxJdEpnVkgKZDdSaDRuRU8rRDVhVEszNkZNSk9TM0VUL0Y1RksrT2QzVmgwMW9RTTJwR0dqQ3A3d2dMeWxNNTVaMWhsTnVXMQo4YWp2eHJDNWVVd2RkaTA4WWFBQjlpR1VRLzRmeUFmQkNkNjJZVVRXSUhib1NMKy84MEpySFZIQnhTaWZ0NmRFClI0SVBtbmtoakovOFcvK1g1WThvOFdrVUF3Zm00QWpOL3ZOZGUxWm1NSVFsSEZhQVRuWUJEaXBaOElSUnBndFEKcnc4d044U0NOV0plZHppdlVoYjdXdUdHbndCcDZ1Wjg1TDUzblN3SFBBS2Y3eGNhTXROVnpuZ1VJaXU1bm9PNwpZSFFzcG1xRVhQQzQ4NERmMHhTUWllUUhTWGxPUVFXS285QTQ5ZXM1NnV2ZGw0c0pTbW9uUTZMblgzV21sUGFGCmdxQSs2ZXZHanVqQmppaTBybncvUWpxY3NteHNtWU84alpGY21pZWpJL3AxUE1OalBpRWJjZGhrNFNrdDlxdnkKcDhvTXVLVzFLNHFpRUp0R1VOT3hkUEt2b1V1MWllQUtuY0FtdUhxdWNHWDBva2JrdmZjT0tYQjFoRi9kK1gwPQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
```
AlexanderYastrebov, line 67: Where do these values come from?

dataclients/kubernetes/testdata/ingressV1/tls/tls-missing-host.log
```diff
@@ -1 +1 @@
-level=info msg="No matching tls hosts found" kind=Ingress name=myapp-ingress ns=default
+level=info msg="No matching tls hosts found - tls hosts: \[foo.org bar.org\], ingress hosts: \[example.org\]" kind=Ingress name=myapp-ingress ns=default
```
AlexanderYastrebov, line 1: Missing new line

dataclients/kubernetes/routegroup.go
AlexanderYastrebov: The same is for Ingress - we plan to add ingress validator to dataclient, see https://github.com/zalando/skipper/pull/2757

dataclients/kubernetes/kube.go
rickhlx: Even though the error from generateTLSCertFromSecret is pretty explicit about the reason?

dataclients/kubernetes/testdata/ingressV1/tls/tls-host-mismatch.yaml
rickhlx: These are from existing fixtures, they’re self signed certs used just for testing.

dataclients/kubernetes/routegroup.go
rickhlx: Is this something we could handle in a separate PR to keep the scope limited? Happy to work on getting in validator for both Ingress and RouteGroup.

AlexanderYastrebov: Lets have it like it is now (log mismatch) and add tls validation for ingress and routegroup as a future improvement (makes sense after we land https://github.com/zalando/skipper/pull/2757).
@MustafaSaber FYI

dataclients/kubernetes/routegroup.go
```diff
@@ -478,7 +480,36 @@ func splitHosts(hosts []string, domains []string) ([]string, []string) {
...
+// addRouteGroupTLS compares the RouteGroup host list and the RouteGroup TLS host list
+// and adds the TLS secret to the registry if a match is found.
+func (r *routeGroups) addRouteGroupTLS(ctx *routeGroupContext, tls *definitions.RouteTLSSpec) {
+	// Host in the tls section need to explicitly match the host in the RouteGroup
+	hostlist := compareStringList(tls.Hosts, ctx.routeGroup.Spec.UniqueHosts())
+	if len(hostlist) == 0 {
+		ctx.logger.Infof("No matching tls hosts found - tls hosts: %s, routegroup hosts: %s", tls.Hosts, ctx.routeGroup.Spec.UniqueHosts())
+		return
+	} else if len(hostlist) != len(tls.Hosts) {
+		ctx.logger.Infof("Hosts in TLS and RouteGroup don't match: tls hosts: %s, routegroup hosts: %s", tls.Hosts, ctx.routeGroup.Spec.UniqueHosts())
```
AlexanderYastrebov, line 492: Warning

dataclients/kubernetes/routegroup.go
```diff
@@ -478,7 +480,36 @@ func splitHosts(hosts []string, domains []string) ([]string, []string) {
...
 }
 
-func (r *routeGroups) convert(s *clusterState, df defaultFilters, loggingEnabled bool) ([]*eskip.Route, error) {
+// addRouteGroupTLS compares the RouteGroup host list and the RouteGroup TLS host list
+// and adds the TLS secret to the registry if a match is found.
+func (r *routeGroups) addRouteGroupTLS(ctx *routeGroupContext, tls *definitions.RouteTLSSpec) {
+	// Host in the tls section need to explicitly match the host in the RouteGroup
+	hostlist := compareStringList(tls.Hosts, ctx.routeGroup.Spec.UniqueHosts())
+	if len(hostlist) == 0 {
+		ctx.logger.Infof("No matching tls hosts found - tls hosts: %s, routegroup hosts: %s", tls.Hosts, ctx.routeGroup.Spec.UniqueHosts())
```
AlexanderYastrebov, line 489: Error

dataclients/kubernetes/routegroup.go
szuecs: @AlexanderYastrebov no warning. There should be only error or info. Warning is only for deprecation.

dataclients/kubernetes/kube.go
AlexanderYastrebov: Logger logs namespace and ingress/routegroup so I guess its enough context to debug.

dataclients/kubernetes/testdata/ingressV1/tls/tls-host-mismatch.yaml
AlexanderYastrebov: Ah, ok, from existing ingress fixtures

dataclients/kubernetes/testdata/routegroups/tls/tls-missing-host.log
```diff
@@ -0,0 +1 @@
+level=info msg="No matching tls hosts found - tls hosts: \[foo.org\], routegroup hosts: \[example.org\]" kind=RouteGroup name=myapp ns=default
```
AlexanderYastrebov, line 1: new line please

dataclients/kubernetes/routegroup.go
AlexanderYastrebov: Ok.

dataclients/kubernetes/testdata/routegroups/tls/tls-missing-host.log
AlexanderYastrebov: Please check all files

AlexanderYastrebov:
Please add missing newlines at the end of test fixtures https://github.com/zalando/skipper/pull/2814#discussion_r1444642002 and fix log level https://github.com/zalando/skipper/pull/2814#discussion_r1444634080

LGTM but I think we should land https://github.com/szuecs/routegroup-client/pull/16 first.

AlexanderYastrebov:
> Why?
> I think first data type change then client change.
> We don't use the client here.

But we generate CRD there. I think we should settle all changes in the client, then copy CRD here, then we can land client PR.

AlexanderYastrebov:
:+1:

szuecs:
:+1:

szuecs:
thanks for your work!

szuecs:
I tagged the merged PR commit with https://github.com/szuecs/routegroup-client/releases/tag/v0.23.0

PR #2815, opened 2024-01-02
net: fix redis ZAdd

AlexanderYastrebov:
There was a breaking change to go-redis library https://github.com/redis/go-redis/pull/2818

I've checked how it handles interface{} command arguments and it converts int64 to string so this change should fix the problem.

Closes #2806


RomanZavodskikh:
:+1:

szuecs:
Let's wait some time to see if we get an answer from the referenced go-redis issue

szuecs:
https://github.com/redis/go-redis/issues/2817 has more communication.

AlexanderYastrebov:
The change in redis library was reverted #2830

PR #2819, opened 2024-01-02
refactor: sync.Mutex is fine as value object and needs no initialization

szuecs:
refactor: sync.Mutex is fine as value object and needs no initialization

### DIFF
--- a/secrets/certregistry/certregistry.go
+++ b/secrets/certregistry/certregistry.go
@@ -13,7 +13,7 @@ import (
 // ensuring synchronized access to them.
 type CertRegistry struct {
 	lookup map[string]*tls.Certificate
-	mx     *sync.Mutex
+	mx     sync.Mutex
 }
 
 // NewCertRegistry initializes the certificate registry.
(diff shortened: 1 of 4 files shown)


secrets/certregistry/certregistry.go
```diff
@@ -13,7 +13,7 @@ import (
 // ensuring synchronized access to them.
 type CertRegistry struct {
 	lookup map[string]*tls.Certificate
-	mx     *sync.Mutex
+	mx     sync.Mutex
```
AlexanderYastrebov, line 16: Lets move it above the lookup - I think there is a convention to put mutex above the fields it guards.
Also maybe rename `mx` to `mu`:
```
$ grep -o -I -E '[a-zA-Z0-9]+\s+sync.Mutex' -r | cut -d: -f2 | cut -d' ' -f1 | sort | uniq -c
      1 closeMx
      1 m
     16 mu
      3 mx
```

AlexanderYastrebov:
:+1:

secrets/certregistry/certregistry.go
szuecs: will add another PR

AlexanderYastrebov: Did not notice your comment before I created https://github.com/zalando/skipper/pull/2820

PR #2823, opened 2024-01-04
Use endpointregistry in the hotpath

RomanZavodskikh:
Using endpointregistry component is aimed to:

+ decrease complexity providing unified interface to access dynamic data (detected time, last seen time, inflight requests, etc.)
+ make this data host-wide instead of route-wide to broaden the amount of data available to users (fadein, loadbalancer, etc.)


proxy/proxy.go
```diff
@@ -467,6 +467,33 @@ func setRequestURLForDynamicBackend(u *url.URL, stateBag map[string]interface{})
 	}
 }
 
+func normalizeSchemeHost(s, h string) (string, string, error) {
```
RomanZavodskikh, line 470: There were 2 copies of this function before PR. After this PR in the current state there will be 4(!) copies of it. I better find some way to reuse it.

proxy/proxy.go
```diff
@@ -845,10 +877,9 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 		return res, nil
 	}
 
-	if endpoint != nil {
-		endpoint.Metrics.IncInflightRequest()
-		defer endpoint.Metrics.DecInflightRequest()
-	}
+	entry := p.registry.GetMetrics(host)
```
RomanZavodskikh, line 880: I think it's better not to think about what backend type it is, but add the data available at this point regardless of backend type at all.

routing/endpointregistry.go
```diff
@@ -81,19 +85,59 @@ type RegistryOptions struct {
 	LastSeenTimeout time.Duration
 }
 
+func NormalizeSchemeHost(s, h string) (string, string, error) {
```
szuecs, line 88: likely this wants to be in our net/ package

AlexanderYastrebov: OTOH this used to handle eskip endpoints, i.e. its not a generic helper so maybe its better to keep it inside of routing

proxy/proxy.go
RomanZavodskikh: Resolved in scope of https://github.com/zalando/skipper/pull/2824

proxy/proxy.go
RomanZavodskikh: Resolved in scope of https://github.com/zalando/skipper/pull/2824

proxy/proxy.go
```diff
@@ -512,7 +512,7 @@ func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHea
 
 	rr, err := http.NewRequestWithContext(requestContext, r.Method, u.String(), body)
 	if err != nil {
-		return nil, endpoint, err
+		return nil, u.Host, err
```
AlexanderYastrebov, line 515: Lets just `return nil, "", err`. It was a bad idea to return `return nil, endpoint, err` since caller never checks endpoint if there is an error.

proxy/proxy.go
```diff
@@ -543,7 +543,7 @@ func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHea
 		rr = forwardToProxy(r, rr)
 	}
 
-	return rr, endpoint, nil
+	return rr, u.Host, nil
```
AlexanderYastrebov, line 546: Here `endpoint` was not nil only for loadbalanced endpoint and it is only used to track inflight requests. Now we want to have metrics key for network and loadbalanced backends but I am not sure we want to have it for dynamic backend (at least now).

If we want to have it for dynamic backend we'd need to normalize it as well.

RomanZavodskikh: At the moment endpointregistry data essentially influences only the way how Skipper's load balancing algorithm chooses the backend address to send request to, the data for LBBackend could be modified from LBBackend or NetworkBackend. So, probably, it does not make much sense to normalize dynamic backends' hosts.

Firstly, it will create some additional and not straightforward changes.
Secondly, it will increase CPU and/or memory used in the hotpath.

Therefore I better retrieve metrics only for LBBackend and NetworkBackend.

szuecs: I would also postpone dynamic backend

proxy/proxy.go
```diff
@@ -484,7 +484,7 @@ func selectEndpoint(ctx *context, registry *routing.EndpointRegistry) *routing.L
 
 // creates an outgoing http request to be forwarded to the route endpoint
 // based on the augmented incoming request
-func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHeaders bool, registry *routing.EndpointRegistry) (*http.Request, *routing.LBEndpoint, error) {
+func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHeaders bool, registry *routing.EndpointRegistry) (*http.Request, string, error) {
 	var endpoint *routing.LBEndpoint
```
AlexanderYastrebov, line 488: This could be local now
```go
endpoint := selectEndpoint(ctx, registry)
```

routing/routing.go
```diff
@@ -152,22 +152,37 @@ type RouteFilter struct {
...
+}
+
+func NewLBMetrics(o *LBMeticsOptions) *LBMetrics {
+	return &LBMetrics{
+		endpointRegistryEntry: o.EndpointRegistryEntry,
+	}
 }
 
 // IncInflightRequest increments the number of outstanding requests from the proxy to a given backend.
 func (m *LBMetrics) IncInflightRequest() {
```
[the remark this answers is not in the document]
AlexanderYastrebov: I am wondering if we should return LBMetrics instead of Metrics interface  from EndpointRegistry
```go
func (r *EndpointRegistry) GetMetrics(key string) *LBMetrics
```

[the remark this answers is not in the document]
RomanZavodskikh: That's possible, see https://github.com/zalando/skipper/pull/2841.
However, I stick to merging the PR 2823 as it is at the moment, because non-breaking change add too many boilerplate code: a lot of LBMetrics methods which essentially only call routing.Metrics methods:
https://github.com/zalando/skipper/pull/2841/files#diff-f58538a9de4e0a1e57e1b56ca2ed06fb8e8a149f68f0f5ea17503b3504ccae72R169-R190

loadbalancer/fadein_test.go
```diff
@@ -329,8 +329,7 @@ func benchmarkFadeIn(
 		registry := routing.NewEndpointRegistry(routing.RegistryOptions{})
 		for i := range eps {
 			route.LBEndpoints = append(route.LBEndpoints, routing.LBEndpoint{
-				Host:     eps[i],
-				Detected: detectionTimes[i],
```
RomanZavodskikh, line 333: minor: I need to add `Metrics: ...` here too.

proxy/proxy.go
```diff
@@ -469,6 +469,9 @@ func setRequestURLForDynamicBackend(u *url.URL, stateBag map[string]interface{})
 
 func selectEndpoint(ctx *context, registry *routing.EndpointRegistry) *routing.LBEndpoint {
 	rt := ctx.route
+	if rt.BackendType != eskip.LBBackend {
+		return nil
+	}
```
AlexanderYastrebov, line 474: This way we have type check in two places.
Lets make it like it was before, i.e. selectEndpoint assumes correct type.

proxy/proxy.go
```diff
@@ -543,7 +545,14 @@ func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHea
...
 	}
 
-	return rr, endpoint, nil
+	var endpointMetrics routing.Metrics
+	switch rt.BackendType {
+	case eskip.LBBackend:
+		endpointMetrics = endpoint.Metrics
+	case eskip.NetworkBackend:
+		endpointMetrics = registry.GetMetrics(rt.Host)
+	}
```
AlexanderYastrebov, line 554: Lets do it in the switch above

proxy/proxy.go
```diff
@@ -484,8 +487,8 @@ func selectEndpoint(ctx *context, registry *routing.EndpointRegistry) *routing.L
 
 // creates an outgoing http request to be forwarded to the route endpoint
 // based on the augmented incoming request
-func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHeaders bool, registry *routing.EndpointRegistry) (*http.Request, *routing.LBEndpoint, error) {
-	var endpoint *routing.LBEndpoint
+func mapRequest(ctx *context, requestContext stdlibcontext.Context, removeHopHeaders bool, registry *routing.EndpointRegistry) (*http.Request, routing.Metrics, error) {
+	endpoint := selectEndpoint(ctx, registry)
```
AlexanderYastrebov, line 491: Lets move it into case eskip.LBBackend:

loadbalancer/algorithm_test.go
```diff
@@ -56,6 +58,7 @@ func TestSelectAlgorithm(t *testing.T) {
 
 	t.Run("LB route with explicit round-robin algorithm", func(t *testing.T) {
 		p := NewAlgorithmProvider()
+		epregistry := routing.NewEndpointRegistry(routing.RegistryOptions{})
```
AlexanderYastrebov, line 61: minor: `endpointRegistry` or `registry` for consistent naming

AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

szuecs:
:+1:

routing/endpointregistry.go
```diff
@@ -86,14 +86,21 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
...
+				epi := &route.LBEndpoints[i]
+				epi.Metrics = r.GetMetrics(epi.Host)
+				if epi.Metrics.DetectedTime().IsZero() {
+					epi.Metrics.SetDetected(now)
 				}
 
-				metrics.SetLastSeen(now)
+				epi.Metrics.SetLastSeen(now)
 			}
+		} else {
```
AlexanderYastrebov, line 98: @RomanZavodskikh Here it should be `else if route.BackendType == eskip.NetworkBackend` because there are other types for which `route.Host` will be empty and `r.GetMetrics(route.Host)` will create an entry for empty key.

Please add this condition and a test for `func (r *EndpointRegistry) Do(routes []*Route) []*Route`

RomanZavodskikh: Very nice catch. :-)
Addressed it here: https://github.com/zalando/skipper/pull/2869

PR #2824, opened 2024-01-04
Write normalized scheme and host to routing.Route fields

RomanZavodskikh:
This change is needed for proxy to easily have normalized host and use it in endpointregistry for dynamic host-wide data to be easily fetchable.

This PR should make changes in #2823 easier.

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -1,9 +1,11 @@
 package net
 
 import (
+	"fmt"
 	"net"
 	"net/http"
 	"net/netip"
+	"net/url"
 	"strings"
 
 	"go4.org/netipx"
@@ -153,4 +155,49 @@ func ParseIPCIDRs(cidrs []string) (*netipx.IPSet, error) {
 	}
 
 	return ips, nil
+}
+
+// SchemeHost parses URI string (without #fragment part) and returns schema used in this URI as first return value and
+// host[:port] part as second return value. Port is never omitted for HTTP(S): if no port is specified in URI, default port for given
+// schema is used. If URI is invalid, error is returned.
+func SchemeHost(input string) (string, string, error) {
+	u, err := url.ParseRequestURI(input)
+	if err != nil {
+		return "", "", err
+	}
+	if u.Scheme == "" {
+		return "", "", fmt.Errorf(`parse %q: missing scheme`, input)
+	}
+	if u.Host == "" {
+		return "", "", fmt.Errorf(`parse %q: missing host`, input)
+	}
+
+	// endpoint address cannot contain path, the rest is not case sensitive
+	s, h := strings.ToLower(u.Scheme), strings.ToLower(u.Host)
+
+	hh, p, err := net.SplitHostPort(h)
+	if err != nil {
+		if strings.Contains(err.Error(), "missing port") {
+			// Trim is needed to remove brackets from IPv6 addresses, JoinHostPort will add them in case of any IPv6 address,
+			// so we need to remove them to avoid duplicate pairs of brackets.
+			h = strings.Trim(h, "[]")
+			switch s {
+			case "http":
+				p = "80"
+			case "https":
+				p = "443"
+			default:
+				p = ""
+			}
+		} else {
+			return "", "", err
+		}
+	} else {
+		h = hh
+	}
+
+	if p != "" {
+		h = net.JoinHostPort(h, p)
+	}
+	return s, h, nil
 }
--- a/net/net_test.go
+++ b/net/net_test.go
@@ -1,13 +1,40 @@
 package net
 
 import (
+	"fmt"
 	"net"
 	"net/http"
 	"net/netip"
+	"path/filepath"
 	"reflect"
+	"runtime"
 	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
+
+type tc[T any] struct {
+	location string
+	in       T
+}
+
+// https://github.com/golang/go/issues/52751
+func testCase[T any](in T) tc[T] {
+	_, file, line, _ := runtime.Caller(1)
+	location := fmt.Sprintf("%s:%d", filepath.Base(file), line)
+	return tc[T]{location: location, in: in}
+}
+
+func (tc *tc[T]) logLocation(t *testing.T) {
+	t.Helper()
+	t.Cleanup(func() {
+		t.Helper()
+		if t.Failed() {
+			t.Logf("Test case location: %s", tc.location)
+		}
+	})
+}
 
 func TestRemoteAddr(t *testing.T) {
 	for _, tt := range []struct {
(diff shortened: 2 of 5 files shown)


routing/datasource.go
```diff
@@ -482,6 +482,10 @@ func processRouteDef(cpm map[string]PredicateSpec, fr filters.Registry, def *esk
 		return nil, err
 	}
 
+	scheme, host, err = NormalizeSchemeHost(scheme, host)
+	if err != nil {
+		return nil, err
+	}
```
AlexanderYastrebov, line 488: This can be moved inside of splitBackend.

Then it looks like that all three call sites first parse a string via url.Parse or url.ParseRequestURI. Reading the docs https://pkg.go.dev/net/url#ParseRequestURI I think ParseRequestURI is the right one to use since we expect absolute URI.

So I propose that we create a helper like
```go
func SchemeHost(u string) (scheme, host string, error)
```
and use it everywhere.

AlexanderYastrebov: Please also add a godoc

net/net.go
```diff
@@ -154,3 +156,47 @@ func ParseIPCIDRs(cidrs []string) (*netipx.IPSet, error) {
...
+// schema is used. If URI is invalid or port could not be detected, error is returned.
+func SchemeHost(input string) (string, string, error) {
+	u, err := url.ParseRequestURI(input)
+	if err != nil {
+		return "", "", err
+	}
+
+	// endpoint address cannot contain path, the rest is not case sensitive
+	s, h := strings.ToLower(u.Scheme), strings.ToLower(u.Host)
+	if h == "" {
```
AlexanderYastrebov, line 171: Please move empty check before ToLower.
Also looks like scheme can be empty?

net/net.go
```diff
@@ -154,3 +156,47 @@ func ParseIPCIDRs(cidrs []string) (*netipx.IPSet, error) {
...
+		return "", "", errors.Errorf(`parse "%s": empty host`, input)
```
AlexanderYastrebov, line 172: use %q instead of "%s"

net/net.go
```diff
@@ -154,3 +156,47 @@ func ParseIPCIDRs(cidrs []string) (*netipx.IPSet, error) {
...
+	case p == "" && s == "http":
+		p = "80"
+	case p == "" && s == "https":
+		p = "443"
+	case p == "" && s == "fastcgi":
+		p = "9000"
+	}
+
+	if p == "" {
+		return "", "", errors.Errorf(`parse "%s": empty port`, input)
```
AlexanderYastrebov, line 197: Please use "missing" instead of "empty in all errors

net/net_test.go
```diff
@@ -236,3 +237,61 @@ func TestIPNetsDoNotContain(t *testing.T) {
...
+			scheme, host, err := SchemeHost(tt.input)
+			if tt.err == nil && err != nil {
+				t.Errorf("unexpected error: '%v'", err)
+			}
+			if tt.err != nil && err == nil {
+				t.Errorf("failed to fail")
+			}
+			if tt.err != nil && err != nil && tt.err.Error() != err.Error() {
+				t.Errorf("unexpected error: '%v'", err)
+			}
```
AlexanderYastrebov, line 288: This would be so much cleaner with assert:
```go
if tc.err != "" {
     assert.EqualError(t, err, tc.err)
} else {
     if assert.NoError(t, err) {
          assert.Equal(t, tc.scheme, scheme)
          assert.Equal(t, tc.host, host)
     }
}
```

net/net_test.go
```diff
@@ -236,3 +237,61 @@ func TestIPNetsDoNotContain(t *testing.T) {
 		})
 	}
 }
+
+func TestSchemeHost(t *testing.T) {
+	for _, tt := range []struct {
```
AlexanderYastrebov, line 242: usually we name it `ti` (test item) or `tc` (test case)

net/net_test.go
```diff
@@ -236,3 +237,61 @@ func TestIPNetsDoNotContain(t *testing.T) {
...
+		input  string
+		scheme string
+		host   string
+		err    error
```
AlexanderYastrebov, line 246: I think it would be easier to use string here

net/net.go
RomanZavodskikh: If scheme is empty, the https://pkg.go.dev/net/url#ParseRequestURI above will fail, and this line of code won't be executed:

```
$ cat main.go 
package main

import (
	"fmt"
	"net/url"
)

func main() {
	u, err := url.ParseRequestURI("example.com")
	fmt.Printf("u: %v, err: %v\n", u, err)
}
$ go build main.go 
$ ./main 
u: <nil>, err: parse "example.com": invalid URI for request
```

RomanZavodskikh: However, I will add check for empty scheme as well. Just to be on the safe side.

AlexanderYastrebov:
:+1:

net/net.go
```diff
@@ -154,3 +156,46 @@ func ParseIPCIDRs(cidrs []string) (*netipx.IPSet, error) {
...
+// SchemeHost parses URI string (without #fragment part) and returns schema used in this URI as first return value and
+// host[:port] part as second return value. Port is never omitted for HTTP(S): if no port is specified in URI, default port for given
+// schema is used. If URI is invalid, error is returned.
+func SchemeHost(input string) (string, string, error) {
+	u, err := url.ParseRequestURI(input)
+	if err != nil {
+		return "", "", err
+	}
+	if u.Scheme == "" {
+		return "", "", errors.Errorf(`parse %q: missing scheme`, input)
```
szuecs, line 169: errors.Errorf -> fmt.Errorf()

net/net.go
```diff
@@ -154,3 +156,46 @@ func ParseIPCIDRs(cidrs []string) (*netipx.IPSet, error) {
...
+	}
+	if u.Host == "" {
+		return "", "", errors.Errorf(`parse %q: missing host`, input)
```
szuecs, line 172: errors.Errorf -> fmt.Errorf()

net/net.go
```diff
@@ -4,8 +4,10 @@ import (
 	"net"
 	"net/http"
 	"net/netip"
+	"net/url"
 	"strings"
 
+	"github.com/pkg/errors"
```
szuecs, line 10: that's not needed

net/net.go
```diff
@@ -154,3 +156,46 @@ func ParseIPCIDRs(cidrs []string) (*netipx.IPSet, error) {
...
+
+		p = ""
+	} else {
+		h = hh
+	}
+
+	switch {
+	case p == "" && s == "http":
+		p = "80"
+	case p == "" && s == "https":
```
szuecs, line 193: I guess `"fastcgi"` makes also sense to support

net/net.go
```diff
@@ -154,3 +156,46 @@ func ParseIPCIDRs(cidrs []string) (*netipx.IPSet, error) {
...
+		if !strings.Contains(err.Error(), "missing port") {
+			return "", "", err
+		}
+
+		p = ""
+	} else {
+		h = hh
+	}
+
+	switch {
```
szuecs, line 190: maybe better:
if p == "" {
  switch s {
  case "http":
    p = "80"
  case "https":
    p = "441"
}

net/net.go
RomanZavodskikh: Added them:
https://github.com/zalando/skipper/pull/2824/files#diff-bacdef5a036ec850a0efe0adc1f0288bc03339074b068f5b9b87e7d8532aaa5bR314-R331

net/net.go
RomanZavodskikh: Idea of checking p before switch is nice, I applied it:
https://github.com/zalando/skipper/pull/2824/files#diff-85b6d38e7e77f99d0018695343944555050e58f3b190ffa74041fdd702e4c30aR191-R198

AlexanderYastrebov:
:+1:

net/net.go
```diff
@@ -154,3 +156,49 @@ func ParseIPCIDRs(cidrs []string) (*netipx.IPSet, error) {
...
+	if u.Host == "" {
+		return "", "", fmt.Errorf(`parse %q: missing host`, input)
+	}
+
+	// endpoint address cannot contain path, the rest is not case sensitive
+	s, h := strings.ToLower(u.Scheme), strings.ToLower(u.Host)
+
+	hh, p, err := net.SplitHostPort(h)
+	if err != nil {
+		// what is the actual right way of doing this, considering IPv6 addresses, too?
```
szuecs, line 181: What does the question mean?
We should be ipv6 safe

szuecs (changes requested):

szuecs:
:-1:

szuecs:
As discussed in person, can be a separate pr

net/net.go
AlexanderYastrebov: This is a copy of existing implementation.
I think what is meant here is that it does not handle IPv6 without port case.

From https://www.ietf.org/rfc/rfc3986.txt
```
Appendix A.  Collected ABNF for URI

   URI           = scheme ":" hier-part [ "?" query ] [ "#" fragment ]

   hier-part     = "//" authority path-abempty
                 / path-absolute
                 / path-rootless
                 / path-empty

   URI-reference = URI / relative-ref

   absolute-URI  = scheme ":" hier-part [ "?" query ]

   relative-ref  = relative-part [ "?" query ] [ "#" fragment ]

   relative-part = "//" authority path-abempty
                 / path-absolute
                 / path-noscheme
                 / path-empty

   scheme        = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )

   authority     = [ userinfo "@" ] host [ ":" port ]
   userinfo      = *( unreserved / pct-encoded / sub-delims / ":" )
   host          = IP-literal / IPv4address / reg-name
   port          = *DIGIT

   IP-literal    = "[" ( IPv6address / IPvFuture  ) "]"

   IPvFuture     = "v" 1*HEXDIG "." 1*( unreserved / sub-delims / ":" )

   IPv6address   =                            6( h16 ":" ) ls32
                 /                       "::" 5( h16 ":" ) ls32
                 / [               h16 ] "::" 4( h16 ":" ) ls32
                 / [ *1( h16 ":" ) h16 ] "::" 3( h16 ":" ) ls32
                 / [ *2( h16 ":" ) h16 ] "::" 2( h16 ":" ) ls32
                 / [ *3( h16 ":" ) h16 ] "::"    h16 ":"   ls32
                 / [ *4( h16 ":" ) h16 ] "::"              ls32
                 / [ *5( h16 ":" ) h16 ] "::"              h16
                 / [ *6( h16 ":" ) h16 ] "::"

   h16           = 1*4HEXDIG
   ls32          = ( h16 ":" h16 ) / IPv4address
   IPv4address   = dec-octet "." dec-octet "." dec-octet "." dec-octet
```
It follows that `https://[::0]/foo/bar` IPv6 URI is valid.
The host `[::0]` would be and `net.SplitHostPort` will fail due to missing port.

I think we should trim `[]` in case of `missing port in address` error, see https://go.dev/play/p/iiRpN82LMnA

AlexanderYastrebov:
:+1:

net/net.go
szuecs: Hmm Trim feels like a poor men's solution. I would guess that there is a netip or net package function that covers the problem.

AlexanderYastrebov: Trim is super fast, from "missing port in address" we know address is ok but the port is missing.

szuecs: Maybe https://pkg.go.dev/net/netip#ParseAddr ?

AlexanderYastrebov:
> As discussed in person, can be a separate pr
@szuecs then you should approve :)

net/net.go
AlexanderYastrebov: > address is ok

Of course we don't validate hostname/address so this will pass `https://---invalid---:80/foo/bar` even though hostname is invalid. I don't think we should validate hostname/address here, it should fail later during request.

RomanZavodskikh: Yes, we can use `netip.ParseAddr` to handle IPv6's properly. However, how should `SchemeHost` behave in case of DNS names? I was thinking it should consider DNS names valid and return host the same way it does with IPv4's and IPv6's.

AlexanderYastrebov: Lets add test cases for ipv4 and ipv6 with and without port

RomanZavodskikh: Added them.

net/net.go
```diff
@@ -154,3 +156,52 @@ func ParseIPCIDRs(cidrs []string) (*netipx.IPSet, error) {
 
 	return ips, nil
 }
+
+// SchemeHost parses URI string (without #fragment part) and returns schema used in this URI as first return value and
+// host[:port] part as second return value. Port is never omitted for HTTP(S): if no port is specified in URI, default port for given
+// schema is used. If URI is invalid, error is returned.
+// SchemeHost does not work with IPv6 addresses.
```
AlexanderYastrebov, line 163: `// SchemeHost does not work with IPv6 addresses.` this is outdated now

net/net.go
```diff
@@ -154,3 +156,52 @@ func ParseIPCIDRs(cidrs []string) (*netipx.IPSet, error) {
...
+	if u.Host == "" {
+		return "", "", fmt.Errorf(`parse %q: missing host`, input)
+	}
+
+	// endpoint address cannot contain path, the rest is not case sensitive
+	s, h := strings.ToLower(u.Scheme), strings.ToLower(u.Host)
+
+	hh, p, err := net.SplitHostPort(h)
+	if err != nil {
+		// what is the actual right way of doing this, considering IPv6 addresses, too?
```
AlexanderYastrebov, line 181: I guess we can remove this comment

AlexanderYastrebov:
:+1:

net/net.go
```diff
@@ -154,3 +156,50 @@ func ParseIPCIDRs(cidrs []string) (*netipx.IPSet, error) {
...
+	// endpoint address cannot contain path, the rest is not case sensitive
+	s, h := strings.ToLower(u.Scheme), strings.ToLower(u.Host)
+
+	hh, p, err := net.SplitHostPort(h)
+	if err != nil {
+		if strings.Contains(err.Error(), "missing port") {
+			// Trim is needed to remove brackets from IPv6 addresses, JoinHostPort will add them in case of any IPv6 address,
+			// so we need to remove them to avoid duplicate pairs of brackets.
+			h = strings.Trim(h, "[]")
+			p = ""
```
szuecs, line 184: Maybe this should also check scheme for well-known ports:
scheme == "http" -> port=80 
scheme == "https" -> port=443

Basically we could move the code from line 193 to here because there is no other way to reach p="" at line 192 other than this branch.

RomanZavodskikh: Yes, it makes sense.

AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

szuecs:
:+1:

net/net.go
```diff
@@ -154,3 +156,48 @@ func ParseIPCIDRs(cidrs []string) (*netipx.IPSet, error) {
...
+	}
+	if u.Scheme == "" {
+		return "", "", fmt.Errorf(`parse %q: missing scheme`, input)
+	}
+	if u.Host == "" {
+		return "", "", fmt.Errorf(`parse %q: missing host`, input)
+	}
+
+	// endpoint address cannot contain path, the rest is not case sensitive
+	s, h := strings.ToLower(u.Scheme), strings.ToLower(u.Host)
```
MustafaSaber, line 176: minor: I think variable names here could be more descriptive 

```go
scheme, host := strings.ToLower(u.Scheme), strings.ToLower(u.Host)
isolatedHost, port, err := net.SplitHostPort(h)
```
I feel like the current variable names is easy to forget and harder to debug later

PR #2827, opened 2024-01-05
feature: logBody filter 

szuecs:
feature: logBody() filter to log request or response body while streaming
doc: add logBody()

fixes: https://github.com/zalando/skipper/issues/2282

first part to replace https://github.com/zalando/skipper/pull/2428

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -709,7 +709,7 @@ The histogramResponseLatency adds latency to responses according to the histogra
 
 ### logHeader
 
-The logHeader filter prints the request line and the header, but not the body, to
+The `logHeader` filter prints the request line and the header, but not the body, to
 stderr. Note that this filter should be used only in diagnostics setup and with care,
 since the request headers may contain sensitive data, and they also can explode the
 amount of logs. Authorization headers will be truncated in request and
@@ -728,6 +728,32 @@ Example:
 * -> logHeader("request") -> "https://www.example.org";
 * -> logHeader("response") -> "https://www.example.org";
 * -> logHeader("request", "response") -> "https://www.example.org";
+```
+
+### logBody
+
+The `logBody` filter logs the request or response body in chunks while
+streaming. Chunks start with `logBody("request") $flowid: ` or
+`logBody("response") $flowid: `, such that you can find all chunks
+belonging to a given flow. See also [flowId()](#flowid) filter.
+
+Note that this filter should be used only in diagnostics setup and
+with care, since the request and response body may contain sensitive
+data. Logs can also explode in the amount of bytes, so you have to
+choose a limit. You can log request or response bodies. This filter
+has close to no overhead other than the I/O created by the logger.
+
+Parameters:
+
+* type: "request" or "response" (string)
+* limit: maximum number of bytes to log (int)
+
+Example:
+
+```
+* -> logBody("request", 1024) -> "https://www.example.org";
+* -> logBody("response", 1024) -> "https://www.example.org";
+* -> logBody("request", 1024) -> logBody("response", 1024) -> "https://www.example.org";
 ```
 
 ## Timeout
--- a/filters/diag/logbody.go
+++ b/filters/diag/logbody.go
@@ -1,0 +1,130 @@
+package diag
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/filters/flowid"
+)
+
+type logBody struct {
+	limit    int
+	request  bool
+	response bool
+}
+
+// NewLogBody creates a filter specification for the 'logBody()' filter.
+func NewLogBody() filters.Spec { return logBody{} }
+
+// Name returns the logBody filtern name.
+func (logBody) Name() string {
+	return filters.LogBodyName
+}
+
+func (logBody) CreateFilter(args []interface{}) (filters.Filter, error) {
+	var (
+		request  = false
+		response = false
+	)
+
+	if len(args) != 2 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	opt, ok := args[0].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+	switch opt {
+	case "response":
+		response = true
+	case "request":
+		request = true
+	default:
+		return nil, fmt.Errorf("failed to match %q: %w", opt, filters.ErrInvalidFilterParameters)
+	}
+
+	limit, ok := args[1].(float64)
+	if !ok || float64(int(limit)) != limit {
+		return nil, fmt.Errorf("failed to convert to int: %w", filters.ErrInvalidFilterParameters)
+	}
+
+	return &logBody{
+		limit:    int(limit),
+		request:  request,
+		response: response,
+	}, nil
+}
+
+func (lb *logBody) Request(ctx filters.FilterContext) {
+	if !lb.request {
+		return
+	}
+
+	req := ctx.Request()
+	if req.Body != nil {
+		req.Body = newLogBodyStream(
+			lb.limit,
+			func(chunk []byte) {
+				ctx.Logger().Infof(
+					`logBody("request") %s: %s`,
+					req.Header.Get(flowid.HeaderName),
+					chunk)
+			},
+			req.Body,
+		)
+	}
+}
+
+func (lb *logBody) Response(ctx filters.FilterContext) {
+	if !lb.response {
+		return
+	}
+
+	rsp := ctx.Response()
+	if rsp.Body != nil {
+		rsp.Body = newLogBodyStream(
+			lb.limit,
+			func(chunk []byte) {
+				ctx.Logger().Infof(
+					`logBody("response") %s: %s`,
+					ctx.Request().Header.Get(flowid.HeaderName),
+					chunk)
+			},
+			rsp.Body,
+		)
+	}
+}
+
+type logBodyStream struct {
+	left  int
+	f     func([]byte)
+	input io.ReadCloser
+}
+
+func newLogBodyStream(left int, f func([]byte), rc io.ReadCloser) io.ReadCloser {
+	return &logBodyStream{
+		left:  left,
+		f:     f,
+		input: rc,
+	}
+}
+
+func (lb *logBodyStream) Read(p []byte) (n int, err error) {
+	if lb.left <= 0 {
+		return lb.input.Read(p)
+	}
+
+	n, err = lb.input.Read(p)
+	if n > 0 {
+		lb.f(p[:min(n, lb.left)])
+	}
+	lb.left -= n
+
+	return n, err
+}
+
+func (lb *logBodyStream) Close() error {
+	return lb.input.Close()
+}
--- a/filters/builtin/builtin.go
+++ b/filters/builtin/builtin.go
@@ -180,6 +180,7 @@ func Filters() []filters.Spec {
 		diag.NewAbsorb(),
 		diag.NewAbsorbSilent(),
 		diag.NewLogHeader(),
+		diag.NewLogBody(),
 		diag.NewUniformRequestLatency(),
 		diag.NewUniformResponseLatency(),
 		diag.NewNormalRequestLatency(),
--- a/filters/diag/logbody_test.go
+++ b/filters/diag/logbody_test.go
@@ -1,0 +1,590 @@
+package diag
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/filters/flowid"
+	"github.com/zalando/skipper/proxy/proxytest"
+)
+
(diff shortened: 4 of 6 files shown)


filters/diag/logbody.go
```diff
@@ -0,0 +1,86 @@
...
+	return filters.LogBodyName
+}
+
+func (logBody) CreateFilter(args []interface{}) (filters.Filter, error) {
+	var (
+		request  = false
+		response = false
+	)
+
+	if len(args) == 0 {
```
AlexanderYastrebov, line 32: or > 2

io/log_stream.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+}
+
+func (lb *logBody) Read(p []byte) (int, error) {
+	n, err := lb.input.Read(p)
+	if n > 0 {
+		lb.log("%s%s", lb.fmtstr, p)
+	}
+	return n, err
+}
```
AlexanderYastrebov, line 30: This is really all it is. Why do we need a new package?

docs/reference/filters.md
```diff
@@ -730,6 +730,27 @@ Example:
 * -> logHeader("request", "response") -> "https://www.example.org";
 ```
 
+### logBody
+
+The `logBody` filter logs the request or response body while
```
AlexanderYastrebov, line 735: Should we mention that it may log body in chunks, i.e. multiple log lines per requets/response?

docs/reference/filters.md
```diff
@@ -730,6 +730,27 @@ Example:
...
+Parameters:
+
+* "request" or "response" (string varargs)
+
+Example:
+
+```
+* -> logHeader("request") -> "https://www.example.org";
+* -> logHeader("response") -> "https://www.example.org";
+* -> logHeader("request", "response") -> "https://www.example.org";
```
AlexanderYastrebov, line 751: logBody

docs/reference/filters.md
```diff
@@ -730,6 +730,27 @@ Example:
...
+The `logBody` filter logs the request or response body while
+streaming.  Note that this filter should be used only in diagnostics
+setup and with care, since the request and response body may contain
+sensitive data, and they also can explode the amount of logs. You can
+log request or response bodies. This filter has close to no overhead
+other than the I/O created by the logger.
+
+Parameters:
+
+* "request" or "response" (string varargs)
```
AlexanderYastrebov, line 744: Maybe we should support single argument for simplicity and let users chain `logBody("request") -> logBody("response")` if they need.
Also maybe we should add second limit parameter to cap number of bytes logged (and log all when it is zero).

szuecs: Do you see that we limit it? Why should chaining not work?

AlexanderYastrebov: We may also want to add default limit as a flag and spec option in the future.

filters/diag/logbody.go
szuecs: done

docs/reference/filters.md
szuecs: done

docs/reference/filters.md
szuecs: added a test

io/log_stream.go
szuecs: it's not a lot of code but it makes sense with more context. It's also reusable if someone wants to plug this into their code as it just requires a logger func with a format string.
I think also little code is good. :)

io/log_stream.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (lb *logBody) Close() error {
+	return lb.input.Close()
+}
+
+func LogBody(ctx context.Context, fmtstr string, log func(format string, args ...interface{}), rc io.ReadCloser) io.ReadCloser {
```
AlexanderYastrebov, line 36: Lets just pass `func([]byte)` callback and let caller decide what and how to log

docs/reference/filters.md
szuecs: yes done

io/log_stream.go
szuecs: why? It's such a common interface function for loggers!

io/log_stream.go
AlexanderYastrebov: If someone needs this they can copy-poste 6 lines, no?

io/log_stream.go
AlexanderYastrebov: Maybe you want to count bytes. Also the whole point of format string is to defer evaluation of arguments until logging and not evaluate them if log level is disabled. In your case
```go
		req.Body = io.LogBody(
			req.Context(),
			fmt.Sprintf(`logBody("request") %s: `, req.Header.Get(flowid.HeaderName)),
			ctx.Logger().Infof,
			req.Body,
		)
```
you always spend time creating format string even if there will be no logging (e.g. when log level set to error).
Compare to
```go
		req.Body = io.LogBody(
			req.Context(),
			func(chunk []byte) {
				ctx.Logger().Infof(`logBody("request") %s: %s`, req.Header.Get(flowid.HeaderName), chunk)
			},
			req.Body,
		)
```

io/log_stream.go
```diff
@@ -0,0 +1,38 @@
+package io
+
+import (
+	"context"
+	"io"
+)
+
+type logBody struct {
+	ctx    context.Context
```
AlexanderYastrebov, line 9: Is this used?

szuecs: no, but maybe we should stop processing if it's canceled, wdyt?

docs/reference/filters.md
szuecs: Ok makes sense then I will restrict it to one argument.

szuecs: done

io/log_stream.go
szuecs: ok got it

szuecs: changed

io/log_stream.go
szuecs: I changed to inspect ctx.Done() before reading

szuecs: added testcase to test cancelation

AlexanderYastrebov: Why do we care about context here? we don't care elsewhere and I think we should not use it here for simplicity. Also for response path its unclear which context to use so one has to resort for context.Background, i.e. its unused

io/log_stream.go
AlexanderYastrebov: What is your argument to introduce whole new exported package before we have any other use for it?

szuecs: It's actually in use @AlexanderYastrebov :)

io/log_stream.go
szuecs: We can use the request.Context() form FilterContext, if that's canceled as in client canceled case we can stop processing.

docs/reference/filters.md
```diff
@@ -730,6 +730,31 @@ Example:
...
+
+Example:
+
+```
+* -> logBody("request") -> "https://www.example.org";
+* -> logBody("response") -> "https://www.example.org";
+* -> logBody("request", "response") -> "https://www.example.org";
```
AlexanderYastrebov, line 755: This should be changed to chained example

filters/diag/logbody.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	opt, ok := args[0].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+	switch strings.ToLower(opt) {
+	case "response":
+		response = true
+	case "request":
```
AlexanderYastrebov, line 42: default return err

filters/diag/logbody.go
```diff
@@ -0,0 +1,90 @@
...
+
+	if len(args) != 1 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	opt, ok := args[0].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+	switch strings.ToLower(opt) {
```
AlexanderYastrebov, line 39: I think we should not lower it but rather fail in default case

io/log_stream.go
AlexanderYastrebov: I am not sure its a good idea to introduce a package names as in stdlib. This will always catch an eye or need to be imported with alias like `sio` or `skpio`. Maybe we should name it like that right away.

io/log_stream.go
```diff
@@ -0,0 +1,41 @@
...
+
+func (lb *logBody) Read(p []byte) (int, error) {
+	select {
+	case <-lb.ctx.Done():
+		return 0, lb.ctx.Err()
+	default:
+	}
+	n, err := lb.input.Read(p)
+	if n > 0 {
+		lb.f(p)
```
AlexanderYastrebov, line 30: This is wrong, should be `lb.f(p[:n])`

io/log_stream.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"context"
+	"io"
+)
+
+type logBody struct {
+	ctx   context.Context
+	f     func([]byte)
+	input io.ReadCloser
+}
```
AlexanderYastrebov, line 12: WDYT about 
```go
type ReadCloseInspector struct {
	io.ReadCloser
	AfterRead  func([]byte, error)
	AfterClose func(error)
}

func (rci *ReadCloseInspector) Read(p []byte) (int, error) {
	n, err := rci.ReadCloser.Read(p)
	if n > 0 && rci.AfterRead != nil {
		rci.AfterRead(p[:n], err)
	}
	return n, err
}

func (rci *ReadCloseInspector) Close() error {
	err := rci.ReadCloser.Close()
	if rci.AfterClose != nil {
		rci.AfterClose(err)
	}
	return err
}
```
it can do logging, or e.g. count chunk lengths and log only after Close.

filters/diag/logbody.go
szuecs: done

filters/diag/logbody.go
szuecs: done

szuecs:
:+1:

io/log_stream.go
szuecs: This looks wrong for chunked readers

szuecs: I think I have done a good one

szuecs:
:+1:

filters/diag/logbody.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	n, err = lb.input.Read(p)
+	//println("n:", n, ", err: ", err.Error())
+	if n > 0 {
+		if len(p) > lb.limit {
+			lb.f(p[0:lb.limit])
+		} else {
+			lb.f(p)
+		}
```
AlexanderYastrebov, line 126: Maybe
```go
lb.f(p[:min(len(p), lb.limit)])
```

filters/diag/logbody.go
```diff
@@ -0,0 +1,139 @@
...
+	case "response":
+		response = true
+	case "request":
+		request = true
+	default:
+		return nil, fmt.Errorf("failed to match %q: %w", opt, filters.ErrInvalidFilterParameters)
+	}
+
+	limit, ok := args[1].(float64)
+	if !ok {
```
AlexanderYastrebov, line 49: `if !ok || float64(int(limit)) != limit`

AlexanderYastrebov:
LGTM aside minor comments

filters/diag/logbody.go
```diff
@@ -0,0 +1,139 @@
...
+import (
+	"fmt"
+	"io"
+
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/filters/flowid"
+)
+
+type logBody struct {
+	limit    int
```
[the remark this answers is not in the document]
szuecs: ah no, that's correct for the filter but not for the logBodyStream struct!

filters/diag/logbody.go
szuecs: requires go 1.21, but I guess it's fine to change go.mod to require that :)

filters/diag/logbody.go
```diff
@@ -0,0 +1,139 @@
...
+	lb.limit -= n
+
+	return n, err
+}
+
+func (lb *logBodyStream) Close() error {
+	return lb.input.Close()
+}
+
+func LogBodyStream(limit int, f func([]byte), rc io.ReadCloser) io.ReadCloser {
```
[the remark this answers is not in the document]
szuecs: dropped that one

filters/diag/logbody.go
```diff
@@ -0,0 +1,130 @@
...
+func (lb *logBodyStream) Read(p []byte) (n int, err error) {
+	if lb.left <= 0 {
+		return lb.input.Read(p)
+	}
+
+	n, err = lb.input.Read(p)
+	if n > 0 {
+		lb.f(p[:min(n, lb.left)])
+	}
+	lb.left -= n
```
AlexanderYastrebov, line 123:
```go
n, err = lb.input.Read(p)
if n > 0 && lb.left > 0 {
    m := n
    if m > lb.left { // maybe we shouldn't bump go to 1.22 until its released
        m = lb.left
    }
    lb.f(p[:m])
    lb.left -= m
}
```

szuecs: I have go.mod set to 1.21 and min() seems to exist, so I left it as it was

szuecs:
:+1:

filters/diag/logbody.go
AlexanderYastrebov: yeah, it was added in 1.21 https://pkg.go.dev/builtin#min

PR #2835, opened 2024-01-08
Ensure normalizeSchemeHost is IPv6-safe

RomanZavodskikh:
Addressing the comment https://github.com/zalando/skipper/pull/2824#discussion_r1443503926 about IPv6 safety.
Fixed small issue in `normalizeSchemeHost` function and created tests to prove its IPv4 and IPv6 safety.

### DIFF
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -486,7 +486,12 @@ func normalizeSchemeHost(s, h string) (string, string, error) {
 		p = "443"
 	}
 
-	h = net.JoinHostPort(h, p)
+	if p != "" {
+		// IPv6 addresses are enclosed in brackets by SplitHostPort, so we need to remove
+		// them before JoinHostPort insert one more brackets pair
+		h = strings.Trim(h, "[]")
+		h = net.JoinHostPort(h, p)
+	}
 	return s, h, nil
 }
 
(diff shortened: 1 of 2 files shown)


loadbalancer/algorithm.go
```diff
@@ -486,7 +486,12 @@ func normalizeSchemeHost(s, h string) (string, string, error) {
 		p = "443"
 	}
 
-	h = net.JoinHostPort(h, p)
+	if p != "" {
+		// IPv6 addresses are enclosed in brackets by SplitHostPort, so we need to remove
+		// them before JoinHostPort insert one more brackets pair
```
AlexanderYastrebov, line 491: Are you sure https://go.dev/play/p/gLwbFOOARXm ?

AlexanderYastrebov:
Lets address it within or after https://github.com/zalando/skipper/pull/2824

loadbalancer/algorithm.go
AlexanderYastrebov: See https://github.com/zalando/skipper/pull/2824#discussion_r1444779626

AlexanderYastrebov:
Lets add test cases for ipv4 and ipv6 with and without port to https://github.com/zalando/skipper/pull/2824

PR #2837, opened 2024-01-09
filters/diag/logbody: use correct calculation of bytes left to log

AlexanderYastrebov:
Followup up on #2827


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2838, opened 2024-01-09
feat: deduplicate OIDC cookies for subdomains

PatrickSpies:
Removed subdomains should share the cookie with the corresponding hostname if all other parameters are the same

e.g. foo.bar.baz (with `SubdomainsToRemove` = 1) and bar.baz (with `SubdomainsToRemove` = 0) should share the same OIDC cookie to prevent duplication. Currently this parameter is taken into account for cookie hashing which leads to multiple cookies.

See also d1457a3c7de06f72c614ce194d7cdd0eb6925ccb

### DIFF
--- a/filters/auth/oidc.go
+++ b/filters/auth/oidc.go
@@ -205,6 +205,10 @@ func (s *tokenOidcSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 		if i == paramCallbackURL {
 			continue
 		}
+		// SubdomainsToRemove not taken into account for cookie hashing for additional sub-domain ingresses
+		if i == paramSubdomainsToRemove {
+			continue
+		}
 		h.Write([]byte(s))
 	}
 	byteSlice := h.Sum(nil)


filters/auth/oidc.go
```diff
@@ -205,6 +205,10 @@ func (s *tokenOidcSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 		if i == paramCallbackURL {
 			continue
 		}
+		// SubdomainsToRemove not taken into account for cookie hashing for additional sub-domain ingresses
+		if i == paramSubdomainsToRemove {
+			continue
+		}
```
AlexanderYastrebov, line 211: Please add a test.

AlexanderYastrebov: Created https://github.com/zalando/skipper/pull/2839 to fix the docs

AlexanderYastrebov: There is a plenty of parameters but we exclude only paramCallbackURL (and proposed paramSubdomainsToRemove). I'd guess the reason is that paramCallbackURL is also depends on the domain. I am wondering whether we should _include_ as set of parameters rather then exclude these to.

@universam1 do you have any input here maybe?

universam1: not sure if I got you right, but it sounds like you want to flip from a negative to a positive paramter? Guess not a big deal for us, it would be a breaking change though I guess.

AlexanderYastrebov: Not a breaking change if we include all already existing parameters. What I am trying to understand is whether it makes sense to include add filter parameters:
* **OpenID Connect Provider URL** For example Google OpenID Connect is available on `https://accounts.google.com`
* **Client ID** This value is obtained from the provider upon registration of the application.
* **Client Secret**  Also obtained from the provider
* **Callback URL** The entire path to the callback from the provider on which the token will be received.
    It can be any value which is a subpath on which the filter is applied.
* **Scopes** The OpenID scopes separated by spaces which need to be specified when requesting the token from the provider.
* **Claims** The claims which should be present in the token returned by the provider.
* **Auth Code Options** (optional) Passes key/value parameters to a provider's authorization endpoint. The value can be dynamically set by a query parameter with the same key name if the placeholder `skipper-request-query` is used.
* **Upstream Headers** (optional) The upstream endpoint will receive these headers which values are parsed from the OIDC information. The header definition can be one or more header-query pairs, space delimited. The query syntax is [GJSON](https://github.com/tidwall/gjson/blob/master/SYNTAX.md).
* **SubdomainsToRemove** (optional, default "1") Configures number of subdomains to remove from the request hostname to derive OIDC cookie domain. By default one subdomain is removed, e.g. for the www.example.com request hostname the OIDC cookie domain will be example.com (to support SSO for all subdomains of the example.com). Configure "0" to use the same hostname. Note that value is a string.


I can justify `OpenID Connect Provider URL`+`Client ID` - they form unique "OIDC client id" within Skipper. Then `Client Secret` - if this changes it invalidates cookies - makes sense. `Callback URL` - excluded (unclear reason, maybe due-to domain). 

I am not sure e.g. about `Claims` especially if one wants to have two routes with different claims:
```
r1: Host("foo.bar.test") && Path("/needs-name-email-and-picture")-> oauthOidcUserInfo("https://oidc-provider.example.com", "client_id", "client_secret",
    "http://target.example.com/subpath/callback", "email profile", "name email picture") -> ... ;
    
r2: Host("foo.bar.test") && Path("/needs-only-name")-> oauthOidcUserInfo("https://oidc-provider.example.com", "client_id", "client_secret",
    "http://target.example.com/subpath/callback", "email profile", "name") -> ... ;
```

`Auth Code Options` - likely should be the same. `Upstream Headers` - like claims maybe user needs them different in two routes.

universam1: I can justify scopes, claims - potentially auth code options and headers as well as they might produce a different ID token from the IdP thus requiring a distinct cookie.
Excluding callback url makes sense, to support several ingress that require their valid, relative reachable return path but sharing one root domain cookie.

Also SubdomainsToRemove to exclude makes sense to a degree, if the purpose of the TR is to combine the same domains in one cookie. However it looks not clean to me, that it could be easily configured wrongly when the resulting cropped domain is not equal by mistake.

So I wonder, if we rather should include the **resulting** domain (after cropping) into the hash, that way we'll ensure the matching is deterministic - what do you think @AlexanderYastrebov @PatrickSpies

AlexanderYastrebov: The cookie has domain attribute, isn't it enough?
https://github.com/zalando/skipper/blob/df658308b4e7faf719c73e5cd48247207782e38f/filters/auth/oidc.go#L468-L478

universam1: But that domain value is not what we hash. To match correctly the cookie hash, shouldn't we call the same function like
```suggestion
		if i == paramSubdomainsToRemove {
			h.Write( []byte( extractDomainFromHost( theDomainSomehow, s ) )
			continue
		}
```
to ensure we are hashing the real resulting domain and not some assumptions

AlexanderYastrebov: > to ensure we are hashing the real resulting domain

This would require moving cookie name generation into runtime as the actual host is only known during the request.

PatrickSpies: How should we proceed here?

AlexanderYastrebov: I think we shall rely on your opinion as users of this filter. If @PatrickSpies and @universam1 agree we'll accept this change as it is.

PatrickSpies: How can we move forward here?

AlexanderYastrebov: @universam1 We would merge this as is if you do not have any objections.

universam1: Thanks for asking - I do see the mentioned issue in the wild and I think we should solve it eventually. However, I agree that this change will not worse the situation as it is as of today. So no objections to merge.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2840, opened 2024-01-10
filters/diag: do not use min builtin

AlexanderYastrebov:
min builtin was added in 1.21 and we have to support 1.20 before 1.22 release.

Followup on #2827

### DIFF
--- a/filters/diag/logbody.go
+++ b/filters/diag/logbody.go
@@ -114,7 +114,10 @@ func newLogBodyStream(left int, f func([]byte), rc io.ReadCloser) io.ReadCloser 
 func (lb *logBodyStream) Read(p []byte) (n int, err error) {
 	n, err = lb.input.Read(p)
 	if lb.left > 0 && n > 0 {
-		m := min(n, lb.left)
+		m := lb.left
+		if m > n {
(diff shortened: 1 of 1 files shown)


szuecs:
:-1:

AlexanderYastrebov:
@szuecs Please explain why you do not agree for transparency.

szuecs:
@AlexanderYastrebov thanks yes.
We do not want to limit us if there are new features in new Go versions. For example https://opensource.zalando.com/skipper/reference/filters/#readtimeout was built directly after http.ResponseController was created in Go stdlib.

I created a poll in [Slack #skipper](https://app.slack.com/client/T029RQSE6/C82Q5JNH5), that seems to show that we never broke the build for our library customers. The first poll shows that we should follow the rule as suggested by @xavivars , so we should update the VERSION file to update the skipper minor version in case we have to change go.mod  `go` directive.
I think this makes a lot of sense.

Let's discuss the exact change of when we update the minor version in my proposed readme change: https://github.com/zalando/skipper/pull/2868

szuecs:
@AlexanderYastrebov I guess we can close this PR

PR #2841, opened 2024-01-10
Use endpointregistry in the hotpath (the non-breaking try)

RomanZavodskikh:
The another option to do https://github.com/zalando/skipper/pull/2823 without breaking LBMetrics.


loadbalancer/fadein_test.go
```diff
@@ -332,7 +333,7 @@ func benchmarkFadeIn(
 				Host:     eps[i],
 				Detected: detectionTimes[i],
```
RomanZavodskikh, line 334: minor: I need to add `Metrics: ...` here too.

RomanZavodskikh:
I like https://github.com/zalando/skipper/pull/2823 more.

PR #2844, opened 2024-01-11
predicates/primitive: document False predicates ignores arguments

AlexanderYastrebov:
This feature could be used to a add comment why route has it.

### DIFF
--- a/docs/reference/predicates.md
+++ b/docs/reference/predicates.md
@@ -229,13 +229,13 @@ route2: Path("/test") && True() -> "http://www.github.com";
 
 ## False
 
-Does not match. Can be used to disable certain routes.
+Does not match. Can be used to disable certain routes. Ignores all arguments.
 
 Example where `route2` is disabled.
 
 ```
 route1: Path("/test") -> "http://www.zalando.de";
-route2: Path("/test") && False() -> "http://www.github.com";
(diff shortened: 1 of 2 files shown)


AlexanderYastrebov:
True() also ignored its arguments but I don't thing we should document it.

MustafaSaber:
👍

docs/reference/predicates.md
```diff
@@ -229,13 +229,13 @@ route2: Path("/test") && True() -> "http://www.github.com";
...
 
-Does not match. Can be used to disable certain routes.
+Does not match. Can be used to disable certain routes. Ignores all arguments.
 
 Example where `route2` is disabled.
 
 ```
 route1: Path("/test") -> "http://www.zalando.de";
-route2: Path("/test") && False() -> "http://www.github.com";
+route2: Path("/test") && False("Disabled due to maintenance") -> "http://www.github.com";
```
szuecs, line 238: Why?
I wonder why you relax input to a predicate which normally you would rather be strict about. :)

szuecs:
:-1:

AlexanderYastrebov:
I think its fine to allow a comment for False predicate but this comment can also be a part of route sourse e.g. eskip file or RouteGroup yaml. Lets come to this idea later if we see the need.

PR #2845, opened 2024-01-11
dataclients/kubernetes: add metadata route

### LINKED ISSUE #1559: Endpoint metadata

Load balancing algorithms (#557) require or would benefit from additional endpoint information besides currently available
scheme, ip address and port. For example slow start aka fadeIn algorithm (#1207) has to know when endpoint bacame available to receive traffic. Other algorithms need endpoint metrics and thus endpoint identity is required (ip address is not suitable as identity because it can be reused).

Useful endpoint metadata values are:

* identity (e.g. `targetRef.uid` in k8s)
* creation date (e.g. `createdAt` of `targetRef` in k8s)
* name (e.g. `name` of `targetRef` in k8s)
* weight (for weighted load balancing algorithms)

The mechanism for passing metadata should support possible future metadata extensions.

Endpoint metadata may be queried by skipper via infrastructure API and/or delivered to skipper proxy/algorithms via route definitions.

Within route definition the following ways are possible to pass endpoint metadata:

### Marker filters

Example route: 
```
r0: * 
-> endpointCreated("http://10.0.0.1:8080", "2020-12-18T15:30:00Z01:00") -> endpointId("http://10.0.0.1:8080", "6b6c2568-138b-11eb-835b-9f5a8a575060")
-> endpointCreated("http://10.0.0.2:8080", "2020-12-18T15:31:00Z01:00") -> endpointId("http://10.0.0.2:8080", "7daa5380-138b-11eb-b85d-7739503adcb7")
-> <roundRobin, "http://10.0.0.1:8080", "http://10.0.0.2:8080">;
```

* pro: readily available
* con: semantically incorrect (hacky, its not a filter after all)

### Eskip backend definition

Example route:
```
r0: * 
-> <roundRobin, 
    "http://10.0.0.1:8080" ("createdAt=2020-12-18T15:30:00Z01:00", "id=6b6c2568-138b-11eb-835b-9f5a8a575060"), 
    "http://10.0.0.2:8080" ("createdAt=2020-12-18T15:31:00Z01:00", "id=7daa5380-138b-11eb-b85d-7739503adcb7")
>;
```
Syntax is subject to be discussed.

* pro: established endpoint metadata as a first class route concept
* con: requires eskip extension

### Endpoint url parameters 

It is possible to use url parameters because only scheme and host of endpoint url are currently used in routing.

Example route:
```
r0: * 
-> <roundRobin, 
    "http://10.0.0.1:8080?createdAt=2020-12-18T15:30:00Z01:00&id=6b6c2568-138b-11eb-835b-9f5a8a575060", 
    "http://10.0.0.2:8080?createdAt=2020-12-18T15:31:00Z01:00&id=7daa5380-138b-11eb-b85d-7739503adcb7"
>;
```

* pro: readily available
* pro: standard url parsing machinery is available
* con: hacky?
 
I would like to validate, discuss and possibly extend the ideas first.

AlexanderYastrebov:
Adds a special route with a predefined id `kube__metadata` that contains `False` predicate and no filters.

The first argument of `False` predicate (which it ignores anyway) contains endpoint metadata encoded as JSON.

Example (formatted):
```
kube__metadata: False("{
\"addresses\":[
  {
    \"address\":\"1.2.3.4\",
    \"zone\":\"zoneA\",
    \"nodeName\":\"node1\",
    \"targetRef\":{
      \"kind\":\"Pod\",
      \"name\":\"foo-app\",
      \"namespace\":\"default\",
      \"uid\":\"83a6408f-8a5b-405c-93f4-1199f7712956\"
    }
  },
  ...
]")
```

This route could be used to obtain zone, node and pod names for a given address. E.g. EndpointRegistry postprocessos may check this route and store zone, node and pod name for each endpoint.

TODO:
- [ ] add flag  to enable metadata route
- [ ] tests

Fixes #1559


szuecs:
I don't like it. Maybe write first a doc for this and why we need metadata.

dataclients/kubernetes/endpointslices.go
```diff
@@ -125,6 +127,10 @@ type EndpointSliceEndpoints struct {
 	// https://kubernetes.io/docs/concepts/services-networking/topology-aware-routing/#safeguards
 	// Zone aware routing will be available if https://github.com/zalando/skipper/issues/1446 is closed.
 	Zone string `json:"zone"` // "eu-central-1c"
+	// Node hosting this endpoint. This can be used to determine endpoints local to a node.
+	NodeName string `json:"nodeName"`
+	// TargetRef is a reference to a Kubernetes object that represents this endpoint.
+	TargetRef *objectReference `json:"targetRef"`
```
szuecs, line 133: maybe `PodName string` is enough here
Namespace we get from endpointslice metadata

dataclients/kubernetes/resources.go
```diff
@@ -24,3 +24,10 @@ type secret struct {
 type secretList struct {
 	Items []*secret `json:"items"`
 }
+
+type objectReference struct {
+	Kind      string `json:"kind"`
+	Name      string `json:"name"`
```
szuecs, line 30: Name is the only thing we need, rest is either:
- static: Kind is always "Pod"
- duplicate from endpointslice object: Namespace
- useless: Uid

dataclients/kubernetes/endpointslices.go
AlexanderYastrebov: This is just a datastruct to get data from k8s. We can decide what we use in the metadata route.

AlexanderYastrebov:
> I don't like it. Maybe write first a doc for this and why we need metadata.

Thank you for having a look. The idea is to use pod and node names in error logging and tracing. Zone could also be used to filter endpoints by zone before applying loadbalancing algorithm. Pod uid could be use to detect [pod IP address reuse](https://github.com/zalando/skipper/issues/1559#issuecomment-724845288) and e.g. reset endpoint detected time.
Namespace we have embedded in route id so this could be ommited.

> Why do we need a special route?

The biggest upside is that metadata about routing table is transferred along with the routing table itself so there is no other API or source of inconsistency.

> Why False() with input data?

This was just a proof-of-concept. I changed the format now to use https://en.wikipedia.org/wiki/Data_URI_scheme

> Why we need so much data in that?

We can decide what data of all we pass as metadata. We can start with pod name only and extend it later.

> Why json and not something better?

JSON is extensible so we can add new fields later and separatetly on producer and consumer side. It is also a text format and there are a lot of tools to process it. We could use binary format but the data size is <200kB (and <30kB gzipped) so I do not think its worth the inconvenience.

dataclients/kubernetes/endpointslices.go
```diff
@@ -16,8 +16,10 @@ type skipperEndpointSlice struct {
 
 // Conditions have to be evaluated before creation
 type skipperEndpoint struct {
-	Address string
-	Zone    string
+	Address   string
+	Zone      string
+	NodeName  string
+	TargetRef *objectReference
```
szuecs, line 22: here we can and should only pass PodName and not ugly K8s references.

dataclients/kubernetes/endpointslices.go
szuecs: yes here true but not at https://github.com/zalando/skipper/pull/2845/files#r1450668487

dataclients/kubernetes/metadataroute.go
```diff
@@ -0,0 +1,181 @@
...
+
+type kubeRouteMetadata struct {
+	Addresses []kubeRouteMetadataAddress `json:"addresses"`
+}
+
+type kubeRouteMetadataAddress struct {
+	Address   string           `json:"address"`
+	Zone      string           `json:"zone"`
+	NodeName  string           `json:"nodeName"`
+	TargetRef *objectReference `json:"targetRef"`
```
szuecs, line 32: also not here, please only pass what we need and not "expandable" data links

dataclients/kubernetes/metadataroute.go
```diff
@@ -0,0 +1,181 @@
...
+
+	addr, ok := metadata[host]
+	if !ok {
+		return
+	}
+
+	metrics := pp.options.EndpointRegistry.GetMetrics(host)
+	metrics.SetTag("zone", addr.Zone)
+	metrics.SetTag("nodeName", addr.NodeName)
+	if addr.TargetRef != nil && addr.TargetRef.Kind == "Pod" {
```
szuecs, line 177: this is ugly, because of the data struct that I question :)

dataclients/kubernetes/resources.go
szuecs: I don't want that we pass kubernetes structs. Pass what we need and no references here.

dataclients/kubernetes/resources.go
```diff
@@ -24,3 +24,10 @@ type secret struct {
...
+	Namespace string `json:"namespace"`
+	Uid       string `json:"uid"`
```
szuecs, line 32: this we don't need

dataclients/kubernetes/resources.go
AlexanderYastrebov: Please read my comment, this struct is to unmarshal data from k8s API. We can put subset of data into metadata route, I'll update that later.

proxy/proxy.go
```diff
@@ -1308,11 +1323,10 @@ func (p *Proxy) errorResponse(ctx *context, err error) {
 			uri = uri[:i]
 		}
 		logFunc(
-			`%s after %v, route %s with backend %s %s%s, status code %d: %v, remote host: %s, request: "%s %s %s", host: %s, user agent: "%s"`,
+			`%s after %v, route %s with backend %s%s, status code %d: %v, remote host: %s, request: "%s %s %s", host: %s, user agent: "%s"`,
```
AlexanderYastrebov, line 1326: This mess should be refactored into structured logging

dataclients/kubernetes/resources.go
AlexanderYastrebov: > I'll update that later.

Updated to not use objectReference in the encoded metadata

dataclients/kubernetes/endpointslices.go
AlexanderYastrebov: Its package-internal struct and only used for unmarshaling k8s API data

dataclients/kubernetes/metadataroute.go
AlexanderYastrebov: Fixed

dataclients/kubernetes/metadataroute.go
AlexanderYastrebov: Fixed

dataclients/kubernetes/resources.go
AlexanderYastrebov: This is for unmarshalling k8s API response. 
uid can be used to detect pod rotation with the same name, e.g. from StatefulSet.

szuecs:

PR #2795, opened 2023-12-13
Use sync.Map in the endpointregistry

RomanZavodskikh:
This PR is intended to make endpointregistry component lock-free
+ use sync.Map to get entries by key
+ convert all entry fields to atomics
+ make GetMetrics method return the entry instead of copy to decrease complexity in general and number of calls to sync.Map's Load in particular

The benchmark results show that it could be slow to call `GetMetrics` too much, but if the user will find a way to reuse its return value, the performance will be much better both than current master and new version:
```
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/routing
cpu: AMD Ryzen 7 PRO 4750U with Radeon Graphics
                                       │   master.txt   │     endpointregistrySyncMap.txt     │
                                       │     sec/op     │   sec/op     vs base                │
GetInflightRequests/1_goroutines-16      163.900n ±  7%   6.977n ± 2%  -95.74% (p=0.000 n=15)
GetInflightRequests/2_goroutines-16      202.100n ±  3%   3.792n ± 3%  -98.12% (p=0.000 n=15)
GetInflightRequests/3_goroutines-16      232.500n ±  8%   2.966n ± 4%  -98.72% (p=0.000 n=15)
GetInflightRequests/4_goroutines-16      252.500n ± 13%   2.572n ± 3%  -98.98% (p=0.000 n=15)
GetInflightRequests/5_goroutines-16      253.500n ±  8%   2.484n ± 4%  -99.02% (p=0.000 n=15)
GetInflightRequests/6_goroutines-16      251.900n ± 12%   2.148n ± 4%  -99.15% (p=0.000 n=15)
GetInflightRequests/7_goroutines-16      277.900n ± 10%   2.061n ± 4%  -99.26% (p=0.000 n=15)
GetInflightRequests/8_goroutines-16      268.800n ± 13%   1.834n ± 5%  -99.32% (p=0.000 n=15)
GetInflightRequests/12_goroutines-16     297.600n ± 22%   2.077n ± 3%  -99.30% (p=0.000 n=15)
GetInflightRequests/16_goroutines-16     270.100n ±  8%   2.578n ± 3%  -99.05% (p=0.000 n=15)
GetInflightRequests/24_goroutines-16     270.800n ± 14%   2.588n ± 2%  -99.04% (p=0.000 n=15)
GetInflightRequests/32_goroutines-16     279.400n ± 13%   2.593n ± 1%  -99.07% (p=0.000 n=15)
[...]

PR #2847, opened 2024-01-12
routing: avoid allocating new entry in EndpointRegistry.GetMetrics

AlexanderYastrebov:
EndpointRegistry.GetMetrics always allocates new entry even when data for the key already exists.

sync.Map does not provide lazy initialization for the new value, see https://github.com/golang/go/issues/44159

This change uses suggested best-effort to avoid allocating a new value on each load.

Follow up on #2795


szuecs:
:+1:

PR #2850, opened 2024-01-12
fix codeql error

sepehrdaddev:
Codeql complains about go version being older that the one defined in `go.mod`, this is to address the issue


szuecs:
:+1:

sepehrdaddev:
:+1:

sepehrdaddev:
seems like I broke the workflow here is the fix #2851

PR #2853, opened 2024-01-12
Improve fadeIn CI tests

RomanZavodskikh:
+ added tests for powerOfNRandomChoices monotony
+ added tests for case when all endpoints are new
+ added tests (obviously failing at the moment) for powerOfNRandomChoices to check new and old enpoints receive correct amount of requests

### DIFF
--- a/loadbalancer/fadein_test.go
+++ b/loadbalancer/fadein_test.go
@@ -169,8 +169,21 @@ func newConsistentHashForTest(endpoints []string) routing.LBAlgorithm {
 	return newConsistentHashInternal(endpoints, 1000)
 }
 
+// Those tests check that the amount of requests per period for each endpoint is monotonical over the time.
+// For every endpoint, it could increase, decrease or stay the same.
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

AlexanderYastrebov:
> added tests (obviously failing at the moment) for powerOfNRandomChoices

I am confused, the build is green

RomanZavodskikh:
> I am confused, the build is green


That's because I commented failing test not to fail the whole build because of fadeIn not added to powerOfNRandom choices (and never had been).
The test for three other algorithms (random, round-robin, consistent-hash) are green, so there were no need to comment them.

 When I will fix this problem, I will uncomment the test for powerOfNRandomChoices.

RomanZavodskikh:
:+1:

RomanZavodskikh:
Merging this PR because:

+ I have enough approvals
+ It does not affect production code
+ Mixed with the next PR, it may become too complex, so I better separate them

PR #2865, opened 2024-01-15
dataclients/kubernetes: do not log route index

AlexanderYastrebov:
Do not log ignored route index to avoid logs bloat. The error message should be enough to figure out which routes were ignored. RouteGroup logger logs message without index only once per routegroup (see #2637).

Followup on #2796


MustafaSaber:
👍

PR #2849, opened 2024-01-12
eskip: add comment test cases for ParsePredicates

AlexanderYastrebov:
Add test cases from #2755 discovered by fuzzing.

PR #2866, opened 2024-01-15
wip: eskip: add more comment test cases

AlexanderYastrebov:
Followup on #2849

### DIFF
--- a/eskip/eskip_test.go
+++ b/eskip/eskip_test.go
@@ -348,6 +348,28 @@ func TestParseFilters(t *testing.T) {
 		`filter1(3.14) -> filter2("key", 42)`,
 		[]*Filter{{Name: "filter1", Args: []interface{}{3.14}}, {Name: "filter2", Args: []interface{}{"key", float64(42)}}},
 		false,
+	}, {
+		"a comment produces empty filters without error",
+		"// a comment",
+		nil,
+		false,
+	}, {
+		"a trailing comment is ignored",
+		`filter1(3.14) -> filter2("key", 42) // a trailing comment`,
+		[]*Filter{{Name: "filter1", Args: []interface{}{3.14}}, {Name: "filter2", Args: []interface{}{"key", float64(42)}}},
+		false,
+	}, {
+		"a comment before is ignored",
+		`// a comment on a separate line
+		filter1(3.14) -> filter2("key", 42)`,
+		[]*Filter{{Name: "filter1", Args: []interface{}{3.14}}, {Name: "filter2", Args: []interface{}{"key", float64(42)}}},
+		false,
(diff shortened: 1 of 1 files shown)


eskip/eskip_test.go
```diff
@@ -346,6 +346,16 @@ func TestParseFilters(t *testing.T) {
 		`filter1(3.14) -> filter2("key", 42)`,
 		[]*Filter{{Name: "filter1", Args: []interface{}{3.14}}, {Name: "filter2", Args: []interface{}{"key", float64(42)}}},
 		false,
+	}, {
+		"a comment produces empty filters without error",
+		"// a comment",
```
AlexanderYastrebov, line 351: Its not clear whether we want to support comments or multiline strings for ParseFilters and ParsePredicates helpers.

AlexanderYastrebov: Its tricky to support all test cases consistently due to how `ParseFilters` (and `ParsePredicates` fro that matter) is implemented:
```go
rs, err := parse("* -> " + f + " -> <shunt>")
```
We can make comments on a separate line work by adding a newline:
```go
rs, err := parse("* -> " + f + "\n -> <shunt>")
```
but its hard to make only comment input work:
```go
rs, err := parse("* -> " + "// a comment" + "\n -> <shunt>")
```
because it produces invalid eskip.

We can not prohibit comments easily because they are skipped by lexer - the lexer would need to be changed to support "allow comments" flag.

Note also that due to use of `"* -> "` prefix parse errors contain incorrect token offset value.

The proper fix would require:
* separate parse methods for routes, predicates and filters (see https://www.gnu.org/software/bison/manual/html_node/Multiple-start_002dsymbols.html)
* a lexer support to prohibit comments for `ParseFilters` and `ParsePredicates` if we decide so

AlexanderYastrebov: Created a branch that separates document, predicates and filters parsing https://github.com/zalando/skipper/tree/eskip/separate-parse-predicates-filters

It includes several other changes that I'd like to land first before opening a PR.
Using the same approach of separate start symbols we can add singular `ParseFilter` and `ParsePredicate` for use in RouteGroup dataclient/validator instead of checking that `ParseFilters` returned a single result:
https://github.com/zalando/skipper/blob/e63fe2cea92d6010fd85b8af9c90af765a2ee1e7/dataclients/kubernetes/definitions/routegroupvalidator.go#L69-L74

AlexanderYastrebov: Created draft PR https://github.com/zalando/skipper/pull/2873

AlexanderYastrebov:
I've added these testcases to #2873

PR #2869, opened 2024-01-16
Fixed EndpointRegistry.Do method to do nothing for unrelated backend types

RomanZavodskikh:
https://github.com/zalando/skipper/pull/2823#discussion_r1450672228

### DIFF
--- a/routing/endpointregistry.go
+++ b/routing/endpointregistry.go
@@ -93,7 +93,7 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
 
 				epi.Metrics.SetLastSeen(now)
 			}
-		} else {
+		} else if route.BackendType == eskip.NetworkBackend {
 			entry := r.GetMetrics(route.Host)
 			if entry.DetectedTime().IsZero() {
 				entry.SetDetected(now)
@@ -134,4 +134,13 @@ func (r *EndpointRegistry) GetMetrics(hostPort string) Metrics {
 		e, _ = r.data.LoadOrStore(hostPort, newEntry())
 	}
 	return e.(*entry)
+}
+
+func (r *EndpointRegistry) allMetrics() map[string]Metrics {
+	result := make(map[string]Metrics)
+	r.data.Range(func(k, v any) bool {
+		result[k.(string)] = v.(Metrics)
+		return true
+	})
+	return result
 }
--- a/routing/endpointregistry_test.go
+++ b/routing/endpointregistry_test.go
@@ -114,6 +114,53 @@ func TestDoRemovesOldEntries(t *testing.T) {
 	assert.Equal(t, int64(0), mRemoved.InflightRequests())
 }
 
+func TestEndpointRegistryPostProcessor(t *testing.T) {
+	beginTestTs := time.Now()
+	r := routing.NewEndpointRegistry(routing.RegistryOptions{})
+
+	routing.SetNow(r, func() time.Time {
+		return beginTestTs
+	})
+	routes := []*routing.Route{
+		{
+			Route: eskip.Route{
+				BackendType: eskip.ShuntBackend,
+			},
+		},
+		{
+			Route: eskip.Route{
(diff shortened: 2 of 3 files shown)


routing/endpointregistry_test.go
```diff
@@ -114,6 +114,34 @@ func TestDoRemovesOldEntries(t *testing.T) {
 	assert.Equal(t, int64(0), mRemoved.InflightRequests())
 }
 
+func TestDoDoesNothingForUnrelatedBackends(t *testing.T) {
```
AlexanderYastrebov, line 117: Its better to test that it does X. Maybe also add network and loadbalancer route and then check that metrics were only added for them (see https://github.com/zalando/skipper/pull/2869/files#r1453615534)

AlexanderYastrebov: `TestEndpointRegistryPostProcessor`

routing/endpointregistry_test.go
```diff
@@ -114,6 +114,34 @@ func TestDoRemovesOldEntries(t *testing.T) {
...
+		},
+	}
+	dynamicBackendRoute := &routing.Route{
+		Route: eskip.Route{
+			BackendType: eskip.DynamicBackend,
+		},
+	}
+	r.Do([]*routing.Route{shuntRoute, loopRoute, dynamicBackendRoute})
+
+	m := r.GetMetrics("")
```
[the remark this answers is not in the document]
RomanZavodskikh: Yes, I agree it makes sense not to rely on default string value in this particular case.

routing/endpointregistry_test.go
RomanZavodskikh: Okay, I created LB backend with 2 endpoints, network backend with single host and left routes with all other backend types.
After that, I check that there're only 3 entries in endpointregistry, and they correspond to 2+1 hosts created above.

routing/endpointregistry.go
```diff
@@ -135,3 +135,12 @@ func (r *EndpointRegistry) GetMetrics(hostPort string) Metrics {
 	}
 	return e.(*entry)
 }
+
+func (r *EndpointRegistry) contents() map[string]Metrics {
```
AlexanderYastrebov, line 139: `allMetrics`?

routing/endpointregistry_test.go
```diff
@@ -114,6 +114,51 @@ func TestDoRemovesOldEntries(t *testing.T) {
...
+			{Host: "endpoint2.test:80", Metrics: r.GetMetrics("endpoint2.test:80")},
+		},
+	}
+	networkBackendRoute := &routing.Route{
+		Route: eskip.Route{
+			BackendType: eskip.NetworkBackend,
+		},
+		Host: "endpoint3.test:80",
+	}
+	r.Do([]*routing.Route{shuntRoute, loopRoute, dynamicBackendRoute, lbBackendRoute, networkBackendRoute})
```
AlexanderYastrebov, line 154: Maybe it would be cleaner to define locals variable 
```go
routes := []*routing.Route{
	&routing.Route{
		Route: eskip.Route{
			BackendType: eskip.ShuntBackend,
		},
	},
	&routing.Route{
		Route: eskip.Route{
			BackendType: eskip.LoopBackend,
		},
	}, 
	...
}
r.Do(routes}
```
This way you do not need to invent a lot of local var names for single use

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

routing/endpointregistry.go
```diff
@@ -93,7 +93,7 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
 
 				epi.Metrics.SetLastSeen(now)
 			}
-		} else {
+		} else if route.BackendType == eskip.NetworkBackend {
```
szuecs, line 96: `switch  route.BackendType {}` would read nicer

PR #2872, opened 2024-01-17
feature: choose benchmark binary to test local modifications

szuecs:
feature: choose benchmark binary to test local modifications

### DIFF
--- a/skptesting/benchmark-proxy.sh
+++ b/skptesting/benchmark-proxy.sh
@@ -1,11 +1,11 @@
 #! /bin/bash
 
+source "${GOPATH}/src/github.com/zalando/skipper/skptesting/benchmark.inc"
+
 if [ "$1" == -help ]; then
-	log benchmark-proxy.sh [duration] [connections] [warmup-duration]
+	log "benchmark-proxy.sh [duration] [connections] [warmup-duration] [skipper-binary]"
 	exit 0
 fi
-
-source $GOPATH/src/github.com/zalando/skipper/skptesting/benchmark.inc
 
 check_deps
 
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

skptesting/benchmark-proxy.sh
```diff
@@ -1,12 +1,12 @@
...
 
+source "${GOPATH}/src/github.com/zalando/skipper/skptesting/benchmark.inc"
+
 if [ "$1" == -help ]; then
-	log benchmark-proxy.sh [duration] [connections] [warmup-duration]
+	log "benchmark-proxy.sh [duration] [connections] [warmup-duration] [skipper-binary]"
 	exit 0
 fi
 
-source $GOPATH/src/github.com/zalando/skipper/skptesting/benchmark.inc
```
AlexanderYastrebov, line 8: Is this moved due to missing log()?

The downside is that benchmark.inc now install skipper even if someone just ran it with -help

AlexanderYastrebov: Maybe just change it to `echo ""benchmark-proxy.sh [duration] ..." >&2`. I think it may also print to stdout since its not an error but help message

szuecs: yes it is required to use log()

AlexanderYastrebov:
:+1:

skptesting/benchmark-proxy.sh
szuecs: I guess it does not hurt to install a current local skipper

PR #2873, opened 2024-01-18
eskip: separate parsing of predicates and filters

AlexanderYastrebov:
Introduce start tokens to allow separate parsing of
eskip document, predicates and filters, see
https://www.gnu.org/software/bison/manual/html_node/Multiple-start_002dsymbols.html

Also remove `stringval` and `regexpval` rules.
These rules are aliases for `stringliteral` which can be used directly.
This reduces parser stack depth and item size.

The change fixes parse error position value and speeds up predicates/filters parsing.

```
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/eskip
ParsePredicates-8   8.531µ ± 2%   7.464µ ± 5%  -12.51% (p=0.000 n=10)
Parse-8             257.3m ± 1%   251.4m ± 2%   -2.32% (p=0.000 n=10)
geomean             1.482m        1.370m        -7.55%

                  │     B/op     │     B/op      vs base                │
ParsePredicates-8   2.375Ki ± 0%   1.773Ki ± 0%  -25.33% (p=0.000 n=10)
Parse-8             61.79Mi ± 0%   61.79Mi ± 0%        ~ (p=0.698 n=10)
geomean             387.7Ki        335.0Ki       -13.59%

                  │  allocs/op  │  allocs/op   vs base               │
ParsePredicates-8    53.00 ± 0%    50.00 ± 0%  -5.66% (p=0.000 n=10)
Parse-8             1.510M ± 0%   1.510M ± 0%       ~ (p=0.700 n=10)
geomean             8.946k        8.689k       -2.87%

        │    HEAD~1    │               HEAD               │
        │   bytes/op   │   bytes/op    vs base            │
Parse-8   15.99Mi ± 0%   15.99Mi ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

### DIFF
--- a/dataclients/kubernetes/testdata/routegroups/convert/failing-filter.log
+++ b/dataclients/kubernetes/testdata/routegroups/convert/failing-filter.log
@@ -1,1 +1,1 @@
-\[routegroup\] parse failed after token foo, position 8: syntax error
+\[routegroup\] parse failed after token foo, position 3: syntax error
--- a/eskip/eskip_test.go
+++ b/eskip/eskip_test.go
@@ -239,17 +239,37 @@ func TestParseFilters(t *testing.T) {
 		"error",
 		"trallala",
 		nil,
-		// TODO: fix position
-		"parse failed after token ->, position 16: syntax error",
+		"parse failed after token trallala, position 8: syntax error",
 	}, {
 		"error 2",
 		"foo-bar",
 		nil,
-		// TODO: fix position
-		"parse failed after token foo, position 8: syntax error",
+		"parse failed after token foo, position 3: syntax error",
 	}, {
 		"success",
 		`filter1(3.14) -> filter2("key", 42)`,
+		[]*Filter{{Name: "filter1", Args: []interface{}{3.14}}, {Name: "filter2", Args: []interface{}{"key", float64(42)}}},
+		"",
+	}, {
+		"a comment produces nil filters without error",
+		"// a comment",
+		nil,
+		"",
+	}, {
+		"a trailing comment is ignored",
+		`filter1(3.14) -> filter2("key", 42) // a trailing comment`,
+		[]*Filter{{Name: "filter1", Args: []interface{}{3.14}}, {Name: "filter2", Args: []interface{}{"key", float64(42)}}},
+		"",
+	}, {
+		"a comment before is ignored",
+		`// a comment on a separate line
+		filter1(3.14) -> filter2("key", 42)`,
+		[]*Filter{{Name: "filter1", Args: []interface{}{3.14}}, {Name: "filter2", Args: []interface{}{"key", float64(42)}}},
+		"",
+	}, {
+		"a comment after is ignored",
+		`filter1(3.14) -> filter2("key", 42)
+		// a comment on a separate line`,
 		[]*Filter{{Name: "filter1", Args: []interface{}{3.14}}, {Name: "filter2", Args: []interface{}{"key", float64(42)}}},
 		"",
 	}} {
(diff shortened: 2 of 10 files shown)


dataclients/kubernetes/testdata/routegroups/convert/failing-filter.log
```diff
@@ -1 +1 @@
-\[routegroup\] parse failed after token foo, last route id: , position 8: syntax error
+\[routegroup\] parse failed after token foo, last route id: , position 3: syntax error
```
AlexanderYastrebov, line 1: last route id was added by https://github.com/zalando/skipper/pull/1885
I think we should not add it to the error message if it is empty, will fix this separately.

dataclients/kubernetes/testdata/routegroups/convert/failing-predicate.log
```diff
@@ -1 +1 @@
-\[routegroup\] parse failed after token Header, last route id: Header, position 6: syntax error
```
AlexanderYastrebov, line 1: This shows another problem of previous ParsePredicates implementation: it confused `Header("Foo", "Bar") -> <shunt>` to have route id of `Header`

eskip/parser.y
```diff
@@ -69,17 +67,42 @@ func convertNumber(s string) float64 {
...
 
+start:
+	start_document document {
+		eskiplex.(*eskipLex).routes = $2.routes
+	}
+	|
+	start_predicates {
+		// allow empty or comments only
+		eskiplex.(*eskipLex).predicates = nil
+	}
```
AlexanderYastrebov, line 84: We can disallow comments for ParsePredicates and ParseFilters by removing these empty rules, see also https://github.com/zalando/skipper/pull/2866

dataclients/kubernetes/testdata/routegroups/convert/failing-filter.log
AlexanderYastrebov, line 1: Note that position was wrong due to previous implementation using `"* -> "` as a prefix to parse filters.

AlexanderYastrebov: https://github.com/zalando/skipper/pull/2875

dataclients/kubernetes/testdata/routegroups/convert/failing-predicate.log
AlexanderYastrebov: https://github.com/zalando/skipper/pull/2875

eskip/eskip_test.go
```diff
@@ -239,19 +239,39 @@ func TestParseFilters(t *testing.T) {
...
-		"parse failed after token foo, position 8: syntax error",
+		"parse failed after token foo, position 3: syntax error",
 	}, {
 		"success",
 		`filter1(3.14) -> filter2("key", 42)`,
 		[]*Filter{{Name: "filter1", Args: []interface{}{3.14}}, {Name: "filter2", Args: []interface{}{"key", float64(42)}}},
 		"",
+	}, {
+		"a comment produces nil filters without error",
+		"// a comment",
```
AlexanderYastrebov, line 255: We can disallow comments in ParsePredicates and ParseFilters, needs a decision.

szuecs: Depends on if we would have a benefit from that. I guess it won't be much and then I would just let it like that.

AlexanderYastrebov: Since we added https://github.com/zalando/skipper/pull/2905 I think it makes sense to allow eskip comments for ParsePredicates and ParseFilters

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2874, opened 2024-01-18
eskip: refactor parsing tests to check error message

AlexanderYastrebov:
Check value of error message to simplify followup fixes.

E.g. #1885 introduced last route id to the error message without tests but last route could be empty or should not be present at all for `ParsePredicates` and `ParseFilters`.

### DIFF
--- a/eskip/eskip_test.go
+++ b/eskip/eskip_test.go
@@ -7,98 +7,66 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/assert"
 )
 
-func checkItems(t *testing.T, message string, l, lenExpected int, checkItem func(int) bool) bool {
-	t.Helper()
-	if l != lenExpected {
-		t.Error(message, "length", l, lenExpected)
-		return false
-	}
-
-	for i := 0; i < l; i++ {
-		if !checkItem(i) {
-			t.Error(message, "item", i)
-			return false
-		}
-	}
-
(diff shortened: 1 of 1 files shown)


eskip/eskip_test.go
```diff
@@ -333,47 +210,62 @@ func TestParseFilters(t *testing.T) {
...
 		" \t",
 		nil,
-		false,
+		"",
 	}, {
 		"error",
 		"trallala",
 		nil,
-		true,
+		// TODO: remove empty last route id and fix position
```
AlexanderYastrebov, line 223: Position fix could be done by https://github.com/zalando/skipper/pull/2873

eskip/eskip_test.go
```diff
@@ -7,297 +7,174 @@ import (
...
-}
-
-func TestParseRouteExpression(t *testing.T) {
+func TestParse(t *testing.T) {
 	for _, ti := range []struct {
 		msg        string
 		expression string
-		check      *Route
-		err        bool
+		check      []*Route
```
MustafaSaber, line 17: minor: maybe expectedRoutes would be better naming?

MustafaSaber:
👍

eskip/eskip_test.go
AlexanderYastrebov: I did not rename struct fields to keep diff small. All three test use different naming

AlexanderYastrebov:
:+1:

PR #2885, opened 2024-01-24
webhook: support validation rules

AlexanderYastrebov:
Add new webhook admitter that evaluates Ingresses and RouteGroups against a set of rules.

Each rule defines properties of matching resource and rejection message.

The implementation uses [Common Expression Language](https://github.com/google/cel-spec) to match properties which is also used in [Kubernetes](https://kubernetes.io/docs/reference/using-api/cel/).


cmd/webhook/admission/testdata/rules.yaml
```diff
@@ -0,0 +1,15 @@
...
+        !has(request.object.metadata.labels) ||
+        !has(request.object.metadata.labels.application) ||
+        request.object.metadata.labels.application == ''
+      )
+  # - reject: 'zalando.org/skipper-filter: oauthTokeninfoAnyScope filter uses "uid" scope, see https://opensource.zalando.com/skipper/reference/filters/#oauthtokeninfoanyscope'
+  #   when: |
+  #     request.object.kind == "Ingress" &&
+  #     has(request.object.metadata.annotations) &&
+  #     request.object.metadata.annotations["zalando.org/skipper-filter"] != "" &&
+  #     eskipParseFilters(request.object.metadata.annotations["zalando.org/skipper-filter"])
```
AlexanderYastrebov, line 14: TODO: expose eskip.ParseFilters as eskipParseFilters in CEL environment.

cmd/webhook/admission/testdata/rules.yaml
```diff
@@ -0,0 +1,15 @@
+rules:
+  - reject: Missing application label, see https://example.test/reference/labels-selectors/#application
+    when: |
+      request.kind.kind in ["Ingress", "RouteGroup"] && (
+        !has(request.object.metadata.labels) ||
+        !has(request.object.metadata.labels.application) ||
+        request.object.metadata.labels.application == ''
+      )
```
AlexanderYastrebov, line 8: We may also implement
```yaml
rules:
  - warning: Feature X will be deprecated
    when:  request.kind.kind == "RouteGroup" && ...
```
that would allow request with warnings, see https://github.com/zalando/skipper/commit/9281783bc5242325317cdbf0ada97ebf2a4136ee

szuecs:
I would assume that you would share pros and cons and why it's better than what we have today.
Not being able to parse eskip sounds like a step backwards.

AlexanderYastrebov:
> I would assume that you would share pros and cons

This approach allows operator to define custom rules that may be specific to a given setup or contain sensitive bits.
E.g. a custom rule may reject certain filter or predicate configurations or deny backend addresses.
The downside is usage of another language (although having very simple syntax) and a need to have end-to-end test suite to test custom rules.

Alternative is to deploy closed-source webhook written in go.

> Not being able to parse eskip sounds like a step backwards.

Its possible, see TODO comment https://github.com/zalando/skipper/pull/2885#discussion_r1465455140, I decided noto to invest into it until we decide to move forward with this (or not).

MustafaSaber:
I think matchConditions for the current ValidationWebhook will be enough to cover our current usecases(not eskip parsing) 
https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchconditions
It's the same language, but we then can use the condition with validator

PR #2886, opened 2024-01-25
minor: refactor fadeIn code

RomanZavodskikh:
Some minor changes after https://github.com/zalando/skipper/pull/2634 just to make code a little bit more readable.

### DIFF
--- a/proxy/fadein.go
+++ b/proxy/fadein.go
@@ -5,19 +5,15 @@ import (
 	"math/rand"
 	"time"
 
-	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/routing"
 )
 
-func returnFast(rt *routing.Route) bool {
-	if rt.BackendType != eskip.LBBackend {
-		return true
-	}
-
-	return rt.LBFadeInDuration <= 0
+type fadeIn struct {
+	rnd              *rand.Rand
+	endpointRegistry *routing.EndpointRegistry
 }
 
-func fadeIn(lifetime time.Duration, duration time.Duration, exponent float64) float64 {
+func (f *fadeIn) fadeInScore(lifetime time.Duration, duration time.Duration, exponent float64) float64 {
 	fadingIn := lifetime > 0 && lifetime < duration
 	if !fadingIn {
 		return 1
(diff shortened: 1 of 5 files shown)


proxy/fadein.go
```diff
@@ -5,19 +5,19 @@ import (
...
-func returnFast(rt *routing.Route) bool {
-	if rt.BackendType != eskip.LBBackend {
-		return true
-	}
+type fadeIn struct {
+	rnd              *rand.Rand
+	endpointRegistry *routing.EndpointRegistry
+}
 
+func (f *fadeIn) noFadeIn(rt *routing.Route) bool {
```
AlexanderYastrebov, line 16: Do we need this? Maybe its better to simply inline the condition

proxy/fadein.go
```diff
@@ -26,18 +26,18 @@ func fadeIn(lifetime time.Duration, duration time.Duration, exponent float64) fl
...
 	now := time.Now()
-	threshold := rnd.Float64()
+	threshold := f.rnd.Float64()
 
 	filtered := make([]routing.LBEndpoint, 0, len(endpoints))
 	for _, e := range endpoints {
-		f := fadeIn(
-			now.Sub(registry.GetMetrics(e.Host).DetectedTime()),
+		f := f.fadeInScore(
+			now.Sub(f.endpointRegistry.GetMetrics(e.Host).DetectedTime()),
```
AlexanderYastrebov, line 40: Maybe extract this to a locak variable `age := now.Sub(f.endpointRegistry.GetMetrics(e.Host).DetectedTime())`

proxy/fadein.go
```diff
@@ -5,19 +5,19 @@ import (
 	"math/rand"
 	"time"
 
-	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/routing"
 )
 
-func returnFast(rt *routing.Route) bool {
-	if rt.BackendType != eskip.LBBackend {
```
MustafaSaber, line 13: Why did you remove this condition?

RomanZavodskikh: The filterFadeIn function is called only by `selectEndpoint` from proxy.go. However, the `selectEndpoint` is called only for `eskip.LBBackend`, see https://github.com/zalando/skipper/blob/v0.19.32/proxy/proxy.go#L505 .

This fact means there is no sense to check BackendType one more time, is was already checked before.

proxy/fadein.go
RomanZavodskikh: It was intended to increase readability (literally tell the people reading this code "this is the case when there is no fade-in"), however, I agree it may be too obvious so I decided to inline it.

AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

PR #2888, opened 2024-01-26
Passive health checks

RomanZavodskikh:
There is huge ongoing effort to enable Passive Health Checks is Skipper to check backends for errors and try not to send requests to the backends that tend to fail the requests.

The initial idea I have at the moment is the following:
+ leverage the endpointregistry component (https://github.com/zalando/skipper/pull/2535) developed recently to collect more data about endpoints (total amount of reqs sent to the endpoint, amount of reqs that were timed out because of backend outage)
+ if ratio of requests that were timed out is too high for particular endpoint, then send lower amount of requests to this endpoint, the decrease of RPS is configurable
+ once in a while (once in `statsResetPeriod`) update the stats to check endpoints' states change

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -285,6 +285,8 @@ type Config struct {
 	OpenPolicyAgentStartupTimeout       time.Duration `yaml:"open-policy-agent-startup-timeout"`
 	OpenPolicyAgentMaxRequestBodySize   int64         `yaml:"open-policy-agent-max-request-body-size"`
 	OpenPolicyAgentMaxMemoryBodyParsing int64         `yaml:"open-policy-agent-max-memory-body-parsing"`
+
+	PassiveHealthCheck mapFlags `yaml:"passive-health-check"`
 }
 
 const (
@@ -570,6 +572,9 @@ func NewConfig() *Config {
 
 	flag.Var(cfg.LuaModules, "lua-modules", "comma separated list of lua filter modules. Use <module>.<symbol> to selectively enable module symbols, for example: package,base._G,base.print,json")
 	flag.Var(cfg.LuaSources, "lua-sources", `comma separated list of lua input types for the lua() filter. Valid sources "", "file", "inline", "file,inline" and "none". Use "file" to only allow lua file references in lua filter. Default "" is the same as "file","inline". Use "none" to disable lua filters.`)
+
+	// Passive Health Checks
+	flag.Var(&cfg.PassiveHealthCheck, "passive-health-check", "sets the parameters for passive health check feature")
 
 	cfg.flags = flag
 	return cfg
@@ -912,6 +917,8 @@ func (c *Config) ToOptions() skipper.Options {
 		OpenPolicyAgentStartupTimeout:       c.OpenPolicyAgentStartupTimeout,
 		OpenPolicyAgentMaxRequestBodySize:   c.OpenPolicyAgentMaxRequestBodySize,
 		OpenPolicyAgentMaxMemoryBodyParsing: c.OpenPolicyAgentMaxMemoryBodyParsing,
+
+		PassiveHealthCheck: c.PassiveHealthCheck.values,
 	}
 	for _, rcci := range c.CloneRoute {
 		eskipClone := eskip.NewClone(rcci.Reg, rcci.Repl)
--- a/docs/operation/operation.md
+++ b/docs/operation/operation.md
@@ -893,6 +893,34 @@ to get the results paginated or getting all of them at the same time.
 curl localhost:9911/routes?offset=200&limit=100
 ```
 
+## Passive health check (*experimental*)
+
+Skipper has an option to automatically detect and mitigate faulty backend endpoints, this feature is called
+Passive Health Check(PHC).
+
+PHC works the following way: the entire uptime is divided in chunks of `period`, per every period Skipper calculates
+the total amount of requests and amount of requests failed per every endpoint. While next period is going on,
+the Skipper takes a look at previous period and if the amount of requests in the previous period is more than `min-requests`
+for the given endpoints then Skipper will send reduced (the more `max-drop-probability`
+and failed requests ratio in previous period are, the stronger reduction is)
+amount of requests compared to amount sent without PHC.
+
+Having this, we expect less requests to fail because a lot of them would be sent to endpoints that seem to be healthy instead.
+
+To enable this feature, you need to provide `-passive-health-check` option having all forementioned parameters
+(`period`, `min-requests`, `max-drop-probability`) defined,
+for instance: `-passive-health-check=period=1s,min-requests=10,max-drop-probability=0.9`.
+
+You need to define all parameters on your side, there are no defaults, and skipper will not run if PHC params are passed only partially.
+
+However, Skipper will run without this feature, if no `-passive-health-check` is provided at all.
+
+The parameters of `-passive-health-check` option are:
++ `period=<duration>` - the duration of stats reset period
++ `min-requests=<int>` - the minimum number of requests per `period` per backend endpoint required to activate PHC for this endpoint
++ `max-drop-probabilty=<float more than/equal to 0 and less than/equal to 1>` - the maximum possible probability of unhealthy endpoint being not considered
+while choosing the endpoint for the given request
+
 ## Memory consumption
 
 While Skipper is generally not memory bound, some features may require
--- a/proxy/healthy_endpoints.go
+++ b/proxy/healthy_endpoints.go
@@ -1,0 +1,34 @@
+package proxy
+
+import (
+	"math/rand"
+
+	"github.com/zalando/skipper/routing"
+)
+
+type healthyEndpoints struct {
+	rnd              *rand.Rand
+	endpointRegistry *routing.EndpointRegistry
+}
+
+func (h *healthyEndpoints) filterHealthyEndpoints(endpoints []routing.LBEndpoint, rt *routing.Route) []routing.LBEndpoint {
+	if h == nil {
+		return endpoints
+	}
+
+	p := h.rnd.Float64()
+
+	filtered := make([]routing.LBEndpoint, 0, len(endpoints))
+	for _, e := range endpoints {
+		if p < e.Metrics.HealthCheckDropProbability() {
+			/* drop */
+		} else {
+			filtered = append(filtered, e)
+		}
+	}
+
+	if len(filtered) == 0 {
+		return endpoints
+	}
+	return filtered
+}
--- a/proxy/healthy_endpoints_test.go
+++ b/proxy/healthy_endpoints_test.go
@@ -14,13 +14,53 @@ import (
 )
 
 const (
-	nRequests            = 10_000
+	nRequests            = 15_000
 	rtFailureProbability = 0.8
-	period               = 1 * time.Second
+	period               = 100 * time.Millisecond
 )
 
 func defaultEndpointRegistry() *routing.EndpointRegistry {
-	return routing.NewEndpointRegistry(routing.RegistryOptions{})
+	return routing.NewEndpointRegistry(routing.RegistryOptions{
+		PassiveHealthCheckEnabled:     true,
+		StatsResetPeriod:              period,
+		MinRequests:                   10,
+		MaxHealthCheckDropProbability: 1.0,
+	})
+}
+
+func TestPHCWithoutRequests(t *testing.T) {
+	services := []*httptest.Server{}
+	for i := 0; i < 3; i++ {
+		service := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusOK)
+		}))
+		services = append(services, service)
+		defer service.Close()
+	}
+	endpointRegistry := defaultEndpointRegistry()
+
+	doc := fmt.Sprintf(`* -> <random, "%s", "%s", "%s">`, services[0].URL, services[1].URL, services[2].URL)
+	tp, err := newTestProxyWithParams(doc, Params{
+		EnablePassiveHealthCheck: true,
+		EndpointRegistry:         endpointRegistry,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tp.close()
+
+	ps := httptest.NewServer(tp.proxy)
+	defer ps.Close()
+
+	rsp, err := ps.Client().Get(ps.URL)
+	if err != nil {
+		t.Fatal(err)
+	}
+	assert.Equal(t, http.StatusOK, rsp.StatusCode)
+	rsp.Body.Close()
+
+	time.Sleep(10 * period)
+	/* this test is needed to check PHC will not crash without requests sent during period at all */
 }
 
 func TestPHCForSingleHealthyEndpoint(t *testing.T) {
@@ -32,7 +72,8 @@ func TestPHCForSingleHealthyEndpoint(t *testing.T) {
 
 	doc := fmt.Sprintf(`* -> "%s"`, service.URL)
 	tp, err := newTestProxyWithParams(doc, Params{
-		EndpointRegistry: endpointRegistry,
+		EnablePassiveHealthCheck: true,
+		EndpointRegistry:         endpointRegistry,
 	})
 	if err != nil {
 		t.Fatal(err)
@@ -43,27 +84,6 @@ func TestPHCForSingleHealthyEndpoint(t *testing.T) {
 	defer ps.Close()
 
 	failedReqs := 0
-	for i := 0; i < nRequests; i++ {
-		rsp, err := ps.Client().Get(ps.URL)
-		if err != nil {
-			t.Fatal(err)
-		}
-
-		if rsp.StatusCode != http.StatusOK {
-			failedReqs++
-		}
-		rsp.Body.Close()
-	}
-	assert.Equal(t, 0, failedReqs)
-
-	// Let endpointregistry update all stats
-	time.Sleep(period + time.Millisecond)
-	dummy := fmt.Sprintf(`Header("Foo", "Bar") -> "%s"`, service.URL)
-	tp.dc.UpdateDoc(dummy, nil)
-	tp.dc.UpdateDoc(doc, nil)
-	time.Sleep(10 * time.Millisecond)
-
-	failedReqs = 0
 	for i := 0; i < nRequests; i++ {
 		rsp, err := ps.Client().Get(ps.URL)
 		if err != nil {
@@ -91,7 +111,8 @@ func TestPHCForMultipleHealthyEndpoints(t *testing.T) {
 
 	doc := fmt.Sprintf(`* -> <random, "%s", "%s", "%s">`, services[0].URL, services[1].URL, services[2].URL)
 	tp, err := newTestProxyWithParams(doc, Params{
-		EndpointRegistry: endpointRegistry,
+		EnablePassiveHealthCheck: true,
+		EndpointRegistry:         endpointRegistry,
 	})
 	if err != nil {
 		t.Fatal(err)
@@ -102,27 +123,6 @@ func TestPHCForMultipleHealthyEndpoints(t *testing.T) {
 	defer ps.Close()
 
 	failedReqs := 0
-	for i := 0; i < nRequests; i++ {
-		rsp, err := ps.Client().Get(ps.URL)
-		if err != nil {
-			t.Fatal(err)
-		}
-
-		if rsp.StatusCode != http.StatusOK {
-			failedReqs++
-		}
-		rsp.Body.Close()
-	}
-	assert.Equal(t, 0, failedReqs)
-
-	// Let endpointregistry update all stats
-	time.Sleep(period + time.Millisecond)
-	dummy := fmt.Sprintf(`* -> <random, "%s", "%s">`, services[0].URL, services[1].URL)
-	tp.dc.UpdateDoc(dummy, nil)
-	tp.dc.UpdateDoc(doc, nil)
-	time.Sleep(10 * time.Millisecond)
-
-	failedReqs = 0
 	for i := 0; i < nRequests; i++ {
 		rsp, err := ps.Client().Get(ps.URL)
 		if err != nil {
@@ -144,7 +144,7 @@ type roundTripperUnhealthyHost struct {
 	rnd         *rand.Rand
 }
 
-type RoundTripperUnhealthyHostParams struct {
+type RoundTripperUnhealthyHostOptions struct {
 	Host        string
 	Probability float64
 }
@@ -158,7 +158,7 @@ func (rt *roundTripperUnhealthyHost) RoundTrip(r *http.Request) (*http.Response,
 	return rt.inner.RoundTrip(r)
 }
 
-func newRoundTripperUnhealthyHost(o *RoundTripperUnhealthyHostParams) func(r http.RoundTripper) http.RoundTripper {
+func newRoundTripperUnhealthyHost(o *RoundTripperUnhealthyHostOptions) func(r http.RoundTripper) http.RoundTripper {
 	return func(r http.RoundTripper) http.RoundTripper {
 		return &roundTripperUnhealthyHost{inner: r, rnd: rand.New(loadbalancer.NewLockedSource()), host: o.Host, probability: o.Probability}
 	}
@@ -177,8 +177,9 @@ func TestPHCForMultipleHealthyAndOneUnhealthyEndpoints(t *testing.T) {
 
 	doc := fmt.Sprintf(`* -> <random, "%s", "%s", "%s">`, services[0].URL, services[1].URL, services[2].URL)
 	tp, err := newTestProxyWithParams(doc, Params{
+		EnablePassiveHealthCheck:   true,
 		EndpointRegistry:           endpointRegistry,
-		CustomHttpRoundTripperWrap: newRoundTripperUnhealthyHost(&RoundTripperUnhealthyHostParams{Host: services[0].URL[7:], Probability: rtFailureProbability}),
+		CustomHttpRoundTripperWrap: newRoundTripperUnhealthyHost(&RoundTripperUnhealthyHostOptions{Host: services[0].URL[7:], Probability: rtFailureProbability}),
 	})
 	if err != nil {
 		t.Fatal(err)
@@ -200,28 +201,5 @@ func TestPHCForMultipleHealthyAndOneUnhealthyEndpoints(t *testing.T) {
 		}
 		rsp.Body.Close()
 	}
-	assert.InDelta(t, 0.33*rtFailureProbability*nRequests, failedReqs, 0.05*nRequests)
-
-	// Let endpointregistry update all stats
-	time.Sleep(period + time.Millisecond)
-	dummy := fmt.Sprintf(`* -> <random, "%s", "%s">`, services[0].URL, services[1].URL)
-	tp.dc.UpdateDoc(dummy, nil)
-	tp.dc.UpdateDoc(doc, nil)
-	time.Sleep(10 * time.Millisecond)
-
-	failedReqs = 0
-	for i := 0; i < nRequests; i++ {
-		rsp, err := ps.Client().Get(ps.URL)
-		if err != nil {
-			t.Fatal(err)
-		}
-
-		if rsp.StatusCode != http.StatusOK {
-			failedReqs++
-		}
-		rsp.Body.Close()
-	}
-	assert.InDelta(t, 0.33*rtFailureProbability*nRequests, failedReqs, 0.05*nRequests)
-	// After PHC is implemented, I expect failed requests to decrease like this:
-	// assert.InDelta(t, 0.33*rtFailureProbability*(1.0-rtFailureProbability)*nRequests, failedReqs, 0.05*nRequests)
+	assert.InDelta(t, 0.33*rtFailureProbability*(1.0-rtFailureProbability)*float64(nRequests), failedReqs, 0.1*float64(nRequests))
 }
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -145,6 +145,69 @@ type OpenTracingParams struct {
 	ExcludeTags []string
 }
 
+type PassiveHealthCheck struct {
+	// The period of time after which the endpointregistry begins to calculate endpoints statistics
+	// from scratch
+	Period time.Duration
+
+	// The minimum number of total requests that should be sent to an endpoint in a single period to
+	// potentially opt out the endpoint from the list of healthy endpoints
+	MinRequests int64
+
+	// The maximum probability of unhealthy endpoint to be dropped out from load balancing for every specific request
+	MaxDropProbability float64
+}
+
+func InitPassiveHealthChecker(o map[string]string) (bool, *PassiveHealthCheck, error) {
+	if len(o) == 0 {
+		return false, &PassiveHealthCheck{}, nil
+	}
+
+	result := &PassiveHealthCheck{}
+	keysInitialized := make(map[string]struct{})
+
+	for key, value := range o {
+		switch key {
+		case "period":
+			period, err := time.ParseDuration(value)
+			if err != nil {
+				return false, nil, fmt.Errorf("passive health check: invalid period value: %s", value)
+			}
+			if period < 0 {
+				return false, nil, fmt.Errorf("passive health check: invalid period value: %s", value)
+			}
+			result.Period = period
+		case "min-requests":
+			minRequests, err := strconv.Atoi(value)
+			if err != nil {
+				return false, nil, fmt.Errorf("passive health check: invalid minRequests value: %s", value)
+			}
+			if minRequests < 0 {
+				return false, nil, fmt.Errorf("passive health check: invalid minRequests value: %s", value)
+			}
+			result.MinRequests = int64(minRequests)
+		case "max-drop-probability":
+			maxDropProbability, err := strconv.ParseFloat(value, 64)
+			if err != nil {
+				return false, nil, fmt.Errorf("passive health check: invalid maxDropProbability value: %s", value)
+			}
+			if maxDropProbability < 0 || maxDropProbability > 1 {
+				return false, nil, fmt.Errorf("passive health check: invalid maxDropProbability value: %s", value)
+			}
+			result.MaxDropProbability = maxDropProbability
+		default:
+			return false, nil, fmt.Errorf("passive health check: invalid parameter: key=%s,value=%s", key, value)
+		}
+
+		keysInitialized[key] = struct{}{}
+	}
+
+	if len(keysInitialized) != 3 {
+		return false, nil, fmt.Errorf("passive health check: missing required parameters")
+	}
+	return true, result, nil
+}
+
 // Proxy initialization options.
 type Params struct {
 	// The proxy expects a routing instance that is used to match
@@ -247,6 +310,12 @@ type Params struct {
 	// and returns some metadata about endpoint. Information about the metadata
 	// returned from the registry could be found in routing.Metrics interface.
 	EndpointRegistry *routing.EndpointRegistry
+
+	// EnablePassiveHealthCheck enables the healthy endpoints checker
+	EnablePassiveHealthCheck bool
+
+	// PassiveHealthCheck defines the parameters for the healthy endpoints checker.
+	PassiveHealthCheck *PassiveHealthCheck
 }
 
 type (
@@ -324,6 +393,7 @@ type Proxy struct {
 	routing                  *routing.Routing
 	registry                 *routing.EndpointRegistry
 	fadein                   *fadeIn
+	heathlyEndpoints         *healthyEndpoints
 	roundTripper             http.RoundTripper
 	priorityRoutes           []PriorityRoute
 	flags                    Flags
@@ -466,6 +536,7 @@ func (p *Proxy) selectEndpoint(ctx *context) *routing.LBEndpoint {
 	rt := ctx.route
 	endpoints := rt.LBEndpoints
 	endpoints = p.fadein.filterFadeIn(endpoints, rt)
+	endpoints = p.heathlyEndpoints.filterHealthyEndpoints(endpoints, rt)
 
 	lbctx := &routing.LBContext{
 		Request:     ctx.request,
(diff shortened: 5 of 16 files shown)


proxy/healthy_endpoints_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/loadbalancer"
+	"github.com/zalando/skipper/routing"
+)
+
+func initiliaze(hosts []string) (*routing.EndpointRegistry, *routing.Route) {
```
AlexanderYastrebov, line 13: initialize

config/config.go
```diff
@@ -283,6 +283,10 @@ type Config struct {
 	OpenPolicyAgentEnvoyMetadata   string        `yaml:"open-policy-agent-envoy-metadata"`
 	OpenPolicyAgentCleanerInterval time.Duration `yaml:"open-policy-agent-cleaner-interval"`
 	OpenPolicyAgentStartupTimeout  time.Duration `yaml:"open-policy-agent-startup-timeout"`
+
+	// Passive Health Checks
+	EnablePassiveHealthCheck bool     `yaml:"enable-passive-health-check"`
+	PassiveHealthCheckParams mapFlags `yaml:"passive-health-check-params"`
```
AlexanderYastrebov, line 289: Maybe we can just have a single new flag
```go
	PassiveHealthCheck mapFlags `yaml:"passive-health-check"`
```
Then we do not have any defaults (proposed defaults are arbitrary) to make user to think about the configuration and enable PHC when `len(PassiveHealthCheck) > 0`

config/config.go
```diff
@@ -567,6 +571,14 @@ func NewConfig() *Config {
 	flag.Var(cfg.LuaModules, "lua-modules", "comma separated list of lua filter modules. Use <module>.<symbol> to selectively enable module symbols, for example: package,base._G,base.print,json")
 	flag.Var(cfg.LuaSources, "lua-sources", `comma separated list of lua input types for the lua() filter. Valid sources "", "file", "inline", "file,inline" and "none". Use "file" to only allow lua file references in lua filter. Default "" is the same as "file","inline". Use "none" to disable lua filters.`)
 
+	// Passive Health Checks
+	flag.BoolVar(&cfg.EnablePassiveHealthCheck, "enable-passive-health-check", false, "enables passive health check")
+	cfg.PassiveHealthCheckParams = mapFlags{values: map[string]string{
+		"total-requests-threshold":           "10",
+		"timed-out-requests-ratio-threshold": "0.1",
```
AlexanderYastrebov, line 578: Why is it time out and not failed?

docs/operation/operation.md
```diff
@@ -893,6 +893,19 @@ to get the results paginated or getting all of them at the same time.
 curl localhost:9911/routes?offset=200&limit=100
 ```
 
+## Passive health check
+
+Skipper has an option to automatically detect and mitigate faulty backend enpoints, this feature is called
+Passive Health Check(PHC). To enable this feature, you need to provide `-enable-passive-health-check` command line flag to Skipper.
```
AlexanderYastrebov, line 899: https://sive.rs/1s

proxy/healthy_endpoints_test.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	rt := loadbalancer.NewAlgorithmProvider().Do([]*routing.Route{route})
+	route = rt[0]
+	endpointRegistry.Do([]*routing.Route{route})
+
+	return endpointRegistry, route
+}
+
+func TestFilterHealthyEndpointsReturnsAllEndpointsWhenAllAreHealthy(t *testing.T) {
+	hosts := []string{"host1:80", "host2:80", "host3:80"}
```
AlexanderYastrebov, line 32: You may use .test domain here e.g. host1.test, see https://www.rfc-editor.org/rfc/rfc2606.html#section-2

proxy/proxy.go
```diff
@@ -67,6 +67,16 @@ const (
...
+
+	// DefaultHealthyEndpointsTotalRequestsThreshold, the default minimum amount of
+	// requests to the single endpoint for healthy endpoints checker to considering
+	// opting endpoint out
+	DefaultHealthyEndpointsTotalRequestsThreshold = 10
+
+	// DefaultHealthyEndpointsTimeoutedRequestsRatioThreshold, the default minimum ratio of
+	// timeouted requests to total requests to the single endpoint for healthy endpoints
+	// checker to considering opting endpoint out
+	DefaultHealthyEndpointsTimeoutedRequestsRatioThreshold = 0.1
```
AlexanderYastrebov, line 79: Can we come up with defaults that fit all? I think this values are arbitrary so maybe we better require user to understand the logic and specify values.

proxy/proxy.go
```diff
@@ -865,6 +945,9 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 
 	roundTripper, err := p.getRoundTripper(ctx, req)
 	if err != nil {
+		if endpointMetrics != nil {
+			endpointMetrics.IncTimeoutedRequests()
```
AlexanderYastrebov, line 949: This is a wrong place, it should increment error counter after RoundTrip

docs/operation/operation.md
```diff
@@ -893,6 +893,19 @@ to get the results paginated or getting all of them at the same time.
...
+Given only this flag, Skipper will try to provide PHC functionality with default parameters.
+You can configure those parameters with `-passive-health-check-params` command line flag,
+like this: `-passive-health-check-params=total-requests-threshold=10,timed-out-requests-ratio-threshold=0.1`.
+
+The possible sub-options for `-passive-health-check-params` are:
++ `total-requests-threshold=<int>` - the minimum number of requests per minute per backend pod required to activate PHC for this pod
++ `timed-out-requests-ratio-threshold=<float>` - the minimum ratio of failed requests for every backend pod to make PHC avoid
+sending requests to this pod
```
AlexanderYastrebov, line 907: We should describe the logic how it works here. When it decides to disable endpoint and when it enables it back based on the configured and implicit parameters.

routing/endpointregistry.go
```diff
@@ -8,7 +8,11 @@ import (
 	"github.com/zalando/skipper/eskip"
 )
 
-const defaultLastSeenTimeout = 1 * time.Minute
+const (
+	defaultLastSeenTimeout  = 1 * time.Minute
+	defaultStatsResetPeriod = 1 * time.Minute
+	defaultDiscountFactor   = 0.5
```
AlexanderYastrebov, line 14: We should reflect in the operation guide how this values affect healthcheck logic.

routing/endpointregistry.go
```diff
@@ -109,6 +144,12 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
 			r.data.Delete(key)
 		}
 
+		if e.lastStatsReset.Load().(time.Time).Add(r.statsResetPeriod).Before(now) {
```
AlexanderYastrebov, line 147: Lets make this the same way as LastSeen() logic above

proxy/healthy_endpoints.go
```diff
@@ -0,0 +1,35 @@
...
+func (h *healthyEndpoints) filterHealthyEndpoints(endpoints []routing.LBEndpoint, rt *routing.Route) []routing.LBEndpoint {
+	if !h.enabled {
+		return endpoints
+	}
+
+	filtered := make([]routing.LBEndpoint, 0, len(endpoints))
+	for _, e := range endpoints {
+		if h.healthyEndpoint(e) {
+			filtered = append(filtered, e)
+		}
```
AlexanderYastrebov, line 28: Once endpoint becomes unhealthy it stops getting requests and thus total and failed requests counters do not change. How will it recover?

szuecs:
Let me comment on the idea.

> The initial idea I have at the moment is the following:
> 
>     * leverage the endpointregistry component ([Introduce endpointregistry component #2535](https://github.com/zalando/skipper/pull/2535)) developed recently to collect more data about endpoints (total amount of reqs sent to the endpoint, amount of reqs that were timed out because of backend outage)

yes you need requests total and errors total of a chosen time period.
Timeouts are only one error case. Another could be also connection reset by peer and others. I think we should count errors from skipper proxy point of view ( `rsp, err := tr.RoundTrip()`, all of those where `err` is not `nil` should be good ).

>     * if ratio of requests that were timed out is too high for particular endpoint, that filter it out, the same way we filter out fading in endpoint (https://github.com/zalando/skipper/blob/v0.19.32/proxy/proxy.go#L475)
> 
>     * once in a while (once in `statsResetPeriod`) half the amount of both total and timed out request so every single request sent now will become less and less important in the future (1 at the moment, 1/2 after `statsResetPeriod`, 1/4 after `2*statsResetPeriod`, etc.)

For the stats collection I would rather have 2 counters for each period in time and swap them.
Example you have observation period `d` and counters `total` and `errors`

```

// let's assume these are sync.Atomic and package global for simplicity of this example
var totals [2]int
var errors [2]int
var cur int // current pointer to access the arrays

// called for each request-response observation
func measure(ok bool) {
   totals[cur]++
   if !ok {
      errors[cur]++
   }
}

func background() {
  d := time.Second
  ticker := time.NewTicker(d)

  for {
    select {
      case <-quit:
           ticker.Stop()
           return
      case <- ticker.C:
    }
    ptr = cur
    cur = (cur + 1) % 2 // swap to new observation window and have stable buckets for our compute
    compute(totals[ptr], errors[ptr]) // this should store for fast lookup in the hotpath
  }   
}
```

proxy/proxy.go
RomanZavodskikh: Who are "users"?

If we are talking about Zalando, it is our team configuring `deployment.yaml`, https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/skipper/deployment.yaml, so basically we are the users, and it is us who will configure those default params (regardless of them being set in this repo or kubernetes-on-aws repo).

If we are talking about open-source users, I'm not sure it would be easy for them to configure params which will create huge entry level barrier for using this feature.

config/config.go
RomanZavodskikh: I have thought about this, and yes, I agree it makes sense to get all values from the user. However, I still provide some example with arbitrary numbers in the `operation.md` which users can use as a baseline.

config/config.go
RomanZavodskikh: Yeah, naming is hard. I came up to `failed-round-trips-ratio-threshold` as the best option I could propose.

docs/operation/operation.md
RomanZavodskikh: I have described the logic in this file.

> When it decides to disable endpoint
When the round trip ratio is too high AND amount of requests is big enough

> when it enables it back
When the round trip ration is not too high any more. For skipper to be able to compute this ratio. PHC will still send some minimal amount of request even if endpoint is considered unhealthy.

proxy/healthy_endpoints.go
RomanZavodskikh: > it stops getting requests

I agree with your point, skipper still should send some minimum amount of requests even to unhealthy endpoints, see the latest `operation.md`.

config/config.go
szuecs: We could also think the other way: provide only the Enable flag and hard code rest.
We could add env variables to switch all the "defaults" deep in the package and test in our dev cluster what values make sense. Thinking about NLB, they also have no input from the user and have PHC working.

proxy/proxy.go
```diff
@@ -891,6 +988,10 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 
 	ctx.proxySpan.LogKV("http_roundtrip", EndEvent)
 	if err != nil {
+		if endpointMetrics != nil {
+			endpointMetrics.IncTotalFailedRoundTrips()
```
RomanZavodskikh, line 992: What if there will be more than `2*statsResetPeriod` (2 minutes at the moment) between `endpointMetrics.IncTotalRequests()` and `endpointMetrics.IncTotalFailedRoundTrips()`.

In the current version of code, the total increment and failed round trips counter increment should increment the same slot corresponding to the same period of time. However, the slot for total counter would be rotated at the moment this line of code would be reached, wouldn't it? If so, the failed round trips counter will increase the slot for other time period (despite the fact this slot will be reached by the same memory address).

proxy/proxy.go
```diff
@@ -145,6 +145,82 @@ type OpenTracingParams struct {
 	ExcludeTags []string
 }
 
+type PassiveHealthCheckParams struct {
+	// The period of time after which the endpointregistry begin to calculate endpoints statistics
+	// for scratch
```
RomanZavodskikh, line 150: minor: **from** scratch

proxy/healthy_endpoints.go
```diff
@@ -0,0 +1,34 @@
...
+	for _, e := range endpoints {
+		if p < e.Metrics.HealthCheckDropProbability() {
+			/* drop */
+		} else {
+			filtered = append(filtered, e)
+		}
+	}
+
+	if len(filtered) == 0 {
+		return endpoints
```
RomanZavodskikh, line 31: This line is not covered by `*_test.go` files by some reason.

```
go test -test.short ./proxy/ -coverprofile=/tmp/coverage.out
go tool cover -html=/tmp/coverage.out
```

docs/operation/operation.md
```diff
@@ -893,6 +893,38 @@ to get the results paginated or getting all of them at the same time.
 curl localhost:9911/routes?offset=200&limit=100
 ```
 
+## Passive health check
```
RomanZavodskikh, line 896: Please mention it's **experimental** feature.

docs/operation/operation.md
```diff
@@ -893,6 +893,38 @@ to get the results paginated or getting all of them at the same time.
...
+
+To enable this feature, you need to provide `-passive-health-check-params` having all forementioned parameters
+(`stats-reset-period`, `total-requests-threshold`, `failed-round-trips-ratio-threshold`, `unhealthy-endpoints-requests-discount`) defined,
+for instance: `-passive-health-check-params=stats-reset-period=1m,total-requests-threshold=10,failed-round-trips-ratio-threshold=0.2,unhealthy-endpoints-requests-discount=0.9`.
+
+You need to define all parameters on your side, there are no defaults, and skipper will not run if PHC params are passed only partially.
+
+However, Skipper will run without this feature, if no `-passive-health-check-params` is provided at all.
+
+The parameters of `-passive-health-check-params` are:
```
RomanZavodskikh, line 920: drop -params from name.

config/config.go
RomanZavodskikh: We can call it `failed-ratio-threshold`.

docs/operation/operation.md
```diff
@@ -893,6 +893,38 @@ to get the results paginated or getting all of them at the same time.
...
++ `stats-reset-period=<duration>` - the duration of stats reset period
++ `total-requests-threshold=<int>` - the minimum number of requests per minute per backend endpoint required to activate PHC for this endpoint
```
RomanZavodskikh, line 922: `total-requests`

docs/operation/operation.md
```diff
@@ -893,6 +893,38 @@ to get the results paginated or getting all of them at the same time.
...
+To enable this feature, you need to provide `-passive-health-check-params` having all forementioned parameters
+(`stats-reset-period`, `total-requests-threshold`, `failed-round-trips-ratio-threshold`, `unhealthy-endpoints-requests-discount`) defined,
+for instance: `-passive-health-check-params=stats-reset-period=1m,total-requests-threshold=10,failed-round-trips-ratio-threshold=0.2,unhealthy-endpoints-requests-discount=0.9`.
+
+You need to define all parameters on your side, there are no defaults, and skipper will not run if PHC params are passed only partially.
+
+However, Skipper will run without this feature, if no `-passive-health-check-params` is provided at all.
+
+The parameters of `-passive-health-check-params` are:
++ `stats-reset-period=<duration>` - the duration of stats reset period
```
RomanZavodskikh, line 921: `[stats-]period`.

proxy/proxy.go
RomanZavodskikh: We expect to try this feature at Zalando at first, and then we will document it for open-source user.

proxy/proxy.go
RomanZavodskikh: IncTotalRequests could be done **after** round trip.

docs/operation/operation.md
```diff
@@ -893,6 +893,38 @@ to get the results paginated or getting all of them at the same time.
...
+Skipper has an option to automatically detect and mitigate faulty backend endpoints, this feature is called
+Passive Health Check(PHC).
+
+PHC works the following way: the entire uptime is divided in chunks of `stats-reset-period`, per every period Skipper calculates
+the total amount of requests and amount of requests for which HTTP round trip had failed per every endpoint. While next period is going on,
+the Skipper takes a look at previous period and if for the given endpoint:
++ the amount of requests in the previous period is more than `total-requests-threshold`
+AND
++ the share of requests with failed HTTP round trip is more than `failed-round-trips-ratio-threshold`
+then Skipper will send reduced (multiplied by `unhealthy-endpoints-requests-discount`, but not less than `total-requests-threshold`)
```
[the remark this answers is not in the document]
RomanZavodskikh: We expect failedRoundTripsRatio to matter here.

routing/endpointregistry.go
```diff
@@ -103,12 +148,31 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
...
 		if e.LastSeen().Before(removeOlder) {
 			r.data.Delete(key)
 		}
 
+		if r.passiveHealthCheckEnabled && e.lastStatsResetTime().Before(resetOlder) {
+			e.setLastStatsReset(now)
+			curSlot := e.curSlot.Load()
+			nextSlot := (curSlot + 1) % 2
+			e.totalFailedRoundTrips[nextSlot].Store(0)
+			e.totalRequests[nextSlot].Store(0)
```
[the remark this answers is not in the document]
szuecs: @AlexanderYastrebov the basic idea is that one slot is to write current observations and the other is to use for reading for the hot traffic algorithm. If we swap the slot every time.Duration (think "1s"), then our PHC will have reaction time of 1s.

routing/endpointregistry.go
```diff
@@ -103,12 +148,31 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
...
 
 	removeOlder := now.Add(-r.lastSeenTimeout)
+	resetOlder := now.Add(-r.statsResetPeriod)
 	r.data.Range(func(key, value any) bool {
 		e := value.(*entry)
 		if e.LastSeen().Before(removeOlder) {
 			r.data.Delete(key)
 		}
 
+		if r.passiveHealthCheckEnabled && e.lastStatsResetTime().Before(resetOlder) {
```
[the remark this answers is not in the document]
RomanZavodskikh: That's a very good catch, I've created the separate `updateStats` goroutine for this.
https://github.com/zalando/skipper/pull/2888/files#diff-1fc6636eb6f373e324ec8c39e7265fa698a5b40f804cd3de60f8400f960ff49eR162

routing/endpointregistry.go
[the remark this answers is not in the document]
RomanZavodskikh: I would kindly ask to describe this race in more detail, because I was taking quite a lot of looks, and I don't see the fatal scenarios at the moment.

routing/endpointregistry.go
```diff
@@ -60,23 +73,54 @@ func (e *entry) SetLastSeen(ts time.Time) {
...
+	if o.FailedRoundTrip {
+		e.totalFailedRoundTrips[curSlot].Add(1)
+	}
+}
+
+func (e *entry) HealthCheckDropProbability() float64 {
+	totalRequests := e.totalRequests[e.curSlot.Load()].Load()
+	if totalRequests < e.minRequests {
+		return 0.0
+	}
```
AlexanderYastrebov, line 88: Why do we need this if calculation of healthCheckDropProbability accounts for this?

szuecs: no we do not, we should just return the value as in line 89

routing/endpointregistry.go
```diff
@@ -115,23 +159,72 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
...
+
+			curSlot := e.curSlot.Load()
+			nextSlot := (curSlot + 1) % 2
+			e.totalFailedRoundTrips[nextSlot].Store(0)
+			e.totalRequests[nextSlot].Store(0)
+
+			failed := e.totalFailedRoundTrips[curSlot].Load()
+			requests := e.totalRequests[curSlot].Load()
+			failedRoundTripsRatio := float64(failed) / float64(requests)
+			if failedRoundTripsRatio > r.minFailed && requests > r.minRequests {
```
AlexanderYastrebov, line 180: Lets check whether we need `failedRoundTripsRatio > r.minFailed`

routing/endpointregistry.go
```diff
@@ -115,23 +159,72 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
...
+			e := value.(*entry)
+
+			curSlot := e.curSlot.Load()
+			nextSlot := (curSlot + 1) % 2
+			e.totalFailedRoundTrips[nextSlot].Store(0)
+			e.totalRequests[nextSlot].Store(0)
+
+			failed := e.totalFailedRoundTrips[curSlot].Load()
+			requests := e.totalRequests[curSlot].Load()
+			failedRoundTripsRatio := float64(failed) / float64(requests)
```
AlexanderYastrebov, line 179: Can this become > 1?

AlexanderYastrebov: Lets acknowledge that here we have a race reading two values and add a check failedRoundTripsRatio = min(failedRoundTripsRatio, 1.0)

routing/endpointregistry.go
```diff
@@ -22,12 +22,25 @@ type Metrics interface {
...
+type IncRequestsOptions struct {
+	FailedRoundTrip bool
 }
 
 type entry struct {
 	detected         atomic.Value // time.Time
 	lastSeen         atomic.Value // time.Time
 	inflightRequests atomic.Int64
+
+	minRequests                int64
```
AlexanderYastrebov, line 39: The same minRequests in each entry does not seem right

routing/endpointregistry.go
RomanZavodskikh: I still would like to keep it, because I can imagine some healthy node to fail only single request because of some glitch. Therefore failedRoundTripsRatio would be equal to something small, like 0.005. Giving such small values, I would like to ignore such values at all, and that's why I introduce `minFailed` parameter (which I will rename to `minFailedRatio`.

routing/endpointregistry.go
RomanZavodskikh: https://github.com/zalando/skipper/pull/2888#discussion_r1499033713
It was the only place where it was used. As I will remove that code, this field is not needed anymore.

routing/endpointregistry.go
RomanZavodskikh: We decided on the meeting that I will remove those lines.

routing/endpointregistry.go
```diff
@@ -60,23 +73,54 @@ func (e *entry) SetLastSeen(ts time.Time) {
 	e.lastSeen.Store(ts)
 }
 
-func newEntry() *entry {
+func (e *entry) IncRequests(o IncRequestsOptions) {
+	curSlot := e.curSlot.Load()
+	e.totalRequests[curSlot].Add(1)
+	if o.FailedRoundTrip {
+		e.totalFailedRoundTrips[curSlot].Add(1)
+	}
```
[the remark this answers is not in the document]
RomanZavodskikh: Such situation is possible, but has no critical circumstances: in this case statsUpdater will just calculate dropProbability ignoring a little amount of information from the very last requests in the period.
It may change the probability value a little bit, but it still will be reasonable and changed atomically.
Moveover, curSlot variable is read at the very beginning of this function, and it guarantees that all information about the given requests will go to the same slot.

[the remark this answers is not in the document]
RomanZavodskikh: >  they can be changed by stats updater

Stats updater stores the data to the `nextSlot`, not the `curSlot`, which makes possibility of zeroing data request handler increments right now not practically possible.

routing/endpointregistry.go
szuecs: @RomanZavodskikh the probability would be so low that we would not have a lot of traffic impaired by that and other pods will be able to take the load.

docs/operation/operation.md
```diff
@@ -893,6 +893,34 @@ to get the results paginated or getting all of them at the same time.
...
+the Skipper takes a look at previous period and if the amount of requests in the previous period is more than `min-requests`
+for the given endpoints then Skipper will send reduced (the more `max-drop-probability`
+and failed requests ratio in previous period are, the stronger reduction is)
+amount of requests compared to amount sent without PHC.
+
+Having this, we expect less requests to fail because a lot of them would be sent to endpoints that seem to be healthy instead.
+
+To enable this feature, you need to provide `-passive-health-check` option having all forementioned parameters
+(`period`, `min-requests`, `max-drop-probability`) defined,
+for instance: `-passive-health-check=period=1m,min-requests=10,max-drop-probability=0.9`.
```
szuecs, line 912: is 1m a resaonable choice?
I would propose something like "1s" instead

proxy/proxy.go
```diff
@@ -717,11 +788,22 @@ func WithParams(p Params) *Proxy {
 	}
 
 	hostname := os.Getenv("HOSTNAME")
+	healthyEndpoints := &healthyEndpoints{
+		rnd:              rand.New(loadbalancer.NewLockedSource()),
+		endpointRegistry: p.EndpointRegistry,
+	}
+	if !p.EnablePassiveHealthCheck {
+		healthyEndpoints = nil
```
szuecs, line 796: this code is a bit unobvious, I think the patter we normally use is this:
```
var healthyEndpoints *healthyEndpoints
if p.EnablePassiveHealthCheck {
    healthyEndpoints = &healthyEndpoints{
        ...
    }
}
```

routing/endpointregistry.go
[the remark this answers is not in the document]
szuecs: As discussed, we think the race might be ok

routing/endpointregistry.go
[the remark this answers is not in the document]
szuecs: Discussed in person, we think it's a good approach, maybe premature optimization, but it's the fast data-path solution.

routing/endpointregistry.go
```diff
@@ -115,16 +151,71 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
...
+			e.totalFailedRoundTrips[nextSlot].Store(0)
+			e.totalRequests[nextSlot].Store(0)
+
+			failed := e.totalFailedRoundTrips[curSlot].Load()
+			requests := e.totalRequests[curSlot].Load()
+			failedRoundTripsRatio := float64(failed) / float64(requests)
+			// Given the current implementation, I expect failedRoundTripsRatio to be always less than 1
+			// because both failed and total requests counter may only increase over time and failed is loaded
+			// before total requests. However, after some code refactoring this assumption may not hold anymore,
+			// that's why I'm adding this line.
```
szuecs, line 175: the comment: replace `I` by `we`"

skipper.go
```diff
@@ -1915,8 +1917,18 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	})
 	defer schedulerRegistry.Close()
 
+	passiveHealthCheckEnabled, passiveHealthCheck, err := proxy.InitPassiveHealthChecker(o.PassiveHealthCheck)
```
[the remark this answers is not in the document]
szuecs: I see that this encloses a bit of complexity for the default case with 0 PassiveHealthCheck options.

routing/endpointregistry.go
```diff
@@ -115,16 +151,71 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
...
+		statsResetPeriod:              o.StatsResetPeriod,
+		minRequests:                   o.MinRequests,
+		maxHealthCheckDropProbability: o.MaxHealthCheckDropProbability,
+
+		quit: make(chan struct{}),
+
+		now:  time.Now,
+		data: sync.Map{},
 	}
+	go registry.updateStats()
```
szuecs, line 216: I guess we want to check here the `o.PassiveHealthCheckEnabled` and remove the check and member `passiveHealthCheckEnabled` in the registry.

routing/endpointregistry.go
```diff
@@ -115,16 +151,71 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
 	return routes
 }
 
+func (r *EndpointRegistry) updateStats() {
+	if !r.passiveHealthCheckEnabled {
```
szuecs, line 155: this can be dropped if you follow my suggestion below

routing/endpointregistry.go
```diff
@@ -60,23 +72,47 @@ func (e *entry) SetLastSeen(ts time.Time) {
...
 	result.SetLastSeen(time.Time{})
 	return result
 }
 
 type EndpointRegistry struct {
-	lastSeenTimeout time.Duration
-	now             func() time.Time
-	data            sync.Map // map[string]*entry
+	lastSeenTimeout               time.Duration
+	passiveHealthCheckEnabled     bool
```
szuecs, line 97: this can be dropped if you follow my suggestion below

docs/operation/operation.md
RomanZavodskikh: It's still experimental, so let's find out.

szuecs:
I am testing this on my computer:

```
# failing backend
./bin/skipper -inline-routes='fail: Traffic(0.9) -> latency("600ms") -> status(502) -> <shunt>; r: * -> status(200) -> repeatContent("foo",1) -> <shunt>' -address :9001

# good backends
./bin/skipper -inline-routes='r: * -> status(200) -> repeatContent("foo",1) -> <shunt>' -address :9002 
./bin/skipper -inline-routes='r: * -> status(200) -> repeatContent("foo",1) -> <shunt>' -address :9003 
./bin/skipper -inline-routes='r: * -> status(200) -> repeatContent("foo",1) -> <shunt>' -address :9004 

# proxy
./bin/skipper -inline-routes='* -> backendTimeout("100ms") -> <roundRobin , "http://127.0.0.1:9001", "http://127.0.0.1:9002", "http://127.0.0.1:9003", "http://127.0.0.1:9004">' -passive-health-check='period=1s,min-requests=10,max-drop-probability=0.99'

# client
echo "GET http://127.0.0.1:9090" | vegeta attack -rate 100/s | vegeta report -every 1s
```

routing/endpointregistry.go
```diff
@@ -115,16 +150,69 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
...
+			e := value.(*entry)
+
+			curSlot := e.curSlot.Load()
+			nextSlot := (curSlot + 1) % 2
+			e.totalFailedRoundTrips[nextSlot].Store(0)
+			e.totalRequests[nextSlot].Store(0)
+
+			failed := e.totalFailedRoundTrips[curSlot].Load()
+			requests := e.totalRequests[curSlot].Load()
+			failedRoundTripsRatio := float64(failed) / float64(requests)
```
AlexanderYastrebov, line 167: Is division by zero possible here?

routing/endpointregistry.go
```diff
@@ -115,16 +150,69 @@ func (r *EndpointRegistry) Do(routes []*Route) []*Route {
...
+			// Given the current implementation, I expect failedRoundTripsRatio to be always less than 1
+			// because both failed and total requests counter may only increase over time and failed is loaded
+			// before total requests. However, after some code refactoring this assumption may not hold anymore,
+			// that's why we are adding this line.
+			failedRoundTripsRatio = min(failedRoundTripsRatio, 1.0)
```
AlexanderYastrebov, line 172: This seems redundant as we cap the value below anyways

AlexanderYastrebov:
:+1:

routing/endpointregistry.go
RomanZavodskikh: Yes, I must move this code under if block below.

RomanZavodskikh: The division by zero is possible, it will make failedRoundTripsRatio equal to 'NaN', and this will make drop probability equal to None. That will cause no problems by itself, but it is very implicit, of course I need:
+ move this code under if block to avoid NaNs
+ write a test that will check PHC won't crash if there're no requests during period (to be on the safe side).

RomanZavodskikh:
+ added test to check PHC won't crash if no requests sent during period https://github.com/zalando/skipper/pull/2888/files#diff-b1bbc8a9c24c3b87d0de99e08ebcdd6df2da583d27fa530049a3cc12d496435cR31
+ avoid division by zero during failedRoundTripsRatio calculation https://github.com/zalando/skipper/pull/2888/files#diff-1fc6636eb6f373e324ec8c39e7265fa698a5b40f804cd3de60f8400f960ff49eR168

szuecs:
:+1:

RomanZavodskikh:
:+1:

RomanZavodskikh:
Load tested the latest image from this PR with such an options:
```
# no options
Requests      [total, rate, throughput]         18005, 100.00, 49.74
Duration      [total, attack, wait]             3m0s, 3m0s, 12.291ms
Latencies     [min, mean, 50, 90, 95, 99, max]  2.331ms, 33.466ms, 39.489ms, 63.35ms, 63.878ms, 67.248ms, 306.544ms
Bytes In      [total, mean]                     207467, 11.52
Success       [ratio]                           49.75%
Status Codes  [code:count]                      200:8957  504:9048  

# -passive-health-check=period=1s,min-requests=10,max-drop-probability=0.99999
Requests      [total, rate, throughput]         20001, 100.01, 77.58
Duration      [total, attack, wait]             3m20s, 3m20s, 61.625ms
Latencies     [min, mean, 50, 90, 95, 99, max]  1.338ms, 16.912ms, 3.436ms, 63.252ms, 63.822ms, 65.332ms, 277.025ms
Bytes In      [total, mean]                     180336, 9.02
Success       [ratio]                           77.60%
Status Codes  [code:count]                      200:15520  504:4481  

# -passive-health-check=period=5s,min-requests=5,max-drop-probability=0.99
Requests      [total, rate, throughput]         18007, 100.01, 78.71
Duration      [total, attack, wait]             3m0s, 3m0s, 3.501ms
Latencies     [min, mean, 50, 90, 95, 99, max]  1.76ms, 16.075ms, 3.222ms, 62.711ms, 63.102ms, 64.696ms, 290.57ms
Bytes In      [total, mean]                     160564, 8.92
Success       [ratio]                           78.70%
Status Codes  [code:count]                      200:14172  504:3835  

# -passive-health-check=period=30s,min-requests=5,max-drop-probability=0.99
Requests      [total, rate, throughput]         18007, 100.01, 99.34
Duration      [total, attack, wait]             3m0s, 3m0s, 2.803ms
Latencies     [min, mean, 50, 90, 95, 99, max]  2.167ms, 3.733ms, 3.06ms, 4.018ms, 4.541ms, 14.933ms, 218.219ms
Bytes In      [total, mean]                     127129, 7.06
Success       [ratio]                           99.33%
Status Codes  [code:count]                      200:17887  504:120  
```

Seems fair enough to merge.

proxy/healthy_endpoints.go
```diff
@@ -0,0 +1,34 @@
...
+
+type healthyEndpoints struct {
+	rnd              *rand.Rand
+	endpointRegistry *routing.EndpointRegistry
+}
+
+func (h *healthyEndpoints) filterHealthyEndpoints(endpoints []routing.LBEndpoint, rt *routing.Route) []routing.LBEndpoint {
+	if h == nil {
+		return endpoints
+	}
```
MustafaSaber, line 17: I don't understand this condition, if h is nil `h.filterHealthyEndpoints()` call should fail before even going into function, right?

szuecs: @MustafaSaber test it yourself and you will see :)

proxy/healthy_endpoints.go
```diff
@@ -0,0 +1,34 @@
...
+
+	p := h.rnd.Float64()
+
+	filtered := make([]routing.LBEndpoint, 0, len(endpoints))
+	for _, e := range endpoints {
+		if p < e.Metrics.HealthCheckDropProbability() {
+			/* drop */
+		} else {
+			filtered = append(filtered, e)
+		}
```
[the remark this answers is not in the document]
szuecs: I think we want to be clear about the decision that if `p < e.Metrics.HealthCheckDropProbability()` holds true we want to skip the endpoint.

proxy/proxy.go
```diff
@@ -145,6 +145,69 @@ type OpenTracingParams struct {
 	ExcludeTags []string
 }
 
+type PassiveHealthCheck struct {
+	// The period of time after which the endpointregistry begins to calculate endpoints statistics
```
MustafaSaber, line 149: I think for godocs you need to start the comment with field name 
`// Period ....`

proxy/proxy.go
```diff
@@ -145,6 +145,69 @@ type OpenTracingParams struct {
...
+
+	// The minimum number of total requests that should be sent to an endpoint in a single period to
+	// potentially opt out the endpoint from the list of healthy endpoints
+	MinRequests int64
+
+	// The maximum probability of unhealthy endpoint to be dropped out from load balancing for every specific request
+	MaxDropProbability float64
+}
+
+func InitPassiveHealthChecker(o map[string]string) (bool, *PassiveHealthCheck, error) {
```
MustafaSaber, line 161: maybe rename o to opts?

proxy/proxy.go
```diff
@@ -888,7 +970,9 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 	req = injectClientTrace(req, ctx.proxySpan)
 
 	response, err := roundTripper.RoundTrip(req)
-
+	if endpointMetrics != nil {
+		endpointMetrics.IncRequests(routing.IncRequestsOptions{FailedRoundTrip: err != nil})
```
MustafaSaber, line 974: nice

proxy/healthy_endpoints.go
MustafaSaber: note for myself, check nil in Go https://go.dev/play/p/8rZ9UOcEY7t

PR #2894, opened 2024-01-29
proxy: add last route id to the 'max loopbacks reached' error

AlexanderYastrebov:
Keep `max loopbacks reached` part of the error message intact in case someone greps logs for it.


MustafaSaber:
👍

RomanZavodskikh:
:+1:

PR #2905, opened 2024-02-06
feature: comment() filter to comment a given filter chain.

szuecs:
One example use case is to use it in more complex objects like RouteGroups and have something that is shown also on kubectl edit rg case

### DIFF
--- a/filters/builtin/comment.go
+++ b/filters/builtin/comment.go
@@ -1,0 +1,40 @@
+package builtin
+
+import (
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/routing"
+)
+
+type comment struct{}
+
+// NewComment is a filter to comment a filter chain. It does nothing
+func NewComment() filters.Spec {
+	return &comment{}
+}
+
+func (*comment) Name() string {
+	return filters.CommentName
+}
+
+func (c *comment) CreateFilter(args []interface{}) (filters.Filter, error) {
+	return c, nil
+}
+
+func (*comment) Request(filters.FilterContext) {}
+
+func (*comment) Response(filters.FilterContext) {}
+
+type CommentPostProcessor struct{}
+
+func (CommentPostProcessor) Do(routes []*routing.Route) []*routing.Route {
+	for _, r := range routes {
+		var ff []*routing.RouteFilter
+		for _, f := range r.Filters {
+			if f.Name != filters.CommentName {
+				ff = append(ff, f)
+			}
+		}
+		r.Filters = ff
+	}
+	return routes
+}
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -63,6 +63,22 @@ Example:
 route1: Host(/^all401\.example\.org$/) -> status(401) -> <shunt>;
 ```
 
+## comment
+
+No operation, only to comment the route or a group of filters of a route
+
+Parameters:
+
+* msg (string)
+
+Example:
+
+```
+route1: *
+    -> comment("nothing to see")
+    -> <shunt>;
+```
+
 ## HTTP Headers
 ### preserveHost
 
--- a/filters/builtin/builtin.go
+++ b/filters/builtin/builtin.go
@@ -115,6 +115,7 @@ const (
 func Filters() []filters.Spec {
 	return []filters.Spec{
 		NewBackendIsProxy(),
+		NewComment(),
 		NewRequestHeader(),
 		NewSetRequestHeader(),
 		NewAppendRequestHeader(),
--- a/filters/filters.go
+++ b/filters/filters.go
@@ -207,6 +207,7 @@ func (r Registry) Register(s Spec) {
 // All Skipper filter names
 const (
 	BackendIsProxyName                         = "backendIsProxy"
+	CommentName                                = "comment"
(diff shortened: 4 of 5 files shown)


filters/builtin/comment.go
```diff
@@ -0,0 +1,23 @@
+package builtin
+
+import "github.com/zalando/skipper/filters"
+
+type comment struct{}
+
+// NewComment is a filter to comment a filter chain. It does nothing
```
AlexanderYastrebov, line 7: We'd need an entry in filters.md docs.

Another/additional idea is to introduce `annotation(<key>, <value>)` filter that represents metadata and could be used by route pre-processing for something. The subject usecase could then become `annotation("comment", "foo bar baz")`.

For this filter we may also add a pre-processor that will skip this filter and therefore there will be no penalty in runtime (e.g. makes no sense to update filter metrics for it).

filters/builtin/comment.go
```diff
@@ -0,0 +1,23 @@
...
+func NewComment() filters.Spec {
+	return &comment{}
+}
+
+func (*comment) Name() string {
+	return filters.CommentName
+}
+
+func (c *comment) CreateFilter(args []interface{}) (filters.Filter, error) {
+	return c, nil
```
AlexanderYastrebov, line 17: Makes sense to restrict args to single string value

filters/builtin/comment.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+func (*comment) Request(filters.FilterContext) {}
+
+func (*comment) Response(filters.FilterContext) {
```
AlexanderYastrebov, line 22: stray newline here

filters/builtin/comment.go
AlexanderYastrebov: Also we should start a countdown until someone puts json into the value :)

szuecs: I think it overcomplicates the problem. Given that the implementation at runtime will be literally:

```
TEXT local.github.com/zalando/skipper/filters/builtin.(*comment).Request(SB) /home/sszuecs/go/src/github.com/zalando/skipper/filters/builtin/comment.go
  comment.go:20         0x15f12c0               c3                      RET
```

filters/builtin/comment.go
szuecs: why limit?

filters/builtin/comment.go
szuecs: done

szuecs:
:+1:

filters/builtin/comment.go
AlexanderYastrebov: because filter has a purpose, I dont think its a good idea to allow `comment("foo", "bar", "baz")` or `comment(1, 2, 3)`

filters/builtin/comment.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (cpp *CommentPreProcessor) Do(routes []*eskip.Route) []*eskip.Route {
+	for _, r := range routes {
+		toDelete := make(map[int]struct{})
+
+		for i, f := range r.Filters {
+			if f.Name == filters.CommentName {
+				toDelete[i] = struct{}{}
+			}
+		}
```
AlexanderYastrebov, line 37: Why not
```go
for _, r := range routes {
	var ff []eskip.Filter
	for _, f := range r.Filters {
		if f.Name != filters.CommentName {
			ff = append(ff, f)
		}
	}
	r.Filters = ff
}
```

filters/builtin/comment.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (c *comment) CreateFilter(args []interface{}) (filters.Filter, error) {
+	return c, nil
+}
+
+func (*comment) Request(filters.FilterContext) {}
+
+func (*comment) Response(filters.FilterContext) {}
+
+type CommentPreProcessor struct{}
```
AlexanderYastrebov, line 27: There is no need to use pointer receiver for empty structs

filters/builtin/comment.go
szuecs: So because we want to save a filter metric we will copy all filters :)

filters/builtin/comment.go
AlexanderYastrebov: > For this filter we may also add a pre-processor that will skip this filter

It should be post-processor I think, otherwise routesrv will not show this filter in the routing table (need to check)

filters/builtin/comment.go
szuecs: done

filters/builtin/comment.go
szuecs: implemented, but I think it's not worth it

szuecs:
:+1:

filters/builtin/comment.go
szuecs: Maybe someone wants to write:
```
comment(
1, "foo bar",
2, "baz",
)
```
Why should we care about the input?

AlexanderYastrebov:
:+1:

filters/builtin/comment.go
AlexanderYastrebov: The same is true for the filter struct

szuecs:
:+1:

PR #2906, opened 2024-02-06
remove deadcode loadbalancer healthchecker

szuecs:
remove deadcode that is not in use for a while, mark all public configuration as deprecated to delete it later


config/config.go
```diff
@@ -340,7 +340,7 @@ func NewConfig() *Config {
 	flag.IntVar(&cfg.MaxLoopbacks, "max-loopbacks", proxy.DefaultMaxLoopbacks, "maximum number of loopbacks for an incoming request, set to -1 to disable loopbacks")
 	flag.IntVar(&cfg.DefaultHTTPStatus, "default-http-status", http.StatusNotFound, "default HTTP status used when no route is found for a request")
 	flag.StringVar(&cfg.PluginDir, "plugindir", "", "set the directory to load plugins from, default is ./")
-	flag.DurationVar(&cfg.LoadBalancerHealthCheckInterval, "lb-healthcheck-interval", 0, "use to set the health checker interval to check healthiness of former dead or unhealthy routes")
+	flag.DurationVar(&cfg.LoadBalancerHealthCheckInterval, "lb-healthcheck-interval", 0, "This is *deprecated* and not in use anymore")
```
AlexanderYastrebov, line 343: We can add it to `c.checkDeprecated` to print warning

AlexanderYastrebov:
Makes sense to also cleanup commented-out usages
https://github.com/zalando/skipper/blob/3082a04bef0482b330ca6acbf5c8d59ffa5f99d9/proxy/proxy.go#L919-L926

config/config.go
szuecs: oh we have something like this :)

szuecs: done

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #2907, opened 2024-02-07
Address vet issues found in Go 1.22

mikkeloscar:
This fixes `go vet` issues discovered in Go 1.22: https://go.dev/doc/go1.22

```
cmd/webhook/admission/admission.go:123:61: call to time.Since is not deferred
# github.com/zalando/skipper/proxy
# [github.com/zalando/skipper/proxy]
proxy/longpoll_test.go:122:2: the cancel function is not used on all paths (possible context leak)
proxy/longpoll_test.go:125:3: this return statement may be reached without using the cancel var defined on line 122
```

### DIFF
--- a/cmd/webhook/admission/admission.go
+++ b/cmd/webhook/admission/admission.go
@@ -120,7 +120,9 @@ func Handler(admitter admitter) http.HandlerFunc {
 		}
 
 		start := time.Now()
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

szuecs:
:+1:

PR #2909, opened 2024-02-07
filters/builtin: preallocate filter slice

AlexanderYastrebov:
Also use value receiver for empty strtuct type.

Followup on #2905

### DIFF
--- a/filters/builtin/comment.go
+++ b/filters/builtin/comment.go
@@ -9,26 +9,26 @@ type comment struct{}
 
 // NewComment is a filter to comment a filter chain. It does nothing
 func NewComment() filters.Spec {
-	return &comment{}
+	return comment{}
 }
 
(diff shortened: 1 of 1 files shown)


filters/builtin/comment.go
```diff
@@ -9,26 +9,26 @@ type comment struct{}
...
 
-func (*comment) Response(filters.FilterContext) {}
+func (comment) Response(filters.FilterContext) {}
 
 type CommentPostProcessor struct{}
 
 func (CommentPostProcessor) Do(routes []*routing.Route) []*routing.Route {
 	for _, r := range routes {
-		var ff []*routing.RouteFilter
+		ff := make([]*routing.RouteFilter, 0, len(r.Filters))
```
AlexanderYastrebov, line 31: See alternative #2910

PR #2910, opened 2024-02-07
filters/builtin: do not allocate filter slice

AlexanderYastrebov:
Also use value receiver for empty struct type.

Followup on #2905

Alternative to #2909


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2916, opened 2024-02-09
Remove dead function in swarm.go

trkohler:
I was reading code source and noticed that Broadcast is not used anywhere. I thought to propose to clean out this piece of code which is not used anywhere.


szuecs:
It's likely an external interface function, but not sure.
I have an old branch that I want to follow up on to rebuild the whole (swim/memberlist based) swarm code. Please just ignore it for now.

Issue #2476, opened 2023-07-20
Allow redis shards discovery without full-fledged k8s dataclient


AlexanderYastrebov:
**Is your feature request related to a problem? Please describe.**

https://github.com/zalando/skipper/pull/2009 added dynamic discovery of Redis shards via Kubernetes service endpoints.
This implementation relies on k8s dataclient present as it uses the endpoints data that is updated periodically by routing datasource poll cycle.

This works fine for k8s cluster-ingress type of deployment where Skipper needs k8s dataclient anyway.
It is not great for application type deployment where full k8s dataclient is not needed. Most likely users would need to ~hack~ configure k8s dataclient and e.g. use a fake selector (#2045) or ingress/routegroup class (#334, #1456) to not fetch Ingresses and RouteGroups from the cluster.

**Describe the solution you would like**

Skipper supports Redis shards discovery in k8s without fetching other cluster resources.
Ideally k8s dataclient is not registered into the routing to not produce empty/default routes (like https redirect).
Maybe it is possible to use k8s dataclient's cluster client to fetch redis endpoints and not rely on datasource poll cycle.

**Describe alternatives you've considered (optional)**

**Additional context (optional)**

**Would you like to work on it?**
Yes

AlexanderYastrebov:
Skipper also supports pulling endpoints via HTTP (e.g. from routesrv) #2242
Might be nice to support pulling directly from k8s apiserver (with ServiceAccount token).

Users may also deploy routesrv and only use redis endpoints API but that adds another moving part.

AlexanderYastrebov:
> Users may also deploy routesrv and only use redis endpoints API

`routesrv` may be configured to produce no routes, e.g. to only serve Redis endpoints from the namespace without
Ingresses/RouteGroups:
```sh
bin/routesrv \
-enable-kubernetes-endpointslices \
-kubernetes-namespace=my-redis-ns \
-kubernetes-redis-service-namespace=my-redis-ns \
-kubernetes-redis-service-name=my-redis \
-kubernetes-https-redirect=false \
-kubernetes-healthcheck=false \
...
```

PR #2917, opened 2024-02-09
dataclients/kubernetes: change GetEndpointAddresses to return addresses without scheme and port

AlexanderYastrebov:
* use -kubernetes-redis-service-port flag to explicitly configure redis port
* use separate cluster state cache for addresses since cache key does not contain port
* rename helpers protocol argument to scheme

This is a breaking change for users that:

* use kubernetes dataclient GetEndpointAddresses method
* use dynamic redis shards discovery AND use non-standard redis port AND do not specify -kubernetes-redis-service-port

For https://github.com/zalando/skipper/issues/2476


dataclients/kubernetes/clusterstate.go
```diff
@@ -83,38 +84,35 @@ func (state *clusterState) GetEndpointsByService(namespace, name, protocol strin
 	return targets
 }
 
-// GetEndpointsByName returns the skipper endpoints for kubernetes endpoints or endpointslices.
-// This function works only correctly for endpointslices (and likely endpoints) with one port with the same protocol ("TCP", "UDP").
-func (state *clusterState) GetEndpointsByName(namespace, name, protocol, scheme string) []string {
-	epID := endpointID{
-		ResourceID: newResourceID(namespace, name),
-		Protocol:   protocol,
-	}
```
AlexanderYastrebov, line 92: This was wrong because it does not contain target unlike other places that use state.cachedEndpoints

dataclients/kubernetes/endpoints.go
```diff
@@ -22,18 +22,18 @@ type endpointList struct {
 	Items []*endpoint `json:"items"`
 }
 
-func formatEndpointString(ip, protocol string, port int) string {
-	return protocol + "://" + net.JoinHostPort(ip, strconv.Itoa(port))
+func formatEndpointString(ip, scheme string, port int) string {
```
AlexanderYastrebov, line 25: This is not required by this PR but I noticed the inconsistency and decided to fix it along.

dataclients/kubernetes/endpointslices_test.go
```diff
@@ -1,51 +1,91 @@
...
+				Protocol: "TCP",
+			},
+		},
+	}).addresses())
 
-	ses := &skipperEndpointSlice{
-		Meta: &definitions.Metadata{
-			Namespace: "ns1",
-			Name:      "a-slice",
+	assert.Equal(t, []string{"10.0.0.1:8080", "10.0.0.1:8081"}, (&skipperEndpointSlice{
```
szuecs, line 26: This looks wrong to me, because if you have a multi port service this could lead to mixed up ports as far as I understand.

dataclients/kubernetes/endpointslices.go
```diff
@@ -74,20 +77,13 @@ func (eps *skipperEndpointSlice) targetsByServiceTarget(protocol, scheme string,
 	return result
 }
 
-func (eps *skipperEndpointSlice) targets(protocol, scheme string) []string {
-	result := make([]string, 0, len(eps.Endpoints))
-
-	var port int
-	for _, p := range eps.Ports {
-		if p.Protocol == protocol {
```
szuecs, line 82: this could mix STCP, UDP and TCP (that's iirc the protocol meaning in kubernetes)

dataclients/kubernetes/endpointslices_test.go
AlexanderYastrebov: yes, and this is how it is implemented for endpoints. I.e. this function is usable for single-port services

Another option is to return only IPs without port and make port explicit (I think we even have a flag for it) or get first port from the service explicitly.

dataclients/kubernetes/clusterstate.go
```diff
@@ -83,38 +84,35 @@ func (state *clusterState) GetEndpointsByService(namespace, name, protocol strin
...
+// getEndpointAddresses returns the list of all addresses for the given service using endpoints or endpointslices.
+func (state *clusterState) getEndpointAddresses(namespace, name string) []string {
+	rID := newResourceID(namespace, name)
+
 	state.mu.Lock()
 	defer state.mu.Unlock()
-	if cached, ok := state.cachedEndpoints[epID]; ok {
+	if cached, ok := state.cachedAddresses[rID]; ok {
```
AlexanderYastrebov, line 93: I think we can drop the cache as implementation is straight forward

dataclients/kubernetes/endpointslices.go
AlexanderYastrebov: Decided to return addresses only

dataclients/kubernetes/endpointslices_test.go
AlexanderYastrebov: Decided to return addresses without ports

RomanZavodskikh:
I would like to kindly ask why we should provide format different from what we already have in the `routing.Route`? Does it make sense to have the same format there and here for increasing readability?

https://github.com/zalando/skipper/pull/2824/files#diff-bacdef5a036ec850a0efe0adc1f0288bc03339074b068f5b9b87e7d8532aaa5bR274

AlexanderYastrebov:
> I would like to kindly ask why we should provide format different from what we already have in the routing.Route

Redis endpoints are used by redis client, these are not backend endpoints for the proxy.

AlexanderYastrebov:
TestConcurrentKubernetesClusterStateAccessWithRemoteRedis fails because it creates an endpoint with single address and multiple ports 
https://github.com/zalando/skipper/blob/be4ef50b259efc02d1cd11e53ebd436eb7fdb112/redis_test.go#L150-L166
while in k8s cluster endpoints have multiple addresses and a single port.

I've created https://github.com/zalando/skipper/pull/2919 to address this.

szuecs:
:+1:

szuecs:
Please make sure that the release notes contain all information what needs to be done to migrate (breaking change with simple migration path if documented).

skipper.go
```diff
@@ -1357,15 +1357,17 @@ func findKubernetesDataclient(dataClients []routing.DataClient) *kubernetes.Clie
...
 	// call to kubernetes API
 	return func() ([]string, error) {
-		a := kdc.GetEndpointAddresses(namespace, name)
+		a := kdc.GetEndpointAddresses(opts.KubernetesRedisServiceNamespace, opts.KubernetesRedisServiceName)
 		log.Debugf("Redis updater called and found %d redis endpoints", len(a))
+
+		port := strconv.Itoa(opts.KubernetesRedisServicePort)
 		for i := 0; i < len(a); i++ {
-			a[i] = strings.TrimPrefix(a[i], "http://")
+			a[i] = net.JoinHostPort(a[i], port)
```
AlexanderYastrebov, line 1370: Here is an obscure bug - the code modifies cached value of `a` returned from kdc.GetEndpointAddresses

AlexanderYastrebov: It was not observed in production becase k8s poll interval is 3 seconds and redis endpoints update interval is 10s

PR #2870, opened 2024-01-16
eskip: improve lexer performance 2

AlexanderYastrebov:
* use plain ascii instead of unicode package
* use loop for scanSymbol
* call scan functions directly instead of selectScanner to aid inlining

```
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/eskip
                  │    HEAD~1    │                HEAD                 │
                  │    sec/op    │   sec/op     vs base                │
ParsePredicates-8   9.637µ ± 11%   8.894µ ± 4%   -7.71% (p=0.001 n=10)
Parse-8             329.1m ±  4%   272.7m ± 2%  -17.15% (p=0.000 n=10)
geomean             1.781m         1.557m       -12.56%

                  │    HEAD~1    │                HEAD                 │
                  │     B/op     │     B/op      vs base               │
ParsePredicates-8   2.008Ki ± 0%   2.008Ki ± 0%       ~ (p=1.000 n=10)
Parse-8             49.94Mi ± 0%   49.94Mi ± 0%       ~ (p=0.926 n=10)
geomean             320.4Ki        320.4Ki       -0.00%

                  │   HEAD~1    │                 HEAD                 │
                  │  allocs/op  │  allocs/op   vs base                 │
ParsePredicates-8    33.00 ± 0%    33.00 ± 0%       ~ (p=1.000 n=10) ¹
Parse-8             1.100M ± 0%   1.100M ± 0%       ~ (p=0.367 n=10)
geomean             6.025k        6.025k       +0.00%
¹ all samples are equal
```

See previous https://github.com/zalando/skipper/pull/2755

PR #2918, opened 2024-02-09
Optimise sanitize with string builder

trkohler:
not a big difference, but on large query strings it's noticeable.
can be minor and safe enhancement?

`go test -benchmem -run=^$ -bench ^BenchmarkStripQuery$ -benchtime=10s  github.com/zalando/skipper/filters/builtin`

```
BenchmarkStripQuery/[old_sanitize]_url_1-8              85732707               118.6 ns/op            80 B/op          4 allocs/op
BenchmarkStripQuery/[old_sanitize]_url_2-8              14885370               808.2 ns/op           472 B/op         15 allocs/op
BenchmarkStripQuery/[old_sanitize]_url_3-8              73297474               162.9 ns/op           120 B/op          4 allocs/op
BenchmarkStripQuery/[old_sanitize]_url_4-8              100000000              114.1 ns/op            80 B/op          4 allocs/op


BenchmarkStripQuery/[new_sanitize]_url_1-8              100000000              103.5 ns/op            24 B/op          3 allocs/op
BenchmarkStripQuery/[new_sanitize]_url_2-8              15854868               778.5 ns/op           192 B/op         14 allocs/op
BenchmarkStripQuery/[new_sanitize]_url_3-8              71689101               166.3 ns/op            64 B/op          4 allocs/op
BenchmarkStripQuery/[new_sanitize]_url_4-8              100000000              102.0 ns/op            24 B/op          3 allocs/op
```

### DIFF
--- a/filters/builtin/stripquery.go
+++ b/filters/builtin/stripquery.go
@@ -51,7 +51,8 @@ func validHeaderFieldByte(b byte) bool {
 }
 
 // make sure we don't generate invalid headers
-func sanitize(input string) string {
+// temporary public function to benchmark it
+func Sanitize(input string) string {
 	toAscii := strconv.QuoteToASCII(input)
 	var b bytes.Buffer
 	for _, i := range toAscii {
@@ -60,6 +61,18 @@ func sanitize(input string) string {
 		}
 	}
 	return b.String()
+}
+
+// temporary public function to benchmark it
+func NewSanitize(input string) string {
+	toAscii := strconv.QuoteToASCII(input)
+	var s strings.Builder
+	for _, i := range toAscii {
+		if validHeaderFieldByte(byte(i)) {
+			s.WriteRune(i)
+		}
+	}
+	return s.String()
 }
 
 // Strips the query parameters and optionally preserves them in the X-Query-Param-xyz headers.
@@ -85,7 +98,7 @@ func (f *stripQuery) Request(ctx filters.FilterContext) {
 			if r.Header == nil {
 				r.Header = http.Header{}
 			}
-			r.Header.Add(fmt.Sprintf("X-Query-Param-%s", sanitize(k)), v)
+			r.Header.Add(fmt.Sprintf("X-Query-Param-%s", Sanitize(k)), v)
 		}
 	}
 
--- a/filters/builtin/stripquery_test.go
+++ b/filters/builtin/stripquery_test.go
@@ -15,11 +15,13 @@
 package builtin
 
 import (
-	"github.com/zalando/skipper/filters/filtertest"
+	"fmt"
 	"net/http"
 	"net/url"
 	"strings"
 	"testing"
+
+	"github.com/zalando/skipper/filters/filtertest"
 )
 
 func TestCreateStripQueryFilter(t *testing.T) {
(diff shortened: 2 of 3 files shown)


filters/builtin/stripquery.go
```diff
@@ -51,7 +51,8 @@ func validHeaderFieldByte(b byte) bool {
 }
 
 // make sure we don't generate invalid headers
-func sanitize(input string) string {
+// temporary public function to benchmark it
```
AlexanderYastrebov, line 54: There is no need to export it as benchmarks sits in the same package.
Usually we introduce benchmark as a first commit and include baseline benchmark results in the commit message.
Then change implementation, re-run benchmark and compare results to the baseline via https://pkg.go.dev/golang.org/x/perf/cmd/benchstat

To get a statistically meaningful benchmark results it should be run several times, usually we run 10.
See this PR as an example of performance-related change https://github.com/zalando/skipper/pull/2870

I've also created a small script to automate benchmarking two commits (it important that benchmark function is added in a separate commit to establish the baseline), see https://github.com/golang/go/issues/63233#issuecomment-1736468193

filters/builtin/stripquery.go
```diff
@@ -62,6 +63,18 @@ func sanitize(input string) string {
...
+func NewSanitize(input string) string {
+	toAscii := strconv.QuoteToASCII(input)
+	var s strings.Builder
+	for _, i := range toAscii {
+		if validHeaderFieldByte(byte(i)) {
+			s.WriteRune(i)
+		}
+	}
+	return s.String()
```
AlexanderYastrebov, line 75: To get ultimate performance we can use byte version https://pkg.go.dev/strconv#AppendQuoteToASCII here and then replace invalid bytes with `-` instead of allocating another slice.

filters/builtin/stripquery_test.go
```diff
@@ -91,3 +93,38 @@ func TestPreserveQuery(t *testing.T) {
...
+		url string
+	}{
+		{url: "http://example.org/foo?bar=baz"},
+		{url: "http://example.org/foo?query=cXVlcnkgewogIG5hbWUKfQo%253D&variables=&apiEndpoint=https%253A%252F%252Fwww.zalando.de%252Fapi%252Fgraphql%252Fpreview&frontendType=web&zalandoFeature=manual"},
+		{url: "http://example.org/foo?馬鹿=123"},
+		{url: "http://example.org/foo?a%20b=123"},
+	}
+
+	for i, v := range table {
+		url, _ := url.ParseRequestURI(v.url)
```
AlexanderYastrebov, line 108: This is redundant, test cases should contain query string instead of url

filters/builtin/stripquery_test.go
```diff
@@ -91,3 +93,38 @@ func TestPreserveQuery(t *testing.T) {
...
+		q := url.Query()
+		b.Run(fmt.Sprintf("[new sanitize] url %d", i + 1), func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				for k := range q {
+					NewSanitize(k)
+				}
+			}
+		})
+	}
```
AlexanderYastrebov, line 129: There is no need to have a second benchmark, see comment above.

results.md
```diff
@@ -0,0 +1,14 @@
+`go test -benchmem -run=^$ -bench ^BenchmarkStripQuery$ -benchtime=10s  github.com/zalando/skipper/filters/builtin`
```
AlexanderYastrebov, line 1: Please add results to the commit message as described in the comment above.

filters/builtin/stripquery_test.go
```diff
@@ -91,3 +93,38 @@ func TestPreserveQuery(t *testing.T) {
...
+		{url: "http://example.org/foo?bar=baz"},
+		{url: "http://example.org/foo?query=cXVlcnkgewogIG5hbWUKfQo%253D&variables=&apiEndpoint=https%253A%252F%252Fwww.zalando.de%252Fapi%252Fgraphql%252Fpreview&frontendType=web&zalandoFeature=manual"},
+		{url: "http://example.org/foo?馬鹿=123"},
+		{url: "http://example.org/foo?a%20b=123"},
+	}
+
+	for i, v := range table {
+		url, _ := url.ParseRequestURI(v.url)
+		q := url.Query()
+		b.Run(fmt.Sprintf("[old sanitize] url %d", i+1), func(b *testing.B) {
```
AlexanderYastrebov, line 110: Makes sense to add `b.ReportAllocs()` instead of  `-benchmem` flag

PR #2927, opened 2024-02-12
net: fix flaky TestRedisClient

AlexanderYastrebov:
Add delay before checking updater calls after close.

Followup on #2177


szuecs:
:+1:

PR #2929, opened 2024-02-12
dataclients/kubernetes: remove endpoint addresses cache

AlexanderYastrebov:
Returning cached endpoint addreses from exported GetEndpointAddresses makes it possible for caller to modify cached values.

Since the logic of obtaining endpoint addresses is lightweight this change simply drops caching altogether.

Follow up on #2917


RomanZavodskikh:
> the logic of obtaining endpoint addresses is lightweight 

`BenchmarkCachedEndpoint` seems to measure performance impact of removing cache, did you consider running it?
https://github.com/zalando/skipper/blob/v0.19.56/dataclients/kubernetes/clusterstate_test.go#L65

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2933, opened 2024-02-13
dataclients/kubernetes: export ClusterClient

AlexanderYastrebov:
Refactor and export cluster client to allow usage without dataclient.


AlexanderYastrebov:
Decided to proceed with #2934

PR #2935, opened 2024-02-13
dataclients/kubernetes: show broken TestGetEndpointAddresses

AlexanderYastrebov:
TestGetEndpointAddresses is broken because it uses shared TestAPI instance and does not re-initialize it for each test.

In partucular "Ingress with endpointslices" was actually testing endpoints and not endpointslices.


AlexanderYastrebov:
Addressed by https://github.com/zalando/skipper/pull/2936

PR #2937, opened 2024-02-14
dataclients/kubernetes/kubernetestest: support getting resource by name

AlexanderYastrebov:
This is useful for testing getting endpoints by service name.

For #2476

### DIFF
--- a/dataclients/kubernetes/kubernetestest/api.go
+++ b/dataclients/kubernetes/kubernetestest/api.go
@@ -3,6 +3,7 @@ package kubernetestest
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"regexp"
@@ -43,8 +44,9 @@ type api struct {
 func NewAPI(o TestAPIOptions, specs ...io.Reader) (*api, error) {
 	a := &api{
 		namespaces: make(map[string]namespace),
+		// see https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-uris
 		pathRx: regexp.MustCompile(
-			"(/namespaces/([^/]+))?/(services|ingresses|routegroups|endpointslices|endpoints|secrets)",
+			"(?:/namespaces/([^/]+))?/(services|ingresses|routegroups|endpointslices|endpoints|secrets)(?:/(.+))?",
 		),
 	}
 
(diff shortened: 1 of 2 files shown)


dataclients/kubernetes/kubernetestest/api.go
```diff
@@ -43,8 +44,9 @@ type api struct {
 func NewAPI(o TestAPIOptions, specs ...io.Reader) (*api, error) {
 	a := &api{
 		namespaces: make(map[string]namespace),
+		// see https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-uris
 		pathRx: regexp.MustCompile(
-			"(/namespaces/([^/]+))?/(services|ingresses|routegroups|endpointslices|endpoints|secrets)",
+			"(?:/namespaces/([^/]+))?/(services|ingresses|routegroups|endpointslices|endpoints|secrets)(?:/(.+))?",
```
AlexanderYastrebov, line 49: We may consider refactoring this to use https://pkg.go.dev/net/http#hdr-Patterns added in go1.22

RomanZavodskikh: What do you think about having test like this? For you not to forget to have a look when we will update Go version.
```
func TestPathRx(t *testing.T) {
    if runtime.Version() == "go1.22" {
        t.Error("please pay attention to dataclients/kubernetes/kubernetestest/api.go:49 using http.Patterns")
    }
}
```


https://pkg.go.dev/runtime#Version

dataclients/kubernetes/kubernetestest/api.go
```diff
@@ -220,35 +219,56 @@ func parseSelectors(r *http.Request) map[string]string {
...
+	if name == "" && len(selectors) == 0 {
+		w.Write(resources)
+		return
 	}
 
-	labels := struct {
+	itemsMetadata := struct {
 		Items []struct {
 			Metadata struct {
+				Name   string            `json:"name"`
```
RomanZavodskikh, line 232: Should we consider namespace as well?

dataclients/kubernetes/kubernetestest/api_test.go
```diff
@@ -235,117 +256,108 @@ func TestTestAPI(t *testing.T) {
...
 
 		var sec map[string]interface{}
-		if err := get(fmt.Sprintf(kubernetes.SecretsNamespaceFmt, namespace), &sec); err != nil {
-			t.Fatal(err)
-		}
+		get(t, fmt.Sprintf(kubernetes.SecretsNamespaceFmt, namespace), &sec)
 		check(t, sec, 0, "Secret")
 	})
+
+	t.Run("resource by name", func(t *testing.T) {
```
RomanZavodskikh, line 337: Should we add "resource with this name does not exist in current namespace, but exists in another"?

dataclients/kubernetes/kubernetestest/api.go
AlexanderYastrebov: We can add it but is not used as the test api only filters by name or labels

dataclients/kubernetes/kubernetestest/api.go
AlexanderYastrebov: This will fail as we use 1.22 already.
We can use 1.22 features after we bump min version in [go.mod](https://github.com/zalando/skipper/blob/ca9abef2b36b1e650365efe2d4b8e8dac69dae59/go.mod#L184)

The proposed change is optional it may wait until we touch this code again.

dataclients/kubernetes/kubernetestest/api_test.go
AlexanderYastrebov: Why? The result will be the same as "resource name does not exist" below

RomanZavodskikh: I expected to have some probability of some bug related to it in k8s API, but maybe I am too paranoid.

RomanZavodskikh:
:+1:

dataclients/kubernetes/kubernetestest/api_test.go
```diff
@@ -208,11 +226,14 @@ func TestTestAPI(t *testing.T) {
 	s := httptest.NewServer(a)
 	defer s.Close()
 
-	get := func(uri string, o interface{}) error {
-		return getJSON(s.URL+uri, o)
+	get := func(t *testing.T, uri string, o interface{}) {
+		t.Helper()
+		err := getJSON(s.URL+uri, o)
+		require.NoError(t, err)
```
szuecs, line 232: `get` function looks a bit weird as a name if we start to check for error. Maybe `mustGet`

AlexanderYastrebov:
:+1:

PR #2261, opened 2023-03-30
Refactor RedisUpdater func 

AKARSHITJOSHI:
fixes : #2258

PR #2939, opened 2024-02-16
net/redisclient: fix nil pointer dereference

AlexanderYastrebov:
Fix use of logger before initialization.

Followup on #2261

### DIFF
--- a/net/redisclient.go
+++ b/net/redisclient.go
@@ -222,6 +222,10 @@ func NewRedisRingClient(ro *RedisOptions) *RedisRingClient {
 			ringOptions.NewConsistentHash = NewMultiprobe
 		}
 
+		if ro.Log == nil {
+			ro.Log = &logging.DefaultLog{}
+		}
+
 		if ro.AddrUpdater != nil {
 			address, err := ro.AddrUpdater()
 			for i := 0; i < retryCount; i++ {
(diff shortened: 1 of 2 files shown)


net/redisclient.go
```diff
@@ -222,6 +222,10 @@ func NewRedisRingClient(ro *RedisOptions) *RedisRingClient {
 			ringOptions.NewConsistentHash = NewMultiprobe
 		}
 
+		if ro.Log == nil {
+			ro.Log = &logging.DefaultLog{}
+		}
+
 		if ro.AddrUpdater != nil {
 			address, err := ro.AddrUpdater()
 			for i := 0; i < retryCount; i++ {
```
AlexanderYastrebov, line 231: I think we should remove this from the client - it simply delays creation of the client and client will recover after redis is up.

We also check AddrUpdater once at startup https://github.com/zalando/skipper/blob/37c474ad7ce52f529028cd8de8dcbe722df0e6ed/skipper.go#L1717-L1730

which might be bigger problem - imagine permanent redis failure prevents new pods startup leading to cascading failure. I think its better to fail open/close with ratelimits then fail startup when redis is not avaialble.

MustafaSaber: I get your point but 10 sec extra startup time is not that much and it will take longer with container restart. Maybe we don't need to fail startup but it's still good to have the retries, wdyt?

szuecs: > which might be a bigger problem - imagine permanent redis failure prevents new pods startup leading to cascading failure. I think its better to fail open/close with ratelimits then fail startup when redis is not avaialble.

I disagree here and you should page if this is a permanent issue. A new bad deployment will block and you get a page.

AlexanderYastrebov: > I get your point but 10 sec extra startup time is not that much and it will take longer with container restart. 

Check the code, its just a delay, it only logs an error.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2940, opened 2024-02-16
filters/builtin: minor performance optimization for sanitize in StripQuery Filter

trkohler:
This is continuation of https://github.com/zalando/skipper/pull/2918/files

I have discovered that for certain cases string builder performs better than buffer, which is usually a go-to option for Go Developers. String builder was introduced in 1.10+ Go.

results (also attached to last commit message)

```
pkg: github.com/zalando/skipper/filters/builtin
           │   sec/op    │   sec/op     vs base               │
Sanitize-8   802.1n ± 0%   748.3n ± 0%  -6.71% (p=0.000 n=10)

           │    B/op    │    B/op     vs base                │
Sanitize-8   472.0 ± 0%   192.0 ± 0%  -59.32% (p=0.000 n=10)

           │   HEAD~1   │               HEAD                │
           │ allocs/op  │ allocs/op   vs base               │
Sanitize-8   15.00 ± 0%   14.00 ± 0%  -6.67% (p=0.000 n=10)
```

### DIFF
--- a/filters/builtin/stripquery_test.go
+++ b/filters/builtin/stripquery_test.go
@@ -15,11 +15,12 @@
 package builtin
 
 import (
-	"github.com/zalando/skipper/filters/filtertest"
 	"net/http"
 	"net/url"
 	"strings"
 	"testing"
+
+	"github.com/zalando/skipper/filters/filtertest"
 )
 
 func TestCreateStripQueryFilter(t *testing.T) {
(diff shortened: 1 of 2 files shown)


AlexanderYastrebov:
Please include benchstat results (without 10+10 before/after benchmark outputs) into commit message (and update PR description).

I'd also suggest to prefix commit message with a package name, e.g.:
`filters/builtin: introduce benchmark for stripQuery filter`

filters/builtin/stripquery_test.go
```diff
@@ -91,3 +92,21 @@ func TestPreserveQuery(t *testing.T) {
 		}
 	}
 }
+
+func BenchmarkStripQuery(b *testing.B) {
```
AlexanderYastrebov, line 96: Rename to BenchmarkSanitize OR rewrite to benchmark the filter, see e.g. https://github.com/zalando/skipper/blob/37c474ad7ce52f529028cd8de8dcbe722df0e6ed/filters/builtin/header_test.go#L547-L559 (make sure each run gets the same input though)

filters/builtin/stripquery_test.go
```diff
@@ -91,3 +92,21 @@ func TestPreserveQuery(t *testing.T) {
...
+	piece := "query=cXVlcnkgewogIG5hbWUKfQo%253D&variables=&apiEndpoint=https%253A%252F%252Fwww.zalando.de%252Fapi%252Fgraphql%252Fpreview&frontendType=web&zalandoFeature=manual"
+	q := strings.Repeat(piece, 1000)
```
AlexanderYastrebov, line 98: Is 1000 realistic? There is a limit to the length of query string anyways, I think we should use something more realistic (say 2x size of max length you see in your system)

AlexanderYastrebov: We also prefer to not put system-specific values into tests, you can use https://en.wikipedia.org/wiki/Metasyntactic_variable and `xxxxxx` for names or values as this does not affect the performance.

filters/builtin/stripquery_test.go
trkohler: renamed to `BenchmarkSanitize`

trkohler:
I attached benchmark results in last commit (https://github.com/zalando/skipper/pull/2940/commits/afdd4268f0b22934288e2d165f7e901f0f586b81)

I have also tried to use `AppendQuoteToASCII`, but it resulted in some failed tests. And when I do atempts to fix it properly, there are more allocations happening and it degrades benchmark result. Probably, I use it incorrectly... For now, I let the line be unchanged. 

I've also re-wrote description pointing out my motivation for a contribution. I'm sorry if previously it was thought as some issue with time-pressure. It is not pressuring at all, it's just my attempt to do an enhancement on some small part of source code. I thought that if it would succeed, I can try something bigger next.

AlexanderYastrebov:
You'd need to `git commit --amend -s` to add DCO tothe last commit.
Please include benchstat results (**without** individual benchmark outputs) to see better see the improvement.

filters/builtin/stripquery_test.go
```diff
@@ -91,3 +92,21 @@ func TestPreserveQuery(t *testing.T) {
...
+	piece := "query=cXVlcnkgUGRwKCRjb25maWdTa3U6IElEIS&variables=%7B%0A%20%20%20%20%22beautyColorImageWidth%22:%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20200,%0A%20%20%20%20%22portraitGalleryWidth%22:%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20828,%0A%20%20%20%20%22segmentedBannerHeaderLogoWidth%22:%20%20%20%20%20%20%20%20%20%20%20%20%20%2084,%0A%20%20%20%20%22shouldIncludeCtaTrackingKey%22:%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20false,%0A%20%20%20%20%22shouldIncludeFlagInfo%22:%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20false,%0A%20%20%20%20%22shouldIncludeHistogramValues%22:%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20false,%0A%20%20%20%20%22shouldIncludeOfferSelectionValues%22:%20%20%20%20%20%20%20%20%20%20%20true,%0A%20%20%20%20%22shouldIncludeOmnibusConfigModeChanges%22:%20%20%20%20%20%20%20false,%0A%20%20%20%20%22shouldIncludeOmnibusPriceLabelChanges%22:%20%20%20%20%20%20%20false,&apiEndpoint=https%253A%252F%252Fmodified%252Fsecret%252Fgraphql%252Fsecret&frontendType=secret&zalandoFeature=secret"
+	q := strings.Repeat(piece, 2)
+	v, e := url.ParseQuery(q)
```
AlexanderYastrebov, line 99: Please use err instead of e like elsewhere.

AlexanderYastrebov:
:+1:

trkohler:
👍

PR #1613, opened 2020-11-18
Fixes global ratelimits registry initialization

AlexanderYastrebov:
https://github.com/zalando/skipper/pull/1549 broke global ratelimits due to registry misinitialization (short variable declaration was used instead of assignment)

Signed-off-by: Alexander Yastrebov <alexander.yastrebov@zalando.de>

PR #2945, opened 2024-02-19
Fix Open Policy Agent filter in standalone use

mjungsbluth:
PR #2518 introduced a variable re-declaration that effectively renders the Open Policy Agent filters useless when used in standalone mode (this does not apply to use as a go library).


AlexanderYastrebov:
Been there, done that https://github.com/zalando/skipper/pull/1613 :)

szuecs:
:+1:

PR #1454, opened 2020-06-24
tee loopback documentation

aryszka:
Signed-off-by: Arpad Ryszka <arpad.ryszka@gmail.com>

PR #2948, opened 2024-02-19
docs/reference/filters: fix teeLoopback example

AlexanderYastrebov:
Add missing quote and fix invalid id (dash is not allowed).

Follow up on #1454


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2895, opened 2024-01-30
net: use redis 7

AlexanderYastrebov:
https://raw.githubusercontent.com/redis/redis/7.2/00-RELEASENOTES

PR #2949, opened 2024-02-20
net/redistest: update go-redis client to work with redis:6

AlexanderYastrebov:
Update https://github.com/redis/go-redis/releases/tag/v9.5.0 broke support of redis:6

Unfortunately we did not catch it because we updated test redis to redis:7 recently, see #2895

This change adds test to verify client supports both redis:6 and redis:7

The issue seems to be https://github.com/redis/go-redis/pull/2915

### DIFF
--- a/net/redistest/redistest_test.go
+++ b/net/redistest/redistest_test.go
@@ -15,3 +15,13 @@ func TestRedistest(t *testing.T) {
 		t.Fatalf("Failed to ping redis: %v", err)
 	}
 }
+
+func TestRedistestRedis6(t *testing.T) {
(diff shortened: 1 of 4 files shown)


net/redistest/redistest_test.go
```diff
@@ -15,3 +15,13 @@ func TestRedistest(t *testing.T) {
 		t.Fatalf("Failed to ping redis: %v", err)
 	}
 }
+
+func TestRedistestRedis6(t *testing.T) {
```
AlexanderYastrebov, line 19: This passes for 9.4.0 and 9.5.1 but fails for 9.5.0

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2950, opened 2024-02-21
routesrv: require kubernetes option

AlexanderYastrebov:
Require kubernetes option (explicitly or via -kubernetes flag).

Initially routesrv implied this option as it was envisioned for serving
Kubernetes dataclient routes.

Later an endpoint to serve Redis endpoints was added by #2237

Within #2476 skipper allows Redis endpoints discovery without using
kubernetes dataclient for routing.
This behaviour is configured by the -kubernetes flag.

The plan is to allow routesrv the same - provide Redis endpoints without
using Kubernetes dataclient to create routes.

This change is the first step that converts implied Kubernetes option
into required.

This is a breaking change for users that run routesrv without
-kubernetes flag or use it from the library without setting Kubernetes
option.

For #2476


RomanZavodskikh:
From my side is okay, because I see no obvious reasons to use routesrv outside of kubernetes.

RomanZavodskikh:
:+1:

routesrv/routesrv.go
```diff
@@ -85,6 +86,10 @@ func New(opts skipper.Options) (*RouteServer, error) {
 		supportHandler.Handle("/debug/pprof/", metricsHandler)
 	}
 
+	if !opts.Kubernetes {
+		return nil, fmt.Errorf("kubernetes option is required")
```
szuecs, line 90: maybe better:
```
`option "Kubernetes" is required`
```

AlexanderYastrebov:
:+1:

PR #2953, opened 2024-02-22
filters/auth: add grant cookie encoder

AlexanderYastrebov:
Add CookerEncoder interface to allow custom implementation of grant cookie encoding.

For example custom implementation may store token value in some permanent key-value storage and encode key into the cookie.

Another implementation may encode token value into multiple cookies.

### DIFF
--- a/filters/auth/grantcookie.go
+++ b/filters/auth/grantcookie.go
@@ -21,27 +21,31 @@ type CookieEncoder interface {
 	Read(request *http.Request) (*oauth2.Token, error)
 }
 
+// EncryptedCookieEncoder is a CookieEncoder that encrypts the token before storing it in a cookie.
 type EncryptedCookieEncoder struct {
-	config *OAuthConfig
+	Encryption       secrets.Encryption
+	CookieName       string
+	RemoveSubdomains int
+	Insecure         bool
 }
 
 var _ CookieEncoder = &EncryptedCookieEncoder{}
 
 func (ce *EncryptedCookieEncoder) Update(request *http.Request, token *oauth2.Token) ([]*http.Cookie, error) {
 	if token != nil {
-		c, err := createCookie(ce.config, request.Host, token)
+		c, err := ce.createCookie(request.Host, token)
 		if err != nil {
 			return nil, err
 		}
 		return []*http.Cookie{c}, nil
 	} else {
-		c := createDeleteCookie(ce.config, request.Host)
+		c := ce.createDeleteCookie(request.Host)
 		return []*http.Cookie{c}, nil
 	}
 }
 
 func (ce *EncryptedCookieEncoder) Read(request *http.Request) (*oauth2.Token, error) {
-	c, err := extractCookie(request, ce.config)
+	c, err := ce.extractCookie(request)
 	if err != nil {
 		return nil, err
 	}
@@ -61,19 +65,14 @@ type cookie struct {
 	Domain       string    `json:"domain,omitempty"`
 }
 
-func decodeCookie(cookieHeader string, config *OAuthConfig) (c *cookie, err error) {
+func (ce *EncryptedCookieEncoder) decodeCookie(cookieHeader string) (c *cookie, err error) {
 	var eb []byte
 	if eb, err = base64.StdEncoding.DecodeString(cookieHeader); err != nil {
 		return
 	}
 
-	var encryption secrets.Encryption
-	if encryption, err = config.Secrets.GetEncrypter(secretsRefreshInternal, config.SecretFile); err != nil {
-		return
-	}
-
 	var b []byte
-	if b, err = encryption.Decrypt(eb); err != nil {
+	if b, err = ce.Encryption.Decrypt(eb); err != nil {
 		return
 	}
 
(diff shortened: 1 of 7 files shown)


filters/auth/grantcookie.go
```diff
@@ -12,6 +12,33 @@ import (
 	"golang.org/x/oauth2"
 )
 
+type CookieEncoder interface {
```
AlexanderYastrebov, line 15: Naming?

filters/auth/grantcookie.go
```diff
@@ -12,6 +12,33 @@ import (
...
+	Create(config *OAuthConfig, host string, t *oauth2.Token) (*http.Cookie, error)
```
szuecs, line 16: I am not sure if the interface is really useful, because OAuthConfig is likely not working for oidc filters so it's only for grantflow and why should every filter invent their own?

AlexanderYastrebov: Interface is for users to provide pluggable implementation. I names OAuthConfig field `GrantCookieEncoder` to highlight that this is only fro grant flow.

filters/auth/grantcookie_test.go
```diff
@@ -22,8 +22,11 @@ func NewGrantCookieWithExpiration(config *OAuthConfig, expiry time.Time) (*http.
 		Expiry:       expiry,
 	}
 
-	cookie, err := createCookie(config, "", token)
-	return cookie, err
+	cookies, err := config.GrantCookieEncoder.Create(config, "", token)
```
AlexanderYastrebov, line 25: Lets have https://github.com/zalando/skipper/pull/2960 first

AlexanderYastrebov: Done.

filters/auth/grantcookie.go
```diff
@@ -12,6 +12,37 @@ import (
 	"golang.org/x/oauth2"
 )
 
+type CookieEncoder interface {
+	Create(config *OAuthConfig, host string, t *oauth2.Token) ([]*http.Cookie, error)
+	Extract(config *OAuthConfig, request *http.Request) (*oauth2.Token, error)
```
AlexanderYastrebov, line 17: Missing: delete/cleanup cookies

filters/auth/grantcookie.go
AlexanderYastrebov: Do we want cookie encoder or a generic token encoder that works with http.Response and may encode/extract token into/from cookie or e.g. headers?

AtorVdm: The existing Oauth2 approach here is designed to work with a cookie. Cookie-specific is fine for us unless you think it can be used somewhere else

filters/auth/grantcookie.go
szuecs: I understand what "interface" is, but why do we need it?
Do we have a user that wants to have its own implementation or do you think it might make sense to have as a user your own implementation?

filters/auth/grantcookie.go
AlexanderYastrebov: @AtorVdm I've updated CookieEncoder interface, please check if its suitable for the implementation you have in mind.

MustafaSaber:
👍

filters/auth/grantcookie.go
szuecs: We discussed that this interface is here such that library users can add their own implementation.

skipper.go
```diff
@@ -957,6 +961,33 @@ func (o *Options) KubernetesDataClientOptions() kubernetes.Options {
...
+	oauthConfig.RevokeTokenURL = o.OAuth2RevokeTokenURL
+	oauthConfig.TokeninfoURL = o.OAuthTokeninfoURL
+	oauthConfig.SecretFile = o.OAuth2SecretFile
+	oauthConfig.ClientID = o.OAuth2ClientID
+	oauthConfig.ClientSecret = o.OAuth2ClientSecret
+	oauthConfig.ClientIDFile = o.OAuth2ClientIDFile
+	oauthConfig.ClientSecretFile = o.OAuth2ClientSecretFile
+	oauthConfig.CallbackPath = o.OAuth2CallbackPath
+	oauthConfig.AuthURLParameters = o.OAuth2AuthURLParameters
+	oauthConfig.Secrets = o.SecretsRegistry
```
RomanZavodskikh, line 978:
```
oauthConfig.SecretsProvider = grantSecrets
```
missing?

skipper.go
```diff
@@ -957,6 +961,33 @@ func (o *Options) KubernetesDataClientOptions() kubernetes.Options {
...
+	oauthConfig.AccessTokenHeaderName = o.OAuth2AccessTokenHeaderName
+	oauthConfig.TokeninfoSubjectKey = o.OAuth2TokeninfoSubjectKey
+	oauthConfig.GrantTokeninfoKeys = o.OAuth2GrantTokeninfoKeys
+	oauthConfig.TokenCookieName = o.OAuth2TokenCookieName
+	oauthConfig.TokenCookieRemoveSubdomains = &o.OAuth2TokenCookieRemoveSubdomains
+	oauthConfig.Insecure = o.OAuth2GrantInsecure
+	oauthConfig.ConnectionTimeout = o.OAuthTokeninfoTimeout
+	oauthConfig.MaxIdleConnectionsPerHost = o.IdleConnectionsPerHost
```
RomanZavodskikh, line 986:
```
oauthConfig.Tracer = tracer
```
missing here as well?

skipper.go
RomanZavodskikh: Okay, I see it below
https://github.com/zalando/skipper/pull/2953/files#diff-9b0b16d06f27b640eb77f6ba0dee84df6a660528e5cccd81839e0330c4b92dd2R1815

skipper.go
RomanZavodskikh: https://github.com/zalando/skipper/pull/2953/files#diff-9b0b16d06f27b640eb77f6ba0dee84df6a660528e5cccd81839e0330c4b92dd2R1815

RomanZavodskikh:
:+1:

PR #2954, opened 2024-02-23
cmd/skipper: fix version and commit for go install

AlexanderYastrebov:
When skipper is installed via `go install` it has empty version and commit values:
```
~$ go install github.com/zalando/skipper/cmd/skipper@latest
go: downloading github.com/zalando/skipper v0.21.2
~$ skipper -version
Skipper version  (commit: , runtime: go1.22.0)
```

Current build process of binaries specifies version and commit via `-ldflags`.

This change uses debug buildinfo to get version and commit if they are not specified by `-ldflags` which is compatible with the current build process but should also work for `go install` case.

If this change fixes `go install` case (which can only be tested after package version is published) then we can add the same snippet to other binaries and get rid of COMMIT_HASH ldflag.

We can not get rid of `-ldflags` fully though because `go build` does not stamp version (see https://github.com/golang/go/issues/50603).

### DIFF
--- a/cmd/skipper/main.go
+++ b/cmd/skipper/main.go
@@ -17,6 +17,7 @@ package main
 import (
 	"fmt"
 	"runtime"
+	"runtime/debug"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/zalando/skipper"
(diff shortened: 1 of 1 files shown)


cmd/skipper/main.go
```diff
@@ -28,6 +29,21 @@ var (
 	commit  string
 )
 
+func init() {
+	if info, ok := debug.ReadBuildInfo(); ok {
+		if version == "" {
+			version = info.Main.Version
```
AlexanderYastrebov, line 35: go install should populate this from the package version that matches git tag.

szuecs:
:+1:

RomanZavodskikh:
:+1:

AlexanderYastrebov:
The version is set now but commit is not
```
~$ rm $(which skipper)
~$ go install github.com/zalando/skipper/cmd/skipper@latest
~$ skipper -version
Skipper version v0.21.3 (commit: , runtime: go1.22.0)

```

PR #2955, opened 2024-02-23
cmd/skipper: do not show empty commit

AlexanderYastrebov:
Binary installed via `go install github.com/zalando/skipper/cmd/skipper@latest` does not contain `vcs.revision`, see https://github.com/golang/go/issues/65904

Followup on #2954


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2934, opened 2024-02-13
skipper: use kubernetes cluster client to discover redis shards

AlexanderYastrebov:
Add LoadEndpointAddresses to kubernetes dataclient that is similar to caching GetEndpointAddresses but does not rely on previous call to Load or LoadAll.

Fixes #2476

PR #2956, opened 2024-02-23
routesrv: discover redis shards without fetching full cluster state

AlexanderYastrebov:
- [ ] tests

Routesrv supports dynamic discovery of Redis instances using Kubernetes Service, see https://github.com/zalando/skipper/blob/master/docs/tutorials/ratelimit.md#redis-based-cluster-ratelimits

Dynamic discovery relies on Kubernetes Dataclient that also fetches Ingresses and RouteGroups and creates routes from them.

This change enables dynamic discovery of Redis instances using Kubernetes without fetching Ingresses and RouteGroups.

When -kubernetes flag is not set but
-kubernetes-redis-service-namespace, -kubernetes-redis-service-name and -kubernetes-redis-service-port are provided then `routesrv` creates Kubernetes dataclient to discover Redis instances
but does not use this dataclient to load routes.

See previous #2934

### DIFF
--- a/routesrv/routesrv.go
+++ b/routesrv/routesrv.go
@@ -2,7 +2,6 @@ package routesrv
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"os"
 	"os/signal"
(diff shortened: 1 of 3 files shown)


routesrv/routesrv.go
```diff
@@ -86,14 +87,20 @@ func New(opts skipper.Options) (*RouteServer, error) {
...
 	}
 
-	if !opts.Kubernetes {
-		return nil, fmt.Errorf(`option "Kubernetes" is required`)
+	var (
+		kdc        *kubernetes.Client
+		dataclient routing.DataClient
+	)
+	if opts.Kubernetes {
+		kdc, err = kubernetes.New(opts.KubernetesDataClientOptions())
```
AlexanderYastrebov, line 95: TODO: close

PR #2961, opened 2024-02-28
minor: countPathRoutes function

RomanZavodskikh:
The name could be a little bit misleading because actually this function calculates the number of Path{,Subtree} predicates in the single route. Nobody calculates the number of routes here, there is only one route which this function takes as input.
Therefore n variable corresponds to the number of predicates, not the routes how it seems to be when taking a look at the code.

### DIFF
--- a/dataclients/kubernetes/ingress.go
+++ b/dataclients/kubernetes/ingress.go
@@ -195,11 +195,11 @@ func addExtraRoutes(ic *ingressContext, ruleHost, path, pathType, eastWestDomain
 			ruleHost+strings.ReplaceAll(path, "/", "_"),
 			extraIndex)
 		setPathV1(ic.pathMode, &route, pathType, path)
-		if n := countPathRoutes(&route); n <= 1 {
+		if n := countPathPredicates(&route); n <= 1 {
 			ic.addHostRoute(ruleHost, &route)
 			ic.redirect.updateHost(ruleHost)
 		} else {
-			ic.logger.Errorf("Failed to add route having %d path routes: %v", n, r)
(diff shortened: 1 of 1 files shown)


dataclients/kubernetes/ingress.go
```diff
@@ -195,11 +195,11 @@ func addExtraRoutes(ic *ingressContext, ruleHost, path, pathType, eastWestDomain
 			ruleHost+strings.ReplaceAll(path, "/", "_"),
 			extraIndex)
 		setPathV1(ic.pathMode, &route, pathType, path)
-		if n := countPathRoutes(&route); n <= 1 {
+		if n := countPathPredicates(&route); n <= 1 {
 			ic.addHostRoute(ruleHost, &route)
 			ic.redirect.updateHost(ruleHost)
 		} else {
-			ic.logger.Errorf("Failed to add route having %d path routes: %v", n, r)
+			ic.logger.Errorf("Failed to add route having %d path predicates: %v", n, r)
```
AlexanderYastrebov, line 202: This should log `route` not `r` because second path was added by setPathV1
Also maybe rephrase it like `Ignoring route %v due to multiple path predicates` (since we print route we don't care about n)

RomanZavodskikh: I agree with `route` instead of `r`, good catch. :-)
As for log message, I reprased it somewhat like you proposed. I still think we should print `n` as well to have a certainty about how many path predicates skipper counted. During the debugging, there could be some doubt whether this `n` is equal to the amount of predicate developer debugging this error will count. I would like to reduce this doubt.

AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

PR #2984, opened 2024-03-14
proxy: fix proxy panic caused by broken backend

AlexanderYastrebov:
Broken backend that closes connection before sending response body may trigger proxy panic if client sends `Expect: 100-continue` header, see https://github.com/golang/go/issues/53808


proxy/context.go
```diff
@@ -192,9 +192,9 @@ func (c *context) shunted() bool {
 	return c.servedWithResponse
 }
 
-func (c *context) setResponse(r *http.Response, preserveOriginal bool) {
```
AlexanderYastrebov, line 195: Decided to cleanup this although its unrelated

proxy/proxy.go
```diff
@@ -1540,15 +1542,6 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	ctx.parentSpan = span
 
-	defer func() {
-		if ctx.response != nil && ctx.response.Body != nil {
-			err := ctx.response.Body.Close()
-			if err != nil {
-				ctx.Logger().Errorf("error during closing the response body: %v", err)
-			}
-		}
-	}()
```
AlexanderYastrebov, line 1550: Decided not to pile-up defers but to use straightforward cleanup sequence after streaming the response.
Would like to refactor other defers as well separately.

PR #2986, opened 2024-03-15
Refactor healthy_endpoints_test.go

RomanZavodskikh:

### DIFF
--- a/proxy/healthy_endpoints_test.go
+++ b/proxy/healthy_endpoints_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/zalando/skipper/routing"
 )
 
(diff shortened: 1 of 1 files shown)


proxy/healthy_endpoints_test.go
```diff
@@ -26,6 +26,39 @@ func defaultEndpointRegistry() *routing.EndpointRegistry {
 	})
 }
 
+func sendGetRequest(t *testing.T, ps *httptest.Server, consistentHashKey int) (*http.Response, error) {
+	req, err := http.NewRequest("GET", ps.URL, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
```
AlexanderYastrebov, line 33: Since you are using assert you may also use `require.NoError(t, err)` for consistency.

proxy/healthy_endpoints_test.go
```diff
@@ -26,6 +26,39 @@ func defaultEndpointRegistry() *routing.EndpointRegistry {
...
+
+	rsp, err := ps.Client().Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	return rsp, nil
+}
+
+func sendGetRequests(t *testing.T, ps *httptest.Server) (int, error) {
```
AlexanderYastrebov, line 44: It never returns error afaict

AlexanderYastrebov: minor: maybe merge two helpers into one `getN(t *testing.T, ps *httptest.Server, n int) (failed int)`

RomanZavodskikh: About never returning error: yes, I agree, it is better to fail in the most inner function everywhere to avoid a huge amount of `if err != nil {...` everywhere, fixed it.

RomanZavodskikh: About two helpers: I still would like to keep to two helpers, because there are tests:
+ where only one request is sent https://github.com/zalando/skipper/pull/2986/files#diff-b1bbc8a9c24c3b87d0de99e08ebcdd6df2da583d27fa530049a3cc12d496435cR82
+ and where a lot requests are sent https://github.com/zalando/skipper/pull/2986/files#diff-b1bbc8a9c24c3b87d0de99e08ebcdd6df2da583d27fa530049a3cc12d496435cR183

Therefore there is one helper for first test and another helper for the second one.

AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

PR #2998, opened 2024-03-26
Fixed PHC for consistentHash with balance factor

RomanZavodskikh:
Fixed PHC for consistentHash with balance factor and made unit test cover this type of load balancing as well.

### DIFF
--- a/proxy/healthy_endpoints_test.go
+++ b/proxy/healthy_endpoints_test.go
@@ -63,30 +63,42 @@ func TestPHCWithoutRequests(t *testing.T) {
 	}
 
 	for _, algorithm := range []string{"random", "consistentHash", "roundRobin", "powerOfRandomNChoices"} {
-		t.Run(algorithm, func(t *testing.T) {
-			endpointRegistry := defaultEndpointRegistry()
-			doc := fmt.Sprintf(`* -> <%s, "%s", "%s", "%s">`,
-				algorithm, services[0].URL, services[1].URL, services[2].URL)
+		balanceFactors := []bool{false}
+		if algorithm == "consistentHash" {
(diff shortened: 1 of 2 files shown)


proxy/healthy_endpoints_test.go
```diff
@@ -125,26 +137,38 @@ func TestPHCForMultipleHealthyEndpoints(t *testing.T) {
...
+					EnablePassiveHealthCheck: true,
+					EndpointRegistry:         endpointRegistry,
+				})
+				if err != nil {
+					t.Fatal(err)
+				}
+				defer tp.close()
+
+				ps := httptest.NewServer(tp.proxy)
+				defer ps.Close()
```
AlexanderYastrebov, line 166: This could be a helper that sets up backends and receives a route template to make more readable.
Then each combination could be a nested t.Run:
```go
t.Run("consistent hash with balance factor", func(t *testing.T) {
  setupProxy(t, `* -> consistentHashKey("${request.header.ConsistentHashKey}") -> consistentHashBalanceFactor(1.25) -> <consitentHash, "%s", "%s", "%s">`, services[0].URL, services[1].URL, services[2].URL)
  failedReqs := sendGetRequests(t, ps)
  assert.Equal(t, 0, failedReqs)
})
```

RomanZavodskikh:
:+1:

proxy/healthy_endpoints_test.go
RomanZavodskikh: It seems to be slightly more compact, but I'm not sure which code looks better. We can continue discussion here not to block deploying fix:
https://github.com/zalando/skipper/pull/2999

PR #3013, opened 2024-04-08
script: support multiple header values

AlexanderYastrebov:
Introduce `add` and `values` methods to set and get multiple header values.

### DIFF
--- a/docs/reference/scripts.md
+++ b/docs/reference/scripts.md
@@ -137,6 +137,32 @@ ctx.request.header["Authorization"] = nil -- delete authorization header
 
 Response headers `ctx.response.header` work the same way - this is of course only valid in the `response()` phase.
 
+### Multiple header values
+
+Request and response header tables provide access to a first value of a header.
+
+To access multiple values use `add` and `values` methods:
+
+```lua
+function request(ctx, params)
+	ctx.request.header.add("X-Foo", "Bar")
+	ctx.request.header.add("X-Foo", "Baz")
+
+	-- all X-Foo values
+	for _, v in pairs(ctx.request.header.values("X-Foo")) do
+		print(v)
+	end
+
+	-- all values
+	for k, _ in ctx.request.header() do
+		for _, v in pairs(ctx.request.header.values(k)) do
+			print(k, "=", v)
+		end
+	end
+end
+```
+
+
 ## Other request fields
 
 * `backend_url` - (read only) returns the backend url specified in the route
--- a/script/script.go
+++ b/script/script.go
@@ -465,7 +465,10 @@ func getRequestValue(f filters.FilterContext) func(*lua.LState) int {
 		switch key {
 		case "header":
 			if header == nil {
-				header = s.CreateTable(0, 0)
+				header = s.CreateTable(0, 2)
+				header.RawSetString("add", s.NewFunction(addRequestHeader(f)))
+				header.RawSetString("values", s.NewFunction(requestHeaderValues(f)))
+
 				mt := s.CreateTable(0, 3)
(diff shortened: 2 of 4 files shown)


script/testdata/log_header.lua
```diff
@@ -18,16 +18,16 @@ function response(ctx, params)
...
-	for k, v in iter() do
+	for k, _ in header() do
 		if k:lower() == "authorization" then
 			table.insert(t, k)
 			table.insert(t, ": TRUNCATED\r\n")
 		else
 			table.insert(t, k)
 			table.insert(t, ": ")
-			table.insert(t, v)
+			table.insert(t, table.concat(header.values(k), " "))
```
AlexanderYastrebov, line 30: This is what `logHeader` filter does.

script/script.go
```diff
@@ -677,6 +683,30 @@ func setRequestHeader(f filters.FilterContext) func(*lua.LState) int {
 	}
 }
 
+func addRequestHeader(f filters.FilterContext) func(*lua.LState) int {
```
AlexanderYastrebov, line 686: It should be possible to generalize this and reuse for request and response headers but for simplicity I decided to introduce separate functions to follow the style of existing single value setters/getters and also because they return closures.

docs/reference/scripts.md
```diff
@@ -137,6 +137,32 @@ ctx.request.header["Authorization"] = nil -- delete authorization header
...
+### Multiple header values
+
+Request and response header tables provide access to a first value of a header.
+
+To access multiple values use `add` and `values` methods:
+
+```lua
+function request(ctx, params)
+	ctx.request.header.add("X-Foo", "Bar")
+	ctx.request.header.add("X-Foo", "Baz")
```
szuecs, line 149: Will this result in:
```
X-Foo: Bar, Baz
```
or 
```
X-Foo: Bar
X-Foo: Baz
```

Maybe this would be good to clarify

script/script.go
```diff
@@ -677,6 +683,30 @@ func setRequestHeader(f filters.FilterContext) func(*lua.LState) int {
...
+	return func(s *lua.LState) int {
+		value := s.ToString(-1)
```
RomanZavodskikh, line 688: minor: where did -1 and -2 came from? Can I use -3 and -4, for example?

RomanZavodskikh: Ah, okay, I see, there is stack-based API.
https://pkg.go.dev/github.com/yuin/gopher-lua@v1.1.1#section-readme

docs/reference/scripts.md
szuecs: Tested: works as expected
```
X-Foo: Bar
X-Foo: Baz
```

script/script_test.go
```diff
@@ -61,16 +62,75 @@ func TestScript(t *testing.T) {
...
+					ctx.request.header.add("Foo", "Baz")
+
+					ctx.request.header["Qux"] = table.concat(ctx.request.header.values("Foo"), ", ")
+					ctx.request.header["Absent-Length"] = table.getn(ctx.request.header.values("Absent"))
+				end`,
+			},
+			expectedRequestHeader: http.Header{
+				"Foo":           []string{"Bar", "Baz"},
+				"Qux":           []string{"Bar, Baz"},
+				"Absent-Length": []string{"0"},
```
RomanZavodskikh, line 91: Should we test `Foo-Length` (I expect it to be 2) and `Qux-Length` (I expect it to be 1) here as well?

AlexanderYastrebov: I added this case to show that length of empty values is zero

szuecs:
:+1:

RomanZavodskikh:
:+1:

PR #3014, opened 2024-04-08
fix: vulnerable dependency go-jose

szuecs:
https://github.com/square/go-jose was archived in favor of https://github.com/go-jose/go-jose, so dependency change makes sense


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #3020, opened 2024-04-15
filters/auth: add jwtMetrics

AlexanderYastrebov:
Add jwtMetrics filter that parses JWT token and increments a set of counters,
see documentation for details.

### DIFF
--- a/filters/auth/jwt_metrics.go
+++ b/filters/auth/jwt_metrics.go
@@ -1,0 +1,92 @@
+package auth
+
+import (
+	"fmt"
+	"regexp"
+	"slices"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/jwt"
+)
+
+type (
+	jwtMetricsSpec struct{}
+
+	jwtMetricsFilter struct {
+		Issuers []string `json:"issuers,omitempty"`
+	}
+)
+
+func NewJwtMetrics() filters.Spec {
+	return &jwtMetricsSpec{}
+}
+
+func (s *jwtMetricsSpec) Name() string {
+	return filters.JwtMetricsName
+}
+
+func (s *jwtMetricsSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	f := &jwtMetricsFilter{}
+
+	if len(args) == 1 {
+		if config, ok := args[0].(string); !ok {
+			return nil, fmt.Errorf("requires single string argument")
+		} else if err := yaml.Unmarshal([]byte(config), f); err != nil {
+			return nil, fmt.Errorf("failed to parse configuration")
+		}
+	} else if len(args) > 1 {
+		return nil, fmt.Errorf("requires single string argument")
+	}
+
+	return f, nil
+}
+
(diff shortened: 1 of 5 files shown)


filters/auth/jwt_metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	response := ctx.Response()
+
+	if response.StatusCode >= 400 && response.StatusCode < 500 {
+		return // ignore invalid requests
+	}
+
+	request := ctx.Request()
+
+	metrics := ctx.Metrics()
+	metricsPrefix := fmt.Sprintf("%s.%s.%d.", request.Method, escapeMetricKeySegment(request.Host), response.StatusCode)
```
szuecs, line 58: Maybe you can ad to the docs the expected memory consumption:
`5 * 9 * #hosts_configured_in_routes * #status_codes * int64`

filters/auth/jwt_metrics.go
```diff
@@ -0,0 +1,92 @@
...
+			metrics.IncCounter(metricsPrefix + "invalid-token")
+			return
+		}
+
+		// https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1
+		if issuer, ok := token.Claims["iss"].(string); !ok {
+			metrics.IncCounter(metricsPrefix + "missing-issuer")
+		} else if !slices.Contains(f.Issuers, issuer) {
+			metrics.IncCounter(metricsPrefix + "invalid-issuer")
+		}
```
szuecs, line 84: You do not count valid tokens, why?

AlexanderYastrebov: Its not a goal to reveal valid tokens, besides we don't validate so we could only know that it parses as JWT token.

filters/auth/jwt_metrics.go
AlexanderYastrebov: I described cardinality of the metrics key and I think the exact memory requirement could be hard to establish (where does 5 * 9 come from?)

filters/auth/jwt_metrics.go
szuecs: As a user I would wonder why we do not have a counter for valid tokens

filters/auth/jwt_metrics.go
szuecs: 5 are your error types
9 are the http methods

filters/auth/jwt_metrics.go
AlexanderYastrebov: The idea is to count configured claims (e.g. token realms) but this may be implemented later.

AlexanderYastrebov: At this stage `jwtMetrics` is essentially `invalidJwtMetrics` because it only counts wrong tokens. I do not want to put "invalid" into the name and we may extend it in the future to count other claims.

filters/auth/jwt_metrics.go
AlexanderYastrebov: Added.

filters/auth/jwt_metrics.go
szuecs: ok

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #3022, opened 2024-04-18
filters/builtin: add annotate filter

AlexanderYastrebov:
Add `annotate` filter to annotate routes and a helper to retrieve annotations.


szuecs:
:+1:

PR #3023, opened 2024-04-18
filters: move annotate into own package

AlexanderYastrebov:
This avoids import cycle when GetAnnotations helper is imported from subpackage of filters.

Followup on #3022


AlexanderYastrebov:
:+1:

Issue #3026, opened 2024-04-19
Deprecate metrics.Default


AlexanderYastrebov:
**Is your feature request related to a problem? Please describe.**

There are [quite a few places](https://github.com/search?q=repo%3Azalando%2Fskipper%20metrics.Default&type=code) that rely on global metrics.Default variable
https://github.com/zalando/skipper/blob/0ce11b9d18fddc00f65194e5338984d2aeb51a4b/metrics/metrics.go#L199-L202

This is a bad design which is not thread-safe and complicates tests, e.g.:
https://github.com/zalando/skipper/blob/0ce11b9d18fddc00f65194e5338984d2aeb51a4b/proxy/metrics_test.go#L19-L24

**Describe the solution you would like**

Components that update metrics receive metrics instance via configuration.
First we may refactor components to make metrics configurable and fallback to metrics.Default for backwards compatibility.
Then deprecate and later remove `metrics.Default` completely.

**Would you like to work on it?**

Yes

PR #3027, opened 2024-04-19
proxy: support configurable metrics

AlexanderYastrebov:
Use configed metrics and fallback to global for backwards compatibility.

Updates #3026

### DIFF
--- a/skipper.go
+++ b/skipper.go
@@ -1309,9 +1309,8 @@ func listenAndServeQuit(
 	}
 
 	if o.EnableConnMetricsServer {
-		m := metrics.Default
 		srv.ConnState = func(conn net.Conn, state http.ConnState) {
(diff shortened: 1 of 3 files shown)


skipper.go
```diff
@@ -1309,9 +1309,8 @@ func listenAndServeQuit(
 	}
 
 	if o.EnableConnMetricsServer {
-		m := metrics.Default
```
RomanZavodskikh, line 1312: minor: why were `metrics.Default` and `mtr` were mixed in this function at all?

RomanZavodskikh:
:+1:

skipper.go
AlexanderYastrebov: :shrug:

AlexanderYastrebov:
:+1:

PR #3029, opened 2024-04-19
refactor: remove EnablePassiveHealthCheck flag

RomanZavodskikh:
rely on PassiveHealthCheck struct pointer to be nil or not nil instead


szuecs:
I feel it's a bit too young feature to know how to set it up without operator input.

PR #3034, opened 2024-04-22
OPA: Add tracing for outbound http calls

mjungsbluth:
For the Open Policy Agent filters, tracing information is currently mainly added for the evaluation path of the filters. This is useful for monitoring everything that is related to requests passing through Skipper.
The PR adds tracing for asynchronous http calls that happen during discovery, bundle download, status updates and decision uppload. This allows to monitor how the OPA instances interact with their control plane.

### DIFF
--- a/filters/openpolicyagent/evaluation.go
+++ b/filters/openpolicyagent/evaluation.go
@@ -11,7 +11,6 @@ import (
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/server"
-	"github.com/open-policy-agent/opa/tracing"
 	"github.com/opentracing/opentracing-go"
 )
 
@@ -58,7 +57,7 @@ func (opa *OpenPolicyAgentInstance) Eval(ctx context.Context, req *ext_authz_v3.
 		return nil, err
 	}
 
-	err = envoyauth.Eval(ctx, opa, inputValue, result, rego.DistributedTracingOpts(tracing.Options{opa}))
+	err = envoyauth.Eval(ctx, opa, inputValue, result, rego.DistributedTracingOpts(buildTracingOptions(opa.registry.tracer, opa.bundleName, opa.manager)))
 	if err != nil {
 		return nil, err
 	}
--- a/filters/openpolicyagent/evaluation.go
+++ b/filters/openpolicyagent/evaluation.go
@@ -57,7 +57,7 @@ func (opa *OpenPolicyAgentInstance) Eval(ctx context.Context, req *ext_authz_v3.
 		return nil, err
 	}
 
-	err = envoyauth.Eval(ctx, opa, inputValue, result, rego.DistributedTracingOpts(buildTracingOptions(opa.registry.tracer, opa.bundleName, opa.manager)))
+	err = envoyauth.Eval(ctx, opa, inputValue, result, rego.DistributedTracingOpts(opa.DistributedTracing()))
 	if err != nil {
 		return nil, err
 	}
--- a/filters/openpolicyagent/openpolicyagent.go
+++ b/filters/openpolicyagent/openpolicyagent.go
@@ -47,6 +47,8 @@ const (
 	DefaultMaxRequestBodySize   = 1 << 20 // 1 MB
 	DefaultMaxMemoryBodyParsing = 100 * DefaultMaxRequestBodySize
 	defaultBodyBufferSize       = 8192 * 1024
+
+	spanNameEval = "open-policy-agent"
 )
 
 type OpenPolicyAgentRegistry struct {
@@ -404,9 +406,13 @@ func interpolateConfigTemplate(configTemplate []byte, bundleName string) ([]byte
 	return buf.Bytes(), nil
 }
 
+func buildTracingOptions(tracer opentracing.Tracer, bundleName string, manager *plugins.Manager) opatracing.Options {
+	return opatracing.NewOptions(WithTracingOptTracer(tracer), WithTracingOptBundleName(bundleName), WithTracingOptManager(manager))
+}
+
 func (registry *OpenPolicyAgentRegistry) withTracingOptions(bundleName string) func(*plugins.Manager) {
 	return func(m *plugins.Manager) {
-		options := opatracing.NewOptions(
+		options := buildTracingOptions(
 			registry.tracer,
 			bundleName,
 			m,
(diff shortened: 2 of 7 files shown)


filters/openpolicyagent/tracing.go
```diff
@@ -14,14 +20,43 @@ func init() {
...
+	manager    *plugins.Manager
+
 	wrapped http.RoundTripper
 }
 
 func (*tracingFactory) NewTransport(tr http.RoundTripper, opts opatracing.Options) http.RoundTripper {
+	var tracer opentracing.Tracer
+
+	if len(opts) < 3 {
+		panic("insufficient tracing options provided for outbound http calls from Open Policy Agent")
```
szuecs, line 34: panic -> log.Fatal, but only fatal in case it's detected at startup. Everything that could happen at runtime -> Error instead.

AlexanderYastrebov: I do not like panic/fatal - we dont use any where exept command as this is not library friendly. Either we should change to return error or make it such that it works without options or ignores wrong options (and logs an error).

filters/openpolicyagent/tracing.go
```diff
@@ -14,14 +20,43 @@ func init() {
...
+
+	bundleName, ok := opts[1].(string)
+	if !ok {
+		panic("invalid type for second argument of tracing options, expected string")
+	}
+
+	manager, ok := opts[2].(*plugins.Manager)
+	if !ok {
+		panic("invalid type third argument of tracing options, expected *plugins.Manager")
+	}
```
AlexanderYastrebov, line 53: This imposes an order on the options which is error-prone.
One way is to process them in loop (and for string arg define typed alias `type BundleName string`):
```go
for _, o := range opts {
   switch o := o.(type) {
    ...
   }
}
```

Another way is to use functional options, see https://golang.cafe/blog/golang-functional-options-pattern.html

AlexanderYastrebov: I guess we can also change signature and supply Options struct

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -553,11 +583,7 @@ func (opa *OpenPolicyAgentInstance) startSpanFromContextWithTracer(tr opentracin
 		span = tracing.CreateSpan("open-policy-agent", ctx, tr)
```
AlexanderYastrebov, line 583: Can we move this span name to the const as well?

filters/openpolicyagent/tracing.go
mjungsbluth: We sadly cannot change the interface to return an error as this is coming from OPA itself. I also do not like panic/fatal but this way pretty much all OPA related tests will fail if the options are passed wrongly. 
I can also change this to just skip tracing if there is an error over here and log warnings but it will be tricky to have a meaningful test that actually validates this, the error will only happen in full integration testing with a filter and tracing.

filters/openpolicyagent/openpolicyagent.go
mjungsbluth: Absolutely!

filters/openpolicyagent/tracing.go
szuecs: Maybe reach out to OPA folks and create a PR that returns an error

mjungsbluth: I changed this so that it warns instead of failing which might degrade the level of tracing that is done. I also centralised the construction of the `Options` so that a mismatch would be covered with the tests for tracing.

The interface is public so would break other OPA users, filed an upstream issue here: https://github.com/open-policy-agent/opa/issues/6710 to see if we can change the interface...

filters/openpolicyagent/tracing.go
mjungsbluth: Changed this to functional options as this fits with the rest of the filters which also use the same pattern...

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -730,7 +766,7 @@ func (opa *OpenPolicyAgentInstance) Config() *config.Config { return opa.opaConf
 
 // DistributedTracing is an implementation of the envoyauth.EvalContext interface
 func (opa *OpenPolicyAgentInstance) DistributedTracing() opatracing.Options {
-	return opatracing.NewOptions(opa)
+	return opatracing.NewOptions(opa.registry.tracer, opa.bundleName, opa.manager)
```
AlexanderYastrebov, line 769: I think this should wrap params via `With*`

filters/openpolicyagent/evaluation.go
```diff
@@ -58,7 +57,7 @@ func (opa *OpenPolicyAgentInstance) Eval(ctx context.Context, req *ext_authz_v3.
 		return nil, err
 	}
 
-	err = envoyauth.Eval(ctx, opa, inputValue, result, rego.DistributedTracingOpts(tracing.Options{opa}))
+	err = envoyauth.Eval(ctx, opa, inputValue, result, rego.DistributedTracingOpts(buildTracingOptions(opa.registry.tracer, opa.bundleName, opa.manager)))
```
AlexanderYastrebov, line 60: Shouldn't this use `func (opa *OpenPolicyAgentInstance) DistributedTracing() opatracing.Options`?

AlexanderYastrebov:
Looks good modulo minor comments.

filters/openpolicyagent/openpolicyagent.go
mjungsbluth: Good catch. Changed this...

filters/openpolicyagent/evaluation.go
mjungsbluth: Yes, that would be simpler. Made that change...

AlexanderYastrebov:
:+1:

filters/openpolicyagent/tracing.go
```diff
@@ -14,15 +21,48 @@ func init() {
...
+	log := &logging.DefaultLog{}
+
+	wrapper := &transport{
 		wrapped: tr,
 	}
+
+	for _, o := range opts {
+		opt, ok := o.(func(*transport))
+		if !ok {
+			log.Warn("invalid type for OPA tracing option, expected func(*transport), tracing information might be incomplete")
```
szuecs, line 59: should we log "%T" for "o" ?

mjungsbluth: Done...

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #3035, opened 2024-04-22
phc: add a metric counter for endpoints that opted out before load balancing

MustafaSaber:
related to https://github.com/zalando/skipper/issues/2346

### DIFF
--- a/docs/operation/operation.md
+++ b/docs/operation/operation.md
@@ -925,7 +925,7 @@ while choosing the endpoint for the given request
 
 A set of metrics will be exposed to track passive health checks:
 
-* `routing.endpoint.drop.loadbalancer`: Number of endpoints dropped before load balancing the requests 
+* `routing.endpoint.drop.loadbalancer`: Number of all endpoints dropped before load balancing a request, so after N requests and M endpoints are being dropped this counter would be N*M.
 
 ## Memory consumption
 
--- a/docs/operation/operation.md
+++ b/docs/operation/operation.md
@@ -921,6 +921,12 @@ The parameters of `-passive-health-check` option are:
 + `max-drop-probabilty=<float more than/equal to 0 and less than/equal to 1>` - the maximum possible probability of unhealthy endpoint being not considered
 while choosing the endpoint for the given request
 
+### Metrics
+
+A set of metrics will be exposed to track passive health checks:
+
+* `routing.endpoint.drop.loadbalancer`: Number of endpoints dropped before load balancing the requests 
+
 ## Memory consumption
 
 While Skipper is generally not memory bound, some features may require
--- a/docs/operation/operation.md
+++ b/docs/operation/operation.md
@@ -923,9 +923,9 @@ while choosing the endpoint for the given request
 
 ### Metrics
 
-A set of metrics will be exposed to track passive health checks:
+A set of metrics will be exposed to track passive health check:
 
-* `routing.endpoint.drop.loadbalancer`: Number of all endpoints dropped before load balancing a request, so after N requests and M endpoints are being dropped this counter would be N*M.
+* `passive-health-check.endpoints.dropped`: Number of all endpoints dropped before load balancing a request, so after N requests and M endpoints are being dropped this counter would be N*M.
 
 ## Memory consumption
 
--- a/proxy/healthy_endpoints.go
+++ b/proxy/healthy_endpoints.go
@@ -3,6 +3,7 @@ package proxy
 import (
 	"math/rand"
 
+	"github.com/zalando/skipper/metrics"
 	"github.com/zalando/skipper/routing"
 )
 
@@ -11,7 +12,7 @@ type healthyEndpoints struct {
 	endpointRegistry *routing.EndpointRegistry
 }
 
-func (h *healthyEndpoints) filterHealthyEndpoints(ctx *context, endpoints []routing.LBEndpoint) []routing.LBEndpoint {
+func (h *healthyEndpoints) filterHealthyEndpoints(ctx *context, endpoints []routing.LBEndpoint, metrics metrics.Metrics) []routing.LBEndpoint {
 	if h == nil {
 		return endpoints
 	}
@@ -24,6 +25,7 @@ func (h *healthyEndpoints) filterHealthyEndpoints(ctx *context, endpoints []rout
 		if p < dropProbability {
 			ctx.Logger().Infof("Dropping endpoint %q due to passive health check: p=%0.2f, dropProbability=%0.2f",
 				e.Host, p, dropProbability)
+			metrics.IncCounter("routing.endpoint.drop.loadbalancer")
 		} else {
 			filtered = append(filtered, e)
 		}
--- a/proxy/healthy_endpoints.go
+++ b/proxy/healthy_endpoints.go
@@ -25,7 +25,7 @@ func (h *healthyEndpoints) filterHealthyEndpoints(ctx *context, endpoints []rout
 		if p < dropProbability {
 			ctx.Logger().Infof("Dropping endpoint %q due to passive health check: p=%0.2f, dropProbability=%0.2f",
 				e.Host, p, dropProbability)
-			metrics.IncCounter("routing.endpoint.drop.loadbalancer")
+			metrics.IncCounter("passive-health-check.endpoints.dropped")
 		} else {
 			filtered = append(filtered, e)
 		}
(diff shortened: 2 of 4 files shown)


proxy/healthy_endpoints.go
```diff
@@ -22,6 +23,7 @@ func (h *healthyEndpoints) filterHealthyEndpoints(endpoints []routing.LBEndpoint
 	for _, e := range endpoints {
 		if p < e.Metrics.HealthCheckDropProbability() {
 			/* drop */
+			metrics.IncCounter("routing.endpoint." + e.Host + ".optout.loadbalancer")
```
RomanZavodskikh, line 26: I think it is better to use Sprintf here:
`fmt.Sprintf("routing.endpoint.%s.optout.loadbalancer", e.Host)`.

proxy/healthy_endpoints.go
```diff
@@ -11,7 +12,7 @@ type healthyEndpoints struct {
 	endpointRegistry *routing.EndpointRegistry
 }
 
-func (h *healthyEndpoints) filterHealthyEndpoints(endpoints []routing.LBEndpoint, rt *routing.Route) []routing.LBEndpoint {
+func (h *healthyEndpoints) filterHealthyEndpoints(endpoints []routing.LBEndpoint, metrics metrics.Metrics) []routing.LBEndpoint {
```
RomanZavodskikh, line 15: I am a little bit confused about adding `metrics` as an argument here. I am working on another PR in parallel (to add logs), and at the moment I don't see any way to pass logger from `proxy.go` here other than adding logger inside arguments list.

proxy/healthy_endpoints.go
AlexanderYastrebov: Conactenation is fine but hostname should be escaped, see https://github.com/zalando/skipper/pull/3020

AlexanderYastrebov: Or maybe use https://github.com/zalando/skipper/blob/708e97af1d1e1511d7bc7803e1b2a0d5599df52b/metrics/utils.go#L21-L25

szuecs: How much memory we think this will take? IIRC int64
What if we have a non trivial number of all day crashing pods? what if we have 10k pods?
Why do we need `e.Host` in the metric?

We don't clean up metrics, so we store it for lifetime of the process. This host is not set by a client (unlike Host header for example), so maybe it's fine.

AlexanderYastrebov: This host is an IP of the pod so metrics size concern is very valid.

Maybe `optout -> drop` to stick to the parlance.

RomanZavodskikh: > (I think its always host+port)

Yes, it is, see https://github.com/zalando/skipper/pull/2824.

RomanZavodskikh, line 26: I think it makes sense to store not only amount of requests being dropped from particular endpoint (as you do in this line) but also to count amount of requests which have at least one endpoint being dropped. It could be done after all iterations of this loop.

MustafaSaber: I added `e.Host` to define which IP we escape then we can track on AWS

szuecs:
Please also add the metrics to operations.md

docs/operation/operation.md
```diff
@@ -921,6 +921,12 @@ The parameters of `-passive-health-check` option are:
 + `max-drop-probabilty=<float more than/equal to 0 and less than/equal to 1>` - the maximum possible probability of unhealthy endpoint being not considered
 while choosing the endpoint for the given request
 
+### PHC Metrics
```
RomanZavodskikh, line 924: minor: as this is already subsection of "Passive health check (*experimental*)", "PHC" here could be omitted. Just "Metrics", and that's it.

docs/operation/operation.md
```diff
@@ -921,6 +921,12 @@ The parameters of `-passive-health-check` option are:
 + `max-drop-probabilty=<float more than/equal to 0 and less than/equal to 1>` - the maximum possible probability of unhealthy endpoint being not considered
 while choosing the endpoint for the given request
 
+### Metrics
+
+A set of metrics will be exposed to track passive health checks:
+
+* `routing.endpoint.drop.loadbalancer`: Number of endpoints dropped before load balancing the requests
```
RomanZavodskikh, line 928: It seems to me such sentence could be misunderstood by external reader: it implicitly states that this is number of **distinct** endpoints dropped. However, this counter is incremented once per endpoint and per request.
So I would tell it is something like 'number of times some endpoint was dropped from endpoints' list before applying load balancer algorithm'.

MustafaSaber: hope the change address your concern

RomanZavodskikh:
:+1:

docs/operation/operation.md
```diff
@@ -921,6 +921,12 @@ The parameters of `-passive-health-check` option are:
 + `max-drop-probabilty=<float more than/equal to 0 and less than/equal to 1>` - the maximum possible probability of unhealthy endpoint being not considered
 while choosing the endpoint for the given request
 
+### Metrics
+
+A set of metrics will be exposed to track passive health checks:
```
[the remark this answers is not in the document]
MustafaSaber: I think it's ok if we use plural here.

proxy/healthy_endpoints.go
```diff
@@ -22,6 +23,7 @@ func (h *healthyEndpoints) filterHealthyEndpoints(endpoints []routing.LBEndpoint
 	for _, e := range endpoints {
 		if p < e.Metrics.HealthCheckDropProbability() {
 			/* drop */
+			metrics.IncCounter("routing.endpoint.drop.loadbalancer")
```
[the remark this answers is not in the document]
MustafaSaber: The idea was to indicate that an endpoint is dropped from load balancing, but naming could be changed if you think it's better.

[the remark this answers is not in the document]
szuecs: Please change.
Metrics are for users and don't need to reflect the code.

szuecs:
:+1:

proxy/healthy_endpoints_test.go
```diff
@@ -173,20 +171,19 @@ func TestPHCForMultipleHealthyAndOneUnhealthyEndpoints(t *testing.T) {
...
 	for _, algorithm := range []string{"random", "consistentHash", "roundRobin", "powerOfRandomNChoices"} {
 		t.Run(algorithm, func(t *testing.T) {
-			tp, ps := setupProxy(t, fmt.Sprintf(`* -> backendTimeout("5ms") -> consistentHashKey("${request.header.ConsistentHashKey}") -> <%s, "%s", "%s", "%s">`,
+			mockMetrics, ps := setupProxy(t, fmt.Sprintf(`* -> backendTimeout("5ms") -> consistentHashKey("${request.header.ConsistentHashKey}") -> <%s, "%s", "%s", "%s">`,
 				algorithm, services[0].URL, services[1].URL, services[2].URL))
-			defer tp.close()
-			defer ps.Close()
 			failedReqs := sendGetRequests(t, ps)
+			mockMetrics.WithCounters(func(counters map[string]int64) {
+				assert.InDelta(t, float64(nRequests), float64(counters["passive-health-check.endpoints.dropped"]), 0.1*float64(nRequests)) // allow 10% error
```
AlexanderYastrebov: Shouldn't it be equal to `failedReqs`?
Also maybe move it after existing assert.InDelta

AlexanderYastrebov:
Looks good, please rebase, address minor comments and resolve conversations.

proxy/healthy_endpoints_test.go
MustafaSaber: > Shouldn't it be equal to `failedReqs`?

No because this metric will count number of requests drop across all endpoints, since we have 15000 request and one unhealthy endpoint then the counter should be equal 15000, however according to @RomanZavodskikh the code takes couple of requests before probability adjustment and dropping the endpoint hence it's less than 15000 and also the delta.

proxy/proxy.go
```diff
@@ -1193,9 +1193,7 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 		if err := p.do(loopCTX, loopSpan); err != nil {
 			// in case of error we have to copy the response in this recursion unwinding
 			ctx.response = loopCTX.response
-			if err != nil {
```
MustafaSaber, line 1196: not related to PR but decided to fix to remove linter warning, the if condition ensures we have error so the inner one is redundant

RomanZavodskikh: Okay, I see outer `err != nil` on the line 1193, I agree.

RomanZavodskikh:
:+1:

AlexanderYastrebov:
:+1:

PR #3036, opened 2024-04-22
Added PHC-related logs

RomanZavodskikh:

### DIFF
--- a/proxy/healthy_endpoints.go
+++ b/proxy/healthy_endpoints.go
@@ -11,7 +11,7 @@ type healthyEndpoints struct {
 	endpointRegistry *routing.EndpointRegistry
 }
 
-func (h *healthyEndpoints) filterHealthyEndpoints(endpoints []routing.LBEndpoint, rt *routing.Route) []routing.LBEndpoint {
+func (h *healthyEndpoints) filterHealthyEndpoints(ctx *context, endpoints []routing.LBEndpoint) []routing.LBEndpoint {
 	if h == nil {
 		return endpoints
 	}
@@ -20,8 +20,10 @@ func (h *healthyEndpoints) filterHealthyEndpoints(endpoints []routing.LBEndpoint
 
 	filtered := make([]routing.LBEndpoint, 0, len(endpoints))
 	for _, e := range endpoints {
-		if p < e.Metrics.HealthCheckDropProbability() {
-			/* drop */
+		dropProbability := e.Metrics.HealthCheckDropProbability()
+		if p < dropProbability {
+			ctx.Logger().Infof("Dropping endpoint %q due to passive health check: p=%0.2f, dropProbability=%0.2f",
+				e.Host, p, dropProbability)
 		} else {
 			filtered = append(filtered, e)
 		}
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -540,7 +540,7 @@ func (p *Proxy) selectEndpoint(ctx *context) *routing.LBEndpoint {
 	rt := ctx.route
 	endpoints := rt.LBEndpoints
 	endpoints = p.fadein.filterFadeIn(endpoints, rt)
-	endpoints = p.heathlyEndpoints.filterHealthyEndpoints(endpoints, rt)
+	endpoints = p.heathlyEndpoints.filterHealthyEndpoints(ctx, endpoints)
 
 	lbctx := &routing.LBContext{
 		Request:     ctx.request,
--- a/routing/endpointregistry.go
+++ b/routing/endpointregistry.go
@@ -5,6 +5,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	log "github.com/sirupsen/logrus"
+
 	"github.com/zalando/skipper/eskip"
 )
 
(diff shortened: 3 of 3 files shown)


proxy/healthy_endpoints.go
```diff
@@ -21,7 +21,7 @@ func (h *healthyEndpoints) filterHealthyEndpoints(endpoints []routing.LBEndpoint
 	filtered := make([]routing.LBEndpoint, 0, len(endpoints))
 	for _, e := range endpoints {
 		if p < e.Metrics.HealthCheckDropProbability() {
-			/* drop */
+			ctx.Logger().Infof("dropping endpoint %s due to passive health check", e.Host)
```
RomanZavodskikh, line 24: There must be both Host and Port.

proxy/healthy_endpoints.go
```diff
@@ -21,7 +21,7 @@ func (h *healthyEndpoints) filterHealthyEndpoints(endpoints []routing.LBEndpoint
 	filtered := make([]routing.LBEndpoint, 0, len(endpoints))
 	for _, e := range endpoints {
 		if p < e.Metrics.HealthCheckDropProbability() {
-			/* drop */
+			ctx.Logger().Infof("dropping endpoint %q due to passive health check", e.Host)
```
szuecs, line 24: maybe also log p and the probability?

routing/endpointregistry.go
```diff
@@ -166,6 +169,9 @@ func (r *EndpointRegistry) updateStats() {
 			requests := e.totalRequests[curSlot].Load()
 			if requests > r.minRequests {
 				failedRoundTripsRatio := float64(failed) / float64(requests)
+				if failedRoundTripsRatio > 0 {
+					r.logger.Infof("passive health check: marking %q as unhealthy due to failed round trips ratio: %f", key, failedRoundTripsRatio)
```
szuecs, line 173: %f -> %0.2f

proxy/healthy_endpoints.go
RomanZavodskikh: Yes, I think it won't make any harm.

proxy/healthy_endpoints.go
```diff
@@ -21,7 +21,8 @@ func (h *healthyEndpoints) filterHealthyEndpoints(endpoints []routing.LBEndpoint
 	filtered := make([]routing.LBEndpoint, 0, len(endpoints))
 	for _, e := range endpoints {
 		if p < e.Metrics.HealthCheckDropProbability() {
-			/* drop */
+			ctx.Logger().Infof("dropping endpoint %q due to passive health check: p=%0.2f, dropProbability=%0.2f",
```
MustafaSaber, line 24: maybe use normal logrus? No need to pass it I mean

RomanZavodskikh: I still would like to have `ctx.Logger()` here: after https://github.com/zalando/skipper/pull/2269/ it has useful feature, it adds traceId inside log entry. It may be very helpful to see the relation between trace and log entry during debugging.

proxy/proxy.go
```diff
@@ -791,8 +791,9 @@ func WithParams(p Params) *Proxy {
 		defaultHTTPStatus = p.DefaultHTTPStatus
 	}
 
+	logger := &logging.DefaultLog{}
```
AlexanderYastrebov, line 794: I am not sure if we should use this, maybe just use logrus logger.
I think we do not use it consistently.

proxy/healthy_endpoints.go
```diff
@@ -21,7 +21,8 @@ func (h *healthyEndpoints) filterHealthyEndpoints(endpoints []routing.LBEndpoint
...
+				e.Host, p, e.Metrics.HealthCheckDropProbability())
```
AlexanderYastrebov, line 25: To print `e.Metrics.HealthCheckDropProbability()` you need to capture its value as it can change between check and log.

AlexanderYastrebov:
My concern is that we can flood logs with this. Lets try but we need to keep an eye and may need to disable/move to debug level these logs.

RomanZavodskikh:
> My concern is that we can flood logs with this. Lets try but we need to keep an eye and may need to disable/move to debug level these logs.

We discussed this topic in the meeting:

+ the logs will be written only in case some endpoint will have non-zero requrest failure ratio, not all the time
+ if there will be too many logs, we will change the mode to the Debug

routing/endpointregistry.go
```diff
@@ -166,6 +168,9 @@ func (r *EndpointRegistry) updateStats() {
 			requests := e.totalRequests[curSlot].Load()
 			if requests > r.minRequests {
 				failedRoundTripsRatio := float64(failed) / float64(requests)
+				if failedRoundTripsRatio > 0.0 {
+					log.Infof("passive health check: marking %q as unhealthy due to failed round trips ratio: %0.2f", key, failedRoundTripsRatio)
```
AlexanderYastrebov, line 172: Log messages should start from capital, error messages from small letter.

RomanZavodskikh: Changed both log messages.

RomanZavodskikh:
:+1:

szuecs:
:+1:

PR #3038, opened 2024-04-23
Removed rtFailureProbability variable

RomanZavodskikh:
This variable is actually is not being used during sending requests to the test proxy, but left in the assert statement. The test were working before, because the result of multiplication is close enough to zero.


RomanZavodskikh:
:+1:

Issue #2104, opened 2022-10-14
Use OpenTelemetry instead of OpenTracing


AlexanderYastrebov:
OpenTracing is deprecated, see https://github.com/opentracing/specification/issues/163

## Migration plan 1

Initial idea implemented by https://github.com/zalando/skipper/pull/2578 was to refactor all the code to use OpenTelemetry tracer and provide an adapter that implements OpenTelemetry API but uses OpenTracing tracer to send tracing data.
Then in subsequent steps introduce first-class OpenTelemetry tracer and switch to it.

## Migration plan 2

Plan 1 has two downsides:
* refactors a lot of code and hence increases chances to introduce bugs
* does not use OpenTelemetry protocol to talk to telemetry backend, i.e. does not migrate to OpenTelemetry from external point of view.

I propose the following migration plan:

### 1. Enable OpenTelemetry SDK

This is a suggested path in [the official migration guide](https://opentelemetry.io/docs/migration/opentracing/#migration-overview). 
The idea is to make Skipper talk OpenTelemetry protocol to the telemetry backend (Lightstep in our case) and keep existing components intact and continue using OpenTracing API via [bridge component](https://pkg.go.dev/go.opentelemetry.io/otel/bridge/opentracing).
OpenTracing API [is deprecated](https://github.com/opentracing/specification/issues/163) which in other words means **it is stable** and is very unlikely to be changed:
> No more new pull requests or feature requests will be accepted, only security fixes (which are unlikely given the nature of the API libraries).

This will achieve the goal of using OpenTelemetry from external point of view without introducing breaking changes for users.

### 2. Use OpenTelemetry API in new components

For all new components use OpenTelemetry API. These new components will not work for users that use OpenTracing but this will be an incentive to migrate.

### 3. Refactor existing components one-by-one to use OpenTelemetry API

This should be done case-by-case and depends on the components.
We may deprecated [OpenTracing filters](https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#open-tracing) and add complementary OpenTelemtry filters that use OpenTelemetry API and proper naming (e.g. for `tracingTag` add `tracingAttribute`, etc)

We may refactor HTTP and Redis clients to use OpenTelemetry API or support both OpenTracing and OpenTelemetry tracers.

The most complex is the Proxy component which we may tackle the last after learning from migration of other components.

fpiche:
[...]

PR #3039, opened 2024-04-24
Specify tracing span kind on creation

AlexanderYastrebov:
OpenTelemetry-OpenTracing bridge span kind can not be changed after creation, see https://github.com/open-telemetry/opentelemetry-go/issues/3953

The workaround is to specify span kind on creation which works for both Open Tracing and Open Telemetry bridge spans.

For #2104

### DIFF
--- a/filters/shedder/admission.go
+++ b/filters/shedder/admission.go
@@ -456,7 +456,6 @@ func (ac *admissionControl) startSpan(ctx context.Context) (span opentracing.Spa
 	if parent != nil {
 		span = ac.tracer.StartSpan(admissionControlSpanName, opentracing.ChildOf(parent.Context()))
 		ext.Component.Set(span, "skipper")
-		ext.SpanKind.Set(span, "shedder")
 		span.SetTag("mode", ac.mode.String())
 	}
 	return
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -958,14 +958,19 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 	if !ok {
 		spanName = "proxy"
 	}
-	ctx.proxySpan = tracing.CreateSpan(spanName, req.Context(), p.tracing.tracer)
+
+	proxySpanOpts := []ot.StartSpanOption{ot.Tags{
+		SpanKindTag:       SpanKindClient,
+		SkipperRouteIDTag: ctx.route.Id,
+	}}
+	if parentSpan := ot.SpanFromContext(req.Context()); parentSpan != nil {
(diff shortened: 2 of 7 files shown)


filters/shedder/admission.go
```diff
@@ -456,7 +456,6 @@ func (ac *admissionControl) startSpan(ctx context.Context) (span opentracing.Spa
 	if parent != nil {
 		span = ac.tracer.StartSpan(admissionControlSpanName, opentracing.ChildOf(parent.Context()))
 		ext.Component.Set(span, "skipper")
-		ext.SpanKind.Set(span, "shedder")
```
AlexanderYastrebov, line 459: I am not sure if we really need non-standard kind here

proxy/proxy.go
```diff
@@ -958,14 +958,20 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
...
 	}
-	ctx.proxySpan = tracing.CreateSpan(spanName, req.Context(), p.tracing.tracer)
+
+	proxySpanOpts := []ot.StartSpanOption{ot.Tags{
+		SpanKindTag:       SpanKindClient,
+		SkipperRouteIDTag: ctx.route.Id,
+	}}
+	if parentSpan := ot.SpanFromContext(req.Context()); parentSpan != nil {
+		proxySpanOpts = append(proxySpanOpts, ot.ChildOf(parentSpan.Context()))
+
```
szuecs, line 968: whitespace

filters/shedder/admission.go
szuecs: Can you check if we have streams that might select it and propose a change?

proxy/proxy.go
AlexanderYastrebov: Fixed

filters/shedder/admission.go
AlexanderYastrebov: Found one that also uses `operation IN ("admission_control")` which should be enough.

AlexanderYastrebov: Reached out to the owner and this was fixed.

szuecs:
let's land it, no?

RomanZavodskikh:
:+1:

AlexanderYastrebov:
Rebased and fixed span creation added by #3034

filters/shedder/admission.go
szuecs: Can you add this to the release notes, too?
I don't know who will hit by that but that we do tell about.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #3040, opened 2024-04-25
phc: add metrics for number of unhealthy endpoints

MustafaSaber:
related to https://github.com/zalando/skipper/issues/2346

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -539,8 +539,10 @@ func setRequestURLForDynamicBackend(u *url.URL, stateBag map[string]interface{})
 func (p *Proxy) selectEndpoint(ctx *context) *routing.LBEndpoint {
 	rt := ctx.route
 	endpoints := rt.LBEndpoints
+	beforefiltering := len(endpoints)
 	endpoints = p.fadein.filterFadeIn(endpoints, rt)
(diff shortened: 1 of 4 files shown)


proxy/proxy.go
```diff
@@ -539,8 +539,10 @@ func setRequestURLForDynamicBackend(u *url.URL, stateBag map[string]interface{})
 func (p *Proxy) selectEndpoint(ctx *context) *routing.LBEndpoint {
 	rt := ctx.route
 	endpoints := rt.LBEndpoints
+	beforefiltering := len(endpoints)
 	endpoints = p.fadein.filterFadeIn(endpoints, rt)
 	endpoints = p.heathlyEndpoints.filterHealthyEndpoints(ctx, endpoints, p.metrics)
+	p.metrics.UpdateGauge("passive-health-check.endpoints.dropped", float64(beforefiltering-len(endpoints)))
```
MustafaSaber, line 545: After some discussions with @RomanZavodskikh, this gauge will be overwritten by different services, for example svc A has 2 unhealthy endpoints and svc B has 1 both will write on the same metric.

What we can do is append routeId or Name+Namespace combination. we aren't sure that's a good thing memory wise, wdyt @AlexanderYastrebov @szuecs?

PR #3043, opened 2024-04-26
proxy: use proxy tracing setTag helper

AlexanderYastrebov:
Use setTag helper to enable tag exclusion

This is follow up fix for #3039


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #3044, opened 2024-04-26
filters: deprecate OpenTracing in FilterContext

AlexanderYastrebov:
Deprecate OpenTracing FilterContext getters and remove their usage from filters.

For #2104

### DIFF
--- a/filters/filters.go
+++ b/filters/filters.go
@@ -108,9 +108,16 @@ type FilterContext interface {
 	Metrics() Metrics
 
 	// Allow filters to add Tags, Baggage to the trace or set the ComponentName.
+	//
+	// Deprecated: OpenTracing is deprecated, see https://github.com/zalando/skipper/issues/2104.
+	// Use opentracing.SpanFromContext(ctx.Request().Context()).Tracer() to get the proxy span Tracer.
 	Tracer() opentracing.Tracer
 
 	// Allow filters to create their own spans
+	//
+	// Deprecated: OpenTracing is deprecated, see https://github.com/zalando/skipper/issues/2104.
+	// Filter spans should be children of the proxy span,
+	// use opentracing.SpanFromContext(ctx.Request().Context()) to get the proxy span.
 	ParentSpan() opentracing.Span
 
 	// Returns a clone of the FilterContext including a brand new request object.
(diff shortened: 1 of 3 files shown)


filters/openpolicyagent/openpolicyagent.go
```diff
@@ -599,7 +599,7 @@ func (opa *OpenPolicyAgentInstance) startSpanFromContextWithTracer(tr opentracin
 }
 
 func (opa *OpenPolicyAgentInstance) StartSpanFromFilterContext(fc filters.FilterContext) (opentracing.Span, context.Context) {
-	return opa.startSpanFromContextWithTracer(fc.Tracer(), fc.ParentSpan(), fc.Request().Context())
+	return opa.StartSpanFromContext(fc.Request().Context())
```
AlexanderYastrebov, line 602: @mjungsbluth BTW I am not sure why OPA needs a whole set of opentracing helpers.

I think it should get tracer from its options and for synchronous filter requests it should get proxy span (and its tracer if needed) via `opentracing.SpanFromContext(ctx.Request().Context())`

szuecs:
:+1:

filters/filters.go
```diff
@@ -108,9 +108,16 @@ type FilterContext interface {
...
+	//
+	// Deprecated: OpenTracing is deprecated, see https://github.com/zalando/skipper/issues/2104.
+	// Use opentracing.SpanFromContext(ctx.Request().Context()).Tracer() to get the proxy span Tracer.
 	Tracer() opentracing.Tracer
 
 	// Allow filters to create their own spans
+	//
+	// Deprecated: OpenTracing is deprecated, see https://github.com/zalando/skipper/issues/2104.
+	// Filter spans should be children of the proxy span,
+	// use opentracing.SpanFromContext(ctx.Request().Context()) to get the proxy span.
```
AlexanderYastrebov, line 120: This is not a proxy span but a "filters" span added by https://github.com/zalando/skipper/pull/890 if enabled https://github.com/zalando/skipper/blob/f6bf0336bafcaab3a7e71447d6925fdaf54d5c2a/proxy/tracing.go#L111-L119

Maybe we can change it to return ingress span.
Anyway it is not used by anything, e.g. `tracingTag` uses ingress span (which is `opentracing.SpanFromContext(ctx.Request().Context())`)

filters/openpolicyagent/openpolicyagent.go
mjungsbluth: @AlexanderYastrebov It should be fine the way you outlined. It was probably overly defensive to include the tracer for cases where the context does not already contain a span (which actually don't provide much value for the synchronous cases).

filters/filters.go
AlexanderYastrebov: See https://github.com/zalando/skipper/pull/3052

PR #3046, opened 2024-04-26
Change condition of PHC to reduce amount of produced logs 

RomanZavodskikh:
Logging every request produces too many log lines.
However, we can still keep the logs of stats updater https://github.com/zalando/skipper/blob/v0.21.65/routing/endpointregistry.go#L172 because their amount does not depend on requests count.

### DIFF
--- a/proxy/healthy_endpoints.go
+++ b/proxy/healthy_endpoints.go
@@ -22,7 +22,7 @@ func (h *healthyEndpoints) filterHealthyEndpoints(ctx *context, endpoints []rout
 	filtered := make([]routing.LBEndpoint, 0, len(endpoints))
 	for _, e := range endpoints {
 		dropProbability := e.Metrics.HealthCheckDropProbability()
-		if p < dropProbability {
+		if dropProbability > 0.05 && p < dropProbability {
 			ctx.Logger().Infof("Dropping endpoint %q due to passive health check: p=%0.2f, dropProbability=%0.2f",
 				e.Host, p, dropProbability)
 			metrics.IncCounter("passive-health-check.endpoints.dropped")


RomanZavodskikh:
:+1:

proxy/healthy_endpoints.go
```diff
@@ -23,8 +23,10 @@ func (h *healthyEndpoints) filterHealthyEndpoints(ctx *context, endpoints []rout
 	for _, e := range endpoints {
 		dropProbability := e.Metrics.HealthCheckDropProbability()
 		if p < dropProbability {
-			ctx.Logger().Infof("Dropping endpoint %q due to passive health check: p=%0.2f, dropProbability=%0.2f",
-				e.Host, p, dropProbability)
+			if dropProbability > 0.05 {
```
szuecs, line 26: Maybe it makes more sense to "enhance" the condition above!?
```
if dropProbability > 0.05 && p < dropProbability {
..
} else {
   filtered = append(filtered, e)
...
```

szuecs:
:+1:

RomanZavodskikh:
:+1:

proxy/healthy_endpoints.go
```diff
@@ -22,7 +22,7 @@ func (h *healthyEndpoints) filterHealthyEndpoints(ctx *context, endpoints []rout
 	filtered := make([]routing.LBEndpoint, 0, len(endpoints))
 	for _, e := range endpoints {
 		dropProbability := e.Metrics.HealthCheckDropProbability()
-		if p < dropProbability {
+		if dropProbability > 0.05 && p < dropProbability {
```
AlexanderYastrebov, line 25: Hm, this does not look right as it not only influences logging but a logic of dropping the endpoint.

We can enhance logging to log once a second/minute and also keep drop counter in the e.Metrics for logging but maybe we should just remove logs and rely on the `passive-health-check.endpoints.dropped`.

AlexanderYastrebov: And when probability goes up this will flood logs anyway.

AlexanderYastrebov: The min probability cutoff (if we want it) should be implemented in updateStats()

szuecs: As long as we have no metrics that are good enough we let the logs.
The idea was that we cut the min. There was a lot of logs that were way too close to 0 to make sense. If you have all endpoints having a low error rate you just cycle through all of them, likely something we also want to address.

PR #3052, opened 2024-04-29
filters: update godoc for deprecated FilterContext OpenTracing fields

AlexanderYastrebov:
The ParentSpan is not a proxy span but either ingress span of "filters" span when enabled.

Follow up on #3044


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #3053, opened 2024-04-29
Add -min-drop-probability cmdline option for the PHC

RomanZavodskikh:


szuecs:
:+1:

RomanZavodskikh:
:+1:

routing/endpointregistry.go
```diff
@@ -164,15 +166,19 @@ func (r *EndpointRegistry) updateStats() {
...
 					log.Infof("Passive health check: marking %q as unhealthy due to failed round trips ratio: %0.2f", key, failedRoundTripsRatio)
+					e.healthCheckDropProbability.Store(min(failedRoundTripsRatio, r.maxHealthCheckDropProbability))
+					newProbabilityStored = true
 				}
-				e.healthCheckDropProbability.Store(min(failedRoundTripsRatio, r.maxHealthCheckDropProbability))
-			} else {
+			}
+
+			if !newProbabilityStored {
 				e.healthCheckDropProbability.Store(0.0)
```
AlexanderYastrebov, line 182: Shouldn't we store minHealthCheckDropProbability instead of zero here?

AlexanderYastrebov: Ah, ok, we should change it when it is above min. But why do we changed else branch instead of just changing `if failedRoundTripsRatio > 0.0`?

RomanZavodskikh: I think this way would be the most convenient. This way we will get rid of else branch at all.

```
newDropProbability := 0.0 // default value
if manyRequests {
  if failedRatioIsBig {
    newDropProbability = failedRatio //non-default value
  }
}
e.healthCheckDropProbability.Store(newDropProbability)
```

routing/endpointregistry.go
```diff
@@ -164,17 +166,17 @@ func (r *EndpointRegistry) updateStats() {
...
 
+			newDropProbability := 0.0
 			failed := e.totalFailedRoundTrips[curSlot].Load()
 			requests := e.totalRequests[curSlot].Load()
 			if requests > r.minRequests {
 				failedRoundTripsRatio := float64(failed) / float64(requests)
-				if failedRoundTripsRatio > 0.0 {
+				if failedRoundTripsRatio > r.minHealthCheckDropProbability {
 					log.Infof("Passive health check: marking %q as unhealthy due to failed round trips ratio: %0.2f", key, failedRoundTripsRatio)
+					newDropProbability = failedRoundTripsRatio
```
AlexanderYastrebov, line 176: Should we keep `min(failedRoundTripsRatio, r.maxHealthCheckDropProbability)` here?

docs/operation/operation.md
```diff
@@ -918,6 +918,7 @@ However, Skipper will run without this feature, if no `-passive-health-check` is
 The parameters of `-passive-health-check` option are:
 + `period=<duration>` - the duration of stats reset period
 + `min-requests=<int>` - the minimum number of requests per `period` per backend endpoint required to activate PHC for this endpoint
++ `min-drop-probabilty=<float more than/equal to 0 and less than/equal to 1>` - the minimum possible probability of unhealthy endpoint being not considered while choosing the endpoint for the given request. The same value is in fact used as minimal failed requests ratio for PHC to be enabled for this endpoint
```
MustafaSaber, line 921: maybe simple `0.0 <= n <= 1.0` would be enough and would reduce text, same for other variables

MustafaSaber: I mean instead of `float more than/equal to 0 and less than/equal to 1`

RomanZavodskikh: `n` is assosiated with natural number, which is not the case here, I will pick up `p` instead.

RomanZavodskikh:
:+1:

docs/operation/operation.md
MustafaSaber: anyway I meant a variable name :)

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #3055, opened 2024-04-30
filters/auth: extend jwtMetrics opt-out options

AlexanderYastrebov:
Allow opt-out via state bag key.
The usecase is to opt-out metrics collection for routes that use `oauthTokeninfo*` and `oauthGrant` filters.


szuecs:
:+1:

AlexanderYastrebov:
:+1:

Issue #2911, opened 2024-02-07
AWS sigv4 auth filter


szuecs:
Sometimes you want to proxy to an aws service and aws uses sigv4 to do authnz. 
It would be great to be able to sign with sigv4 the request with https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws/signer/v4#HTTPSigner to be able to call AWS services from the proxy itself.

Maybe leverage a similar kind of roundtripper which depends also on aws sdk https://github.com/prometheus/common/blob/main/sigv4/sigv4.go

Anurag252:
Hey @szuecs 
I can take a look at this one. My thoughts around this 
- Create a new request filter named `AWSSignV4` 
- add a dependency to aws-sdk-go-v2 and call [Sign](https://github.com/aws/aws-sdk-go-v2/blob/v1.25.0/aws/signer/v4/v4.go#L280). 
- Allow createfilter to accept `credentials`, take `service` and `region` from eskip file, and keep `signtime` as now [documentation](https://github.com/aws/aws-sdk-go-v2/blob/v1.25.0/aws/signer/v4/v4.go#L256). 

I believe we need to consider that 
- An instance of skipper would be able to sign requests for a single AWS account , but may target different services as credentials are being passed to create filter. Maybe there are other ways to generalize the credentials.

- `SignHTTP` accepts `payloadHash` which is essentially hex encoded SHA-256 hash of the request [payload](https://github.com/aws/aws-sdk-go-v2/blob/v1.25.0/aws/signer/v4/v4.go#L261C27-L261C74). This would mean request body would need to be read completely in the filter and then a new ReadCloser would need to be assigned to request so that it can be propagated further. This could increase memory usage of skipper significantly. Although as docs says there are a few AWS services that do not need body to be signed. 

- Since request is signed here, it makes sense if this is the last filter before proxy as any other filter changing headers, query params or body may invalidate the signing.

szuecs:
@Anurag252 I am not sure if I want to have the AWS sdk as dependency, rather not. I think it seems to be open enough to build this without AWS SDK, but I am not sure.

Anurag252:
> @Anurag252 I am not sure if I want to have the AWS sdk as dependency, rather not. I think it seems to be open enough to build this without AWS SDK, but I am not sure.

[...]

PR #3056, opened 2024-05-01
feature: adding AWS sigv4 auth filter

Anurag252:
### Work in progress

- This filter reads whole body in memory to generate signature. Special care needs to be taken in terms of memory when operating skipper taking into account concurrent requests
- This filter resets `read()` and `close()` on body. Hence any custom implementations of ReadCloser by previous filters will be overwritten and will have no effect. 

https://github.com/zalando/skipper/issues/2911

implementation of https://docs.aws.amazon.com/IAM/latest/UserGuide/create-signed-request.html .


Anurag252:
@szuecs Moving conversation to draft PR. On second look, implementation is not simply what is described at above link
- There is a custom list of allowed and ignored headers [see](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/signer/internal/v4/headers.go#L3C1-L13C2)
- The urlEncode is not a standard url encode and has special AWS style to it see [this](https://github.com/aws/aws-sdk-go-v2/blob/v1.25.0/aws/signer/v4/v4.go#L196) and [this](https://github.com/aws/smithy-go/blob/v0.4.0/httpbinding/path_replace.go#L82)

these are just the two that I came across till now. Since these are not documented at above link and only found in code, I would argue that not using SDK is error prone and very difficult to maintain . What are your thoughts about including SDK ?

szuecs:
Thanks for the heads up!

I think that aws won't break it because it will be a pain for their users to update aws sdk, so they will have a migration path and do a lot of work to omit this. Maybe check the git blame to see the age of the code to judge about it.

Anurag252:
This PR is messed up beyond repair in an attempt to add sign off message. Closing this and will raise a new one

PR #3058, opened 2024-05-02
phc: test multiple unhealthy endpoints

MustafaSaber:
related to https://github.com/zalando/skipper/pull/3035

With multiple unhealthy endpoints test case fails

### DIFF
--- a/proxy/healthy_endpoints_test.go
+++ b/proxy/healthy_endpoints_test.go
@@ -68,15 +68,28 @@ func setupProxy(t *testing.T, doc string) (*metricstest.MockMetrics, *httptest.S
 	return m, ps
 }
 
-func TestPHCWithoutRequests(t *testing.T) {
+func setupServices(t *testing.T, healthy, unhealthy int) []*httptest.Server {
 	services := []*httptest.Server{}
-	for i := 0; i < 3; i++ {
+	for i := 0; i < healthy; i++ {
 		service := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			w.WriteHeader(http.StatusOK)
 		}))
 		services = append(services, service)
-		defer service.Close()
+		t.Cleanup(service.Close)
 	}
+	for i := 0; i < unhealthy; i++ {
+		service := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			time.Sleep(100 * time.Millisecond)
+			w.WriteHeader(http.StatusOK)
+		}))
+		services = append(services, service)
+		t.Cleanup(service.Close)
+	}
+	return services
+}
+
+func TestPHCWithoutRequests(t *testing.T) {
+	services := setupServices(t, 3, 0)
 
 	for _, algorithm := range []string{"random", "consistentHash", "roundRobin", "powerOfRandomNChoices"} {
 		t.Run(algorithm, func(t *testing.T) {
@@ -104,10 +117,7 @@ func TestPHCWithoutRequests(t *testing.T) {
 }
 
 func TestPHCForSingleHealthyEndpoint(t *testing.T) {
-	service := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.WriteHeader(http.StatusOK)
-	}))
-	defer service.Close()
+	service := setupServices(t, 1, 0)[0]
 	endpointRegistry := defaultEndpointRegistry()
 
 	doc := fmt.Sprintf(`* -> "%s"`, service.URL)
@@ -128,14 +138,7 @@ func TestPHCForSingleHealthyEndpoint(t *testing.T) {
 }
 
 func TestPHCForMultipleHealthyEndpoints(t *testing.T) {
-	services := []*httptest.Server{}
-	for i := 0; i < 3; i++ {
-		service := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			w.WriteHeader(http.StatusOK)
-		}))
-		services = append(services, service)
-		defer service.Close()
-	}
+	services := setupServices(t, 3, 0)
 
 	for _, algorithm := range []string{"random", "consistentHash", "roundRobin", "powerOfRandomNChoices"} {
 		t.Run(algorithm, func(t *testing.T) {
@@ -155,20 +158,7 @@ func TestPHCForMultipleHealthyEndpoints(t *testing.T) {
 }
 
 func TestPHCForMultipleHealthyAndOneUnhealthyEndpoints(t *testing.T) {
-	services := []*httptest.Server{}
-	for i := 0; i < 3; i++ {
-		serviceNum := i
-		service := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			if serviceNum == 0 {
-				// emulating unhealthy endpoint
-				time.Sleep(100 * time.Millisecond)
-			}
-			w.WriteHeader(http.StatusOK)
-		}))
-		services = append(services, service)
-		defer service.Close()
-	}
-
+	services := setupServices(t, 2, 1)
 	for _, algorithm := range []string{"random", "consistentHash", "roundRobin", "powerOfRandomNChoices"} {
 		t.Run(algorithm, func(t *testing.T) {
 			mockMetrics, ps := setupProxy(t, fmt.Sprintf(`* -> backendTimeout("5ms") -> consistentHashKey("${request.header.ConsistentHashKey}") -> <%s, "%s", "%s", "%s">`,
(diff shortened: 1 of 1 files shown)


proxy/healthy_endpoints_test.go
```diff
@@ -128,62 +137,80 @@ func TestPHCForSingleHealthyEndpoint(t *testing.T) {
...
-				assert.InDelta(t, float64(nRequests), float64(counters["passive-health-check.endpoints.dropped"]), 0.3*float64(nRequests)) // allow 30% error
+				assert.InDelta(t, float64(nRequests)*2.0, float64(counters["passive-health-check.endpoints.dropped"]), 0.3*float64(nRequests)) // allow 30% error
 			})
 		})
 	}
 
 	t.Run("consistent hash with balance factor", func(t *testing.T) {
-		_, ps := setupProxy(t, fmt.Sprintf(`* -> backendTimeout("5ms") -> consistentHashKey("${request.header.ConsistentHashKey}") -> consistentHashBalanceFactor(1.25) -> <consistentHash, "%s", "%s", "%s">`,
-			services[0].URL, services[1].URL, services[2].URL))
+		endpointRegistry := routing.NewEndpointRegistry(routing.RegistryOptions{
```
RomanZavodskikh, line 205: minor for now: I still would like to see this test working with default endpointRegistry.
To achieve this, we will need to speed up this test somehow, probably, we will need to send GET requests concurrently, not sequentially.

proxy/healthy_endpoints_test.go
```diff
@@ -128,62 +137,80 @@ func TestPHCForSingleHealthyEndpoint(t *testing.T) {
...
+	})
+}
+
+func TestPHCForMultipleHealthyAndMultipleUnhealthyEndpoints(t *testing.T) {
+	services := setupServices(t, 2, 2)
 	for _, algorithm := range []string{"random", "consistentHash", "roundRobin", "powerOfRandomNChoices"} {
 		t.Run(algorithm, func(t *testing.T) {
-			mockMetrics, ps := setupProxy(t, fmt.Sprintf(`* -> backendTimeout("5ms") -> consistentHashKey("${request.header.ConsistentHashKey}") -> <%s, "%s", "%s", "%s">`,
-				algorithm, services[0].URL, services[1].URL, services[2].URL))
+
```
RomanZavodskikh, line 185: minor: could you please remove this empty?

RomanZavodskikh:
:+1:

MustafaSaber:
we were able to fix with custom options for endpointregistry but still with the minimum value race still fails

proxy/healthy_endpoints_test.go
```diff
@@ -188,3 +181,41 @@ func TestPHCForMultipleHealthyAndOneUnhealthyEndpoints(t *testing.T) {
...
+				assert.InDelta(t, float64(nRequests)*2.0, float64(counters["passive-health-check.endpoints.dropped"]), 0.3*float64(nRequests)) // allow 30% error
+			})
+		})
+	}
+
+	t.Run("consistent hash with balance factor", func(t *testing.T) {
+		endpointRegistry := routing.NewEndpointRegistry(routing.RegistryOptions{
+			PassiveHealthCheckEnabled:     true,
+			StatsResetPeriod:              period,
+			MinRequests:                   1, // with 3 test case fails
```
szuecs, line 212: I guess `sendGetRequests()` is too slow or the backend is too slow.
Maybe use vegeta as library to create a stable amount of requests like 10rps.

szuecs: So it does not do concurrent requests and it reads the response and checks for error and status code.
In a CI environment that's likely too slow for having MinRequests increased to 3. Vegeta would be the better option in this case.

MustafaSaber: yes that was the problem and we were looking how to make send them concurrently Vegeta seems like a good option

proxy/healthy_endpoints_test.go
```diff
@@ -68,15 +71,28 @@ func setupProxy(t *testing.T, doc string) (*metricstest.MockMetrics, *httptest.S
...
 	}
+	for i := 0; i < unhealthy; i++ {
+		service := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			time.Sleep(100 * time.Millisecond)
+			w.WriteHeader(http.StatusOK)
+		}))
+		services = append(services, service)
+		t.Cleanup(service.Close)
+	}
+	return services
```
AlexanderYastrebov, line 91: Since you are doing automatic cleanup, it might be handy to return slice of service URLs if its the only thing that is used by the caller.

proxy/healthy_endpoints_test.go
```diff
@@ -128,63 +151,134 @@ func TestPHCForSingleHealthyEndpoint(t *testing.T) {
...
+	services := setupServices(t, 3, 0)
 
 	for _, algorithm := range []string{"random", "consistentHash", "roundRobin", "powerOfRandomNChoices"} {
 		t.Run(algorithm, func(t *testing.T) {
 			_, ps := setupProxy(t, fmt.Sprintf(`* -> consistentHashKey("${request.header.ConsistentHashKey}") -> <%s, "%s", "%s", "%s">`,
-				algorithm, services[0].URL, services[1].URL, services[2].URL))
-			failedReqs := sendGetRequests(t, ps)
-			assert.Equal(t, 0, failedReqs)
+				algorithm, services[0], services[1], services[2]))
+			va := fireVegeta(t, ps, 3000, 1*time.Second, 5*time.Second)
```
szuecs, line 160: 3000/s is likely too much to achieve in CI environments. Please test it and see what you can achieve.

proxy/healthy_endpoints_test.go
```diff
@@ -49,8 +52,18 @@ func sendGetRequests(t *testing.T, ps *httptest.Server) (failed int) {
 	return
 }
 
+func fireVegeta(t *testing.T, ps *httptest.Server, ferq int, per time.Duration, timeout time.Duration) *zhttptest.VegetaAttacker {
```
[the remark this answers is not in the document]
MustafaSaber: yes, will fix

proxy/healthy_endpoints_test.go
```diff
@@ -128,63 +151,134 @@ func TestPHCForSingleHealthyEndpoint(t *testing.T) {
...
 		t.Run(algorithm, func(t *testing.T) {
 			_, ps := setupProxy(t, fmt.Sprintf(`* -> consistentHashKey("${request.header.ConsistentHashKey}") -> <%s, "%s", "%s", "%s">`,
-				algorithm, services[0].URL, services[1].URL, services[2].URL))
-			failedReqs := sendGetRequests(t, ps)
-			assert.Equal(t, 0, failedReqs)
+				algorithm, services[0], services[1], services[2]))
+			va := fireVegeta(t, ps, 3000, 1*time.Second, 5*time.Second)
+			count200, ok := va.CountStatus(200)
+			assert.True(t, ok)
+			assert.Equal(t, count200, int(va.TotalRequests()))
```
szuecs, line 163: you should also check that the amount is not 0 or maybe >=N for an N that you expect

proxy/healthy_endpoints_test.go
```diff
@@ -128,63 +151,135 @@ func TestPHCForSingleHealthyEndpoint(t *testing.T) {
...
+			assert.Equal(t, float64(count504), failedRequests)
+			assert.InDelta(t, 0, float64(failedRequests), 0.1*float64(va.TotalRequests()))
 			mockMetrics.WithCounters(func(counters map[string]int64) {
-				assert.InDelta(t, float64(nRequests), float64(counters["passive-health-check.endpoints.dropped"]), 0.3*float64(nRequests)) // allow 30% error
+				assert.InDelta(t, float64(va.TotalRequests()), float64(counters["passive-health-check.endpoints.dropped"]), 0.6*float64(va.TotalRequests())) // allow 60% error
 			})
 		})
 	}
 
+	t.Run("consistentHash", func(t *testing.T) {
```
MustafaSaber, line 204: ConsistentHash was separated because it's a little bit tricky to set headers for different set of requests with Vegeta, we use the hash to map the request to endpoint.

MustafaSaber:
👍

proxy/healthy_endpoints_test.go
```diff
@@ -128,63 +151,135 @@ func TestPHCForSingleHealthyEndpoint(t *testing.T) {
...
+			assert.True(t, ok)
+
+			failedRequests := math.Abs(float64(va.TotalRequests())) - float64(count200)
+			t.Logf("total requests: %d, count200: %d, count504: %d, failedRequests: %f", va.TotalRequests(), count200, count504, failedRequests)
+
+			assert.Equal(t, float64(count504), failedRequests)
+			assert.InDelta(t, 0, float64(failedRequests), 0.1*float64(va.TotalRequests()))
 			mockMetrics.WithCounters(func(counters map[string]int64) {
-				assert.InDelta(t, float64(nRequests), float64(counters["passive-health-check.endpoints.dropped"]), 0.3*float64(nRequests)) // allow 30% error
+				assert.InDelta(t, float64(va.TotalRequests()), float64(counters["passive-health-check.endpoints.dropped"]), 0.6*float64(va.TotalRequests())) // allow 60% error
```
RomanZavodskikh, line 199: minor: can we change delta to 30% here?

RomanZavodskikh:
:+1:

MustafaSaber:
👍

PR #3065, opened 2024-05-06
eskip: remove flowid dependency

AlexanderYastrebov:
eskip package is generally useful without proxy or filters and therefore ideally should have minimal number of dependencies.

This change deprecates GenerateIfNeeded which is only used by etcd dataclient and reimplements it to remove dependency on github.com/zalando/skipper/filters/flowid

See 35ba6cc76dee89a57cc7869e79824bd9fc56e750

### DIFF
--- a/eskip/eskip.go
+++ b/eskip/eskip.go
@@ -5,13 +5,13 @@ package eskip
 import (
 	"errors"
 	"fmt"
+	"math/rand"
 	"net/url"
 	"regexp"
 	"strings"
 	"sync"
 
 	log "github.com/sirupsen/logrus"
-	"github.com/zalando/skipper/filters/flowid"
 )
 
 const duplicateHeaderPredicateErrorFmt = "duplicate header predicate: %s"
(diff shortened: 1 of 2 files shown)


eskip/eskip.go
```diff
@@ -712,29 +712,28 @@ func ParsePredicates(p string) ([]*Predicate, error) {
...
+const (
+	randomIdLength = 16
+	// does not contain underscore to produce compatible output with previously used flow id generator
+	alphabet = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+)
 
 // generate weak random id for a route if
 // it doesn't have one.
+//
+// Deprecated: do not use, generate valid route id that matches [a-zA-Z_] yourself.
```
szuecs, line 724: deprecate in the sense of stable?
If so, why do we need to deprecate it?
If we do not plan to delete it, then it's just fine and we do not need to apply the ignore SA1019

AlexanderYastrebov: To hint lib users to avoid it.
I'd also delete it in the future as it has strange API contract.

Or do you propose that we keep it or replace with something else?

szuecs: I just wanted to ask what the plan is, because the PR description nor commit does show what the plan is.
And you talked about deprecated is good because it means it is stable in the last days.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #3067, opened 2024-05-07
Make min-drop-probability of PHC optional

RomanZavodskikh:
The default value (0.0) makes some sense and is equivalent to how PHC was working before.


proxy/proxy.go
```diff
@@ -169,10 +169,11 @@ func InitPassiveHealthChecker(o map[string]string) (bool, *PassiveHealthCheck, e
 	}
 
 	result := &PassiveHealthCheck{}
-	keysInitialized := make(map[string]struct{})
+	requiredKeysInitialized := make(map[string]struct{})
```
AlexanderYastrebov, line 172: Maybe
```diff
diff --git a/proxy/proxy.go b/proxy/proxy.go
index 2de0161d..7c8b2e4d 100644
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -16,6 +16,7 @@ import (
        "net/url"
        "os"
        "runtime"
+       "slices"
        "strconv"
        "strings"
        "time"
@@ -169,9 +170,10 @@ func InitPassiveHealthChecker(o map[string]string) (bool, *PassiveHealthCheck, e
        }
 
        result := &PassiveHealthCheck{}
-       keysInitialized := make(map[string]struct{})
+       requiredOptions := []string{"max-drop-probability"}
 
        for key, value := range o {
+               requiredOptions = slices.DeleteFunc(requiredOptions, func(s string) bool { return s == key })
                switch key {
                case "period":
                        period, err := time.ParseDuration(value)
@@ -212,13 +214,12 @@ func InitPassiveHealthChecker(o map[string]string) (bool, *PassiveHealthCheck, e
                default:
                        return false, nil, fmt.Errorf("passive health check: invalid parameter: key=%s,value=%s", key, value)
                }
-
-               keysInitialized[key] = struct{}{}
        }
 
-       if len(keysInitialized) != 4 {
-               return false, nil, fmt.Errorf("passive health check: missing required parameters")
+       if len(requiredOptions) != 0 {
+               return false, nil, fmt.Errorf("passive health check: missing required parameters %v", requiredOptions)
        }
+
        if result.MinDropProbability >= result.MaxDropProbability {
                return false, nil, fmt.Errorf("passive health check: minDropProbability should be less than maxDropProbability")
        }
```

proxy/proxy.go
```diff
@@ -209,14 +203,24 @@ func InitPassiveHealthChecker(o map[string]string) (bool, *PassiveHealthCheck, e
...
+			result.MinDropProbability = minDropProbability
 		default:
 			return false, nil, fmt.Errorf("passive health check: invalid parameter: key=%s,value=%s", key, value)
 		}
-
-		keysInitialized[key] = struct{}{}
 	}
 
-	if len(keysInitialized) != 4 {
+	if len(requiredKeysInitialized) != 3 {
```
AlexanderYastrebov, line 223: If we only use length then integer counter would be enough but I think we can also show which required are missing in the error message, see snippet.

proxy/proxy.go
RomanZavodskikh: I like the code snippet. What about the exact name: I still would like some word after `required` to be specified, so I'll use `requiredParams`.

proxy/proxy.go
```diff
@@ -209,14 +206,23 @@ func InitPassiveHealthChecker(o map[string]string) (bool, *PassiveHealthCheck, e
...
+			result.MinDropProbability = minDropProbability
 		default:
 			return false, nil, fmt.Errorf("passive health check: invalid parameter: key=%s,value=%s", key, value)
 		}
-
-		keysInitialized[key] = struct{}{}
 	}
 
-	if len(keysInitialized) != 4 {
+	if len(requiredParams) != 0 {
```
AlexanderYastrebov, line 225: Lets add `maps.Keys(requiredParams)` to the error message

AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

proxy/proxy.go
```diff
@@ -21,6 +21,7 @@ import (
 	"time"
 	"unicode/utf8"
 
+	"golang.org/x/exp/maps"
```
AlexanderYastrebov, line 24: This is not in stdlib, see https://pkg.go.dev/maps

RomanZavodskikh: Yes, but there is no `maps.Keys` method on the stdlib page.
image
https://pkg.go.dev/maps

The experimental for comparison.
image
https://pkg.go.dev/golang.org/x/exp/maps

AlexanderYastrebov: Right, indeed https://github.com/golang/go/issues/61538. Thanks for checking

PR #3069, opened 2024-05-07
config: add yaml flag

AlexanderYastrebov:
YAML flag parses (preferably flow-style) YAML value from the command line or unmarshals object yaml value from the yaml config file.

Example value:
```
bin/skipper -foo-flag='{foo: hello, bar: [world, "1"], baz: 2, qux: {baz: 3}}'
```

and equivalent branch in config yaml:
```yaml
foo-flag:
  foo: hello
  bar:
    - world
    - "1"
  baz: 2
  qux:
    baz: 3
```

This will be useful for #2104

It is also a better alternative to manual parsing of micro-syntax like e.g. implemented in #2888

### DIFF
--- a/config/yamlflag.go
+++ b/config/yamlflag.go
@@ -1,0 +1,42 @@
+package config
+
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v2"
+)
+
+type yamlFlag[T any] struct {
+	Ptr   **T
+	value string // only for Set
+}
+
+func newYamlFlag[T any](ptr **T) *yamlFlag[T] {
+	return &yamlFlag[T]{Ptr: ptr}
+}
+
+func (yf *yamlFlag[T]) Set(value string) error {
+	var opts T
+	if err := yaml.Unmarshal([]byte(value), &opts); err != nil {
+		return fmt.Errorf("failed to parse yaml: %w", err)
+	}
+	*yf.Ptr = &opts
+	yf.value = value
+	return nil
+}
+
+func (yf *yamlFlag[T]) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var opts T
+	if err := unmarshal(&opts); err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


config/yamlflag_test.go
```diff
@@ -0,0 +1,83 @@
...
+		}{}
+
+		f := newYamlFlag(&cfg.Field)
+		f.Set("{}")
+
+		assert.Equal(t, &yamlFlagTestConfig{}, cfg.Field)
+		assert.Equal(t, "{}", f.String())
+	})
+
+	t.Run("unmarshal YAML", func(t *testing.T) {
```
MustafaSaber, line 61: I think we can another test case for non-valid yaml, wdyt?

AlexanderYastrebov: Makes sense, added.

config/yamlflag.go
```diff
@@ -0,0 +1,42 @@
+package config
+
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v2"
+)
+
+type yamlFlag[T any] struct {
+	Ptr   **T
```
szuecs, line 10: ouch :)
But I see why

szuecs:
:+1:

config/yamlflag.go
AlexanderYastrebov: The usage looks like
```go
type Config struct {
  OpenTelemetry *otel.Options `yaml:"open-telemetry"`
  ...
}
...
flag.Var(newYamlFlag(&cfg.OpenTelemetry), "open-telemetry", "OpenTelemetry configuration in YAML format, use flow-style for convenience")
...

if o.OpenTelemetry != nil {
...
}
```

The value is allocated when flag is set and stays nil when it is not, i.e. this allows to tell if flag was provided.

szuecs: Yes I saw it, all good  :)

AlexanderYastrebov:
:+1:

PR #3071, opened 2024-05-08
phc: refactor testcases

MustafaSaber:
followup on https://github.com/zalando/skipper/pull/3058

Tried to add test case for all unhealthy but was a little bit complicated, maybe will give another try later.

### DIFF
--- a/proxy/healthy_endpoints_test.go
+++ b/proxy/healthy_endpoints_test.go
@@ -101,6 +101,17 @@ func setupServices(t *testing.T, healthy, unhealthy int) []string {
 	return services
 }
 
+func formatServicesString(services []string) string {
+	backends := ""
+	for i, service := range services {
+		backends += fmt.Sprintf(`"%s"`, service)
+		if i < len(services)-1 {
+			backends += ", "
+		}
+	}
+	return backends
+}
+
 func TestPHCWithoutRequests(t *testing.T) {
 	services := setupServices(t, 3, 0)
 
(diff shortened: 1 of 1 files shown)


proxy/healthy_endpoints_test.go
```diff
@@ -150,142 +161,84 @@ func TestPHCForSingleHealthyEndpoint(t *testing.T) {
...
+						algorithm, servicesString))
+
+					va := fireVegeta(t, ps, 3000, 1*time.Second, 5*time.Second)
+
+					count200, ok := va.CountStatus(200)
+					assert.True(t, ok)
+
+					count504, ok := va.CountStatus(504)
+					assert.Condition(t, func() bool {
+						if tt.unhealthy == 0 && (ok || count504 == 0) {
```
RomanZavodskikh, line 189: Could you kindly describe in English language (here or in comment to this code), what this function should check?

proxy/healthy_endpoints_test.go
```diff
@@ -150,142 +161,84 @@ func TestPHCForSingleHealthyEndpoint(t *testing.T) {
...
+					MinRequests:                   1, // with 2 test case fails on github actions
+					MaxHealthCheckDropProbability: 0.95,
+					MinHealthCheckDropProbability: 0.01,
+				})
+				mockMetrics, ps := setupProxyWithCustomEndpointRegisty(t, fmt.Sprintf(`* -> backendTimeout("5ms") -> consistentHashKey("${request.header.ConsistentHashKey}") -> consistentHashBalanceFactor(1.25) -> <consistentHash, %s>`,
+					servicesString), endpointRegistery)
+				failedReqs := sendGetRequests(t, ps)
+				assert.InDelta(t, 0, failedReqs, 0.2*float64(nRequests))
+				mockMetrics.WithCounters(func(counters map[string]int64) {
+					assert.InDelta(t, float64(tt.unhealthy*nRequests), float64(counters["passive-health-check.endpoints.dropped"]), 0.3*float64(nRequests)*float64(tt.unhealthy))
```
RomanZavodskikh, line 239: minor: I see `0.3*float64(nRequests)*float64(tt.unhealthy)` here and `0.3*float64(tt.unhealthy*nRequests)` above. Both options are fine for me, but I think they should be the same for all test cases to increase readability.

proxy/healthy_endpoints_test.go
```diff
@@ -150,142 +161,84 @@ func TestPHCForSingleHealthyEndpoint(t *testing.T) {
...
+						}
+					})
+
+					failedRequests := math.Abs(float64(va.TotalRequests())) - float64(count200)
+					t.Logf("total requests: %d, count200: %d, count504: %d, failedRequests: %f", va.TotalRequests(), count200, count504, failedRequests)
+
+					assert.InDelta(t, float64(count504), failedRequests, 5)
+					assert.InDelta(t, 0, float64(failedRequests), 0.3*float64(va.TotalRequests()))
+					mockMetrics.WithCounters(func(counters map[string]int64) {
+						assert.InDelta(t, float64(tt.unhealthy)*float64(va.TotalRequests()), float64(counters["passive-health-check.endpoints.dropped"]), 0.6*float64(va.TotalRequests()))
```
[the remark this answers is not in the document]
RomanZavodskikh: Or `0.3*float64(tt.unhealthy*va.TotalRequests())` if you like.

proxy/healthy_endpoints_test.go
```diff
@@ -101,6 +101,17 @@ func setupServices(t *testing.T, healthy, unhealthy int) []string {
 	return services
 }
 
+func formatServicesString(services []string) string {
```
RomanZavodskikh, line 104: I like this function. :-) Really helpful to avoid fmt.Printf arguments mess.

proxy/healthy_endpoints_test.go
```diff
@@ -101,6 +101,17 @@ func setupServices(t *testing.T, healthy, unhealthy int) []string {
...
+	backends := ""
+	for i, service := range services {
+		backends += fmt.Sprintf(`"%s"`, service)
```
szuecs, line 107: `+=` for string is not great in Go.
Use strings.Builder instead of string and return builder.String()

proxy/healthy_endpoints_test.go
MustafaSaber: if we don't have unhealthy endpoints, then we don't have 504s this mean that `assert.True` would fail which we don't want. The condition here check if we have don't have unhealthy endpoints then the metric shouldn't find the key for 504 or at least return 0, the other case can be simplified to `return count504 > 0`, does this make more sense?

proxy/healthy_endpoints_test.go
szuecs: Can't you use strings.Join() to achieve the same?

MustafaSaber: I don't know will check, here we need to include '"' in the string I don't think that's feasible from strings.Join but will check

proxy/healthy_endpoints_test.go
RomanZavodskikh: Ah, I see, in other words, this condition checks:

+ if there're no unhealthy endpoints: check whether `ok` or `count504`
+ if there's at least one unhealthy endpoint: check `ok` only

Okay, now makes sense to me.

RomanZavodskikh:
:+1:

MustafaSaber:
👍

PR #3081, opened 2024-05-14
Added max-unhealthy-endpoints-ratio cmdline parameter for PHC

RomanZavodskikh:
This parameter is nice to have because PHC feature was designed to mitigate some small amount of unhealthy endpoints.
So, it is better to fail open if the amount of unhealthy endpoints is "too much".

The introduced parameter configures this "too much".


proxy/proxy.go
```diff
@@ -162,14 +162,20 @@ type PassiveHealthCheck struct {
...
 }
 
 func InitPassiveHealthChecker(o map[string]string) (bool, *PassiveHealthCheck, error) {
 	if len(o) == 0 {
 		return false, &PassiveHealthCheck{}, nil
 	}
 
-	result := &PassiveHealthCheck{}
+	result := &PassiveHealthCheck{
+		MaxUnhealthyEndpointsRatio: 1.0,
```
AlexanderYastrebov, line 177: Nice.

Could you also add another default optional explicitly (`MinDropProbability: 0.0`)?

proxy/healthy_endpoints.go
```diff
@@ -19,20 +19,22 @@ func (h *healthyEndpoints) filterHealthyEndpoints(ctx *context, endpoints []rout
...
 			ctx.Logger().Infof("Dropping endpoint %q due to passive health check: p=%0.2f, dropProbability=%0.2f",
 				e.Host, p, dropProbability)
 			metrics.IncCounter("passive-health-check.endpoints.dropped")
+			unhealthyEndpointsCount++
 		} else {
 			filtered = append(filtered, e)
 		}
-	}
 
-	if len(filtered) == 0 {
```
AlexanderYastrebov, line 34: I think we need either to keep this check or use `>=` condition

AlexanderYastrebov: Please add a test for this edge case.

docs/operation/operation.md
```diff
@@ -926,6 +931,7 @@ The parameters of `-passive-health-check` option are:
 + `min-drop-probabilty=<0.0 <= p <= 1.0>` - the minimum possible probability of unhealthy endpoint being not considered while choosing the endpoint for the given request. The same value is in fact used as minimal failed requests ratio for PHC to be enabled for this endpoint
 + `max-drop-probabilty=<0.0 <= p <= 1.0>` - the maximum possible probability of unhealthy endpoint being not considered
```
szuecs, line 932: I think we also miss here that `min-drop-probabilty < max-drop-probabilty` should hold true. Not sure if we check this in code somewhere but sounds like we should.

docs/operation/operation.md
```diff
@@ -926,6 +931,7 @@ The parameters of `-passive-health-check` option are:
...
 while choosing the endpoint for the given request
++ `max-unhealthy-endpoints-ratio=<0.0 <= r <= 1.0>` - the maximum ratio of unhealthy endpoints for PHC to try to mitigate ongoing requests
```
szuecs, line 934: Maybe we should not use angle brackets `<>` but `[]`. At least in my experience in math we use these which makes more sense to at least me. :)

docs/operation/operation.md
```diff
@@ -904,15 +904,20 @@ the Skipper takes a look at previous period and if the amount of requests in the
...
+if there are too many unhealthy endpoints PHC does not try to mitigate them any more and requests are sent like there is no PHC at all.
 
 Having this, we expect less requests to fail because a lot of them would be sent to endpoints that seem to be healthy instead.
 
 To enable this feature, you need to provide `-passive-health-check` option having forementioned parameters
-(`period`, `min-requests`, `min-drop-probability`, `max-drop-probability`) defined.
+(`period`, `min-requests`, `min-drop-probability`, `max-drop-probability`, `max-unhealthy-endpoints-ratio`) defined.
 `period`, `min-requests`, `max-drop-probability` are required parameters, it is not possible for PHC to be enabled without
 them explicitly defined by user. `min-drop-probability` is implicitly defined as `0.0` if not explicitly set by user.
+`max-unhealthy-endpoints-ratio` is defined as `1.0` if not expllicilty set by user.
```
szuecs, line 916: expllicilty -> explicitly

proxy/proxy.go
```diff
@@ -162,14 +162,20 @@ type PassiveHealthCheck struct {
 
 	// The maximum probability of unhealthy endpoint to be dropped out from load balancing for every specific request
 	MaxDropProbability float64
+
+	// The maximum ratio of unhealthy endpoints in the list of all endpoints PHC will check in case of all endpoints
```
szuecs, line 166: Proper go doc starts with the var name: `MaxUnhealthyEndpointsRatio is the maximum ratio of unhealthy ...`

proxy/proxy.go
```diff
@@ -218,6 +224,15 @@ func InitPassiveHealthChecker(o map[string]string) (bool, *PassiveHealthCheck, e
 				return false, nil, fmt.Errorf("passive health check: invalid minDropProbability value: %s", value)
 			}
 			result.MinDropProbability = minDropProbability
+		case "max-unhealthy-endpoints-ratio":
+			maxUnhealthyEndpointsRatio, err := strconv.ParseFloat(value, 64)
+			if err != nil {
+				return false, nil, fmt.Errorf("passive health check: invalid maxUnhealthyEndpointsRatio value: %s", value)
```
szuecs, line 230: %s -> %q

proxy/proxy.go
```diff
@@ -218,6 +224,15 @@ func InitPassiveHealthChecker(o map[string]string) (bool, *PassiveHealthCheck, e
...
+			}
+			if maxUnhealthyEndpointsRatio < 0 || maxUnhealthyEndpointsRatio > 1 {
+				return false, nil, fmt.Errorf("passive health check: invalid maxUnhealthyEndpointsRatio value: %s", value)
```
szuecs, line 233: %s -> %q

proxy/proxy_test.go
```diff
@@ -2296,148 +2296,162 @@ func TestInitPassiveHealthChecker(t *testing.T) {
...
 		{
 			inputArg: map[string]string{
-				"period":               "1m",
-				"min-requests":         "10",
-				"max-drop-probability": "0.05",
-				"min-drop-probability": "0.9",
+				"period":                        "1m",
+				"min-requests":                  "10",
+				"max-drop-probability":          "0.05",
+				"min-drop-probability":          "0.9",
```
szuecs, line 2440: here you test `min > max`, so just missing docs as far as I see

proxy/healthy_endpoints_test.go
```diff
@@ -241,3 +244,39 @@ func TestPHC(t *testing.T) {
 		})
 	}
 }
+
+func TestPHCMaxUnhealthyEndpointsEdgeCase(t *testing.T) {
```
AlexanderYastrebov, line 248: EdgeCase -> AllUnhealthy

proxy/healthy_endpoints_test.go
```diff
@@ -241,3 +244,39 @@ func TestPHC(t *testing.T) {
...
+	)
+
+	servicesString := setupServices(t, healthy, unhealthy)
+	endpointregistry := routing.NewEndpointRegistry(routing.RegistryOptions{
+		PassiveHealthCheckEnabled:     true,
+		StatsResetPeriod:              period,
+		MinRequests:                   2,
+		MaxHealthCheckDropProbability: 0.95,
+		MinHealthCheckDropProbability: 0.01,
+		MaxUnhealthyEndpointsRatio:    0.5, // because of this setting PHC expected effectively not to work here
```
AlexanderYastrebov, line 261: I think the edge case is when this is 1.0, then
```go
unhealthyEndpointsCount == len(endpoints)*e.Metrics.MaxUnhealthyEndpointsRatio()
```

RomanZavodskikh: Yeah, I just realized that the "edge case" is "all endpoints are unhealthy", implemented such case instead here:
https://github.com/zalando/skipper/pull/3081/files#diff-b1bbc8a9c24c3b87d0de99e08ebcdd6df2da583d27fa530049a3cc12d496435cR248

proxy/healthy_endpoints_test.go
RomanZavodskikh: I think `TestPHCNoHealthyEndpoints` makes much sense too.

proxy/healthy_endpoints.go
```diff
@@ -19,16 +19,22 @@ func (h *healthyEndpoints) filterHealthyEndpoints(ctx *context, endpoints []rout
...
 		if p < dropProbability {
 			ctx.Logger().Infof("Dropping endpoint %q due to passive health check: p=%0.2f, dropProbability=%0.2f",
 				e.Host, p, dropProbability)
 			metrics.IncCounter("passive-health-check.endpoints.dropped")
+			unhealthyEndpointsCount++
 		} else {
 			filtered = append(filtered, e)
 		}
+
+		if float64(unhealthyEndpointsCount) > float64(len(endpoints))*e.Metrics.MaxUnhealthyEndpointsRatio() {
```
AlexanderYastrebov, line 35: I think we can pre-compute allowed ratio outside of the loop.

routing/endpointregistry.go
```diff
@@ -227,6 +230,10 @@ func (r *EndpointRegistry) GetMetrics(hostPort string) Metrics {
 	return e.(*entry)
 }
 
+func (r *EndpointRegistry) MaxUnhealthyEndpointsRatio() float64 {
```
AlexanderYastrebov, line 233: Why do we need this to be a part of Endpoint Registry?
I think proxy healthchecker should have access to this value from PassiveHealthCheck config.

AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

szuecs:
:+1:

PR #3086, opened 2024-05-29
Reduce errors in decision logs that are operational

mjungsbluth:
Currently the Open Policy Agent filters send decision logs with an error to the control plane if the http path is invalid UTF-8 or if the evaluation is cancelled (the client closes the connection before the evaluation finishes). 

This disguises real errors and leads to a non-zero baseline of decision errors in the control plane. 

This PR pre-validates if paths are valid UTF-8 and do not send decisions for cancelled evalutations to the control plane.


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #3089, opened 2024-05-30
metrics: add start label for prometheus counters

### LINKED ISSUE #3087: Missing Start Timestamp for Cumulative Metrics in Skipper Metrics
In OpenTelemetry, every [metric of cumulative type](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#temporality) is required to have a `StartTimestamp`, which records when a metric is first recorded. However, Prometheus exporter doesn't provide `StartTimestamp` by default. The absence of `StartTimestamp` leads to inaccuracies that distort the telemetry data.

Example 01: When new `StartTimestamp` is less than actual `StartTimestamp`, it creates "fake tail" in metric values

image

Example 02: When new `StartTimestamp` is more than actual `StartTimestamp`, it creates "fake peak" in metric values

image

As you can see from the examples above, incorrect choice of `StartTimestamp` on OpenTelemetry side always leads to distortions in collected metrics data. And without `StartTimestamp` on Skipper side there is no chance to correctly guess `StartTimestamp` on telemetry collector side.

To remove these side effects in telemetry data, we need StartTimestamp to be set for each cumulative metric by Skipper. The value of StartTimestamp for a particular metric must be equal to the UNIX timestamp (in nanoseconds) of when this metric, with the exact same combination of attributes, was first seen by Skipper.

image

In our setup we also tested out start timestamp normalization algorithm, [introduced by OpenTelemetry in this article](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#cumulative-streams-handling-unknown-start-time). It describes a set of transformations to the cumulative sum metrics that helps handling unknown start timestamp. This approach can be used as a workaround in situations, where getting start timestamp of stream is not possible at all. But like any other hacks, this one has its own disadvantages. During collector restarts metric data will be completely lost:

image

This is a critical problem, since usually one OpenTelemetry collector is used per cluster / installation, so restart will cause loss of data. This can be definetly avoided in case we have `StartTimestamp` for each cumulative metric on Skipper side.

Prometheus Data Model: https://prometheus.io/docs/concepts/metric_types/
OpenTelemetry Data Model: https://opentelemetry.io/docs/specs/otel/metrics/data-model/

Yes

AlexanderYastrebov:
Add start label to each counter with the value of counter creation timestamp as unix nanoseconds.

This enables OpenTelemetry cumulative temporality, see https://opentelemetry.io/docs/specs/otel/metrics/data-model/#temporality

Example:
```
~$ curl -s localhost:9911/metrics | grep host_count
 # HELP skipper_serve_host_count Total number of requests of serving a host.
 # TYPE skipper_serve_host_count counter
skipper_serve_host_count{code="200",host="bar_test",method="GET",start="1717066533598500794"} 1
skipper_serve_host_count{code="200",host="foo_test",method="GET",start="1717066538031805059"} 2
```

Fixes #3087

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -70,6 +70,7 @@ type Config struct {
 
 	// logging, metrics, profiling, tracing:
 	EnablePrometheusMetrics             bool      `yaml:"enable-prometheus-metrics"`
+	EnablePrometheusStartLabel          bool      `yaml:"enable-prometheus-start-label"`
 	OpenTracing                         string    `yaml:"opentracing"`
 	OpenTracingInitialSpan              string    `yaml:"opentracing-initial-span"`
 	OpenTracingExcludedProxyTags        string    `yaml:"opentracing-excluded-proxy-tags"`
@@ -378,6 +379,7 @@ func NewConfig() *Config {
 	flag.IntVar(&cfg.BlockProfileRate, "block-profile-rate", 0, "block profile sample rate, see runtime.SetBlockProfileRate")
 	flag.IntVar(&cfg.MutexProfileFraction, "mutex-profile-fraction", 0, "mutex profile fraction rate, see runtime.SetMutexProfileFraction")
 	flag.IntVar(&cfg.MemProfileRate, "memory-profile-rate", 0, "memory profile rate, see runtime.SetMemProfileRate, keeps default 512 kB")
+	flag.BoolVar(&cfg.EnablePrometheusStartLabel, "enable-prometheus-start-label", false, "adds start label to each prometheus counter with the value of counter creation timestamp as unix nanoseconds")
 	flag.BoolVar(&cfg.DebugGcMetrics, "debug-gc-metrics", false, "enables reporting of the Go garbage collector statistics exported in debug.GCStats")
 	flag.BoolVar(&cfg.RuntimeMetrics, "runtime-metrics", true, "enables reporting of the Go runtime statistics exported in runtime and specifically runtime.MemStats")
 	flag.BoolVar(&cfg.ServeRouteMetrics, "serve-route-metrics", false, "enables reporting total serve time metrics for each route")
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -378,6 +379,7 @@ func NewConfig() *Config {
 	flag.IntVar(&cfg.BlockProfileRate, "block-profile-rate", 0, "block profile sample rate, see runtime.SetBlockProfileRate")
 	flag.IntVar(&cfg.MutexProfileFraction, "mutex-profile-fraction", 0, "mutex profile fraction rate, see runtime.SetMutexProfileFraction")
 	flag.IntVar(&cfg.MemProfileRate, "memory-profile-rate", 0, "memory profile rate, see runtime.SetMemProfileRate, keeps default 512 kB")
+	flag.BoolVar(&cfg.EnableStartTimestamp, "enable-start-timestamp", false, "only for prometheus: adds start label to each counter with the value of counter creation timestamp as unix nanoseconds")
```
szuecs, line 382: Maybe better to have also `label` in this name and flag

skipper.go
```diff
@@ -458,6 +458,10 @@ type Options struct {
 	// MemProfileRate calls runtime.SetMemProfileRate(MemProfileRate) if non zero value, deactivate with <0
 	MemProfileRate int
 
+	// EnableStartTimestamp adds start label to each counter with the value of counter creation
+	// timestamp as unix nanoseconds.
+	EnableStartTimestamp bool
```
szuecs, line 463: Maybe include `Prometheus`?!

szuecs:
:+1:

config/config.go
AlexanderYastrebov: Fixed, thanks

skipper.go
AlexanderYastrebov: Fixed, thanks

szuecs:
:+1:

PR #3095, opened 2024-06-03
build: trim path

AlexanderYastrebov:
Add -trimpath flag to remove filesystem paths and make stacktraces shorter, see https://pkg.go.dev/cmd/go


MustafaSaber:
👍

AlexanderYastrebov:
:+1:

PR #3097, opened 2024-06-05
build: implement DCO check

AlexanderYastrebov:
Add dco make target and run it in the build pipeline to replace DCO application that is currently unavailable due to https://github.com/dcoapp/app/issues/211


szuecs:
DCO app seems to be fixed, so I think we can close this one

PR #3098, opened 2024-06-06
Set log level of PHC dropping an endpoint to debug

RomanZavodskikh:
This is needed not to log too many lines. Moreover, the information about requests amount could be seen in the metrics, the information about IPs being mitigated could be get from updateStats logs


RomanZavodskikh:
:+1:

PR #3112, opened 2024-06-17
routing: log route update id

AlexanderYastrebov:
Add and log route update counter to observe delay between receiving and applying routes.

### DIFF
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -511,6 +511,7 @@ func processRouteDefs(o Options, fr filters.Registry, defs []*eskip.Route) (rout
 }
 
 type routeTable struct {
+	id            int
 	m             *matcher
 	once          sync.Once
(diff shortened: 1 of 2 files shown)


routing/datasource.go
```diff
@@ -543,10 +544,10 @@ func receiveRouteMatcher(o Options, out chan<- *routeTable, quit <-chan struct{}
 		updatesRelay <-chan []*eskip.Route
 	)
 	updatesRelay = updates
-	for {
+	for id := 1; ; id++ {
```
RomanZavodskikh, line 547: Is is safe to use non-atomic increment here?

RomanZavodskikh: Ah, okay, there is no concurrent access to this variable. :-)

RomanZavodskikh:
:+1:

PR #3113, opened 2024-06-17
routing: measure route update latency

AlexanderYastrebov:
* add route update latency metric - the time it takes to apply routes update after receival.
* remove redundant setting of `created` location to UTC - its value is only exposed via Unix() which is always in UTC.


szuecs:
:+1:

MustafaSaber:
👍

AlexanderYastrebov:
:+1:

PR #3116, opened 2024-06-18
routing: refactor update id logging

AlexanderYastrebov:
Increment update id only on update receival instead of each loop cycle which results in odd update id values due to second select case branch.

Follow up on #3112


RomanZavodskikh:
:+1:

AlexanderYastrebov:
:+1:

PR #3118, opened 2024-06-19
OPA: Add response status to control plane traces

mjungsbluth:
The current span for control plane requests of the Open Policy Agent based filters does not include the actual response from the control plane. 

Apart from hard network errors this makes it hard to pinpoint authN / authZ failures or intermittent unavailability (5xx status) of the control plane. 

This PR both adds the control plane responses' http status code to the spans and also flags them as erroneous for all status with a value of 4xx or 5xx.


AlexanderYastrebov:
:+1:

MustafaSaber:
Does this needs to be reflected in `Open Policy Agent span` docs?

szuecs:
> Does this needs to be reflected in `Open Policy Agent span` docs?

let's document operations things if we have more experience with operating it.

Issue #3119, opened 2024-06-20
OPA: Reduce impact of filter creation time on routes


mjungsbluth:
**Is your feature request related to a problem? Please describe.**
The Open Policy Agent filters wait for a default of up to 30s to get functional, filter creation for example fails if the bundle cannot be loaded in time or the control plane for bundle / discovery download is unavailable / rejects the request. 

In the current there are multiple issues this can cause:
- OPA instances are shared amongst filter instances and cached but created serially. In case routes use a lot of different bundles, this can slow down the creation (due to a fairly broad mutex)
- Even if the discovery and bundle plugin have already reported errors, the way OPA reports plugin success, creation waits until timeout (could be fail fast)
- Blocking filter creation too long can block Endpoint updates, i.e. can lead to intermittent 502s on the current routes 

**Describe the solution you would like**

For the startup, the the implementation can be changed to fail fast if errors in discovery / bundle download are detected. The consequence of this would be that the filter would also fail fast even on intermittent 5xx in bundle download or network timeouts. 

OPA instance creation can be parallelized (i.e. the current mutex on the registry could just guard creation of a `sync.Once` per instance and the actually startup of the instances can run in parallel for multiple filters using different bundles. 

We can optionally make the creation of the filter asynchronous in the sense that bootstrap is done in the context of the filter but there is no wait on successfully loading the bundle / finishing discovery. The consequences of this would be that route creation is very fast but potential problems would be pushed to the data plane, i.e. requests would either block or return 5xx status for a while. The latter would be especially problematic if you introduce the filter into an existing route. 

**Would you like to work on it?**
Yes

PR #3120, opened 2024-06-20
OPA: Fail fast on discovery or bundle download errors

mjungsbluth:
As part of #3119, this PR changes the behaviour on starting up OPA instances inside the OPA based Skipper filters.

Currently even in the face of download errors of bundles, the default timeout of 30s is applied. 

By using listeners both for bundle download and discovery bundle download, we can detect a failure early despite the respective plugins still waiting for the bundles to become available.

Still a draft because tests are missing, there are some cleanups in the code that need to be done. 

Pending: Unregister discovery and bundle plugin listeners.

### DIFF
--- a/filters/openpolicyagent/openpolicyagent.go
+++ b/filters/openpolicyagent/openpolicyagent.go
@@ -21,6 +21,7 @@ import (
 	"github.com/open-policy-agent/opa/config"
 	"github.com/open-policy-agent/opa/logging"
 	"github.com/open-policy-agent/opa/plugins"
+	"github.com/open-policy-agent/opa/plugins/bundle"
 	"github.com/open-policy-agent/opa/plugins/discovery"
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/runtime"
@@ -480,25 +481,50 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, configBytes []
 // Start asynchronously starts the policy engine's plugins that download
 // policies, report status, etc.
 func (opa *OpenPolicyAgentInstance) Start(ctx context.Context, timeout time.Duration) error {
-	err := opa.manager.Start(ctx)
+	discoveryPlugin := discovery.Lookup(opa.manager)
+	bundlePlugin := bundle.Lookup(opa.manager)
 
+	done := make(chan struct{})
+	defer close(done)
+	failed := make(chan error)
+	defer close(failed)
+
+	discoveryPlugin.RegisterListener("startuplistener", func(status bundle.Status) {
+		if len(status.Errors) > 0 {
+			failed <- fmt.Errorf("discovery download failed: %w", errors.Join(status.Errors...))
+		}
+	})
+
+	bundlePlugin.Register("startuplistener", func(status bundle.Status) {
+		if len(status.Errors) > 0 {
+			failed <- fmt.Errorf("bundle activation failed: %w", errors.Join(status.Errors...))
+		}
+	})
+	defer bundlePlugin.Unregister("startuplistener")
+
+	opa.manager.RegisterPluginStatusListener("startuplistener", func(status map[string]*plugins.Status) {
+		for _, pluginstatus := range status {
+			if pluginstatus != nil && pluginstatus.State != plugins.StateOK {
+				return
+			}
+		}
+		close(done)
+	})
+	defer opa.manager.UnregisterPluginStatusListener("startuplistener")
+
+	err := opa.manager.Start(ctx)
 	if err != nil {
 		return err
 	}
 
-	// check readiness of all plugins
-	pluginsReady := func() bool {
-		for _, status := range opa.manager.PluginStatus() {
-			if status != nil && status.State != plugins.StateOK {
-				return false
-			}
-		}
-		return true
-	}
+	select {
+	case <-done:
+		return nil
+	case err := <-failed:
+		opa.Close(ctx)
 
-	err = waitFunc(ctx, pluginsReady, 100*time.Millisecond)
-
-	if err != nil {
+		return err
+	case <-ctx.Done():
 		for pluginName, status := range opa.manager.PluginStatus() {
 			if status != nil && status.State != plugins.StateOK {
 				opa.Logger().WithFields(map[string]interface{}{
@@ -511,7 +537,6 @@ func (opa *OpenPolicyAgentInstance) Start(ctx context.Context, timeout time.Dura
 		opa.Close(ctx)
 		return fmt.Errorf("one or more open policy agent plugins failed to start in %v with error: %w", timeout, err)
 	}
-	return nil
 }
 
 func (opa *OpenPolicyAgentInstance) Close(ctx context.Context) {
--- a/filters/openpolicyagent/openpolicyagent.go
+++ b/filters/openpolicyagent/openpolicyagent.go
@@ -481,36 +481,47 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, configBytes []
 // Start asynchronously starts the policy engine's plugins that download
 // policies, report status, etc.
 func (opa *OpenPolicyAgentInstance) Start(ctx context.Context, timeout time.Duration) error {
+
 	discoveryPlugin := discovery.Lookup(opa.manager)
-	bundlePlugin := bundle.Lookup(opa.manager)
 
 	done := make(chan struct{})
-	defer close(done)
-	failed := make(chan error)
-	defer close(failed)
+	failed := make(chan error, 1)
 
-	discoveryPlugin.RegisterListener("startuplistener", func(status bundle.Status) {
-		if len(status.Errors) > 0 {
-			failed <- fmt.Errorf("discovery download failed: %w", errors.Join(status.Errors...))
+	var registerDiscoveryListenerOnce sync.Once
+	registerDiscoveryListenerOnce.Do(func() {
+		discoveryPlugin.RegisterListener("skipper-instance-startup-discovery", func(status bundle.Status) {
+			handleStatusErrors(status, failed, "discovery plugin")
+		})
+	})
+	defer discoveryPlugin.Unregister("skipper-instance-startup-discovery")
+
+	// Register listener for "bundle" status
+	var registerBundleListenerOnce sync.Once
+	opa.manager.RegisterPluginStatusListener("skipper-instance-startup-plugin", func(status map[string]*plugins.Status) {
+		if _, exists := status["bundle"]; exists {
+			bundlePlugin := bundle.Lookup(opa.manager)
+			if bundlePlugin != nil {
+				registerBundleListenerOnce.Do(func() {
+					bundlePlugin.Register("skipper-instance-startup-bundle", func(status bundle.Status) {
+						handleStatusErrors(status, failed, "bundle plugin")
+					})
+				})
+				defer bundlePlugin.Unregister("skipper-instance-startup-bundle")
+			}
 		}
 	})
+	defer opa.manager.UnregisterPluginStatusListener("skipper-instance-startup-plugin")
 
-	bundlePlugin.Register("startuplistener", func(status bundle.Status) {
-		if len(status.Errors) > 0 {
-			failed <- fmt.Errorf("bundle activation failed: %w", errors.Join(status.Errors...))
-		}
-	})
-	defer bundlePlugin.Unregister("startuplistener")
-
-	opa.manager.RegisterPluginStatusListener("startuplistener", func(status map[string]*plugins.Status) {
-		for _, pluginstatus := range status {
-			if pluginstatus != nil && pluginstatus.State != plugins.StateOK {
+	// Register listener for general plugin status checks
+	opa.manager.RegisterPluginStatusListener("generalStartUpListener", func(status map[string]*plugins.Status) {
+		for _, pluginStatus := range status {
+			if pluginStatus != nil && pluginStatus.State != plugins.StateOK {
 				return
 			}
 		}
 		close(done)
 	})
-	defer opa.manager.UnregisterPluginStatusListener("startuplistener")
+	defer opa.manager.UnregisterPluginStatusListener("generalStartUpListener")
 
 	err := opa.manager.Start(ctx)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


filters/openpolicyagent/openpolicyagent.go
```diff
@@ -476,25 +477,71 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, configBytes []
 // Start asynchronously starts the policy engine's plugins that download
 // policies, report status, etc.
 func (opa *OpenPolicyAgentInstance) Start(ctx context.Context, timeout time.Duration) error {
-	err := opa.manager.Start(ctx)
+	discoveryPlugin := discovery.Lookup(opa.manager)
+	bundlePlugin := bundle.Lookup(opa.manager)
```
mjungsbluth, line 481: I guess this needs to happen later when the discovery plugin is actually ready

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -476,25 +477,71 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, configBytes []
...
-			if status != nil && status.State != plugins.StateOK {
-				return false
+	done := make(chan struct{})
+	defer close(done)
+	failed := make(chan error)
+	defer close(failed)
+
+	discoveryPlugin.RegisterListener("startuplistener", func(status bundle.Status) {
+		if len(status.Errors) > 0 {
+			select {
```
mjungsbluth, line 497: I don't think the select is needed here (I wouldn't know what it even does in this case), sending to a channel can be done without it.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -476,25 +477,71 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, configBytes []
...
 			}
+			return
 		}
-		return true
-	}
 
-	err = waitFunc(ctx, pluginsReady, 100*time.Millisecond)
+		// Add listener for bundle plugin on discovery plugin OK, because bundlePlugin is dependent on the configuration
+		// discovered using the discovery configs
+		bundlePlugin.Register("startuplistener", func(status bundle.Status) {
```
mjungsbluth, line 506: Given the comment above, the check on status seems to be missing...

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -43,6 +44,7 @@ const (
 	defaultReuseDuration       = 30 * time.Second
 	defaultShutdownGracePeriod = 30 * time.Second
 	DefaultOpaStartupTimeout   = 30 * time.Second
+	DefaultWaitToDiscover      = 5 * time.Second
```
mjungsbluth, line 47: Should the Discovery events exactly avoid the need for yet another timeout?

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -476,38 +477,74 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, configBytes []
...
+
+	select {
+	case <-done:
+		return nil
+	case err := <-failed:
+		opa.Close(ctx)
+
+		return err
+	case <-ctx.Done():
+		timeoutErr := ctx.Err()
```
ZunairaSid, line 531: Here in context canceled or times-out case, we should clean up resources (opa.Close(ctx)) and return the appropriate error.

szuecs: As far as I see that will be done in line 542.

Pushpalanka:
Addressing the raise condition in the upstream as a part of [this reported issue](https://github.com/open-policy-agent/opa/issues/6849).

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -476,38 +477,74 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, configBytes []
...
+	done := make(chan struct{})
+	failed := make(chan error, 1)
 
-	// check readiness of all plugins
-	pluginsReady := func() bool {
-		for _, status := range opa.manager.PluginStatus() {
-			if status != nil && status.State != plugins.StateOK {
-				return false
+	discoveryPlugin.RegisterListener("startuplistener", func(status bundle.Status) {
+		handleStatusErrors(status, failed, "discovery plugin")
```
mjungsbluth, line 486: I wonder if this should also be wrapped in a `sync.Once`. We cannot [currently unregister](https://github.com/open-policy-agent/opa/pull/6851) the listener and if it somehow triggers multiple times, it will write to a closed channel which will panic.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -801,3 +819,13 @@ func (l *QuietLogger) Error(fmt string, a ...interface{}) {
...
 	l.target.Warn(fmt, a)
 }
+
+func handleStatusErrors(status bundle.Status, failed chan error, prefix string) {
+	if len(status.Errors) > 0 {
+		failed <- fmt.Errorf("%s failed: %w", prefix, errors.Join(status.Errors...))
+	}
+
+	if status.Code != "" {
+		failed <- fmt.Errorf("%s failed: %s", prefix, status.Code)
```
mjungsbluth, line 829: I suspect that in this case, `len(status.Errors) > 0`, i.e. you'll write two errors to the channel which means the second will never be shown. Can you verify that this is actually visible in tests?

Pushpalanka: It seems more reliable to depend on status.Code for errors as for [these conditions](https://github.com/open-policy-agent/opa/blob/v0.62.0/plugins/bundle/status.go#L76-L96). Modified accordingly with if else which will avoid writing twice. Will do a test case as well.

mjungsbluth: You will hide the actual error message which will lead to hard to debug situations. I‘d rather combine checking for errrors and the http status code (if set)

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -476,38 +477,77 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, configBytes []
...
 
-	err = waitFunc(ctx, pluginsReady, 100*time.Millisecond)
+	// Register listener for general plugin status checks
+	opa.manager.RegisterPluginStatusListener("generalStartUpListener", func(status map[string]*plugins.Status) {
+		for _, pluginStatus := range status {
+			if pluginStatus != nil && pluginStatus.State != plugins.StateOK {
+				return
+			}
+		}
+		done <- struct{}{}
```
AlexanderYastrebov, line 516: done channel is unbuffered and this will block until another thread reads from it.
For signalling it might be better to `close(done)`.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -476,38 +477,77 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, configBytes []
...
-	pluginsReady := func() bool {
-		for _, status := range opa.manager.PluginStatus() {
-			if status != nil && status.State != plugins.StateOK {
-				return false
+	done := make(chan struct{})
+	failed := make(chan error, 1)
+
+	var registerDiscoveryListenerOnce sync.Once
+	registerDiscoveryListenerOnce.Do(func() {
+		discoveryPlugin.RegisterListener("startuplistener", func(status bundle.Status) {
```
mjungsbluth, line 488: Since OPA 0.67 was just released, you can add a defer unregister…

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -476,38 +477,77 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, configBytes []
...
+			handleStatusErrors(status, failed, "discovery plugin")
+		})
+	})
+
+	// Register listener for "bundle" status
+	var registerBundleListenerOnce sync.Once
+	opa.manager.RegisterPluginStatusListener("bundlelistener", func(status map[string]*plugins.Status) {
```
mjungsbluth, line 495: This applies to all the other listeners as well: the names (first parameter) are global names in the opa instance, so might yield collisions. 
I‘d suggest sth more unique: `skipper-instance-startup`, also move it to an internal constant..

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -801,3 +822,46 @@ func (l *QuietLogger) Error(fmt string, a ...interface{}) {
...
+	_, exists := temporaryClientErrorHTTPCodes[code]
+	return exists
+}
+
+func handleStatusErrors(
+	status bundle.Status,
+	failed chan error,
+	prefix string,
+) {
+	if status.Code == "bundle_error" {
```
mjungsbluth, line 841: This still feels like we are only handling a subset of errors, is this really reliable? My understanding was that the Code only gives information about what went wrong and that length(status.Errors) essentially gives yoj if anything goes wrong…

filters/openpolicyagent/openpolicyagent_test.go
```diff
@@ -696,3 +697,141 @@ func TestBodyExtractionUnknownBody(t *testing.T) {
...
+	cfg, err := NewOpenPolicyAgentConfig(WithConfigTemplate(config))
+	assert.NoError(t, err)
+
+	_, err = registry.NewOpenPolicyAgentInstance("test", *cfg, "testfilter")
+	assert.Error(t, err)
+	assert.Equal(t, "discovery plugin failed: Name: discovery, Code: bundle_error, Message: server replied with Not Found, HTTPCode: 404, Errors: []", err.Error())
+}
+
+func TestDiscoveryRetryTemporaryError429(t *testing.T) {
+	testDiscoveryBundleError(t, http.StatusTooManyRequests, "one or more open policy agent plugins failed to start in 1s with error: context deadline exceeded")
```
mjungsbluth, line 715: To make this more idiomatic Go, maybe capture all the cases in an array of struct and loop over them. You can find this pattern in other OPA related tests

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -476,38 +477,77 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, configBytes []
...
-	// check readiness of all plugins
-	pluginsReady := func() bool {
-		for _, status := range opa.manager.PluginStatus() {
-			if status != nil && status.State != plugins.StateOK {
-				return false
+	done := make(chan struct{})
+	failed := make(chan error, 1)
+
+	var registerDiscoveryListenerOnce sync.Once
+	registerDiscoveryListenerOnce.Do(func() {
```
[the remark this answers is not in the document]
Pushpalanka: As discussed offline, we couldn't gain much removing this duplication in readability wise.

filters/openpolicyagent/openpolicyagent.go
Pushpalanka: That indeed is more intuitive, but had below concerns.

When looking at the OPA implementation below, `status.Errors` it is set only in 1 occasion out of the 3. But the other 2 occasions set `status.Errors` to nil even though it is setting an error. Based on the existing implementation status.Code is most reliable as I see. 

I also understand if the `bundle_error` constant is ever modified then we are in trouble. But then again, just relying on code to be non-empty is also not good if the code got ever set in an non-errornous scenario.

<details>
<summary>Status Error code segment, /plugins/bundle/status.go#L64-L97</summary>

```language
const (
	errCode = "bundle_error"
)

func (s *Status) SetError(err error) {
	var (
		astErrors ast.Errors
		httpError download.HTTPError
	)
	switch {
	case err == nil:
		s.Code = ""
		s.HTTPCode = ""
		s.Message = ""
		s.Errors = nil

	case errors.As(err, &astErrors):
		s.Code = errCode
		s.HTTPCode = ""
		s.Message = types.MsgCompileModuleError
		s.Errors = make([]error, len(astErrors))
		for i := range astErrors {
			s.Errors[i] = astErrors[i]
		}

	case errors.As(err, &httpError):
		s.Code = errCode
		s.HTTPCode = json.Number(strconv.Itoa(httpError.StatusCode))
		s.Message = err.Error()
		s.Errors = nil

	default:
		s.Code = errCode
		s.HTTPCode = ""
		s.Message = err.Error()
		s.Errors = nil
	}
}

```
</details>

Pushpalanka:
Moving to draft as using OPA 0.67.0 which is yet to be taken by Skipper dependency upgrades.

MustafaSaber:
> Moving to draft as using OPA 0.67.0, which is yet to be taken by Skipper dependency upgrades.

OPA 0.67.1 is merged now.

filters/openpolicyagent/openpolicyagent.go
szuecs: Sounds like we want to reach out to styra to make a proper error handling possible.
I also wonder about the typing here. Status.HTTPCode being a string and httpError not being an error, what the hack is this code?

Pushpalanka: Will raise this in the OPA community and get back.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -476,38 +477,77 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, configBytes []
...
 
-	// check readiness of all plugins
-	pluginsReady := func() bool {
-		for _, status := range opa.manager.PluginStatus() {
-			if status != nil && status.State != plugins.StateOK {
-				return false
+	done := make(chan struct{})
+	failed := make(chan error, 1)
+
+	var registerDiscoveryListenerOnce sync.Once
```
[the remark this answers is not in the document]
Pushpalanka: Addressed.

filters/openpolicyagent/openpolicyagent.go
Pushpalanka: Went ahead with the `bundle_error` constant as a change there will be communicated in OPA release notes.

Pushpalanka:
Requesting a review as the PR goes green without the 2 ToDo's to be addressed to unregister the listeners. 
Unregistering is removed as it prematurely unregistering the listeners without handling the errors. Appreciate inputs.

mjungsbluth:
Closed in favour of #3464

Issue #3121, opened 2024-06-20
Insecure cipher suites used for TLS 1.2


rickhlx:
**Describe the bug**
The config used for TLS 1.2 when terminating at Skipper is allowing clients to use two insecure cipher suites:

* [`TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA`](https://ciphersuite.info/cs/TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA/)
* [`TLS_RSA_WITH_3DES_EDE_CBC_SHA`](https://ciphersuite.info/cs/TLS_RSA_WITH_3DES_EDE_CBC_SHA/)

**To Reproduce**
Run nmap script `ssl-enum-ciphers` against an https ingress using skipper for TLS termination.

**Expected behavior**
```
PORT    STATE SERVICE
443/tcp open  https
| ssl-enum-ciphers: 
|   TLSv1.2: 
|     ciphers: 
|       TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA (secp256r1) - A
|       TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 (secp256r1) - A
|       TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA (secp256r1) - A
|       TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 (secp256r1) - A
|       TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256 (secp256r1) - A
|       TLS_RSA_WITH_AES_128_CBC_SHA (rsa 2048) - A
|       TLS_RSA_WITH_AES_128_GCM_SHA256 (rsa 2048) - A
|       TLS_RSA_WITH_AES_256_CBC_SHA (rsa 2048) - A
|       TLS_RSA_WITH_AES_256_GCM_SHA384 (rsa 2048) - A
|     compressors: 
|       NULL
|     cipher preference: client
|   TLSv1.3: 
|     ciphers: 
|       TLS_AKE_WITH_AES_128_GCM_SHA256 (ecdh_x25519) - A
|       TLS_AKE_WITH_AES_256_GCM_SHA384 (ecdh_x25519) - A
|       TLS_AKE_WITH_CHACHA20_POLY1305_SHA256 (ecdh_x25519) - A
|     cipher preference: server
|_  least strength: A
```

**Observed behavior**
```
PORT    STATE SERVICE
443/tcp open  https
[...]

PR #3123, opened 2024-06-21
cmd/skipper: allow exclusion of insecure cipher suites

rickhlx:
Golang maintains a list of cipher suites considered insecure, which are still allowed if requested. This flag will allow those cipher suites to be completely excluded.

Options considered:

**Use a list of allowed cipher suites**
This may need some maintenance over time as cipher suites are updated, introduced or deprecated. 

**Exclude used cipher suites based on name**
Less maintenance overhead than maintaining desired list of cipher suites, excluding the ones not desired would also require some maintenance overtime as cipher suites are considered insecure.

**Exclude known insecure cipher suites** 
Using golang's list of [InsecureCipherSuites](https://pkg.go.dev/crypto/tls#InsecureCipherSuites) reducing maintenance overhead by allowing list to be maintained by golang.

Fixes https://github.com/zalando/skipper/issues/3121

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -219,6 +219,9 @@ type Config struct {
 	// TLS version
 	TLSMinVersion string `yaml:"tls-min-version"`
 
+	// Exclude insecure cipher suites
+	ExcludeInsecureCipherSuites bool `yaml:"exclude-insecure-cipher-suites"`
+
 	// TLS Config
 	KubernetesEnableTLS bool `yaml:"kubernetes-enable-tls"`
 
@@ -517,6 +520,9 @@ func NewConfig() *Config {
 	// TLS version
 	flag.StringVar(&cfg.TLSMinVersion, "tls-min-version", defaultMinTLSVersion, "minimal TLS Version to be used in server, proxy and client connections")
 
+	// Exclude insecure cipher suites
+	flag.BoolVar(&cfg.ExcludeInsecureCipherSuites, "exclude-insecure-cipher-suites", false, "excludes insecure cipher suites")
+
 	// API Monitoring:
 	flag.BoolVar(&cfg.ApiUsageMonitoringEnable, "enable-api-usage-monitoring", false, "enables the apiUsageMonitoring filter")
 	flag.StringVar(&cfg.ApiUsageMonitoringRealmKeys, "api-usage-monitoring-realm-keys", "", "name of the property in the JWT payload that contains the authority realm")
@@ -715,6 +721,7 @@ func (c *Config) ToOptions() skipper.Options {
 		DebugListener:             c.DebugListener,
 		CertPathTLS:               c.CertPathTLS,
 		KeyPathTLS:                c.KeyPathTLS,
+		CipherSuites:              c.filterCipherSuites(),
 		MaxLoopbacks:              c.MaxLoopbacks,
 		DefaultHTTPStatus:         c.DefaultHTTPStatus,
 		ReverseSourcePredicate:    c.ReverseSourcePredicate,
@@ -1029,6 +1036,19 @@ func (c *Config) getMinTLSVersion() uint16 {
 	}
 	log.Infof("No valid minimal TLS version confiured (set to '%s'), fallback to default: %s", c.TLSMinVersion, defaultMinTLSVersion)
 	return tlsVersionTable[defaultMinTLSVersion]
+}
+
+func (c *Config) filterCipherSuites() []uint16 {
+	if !c.ExcludeInsecureCipherSuites {
+		return nil
+	}
+
+	cipherSuites := make([]uint16, 0)
+	for _, suite := range tls.CipherSuites() {
+		cipherSuites = append(cipherSuites, suite.ID)
+	}
+
+	return cipherSuites
 }
 
 func (c *Config) parseHistogramBuckets() ([]float64, error) {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"os"
 	"reflect"
+	"slices"
 	"strings"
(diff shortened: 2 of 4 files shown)


config/config.go
```diff
@@ -219,6 +219,9 @@ type Config struct {
 	// TLS version
 	TLSMinVersion string `yaml:"tls-min-version"`
 
+	// Exclude insecure cipher suites
+	ExcludeInsecureCipherSuites bool `yaml:"exclude-insecure-cipher-suites"`
```
AlexanderYastrebov, line 223: Do we want to exclude from a default set or rather enumerate allowed?

rickhlx: It might be worthy of a discussion. I've updated the description with the options considered to allow for secure cipher suites.

I don't have a strong preference and went with the one requiring less maintenance.

config/config_test.go
```diff
@@ -405,6 +405,23 @@ func TestMinTLSVersion(t *testing.T) {
...
+		}
+	})
+	t.Run("test excluded insecure cipher suites", func(t *testing.T) {
+		cfg := new(Config)
+		cfg.ExcludeInsecureCipherSuites = true
+		if cfg.filterCipherSuites() == nil {
+			t.Error(`Failed to get list of filtered cipher suites`)
+		}
+	})
+}
```
AlexanderYastrebov, line 423: I think its better to test that ToOptions produces a list with excluded insecure suites rather than test unexported method.

config/config_test.go
```diff
@@ -231,6 +232,17 @@ func TestToOptions(t *testing.T) {
...
+	if opt.CipherSuites == nil {
+		t.Errorf("Failed to get the filtered cipher suites")
+	} else {
+		for _, c := range opt.CipherSuites {
+			for _, i := range tls.InsecureCipherSuites() {
+				if c == i.ID {
+					t.Errorf("Insecure cipher found in list: %s", i.Name)
+				}
+			}
+		}
```
AlexanderYastrebov, line 244: minor: maybe use https://pkg.go.dev/slices#Contains
```go
for _, c := range tls.InsecureCipherSuites() {
  if slices.Contains(opt.CipherSuites, i.ID) {
	t.Errorf("Insecure cipher found in list: %s", i.Name)
  }
}
```

config/config.go
szuecs: So if Go stdlib adds more "secure" ciphers than nobody needs to do anything and it will deployed automatically as a side effect.
This is good because likely Go stdlib will include only more secure cipher suites and not less secure ones. Of course these could become broken and you have to know the side effect and exclude these, too.
The bad thing is for people that have regulations like FIPS, which may or may not be up to date with secure cipher suite definitions.

Right now I have no opinion on which one is better, but maybe rather to say we should have an allow list and an exclude list for giving the user a choice based on their needs.

skipper.go
```diff
@@ -614,6 +614,9 @@ type Options struct {
 	// TLSMinVersion to set the minimal TLS version for all TLS configurations
 	TLSMinVersion uint16
 
+	// List of cipher suites to use for TLS 1.2
```
szuecs, line 617: proper godoc starts by the name: `CipherSuites sets the list of cipher suites to use for TLS 1.2`

config/config.go
rickhlx: I'd favor having an optional allow list for that case, since it's likely one would want to be specific about what's allowed.

skipper_test.go
```diff
@@ -186,6 +186,14 @@ func TestOptionsTLSConfig(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, uint16(tls.VersionTLS13), c.MinVersion)
 	require.Equal(t, []tls.Certificate{cert, cert2}, c.Certificates)
+
+	// TLS Cipher Suites
+	o = &Options{CipherSuites: []uint16{1}}
+	c, err = o.tlsConfig(cr)
+	require.NoError(t, err)
+	require.NotEqual(t, len(c.CipherSuites), nil)
```
AlexanderYastrebov, line 194: minor: this is redundant or should be `require.NotNil`.
Also I think it should be `require.NoError` and then `assert.Equal(t, len(c.CipherSuites), 1)` - require preconditions and assert expectations.

config/config.go
```diff
@@ -517,6 +521,10 @@ func NewConfig() *Config {
 	// TLS version
 	flag.StringVar(&cfg.TLSMinVersion, "tls-min-version", defaultMinTLSVersion, "minimal TLS Version to be used in server, proxy and client connections")
 
+	// Cipher Suites
+	flag.BoolVar(&cfg.ExcludeInsecureCipherSuites, "exclude-insecure-cipher-suites", false, "excludes insecure cipher suites")
+	flag.StringVar(&cfg.AllowedCipherSuites, "allowed-cipher-suites", "", "comma separated list of exclusively allowed cipher suites")
```
AlexanderYastrebov, line 526: Please check listFlag and multiFlag types

AlexanderYastrebov: I think we should start with a single boolean flag to disable insecure suites and then add a list flag if we even find a need for it.

config/config.go
AlexanderYastrebov: I think we should start with a single boolean flag to disable insecure suites and then add a list flag if we even find a need for it.

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
:+1:

rickhlx:
@szuecs @AlexanderYastrebov any chance we can get this merged? We're looking to solve some security issues by deploying this change.

MustafaSaber:
👍

MustafaSaber:
Thanks for your contribution!

PR #3126, opened 2024-06-24
filters/auth: add token validator filter

AlexanderYastrebov:
The `oauthTokeninfoValidate` filter obtains token info and allows request if there was no error otherwise it responds with
`401 Unauthorized` status and configured response body.

It does nothing if any preceding filter already validated the token or if route is annotated with configured annotations.

It is useful as a default filter to ensure each request has a valid token.

Since its logic is diffrent from existing oauthTokeninfo* filters it is implemented as a separate filter type.

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -1343,6 +1343,35 @@ Examples:
 oauthTokeninfoAllKV("k1", "v1", "k2", "v2")
 ```
 
+#### oauthTokeninfoValidate
+
+> This filter is experimental and may change in the future, please see tests for example usage.
+
+The filter obtains token info and allows request if there was no error
+otherwise it responds with `401 Unauthorized` status and configured response body.
+
+It does nothing if any preceding filter already validated the token or if route is annotated with configured annotations.
+
+It is useful as a default filter to ensure each request has a valid token.
+[jwtMetrics](#jwtmetrics) filter may be used to discover routes serving requests without a valid token.
+
+The filter requires single string argument that is parsed as YAML.
+For convenience use [flow style format](https://yaml.org/spec/1.2.2/#chapter-7-flow-style-productions).
+
+Examples:
+
+```
+// without opt-out annotation validates the token
+oauthTokeninfoValidate("{optOutAnnotations: [oauth.disabled], unauthorizedResponse: 'Authentication required, see https://auth.test/foo'}")
+```
+
+```
+// with opt-out annotation does not validate the token
+annotate("oauth.disabled", "this endpoint is public") ->
+oauthTokeninfoValidate("{optOutAnnotations: [oauth.disabled], unauthorizedResponse: 'Authentication required, see https://auth.test/foo'}")
+```
+
+
 ### Tokenintrospection
 
 Tokenintrospection handled by another service.
--- a/filters/auth/tokeninfo.go
+++ b/filters/auth/tokeninfo.go
@@ -2,11 +2,16 @@ package auth
 
 import (
 	"fmt"
+	"io"
+	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
+	"github.com/ghodss/yaml"
 	"github.com/opentracing/opentracing-go"
 	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/filters/annotate"
 )
 
 const (
@@ -50,6 +55,14 @@ type (
 		client tokeninfoClient
 		scopes []string
 		kv     kv
+	}
+
+	tokeninfoValidateFilter struct {
+		client tokeninfoClient
+		config struct {
+			OptOutAnnotations    []string `json:"optOutAnnotations,omitempty"`
+			UnauthorizedResponse string   `json:"unauthorizedResponse,omitempty"`
+		}
 	}
 )
 
@@ -150,6 +163,13 @@ func NewOAuthTokeninfoAnyKVWithOptions(to TokeninfoOptions) filters.Spec {
 	}
 }
 
+func NewOAuthTokeninfoValidate(to TokeninfoOptions) filters.Spec {
+	return &tokeninfoSpec{
+		typ:     checkOAuthTokeninfoValidate,
+		options: to,
+	}
+}
+
 // NewOAuthTokeninfoAnyKV creates a new auth filter specification
 // to validate authorization for requests. Current implementation uses
 // Bearer tokens to authorize requests and checks that the token
@@ -192,6 +212,8 @@ func (s *tokeninfoSpec) Name() string {
 		return filters.OAuthTokeninfoAnyKVName
 	case checkOAuthTokeninfoAllKV:
 		return filters.OAuthTokeninfoAllKVName
+	case checkOAuthTokeninfoValidate:
+		return filters.OAuthTokeninfoValidateName
 	}
 	return AuthUnknown
 }
@@ -216,6 +238,18 @@ func (s *tokeninfoSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 	ac, err := s.options.getTokeninfoClient()
 	if err != nil {
 		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	if s.typ == checkOAuthTokeninfoValidate {
+		if len(sargs) != 1 {
+			return nil, fmt.Errorf("requires single string argument")
+		}
+
+		f := &tokeninfoValidateFilter{client: ac}
+		if err := yaml.Unmarshal([]byte(sargs[0]), &f.config); err != nil {
+			return nil, fmt.Errorf("failed to parse configuration")
+		}
+		return f, nil
 	}
 
 	f := &tokeninfoFilter{typ: s.typ, client: ac, kv: make(map[string][]string)}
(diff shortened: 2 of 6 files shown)


filters/auth/tokeninfo.go
```diff
@@ -218,6 +240,18 @@ func (s *tokeninfoSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 		return nil, filters.ErrInvalidFilterParameters
 	}
 
+	if s.typ == checkOAuthTokeninfo {
+		if len(sargs) != 1 {
+			return nil, fmt.Errorf("requires single string argument")
```
szuecs, line 245: maybe add the filter name

filters/auth/tokeninfo.go
```diff
@@ -218,6 +240,18 @@ func (s *tokeninfoSpec) CreateFilter(args []interface{}) (filters.Filter, error)
...
+		}
+
+		f := &tokeninfoValidateFilter{client: ac}
+		if err := yaml.Unmarshal([]byte(sargs[0]), &f.config); err != nil {
+			return nil, fmt.Errorf("failed to parse configuration")
```
szuecs, line 250: same as above use `s.Name()` in error message

filters/auth/tokeninfo.go
```diff
@@ -395,3 +429,39 @@ func (f *tokeninfoFilter) Request(ctx filters.FilterContext) {
...
+		for _, annotation := range f.config.OptOutAnnotations {
+			if _, ok := annotations[annotation]; ok {
+				return // opt-out from validation
+			}
+		}
+	}
+
+	if token, ok := getToken(ctx.Request()); ok {
+		tokeninfo, err := f.client.getTokeninfo(token, ctx)
+		if err == nil { // token is valid
```
szuecs, line 449: hmm rather not obvious condition, also the `ok` check instead of `!ok` is not really obvious by the rule of "errors to the right and ok path to the left."

filters/auth/tokeninfo_test.go
```diff
@@ -575,3 +578,126 @@ func TestOAuthTokeninfoAllocs(t *testing.T) {
...
+	}
+
+	for _, tc := range []struct {
+		name               string
+		precedingFilters   string
+		authHeader         string
+		expectStatus       int
+		expectAuthRequests int32
+	}{
+		// reject
```
szuecs, line 613: these kind of comments will be outdated to soon to make sense, IMO

filters/auth/tokeninfo_test.go
```diff
@@ -575,3 +578,126 @@ func TestOAuthTokeninfoAllocs(t *testing.T) {
...
+			expectAuthRequests: 0,
+		},
+		{
+			name:               "reject missing token when opt-out is invalid",
+			precedingFilters:   `annotate("oauth.invalid", "invalid opt-out")`,
+			authHeader:         "",
+			expectStatus:       http.StatusUnauthorized,
+			expectAuthRequests: 0,
+		},
+		// allow
```
szuecs, line 639: these kind of comments will be outdated to soon to make sense, IMO

filters/auth/tokeninfo.go
AlexanderYastrebov: > maybe add the filter name to the error message

https://github.com/zalando/skipper/pull/1908

filters/auth/tokeninfo.go
AlexanderYastrebov: Please suggest how to rewrite to keep the same logic.

filters/auth/tokeninfo_test.go
AlexanderYastrebov: These were to mark two test case groups as I did not want to add two t.Run, removed.

docs/reference/filters.md
```diff
@@ -1343,6 +1343,31 @@ Examples:
...
+The filter requires single string argument that is parsed as YAML.
+For convenience use [flow style format](https://yaml.org/spec/1.2.2/#chapter-7-flow-style-productions).
+
+Examples:
+
+```
+oauthTokeninfo("{optOutAnnotations: [oauth.disabled], unauthorizedResponse: 'Authentication required, see https://auth.test/foo'}")
+
+// opt-out by annotation
+annotate("oauth.disabled", "this endpoint is public") ->
```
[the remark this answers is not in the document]
AlexanderYastrebov: Please read examples along along with filter description

filters/auth/tokeninfo_test.go
```diff
@@ -575,3 +578,124 @@ func TestOAuthTokeninfoAllocs(t *testing.T) {
...
+			expectStatus:       http.StatusOK,
+			expectAuthRequests: 1, // called once by oauthTokeninfoAllScope
+		},
+		{
+			name:               "allow missing token when opted-out",
+			precedingFilters:   `annotate("oauth.disabled", "this endpoint is public")`,
+			authHeader:         "",
+			expectStatus:       http.StatusOK,
+			expectAuthRequests: 0,
+		},
```
RomanZavodskikh, line 657: Should we test invalid token+opt-out? What is expected behavior in this case?

AlexanderYastrebov: Added, thanks

docs/reference/filters.md
```diff
@@ -1343,6 +1343,31 @@ Examples:
...
+
+It is useful as a default filter to ensure each request has a valid token.
+[jwtMetrics](#jwtmetrics) filter may be used to discover routes serving requests without a valid token.
+
+The filter requires single string argument that is parsed as YAML.
+For convenience use [flow style format](https://yaml.org/spec/1.2.2/#chapter-7-flow-style-productions).
+
+Examples:
+
+```
```
RomanZavodskikh, line 1363: I would just add the comment like "// has no opt-out annotation, does validate the token".

RomanZavodskikh, line 1363: There are two examples out here, maybe it's better to create two separate code blocks for them to separate them.

filters/auth/tokeninfo.go
RomanZavodskikh: As far as I understood Sandor, something like this:
```go
func (f *tokeninfoValidateFilter) serveOnError(ctx filters.FilterContext) {
	ctx.Serve(&http.Response{
		StatusCode: http.StatusUnauthorized,
		Header: http.Header{
			"Content-Length": []string{strconv.Itoa(len(f.config.UnauthorizedResponse))},
		},
		Body: io.NopCloser(strings.NewReader(f.config.UnauthorizedResponse)),
	})
}
```

```go
	token, ok := getToken(ctx.Request())
	if !ok {
		f.serveOnError(ctx)
		return
	}

	tokeninfo, err := f.client.getTokeninfo(token, ctx)
	if err != nil {
		f.serveOnError(ctx)
		return
	}

	uid, _ := tokeninfo[uidKey].(string)
	authorized(ctx, uid)
	ctx.StateBag()[tokeninfoCacheKey] = tokeninfo
```

AlexanderYastrebov: I do not see how it is better, moreover I see an opportunity to forget one of the early returns.

docs/reference/filters.md
AlexanderYastrebov: Fixed

AlexanderYastrebov: Fixed

docs/reference/filters.md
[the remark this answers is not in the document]
AlexanderYastrebov: Split into two blocks and added a comment.

filters/auth/tokeninfo.go
RomanZavodskikh: That's how I see Sandor's suggestion:
```go
func (f *tokeninfoValidateFilter) serveOnError(ctx filters.FilterContext) {
	ctx.Serve(&http.Response{
		StatusCode: http.StatusUnauthorized,
		Header: http.Header{
			"Content-Length": []string{strconv.Itoa(len(f.config.UnauthorizedResponse))},
		},
		Body: io.NopCloser(strings.NewReader(f.config.UnauthorizedResponse)),
	})
}
```

```go
token, ok := getToken(ctx.Request())
	if !ok {
		f.serveOnError(ctx)
		return
	}

	tokeninfo, err := f.client.getTokeninfo(token, ctx)
	if err != nil {
		f.serveOnError(ctx)
		return
	}

	uid, _ := tokeninfo[uidKey].(string)
	authorized(ctx, uid)
	ctx.StateBag()[tokeninfoCacheKey] = tokeninfo
	return
```

RomanZavodskikh:
:+1:

filters/auth/tokeninfo.go
RomanZavodskikh: Okay, I think I'm fine with both options.

MustafaSaber:
👍

docs/reference/filters.md
```diff
@@ -1343,6 +1343,35 @@ Examples:
 oauthTokeninfoAllKV("k1", "v1", "k2", "v2")
 ```
 
+#### oauthTokeninfoValidate
+
+> This filter is experimental and may change in the future, please see tests for example usage.
+
+The filter obtains token info and allows request if there was no error
```
szuecs, line 1350: We used tokeninfo as tool name, so it might make sense to stay with tokeninfo here and not use "token info".

filters/auth/tokeninfo.go
szuecs: yes @RomanZavodskikh code is what I would prefer.

AlexanderYastrebov: I don't think it is better but I changed to the suggested variant.

docs/reference/filters.md
AlexanderYastrebov: Its an info about token which it obtains. We also use "token info" in a few other places.

MustafaSaber:
👍

AlexanderYastrebov:
:+1:

PR #3129, opened 2024-06-26
metrics: register skipper_filter_create_duration_seconds

AlexanderYastrebov:
Follow up on #3115

### DIFF
--- a/metrics/prometheus_test.go
+++ b/metrics/prometheus_test.go
@@ -74,6 +74,29 @@ func TestPrometheusMetrics(t *testing.T) {
 			expCode: http.StatusOK,
 		},
 		{
+			name: "Filter creation latency",
+			addMetrics: func(pm *metrics.Prometheus) {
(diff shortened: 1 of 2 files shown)


metrics/prometheus_test.go
```diff
@@ -73,6 +73,29 @@ func TestPrometheusMetrics(t *testing.T) {
 			},
 			expCode: http.StatusOK,
 		},
+		{
+			name: "Filter creation latency",
+			addMetrics: func(pm *metrics.Prometheus) {
+				pm.MeasureFilterCreate("filter1", time.Now().Add(-15*time.Millisecond))
```
RomanZavodskikh, line 79: Maybe add some second value, -time.Second, for example?

AlexanderYastrebov: It does not matter, we don't test that prometheus works correctly but rather that metric is registered.

MustafaSaber:
👍

RomanZavodskikh:
:+1:

szuecs:
👍

Issue #3130, opened 2024-06-27
Handle `302 Found` Redirects when using the webhook filter


FergusInLondon:
**Is your feature request related to a problem? Please describe.**
The Skipper [`webhook` filter](https://opensource.zalando.com/skipper/reference/filters/#webhook) allows requests to be filtered based on the status code received from an authorisation/authentication endpoint. By default, Skipper will return an empty response with a status code of 401 or 403 if a request is rejected. 

*This behaviour is well-suited to API traffic, but doesn't work for frontend traffic where the user agent should be **redirected** upon rejection.*

**Describe the solution you would like**
When an authentication endpoint returns a HTTP redirect - i.e. a HTTP `302 Found` - then Skipper should return that redirect to the user agent.

These changes would be limited to:

1. Introducing a new conditional to catch responses with a status code of 302 - in [webhook.go#L124-L128](https://github.com/zalando/skipper/blob/master/filters/auth/webhook.go#L124-L128).
2. Introducing a new function - `redirect` - which is the equivalent of the existing `reject` function - in [auth.go#L136-L142](https://github.com/zalando/skipper/blob/master/filters/auth/auth.go#L136-L142) - but copies the `Location` header from the authentication endpoint response.

This should not be a breaking change as currently `302 Found` has no specific meaning to the `webhook` filter: it's caught by the default behaviour which returns `401 Unauthorized`.

**Would you like to work on it?**
Yes, I have a Pull Request prepared - #3131

szuecs:
thanks, reviewed the PR, feature wise I think it's fine but implementation needs to be changed as I wrote in PR comment. Discussion in PR.

AlexanderYastrebov:
> This behaviour is well-suited to API traffic, but doesn't work for frontend traffic where the user agent should be redirected upon rejection.

I can see the point, e.g. if webhook authenticates Cookie and issues redirect for new visitors.

From #3131
> Incidental: Prevent the webhook client from following redirects from the AuthN/AuthZ endpoint: during testing I realised that the default net/http behaviour was in use - i.e. redirects were followed. Is this an intended behaviour of the filter: it's not documented, and it seems potentially risky?

Its not risky assuming operator trusts the webhook server. Following redirects may be used e.g. for migrating webhook url to another domain.


I am not sure I like the idea of forwarding webhook redirect to the client on the other hand there is a usecase.

[...]

PR #3131, opened 2024-06-27
webhook filter: allow client redirects via 302 responses (#3130)

FergusInLondon:
This commit changes the behaviour of the webhook filter when a 302 Found response is received from the AuthN/AuthZ endpoint. As a result, it allows front-end facing (i.e. non-API) traffic to be filtered via the webhook.

Documentation updates and increased test coverage are included.

**Incidental**: Prevent the webhook client from following redirects from the AuthN/AuthZ endpoint: during testing I realised that the default `net/http` behaviour was in use - i.e. redirects were followed. Is this an intended behaviour of the filter: it's not documented, and it seems potentially risky?

Please take a look at #3130 for more details.

---

Changes:

1. Update to the documentation - `filters.md`.
2. Update to the auth `reject()` function, whereby the `Location` header is forwarded to the client if a `302` is encountered from the AuthN/AuthZ endpoint.
3. Prevent the webhook filter from following 302 redirects.
4. Update tests to cover 302 redirect forwarding.

### DIFF
--- a/filters/auth/auth.go
+++ b/filters/auth/auth.go
@@ -105,6 +105,7 @@ func reject(
 	reason rejectReason,
 	hostname,
 	debuginfo string,
+	destination string,
 ) {
 	if debuginfo == "" {
 		ctx.Logger().Debugf(
@@ -125,6 +126,10 @@ func reject(
 		Header:     make(map[string][]string),
 	}
 
+	if status == http.StatusFound && destination != "" {
+		rsp.Header.Add("Location", destination)
+	}
+
 	if hostname != "" {
 		// https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.2
 		rsp.Header.Add("WWW-Authenticate", hostname)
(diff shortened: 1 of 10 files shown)


FergusInLondon:
I'm more than happy to adapt or change the approach if required - especially if there's any reason why the `webhook` filter is expected to follow redirects itself.

Bringing the `webhook` filter closer to traefik's [ForwardAuth](https://doc.traefik.io/traefik/middlewares/http/forwardauth/) functionality - i.e. having more control over the response returned to the client - would be very useful though!

filters/auth/auth.go
```diff
@@ -133,12 +138,16 @@ func reject(
 	ctx.Serve(rsp)
 }
 
+func redirect(ctx filters.FilterContext, username string, reason rejectReason, destination, debuginfo string) {
+	reject(ctx, http.StatusFound, username, reason, "", debuginfo, destination)
```
szuecs, line 142: please provide the body of the function and not re-use `redirect`.
This will add a bit of code, but reduce the size of change.
`reject` should only be used for rejection as 4xx, thanks!

filters/auth/webhook_test.go
```diff
@@ -15,15 +15,17 @@ import (
 )
 
 const headerToCopy = "X-Copy-Header"
+const webhookRedirectLocation = "https://example.com/auth"
```
szuecs, line 18: Please use: `"https://redirect.example/auth"` as by RFC this is in `.example`, so better than in `.com`.

ppoloskov:
Any plans to merge this change?

szuecs:
> Any plans to merge this change? 

Can you share your use case?
In my point of view it has a potential problem but also a useful change. In general who follows the redirect should not be an issue, but behavior change is always a higher risk factor.

ppoloskov:
Sorry for bumping this after such a long time. I want to use skipper for front-end traffic and use authelia for authentication. The original problem described here - https://github.com/zalando/skipper/issues/3329

szuecs:
There are open comments about it. Maybe you can check if you can fix it, such that it is usable to integrate with authelia webhook.

PR #3132, opened 2024-06-27
metrics: refactor prometheus metric registration

AlexanderYastrebov:
Register prometheus metrics right after creation
to reduce likelihood of missing registration.

Follow up on #3129


MustafaSaber:
👍

AlexanderYastrebov:
:+1:

PR #3140, opened 2024-07-02
feature: httpclient before and after hooks

szuecs:
feature: net.Transport got beforeSend and afterResponse hooks that a user can set via net.Options 
```go
	// BeforeSend is a hook function that runs just before executing RoundTrip(*http.Request)
	BeforeSend func(*http.Request)
	// AfterResponse is a hook function that runs just after executing RoundTrip(*http.Request)
	AfterResponse func(*http.Response, error)
```

replaces https://github.com/zalando/skipper/pull/3137

### DIFF
--- a/net/httpclient.go
+++ b/net/httpclient.go
@@ -207,6 +207,11 @@ type Options struct {
 
 	// Log is used for error logging
 	Log logging.Logger
+
+	// BeforeSend is a hook function that runs just before executing RoundTrip(*http.Request)
+	BeforeSend func(*http.Request)
+	// AfterResponse is a hook function that runs just after executing RoundTrip(*http.Request)
+	AfterResponse func(*http.Response, error)
 }
 
 // Transport wraps an http.Transport and adds support for tracing and
@@ -219,6 +224,8 @@ type Transport struct {
 	spanName      string
 	componentName string
 	bearerToken   string
+	beforeSend    func(*http.Request)
+	afterResponse func(*http.Response, error)
 }
 
 // NewTransport creates a wrapped http.Transport, with regular DNS
@@ -275,10 +282,12 @@ func NewTransport(options Options) *Transport {
 	}
 
 	t := &Transport{
-		once:   sync.Once{},
-		quit:   make(chan struct{}),
-		tr:     htransport,
-		tracer: options.Tracer,
+		once:          sync.Once{},
+		quit:          make(chan struct{}),
+		tr:            htransport,
+		tracer:        options.Tracer,
+		beforeSend:    options.BeforeSend,
+		afterResponse: options.AfterResponse,
 	}
 
 	if t.tracer != nil {
@@ -342,6 +351,8 @@ func (t *Transport) shallowCopy() *Transport {
 		spanName:      t.spanName,
 		componentName: t.componentName,
 		bearerToken:   t.bearerToken,
+		beforeSend:    t.beforeSend,
+		afterResponse: t.afterResponse,
 	}
 }
 
@@ -369,7 +380,13 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	if t.bearerToken != "" {
 		req.Header.Set("Authorization", "Bearer "+t.bearerToken)
 	}
+	if t.beforeSend != nil {
+		t.beforeSend(req)
+	}
 	rsp, err := t.tr.RoundTrip(req)
+	if t.afterResponse != nil {
+		t.afterResponse(rsp, err)
+	}
 	if span != nil {
 		span.LogKV("http_do", "stop")
 		if rsp != nil {
@@ -388,10 +405,10 @@ func (t *Transport) injectSpan(req *http.Request) (*http.Request, opentracing.Sp
 		string(ext.HTTPUrl):    req.URL.String(),
 	}}
 	if parentSpan := opentracing.SpanFromContext(req.Context()); parentSpan != nil {
-		req = req.WithContext(opentracing.ContextWithSpan(req.Context(), parentSpan))
 		spanOpts = append(spanOpts, opentracing.ChildOf(parentSpan.Context()))
 	}
 	span := t.tracer.StartSpan(t.spanName, spanOpts...)
+	req = req.WithContext(opentracing.ContextWithSpan(req.Context(), span))
 
 	_ = t.tracer.Inject(span.Context(), opentracing.HTTPHeaders, opentracing.HTTPHeadersCarrier(req.Header))
 
--- a/net/httpclient_example_test.go
+++ b/net/httpclient_example_test.go
@@ -10,10 +10,23 @@ import (
 
 	"github.com/lightstep/lightstep-tracer-go"
 	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
 	"github.com/opentracing/opentracing-go/mocktracer"
+	"github.com/sirupsen/logrus"
 	"github.com/zalando/skipper/net"
 	"github.com/zalando/skipper/secrets"
 )
+
+func waitForSpanViaMockTracer(mockTracer *mocktracer.MockTracer) {
+	for i := 0; i < 20; i++ {
+		if n := len(mockTracer.FinishedSpans()); n > 0 {
+			logrus.Printf("found %d spans", n)
+			return
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+	logrus.Println("no span found")
+}
 
 func ExampleTransport() {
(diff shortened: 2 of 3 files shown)


net/httpclient.go
```diff
@@ -207,6 +207,11 @@ type Options struct {
 
 	// Log is used for error logging
 	Log logging.Logger
+
+	// BeforeSend is a hook function that runs just before executing RoundTrip(*http.Request)
+	BeforeSend func(opentracing.Span, *http.Request)
+	// AfterResponse is a hook function that runs just after executing RoundTrip(*http.Request)
+	AfterResponse func(opentracing.Span, *http.Response, error)
```
AlexanderYastrebov, line 214: Request context already contains the span so there is no need to pass it as an argument.
It also ties API to opentracing. To pass something else than span along the request we would need to put it into requets context anyways.

The same thing could be implemented by wrapping http.Transport into custom roundTripper and passing it in the Options (we may need to change Options.Transport to Options.RoundTripper (or add Options.RoundTripper))

szuecs: > Request context already contains the span so there is no need to pass it as an argument. It also ties API to opentracing. To pass something else than span along the request we would need to put it into requets context anyways.

1. http.Response has no access to the Span, so it makes IMO more sense to have it analog in Before / After case.
2. simplicity for the user
3. why should we pack/unpack twice?

I agree that opentracing.Span is not ideal, but your choice was to stay with opentracing and not to move to OTel... 

> The same thing could be implemented by wrapping http.Transport into custom roundTripper and passing it in the Options (we may need to change Options.Transport to Options.RoundTripper (or add Options.RoundTripper))

Of course you can always wrap it but this should be the rare case and not the general case. The idea was to have a client that has all these batteries included and fix all stdlib bugs we know which would take long to fix in stdlib.

AlexanderYastrebov: > http.Response has no access to the Span

Response holds a link to the request, You can also pass request+response to AfterResponse

net/httpclient_example_test.go
```diff
@@ -325,3 +326,74 @@ func ExampleClient_hostSecret() {
...
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer srv.Close()
+
+	rsp, err := cli.Get("http://" + srv.Listener.Addr().String() + "/")
+	if err != nil {
+		log.Fatalf("Failed to get: %v", err)
+	}
+	fmt.Printf("response code: %d\n", rsp.StatusCode)
+	fmt.Printf("span status.code: %d", mockTracer.FinishedSpans()[0].Tags()["status.code"])
```
AlexanderYastrebov, line 394: Client may return before span is available in finished spans.
To mitigate you'd need to wait for it, see example https://github.com/zalando/skipper/commit/34b188cab70fe01bed88b6a9a16db7a34559a70d#diff-a12e7212c817eb6b6e6168aaeeb1e1faefd34de890a929adc152505bf34bf097R268-R269

net/httpclient.go
szuecs: this was done

net/httpclient_example_test.go
szuecs: Because we have no testing.T, I will create a loop that waits for some time similar to the linked code.

szuecs: I added some debug logs to figure out what's going on, not sure how this can actually happen:
```
% go test ./net -run ExampleClient_withBeforeSendHook
2024/07/03 17:28:31 injectSpan
time="2024-07-03T17:28:31+02:00" level=info msg="BeforeSend: no span found"
time="2024-07-03T17:28:31+02:00" level=info msg="found 1 spans"
--- FAIL: ExampleClient_withBeforeSendHook (0.00s)
got:
X-Foo: qux
request tag "peer.service" set to %!q(<nil>)
want:
X-Foo: qux
request tag "peer.service" set to "my-peer-service"
FAIL
FAIL    github.com/zalando/skipper/net  0.099s
FAIL
```

net/httpclient.go
```diff
@@ -381,6 +399,7 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 }
 
 func (t *Transport) injectSpan(req *http.Request) (*http.Request, opentracing.Span) {
+	log.Println("injectSpan")
```
AlexanderYastrebov, line 402: injectSpan seems to be broken - it updates request context only when there is a parent span.
I think  it should be:
```diff
diff --git a/net/httpclient.go b/net/httpclient.go
index 2a4eed1a..db944f32 100644
--- a/net/httpclient.go
+++ b/net/httpclient.go
@@ -407,10 +407,10 @@ func (t *Transport) injectSpan(req *http.Request) (*http.Request, opentracing.Sp
                string(ext.HTTPUrl):    req.URL.String(),
        }}
        if parentSpan := opentracing.SpanFromContext(req.Context()); parentSpan != nil {
-               req = req.WithContext(opentracing.ContextWithSpan(req.Context(), parentSpan))
                spanOpts = append(spanOpts, opentracing.ChildOf(parentSpan.Context()))
        }
        span := t.tracer.StartSpan(t.spanName, spanOpts...)
+       req = req.WithContext(opentracing.ContextWithSpan(req.Context(), span))
 
        _ = t.tracer.Inject(span.Context(), opentracing.HTTPHeaders, opentracing.HTTPHeadersCarrier(req.Header))
```

szuecs: Oh yeah that makes sense!
Only one case is now failing

net/httpclient_test.go
```diff
@@ -248,6 +255,89 @@ func TestTransport(t *testing.T) {
...
+					}
+					println("AfterResponse code not executed")
+				},
+			},
+			spanName: "myspan",
+			req:      httptest.NewRequest("GET", "http://example.com/", nil),
+			wantErr:  false,
+			checkRequest: func(req *http.Request) error {
+				fmt.Printf("1. checkRequest req: %p\n", req)
+				if span := opentracing.SpanFromContext(req.Context()); span != nil {
```
szuecs, line 321: span is nil at the server and client side. Test log:
```
% go test ./net -run "TestTransport/With_hooks_and_opentracing,_should_have_request_header_and_response_changed" -count 1 -v
=== RUN   TestTransport
DO NOT USE IN PRODUCTION
=== RUN   TestTransport/With_hooks_and_opentracing,_should_have_request_header_and_response_changed
2024/07/10 13:01:20 injectSpan
1. req: 0xc0000f5c20
2. req: 0xc0000f5c20
before roundtrip req: 0xc0000f5c20
1. checkRequest req: 0xc00015e000
    httpclient_test.go:368: Failed to check request: failed get span from request
after roundtrip: 0xc0000f5c20
1. rsp.request: 0xc0000f5c20
2. rsp.request: 0xc0000f5c20
after roundtrip on the client
1. checkRequest req: 0xc0000f5200
    httpclient_test.go:396: Failed to check request: failed get span from request
--- FAIL: TestTransport (0.00s)
    --- FAIL: TestTransport/With_hooks_and_opentracing,_should_have_request_header_and_response_changed (0.00s)
FAIL
FAIL    github.com/zalando/skipper/net  0.180s
FAIL
zsh: exit 1     go test ./net -run  -count 1 -v
```
see that checkRequest() sees always a different pointer for the modified request than everyone else.

szuecs: Yes but it should!?
Also the last log just before the 2nd error `1. checkRequest req: 0xc0000f5200` on the client, that use `rsp.Request` is not the same as the request that we sent.

AlexanderYastrebov: Span comes from the context and I think there is nothing in test server that adds a span to the context.

net/httpclient_example_test.go
```diff
@@ -325,3 +341,83 @@ func ExampleClient_hostSecret() {
...
+		},
+	})
+	defer cli.Close()
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer srv.Close()
+
+	rsp, err := cli.Get("http://" + srv.Listener.Addr().String() + "/")
```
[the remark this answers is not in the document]
szuecs: fixed

net/httpclient_test.go
szuecs: fixed test by running on the client after roundtrip

net/httpclient_example_test.go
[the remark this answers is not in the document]
szuecs: I tested to change it but it gives a nil ptr deref panic

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #3142, opened 2024-07-04
docs: running as ingress backend

szuecs:
you can also use the newer lifecycle hook target "sleep" to let kubelet sleep for you, so no need to have a binary inside your container


AlexanderYastrebov:
:+1:

MustafaSaber:
👍

PR #3041, opened 2024-04-25
Facilitate OPA decision correlation with business flows

JanardhanSharma:
[Issue](https://github.bus.zalan.do/corporate-iam/issues/issues/595)

## Context
Each evaluation yields a distinct decision, identifiable by its unique decision ID. This decision ID can be located within the input at input.attributes.metadataContext.filterMetadata.open_policy_agent.decision_id. Leveraging this ID enables seamless correlation of decisions with the flow ID of the originating request.

## Changes
Decision id is added to the request object and is available to be used in the policies.

PR #3157, opened 2024-07-19
Facilitate opa decision correlation with business flows iteration 2

JanardhanSharma:
## Context
The release of the feature Facilitate OPA decision correlation with business flows ([PR](https://github.com/zalando/skipper/pull/3041/files)) let to the a [bug](https://github.bus.zalan.do/pg9/issues/issues/624) that leads to the skipper restarts. This is because an object of same reference is being shared among the different instances of opa instances. 

## ACs
 Fix the opaauhorizerequest filter restart issue.


JanardhanSharma:
@AlexanderYastrebov  Have a look once.  I wonder why though now I have so many conflicts.

JanardhanSharma:
Not needed, will raise a clean one.

PR #3164, opened 2024-07-24
filters/auth: add host opt-out to jwtMetrics and oauthTokeninfoValidate

AlexanderYastrebov:
Extend configuration of `jwtMetrics` and `oauthTokeninfoValidate` to support opt-out by request host pattern - disable metrics collection and validation when request host matches any of the configured opt-out regular expressions.

This can be used to exclude internal cluster domain (*.ingress.cluster.local hosts).


filters/auth/jwt_metrics_test.go
```diff
@@ -223,7 +225,44 @@ func TestJwtMetrics(t *testing.T) {
...
+			status: http.StatusOK,
+			expected: map[string]int64{
+				"jwtMetrics.custom.GET.foo_test.200.invalid-token": 1,
+			},
+			expectedTag: "invalid-token",
+		},
+		{
+			name:     "no metrics when host matches opted-out domain",
+			filters:  `jwtMetrics("{issuers: [foo, bar], optOutHosts: [ '^.+[.]domain[.]test$', '^exact[.]test$' ]}")`,
+			request:  &http.Request{Method: "GET", Host: "foo.domain.test"},
```
RomanZavodskikh, line 240: Should we test the same filter, but with "domain.test" host?

filters/auth/jwt_metrics_test.go
```diff
@@ -223,7 +225,44 @@ func TestJwtMetrics(t *testing.T) {
...
+			name:     "no metrics when second level host matches opted-out domain",
+			filters:  `jwtMetrics("{issuers: [foo, bar], optOutHosts: [ '^.+[.]domain[.]test$', '^exact[.]test$' ]}")`,
+			request:  &http.Request{Method: "GET", Host: "foo.bar.domain.test"},
+			status:   http.StatusOK,
+			expected: map[string]int64{},
+		},
+		{
+			name:     "no metrics when host matches opted-out host exactly",
+			filters:  `jwtMetrics("{issuers: [foo, bar], optOutHosts: [ '^.+[.]domain[.]test$', '^exact[.]test$' ]}")`,
+			request:  &http.Request{Method: "GET", Host: "exact.test"},
```
RomanZavodskikh, line 254: Should we test the same filter, but with "not.exact.test" and/or "notexact.test" host?

filters/auth/tokeninfo_test.go
```diff
@@ -669,16 +676,35 @@ func TestOAuthTokeninfoValidate(t *testing.T) {
...
+			expectStatus:       http.StatusOK,
+			expectAuthRequests: 0,
+		},
+		{
+			name:               "allow missing token when request host matches opted-out host exactly",
+			host:               "exact.test",
+			authHeader:         "",
+			expectStatus:       http.StatusOK,
+			expectAuthRequests: 0,
+		},
```
RomanZavodskikh, line 699: Should we test "domain.test" and "not.exact.test" and/or "notexact.test" here as well?

filters/auth/tokeninfo_test.go
```diff
@@ -713,3 +743,45 @@ func TestOAuthTokeninfoValidate(t *testing.T) {
...
+	tio := TokeninfoOptions{
+		URL:     "https://auth.test",
+		Timeout: testAuthTimeout,
+	}
+	spec := NewOAuthTokeninfoValidate(tio)
+
+	t.Run("valid", func(t *testing.T) {
+		for _, def := range []string{
+			`oauthTokeninfoValidate("{}")`,
+			`oauthTokeninfoValidate("{optOutAnnotations: [oauth.disabled], optOutHosts: [ '^.+[.]domain[.]test$', '^exact[.]test$' ], unauthorizedResponse: 'Authentication required, see https://auth.test/foo'}")`,
```
RomanZavodskikh, line 757: I would've added oauthTokeninfoValidate fiters with single arguments as well, like this:
+ `oauthTokeninfoValidate("{optOutAnnotations: [oauth.disabled], optOutHosts: [ '^.+[.]domain[.]test$', '^exact[.]test$' ]}")`
+ `oauthTokeninfoValidate("{unauthorizedResponse: 'Authentication required, see https://auth.test/foo'}")`

filters/auth/jwt_metrics_test.go
AlexanderYastrebov: There is "counts invalid-token when host does not match" test case for that.

filters/auth/jwt_metrics_test.go
AlexanderYastrebov: We can but the goal is not to test regexp itself or all edge cases but the logic when host matches any or none of the configured patterns.

filters/auth/tokeninfo_test.go
AlexanderYastrebov: This quickly goes out of hands but I added.

RomanZavodskikh:
:+1:

filters/auth/tokeninfo_test.go
```diff
@@ -713,3 +743,47 @@ func TestOAuthTokeninfoValidate(t *testing.T) {
...
+			`oauthTokeninfoValidate(1)`,
+			`oauthTokeninfoValidate("{optOutAnnotations: [oauth.disabled]}", "extra arg")`,
+			`oauthTokeninfoValidate("{optOutHosts: [ '[' ]}")`, // invalid regexp
+		} {
+			t.Run(def, func(t *testing.T) {
+				f := eskip.MustParseFilters(def)[0]
+				require.Equal(t, spec.Name(), f.Name)
+
+				_, err := spec.CreateFilter(f.Args)
+				t.Logf("%v", err)
```
szuecs, line 784: does not assert.Error show the same thing?

AlexanderYastrebov: I think assert.Error does not log the error, only checks that is not nil

PR #3192, opened 2024-08-15
proxy: handle invalid chunked requests

AlexanderYastrebov:
Respond with 400 instead of 500 in case of invalid chunked request body.


szuecs:
:+1:

PR #3193, opened 2024-08-16
fix: test requires Go version 1.22

szuecs:
fix: test requires Go version 1.22


szuecs:
ClusterFuzzLite fails because it can not use 1.22 but let's fix this later.

szuecs:
:+1:

MustafaSaber:
👍

PR #3201, opened 2024-08-20
filters/auth: handle trailing question mark in grant flow

AlexanderYastrebov:
Do not propagate trailing question mark from the initial request to redirect URI.

Construct redirect url from scratch instead of resetting URL.ForceQuery.

See related https://github.com/golang/go/issues/13488

### DIFF
--- a/filters/auth/grantconfig.go
+++ b/filters/auth/grantconfig.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
 	"path/filepath"
 	"strings"
 	"time"
@@ -370,10 +371,11 @@ func (c *OAuthConfig) RedirectURLs(req *http.Request) (redirect, original string
 
 	original = u.String()
 
-	u.Path = c.CallbackPath
-	u.RawQuery = ""
-
-	redirect = u.String()
+	redirect = (&url.URL{
+		Scheme: u.Scheme,
+		Host:   u.Host,
+		Path:   c.CallbackPath,
+	}).String()
 
 	return
 }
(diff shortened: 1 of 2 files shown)


filters/auth/grantconfig.go
```diff
@@ -371,7 +371,9 @@ func (c *OAuthConfig) RedirectURLs(req *http.Request) (redirect, original string
 	original = u.String()
 
 	u.Path = c.CallbackPath
+	u.RawPath = ""
```
AlexanderYastrebov, line 374: `RawPath` is a hint how to encode `Path` (see https://pkg.go.dev/net/url#URL.EscapedPath), since we change `Path`, reset `RawPath` as well.

filters/auth/grantconfig.go
```diff
@@ -371,7 +371,9 @@ func (c *OAuthConfig) RedirectURLs(req *http.Request) (redirect, original string
...
 	u.RawQuery = ""
+	u.ForceQuery = false
```
AlexanderYastrebov, line 376: This is the fix, see https://github.com/golang/go/issues/13488 for details.

filters/auth/grantconfig.go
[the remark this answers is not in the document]
MustafaSaber: the comment is old I see you have another one to explain

szuecs: maybe we should have these comments as comments in code?

filters/auth/grantconfig.go
szuecs: maybe we should have these comments as comments in code?

filters/auth/grantconfig.go
AlexanderYastrebov: I don't like that we reused original url object to construct redirect url so I rewrote it to be constructed from scratch. This was we don't need to reset any fields.

filters/auth/grantconfig.go
AlexanderYastrebov: https://github.com/zalando/skipper/pull/3201#discussion_r1723728995

MustafaSaber:
👍

PR #3203, opened 2024-08-22
options: Add RegisterFilters callback

AlexanderYastrebov:
Add a callback to register additional filters that can also re-use already registered filters.


MustafaSaber:
👍

szuecs:
:+1:

PR #3207, opened 2024-08-26
opa: pass URL query parameters to OPA policy evaluation

mefarazath:
Improve the envoy request adapting logic to include query parameters sent in the request.

This would allow the policy evaluation `opaAuthorizeRequest*` and `opaServeResponse*` filters to make use of query parameters/values in the policy.

### DIFF
--- a/filters/openpolicyagent/internal/envoy/skipperadapter.go
+++ b/filters/openpolicyagent/internal/envoy/skipperadapter.go
@@ -3,6 +3,7 @@ package envoy
 import (
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 	"unicode/utf8"
 
@@ -11,9 +12,8 @@ import (
 )
 
 func AdaptToExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metadata, contextExtensions map[string]string, rawBody []byte) (*ext_authz_v3.CheckRequest, error) {
-
-	if !utf8.ValidString(req.URL.Path) {
-		return nil, fmt.Errorf("invalid utf8 in path: %q", req.URL.Path)
+	if err := validateURLForInvalidUTF8(req.URL); err != nil {
+		return nil, fmt.Errorf("invalid url: %w", err)
 	}
 
 	headers := make(map[string]string, len(req.Header))
@@ -29,7 +29,7 @@ func AdaptToExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metada
 				Http: &ext_authz_v3.AttributeContext_HttpRequest{
 					Host:    req.Host,
 					Method:  req.Method,
-					Path:    req.URL.RequestURI(),
+					Path:    buildPathWithQueryParams(req),
 					Headers: headers,
 					RawBody: rawBody,
 				},
@@ -40,4 +40,28 @@ func AdaptToExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metada
 	}
 
 	return ereq, nil
+}
+
+func buildPathWithQueryParams(req *http.Request) string {
+	if req.URL.RawQuery == "" {
+		return url.PathEscape(req.URL.Path)
+	}
+	return url.PathEscape(req.URL.Path) + "?" + req.URL.RawQuery
+}
+
+func validateURLForInvalidUTF8(u *url.URL) error {
+	if !utf8.ValidString(u.Path) {
+		return fmt.Errorf("invalid utf8 in path: %q", u.Path)
+	}
+
+	decodedQuery, err := url.QueryUnescape(u.RawQuery)
+	if err != nil {
+		return fmt.Errorf("error decoding query string: %w", err)
+	}
+
+	if !utf8.ValidString(decodedQuery) {
+		return fmt.Errorf("invalid utf8 in query: %q", u.RawQuery)
+	}
+
+	return nil
 }
--- a/filters/openpolicyagent/internal/envoy/skipperadapter.go
+++ b/filters/openpolicyagent/internal/envoy/skipperadapter.go
@@ -29,7 +29,7 @@ func AdaptToExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metada
 				Http: &ext_authz_v3.AttributeContext_HttpRequest{
 					Host:    req.Host,
 					Method:  req.Method,
-					Path:    buildPathWithQueryParams(req),
+					Path:    req.URL.RequestURI(),
 					Headers: headers,
 					RawBody: rawBody,
 				},
@@ -40,13 +40,6 @@ func AdaptToExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metada
 	}
 
 	return ereq, nil
-}
-
-func buildPathWithQueryParams(req *http.Request) string {
-	if req.URL.RawQuery == "" {
-		return url.PathEscape(req.URL.Path)
-	}
-	return url.PathEscape(req.URL.Path) + "?" + req.URL.RawQuery
 }
 
 func validateURLForInvalidUTF8(u *url.URL) error {
--- a/filters/openpolicyagent/internal/envoy/skipperadapter.go
+++ b/filters/openpolicyagent/internal/envoy/skipperadapter.go
@@ -56,7 +56,7 @@ func validateURLForInvalidUTF8(u *url.URL) error {
 
 	decodedQuery, err := url.QueryUnescape(u.RawQuery)
 	if err != nil {
-		return fmt.Errorf("error decoding query string: %w", err)
+		return fmt.Errorf("error unescaping query string %q: %w", u.RawQuery, err)
 	}
 
 	if !utf8.ValidString(decodedQuery) {
--- a/filters/openpolicyagent/internal/envoy/skipperadapter.go
+++ b/filters/openpolicyagent/internal/envoy/skipperadapter.go
@@ -12,6 +12,10 @@ import (
 
 func AdaptToExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metadata, contextExtensions map[string]string, rawBody []byte) (*ext_authz_v3.CheckRequest, error) {
 
+	if !utf8.ValidString(req.URL.Path) {
+		return nil, fmt.Errorf("invalid utf8 in path: %q", req.URL.Path)
+	}
+
 	headers := make(map[string]string, len(req.Header))
 	for h, vv := range req.Header {
 		// This makes headers in the input compatible with what Envoy does, i.e. allows to use policy fragments designed for envoy
@@ -25,7 +29,7 @@ func AdaptToExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metada
 				Http: &ext_authz_v3.AttributeContext_HttpRequest{
 					Host:    req.Host,
 					Method:  req.Method,
-					Path:    req.URL.Path,
+					Path:    req.URL.RequestURI(),
 					Headers: headers,
 					RawBody: rawBody,
 				},
@@ -33,10 +37,6 @@ func AdaptToExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metada
 			ContextExtensions: contextExtensions,
 			MetadataContext:   metadata,
 		},
-	}
-
-	if !utf8.ValidString(ereq.Attributes.Request.Http.Path) {
-		return nil, fmt.Errorf("invalid utf8 in path: %q", ereq.Attributes.Request.Http.Path)
 	}
 
 	return ereq, nil
--- a/filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
+++ b/filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
@@ -58,11 +58,25 @@ func TestAuthorizeRequestFilter(t *testing.T) {
 			removeHeaders:     make(http.Header),
 		},
 		{
+			msg:               "Allow Requests with trailing ? in URL",
+			filterName:        "opaAuthorizeRequest",
+			bundleName:        "somebundle.tar.gz",
+			regoQuery:         "envoy/authz/allow",
+			requestPath:       "/allow?",
+			requestMethod:     "GET",
+			contextExtensions: "",
+			expectedStatus:    http.StatusOK,
+			expectedBody:      "Welcome!",
+			expectedHeaders:   make(http.Header),
+			backendHeaders:    make(http.Header),
+			removeHeaders:     make(http.Header),
+		},
+		{
 			msg:               "Allow Requests with query parameters",
 			filterName:        "opaAuthorizeRequest",
 			bundleName:        "somebundle.tar.gz",
 			regoQuery:         "envoy/authz/allow",
-			requestPath:       "/allow-with-query?pass=yes&q2=v2",
+			requestPath:       "/allow-with-query?pass=yes&id=1&id=2",
 			requestMethod:     "GET",
 			contextExtensions: "",
 			expectedStatus:    http.StatusOK,
@@ -341,6 +355,7 @@ func TestAuthorizeRequestFilter(t *testing.T) {
 						allow {
 							input.parsed_path = [ "allow-with-query" ]
 							input.parsed_query.pass == ["yes"]
+							input.parsed_query.id == ["1", "2"]
 						}
 
 						allow_context_extensions {
--- a/filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
+++ b/filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
@@ -93,7 +93,7 @@ func TestAuthorizeRequestFilter(t *testing.T) {
 			filterName:        "opaAuthorizeRequest",
 			bundleName:        "somebundle.tar.gz",
 			regoQuery:         "envoy/authz/allow",
-			requestPath:       "/allow-with-query?pass=yes&id=1&id=2",
+			requestPath:       "/allow-with-query?pass=yes&id=1&id=2&msg=help%20me",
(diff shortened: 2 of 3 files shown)


szuecs:
:+1:

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
```diff
@@ -57,6 +57,20 @@ func TestAuthorizeRequestFilter(t *testing.T) {
 			backendHeaders:    make(http.Header),
 			removeHeaders:     make(http.Header),
 		},
+		{
+			msg:               "Allow Requests with query parameters",
+			filterName:        "opaAuthorizeRequest",
+			bundleName:        "somebundle.tar.gz",
+			regoQuery:         "envoy/authz/allow",
+			requestPath:       "/allow-with-query?pass=yes&q2=v2",
```
Pushpalanka, line 65: Shall we add a test case to cover, `/allow-with-query?q2=yes&q2=v2` case as well as for the behavior mentioned [here](https://www.openpolicyagent.org/docs/latest/envoy-primer/#example-input).

`/people?lang=en&id=1&id=2` should give `input.parsed_query.id == ["1", "2"]`

MustafaSaber: recently we found a good special case, https://github.com/zalando/skipper/pull/3201 a testcase can also be good for this to see how it will be handled by opa`/allow-with-query?`

mefarazath: > recently we found a good special case, #3201 a testcase can also be good for this to see how it will be handled by opa`/allow-with-query?`

Do you mean the test case for `/allow-with-query?` or `?` trailing in the query params?

mefarazath: > Shall we add a test case to cover, `/allow-with-query?q2=yes&q2=v2` case as well as for the behavior mentioned [here](https://www.openpolicyagent.org/docs/latest/envoy-primer/#example-input) when same param ID has multiple values.
> 
> `/people?lang=en&id=1&id=2` should give `input.parsed_query.id == ["1", "2"]`

Modifed the same test case to include both scenarios

MustafaSaber: > Do you mean the test case for `/allow-with-query?` or `?` trailing in the query params?

yes, I meant the trailing `?` with no query parameters.

mefarazath: > > Do you mean the test case for `/allow-with-query?` or `?` trailing in the query params?
> 
> yes, I meant the trailing `?` with no query parameters.

I assumed this was the case and added a test case :)

mefarazath:
@szuecs I added a few more test cases based on review comments, please take a look approve if everything looks fine.

filters/openpolicyagent/internal/envoy/skipperadapter.go
```diff
@@ -12,6 +12,10 @@ import (
 
 func AdaptToExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metadata, contextExtensions map[string]string, rawBody []byte) (*ext_authz_v3.CheckRequest, error) {
 
+	if !utf8.ValidString(req.URL.Path) {
```
mjungsbluth, line 15: This maintains the current behaviour, but shouldn't this check the `RequestURI()` as well?

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
```diff
@@ -57,6 +57,34 @@ func TestAuthorizeRequestFilter(t *testing.T) {
 			backendHeaders:    make(http.Header),
 			removeHeaders:     make(http.Header),
 		},
+		{
+			msg:               "Allow Requests with trailing ? in URL",
+			filterName:        "opaAuthorizeRequest",
+			bundleName:        "somebundle.tar.gz",
+			regoQuery:         "envoy/authz/allow",
+			requestPath:       "/allow?",
```
mjungsbluth, line 65: Shouldn't this be `allow-with-query?` ?

mjungsbluth: Or rather, I don't this this test is super meaningful as you can also pass `/allow?foo=bar` and it will pass as query parameters are ignored in Rego.

MustafaSaber: > You might want to validate the that path as visible in Reno does not actually have the ? in it, which I think is what you want to make sure of.

yes, that was my idea when I suggested this test case.

filters/openpolicyagent/internal/envoy/skipperadapter.go
mefarazath: `RequestURI()` returns the encoded path, which bypasses our validation.
`req.URL.Path` returns the decoded path. Therefore I decided to use it to preserve the current behaviour.

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
mefarazath: > Or rather, I don't think this test is super meaningful as you can also pass `/allow?foo=bar` and it will pass as query parameters are ignored in Rego. You might want to validate the that path as visible in Reno does not actually have the `?` in it, which I think is what you want to make sure of...

My thinking was that if the client requests `/allow?` (without queries) and the policy only allows only path `/allow` (and does not take into account query params), the test would assert that path passed to rego does not have the consider trailing `?`

Won't this cover what we want to check?

The other possible test case i thought could cover the scenario explicitly was,
```
		{
			msg:               "OPA does not consider trailing ? in URL",
			filterName:        "opaAuthorizeRequest",
			bundleName:        "somebundle.tar.gz",
			regoQuery:         "envoy/authz/allow",
			requestPath:       "/ends-with?",
			requestMethod:     "GET",
			contextExtensions: "",
			expectedStatus:    http.StatusForbidden,
			expectedHeaders:   make(http.Header),
		},
		
		allow {
			input.parsed_path = [ "ends-with?" ]
		}
```

MustafaSaber: > and the policy only allows only path /allow (and does not take into account query params)

I think I made a confusion (sorry about that). My idea was to test with the trailing `?` (with empty query params) to see and document the behaviour of rego with this type of paths.

I was trying to answer this questions
* will the path be `/allow?` or `/allow`?
* If for some reason the rego role allow any query param will this also be a valid query param or not?

filters/openpolicyagent/internal/envoy/skipperadapter.go
mjungsbluth: Ok, that makes sense but we will miss invalid UTF8 in the query the way it is written now.

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
mjungsbluth: @mefarazath I think the added test case makes sense. I'd also add a condition in Rego to check that the `parsed_query` is empty. This will at least document the current behaviour.

I am curious to see if this looks like this in Rego
`input.parsed_path == [ "ends-with?" ]`
or
`input.parsed_path == [ "ends-with" ]`

It should actually be the latter, stripping the question mark.

filters/openpolicyagent/internal/envoy/skipperadapter.go
szuecs: RequestURI() is likely enforced by Go stdlib net/http to be compliant with RFC, so we will not have the issue there.

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
szuecs: `?` should not be part of the "path".

filters/openpolicyagent/internal/envoy/skipperadapter.go
```diff
@@ -25,7 +29,7 @@ func AdaptToExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metada
 				Http: &ext_authz_v3.AttributeContext_HttpRequest{
 					Host:    req.Host,
 					Method:  req.Method,
-					Path:    req.URL.Path,
+					Path:    req.URL.RequestURI(),
```
szuecs, line 32: if you have filter chain:
```
setPath("/foo") -> opaAuthorizeRequest("...")
```

The request will be to http://.../my-path

Would you expect /my-path or /foo that is checked by the opa filter?

RequestURI will be /my-path

mefarazath: Assuming that `req.URL.Path` would give /foo, the current behaviour would be that OPA filter/policy would see /foo as the path. So this change of using RequestURI would break that behaviour.

probably using below would be a way to retain the current behaviour + pass the query params
```
Path:    req.URL.Path + "?" + req.URL.RawQuery
```

szuecs: I don't know what is expected. From my side we can also document the behavior that req.URL.RequestURI() (link godoc stdlib) is used and would be fine for me.

mefarazath: i felt if we use the `req.URL.RequestURI()` just to get the query params in the envoy check request, it will make the OPA filter not usable with setPath() and the behaviour will be a bit confusing for a user.

Decide to use a different approach to build the path with the query params.

filters/openpolicyagent/internal/envoy/skipperadapter.go
```diff
@@ -35,9 +39,29 @@ func AdaptToExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metada
...
-	if !utf8.ValidString(ereq.Attributes.Request.Http.Path) {
-		return nil, fmt.Errorf("invalid utf8 in path: %q", ereq.Attributes.Request.Http.Path)
+	return ereq, nil
+}
+
+func buildPathWithQueryParams(req *http.Request) string {
+	if req.URL.RawQuery == "" {
+		return url.PathEscape(req.URL.Path)
 	}
+	return url.PathEscape(req.URL.Path) + "?" + req.URL.RawQuery
```
szuecs, line 49: do we need to use `url.QueryEscape`?

filters/openpolicyagent/internal/envoy/skipperadapter.go
```diff
@@ -35,9 +39,29 @@ func AdaptToExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metada
...
 
-	return ereq, nil
+func validateURLForInvalidUTF8(u *url.URL) error {
+	if !utf8.ValidString(u.Path) {
+		return fmt.Errorf("invalid utf8 in path: %q", u.Path)
+	}
+
+	decodedQuery, err := url.QueryUnescape(u.RawQuery)
+	if err != nil {
+		return fmt.Errorf("error decoding query string: %w", err)
```
szuecs, line 59: maybe also add %q and u.RawQuery similar to the error above

szuecs:
2 small things.
Can you resolve comments that are "fixed"? Thanks

filters/openpolicyagent/internal/envoy/skipperadapter.go
mefarazath: [req.URL.RawQuery](https://cs.opensource.google/go/go/+/refs/tags/go1.23.1:src/net/url/url.go;l=379) is already encoded, so didn't have to encode it again.

szuecs: Are you sure you want to have this result: https://go.dev/play/p/AOiC7-DQEBs ?

mefarazath: > Are you sure you want to have this result: https://go.dev/play/p/AOiC7-DQEBs ? path and the query (query does not look encoded to me, but maybe at http.Server it is encoded)

To be honest, I am not entirely sure about how Go handles the parsing when there is a UTF invalid character passed in the URL string. (IINM, to be a valid URL, non-ascii characters needs to be percentage encoded)

Prior to processing the URL, i think in skipper we check for UTF invalid characters. The [check](https://github.com/zalando/skipper/pull/3207/files#diff-dee220a324e25f9867826168abe29e559c7291d044fe87d44165c3ef28384c5cR15) was added to do the same to return a 400 for a URLs with invalid UTF characters.

To test the behaviour of query string, I tried an example, where there is a space in the query param value
https://go.dev/play/p/Z1qeKvyfMre

This shows that the req.URL.RawQuery returns the unescaped query string.

szuecs: And the PathEscape() is by intention, right?
So `/` is `%2F`

mefarazath: Yes, it was intentional.

When OPA creates the input to policy it unescapes the path and query. So I didn't see a problem with this. 
The reason to PathEscape() was to avoid a double escape/decode.

filters/openpolicyagent/internal/envoy/skipperadapter.go
```diff
@@ -25,7 +29,7 @@ func AdaptToExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metada
 				Http: &ext_authz_v3.AttributeContext_HttpRequest{
 					Host:    req.Host,
 					Method:  req.Method,
-					Path:    req.URL.Path,
+					Path:    buildPathWithQueryParams(req),
```
mjungsbluth, line 32: Wouldn't this be even simpler: 

```suggestion
					Path:    req.URL.RequestURI(),
```

mefarazath: You mean use `req.URL.RequestURI()`?
That would discard the path set by any setPath() filter in the chain and use the original request path.

mjungsbluth: But it would work, wouldn't it? https://github.com/golang/go/blob/af0c40311e2ee33ecd24971257606f42a49cf593/src/net/url/url.go#L1165

It uses `EscapedPath()` internally which will use `Path` if it deviates from being the decoded form of `RawPath`

filters/openpolicyagent/internal/envoy/skipperadapter.go
mefarazath: > Assuming that `req.URL.Path` would give /foo, the current behaviour would be that OPA filter/policy would see /foo as the path. So this change of using RequestURI would break that behaviour.
> 
> probably using below would be a way to retain the current behaviour + pass the query params
> 
> ```
> Path:    req.URL.Path + "?" + req.URL.RawQuery
> ```

Made a wrong assumption here 😢 
[req.URL.RequestURI()](https://github.com/golang/go/blob/af0c40311e2ee33ecd24971257606f42a49cf593/src/net/url/url.go#L1165-L1181) actually returns the path with the query params.

Therefore using req.URL.RequestURI() will not omit the path set by setPath() ([test case](https://github.com/zalando/skipper/pull/3207/files#diff-bea895a3252dbca936635114f887dfcdd5a66b7360d676ace6dab16e9a0d8099R77-R89)).

ie. 
setPath("/foo") -> opaAuthorizeRequest("...")
with request to http://.../my-path, the OPA filter would pass "foo" as the path.

filters/openpolicyagent/internal/envoy/skipperadapter.go
mefarazath: Yeah, this simplifies things.
I made a wrong assumption previously that this would conflict with the setPath() filter.

mjungsbluth:
👍

filters/openpolicyagent/internal/envoy/skipperadapter.go
```diff
@@ -11,6 +12,9 @@ import (
 )
 
 func AdaptToExtAuthRequest(req *http.Request, metadata *ext_authz_v3_core.Metadata, contextExtensions map[string]string, rawBody []byte) (*ext_authz_v3.CheckRequest, error) {
+	if err := validateURLForInvalidUTF8(req.URL); err != nil {
+		return nil, fmt.Errorf("invalid url: %w", err)
+	}
```
AlexanderYastrebov, line 17: Why do we need this check?

mefarazath: This was done as a part of reducing errors in decision logs when the http path contained invalid UTF-8 characters.
(Introduced in https://github.com/zalando/skipper/pull/3086)

mefarazath: Sorry should have been precise. I meant the decision logs in OPA.

AlexanderYastrebov: So do we want to save a roundtrip here by pre-validating URL?
Can this be done on OPA side, e.g. can it decide to reject request if URL is not valid?

AlexanderYastrebov:
:+1:

filters/openpolicyagent/internal/envoy/skipperadapter.go
mefarazath: > So do we want to save a roundtrip here by pre-validating URL? Can this be done on OPA side, e.g. can it decide to reject request if URL is not valid?

To be honest, saving the round trip was just a side benefit. 

Our main concern was that requests with invalid URLs were failing before even reaching the policy logic (due to parse errors in OPA), but they still generated decision log entries marked as errors. This was creating a lot of noise in some of the application policies, especially when there were frequent malicious requests happening every day. We were worried that this noise could cause the teams to overlook or ignore real policy errors that actually need attention.

PR #3208, opened 2024-08-26
options: test filter registration order

AlexanderYastrebov:
Test that built-in and CustomFilters are registered before RegisterFilters callback.

Follow up on #3203


MustafaSaber:
👍

PR #3209, opened 2024-08-27
use standard `errors.Join` since we already use go1.22

MustafaSaber:
check https://pkg.go.dev/errors#Join
follow up on https://github.com/zalando/skipper/pull/2492


AlexanderYastrebov:
:+1:

RomanZavodskikh:
:+1:

PR #3210, opened 2024-08-27
Remove OPA filter `dependency.base` calculation in optional body parsing

Pushpalanka:
Due to a detected performance concern for CPU as seen below,

image

**Memory profile**

_alloc_objects profile_
image

### Load test results
<details>
<summary>Using a request body as below with a simple policy to check email domain.</summary>

```
{
      "email": "loadtest@zalando.de",
      "user": {
        "id": "12345",
        "name": {
          "first_name": "John",
          "last_name": "Doe"},
        "contact": {
          "phone_number": "+49-123-456-7890",
          "address": {
            "street": "Main Street 123",
            "city": "Berlin",
            "zip_code": "10115",
            "country": "Germany"}}},
      "account": {
        "created_at": "2023-08-28T10:15:30Z",
        "status": "active",
        "preferences": {
          "language": "en",
          "currency": "EUR"},
        "subscriptions": {
          "newsletter": true,
          "alerts": {
            "email": true,
            "sms": false}}},
      "meta": {
        "last_login": "2023-08-28T10:20:00Z",
        "login_count": 42,
        "roles": [
          "user",
          "tester"],
        "tags": [
          "load-test",
          "test-user"]}
    }
```
</details>

**Before this improvement**
23 instances for 1000 rps 
max GC pause : 500 µs (high number of small objects generated by the dependency.base functions seems to have contributed here)
max heap objects : 3.17M
max heap usage: 557 MB

**After this improvement**
11 instances for 1000 rps  (50% reduced of CPU power)
max GC pause: 269 µs. 
max heap objects : 1.28M (50% reduced.)
max heap usage: 236 MB 

### Bench-mark test results
_Before_
```BenchmarkAuthorizeRequest/authorize-request-with-body-8                    10002           1068249 ns/op         8432573 B/op             806 allocs/op```

_After_
```BenchmarkAuthorizeRequest/authorize-request-with-body-8                    13461            587883 ns/op         8418624 B/op             566 allocs/op```

### DIFF
--- a/filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
+++ b/filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
@@ -545,6 +545,54 @@ func BenchmarkAuthorizeRequest(b *testing.B) {
 		}
 	})
 
+	b.Run("authorize-request-with-body", func(b *testing.B) {
+		opaControlPlane := opasdktest.MustNewServer(
+			opasdktest.MockBundle("/bundles/somebundle.tar.gz", map[string]string{
+				"main.rego": `
+					package envoy.authz
+
+					import rego.v1
+
+					default allow = false
+
+					allow if {
+						endswith(input.parsed_body.email, "@zalando.de")
+					}
+				`,
+			}),
+		)
+
+		f, err := createBodyBasedOpaFilter(opaControlPlane)
+		assert.NoError(b, err)
+
+		url, err := url.Parse("http://opa-authorized.test/somepath")
+		assert.NoError(b, err)
+
+		body := `{"email": "bench-test@zalando.de"}`
+		ctx := &filtertest.Context{
+			FStateBag: map[string]interface{}{},
+			FResponse: &http.Response{},
+			FRequest: &http.Request{
+				Method: "POST",
+				Header: map[string][]string{
+					"Authorization": {"Bearer FOOBAR"},
+					"Content-Type":  {"application/json"},
+				},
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by szuecs** [comment]
@Pushpalanka please also --sign-off the commit (DCO).

**Comment by Pushpalanka** [comment]
Seeking discussion,

I would like to still **optionally** parse request body only if the filter is configured specifically as `opaAuthorizeRequestWithBody` or  `opaServeResponseWithReqBody` even if a request body is present. Then a user can have a conscious decision to apply the body filters to a narrowed down route where request body is necessary for the policy.

Code currently check for `opa.EnvoyPluginConfig().SkipRequestBodyParse` which seems to be an OPA configuration and I am doubtful if this take into account the filter specification configuration we have as `bodyParsing`

```
func NewOpaAuthorizeRequestWithBodySpec(registry *openpolicyagent.OpenPolicyAgentRegistry, opts ...func(*openpolicyagent.OpenPolicyAgentInstanceConfig) error) filters.Spec {
	return &spec{
		registry:    registry,
		opts:        opts,
		name:        filters.OpaAuthorizeRequestWithBodyName,
		**bodyParsing: true,**
	}
}
```

mjungsbluth:
👍

MustafaSaber:
maybe the image for the CPU profile after would be great also to visualise the difference

MustafaSaber:
👍

Pushpalanka:
> maybe the image for the CPU profile after would be great also to visualise the difference

Here we go,

image

PR #3217, opened 2024-09-03
fallback to env variable for oidc,oauth2 client id and secret

czhou-brex:
Similar to [secureOauthTokenintrospectionAllKV](https://github.com/search?q=repo%3Azalando%2Fskipper%20os.LookupEnv&type=code), support falling back to environment varible for client id and secret for oidc and oauth2 filters, when the values passed in are empty strings.

This will allow compatibility with external secret stores.

Testing:
- [x] Run command line skipper with empty args for client_id and/or client_secret, in combination with those values set via env vars in oidc filter
- [x] Run command line skipper with empty args for client_id and/or client_secret, in combination with those values set via env vars in oauthgrant filter

### DIFF
--- a/filters/auth/grantconfig.go
+++ b/filters/auth/grantconfig.go
@@ -6,7 +6,6 @@ import (
 	"net"
 	"net/http"
 	"net/url"
-	"os"
 	"path/filepath"
 	"strings"
 	"time"
@@ -201,10 +200,6 @@ func (c *OAuthConfig) Init() error {
 
 	c.flowState = newFlowState(c.Secrets, c.SecretFile)
 
-	if c.ClientID == "" {
-		c.ClientID, _ = os.LookupEnv("OAUTH2_CLIENT_ID")
-	}
-
 	if c.ClientID != "" {
 		c.getClientId = func(*http.Request) (string, error) {
 			return c.ClientID, nil
@@ -230,10 +225,6 @@ func (c *OAuthConfig) Init() error {
 		}
 	} else {
 		return ErrMissingClientID
-	}
-
-	if c.ClientSecret == "" {
-		c.ClientSecret, _ = os.LookupEnv("OAUTH2_CLIENT_SECRET")
 	}
 
 	if c.ClientSecret != "" {
--- a/filters/auth/grantconfig.go
+++ b/filters/auth/grantconfig.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


czhou-brex:
Possibly closes https://github.com/zalando/skipper/issues/1952

filters/auth/grantconfig.go
```diff
@@ -200,6 +201,10 @@ func (c *OAuthConfig) Init() error {
 
 	c.flowState = newFlowState(c.Secrets, c.SecretFile)
 
+	if c.ClientID == "" {
+		c.ClientID, _ = os.LookupEnv("OAUTH2_CLIENT_ID")
```
MustafaSaber, line 205: I think it makes sense to use the bool for logging.

filters/auth/oidc.go
```diff
@@ -242,19 +243,29 @@ func (s *tokenOidcSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 		validity = defaultCookieValidity
 	}
 
+	oidcClientId := sargs[paramClientID]
+	if oidcClientId == "" {
+		oidcClientId, _ = os.LookupEnv("OIDC_CLIENT_ID")
```
szuecs, line 248: we should lookup once in skipper.go and pass the result values via `OidcOptions`

szuecs:
To close #1952 we would need to use `secrets` module and read from file to allow secrets rotation.

filters/auth/oidc.go
szuecs: see also https://github.com/zalando/skipper/blob/master/skipper.go#L1674-L1680

czhou-brex: The existing OidcOptions struct does not have OIDC_Client_ID - should that be added to the field as a struct?
In addition would the arguments parsing in oidc.go have to be updated? https://github.com/zalando/skipper/blob/master/filters/auth/oidc.go#L80

filters/auth/grantconfig.go
czhou-brex: Do you mean to log empty values for client id and secret? What should be the logging level?
The current version always fetches the env var for oidc so logging that might not make sense.

szuecs:
Changes look good. Please update the docs for the filters: see docs/reference/filters.md

table in https://opensource.zalando.com/skipper/reference/filters/#oauth2
listing in https://opensource.zalando.com/skipper/reference/filters/#openid-connect

For CLI flags we would override CLI values with the ENV, which should be written in help messages:
https://github.com/zalando/skipper/blob/master/config/config.go#L484-L485

czhou-brex:
> Changes look good. Please update the docs for the filters: see docs/reference/filters.md
> 
> table in https://opensource.zalando.com/skipper/reference/filters/#oauth2 listing in https://opensource.zalando.com/skipper/reference/filters/#openid-connect
> 
> For CLI flags we would override CLI values with the ENV, which should be written in help messages: https://github.com/zalando/skipper/blob/master/config/config.go#L484-L485

Docs have been updated.

filters/auth/grantconfig.go
MustafaSaber: > What should be the logging level?

log on debug level, that the fail back also failed to fetch the value.

MustafaSaber: but it's a minor anyway.

RomanZavodskikh:
:+1:

MustafaSaber:
👍

MustafaSaber:
thanks @czhou-brex for your contribution 🙏!

PR #3222, opened 2024-09-10
docs: fix HeaderRegexp example

AlexanderYastrebov:
From https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

> Elements are comma-separated, with optional whitespace surrounding the commas.
> If a request goes through multiple proxies, the IP addresses of each successive proxy is listed.
> This means that, given well-behaved client and proxies, the rightmost IP address is the IP address
> of the most recent proxy and the leftmost IP address is the IP address of the originating client.

Fix example by removing trailing space.


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #3225, opened 2024-09-11
filters/auth: cache yaml config

AlexanderYastrebov:
Cache parsed yaml config for `jwtMetrics` and `oauthTokeninfoValidate` filters to avoid parsing the same configuration over and over when these filters are appended as default filters to all routes.

```
                          │     HEAD~1     │                 HEAD                 │
                          │     sec/op     │    sec/op     vs base                │
JwtMetrics_CreateFilter-8   30777.00n ± 6%   18.45n ± 11%  -99.94% (p=0.000 n=10)

                          │     B/op     │    B/op      vs base                 │
JwtMetrics_CreateFilter-8   24.33Ki ± 0%   0.00Ki ± 0%  -100.00% (p=0.000 n=10)

                          │ allocs/op  │ allocs/op  vs base                 │
JwtMetrics_CreateFilter-8   180.0 ± 0%    0.0 ± 0%  -100.00% (p=0.000 n=10)
```

### DIFF
--- a/filters/auth/jwt_metrics.go
+++ b/filters/auth/jwt_metrics.go
@@ -6,7 +6,6 @@ import (
 	"slices"
 	"strings"
 
-	"github.com/ghodss/yaml"
 	"github.com/opentracing/opentracing-go"
 	"github.com/zalando/skipper/filters"
 	"github.com/zalando/skipper/filters/annotate"
@@ -14,8 +13,12 @@ import (
 )
 
 type (
-	jwtMetricsSpec struct{}
+	jwtMetricsSpec struct {
+		yamlConfigParser[jwtMetricsFilter]
+	}
 
+	// jwtMetricsFilter implements [yamlConfig],
+	// make sure it is not modified after initialization.
 	jwtMetricsFilter struct {
 		Issuers           []string `json:"issuers,omitempty"`
 		OptOutAnnotations []string `json:"optOutAnnotations,omitempty"`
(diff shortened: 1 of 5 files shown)


szuecs:
:+1:

filters/auth/jwt_metrics.go
```diff
@@ -27,34 +28,31 @@ type (
 )
 
 func NewJwtMetrics() filters.Spec {
-	return &jwtMetricsSpec{}
+	return &jwtMetricsSpec{
+		yamlConfig: newYamlConfig[jwtMetricsFilter](64),
```
AlexanderYastrebov, line 32: 64 here is an arbitrary cache size here. In case number of distinct yaml configurations exceeds it the caching still has value because it will hold recently parsed configurations.

AlexanderYastrebov: Decided currently it does not make sense to make it configurable, we can do it later if needed.

filters/auth/yamlconfig.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"fmt"
+
+	"github.com/ghodss/yaml"
+)
+
+// yamlConfigParser parses and caches yaml configurations of type T.
+// Use [newYamlConfigParser] to create instances and ensure that *T implements [yamlConfig].
+type yamlConfigParser[T any] struct {
+	initialize func(*T) error
```
MustafaSaber, line 12: why do we need to enforce this?

AlexanderYastrebov: The idea is to ensure that user of yamlConfigParser is aware that config values must be immutable.

MustafaSaber:
👍

AlexanderYastrebov:
:+1:

Issue #2038, opened 2022-07-07
Set host header in order to use it in another route lookup


szuecs:
If you add `dropRequestHeader("Host") -> setRequestHeader("Host", "bar") -> <loopback>` the new host "bar" is not replacing the host header, but appending it. This works differently from the network backend.

Example
```
./bin/skipper -inline-routes='
r: HostAny("foo") -> logHeader() -> preserveHost("true") -> dropRequestHeader("Host") -> setRequestHeader("Host", "bar") -> <loopback>; 
r1: HostAny("bar") -> status(200) -> inlineContent("bar") -> <shunt>;'
[APP]INFO[0002] GET /foo HTTP/1.1
Host: foo                 <----
User-Agent: curl/7.49.0
Accept: */*

[APP]INFO[0002] GET /foo HTTP/1.1
Host: foo                 <----
Accept: */*
Host: bar                 <----
User-Agent: curl/7.49.0

[APP]INFO[0002] GET /foo HTTP/1.1
Host: foo                 <----
Host: bar                 <----
User-Agent: curl/7.49.0
Accept: */*

...
[APP]ERRO[0002] error while proxying after 5.67122ms, route r with backend loopback ://, status code 500: max loopbacks reached, remote host: ::1, request: "GET /foo HTTP/1.1", user agent: "curl/7.49.0"
::1 - - [07/Jul/2022:15:24:03 +0200] "GET /foo HTTP/1.1" 500 22 "-" "curl/7.49.0" 5 foo - -
```

network backend:
```
% ./bin/skipper -inline-routes='
r: HostAny("foo") -> logHeader() -> preserveHost("false") -> setRequestHeader("Host", "bar") -> "http://127.0.0.1:9090"; 
r1: HostAny("bar") -> status(200) -> inlineContent("bar") -> <shunt>;'

[APP]INFO[0001] GET /foo HTTP/1.1
Host: foo
User-Agent: curl/7.49.0
Accept: */*

127.0.0.1 - - [07/Jul/2022:15:29:40 +0200] "GET /foo HTTP/1.1" 200 3 "-" "curl/7.49.0" 1 bar - -
::1 - - [07/Jul/2022:15:29:40 +0200] "GET /foo HTTP/1.1" 200 3 "-" "curl/7.49.0" 3 foo - -
```

AlexanderYastrebov:
[...]

PR #3228, opened 2024-09-16
Add X-Skipper-Redirect-Base-Uri override headers for oauthgrant filter redirect_uri

czhou-brex:


szuecs:
The problem with override by header is that it can easily lead to a security issue ([example](https://github.com/zalando/skipper/security/advisories/GHSA-f2rj-m42r-6jm2)). 
I think we need a different implementation if that's needed. A trigger from a client can not be the way we want to have it.

czhou-brex:
> The problem with override by header is that it can easily lead to a security issue ([example](https://github.com/zalando/skipper/security/advisories/GHSA-f2rj-m42r-6jm2)). I think we need a different implementation if that's needed. A trigger from a client can not be the way we want to have it.

It is true the client can now pass an arbitrary value, but ultimately the IDP will decide which redirect_uri values can be or cannot be allowed.

czhou-brex:
https://github.com/zalando/skipper/issues/3229

kliakhovskii-brex:
> Please describe the problem it intends to solve (maybe raise a ticket).

Will prob follow up on @czhou-brex behalf for now as we're trying to use for oidc/auth-grant authN for our services.

The problems they're running into is having another proxy in front of Skipper. In our case it's a CloudFront that caches some resources, so the routing kind of looks like the following:
- user loads a browser page
- it goes to... let's say `cloudfront.brex.com`
- then it has Skipper endpoint as origin, let's say `skipper.brex.com`
- if `authGrant` needs to redirect users to identity provider, it generates the redirect_uri out of the current host (which is `skipper.brex.com`) but this is not visible host to the user.
- Ideally what we want is to have a redirect_uri that will go back to smth like `cloudfront.brex.com/oauth/callback`

So it's basically how to redirect to the top-most proxy in the chain when Skipper host isn't what directly visible to users due to smth like CDN.

PS Another potential use case we're trying to figure out is localhost experience for things like webpack dev server, technically it is an acting proxy in front of Skipper too, and it can even do smth like `localhost:3000/oauth/callback` redirect, and set a cookie for localhost too. But this is secondary right now (though curious to hear your thoughts on this).

AlexanderYastrebov:
> * it goes to... let's say cloudfront.brex.com
> * then it has Skipper endpoint as origin, let's say skipper.brex.com

Is it possible to configure first proxy to [pass](https://repost.aws/knowledge-center/configure-cloudfront-to-forward-headers) "cloudfront.brex.com" as a Host header?

> The filter will substitute the base URL of redirect_uri, if "X-Skipper-Redirect-Base-Uri" header is passed in the request.

I am not sure I understand who is supposed to send this header? Likely its not browser but first proxy, or?

kliakhovskii-brex:
> Is it possible to configure first proxy to [pass](https://repost.aws/knowledge-center/configure-cloudfront-to-forward-headers) "cloudfront.brex.com" as a Host header?

Yes it's doable, but then Skipper has to have routing rules for each external host. For things like localhost (or any local custom hosts e.g. `dev.example.com` that points to your localhost) i find it somewhat weird.

> I am not sure I understand who is supposed to send this header? Likely its not browser but first proxy, or?

Yes, I would expect a first proxy to pass it somewhat.

What's interesting, @mmichels-brex was testing it with both `authGrant` & `oidc` filters, and the latter seem to be going back to the right host. This is one of the reasons I believe we will likely stick to `oidc` & just finalize filters for logout (as they practically don't exist on Skipper today).

AlexanderYastrebov:
I think the "standard" way is to use [X-Forwarded-Host](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) header.

In case of `oauthGrant` its not enough to patch redirect URI because request domain is also used to configure token cookie (see uses of `extractDomainFromHost`).

If we move towards passing host from upstream proxy then I think we should think about using `X-Forwarded-Host` to override incoming `request.Host` via middleware before routing. And also investigate how other proxies solve the issue.

> What's interesting, @mmichels-brex was testing it with both authGrant & oidc filters, and the latter seem to be going back to the right host. 

oid seems to also consider "host" header (see #1016) but I think this is not correct because go HTTP server removes Host header and puts its value to the [request.Host](https://pkg.go.dev/net/http#Request):
```go
// For incoming requests, the Host header is promoted to the
// Request.Host field and removed from the Header map.
```
So maybe if oidc route has `setRequestHeader("Host", "cloudfront.brex.com")` then it works (I don't know how your route that works looks like).

See also Host header related recent discussion https://github.com/zalando/skipper/issues/2038

AlexanderYastrebov:
> Yes it's doable, but then Skipper has to have routing rules for each external host. For things like localhost (or any local custom hosts e.g. dev.example.com that points to your localhost) i find it somewhat weird.

Why? Maybe you could provide your example routes for better understanding?
By default `oauthGrant` and oidc filters will derive domain from req.Host which comes from Host header.
So if upstream proxy sets proper Host header everything should work out of the box.

PR #3234, opened 2024-09-18
Add constant cookie name for the OIDC filter

czhou-brex:
In order to support the logout filter for OIDC, we are proposing a parameter to make the cookie name a constant, so it can be read accordingly.

This is one of the PRs for https://github.com/zalando/skipper/issues/3235

Thanks to @matheus-michels for the code.


szuecs:
I just want to be clear:
Do you know that this cookie name is only a prefix of the name, because we will split the content into multiple cookies because the size of the data we need in the cookie is often too big for browsers to accept?
So you will only match by "prefix" the cookie names. The number of cookies can change and we have the feature to split into more cookies and merge into less cookies based on data needed.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

czhou-brex:
> I just want to be clear: Do you know that this cookie name is only a prefix of the name, because we will split the content into multiple cookies because the size of the data we need in the cookie is often too big for browsers to accept? So you will only match by "prefix" the cookie names. The number of cookies can change and we have the feature to split into more cookies and merge into less cookies based on data needed.

Thanks for explaining this. So far we were only able to observe one skipper cookie for the OIDC session. Can you elaborate on cases where there may be multiple cookies? Are these cases only within auth or outside of auth?

AlexanderYastrebov:
@czhou-brex Thank you for contribution.

Next time please rebase and squash your PR on top of the latest master so we do not have github confused and its easier to write release notes as they are composed from all commit messages

Screenshot from 2024-09-19 18-21-48

PR #3237, opened 2024-09-19
Add url path normalizer

### LINKED ISSUE #3236: Add a filter that normalizes url paths by removing trailing slashes and empty path segments

At Zalando we have [defined a guiding rule](https://opensource.zalando.com/restful-api-guidelines/#136) for API designers to specify APIs without empty path segments and trailing slashes

However, client implementations in production may be configured poorly to contain trailing slashes which the rule also suggests the server to protect against by

```
We recommend to implement services robust against clients not following this rule. All services
should normalize request paths before processing by removing duplicate and trailing slashes.
```

This can be implemented in a simple filter in Skipper.

A `normalizePath()`-filter which transforms URLs to contain no empty path segments and no trailing slashes.

URLs such as 
```
GET /orders/{order-id}
GET /orders/{order-id}/
GET /orders//{order-id}
```
should be normalized to `/orders/{order-id}`.

We can achieve this by using the existing `modPath`-filter and a regular expression, which can be sufficient but also lead to incidents where the regular expression is poorly written and breaks production environment. 

Yes

mvalkon:
Adds a filter called `normalizePath` which implements URL Path
normalization by removing empty path segments and trailing slashes from
the path. This follows the guidance from the Zalando Restful API
Guidelines rule [136][1].

[1]: https://opensource.zalando.com/restful-api-guidelines/#136

Closes #3236

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -365,6 +365,32 @@ Parameters:
 The replacement may contain [template placeholders](#template-placeholders).
 If a template placeholder can't be resolved then empty value is used for it.
 
+### normalizePath
+
+Normalize the URL path by removing empty path segments and trailing slashes.
+
+Parameters:
+
+- The URL path
+
+Example:
+
+```
+all: * -> normalizePath() -> "https://backend.example.org/api/v1;
+```
+
+Requests to
+
+```
+https://backend.example.org//api/v1/
+https://backend.example.org//api/v1
+https://backend.example.org/api//v1
+https://backend.example.org/api//v1/
+https://backend.example.org/api/v1/
+```
+
+will all be normalized to `https://backend.example.org/api/v1`. 
+
 ## HTTP Redirect
 ### redirectTo
 
--- a/filters/filters.go
+++ b/filters/filters.go
@@ -370,4 +370,5 @@ const (
 	SetFastCgiFilenameName = "setFastCgiFilename"
 	DisableRatelimitName   = "disableRatelimit"
 	UnknownRatelimitName   = "unknownRatelimit"
+	NormalizePath          = "normalizePath"
 )
--- a/filters/normalizepath/normalizepath.go
+++ b/filters/normalizepath/normalizepath.go
@@ -32,11 +32,6 @@ func (f normalizePath) Request(ctx filters.FilterContext) {
 	}
 	normalizedPath := "/" + strings.Join(filteredSegments, "/")
 
-	// Ensure there's no trailing slash, unless the path is just "/"
-	if len(normalizedPath) > 1 && strings.HasSuffix(normalizedPath, "/") {
-		normalizedPath = normalizedPath[:len(normalizedPath)-1]
-	}
-
 	req.URL.Path = normalizedPath
 }
 
--- a/filters/normalizepath/normalizepath.go
+++ b/filters/normalizepath/normalizepath.go
@@ -1,7 +1,7 @@
 package normalizepath
 
 import (
-	"path"
+	"strings"
 
 	"github.com/zalando/skipper/filters"
 )
@@ -22,7 +22,22 @@ func (spec normalizePath) CreateFilter(config []interface{}) (filters.Filter, er
 
 func (f normalizePath) Request(ctx filters.FilterContext) {
 	req := ctx.Request()
-	req.URL.Path = path.Clean(req.URL.Path)
+
+	segments := strings.Split(req.URL.Path, "/")
+	var filteredSegments []string
+	for _, seg := range segments {
+		if seg != "" {
+			filteredSegments = append(filteredSegments, seg)
+		}
+	}
+	normalizedPath := "/" + strings.Join(filteredSegments, "/")
+
+	// Ensure there's no trailing slash, unless the path is just "/"
+	if len(normalizedPath) > 1 && strings.HasSuffix(normalizedPath, "/") {
+		normalizedPath = normalizedPath[:len(normalizedPath)-1]
+	}
+
+	req.URL.Path = normalizedPath
 }
 
 func (f normalizePath) Response(ctx filters.FilterContext) {}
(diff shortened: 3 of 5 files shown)


docs/reference/filters.md
```diff
@@ -365,6 +365,32 @@ Parameters:
...
+Normalize the URL path by removing empty path segments and trailing slashes.
+
+Parameters:
+
+- The URL path
+
+Example:
+
+```
+all: * -> normalizePath() -> "https://backend.example.org/api/v1;
```
szuecs, line 379: missing closing `"` and network backend is not having a path segment.
`all: * -> normalizePath() -> "https://backend.example.org";`

filters/filters.go
```diff
@@ -370,4 +370,5 @@ const (
 	SetFastCgiFilenameName = "setFastCgiFilename"
 	DisableRatelimitName   = "disableRatelimit"
 	UnknownRatelimitName   = "unknownRatelimit"
+	NormalizePath          = "normalizePath"
```
szuecs, line 373: NormalizePath -> NormalizePathName

filters/normalizepath/normalizepath.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+type normalizePath struct{}
+
+func NewNormalizePath() filters.Spec { return normalizePath{} }
+
+func (spec normalizePath) Name() string { return "normalizePath" }
+
+func (spec normalizePath) CreateFilter(config []interface{}) (filters.Filter, error) {
+	return normalizePath{}, nil
```
szuecs, line 20: I think you can `return spec, nil` and if you have a ptr receiver you don't need to have allocations for the filter:

```
func NewNormalizePath() filters.Spec { return &normalizePath{} }

func (f *normalizePath) Name() string { return filters.NormalizePathName }

func (f *normalizePath) CreateFilter(config []interface{}) (filters.Filter, error) {
    return f, nil
}

func (f *normalizePath) Request(ctx filters.FilterContext) {
...
```

filters/normalizepath/normalizepath.go
```diff
@@ -0,0 +1,38 @@
...
+	req := ctx.Request()
+
+	segments := strings.Split(req.URL.Path, "/")
+	var filteredSegments []string
+	for _, seg := range segments {
+		if seg != "" {
+			filteredSegments = append(filteredSegments, seg)
+		}
+	}
+	normalizedPath := "/" + strings.Join(filteredSegments, "/")
```
szuecs, line 33: semgrep asks: 
> did you want path.Join() or filepath.Join()?

AlexanderYastrebov:
Couple of thoughts:
* Filter name is generic and implies normalization but it only does duplicate slash elimination
* Link to Zalando Guidelines mentions https://en.wikipedia.org/wiki/URI_normalization which mentions [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) which describes a lot of other normalization methods but does not describe duplicate slash elimination
* There is [`rfcPath`](https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#rfcpath) filter that performs escape normalization and also mentions RFC 3986
* This filter will be invoked after routing which means Path-related predicates will not benefit from it.
* There is `-rfc-patch-path` flag that reuses the same machinery as existing `rfcPath` at proxy level and thus fixes path before routing.

So I am a bit puzzled here. This filter does not fix routing so maybe we could extend existing `rfcPath` which also can be enabled globally before routing. On the other hand duplicate slash is not against RFC (see [this](https://stackoverflow.com/questions/10161177/url-with-multiple-forward-slashes-does-it-break-anything)) so eliminating duplicate slashes is not "RFC"-compliant. Lastly, what problem does it really solve - maybe answering client 404 (in case they sent wrong path that does not match any route) or passing url as is to the backend is a better approach then silently fixing the path.

mvalkon:
> Couple of thoughts:
> 
> * Filter name is generic and implies normalization but it only does duplicate slash elimination
> * Link to Zalando Guidelines mentions https://en.wikipedia.org/wiki/URI_normalization which mentions [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) which describes a lot of other normalization methods but does not describe duplicate slash elimination

I agree. The normalization is only normalizing the path following the guidance (see below snippet) from the API guidelines, and I've raised the topic internally to see if we should extend this beyond the scope of this.

```
We recommend to implement services robust against clients not following this rule. All services should [normalize](https://en.wikipedia.org/wiki/URI_normalization) request paths before processing by removing duplicate and trailing slashes.
```

> * There is [`rfcPath`](https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#rfcpath) filter that performs escape normalization and also mentions RFC 3986
> * This filter will be invoked after routing which means Path-related predicates will not benefit from it.
> * There is `-rfc-patch-path` flag that reuses the same machinery as existing `rfcPath` at proxy level and thus fixes path before routing.

I'm happy to extend an existing filter. I have to admit my knowledge of skipper is limited - I'm assuming that this filter would work for all paths if you have a catch-all predicate `*` defined, but this might be overly simplistic assumption.

> So I am a bit puzzled here. This filter does not fix routing so maybe we could extend existing `rfcPath` which also can be enabled globally before routing. On the other hand duplicate slash is not against RFC (see [this](https://stackoverflow.com/questions/10161177/url-with-multiple-forward-slashes-does-it-break-anything)) so eliminating duplicate slashes is not "RFC"-compliant. Lastly, what problem does it really solve - maybe answering client 404 (in case they sent wrong path that does not match any route) or passing url as is to the backend is a better approach then silently fixing the path.

While it's not RFC compliant as such, the unfortunate reality is that backend frameworks and server implementations treat trailing slashes in different ways. We had an incident internally recently where an API that previously supported endpoints with trailing slashes stopped doing so after a version upgrade of Spring Boot - the new version no longer treated `/foobar` and `/foobar/` as the same URL. There was of course a client which was incorrectly configured to call `/foobar/` which triggered the incident. We clearly guide (see guidelines) that APIs must not specify paths with duplicate or trailing slashes, but we can't control all the client configurations and from there the same guideline recommends server implementations to be robust against poorly configured clients.

PR #3238, opened 2024-09-19
feature: dataclient kubernetes zone awareness

szuecs:
redis should not be part of the az awareness as used by global rate limit

### DIFF
--- a/dataclients/kubernetes/clusterclient.go
+++ b/dataclients/kubernetes/clusterclient.go
@@ -562,7 +562,7 @@ func collectReadyEndpoints(endpointSlices *endpointSliceList) map[definitions.Re
 }
 
 // loadEndpointAddresses returns the list of all addresses for the given service using endpoints or endpointslices API.
-func (c *clusterClient) loadEndpointAddresses(namespace, name string) ([]string, error) {
+func (c *clusterClient) loadEndpointAddresses(zone, namespace, name string) ([]string, error) {
 	var result []string
 	if c.enableEndpointSlices {
 		url := fmt.Sprintf(EndpointSlicesNamespaceFmt, namespace) +
@@ -579,7 +579,11 @@ func (c *clusterClient) loadEndpointAddresses(namespace, name string) ([]string,
 		}
 
 		for _, eps := range ready {
-			result = eps.addresses()
+			if zone != "" {
+				result = eps.addressesByZone(zone)
+			} else {
+				result = eps.addresses()
+			}
 			break
 		}
 	} else {
--- a/skipper.go
+++ b/skipper.go
@@ -286,6 +286,9 @@ type Options struct {
 	// KubernetesRedisServicePort to be used to lookup ring shards dynamically
 	KubernetesRedisServicePort int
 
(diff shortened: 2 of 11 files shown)


greeshma1196:
ingressContext seems to be missing zone information, I think it should be defined [here](https://github.com/zalando/skipper/blob/38f6711c66552edfa80c1524ddee91b5a771c026/dataclients/kubernetes/ingressv1.go#L430)

skipper.go
```diff
@@ -1531,14 +1535,14 @@ func getKubernetesRedisAddrUpdater(opts *Options, kdc *kubernetes.Client, loaded
 		// has polled the data once or kdc.GetEndpointAddresses should be blocking
 		// call to kubernetes API
 		return func() ([]string, error) {
-			a := kdc.GetEndpointAddresses(opts.KubernetesRedisServiceNamespace, opts.KubernetesRedisServiceName)
+			a := kdc.GetEndpointAddresses("", opts.KubernetesRedisServiceNamespace, opts.KubernetesRedisServiceName)
```
szuecs, line 1538: This is for redis/valkey shards, which we do not want to have zone awareness for, because the shard should be 1 not 1 per zone.

skipper.go
```diff
@@ -1531,14 +1535,14 @@ func getKubernetesRedisAddrUpdater(opts *Options, kdc *kubernetes.Client, loaded
...
 			log.Debugf("GetEndpointAddresses found %d redis endpoints", len(a))
 
 			return joinPort(a, opts.KubernetesRedisServicePort), nil
 		}
 	} else {
 		return func() ([]string, error) {
-			a, err := kdc.LoadEndpointAddresses(opts.KubernetesRedisServiceNamespace, opts.KubernetesRedisServiceName)
+			a, err := kdc.LoadEndpointAddresses("", opts.KubernetesRedisServiceNamespace, opts.KubernetesRedisServiceName)
```
szuecs, line 1545: This is for redis/valkey shards, which we do not want to have zone awareness for, because the shard should be 1 not 1 per zone.

dataclients/kubernetes/clusterclient.go
```diff
@@ -591,6 +597,9 @@ func (c *clusterClient) loadEndpointAddresses(namespace, name string) ([]string,
 		}
 		result = ep.addresses()
 	}
+	if len(resultByZone) > 2 {
```
MustafaSaber: I don't understand why we do this override

szuecs: This dataclient is zone aware so we return only the result of the zone if we have enough endpoints (min 3).
The rest of the function body should be the same, so the override instead of a return from here.

I hope this clarifies, if not please ask more specific. Thanks

MustafaSaber: Taking a fresh look this makes sense, maybe reuse the variable `minEndpointsByZone` instead of the magic number

szuecs: fixed

MustafaSaber (approved):

a4180p:
👍

PR #3240, opened 2024-09-19
filters/auth: fix assert usage in tokeninfocache tests

AlexanderYastrebov:
Second argument of assert.Equal should be expected value.


MustafaSaber:
👍

AlexanderYastrebov:
:+1:

PR #3246, opened 2024-09-23
net: add server connection keepalive limits

AlexanderYastrebov:
Clients may connect to a subset of Skipper fleet which leads to uneven request distribution and increased cpu usage.

Autoscaling of Skipper fleet is not effective because clients stay connected to old instances while new instances are underutilized.

This change adds ConnManager that tracks creation of new connections and closes connections when their age or number of requests served reaches configured limits.


config/config.go
```diff
@@ -243,6 +243,8 @@ type Config struct {
 	ReadHeaderTimeoutServer      time.Duration `yaml:"read-header-timeout-server"`
 	WriteTimeoutServer           time.Duration `yaml:"write-timeout-server"`
 	IdleTimeoutServer            time.Duration `yaml:"idle-timeout-server"`
+	KeepaliveServer              time.Duration `yaml:"keepalive-server"`
```
AlexanderYastrebov, line 246: Use naming similar to existing `keepalive-backend`

net/connmanager.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (cm *ConnManager) connContext(ctx context.Context, _ net.Conn) context.Context {
+	state := &connState{
+		expiresAt: time.Now().Add(cm.Keepalive),
+	}
+	return context.WithValue(ctx, connection, state)
+}
+
+func (cm *ConnManager) connState(_ net.Conn, state http.ConnState) {
+	cm.count(fmt.Sprintf("lb-conn-%s", state))
```
szuecs, line 88: :+1: , I think this is to have the current metrics, right?

AlexanderYastrebov: Yes, moved it here

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #3247, opened 2024-09-24
docs: document single argument variant of `tracingBaggageToTag`

AlexanderYastrebov:
Implementation allows single argument variant which is used in the wild so document it.

Follow up on #1160 and #1168


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #3258, opened 2024-10-01
feature: az awareness via endpointregistry and routesrv - option 4

szuecs:
TODO


szuecs:

PR #3259, opened 2024-10-01
Improvements to OPA filter unit test cases

mefarazath:
- Fix test case that uses an invalid rego query in the test case
- Include a test case to validate policy evaluation based on `input.attributes.request.http.path` (http path set in the envoy check request)


szuecs:
:+1:

MustafaSaber:
👍

PR #3261, opened 2024-10-04
enable opa-wasm

sepehrdaddev:
currently, the rego bundles are distributed using rego which runs in the rego interpreter, with this pr, I enable the wasm evaluation engine of opa which will allow opa to evaluate wasm based bundles which based on this [blogpost](https://blog.openpolicyagent.org/opa-v0-15-1-rego-on-webassembly-81c226c51be4) it should give us a 10x performance boost.

### DIFF
--- a/filters/openpolicyagent/evaluation.go
+++ b/filters/openpolicyagent/evaluation.go
@@ -3,6 +3,8 @@ package openpolicyagent
 import (
 	"context"
 	"fmt"
+	"time"
+
 	ext_authz_v3_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	ext_authz_v3 "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
 	"github.com/open-policy-agent/opa-envoy-plugin/envoyauth"
@@ -13,7 +15,8 @@ import (
 	"github.com/open-policy-agent/opa/topdown"
 	"github.com/opentracing/opentracing-go"
 	pbstruct "google.golang.org/protobuf/types/known/structpb"
-	"time"
+
+	_ "github.com/open-policy-agent/opa/features/wasm"
 )
 
 func (opa *OpenPolicyAgentInstance) Eval(ctx context.Context, req *ext_authz_v3.CheckRequest) (*envoyauth.EvalResult, error) {
@@ -70,7 +73,15 @@ func (opa *OpenPolicyAgentInstance) Eval(ctx context.Context, req *ext_authz_v3.
 		return nil, err
 	}
 
-	err = envoyauth.Eval(ctx, opa, inputValue, result, rego.DistributedTracingOpts(opa.DistributedTracing()))
+	opts := []func(*rego.Rego){rego.DistributedTracingOpts(opa.DistributedTracing())}
+
+	for _, r := range opa.manager.GetWasmResolvers() {
+		for _, e := range r.Entrypoints() {
+			opts = append(opts, rego.Resolver(e, r))
+		}
+	}
+
+	err = envoyauth.Eval(ctx, opa, inputValue, result, opts...)
 	if err != nil {
 		return nil, err
 	}


sepehrdaddev:
the way to test it:

```
# policy.rego
package envoy.http.public

import rego.v1

default allow := false

allow if {
	input.attributes.request.http.method == "GET"
	input.attributes.request.http.path == "/"
}

allow if input.attributes.request.http.headers.authorization == "Basic charlie"
```

```
# opaconfig.yaml
bundles:
  play:
    resource: bundles/{{ .bundlename }}
    polling:
      long_polling_timeout_seconds: 45
services:
  - name: play
    url: http://127.0.0.1:8000 # your http server address
plugins:
  envoy_ext_authz_grpc:
    # This needs to match the package, defaulting to envoy/authz/allow
    path: envoy/http/public/allow
    dry-run: false
decision_logs:
  console: true
```

```sh
# compile policy to wasm

$ opa build -t wasm -e envoy/http/public/allow policy.rego # the entrypoint is the name of the package in the rego file

# prepare the http server to host the bundle
$ mkdir -p opa-server/bundles && mv bundle opa-server/bundles/bundle

# run python simple http in the directory
$ cd opa-server && python3 -m http.server

# run skipper

$ skipper -enable-open-policy-agent -open-policy-agent-config-template opaconfig.yaml -inline-routes 'notfound: * -> opaAuthorizeRequest("play") -> inlineContent("<h1>Authorized Hello</h1>") -> <shunt>'
```

sepehrdaddev:
please ignore the fuzzer compile errors for now, I will fix them later in another PR.

sepehrdaddev:
@szuecs / @AlexanderYastrebov PTAL

sepehrdaddev:
:+1:

szuecs:
@mjungsbluth FYI

MustafaSaber:
Can you add performance tests to show the improvement?

szuecs:
Yes and no feature without a test and optimizations should have a benchmark.

sepehrdaddev:
@MustafaSaber I have asked our IAM guys to have a look and do some benchmarks and loadtests, the results will be posted here,

sepehrdaddev:
I need to highlight that this has no user impact as the opa plugin automatically detects if the bundle is rego or wasm and it will run them in the appropriate environment, so current user base will not be affected as both engines are supported in parallel.

Pushpalanka:
It's a nice finding. In addition to the loadtests we have below points also to be considered.

1. Unsupported built-in functions - https://www.openpolicyagent.org/docs/latest/wasm/#current-status as for this it seems WASM is not supporting some built-in functions. Eventhough `http.send` is discouraged either way, not sure what else is missing.

2. Bundle building is handled by the OPA control plane. So far it doesn't seem to support building WASM bundles. Will need a separate approach to handle this, if taking this path.

mjungsbluth:
Interesting experiment @sepehrdaddev , looking forward to the results!

As for local performance testing, you can add one benchmark to the existing local benchmarks [here](https://github.com/zalando/skipper/blob/0e80fe7db8cafc75d3a2a7f5cced58319a0e62ec/filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go#L628). I would however not expect a massive change as the WASM variant should play out its strengths for more complex policies but you should already see some numbers. 

We'll need to see how this then plays out in total as this optimises the evaluation but leaves decision logging and input conversion as is. To see a flame graph with a minimal policy, look [here](https://github.com/open-policy-agent/opa/issues/5724#issuecomment-2312823968)

> I need to highlight that this has no user impact..

This change still changes one of the core ingredients of the filter, the `Compiler`, so in a "been there, done that" sense I would recommend hiding the change of behaviour behind a flag to enable WASM support.

szuecs:
So then I would say add `enable-open-policy-agent-wasm` via config/config.go and pass it in skipper.go to OPA filter configurations

PR #3267, opened 2024-10-09
filters/auth: use sync.Map for tokeninfo cache

AlexanderYastrebov:
* use sync.Map for tokeninfo cache to avoid synchronizing all callers on a single mutex
* evict stale entries periodically instead of least recently used
* store token expiration time instead of creation time

```
                                                  │      master      │                HEAD                 │
                                                  │      sec/op      │   sec/op     vs base                │
TokeninfoCache/tokens=1,cacheSize=1,p=0-8               275.5n ±  6%   170.1n ± 4%  -38.26% (p=0.000 n=10)
TokeninfoCache/tokens=2,cacheSize=2,p=0-8               492.9n ± 21%   176.8n ± 2%  -64.12% (p=0.000 n=10)
TokeninfoCache/tokens=100,cacheSize=100,p=0-8           455.9n ±  7%   165.5n ± 1%  -63.70% (p=0.000 n=10)
TokeninfoCache/tokens=100,cacheSize=100,p=10000-8       593.4n ±  4%   179.8n ± 4%  -69.71% (p=0.000 n=10)
TokeninfoCache/tokens=4,cacheSize=2,p=0-8           2571424.0n ±  0%   149.7n ± 3%  -99.99% (p=0.000 n=10)
TokeninfoCache/tokens=100,cacheSize=10,p=0-8        2579227.5n ±  0%   139.3n ± 1%  -99.99% (p=0.000 n=10)
geomean                                                 7.903µ         162.9n       -97.94%

                                                  │   master   │                  HEAD                   │
                                                  │    B/op    │    B/op      vs base                    │
TokeninfoCache/tokens=1,cacheSize=1,p=0-8           344.0 ± 0%    344.0 ± 0%          ~ (p=1.000 n=10) ¹
TokeninfoCache/tokens=2,cacheSize=2,p=0-8           344.0 ± 0%    344.0 ± 0%          ~ (p=1.000 n=10) ¹
TokeninfoCache/tokens=100,cacheSize=100,p=0-8       344.0 ± 0%    344.0 ± 0%          ~ (p=1.000 n=10) ¹
TokeninfoCache/tokens=100,cacheSize=100,p=10000-8   368.0 ± 1%    350.0 ± 0%     -4.89% (p=0.000 n=10)
TokeninfoCache/tokens=4,cacheSize=2,p=0-8           27.00 ± 0%   344.00 ± 0%  +1174.07% (p=0.000 n=10)
TokeninfoCache/tokens=100,cacheSize=10,p=0-8        27.00 ± 7%   344.00 ± 0%  +1174.07% (p=0.000 n=10)
geomean                                             149.0         345.0        +131.62%
¹ all samples are equal

                                                  │    master    │              HEAD              │
                                                  │  allocs/op   │ allocs/op   vs base            │
TokeninfoCache/tokens=1,cacheSize=1,p=0-8           3.000 ± 0%     3.000 ± 0%  ~ (p=1.000 n=10) ¹
TokeninfoCache/tokens=2,cacheSize=2,p=0-8           3.000 ± 0%     3.000 ± 0%  ~ (p=1.000 n=10) ¹
TokeninfoCache/tokens=100,cacheSize=100,p=0-8       3.000 ± 0%     3.000 ± 0%  ~ (p=1.000 n=10) ¹
TokeninfoCache/tokens=100,cacheSize=100,p=10000-8   3.000 ± 0%     3.000 ± 0%  ~ (p=1.000 n=10) ¹
TokeninfoCache/tokens=4,cacheSize=2,p=0-8           0.000 ± 0%     3.000 ± 0%  ? (p=0.000 n=10)
TokeninfoCache/tokens=100,cacheSize=10,p=0-8        0.000 ± 0%     3.000 ± 0%  ? (p=0.000 n=10)
geomean                                                        ²   3.000       ?
¹ all samples are equal
² summaries must be >0 to compute geomean
```


filters/auth/tokeninfocache.go
```diff
@@ -95,38 +88,61 @@ func (c *tokeninfoCache) tryCache(token string, info map[string]any) {
...
+}
 
-	// create
-	c.cache[token] = &entry{
-		cachedAt:  now,
-		expiresAt: expiresAt,
-		info:      info,
-		href:      c.history.PushFront(token),
+func (c *tokeninfoCache) evictLoop() {
+	ticker := time.NewTicker(time.Minute)
```
RomanZavodskikh, line 110: Does it make sense this `time.Minute` configurable at the next PR?

filters/auth/tokeninfocache.go
```diff
@@ -1,48 +1,55 @@
...
 	tokeninfoCache struct {
-		client tokeninfoClient
-		size   int
-		ttl    time.Duration
-		now    func() time.Time
-
-		mu    sync.Mutex
-		cache map[string]*entry
-		// least recently used token at the end
-		history *list.List
```
RomanZavodskikh, line 21: I see you would like to keep tokeninfo filter lock-free, that is why you are thinking about not keeping history any more.

However, there is a data structure which acts like a list and has lock-free implementations (at least in theory). https://en.wikipedia.org/wiki/Skip_list

Do you think it makes sense to try to obtain this data structure to have lock-free history access and not evict random items?

AlexanderYastrebov: It was used to evict oldest entries when cache grows over the limit.
This change simplifies this by simply removing random items if number of cached items is over the size limit.
It also adds a metric formonitoring.

In production setup one should have cache size set large enough such that entries are never evicted due to overflow. This way there is no need to complicate eviction algorithm, keep access history or use clever datastructures.

filters/auth/tokeninfocache.go
AlexanderYastrebov: I am not sure which other value would be good.

filters/auth/tokeninfocache.go
```diff
@@ -95,38 +88,61 @@ func (c *tokeninfoCache) tryCache(token string, info map[string]any) {
...
+			if c.cache.CompareAndDelete(key, value) {
+				c.count.Add(-1)
+			}
+		}
+		return true
+	})
+
+	// Evict random entries until the cache size is within limits
+	if c.count.Load() > int64(c.size) {
+		c.cache.Range(func(key, value any) bool {
```
MustafaSaber, line 136: just to confirm, the random entries here comes from the random sort of the map, right?

filters/auth/tokeninfocache.go
```diff
@@ -1,48 +1,55 @@
...
-		history *list.List
+		client  tokeninfoClient
+		metrics metrics.Metrics
+		size    int
+		ttl     time.Duration
+		now     func() time.Time
+
+		cache sync.Map     // map[string]*entry
+		count atomic.Int64 // estimated number of cached entries, see https://github.com/golang/go/issues/20680
+		done  chan struct{}
```
szuecs, line 23: done -> quit? 
I think we most often use quit.

filters/auth/authclient.go
```diff
@@ -32,6 +32,7 @@ type authClient struct {
 
 type tokeninfoClient interface {
 	getTokeninfo(token string, ctx filters.FilterContext) (map[string]any, error)
+	Close()
```
szuecs, line 35: io.Closer has `Close() error` or we could also not expose it and use `close()` ?

filters/auth/tokeninfocache_test.go
```diff
@@ -111,7 +118,7 @@ func TestTokeninfoCache(t *testing.T) {
 
 	assert.Equal(t, int32(1), authRequests, "expected no request to auth sever")
 	assert.Equal(t, token, info["uid"])
-	assert.Equal(t, float64(595), info["expires_in"], "expected TokenTTLSeconds - truncate(delay)")
```
RomanZavodskikh, line 114: It is still quite not obvious why 595 changed to 594?

Is it related to this part of code?
https://github.com/zalando/skipper/pull/3267/files#diff-a2721f4aa66d24557c036d686d4e3344636871b2de93fb75266b8da83d511891L79

filters/auth/tokeninfocache.go
AlexanderYastrebov: Yes, the iteration order is no defined for maps and hence for sync.Map as it uses map internally.

filters/auth/tokeninfocache.go
AlexanderYastrebov: Right, will fix
```
ayastrebov@zamd:~/src/github.com/zalando/skipper$ grep -rn -E 'done\s*chan ' | wc -l
2
ayastrebov@zamd:~/src/github.com/zalando/skipper$ grep -rn -E 'quit\s*chan ' | wc -l
22
```

filters/auth/authclient.go
AlexanderYastrebov: This is an unexported interface for internal use.
Tokeninfo client has Close() method and we need one to close tokeninfocache so I put it into this interface.
It makes cleanup simpler and unified for both client and cache.

filters/auth/tokeninfocache_test.go
AlexanderYastrebov: Yes, previous version adjusted cached `expires_in`  based on the time elapsed since entry was cached.
Since the number must be integer it truncated the elapsed time:
```
new_expires_in = expires_in - truncateToSeconds(now-cachedAt)
```
this version stores expiration date instead of cached date and calculates expires_in from it:
```
new_expires_in = truncateToSeconds(infoExpiresAt-now)
```

This test moves clock by `5.7` seconds so previous version resulted in:
```
new_expires_in = expires_in - truncateToSeconds(now-cachedAt) = 600 - truncateToSeconds(cachedAt+5.7-cachedAt) =  truncateToSeconds(5.7) = 595
```
and this version results in:
```
new_expires_in = truncateToSeconds(infoExpiresAt-now) = truncateToSeconds(cachedAt+600-(cachedAt+5.7)) = truncateToSeconds(594.3) = 594
```

filters/auth/tokeninfocache.go
```diff
@@ -58,35 +65,21 @@ func (c *tokeninfoCache) getTokeninfo(token string, ctx filters.FilterContext) (
...
-			cachedInfo := e.info
-			c.mu.Unlock()
-
 			// It might be ok to return cached value
 			// without adjusting "expires_in" to avoid copy
 			// if caller never modifies the result and
 			// when "expires_in" did not change (same second)
 			// or for small TTL values
-			info := shallowCopyOf(cachedInfo)
+			info := maps.Clone(e.info)
```
szuecs, line 77: Is this ok for 2 goroutines access the same e.info ?
I tried to check the docs but it doesn't tell about concurrent access is fine.
A different way could be to use also sync.Map for info member instead of map.

filters/auth/tokeninfocache.go
```diff
@@ -95,38 +88,61 @@ func (c *tokeninfoCache) tryCache(token string, info map[string]any) {
...
-	// remove least used
-	if len(c.cache) > c.size {
-		leastUsed := c.history.Back()
-		delete(c.cache, leastUsed.Value.(string))
-		c.history.Remove(leastUsed)
+func (c *tokeninfoCache) evict() {
+	// Evict expired entries
+	c.cache.Range(func(key, value any) bool {
+		e := value.(*entry)
+		if c.now().After(e.expiresAt) {
```
szuecs, line 126: We could also reuse c.now() and just run it once.
```
now := c.now()
c.cache.Range ...
...
   if now.After ...
```
This would reduce number of syscalls to get time.

szuecs (commented):
I hope my review makes sense.
Happy to see it happen!

filters/auth/tokeninfocache.go
MustafaSaber: > Is this ok for 2 goroutines access the same e.info ?

Is it a problem if both are readonly?

szuecs: 2 reads would be fine but later it will surely write/replace the object.

szuecs: Ah no it's read only 😀

MustafaSaber:
👍

AlexanderYastrebov:
:+1:

PR #3281, opened 2024-10-18
skipper: add TLS client authentication config

AlexanderYastrebov:
Add an option and a flag to configure TLS Client Authentication policy of the Server.

Fixes https://github.com/zalando/skipper/issues/3280

PR #3282, opened 2024-10-19
skipper: set server minimal TLS version

AlexanderYastrebov:
The tls-min-version flag specifies "minimal TLS Version to be used in server, proxy and client connections" but was not used to configure Server TLS.

Follow up on: #1656


AlexanderYastrebov:
Notices while working on #3281

RomanZavodskikh:
:+1:

MustafaSaber:
👍

szuecs:
:+1:

PR #3287, opened 2024-10-26
Added check to handle token parsing with claims without "sub".

### LINKED ISSUE #3216: oidcclaimquery: better error handling for casting
**Describe the bug**

if `oidcclaimsquery` get a jwt token that doesn't have `sub` for any reason it will fail to cast `nil` to string.

https://github.com/zalando/skipper/blob/00ff74258194a4292e389b00979fc2290a085a91/filters/auth/oidc_introspection.go#L134

**To Reproduce**

use a token without `sub`

**Expected behavior**

Better error handling instead of failing without clue what is happening.

wassafshahzad:
### Description 
In the oidc_introspection.py `func (filter *oidcIntrospectionFilter) Request(ctx filters.FilterContext)` function get value form the Claim map using `sub` key and then casts the value into a string, but this would panic if no sub key is present. I added a check to ensure we only cast values if sub key is present.

### Linked Issue 
closes #3216 

### Approach to solution
As describe in the following [comment](https://github.com/zalando/skipper/issues/3216#issuecomment-2433498646), I dug a bit deeper and found 2 suitable values to use if subject is not present in token. One is the Subject value in tokenContainer struct or UserInfo Struct. I decided to go with the tokenContainer struct. If this is not correct please do guide me  if UserInfo Struct would be better.


szuecs:
Can you please add a test case?
Thanks 🙏🏽

wassafshahzad:
> Can you please add a test case? Thanks 🙏🏽

Could you give me some pointers, I have been looking into multiple filter tests and oidc_intorspection_tests but cant wrap my head around on how to test it.

wassafshahzad:
> hi @wassafshahzad, thanks for your contribution! You can check
> 
> https://github.com/zalando/skipper/blob/8e6c3651874a24ddddcd4d8ab4cb3f5e35bb00e9/filters/auth/oidc_test.go#L224
> 
> 
> &
> https://github.com/zalando/skipper/blob/8e6c3651874a24ddddcd4d8ab4cb3f5e35bb00e9/filters/auth/oidc_introspection_test.go#L295
> 
> This server creates the token with those claims, I would try to find a way to override those claims and create another testcase where I create the server with the override and this case we expect the fail before the fix and succeed after.
> 
> The test case should fail or succeed after triggering
> 
> https://github.com/zalando/skipper/blob/8e6c3651874a24ddddcd4d8ab4cb3f5e35bb00e9/filters/auth/oidc_introspection_test.go#L355

Thank you and on it

wassafshahzad:
@MustafaSaber Sorry for the delay, I had a bit of a tough time resolving this issue. I tried using `token.Subject` and `token.UserInfo.Subject` in case sub was missing but this would still cause authentication issues. I still need help to identify the right key as I am not as familiar with the repo yet.

However I stumbled on to this [piece of code ](https://github.com/zalando/skipper/blob/f6f9efb2c7ff93ff940b6f5d60fda3e89e89cfec/filters/auth/jwt_validation.go#L148) from the jwt_validation file and decided to use it since we already have it for validation and thought consistency would be the right way to go.

MustafaSaber:
Thanks for the update! We will review when we have time, we can't merge in the next couple of days also.

wassafshahzad:
> Thanks for the update! We will review when we have time, we can't merge in the next couple of days also.

No issue, Can I pick another issue ?

MustafaSaber:
hi @wassafshahzad, can you rebase?

wassafshahzad:
> hi @wassafshahzad, can you rebase?

Yeah sure no problem

filters/auth/oidc_introspection.go
```diff
@@ -131,7 +131,12 @@ func (filter *oidcIntrospectionFilter) Request(ctx filters.FilterContext) {
 		return
 	}
 
-	sub := token.Claims["sub"].(string)
+	sub, ok := token.Claims["sub"].(string)
+	if !ok {
+		unauthorized(ctx, sub, invalidSub, "", "")
```
AlexanderYastrebov, line 136: Here the sub will be empty? Maybe then pass empty value (`""`) or a sentinel value.

filters/auth/oidc_introspection_test.go
```diff
@@ -165,6 +166,17 @@ func TestOIDCQueryClaimsFilter(t *testing.T) {
 			expected:  200,
 			expectErr: false,
 		},
+		{
+			msg: "secure sub/path is not permitted",
```
AlexanderYastrebov, line 170: "missing sub claim is not permitted"

filters/auth/oidc_introspection.go
AlexanderYastrebov: I think we can also pass r.Host like in the invocations above.

AlexanderYastrebov:
As a panic fix it looks good to me (with minor comments).

wassafshahzad:
@AlexanderYastrebov  Sorry for the delay was on Vacation but have addressed feedback

filters/auth/oidc_introspection.go
```diff
@@ -139,7 +139,12 @@ func (filter *oidcIntrospectionFilter) Request(ctx filters.FilterContext) {
 		return
 	}
 
-	sub := token.Claims["sub"].(string)
+	sub, ok := token.Claims["sub"].(string)
+	if !ok {
+		unauthorized(ctx, r.Host, invalidSub, "", "")
```
AlexanderYastrebov, line 144: Looking at the function signature
https://github.com/zalando/skipper/blob/e819c038999358a9a79514bf1de11f7132079369/filters/auth/auth.go#L136

used arguments do not look right.

szuecs:
:+1:

MustafaSaber:
👍

PR #3296, opened 2024-11-01
host2regex: doesn't take in consideration `*`

MustafaSaber:
converting host from Ingress/RouteGroup for `Host` predicate doesn't take into consideration that Ingress support wildcard hostnames (see https://kubernetes.io/docs/concepts/services-networking/ingress/#hostname-wildcards) and we produce invalid regex

This PR adds testcases to reproduce the error, **requires fix** before merging.

following ingress produce

```yaml
kind: Ingress
metadata:
  name: simple
  labels:
    name: simple
spec:
  rules:
  - host: "*.example.org"
    http:
      paths:
      - pathType: Prefix
        path: "/"
        backend:
          service:
            name: simple
            port: 
              number: 80
```

Error:
```
[APP]time="2024-11-01T12:43:46Z" level=error msg="kube_default__simple_redirect_app_default_0____example_org_____ [2]: error parsing regexp: missing argument to repetition operator: *
```

Test with

```bash
➜ ./bin/skipper -inline-routes 'r0: Host("*.mu.sa") -> inlineContent("hi from 0") -> <shunt>; r1: Host("ab.mu.sa") -> inlineContent("hi from 1") -> <shunt>;' -address=:8080

➜ curl -i -H "Host: ba.mu.sa" http://127.0.0.1:8080
HTTP/1.1 200 OK
Content-Length: 9
Server: Skipper
Date: Tue, 03 Jun 2025 11:18:23 GMT

hi from 0%

➜ curl -i -H "Host: ab.mu.sa" http://127.0.0.1:8080
HTTP/1.1 200 OK
Content-Length: 9
Server: Skipper
Date: Tue, 03 Jun 2025 11:18:28 GMT

hi from 1%
```

fixes https://github.com/zalando/skipper/issues/3297

### DIFF
--- a/dataclients/kubernetes/hosts.go
+++ b/dataclients/kubernetes/hosts.go
@@ -5,6 +5,7 @@ import (
 	"regexp"
 	"strings"
 
+	log "github.com/sirupsen/logrus"
 	"github.com/zalando/skipper/eskip"
 )
 
@@ -15,6 +16,10 @@ func createHostRx(hosts ...string) string {
 
 	hrx := make([]string, len(hosts))
 	for i, host := range hosts {
+		if strings.HasPrefix(host, "*.") {
+			log.Debugf("Host %q starts with '*.'; replacing with regex", host)
+			host = strings.Replace(host, "*", "[a-z0-9]+((-[a-z0-9]+)?)*", 1)
+		}
 		// trailing dots and port are not allowed in kube
 		// ingress spec, so we can append optional setting
 		// without check
--- a/dataclients/kubernetes/hosts_test.go
+++ b/dataclients/kubernetes/hosts_test.go
@@ -1,0 +1,31 @@
+package kubernetes
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestHostsToRegex(t *testing.T) {
+	for _, ti := range []struct {
+		msg   string
+		host  string
+		regex string
+	}{
+		{
+			msg:   "simple",
+			host:  "simple.example.org",
+			regex: "^(simple[.]example[.]org[.]?(:[0-9]+)?)$",
+		},
+		{
+			msg:   "wildcard",
+			host:  "*.example.org",
+			regex: "^([a-z0-9]+((-[a-z0-9]+)?)*[.]example[.]org[.]?(:[0-9]+)?)$",
+		},
+	} {
+		t.Run(ti.msg, func(t *testing.T) {
+			regex := createHostRx(ti.host)
+			require.Equal(t, ti.regex, regex)
+		})
+	}
+}
--- a/dataclients/kubernetes/testdata/ingressV1/ingress-data/wildcard-ing-prefix.eskip
+++ b/dataclients/kubernetes/testdata/ingressV1/ingress-data/wildcard-ing-prefix.eskip
@@ -1,0 +1,3 @@
+kube_foo__qux____example_org_____qux:
+	Host("^([a-z0-9]+((-[a-z0-9]+)?)*[.]example[.]org[.]?(:[0-9]+)?)$") && PathSubtree("/")
+	-> <roundRobin, "http://10.2.9.103:8080", "http://10.2.9.104:8080">;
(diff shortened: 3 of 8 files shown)


MustafaSaber:
maybe it's not a good idea to have this at all

dataclients/kubernetes/hosts_test.go
```diff
@@ -0,0 +1,31 @@
...
+			regex: "^(simple[.]example[.]org[.]?(:[0-9]+)?)$",
+		},
+		{
+			msg:   "wildcard",
+			host:  "*.example.org",
+			regex: "^([a-z0-9]+(-[a-z0-9]+)?[.]example[.]org[.]?(:[0-9]+)?)$",
+		},
+	} {
+		t.Run(ti.msg, func(t *testing.T) {
+			regex := createHostRx(ti.host)
```
MustafaSaber, line 27: another idea is to update validation webhook to say if the host regex is valid

dataclients/kubernetes/testdata/ingressV1/ingress-data/wildcard-ing-prefix.eskip
```diff
@@ -0,0 +1,3 @@
+kube_foo__qux____example_org_____qux:
+	Host("^([a-z0-9]+(-[a-z0-9]+)?[.]example[.]org[.]?(:[0-9]+)?)$") && PathSubtree("/")
```
szuecs, line 2: I think you want 
`^([a-z0-9][a-z0-9-]*)?[.]example....`

Better also to write down in the issue what kind of match or regexp you want to create.

MustafaSaber: I want to match a valid hostname, it can't end or start with `-` I think this `([a-z0-9][a-z0-9-]*)` can end with `-`

MustafaSaber:
See https://github.com/zalando/skipper/issues/3297#issuecomment-2454827352

dataclients/kubernetes/testdata/ingressV1/ingress-data/wildcard-ing-prefix.eskip
szuecs: What about foo-bar-qux.example , that your regexp doesn't match?

MustafaSaber: updated the PR to match it, the new regex `[a-z0-9]+((-[a-z0-9]+)?)*`

dataclients/kubernetes/hosts.go
```diff
@@ -15,6 +15,9 @@ func createHostRx(hosts ...string) string {
 
 	hrx := make([]string, len(hosts))
 	for i, host := range hosts {
+		if strings.HasPrefix(host, "*.") {
+			host = strings.Replace(host, "*", "[a-z0-9]+((-[a-z0-9]+)?)*", 1)
```
AlexanderYastrebov, line 19: We need to verify this matches behaviour descibed at https://kubernetes.io/docs/concepts/services-networking/ingress/#hostname-wildcards

Makes sense to add this link as a comment.

Subdomain regexp looks strange with two capture groups using ? followed by *.

I guess subdomain shoud start and end with alphanumeric (not dash). Lets check this https://stackoverflow.com/questions/7930751/regexp-for-subdomain and similar answers and reuse/adopt regexp from there.

szuecs: Sounds like https://github.com/zalando/skipper/pull/3296#discussion_r1829017617 would be the one 😀

MustafaSaber: I think this one already works for all cases.
image

MustafaSaber: > Sounds like #3296 (comment) would be the one

yes this also work

Jan-M:
Can this advance, e.g. with a feature toggle to support it? I would say potential conflicts can also be ruled user error and not skipper error.

What would be needed if you wanted to elevate the support? Weights? Warnings?

PR #3303, opened 2024-11-08
feature: allow overriding tls config based on client hello

### LINKED ISSUE #3295: Override TLS Client Auth (potentially other TLSOptions) per Ingress/Route
With the recently implemented TLS Client Auth config in #3281 we are restricted to setting the TLS Client Option to all routes skipper is handling. This unfortunately prevents us from using since we do not want to have browsers request a client auth certificate for all routes.

An ingress annotation and/or route group CRD parameter to enable TLS Client Auth per route.

None.

The traefik project allows per ingress changes to [TLS Options](https://doc.traefik.io/traefik/https/tls/#tls-options) including TLS CLient Auth using ingress annotations. 

> To explicitly use a different TLSOption (and using the Kubernetes Ingress resources) you'll have to add an annotation to the Ingress in the following form: `traefik.ingress.kubernetes.io/router.tls.options: <resource-namespace>-<resource-name>@kubernetescrd`

Yes, but no time

rickhlx:
Fixes #3295

This feature allows setting a custom TLS config based on the client's request. It uses tls [GetConfigForClient](https://pkg.go.dev/crypto/tls#Config.GetConfigForClient) to obtain the correct config for the specific client's hello.

This feature allows the use of ingress annotations or other methods to add custom configs per routes. 

This is especially useful for the issue being fixed where we would like to setup client auth per ingress and/or routegroup.

### DIFF
--- a/dataclients/kubernetes/kube.go
+++ b/dataclients/kubernetes/kube.go
@@ -623,25 +623,16 @@ func compareStringList(a, b []string) []string {
 	return c
 }
 
-// addTLSCertToRegistry adds a TLS certificate to the certificate registry per host using the provided
-// Kubernetes TLS secret
-func addTLSCertToRegistry(cr *certregistry.CertRegistry, logger *logger, hosts []string, secret *secret) {
+// addTLSConfigToRegistry adds a TLS Config to the registry per host using the provided config and secret.
+func addTLSConfigToRegistry(cr *certregistry.CertRegistry, logger *logger, hosts []string, config *certregistry.Config, secret *secret) {
 	cert, err := generateTLSCertFromSecret(secret)
 	if err != nil {
 		logger.Errorf("Failed to generate TLS certificate from secret: %v", err)
 		return
 	}
 	for _, host := range hosts {
-		err := cr.ConfigureCertificate(host, cert)
-		if err != nil {
-			logger.Errorf("Failed to configure certificate: %v", err)
-		}
-	}
-}
-
-func addTLSConfigToRegistry(cr *certregistry.CertRegistry, logger *logger, hosts []string, tlsConfig *tls.Config) {
-	for _, host := range hosts {
-		err := cr.ConfigureTLSConfig(host, tlsConfig)
+		config.Certificate = *cert
+		err := cr.SetTLSConfig(host, config)
 		if err != nil {
 			logger.Errorf("Failed to configure TLS config: %v", err)
 		}
--- a/secrets/certregistry/certregistry.go
+++ b/secrets/certregistry/certregistry.go
@@ -1,6 +1,7 @@
 package certregistry
 
 import (
+	"bytes"
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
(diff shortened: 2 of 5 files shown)


rickhlx:
@szuecs could we get some early feedback? We'd like to know if implementing this in the CertificateRegistry seems like a good idea.

skipper.go
```diff
@@ -1211,6 +1211,7 @@ func (o *Options) tlsConfig(cr *certregistry.CertRegistry) (*tls.Config, error)
 
 	if cr != nil {
 		config.GetCertificate = cr.GetCertFromHello
+		config.GetConfigForClient = cr.GetConfigFromHello
```
rickhlx, line 1214: problem: the `tls.Config` from `GetConfigForClient` will override the previously used `GetCertificate`.  So the two are mutually exclusive, from the docs:

> Server configurations must set one of Certificates, GetCertificate or GetConfigForClient.

szuecs: Check if the new one is non nil and if so only use that one, else fallback to current behavior.
Does it make sense?

rickhlx: Makes sense.

I was also thinking we could also use `GetConfigForClient` only. This would allow us to use custom certs and custom TLS config. Config could be set to a "default" based on global skipper flags allowing it to be overridden. Wdyt?

dataclients/kubernetes/kube.go
```diff
@@ -383,6 +384,25 @@ func ParsePathMode(s string) (PathMode, error) {
 	}
 }
 
+// ParseTLSClientAuth parses the string representations of different
+// client auth types.
+func ParseTLSClientAuth(s string) (tls.ClientAuthType, error) {
+	switch s {
```
universam1, line 390:
```suggestion
	switch strings.ToLower(s) {
```
just an idea to support casing as well

secrets/certregistry/certregistry.go
```diff
@@ -9,62 +10,114 @@ import (
...
-			log.Infof("updating certificate in registry - %s", host)
-			r.lookup[host] = cert
-			return nil
-		} else {
+	if found && bytes.Equal(curr.hash, configHash) {
+		return nil
+	}
+
+	if found && !bytes.Equal(curr.hash, configHash) {
+		if !config.Certificate.Leaf.NotBefore.After(curr.config.Certificates[0].Leaf.NotBefore) {
```
rickhlx, line 71: problem: this causes config to be skipped when separate ingresses have different annotations. We'd need to set the annotation for all hosts regardless of the ingress.

secrets/certregistry/certregistry.go
```diff
@@ -9,62 +10,114 @@ import (
...
-		if cert.Leaf.NotBefore.After(curr.Leaf.NotBefore) {
-			log.Infof("updating certificate in registry - %s", host)
-			r.lookup[host] = cert
-			return nil
-		} else {
+	if found && bytes.Equal(curr.hash, configHash) {
+		return nil
+	}
+
+	if found && !bytes.Equal(curr.hash, configHash) {
```
szuecs, line 70: it's a duplicate check for `found` so maybe reafctor to have only one check for `found`.

secrets/certregistry/certregistry.go
szuecs: I am not sure why you compare the "NotBefore" values. 
As far as I understand `config.Certificate` is the global default and `curr.config.Certificates[0]` is the looked up one. So here we compare if default is not newer than the looked up one.
Should not we match the "better" one?
For example better match like `*.foo.example` vs `www.foo.example` ?

skipper.go
szuecs: I have no strong opinion on that onre

rickhlx:
Closing since I can't continue working and has been stale for a while.

PR #3311, opened 2024-11-15
oid_introspection: add SetOIDCClaims

### LINKED ISSUE #3139: Allow third-party filters to set the state for oidcClaimsQuery
The oidcClaimsQuery filter allows for arbitrary queries on claims using GJSON. Besides being able to be used for OIDC, other filters in Skipper that handle different authentication schemes also set the OIDC claims cache key allowing the oidcClaimsQuery filter to be called.
However, when working with a third-party filter, it is not possible to create a tokenContainer value because this type is private to the filters/auth package. Therefore it is not possible to set the OIDC claims cache key to a value that would be accepted by oidcClaimsQuery, since the filter expects a value of type tokenContainer to be stored there.
This prevents the reuse of oidcClaimsQuery when the authentication is done in a third-party filter.

A function should be exported that sets the OIDC claims cache key (given a `map[string]interface{}` claims value).

An alternative would be to export the tokenContainer type.

Yes

haveo:
This method allows third-party filters to set the oidcClaimsCacheKey which enables the use of the oidcClaimsQuery filter.

Closes #3139

### DIFF
--- a/filters/auth/grant.go
+++ b/filters/auth/grant.go
@@ -175,11 +175,7 @@ func (f *grantFilter) setupToken(token *oauth2.Token, tokeninfo map[string]inter
 
 	// By piggy-backing on the OIDC token container,
 	// we gain downstream compatibility with the oidcClaimsQuery filter.
-	ctx.StateBag()[oidcClaimsCacheKey] = tokenContainer{
-		OAuth2Token: token,
-		Subject:     subject,
-		Claims:      tokeninfo,
-	}
+	SetOIDCClaims(ctx, tokeninfo)
 
 	// Set the tokeninfo also in the tokeninfoCacheKey state bag, so we
 	// can reuse e.g. the forwardToken() filter.
--- a/filters/auth/oidc_introspection.go
+++ b/filters/auth/oidc_introspection.go
@@ -42,6 +42,14 @@ func NewOIDCQueryClaimsFilter() filters.Spec {
 	}
 }
 
+// Sets OIDC claims in the state bag.
+// Intended for use with the oidcClaimsQuery filter.
+func SetOIDCClaims(ctx filters.FilterContext, claims map[string]interface{}) {
+	ctx.StateBag()[oidcClaimsCacheKey] = tokenContainer{
+		Claims: claims,
+	}
+}
+
 func (spec *oidcIntrospectionSpec) Name() string {
 	switch spec.typ {
 	case checkOIDCQueryClaims:
--- a/filters/auth/grant_test.go
+++ b/filters/auth/grant_test.go
@@ -210,6 +210,7 @@ func newAuthProxy(t *testing.T, config *auth.OAuthConfig, routes []*eskip.Route,
 	fr.Register(config.NewGrantCallback())
 	fr.Register(config.NewGrantClaimsQuery())
 	fr.Register(config.NewGrantLogout())
+	fr.Register(auth.NewOIDCQueryClaimsFilter())
 
 	pc := proxytest.Config{
 		RoutingOptions: routing.Options{
@@ -331,6 +332,7 @@ func TestGrantFlow(t *testing.T) {
 	config := newGrantTestConfig(tokeninfo.URL, provider.URL)
 
 	routes := eskip.MustParse(`* -> oauthGrant()
+		-> oidcClaimsQuery("/:sub")
 		-> status(204)
 		-> setResponseHeader("Backend-Request-Cookie", "${request.header.Cookie}")
 		-> <shunt>


filters/auth/oidc_introspection.go
```diff
@@ -42,6 +42,12 @@ func NewOIDCQueryClaimsFilter() filters.Spec {
 	}
 }
 
+func MakeTokenContainer(h map[string]interface{}) tokenContainer {
```
AlexanderYastrebov, line 45: Its an exported method that returns unexpertor type - not good.

The original issue was:
> However, when working with a third-party filter, it is not possible to create a tokenContainer value because this type is private to the filters/auth package. Therefore it is not possible to set the OIDC claims cache key to a value that would be accepted by oidcClaimsQuery, since the filter expects a value of type tokenContainer to be stored there.
This prevents the reuse of oidcClaimsQuery when the authentication is done in a third-party filter.

and proposed solution:
> A function should be exported that sets the OIDC claims cache key (given a map[string]interface{} claims value).

But this change does not implement the proposal - user of MakeTokenContainer still needs to rely on a known value of statebag key.

I think the proper change should provide an utility to store token container into statebag, it should be documented and existing code should be changed to use it.

haveo: "existing code should be changed to use it" do you mean the auth filters in skipper that store the token container in the statebag should use this new method? They wouldn't be able to do that with the proposed interface since they store more than just claims. But on the other hand it doesn't seem like any field other than claims is used when reading from the statebag.

AlexanderYastrebov:
:+1:

filters/auth/oidc_introspection.go
AlexanderYastrebov: Ok, lets have it like this.

filters/auth/oidc_introspection.go
```diff
@@ -42,6 +42,12 @@ func NewOIDCQueryClaimsFilter() filters.Spec {
 	}
 }
 
+func SetOIDCClaims(ctx filters.FilterContext, h map[string]interface{}) {
```
AlexanderYastrebov, line 45: Please add godoc comment, squash and rebase.

AlexanderYastrebov: h -> claims

filters/auth/oidc.go
```diff
@@ -843,7 +843,8 @@ func (f *tokenOidcFilter) Request(ctx filters.FilterContext) {
 	}
 
 	// saving token info for chained filter
-	ctx.StateBag()[oidcClaimsCacheKey] = container
+	// ctx.StateBag()[oidcClaimsCacheKey] = container
+	SetOIDCClaims(ctx, container.Claims)
```
AlexanderYastrebov, line 847: We can not do this because container is used to unmarshal json above which also contains other than claims fields.

filters/auth/oidc_introspection.go
```diff
@@ -42,6 +42,14 @@ func NewOIDCQueryClaimsFilter() filters.Spec {
 	}
 }
 
+// Sets OIDC claims in the state bag.
+// Intended for use with the oidcClaimsQuery filter.
```
AlexanderYastrebov, line 46: Hm, the filter also uses sub claim from the container https://github.com/zalando/skipper/blob/44139f3a87d0fb455d9564af5207546f25ee622a/filters/auth/oidc_introspection.go#L123-L136

AlexanderYastrebov: See also related https://github.com/zalando/skipper/pull/3287

filters/auth/oidc.go
haveo: Sorry this is just a leftover from me trying to get existing tests to validate that SetOIDCClaims works correctly. I'll revert that.

filters/auth/oidc_introspection.go
haveo: I see. I can add a check and set the sub key to empty if it's not a string value.

haveo: I'm not sure why this filter is the one setting the auth-user key (via authorized) though. I feel like it should be up to whatever filter obtained the credentials to begin with.

AlexanderYastrebov:
:+1:

filters/auth/oidc.go
MustafaSaber: > Sorry this is just a leftover from me trying to get existing tests to validate that SetOIDCClaims works correctly. I'll revert that.

Maybe it's better to add a standalone testcase?

haveo: I modified the oauthGrant filter to use SetOIDCClaims and added a use of oidcClaimsQuery in grant_test.go.

MustafaSaber:
👍

filters/auth/grant.go
```diff
@@ -175,11 +175,7 @@ func (f *grantFilter) setupToken(token *oauth2.Token, tokeninfo map[string]inter
 
 	// By piggy-backing on the OIDC token container,
 	// we gain downstream compatibility with the oidcClaimsQuery filter.
-	ctx.StateBag()[oidcClaimsCacheKey] = tokenContainer{
-		OAuth2Token: token,
-		Subject:     subject,
-		Claims:      tokeninfo,
-	}
+	SetOIDCClaims(ctx, tokeninfo)
```
MustafaSaber, line 178: but this won't set OAuth2Token & Subject, right?

haveo: They're not used by `oidcClaimsQuery` anyway.

AlexanderYastrebov:
:+1:

MustafaSaber:
👍

PR #3316, opened 2024-11-21
Add unit tests for HTTP request to Envoy request conversion

Pushpalanka:
Add unit tests for HTTP request to Envoy request conversion and to test the invalid UTF-8 character validation method.


filters/openpolicyagent/internal/envoy/skipperadapter_test.go
```diff
@@ -0,0 +1,199 @@
...
+					}),
+				},
+			},
+			wantErr: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := AdaptToExtAuthRequest(tt.req, tt.metadata, tt.contextExtensions, tt.rawBody)
```
mjungsbluth, line 81: I think it would make more sense to change the tests to convert this as well to an AST input and validate that one. The Envoy request object itself is sth intermediate. 

Looking forward we could also remove the double serialisation / deserialisation of Skipper Request -> Envoy Protobuf -> JSON -> OPA AST without changing the tests.

szuecs: I agree!
Normally this should be part of some bdd style test with invalid input. If it is too complicated a unit test is fine.

Issue #3317, opened 2024-11-22
Add more visibility into admissionControl span tag


RomanZavodskikh:
**Is your feature request related to a problem? Please describe.**
Some visibility of admissionControl is missing, for example, it is impossible to filter all spans for specific admissionController filter at the moment.

**Describe the solution you would like**
Add additional span tag(s) in this span
https://github.com/zalando/skipper/blob/v0.21.223/filters/shedder/admission.go#L413

**Would you like to work on it?**
Yes

PR #1964, opened 2022-02-21
filters/scheduler: wait for spans finishing in lifo errors test

AlexanderYastrebov:
The tests was introduced by #1944
It may happen that not all spans are finished by the time test verifies response results (because spans are closed after response has been served).
This change waits for all mock spans to finish and forces scheduler metrics update.

Signed-off-by: Alexander Yastrebov <alexander.yastrebov@zalando.de>

PR #3318, opened 2024-11-22
Add more tags to the admissionControl span to increase visibility

RomanZavodskikh:
Updates https://github.com/zalando/skipper/issues/3317

### DIFF
--- a/filters/shedder/admission.go
+++ b/filters/shedder/admission.go
@@ -370,6 +370,13 @@ func sum(a []int64) int64 {
 	return result
 }
 
+func (ac *admissionControl) setCommonTags(span opentracing.Span) {
+	span.SetTag("admissionControl.group", ac.metricSuffix)
+	span.SetTag("admissionControl.mode", ac.mode.String())
+	span.SetTag("admissionControl.duration", ac.d.String())
+	span.SetTag("admissionControl.windowSize", ac.windowSize)
+}
+
 // calculates P_{reject} see https://opensource.zalando.com/skipper/reference/filters/#admissioncontrol
 func (ac *admissionControl) pReject() float64 {
 	var rejectP float64
@@ -412,6 +419,8 @@ func (ac *admissionControl) shouldReject() bool {
 func (ac *admissionControl) Request(ctx filters.FilterContext) {
 	span := ac.startSpan(ctx.Request().Context())
 	defer span.Finish()
+
+	ac.setCommonTags(span)
 	ac.metrics.IncCounter(counterPrefix + "total." + ac.metricSuffix)
 
 	if ac.shouldReject() {
--- a/filters/shedder/admission_test.go
+++ b/filters/shedder/admission_test.go
@@ -626,8 +626,6 @@ func TestAdmissionControlSetsSpansTags(t *testing.T) {
 
 	fr := make(filters.Registry)
 	fr.Register(fspec)
-	fr.Register(builtin.NewSetRequestHeader())
-	fr.Register(builtin.NewStatus())
 
 	dc := testdataclient.New([]*eskip.Route{r})
 	defer dc.Close()
--- a/filters/shedder/admission_test.go
+++ b/filters/shedder/admission_test.go
@@ -13,15 +13,18 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/filters"
 	"github.com/zalando/skipper/filters/builtin"
 	"github.com/zalando/skipper/metrics"
 	"github.com/zalando/skipper/metrics/metricstest"
 	"github.com/zalando/skipper/net"
+	"github.com/zalando/skipper/proxy"
 	"github.com/zalando/skipper/proxy/proxytest"
 	"github.com/zalando/skipper/routing"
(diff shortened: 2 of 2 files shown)


filters/shedder/admission_test.go
```diff
@@ -594,6 +598,127 @@ func TestAdmissionControlCleanupMultipleFilters(t *testing.T) {
 	}
 }
 
+type testTracer struct {
```
AlexanderYastrebov, line 601: This looks borrowed from https://github.com/zalando/skipper/pull/1964

Lets think ho we can extract it into e.g. tracing/tracingtest package

RomanZavodskikh: Yes, it was copied from there.

RomanZavodskikh: Addressed in this commit, see https://github.com/zalando/skipper/pull/3318/commits/bdea461506235c79299994c135a632cd35902862 .

tracing/tracingtest/mocktracer.go
```diff
@@ -0,0 +1,41 @@
...
+	"time"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/mocktracer"
+)
+
+type MockTracer struct {
+	*mocktracer.MockTracer
+	spans int32
+}
```
AlexanderYastrebov, line 14: Here we can use https://pkg.go.dev/sync/atomic#Int32 to count spans.

I think it also makes sense to add a constructor that would init *mocktracer.MockTracer to simplify caller side to:
```go
tracer := tracingtest.NewTracer()
```

tracing/tracingtest/mocktracer.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"sync/atomic"
+	"time"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/mocktracer"
+)
+
+type MockTracer struct {
+	*mocktracer.MockTracer
```
AlexanderYastrebov, line 12: Do we need this field to be exported?

filters/scheduler/fifo_test.go
```diff
@@ -296,7 +297,7 @@ func TestFifoWithBody(t *testing.T) {
 			rt := routing.New(ro)
 			defer rt.Close()
 			<-rt.FirstLoad()
-			tracer := &testTracer{MockTracer: mocktracer.New()}
+			tracer := &tracingtest.MockTracer{MockTracer: mocktracer.New()}
```
AlexanderYastrebov, line 300: Lets check all callsites using `mocktracer.New()` - I think we have multiple various ways to wait for finished spans.
Replacing them with this new tracingtest.NewTracer should simplify tests.

RomanZavodskikh: Actually, three tracers created and were not used, I have removed them: https://github.com/zalando/skipper/pull/3322/files#diff-703dc4a90e39e2396aa495f14a94d9132ffc097b1be0cb2c01fb244d116289a1L299-L642


The only one left is using FinishedSpans() method of MockTracer in an expected way, no need to change that.

tracing/tracingtest/mocktracer.go
RomanZavodskikh: Fixed there https://github.com/zalando/skipper/pull/3322 .

tracing/tracingtest/mocktracer.go
RomanZavodskikh: Actually, this was a good idea because we need to explicitly code every method of Tracer interface. https://pkg.go.dev/github.com/opentracing/opentracing-go#Tracer

I have defined couple more of them, it's not a big deal at all, but it makes the calls much more understandable: we know which exactly functions are called by mock tests.
https://github.com/zalando/skipper/pull/3322/files#diff-3f5e5fd69ecb51b0c383da632e30fe4bd12ffb855d983235b063b70d0a17ac00R47-R53

filters/scheduler/fifo_test.go
RomanZavodskikh: After this change was made https://github.com/zalando/skipper/pull/3318#discussion_r1855180957 we can be sure only this FinishedSpans() method is called https://github.com/zalando/skipper/pull/3322/files#diff-3f5e5fd69ecb51b0c383da632e30fe4bd12ffb855d983235b063b70d0a17ac00R30-R45 .

MustafaSaber:
Can we rebase, remove the do-not-merge label and push this forward?

filters/shedder/admission.go
```diff
@@ -370,6 +370,13 @@ func sum(a []int64) int64 {
 	return result
 }
 
+func (ac *admissionControl) setCommonTags(span opentracing.Span) {
+	span.SetTag("admissionControl.group", ac.metricSuffix)
+	span.SetTag("admissionControl.mode", ac.mode.String())
+	span.SetTag("admissionControl.d", ac.d.String())
```
MustafaSaber, line 376: what is d? I think the tag in the span should be a bit descriptive.

RomanZavodskikh: It's actually the duration during which admissionControll count requests amount. I have renamed "d" to "duration".

RomanZavodskikh:
:+1:

PR #3328, opened 2024-12-04
dataclients/kubernetes: append predicates to routes of Ingress/RouteGroup having specific annotation

MustafaSaber:
The main purpose of this change is to allow skipper operators to dynamically add predicates to routes based on annotations key-value pairs in Kubernetes resources (metadata.annotation). This provides more flexibility in routing configuration without modifying the core Ingress or RouteGroup definitions.

Key changes:
- Added KubernetesAnnotationPredicates option & `-kubernetes-annotation-predicates` flag for public hostnames, which can be specified multiple times.
- Added KubernetesEastWestRangeAnnotationPredicates option & `-kubernetes-east-west-range-annotation-predicates` flag for internal hostnames, which can be specified multiple times.
- Modified Ingress and RouteGroup clients to apply annotation predicates depending on the host
- Added test cases for new functionality

Run & test with

```sh

➜ ./bin/skipper -kubernetes -kubernetes-annotation-predicates="zalando.org/traffic-disabled=true=False()" -kubernetes-east-west-range-annotation-predicates="zalando.org/traffic-disabled=true=False()"
➜ #or
➜ ./bin/routesrv -kubernetes -kubernetes-annotation-predicates="zalando.org/traffic-disabled=true=False()" -kubernetes-east-west-range-annotation-predicates="zalando.org/traffic-disabled=true=False()"
```

Having both flags with same value will append configured predicates to all routes of Ingress/RouteGroup resources annotated with `zalando.org/traffic-disabled: true`, routes defined in:

* The `spec.rules` section of Ingress resources.
* Routes specified using the `zalando.org/skipper-routes` annotation of Ingress resources.
* The `spec.routes` section of RouteGroup resources.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -149,34 +149,38 @@ type Config struct {
 	RefusePayload    multiFlag `yaml:"refuse-payload"`
 
 	// Kubernetes:
-	KubernetesIngress                       bool                               `yaml:"kubernetes"`
-	KubernetesInCluster                     bool                               `yaml:"kubernetes-in-cluster"`
-	KubernetesURL                           string                             `yaml:"kubernetes-url"`
-	KubernetesTokenFile                     string                             `yaml:"kubernetes-token-file"`
-	KubernetesHealthcheck                   bool                               `yaml:"kubernetes-healthcheck"`
-	KubernetesHTTPSRedirect                 bool                               `yaml:"kubernetes-https-redirect"`
-	KubernetesHTTPSRedirectCode             int                                `yaml:"kubernetes-https-redirect-code"`
-	KubernetesDisableCatchAllRoutes         bool                               `yaml:"kubernetes-disable-catchall-routes"`
-	KubernetesIngressClass                  string                             `yaml:"kubernetes-ingress-class"`
-	KubernetesRouteGroupClass               string                             `yaml:"kubernetes-routegroup-class"`
-	WhitelistedHealthCheckCIDR              string                             `yaml:"whitelisted-healthcheck-cidr"`
-	KubernetesPathModeString                string                             `yaml:"kubernetes-path-mode"`
-	KubernetesPathMode                      kubernetes.PathMode                `yaml:"-"`
-	KubernetesNamespace                     string                             `yaml:"kubernetes-namespace"`
-	KubernetesEnableEndpointSlices          bool                               `yaml:"enable-kubernetes-endpointslices"`
-	KubernetesEnableEastWest                bool                               `yaml:"enable-kubernetes-east-west"`
-	KubernetesEastWestDomain                string                             `yaml:"kubernetes-east-west-domain"`
-	KubernetesEastWestRangeDomains          *listFlag                          `yaml:"kubernetes-east-west-range-domains"`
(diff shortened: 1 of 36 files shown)


skipper.go
```diff
@@ -963,6 +970,7 @@ func (o *Options) KubernetesDataClientOptions() kubernetes.Options {
 		KubernetesEastWestDomain:          o.KubernetesEastWestDomain,
 		KubernetesEastWestRangeDomains:    o.KubernetesEastWestRangeDomains,
 		KubernetesEastWestRangePredicates: o.KubernetesEastWestRangePredicates,
+		KubernetesAnnotationPredicates:    o.KubernetesAnnotationPredicates,
```
AlexanderYastrebov, line 973: Should we second field here as well?

config/config.go
```diff
@@ -470,6 +474,8 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.KubernetesEastWestDomain, "kubernetes-east-west-domain", "", "*Deprecated*: use kubernetes-east-west-range feature. Sets the east-west domain, defaults to .skipper.cluster.local")
 	flag.Var(cfg.KubernetesEastWestRangeDomains, "kubernetes-east-west-range-domains", "set the the cluster internal domains for east west traffic. Identified routes to such domains will include the -kubernetes-east-west-range-predicates")
 	flag.StringVar(&cfg.KubernetesEastWestRangePredicatesString, "kubernetes-east-west-range-predicates", "", "set the predicates that will be appended to routes identified as to -kubernetes-east-west-range-domains")
+	flag.Var(&cfg.KubernetesAnnotationPredicatesString, "kubernetes-annotation-predicates", "predicates that will be appended to public hostnames routes identified by the annotations, for east-west routes use -kubernetes-east-west-range-annotation-predicates. E.g. -kubernetes-annotation-predicates='zone-a=true=Foo() && Bar()' will add 'Foo() && Bar()' predicates to all routes of ingress or routegroup annotated with 'zone-a: true'")
```
AlexanderYastrebov, line 477: Maybe rephrase like:
```
configures predicates appended to non east-west routes of annotated resources.
E.g. -kubernetes-annotation-predicates='zone-a=true=Foo() && Bar()' will add 'Foo() && Bar()' predicates to all non east-west routes of ingress or routegroup annotated with 'zone-a: true'.
For east-west routes use -kubernetes-east-west-range-annotation-predicates. 
```

config/config.go
```diff
@@ -470,6 +474,8 @@ func NewConfig() *Config {
...
+	flag.Var(&cfg.KubernetesEastWestRangeAnnotationPredicatesString, "kubernetes-east-west-range-annotation-predicates", "as -kubernetes-annotation-predicates, but for east-west range domains that match -kubernetes-east-west-range-domains")
```
AlexanderYastrebov, line 478:
```
Similar to -kubernetes-annotation-predicates configures predicates appended to east-west routes of annotated resources. See also -kubernetes-east-west-range-domains
```

MustafaSaber:
👍

szuecs:
Can you please also write in the description what "internal" hostname means?
There is a flag that defines the east-west domain and that's what we use (I didn't review the code but everything else would be a bug).

PR #3341, opened 2024-12-11
proxy: always count route lookup failures

AlexanderYastrebov:
Remove debug mode check for clarity and because
other metrics are updated without checking for debug mode.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1211,10 +1211,7 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 	route, params := p.lookupRoute(ctx)
 	p.metrics.MeasureRouteLookup(lookupStart)
 	if route == nil {
-		if !p.flags.Debug() {
-			p.metrics.IncRoutingFailures()
-		}
-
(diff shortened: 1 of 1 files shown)


szuecs:
I didn't check but if you create 404 counts for unknown hosts than it's a DoS vector for memory usage.

AlexanderYastrebov:
> I didn't check but if you create 404 counts for unknown hosts then it's a DoS vector for memory usage.
> If it's a single counter without host cardinality then it's fine.

Yes, this metric does not have client-control dimensions (like hostname):
```console
$ bin/skipper -inline-routes='Host("foo.test") -> <shunt>' -metrics-flavour=prometheus
[APP]INFO[0000] Expose metrics in prometheus format
...
```
```console
$ curl localhost:9090
Not Found
$ curl -s localhost:9911/metrics | grep skipper_route_error_total
# HELP skipper_route_error_total The total of route lookup errors.
# TYPE skipper_route_error_total counter
skipper_route_error_total 1
```

MustafaSaber:
👍

AlexanderYastrebov:
:+1:

PR #3365, opened 2025-01-10
dataclients/kubernetes: add inlineContent to healthcheck routes

AlexanderYastrebov:
Add shunting `inlineContent` filter to healthcheck routes so that filters added via `-default-filters-append` do not modify healthcheck response.

Note that users may disable healthcheck routes via `-kubernetes-healthcheck=false` and define custom healthcheck routes if needed.


AlexanderYastrebov:
Using `inlineContent` to deal with `-default-filters-append` feels hacky, back to the drawing board then.

Issue #169, opened 2016-06-21
Method predicate with multiple http methods is not behaving as expected


mohabusama:
Assuming the following eskip example:

``Path("/endpoint") && Method("GET", "POST")``

``eskip check`` passes with no errors.

So the expected behavior was to allow only ``GET`` and ``POST`` http methods, but ``DELETE`` was allowed too.

PR #3380, opened 2025-01-20
eskip: improve invalid predicate arguments error message

AlexanderYastrebov:
Add route id and predicate name to the error message.

### DIFF
--- a/eskip/eskip.go
+++ b/eskip/eskip.go
@@ -17,8 +17,6 @@ import (
 const duplicateHeaderPredicateErrorFmt = "duplicate header predicate: %s"
 
 var (
-	invalidPredicateArgError        = errors.New("invalid predicate arg")
-	invalidPredicateArgCountError   = errors.New("invalid predicate count arg")
 	duplicatePathTreePredicateError = errors.New("duplicate path tree predicate")
 	duplicateMethodPredicateError   = errors.New("duplicate method predicate")
 )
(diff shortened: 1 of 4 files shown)


eskip/eskip.go
```diff
@@ -531,9 +533,13 @@ func applyPredicates(route *Route, proute *parsedRoute) error {
...
 			route.Predicates = append(route.Predicates, p)
 		}
+
+		if err != nil {
+			return fmt.Errorf("invalid route %q: %w", proute.id, err)
+		}
 	}
 
-	return err
+	return nil
```
AlexanderYastrebov, line 542: Moved error check inside of the loop to add route id. 

Its also as better fix then #170 was for #169

eskip/eskip.go
```diff
@@ -446,17 +444,25 @@ func (t BackendType) String() string {
 }
 
 // Expects exactly n arguments of type string, or fails.
-func getStringArgs(n int, args []interface{}) ([]string, error) {
-	if len(args) != n {
-		return nil, invalidPredicateArgCountError
+func getStringArgs(p *Predicate, n int) ([]string, error) {
```
MustafaSaber, line 447: Is it worth changing the param order?

MustafaSaber:
👍

eskip/eskip.go
AlexanderYastrebov: I think its a matter of taste, I somehow like `getStringArgs(p, 2)` more than  `getStringArgs(2, p)` at the call site

AlexanderYastrebov: Its a helper, we could also make it a member although very specific to built-in predicates. Then the call would be `p.getStringArgs(2)`

AlexanderYastrebov:
:+1:

PR #3381, opened 2025-01-20
all: re-enable staticcheck rule ST1012

AlexanderYastrebov:
Rename error variables and re-enable staticheck
[ST1012 - Poorly chosen name for error variable](https://staticcheck.dev/docs/checks/#ST1012) rule.

All renamed variables are non-exported except errInvalidArgs in predicates/source which was made unexported.

Follow up on #897 and #1642


eskip/eskip.go
```diff
@@ -17,8 +17,8 @@ import (
 const duplicateHeaderPredicateErrorFmt = "duplicate header predicate: %s"
 
 var (
-	duplicatePathTreePredicateError = errors.New("duplicate path tree predicate")
-	duplicateMethodPredicateError   = errors.New("duplicate method predicate")
+	errDuplicatePathTreePredicate = errors.New("duplicate path tree predicate")
```
MustafaSaber, line 20: do we need to regenerate parser.y?

AlexanderYastrebov: Theses are not used in generated code but I'll double check if anything changes after regen.

AlexanderYastrebov: Its OK:
```console
$ go install golang.org/x/tools/cmd/goyacc@latest
$ go generate ./eskip
$ git diff
```

AlexanderYastrebov: Actually we can now remove `lint:file-ignore` directives from it.

MustafaSaber:
👍

AlexanderYastrebov:
:+1:

PR #3382, opened 2025-01-22
all: remove unused fields

AlexanderYastrebov:
Remove (some) unused fields detected via:
```
$ go install honnef.co/go/tools/internal/cmd/unused@latest
$ unused -field-writes-are-uses=false ./... 2>/dev/null
```


AlexanderYastrebov:
:+1:

MustafaSaber:
👍

PR #3383, opened 2025-01-23
config: allow multiple -inline-routes value

### LINKED ISSUE #3366: Support multiple values for -inline-routes flag

Currently `-inline-routes` flag can only be specified once:
https://github.com/zalando/skipper/blob/8d4721fb0d1787eb6d4c7159d2d22fa5341969f7/config/config.go#L128

which is not convenient for [templated yaml configurations](https://github.com/zalando-incubator/kubernetes-on-aws).

Since its value is in the eskip format one would need to resort to nested conditional logic like:
```yaml
args:
{{ if or (.config.foo .config.bar)}}
  - -inline-routes={{ if .config.foo }}{{ .Cluster.ConfigItems.foo }}{{ end }}
    {{- if .config.bar }}{{ if .config.foo }};{{ end }}{{ .Cluster.ConfigItems.bar }}{{ end }}
{{ end }}
```

`-inline-routes` supports multiple values

**Additional context**

See similar https://github.com/zalando/skipper/pull/3084

No

AlexanderYastrebov:
Extend config.Config and skipper.Options to allow multiple sets of routes in eskip format for convenience.

Fixes #3366

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -125,7 +125,7 @@ type Config struct {
 	EtcdPassword       string               `yaml:"etcd-password"`
 	RoutesFile         string               `yaml:"routes-file"`
 	RoutesURLs         *listFlag            `yaml:"routes-urls"`
-	InlineRoutes       string               `yaml:"inline-routes"`
+	InlineRoutes       multiFlag            `yaml:"inline-routes"`
 	AppendFilters      *defaultFiltersFlags `yaml:"default-filters-append"`
 	PrependFilters     *defaultFiltersFlags `yaml:"default-filters-prepend"`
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -125,7 +125,7 @@ type Config struct {
 	EtcdPassword       string               `yaml:"etcd-password"`
 	RoutesFile         string               `yaml:"routes-file"`
 	RoutesURLs         *listFlag            `yaml:"routes-urls"`
-	InlineRoutes       string               `yaml:"inline-routes"`
+	InlineRoutes       multiFlag            `yaml:"inline-routes"`
```
AlexanderYastrebov, line 128: This is a breaking change for yaml-based configuration.

PR #3405, opened 2025-02-11
predicates/forwarded: use strings.Index to parse header

AlexanderYastrebov:
Use strings.Index instead of strings.Split to reduce memory allocations.

See #3403 for details.

### DIFF
--- a/predicates/forwarded/forwarded.go
+++ b/predicates/forwarded/forwarded.go
@@ -130,11 +130,10 @@ type forwarded struct {
 }
 
 func parseForwarded(fh string) *forwarded {
-
 	f := &forwarded{}
 
-	for _, forwardedFull := range strings.Split(fh, ",") {
-		for _, forwardedPair := range strings.Split(strings.TrimSpace(forwardedFull), ";") {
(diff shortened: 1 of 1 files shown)


predicates/forwarded/forwarded.go
```diff
@@ -133,8 +133,18 @@ func parseForwarded(fh string) *forwarded {
...
 
-	for _, forwardedFull := range strings.Split(fh, ",") {
-		for _, forwardedPair := range strings.Split(strings.TrimSpace(forwardedFull), ";") {
+	// TODO: use range over func added in go1.23 like:
+	// for forwardedFull := range split(fh, ",") {
+	// 	for forwardedPair := range split(strings.TrimSpace(forwardedFull), ";") {
+	// 		// ...
+	// 	}
+	// }
+	// once go1.24 is released
```
AlexanderYastrebov, line 142: Actually it was released yesterday so we may bump min version https://github.com/zalando/skipper/pull/3406 and use range over func.

AlexanderYastrebov: Updated.

MustafaSaber:
👍

PR #3421, opened 2025-02-26
proxy: trace response body size and request header length

AlexanderYastrebov:
Add response body size and request headers length rounded to the next power of two as proxy span tags.

The rounding is done to reduce cardinality of the tags and avoid introducing a new configuration parameter.

Since it is tricky to measure request header length in bytes it is calculated as the sum of lengths of header names and values.

Note that OpenTelemetry [Semantic Conventions for HTTP Spans](https://opentelemetry.io/docs/specs/semconv/http/http-spans/) define the following related attributes:
* experimental [`http.response.body.size`](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/#http-response-body-size) for body size in bytes
* stable [`http.request.header.<key>`](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/#http-request-header) for header value

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -521,6 +521,16 @@ func cloneHeaderExcluding(h http.Header, excludeList map[string]bool) http.Heade
 	return hh
 }
 
+func sizeOfHeader(h http.Header) (size int) {
+	for k, vv := range h {
+		size += len(k)
+		for _, v := range vv {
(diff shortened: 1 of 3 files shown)


MustafaSaber:
👍

AlexanderYastrebov:
Should start with metrics.

PR #3431, opened 2025-03-03
Consistent QueryStrings on redirect (fixes #3430)

### LINKED ISSUE #3430: `redirectTo` filter is inconsistent.
**Describe the bug**
`redirectTo` filter behaves inconsistently when dealing with query strings.

[slack conversation](https://gophers.slack.com/archives/C82Q5JNH5/p1741018631409179)

Test case 1.
If I define the filter like `redirectTo(301, "https://example.com/")`, for a route that matches `mydomain.com`, with url `https://mydomain.com/incoming=true`, I'll get a redirect to `https://mydomain.com/?incoming=true`. That is, **query strings from the incoming request are propagated**. There's no way to disable query string propagation.

Test case 2.
If I define the filter like `redirectTo(301, "https://example.com/filter=true")`, for a route that matches `mydomain.com`, with url `https://mydomain.com/`, I'll get a redirect to `https://mydomain.com/?filter=true`. That is, **query strings from the filter are propagated**.

**Observed behavior**

Test case 3. (the one broken, IMHO)

If I define the filter like `redirectTo(301, "https://example.com/filter=true")`, for a route that matches `mydomain.com`, with url `https://mydomain.com/?incoming=true`, I'll get a redirect to `https://mydomain.com/?filter=true`. That is, **incoming query strings are lost**. 

**Expected behavior**

My expectation is that **incoming query strings** are propagated, and the final URL is a combination of query strings from filter and from incoming URL. 

So, if I define the filter like `redirectTo(301, "https://example.com/filter=true")`, for a route that matches `mydomain.com`, with url `https://mydomain.com/?incoming=true`, my expectation would be to get  a redirect to `https://mydomain.com/?filter=true&incoming=true`.

xavivars:


filters/builtin/redirect.go
```diff
@@ -149,6 +149,8 @@ func getLocation(ctx filters.FilterContext, location *url.URL, typ redirectType)
 
 	if u.RawQuery == "" {
 		u.RawQuery = r.URL.RawQuery
+	} else {
+		u.RawQuery = r.URL.RawQuery + "&" + u.RawQuery
```
szuecs, line 153: Please see the issue comment.
I think we need to have a separate input if you need this behavior, because we do not know how many users could have critical redirects that expect the current behavior.
I think the current behavior is correct.

We could have a 3rd optional parameter to mix the input query and configured query, see also [doc](https://opensource.zalando.com/skipper/reference/filters/#http-redirect).

xavivars: Yeah, I had read the docs, but they say nothing about query (I actually thought when reading the docs they were dropping whatever query was coming from the original request)

PR #3432, opened 2025-03-04
filters/openpolicyagent: Resolve race conditions in OPA filter benchmark tests

mefarazath:
Observed a race condition in parallel benchmark execution in OPA filter benchmark tests when testing it out with policy bundles that manipulate headers (leading to concurrent modification of the filter context created and used in tests with the way it is written now).

Changes,
- Moved context creation to b.RunParallel loop to prevent concurrent modification.
- Added some missing assertions in benchmark tests.
- Removed duplicate test resource.


AlexanderYastrebov:
:+1:

mjungsbluth:
👍

PR #3406, opened 2025-02-12
build: update min go version to 1.23

AlexanderYastrebov:
As go1.24 is out [https://go.dev/doc/go1.24] this change updates minimal required go version.

PR #3434, opened 2025-03-05
go.mod: use latest 1.23 version

AlexanderYastrebov:
Current go version is 1.24 and we support two latest go versions. This change updates go.mod to use latest go 1.23 version to include all available go runtime security fixes.

It also bumps minor skipper version as per https://github.com/zalando/skipper/?tab=readme-ov-file#version-promise (overlooked within #3406).


MustafaSaber:
👍

AlexanderYastrebov:
:+1:

PR #1710, opened 2021-02-08
Wait for multiple dataclients. Fixes #1709

xavivars:
Fixes: #1709

PR #3447, opened 2025-03-21
routing: fix first route load signaling

AlexanderYastrebov:
Dataclients may load routes with arbitrary delays so it may happen
that of two dataclients the first loads routes twice and the second zero
times which will signal first route loaded condition although
routing table will only contain routes from the first dataclient.

This change fixes route update logic to signal first load only after
routes from all configured dataclients are received at least once.

Updates https://github.com/zalando/skipper/issues/1709
Updates https://github.com/zalando/skipper/pull/1710


MustafaSaber:
👍

a4180p:
👍

routing/datasource.go
```diff
@@ -140,10 +140,17 @@ func applyIncoming(defs routeDefs, d *incomingData) routeDefs {
...
+}
+
 // merges the route definitions from multiple data clients by route id
-func mergeDefs(defsByClient map[DataClient]routeDefs) []*eskip.Route {
+func mergeDefs(defsByClient map[DataClient]routeDefs) mergedDefs {
+	clients := make(map[DataClient]struct{}, len(defsByClient))
 	mergeByID := make(routeDefs)
-	for _, defs := range defsByClient {
+	for c, defs := range defsByClient {
+		clients[c] = struct{}{}
```
MustafaSaber, line 153: This assumes no client is repeated (which actually how skipper works now), but can override if we use multiple same dataclient multiple times

MustafaSaber:
👍

routing/datasource.go
AlexanderYastrebov: Data clients we have are pointers so there could be multiple clients of the same type, e.g. the test uses two clients of the same type.
But value-type dataclients won't work, e.g. we can not have two of these:
```go
type valueDataClient struct{}

var _ routing.DataClient = valueDataClient{}

func (v valueDataClient) LoadAll() ([]*eskip.Route, error)              { return nil, nil }
func (v valueDataClient) LoadUpdate() ([]*eskip.Route, []string, error) { return nil, nil, nil }
```
I think we can add a check for this case.

See also https://go.dev/ref/spec#Map_types

AlexanderYastrebov:
:+1:

routing/datasource.go
MustafaSaber: yes, thanks! I overlooked that they are pointers.

AlexanderYastrebov: I've created follow up #3527

Issue #2803, opened 2023-12-20
Does the oauthOidcAllClaims filter provide the possibility to lookup sensitive data from Kubernetes Secrets


lukibahr:
Hi folks, 

we have several RouteGroups with the oauthOidcAllClaims filter set. 

```yaml
oauthOidcAllClaims("https://oidc-provider.example.com", "client_id", "client_secret",
    "http://target.example.com/subpath/callback", "email profile", "name email picture",
    "parameter=value", "X-Auth-Authorization:claims.email")
```

Is it possible or do you know a way on how to lookup the "client_secret" value from another source, like a kubernetes secret? Or is there a recommended way on how to handle these resources especially when dealing with GitOps?

Thanks and kind regards,

Lukas

szuecs:
@lukibahr That would be a feature to support which makes a lot of sense!

lukibahr:
@szuecs Is there a reference implementation already in skipper where we use the secrets module?

szuecs:
In many auth filters it is in use or also in net/Client (skipper's http client)

lukibahr:
@szuecs Is there any chance to get this implemented in the next couple of days or weeks? I'm currently not familiar enough with the entire codebase, to get this implemented and propose a MR for this. 

Thanks

szuecs:
MR?

I will have holidays until January, so I am not really available next days.
Maybe I can give a bit of context to the code. We implement these kind of features in filters. 
Filters are created by an instance of a ["spec"](https://github.com/zalando/skipper/blob/master/filters/auth/oidc.go#L152). You likely want to create a new spec for this similar to the others (share spec with different new kind of "typ").
A filter instance is created by [CreateFilter](https://github.com/zalando/skipper/blob/master/filters/auth/oidc.go#L181)().

The entry point of the code that does the call path to work on the request is https://github.com/zalando/skipper/blob/master/filters/auth/oidc.go#L756.

AlexanderYastrebov:
Looks like duplicate of https://github.com/zalando/skipper/issues/1952

lukibahr:
I will have a look into it, after getting familiar with the codebase.

lukibahr:
solved in https://github.com/zalando/skipper/releases/tag/v0.24.40.

PR #3454, opened 2025-03-24
default config for OIDC SubdomainsToRemove

universam1:
Follows: 
https://github.com/zalando/skipper/pull/1923  
https://github.com/zalando/skipper/issues/1909

We would appreciate having a default for [SubdomainsToRemove](https://opensource.zalando.com/skipper/reference/filters/#oauthoidcanyclaims) 
Currently we need to train many teams to configure and change the default.

We have many hosts under one domain with different OIDC filter settings, and the _default_ causes still the original issue https://github.com/zalando/skipper/issues/1909


webframp:
Definitely solves an issue we've encountered, maybe useful for others as well

filters/auth/oidc.go
```diff
@@ -234,7 +235,7 @@ func (s *tokenOidcSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 		return nil, err
 	}
 
-	subdomainsToRemove := 1
+	subdomainsToRemove := s.options.SubdomainsToRemove // default from config
```
AlexanderYastrebov, line 238: Hm, this will break for those who configure auth.OidcOptions manually - now it becomes zero by default when not specified. Here https://github.com/zalando/skipper/pull/2203#discussion_r1073836873we decided to go with a pointer.

universam1: Thanks for the hint, good point!

AlexanderYastrebov:
Looks good modulo comments

filters/auth/oidc.go
AlexanderYastrebov: In general I think we have a lot of overlapping functionality between OIDC and OAuth Grant Flow related to cookie handling so I created https://github.com/zalando/skipper/issues/3455

config/config.go
```diff
@@ -216,6 +216,7 @@ type Config struct {
 	OidcSecretsFile                   string        `yaml:"oidc-secrets-file"`
 	OIDCCookieValidity                time.Duration `yaml:"oidc-cookie-validity"`
 	OidcDistributedClaimsTimeout      time.Duration `yaml:"oidc-distributed-claims-timeout"`
+	OIDCSubdomainsToRemove            int           `yaml:"oidc-subdomains-to-remove"`
```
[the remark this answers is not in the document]
universam1: good hint, applied!

config/config.go
```diff
@@ -216,6 +216,7 @@ type Config struct {
 	OidcSecretsFile                   string        `yaml:"oidc-secrets-file"`
 	OIDCCookieValidity                time.Duration `yaml:"oidc-cookie-validity"`
 	OidcDistributedClaimsTimeout      time.Duration `yaml:"oidc-distributed-claims-timeout"`
+	OIDCTokenCookieRemoveSubdomains   int           `yaml:"oidc-token-cookie-remove-subdomains"`
```
AlexanderYastrebov, line 219: Do we want `-token` part really? For OAuth Grant we ended up with it but here we have another cookie option `oidc-cookie-validity` so this could be `oidc-cookie-remove-subdomains` without implying any token stored in it. WDYT?

AlexanderYastrebov:
> Currently we need to train many teams to configure and change the default.

It is interesting to see OIDC in use in multi-tenant setup :+1: 
Do you allow users to configure arbitrary issuers or how do you validate them?
How do you handle client secret management (https://github.com/zalando/skipper/issues/2803 and https://github.com/zalando/skipper/issues/1952)?

BTW check out my related weekend project https://github.com/AlexanderYastrebov/bot-idp/

config/config.go
universam1: That makes sense, the shorter the better. I removed the token from the name.

AlexanderYastrebov:
Thank you for improving Skipper :+1:

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
It would be nice to update the docs.

szuecs:
:+1:

universam1:
> It would be nice to update the docs.

added some docs for the command line flag

universam1:
> It is interesting to see OIDC in use in multi-tenant setup 👍 Do you allow users to configure arbitrary issuers or how do you validate them? 

The dev teams are trained how to configure the auth via RGs after they have been provided an app client for their application. We usually don't do multitenant but every app has its dedicated client.

> How do you handle client secret management (#2803 and #1952)?

This is indeed probably the probably biggest technical debt of Skipper auth concept. We would like to use Bank-Vaults webhook secret injection in memory if possible, but the current security posture relies on namespace RBAC access isolation. K8s secrets might be slightly better as of obscurity, but not a real win.

> BTW check out my related weekend project https://github.com/AlexanderYastrebov/bot-idp/ ⭐

This is a cool idea, thanks for the hint!

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #3457, opened 2025-03-24
filters/builtin: drop specified header value

AlexanderYastrebov:
Extend `dropRequestHeader` and `dropResponseHeader` to support optional second value argument and drop only this value when specified.


filters/builtin/header_test.go
```diff
@@ -85,12 +86,7 @@ func testHeaders(t *testing.T, got, expected http.Header) {
 			delete(got, n)
 		}
 	}
-
-	if !compareHeaders(got, expected) {
-		printHeader(t, expected, "invalid header", "expected")
-		printHeader(t, got, "invalid header", "got")
-		t.Error("invalid header")
-	}
+	assert.Equal(t, expected, got)
```
AlexanderYastrebov, line 89: This gives a better diff and tells empty from nil apart.

filters/builtin/header_test.go
```diff
@@ -186,19 +184,40 @@ func TestHeader(t *testing.T) {
...
+			args:           []interface{}{"X-Test-Name", "bar"},
+			valid:          true,
+			requestHeader:  http.Header{"X-Test-Name": []string{"foo", "bar", "baz"}},
+			expectedHeader: http.Header{"X-Test-Request-Name": []string{"foo", "baz"}},
+		}, {
+			msg:            "ignore non-matching",
+			args:           []interface{}{"X-Test-Name", "qux"},
+			valid:          true,
+			requestHeader:  http.Header{"X-Test-Name": []string{"foo", "bar", "baz"}},
+			expectedHeader: http.Header{"X-Test-Request-Name": []string{"foo", "bar", "baz"}},
```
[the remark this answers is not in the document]
AlexanderYastrebov: Also removal of a single value.

filters/builtin/headerfilter.go
```diff
@@ -281,7 +283,12 @@ func (f *headerFilter) Request(ctx filters.FilterContext) {
 			ctx.SetOutgoingHost(f.value)
 		}
 	case dropRequestHeader:
-		header.Del(f.key)
+		if f.value == "" {
+			header.Del(f.key)
+		} else {
+			k := textproto.CanonicalMIMEHeaderKey(f.key)
+			header[k] = slices.DeleteFunc(header[k], func(v string) bool { return v == f.value })
```
szuecs, line 290: I think value should also be case insensitive.

AlexanderYastrebov: My idea was to have a low-level primitive to drop a specific value. 
For setting and adding header we do not do any modification to the value.

To support regexp we may consider changing modRequestHeader such that it _deletes_ the matching header if "the replacement (string)" parameter is empty (or absent) or maybe add a whole new filter.

szuecs: I think we should add another filter dropRequestHeaderRegexp maybe?

ponimas: Could a third boolean parameter to enable case-insensitive comparison be an alternative to regex?

AlexanderYastrebov: It could be done like that of course but lets wait for the usecase.

I initially considered to add a new filter `dropRequestHeaderValue` then decided to extend existing one but we should wary about how much logic we cram into it.

These header filters are basic primitives and building blocks and I think we should keep them simple. For higher-level or interesting logic we can always add new filters.

AlexanderYastrebov: > I think we should add another filter dropRequestHeaderRegexp maybe?

Indeed, thanks, a new dropRequestHeaderRegexp would be better because exact value match is a very simple regexp (`foo -> ^foo$` ), stdlib handles it as a special case of fast prefix check, we already use it e.g. in HeaderRegexp and it covers many other usecases.

AlexanderYastrebov: > Could a third boolean parameter to enable case-insensitive comparison be an alternative to regex?

This could be achieved by regexp flags (e.g. `^(?i)foo$`)

MustafaSaber: https://github.com/zalando/skipper/pull/3778

a4180p:
https://github.com/zalando/skipper/pull/3778 was merged. 
Should we close this one? @MustafaSaber

PR #3464, opened 2025-03-31
Allow configuration of bundle download behaviour of open-policy-agent filter

torwunder:
Addresses: https://github.com/zalando/skipper/issues/3119

**Solution**
add option to override the period trigger for discovery and bundle plugins
- control timing of f.ex. bundle download directly (also adding jitter to avoid different opa instances updating at the same time)
- fail fast for non-recoverable errors

**Context**
The current (and future default) behaviour is relying on the configurable startup timeout to determine if the open-policy-agent filter can successfully serve requests (if it was able to download the bundles, aso).
The solution is based on a [change](add option to override period trigger for plugin) in upstream opa that exposes the httpError struct on the trigger method.


config/config.go
```diff
@@ -524,9 +527,12 @@ func NewConfig() *Config {
 	flag.Var(cfg.CredentialPaths, "credentials-paths", "directories or files to watch for credentials to use by bearerinjector filter")
 	flag.DurationVar(&cfg.CredentialsUpdateInterval, "credentials-update-interval", 10*time.Minute, "sets the interval to update secrets")
 	flag.BoolVar(&cfg.EnableOpenPolicyAgent, "enable-open-policy-agent", false, "enables Open Policy Agent filters")
+	flag.BoolVar(&cfg.EnableOpenPolicyAgentOverridePeriodTriggers, "enable-open-policy-agent-override-period-triggers", false, "when enabled skipper will set all plugin triggers to manual and trigger all plugins directly, which includes f.ex. to download of new bundles")
```
mjungsbluth, line 530:
```suggestion
	flag.BoolVar(&cfg.EnableOpenPolicyAgentOverridePeriodTriggers, "enable-open-policy-agent-skipper-controlled-triggers", false, "when enabled, Skipper will use a central control loop trigger plugins, which includes f.ex. to download of new bundles")
```

filters/openpolicyagent/internal/confighook.go
```diff
@@ -0,0 +1,80 @@
...
+
+func discoveryPluginOverride(config *config.Config) (*config.Config, error) {
+	var (
+		discoveryConfig discovery.Config
+		triggerManual   = plugins.TriggerManual
+		message         []byte
+	)
+
+	if config.Discovery != nil {
+		if err := json.Unmarshal(config.Discovery, &discoveryConfig); err == nil {
```
mjungsbluth, line 44: While this is more compact, it is a little less idomatic Go

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -289,6 +326,37 @@ func (registry *OpenPolicyAgentRegistry) startCleanerDaemon() {
 	}
 }
 
+func (registry *OpenPolicyAgentRegistry) startPluginTriggerDaemon() {
+	interval := registry.pluginTriggerInterval
+
+	// Prevent pods from triggering plugins (f.ex. downloading new bundles) at the same time
```
mjungsbluth, line 332:
```suggestion
	// Prevent multiple OPA instances from triggering plugins (f.ex. downloading new bundles) at the same time
```

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -289,6 +326,37 @@ func (registry *OpenPolicyAgentRegistry) startCleanerDaemon() {
...
+	if registry.maxPluginTriggerJitter > 0 {
+		jitter := time.Duration(rand.Int63n(registry.maxPluginTriggerJitter.Milliseconds())) * time.Millisecond
+		interval += jitter
```
mjungsbluth, line 335: I think this jitter might have some unexpected statistical properties: Over time I would expect the interval be independent of the jitter, i.e. after 10 repetitions at aroung 600s +/- jitter. However the jitter is also taken once upon start and then amplifies. Typical jitter implementations would also center around the time point that it would hit without (i.e. use (period - max_jitter, period + max_jitter) as range to pull from. Like here: https://github.com/mroth/jitter

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -289,6 +326,37 @@ func (registry *OpenPolicyAgentRegistry) startCleanerDaemon() {
...
+	}
+
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	registry.mu.Lock()
+	instances := make([]*OpenPolicyAgentInstance, 0, len(registry.instances))
```
mjungsbluth, line 342: This list is dynamic, and the for loop below would run on outdated instances...

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -516,6 +596,102 @@ func (opa *OpenPolicyAgentInstance) Start(ctx context.Context, timeout time.Dura
...
+	isHttpError := errors.As(err, httpError)
+	if isHttpError && (httpError.StatusCode == 429 || httpError.StatusCode >= 500) {
+		return true
+	}
+	return false
+}
+
+func (opa *OpenPolicyAgentInstance) verifyAllPluginsStarted() error {
+	allPluginsReady := true
+	for pluginName, status := range opa.manager.PluginStatus() {
```
mjungsbluth, line 658: Maybe we make it explicit that we only wait for the bundle and discovery plugin?

filters/openpolicyagent/openpolicyagent.go
torwunder: yes but that avoids having a look that spans http calls
also an instance coming up will be triggered on startup

filters/openpolicyagent/internal/confighook.go
mjungsbluth, line 44:
```
err := json.Unmarshal(config.Discovery, &discoveryConfig) 

if err != nil {
	return nil, err
}

discoveryConfig.Trigger = &triggerManual

message, err := json.Marshal(discoveryConfig)

if err != nil {
	return nil, err
}

config.Discovery = message
```

config/config.go
torwunder: I will leave this open for other opinions on the name of the parameter, as changing it will reflect in quite some test changes.
current options:
1. `enable-open-policy-agent-override-period-triggers`
2. `enable-open-policy-agent-skipper-controlled-triggers`
3. `enable-open-policy-agent-custom-plugin-triggers`

filters/openpolicyagent/openpolicyagent.go
torwunder: I agree with the assessment but IMO this is more about expectation management. I would not expect operational impact if the update/trigger cycles diverge.
Having a more correct implementation will come with either a more complex implementation or the addition of a library.

filters/openpolicyagent/openpolicyagent.go
torwunder: I misunderstood, ofc. moving it into the ticker loop is the right thing to do here.

torwunder: addressed

AlexanderYastrebov (commented):
I did a first pass

AlexanderYastrebov:
@torwunder  FYI, I am trying what Copilot review is capable of so I've requested a review from it.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -516,6 +596,102 @@ func (opa *OpenPolicyAgentInstance) Start(ctx context.Context, timeout time.Dura
...
+			}
+
+			opa.Logger().WithFields(map[string]interface{}{
+				"retry":  retry,
+				"status": httpError.StatusCode,
+			}).Warn("Triggering bundles failed. Retrying %d", retry)
+
+			time.Sleep(backoff)
+			backoff *= 2
+			retry++
```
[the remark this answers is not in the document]
torwunder: the retry is cancelled via the surrounding context that uses the startup timeout
if either discovery or bundle plugin cannot be triggered without error the filter creation will fail
the loop will exit if there is no retryable error (http error with status <500 and not 429 or no error at all)

will have a look at an implementation with a ticker

filters/openpolicyagent/openpolicyagent.go
torwunder: addressed as suggested by resetting the trigger

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -289,6 +326,37 @@ func (registry *OpenPolicyAgentRegistry) startCleanerDaemon() {
...
+		case <-ticker.C:
+			registry.mu.Lock()
+			instances := make([]*OpenPolicyAgentInstance, 0, len(registry.instances))
+			for _, opa := range registry.instances {
+				instances = append(instances, opa)
+			}
+			registry.mu.Unlock()
+
+			for _, opa := range instances {
+				opa.triggerPlugins(context.Background())
```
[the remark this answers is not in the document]
torwunder: it is running in it's own go routing (see https://github.com/zalando/skipper/blob/ebd3bea1a8d6ff6609010c120a3d15b1537644ce/filters/openpolicyagent/openpolicyagent.go#L183) 
if we add a timeout I think it could be sensible to use the same timeout as the startup timeout.

config/config.go
torwunder: the configuration will switch from the opa default behaviour of periodically triggering the bundle to a setup where the opa-filter is triggering the bundles, allowing for better error handling (like retries) and more control (like adding the jitter)

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -289,6 +328,35 @@ func (registry *OpenPolicyAgentRegistry) startCleanerDaemon() {
...
+			}
+			ticker.Reset(pluginTriggerIntervalWithJitter(registry.pluginTriggerInterval, registry.maxPluginTriggerJitter))
+		}
+	}
+}
+
+// Prevent different opa instances from triggering plugins (f.ex. downloading new bundles) at the same time
+func pluginTriggerIntervalWithJitter(interval, maxJitter time.Duration) time.Duration {
```
AlexanderYastrebov, line 353: Could be a member method if registry and then it won't require any parameters.

config/config.go
```diff
@@ -294,14 +294,17 @@ type Config struct {
...
-	EnableOpenPolicyAgent                bool          `yaml:"enable-open-policy-agent"`
-	OpenPolicyAgentConfigTemplate        string        `yaml:"open-policy-agent-config-template"`
-	OpenPolicyAgentEnvoyMetadata         string        `yaml:"open-policy-agent-envoy-metadata"`
-	OpenPolicyAgentCleanerInterval       time.Duration `yaml:"open-policy-agent-cleaner-interval"`
-	OpenPolicyAgentStartupTimeout        time.Duration `yaml:"open-policy-agent-startup-timeout"`
-	OpenPolicyAgentRequestBodyBufferSize int64         `yaml:"open-policy-agent-request-body-buffer-size"`
-	OpenPolicyAgentMaxRequestBodySize    int64         `yaml:"open-policy-agent-max-request-body-size"`
-	OpenPolicyAgentMaxMemoryBodyParsing  int64         `yaml:"open-policy-agent-max-memory-body-parsing"`
+	EnableOpenPolicyAgent                       bool          `yaml:"enable-open-policy-agent"`
+	EnableOpenPolicyAgentOverridePeriodTriggers bool          `yaml:"enable-open-policy-agent-override-period-triggers"`
```
AlexanderYastrebov, line 298: Should it have `-plugin` in the name?
Also please put related configs next to each other:
```
enable-open-policy-agent[-plugin]-override-period-triggers
open-policy-agent-plugin-trigger-interval
```

config/config.go
mjungsbluth: I like this one best

AlexanderYastrebov: Lets have it like that then

filters/openpolicyagent/openpolicyagent.go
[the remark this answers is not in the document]
torwunder: addressed

filters/openpolicyagent/openpolicyagent.go
torwunder: it is checking the status of all plugins to align with the normal startup without the control loop

config/config.go
torwunder: addressed

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -516,6 +604,109 @@ func (opa *OpenPolicyAgentInstance) Start(ctx context.Context, timeout time.Dura
...
+	}
+	if !allPluginsReady {
+		return fmt.Errorf("open policy agent plugins failed to start")
+	}
+	return nil
+}
+
+func (opa *OpenPolicyAgentInstance) triggerPlugins(ctx context.Context) error {
+
+	if opa.stopped {
```
mjungsbluth, line 694: "stopped" is true as soon as OPA has been shut down. During the shutdown process we might still trigger the plugins. 

I wonder I we should add a new attribute "stopping" that we set before we initiate the shutdown on the plugin manager

mjungsbluth:
Looks good to me minus one minor nitpick that should be quick to address...

filters/openpolicyagent/openpolicyagent.go
torwunder: addressed with https://github.com/zalando/skipper/pull/3464/commits/50bacbe70463bdd58764926cc5ec4eda49934c22

mjungsbluth:
👍

AlexanderYastrebov:
:+1:

PR #3466, opened 2025-04-01
Add context to log entry

xavivars:
This change allows accesslog entries to be enriched with requests' context, if a custom formatter uses that context.

### DIFF
--- a/logging/access.go
+++ b/logging/access.go
@@ -170,5 +170,9 @@ func LogAccess(entry *AccessEntry, additional map[string]interface{}) {
 		logData[k] = v
 	}
 
-	accessLog.WithFields(logData).Infoln()
+	logEntry := accessLog.WithFields(logData)
+	if entry.Request != nil {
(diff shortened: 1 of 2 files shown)


szuecs:
I miss a test :)

xavivars:
Test added!

MustafaSaber:
Just curious, do we know how much extra data will be logged by this? Can we do benchmark and compare before and after? I'm thinking about how much logging storage will this increase.

AlexanderYastrebov:
staticcheck is not happy for some reason
```
staticcheck -checks "all,-ST1000,-ST1003,-ST1020,-ST1021" ./...
Error: logging/access_test.go:82:87: should not use built-in type string as key for value; define your own type to avoid collisions (SA1029)
```

xavivars:
All tests passed now!

MustafaSaber:
👍

AlexanderYastrebov:
Thank you for improving Skipper.

AlexanderYastrebov:
:+1:

Issue #3327, opened 2024-12-04
basicAuth filter panics when htpasswd file does not exist


AlexanderYastrebov:
**Describe the bug**

[basicAuth](https://github.com/zalando/skipper/blob/master/docs/reference/filters.md#basicauth) filter panics when htpasswd file does not exist

**To Reproduce**

```console
$ bin/skipper -inline-routes='*-> basicAuth("/does-not-exist") -> inlineContent("OK\n") -> <shunt>'
[APP]INFO[0000] Expose metrics in codahale format            
[APP]INFO[0000] enable swarm: false                          
[APP]INFO[0000] Replacing tee filter specification           
[APP]INFO[0000] Replacing teenf filter specification         
[APP]INFO[0000] Replacing lua filter specification           
[APP]INFO[0000] route settings, reset, route: : * -> basicAuth("/does-not-exist") -> inlineContent("OK\n") -> <shunt> 
[APP]INFO[0000] route settings received, id: 1               
[APP]INFO[0000] support listener on :9911                    
[APP]INFO[0000] route settings applied, id: 1                
[APP]INFO[0000] Dataclients are updated once, first load complete 
[APP]INFO[0000] Listen on :9090                              
[APP]INFO[0000] TLS settings not found, defaulting to HTTP
```

Make a request:
```console
$ curl -v test:test@localhost:9090
* Host localhost:9090 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:9090...
* Connected to localhost (::1) port 9090
* Server auth using Basic with user 'test'
> GET / HTTP/1.1
> Host: localhost:9090
> Authorization: Basic dGVzdDp0ZXN0
> User-Agent: curl/8.5.0
> Accept: */*
> 
< HTTP/1.1 500 Internal Server Error
< Content-Length: 22
< Content-Type: text/plain; charset=utf-8
< Server: Skipper
< X-Content-Type-Options: nosniff
< Date: Wed, 04 Dec 2024 12:02:37 GMT
< 
Internal Server Error
* Connection #0 to host localhost left intact
```

Observe panic message in the logs:
```
[APP]ERRO[0090] stacktrace of panic caused by: stat /does-not-exist: no such file or directory:
goroutine 57 [running]:
github.com/zalando/skipper/proxy.stack()
        /home/ayastrebov/src/github.com/zalando/skipper/proxy/proxy.go:1167 +0x5e
[...]

PR #3467, opened 2025-04-02
bugfix - check config file existence for basicAuth filter

ayruslore:
filters/auth: check config file existence for basicAuth filter

Missing config file causes basicAuth filter to panic.
This change adds file existence check during filter creation.

Fixes https://github.com/zalando/skipper/issues/3327


filters/auth/basic.go
```diff
@@ -65,6 +75,13 @@ func (spec *basicSpec) CreateFilter(config []interface{}) (filters.Filter, error
 		}
 	}
 
+	if _, err := os.Stat(configFile); err != nil {
+		return &basic{
+			authenticator:   nil,
+			realmDefinition: ForceBasicAuthHeaderValue + `"` + realmName + `"`,
+		}, nil
+	}
```
AlexanderYastrebov, line 83: I think we should fail early here if file does not exist (and the datasource will discard the whole route).
If file was removed later in runtime then its fine to return 500 like it is now.

ayruslore: > I think we should fail early here if file does not exist (and the datasource will discard the whole route)

But If the route is discarded then the response would be 404 instead of 401. What do we want as expected behaviour?

To return a 401 we need to match the route and then fail the auth, and if we discard the route, then this will never match returning 404 or status code something other route that matches....

> If file was removed later in runtime then its fine to return 500 like it is now.

this makes sense.

AlexanderYastrebov: We need to fail somehow as current approach simply masks the problem and user/operator won't tell why they get 401.

ayruslore: i added a body message to show to the user. See line https://github.com/zalando/skipper/pull/3467/files#diff-eef77465b5f737710ce3478e21e86d18fa5ce13979926f4f19cd38a8457d6e5dR41

Now with a 401 we indicate that the credential file is missing. Would that suffice or do you have any other suggestions indicating why the 401?

This is what current response looks like now.
```
❯ curl -v test:test@localhost:9090
* Host localhost:9090 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:9090...
* Connected to localhost (::1) port 9090
* Server auth using Basic with user 'test'
> GET / HTTP/1.1
> Host: localhost:9090
> Authorization: Basic dGVzdDp0ZXN0
> User-Agent: curl/8.7.1
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 401 Unauthorized
< Server: Skipper
< Date: Fri, 04 Apr 2025 12:38:51 GMT
< Transfer-Encoding: chunked
< 
missing credential file%    
```

MustafaSaber:
I think last commit wasn't supposed to be on this branch

ayruslore:
> I think last commit wasn't supposed to be on this branch

Yes. I have removed that commit. I pushed to the wrong branch. Now the commit is excluded from here and added to the right place.

filters/auth/basic.go
ayruslore: Update : I have simplified and now it fails early in Create Filter and ignores the route. Consistent with the behaviour of other filters.

filters/auth/basic.go
```diff
@@ -65,6 +68,10 @@ func (spec *basicSpec) CreateFilter(config []interface{}) (filters.Filter, error
 		}
 	}
 
+	if _, err := os.Stat(configFile); err != nil {
+		return nil, errors.New("htpasswd file not found: " + configFile)
```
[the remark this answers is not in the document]
ayruslore: done. updated.

filters/auth/basic_test.go
```diff
@@ -71,6 +71,14 @@ func TestWithSuccessfulAuth(t *testing.T) {
 	}
 }
 
+func TestWithMissingAuthFile(t *testing.T) {
+	spec := NewBasicAuth()
+	_, err := spec.CreateFilter([]interface{}{"testdata/missingfile"})
+	if err == nil {
```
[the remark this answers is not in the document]
ayruslore: fixed.

filters/auth/basic.go
```diff
@@ -65,6 +68,10 @@ func (spec *basicSpec) CreateFilter(config []interface{}) (filters.Filter, error
 		}
 	}
 
+	if _, err := os.Stat(configFile); err != nil {
+		return nil, fmt.Errorf("htpasswd file not found: %s", configFile)
```
AlexanderYastrebov, line 72: Lets wrap error here and also make message generic since we don't know what the problem exactly is:
```go
return nil, fmt.Errorf("stat failed for %q: %w", configFile, err)
```

ayruslore: updated the code.

filters/auth/basic_test.go
```diff
@@ -71,6 +72,13 @@ func TestWithSuccessfulAuth(t *testing.T) {
 	}
 }
 
+func TestWithMissingAuthFile(t *testing.T) {
```
MustafaSaber, line 75: I think this can be added under `TestCreateFilterBasicAuthErrorCases`, but can also be done in another PR.

PR #3471, opened 2025-04-07
filters/accesslog:  Add mask access log query filter

ponimas:
See the previous PR for discussion details - https://github.com/zalando/skipper/pull/2674

Closing https://github.com/zalando/skipper/issues/2156

### DIFF
--- a/filters/accesslog/control.go
+++ b/filters/accesslog/control.go
@@ -105,7 +105,7 @@ func NewMaskAccessLogQuery() filters.Spec {
 func (*maskAccessLogQuery) Name() string { return filters.MaskAccessLogQueryName }
 
 func (al *maskAccessLogQuery) CreateFilter(args []interface{}) (filters.Filter, error) {
-	if len(args) <= 0 {
+	if len(args) == 0 {
 		return nil, filters.ErrInvalidFilterParameters
 	}
 
--- a/filters/accesslog/control.go
+++ b/filters/accesslog/control.go
@@ -18,8 +18,12 @@ const (
 
 // Common filter struct for holding access log state
 type AccessLogFilter struct {
-	Enable   bool
+	// Enable represents whether or not the access log is enabled.
+	Enable bool
+	// Prefixes contains the list of response code prefixes.
 	Prefixes []int
+	// MaskedQueryParams contains the list of query parameters (keys) that are masked/obfuscated in the access log.
+	MaskedQueryParams []string
 }
 
 func (al *AccessLogFilter) Request(ctx filters.FilterContext) {
@@ -86,4 +90,33 @@ func (*enableAccessLog) Name() string { return filters.EnableAccessLogName }
 
 func (al *enableAccessLog) CreateFilter(args []interface{}) (filters.Filter, error) {
 	return extractFilterValues(args, true)
+}
+
+type maskAccessLogQuery struct{}
+
+// NewMaskAccessLogQuery creates a filter spec to mask specific query parameters from the access log for a specific route.
+// Takes in query param keys as arguments. When provided, the value of these keys are masked (i.e., hashed).
+//
+//	maskAccessLogQuery("key_1", "key_2") to mask the value of provided keys in the access log.
+func NewMaskAccessLogQuery() filters.Spec {
+	return &maskAccessLogQuery{}
+}
+
+func (*maskAccessLogQuery) Name() string { return filters.MaskAccessLogQueryName }
+
+func (al *maskAccessLogQuery) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) <= 0 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	keys := make([]string, len(args))
+	for i := range args {
+		if key, ok := args[i].(string); ok && key != "" {
+			keys[i] = key
+		} else {
+			return nil, filters.ErrInvalidFilterParameters
+		}
+	}
+
+	return &AccessLogFilter{Enable: true, MaskedQueryParams: keys}, nil
 }
--- a/filters/accesslog/control_test.go
+++ b/filters/accesslog/control_test.go
@@ -1,8 +1,9 @@
 package accesslog
 
 import (
-	"github.com/google/go-cmp/cmp"
 	"testing"
+
+	"github.com/google/go-cmp/cmp"
 
 	"github.com/zalando/skipper/filters"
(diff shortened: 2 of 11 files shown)


filters/accesslog/control.go
```diff
@@ -87,3 +99,32 @@ func (*enableAccessLog) Name() string { return filters.EnableAccessLogName }
...
+
+func (*maskAccessLogQuery) Name() string { return filters.MaskAccessLogQueryName }
+
+func (al *maskAccessLogQuery) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) == 0 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	keys := make(map[string]bool, len(args))
+	for i := range args {
```
MustafaSaber, line 121: minor:

```go
for _, arg := args {
     if key, ok := arg.(string);
}
```

filters/accesslog/control_test.go
```diff
@@ -41,34 +42,44 @@ func TestAccessLogControl(t *testing.T) {
...
 			f, err := ti.state.CreateFilter(ti.args)
 
 			if ti.isError {
-				if err == nil {
-					t.Errorf("Unexpected error creating filter %v", err)
-					return
-				} else {
-					return
-				}
+				require.Error(t, err, "Expected error creating filter")
```
MustafaSaber, line 81: minor: you can also use `require.Condition` to remove the outer if.

logging/access.go
```diff
@@ -109,6 +115,29 @@ func stripQueryString(u string) string {
...
+
+	params := req.URL.Query()
+	for k := range maps.Keys(maskedQueryParams) {
+		val := params.Get(k)
+		if val == "" {
+			continue
+		}
+
+		hashed := hash(val)
+		params.Set(k, fmt.Sprintf("%d", hashed))
```
MustafaSaber, line 131: minor: maybe just inline since we don't use the variable

MustafaSaber: params.Set(k, fmt.Sprintf("%d", hash(val)))

logging/access.go
```diff
@@ -109,6 +115,29 @@ func stripQueryString(u string) string {
 	}
 }
 
+// maskQueryParams masks (i.e., hashing) specific query parameters in the provided request's URI.
+// Returns the obfuscated URI.
+func maskQueryParams(req *http.Request, maskedQueryParams map[string]bool) string {
```
MustafaSaber: I think we can also inline map len check here and return early

logging/access.go
```diff
@@ -144,6 +173,10 @@ func LogAccess(entry *AccessEntry, additional map[string]interface{}) {
 		uri = entry.Request.RequestURI
 		if stripQuery {
 			uri = stripQueryString(uri)
+		} else if keys, ok := additional[KeyMaskedQueryParams].(map[string]bool); ok {
+			if len(keys) > 0 {
```
MustafaSaber, line 177: see https://github.com/zalando/skipper/pull/3471/files#r2035043394

logging/access_test.go
```diff
@@ -293,3 +323,12 @@ func TestAccessLogStripQuery(t *testing.T) {
 	entry.Request.RequestURI += "?foo=bar"
 	testAccessLog(t, entry, logOutput, Options{AccessLogStripQuery: true})
 }
+
+func TestHashQueryParamValue(t *testing.T) {
+	want := uint64(3728699739546630719)
+	got := hash("foo")
+
+	if got != want {
```
MustafaSaber, line 331: why not using require here also?

ponimas: I think these tests were created before testify dependency was added to skipper.
This test makes no sense. It just tests the library function, i'll remove.

logging/access.go
ponimas: the link is broken for me

logging/access_test.go
MustafaSaber: it's nice for coverage

logging/access.go
MustafaSaber: was this one https://github.com/zalando/skipper/pull/3471#discussion_r2035043394

logging/access.go
ponimas: But this function returns string, means we need to allocate new one even for early return.

filters/accesslog/control.go
```diff
@@ -18,13 +21,22 @@ const (
...
+	// MaskedQueryParams contains the set of query parameters (keys) that are masked/obfuscated in the access log.
+	MaskedQueryParams map[string]struct{}
 }
 
 func (al *AccessLogFilter) Request(ctx filters.FilterContext) {
 	bag := ctx.StateBag()
-	bag[AccessLogEnabledKey] = al
+	if f, ok := bag[AccessLogEnabledKey]; ok {
+		a := f.(*AccessLogFilter)
+		maps.Copy(a.MaskedQueryParams, al.MaskedQueryParams)
```
AlexanderYastrebov, line 36: This modifies filter instance from parallel requests and will panic.

ponimas: @AlexanderYastrebov seems that I miss some knowlege here.
How could this modification cross the boundary between requests if the state bag is unique to each request and filters are executed sequentially?
https://github.com/zalando/skipper/blob/46828d43055dab3a93a37e21fddab44b49bfe5ce/filters/filters.go#L85-L87

ponimas: nvm, I got it 🤦

filters/accesslog/control_test.go
szuecs: I guess @MustafaSaber meant to remove the `if ti.isError` and replace it with a oneliner err check.
I don't like require/assert lib but feel free to let or change it.

szuecs:
@ponimas can you please create user visible docs: docs/reference/filters.md in [logs section](https://opensource.zalando.com/skipper/reference/filters/#logs) ?

Rest seems good to go. I classify this as major change because it modifies code paths that are already in use, which means it does not classify as completely "new" filter.

a4180p:
@szuecs updated

szuecs:
@ponimas do you want to rebase in `cw2025`? Then we can merge it or we wait after code freeze is over

szuecs:
DCO :)

a4180p:
@szuecs that's weird. All commits have sign-offs

a4180p:
👍

szuecs:
:+1:

a4180p:
👍

PR #3473, opened 2025-04-08
routing: Improve first route load signaling

ponimas:
Dataclients may load routes with arbitrary delays. When multiple dataclients are used, it's possible that one dataclient loads routes multiple times while another doesn't load any routes at all. This could incorrectly signal that the first route has been loaded, even though the routing table only contains routes from a single dataclient.

This change fixes route update logic to signal first load only after routes from all configured dataclients are received at least once.

An alternative to https://github.com/zalando/skipper/pull/3447

### DIFF
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -65,7 +65,8 @@ func (d *incomingData) log(l logging.Logger, suppress bool) {
 // communication error occurs, it re-requests the whole valid set, and continues polling.
 // Currently, the routes with the same id coming from different sources are merged in an
 // undeterministic way, but this may change in the future.
-func receiveFromClient(c DataClient, o Options, out chan<- *incomingData, quit <-chan struct{}) {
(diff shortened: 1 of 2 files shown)


routing/datasource.go
```diff
@@ -96,6 +97,7 @@ func receiveFromClient(c DataClient, o Options, out chan<- *incomingData, quit <
 			continue
 		case initial || len(routes) > 0 || len(deletedIDs) > 0:
 			var incoming *incomingData
+			once.Do(firstLoad.Done)
```
AlexanderYastrebov, line 100: I think there is a race condition - this will signal first load before actual table is updated (which happens around `r.routeTable.Store(rt)` )

You can play/test this by adding time.Sleep here.

ponimas: Yup, tests are failing exactly because of this race.

PR #3478, opened 2025-04-22
feat/metric: add new metric for skipper latency

ayruslore:
feat/metric: add new metric for skipper latency

This change involves adding a new metric to measure the overall skipper latency of a request. i.e., the time taken by a request in skipper itself, so this does not include time taken for serving for the response to client and the time taken for the backend round trip.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1335,6 +1335,7 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 		}
 
 		ctx.setResponse(rsp, p.flags.PreserveOriginal())
+		ctx.backendTime = time.Since(backendStart)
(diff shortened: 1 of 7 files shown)


proxy/proxy.go
```diff
@@ -1335,6 +1335,7 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 		}
 
 		ctx.setResponse(rsp, p.flags.PreserveOriginal())
+		ctx.backendTime = time.Since(backendStart)
```
AlexanderYastrebov, line 1338: Lets discuss. It feels like this is collected by two lines above and also it does not have request host dimension (ctx.metricsHost())

ayruslore:
PR was just to understand the metrics. We can do a proposal to expose different set of metrics

PR #3491, opened 2025-05-05
fix: Copy headers from backend in upgrade proxy

ponimas:
When a backend server responded with a non-101 status code during WebSocket upgrade attempts, the response headers were not forwarded to the client. This fix ensures all headers from the backend response are copied to the client response before writing the status code.

### DIFF
--- a/proxy/upgrade.go
+++ b/proxy/upgrade.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"maps"
 	"net"
 	"net/http"
 	"net/http/httputil"
@@ -124,11 +125,7 @@ func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
 
 	if resp.StatusCode != http.StatusSwitchingProtocols {
 		log.Debugf("Got invalid status code from backend: %d", resp.StatusCode)
-		for key, values := range resp.Header {
-			for _, value := range values {
-				w.Header().Add(key, value)
-			}
-		}
+		maps.Copy(w.Header(), resp.Header)
 		w.WriteHeader(resp.StatusCode)
 		_, err := io.Copy(w, resp.Body)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


proxy/upgrade.go
```diff
@@ -124,6 +124,11 @@ func (p *upgradeProxy) serveHTTP(w http.ResponseWriter, req *http.Request) {
 
 	if resp.StatusCode != http.StatusSwitchingProtocols {
 		log.Debugf("Got invalid status code from backend: %d", resp.StatusCode)
+		for key, values := range resp.Header {
+			for _, value := range values {
+				w.Header().Add(key, value)
+			}
+		}
```
AlexanderYastrebov, line 131: Why not https://pkg.go.dev/maps#Copy?

Also, this and below resembles why proxy.go is doing so I am wondering whether its worth it to extract the "serve backend response" logic and reuse it here.

proxy/upgrade_test.go
```diff
@@ -125,6 +126,16 @@ func TestServeHTTP(t *testing.T) {
...
 			expectedResponseStatusCode: http.StatusServiceUnavailable,
 		},
+		{
+			msg:                        "Backend responds 200 with headers",
+			route:                      `route: Path("/ws") -> "%s";`,
+			method:                     http.MethodGet,
+			backendStatusCode:          http.StatusOK,
+			backendClosesConnection:    true,
+			expectedResponseStatusCode: http.StatusOK,
+			backendHeaders:             map[string]string{"X-Header": "value", "Content-Type": "application/json"},
```
szuecs, line 136: We should also Test that we can have multiple header values for the same name

proxy/upgrade_test.go
```diff
@@ -163,6 +174,11 @@ func TestServeHTTP(t *testing.T) {
 			var clientConnClosed atomic.Bool
 
 			backend := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if ti.backendHeaders != nil {
+					for k, v := range ti.backendHeaders {
+						w.Header().Set(k, v)
```
szuecs, line 179: This won't allow multiple values for a header name.

szuecs (changes requested):

AlexanderYastrebov:
:+1:

ponimas:
👍

PR #3498, opened 2025-05-14
tracing: set OTel network.peer.address in proxy span

szuecs:
To be able to find bad backend application instances we can use group_by network.peer.address queries.


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #2220, opened 2023-02-07
proxy/proxytest: add TLS support

AlexanderYastrebov:
* introduces Config to simplify creation of test proxy
* adds helper to create certificate for given domains
* starts TLS server when certificate is present

Signed-off-by: Alexander Yastrebov <alexander.yastrebov@zalando.de>

PR #3506, opened 2025-05-19
feat/metrics : add new metrics for measuring skipper latency

ayruslore:
This change involves adding a new metric to measure the overall skipper latency of a request. i.e., the time taken by a request in skipper itself, so this does not include time taken for serving for the response to client and the time taken for the backend round trip.

skipper latency = serve duration - response duration - backend duration


proxy/proxy.go
```diff
@@ -1336,6 +1336,7 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 		}
 
 		ctx.setResponse(rsp, p.flags.PreserveOriginal())
+		ctx.roundTripDuration = time.Since(backendStart)
 		p.metrics.MeasureBackend(ctx.route.Id, backendStart)
 		p.metrics.MeasureBackendHost(ctx.route.Host, backendStart)
```
ayruslore, line 1341: Here the duration is being calculated 3 times overall. To reduce this we need to update the api to use the duration directly. Should we also do this in the same PR?

proxy/proxy.go
```diff
@@ -1391,8 +1392,10 @@ func (p *Proxy) serveResponse(ctx *context) {
 		p.tracing.setTag(ctx.proxySpan, StreamBodyEvent, StreamBodyError)
 		p.tracing.logStreamEvent(ctx.proxySpan, StreamBodyEvent, fmt.Sprintf("Failed to stream response: %v", err))
 	} else {
+		ctx.responseDuration = time.Since(start)
 		p.metrics.MeasureResponse(ctx.response.StatusCode, ctx.request.Method, ctx.route.Id, start)
```
ayruslore, line 1396: Same as above. Value is calculated twice.

skipper.go
```diff
@@ -522,6 +522,24 @@ type Options struct {
...
+	// If set, the detailed total skipper time metrics will contain the
+	// HTTP Response status code as a domain of the metric. It affects
+	// both route and host split metrics.
+	EnableSkipperLatencyStatusCodeMetric bool
+
+	// If set, the detailed total skipper time metrics will contain the
+	// HTTP method as a domain of the metric. It affects both route and
+	// host split metrics.
+	EnableSkipperLatencyMethodMetric bool
+
```
ayruslore, line 542: I made sure to expose the same metrics data as done for Serve metrics. The Skipper latency (i.e., the actual time a request took inside skipper would be very useful to be collected on all fronts.)

metrics/all_kind.go
```diff
@@ -89,6 +89,11 @@ func (a *All) MeasureResponse(code int, method string, routeId string, start tim
 	a.codaHale.MeasureResponse(code, method, routeId, start)
 }
 
+func (a *All) MeasureSkipperLatency(routeId, host, method string, code int, start time.Time, backendDuration time.Duration, responseDuration time.Duration) {
```
szuecs, line 92: why do we need skipper latency by route/host/method/status/path ?

proxy/proxy.go
szuecs: I guess "refactoring" means increasing the metrics API to also allow setting a time.Duration for these calls to only do once the time.Since() call, right?
If not we would break our API.

proxy/proxy.go
```diff
@@ -1391,8 +1392,10 @@ func (p *Proxy) serveResponse(ctx *context) {
...
 	}
+	p.metrics.MeasureSkipperLatency(ctx.route.Id, ctx.metricsHost(), ctx.request.Method, ctx.response.StatusCode, ctx.startServe, ctx.roundTripDuration, ctx.responseDuration)
```
szuecs, line 1398: in case of an error the `ctx.responseDuration` is 0.

proxy/proxy.go
ayruslore: Yes just increase in this PR. But maybe we can do a Breaking change in the next version.

proxy/proxy.go
ayruslore: Yes and I use 0 here for error response 

https://github.com/zalando/skipper/blob/c208b09485279c21ddafedd7d6e393a256386328/proxy/proxy.go#L1485-L1493

proxy/proxy.go
szuecs: We are a library and therefore we do not break our APIs

ayruslore: Then are we okay with calculating this value three times. OR do we want to add two new or three new (also including the same for response) APIs to avoid the extra calculations.

ayruslore: We can address the APIs in the next PR

proxy/context.go
```diff
@@ -51,6 +51,7 @@ type context struct {
 	routeLookup          *routing.RouteLookup
 	cancelBackendContext stdlibcontext.CancelFunc
 	logger               filters.FilterContextLogger
+	skipperDuration      time.Duration
```
AlexanderYastrebov: It is hard to follow when this changes. I think we may consider adding a stop watch field, start it when we receive request, pause if around operations we do not want to measure and report elapsed time after we served the response, see this outline https://github.com/zalando/skipper/compare/metrics/stopwatch

metrics/prometheus.go
```diff
@@ -19,6 +19,7 @@ const (
 	promFilterSubsystem    = "filter"
 	promProxySubsystem     = "backend"
 	promStreamingSubsystem = "streaming"
+	promRequestSubsystem   = "request"
```
AlexanderYastrebov: I think it should be "proxy" subsystem since we are measuring proxy component latency and `promProxySubsystem` should be renamed to `promBackendSubsystem` (need to check).

proxy/context.go
ayruslore: done. using a simple stopwatch for now

metrics/prometheus.go
ayruslore: renamed `promProxySubsystem` to `promBackendSubsystem`. Just a rename to reflect the value `backend`, all metrics remain the same. 

And then I added `promProxySubsystem = proxy` and used it for measuring skipper latency

szuecs: I guess this changes all metric names?
If so it's a breaking change for the metrics.

proxy/proxy.go
szuecs: does it make sense to have a 0 on error case?

metrics/all_kind.go
szuecs: I would rather have no dimension.
If we do not include filter latency in our measurement (I think I would prefer not to have added filter latency here), then there should not be a dependency for any kind of route specifics, so why should we have it?

proxy/proxy.go
```diff
@@ -1023,7 +1023,10 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 	ctx.proxySpan.LogKV("http_roundtrip", StartEvent)
 	req = injectClientTrace(req, ctx.proxySpan)
 
+	ctx.timer.Stop()
 	response, err := roundTripper.RoundTrip(req)
+	ctx.timer.Start()
```
szuecs: We should Reset the timer to measure skipper_response_latency and then we can sum `skipper_latency = skipper_request_latency + skipper_response_latency`

proxy/proxy.go
```diff
@@ -1023,7 +1023,10 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 	ctx.proxySpan.LogKV("http_roundtrip", StartEvent)
 	req = injectClientTrace(req, ctx.proxySpan)
 
+	ctx.timer.Stop()
```
szuecs: I guess we should store the time.Duration for skipper_request_latency and skipper_latency:
```
ctx.skipperRequestLatency = ctx.timer.Elapsed()
```

proxy/proxy.go
```diff
@@ -1606,6 +1615,12 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx.initialSpan = span
 	ctx.parentSpan = span
 
+	ctx.timer.Start()
+	defer func() {
+		ctx.timer.Stop()
+		p.metrics.MeasureSkipperLatency(ctx.route.Id, ctx.metricsHost(), ctx.request.Method, ctx.response.StatusCode, ctx.timer.Elapsed())
```
szuecs: My suggestion would be something like:
```
skipperResponseLatency := ctx.timer.Elapsed()
skipperLatency := ctx.skipperRequestLatency + skipperResponseLatency
p.metrics.MeasureSkipperLatency(skipperLatency, "total")
p.metrics.MeasureSkipperLatency(skipperResponseLatency, "response")
p.metrics.MeasureSkipperLatency(ctx.skipperRequestLatency, "request")
```

proxy/proxy.go
ayruslore: with the timer now we dont have this 0.

metrics/all_kind.go
ayruslore: okay I will remove all dimensions.

metrics/prometheus.go
ayruslore: > I guess this changes all metric names?

No. The metrics are same. Its a variable rename. Using the right variable names. 
Previously `promProxySubsystem = backend`. And I have just renamed the variable `promBackendSubsytem = backend` and also accordingly updated the usages in the metrics to keep the same metric name. 

A simple rename symbol in vs code.


And then I added a new subsystem `promProxySubsystem = proxy`. This value did not exist before. Now `promProxySubsystem` is only used for measuring skipper latency.

proxy/proxy.go
ayruslore: done. calculating both now.

proxy/proxy.go
ayruslore: good idea done

proxy/proxy.go
ayruslore: done. Using this now

metrics/stopwatch.go
```diff
@@ -0,0 +1,34 @@
+package metrics
+
+import "time"
+
+type StopWatch struct {
```
szuecs: We should also write a stopwatch_test.go to test that we measure something and Reset() works as expected and so on.
Keep in mind to have some `+/- \epsilon` and not use `==` for timings. It will be anyways a bit flaky.
Yes we spoke a bit about it today and welcome to the guilty people :)

skipper.go
```diff
@@ -522,6 +522,14 @@ type Options struct {
...
 	EnableServeStatusCodeMetric bool
 
+	// If set, detailed request handling time taken by skipper
+	// will be collected.
+	EnableSkipperLatencyRequestMetrics bool
+
+	// If set, detailed response handling time take by skipper
+	// will be collected.
+	EnableSkipperLatencyResponseMetrics bool
+
```
szuecs, line 536: A proxy table tests that shows that we measure / not measure the different configuration possibilities is a duty for this change :)

metrics/stopwatch.go
AlexanderYastrebov: We don't need to depend on the real clock and may use mocked (i.e. add `now func() time.Time` field), see e.g. leaky bucket or tokeninfo cache tests.

proxy/proxy.go
```diff
@@ -1606,6 +1617,13 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx.initialSpan = span
 	ctx.parentSpan = span
 
+	ctx.timer.Start()
+	defer func() {
+		ctx.timer.Stop()
+		skipperResponseLatency := ctx.timer.Elapsed()
+		p.metrics.MeasureSkipperLatency(ctx.skipperRequestLatency, skipperResponseLatency)
+	}()
```
AlexanderYastrebov: I have a long dream of cleaning pile of defers from ServeHTTP - lets put this to the end of the function after cancelBackendContext.

AlexanderYastrebov: OTOH this does not measure access logging, span finishing and some other stuff. If we want that then we'd need to create stopwatch at entry, measure via first defer and pass it as a parameter to the newContext.

metrics/stopwatch.go
AlexanderYastrebov: There is a single usage, not sure whether we want this as an exported type right away. 
We may consider moving this to proxy package (which is bloated though) and export when we need it a second time.

AlexanderYastrebov: Maybe it's time to start adopting internal packages https://go.dev/doc/modules/layout#package-or-command-with-supporting-packages

proxy/proxy.go
szuecs: > I have a long dream of cleaning pile of defers from ServeHTTP - lets put this to the end of the function after cancelBackendContext.
> 


Hmm why don't you like defer?
It's pretty fast and is great for making sure we run it in all kind of possible branch scenarios.

metrics/stopwatch.go
szuecs: Maybe someone wants to use it in their skipper proxy.
I really don't like internal packages.

proxy/proxy.go
ayruslore: It is fast, But now we have multiple defers after every few statements. We should consolidate and make it more readable. 

i think defer poses the same problem as with goto statements, readability. With multiple defers (yes it works like a stack), but still single defer is more preferable or just write in proper structural paradigm itself without any defers. Much better.

metrics/stopwatch.go
AlexanderYastrebov: If someone wants to use it they may copy these 20 lines. For us making it exported means it will be hard to change due to our compatibility promise.

skipper.go
ayruslore: I added tests for this accordingly in the prometheus and codahale test files

proxy/proxy.go
ayruslore: > OTOH this does not measure access logging, span finishing and some other stuff. If we want that then we'd need to create stopwatch at entry, measure via first defer and pass it as a parameter to the newContext.

I am doing the same as above. Moved the timer creating and the stop to first defer after span finish. So everything is measured now.

metrics/stopwatch.go
ayruslore: I moved it to proxy and not exporting it. I did not use any internal packages for now. I think this is okay.

ayruslore: I added tests. In my local i didnt see any flakyness but in the pipeline it is flaky. I will add mock now function

ayruslore: Fixed it now, using a mock clock

proxy/stopwatch.go
```diff
@@ -0,0 +1,38 @@
...
+
+import "time"
+
+type stopWatch struct {
+	now     func() time.Time
+	started time.Time
+	elapsed time.Duration
+}
+
+func NewStopWatch(now func() time.Time) *stopWatch {
```
AlexanderYastrebov: Let make it unexported. Also we can make it use time.Now by default (i.e. `func NewStopWatch()`) and set `now` field in the test as we dont need to mock clock in production code.

proxy/stopwatch_test.go
```diff
@@ -0,0 +1,57 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.Time = c.Time.Add(d)
+}
+
+func (c *testClock) now() time.Time {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.Time
+}
```
AlexanderYastrebov: Is ok, but we don't need this to be thread-safe and could also implement it in-place like:
```go
now, _ := time.Parse(time.RFC3339, "2022-11-10T00:36:41+01:00")

watch := NewStopWatch()
watch.now = func() time.Time {
  return now
}
...
now = now.Add(d)

```

metrics/codahale.go
```diff
@@ -27,6 +27,9 @@ const (
 	KeyResponse                   = "response.%d.%s.skipper.%s"
 	KeyResponseCombined           = "all.response.%d.%s.skipper"
 	Key5xxsBackend                = "all.backend.5xx"
+	KeySkipperLatencyTotal        = "skipperlatency.total"
+	KeySkipperLatencyRequest      = "skipperlatency.request"
+	KeySkipperLatencyResponse     = "skipperlatency.response"
```
AlexanderYastrebov: I think we should name these similar to prometheus metrics (we have `skipper_proxy_total_duration_seconds_bucket`), i.e. something like proxy.request, proxy.total, etc...

config/config.go
```diff
@@ -409,6 +411,8 @@ func NewConfig() *Config {
 	flag.BoolVar(&cfg.ServeMethodMetric, "serve-method-metric", true, "enables the HTTP method as a domain of the total serve time metric. It affects both route and host split metrics")
 	flag.BoolVar(&cfg.ServeStatusCodeMetric, "serve-status-code-metric", true, "enables the HTTP response status code as a domain of the total serve time metric. It affects both route and host split metrics")
 	flag.BoolVar(&cfg.BackendHostMetrics, "backend-host-metrics", false, "enables reporting total serve time metrics for each backend")
+	flag.BoolVar(&cfg.SkipperLatencyRequestMetrics, "skipper-latency-request-metrics", false, "enables reporting request latency metrics for skipper")
+	flag.BoolVar(&cfg.SkipperLatencyResponseMetrics, "skipper-latency-response-metrics", false, "enables reporting response latency metrics for skipper")
```
AlexanderYastrebov: I think these should be about proxy and not skipper, also singular so maybe `proxy-request-metric` and `proxy-response-metric` such that we have common naming for flags, codahale and prometheus metrics?

metrics/codahale.go
```diff
@@ -182,6 +185,17 @@ func (c *CodaHale) MeasureResponse(code int, method string, routeId string, star
 	}
 }
 
+func (c *CodaHale) MeasureSkipperLatency(requestDuration, responseDuration time.Duration) {
```
AlexanderYastrebov: MeasureProxy?

proxy/stopwatch_test.go
```diff
@@ -0,0 +1,57 @@
...
+	watch := NewStopWatch(clock.now)
+
+	watch.Start()
+	clock.add(5 * time.Millisecond)
+	watch.Stop()
+
+	elapsed := watch.Elapsed()
+	if elapsed != 5*time.Millisecond {
+		t.Errorf("Expected elapsed time to be 5ms, got %v", elapsed)
+	}
```
AlexanderYastrebov: I personally prefer to use assert.Equal, etc for brevity in the new code.

AlexanderYastrebov:
We test metric collection and stopwatch but do not test that proxy measures correctly.
To do that we can create a test route `*  -> latency("100ms") -> backendLatency("200ms") -> status(200) -> <shunt>` and test that proxy reports ~200ms for request and ~100ms for response (e.g. via [assert.InDelta](https://pkg.go.dev/github.com/stretchr/testify/assert#InDelta) or [assert.InEpsilon](https://pkg.go.dev/github.com/stretchr/testify/assert#InEpsilon)).

proxy/context.go
```diff
@@ -145,6 +148,7 @@ func newContext(
 		metrics:        &filterMetrics{impl: p.metrics},
 		proxy:          p,
 		routeLookup:    p.routing.Get(),
+		timer:          *watch,
```
AlexanderYastrebov: timer-> watch or watch -> timer

proxy/proxy.go
```diff
@@ -1536,6 +1547,9 @@ func shouldLog(statusCode int, filter *al.AccessLogFilter) bool {
 
 // http.Handler implementation
 func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	skipperStopWatch := NewStopWatch(time.Now)
+	skipperStopWatch.Start()
```
AlexanderYastrebov: timer?

proxy/stopwatch.go
ayruslore: Missed this. Not exporting the method too. 

Also Makes sense, made it to use `time.Now` as default for `now` field

proxy/stopwatch_test.go
ayruslore: sounds good. test is now simplified.

metrics/codahale.go
ayruslore: using proxyTotal, proxyRequest and proxyResponse as metrics keys now. Also updated all the naming to be consistent everywhere.

config/config.go
ayruslore: done. Since the metrics and fields all use proxy now, also update the flags accordingly

metrics/codahale.go
ayruslore: done. Everywhere updated to Proxy

proxy/stopwatch_test.go
ayruslore: done.

proxy/context.go
ayruslore: using the name proxyWatch now. It is consistent everywhere

proxy/proxy.go
ayruslore: its all using watch everywhere consistent with the type name too.

proxy/proxy_test.go
```diff
@@ -1247,6 +1248,77 @@ func TestFlusherImplementation(t *testing.T) {
...
+				continue
+			}
+			val, err := strconv.ParseFloat(parts[1], 64)
+			if err != nil {
+				t.Errorf("invalid float value in metrics line: %v", err)
+				continue
+			}
+			data[parts[0]] = val
+		}
+	}
```
AlexanderYastrebov: I am not sure its a good idea to implement prometheus parsing ad-hoc here. I think we should rather implement MockMetrics.MeasureProxy and then use m.WithMeasures like in other tests.

proxy/proxy_test.go
```diff
@@ -1247,6 +1248,77 @@ func TestFlusherImplementation(t *testing.T) {
 	}
 }
 
+func TestMeasureProxyWatch(t *testing.T) {
```
AlexanderYastrebov: Lets move this to proxy/metrics_test.go?

proxy/proxy_test.go
```diff
@@ -1247,6 +1248,77 @@ func TestFlusherImplementation(t *testing.T) {
...
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	defer tp.close()
+
+	u, _ := url.ParseRequestURI("https://www.example.org/hello")
+	r := &http.Request{URL: u, Method: "GET"}
+	w := httptest.NewRecorder()
+	tp.proxy.ServeHTTP(w, r)
```
AlexanderYastrebov: While this works, I think we prefer to issue a real http request to the test server via test client.

szuecs:
@ayruslore we need also a change in docs/operation/operation.md

proxy/proxy_test.go
```diff
@@ -1247,6 +1248,77 @@ func TestFlusherImplementation(t *testing.T) {
...
+	mo := metrics.Options{
+		Format:                     metrics.PrometheusKind,
+		EnableProxyRequestMetrics:  true,
+		EnableProxyResponseMetrics: true,
+	}
+	m := metrics.NewMetrics(mo)
+	defer m.Close()
+
+	doc := `*  -> latency("10ms") -> backendLatency("20ms") -> status(200) -> <shunt>`
+	tp, err := newTestProxyWithParams(doc, Params{Metrics: m})
```
AlexanderYastrebov: This is ok but it is preferable to use `proxytest.Config` (see https://github.com/zalando/skipper/pull/2220) see e.g. TestMetricsUncompressed

skipper.go
```diff
@@ -522,6 +522,14 @@ type Options struct {
 	// both route and host split metrics.
 	EnableServeStatusCodeMetric bool
 
+	// If set, detailed request handling time taken by skipper
+	// will be collected.
+	EnableProxyRequestMetrics bool
+
+	// If set, detailed response handling time take by skipper
+	// will be collected.
+	EnableProxyResponseMetrics bool
```
AlexanderYastrebov: Lets make these close/equal to flag help texts

szuecs: and use proper godoc here

proxy/proxy_test.go
ayruslore: Done. Makes sense. Updated it.

proxy/proxy_test.go
ayruslore: Ohh i didnt see this file. I was wondering if we dont have any metrics end-to-end tests, its in this file. Okay.

I moved it there now.

proxy/proxy_test.go
ayruslore: done.

proxy/proxy_test.go
ayruslore: Yeah the new test now follows the existing tests pattern

skipper.go
ayruslore: I have adjusted the description. Please check and let me know if it clear enough.

ayruslore:
> @ayruslore we need also a change in docs/operation/operation.md

done. Updated the operation docs too.

metrics/metricstest/metricsmock.go
```diff
@@ -7,6 +7,11 @@ import (
 	"time"
 )
 
+type MockMetricsOptions struct {
```
AlexanderYastrebov: Do we need mock options? Lets just record all three measurements unconditionally for simplicity.

proxy/metrics_test.go
```diff
@@ -62,3 +63,53 @@ func TestMetricsUncompressed(t *testing.T) {
...
+	dc := testdataclient.New([]*eskip.Route{
+		{
+			Id: "test",
+			Filters: []*eskip.Filter{
+				{Name: "latency", Args: []interface{}{"10ms"}},
+				{Name: "backendLatency", Args: []interface{}{"20ms"}},
+				{Name: "status", Args: []interface{}{200}},
+			},
+			BackendType: eskip.ShuntBackend,
+		},
```
AlexanderYastrebov: Why not use 
```go
eskip.MustParse(`test: * -> ... -> <shunt>;`)
```

AlexanderYastrebov: You can also pass routes directly to config like
```go
tp := proxytest.Config{
   Routes: eskip.MustParse(`test: * -> ... -> <shunt>;`),
   ...
}
```

metrics/metricstest/metricsmock.go
ayruslore: I am trying to just replicate the Mock as close to the original one. But okay, we can collect all three

proxy/metrics_test.go
ayruslore: I think its okay to pass the filters in a struct. Does it make a difference.?

ayruslore: updated

metrics/metricstest/metricsmock.go
ayruslore: removed

AlexanderYastrebov:
:+1:

ayruslore:
👍

PR #3514, opened 2025-05-26
Enabling OPA AST optimizations in Skipper to store AST values in in-memory store

Pushpalanka:
**Context:**
With https://github.com/open-policy-agent/opa/pull/7125 Open Policy Agent introduced an option to read AST values from in-memory storage instead of interfaces/slices.

When enabled it is expected to improve latency as this removes the time spent converting raw data values to AST during policy evaluation. The memory footprint of the store will increase, as processed AST values would take more space in memory than the corresponding raw data values, but overall memory usage of OPA might remain more stable over time, as pre-converted data is shared across evaluations and isn't recomputed for each evaluation.

With this PR, the option `enable-open-policy-agent-data-preprocessing-optimization` is supported as a Skipper configuration at OPA registry level. In other words, when enabled it will be enabled for all the OPA systems running within the same Skipper ingress.

### Benchmark test results: Negligible Impact
(against the provided split bundle data and policy test resources with minimal complexity data and policy)
```go test -bench=^BenchmarkSplitPolicyAndDataBundles$ -cpu=1 -benchmem ./filters/openpolicyagent/opaauthorizerequest ```

Attached the result from 10 runs. 
[bench_results.txt](https://github.com/user-attachments/files/20747428/bench_results.txt)

```
goos: darwin
goarch: arm64
pkg: github.com/zalando/skipper/filters/openpolicyagent/opaauthorizerequest
cpu: Apple M1
                                                               │ bench_results.txt │
                                                               │      sec/op       │
SplitPolicyAndDataBundles/default_filter                               24.20µ ± 1%
SplitPolicyAndDataBundles/with_data_preprocessing_optimization         24.39µ ± 1%
geomean                                                                24.30µ

                                                               │ bench_results.txt │
                                                               │       B/op        │
SplitPolicyAndDataBundles/default_filter                              18.50Ki ± 0%
SplitPolicyAndDataBundles/with_data_preprocessing_optimization        18.52Ki ± 0%
geomean                                                               18.51Ki

                                                               │ bench_results.txt │
                                                               │     allocs/op     │
SplitPolicyAndDataBundles/default_filter                                308.0 ± 0%
SplitPolicyAndDataBundles/with_data_preprocessing_optimization          309.0 ± 0%
geomean                                                                 308.5

```

### Benchmark test results: Visible optimization
(against the provided split bundle data and policy test resources with large datasource passed onto a function)

Attached the result from 10 runs. 
[bench_results_pre-dataprocessing.txt](https://github.com/user-attachments/files/20758114/bench_results_pre-dataprocessing.txt)

```
goos: darwin
goarch: arm64
pkg: github.com/zalando/skipper/filters/openpolicyagent/opaauthorizerequest
cpu: Apple M1
                                                               │ bench_results_pre-dataprocessing.txt │
                                                               │                sec/op                │
SplitPolicyAndDataBundles/default_filter                                                  993.8µ ± 1%
SplitPolicyAndDataBundles/with_data_preprocessing_optimization                            27.46µ ± 1%
geomean                                                                                   165.2µ

                                                               │ bench_results_pre-dataprocessing.txt │
                                                               │                 B/op                 │
SplitPolicyAndDataBundles/default_filter                                                 813.2Ki ± 0%
SplitPolicyAndDataBundles/with_data_preprocessing_optimization                           19.91Ki ± 0%
geomean                                                                                  127.2Ki

                                                               │ bench_results_pre-dataprocessing.txt │
                                                               │              allocs/op               │
SplitPolicyAndDataBundles/default_filter                                                  13.99k ± 0%
SplitPolicyAndDataBundles/with_data_preprocessing_optimization                             336.0 ± 0%
geomean                                                                                   2.168k

```

### CODE REVIEW

**Comment by Pushpalanka** [comment]
Memory profile with pre-processing,
image

Memory profile with default settings,
image

Overall memory consumption at the benchmark request rate, has reduced with pre processing, when compared to default settings.

File: filters/openpolicyagent/opaauthorizerequest/benchmark_test.go [hunk 1]
```diff
@@ -333,6 +333,70 @@ func BenchmarkMinimalPolicyBundle(b *testing.B) {
 	})
 }
 
+// Assume a policy bundle is given in the testResources directory and named "policy.tgz". It will evaluate the
```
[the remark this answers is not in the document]
Pushpalanka: I was thinking more towards, if someone needs to consider enabling the flag, they can place the bundle and compare the impact on that bundle as the impact is sensitive to policy characteristics. 

What was used in the shared results is internal demo app which has some sensitive data. Will try to craft one with anonymous data and possibly adjust the test to run with bundle split(separate data and policy bundles) in place as well.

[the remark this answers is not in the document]
mjungsbluth: I think that is generally fine but even with synthetic data you should see some improvement. The alternative is just prove it does not decrease performance at least and point to the upstream tests. 

Otherwise there is a place already to place sample bundles and one is provided here: https://github.com/zalando/skipper/tree/master/skptesting/simple-opa-bundle

To make this reproducible, it would indeed help to have sth you can just run. Bringing your own bundle (locally) is an added benefit that is also helpful but I think the other should also happen.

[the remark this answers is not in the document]
Pushpalanka: Thanks for the inputs and addressed on top of [contributions from this PR](https://github.com/zalando/skipper/pull/3524).

Noticed we have another location for sample bundles here: https://github.com/zalando/skipper/tree/master/filters/openpolicyagent/opaauthorizerequest/testResources/simple-opa-bundle with the same policies. Do we need this redundant for some reason? 

Updated the new results in the description. Based on the 10 iterations, got slightly deviating results in local machine for the same policy. 
Could we consider the option of not enabling the flag for everyone, but only for a pre-evaluated clusters when utilizing this flag?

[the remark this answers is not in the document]
Pushpalanka: Additionally updated with a sample rule + data with clear optimization impact.

mefarazath:
👍🏻

config/config.go
```diff
@@ -536,6 +537,7 @@ func NewConfig() *Config {
 	flag.BoolVar(&cfg.EnableOpenPolicyAgentCustomControlLoop, "enable-open-policy-agent-custom-control-loop", false, "when enabled skipper will use a custom control loop to orchestrate certain opa behaviour (like the download of new bundles) instead of relying on periodic plugin triggers")
 	flag.DurationVar(&cfg.OpenPolicyAgentControlLoopInterval, "open-policy-agent-control-loop-interval", openpolicyagent.DefaultControlLoopInterval, "Interval between the execution of the control loop. Only applies if the custom control loop is enabled")
 	flag.DurationVar(&cfg.OpenPolicyAgentControlLoopMaxJitter, "open-policy-agent-control-loop-max-jitter", openpolicyagent.DefaultControlLoopMaxJitter, "Maximum jitter to add to the control loop interval. Only applies if the custom control loop is enabled")
+	flag.BoolVar(&cfg.EnableOpenPolicyAgentDataPreProcessingOptimization, "enable-open-policy-agent-data-preprocessing-optimization", false, "As a latency optimization, open policy agent will read values from in-memory storage as pre converted ASTs, removing conversion overhead at evaluation time.")
```
szuecs: If this is proven valuable, does it make sense to have this flag forever?
We could also tell that this config is experimental and if successful we will have this as default without a disable possible.

Pushpalanka: I updated the latest results for two policy versions in the description. This introduces an overhead of a fraction of a micro second to the simplest policy, but gains a lot on slightly complex policies where some amount of data is passed among functions. Within rego global libraries(authorization) we use this data passing between functions a lot, hence I would expect considerable number of systems to be benefitted from this. 

I like the approach, it will allow to gauge the impact on different policies and observe the memory demand etc.. will adjust the wording to indicate experimental and coordinate to get the teams to experiment the flag.

filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
```diff
@@ -357,49 +359,72 @@ func BenchmarkMinimalPolicyBundle(b *testing.B) {
...
+		{
+			name: "default_filter",
+			createFunc: func(opts FilterOptions) (filters.Filter, error) {
+				return createOpaFilterForMultipleBundles(opts)
+			},
+		},
+		{
+			name: "with_data_preprocessing_optimization",
+			createFunc: func(opts FilterOptions) (filters.Filter, error) {
+				return createOpaFilterWithDataProcessingOptimization(opts)
```
amak3, line 377: I have the impression that the description could introduce a little bit of confusion after these changes, it says "BenchmarkSplitPolicyAndDataBundles measures the performance of the OPA authorization filter when using a pre-compiled split policy and data bundles".
It's not entirely true as the user will get two results, one with AST optimization. Wouldn't it make sense to have separate tests for these 2 use cases? One to measure the performance of the OPA authorization filter for split bundles and one for comparing it with AST optimization? What do you think?

Pushpalanka: It's not false, in the sense it's evaluating the split bundles with two different OPA configurations. 
One benefit I saw with this was it can run multiple iterations as sets of both setting, diluting any external impact from local machine processes. 
To avoid the confusion we can renaming the test. WDYT? (If we have a need to compare the results of bundle split against a single bundle, test separation make sense, but I don't see we are after that.)

Pushpalanka:
Additionally made a 1000 run of the benchmark comparison,

For the simple policy,
image

<google-sheets-html-origin>
  | ns/op _default | ns/op _optimized | B/op _default | B/op _optimized | allocs/op _default | allocs/op _optimized
-- | -- | -- | -- | -- | -- | --
Median | 24727.5 | 25016 | 18945 | 18961 | 308 | 309
Average | 26727.107 | 27058.068 | 18945.996 | 18961.234 | 308 | 309
P99 | 37737.5 | 39558.05 | 18951 | 18967 | 308 | 309
Max | 44914 | 49197 | 18953 | 18968 | 308 | 309
Min | 23375 | 23899 | 18945 | 18961 | 308 | 309
Deviation | 3703.249928 | 3840.160723 | 1.550609038 | 0.9110840104 | 0 | 0


For the slightly complex policy,
image

**Complex Policy**
<google-sheets-html-origin>
  | ns/op _default | ns/op _optimized | B/op _default | B/op _optimized | allocs/op _default | allocs/op _optimized
-- | -- | -- | -- | -- | -- | --
Median | 1021329 | 28488.5 | 832732 | 20386 | 13990 | 336
Average | 1083208.684 | 29921.284 | 832734.342 | 20386.162 | 13990.147 | 336
P99 | 1612204.06 | 36141.25 | 832751 | 20388 | 13991 | 336
Max | 2908902 | 53455 | 832797 | 20394 | 13991 | 336
Min | 985625 | 26690 | 832726 | 20385 | 13990 | 336
Deviation | 140331.5701 | 3317.752934 | 7.238607147 | 1.144143813 | 0.3542831022 | 0

filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
amak3: that makes sense, thank you 🙏

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -151,6 +153,13 @@ func WithEnableCustomControlLoop(enabled bool) func(*OpenPolicyAgentRegistry) er
 	}
 }
 
+func WithEnableDataPreProcessingOptimization(enabled bool) func(*OpenPolicyAgentRegistry) error {
+	return func(cfg *OpenPolicyAgentRegistry) error {
+		cfg.enableDataPreProcessingOptimization = enabled
```
szuecs, line 158: :+1:

szuecs:
:+1:

mefarazath:
👍🏻

PR #3524, opened 2025-06-04
Benchmark test for split opa policy and data bundles

amak3:
BenchmarkSplitPolicyAndDataBundles measures the performance of the OPA authorization filter when using a pre-compiled split policy and data bundles loaded from a .tar.gz file.

This benchmark evaluates the filter's authorization decision overhead with a pre-compiled bundles, serving as a representative use case.

A sample bundles for this benchmark are located at testResources/split-bundles.
Users can also use their own bundles.


szuecs:
:+1:

filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
```diff
@@ -432,11 +528,74 @@ func generateConfig(opaControlPlane string, decisionLogConsumer string, decision
...
+		}
+	}`, opaControlPlane, decisionLogConsumer, decisionPlugin, decisionPath))
+}
+
 type FilterOptions struct {
 	OpaControlPlaneUrl  string
 	DecisionConsumerUrl string
 	DecisionPath        string
 	BundleName          string
+	BundleNames         []string
```
mefarazath: what if we get rid of `BundleName` and always take an array of bundle names?

filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
```diff
@@ -333,6 +333,102 @@ func BenchmarkMinimalPolicyBundle(b *testing.B) {
...
+//   cd testResources/split-bundles
+//   opa build -b data -o context-data.tgz
+//   opa build -b policy -o policy.tgz
+
+// You can also use your own bundles.  If you do so, ensure that the bundlePaths,
+// and filterOpts variables are correctly configured to match your bundle and
+// the roots in .manifest files in your bundles do not overlap.
+func BenchmarkSplitPolicyAndDataBundles(b *testing.B) {
+	policyBundle := "policy"
+	dataBundle := "context-data"
```
mefarazath, line 361: We use these names in filter configuration hardcoded. Would be good to add a comment that we expect the same names or make the `generateConfigForMultipleBundles` function to use the passed bundleNames (without hardcoding the values).

filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
```diff
@@ -333,6 +333,102 @@ func BenchmarkMinimalPolicyBundle(b *testing.B) {
...
+			}
+			f.Request(ctx)
+			assert.False(b, ctx.FServed)
+			assert.NotEqual(b, 403, ctx.FResponse.StatusCode, "Expected 403 Forbidden response")
+		}
+	})
+
+}
+
+func newOpaControlPlaneServingDataAndPolicyBundles(b *testing.B, bundleFiles map[string]string) *httptest.Server {
```
mefarazath, line 405: Now that we have two methods, one that takes a map of bundles and another that takes a single bundle. My suggestion would be to keep this function (that takes the map of bundles), and refactor the usages of other function.
(This can be done in another PR too)

filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
amak3: good point, I make the `generateConfigForMultipleBundles` to use the passed bundle names

Pushpalanka (approved):

PR #3525, opened 2025-06-04
docs: Fix ingress paths mode documentation

ponimas:
Marking the default version as **default** in the description of the default kubernetes path mode because mkdocs rendering of nested list differs from Github.
Also nested list requires 4 spaces to be rendered properly. See this issue in python-markdown - https://github.com/Python-Markdown/markdown/issues/3
Screenshot 2025-06-04 at 20 54 52


AlexanderYastrebov:
I think we also want this to be readable on Github

image


I guess the best would be to get rid of a nested list or mark the default value in the sublist item instead of referencing it from below.

a4180p:
@AlexanderYastrebov  updated

szuecs:
:+1:

a4180p:
👍

PR #3527, opened 2025-06-06
routing: ignore duplicate dataclients

AlexanderYastrebov:
Ignore duplicate dataclients since this is likely a misconfiguration. 

Note that datasource still allows multiple pointer dataclients of the same type or non-equal value-type dataclients.

Routing constructor does not return error and we generally do not use panic therefore ignoring and logging an error is the only simple option.

This is also needed because datasource manages routes keyed by DataClient and to correctly signal first route load.

Follow up on #3447

### DIFF
--- a/routing/routing.go
+++ b/routing/routing.go
@@ -3,7 +3,9 @@ package routing
 import (
 	"encoding/json"
 	"fmt"
+	"maps"
 	"net/http"
+	"slices"
 	"strconv"
 	"strings"
 	"sync/atomic"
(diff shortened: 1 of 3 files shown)


routing/routing.go
```diff
@@ -259,6 +261,19 @@ func New(o Options) *Routing {
 		o.Log = &logging.DefaultLog{}
 	}
 
+	uniqueClients := make(map[DataClient]struct{}, len(o.DataClients))
+	for i, c := range o.DataClients {
+		if _, ok := uniqueClients[c]; ok {
+			o.Log.Errorf("Duplicate data clients are not allowed, ignoring client #%d: %T", i, c)
```
AlexanderYastrebov, line 267: We can make it work with duplicates as well but I don't see a usecase for it - its likely a misconfiguration that currently goes unnoticed.

szuecs: I might have an internal use case supporting eks migration, which effectively requires the knowledge of more than one clusters.
So I would prefer if you change it to allow multiple instances of the same dataclient.

AlexanderYastrebov: This supports multiple dataclient instances of the same type but excludes **equal** duplicates i.e. equal pointers or value types which is very likely a misconfguration.

szuecs: Ah ok it wasn't clear to me.

routing/routing.go
```diff
@@ -259,6 +261,19 @@ func New(o Options) *Routing {
...
+			continue
+		}
+		uniqueClients[c] = struct{}{}
+	}
+	if len(uniqueClients) != len(o.DataClients) {
+		o.Log.Errorf("Ignored %d duplicate data clients", len(o.DataClients)-len(uniqueClients))
+		o.DataClients = slices.Collect(maps.Keys(uniqueClients))
```
szuecs, line 274: Oh I see some newer Go constructs here. It looks pretty similar to ruby style chaining. 
I have to say that I have to get used to it again 😀

szuecs:
:+1:

AlexanderYastrebov:
:+1:

Issue #3530, opened 2025-06-12
RFC: deprecation of codahale metrics


szuecs:
In skipper we do not intend to break compatibility for our users (library or binary/container), but we found that the codahale metrics library was archived and https://github.com/rcrowley/go-metrics?tab=readme-ov-file#archived-as-of-april-1-2025 tells about [Prometheus](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus) or [OTel metrics](https://opentelemetry.io/docs/languages/go/instrumentation/#metrics) being replacements, which aligns with our thoughts and likely it aligns with expectations of our community.

For this change I would also re-think the general metrics.Metrics package. In order to support custom Prometheus labels we hacked around the fact that "name" and "key" of a label have to be static assigned value, because of limitations of the Prometheus client library. 
For example:
```
skipper_custom_duration_seconds_bucket{key="routes.update_latency",le="300"} 420
skipper_custom_duration_seconds_bucket{key="routes.update_latency",le="600"} 420
skipper_custom_duration_seconds_bucket{key="routes.update_latency",le="+Inf"} 420
skipper_custom_duration_seconds_sum{key="routes.update_latency"} 4.013618892000004
skipper_custom_duration_seconds_count{key="routes.update_latency"} 420

skipper_custom_gauges{key="fifo.kube_rg__example_https_redirect.active"} 0
skipper_custom_gauges{key="fifo.kube_rg__example_https_redirect.queued"} 0
skipper_custom_gauges{key="listener.accepted.connections"} 158
skipper_custom_gauges{key="listener.queued.connections"} 0
skipper_custom_gauges{key="routes.total"} 390
skipper_custom_gauges{key="routes.updated_timestamp"} 1.749711365e+09
skipper_custom_gauges{key="swarm.redis.hits"} 119640
skipper_custom_gauges{key="swarm.redis.idleconns"} 25
skipper_custom_gauges{key="swarm.redis.misses"} 0
skipper_custom_gauges{key="swarm.redis.shards"} 1
skipper_custom_gauges{key="swarm.redis.staleconns"} 0
skipper_custom_gauges{key="swarm.redis.timeouts"} 0
skipper_custom_gauges{key="swarm.redis.totalconns"} 26

[...]

PR #3531, opened 2025-06-12
metrics: add route validation metrics to track processing success/failure rates

vlktna:
### New Metrics Added

**Prometheus metrics:**
- `skipper_route_valid_routes` (counter) - Total number of successfully processed routes
- `skipper_route_invalid_routes{reason}` (counter) - Total number of invalid routes with failure reason labels

**CodaHale metrics:**
- `route.valid` (counter) - Successfully processed routes
- `route.invalid.{reason}` (counter) - Invalid routes with reason suffix

### Failure Reasons Tracked

The `reason` label/suffix captures different validation failures:
- `failed_backend_split` - Backend URL parsing errors
- `unknown_filter` - Unknown or disabled filters used
- `unknown_predicate` - Unknown predicates used
- `invalid_filter_params` - Invalid filter parameters
- `invalid_predicate_params` - Invalid predicate parameters


szuecs:
please rebase with --sign-off and force push again (DCO check)

AlexanderYastrebov:
The counters would be ever growing so I think we should rather use gauges.

Then instead of spreading metrics update all over the place like
```go
if o.Metrics != nil {
    o.Metrics.IncInvalidRoutes("unknown_filter")
}
return nil, fmt.Errorf("trying to use %q as filter, but it is only available as predicate", def.Name)
```
we may do something like:
```diff
diff --git a/routing/datasource.go b/routing/datasource.go
index 4f87c6ed..cca81782 100644
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -23,6 +23,14 @@ const (
 
 var errInvalidWeightParams = errors.New("invalid argument for the Weight predicate")
 
+type invalidDefinitionError string
+
+func (e invalidDefinitionError) Error() string { return e.Code() }
+func (e invalidDefinitionError) Code() string  { return string(e) }
+
+var errUnknownFilter = invalidDefinitionError("unknown_filter")
+// other error kinds follow...
+
 func (it incomingType) String() string {
        switch it {
        case incomingReset:
@@ -212,14 +220,14 @@ func createFilter(o *Options, def *eskip.Filter, cpm map[string]PredicateSpec) (
        spec, ok := o.FilterRegistry[def.Name]
        if !ok {
                if isTreePredicate(def.Name) || def.Name == predicates.HostName || def.Name == predicates.PathRegexpName || def.Name == predicates.MethodName || def.Name == predicates.HeaderName || def.Name == predicates.HeaderRegexpName {
-                       return nil, fmt.Errorf("trying to use %q as filter, but it is only available as predicate", def.Name)
+                       return nil, fmt.Errorf("%w: trying to use %q as filter, but it is only available as predicate", errUnknownFilter, def.Name)
                }
 
                if _, ok := cpm[def.Name]; ok {
-                       return nil, fmt.Errorf("trying to use %q as filter, but it is only available as predicate", def.Name)
+                       return nil, fmt.Errorf("%w: trying to use %q as filter, but it is only available as predicate", errUnknownFilter, def.Name)
                }
 
-               return nil, fmt.Errorf("filter %q not found", def.Name)
+               return nil, fmt.Errorf("%w: filter %q not found", errUnknownFilter, def.Name)
        }
 
        start := time.Now()
@@ -512,6 +520,11 @@ func processRouteDefs(o *Options, defs []*eskip.Route) (routes []*Route, invalid
                } else {
                        invalidDefs = append(invalidDefs, def)
                        o.Log.Errorf("failed to process route %s: %v", def.Id, err)
+
+                       var defErr invalidDefinitionError
+                       if errors.As(err, &defErr) && o.Metrics != nil {
+                               o.Metrics.IncCounter(defErr.Code()) // actually count them per code and update gauges outside of the loop
+                       }
                }
        }
        return
```
to handle metrics update in one place

vlktna:
> The counters would be ever growing

@AlexanderYastrebov The point is precisely to let the counters grow over time because I’ll calculate the error rate over a specific period. Gauges only allow calculating the error rate at the exact moment the measurement is taken, not over intervals.

routing/datasource_test.go
```diff
@@ -340,6 +340,49 @@ func TestMetrics(t *testing.T) {
 	})
 }
 
+func TestRouteValidationMetrics(t *testing.T) {
```
szuecs, line 348: please also test with Prometheus, because https://github.com/zalando/skipper/issues/3530

szuecs:
:+1:

szuecs:
Let's try this, but for the records, I think gauge would be better to use in this case, because I think we want no rate that depends on route update latency, but just values that can calculate a percentage.

routing/datasource.go
```diff
@@ -519,6 +536,11 @@ func processRouteDefs(o *Options, defs []*eskip.Route) (routes []*Route, invalid
 		} else {
 			invalidDefs = append(invalidDefs, def)
 			o.Log.Errorf("failed to process route %s: %v", def.Id, err)
+
+			var defErr invalidDefinitionError
+			if errors.As(err, &defErr) && o.Metrics != nil {
+				o.Metrics.IncInvalidRoutes(defErr.Code())
+			}
```
AlexanderYastrebov, line 545: Here I think we should have else branch to count other types of errors (that may appear in the future) with an "other" or "generic" error code.

szuecs:
:+1:

ayruslore:
👍

PR #3535, opened 2025-06-22
feature: letsencrypt http.Handler

### LINKED ISSUE #1786: ACME support
We should build a filter that supports ACME to do the letsencrypt challenge automatically

Pointers to some docs, that might help
- golang.org/x/crypto/acme
- golang.org/x/crypto/acme/autocert
- Example: https://github.com/bradfitz/autocertdelegate/blob/master/autocertdelegate.go

szuecs:
feature: letsencrypt http.Handler integration via autocert with different storage/cache systems to be used by different deployments

Test:
I am running this now since a while on my personal websites and completely replaced certbot+apache+cron to automate certs via skipper+curl+cron.

ref: closes https://github.com/zalando/skipper/issues/1786

### DIFF
--- a/net/letsencrypt.go
+++ b/net/letsencrypt.go
@@ -1,0 +1,137 @@
+package net
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"net"
+	"net/http"
+	"regexp"
+	"strings"
+	"sync"
+
+	"golang.org/x/crypto/acme"
+	"golang.org/x/crypto/acme/autocert"
+)
+
+type InmemoryCache struct {
+	m sync.Map
+}
+
+func (ic *InmemoryCache) Get(ctx context.Context, key string) ([]byte, error) {
+	if dat, ok := ic.m.Load(key); !ok {
+		return nil, fmt.Errorf("missing key %q", key)
+	} else {
+		if data, ok := dat.([]byte); !ok {
+			return nil, fmt.Errorf("failed to convert %q to []byte", dat)
+		} else {
+			return data, nil
+		}
+	}
+}
+
+func (ic *InmemoryCache) Put(ctx context.Context, key string, data []byte) error {
+	ic.m.Store(key, data)
+	return nil
+}
+
+func (ic *InmemoryCache) Delete(ctx context.Context, key string) error {
+	ic.m.Delete(key)
+	return nil
+}
+
+type RemoteCache struct {
+	Client *RedisRingClient
+}
+
+func (rc *RemoteCache) Get(ctx context.Context, key string) ([]byte, error) {
+	res, err := rc.Client.Get(ctx, key)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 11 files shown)


skipper.go
```diff
@@ -1511,6 +1517,11 @@ func (o *Options) filterRegistry() filters.Registry {
 
 func (o *Options) TlsConfig(cr *certregistry.CertRegistry) (*tls.Config, error) {
 
+	// TODO(sszuecs): does it make sense or do we want to chain TLSConfigs?
+	if o.Letsencrypt != nil {
+		return o.Letsencrypt.TLSConfig(), nil
```
szuecs: @MustafaSaber @a4180p wdyt?
chain it or return here?

szuecs: Please check this function. I wrote a one letsencrypt test for it and did not change the tests that existed so we do not break anyone.

skipper.go
```diff
@@ -2293,6 +2308,27 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
+					Client: valkeyRing,
+				})
+
+			case redisRing != nil:
+				o.Letsencrypt.SetCache(&skpnet.RemoteCache{
+					Client: redisRing,
+				})
+
+			default:
+				// nothing to do, see config/config.go for other autocert.Cache implementations
```
a4180p, line 2327: the cache will be nil in that case, we postponed it in config.go

net/letsencrypt.go
```diff
@@ -0,0 +1,184 @@
...
+	return &DirCache{
+		cache: autocert.DirCache(dir),
+		dir:   dir,
+	}
+}
+
+func (d *DirCache) Get(ctx context.Context, key string) ([]byte, error) {
+	val, err := d.cache.Get(ctx, key)
+	if err != nil {
+		logrus.Errorf("Get %q -> %v", key, err)
```
a4180p, line 33: we must return `autocert.ErrMissCache` if cerificate is not found in cache or certificate will not be created in case of cache miss
https://cs.opensource.google/go/x/crypto/+/refs/tags/v0.57.0:acme/autocert/autocert.go;l=304-314

a4180p (commented):
I think we should also add docs for the feature, but it could be done later.

net/letsencrypt.go
szuecs: Interesting, but I think we do because we wrap here the autocert.DirCache and it will return this kind of error. So the error returned will be this.

https://cs.opensource.google/go/x/crypto/+/refs/tags/v0.57.0:acme/autocert/cache.go;l=59

szuecs: fixed in the other implementations: RemoteCache and InmemoryCache

skipper.go
szuecs: oh wrong indentation of the default :)

szuecs:
:+1:

a4180p (approved):

PR #3539, opened 2025-06-23
docs: improve clusterRatelimit filter documentation

vlktna:
- fix clusterRatelimit filter example
- mention that filters having the same ratelimit group must use the same number of requests and period parameter values.


AlexanderYastrebov:
:+1:

vlktna:
👍

PR #3540, opened 2025-06-26
opa-jwt-token-cache-verification-with-benchmark-tests

amak3:
With https://github.com/open-policy-agent/opa/pull/7274 Open Policy Agent added configurable token cache to io.jwt token verification built-in functions to improve performance. See [changelog](https://github.com/open-policy-agent/opa/releases/tag/v1.1.0). 

This PR contains benchmark test changes to compare JWT validation with and without cache.

### DIFF
--- a/filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
+++ b/filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
@@ -124,6 +124,7 @@ func BenchmarkMinimalPolicyWithDecisionLogs(b *testing.B) {
 		BundleNames:         []string{testBundleName},
 		DecisionLogging:     true,
 		ContextExtensions:   "",
+		JwtCache:            false,
 	}
 	f, err := createOpaFilter(filterOpts)
 	require.NoError(b, err)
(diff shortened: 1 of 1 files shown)


filters/openpolicyagent/openpolicyagent.go
```diff
@@ -174,6 +188,29 @@ func WithControlLoopMaxJitter(maxJitter time.Duration) func(*OpenPolicyAgentRegi
 	}
 }
 
+func (registry *OpenPolicyAgentRegistry) initializeCache() {
+	interQueryBuiltinCacheConfig, err := cache.ParseCachingConfig(createJwtCacheConfig(registry.jwtCacheMaxNumEntries))
```
mjungsbluth: This seems a bit indirect (parsing a JSON to produce a Go struct). Can we create the config directly?

config/config.go
```diff
@@ -545,6 +546,7 @@ func NewConfig() *Config {
 	flag.Int64Var(&cfg.OpenPolicyAgentMaxRequestBodySize, "open-policy-agent-max-request-body-size", openpolicyagent.DefaultMaxRequestBodySize, "Maximum number of bytes from a http request body that are passed as input to the policy")
 	flag.Int64Var(&cfg.OpenPolicyAgentRequestBodyBufferSize, "open-policy-agent-request-body-buffer-size", openpolicyagent.DefaultRequestBodyBufferSize, "Read buffer size for the request body")
 	flag.Int64Var(&cfg.OpenPolicyAgentMaxMemoryBodyParsing, "open-policy-agent-max-memory-body-parsing", openpolicyagent.DefaultMaxMemoryBodyParsing, "Total number of bytes used to parse http request bodies across all requests. Once the limit is met, requests will be rejected.")
+	flag.IntVar(&cfg.OpenPolicyAgentJwtCacheMaxNumEntries, "open-policy-agent-jwt-cache-max-num-entries", openpolicyagent.DefaultJwtCacheMaxNumEntries, "Maximum number of JWTs to cache for OPA JWT filters")
```
mjungsbluth: This flag is very specific. Do we have other built-ins that actually use the value cache. Enabling this also implicitly disables that we can configure the value cache via the OPA config. 

I am wondering if we should do this a bit more generically (like enable central value cache). 

Apart from this here, we could also read the configuration from the opaconfig that is managed in the registry.

mjungsbluth:
So be basically see a 50% improvement with the cache enabled. 

Could we compare this with `benchstat` (i.e. do two runs producing two different files) so that we can output the comparison side by side?

szuecs:
> So be basically see a 50% improvement with the cache enabled.
> 
> Could we compare this with `benchstat` (i.e. do two runs producing two different files) so that we can output the comparison side by side?

I would recommend to run benchmarks like this to show also that we can scale by number of CPU:

```
% go test -bench=BenchmarkStackListener -benchmem ./queuelistener -run='^$' -cpu 1,2,4,8 -count 1 2>/dev/null | tee old.txt
```

amak3:
New pull request created: https://github.com/zalando/skipper/pull/3572

PR #3544, opened 2025-07-04
proxy/tracing: loopback spans bugfix

ponimas:
When a request is looped back, all tracing operations (such as tagging and creating new spans) are performed on the original request's span rather than on the loopback span. This results in inaccurate tracing data: tags are overwritten, and the ingress span incorrectly appears to have child spans that actually belong to the inner loopback spans.

This PR adds a test case to illustrate the issue, along with a potential bug fix.

The trace produced by the request to the routegroup below

Screenshot 2025-07-09 at 09 17 24
Both loopbacks are children of the ingress span
<details>
<summary>routegroup definition</summary>

```yaml
---
apiVersion: zalando.org/v1
kind: RouteGroup
metadata:
  name: loopback-test
spec:
  hosts:
  - loopback-test.playground.zalan.do
  backends:
  - name: loopback
    type: loopback
  - name: shunt
    type: shunt
  routes:
  - pathSubtree: /
    filters:
    - appendRequestHeader("A", "B")
    backends:
    - backendName: loopback
  - pathSubtree: /
    predicates:
    - Header("A","B")
    filters:
    - setRequestHeader("A", "C")
    - tracingTag("loop", "first")
    backends:
    - backendName: loopback
  - pathSubtree: /
    predicates:
    - Header("A","C")
    filters:
    - tracingTag("loop", "second")
    - inlineContent("boo")
    - status(200)
    backends:
    - backendName: shunt
```

</details>

The trace after the bugfix
image

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1255,6 +1255,10 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 		p.setCommonSpanInfo(ctx.Request().URL, ctx.Request(), loopSpan)
 		ctx.parentSpan = loopSpan
 
+		r := loopCTX.Request()
+		r = r.WithContext(ot.ContextWithSpan(r.Context(), loopSpan))
+		loopCTX.request = r
+
 		defer loopSpan.Finish()
 
 		if err := p.do(loopCTX, loopSpan); err != nil {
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

a4180p:
👍

PR #3547, opened 2025-07-07
Add LoopbackWithResponse to proxy/context

xavivars:


filters/builtin/loopbackifstatus.go
```diff
@@ -0,0 +1,63 @@
...
+// The filter replaces the response coming from the backend or the following filters.
+func NewLoopbackIfStatus() filters.Spec {
+	return &loopbackIfStatusSpec{}
+}
+
+func (c loopbackIfStatusFilter) Request(ctx filters.FilterContext) {
+
+}
+
+func (c loopbackIfStatusFilter) Response(context filters.FilterContext) {
```
szuecs: You want to use ptr receiver here or you copy string+int all the time.

AlexanderYastrebov:
Hello. What is the usecase and example usage of it?

xavivars:
@AlexanderYastrebov: an example usage can be demonstrated with the filter (that I was initially not planning to submit): I need to perform an "internal redirect" (without it being exposed to the client) given some information I only have from the response. 

Conceptually, very similar to the `inlineContentIfStatus`, but instead of returning inline content, I need to loopback the request after changing some things from the original request (if using the filter in the PR, the path).

`teeLoopback`, [introduced 5 years ago](https://github.com/zalando/skipper/pull/1443), filter already added all the building blocks required for this use case, but it was blindly discarding the response. So my original proposal (see first commit) was to expose the response. My second commit adds a filter to showcase how it could be used (not necessarily how we plan to use it internally, but very close to)

filters/builtin/loopbackifstatus.go
xavivars: Fixed.

szuecs:
> @AlexanderYastrebov: an example usage can be demonstrated with the filter (that I was initially not planning to submit): I need to perform an "internal redirect" (without it being exposed to the client) given some information I only have from the response.
> 
> Conceptually, very similar to the `inlineContentIfStatus`, but instead of returning inline content, I need to loopback the request after changing some things from the original request (if using the filter in the PR, the path).
> 
> `teeLoopback`, [introduced 5 years ago](https://github.com/zalando/skipper/pull/1443), filter already added all the building blocks required for this use case, but it was blindly discarding the response. So my original proposal (see first commit) was to expose the response. My second commit adds a filter to showcase how it could be used (not necessarily how we plan to use it internally, but very close to)

Yeah teeLoopback is an implementation of shadow traffic. https://opensource.zalando.com/skipper/tutorials/shadow-traffic/
So the idea is different because you want to test a not production application with current production traffic, so you do  want to discard the response.

A simple use case I understand from your filter would be a fallback if preferred option failed.
For example you have /old-login and /new-login. You try first /new-login, but if the cookie is not encrypted by the new scheme you try /old-login

xavivars:
> A simple use case I understand from your filter would be a fallback if preferred option failed.

Yes, that's exactly the use case we have, and if Loopback wasn't discarding the response* we would have everything we need.

*The reason I'm saying if Loopback wasn't discarding the response is because, IMHO, while it makes total sense that in something like the `teeLoopback` filter you want to discard the response, that could have been implemented in the filter itself, and not in the `loopback` function. 

And I think that would make much more sense conceptually:
- Loopback in filterContext behaving similarly to `<loopback>` on routes definition
- `teeLoopback` filter discarding the response itself, exactly how `tee` [is also discarding the response itself](https://github.com/zalando/skipper/blob/master/filters/tee/tee.go#L227)

The reason I didn't do that was because introducing that change would potentially be a backwards incompatible change, but I think it would make more sense. Even, as a last resource, fully deprecate `Loopback` and create a new `DoLoopback` action on filterContext (without discarding response) and use it also in `teeLoopback`.

> So the idea is different because you want to test a not production application with current production traffic, so you do want to discard the response.

xavivars:
Any updates?

xavivars:
@AlexanderYastrebov  could you take a look at this? @szuecs asked me to ask you (in the Skipper slack 😅 )

filters/builtin/loopbackifstatus.go
```diff
@@ -0,0 +1,63 @@
...
+// NewLoopbackIfStatus Creates a filter spec for the loopbackIfStatus() filter.
+//
+//	r: * -> loopbackIfStatus(401, "/new-path") -> "https://www.example.org";
+//
+// It accepts two arguments: the statusCode code to and the path to change the request to.
+//
+// The filter replaces the response coming from the backend or the following filters.
+func NewLoopbackIfStatus() filters.Spec {
+	return &loopbackIfStatusSpec{}
+}
```
AlexanderYastrebov, line 21: Please add tests for this new filter to verify that it works as expected.

filters/builtin/loopbackifstatus.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (c *loopbackIfStatusFilter) Response(context filters.FilterContext) {
+	if context.Response().StatusCode == c.statusCode {
+		context.Request().URL.Path = c.path
+		context.LoopbackWithResponse()
+	}
+}
+
+func (s *loopbackIfStatusSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
```
AlexanderYastrebov: Please move CreateFilter and Name to the top before filter implementation like we have it for all other filters.

proxy/context.go
```diff
@@ -300,10 +300,18 @@ func (c *context) Split() (filters.FilterContext, error) {
...
+func (c *context) LoopbackWithResponse() {
+	c.loopbackInternal(true)
+}
+
+func (c *context) loopbackInternal(keepResponse bool) {
 	loopSpan := c.tracer.StartSpan(c.proxy.tracing.initialOperationName, opentracing.ChildOf(c.parentSpan.Context()))
 	defer loopSpan.Finish()
 	err := c.proxy.do(c, loopSpan)
-	if c.response != nil && c.response.Body != nil {
+	if c.response != nil && c.response.Body != nil && !keepResponse {
```
AlexanderYastrebov: minor: I'd name it `discardResponse` and have condition like 
```go
if discardResponse && c.response != nil && c.response.Body != nil {
```

xavivars:
Thanks for the review, @AlexanderYastrebov !

I fixed all your suggestions

patch.diff
```diff
@@ -0,0 +1,75 @@
+diff --git a/eskipfile/remote.go b/eskipfile/remote.go
```
AlexanderYastrebov: This seems like a leftover file

filters/filters.go
```diff
@@ -131,6 +131,9 @@ type FilterContext interface {
 	// Performs a new route lookup and executes the matched route if any
 	Loopback()
 
+	// Performs a new route lookup and executes the matched route if , keeping the response
```
AlexanderYastrebov: `if any`

xavivars:
Sorry about that! fixed!

proxy/proxytest/loopbackifstatus_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/filters/builtin"
+	"io"
+	"net/url"
+	"testing"
+)
+
+func TestLoopbackWithResponse(t *testing.T) {
```
AlexanderYastrebov: Please move this test to proxy/loopback_test.go

AlexanderYastrebov:
The build fails, please run tests locally before pushing
```
$ go test ./proxy/...
# github.com/zalando/skipper/proxy/proxytest [github.com/zalando/skipper/proxy/proxytest.test]
proxy/proxytest/loopbackifstatus_test.go:6:2: "github.com/zalando/skipper/filters" imported and not used
proxy/proxytest/loopbackifstatus_test.go:18:11: undefined: loopbackIfStatusSpec
```

xavivars:
Tests should be fixed now

AlexanderYastrebov:
:+1:

filters/builtin/loopbackifstatus.go
```diff
@@ -0,0 +1,67 @@
...
+//
+// It accepts two arguments: the statusCode code to and the path to change the request to.
+//
+// The filter replaces the response coming from the backend or the following filters.
+func NewLoopbackIfStatus() filters.Spec {
+	return &loopbackIfStatusSpec{}
+}
+
+func (s *loopbackIfStatusSpec) Name() string {
+	return "loopbackIfStatus"
```
MustafaSaber: It's better to add this into `filters.go` as a constant and use it her.

filters/builtin/loopbackifstatus_test.go
```diff
@@ -0,0 +1,115 @@
...
+	// Not enough arguments
+	_, err = spec.CreateFilter([]interface{}{401})
+	if err == nil {
+		t.Error("expected error for missing path argument")
+	}
+}
+
+func TestLoopbackIfStatus(t *testing.T) {
+
+	br := eskip.MustParse(`LOOPBACK_404: Path("/loopback-404") -> inlineContent("404 Not Found") -> status(200) -> <shunt>`)
```
MustafaSaber: `status(200)` after `inlineContent` doesn't really make sense so it's better to remove.

MustafaSaber (commented):
Added some nitpicks

MustafaSaber:
retriggered pipeline to check tests

MustafaSaber:
Can you also please rebase? There is a test failing with check-race that should be fixed

xavivars:
Thanks @MustafaSaber ! 

It should be solved now.

MustafaSaber:
👍

MustafaSaber:
Will request another review from the team, thanks for the contribution!

xavivars:
It seems there's still a review missing

MustafaSaber:
> It seems there's still a review missing

sorry the team is a bit busy right now 😅, in the mean time can you squash commits into a single commit with a nice description?

filters/builtin/loopbackifstatus.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	return &f, nil
+}
+
+func (c *loopbackIfStatusFilter) Request(ctx filters.FilterContext) {
+
+}
+
+func (c *loopbackIfStatusFilter) Response(context filters.FilterContext) {
```
szuecs: Please change `c` to `f` for readability and `context` to `ctx`

szuecs (commented):
Not much to change from my side. Please squash the commits into one and change names as suggested.

filters/filtertest/example_test.go
```diff
@@ -16,19 +16,22 @@ package filtertest_test
...
 	"github.com/zalando/skipper/routing/testdataclient"
-	"log"
 )
 
 type customFilter struct{}
 
 func (f *customFilter) Request(ctx filters.FilterContext) {
 	ctx.StateBag()["filter called"] = true
+
+	ctx.(proxy.context)
```
szuecs: ?

xavivars: I don't know how this ended up here after my squash....

MustafaSaber: It's not removed

szuecs: yes and GH Actions also see it so not only UI cache. @xavivars

xavivars: I'm really sorry about that.

xavivars: Should be fixed now.

szuecs:
:+1:

MustafaSaber:
👍

PR #3549, opened 2025-07-07
doc: metrics-flavour args were not clear how to set "both"

szuecs:
We can not set  `skipper -metrics-flavour codahale -metrics-flavour prometheus` as I would expect from the args doc.

We have to set `skipper -metrics-flavour codahale,prometheus`


AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #3550, opened 2025-07-07
metrics: add route validation gauge metrics

vlktna:
**Prometheus metrics:**
- `skipper_route_invalid_routes_gauge{reason}` (gauge) - Current number of invalid routes by reason

**CodaHale metrics:**
- `route.invalid.gauge.{reason}` (gauge) - Current invalid routes count by reason

**Failure Reasons Tracked:**
- `failed_backend_split` - Backend URL parsing errors
- `unknown_filter` - Unknown or disabled filters used
- `unknown_predicate` - Unknown predicates used
- `invalid_filter_params` - Invalid filter parameters
- `invalid_predicate_params` - Invalid predicate parameters
- Any other validation errors that may occur

## How error rate will be calculated

- **Graph Representation**: Displays a graph with points over specific time intervals.

- **Calculation**: `(routes.total - sum(route.invalid.gauge.%s)) / routes.total`, where `routes.total` is an existed gauge metric

### DIFF
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -34,6 +34,7 @@ var (
 	errUnknownPredicate       = invalidDefinitionError("unknown_predicate")
 	errInvalidPredicateParams = invalidDefinitionError("invalid_predicate_params")
 	errFailedBackendSplit     = invalidDefinitionError("failed_backend_split")
+	errInvalidMatcher         = invalidDefinitionError("invalid_matcher")
 )
 
 func (it incomingType) String() string {
@@ -509,10 +510,6 @@ func processRouteDef(o *Options, cpm map[string]PredicateSpec, def *eskip.Route)
 		return nil, err
 	}
 
-	if o.Metrics != nil {
-		o.Metrics.IncValidRoutes()
-	}
-
 	return r, nil
 }
 
@@ -529,6 +526,8 @@ func mapPredicates(cps []PredicateSpec) map[string]PredicateSpec {
 // processes a set of route definitions for the routing table
 func processRouteDefs(o *Options, defs []*eskip.Route) (routes []*Route, invalidDefs []*eskip.Route) {
 	cpm := mapPredicates(o.Predicates)
+	reasonCounts := make(map[string]int)
+
 	for _, def := range defs {
 		route, err := processRouteDef(o, cpm, def)
 		if err == nil {
@@ -538,13 +537,19 @@ func processRouteDefs(o *Options, defs []*eskip.Route) (routes []*Route, invalid
 			o.Log.Errorf("failed to process route %s: %v", def.Id, err)
 
 			var defErr invalidDefinitionError
-			if errors.As(err, &defErr) && o.Metrics != nil {
-				o.Metrics.IncInvalidRoutes(defErr.Code())
-			} else if o.Metrics != nil {
-				o.Metrics.IncInvalidRoutes("other")
+			reason := "other"
+			if errors.As(err, &defErr) {
+				reason = defErr.Code()
 			}
+
+			reasonCounts[reason]++
 		}
 	}
+
+	if o.Metrics != nil {
+		o.Metrics.UpdateInvalidRoute(reasonCounts)
+	}
+
 	return
 }
 
@@ -611,6 +616,10 @@ func receiveRouteMatcher(o Options, out chan<- *routeTable, quit <-chan struct{}
 			for _, err := range errs {
 				o.Log.Error(err)
 				invalidRouteIds[err.ID] = struct{}{}
+			}
+
+			if o.Metrics != nil {
+				o.Metrics.UpdateInvalidRoute(map[string]int{errInvalidMatcher.Code(): len(errs)})
 			}
 
 			for i := range routes {
--- a/routing/datasource_test.go
+++ b/routing/datasource_test.go
@@ -404,15 +404,15 @@ func testRouteValidationMetricsWithMock(t *testing.T, routes string, expectedVal
 	defer r.Close()
 	<-r.FirstLoad()
 
-	metrics.WithCounters(func(counters map[string]int64) {
-		if counters["route.valid"] != expectedValid {
-			t.Errorf("Expected %d valid routes, got %d", expectedValid, counters["route.valid"])
+	metrics.WithGauges(func(gauges map[string]float64) {
+		if gauges["routes.total"] != float64(expectedValid) {
+			t.Errorf("Expected %d valid routes, got %f", expectedValid, gauges["routes.total"])
 		}
 
 		for reason, expected := range expectedInvalid {
 			key := "route.invalid." + reason
-			if counters[key] != expected {
-				t.Errorf("Expected %d %s errors, got %d", expected, reason, counters[key])
+			if gauges[key] != float64(expected) {
+				t.Errorf("Expected %d %s errors, got %f", expected, reason, gauges[key])
 			}
 		}
 	})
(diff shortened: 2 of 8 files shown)


szuecs:
please add also a section to docs/operations/operations.md

vlktna:
UPD 

- Deleted Counter Metrics. Removed the `route.valid` and `route.invalid` metrics, along with their associated tests. 
- Renamed the `skipper_route_invalid_routes_gauge` metric to `skipper_route_invalid` in both the code and tests.

routing/datasource.go
```diff
@@ -529,22 +525,32 @@ func mapPredicates(cps []PredicateSpec) map[string]PredicateSpec {
...
+	reasonCounts := make(map[string]int)
+
 	for _, def := range defs {
 		route, err := processRouteDef(o, cpm, def)
 		if err == nil {
 			routes = append(routes, route)
 		} else {
 			invalidDefs = append(invalidDefs, def)
-			o.Log.Errorf("failed to process route %s: %v", def.Id, err)
+			if o.Log != nil {
```
AlexanderYastrebov: Why do we add nil check here? It creates confusion signalling that log can be nil while it was never nil before.

routing/datasource.go
```diff
@@ -529,22 +525,32 @@ func mapPredicates(cps []PredicateSpec) map[string]PredicateSpec {
...
+			if errors.As(err, &defErr) {
+				reason = defErr.Code()
 			}
+
+			reasonCounts[reason]++
 		}
 	}
+
+	if o.Metrics != nil {
+		o.Metrics.UpdateInvalidRoute(reasonCounts)
```
AlexanderYastrebov, line 550: We can also have invalid routes in this code path after returning from processRouteDefs
https://github.com/zalando/skipper/blob/0ad4457c1757bba8cb6d07c805a3e86e4c749960/routing/datasource.go#L606-L614

I think we can re-use UpdateInvalidRoute if we choose different reason codes there.

routing/datasource.go
```diff
@@ -608,9 +612,15 @@ func receiveRouteMatcher(o Options, out chan<- *routeTable, quit <-chan struct{}
...
-				o.Log.Error(err)
-				invalidRouteIds[err.ID] = struct{}{}
+			if len(errs) > 0 {
+				for _, err := range errs {
+					o.Log.Error(err)
+					invalidRouteIds[err.ID] = struct{}{}
+				}
+
+				if o.Metrics != nil {
+					o.Metrics.UpdateInvalidRoute(map[string]int{"matcher_error": len(errs)})
```
AlexanderYastrebov: Lets define `errInvalidMatcher` to have this all unified and use like:
```go
var errInvalidMatcher         = invalidDefinitionError("invalid_matcher")
...
if o.Metrics != nil && len(errs) > 0 {
    o.Metrics.UpdateInvalidRoute(map[string]int{errInvalidMatcher.Code(): len(errs)})
}
```

In theory we could also add `Unwrap` to the `definitionError` and then use `invalidDefinitionError` wrapper for `Original` to have distinct errors within `newMatcher` but lets keep it simple for now.

AlexanderYastrebov: Thinking about `len(errs) > 0` condition I think we have an edge case here and in processRoutesDefs that we do not reset gauges to zero. 

So if value for a certain error code was not zero and then routes were fixed we should set value to zero. 
Here we can remove `len(errs) > 0` and in processRoutesDefs set zeroes explicitly for all defined error codes.

routing/datasource_test.go
```diff
@@ -510,3 +512,227 @@ func (s slowCreateSpec) CreateFilter(args []interface{}) (filters.Filter, error)
...
+		routes         string
+		expectedValid  int
+		expectedCounts map[string]int
+	}{
+		{
+			name: "various error types",
+			routes: `
+				validRoute: Path("/valid") -> "https://example.org";
+				invalidBackend1: Path("/bad1") -> "invalid-url";
+				invalidBackend2: Path("/bad2") -> "another-invalid-url";
```
szuecs: A second test to detect another "network backend" does not make sense. If you want you could add `<unknown_backend>` as backend to have a different backend error tested.

routing/datasource_test.go
```diff
@@ -510,3 +512,227 @@ func (s slowCreateSpec) CreateFilter(args []interface{}) (filters.Filter, error)
...
+				unknownFilter1: Path("/uf1") -> unknownFilter() -> "https://example.org";
+				unknownFilter2: Path("/uf2") -> anotherUnknownFilter() -> "https://example.org";
```
szuecs: unknownFilter2 does not provide more value than the unknownFilter1, so I would drop this test.

routing/datasource_test.go
```diff
@@ -510,3 +512,227 @@ func (s slowCreateSpec) CreateFilter(args []interface{}) (filters.Filter, error)
...
+		if allReasonCountsFound {
+			break
+		}
+
+		select {
+		case <-timeout:
+			// Final check with error reporting
+			for reason, expectedCount := range expectedCounts {
+				if expectedCount > 0 {
+					expectedLine := fmt.Sprintf("skipper_route_invalid{reason=\"%s\"} %d", reason, expectedCount)
```
szuecs: better:
```
fmt.Sprintf(`skipper_route_invalid{reason="%s"} %d`, reason, expectedCount)
```

routing/datasource_test.go
```diff
@@ -510,3 +512,227 @@ func (s slowCreateSpec) CreateFilter(args []interface{}) (filters.Filter, error)
...
+				return
+			case <-ticker.C:
+				continue
+			}
+		}
+
+		allReasonCountsFound := true
+		for reason, expectedCount := range expectedCounts {
+			if expectedCount > 0 {
+				expectedLine := fmt.Sprintf("skipper_route_invalid{reason=\"%s\"} %d", reason, expectedCount)
```
szuecs: similar to https://github.com/zalando/skipper/pull/3550/files#r2207070355

routing/datasource_test.go
```diff
@@ -510,3 +512,227 @@ func (s slowCreateSpec) CreateFilter(args []interface{}) (filters.Filter, error)
...
+		}
+
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			t.Fatal(err)
+		}
+		output = string(body)
+
+		// Check if metrics match expected values
+		expectedRoutesTotalLine := fmt.Sprintf("skipper_custom_gauges{key=\"routes.total\"} %d", expectedValid)
```
szuecs: similar to https://github.com/zalando/skipper/pull/3550/files#r2207070355

routing/datasource_test.go
```diff
@@ -457,14 +457,16 @@ func testRouteValidationMetricsWithPrometheus(t *testing.T, routes string, expec
 	if err != nil {
 		t.Fatal(err)
 	}
+	output := string(body)
 
-	validPattern := fmt.Sprintf(`skipper_route_valid_routes %d`, expectedValid)
-	if !strings.Contains(string(body), validPattern) {
-		t.Errorf("Expected to find %q in metrics output", validPattern)
+	// Check valid route count (via existing routes.total metric)
+	expectedRoutesTotalLine := fmt.Sprintf("skipper_custom_gauges{key=\"routes.total\"} %d", expectedValid)
```
szuecs, line 463: similar to https://github.com/zalando/skipper/pull/3550/files#r2207070355

szuecs:
:+1:

vlktna:
👍

PR #3551, opened 2025-07-07
fix: cache apiUsageMonitoring filter creation based on args

szuecs:
fix: creation of apiUsageMonitoring filter is costly so cache filter creation based on args

```
% benchstat old.txt new.txt
goos: linux
goarch: amd64
pkg: github.com/zalando/skipper/filters/apiusagemonitoring cpu: AMD Ryzen 7 PRO 4750U with Radeon Graphics
                                           │     old.txt     │               new.txt               │
                                           │     sec/op      │   sec/op     vs base                │
_CreateFilter_FullConfigSingleApiNakadi-16   452971.0n ± 36%   670.7n ± 1%  -99.85% (p=0.000 n=10)

                                           │    old.txt    │               new.txt                │
                                           │     B/op      │     B/op      vs base                │
_CreateFilter_FullConfigSingleApiNakadi-16   49.322Ki ± 0%   1.125Ki ± 0%  -97.72% (p=0.000 n=10)

                                           │    old.txt    │              new.txt               │
                                           │   allocs/op   │ allocs/op   vs base                │
_CreateFilter_FullConfigSingleApiNakadi-16   1379.000 ± 0%   1.000 ± 0%  -99.93% (p=0.000 n=10)
```

### DIFF
--- a/filters/apiusagemonitoring/spec.go
+++ b/filters/apiusagemonitoring/spec.go
@@ -103,6 +103,7 @@ func NewApiUsageMonitoring(
 		unknownPath:           unknownPath,
 		realmsTrackingMatcher: realmsTrackingMatcher,
 		sometimes:             rate.Sometimes{First: 3, Interval: 1 * time.Minute},
+		filterMap:             make(map[string]*apiUsageMonitoringFilter),
 	}
 	log.Debugf("created filter spec: %+v", spec)
 	return spec
@@ -124,6 +125,9 @@ type apiUsageMonitoringSpec struct {
 	realmsTrackingMatcher *regexp.Regexp
 	unknownPath           *pathInfo
 	sometimes             rate.Sometimes
+
+	mu        sync.Mutex
+	filterMap map[string]*apiUsageMonitoringFilter
 }
 
 func (s *apiUsageMonitoringSpec) errorf(format string, args ...interface{}) {
@@ -154,7 +158,31 @@ func (s *apiUsageMonitoringSpec) Name() string {
 	return filters.ApiUsageMonitoringName
 }
 
-func (s *apiUsageMonitoringSpec) CreateFilter(args []interface{}) (filter filters.Filter, err error) {
+func keyFromArgs(args []interface{}) string {
+	var sb strings.Builder
+	for _, a := range args {
+		s, ok := a.(string)
+		if !ok {
+			sb.Reset()
+			break
+		}
+		sb.WriteString(s)
+	}
+	return sb.String()
+}
+
+func (s *apiUsageMonitoringSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	key := keyFromArgs(args)
+	if key != "" {
+		s.mu.Lock()
+		f, ok := s.filterMap[key]
+		if ok {
+			s.mu.Unlock()
+			return f, nil
+		}
+		s.mu.Unlock()
+	}
+
 	apis := s.parseJsonConfiguration(args)
 	paths := s.buildPathInfoListFromConfiguration(apis)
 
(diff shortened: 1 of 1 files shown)


filters/apiusagemonitoring/spec.go
```diff
@@ -154,7 +158,31 @@ func (s *apiUsageMonitoringSpec) Name() string {
...
+	for _, a := range args {
+		s, ok := a.(string)
+		if !ok {
+			sb.Reset()
+			break
+		}
+		sb.WriteString(s)
+	}
+	return sb.String()
+}
```
AlexanderYastrebov: Let's be strict and return error if any arg is not a string. 
Also its better to join values with a separator to have distinct keys for `[A, BC]` and `[AB, C]` (likely does not matter in this case).

filters/apiusagemonitoring/spec.go
```diff
@@ -154,7 +158,31 @@ func (s *apiUsageMonitoringSpec) Name() string {
...
+
+func (s *apiUsageMonitoringSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	key := keyFromArgs(args)
+	if key != "" {
+		s.mu.Lock()
+		f, ok := s.filterMap[key]
+		if ok {
+			s.mu.Unlock()
+			return f, nil
```
AlexanderYastrebov, line 183: I don't like that filter instance references pathInfo which is mutable but I guess its ok in this case.

filters/apiusagemonitoring/spec.go
```diff
@@ -124,6 +125,9 @@ type apiUsageMonitoringSpec struct {
 	realmsTrackingMatcher *regexp.Regexp
 	unknownPath           *pathInfo
 	sometimes             rate.Sometimes
+
+	mu        sync.Mutex
+	filterMap map[string]*apiUsageMonitoringFilter
```
AlexanderYastrebov, line 131: This will have the usual problem of potentially growing unbounded because it is never cleaned up.

filters/apiusagemonitoring/spec.go
tkrop: I support this approach, however, I'm not even sure that there is more than one argument to the filter.

filters/apiusagemonitoring/spec.go
```diff
@@ -154,7 +158,31 @@ func (s *apiUsageMonitoringSpec) Name() string {
 	return filters.ApiUsageMonitoringName
 }
 
-func (s *apiUsageMonitoringSpec) CreateFilter(args []interface{}) (filter filters.Filter, err error) {
+func keyFromArgs(args []interface{}) string {
```
tkrop: Nit: from OO-design perspective I would define this private function of the `apiUsageMonitoringSpec`, however, I know that this can be seen different.

filters/apiusagemonitoring/spec.go
tkrop: Agreed, but unless we start to automatically change APIs, this should have limited impact. It also would be possible to have something like a timeout mechanism that removes entries when they have not been accessed for a certain time. The duplication of filters that happen in this case should be acceptable compared to what you experience now.

I also think, that filters could be even cached on a global level across specs - if considering the spec parameters going into the filter in addition. This would be helpful/necessary, if more than one spec with the same arguments can be created in skipper.

filters/apiusagemonitoring/spec.go
tkrop: The whole filter and filter handling is designed to be thread-safe. So I don't see a problem here.

filters/apiusagemonitoring/spec.go
szuecs: I think we can add the cleanup as preProcessor if we see that we have an unused key we can clean it up.
@AlexanderYastrebov do you have a strong opinion to have it, right now or do you think it's fine to do it later?

filters/apiusagemonitoring/spec.go
szuecs: As long as we have no reference in use for apiUsageMonitoringSpec the Go design prefers not to carry a ptr with us.

filters/apiusagemonitoring/spec.go
szuecs: I think the discussion is nice but not part of this change.

filters/apiusagemonitoring/spec.go
szuecs: This means we will change the behavior of the current filter.
The input is json anyways so there won't really be the case that we have associative change as suggested `[A, BC]` and `[AB, C]`, because of json syntax.

filters/apiusagemonitoring/spec.go
AlexanderYastrebov: Should be fine to keep it like that

filters/apiusagemonitoring/spec.go
AlexanderYastrebov: Then maybe just do fmt.Sprintf("%v", a) to convert to string regardless of the type

AlexanderYastrebov:
:+1:

szuecs:
:+1:

AlexanderYastrebov:
:+1:

Issue #3548, opened 2025-07-07
Flag -remove-hop-headers breaks Websocket requests


wndhydrnt:
**Describe the bug**
Websocket requests routed by Skipper fail when an operator sets the `-remove-hop-headers` flag.

**To Reproduce**
Prerequisites: 

[websocat](https://github.com/vi/websocat)

`routes.eskip`:
```
r_0: Host(/^localhost:9090$/) && Path("/example") -> "https://echo.websocket.org/";
```

1. Start Skipper: `skipper -routes-file /routes.eskip -experimental-upgrade`
2. Connect via Websocket: `websocat ws://localhost:9090/example`
3. Type arbitrary strings, these should be echoed back.
4. Stop Skipper, start it again: `skipper -routes-file /routes.eskip -remove-hop-headers -experimental-upgrade `
5. Connect via Websocket: `websocat ws://localhost:9090/example`

**Expected behavior**
The connection succeeds, as it did in step 2.

**Observed behavior**
The connection fails:
```text
websocat: WebSocketError: WebSocketError: Received unexpected status code (200 OK)
websocat: error running
```

Tested with Skipper `v0.22.57`.

wndhydrnt:
It looks to me like this is happening because [mapRequest()](https://github.com/zalando/skipper/blob/6a140c154bce660c30c8522ea478ec1441a59585/proxy/proxy.go#L599) drops the `Upgrade` header.

The check for `isUpgradeRequest` [in makeBackendRequest()](https://github.com/zalando/skipper/blob/6a140c154bce660c30c8522ea478ec1441a59585/proxy/proxy.go#L983) then doesn't succeed.

I believe that Skipper should respect the `Upgrade` header of the original request, which is why this issue has been created as a bug.
Happy to discuss the expected behavior.

szuecs:
@wndhydrnt yes of course we should respect that!
Thanks for the bug report!

szuecs:
@wndhydrnt please check this will fix it but what would be your proposal?

wndhydrnt:
@szuecs Thank you for providing a fix so quickly. I've rolled the branch out to a test cluster. Will keep it running for a while, but first tests look good.

I'll report back.

wndhydrnt:
@szuecs The change didn't cause any issues in our cluster. Websocket connections succeeded. Looks good to me.

szuecs:
[...]

PR #3553, opened 2025-07-08
fix: upgrade headers if hop headers are removed

szuecs:
fix: upgrade headers if hop headers are removed

fix: https://github.com/zalando/skipper/issues/3548

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -966,6 +966,8 @@ func (p *Proxy) makeUpgradeRequest(ctx *context, req *http.Request) {
 }
 
 func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Context) (*http.Response, *proxyError) {
+	payloadProtocol := getUpgradeRequest(ctx.Request())
+
 	req, endpointMetrics, err := p.mapRequest(ctx, requestContext)
 	if err != nil {
 		return nil, &proxyError{err: fmt.Errorf("could not map backend request: %w", err)}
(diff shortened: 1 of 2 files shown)


proxy/proxy.go
```diff
@@ -966,6 +966,12 @@ func (p *Proxy) makeUpgradeRequest(ctx *context, req *http.Request) {
 }
 
 func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Context) (*http.Response, *proxyError) {
+	var payloadProtocol string
+	isUpgrade := isUpgradeRequest(ctx.Request())
+	if isUpgrade {
+		payloadProtocol = ctx.Request().Header.Get("Upgrade")
```
AlexanderYastrebov: I think there is getUpgradeRequest helper for this.

proxy/proxy.go
```diff
@@ -980,7 +986,9 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 		defer endpointMetrics.DecInflightRequest()
 	}
 
-	if p.experimentalUpgrade && isUpgradeRequest(req) {
+	if p.experimentalUpgrade && isUpgrade {
+		req.Header.Set("Upgrade", payloadProtocol)
+		req.Header.Set("Connection", "Upgrade")
```
AlexanderYastrebov, line 987: Maybe capture "Connection" header value to pass it intact here?

AlexanderYastrebov: reverseproxy.go [does the same](https://github.com/golang/go/blob/9159cd4ec6b0e9475dc9c71c830035c1c4c13483/src/net/http/httputil/reverseproxy.go#L423-L428)

proxy/proxy.go
szuecs: ah yeah makes sense to use. At first I thought we don't need the value and then I forgot the func.

proxy/proxy.go
szuecs: Oh yeah good to know so we just do the same as stdlib, which seems to be fine.

proxy/proxy.go

MustafaSaber:
👍

szuecs:
:+1:

AndreasSko:
Just FYI: We upgraded Skipper from v0.22.65 to v0.22.67, and since then, Websocket connections are established as usual, but don't seem to respond anymore. We had to downgrade again to fix the issue. Still trying to figure out what the exact problem is.

szuecs:
@AndreasSko can you create an issue and tell a bit more about configuration and maybe logs?

PR #3559, opened 2025-07-14
Support OpenTelemetry SDK

AlexanderYastrebov:
When OpenTelemetry is enabled Skipper registers global OpenTelemetry TracerProvider and creates a bridge OpenTracing tracer.

This follows recommended migration strategy https://opentelemetry.io/docs/migration/opentracing/#migration-overview

OpenTelemetry Init reads configuration options from standard environment variables and allows additional options as a command line flag in yaml format to simplify parsing.

Updates #2104


tracing/tracing.go
```diff
@@ -173,6 +178,10 @@ func GetTraceID(span ot.Span) string {
 		return spanContextType.TraceID().String()
 	case origlightstep.SpanContext:
 		return fmt.Sprintf("%x", spanContextType.TraceID)
+	case otelSpanContext:
+		// Use last 64 bits of the trace id (i.e. 16 hexadecimal characters)
+		traceId := spanContextType.TraceID().String()
+		return traceId[len(traceId)-16:]
```
szuecs: This can panic https://go.dev/play/p/r2HJ6le1rt8

config/config.go
```diff
@@ -117,6 +118,8 @@ type Config struct {
 	AccessLogStripQuery                 bool      `yaml:"access-log-strip-query"`
 	SuppressRouteUpdateLogs             bool      `yaml:"suppress-route-update-logs"`
 
+	OpenTelemetry *otel.Options `yaml:"open-telemetry"`
```
szuecs, line 121: Maybe better to tell it's OTelOptions or so

otel/otel.go
```diff
@@ -0,0 +1,138 @@
+package otel
```
szuecs, line 4: package docs would be great to read to understand the idea

otel/otel.go
```diff
@@ -0,0 +1,138 @@
...
+
+// Options configure OpenTelemetry pipeline.
+type Options struct {
+	// Initialized indicates whether the OpenTelemetry pipeline has been initialized externally.
+	// If true Skipper will use via [otel.Tracer] to get OpenTelemetry tracer from pre-configured global tracer provider.
+	Initialized bool
+}
+
+// Init bootstraps the OpenTelemetry pipeline using environment variables and provided options.
+// Make sure to call shutdown for proper cleanup when err is nil.
```
szuecs: when -> if

szuecs:
Would this configure some added span Tags or something like this?
I am a bit skeptical about everything "auto"-magic things, but likely it's fine as far as I understand.

tracing/tracing.go
AlexanderYastrebov: It can't - the trace id type is 16-byte array. But I changed it to simply return spanContextType.TraceID().String().

config/config.go
AlexanderYastrebov: I'd like it to match the flag name which looks ok. Let me think about it and I'll change it later if needed.

tracing/tracing.go
szuecs: Sgtm

config/config.go
szuecs: It's just not obvious so maybe change the flag.
It's also kind of new so I would prefer to change it if we have it already.

skipper.go
```diff
@@ -732,6 +737,12 @@ type Options struct {
 	// tracer instance overrides options provided under OpenTracing property.
 	OpenTracingTracer ot.Tracer
 
+	// OpenTelemetry configures the OpenTelemetry pipeline.
+	// When set Skipper ignores OpenTracing and OpenTracingTracer options and configures OpenTracing bridge.
+	// If not specified, Skipper will use [otel.Tracer] function to get the OpenTelemetry tracer
+	// which could be a noop tracer or a wrapper tracer if library user configured OpenTracing bridge tracer.
+	OpenTelemetry *sotel.Options
```
a4180p, line 744: Shouldn't we have an ability to pass initialized tracer in options?
Like the OpenTracingTracer above?

otel/otel.go
```diff
@@ -0,0 +1,145 @@
...
+		return err
+	}
+
+	// handleErr calls shutdown for cleanup and makes sure that all errors are returned.
+	handleErr := func(inErr error) {
+		err = errors.Join(inErr, shutdown(ctx))
+	}
+
+	// Similar to "console" but writes into debug logger
+	autoexport.RegisterSpanExporter("skipper-debug", func(ctx context.Context) (trace.SpanExporter, error) {
```
a4180p, line 94: Maybe we can register this exporter only if the log level is Debug or higher?

skipper.go
a4180p: Ok, so `Options.Initalized` means exactly that, we will use preinitialised global tracer

skipper.go
```diff
@@ -1640,11 +1651,39 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
+		defer shutdown(context.Background())
+
+		// Setup OpenTracing bridge tracer
+		bridgeTracer, wrapperTracerProvider := otBridge.NewTracerPair(otel.Tracer(otelTracerName))
+
+		bridgeTracer.SetWarningHandler(func(msg string) { log.Warnf("OpenTracing bridge warning: %s", msg) })
+		otel.SetTracerProvider(wrapperTracerProvider)
+
+		tracer = bridgeTracer
+		otelTracer = otel.Tracer(otelTracerName)
```
a4180p, line 1673: why do we need this variable?

proxy/proxy.go
```diff
@@ -1631,7 +1632,15 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	p.tracing.setTag(span, HTTPRemoteIPTag, stripPort(r.RemoteAddr))
 	p.setCommonSpanInfo(r.URL, r, span)
-	r = r.WithContext(ot.ContextWithSpan(r.Context(), span))
+
+	ctxWithSpan := ot.ContextWithSpan(r.Context(), span)
+
+	// https://pkg.go.dev/go.opentelemetry.io/otel/bridge/opentracing#readme-interop-from-trace-context-from-opentracing-to-opentelemetry
+	if bridgeTracer, ok := p.tracing.tracer.(*otBridge.BridgeTracer); ok {
+		ctxWithSpan = bridgeTracer.ContextWithSpanHook(ctxWithSpan, span)
+	}
```
a4180p, line 1641: hm.. that's interesting.
I don't understand how it works now, when we provide bridgeTracer through skipper options.

otel/otel.go
AlexanderYastrebov: Registering does not hurt - its not used unless OTEL_TRACES_EXPORTER=skipper-debug

skipper.go
AlexanderYastrebov: We don't use it now, I just define it to mark that we can pass it to the components. OTOH we can probably export otelTracerName and  use otel.Tracer(OtelTracerName) to get it elsewhere...

a4180p: we can create it earlier and use in https://github.com/zalando/skipper/pull/3559/files#diff-9b0b16d06f27b640eb77f6ba0dee84df6a660528e5cccd81839e0330c4b92dd2R1667

AlexanderYastrebov: They can be different in case Initialized: true . The first one would be externally initialized and then this will get wrapper tracer coming from otel.SetTracerProvider(wrapperTracerProvider)

AlexanderYastrebov:
Closing in favour of #3627 which supports environment variable overrides via config.

Issue #3526, opened 2025-06-06
OPA Filters: Move Startup / Bundle Download out of Filter Creation


mjungsbluth:
**Is your feature request related to a problem? Please describe.**
The Open Policy Agent based filters today download bundles and fire up a virtual OPA instance during filter creation. This challenges some implicit assumptions in the code base that filter creation is a very fast and quick operation. Blocking filter creation for too long can block route processing which can cause unavailability of other routes and not update the endpoints that are targeted. 

While there have been improvements to fail fast on bundle download errors, network delays and big bundles remain a lingering issue. In steady state we currently see no problem but this should be addressed. 

**Describe the solution you would like**

We have discussed doing the following changes to make filter creation fast and not depend on external factors like network performance or bundle size:

Implement a route preprocessor that
* On first use (Skipper startup): Pre-Loads all OPA instances via the OPA Registry across all routes and waits until they have successfully finished or failed permanently
* On sub-sequent use: pre-load new OPA instances when they are detected but do not block the preprocessing further

Filter creation will be changed such that:
* It will lookup up if a ready instance can be obtained from the OPA Registry
* If none is found (not ready yet, errored out), it will just error out
* This will render the route erroneous and basically means that traffic cannot be served. 

A note on CPU consumption
* On the initial start we can parallelize starting the OPA instances and bundle downloads because we want to start fast and the CPU consumption cannot have an effect on other I/O bound requests
* For the sub-sequent processing, we need to be careful to load new OPA instances in sequence to not overconsume CPU. This should ideally be combined if the central control loop for OPA bundle updates is enabled. 

**Additional context (optional)**
An initial skeleton for the implementation can be found here: https://github.com/zalando/skipper/compare/filters/openpolicyagent/initial-bundle-download

**Would you like to work on it?**
Yes

Pushpalanka:
PR #3562 is merged addressing this.

PR #3561, opened 2025-07-16
Move OPA instance configuration to registry level

mjungsbluth:
This PR is a no-op refactoring from a user perspective that moves the Open Policy Agent configuration to the registry. 

Originally it was designed to have a per filter configuration but practically it is always the same.

This is the precondition to allow instance pre-loading #3526 and will also to re-use config across filters.

### DIFF
--- a/filters/openpolicyagent/openpolicyagent.go
+++ b/filters/openpolicyagent/openpolicyagent.go
@@ -187,7 +187,7 @@ func WithControlLoopMaxJitter(maxJitter time.Duration) func(*OpenPolicyAgentRegi
 	}
 }
 
-func NewOpenPolicyAgentRegistry(opts ...func(*OpenPolicyAgentRegistry) error) *OpenPolicyAgentRegistry {
+func NewOpenPolicyAgentRegistry(opts ...func(*OpenPolicyAgentRegistry) error) (*OpenPolicyAgentRegistry, error) {
 	registry := &OpenPolicyAgentRegistry{
 		reuseDuration:          defaultReuseDuration,
 		cleanInterval:          DefaultCleanIdlePeriod,
@@ -205,6 +205,16 @@ func NewOpenPolicyAgentRegistry(opts ...func(*OpenPolicyAgentRegistry) error) *O
 		opt(registry)
 	}
 
+	if registry.configTemplate == nil {
+		config, err := NewOpenPolicyAgentConfig()
+
+		if err != nil {
+			return nil, err
+		}
+
+		registry.configTemplate = config
+	}
+
 	if registry.maxMemoryBodyParsingSem == nil {
 		registry.maxMemoryBodyParsingSem = semaphore.NewWeighted(DefaultMaxMemoryBodyParsing)
 	}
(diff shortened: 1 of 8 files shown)


mjungsbluth:
@amak3 fyi

amak3:
looks good to me 👍

mjungsbluth:
@torwunder could you have a look?

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -79,6 +79,7 @@ type OpenPolicyAgentRegistry struct {
 	reuseDuration          time.Duration
 	cleanInterval          time.Duration
 	instanceStartupTimeout time.Duration
+	config                 *OpenPolicyAgentInstanceConfig
```
torwunder: nitpick: do we want to make it transparent that this does not hold the complete configuration for an instance but the general config that is the same for all instances?

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -404,7 +418,7 @@ func (registry *OpenPolicyAgentRegistry) NewOpenPolicyAgentInstance(bundleName s
 		return instance, nil
 	}
 
-	instance, err := registry.newOpenPolicyAgentInstance(bundleName, config, filterName)
+	instance, err := registry.newOpenPolicyAgentInstance(bundleName, *registry.config, filterName)
```
torwunder: the interpolation function could be put on the instance config now for more encapsulation, it would also allow to remove one parameter from the `new` method as this takes both the interpolated config and the config itself

torwunder (commented):
looks good, some minor comments about the config itself

filters/openpolicyagent/openpolicyagent.go
mjungsbluth: You mean: call it `configTemplate`, yes. that is a good idea I guess...

mjungsbluth:
👍

torwunder:
👍

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -557,7 +571,7 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, configBytes []
 
 	opa := &OpenPolicyAgentInstance{
 		registry:       registry,
-		instanceConfig: instanceConfig,
+		instanceConfig: *registry.configTemplate,
```
AlexanderYastrebov, line 584: This assumes registry.configTemplate is not nil which implies/requires that WithOpenPolicyAgentInstanceConfig has been called.
Can we check that it is not nil and return an error before instead of potentially failing with nil dereference?

szuecs: I agree it's simpler to find the error, if you don't panic

mjungsbluth: Good catch. Let me address this...

AlexanderYastrebov:
:+1:

mjungsbluth:
👍

PR #3562, opened 2025-07-18
Open Policy Agent: Instance Preloading

mjungsbluth:
## Overview

This feature implements issue #3526 by moving OPA instance startup and bundle downloading out of filter creation to make route processing faster and more reliable.

WIP: Requires #3561 to be merged, which will also make the review easier. 

## Problem

Previously, OPA filters created instances synchronously during filter creation, which:
- Blocked route processing during bundle downloads
- Could cause timeouts with large bundles or slow networks
- Violated the assumption that filter creation should be fast

## Solution

When enabled via the `--enable-open-policy-agent-preloading` flag, OPA instances are pre-loaded during route processing instead of filter creation:

1. **Route PreProcessor**: Scans routes for OPA filter usage and pre-loads instances
2. **Non-blocking Filter Creation**: Filters only lookup ready instances, failing fast if not available
3. **Loading Strategy**: Parallel loading on startup, sequential on route updates

### Enable the Feature

```bash
skipper --enable-open-policy-agent --enable-open-policy-agent-preloading
```

### Configuration

- `--enable-open-policy-agent-preloading`: Enable the pre-loading behavior (default: false)
- All existing OPA configuration flags remain the same


docs/operation/opa-preloading.md
```diff
@@ -0,0 +1,116 @@
...
+- All existing OPA configuration flags remain the same
+
+### Behavior Changes
+
+#### With Pre-loading Disabled (Default)
+- **Filter Creation**: Synchronous, blocks until OPA instance is ready
+- **Route Processing**: Fast, instances already created
+- **Backward Compatibility**: 100% compatible with existing behavior
+
+#### With Pre-loading Enabled
```
torwunder: it might help to explain how new routes would be added, by being picked up be the route processing (that runs regularly) which triggers the instance creation and the filter might only become active after the second route processing, as instance creation is async (if I understand the process correctly)

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -428,13 +497,77 @@ func (registry *OpenPolicyAgentRegistry) NewOpenPolicyAgentInstance(bundleName s
...
-		return nil, err
-	}
-	registry.instances[bundleName] = instance
+	return nil, nil
+}
 
-	return instance, nil
+// PrepareInstanceLoader returns a function that when called will create an OPA instance
+// This allows the preprocessor to control when and how the instance creation happens
+// Prevents concurrent creation of the same bundle by tracking in-flight operations
```
AlexanderYastrebov: Please check if we can use https://pkg.go.dev/golang.org/x/sync/singleflight for this

proxy/proxytest/proxytest.go
```diff
@@ -123,6 +123,24 @@ func (p *TestProxy) Client() *TestClient {
...
+
+	p.dc.Update(routes, deleted)
+	_, _, err := p.dc.LoadUpdate()
+
+	return err
+}
+
+func (p *TestProxy) GetRoutes() map[string]*eskip.Route {
+	if p.dc == nil {
+		return nil
```
AlexanderYastrebov: Either this should panic like the UpdateRoutes or the UpdateRoutes should return error.

filters/openpolicyagent/openpolicyagent.go
Pushpalanka: Addressed and added a test to confirm only instance is created for multiple calls in registry.

proxy/proxytest/proxytest.go
Pushpalanka: Made it panic. Thanks.

docs/operation/opa-preloading.md
Pushpalanka: Sharing my understanding as of now.

For a fresh OPA route addition, 
once the data clients detect the change, 
- If the referred bundle was already present in another route -> Pre processor see OPA instance exists ->  filter created
- If the referred bundle is fresh -> Pre processor initiate a background task to create OPA instance  -> successful filter creation if background task complete on time -> (else filter creation fails -> route marked invalid)

This invalid route is then not re-attempted unless there is an effective change detected by the data clients. Regular watch on routes doesn't provide automatic recovery for failed routes as I saw, unless there is an effective change in the route definitions.

Pushpalanka: I added a log line at https://github.com/zalando/skipper/blob/master/routing/datasource.go#L612 to confirm if pre processors are getting called for regular route update check and it was only printed if there was an effective change to route definitions in .eskip file. In that sense a failed OPA route will not become active again, unless there is a route modification. I understand it could have been an overhead if it indefinitely keep on re-trying with pre processor getting called periodically as well.

@AlexanderYastrebov @szuecs could you shed some light on expected execution of pre-processors for route updates and whether invalid routes get re-attempted.

szuecs: That's correct, but I think autoscaling will make sure it will get triggered more often in production clusters that you may think.

filters/openpolicyagent/opaauthorizerequest/openpolicyagent_bootstrap_routeupdate_test.go
```diff
@@ -0,0 +1,309 @@
...
+
+	updatedRoutes := eskip.MustParse(fmt.Sprintf("r1: %s", opaRouteTemplate("/initial", bundleName)))
+
+	ts.preprocessor.Do(updatedRoutes)
+	require.NoError(t, ts.waitForInstanceLoad())
+
+	ts.dataClient.Update(updatedRoutes, nil)
+	require.NoError(t, ts.routing.waitForRouteSettings(2))
+
+	route := ts.routing.requireRoute(t, "https://www.z-opa.org/initial")
```
szuecs: maybe better to have an example `https://opa.example/initial` or  `https://opa.test/initial`  URL. Someone could register the name and `.org` is not good for test cases.

filters/openpolicyagent/preprocessor.go
```diff
@@ -0,0 +1,126 @@
...
+	filterName string
+}
+
+func (p *opaPreProcessor) isOpaFilter(filterName string) bool {
+	return strings.HasPrefix(filterName, "opa")
+}
+
+func (p *opaPreProcessor) preloadInstancesParallel(requests map[string]bundleRequest) {
+	var wg sync.WaitGroup
+	log := &logging.DefaultLog{}
```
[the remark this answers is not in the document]
szuecs: yes we should pass it on create of the struct instance as member to the struct.

docs/operation/opa-preloading.md
Pushpalanka: We had an offline discussion and concluded to have a retry go routine, attempting failed instances either by deducing it using required instances vs available instances or tracking the instance status in OPA registry.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -453,6 +453,9 @@ func (registry *OpenPolicyAgentRegistry) startCustomControlLoopDaemon() {
 			registry.mu.Unlock()
 
 			for _, opa := range instances {
+				if opa != nil && opa.Healthy() { // Skip retriggering plugins of healthy instance
```
mjungsbluth: This would prevent bundle updates for healthy instances, i.e. they would stop fetching. I don’t think we want this…

Pushpalanka: Thanks for catching this. It indeed depend on this loop since trigger is set to manual mode.

There was a data race without this between control loop and parallel start. 
[parallelVSloopRACE.txt](https://github.com/user-attachments/files/22454030/parallelVSloopRACE.txt) will see if `!opa.Started` can be a solution here. 

Still another two race condition remained after this coming for OPA code, not sure if we can do anything from the flows in ourside to avoid this from getting triggered. 

1. [DownloadRACE.txt](https://github.com/user-attachments/files/22454047/DownloadRACE.txt) 
The etag value of the Downloader is 
written as
 
```
func (d *Downloader) SetCache(etag string) {
	d.etag = etag
}
``` 

and read as
```	
d.client = d.client.WithHeader("If-None-Match", d.etag)

```
without synchronization in OPA download.go file.

2. [ManagerInitializedRace.txt](https://github.com/user-attachments/files/22460310/ManagerInitializedRace.txt)
plugins.go file, m.initialized field read/write race between Start and Init methods.

Possible fixes in OPA : [opa_race_conditions.patch](https://github.com/user-attachments/files/22460363/opa_race_conditions.patch)

Pushpalanka: We could handle it by adding the `Start` task to the background queue itself than immediately attempting the start which was overlapping with task execution from the already added tasks in the queue.

szuecs: You could also handle "nil" in Healthy():
```
func (o *opa) Healthy() bool {
    if o == nil {
        return false
    }
...
}
```

filters/openpolicyagent/internal/opatestutils/openpolicyagent_helpers.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (c *ControllableBundleServer) Stop() {
+	c.proxyServer.Close()
+	c.realServer.Stop()
+}
+
+type BundleServerConfig struct {
+	BundleName string
+	RespCode   int
```
mjungsbluth: Is this really needed? It seems quite duplicate to pass this and the duration here and then also on the ControllableBundleServer..

filters/openpolicyagent/opaauthorizerequest/openpolicyagent_routeupdate_test.go
```diff
@@ -0,0 +1,667 @@
...
+		targetPhase    *testPhase
+	}{
+		{
+			msg:        "Bootstrap with OPA route",
+			bundleName: "somebundle",
+			bootstrapPhase: testPhase{
+				routes:         `r1: Path("/secure") -> opaAuthorizeRequest("somebundle", "") -> status(204) -> <shunt>`,
+				testPath:       "/secure",
+				expectedStatus: http.StatusNoContent,
+				healthy:        true,
```
mjungsbluth: This is an expected value?

```suggestion
				expectedHealthy:        true,
```

filters/openpolicyagent/opaserveresponse/opaserveresponse.go
```diff
@@ -66,7 +66,8 @@ func (s *spec) CreateFilter(args []interface{}) (filters.Filter, error) {
 		}
 	}
 
-	opa, err := s.registry.NewOpenPolicyAgentInstance(bundleName, s.Name())
+	// Try to get instance with new non-blocking approach
+	opa, err := s.registry.GetOrStartInstance(bundleName)
 	if err != nil {
 		return nil, err
```
mjungsbluth:
```suggestion
		return nil, fmt.Errorf("open policy agent instance for bundle name '%s' could not be obtained: %w", bundleName, err)
```

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -436,8 +492,39 @@ func (registry *OpenPolicyAgentRegistry) Do(routes []*routing.Route) []*routing.
...
+	}
+
+	if instance != nil {
+		// Instance already exists, return it
+		return instance, nil
+	}
+
+	if registry.preloadingEnabled {
+		// This is a hard error in the sense that even an unhealthy instance could not be created.
+		// In preloading mode, if instance doesn't exist, it means it's not ready yet
```
mjungsbluth: This comment needs clean-up

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -582,7 +670,7 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, filterName str
 	runtime.RegisterPlugin(envoy.PluginName, envoy.Factory{})
 
 	var logger logging.Logger = &QuietLogger{target: logging.Get()}
-	logger = logger.WithFields(map[string]interface{}{"skipper-filter": filterName, "bundle-name": bundleName})
+	logger = logger.WithFields(map[string]interface{}{"opa-bundle-name": bundleName})
```
mjungsbluth: we shouldn't change the field name for the bundle name attribute

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -617,16 +705,52 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, filterName str
 		interQueryBuiltinValueCache: registry.valueCache,
 
 		idGenerator: uniqueIDGenerator,
+		logger:      logging.Get().WithFields(map[string]any{"opa-bundle-name": bundleName}),
```
mjungsbluth:
```suggestion
		logger:      logging.Get().WithFields(map[string]any{"bundle-name": bundleName}),
```

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -1039,18 +1187,59 @@ func (l *QuietLogger) GetLevel() logging.Level {
 	return l.target.GetLevel()
 }
 
-func (l *QuietLogger) Debug(fmt string, a ...interface{}) {
-	l.target.Debug(fmt, a)
+func logWithArgs(logFunc func(string, ...interface{}), msg string, args ...interface{}) {
```
mjungsbluth: @Pushpalanka what does this do or why was it needed?

Pushpalanka: @mjungsbluth Without this it was printing  character similar to sequence(!EXTRA interface{}) when args are not present. Wanted to make it neat. So it is checking if args are present or not and printing at the selected log level. Function was just to generalize the logic rather than having the args length check in each log level.

mjungsbluth: Hmm, ok that makes sense but I guess we could just replace this by appending `...` to the original call so `l.target.Debug(fmt, a...)`?

filters/openpolicyagent/internal/opatestutils/openpolicyagent_helpers.go
Pushpalanka: Was thinking this will simplify code in multiple bundle servers case to pass the input and then keeping ControllableBundleServer for runtime state. It does seem duplicate when look back, will simplify and keep only ControllableBundleserver.

filters/openpolicyagent/opaauthorizerequest/openpolicyagent_routeupdate_test.go
Pushpalanka: Yes, this will be more expressive. Will commit with other required changes.

filters/openpolicyagent/openpolicyagent.go
Pushpalanka: Will revert back to original, is this because it's used for alerting or monitoring?

filters/openpolicyagent/openpolicyagent.go
Pushpalanka: Yes, this works. That was overcomplicating this. :)

filters/openpolicyagent/preprocessor.go
```diff
@@ -0,0 +1,150 @@
...
+// Do implements routing.PreProcessor
+func (p *opaPreProcessor) Do(routes []*eskip.Route) []*eskip.Route {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	// Extract OPA bundle requirements from routes
+	bundleConfigs := p.extractOpaBundleRequests(routes)
+
+	// Use sync.Once to ensure initial load happens exactly once
+	p.initialLoad.Do(func() {
```
szuecs: maybe rename `initialLoad` to `once` and we do understand without comment.
I think we use `once` (sync.Once) similar to `mu` (sync.Mutex)

szuecs:
:+1:

Pushpalanka:
👍

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -436,8 +492,38 @@ func (registry *OpenPolicyAgentRegistry) Do(routes []*routing.Route) []*routing.
...
+	}
+
+	if instance != nil {
+		// Instance already exists, return it
+		return instance, nil
+	}
+
+	if registry.preloadingEnabled {
+		// In preloading mode, if instance doesn't exist, it means a fatal error occurred while creating the opa instance
+		return nil, fmt.Errorf("open policy agent instance for bundle '%s' could not be created", bundleName)
```
szuecs: If you want to have quotes in error strings or log strings, use %q instead of '%s'

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -973,56 +1116,60 @@ func (opa *OpenPolicyAgentInstance) ExtractHttpBodyOptionally(req *http.Request)
...
 }
 
+type evalContext struct {
+	opa *OpenPolicyAgentInstance
+}
+
 // ParsedQuery is an implementation of the envoyauth.EvalContext interface
-func (opa *OpenPolicyAgentInstance) ParsedQuery() ast.Body {
-	return opa.EnvoyPluginConfig().ParsedQuery
+func (evalctx *evalContext) ParsedQuery() ast.Body {
```
szuecs: Maybe:
```
func (e *evalContext) ParsedQuery() ast.Body {
```
or 
```
func (ec *evalContext) ParsedQuery() ast.Body {
```
In go var names should be short if it's close to the reader.

filters/openpolicyagent/preprocessor.go
```diff
@@ -0,0 +1,149 @@
...
+
+	"github.com/zalando/skipper/eskip"
+	"github.com/zalando/skipper/logging"
+	"github.com/zalando/skipper/routing"
+)
+
+type opaPreProcessor struct {
+	registry *OpenPolicyAgentRegistry
+	once     sync.Once
+	mu       sync.Mutex
```
szuecs, line 15: normally we put the Mutex in the struct before the things it protects and then one newline to show that the other things are not related to the Mutex.

szuecs:
:+1:

szuecs:
I added some minor comments, but feel free to merge and fix these later in a separate PR

filters/openpolicyagent/openpolicyagent.go
Pushpalanka: Thanks, addressed.

filters/openpolicyagent/openpolicyagent.go
Pushpalanka: Thanks, addressed.

filters/openpolicyagent/preprocessor.go
Pushpalanka: Thanks, reordered. From the list of variables here, I felt `once` can also stay related to the same group.

Pushpalanka:
👍

szuecs:
:+1:

PR #3563, opened 2025-07-18
fix: Websocket requests fail if case of value in Connection header and case of key in Upgrade header do not match

wndhydrnt:
When a client sends a request to upgrade to a Websocket connection
and presents the value of the header `Connection` in lower-case
then Skipper fails to upgrade the request.

This change fixes the issue by using a function to read the header that is case insensitive.

The bug was likely introduced in #3553.

### DIFF
--- a/proxy/upgrade.go
+++ b/proxy/upgrade.go
@@ -32,7 +32,7 @@ func isUpgradeRequest(req *http.Request) bool {
 func getUpgradeRequest(req *http.Request) string {
 	for _, h := range req.Header[http.CanonicalHeaderKey("Connection")] {
 		if strings.Contains(strings.ToLower(h), "upgrade") {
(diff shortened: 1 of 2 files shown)


wndhydrnt:
The issue has been observed when Skipper was running behind an AWS ALB.
The ALB sends `Connection: upgrade`, which does not match the key in `Upgrade: websocket`.

szuecs:
:+1:

szuecs:
Thanks @wndhydrnt

wndhydrnt:
@szuecs Thank you for the review.
This fix enables us to activate the `-remove-hop-headers` flag.
Would it be possible to find a second reviewer and merge the change any time soon?

AlexanderYastrebov:
:+1:

PR #3572, opened 2025-07-28
OPA JWT token cache configuration

amak3:
### Context
With https://github.com/open-policy-agent/opa/pull/7274 Open Policy Agent added configurable token cache to io.jwt token verification built-in functions to improve performance. See [changelog](https://github.com/open-policy-agent/opa/releases/tag/v1.1.0). 

### Changes
With this pull request JWT cache reading and writing is enabled and benchmark tests are added to measure JWT validation performance with cache configuration. 

**Fix in opa-envoy-plugin**
After initial round of benchmark tests that showed no impact on perfomence the bug was found in opa-envoy-plugin and [contribution with the fix was made](https://github.com/open-policy-agent/opa-envoy-plugin/pull/748). 

### Benchmark tests result after fix

- results for test with vs without cache configuration

for 10 runs, each 20 s:
attached: [jwt-validation-cache-benchmark-results-10runs-20s.txt](https://github.com/user-attachments/files/21157031/jwt-validation-cache-benchmark-results-10runs-20s.txt)
```goos: darwin
goarch: arm64
pkg: github.com/zalando/skipper/filters/openpolicyagent/opaauthorizerequest
cpu: Apple M1
                                  │       sec/op       │
JwtValidation/without_jwt_cache-8       20.10µ ± 4%
JwtValidation/with_jwt_cache-8          10.65µ ± 3%
geomean                                 14.63µ

                                  │        B/op        │
JwtValidation/without_jwt_cache-8      38.11Ki ± 0%
JwtValidation/with_jwt_cache-8         29.89Ki ± 0%
geomean                                33.75Ki

                                  │     allocs/op      │
JwtValidation/without_jwt_cache-8        621.0 ± 0%
JwtValidation/with_jwt_cache-8           512.0 ± 0%
geomean                                  563.9
```

- results for test with cache set for max number of 5 tokens in the cache and varying number of tokens in the test (1, 5, 6, 10, 100)

```goos: darwin
goarch: arm64
pkg: github.com/zalando/skipper/filters/openpolicyagent/opaauthorizerequest
cpu: Apple M1
                                                       │        B/op        │
JwtValidationWithVaryingTokenCounts/tokens_count_1-8           9.892µ ± 14%
JwtValidationWithVaryingTokenCounts/tokens_count_5-8           10.38µ ±  9%
JwtValidationWithVaryingTokenCounts/tokens_count_6-8           13.38µ ±  3%
JwtValidationWithVaryingTokenCounts/tokens_count_10-8          16.23µ ±  7%
JwtValidationWithVaryingTokenCounts/tokens_count_100-8         22.25µ ± 14%

                                                       
                                                       │        B/op        │
JwtValidationWithVaryingTokenCounts/tokens_count_1-8           29.89Ki ± 0%
JwtValidationWithVaryingTokenCounts/tokens_count_5-8           29.90Ki ± 0%
JwtValidationWithVaryingTokenCounts/tokens_count_6-8           31.78Ki ± 0%
JwtValidationWithVaryingTokenCounts/tokens_count_10-8          34.57Ki ± 0%
JwtValidationWithVaryingTokenCounts/tokens_count_100-8         38.27Ki ± 0%
geomean                                                        32.73Ki

                                
                                                       │     allocs/op      │
JwtValidationWithVaryingTokenCounts/tokens_count_1-8             512.0 ± 0%
JwtValidationWithVaryingTokenCounts/tokens_count_5-8             512.0 ± 0%
JwtValidationWithVaryingTokenCounts/tokens_count_6-8             538.0 ± 0%
JwtValidationWithVaryingTokenCounts/tokens_count_10-8            577.0 ± 0%
JwtValidationWithVaryingTokenCounts/tokens_count_100-8           631.5 ± 0%
```

### DIFF
--- a/filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
+++ b/filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
@@ -4,6 +4,7 @@ import (
 	_ "embed"
 	"fmt"
 	"io"
+	"math/rand"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
@@ -125,6 +126,7 @@ func BenchmarkMinimalPolicyWithDecisionLogs(b *testing.B) {
 		BundleNames:         []string{testBundleName},
 		DecisionLogging:     true,
 		ContextExtensions:   "",
+		JwtCache:            false,
 	}
 	f, err := createOpaFilter(filterOpts)
 	require.NoError(b, err)
@@ -199,7 +201,119 @@ func BenchmarkAllowWithReqBody(b *testing.B) {
 
 }
 
+// BenchmarkJwtValidation measures the performance of the OPA authorization filter with a JWT validation
+// with and without JWT cache configuration enabled.
 func BenchmarkJwtValidation(b *testing.B) {
+	type benchmarkCase struct {
+		name     string
+		jwtCache bool
+	}
+
+	cases := []benchmarkCase{
+		{
+			name:     "without_jwt_cache",
+			jwtCache: false,
+		},
+		{
+			name:     "with_jwt_cache",
+			jwtCache: true,
+		},
+	}
+
+	for _, bc := range cases {
+		opaControlPlane := opasdktest.MustNewServer(
+			opasdktest.MockBundle(testBundleEndpoint, map[string]string{
+				"main.rego": fmt.Sprintf(`
+					package envoy.authz
+
+					import rego.v1
+
+					default allow = false
+
+					public_key_cert := %q
+
+					bearer_token := t if {
+						v := input.attributes.request.http.headers.authorization
+						startswith(v, "Bearer ")
+						t := substring(v, count("Bearer "), -1)
+					}
+
+					allow if {
+						[valid, _, payload] :=  io.jwt.decode_verify(bearer_token, {
+							"cert": public_key_cert,
+							"aud": "nqz3xhorr5"
+						})
+					
+						valid
+						
+						payload.sub == "5974934733"
+					}				
+				`, publicKey),
+			}),
+		)
+		defer opaControlPlane.Stop()
+
+		filterOpts := FilterOptions{
+			OpaControlPlaneUrl:  opaControlPlane.URL(),
+			DecisionConsumerUrl: "",
+			DecisionPath:        testDecisionPath,
+			BundleNames:         []string{testBundleName},
+			DecisionLogging:     false,
+			ContextExtensions:   "",
+			JwtCache:            bc.jwtCache,
+		}
+
+		f, err := createOpaFilter(filterOpts)
+		require.NoError(b, err)
+
+		reqUrl, err := url.Parse("http://opa-authorized.test/somepath")
+		require.NoError(b, err)
+
+		claims := jwt.MapClaims{
+			"iss":   "https://some.identity.acme.com",
+			"sub":   "5974934733",
+			"aud":   "nqz3xhorr5",
+			"iat":   time.Now().Add(-time.Minute).UTC().Unix(),
+			"exp":   time.Now().Add(tokenExp).UTC().Unix(),
+			"email": "someone@example.org",
+		}
+
+		token := jwt.NewWithClaims(jwt.SigningMethodRS256, claims)
+
+		key, err := jwt.ParseRSAPrivateKeyFromPEM(privateKey)
+		require.NoError(b, err, "Failed to parse RSA PEM")
+
+		// Sign and get the complete encoded token as a string using the secret
+		signedToken, err := token.SignedString(key)
+		require.NoError(b, err, "Failed to sign token")
+
+		b.Run(bc.name, func(b *testing.B) {
+			ctx := &filtertest.Context{
+				FStateBag: map[string]interface{}{},
+				FResponse: &http.Response{},
+				FRequest: &http.Request{
+					Header: map[string][]string{
+						"Authorization": {fmt.Sprintf("Bearer %s", signedToken)},
+					},
+					URL: reqUrl,
+				},
+				FMetrics: &metricstest.MockMetrics{},
+			}
+
+			b.ResetTimer()
+			b.RunParallel(func(pb *testing.PB) {
+				for pb.Next() {
+					f.Request(ctx)
+					assert.False(b, ctx.FServed)
+				}
+			})
+		})
+	}
+}
+
+// BenchmarkJwtValidationWithVaryingTokenCounts measures the performance of the OPA authorization filter with a JWT validation
+// with JWT cache configured to keep 5 entries and with varying number of tokens.
+func BenchmarkJwtValidationWithVaryingTokenCounts(b *testing.B) {
 	opaControlPlane := opasdktest.MustNewServer(
 		opasdktest.MockBundle(testBundleEndpoint, map[string]string{
 			"main.rego": fmt.Sprintf(`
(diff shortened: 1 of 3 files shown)


amak3:
Additionally more benchmark tests were run to compare JWT cache performance improvement for OPA JWT validation and Styra Caller Identity with and without decision logging.

Test configuration: Apple M1 Max with 16 GB RAM, 8 CPUs, 1 goroutines/parallelism, 10-second duration, 10 repetitions per benchmark

image

amak3:
Additionally run with varying parallelism CPU = 1,2,4,8 to show how scaling by number of CPU looks like:

image

filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
```diff
@@ -199,7 +201,119 @@ func BenchmarkAllowWithReqBody(b *testing.B) {
...
+// with and without JWT cache configuration enabled.
 func BenchmarkJwtValidation(b *testing.B) {
+	type benchmarkCase struct {
+		name     string
+		jwtCache bool
+	}
+
+	cases := []benchmarkCase{
+		{
+			name:     "without_jwt_filter",
```
Pushpalanka:
```suggestion
			name:     "without_jwt_cache",
```

filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
```diff
@@ -232,50 +346,72 @@ func BenchmarkJwtValidation(b *testing.B) {
...
+	tokenCounts := []int{1, 5, 6, 10, 100}
+	for _, tokenCount := range tokenCounts {
+		tokens := make([]string, tokenCount)
+		for i := 0; i < tokenCount; i++ {
+			claims := jwt.MapClaims{
+				"iss":   "https://some.identity.acme.com",
+				"sub":   "5974934733",
+				"aud":   "nqz3xhorr5",
+				"iat":   time.Now().Add(-time.Minute).UTC().Unix(),
+				"exp":   time.Now().Add(tokenExp).UTC().Unix(),
```
Pushpalanka, line 381: For my understanding:
With time time.Now() I would expect all the generated to tokens to be different. But within the capacity of max entries of 5, we see stable latency. Wondering how it hit the cache when token is different. 
May be it depends on what is exactly cached. Do we some insights? (also trying to rule out the expired tokens are not served from cache)

Pushpalanka:
Applied `don-not-merge` as awaiting `opa-envoy-plugin` release

olevchyk:
> Applied `don-not-merge` as awaiting `opa-envoy-plugin` release

https://github.com/open-policy-agent/opa-envoy-plugin/releases/tag/v1.7.1-envoy
looks like there was a release 3 days ago

filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
amak3: That is correct, all the generated tokens will be different. The test is configured in below way:
- cache is configured to keep 5 entries
- we are running test with varying number of tokens, we are generating tokens and store them in the collection (first run with 1 token generated, second run with 5 tokens generated, third run with 6 tokens generated, etc)
- each request is taking random token from the list of generated tokens (first run will take random token from the list of 1 token, second run will take random token from the list of 5 tokens, third run will take random token from the list of 6 tokens, etc)
- in the result we can see the performance improvement when all the tokens are cached (run with 1 token and 5 tokens generated) and how latency increases for more than 5 tokens generated (because only 5 tokens can be stored in the cache) 

Regarding the expired tokens concern - when the token is fetched from the cache [all the constraints are revalidated including expiration](https://github.com/johanfylling/opa/blob/e47b77f94383e1de84755d841f04b5901c4353cc/v1/topdown/tokens.go#L1151)

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -187,6 +189,21 @@ func WithControlLoopMaxJitter(maxJitter time.Duration) func(*OpenPolicyAgentRegi
 	}
 }
 
+func (registry *OpenPolicyAgentRegistry) initializeCache() {
+	id := uuid.New().String()
+	parsedConfig, err := config.ParseConfig(registry.configTemplate.configTemplate, id)
+	if err != nil {
+		fmt.Printf("failed to parse opa config template: %v", err)
+		return
```
AlexanderYastrebov: Lets return error here 
```go
return fmt.Errorf("failed to parse opa config template: %w", err)
```
and handle it on the caller side.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -187,6 +189,21 @@ func WithControlLoopMaxJitter(maxJitter time.Duration) func(*OpenPolicyAgentRegi
...
+	}
+	interQueryBuiltinValueCache, err := iCache.ParseCachingConfig(parsedConfig.Caching)
+	if err != nil {
+		return
+	}
+
+	registry.valueCache = iCache.NewInterQueryValueCache(context.Background(), interQueryBuiltinValueCache)
```
AlexanderYastrebov, line 203: Looks like registry.valueCache field is only used temporary to return the value. If so then lets return (value, error) pair and get rid of the field.

filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
```diff
@@ -232,50 +346,72 @@ func BenchmarkJwtValidation(b *testing.B) {
...
-		for pb.Next() {
-			f.Request(ctx)
-			assert.False(b, ctx.FServed)
-		}
-	})
+			b.ResetTimer()
+			b.RunParallel(func(pb *testing.PB) {
+				for pb.Next() {
+					rnd := rand.New(rand.NewSource(time.Now().UnixNano() + int64(rand.Int())))
+					token := tokens[rnd.Intn(tokenCount)]
```
AlexanderYastrebov: Why do we need source here? Just use https://pkg.go.dev/math/rand#Intn

filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
```diff
@@ -232,50 +346,72 @@ func BenchmarkJwtValidation(b *testing.B) {
...
+						FStateBag: map[string]interface{}{},
+						FResponse: &http.Response{},
+						FRequest: &http.Request{
+							Header: map[string][]string{
+								"Authorization": {fmt.Sprintf("Bearer %s", token)},
+							},
+							URL: reqUrl,
+						},
+						FMetrics: &metricstest.MockMetrics{},
+					}
```
AlexanderYastrebov: I guess we could prepare an array of test contexts instead of tokens 
so that benchmark body only calls `f.Request(ctxs[rand.Intn(tokenCount)])`

filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
```diff
@@ -199,7 +201,119 @@ func BenchmarkAllowWithReqBody(b *testing.B) {
...
+			jwtCache: false,
+		},
+		{
+			name:     "with_jwt_cache",
+			jwtCache: true,
+		},
+	}
+
+	for _, bc := range cases {
+		opaControlPlane := opasdktest.MustNewServer(
```
AlexanderYastrebov: Should this be moved outside of the loop?

filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
```diff
@@ -199,7 +201,119 @@ func BenchmarkAllowWithReqBody(b *testing.B) {
...
+			"aud":   "nqz3xhorr5",
+			"iat":   time.Now().Add(-time.Minute).UTC().Unix(),
+			"exp":   time.Now().Add(tokenExp).UTC().Unix(),
+			"email": "someone@example.org",
+		}
+
+		token := jwt.NewWithClaims(jwt.SigningMethodRS256, claims)
+
+		key, err := jwt.ParseRSAPrivateKeyFromPEM(privateKey)
+		require.NoError(b, err, "Failed to parse RSA PEM")
```
AlexanderYastrebov: This could be moved out of the loop

AlexanderYastrebov: I think we can even move it to init() and reuse in all tests/benchmarks
like
```go
var (
	//go:embed testResources/key.pem
	privateKeyPem []byte
	privateKey    = func() *rsa.PrivateKey {
		key, err := jwt.ParseRSAPrivateKeyFromPEM(privateKeyPem)
		if err != nil {
			panic(fmt.Sprintf("Failed to parse RSA PEM: %v", err))
		}
		return key
	}()
)
```

filters/openpolicyagent/openpolicyagent.go
amak3: The idea is to read the configuration and create the cash in the registry (so we could have one cache for all the opa instances). To do that the registry.valueCache filed is introduced to store the cache in registry and later [when opa instances are created](https://github.com/zalando/skipper/pull/3572/files#diff-c149d1342ef3074c13f6b2397eb74e17bdbea1e557e4c93ad7fd083efe1ee1abR614:~:text=%3A%20registry.-,valueCache,-%2C) we are passing the cache that was created in registry (we don't want to create cache here to not to create separate cache for each opa instance) .

filters/openpolicyagent/opaauthorizerequest/benchmark_test.go
amak3: nice idea :)

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -219,6 +237,8 @@ func NewOpenPolicyAgentRegistry(opts ...func(*OpenPolicyAgentRegistry) error) (*
 		registry.maxMemoryBodyParsingSem = semaphore.NewWeighted(DefaultMaxMemoryBodyParsing)
 	}
 
+	_ = registry.initializeCache()
```
AlexanderYastrebov: Here we should return the error. 
Generally there are really very few cases when you can ignore the error, e.g. https://pkg.go.dev/crypto/rand#Read is documented to never return an error or e.g. HTTP response.Body.Close() error [is usually ignored ](https://github.com/golang/go/issues/58406) - in all other cases handle or propagate error to the caller.

amak3: I made changes as we discussed in priv: this implementation will intentionally fail fast in order to surface potential issues early. If it becomes a frequent problem, we’ll revisit and adjust to make it less strict in case of cache misconfiguration

AlexanderYastrebov:
:+1:

Pushpalanka (approved):

PR #3579, opened 2025-08-06
metrics: fix invalid matcher metric

vlktna:
Fix skip_route_invalid metric to only emit when count > 0 instead of always exposing zero values.


routing/datasource.go
```diff
@@ -618,7 +618,7 @@ func receiveRouteMatcher(o Options, out chan<- *routeTable, quit <-chan struct{}
 				invalidRouteIds[err.ID] = struct{}{}
 			}
 
-			if o.Metrics != nil {
+			if o.Metrics != nil && len(errs) > 0 {
```
AlexanderYastrebov: I think it was correct, otherwise invalid count ever becomes zero even after invalid routes are fixed/removed.

vlktna: it’s a bug because it collected `invalid_matcher` with a value of 0 in all clusters by default, even though all routes were valid. I changed the metric logic a bit, now it won’t send a 0 value if the route is valid, but if the route was previously invalid, the metric will be set to 0

vlktna:
the bug is with reasonCounts in processRouteDefs. If a route was fixed, the metric wouldn’t be updated to 0 because the error type was no longer present in the reasonCounts map.

another approach might be to initialize reasonCounts with all error types set to 0

AlexanderYastrebov:
> another approach might be to initialize reasonCounts with all error types set to 0

I like this more because then we don't need to track previous state in the metric backends.

metrics/codahale.go
```diff
@@ -44,20 +44,23 @@ const (
...
-	handler       http.Handler
-	quit          chan struct{}
+	reg                 metrics.Registry
+	createTimer         func() metrics.Timer
+	createCounter       func() metrics.Counter
+	createGauge         func() metrics.GaugeFloat64
+	options             Options
+	handler             http.Handler
+	quit                chan struct{}
+	invalidRouteReasons map[string]bool
```
AlexanderYastrebov: If you want to track previous state lets do it once in the datasource instead of metrics backends.

vlktna: > another approach might be to initialize reasonCounts with all error types set to 0

I would do it this way instead, it less changes in the code

AlexanderYastrebov:
:+1:

vlktna:
👍

PR #3582, opened 2025-08-06
cmd/webhook: start validation webhook in the skipper

vlktna:
- Enhanced routegroup and ingress admitters with advanced validators and shared admission package
- Introduced reusable validators for filters, predicates, routes, and backend definitions
- Started an optional TLS validation webhook server alongside the proxy with dedicated tests

part of https://github.com/zalando/skipper/issues/1618


internal/validation/comprehensive.go
```diff
@@ -0,0 +1,134 @@
...
+			eskipRoute.Predicates = append(eskipRoute.Predicates, predicates...)
+		}
+
+		routes = append(routes, eskipRoute)
+	}
+
+	return routes
+}
+
+func (cv *ComprehensiveValidator) convertIngressToMinimalEskip(ing *definitions.IngressV1Item) []*eskip.Route {
```
AlexanderYastrebov: Looks like we replicate Kubernetes dataclient logic here (and for routegroup as well).
I am wondering if we could refactor dataclient instead to provide validation and reuse existing internal logic for it.

routing/datasource.go
```diff
@@ -483,6 +483,12 @@ func processTreePredicates(r *Route, predicateList []*eskip.Predicate) error {
 	return nil
 }
 
+// ProcessRouteDef processes a route definition for the routing table.
+// This function is exported to be used by validation webhooks.
+func ProcessRouteDef(o *Options, cpm map[string]PredicateSpec, def *eskip.Route) (*Route, error) {
+	return processRouteDef(o, cpm, def)
+}
+
```
AlexanderYastrebov, line 480: This exports internal function which is not good and will block us from refactoring it later.
I think we should provide a dedicated validation function like
```go
func ValidateRoute(o *Options, r *eskip.Route) error 
```

internal/validation/comprehensive.go
AlexanderYastrebov: Alternatively, since we fake backend and ignore other k8s details (like PathMode option which could also be overridden by ingress annotation) anyway, maybe we should extend IngressV1Validator and RouteGroupValidator like:
```diff
-type IngressV1Validator struct{}
+type EskipValidator interface {
+       ValidateFilters(filters []*eskip.Filter) error
+       ValidatePredicates(predicates []*eskip.Predicate) error
+}
+
+type IngressV1Validator struct {
+       EskipValidator
+}
 
 func (igv *IngressV1Validator) Validate(item *IngressV1Item) error {
        var errs []error
@@ -21,22 +28,30 @@ func (igv *IngressV1Validator) Validate(item *IngressV1Item) error {
 
 func (igv *IngressV1Validator) validateFilterAnnotation(annotations map[string]string) error {
        if filters, ok := annotations[IngressFilterAnnotation]; ok {
-               _, err := eskip.ParseFilters(filters)
+               filters, err := eskip.ParseFilters(filters)
                if err != nil {
-                       err = fmt.Errorf("invalid \"%s\" annotation: %w", IngressFilterAnnotation, err)
+                       return fmt.Errorf("invalid \"%s\" annotation: %w", IngressFilterAnnotation, err)
+               }
+               if igv.EskipValidator != nil {
+                       if err := igv.EskipValidator.ValidateFilters(filters); err != nil {
+                               return fmt.Errorf("invalid \"%s\" annotation: %w", IngressFilterAnnotation, err)
+                       }
                }
-               return err
        }
        return nil
 }
 
 func (igv *IngressV1Validator) validatePredicateAnnotation(annotations map[string]string) error {
        if predicates, ok := annotations[IngressPredicateAnnotation]; ok {
-               _, err := eskip.ParsePredicates(predicates)
+               predicates, err := eskip.ParsePredicates(predicates)
                if err != nil {
-                       err = fmt.Errorf("invalid \"%s\" annotation: %w", IngressPredicateAnnotation, err)
+                       return fmt.Errorf("invalid \"%s\" annotation: %w", IngressPredicateAnnotation, err)
+               }
+               if igv.EskipValidator != nil {
+                       if err := igv.EskipValidator.ValidatePredicates(predicates); err != nil {
+                               return fmt.Errorf("invalid \"%s\" annotation: %w", IngressPredicateAnnotation, err)
+                       }
                }
-               return err
        }
        return nil
 }
```
and provide them eskip validator that delegates to new functions like `routing.ValidateFilters`
(see https://github.com/zalando/skipper/pull/3582/files#r2262219202).

skipper.go
```diff
@@ -2224,9 +2231,46 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
+	log.Infof("Starting validation webhook server on %s", o.ValidationWebhookAddress)
+
+	mux := http.NewServeMux()
+
+	routeGroupAdmitter := admission.NewRouteGroupAdmitter(filterRegistry, predicateSpecs)
+	ingressAdmitter := admission.NewIngressAdmitter(filterRegistry, predicateSpecs)
+
+	mux.Handle("/routegroups", admission.Handler(routeGroupAdmitter))
+	mux.Handle("/ingresses", admission.Handler(ingressAdmitter))
+	mux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
```
AlexanderYastrebov: Do we need this?

skipper.go
```diff
@@ -2224,9 +2231,46 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	<-routing.FirstLoad()
 	log.Info("Dataclients are updated once, first load complete")
 
+	// start validation webhook server if enabled
+	if o.ValidationWebhookEnabled {
+		go func() {
+			if err := startValidationWebhook(o, o.filterRegistry(), o.CustomPredicates); err != nil {
```
AlexanderYastrebov: Here it might be better to reference routing options instead (ro.FilterRegistry, ro.Predicates).

cmd/webhook/admission/ingress.go
```diff
@@ -2,34 +2,71 @@ package admission
...
-	if err := iga.IngressValidator.Validate(&ingressItem); err != nil {
-		return &admissionResponse{
-			UID:     req.UID,
-			Allowed: false,
-			Result:  &status{Message: err.Error()},
-		}, nil
-	}
+func (ia *IngressAdmitter) admit(req *admissionRequest) (*admissionResponse, error) {
+	switch req.Operation {
+	case "CREATE", "UPDATE":
```
MustafaSaber: why do we introduce this?

cmd/webhook/admission/routegroup.go
```diff
@@ -2,34 +2,71 @@ package admission
...
 
+func NewRouteGroupAdmitter(filterRegistry filters.Registry, predicateSpecs []routing.PredicateSpec) *RouteGroupAdmitter {
+	return &RouteGroupAdmitter{
+		validator: validation.NewComprehensiveValidator(filterRegistry, predicateSpecs),
+	}
+}
+
 func (rga *RouteGroupAdmitter) name() string {
-	return "routegroup"
+	return "routegroup-admitter"
```
MustafaSaber: Are we sure this won't break anything?

vlktna: fixed

cmd/webhook/admission/ingress.go
vlktna: removed

skipper.go
vlktna: it needs to receive Kubernetes admission webhook requests

vlktna:
Ready for review, working on unit tests

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -7,7 +7,9 @@ import (
 	"github.com/zalando/skipper/eskip"
 )
 
-type IngressV1Validator struct{}
+type IngressV1Validator struct {
+	EskipValidator
```
MustafaSaber: Why do we add the field instead of implementing the interface? Implement the interface and you won't need the nil check for the object

vlktna: fixed

vlktna: I decided to use composition instead of direct implementation because it would require more code changes and methods duplication. A nil check would still be needed because EskipValidator is nil in the in the [webhook](https://github.com/zalando/skipper/blob/master/cmd/webhook/main.go#L56-L57) by default.

dataclients/kubernetes/definitions/eskipvalidator.go
```diff
@@ -0,0 +1,16 @@
+package definitions
+
+import "github.com/zalando/skipper/eskip"
+
+type ResourceContext struct {
+	Namespace    string
+	Name         string
+	ResourceType string // "rg" or "ing"
```
szuecs: I guess you mean "RouteGroup" or "Ingress" right? (we should spell this out if you add such docs)

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -7,47 +7,76 @@ import (
...
+		filters, err := eskip.ParseFilters(filters)
 		if err != nil {
-			err = fmt.Errorf("invalid \"%s\" annotation: %w", IngressFilterAnnotation, err)
+			return fmt.Errorf("invalid \"%s\" annotation: %w", IngressFilterAnnotation, err)
+		}
+		if igv.EskipValidator != nil {
+			ctx := ResourceContext{
+				Namespace:    namespaceString(item.Metadata.Namespace),
+				Name:         item.Metadata.Name,
+				ResourceType: "ing",
```
szuecs: We should better use the real value from `kind` field.

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -7,47 +7,76 @@ import (
...
+		predicates, err := eskip.ParsePredicates(predicates)
 		if err != nil {
-			err = fmt.Errorf("invalid \"%s\" annotation: %w", IngressPredicateAnnotation, err)
+			return fmt.Errorf("invalid \"%s\" annotation: %w", IngressPredicateAnnotation, err)
+		}
+		if igv.EskipValidator != nil {
+			ctx := ResourceContext{
+				Namespace:    namespaceString(item.Metadata.Namespace),
+				Name:         item.Metadata.Name,
+				ResourceType: "ing",
```
szuecs: We should better use the real value from `kind` field.

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -7,47 +7,76 @@ import (
...
+		route, err := eskip.Parse(routes)
 		if err != nil {
-			err = fmt.Errorf("invalid \"%s\" annotation: %w", IngressRoutesAnnotation, err)
+			return fmt.Errorf("invalid \"%s\" annotation: %w", IngressRoutesAnnotation, err)
+		}
+		if igv.EskipValidator != nil {
+			ctx := ResourceContext{
+				Namespace:    namespaceString(item.Metadata.Namespace),
+				Name:         item.Metadata.Name,
+				ResourceType: "ing",
```
szuecs: We should better use the real value from `kind` field.

dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -88,6 +99,15 @@ func (rgv *RouteGroupValidator) validatePredicates(item *RouteGroupItem) error {
 				errs = append(errs, err)
 			} else if len(predicates) != 1 {
 				errs = append(errs, fmt.Errorf("%w at %q", errSinglePredicateExpected, p))
+			} else if rgv.EskipValidator != nil {
+				ctx := ResourceContext{
+					ResourceType: "RouteGroup",
```
szuecs: It's not consistent that you chose "ing" for Ingress and here the IMO right way, but documented "ing" vs "rg" :)

internal/validation/comprehensive.go
```diff
@@ -0,0 +1,115 @@
+package validation
```
szuecs: Why do we need to use "internal" package ?
We don't use internal packages and IMO it's a bad way.

internal/validation/comprehensive.go
```diff
@@ -0,0 +1,115 @@
...
+	_, err := routing.ValidateRoute(cv.routingOptions, route)
+	if validationErr := cv.handleValidationError(err, routeId); validationErr != nil {
+		return fmt.Errorf("predicate validation failed: %w", validationErr)
+	}
+	return nil
+}
+
+func (cv *ComprehensiveValidator) ValidateRoute(ctx definitions.ResourceContext, routes []*eskip.Route) error {
+	for _, route := range routes {
+		routeId := fmt.Sprintf("validation-%s-%s-%s-%s", ctx.ResourceType, ctx.Namespace, ctx.Name, route.Id)
```
szuecs: I think it's bad to merge the resource identifiers (Kubernetes 3-Tuple) and the final outcome (route.Id).

I would rather specify it as:
```
fmt.Sprintf("validation %q %s/%s %s", ctx.ResourceType, ctx.Namespace, ctx.Name, route.Id)
```
Because instead of "-",  "/" is not valid in kubenetes namespace and name and used in kubernetes a lot to identify it.

internal/validation/comprehensive.go
```diff
@@ -0,0 +1,115 @@
...
+	return nil
+}
+
+func (cv *ComprehensiveValidator) ValidatePredicates(ctx definitions.ResourceContext, predicates []*eskip.Predicate) error {
+	predicateNames := make([]string, len(predicates))
+	for i, predicate := range predicates {
+		predicateNames[i] = predicate.Name
+	}
+	predicateNameString := strings.Join(predicateNames, ",")
+	routeId := fmt.Sprintf("validation-%s-%s-%s-predicates-%s", ctx.ResourceType, ctx.Namespace, ctx.Name, predicateNameString)
```
szuecs: similar to https://github.com/zalando/skipper/pull/3582/files#r2329605339

internal/validation/comprehensive.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+}
+
+func (cv *ComprehensiveValidator) ValidateFilters(ctx definitions.ResourceContext, filters []*eskip.Filter) error {
+	filterNames := make([]string, len(filters))
+	for i, filter := range filters {
+		filterNames[i] = filter.Name
+	}
+	filterNameString := strings.Join(filterNames, ",")
+	routeId := fmt.Sprintf("validation-%s-%s-%s-filters-%s", ctx.ResourceType, ctx.Namespace, ctx.Name, filterNameString)
```
szuecs: similar to https://github.com/zalando/skipper/pull/3582/files#r2329605339

internal/validation/comprehensive.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (cv *ComprehensiveValidator) ValidateBackend(ctx definitions.ResourceContext, backend string, backendType eskip.BackendType) error {
+	_, _, err := routing.SplitBackend(backend, backendType, false)
+
+	// Sanitize the backend name for route ID
+	sanitizedBackend := strings.ReplaceAll(backend, "://", "-")
+	sanitizedBackend = strings.ReplaceAll(sanitizedBackend, "/", "-")
+	sanitizedBackend = strings.ReplaceAll(sanitizedBackend, ":", "-")
+	routeId := fmt.Sprintf("validation-%s-%s-%s-backend-%s", ctx.ResourceType, ctx.Namespace, ctx.Name, sanitizedBackend)
```
szuecs: similar to https://github.com/zalando/skipper/pull/3582/files#r2329605339

skipper.go
```diff
@@ -17,13 +17,17 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/zalando/skipper/dataclients/kubernetes/definitions"
+	"github.com/zalando/skipper/internal/validation"
```
szuecs: these 2 lines leak the abstraction of a dataclient

skipper.go
```diff
@@ -17,13 +17,17 @@ import (
...
+
 	stdlog "log"
 
 	ot "github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/zalando/skipper/circuit"
+	"github.com/zalando/skipper/cmd/webhook/admission"
```
szuecs: why "cmd" should be imported at the library core "skipper"?

skipper.go
```diff
@@ -2224,9 +2234,51 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	<-routing.FirstLoad()
 	log.Info("Dataclients are updated once, first load complete")
 
+	// start validation webhook server if enabled
+	if o.ValidationWebhookEnabled {
+		go func() {
+			if err := startValidationWebhook(o, ro.FilterRegistry, ro.Predicates, mtr); err != nil {
+				log.Errorf("Failed to start validation webhook: %v", err)
```
szuecs: Why Errorf and not Fatalf?

vlktna: Yes it makes more sense to fail fast here when `ValidationWebhookEnabled` is true, removed `startValidationWebhook` from goroutine and changed to Fatalf

internal/validation/comprehensive.go
vlktna: changed to validation package

skipper.go
vlktna: creating a common definitions.go out of webhook/admission would require a lot of changes to the existing code, such as exporting structs and introducing an interface for the Handler function

in skipper’s main, we import only validation/webhook, which contains the webhook runner implementation, and that package imports webhook/admission

imho, there’s not enough logic in the admission package to justify another abstraction layer (e.g., importing github.com/zalando/skipper/cmd/webhook/definitions in runner.go instead of github.com/zalando/skipper/cmd/webhook/admission).

cmd/validation/webhook/main.go
```diff
@@ -0,0 +1,71 @@
+package webhook
```
szuecs: this looks like it should be in "cmd"

dataclients/kubernetes/definitions/ingressvalidator.go
vlktna: here is a draft for the approach with interfaces, there are two commits with two different versions

the first one is an implementation without changes to main.go in the webhook
https://github.com/zalando/skipper/pull/3629/commits/2e83042855902a6c6ae58cf6025eb275cfbe1605

in the second version, I added a noopValidator in main.go in the webhook. it’s needed because we can’t use webhook validation as the default option since it requires extra TLS configuration
https://github.com/zalando/skipper/pull/3629/commits/e6847497d17a687ee66395f02de7b51507221faa

imho the approach with composition looks cleaner

MustafaSaber:
One general comment, the change should ideally replace webhook validation in /cmd so we can't have both implemetations (runner and webhook/admission/main) only one should exist as both do the same thing and we need both for validation webhook.

cmd/validation/eskipvalidator.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+type EskipValidator interface {
+	ValidateFilters(ctx ResourceContext, filters []*eskip.Filter) error
+	ValidatePredicates(ctx ResourceContext, predicates []*eskip.Predicate) error
+	ValidateRoute(ctx ResourceContext, routes []*eskip.Route) error
+	ValidateBackend(ctx ResourceContext, backend string, backendType eskip.BackendType) error
+}
+
+type Runner interface {
```
MustafaSaber: Why do you need this to be an interface? If we remove this nothing will fail and we don't have multiple runners

MustafaSaber: yes, but we don't need the interface to do that, right? It can be just a struct with function

cmd/validation/definitions.go
```diff
@@ -0,0 +1,7 @@
+package validation
+
+type Config struct {
```
szuecs: this should be moved to config/config.go and renamed to ValidationConfig or better ValidationWebhook, then we do not have so weird dependencies.

cmd/validation/webhook/main.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"errors"
+	log "github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/cmd/validation"
+	"github.com/zalando/skipper/cmd/webhook/admission"
```
szuecs: "admission" should  be moved to dataclients/kubernetes.

cmd/validation/webhook/main.go
```diff
@@ -0,0 +1,80 @@
+package webhook
```
szuecs: main.go should be moved to cmd/validation/ and therefore in "main" package

config/config.go
```diff
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"flag"
 	"fmt"
+	"github.com/zalando/skipper/cmd/validation"
```
szuecs: This should not be done and if you follow the other comments it will be deleted.

skipper.go
```diff
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"github.com/zalando/skipper/cmd/validation/webhook"
```
szuecs: no imports from cmd/ here, if you follow my comments this will be deleted

skipper.go
```diff
@@ -17,6 +18,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/zalando/skipper/cmd/validation"
```
szuecs: no imports from cmd/ here, if you follow my comments this will be deleted

cmd/validation/definitions.go
vlktna: removed Config struct to avoid overcomplicating imports

dataclients/kubernetes/definitions/ingressvalidator.go
vlktna: Implemented the interface approach

vlktna:
The goal is to run the validation webhook inside Skipper itself, not spin up an additional binary. That's why I moved validation runner in the /validation package and deleted /cmd/validation

cmd/validation/webhook/main.go
vlktna: see https://github.com/zalando/skipper/pull/3582#issuecomment-3333955844

skipper.go
vlktna: see https://github.com/zalando/skipper/pull/3582#issuecomment-3333955844

skipper.go
vlktna: see https://github.com/zalando/skipper/pull/3582#issuecomment-3333955844

cmd/webhook/main.go
```diff
@@ -53,8 +53,16 @@ func main() {
 	var cfg = &config{}
 	cfg.parse()
 
-	rgAdmitter := &admission.RouteGroupAdmitter{RouteGroupValidator: &definitions.RouteGroupValidator{}}
-	ingressAdmitter := &admission.IngressAdmitter{IngressValidator: &definitions.IngressV1Validator{}}
+	rgAdmitter := &admission.RouteGroupAdmitter{
```
MustafaSaber, line 57: Why do we still need that file here? We should have all things removed from `cmd/webhook`

config/config.go
```diff
@@ -527,6 +533,11 @@ func NewConfig() *Config {
 	flag.IntVar(&cfg.Oauth2TokenCookieRemoveSubdomains, "oauth2-token-cookie-remove-subdomains", 1, "sets the number of subdomains to remove from the callback request hostname to obtain token cookie domain")
 	flag.BoolVar(&cfg.Oauth2GrantInsecure, "oauth2-grant-insecure", false, "omits Secure attribute of the token cookie and uses http scheme for callback url")
 	flag.DurationVar(&cfg.WebhookTimeout, "webhook-timeout", 2*time.Second, "sets the webhook request timeout duration")
+	flag.BoolVar(&cfg.ValidationWebhookEnabled, "validation-webhook-enabled", false, "enables validation webhook for incoming requests")
+	flag.StringVar(&cfg.ValidationWebhookAddress, "validation-webhook-address", "0.0.0.0:9000", "address of the validation webhook service")
```
MustafaSaber: I think `:9000` would be better for IPv6 compatibility

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,87 @@ import (
...
+	"github.com/zalando/skipper/metrics"
+	"github.com/zalando/skipper/routing"
 )
 
-type IngressV1Validator struct{}
+type IngressV1Validator struct {
+	FilterRegistry          filters.Registry
+	PredicateSpecs          []routing.PredicateSpec
+	Metrics                 metrics.Metrics
+	EnableWebhookValidation bool
```
MustafaSaber: Better rename to `EnableAdvancedValidation`, this naming implies it enables the webhook itself which is not true

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,87 @@ import (
...
-func (igv *IngressV1Validator) validateFilterAnnotation(annotations map[string]string) error {
-	if filters, ok := annotations[IngressFilterAnnotation]; ok {
-		_, err := eskip.ParseFilters(filters)
+func (igv *IngressV1Validator) validateFilterAnnotation(item *IngressV1Item) error {
+	if filters, ok := item.Metadata.Annotations[IngressFilterAnnotation]; ok {
+		filters, err := eskip.ParseFilters(filters)
 		if err != nil {
 			err = fmt.Errorf("invalid \"%s\" annotation: %w", IngressFilterAnnotation, err)
 		}
+		if igv.EnableWebhookValidation && err == nil {
```
MustafaSaber:
```suggestion
		if err != nil {
			return fmt.Errorf("invalid \"%s\" annotation: %w", IngressFilterAnnotation, err)
		}
		if igv.EnableWebhookValidation {
		...
		}
```

dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -73,6 +83,16 @@ func (rgv *RouteGroupValidator) validateFilters(item *RouteGroupItem) error {
 			} else if len(filters) != 1 {
 				errs = append(errs, fmt.Errorf("%w at %q", errSingleFilterExpected, f))
 			}
+			if rgv.EnableWebhookValidation && len(filters) == 1 {
```
MustafaSaber: also here would return if basic validation fails

dataclients/kubernetes/definitions/validator.go
```diff
@@ -0,0 +1,52 @@
...
+	Namespace    string
+	Name         string
+	ResourceType ResourceType
+}
+
+type Validator[R KubernetesResource] interface {
+	Validate(resource R) error
+}
+
+func validateFilters(resourceCtx ResourceContext, filterRegistry filters.Registry, filters []*eskip.Filter) error {
```
MustafaSaber: I think this is missing metrics update somewhere

routing/datasource.go
```diff
@@ -216,15 +216,18 @@ func receiveRouteDefs(o Options, quit <-chan struct{}) <-chan mergedDefs {
...
 }
 
-// splits the backend address of a route definition into separate
-// scheme and host variables.
-func splitBackend(r *eskip.Route) (string, string, error) {
-	if r.Shunt || r.BackendType == eskip.ShuntBackend || r.BackendType == eskip.LoopBackend ||
-		r.BackendType == eskip.DynamicBackend || r.BackendType == eskip.LBBackend {
+// SplitBackend splits the backend address into separate scheme and host variables.
+// This function is exported to be used by validation components.
+func SplitBackend(backend string, backendType eskip.BackendType, shunt bool) (string, string, error) {
```
MustafaSaber, line 207: why this was changed?

skipper.go
```diff
@@ -2224,6 +2231,14 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	<-routing.FirstLoad()
 	log.Info("Dataclients are updated once, first load complete")
 
+	// start validation webhook server if enabled
+	if o.ValidationWebhookEnabled {
+		runner := validation.NewRunner(mtr)
```
MustafaSaber: I think it's better to have, something like

```golang
Config {
  address: o.ValidationWebhookAddress,
  certFile: o.ValidationWebhookCertFile,
  keyFile: o.ValidationWebhookKeyFile,
  filterRegistry: ro.FilterRegistry,
  predictes: ro.Predicates,
}

startValidation(config *validator.Config)
```

it matches more skipper style in handling configs

MustafaSaber: No need for runner at all

validation/runner.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/zalando/skipper/dataclients/kubernetes/definitions"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/metrics"
+	"github.com/zalando/skipper/routing"
+)
+
+type Runner struct {
+	metrics  metrics.Metrics
+	serverMu sync.Mutex
+	server   *http.Server
```
MustafaSaber: Why to we need to store/set the server? We should run this once no?

validation/runner.go
```diff
@@ -0,0 +1,128 @@
+package validation
```
MustafaSaber: This whole file can be replaced by `cmd/webhook/main.go`

cmd/webhook/main.go
MustafaSaber: For backward compatibility we still need to leave this binary running so this comment is not relevant anymore

validation/runner.go
MustafaSaber: or duplicated

routing/datasource.go
vlktna: to reuse the backend validation logic in datasource and RouteGroupValidator

cmd/webhook/main.go
```diff
@@ -53,8 +53,16 @@ func main() {
...
+		RouteGroupValidator: &definitions.RouteGroupValidator{
+			EnableAdvancedValidation: false,
```
MustafaSaber: I think it's better to add a flag to (webhook & skipper binary) to enable advanced validation. We don't need to make code changes to enable or disable, and will allow tier based rollout. It also should be disabled by default in both places.

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,90 @@ import (
...
-func (igv *IngressV1Validator) validateFilterAnnotation(annotations map[string]string) error {
-	if filters, ok := annotations[IngressFilterAnnotation]; ok {
-		_, err := eskip.ParseFilters(filters)
+func (igv *IngressV1Validator) validateFilterAnnotation(item *IngressV1Item) error {
+	if filters, ok := item.Metadata.Annotations[IngressFilterAnnotation]; ok {
+		filters, err := eskip.ParseFilters(filters)
 		if err != nil {
 			err = fmt.Errorf("invalid \"%s\" annotation: %w", IngressFilterAnnotation, err)
 		}
+		if igv.EnableAdvancedValidation {
```
MustafaSaber, line 39: We should return the error in the preceding error check, else validateFilter will also always fail.

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,90 @@ import (
...
-func (igv *IngressV1Validator) validatePredicateAnnotation(annotations map[string]string) error {
-	if predicates, ok := annotations[IngressPredicateAnnotation]; ok {
-		_, err := eskip.ParsePredicates(predicates)
+func (igv *IngressV1Validator) validatePredicateAnnotation(item *IngressV1Item) error {
+	if predicates, ok := item.Metadata.Annotations[IngressPredicateAnnotation]; ok {
+		predicates, err := eskip.ParsePredicates(predicates)
 		if err != nil {
 			err = fmt.Errorf("invalid \"%s\" annotation: %w", IngressPredicateAnnotation, err)
 		}
+		if igv.EnableAdvancedValidation {
```
MustafaSaber, line 60: same as https://github.com/zalando/skipper/pull/3582/files#r2415999548

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,90 @@ import (
...
+				Predicates:     igv.PredicateSpecs,
+			}
+			for _, r := range routes {
+				err := validateRoute(ResourceContext{
+					Namespace:    item.Metadata.Namespace,
+					Name:         item.Metadata.Name,
+					ResourceType: ResourceTypeIngress,
+				}, opts, igv.Metrics, r)
+				if err != nil {
+					return fmt.Errorf("invalid \"%s\" annotation: %w", IngressRoutesAnnotation, err)
```
MustafaSaber: This we can do the join and return and the end. The idea from returning or joining is to reduce feedback cycle. If basic validation fails it doesn't make sense to do advanced validation. If advanced validation starts we should collect all errors to user instead of returning error route by route.

dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -73,6 +83,19 @@ func (rgv *RouteGroupValidator) validateFilters(item *RouteGroupItem) error {
 			} else if len(filters) != 1 {
 				errs = append(errs, fmt.Errorf("%w at %q", errSingleFilterExpected, f))
 			}
+			if len(errs) > 0 {
+				continue
+			}
+			if rgv.EnableAdvancedValidation && len(filters) == 1 {
```
MustafaSaber: `len(filters) == 1` doesn't fit anymore with the error condition to continue above.

dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -82,13 +105,30 @@ func (rgv *RouteGroupValidator) validateFilters(item *RouteGroupItem) error {
 func (rgv *RouteGroupValidator) validatePredicates(item *RouteGroupItem) error {
 	var errs []error
 	for _, r := range item.Spec.Routes {
-		for _, p := range r.Predicates {
+		for routePredicate, p := range r.Predicates {
+			//if strings.TrimSpace(p) == "" {
```
MustafaSaber: please remove the comments

validation/validation.go
```diff
@@ -0,0 +1,101 @@
...
+			EnableAdvancedValidation: true,
+		},
+	}
+
+	ingressAdmitter := &admission.IngressAdmitter{
+		IngressValidator: &definitions.IngressV1Validator{
+			FilterRegistry:           filterRegistry,
+			PredicateSpecs:           predicateSpecs,
+			Metrics:                  mtr,
+			EnableAdvancedValidation: true,
```
MustafaSaber: This should be provided by a flag and disabled by default.

MustafaSaber:
in `clusterclient` we also use validator objects I think we should set the advanced value explicitly.

vlktna:
> in clusterclient we also use validator objects I think we should set the advanced value explicitly.

Created a separate ticket
https://github.com/zalando/skipper/issues/3659

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,106 @@ import (
...
-		_, err := eskip.ParseFilters(filters)
+func (igv *IngressV1Validator) validateFilterAnnotation(item *IngressV1Item) error {
+	if filters, ok := item.Metadata.Annotations[IngressFilterAnnotation]; ok {
+		var errs []error
+		parsedFilters, err := eskip.ParseFilters(filters)
 		if err != nil {
-			err = fmt.Errorf("invalid \"%s\" annotation: %w", IngressFilterAnnotation, err)
+			errs = append(errs, fmt.Errorf("invalid \"%s\" annotation: %w", IngressFilterAnnotation, err))
+		}
+		if len(errs) > 0 {
```
MustafaSaber: This seems redundant, if `err != nil` return

dataclients/kubernetes/definitions/routegroupvalidator.go
```diff
@@ -73,6 +83,19 @@ func (rgv *RouteGroupValidator) validateFilters(item *RouteGroupItem) error {
 			} else if len(filters) != 1 {
 				errs = append(errs, fmt.Errorf("%w at %q", errSingleFilterExpected, f))
 			}
+			if len(errs) > 0 {
+				continue
+			}
+			if rgv.EnableAdvancedValidation {
```
MustafaSaber: We can improve this by calling only once outside the loop.

dataclients/kubernetes/definitions/validator.go
```diff
@@ -0,0 +1,124 @@
...
+	originalID := route.Id
+	routeId := buildValidationRouteID(resourceCtx, "route", originalID)
+	validationRoute := *route
+	validationRoute.Id = routeId
+
+	options := baseOptions
+	options.Metrics = mtr
+
+	_, err := routing.ValidateRoute(&options, &validationRoute)
+	if validationErr := handleValidationError(mtr, err, routeId); validationErr != nil {
```
MustafaSaber: this can be

```go
return handleValidationError(mtr, err, routeId)
```

And do the error format in the function

dataclients/kubernetes/definitions/validator.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	if validationErr := handleValidationError(mtr, err, routeId); validationErr != nil {
+		return fmt.Errorf("backend validation failed: %w", validationErr)
+	}
+
+	return nil
+}
+
+func handleValidationError(mtr metrics.Metrics, err error, routeId string) error {
```
MustafaSaber: Maybe we can move this function into `datasource.go` to not export the error which is only used here 🤔

config/config.go
```diff
@@ -534,6 +536,11 @@ func NewConfig() *Config {
 	flag.IntVar(&cfg.Oauth2TokenCookieRemoveSubdomains, "oauth2-token-cookie-remove-subdomains", 1, "sets the number of subdomains to remove from the callback request hostname to obtain token cookie domain")
 	flag.BoolVar(&cfg.Oauth2GrantInsecure, "oauth2-grant-insecure", false, "omits Secure attribute of the token cookie and uses http scheme for callback url")
 	flag.DurationVar(&cfg.WebhookTimeout, "webhook-timeout", 2*time.Second, "sets the webhook request timeout duration")
+	flag.BoolVar(&cfg.ValidationWebhookEnabled, "validation-webhook-enabled", false, "enables validation webhook for incoming requests")
+	flag.StringVar(&cfg.ValidationWebhookAddress, "validation-webhook-address", ":9000", "address of the validation webhook service")
+	flag.StringVar(&cfg.ValidationWebhookCertFile, "validation-webhook-cert-file", "", "path to the certificate file for the validation webhook")
+	flag.StringVar(&cfg.ValidationWebhookKeyFile, "validation-webhook-key-file", "", "path to the key file for the validation webhook")
```
szuecs, line 542: I think we miss here `EnableAdvancedValidation`

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,106 @@ import (
...
-func (igv *IngressV1Validator) validateFilterAnnotation(annotations map[string]string) error {
-	if filters, ok := annotations[IngressFilterAnnotation]; ok {
-		_, err := eskip.ParseFilters(filters)
+func (igv *IngressV1Validator) validateFilterAnnotation(item *IngressV1Item) error {
+	if filters, ok := item.Metadata.Annotations[IngressFilterAnnotation]; ok {
+		var errs []error
+		parsedFilters, err := eskip.ParseFilters(filters)
 		if err != nil {
-			err = fmt.Errorf("invalid \"%s\" annotation: %w", IngressFilterAnnotation, err)
+			errs = append(errs, fmt.Errorf("invalid \"%s\" annotation: %w", IngressFilterAnnotation, err))
```
szuecs: `\"%s\"` -> `%q`

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,106 @@ import (
...
+			return errors.Join(errs...)
 		}
-		return err
+		if igv.EnableAdvancedValidation {
+			if err := validateFilters(ResourceContext{
+				Namespace:    item.Metadata.Namespace,
+				Name:         item.Metadata.Name,
+				ResourceType: ResourceTypeIngress,
+			}, igv.FilterRegistry, igv.Metrics, parsedFilters); err != nil {
+				errs = append(errs, fmt.Errorf("invalid \"%s\" annotation: %w", IngressFilterAnnotation, err))
```
szuecs: `\"%s\"` -> `%q`

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,106 @@ import (
...
-func (igv *IngressV1Validator) validatePredicateAnnotation(annotations map[string]string) error {
-	if predicates, ok := annotations[IngressPredicateAnnotation]; ok {
-		_, err := eskip.ParsePredicates(predicates)
+func (igv *IngressV1Validator) validatePredicateAnnotation(item *IngressV1Item) error {
+	if predicates, ok := item.Metadata.Annotations[IngressPredicateAnnotation]; ok {
+		var errs []error
+		parsedPredicates, err := eskip.ParsePredicates(predicates)
 		if err != nil {
-			err = fmt.Errorf("invalid \"%s\" annotation: %w", IngressPredicateAnnotation, err)
+			errs = append(errs, fmt.Errorf("invalid \"%s\" annotation: %w", IngressPredicateAnnotation, err))
```
szuecs: `\"%s\"` -> `%q`

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,106 @@ import (
...
+		if len(errs) > 0 {
+			return errors.Join(errs...)
+		}
+		if igv.EnableAdvancedValidation {
+			if err := validatePredicates(ResourceContext{
+				Namespace:    item.Metadata.Namespace,
+				Name:         item.Metadata.Name,
+				ResourceType: ResourceTypeIngress,
+			}, igv.PredicateSpecs, igv.Metrics, parsedPredicates); err != nil {
+				errs = append(errs, fmt.Errorf("invalid \"%s\" annotation: %w", IngressPredicateAnnotation, err))
```
szuecs: `\"%s\"` -> `%q`

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,106 @@ import (
...
-func (igv *IngressV1Validator) validateRoutesAnnotation(annotations map[string]string) error {
-	if routes, ok := annotations[IngressRoutesAnnotation]; ok {
-		_, err := eskip.Parse(routes)
+func (igv *IngressV1Validator) validateRoutesAnnotation(item *IngressV1Item) error {
+	if routes, ok := item.Metadata.Annotations[IngressRoutesAnnotation]; ok {
+		var errs []error
+		parsedRoutes, err := eskip.Parse(routes)
 		if err != nil {
-			err = fmt.Errorf("invalid \"%s\" annotation: %w", IngressRoutesAnnotation, err)
+			errs = append(errs, fmt.Errorf("invalid \"%s\" annotation: %w", IngressRoutesAnnotation, err))
```
szuecs: `\"%s\"` -> `%q`

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,106 @@ import (
...
+				FilterRegistry: igv.FilterRegistry,
+				Predicates:     igv.PredicateSpecs,
+			}
+			for _, r := range parsedRoutes {
+				if err := validateRoute(ResourceContext{
+					Namespace:    item.Metadata.Namespace,
+					Name:         item.Metadata.Name,
+					ResourceType: ResourceTypeIngress,
+				}, opts, igv.Metrics, r); err != nil {
+					errs = append(errs, fmt.Errorf("invalid \"%s\" annotation: %w", IngressRoutesAnnotation, err))
```
szuecs: `\"%s\"` -> `%q`

cmd/webhook/main.go
```diff
@@ -11,7 +12,6 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	log "github.com/sirupsen/logrus"
-	"github.com/zalando/skipper/cmd/webhook/admission"
```
szuecs, line 14: do you use goimports?
Somehow your imports are a bit weird structured.
Normally goimports does sort by name sections. Sections are 
1. stdlib
2. non stdlib dependencies

In this file `"github.com/zalando/skipper/dataclients/kubernetes/admission"` moved to the stdlib section.
We see this also in other files.

validation/validation.go
```diff
@@ -0,0 +1,101 @@
...
+	mux.Handle("/metrics", promhttp.Handler())
+	mux.HandleFunc("/healthz", healthCheck)
+
+	return mux
+}
+
+func healthCheck(writer http.ResponseWriter, _ *http.Request) {
+	writer.WriteHeader(http.StatusOK)
+	if _, err := writer.Write([]byte("ok")); err != nil {
+		log.Errorf("failed to write health check: %v", err)
```
szuecs: log starts upper case: `"Failed ..`

validation/validation.go
```diff
@@ -0,0 +1,101 @@
...
+		err := server.Shutdown(context.Background())
+		if err != nil {
+			return
+		}
+	}()
+
+	err := server.ListenAndServeTLS(certFile, keyFile)
+
+	if err != nil && !errors.Is(err, http.ErrServerClosed) {
+		log.Fatalf("Listener error: %v.", err)
```
szuecs: maybe this would be nicer to have the fatal/error comment pattern: `"Failed to listen: %v", err`

cmd/webhook/main.go
vlktna: fixed with goimports

validation/validation.go
vlktna: also changed comment in the webhook/main
https://github.com/zalando/skipper/pull/3582/files#diff-d66a55a199ff817b0b9bfbe228afc8185087487daf79edd85e2a1a7d26a5a49eR113

dataclients/kubernetes/definitions/ingressvalidator.go
szuecs: Then the question would be why not return the error instead of joining a single error slice?

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,106 @@ import (
...
+			return errors.Join(errs...)
 		}
-		return err
+		if igv.EnableAdvancedValidation {
+			if err := validateFilters(ResourceContext{
+				Namespace:    item.Metadata.Namespace,
+				Name:         item.Metadata.Name,
+				ResourceType: ResourceTypeIngress,
+			}, igv.FilterRegistry, igv.Metrics, parsedFilters); err != nil {
+				errs = append(errs, fmt.Errorf("invalid %q annotation: %w", IngressFilterAnnotation, err))
```
szuecs: why not return the err here?

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,106 @@ import (
...
-func (igv *IngressV1Validator) validatePredicateAnnotation(annotations map[string]string) error {
-	if predicates, ok := annotations[IngressPredicateAnnotation]; ok {
-		_, err := eskip.ParsePredicates(predicates)
+func (igv *IngressV1Validator) validatePredicateAnnotation(item *IngressV1Item) error {
+	if predicates, ok := item.Metadata.Annotations[IngressPredicateAnnotation]; ok {
+		var errs []error
+		parsedPredicates, err := eskip.ParsePredicates(predicates)
 		if err != nil {
-			err = fmt.Errorf("invalid \"%s\" annotation: %w", IngressPredicateAnnotation, err)
+			errs = append(errs, fmt.Errorf("invalid %q annotation: %w", IngressPredicateAnnotation, err))
```
szuecs: here we should return the `err` instead of joining and return later

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,106 @@ import (
...
+		if len(errs) > 0 {
+			return errors.Join(errs...)
+		}
+		if igv.EnableAdvancedValidation {
+			if err := validatePredicates(ResourceContext{
+				Namespace:    item.Metadata.Namespace,
+				Name:         item.Metadata.Name,
+				ResourceType: ResourceTypeIngress,
+			}, igv.PredicateSpecs, igv.Metrics, parsedPredicates); err != nil {
+				errs = append(errs, fmt.Errorf("invalid %q annotation: %w", IngressPredicateAnnotation, err))
```
szuecs: same here `return err`

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,106 @@ import (
...
-func (igv *IngressV1Validator) validateRoutesAnnotation(annotations map[string]string) error {
-	if routes, ok := annotations[IngressRoutesAnnotation]; ok {
-		_, err := eskip.Parse(routes)
+func (igv *IngressV1Validator) validateRoutesAnnotation(item *IngressV1Item) error {
+	if routes, ok := item.Metadata.Annotations[IngressRoutesAnnotation]; ok {
+		var errs []error
+		parsedRoutes, err := eskip.Parse(routes)
 		if err != nil {
-			err = fmt.Errorf("invalid \"%s\" annotation: %w", IngressRoutesAnnotation, err)
+			errs = append(errs, fmt.Errorf("invalid %q annotation: %w", IngressRoutesAnnotation, err))
```
szuecs: same here `return err`

dataclients/kubernetes/definitions/ingressvalidator.go
```diff
@@ -5,49 +5,106 @@ import (
...
+				FilterRegistry: igv.FilterRegistry,
+				Predicates:     igv.PredicateSpecs,
+			}
+			for _, r := range parsedRoutes {
+				if err := validateRoute(ResourceContext{
+					Namespace:    item.Metadata.Namespace,
+					Name:         item.Metadata.Name,
+					ResourceType: ResourceTypeIngress,
+				}, opts, igv.Metrics, r); err != nil {
+					errs = append(errs, fmt.Errorf("invalid %q annotation: %w", IngressRoutesAnnotation, err))
```
szuecs, line 92: same here `return err`

dataclients/kubernetes/definitions/routegroupvalidator.go
szuecs: Yes do it in the outer loop and pass `r.Filters`

dataclients/kubernetes/definitions/ingressvalidator.go
MustafaSaber: In this place it make sense to return yes, but in other places it makes sense to join to reduce feedback cycle.

dataclients/kubernetes/definitions/routegroupvalidator.go
vlktna: we can't pass `r.Filters` because the existing validation needs filters and predicates in eskip format to avoid extra steps converting between formats

your suggestion would require storing all parsed filters `parsed, err := eskip.ParseFilters(f)` in one slice and pass it all together. However, since there might be many routes, creating an additional slice could be memory-consuming. The `validateFilters` call runs inside the binary and doesn’t make any external requests, so it executes quickly, and the additional calls are not performance-critical given the stack depth

dataclients/kubernetes/definitions/ingressvalidator.go
vlktna: here `ValidateRoute` is called inside the loop to reuse the validation logic from datasource for a single route

dataclients/kubernetes/definitions/validator.go
vlktna: added HandleValidationError in the separate `routing/errors.go` file

dataclients/kubernetes/definitions/routegroupvalidator.go
szuecs: `r.Filters` is `[]string`, `eskip.ParseFilters` signature:
```
func ParseFilters(f string) ([]*Filter, error) {
```

you can do something like:
```
eskip.ParseFilters(strings.Join(r.Filters, " -> "))
```
see also https://go.dev/play/p/CWNtF_ROHBo

MustafaSaber:
Thanks for the great job! We can land this and address other concerns in follow up PRs if needed.

MustafaSaber (approved):

MustafaSaber:
👍

MustafaSaber:
👍

szuecs:
:+1:

PR #3600, opened 2025-08-26
Optimize: jwt regexp predicates

szuecs:
optimize: JWT regexp based predicates

Context:
Normally clients reuse their tokens so we can easily cache the matching result by token for each predicate.
This will speedup significantly FabricGateway matching lookups.

Adds a sync.Map that we use to perform a cache lookup of the wire string to the result.
A new type `registry` is used to cleanup all entries every hour. For simplicity we do not track timestamps. Tested by a benchmark to cleanup every Millisecond and the result does not change significantly, so we should not care about the recreation of the cache once in a while.

Benchmarks for benchstat comparision ran with count 20
```
goos: darwin
goarch: arm64
pkg: github.com/zalando/skipper/predicates/auth
cpu: Apple M2
                        │    old.txt    │               new.txt               │
                        │    sec/op     │   sec/op     vs base                │
JWTPayloadAllKVRegexp     4079.00n ± 1%   45.59n ± 0%  -98.88% (p=0.000 n=20)
JWTPayloadAllKVRegexp-2   3600.00n ± 0%   45.57n ± 0%  -98.73% (p=0.000 n=20)
JWTPayloadAllKVRegexp-4   3607.00n ± 0%   45.59n ± 0%  -98.74% (p=0.000 n=20)
JWTPayloadAllKVRegexp-8   3603.00n ± 0%   45.70n ± 0%  -98.73% (p=0.000 n=20)
geomean                     3.717µ        45.61n       -98.77%

                        │   old.txt    │                  new.txt                  │
                        │     B/op     │     B/op      vs base                     │
JWTPayloadAllKVRegexp     2.250Ki ± 0%   0.000Ki ± 0%  -100.00% (p=0.000 n=20)
JWTPayloadAllKVRegexp-2   2.250Ki ± 0%   0.000Ki ± 0%  -100.00% (p=0.000 n=20)
JWTPayloadAllKVRegexp-4   2.250Ki ± 0%   0.000Ki ± 0%  -100.00% (p=0.000 n=20)
JWTPayloadAllKVRegexp-8   2.250Ki ± 0%   0.000Ki ± 0%  -100.00% (p=0.000 n=20)
geomean                   2.250Ki                      ?                       ¹ ²
¹ summaries must be >0 to compute geomean
² ratios must be >0 to compute geomean

                        │  old.txt   │                new.txt                 │
                        │ allocs/op  │ allocs/op  vs base                     │
JWTPayloadAllKVRegexp     51.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=20)
JWTPayloadAllKVRegexp-2   51.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=20)
JWTPayloadAllKVRegexp-4   51.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=20)
JWTPayloadAllKVRegexp-8   51.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=20)
geomean                   51.00                   ?                       ¹ ²
¹ summaries must be >0 to compute geomean
² ratios must be >0 to compute geomean
```

old:
```
% go test -run '^$' -benchmem -cpu 1,2,4,8 -bench BenchmarkJWTPayloadAnyKVRegexp ./predicates/auth -count 1
goos: darwin
goarch: arm64
pkg: github.com/zalando/skipper/predicates/auth
cpu: Apple M2
BenchmarkJWTPayloadAnyKVRegexp            311738              3985 ns/op            2304 B/op         51 allocs/op
BenchmarkJWTPayloadAnyKVRegexp-2          338103              3524 ns/op            2304 B/op         51 allocs/op
BenchmarkJWTPayloadAnyKVRegexp-4          343358              3519 ns/op            2304 B/op         51 allocs/op
BenchmarkJWTPayloadAnyKVRegexp-8          342327              3543 ns/op            2304 B/op         51 allocs/op
PASS
ok      github.com/zalando/skipper/predicates/auth      5.013s
```
new:
```
% go test -run '^$' -benchmem -cpu 1,2,4,8 -bench BenchmarkJWTPayloadAnyKVRegexp ./predicates/auth -count 1               
goos: darwin
goarch: arm64
pkg: github.com/zalando/skipper/predicates/auth
cpu: Apple M2
BenchmarkJWTPayloadAnyKVRegexp          24621844                45.77 ns/op            0 B/op          0 allocs/op
BenchmarkJWTPayloadAnyKVRegexp-2        26927236                45.68 ns/op            0 B/op          0 allocs/op
BenchmarkJWTPayloadAnyKVRegexp-4        26123988                45.59 ns/op            0 B/op          0 allocs/op
BenchmarkJWTPayloadAnyKVRegexp-8        26365248                45.65 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/zalando/skipper/predicates/auth      4.952s
```

### DIFF
--- a/predicates/auth/jwt.go
+++ b/predicates/auth/jwt.go
@@ -20,6 +20,8 @@ import (
 	"net/http"
 	"regexp"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/zalando/skipper/jwt"
 	"github.com/zalando/skipper/predicates"
@@ -49,14 +51,21 @@ const (
 )
 
 type (
+	registry struct {
+		quit       chan struct{}
+		predicates []*predicate
+	}
+
 	spec struct {
 		name          string
 		matchBehavior matchBehavior
 		matchMode     matchMode
+		reg           *registry
 	}
 	predicate struct {
 		kv            map[string][]valueMatcher
 		matchBehavior matchBehavior
+		cache         sync.Map
 	}
 	exactMatcher struct {
 		expected string
@@ -65,6 +74,26 @@ type (
 		regexp *regexp.Regexp
 	}
 )
+
+func (r *registry) Close() {
+	close(r.quit)
+}
+
+func (r *registry) clean() {
+	tick := time.NewTicker(time.Hour)
+	defer tick.Stop()
+
+	for {
+		select {
+		case <-r.quit:
+			return
+		case <-tick.C:
+			for _, p := range r.predicates {
+				p.cache.Clear()
+			}
+		}
+	}
+}
 
 func NewJWTPayloadAnyKV() routing.PredicateSpec {
 	return &spec{
@@ -83,18 +112,32 @@ func NewJWTPayloadAllKV() routing.PredicateSpec {
 }
 
 func NewJWTPayloadAnyKVRegexp() routing.PredicateSpec {
+	reg := &registry{
+		quit:       make(chan struct{}),
+		predicates: make([]*predicate, 0),
+	}
+	go reg.clean()
+
 	return &spec{
 		name:          predicates.JWTPayloadAnyKVRegexpName,
 		matchBehavior: matchBehaviorAny,
 		matchMode:     matchModeRegexp,
+		reg:           reg,
 	}
 }
 
 func NewJWTPayloadAllKVRegexp() routing.PredicateSpec {
+	reg := &registry{
+		quit:       make(chan struct{}),
+		predicates: make([]*predicate, 0),
+	}
+	go reg.clean()
+
 	return &spec{
 		name:          predicates.JWTPayloadAllKVRegexpName,
 		matchBehavior: matchBehaviorAll,
 		matchMode:     matchModeRegexp,
+		reg:           reg,
 	}
 }
 
@@ -131,10 +174,17 @@ func (s *spec) Create(args []interface{}) (routing.Predicate, error) {
 		kv[key] = append(kv[key], matcher)
 	}
 
-	return &predicate{
+	if s.matchMode == matchModeRegexp {
+
+	}
+
+	p := &predicate{
 		kv:            kv,
 		matchBehavior: s.matchBehavior,
-	}, nil
+	}
+	s.reg.predicates = append(s.reg.predicates, p)
+
+	return p, nil
 }
 
 func (m exactMatcher) Match(jwtValue string) bool {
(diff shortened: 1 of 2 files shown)


a4180p:
👍

greeshma1196:
👍

predicates/auth/jwt.go
```diff
@@ -66,6 +75,26 @@ type (
 	}
 )
 
+func (r *registry) Close() {
+	close(r.quit)
+}
+
+func (r *registry) clean() {
+	tick := time.NewTicker(time.Hour)
```
MustafaSaber, line 83: Should we make this customisable? Or at least a const somewhere to read from?

MustafaSaber (commented):
Left one comment that can be addressed in another PR if needed.

predicates/auth/jwt.go
```diff
@@ -131,10 +174,15 @@ func (s *spec) Create(args []interface{}) (routing.Predicate, error) {
...
 
-	return &predicate{
+	p := &predicate{
 		kv:            kv,
 		matchBehavior: s.matchBehavior,
-	}, nil
+	}
+	if s.matchMode == matchModeRegexp {
+		s.reg.predicates = append(s.reg.predicates, p)
+	}
```
AlexanderYastrebov, line 183: 1. Here we track only regexp predicates but cache is used by all predicate types, isn't it?
2. Predicate instances are created on every route table update (potentially each 3s by default) and this list is never trimmed so I think there is a potential leak here.
3. Each predicate instance uses its own cache which means the same value might be cached many times and does not survive routing table update. A different strategy could be to have registry-global two-level cache `map[<predicate eskip>][<token value>]`. This way cached values would be reused by different instances of equal predicates (i.e. those that have the same name+arguments) and will survive route table update.

szuecs: > 1. Here we track only regexp predicates but cache is used by all predicate types, isn't it?

Yes, and that should be fine, because if we test against a fixed value, then we do not need to clear the cache.

>     2. Predicate instances are created on every route table update (potentially each 3s by default) and this list is never trimmed so I think there is a potential leak here.

Yes you are right that we missed.

>     3. Each predicate instance uses its own cache which means the same value might be cached many times and does not survive routing table update. A different strategy could be to have registry-global two-level cache `map[<predicate eskip>][<token value>]`. This way cached values would be reused by different instances of equal predicates (i.e. those that have the same name+arguments) and will survive route table update.

So we could have one level if we would cache, similar to kubernetes, a 2-tuple:
```
type cacheKey struct {
    p eskip.Predicate
    val string
}

func (m exactMatcher) Match(jwtValue string) bool {
    ahead := r.Header.Get(authHeaderName)
....
    p.cache.Store(&cacheKey{m.p, ahead}, false)
....
}
```

szuecs: Let's do a further discussion https://github.com/zalando/skipper/issues/3609

PR #3601, opened 2025-08-27
Fix config template parsing issue

amak3:
Context:
The config template in registry is parsed as the real config to extract caching configuration. It's possible that config template is not valid yaml file and then the parsing fails. 

Fix:
To fix the issue the config template is being interpolated before parsing.


filters/openpolicyagent/openpolicyagent.go
```diff
@@ -190,8 +190,13 @@ func WithControlLoopMaxJitter(maxJitter time.Duration) func(*OpenPolicyAgentRegi
 }
 
 func (registry *OpenPolicyAgentRegistry) initializeCache() error {
+	configBytes, err := registry.configTemplate.interpolateConfigTemplate("dummy-bundle-name.tar.gz")
```
mjungsbluth: A comment on why this is ok in production code would be nice (also you can strip the `.tar.gz` suffix): This is not used apart from making sure the config is parseable.

amak3:
👍

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -190,8 +190,15 @@ func WithControlLoopMaxJitter(maxJitter time.Duration) func(*OpenPolicyAgentRegi
 }
 
 func (registry *OpenPolicyAgentRegistry) initializeCache() error {
+	// This line interpolates the config template with a dummy bundle name to make sure the config is parseable.
+	// It is safe in production because the result is not used for anything except caching configuration.
+	configBytes, err := registry.configTemplate.interpolateConfigTemplate("dummy-bundle-name")
```
MustafaSaber, line 195: can we generate this random name?

szuecs: Is it random @MustafaSaber ?
I think it's just a fixed name for testing that a known bundle works.

My question would be what this bundle is and where does it come from?
It sounds as if we rely on our Zalando configuration or will it work in general for everyone?

amak3: It's not random, it's dummy name only used to be able to parse config. 

The only configuration that we need on this (registry) level is caching. To parse the config to access the caching configuration we need to interpolate it and at this point we don't know bundle name yet, so we use dummy name. The result is not used anywhere at this point, we only get caching configuration. 

Later on OPA instance creation we already know the bundle name and we need all config file values, so we interpolate it with correct bundle name:
https://github.com/zalando/skipper/blob/40a816da754db2a13b871097704e4603d3933b3e/filters/openpolicyagent/openpolicyagent.go#L572

MustafaSaber: I suggested the random name, as I think this will generate temp file with this bundle name, is this wrong? If it's going to generate the file and later we try to init but this file is there will it override or reuse?

szuecs: @MustafaSaber I will merge it, I think it's fine and let's see if there is an issue and fix later if needed.

szuecs:
:+1:

PR #3603, opened 2025-08-29
proxy: fix timings calculation

ponimas:
Refactored proxy timing measurement to track request and response phases separately.

The previous implementation used a single stopwatch that failed to
properly handle shunt and loopback scenarios. For loopback routes,
the entire duration of subsequent loopbacks was incorrectly attributed
to response time since each loopback creates its own context.
Additionally, shunt routes never triggered the request-to-response
timer transition, which was located in makeBackendRequest and
therefore unreachable for shunt routes.

Additonally fixing another bug:
The defer statement for finishing the loopSpan was causing response
filter processing time to be incorrectly included in the loopback span
timings. This change moves the loopSpan.Finish() calls to execute
immediately after the recursive do() call completes.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1258,7 +1258,11 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 		if parentSpan := ot.SpanFromContext(ctx.request.Context()); parentSpan != nil {
 			loopSpanOpts = append(loopSpanOpts, ot.ChildOf(parentSpan.Context()))
 		}
+
+		// defer can't be used to finish the loopSpan, because it will be called only at the end of the outer function
+		// leading to distortion of the tracing picture, because response filters processing time will be included into the loopSpan timings
 		loopSpan := p.tracing.tracer.StartSpan("loopback", loopSpanOpts...)
+
 		p.tracing.setTag(loopSpan, SkipperRouteIDTag, ctx.route.Id)
 		p.setCommonSpanInfo(ctx.Request().URL, ctx.Request(), loopSpan)
 		ctx.parentSpan = loopSpan
@@ -1266,16 +1270,14 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 		r := loopCTX.Request()
 		r = r.WithContext(ot.ContextWithSpan(r.Context(), loopSpan))
 		loopCTX.request = r
-
-		defer loopSpan.Finish()
-
 		if err := p.do(loopCTX, loopSpan); err != nil {
 			// in case of error we have to copy the response in this recursion unwinding
 			ctx.response = loopCTX.response
 			p.applyFiltersOnError(ctx, processedFilters)
+			loopSpan.Finish()
 			return err
 		}
-
+		loopSpan.Finish()
 		ctx.setResponse(loopCTX.response, p.flags.PreserveOriginal())
 		ctx.proxySpan = loopCTX.proxySpan
 	} else if p.flags.Debug() {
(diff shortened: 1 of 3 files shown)


proxy/proxy.go
```diff
@@ -1422,6 +1477,14 @@ func (p *Proxy) serveResponse(ctx *context) {
 }
 
 func (p *Proxy) errorResponse(ctx *context, err error) {
+	responseStopWatch := newStopWatch()
+	responseStopWatch.Start()
+	defer func() {
+		responseStopWatch.Stop()
+		ctx.proxyRequestElapsed = 0
```
szuecs: why do we need to set it to `0` ?

proxy/proxy.go
```diff
@@ -1659,13 +1732,17 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
 		p.errorResponse(ctx, err)
+		responseElapsed += ctx.proxyResponseElapsed
 	} else {
 		p.serveResponse(ctx)
+		responseElapsed += ctx.proxyResponseElapsed
 	}
 
+	responseStopWatch.Start()
 	// fifoWtihBody() filter
```
szuecs: fifoWtihBody -> fifoWithBody (as I see it here :) )

proxy/proxy.go
a4180p: It's not needed. Removed.

proxy/proxy.go
```diff
@@ -1659,14 +1730,18 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	responseStopWatch.Stop()
 	// stream response body to client
 	if err != nil {
 		p.errorResponse(ctx, err)
+		responseElapsed += ctx.proxyResponseElapsed
```
szuecs: This can be moved down just behind the if..else block and you can delete one line.

a4180p:
👍

proxy/proxy.go
```diff
@@ -1031,11 +1034,11 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
...
 
-	ctx.proxyWatch.Stop()
-	ctx.proxyRequestLatency = ctx.proxyWatch.Elapsed()
+	requestStopWatch.Stop()
+
 	response, err := roundTripper.RoundTrip(req)
-	ctx.proxyWatch.Reset()
-	ctx.proxyWatch.Start()
+
+	responseStopWatch.Start()
```
MustafaSaber, line 1041: I'm struggling to understand, where do we stop this responseStopWatch?

proxy/proxy.go
```diff
@@ -1199,6 +1207,10 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 			p.makeErrorResponse(ctx, perr)
 			err = perr
 		}
+		requestStopWatch.Stop()
+		responseStopWatch.Stop()
```
MustafaSaber, line 1211: Should this stop the previous responseStopWatch? Aren't they 2 different instances?

a4180p: If by "previous" you mean the one in your comment above - the answer is "no" it should not. These are 2 diffferent instances.

proxy/proxy.go
a4180p: seems the whole defer function is lost
https://github.com/zalando/skipper/pull/3634

MustafaSaber: perfect, thanks!

a4180p: Thanks to you 👍  for drawning attention.

PR #3612, opened 2025-09-04
Keep OPA involved routes valid, while retrying failed OPA instance creation

Pushpalanka:
### Context

Extending from https://github.com/zalando/skipper/pull/3562

The approach is slightly modified as below.
- OPA instance failures will no longer mark the routes as invalid
- Instead the route is added to the routing table and recoverable errors will be retried in a background task
- When requests come to the added route, but OPA instance not ready yet, -> the filter will serve 503 indicating the OPA instance unavailability

### Tasks
- [x] Filter creation does not block routing table additions when OPA instance creation is failed or still in progress.
- [ ] Recoverable failures in OPA instance creation are retried without interfering with traffic (using map that stores the status of each OPA instance)
- [ ] If OPA becomes available after initial failure, ensure routes are serving traffic fine.


Pushpalanka:
We moved out of this approach due to significant number of changes it introduces.
Moved ahead with introducing a new field `healthy` inside the OPA instance itself than this wrapper object approach.

PR #3621, opened 2025-09-11
logging: use same error variable

MustafaSaber:
see subject

### DIFF
--- a/dataclients/kubernetes/clusterstate.go
+++ b/dataclients/kubernetes/clusterstate.go
@@ -43,7 +43,7 @@ func (state *clusterState) getServiceRG(namespace, name string) (*service, error
 	defer state.mu.Unlock()
 	s, ok := state.services[newResourceID(namespace, name)]
 	if !ok {
-		return nil, fmt.Errorf("service not found: %s/%s", namespace, name)
+		return nil, fmt.Errorf("%s/%s: %w", namespace, name, errServiceNotFound)
 	}
 
 	return s, nil
--- a/dataclients/kubernetes/clusterclient_test.go
+++ b/dataclients/kubernetes/clusterclient_test.go
@@ -266,7 +266,7 @@ func TestLoggingInterval(t *testing.T) {
 	defer log.SetOutput(os.Stderr)
 
 	countMessages := func() int {
-		return strings.Count(out.String(), "Ignoring route: service not found: default/myapp")
+		return strings.Count(out.String(), "Ignoring route: default/myapp: service not found")
 	}
 
 	a, err := kubernetestest.NewAPI(kubernetestest.TestAPIOptions{}, manifest)
--- a/dataclients/kubernetes/testdata/routegroups/convert/failing-route-groups-ignored.log
+++ b/dataclients/kubernetes/testdata/routegroups/convert/failing-route-groups-ignored.log
@@ -1,2 +1,2 @@
-Ignoring route: service not found: foo/non-existent
+Ignoring route: foo/non-existent: service not found
 kind=RouteGroup name=myapp-no-service ns=foo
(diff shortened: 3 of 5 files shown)


dataclients/kubernetes/clusterstate.go
```diff
@@ -43,7 +43,7 @@ func (state *clusterState) getServiceRG(namespace, name string) (*service, error
 	defer state.mu.Unlock()
 	s, ok := state.services[newResourceID(namespace, name)]
 	if !ok {
-		return nil, fmt.Errorf("service not found: %s/%s", namespace, name)
+		return nil, fmt.Errorf("%s: %s/%s", errServiceNotFound, namespace, name)
```
szuecs: This should be a `%w`

dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-prule-wrong-ep-port.log
```diff
@@ -1 +1 @@
-level=error msg="Failed to get service bar, baz" kind=Ingress name=qux ns=foo
+level=error msg="Failed to get service foo/bar:baz" kind=Ingress name=qux ns=foo
```
AlexanderYastrebov: There is a namespace already available as structured field (`ns=foo`)

dataclients/kubernetes/testdata/routegroups/with-ingress/failing-ingress-item.log
```diff
@@ -1,2 +1,2 @@
-Failed to get service myappv1, 80
+Failed to get service default/myappv1:80
```
AlexanderYastrebov: Not sure why this is two-line log but ns=default is also already there

AlexanderYastrebov:
PR title does not seem to match the actual change...

AlexanderYastrebov:
:+1:

dataclients/kubernetes/clusterstate.go
```diff
@@ -43,7 +43,7 @@ func (state *clusterState) getServiceRG(namespace, name string) (*service, error
 	defer state.mu.Unlock()
 	s, ok := state.services[newResourceID(namespace, name)]
 	if !ok {
-		return nil, fmt.Errorf("service not found: %s/%s", namespace, name)
+		return nil, fmt.Errorf("%w: %s/%s", errServiceNotFound, namespace, name)
```
szuecs: please use always the pattern: 
```
fmt.Errorf("blah, %s/%s blah: %w", namespace, name, err)
```

dataclients/kubernetes/ingressv1.go
```diff
@@ -96,7 +96,7 @@ func convertPathRuleV1(
 
 	svc, err = state.getService(ns, svcName)
 	if err != nil {
-		ic.logger.Errorf("Failed to get service %s, %s", svcName, svcPort)
+		ic.logger.Errorf("Failed to get service %s:%s", svcName, svcPort)
```
szuecs: not sure if we need a change or why

MustafaSaber: I changed it, to always indicate it's a port even when it's a named port; Something like `my-service:main` for me is better as indicating it's service port

dataclients/kubernetes/clusterstate.go
MustafaSaber: This format for this specific error will be confusing, `Ignoring route: my-namespace/my-service service not found` is a bit confusing, no?

szuecs: No, it's right now confusing because the call stack should not change the order of the error. The last is the deepest error, please stay always with this pattern in all applications we do.

dataclients/kubernetes/ingressv1.go
szuecs: In general  I don't really like changes like this. People can have tools that already work with this pattern and we would break the parsing if we change it.

MustafaSaber: But by this idea we won't be able change any log message. Even the one above where we restructure the error.

szuecs: If we see a benefit we change but not necessarily here, because I don't see the benefit.

Also `:` is used to split error wrap, so if we have a port name it's not obvious and breaks the error wrapping log reading.

MustafaSaber:
👍

szuecs:
:+1:

PR #3622, opened 2025-09-13
fix: ST1000 missing package docs

szuecs:
fix: ST1000 missing package docs

not fully but a good chunk


szuecs:
:+1:

eskip/parser.go
```diff
@@ -1,9 +1,13 @@
+//
 // Code generated by goyacc -l -v  -o parser.go -p eskip parser.y. DO NOT EDIT.
-package eskip
+//
 
-import __yyfmt__ "fmt"
+package eskip
 
-import "strconv"
+import (
```
MustafaSaber: was autogenerated or manually edited?

szuecs: I think it was gopls/go tooling.
The thing is that grouping imports is now the normal and seems to be fine or what do you think?

I had to change the comment lines above to stop the linter to see this as "package comment".

MustafaSaber: I think once this file is regenerated it will be deleted.

szuecs: removed the change
I checked how it was generated and yes you are right.

szuecs:
:+1:

MustafaSaber:
👍

Issue #3596, opened 2025-08-22
Use Go routine labels to add more context to profiling


szuecs:
You can use runtime package to add labels like SpanID to a Go routine metadata that is read by the profiler.
With that we have a link from Tracing to Profiling, so a spike in latency can show the actual profile for it. 

**Is your feature request related to a problem? Please describe.**

To understand what happened inside a process at a certain point of time is great and this tasks allows to link for a SpanID in OTel/OT to a profile, which is even better, because no need of time to be in sync across the systems.

**Describe the solution you would like**

The go runtime/pprof has the abilty of having labels, example:
```
pprof.Do(ctx, pprof.Labels("otel_traceid", traceID), func(ctx context.Context) {
				h.ServeHTTP(w, r.WithContext(ctx))
			}) 
```

I think we want to have `spanID` and set it in our proxy.ServeHTTP and change it on every span change.
All background goroutines can also have labels to identify them.

**Additional context (optional)**

https://www.polarsignals.com/blog/posts/2024/10/16/custom-labels-for-rust-go-and-cpp
https://www.polarsignals.com/blog/posts/2021/04/13/demystifying-pprof-labels-with-go
https://github.com/polarsignals/otel-profiling-go/blob/main/otelhttpprofiling/http.go

szuecs:
change https://github.com/zalando/skipper/pull/3626 was merged and released

PR #3626, opened 2025-09-15
Add trace_id to request ctx

greeshma1196:
updates: https://github.com/zalando/skipper/issues/3596


proxy/proxy.go
```diff
@@ -1698,6 +1699,14 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	r = r.WithContext(ot.ContextWithSpan(r.Context(), span))
 	r = r.WithContext(routing.NewContext(r.Context()))
 
+	rCtx := r.Context()
+	defer pprof.SetGoroutineLabels(rCtx)
+	defer r.WithContext(rCtx)
```
a4180p: What does this line do?

greeshma1196: Resetting the request context to the way it was before adding the labels.

a4180p: Hm... I don't think this function mutates the request.
It returns a copy of the request with the new context. `r` will remain the same.

greeshma1196: Thanks for pointing it out, as discussed I have removed the last defer

szuecs:
:+1:

a4180p:
👍

AlexanderYastrebov:
Its hard to come up with a test for this because there is no API to return goroutine labels, see https://github.com/golang/go/issues/68785
Alternative could be to collect goroutine profile inside of the test filter and then parse it to extract labels but that seems like an overkill for the task:
```diff
diff --git a/proxy/tracing_test.go b/proxy/tracing_test.go
index daaaac81..c0609250 100644
--- a/proxy/tracing_test.go
+++ b/proxy/tracing_test.go
@@ -1,18 +1,24 @@
 package proxy
 
 import (
+       "bytes"
        "fmt"
        "io"
+       "maps"
        "math/rand"
        "net/http"
        "net/http/httptest"
        "net/url"
+       "runtime/pprof"
        "strings"
        "testing"
        "time"
 
+       "github.com/google/pprof/profile"
        ot "github.com/opentracing/opentracing-go"
        "github.com/opentracing/opentracing-go/mocktracer"
+       "github.com/zalando/skipper/filters"
+       "github.com/zalando/skipper/filters/builtin"
        "github.com/zalando/skipper/tracing/tracingtest"
 
        "github.com/stretchr/testify/assert"
@@ -609,10 +615,52 @@ func TestSetTagWithEmptySpan(t *testing.T) {
        tracing.setTag(nil, "test", "val")
 }
 
+type funcFilterSpec func(filters.FilterContext)
+
+func (s funcFilterSpec) Name() string                                              { return "funcFilter" }
+func (s funcFilterSpec) CreateFilter(config []interface{}) (filters.Filter, error) { return s, nil }
+func (s funcFilterSpec) Request(ctx filters.FilterContext)                         { s(ctx) }
+func (s funcFilterSpec) Response(filters.FilterContext)                            {}
+
 func TestTraceIdGoroutineLabel(t *testing.T) {
-       s := startTestServer(nil, 0, func(r *http.Request) {})
-       defer s.Close()
+       fr := make(filters.Registry)
+       fr.Register(builtin.NewStatus())
+       fr.Register(funcFilterSpec(func(fc filters.FilterContext) {
+               buf := new(bytes.Buffer)
+               pprof.Lookup("goroutine").WriteTo(buf, 0)
+
+               p, err := profile.Parse(buf)
+               require.NoError(t, err)
+
+               labels := make(map[string][]string)
+               for _, s := range p.Sample {
+                       maps.Copy(labels, s.Label)
+               }
+               assert.Contains(t, labels, "trace_id")
+       }))
+
+       tracer := tracingtest.NewTracer()
+       params := Params{
+               OpenTracing: &OpenTracingParams{
+                       Tracer: tracer,
+               },
+               Flags: FlagsNone,
+       }
+
+       tp, err := newTestProxyWithFiltersAndParams(fr, `* -> funcFilter() -> status(200) -> <shunt>`, params, nil)
+       if err != nil {
+               t.Fatal(err)
+       }
+       defer tp.close()
+
+       ps := httptest.NewServer(tp.proxy)
+       defer ps.Close()
+
+       res, err := ps.Client().Get(ps.URL + "/")
+       require.NoError(t, err)
+       res.Body.Close()
 
+       assert.Equal(t, res.StatusCode, http.StatusOK)
 }
 
 func findSpanByRouteID(tracer *tracingtest.MockTracer, routeID string) (*tracingtest.MockSpan, bool) {
diff --git a/tracing/tracing.go b/tracing/tracing.go
index be66d209..d291de74 100644
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -175,5 +175,10 @@ func GetTraceID(span ot.Span) string {
                return fmt.Sprintf("%x", spanContextType.TraceID)
        }
 
+       // For testing mocktracer span
+       if traceIdAware, ok := span.(interface{ TraceID() string }); ok {
+               return traceIdAware.TraceID()
+       }
+
        return ""
 }
diff --git a/tracing/tracingtest/mocktracer.go b/tracing/tracingtest/mocktracer.go
index f848d61f..52bf9c00 100644
--- a/tracing/tracingtest/mocktracer.go
+++ b/tracing/tracingtest/mocktracer.go
@@ -74,3 +74,7 @@ func (t *MockTracer) Extract(format any, carrier any) (opentracing.SpanContext,
 func (s *MockSpan) Tracer() opentracing.Tracer {
        return s.t
 }
+
+func (s *MockSpan) TraceID() string {
+       return fmt.Sprintf("%x", s.SpanContext.TraceID)
+}
```

See also related:
* https://github.com/golang/vscode-go/issues/3247
* https://stackoverflow.com/questions/79395470/vscode-launch-config-for-specific-golang-test?noredirect=1&lq=1

PR #3630, opened 2025-09-19
Fix Open Policy Agent quick start after Playground no longer serves bundles

mjungsbluth:
The currently documented Open Policy Agent quick start no longer works as the Rego Playground can no longer serve bundles directly. 

This changes the quick start to a fully local setup.


szuecs (approved):

mjungsbluth:
👍

PR #3631, opened 2025-09-19
Enable native OPA Prometheus metrics

mjungsbluth:
This PR passes through existing Prometheus metrics from Open Policy Agent by prefixing them to fit with remainging Skipper metrics and labeling them per virtual OPA instance. 

This can only be supported for the Prometheus metrics flavour and is breaking the metrics abstraction as it pulls the Prometheus `Registerer` to the outside but avoids maintaining and copying existing work that has happened upstream in Open Policy Agent. 

Should only be merged after #3562 is merged

### DIFF
--- a/filters/openpolicyagent/openpolicyagent_test.go
+++ b/filters/openpolicyagent/openpolicyagent_test.go
@@ -1106,33 +1106,51 @@ func TestBodyExtractionUnknownBody(t *testing.T) {
 func TestPrometheusPluginStatusGaugeRegistered(t *testing.T) {
 	_, config := mockControlPlaneWithResourceBundle(WithStatusPluginEnabled(true))
 
-	reg := prometheus.NewRegistry()
-	registry, err := NewOpenPolicyAgentRegistry(
+	promRegistry := prometheus.NewRegistry()
+	opaRegistry, err := NewOpenPolicyAgentRegistry(
 		WithReuseDuration(1*time.Second),
 		WithCleanInterval(1*time.Second),
 		WithOpenPolicyAgentInstanceConfig(WithConfigTemplate(config)),
-		WithPrometheusRegisterer(reg),
+		WithPrometheusRegisterer(promRegistry),
 	)
 	assert.NoError(t, err)
 
-	inst, err := registry.GetOrStartInstance("test")
+	opaInst, err := opaRegistry.GetOrStartInstance("test")
 	assert.NoError(t, err)
 
 	// Simulate an HTTP request evaluation
 	ctx := context.Background()
-	_, err = inst.Eval(ctx, &authv3.CheckRequest{
+	_, err = opaInst.Eval(ctx, &authv3.CheckRequest{
 		Attributes: &authv3.AttributeContext{},
 	})
 	assert.NoError(t, err)
 
 	// Gather metrics and check for plugin_status_gauge
-	metricsFamilies, err := reg.Gather()
+	metricsFamilies, err := promRegistry.Gather()
 	assert.NoError(t, err)
 
 	found := false
+	foundLabels := map[string]string{
+		"name":              "bundle",
+		"opa_instance_id":   "34b78d90-6d9a-4ecc-8ff9-393b9dafa40d", // random ID
+		"opa_instance_name": "test",
+		"status":            "OK",
+	}
 	for _, mf := range metricsFamilies {
 		if mf.GetName() == "plugin_status_gauge" && len(mf.Metric) > 0 {
-			found = true
+			if len(mf.Metric) > 0 {
+				met := mf.Metric[0]
+				for _, lPair := range met.GetLabel() {
+					switch lPair.GetName() {
+					case "opa_instance_id":
+						found = true // ok, because random ID should always be there
+					default:
+						if foundLabels[lPair.GetName()] != lPair.GetValue() {
+							t.Fatalf("Failed to find label %q: %q, got: %q", lPair.GetName(), foundLabels[lPair.GetName()], lPair.GetValue())
+						}
+					}
+				}
+			}
 			break
 		}
 	}
--- a/metrics/all_kind.go
+++ b/metrics/all_kind.go
@@ -143,7 +143,7 @@ func (a *All) Close() {
 	a.prometheus.Close()
 }
 
-// Implements the PrometheusMetrics interface
+// ScopedPrometheusRegisterer implements the PrometheusMetrics interface
 func (a *All) ScopedPrometheusRegisterer(subsystem string) prometheus.Registerer {
 	return a.prometheus.ScopedPrometheusRegisterer(subsystem)
 }
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -84,9 +84,8 @@ func NewPrometheus(opts Options) *Prometheus {
 	opts = applyCompatibilityDefaults(opts)
 
 	p := &Prometheus{
-		registry:  opts.PrometheusRegistry,
-		opts:      opts,
-		namespace: namespace,
+		registry: opts.PrometheusRegistry,
+		opts:     opts,
 	}
 
 	if p.registry == nil {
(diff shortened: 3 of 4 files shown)


metrics/metrics.go
```diff
@@ -85,6 +85,14 @@ type Metrics interface {
 	Close()
 }
 
+// OpaMetrics is an optional interface that a Metrics implementation can implement. It is used
+// to expose a Prometheus registerer that applies namespace prefixing for use in Open Policy Agent instances
+// This can only credibly be implemented with Prometheus as a backend and can also be removed if Prometheus
+// is the only supported backend and Codahale support is dropped
+type OpaMetrics interface {
```
AlexanderYastrebov: I'd just do `PrometheusMetrics` if we go that way.
Ideally we should embrace OTEL metrics someday.

metrics/prometheus.go
```diff
@@ -555,6 +558,11 @@ func (p *Prometheus) DeleteInvalidRoute(routeId string) {
 
 func (p *Prometheus) Close() {}
 
+// Implements the OpaMetrics interface
+func (p *Prometheus) OpaScopedPrometheusRegisterer() prometheus.Registerer {
+	return prometheus.WrapRegistererWithPrefix(p.namespace+"_"+promOpenPolicyAgentSubsystem+"_", p.registry)
+}
```
AlexanderYastrebov: I'd prefer that OPA side takes care of using its own namespace

skipper.go
```diff
@@ -1941,6 +1941,7 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 			openpolicyagent.WithCleanInterval(o.OpenPolicyAgentCleanerInterval),
 			openpolicyagent.WithInstanceStartupTimeout(o.OpenPolicyAgentStartupTimeout),
 			openpolicyagent.WithTracer(tracer),
+			openpolicyagent.WithMetrics(mtr),
```
AlexanderYastrebov: Maybe just declare WithPrometheusMetrics and pass prometheus instance if Skipper was configured to use it, otherwise no OPA metrics :) Then we would not need any type discovery magic.

metrics/metrics.go
szuecs: I think OTel has to work on efficiency https://promlabs.com/blog/2025/07/17/why-i-recommend-native-prometheus-instrumentation-over-opentelemetry/ , so let's go with Prometheus.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -189,6 +192,13 @@ func WithControlLoopMaxJitter(maxJitter time.Duration) func(*OpenPolicyAgentRegi
 	}
 }
 
+func WithMetrics(metrics metrics.Metrics) func(*OpenPolicyAgentRegistry) error {
```
szuecs: We should here already make sure that we have PrometheusMetrics.
If not in the signature, then return error in case non-PrometheusMetrics are passed.

metrics/prometheus.go
szuecs: What about this example:
```
func (p *Prometheus) ScopedPrometheusRegisterer(scope string) prometheus.Registerer {
    return prometheus.WrapRegistererWithPrefix(p.namespace+"_"+scope+"_", p.registry)
}
```

We would not need to name things "Opa" and enable others to add their own $namespace_$scope

Maybe even:
```
func (p *Prometheus) ScopedPrometheusRegisterer(namespace, scope string) prometheus.Registerer {
    return prometheus.WrapRegistererWithPrefix(namespace+"_"+scope+"_", p.registry)
}
```

metrics/metrics.go
szuecs: @AlexanderYastrebov I am not sure if I do like more `OpaMetrics`

metrics/all_kind.go
```diff
@@ -3,6 +3,8 @@ package metrics
 import (
 	"net/http"
 	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
```
szuecs, line 7: I think this you do not need

metrics/all_kind.go
```diff
@@ -146,6 +148,11 @@ func (a *All) Close() {
 	a.prometheus.Close()
 }
 
+// Implements the OpaMetrics interface
+func (a *All) OpaScopedPrometheusRegisterer() prometheus.Registerer {
```
szuecs: if we make sure we only pass PrometheusMetrics then we can likely omit this.

metrics/metrics.go
szuecs: If we do https://github.com/zalando/skipper/pull/3631/files#r2372746228 then let's use `PrometheusMetrics` and likely we want to do:

```
type PrometheusMetrics interface {
    Metrics
    PrometheusRegisterer(namespace, scope string) prometheus.Registerer
}
```

skipper.go
szuecs: https://github.com/zalando/skipper/pull/3631/files#r2372761112

metrics/all_kind.go
mjungsbluth: I believe there is a corner case if you start Skipper with both Codahale and Prometheus as metrics flavours. In that case I would still like to support the Prometheus side more fully but I can verify that...

metrics/prometheus.go
mjungsbluth: I like that. I will make the change to see how it looks like.

I would however prefer the first method signature as the namespace can also be globally configured for skipper and it would probably be strange to suddenly have a different namespace (especially for library use of Skipper)

szuecs: Makes sense to me

metrics/all_kind.go
szuecs: ok

metrics/all_kind.go
mjungsbluth: This is actually where the `prometheus.Registerer` is defied so I guess this has to be here.

szuecs: ok

szuecs:
:+1:

mjungsbluth:
@torwunder could you also have a look?

filters/openpolicyagent/openpolicyagent_test.go
```diff
@@ -1078,6 +1097,42 @@ func TestBodyExtractionUnknownBody(t *testing.T) {
...
+	})
+	assert.NoError(t, err)
+
+	// Gather metrics and check for plugin_status_gauge
+	metricsFamilies, err := reg.Gather()
+	assert.NoError(t, err)
+
+	found := false
+	for _, mf := range metricsFamilies {
+		if mf.GetName() == "plugin_status_gauge" && len(mf.Metric) > 0 {
```
torwunder, line 1140: you could also verify the labels for the metrics as this is also part of the interface to the outside

mjungsbluth: Good catch

metrics/prometheus.go
```diff
@@ -555,6 +557,11 @@ func (p *Prometheus) DeleteInvalidRoute(routeId string) {
 
 func (p *Prometheus) Close() {}
 
+// Implements the PrometheusMetrics interface
```
szuecs: Godoc should start by the func name

skipper.go
```diff
@@ -1982,7 +1982,7 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 			opts = append(opts, openpolicyagent.WithEnvoyMetadataFile(o.OpenPolicyAgentEnvoyMetadata))
 		}
 
-		opaRegistry, err = openpolicyagent.NewOpenPolicyAgentRegistry(
+		registryOpts := []func(*openpolicyagent.OpenPolicyAgentRegistry) error{
```
szuecs: Maybe opaRegistryOpts

filters/openpolicyagent/openpolicyagent_test.go
szuecs: done

szuecs:
:+1:

mjungsbluth:
👍

PR #3634, opened 2025-09-22
fix: Add missing timing metrics in makeBackendRequest

ponimas:
The request and response timing metrics weren't captured in makeBackendRequest. 
Added a defer block to ensure stopwatches are stopped and elapsed times are recorded in the context.
A follow up to https://github.com/zalando/skipper/pull/3603


a4180p:
👍

PR #3639, opened 2025-09-26
feature: implement <forward>

szuecs:
feature: implement <forward>

For migration purposes it is useful that the operator can override the backend target. Users of routing tools like stackset https://github.com/zalando-incubator/stackset-controller can set the backend to something that the operators decide.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -128,7 +128,7 @@ type Config struct {
 	RoutesFile         string               `yaml:"routes-file"`
 	RoutesURLs         *listFlag            `yaml:"routes-urls"`
 	InlineRoutes       string               `yaml:"inline-routes"`
-	ForwardURL         string               `yaml:"forward-url"`
+	ForwardBackendURL  string               `yaml:"forward-backend-url"`
 	AppendFilters      *defaultFiltersFlags `yaml:"default-filters-append"`
 	PrependFilters     *defaultFiltersFlags `yaml:"default-filters-prepend"`
 	DisabledFilters    *listFlag            `yaml:"disabled-filters"`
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -445,6 +446,7 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.RoutesFile, "routes-file", "", "file containing route definitions")
 	flag.Var(cfg.RoutesURLs, "routes-urls", "comma separated URLs to route definitions in eskip format")
 	flag.StringVar(&cfg.InlineRoutes, "inline-routes", "", "inline routes in eskip format")
+	flag.StringVar(&cfg.ForwardURL, "forward-url", "", "target url of the <forward> backend")
```
a4180p: I would name the flag `forward-backend-url`

docs/reference/backends.md
```diff
@@ -246,6 +246,44 @@ target host defaults to either the `Host` header or the host name given in the
...
+## Forward backend
+
+The forward backend, `<forward>`, will set the backend to operators
+choice set by `-forward-url`.  This can be useful for data plane
+migrations. In one case we want to switch from one Kubernetes cluster
+to another Kubernetes cluster, but both cluster data planes can reach
+each other. The route with the `<forward>` will get cleaned all
+filters so there is no duplicate filter execution. This is useful
+because old and new clusters will have the same routing objects, that
+apply a set of routes with filters. Filters for example `modPath` can
```
a4180p:
```suggestion
apply a set of routes with filters. Filters, for example `modPath`, can
```

docs/reference/backends.md
```diff
@@ -246,6 +246,44 @@ target host defaults to either the `Host` header or the host name given in the
...
+create unexpected change to requests and responses passing through a
```
a4180p:
```suggestion
create an unexpected change to requests and responses passing through a
```

AlexanderYastrebov:
This is technically a network backend with the difference that user would not need to specify url, just <forward> placeholder that can be replaced by a single value. 
What do you think about adding a backend rewrite feature (similar to `-edit-route`, maybe even extend it) to rewrite backend? This way user would simply specify a predefined network backend (e.g. `"http://forward.cluster2"`) and operator can specify one or several rewrite rules to replace predefined backends to whatever is needed.

szuecs:
> This is technically a network backend with the difference that user would not need to specify url, just placeholder that can be replaced by a single value. What do you think about adding a backend rewrite feature (similar to `-edit-route`, maybe even extend it) to rewrite backend? This way user would simply specify a predefined network backend (e.g. `"http://forward.cluster2"`) and operator can specify one or several rewrite rules to replace predefined backends to whatever is needed.

Yes that's a network backend under the hood, but using `-edit-route` feels more like a hack for me and the code is small enough to have it more clearly specified. We will need integration in stackset and this will be likely useful for other future migrations and maybe other use cases, so I would prefer the PR as it is.

config/config.go

AlexanderYastrebov:
:+1:

a4180p:
👍

Issue #3649, opened 2025-10-07
Get rid of "ServiceBackend" in Kubernetes Routegroups


szuecs:
There is a hacked backend, that we sooner or later should get rid of:
```

// adding Kubernetes specific backend types here. To be discussed.
// The main reason is to differentiate between service and external, in a way
// where we can also use the current global option to decide whether the service
// should then be converted to LB. Or shall we expect the route group already
// contain the pod endpoints, and ignore the global option for skipper?
// --> As CRD we have to lookup endpoints ourselves, maybe via kube.go
const (
	ServiceBackend = eskip.LBBackend + 1 + iota
)
```

Normally in eskip we do not have it and when we introduced `<forward>` backend we stumbled in our cluster about the validation webhook that denied this backend because somehow `eskip.LBBackend + 1` mapped to `eskip.ForwardBackend`...

AlexanderYastrebov:
🙈 🤣 
```go
const ServiceBackend = -1
```

AlexanderYastrebov:
I guess we can add unexported `isService bool` here https://github.com/zalando/skipper/blob/589f1186cb035078a56253c4e241b15d28991605/dataclients/kubernetes/definitions/routegroups.go#L93
and return it from here https://github.com/zalando/skipper/blob/589f1186cb035078a56253c4e241b15d28991605/dataclients/kubernetes/definitions/routegroups.go#L288-L295

PR #3650, opened 2025-10-07
fix: forward backend in routegroups

szuecs:
fix: forward backend in routegroups

The validation webhook creates a ServiceBackend by adding +1 to eskip.LBBackend, which was ForwardBackend. 
It failed then by wanting to have a service_name which is not required nor makes sense in ForwardBackend. 

follow up ref https://github.com/zalando/skipper/issues/3649 
commit dd70bd65e7f99cfb5dd6b6f71885d9fe3b2707f6 broke the append for eskip backends.

### DIFF
--- a/dataclients/kubernetes/definitions/routegroups.go
+++ b/dataclients/kubernetes/definitions/routegroups.go
@@ -17,7 +17,7 @@ import (
 // contain the pod endpoints, and ignore the global option for skipper?
 // --> As CRD we have to lookup endpoints ourselves, maybe via kube.go
 const (
-	ServiceBackend = eskip.LBBackend + 1 + iota
+	ServiceBackend = -1
 )
 
 var (


AlexanderYastrebov:
I'd rather fix ServiceBackend hack by using negative or large value
https://github.com/zalando/skipper/blob/589f1186cb035078a56253c4e241b15d28991605/dataclients/kubernetes/definitions/routegroups.go#L19-L21

like e.g.:
```go
const ServiceBackend eskip.BackendType = -1
```

AlexanderYastrebov:
I am not even sure why we do need ServiceBackend and not use e.g. LBBackend in k8s dataclient.

AlexanderYastrebov:
> I am not even sure why we do need ServiceBackend and not use e.g. LBBackend in k8s dataclient.

We use it here :(
https://github.com/zalando/skipper/blob/589f1186cb035078a56253c4e241b15d28991605/dataclients/kubernetes/definitions/routegroupvalidator.go#L240-L242

szuecs:
> I am not even sure why we do need ServiceBackend and not use e.g. LBBackend in k8s dataclient.

We use it such that validation webhook knows about backend and can validate it. routegroup `type: service` backend only exists in kubernetes and not in eskip, because there is no use in the data plane of this. It's only to pass relevant data around and validation webhook does not evaluate the backend that is created, but only the kubernetes object.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #3652, opened 2025-10-08
build: use latest version tag

AlexanderYastrebov:
To derive new release version from the latest version build script describes git tags.

If build pipeline was retriggered we end up with two version tags on the HEAD commit which breaks subsequent builds because git describe tags returns first tag from several (see https://stackoverflow.com/questions/8089002/git-describe-with-two-tags-on-the-same-commit) and then build script can not create new tag because it already exists (i.e. second tag on the commit).

This change lists tags differently such that build script always sees latest version tag.


delivery.yaml
```diff
@@ -33,7 +33,7 @@ pipeline:
       MULTIARCH_REGISTRY="container-registry-test.zalando.net"
 
       if [[ $CDP_TARGET_BRANCH == master && ! $CDP_PULL_REQUEST_NUMBER ]]; then
-        LATEST_VERSION=$(git describe --tags --always | awk -F \- '{print $1}')
+        LATEST_VERSION=$(git tag -l | grep 'v[0-9]\+\.[0-9]\+\.[0-9]\+$' | sort -V | tail -1)
```
AlexanderYastrebov, line 36:
```console
$ (echo v0.22.11 ; echo foo ; echo v123 ; echo v0.22.1) | grep 'v[0-9]\+\.[0-9]\+\.[0-9]\+$' | sort -V
v0.22.1
v0.22.11
```

AlexanderYastrebov:
:+1:

MustafaSaber:
👍

Issue #3655, opened 2025-10-09
forward backend creates broken routes in routesrv


szuecs:
The backend `<forward>` or the replacement is missing in routesrv /routes if `- -forward-backend-url=http://skipper-...example` is enabled or not does not matter.

```
% curl http://localhost:9990/routes G migra G -E '[-][>] ;'
kube_rg__default__test_migration_app_v2__all__0_0: PathSubtree("/") && Host("^(test-migration-app-v2[.]example[.]?(:[0-9]+)?)$") -> disableAccessLog(2, 3, 404, 429) -> fifo(2000, 20, "1s") -> stateBagToTag("auth-user", "client.uid") -> ;
kube_rg__default__test_migration_app_v2_traffic_segment__all__0_0: PathSubtree("/") && Host("^(test-migration[.]example[.]?(:[0-9]+)?)$") && TrafficSegment(0.5, 1) -> disableAccessLog(2, 3, 404, 429) -> fifo(2000, 20, "1s") -> stateBagToTag("auth-user", "client.uid") -> ;
```

PR #3656, opened 2025-10-09
fix: forward backend integration in kubernetes dataclient and routesrv

szuecs:
fix: forward backend integration in kubernetes dataclient and routesrv

ref: https://github.com/zalando/skipper/issues/3655


dataclients/kubernetes/routegroups_test.go
```diff
@@ -67,3 +67,7 @@ func TestRouteGroupAnnotationConfig(t *testing.T) {
 		"testdata/routegroups/annotation-predicates",
 		"testdata/routegroups/annotation-filters")
 }
+
+func TestRouteGroupBackends(t *testing.T) {
+	kubernetestest.FixturesToTest(t, "testdata/routegroups/backends")
+}
```
AlexanderYastrebov, line 73: Did we forget this previously? I do not see new fixtures added...

routesrv/testdata/forward-backend.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: my-test
+spec:
+  backends:
+  - name: fwd
+    type: forward
+  routes:
+    - pathSubtree: /
+      backends:
+        - backendName: fwd
```
AlexanderYastrebov: New line is missing

dataclients/kubernetes/routegroups_test.go
szuecs: No I forgot to add them, fixed

routesrv/testdata/forward-backend.yaml
szuecs: Why do we need this?

szuecs: fixed

AlexanderYastrebov: There is another one missing as well.

AlexanderYastrebov:
:+1:

routesrv/testdata/forward-backend.yaml
AlexanderYastrebov: It is a nit pick, yes, but see discussion https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

I think it is beneficial to not have any warnings about missing new line from github or to be able to `cat foo.go bar.go` without melding them.

szuecs:
:+1:

PR #3627, opened 2025-09-15
Support OpenTelemetry SDK

AlexanderYastrebov:
When OpenTelemetry is enabled Skipper registers global OpenTelemetry TracerProvider and creates a bridge OpenTracing tracer.

This follows recommended migration strategy https://opentelemetry.io/docs/migration/opentracing/#migration-overview

OpenTelemetry Init reads configuration options from standard environment variables and allows additional options as a command line flag in yaml format.

Updates #2104

PR #3657, opened 2025-10-09
routesrv: support OpenTelemetry SDK

AlexanderYastrebov:
When OpenTelemetry is enabled routesrv registers global OpenTelemetry TracerProvider and creates a bridge OpenTracing tracer similar to Skipper, see #3627

Updates #2104

### DIFF
--- a/routesrv/routesrv.go
+++ b/routesrv/routesrv.go
@@ -10,6 +10,13 @@ import (
 	"syscall"
 	"time"
 
+	ot "github.com/opentracing/opentracing-go"
+	sotel "github.com/zalando/skipper/otel"
+	"github.com/zalando/skipper/tracing"
+	"go.opentelemetry.io/otel"
+	otBridge "go.opentelemetry.io/otel/bridge/opentracing"
+	"go.opentelemetry.io/otel/trace"
+
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
 
@@ -17,17 +24,19 @@ import (
 	"github.com/zalando/skipper/dataclients/kubernetes"
 	"github.com/zalando/skipper/filters/auth"
 	"github.com/zalando/skipper/metrics"
-	"github.com/zalando/skipper/tracing"
 )
 
 // RouteServer is used to serve eskip-formatted routes,
 // that originate from the polled data source.
 type RouteServer struct {
-	server        *http.Server
-	supportServer *http.Server
-	poller        *poller
-	wg            *sync.WaitGroup
+	server         *http.Server
+	supportServer  *http.Server
+	poller         *poller
+	wg             *sync.WaitGroup
+	tracerShutdown func(context.Context) error
 }
+
+const otelTracerName = "routesrv"
 
 // New returns an initialized route server according to the passed options.
 // This call does not start data source updates automatically. Kept routes
@@ -54,14 +63,11 @@ func New(opts skipper.Options) (*RouteServer, error) {
 
 	rs := &RouteServer{}
 
-	opentracingOpts := opts.OpenTracing
-	if len(opentracingOpts) == 0 {
-		opentracingOpts = []string{"noop"}
-	}
-	tracer, err := tracing.InitTracer(opentracingOpts)
+	tracer, shutdown, err := tracerInstance(&opts)
 	if err != nil {
 		return nil, err
 	}
+	rs.tracerShutdown = shutdown
 
 	b := &eskipBytes{
 		tracer:  tracer,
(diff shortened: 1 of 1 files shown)


madumalt:
👍

a4180p:
👍

routesrv/routesrv.go
```diff
@@ -142,6 +148,52 @@ func New(opts skipper.Options) (*RouteServer, error) {
...
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to setup OpenTracing: %w", err)
+			}
+		}
+		// Could be a noop tracer or a wrapper if library user configured OpenTracing bridge tracer
+		otelTracer = otel.Tracer(otelTracerName)
+		shutdown = func(context.Context) error { return nil }
+	}
+
+	_ = otelTracer // unused for now
```
MustafaSaber, line 192: Just to confirm, this is unused because we use the OpenTracing tracer or OpenTracing bridge tracer and never OpenTelemetry?

routesrv/routesrv.go
```diff
@@ -142,6 +148,52 @@ func New(opts skipper.Options) (*RouteServer, error) {
...
+
+		// Setup OpenTracing bridge tracer
+		bridgeTracer, wrapperTracerProvider := otBridge.NewTracerPair(otel.Tracer(otelTracerName))
+
+		bridgeTracer.SetWarningHandler(func(msg string) { log.Warnf("OpenTracing bridge warning: %s", msg) })
+		otel.SetTracerProvider(wrapperTracerProvider)
+
+		tracer = bridgeTracer
+		// Obtain tracer from wrappedTracerProvider
+		otelTracer = otel.Tracer(otelTracerName)
```
MustafaSaber, line 173: The comment suggests that this should be
```go
otelTracer = wrapperTracerProvider.Tracer(otelTracerName)
```

a4180p: It's set as global in line 169.

MustafaSaber: ~~I think because `WrapperTracerProvider` is deprecated we use `(provide).Tracer` directly in this case~~ Thanks @ponimas!

MustafaSaber: Should we then use something else other that `WrapperTracerProvider` since it's deprecated?

routesrv/routesrv.go
AlexanderYastrebov: yes, its a placeholder. Once we ditch OpenTracing we'll use it.

routesrv/routesrv.go
AlexanderYastrebov: The type WrapperTracerProvider is deprecated but factory method https://pkg.go.dev/go.opentelemetry.io/otel/bridge/opentracing#NewTracerPair is not. Since we do not use any methods/fields of the instance I think it should be fine.

The whole bridge dance should go away once we get rid of OpenTracing.

MustafaSaber: Isn't `.Tracer` considered somehow method of `WrapperTracerProvider` in this case? We just call it from otel.GlobalTracer

AlexanderYastrebov: `Tracer()` is the method of `trace.TracerProvider` interface, `WrapperTracerProvider` implements it but we don't use anything `WrapperTracerProvider` specific.
I agree it is confusing that they deprecated `WrapperTracerProvider`  but not `otBridge.NewTracerPair` which was suggested in https://github.com/open-telemetry/opentelemetry-go/pull/3116

AlexanderYastrebov: Here is a related PR that uses `otBridge.NewTracerProvider` https://github.com/sourcegraph/sourcegraph-public-snapshot/pull/40945

AlexanderYastrebov: At the end the only difference seems to be in the `instrumentation.name` attribute - the other setup method allows using different tracer name for spans created via bridge tracer like:
```go
bridgeTracer.SetOpenTelemetryTracer(bridgeTracerProvider.Tracer(otelTracerName + ".opentracing-bridge"))
```
while `NewTracerPair` [re-uses](https://github.com/open-telemetry/opentelemetry-go/blob/5c78f7c833a08fdd6c96eabeed006f4b86728920/bridge/opentracing/util.go#L21) otel tracer name:
```go
bridgeTracer.SetOpenTelemetryTracer(wrapperProvider.Tracer(""))
```

AlexanderYastrebov: FWIW official docs https://github.com/open-telemetry/opentelemetry-go/tree/main/bridge/opentracing stuugest to use NewTracerPair

MustafaSaber: I have better understanding now thanks!

Issue #3663, opened 2025-10-14
OpenTelemetry: support AWS X-Ray Propagator


universam1:
**Is your feature request related to a problem? Please describe.**
OTEL support arrived https://github.com/zalando/skipper/pull/3627 (thank you @AlexanderYastrebov !) but misses support for AWS X-Ray Propagator. 
The `OTEL_PROPAGATORS=xray` is currently ignored, the ALB's `X-Amzn-Trace-Id` headers are not extracted by Skipper, so Traces effectively "start" at Skipper level, not at ALB.

**Describe the solution you would like**
Adding support for https://pkg.go.dev/go.opentelemetry.io/contrib/propagators/aws/xray propagator

**Would you like to work on it?**
Yes - with guidance

szuecs:
I would say just try to add it

AlexanderYastrebov:
Hi @universam1, thanks for opening an issue. Please note that OpenTelemetry support was just added and we'll likely iron out some details down the line.

> I would say just try to add it

I think it should be as easy as https://github.com/zalando/skipper/pull/3664

universam1:
@AlexanderYastrebov I thought so, but this is 🚀 amazing fast!

AlexanderYastrebov:
Hm, I tried to test it and discovered that it is actually registered here https://github.com/open-telemetry/opentelemetry-go-contrib/blob/613f80e3ffd5cb8de28fcce63fa7dfdf9645bc10/propagators/autoprop/registry.go#L38-L39

AlexanderYastrebov:
Skipper won't start with an unsupported propagator (e.g. `foo`):
```
...
[APP]FATA[0000] failed to setup OpenTelemetry: unknown propagator: foo
```
and it starts with `xray` so I think something else is missing.

AlexanderYastrebov:
> The OTEL_PROPAGATORS=xray is currently ignored,

Could you provide a flag value (and env variables for that matter) of how you init OTEL?
Just in case, even if you want to use env variables to configure, you still need to tell Skipper to use OTEL via:
```
skipper -opentelemetry={}
```
see https://github.com/zalando/skipper/blob/master/docs/operation/operation.md#opentelemetry

universam1:
> discovered that it is actually registered here

OK thank you for this insight, I missed that!

> ```
> skipper -opentelemetry={}
> ```
> 
> see https://github.com/zalando/skipper/blob/master/docs/operation/operation.md#opentelemetry

[...]

PR #3664, opened 2025-10-14
otel: register AWS X-Ray Propagator

AlexanderYastrebov:
Move registration of auto exporters and propagators into init function
that it called iff otel.Init was called.

Updates #3663

### DIFF
--- a/otel/otel.go
+++ b/otel/otel.go
@@ -8,10 +8,12 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"sync"
 	"time"
 
 	"go.opentelemetry.io/contrib/exporters/autoexport"
 	"go.opentelemetry.io/contrib/propagators/autoprop"
+	"go.opentelemetry.io/contrib/propagators/aws/xray"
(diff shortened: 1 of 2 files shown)


otel/otel.go
```diff
@@ -53,6 +55,12 @@ type BatchSpanProcessor struct {
 	MaxExportBatchSize int           `yaml:"maxExportBatchSize"`
 }
 
+// initOnce makes sure that the package-level initialization is done only once if [Init] was called.
+var initOnce = sync.OnceFunc(func() {
+	autoexport.RegisterSpanExporter("skipper-debug", skipperDebugSpanExporter)
+	autoprop.RegisterTextMapPropagator("xray", xray.Propagator{})
```
AlexanderYastrebov, line 61: It is actually already registered here https://github.com/open-telemetry/opentelemetry-go-contrib/blob/613f80e3ffd5cb8de28fcce63fa7dfdf9645bc10/propagators/autoprop/registry.go#L38-L39

PR #3665, opened 2025-10-14
feature: forward backend for ingress as annotation zalando.org/skipper-backend

szuecs:
feature: [forward backend](https://opensource.zalando.com/skipper/reference/backends/#forward-backend) for ingress as annotation `zalando.org/skipper-backend`.

RouteGroup validation makes sure that the backend type is recognized.
dataclients/kubernetes/definitions/testdata/validation/test-forward.log is an empty file, because if log file does not exist test would be skipped.


dataclients/kubernetes/ingressv1.go
```diff
@@ -446,9 +448,11 @@ func (ing *ingress) ingressV1Route(
 	var route *eskip.Route
 	if r, ok, err := ing.convertDefaultBackendV1(ic, ing.forceKubernetesService); ok {
 		route = r
+		ic.applyBackend(route)
 	} else if err != nil {
 		ic.logger.Errorf("Failed to convert default backend: %v", err)
 	}
+
```
szuecs, line 455: for readability, because `route` is not touched anymore until "return".

dataclients/kubernetes/ingressv1.go
```diff
@@ -433,6 +433,8 @@ func (ing *ingress) ingressV1Route(
 		logger:              logger,
 		annotationFilters:   annotationFilter(i.Metadata, logger),
 		annotationPredicate: annotationPredicate(i.Metadata),
+		annotationBackend:   annotationBackendString(i.Metadata),
```
szuecs, line 436: here we need annotationBackendString() to make it more clean

szuecs:
:+1:

MustafaSaber:
I think we can add another test case for filter by annotation & predicate by annotation (feature where we add filter if some annotation exists)

Also we need to extend validation webhook to only allow forward value for now.

szuecs:
> I think we can add another test case for filter by annotation & predicate by annotation (feature where we add filter if some annotation exists)

These test cases are in the PR. Maybe I miss understand your comment.

> Also we need to extend validation webhook to only allow forward value for now.

I don't see the need. This should not be used by ordinary users anyways. 
I plan to use this only as tool to support stackset by orchestrating it https://github.com/zalando-incubator/stackset-controller/pull/716

PR #3668, opened 2025-10-15
otel/xxray: add custom X-Ray propagator that preserves X-Amzn-Trace-Id

AlexanderYastrebov:
Updates #3663

### DIFF
--- a/otel/otel.go
+++ b/otel/otel.go
@@ -8,10 +8,13 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"slices"
+	"strings"
 	"time"
 
 	"go.opentelemetry.io/contrib/exporters/autoexport"
 	"go.opentelemetry.io/contrib/propagators/autoprop"
+	"go.opentelemetry.io/contrib/propagators/aws/xray"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
@@ -23,6 +26,7 @@ import (
 
 	"github.com/bombsimon/logrusr/v4"
 	"github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/otel/xxray"
 )
 
 var log = logrus.WithField("package", "otel")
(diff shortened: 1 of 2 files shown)


otel/otel.go
```diff
@@ -53,6 +57,10 @@ type BatchSpanProcessor struct {
 	MaxExportBatchSize int           `yaml:"maxExportBatchSize"`
 }
 
+func init() {
+	autoprop.RegisterTextMapPropagator("xxray", xxray.NewPropagatorGenerator())
```
AlexanderYastrebov: We can also add a config field e.g .`PropagatorXRayReuseTraceID` instead of a new name and create custom implementation if it is set.

otel/xxray/propagator.go
```diff
@@ -0,0 +1,159 @@
...
+			part = strings.TrimSpace(headerVal[pos:])
+			pos = len(headerVal)
+		}
+		equalsIndex := strings.Index(part, kvDelimiter)
+		if equalsIndex < 0 {
+			return empty, errInvalidTraceHeader
+		}
+		value := part[equalsIndex+1:]
+		switch {
+		case strings.HasPrefix(part, traceIDKey):
```
AlexanderYastrebov, line 103: This is ugly and also allows `RootFooBar=123` key 🤷 . Parsing should be done via strings.Cut or similar to https://github.com/zalando/skipper/pull/3405

universam1:
maybe we can add these samples in the test: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html

otel/xxray/propagator.go
```diff
@@ -0,0 +1,159 @@
+package xxray
```
szuecs, line 1: Why not xray?

universam1:
Tested this in lab and it is promising, the ALB trace-id is forwarded!

example headers forwarded: 

```
"traceparent": "00-68ff8fe86949c39c171cef723838f6ed-2b5fdb1944ad4911-00",
"x-amzn-trace-id": "Root=1-68ff8fe8-6949c39c171cef723838f6ed;Parent=2b5fdb1944ad4911;Sampled=0",
```

Just it looks like the `flow-id` log output is not covered here yet, would it not be correct to fill it at this place as well?
@szuecs @AlexanderYastrebov 
```
skipper-ingress-678759b679-6kvxp {"audit":"","auth-user":"","duration":1,"flow-id":"",...
```

szuecs:
@universam1 can't you use https://opensource.zalando.com/skipper/reference/filters/#tracingtag ?
`tracingTag("http.flow_id", "${request.header.X-Flow-Id}")`

PR #3672, opened 2025-10-16
validation: run skipper in validation mode inside webhook

MustafaSaber:
We leave webhook binary for backward combitability but use skipper in validation mode everywhere as source of truth

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -1,19 +1,12 @@
 package main
 
 import (
-	"context"
 	"flag"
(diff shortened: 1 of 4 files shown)


szuecs:
Even if tests would all pass this will break current installed webhooks if they would update to this version.
For example default port is different.

MustafaSaber:
> Even if tests would all pass this will break current installed webhooks if they would update to this version. For example default port is different.

hmm, why? Default port is the same

PR #3682, opened 2025-10-27
refactor: if..else if to switch

szuecs:
refactor: rename var such that it makes sense
refactor: if..else if to switch


szuecs:
:+1:

MustafaSaber:
👍

PR #3685, opened 2025-10-30
add: tee()/teenf() routing loop detection

szuecs:
add: tee()/teenf() routing loop detection

PR #3686, opened 2025-10-31
feature: set tee loop detection header value to something that makes sense

szuecs:
feature: set tee loop detection header value to something that makes sense

follow up
ref: https://github.com/zalando/skipper/pull/3685


a4180p:
👍

MustafaSaber:
👍

PR #3691, opened 2025-11-03
fix: net.Client Timeout and slow shadow route error cases

szuecs:
fix: net.Client to use Timeout as we pass

test: 3 different cases 
1. slow body reader 
2. slow body writer 
3. slow accept listener (test required the Timeout to be fixed in net.Client)

For test cases I created a couple of test packages to support testing slow read(), write(), accept(), which can be helpful in other tests, too.

### DIFF
--- a/net/nettest/listener.go
+++ b/net/nettest/listener.go
@@ -72,8 +72,6 @@ func (lo *SlowAcceptListener) Close() error {
 	return nil
 }
 
-var _ net.Listener = &SlowAcceptListener{}
-
 func NewSlowAcceptListener(opt *SlowAcceptListenerOptions) (*SlowAcceptListener, error) {
 	lo := &SlowAcceptListener{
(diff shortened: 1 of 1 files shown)


net/nettest/listener.go
```diff
@@ -0,0 +1,93 @@
...
+func (lo *SlowAcceptListener) Close() error {
+	if lo.l != nil {
+		lo.l.Close()
+	}
+	lo.once.Do(func() { close(lo.quit) })
+
+	return nil
+}
+
+var _ net.Listener = &SlowAcceptListener{}
```
MustafaSaber: This is repeated at line 32

szuecs: fixed

MustafaSaber:
👍

szuecs:
:+1:

PR #3693, opened 2025-11-04
chore: Update Go version to 1.24.9

ponimas:
1.24.6 has some vulnerabilities:

https://osv.dev/GO-2025-4006
https://osv.dev/GO-2025-4007
https://osv.dev/GO-2025-4008
https://osv.dev/GO-2025-4009
https://osv.dev/GO-2025-4010
https://osv.dev/GO-2025-4011
https://osv.dev/GO-2025-4012
https://osv.dev/GO-2025-4013
https://osv.dev/GO-2025-4014

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -207,4 +207,4 @@ require (
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
-go 1.24.6
+go 1.24.9


a4180p:
👍

madumalt:
👍

szuecs:
I wonder why we need this change.
The version in go.mod means only that new features from that version are available for skipper code...., not that the binary was built by that version.

a4180p:
@szuecs I think that the version in go.mod means also that it's impossible to build binary with version less than required.

szuecs:
Yes that's true, because of the feature set is only available from that version ongoing.
For example you get in 1.x.y net.SuperListener, then if you are <1.x.y there is no net.SuperListener, but >=1.x.y has it.

However if you are in 1.x1.y1 with x1<=x and y1<y and you do not need net.SuperListener, you can also use the latest version to compile but your gopls and other tooling will be able to tell you that net.SuperListener, is not available.

a4180p:
I understand that. But osv-scanner do not :)

szuecs:
I wonder what we can do.
Maybe we can use 1.24 and it will use the latest patch version or something like that automatically.

a4180p:
That will work for osv-scanner, i guess.
But any run of the `go mod tidy` will transform it into `1.24.0` and add a toolchain with your local go version.

szuecs:
I guess then it would be good to have dependabot update this part, too if possible

Issue #3683, opened 2025-10-30
Handle background task queue errors at pre-loading OPA Instances


Pushpalanka:
**Describe the bug**
The background task queue has a default size of 100 and in order to exhaust this queue there has to be 100 unique bundles existing in the queue. This exhausted error message has been seen while only two unique bundles were available in the route.
So the error message is misleading and other errors are possible to happen while a task is attempted to be added in the queue.

<details>

<summary>Log trace</summary>

level=info msg="Scheduling background task for new OPA instance for bundle: in-application-authorization-demo"
level=info msg="Scheduling background task to start existing OPA instance for bundle: in-application-authorization-demo"
level=info msg="**Scheduling background task to start existing OPA instance for bundle: in-application-authorization-demo**"
level=info msg="Open Policy Agent instance is not ready yet, returning unavailable" bundle-name=in-application-authorization-demo
"GET /dashboards HTTP/1.1" 503 0 "-" "Go-http-client/1.1" 0 in-application-authorization-demo.corporate-iam-test.example.com - -
level=info msg="Scheduling background task to start existing OPA instance for bundle: in-application-authorization-demo"
level=info msg="Open Policy Agent instance is not ready yet, returning unavailable" bundle-name=in-application-authorization-demo
"GET /dashboards HTTP/1.1" 503 0 "-" "Go-http-client/1.1" 0 in-application-authorization-demo.corporate-iam-test.example.com- -
level=info msg="Open Policy Agent instance is not ready yet, returning unavailable" bundle-name=in-application-authorization-demo
"GET /dashboards HTTP/1.1" 503 0 "-" "Go-http-client/1.1" 1 in-application-authorization-demo.corporate-iam-test.example.com - -
level=info msg="Open Policy Agent instance is not ready yet, returning unavailable" bundle-name=in-application-authorization-demo
"GET /dashboards HTTP/1.1" 503 0 "-" "Go-http-client/1.1" 0 in-application-authorization-demo.corporate-iam-test.example.com - -
level=info msg="Scheduling background task to start existing OPA instance for bundle: in-application-authorization-demo"
level=info msg="Scheduling background task to start existing OPA instance for bundle: in-application-authorization-demo"
level=info msg="Open Policy Agent instance is not ready yet, returning unavailable" bundle-name=in-application-authorization-demo
"GET /dashboards HTTP/1.1" 503 0 "-" "Go-http-client/1.1" 0 in-application-authorization-demo.corporate-iam-test.example.com - -
[...]

PR #3694, opened 2025-11-05
fix: duplicate processing of bundle downloads

szuecs:
fix https://github.com/zalando/skipper/issues/3683

### DIFF
--- a/filters/openpolicyagent/preprocessor.go
+++ b/filters/openpolicyagent/preprocessor.go
@@ -15,8 +15,6 @@ type opaPreProcessor struct {
 	mu       sync.Mutex
 	registry *OpenPolicyAgentRegistry
 
-	bundleMap map[string]struct{}
-
 	log logging.Logger
 }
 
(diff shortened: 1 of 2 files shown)


filters/openpolicyagent/preprocessor.go
```diff
@@ -41,9 +44,18 @@ func (p *opaPreProcessor) Do(routes []*eskip.Route) []*eskip.Route {
 		p.preloadInstancesParallel(bundleConfigs)
 	})
 
+	// check already processed bundles
+	bundles := make([]string, 0, len(bundleConfigs))
+	for _, bundleName := range bundleConfigs {
+		if _, ok := p.bundleMap[bundleName]; !ok {
+			bundles = append(bundles, bundleName)
+			p.bundleMap[bundleName] = struct{}{}
```
Pushpalanka: This looked good me until I check the test failure.
We currently have a retry possibility for tasks which failed to get added to the background queue when it is full(those are retried with the next route update), which is lost with this. Except for this I think the behavior is synonyms for sync.Once. 
Given these impacts I think sync.Once is simpler approach and the original issue I reported was misleading.

szuecs:
I had a pairprogramming session with @Pushpalanka and we came up with this solution to fix https://github.com/zalando/skipper/issues/3683

szuecs:
closing in favor of https://github.com/zalando/skipper/pull/3701

PR #3701, opened 2025-11-06
Avoid duplicate start tasks from getting into the queue.

Pushpalanka:
Avoid the same bundle start tasks been added to the queue twice.
co-author: @mjungsbluth


szuecs:
:+1:

Pushpalanka:
👍

PR #3703, opened 2025-11-06
fix: thread-key step reference in workflow

ponimas:
The workflow was incorrectly referencing 'step.thread-key.KEY' instead of 'steps.thread-key.KEY' (missing the 's' in 'steps').


a4180p:
👍

szuecs:
:+1:

PR #3704, opened 2025-11-07
fix: threaded notifications

ponimas:
This commit fixes the reference to the output of the step. 
Also moving the check-race step further in pipeline, it is cheaper to fail earlier in case if vulnerability detected.
The query argument moved to the secret itself.

Screenshot 2025-11-07 at 13 38 52


a4180p:
👍

shyamz-22:
👍🏽

Issue #2713, opened 2023-11-01
flakytest TestLoopbackAndMatchPredicate


szuecs:
```
--- FAIL: TestLoopbackAndMatchPredicate (0.02s)
    teeloopback_test.go:64: teeloopback: expected to receive 1 requests in split and shadow backend but got Split: 1, Shadow: 0
```

rMaxiQp:
Hi, I tried to add some additional logging to understand the current behavior. I used `go test -count 100000 -timeout 20m -run TestLoopbackAndMatchPredicate -failfast ./filters/tee/...` to stress test it and believe it might be a timing issue. I use `fmt.Printf` to print out more information. 

For a success run, the log looks like:
```go
===TestLoopbackAndMatchPredicate begin===
traffic:Create args: []interface {}{1}
traffic:Create args: []interface {}{1}
tee:Create args: []interface {}{"A"}
Proxy: do url=/foo
traffic:takeChance chance 1.000000 match true
tee:Match r.Key  vs predicate.Key A
traffic:takeChance chance 1.000000 match true
context:loopbackInternal discardResponse: true
Proxy: do url=//127.0.0.1:33722/foo
traffic:takeChance chance 1.000000 match true
tee:Match r.Key A vs predicate.Key A
context:loopbackInternal finishing proxySpan
context:loopbackInternal perr: <nil> ok: false 
===TestLoopbackAndMatchPredicate completed===
```

For a failed run, the log looks like:
```go
===TestLoopbackAndMatchPredicate begin===
traffic:Create args: []interface {}{1}
traffic:Create args: []interface {}{1}
tee:Create args: []interface {}{"A"}
Proxy: do url=/foo
traffic:takeChance chance 1.000000 match true
tee:Match r.Key  vs predicate.Key A
traffic:takeChance chance 1.000000 match true
context:loopbackInternal discardResponse: true
Proxy: do url=//127.0.0.1:44984/foo
traffic:takeChance chance 1.000000 match true
tee:Match r.Key A vs predicate.Key A
===TestLoopbackAndMatchPredicate completed===

--- FAIL: TestLoopbackAndMatchPredicate (0.02s)
    teeloopback_test.go:48: original: http://127.0.0.1:35251, split: http://127.0.0.1:37630, shadow: http://127.0.0.1:38552
    teeloopback.go:53: teeloopback: processing request GET /foo
    teeloopback.go:62: teeloopback: issuing loopback request GET //127.0.0.1:44984/foo
[...]

PR #3715, opened 2025-11-10
fix: `TestLoopbackAndMatchPredicate` waitFor requests delivery

rMaxiQp:
fix https://github.com/zalando/skipper/issues/2713

In case of a slow network or limited network resource, the server may shutdown before the Loopback() completes the roundtrip. Here, we use Eventually() to wait for both server resolve the request (or fail assertion on timeout) before continuing.


szuecs:
:+1:

MustafaSaber:
👍

PR #3722, opened 2025-11-10
feat: add support for disabling route polling

MustafaSaber:
- Add ability to disable route polling by setting source-poll-timeout to 0
- Return HTTP 204 No Content when polling is disabled

This will help with running `skipper`binary in webhook mode as we need to use `-kubernetes` flag to enable some filter registries but we don't want skipper to overload/pull data from k8s apiserver

- [ ] Add more tests

### DIFF
--- a/routing/routing.go
+++ b/routing/routing.go
@@ -278,7 +278,7 @@ func New(o Options) *Routing {
 	r := &Routing{log: o.Log, firstLoad: make(chan struct{}), quit: make(chan struct{})}
 	r.metrics = o.Metrics
 
-	if o.PollTimeout == 0 {
+	if o.PollTimeout == -1 {
 		o.Log.Info("Polling routetable is ignored as timeout is set to 0")
 		r.noPolls = true
 		return r
--- a/skipper.go
+++ b/skipper.go
@@ -2045,7 +2045,7 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	}
 
 	// ensure a non-zero poll timeout
-	if o.SourcePollTimeout < 0 {
+	if o.SourcePollTimeout != -1 && o.SourcePollTimeout <= 0 {
 		o.SourcePollTimeout = defaultSourcePollTimeout
 	}
 
--- a/scheduler/proxy_test.go
+++ b/scheduler/proxy_test.go
@@ -37,7 +37,7 @@ func TestFifoChanges(t *testing.T) {
 	spec := filterScheduler.NewFifo()
 	fr := builtin.MakeRegistry()
 	fr.Register(spec)
-	args := []interface{}{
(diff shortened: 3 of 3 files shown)


szuecs:
I think we should not use 0, but maybe -1

skipper.go
```diff
@@ -2046,7 +2045,7 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	}
 
 	// ensure a non-zero poll timeout
-	if o.SourcePollTimeout <= 0 {
+	if o.SourcePollTimeout < 0 {
```
szuecs: This will change defaults

config/config.go
```diff
@@ -460,7 +460,7 @@ func NewConfig() *Config {
 	flag.Var(cfg.RoutesURLs, "routes-urls", "comma separated URLs to route definitions in eskip format")
 	flag.StringVar(&cfg.InlineRoutes, "inline-routes", "", "inline routes in eskip format")
 	flag.StringVar(&cfg.ForwardBackendURL, "forward-backend-url", "", "target url of the <forward> backend")
-	flag.Int64Var(&cfg.SourcePollTimeout, "source-poll-timeout", int64(3000), "polling timeout of the routing data sources, in milliseconds")
+	flag.Int64Var(&cfg.SourcePollTimeout, "source-poll-timeout", int64(3000), "polling timeout of the routing data sources, in milliseconds. Use 0 to not poll routes.")
```
szuecs, line 463: This is not enough to create a default other than 0

routing/routing.go
```diff
@@ -276,6 +277,13 @@ func New(o Options) *Routing {
 
 	r := &Routing{log: o.Log, firstLoad: make(chan struct{}), quit: make(chan struct{})}
 	r.metrics = o.Metrics
+
+	if o.PollTimeout == 0 {
```
vlktna: I think we also need to check the ValidationWebhookEnabled flag to prevent misconfiguration. So if ValidationWebhookEnabled is true or SourcePollTimeout is less than 0, we disable polling

skipper.go
MustafaSaber: The default is already int64(3000) it's what we use now. We only update to use the default in case of invalid input like 0 or negative

config/config.go
MustafaSaber: Why not enough?

routing/routing.go
MustafaSaber: I rather not, it just a small flag and maybe there is a case where someone needs to actually have it as webhook and actually poll routes for some reason. This makes it generic and not associated to validation webhook, which I like more TBH.

MustafaSaber:
close in favor of https://github.com/zalando/skipper/pull/3729

PR #3725, opened 2025-11-10
Fix: flakytest admission control

szuecs:
Fix: flakytest admission control

The core problem was that we have had a non-seeded global random source that generated every time different random numbers so it's easy to have bad luck. Another issue was that if you have a global random source is that the different test cases interfere with each other. The third problem was that we used the same global random to have randomized response codes on the server side in the httptest.Server.
All of it is now fixed. 

We have an option to use a static seeded random source by filter that we use for tests.
In general we use the global random source that we have to guard by a mutex at least `test -race` says so.
Maybe we should also create filter local random source for the production build but I think it's fine like this.

### DIFF
--- a/filters/shedder/admission.go
+++ b/filters/shedder/admission.go
@@ -335,7 +335,6 @@ func (spec *AdmissionControlSpec) CreateFilter(args []interface{}) (filters.Filt
 		counter:          new(atomic.Int64),
 		successCounter:   new(atomic.Int64),
 		rand:             r,
-		//rand:             randWithSeed(), // flakytest
 	}
 	go ac.tickWindows(d)
 	return ac, nil
--- a/filters/shedder/admission_test.go
+++ b/filters/shedder/admission_test.go
@@ -27,24 +27,6 @@ import (
 	"golang.org/x/time/rate"
 )
 
-func TestRand(t *testing.T) {
-	r := randWithSeed()
-
-	for _, ti := range []struct {
-		msg string
-	}{
-		{msg: "1"},
-		{msg: "2"},
-	} {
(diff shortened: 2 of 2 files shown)


szuecs:
another flakytest we already know
```
--- FAIL: TestRedisAddrUpdater (1.26s)
    --- FAIL: TestRedisAddrUpdater/without_kubernetes_dataclient (1.01s)
        redis_test.go:428: timeout waiting for http://[::]:44997/ready
```

I guess ipv4/ipv6 problem that is not reliable listening on both in test, but not sure

szuecs:
:+1:

ayruslore:
👍

MustafaSaber:
Can you please try to explain more why it was flaky in the description?

szuecs:
> Can you please try to explain more why it was flaky in the description?

done

szuecs:
:+1:

ayruslore (approved):

filters/shedder/admission.go
```diff
@@ -94,7 +94,8 @@ const (
 )
 
 type Options struct {
-	Tracer opentracing.Tracer
+	Tracer   opentracing.Tracer
+	testRand bool
```
MustafaSaber, line 98: Minor: I think we should use some other name like `provideRand` or `staticRand` or something, it's currently only used for test and most probably will will only be used by tests but I don't think the naming is good for prod

szuecs: We don't pass a rand so `provideRand` is not good.
I like `testRand` because it shows the purpose of this unexported config option. Maybe `enableTestRand` would be better but I don't think so.

PR #3727, opened 2025-11-11
filters: registered missed awsSigv4

vlktna:
- registered missed awsSigv4

### DIFF
--- a/filters/builtin/builtin.go
+++ b/filters/builtin/builtin.go
@@ -8,6 +8,7 @@ import (
 	"github.com/zalando/skipper/filters/accesslog"
 	"github.com/zalando/skipper/filters/annotate"
 	"github.com/zalando/skipper/filters/auth"
+	"github.com/zalando/skipper/filters/awssigner/awssigv4"
 	"github.com/zalando/skipper/filters/circuit"
 	"github.com/zalando/skipper/filters/consistenthash"
 	"github.com/zalando/skipper/filters/cookie"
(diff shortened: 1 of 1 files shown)


filters/awssigner/awssigv4/sigv4.go
```diff
@@ -30,7 +30,7 @@ type awsSigV4Filter struct {
 	disableSessionToken    bool
 }
 
-func New() filters.Spec {
+func NewAwsSigv4() filters.Spec {
```
MustafaSaber: Changing name will break API, I think this was contributed by external user and maybe they already use it.

We can add both functions and add depreciation notice on the old one then remove after some releases.

filters/awssigner/awssigv4/sigv4.go
```diff
@@ -30,10 +30,15 @@ type awsSigV4Filter struct {
 	disableSessionToken    bool
 }
 
+// Deprecated: use NewAwsSigv4 instead
```
MustafaSaber: Please update to match godocs

// New create a new AWSSigV4 filter, deprecated use New....

filters/awssigner/awssigv4/sigv4.go
szuecs: We don't break this. So deprecation does not help because we never deprecate something like a filter.
Let's just drop the new function.

filters/builtin/builtin.go
```diff
@@ -170,6 +171,7 @@ func Filters() []filters.Spec {
 		NewSetDynamicBackendScheme(),
 		NewSetDynamicBackendUrl(),
 		NewOriginMarkerSpec(),
+		awssigv4.NewAwsSigv4(),
```
szuecs: awssigv4.New()

filters/awssigner/awssigv4/sigv4.go
vlktna: made it awssigv4.New()

szuecs:
:+1:

greeshma1196:
👍

PR #3728, opened 2025-11-11
test-log: reduce logging by leveraging rate.Sometimes

szuecs:
Test: reduce logging in test
Test: fix flaky test that depend on tcp.Listen to find a free port to use in tests. The implementation was done twice and had race conditions caused by a TOCTOU 
Test: fix flakytest TestRedisAddrUpdater


MustafaSaber:
👍

szuecs:
:+1:

PR #3729, opened 2025-11-12
k8s dataclient: run without polling data

MustafaSaber:
This allows k8s dataclient to run without loading routing table (which won't call apisever), this will allow us to use k8s dataclient in webhook mode to enable some filters without calling apiserver and may allow us later to get redis endpoints wihout calling routeSRV.

This should replace https://github.com/zalando/skipper/pull/3722 which disable polling routing across all dataclients which maybe not desriable.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -195,6 +195,7 @@ type Config struct {
 	KubernetesBackendTrafficAlgorithm                    kubernetes.BackendTrafficAlgorithm `yaml:"-"`
 	KubernetesDefaultLoadBalancerAlgorithm               string                             `yaml:"kubernetes-default-lb-algorithm"`
 	KubernetesForceService                               bool                               `yaml:"kubernetes-force-service"`
+	KubernetesNoPoll                                     bool                               `yaml:"kubernetes-no-poll"`
 
 	// Default filters
 	DefaultFiltersDir string `yaml:"default-filters-dir"`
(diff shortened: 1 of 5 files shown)


dataclients/kubernetes/kube.go
```diff
@@ -503,7 +513,7 @@ func healthcheckRoutes(reverseSourcePredicate bool) []*eskip.Route {
 		if i > 0 {
 			cidrs.WriteString(", ")
 		}
-		cidrs.WriteString(fmt.Sprintf("%q", ip))
+		fmt.Fprintf(cidrs, "%q", ip)
```
ayruslore, line 516: why do we need this print here.?

MustafaSaber: This prints to the stringbuilder, it just replaces cidrs.WriteString as gopls gave as a hint.

dataclients/kubernetes/kube.go
```diff
@@ -549,8 +559,7 @@ func (c *Client) LoadUpdate() ([]*eskip.Route, []string, error) {
 	)
 
 	for id := range c.current {
-		// TODO: use eskip.Eq()
-		if r, ok := next[id]; ok && r.String() != c.current[id].String() {
+		if r, ok := next[id]; ok && eskip.Eq(r, c.current[id]) {
```
szuecs, line 562: It has nothing to do with your change.

skipper.go
```diff
@@ -310,6 +310,9 @@ type Options struct {
 	// instead using Kubernetes Services.
 	KubernetesForceService bool
 
+	// KubernetesNoPoll force kubernetes dataclient to not create routing table
```
szuecs, line 313: Dataclient doesn't create a routing table.
It just creates []*eskip.Routes.
So a list of route descriptions

dataclients/kubernetes/kube.go
ayruslore: ohh interesting. Nice.

config/config.go
```diff
@@ -195,6 +195,7 @@ type Config struct {
 	KubernetesBackendTrafficAlgorithm                    kubernetes.BackendTrafficAlgorithm `yaml:"-"`
 	KubernetesDefaultLoadBalancerAlgorithm               string                             `yaml:"kubernetes-default-lb-algorithm"`
 	KubernetesForceService                               bool                               `yaml:"kubernetes-force-service"`
+	KubernetesNoPoll                                     bool                               `yaml:"kubernetes-no-poll"`
```
szuecs, line 198: Maybe better an enum to create a DisableDataClient or something like that.
I think we should not be as specific if we introduce a new configuration.

szuecs (changes requested):

dataclients/kubernetes/kube.go
MustafaSaber: yes thought would be nice to just clean the TODO

config/config.go
MustafaSaber: That's also a nice idea, thanks!

vlktna:
disabled routing table updates by setting the source-poll-timeout flag to the max int64 value

https://github.com/zalando-incubator/kubernetes-on-aws/pull/10181

dataclients/kubernetes/kube.go
szuecs: The question is why the original author didn't do it. We have to carefully check. Maybe something is missing.

MustafaSaber: You are the original author 😄

szuecs: Let's not change it right away. It creates a lot of more computation in a hot inner loop.
We should benchmark it to understand memory and CPU.

PR #3730, opened 2025-11-12
tee: only shadow portion of request

MustafaSaber:
Allow tee filter to shadow only portion of requests. In theory this can be done by multiple routes like:

```
r: Path("/") && Traffic(0.2) -> tee("https://example.com") ->  <shunt>
r: Path("/")  ->  <shunt>
```

However, adding this to tee can reduce number of maintained routes, and hacky yaml solutions to only add `tee` filter if some variable was used. Hence reduce failure impact.

- [ ] todo add tests

### DIFF
--- a/filters/tee/tee.go
+++ b/filters/tee/tee.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"math/rand"
 	"net/http"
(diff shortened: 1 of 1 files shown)


MustafaSaber:
It doesn't really make sense to go against skipper philiosphy of composition in this simple case.

PR #3739, opened 2025-11-18
Eopa decision_log plugin with open policy agent

wisinghe:
[Enterprise OPA or eopa](https://github.com/open-policy-agent/eopa) (now donated to the OPA community) provides an improved behavior of decision log export with a plugin. 
This PR introduces eopa_dl plugin into the open policy agent instance and makes sure the new plugin works with the existing opaconfig and discovery config. To use the new decision login feature, we need modifications in the configuration which is outside of this repo.

With the current opa and discovery configs, there will not be any functional changes.


filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
```diff
@@ -606,6 +987,275 @@ func TestAuthorizeRequestFilter(t *testing.T) {
...
+					t.Fatal(err)
+				}
+				assert.Equal(t, ti.body, string(body))
+			}))
+			defer clientServer.Close()
+
+			logUploadCount := 0
+			s3Server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if strings.Contains(r.URL.Path, "logs-success") {
+					logUploadCount++
```
Pushpalanka: Race detector is complaining of this been written here and read in assertion without synchronization.
It will need to be atomic, for the test reliability.

filters/openpolicyagent/opaserveresponse/opaserveresponse_test.go
```diff
@@ -163,6 +164,152 @@ func TestServerResponseFilter(t *testing.T) {
 			expectedBody:    "",
 			expectedHeaders: make(http.Header),
 		},
+		// Same tests with EOPA plugins enabled
+		{
```
Pushpalanka: This is duplicating lot of lines. Could we have a nested loop for the test cases, which iterate through enableEopaPlugins is true vs false and run the test scenarios for both?

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
```diff
@@ -392,6 +394,385 @@ func TestAuthorizeRequestFilter(t *testing.T) {
 			backendHeaders:    make(http.Header),
 			removeHeaders:     make(http.Header),
 		},
+		// Testing same cases with eopa plugins enabled
```
Pushpalanka: Same as other, this is duplicating lot of lines. 
Shall we introduce a nested loop that iterate over eopa enabling and then iterate over each scenario?

This will provide a structure that can scale for different combination of configuration options we support.

filters/openpolicyagent/opaserveresponse/opaserveresponse_test.go
mjungsbluth: Yes, you should be able to run the same tests one time with eopa enabled, one time without. I guess that this is the intent.

filters/openpolicyagent/openpolicyagent_test.go
```diff
@@ -121,6 +121,44 @@ func mockControlPlaneWithDiscoveryBundle(discoveryBundle string) (*opasdktest.Se
...
+			"data.json": `{
+			  "discovery": {
+				"bundles": {
+				  "bundles/test": {
+					"persist": false,
+					"resource": "bundles/test",
+					"service": "test"
+				  }
+				},
+				"decision_logs": {
```
mjungsbluth: This would typically be rather configured on the configuration and not via discovery, so this might simplify the test by using the same discovery.

filters/openpolicyagent/internal/eopa/eopaplugins.go
```diff
@@ -0,0 +1,15 @@
+// Package eopa provides enterprise opa plugins aggregation.
+package eopa
+
+import (
+	dl "github.com/open-policy-agent/eopa/pkg/plugins/decision_logs"
+	"github.com/open-policy-agent/opa/v1/plugins"
+)
+
+func All() map[string]plugins.Factory {
```
mjungsbluth: This really only cherry picks one plugin from eopa, when I look at 
https://github.com/open-policy-agent/eopa/blob/7b4e2ef7a83a452ccbb24c697867dd49ffd3d88e/pkg/sdk/sdk.go#L22

We are leaving out quite a few things, so the flag announces sth that is a bit overselling what is actually done (enabling S3 uploads)

filters/openpolicyagent/opaserveresponse/opaserveresponse_test.go
Pushpalanka: Yes, the intent is correct, my suggestion is more towards implementing it without duplicating the test scenario definitions. 
Something like,
```
for _, enableEopa := range []bool{false, true} {
		....
		for _, ti := range testCases {
          ....
         }
}
```
this way we can make it concise. If we have more configurations in the future, those can be more outer for loops and we will be covering all combinations.

filters/openpolicyagent/internal/eopa/eopaplugins.go
wisinghe: Will make the flag more specific.

The initial idea was to enable extra plugins which are defined here https://github.com/open-policy-agent/eopa/blob/main/pkg/plugins/all.go#L17 with the flag. Hence the name `enable-enterprise-open-policy-agent-plugins`. But we have our early implementation of the envoy plugin and the data plugin had a conflict with our body parse filter.

filters/openpolicyagent/opaserveresponse/opaserveresponse_test.go
wisinghe: Tried the wrapped loop approach suggested by @Pushpalanka which in-fact helps to manage the flag on and off situation without having to duplicate the test cases. However in my personal opinion, this makes it less readable. Also if we have more such flags, creating more and more for loops wrapping the test will be even messier. In addition it takes away the possibility to simply provide a specific message for each flag combination. So if there are no objections, I will keep the repeated test cases.

I am also not sure whether we need to run all the filter specific tests with opa config flags. I only added them in this special scenario as the usage of eopa plugins might have unforeseen impact on the existing behaviors.

filters/openpolicyagent/internal/eopa/eopaplugins.go
mjungsbluth: If you look at the link I shared there is more than the plugins but also some hooks.

If you‘re cherrypicking just the s3 plugin you might as well just add it always. It will not be active without configuration.

what specifically is the „conflict“ with the body parsing?

filters/openpolicyagent/opaserveresponse/opaserveresponse_test.go
Pushpalanka: Yes, I agree that we don't need to run these functional tests with all the OPA config flags (there are several optimization flags where we didn't run these).
However on the specific message concern, if we give a test name, it will indicate it in the test results such as ```TestAuthorizeRequestFilter/EOPA_Enabled/Allow_Requests``` which will help interpreting the failure.

No objection from me either way. :)

filters/openpolicyagent/internal/eopa/eopaplugins.go
wisinghe: Yes, the PR completely focuses on S3 decision log functionality even though there are more features provided with eopa. Currently working on a POC to see how we can switch to S3 decision logging and how to roll this out in Zalando as this will be time critical.

So we can either skip the flag or check on introducing all additional eopa features with a flag? 

The new data plugin package "github.com/open-policy-agent/eopa/pkg/plugins/data" changed the unmarshalling behavior of the of the json body, resulting in a broken json body causing 500 response code. [This](https://github.com/zalando/skipper/blob/27e58036af08e3cc3f6c0dc1d5e35c4fd710e495/filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go#L308) test fails if the new package is imported in the project. The error thrown from request body parsing differs with the new data package. Could not figure out why without spending more time on it. AFAIU with or without the new data package, the same unmarshalling logic from the same version of open-policy-agent/opa/v1/util should be used. I will create an issue to track this.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -690,7 +692,15 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, bundleName str
...
 	}
 
-	discoveryPlugin, err := discovery.New(manager, discovery.Factories(map[string]plugins.Factory{envoy.PluginName: envoy.Factory{}}), discovery.Hooks(configHooks))
+	discoveryOpts := map[string]plugins.Factory{envoy.PluginName: envoy.Factory{}, dl.DLPluginName: dl.Factory()}
+
+	var bootConfig map[string]any
+	err = util.Unmarshal(configBytes, &bootConfig)
+	if err != nil {
+		return nil, err
+	}
```
wisinghe, line 701: This part is not exactly related to enabling the eopa_dl plugin.

It seems unless we add `discovery.BootConfig` function during the discovery initialization, the `decision_log` or `plugins` configuration provided in the opaconfig.yaml is not effective.
As we are planning to work with both bootstrapped and discovered configuration (Possible from opa v0.64.0 onwards) we will need this.

If I did not miss anything, [This](https://github.com/zalando-incubator/kubernetes-on-aws/blob/91d8220721bfb1cd62251ae7f2bdc3f458eda346/cluster/manifests/skipper/configmap-open-policy-agent.yaml#L35) configuration we have for the clusters is not currently doing anything and will start to make an impact once this change is merged.

filters/openpolicyagent/internal/eopa/eopaplugins.go
wisinghe: I removed the flag as it is a harmless plugin with no impact unless we change the configuration.

Created https://github.com/zalando-infosec/corporate-iam-issues/issues/1395 to enable all eopa capabilities

filters/openpolicyagent/openpolicyagent.go
torwunder: so regardless the decision log config will be overwritten by Styra DAS correct?
So to test it in integration we would need to change the discovery config?

wisinghe: With the current filter implementation, yes.

This code change takes away the requirement to change the discovery config, by passing the bootstrap config into the discovery initialization (at line 703)

wisinghe:
Development over time and multiple master merges on the way made it almost impossible to rebase. Closing this PR and changes are moved here https://github.com/zalando/skipper/pull/3792

PR #3743, opened 2025-11-18
fix: flaky routing tests and speedup

szuecs:
See commit messages if you want to see more progress on test speedups
For an overview you can check https://github.com/zalando/skipper/actions/workflows/pr.yaml and compare test running time from this PR to other PRs and you can see this PR takes 5-6m and other PRs 8-9 so 25%-45% speedup

Example speedup: 13x

before
% go test -race ./routing -run TestSignalFirstLoad -count 100
ok      github.com/zalando/skipper/routing      13.615s

after:
% go test -race ./routing -run TestSignalFirstLoad -count 100
ok      github.com/zalando/skipper/routing      1.536s

speedup routing datasource tests

### DIFF
--- a/filters/scheduler/fifo_test.go
+++ b/filters/scheduler/fifo_test.go
@@ -394,7 +394,7 @@ func TestFifo(t *testing.T) {
 	}{
 		{
 			name:        "fifo simple ok",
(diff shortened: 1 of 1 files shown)


a4180p:
👍

PR #3746, opened 2025-11-19
gh-action: fix security scan and labeler syntax

MustafaSaber:
follow up on https://github.com/zalando/skipper/pull/3740


MustafaSaber:
It's currently working but on the old config, merging this should make it work

MustafaSaber:
👍

PR #3768, opened 2025-12-01
Telemetry for passive health check

ponimas:
Having PHC tags on the ~proxy~ ingress spans will allow us to easily detect a suffering route.

### DIFF
--- a/proxy/healthy_endpoints.go
+++ b/proxy/healthy_endpoints.go
@@ -3,6 +3,7 @@ package proxy
 import (
 	"math/rand"
 
+	ot "github.com/opentracing/opentracing-go"
 	"github.com/zalando/skipper/metrics"
 	"github.com/zalando/skipper/routing"
 )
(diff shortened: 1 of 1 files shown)


proxy/healthy_endpoints.go
```diff
@@ -17,6 +18,8 @@ func (h *healthyEndpoints) filterHealthyEndpoints(ctx *context, endpoints []rout
 		return endpoints
 	}
 
+	span := ot.SpanFromContext(ctx.request.Context())
```
szuecs, line 21: we should pass the span or lookup from context. A bit later in the proxy we do exactly the same.

a4180p: We lookup it from context here just like we [do in proxy](https://github.com/zalando/skipper/blob/91237ea140592617b11d42c2c1e671b54560fc80/proxy/proxy.go#L1022)

szuecs: Exactly so we could do it once and pass it.

a4180p: But what advantages does it have? 
We pass context around anyway, and now we should change signatures of N functions just to explicitly pass span.

szuecs: My thought was that we don't have to restore it from context but likely it's only a map lookup anyways so, fine.

szuecs:
:+1:

a4180p:
👍

PR #3775, opened 2025-12-04
validation: fix memory leak

vlktna:
- closed the request body in admission webhook
- closed filters after validation
- added a check in the lifo filter to ensure the queue is not nil during closing

### DIFF
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -475,7 +475,13 @@ func processTreePredicates(r *Route, predicateList []*eskip.Predicate) error {
 // ValidateRoute processes a route definition for the routing table.
 // This function is exported to be used by validation webhooks.
 func ValidateRoute(o *Options, def *eskip.Route) (*Route, error) {
-	return processRouteDef(o, mapPredicates(o.Predicates), def)
+	route, err := processRouteDef(o, mapPredicates(o.Predicates), def)
+	if err != nil {
+		return nil, err
+	}
+
+	defer closeFilters(route.Filters)
+	return route, nil
 }
 
 // processes a route definition for the routing table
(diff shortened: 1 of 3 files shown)


routing/datasource.go
```diff
@@ -489,6 +495,11 @@ func processRouteDef(o *Options, cpm map[string]PredicateSpec, def *eskip.Route)
 	if err != nil {
 		return nil, err
 	}
+	defer func() {
+		if err != nil {
+			closeFilterClosers(fs)
```
szuecs: I think here we don't need it.
It feel safer to not touch `processRouteDef` and in the `ValidateRoute` case we move the defer to the top.

I think this would close all filters after having created a routing tree and then every filter that is a FilterCloser is closed instead of working.

routing/datasource.go
```diff
@@ -563,15 +574,19 @@ type routeTable struct {
...
-				if fc, ok := f.Filter.(filters.FilterCloser); ok {
-					fc.Close()
-				}
-			}
+			closeFilterClosers(route.Filters)
 		}
 	})
 }
 
+func closeFilterClosers(rfs []*RouteFilter) {
```
MustafaSaber: nitpick: `closeFilters` I think is a better name

routing/datasource.go
```diff
@@ -563,15 +574,19 @@ type routeTable struct {
...
+	for _, f := range rfs {
+		if fc, ok := f.Filter.(filters.FilterCloser); ok {
+			_ = fc.Close()
```
MustafaSaber: just 
```suggestion
			fc.Close()
```

szuecs:
:+1:

MustafaSaber:
👍

PR #3776, opened 2025-12-05
metrics: remove invalid routes deletion

vlktna:
When validation is enabled and a new invalid route is created, it gets rejected and a metric is emitted. But when the route is later fixed, the old metric is not removed. Once the route is fixed, there is no error, so the corrected route no longer matches the old metric key.

### DIFF
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -531,9 +531,6 @@ func processRouteDefs(o *Options, defs []*eskip.Route) (routes []*Route, invalid
 		route, err := processRouteDef(o, cpm, def)
 		if err == nil {
 			routes = append(routes, route)
-			if o.Metrics != nil {
-				o.Metrics.DeleteInvalidRoute(def.Id)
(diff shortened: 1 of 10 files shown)


routing/datasource.go
```diff
@@ -475,7 +475,7 @@ func processTreePredicates(r *Route, predicateList []*eskip.Predicate) error {
 // ValidateRoute processes a route definition for the routing table.
 // This function is exported to be used by validation webhooks.
 func ValidateRoute(o *Options, def *eskip.Route) (*Route, error) {
-	return processRouteDef(o, mapPredicates(o.Predicates), def)
+	return processRouteDef(o, mapPredicates(o.Predicates), def) //
```
szuecs: cleanup

szuecs:
:+1:

vlktna:
👍

MustafaSaber:
👍

PR #3778, opened 2025-12-08
feature/builtin: drop specific header value

MustafaSaber:
Add `dropRequestHeaderValueRegexp` and `dropResponseHeaderValueRegexp` filters that allow selective removal of header values based on regex patterns, unlike the existing drop filters that remove entire headers.

- [ ] **Can Improve** Add local caching for regex compilation

Replaces https://github.com/zalando/skipper/pull/3457


filters/builtin/headerfilter.go
```diff
@@ -38,43 +42,50 @@ const (
...
-	template   *eskip.Template
+	typ         headerType
+	key, value  string
+	template    *eskip.Template
+	valueRegexp *regexp.Regexp
 }
 
 // verifies that the filter config has two string parameters
-func headerFilterConfig(typ headerType, config []interface{}) (string, string, *eskip.Template, error) {
+func headerFilterConfig(typ headerType, config []interface{}) (string, string, *regexp.Regexp, *eskip.Template, error) {
```
szuecs, line 52: IMO >2 return values is a code smell.
Maybe refactoring would be good.

filters/builtin/headerfilter.go
```diff
@@ -314,6 +341,10 @@ func (f *headerFilter) Response(ctx filters.FilterContext) {
 		header.Add(f.key, f.value)
 	case dropResponseHeader:
 		header.Del(f.key)
+	case dropResponseHeaderRegexp:
+		if headerValue := header.Get(f.key); headerValue != "" {
+			header[f.key] = slices.DeleteFunc(header.Values(f.key), func(value string) bool { return f.valueRegexp.MatchString(value) })
```
szuecs: Shouldn't you also delete the key from the map?

filters/builtin/compress_test.go
```diff
@@ -543,13 +544,7 @@ func TestCompress(t *testing.T) {
...
 			}
 
-			if !compareHeaders(rsp.Header, ti.expectedHeader) {
-				printHeader(t, ti.expectedHeader, "invalid header", "expected")
-				printHeader(t, rsp.Header, "invalid header", "got")
-
-				t.Error("invalid header")
-				return
-			}
+			assert.Equal(t, ti.expectedHeader, rsp.Header)
```
szuecs: This seems to be not scope of the feature.

filters/builtin/headerfilter.go
MustafaSaber: Prefer to do this in a separate PR, but I agree with you.

filters/builtin/headerfilter.go
MustafaSaber: We can have multiple values, didn't want to remove the key because we only want to drop some values (which may or may not become empty after removal) not the full key, if someone wants to remove the key then it's better to use `dropRequestHeader` directly

filters/builtin/compress_test.go
MustafaSaber: I removed the function initiation from header_test.go so I removed all the usages, and this is one of them.

filters/builtin/headerfilter.go
szuecs: that's fine

filters/builtin/compress_test.go
szuecs: that's a refactoring not a feature -> separate PR

filters/builtin/headerfilter.go
szuecs: Then the filter name should be changed to something "Value"

filters/builtin/compress_test.go
MustafaSaber: Will add it with the PR to change return type

szuecs:
:+1:

MustafaSaber:
👍

filters/builtin/compress_test.go
MustafaSaber: the follow up https://github.com/zalando/skipper/pull/3779

PR #3779, opened 2025-12-08
refactor: update `CreateFilter` function for header filters

MustafaSaber:
Inline headerFilterConfig to reduce function return params and reduce complexity. Remove some test functions in favour of `assert` as it's already implementing same functionality.

Use `any` instead of `interface{}` when possible just for less linters suggesstions.

Follow up on https://github.com/zalando/skipper/pull/3778

### DIFF
--- a/filters/builtin/header_test.go
+++ b/filters/builtin/header_test.go
@@ -44,43 +44,6 @@ func (c testContext) Request(ctx filters.FilterContext) {
 
 func (c testContext) Response(filters.FilterContext) {}
 
-func printHeader(t *testing.T, h http.Header, msg ...interface{}) {
-	for k, v := range h {
-		for _, vi := range v {
-			t.Log(append(msg, k, vi)...)
-		}
-	}
-}
-
-func compareHeaders(left, right http.Header) bool {
-	if len(left) != len(right) {
-		return false
-	}
-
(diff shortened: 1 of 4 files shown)


filters/builtin/header_test.go
```diff
@@ -20,7 +20,7 @@ type testContext struct {
 
 func (c testContext) Name() string { return "testContext" }
 
-func (c testContext) CreateFilter(args []interface{}) (filters.Filter, error) {
+func (c testContext) CreateFilter(args []any) (filters.Filter, error) {
```
szuecs: Not sure if this is compatible with our filter interface, I would say no.

MustafaSaber: It works fine, but will remove this from CreateFilter

szuecs: Ah nice https://go.dev/play/p/TgjUOiFgsqH
I guess we can do a refactoring to do change it for all parts (all filters, tests, ..)

szuecs:
:+1:

szuecs:
The fuzzer fails with null ptr deref and:
```
2025-12-09 07:51:44,587 - root - INFO - crash_type: no-crashes.
```
so fine

MustafaSaber:
👍

PR #3780, opened 2025-12-09
fix: outdated env syntax in dockerfiles

ponimas:
We have `#  - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 26)` in build logs.
See https://docs.docker.com/reference/dockerfile/#env


szuecs:
:+1:

a4180p:
👍

PR #3783, opened 2025-12-10
refactor: admission test to use synctest

ponimas:
- Replace real network connections with fake network pipes and listeners to enable controlled simulation of network behavior. Implementation is taken from https://go.dev/src/net/http/netconn_test.go
- Fix the loop in admissionControl ticker, we were checking the quit channel only on tick instead of waiting for one of the two conditions.
- Simplify `TestAdmissionControl`. We control rand and time and can expect exact values.

ref https://github.com/zalando/skipper/issues/2386

### DIFF
--- a/filters/shedder/admission_test.go
+++ b/filters/shedder/admission_test.go
@@ -241,7 +241,7 @@ func TestAdmissionControl(t *testing.T) {
 
 				N := ti.N * ti.windowsize * int(ti.d.Milliseconds())
 
-				for _ = range N {
+				for range N {
 					time.Sleep(time.Millisecond)
 
 					rsp, err := client.Do(req)


filters/shedder/admission_test.go
```diff
@@ -1,15 +1,18 @@
 package shedder
 
 import (
+	"context"
 	"fmt"
 	"io"
+	gonet "net"
```
szuecs: I think we either use stdlibnet or use net and change the skipper net package to alias to skpnet.
So for consistency I think alias the current skipper/net to skpnet would be best

a4180p:
👍

szuecs (approved):

PR #1550, opened 2020-10-12
handle the case smoother when all endpoints are fading in

aryszka:
use weighted shifting instead of shifting to oldest when there are no available stable endpoints

PR #3786, opened 2025-12-11
Optimize/rand

szuecs:
optimize: Go stdlib math/rand is not safe for concurrent access. By using math/rand/v2 we can speedup random generator code to increase concurrency.

The commit messages show a speedup of 2x in micro-benchmarks of some hot code paths.

This also fixes a flaky test introduced when the case was fixed https://github.com/zalando/skipper/pull/1550

### DIFF
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -58,7 +58,7 @@ type roundRobin struct {
 
 func newRoundRobin(endpoints []string) routing.LBAlgorithm {
 	return &roundRobin{
-		index: rand.New(rand.NewPCG(uint64(time.Now().UnixNano()), 0)).Int64N(int64(len(endpoints))), // #nosec
+		index: rand.Int64N(int64(len(endpoints))), // #nosec
 	}
 }
 
@@ -78,6 +78,7 @@ type random struct {
 }
 
 func newRandom(endpoints []string) routing.LBAlgorithm {
+	// #nosec
 	return &random{
 		rnd: rand.New(rand.NewPCG(uint64(time.Now().UnixNano()), 0)),
 	}
@@ -225,16 +226,15 @@ func (ch *consistentHash) chooseConsistentHashEndpoint(ctx *routing.LBContext) i
 }
 
 type powerOfRandomNChoices struct {
-	mu  sync.Mutex
-	rnd *rand.Rand
-
+	mu              sync.Mutex
+	rnd             *rand.Rand
 	numberOfChoices int
 }
 
 // newPowerOfRandomNChoices selects N random backends and picks the one with less outstanding requests.
 func newPowerOfRandomNChoices([]string) routing.LBAlgorithm {
 	return &powerOfRandomNChoices{
-		rnd:             rand.New(rand.NewPCG(uint64(time.Now().UnixNano()), 0)),
+		rnd:             rand.New(rand.NewPCG(uint64(time.Now().UnixNano()), 0)), // #nosec
 		numberOfChoices: powerOfRandomNChoicesDefaultN,
 	}
 }


filters/diag/diag.go
```diff
@@ -108,7 +110,16 @@ type jitter struct {
...
+func genCharArray(s string) [62]byte {
+	res := [62]byte{}
+	for i := 0; i < len(s); i++ {
+		res[i] = s[i]
+	}
+
+	return res
+}
+
+var randomChars = genCharArray("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789")
```
szuecs, line 122: In Go arrays can be concurrently accessed without lock, slices can grow so they can not access concurrently

filters/flowid/ulid.go
```diff
@@ -26,7 +23,7 @@ var ulidFlowIDRegex = regexp.MustCompile(`^[0123456789ABCDEFGHJKMNPQRSTVWXYZ]{26
 // It uses a shared, pseudo-random source of entropy, seeded with the current timestamp.
 // It is safe for concurrent usage.
 func NewULIDGenerator() Generator {
-	return NewULIDGeneratorWithEntropyProvider(rand.New(rand.NewSource(time.Now().UTC().UnixNano()))) // #nosec
+	return NewULIDGeneratorWithEntropyProvider(&rand.ChaCha8{})
```
szuecs, line 26: ChaCha8 does not need to be guarded by locks and has an implementation of io.Reader interface

eskip/eskip.go
```diff
@@ -769,7 +769,7 @@ func GenerateIfNeeded(existingId string) string {
 	sb.WriteString("route")
 
 	for i := 0; i < randomIdLength; i++ {
-		ai := rand.Intn(len(alphabet))
+		ai := rand.IntN(len(alphabet)) // #nosec
```
MustafaSaber, line 772: Why the optout here and why it didn't fail before before? I also think it's better to not update the deprecated function and remove it, if it's too old.

filters/diag/diag.go
```diff
@@ -217,25 +228,23 @@ func NewUniformResponseLatency() filters.Spec { return &jitter{typ: uniformRespo
...
 
-func (r *random) CreateFilter(args []interface{}) (filters.Filter, error) {
+func (r *randomSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
 	if len(args) != 1 {
 		return nil, filters.ErrInvalidFilterParameters
 	}
 
 	if l, ok := args[0].(float64); ok {
-		return &random{rand: rand.New(rand.NewSource(time.Now().UnixNano())), len: int64(l)}, nil // #nosec
+		return &random{rand: rand.New(rand.NewPCG(uint64(time.Now().UnixNano()), 0)), len: int64(l)}, nil // #nosec
```
MustafaSaber, line 239: Why do we use PCG here instead of ChaCha8, speed?

eskip/eskip.go
MustafaSaber: There is multiple optouts but I can't figure why it fails now

szuecs: We don't run gosec in pipelines anymore:

```
% grep -rn gosec .github
zsh: exit 1
```

filters/diag/diag.go
szuecs: I did not check if it would make sense to use ChaCha8. Maybe I should benchmark this.

szuecs: I checked it and ChaCha8 does not have IntN() but only Uint64().
So it would be a bigger change or if we would use Read() it would change the alphabet of the returned random data, because right now we return character sequence of the alphabet `[A-Za-z0-9]` and ChaCha8.Read will return random bytes.

eskip/eskip.go
szuecs: It was too noisy and a discussion with the maintainers showed that they do not care about users that want not to see false positives (Go style is omit false positives, see also why go vet or staticcheck have basically 0% false positives to not annoy the developer). Because of this I decided to drop gosec from our pipelines. We have it in our Makefile to check once in a while but there are better scanners available that we run.

szuecs:
:+1:

MustafaSaber:
👍

PR #3792, opened 2025-12-15
Include eopa_dl plugin in opa

wisinghe:
This PR introduces the `eopa_dl` plugin ([documentation](https://docs.styra.com/enterprise-opa/reference/configuration/decision-logs)) into the open policy agent instance running in skipper filter. `eopa_dl` plugin allows multiple types of outputs. AWS S3 is one of the allowed output types. If the running skipper instance has an already assumed role with correct permission to a S3 bucket, opa can use the same credentials to upload decision logs to the S3.

The eopa_dl plugin configuration can be either provided by opaconfig.yaml or discovery

Old PR: https://github.com/zalando/skipper/pull/3739 (had to close due to messed up git history)


filters/openpolicyagent/openpolicyagent.go
```diff
@@ -690,7 +692,15 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, bundleName str
 		return nil, err
 	}
 
-	discoveryPlugin, err := discovery.New(manager, discovery.Factories(map[string]plugins.Factory{envoy.PluginName: envoy.Factory{}}), discovery.Hooks(configHooks))
+	discoveryOpts := map[string]plugins.Factory{envoy.PluginName: envoy.Factory{}, dl.DLPluginName: dl.Factory()}
```
torwunder: `pluginFactories`? or simply `plugins`?

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
```diff
@@ -607,6 +607,270 @@ func TestAuthorizeRequestFilter(t *testing.T) {
...
+		requestMethod     string
+		body              string
+		contextExtensions string
+		expectedBody      string
+		expectedHeaders   http.Header
+		expectedStatus    int
+		backendHeaders    http.Header
+		removeHeaders     http.Header
+		discoveryPath     string
+		discoveryBundle   string
```
torwunder: you could trim down the test case params as a lot don't seem to be used or initialized with default values (liek the headers)

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
```diff
@@ -607,6 +607,270 @@ func TestAuthorizeRequestFilter(t *testing.T) {
...
+				} else if strings.Contains(r.URL.Path, "logs-timeout") {
+					time.Sleep(5 * time.Second)
+					w.WriteHeader(http.StatusOK)
+				} else if strings.Contains(r.URL.Path, "logs-5xx") {
+					w.WriteHeader(http.StatusInternalServerError)
+				} else {
+					w.WriteHeader(http.StatusNotFound)
+				}
+			}))
+			defer s3Server.Close()
```
torwunder, line 702: is there a check that the decisions are actually sent?
it would also be nice to check at least some of the content

wisinghe: Verifying the data from async process was tricky. I will try a bit more

filters/openpolicyagent/openpolicyagent.go
wisinghe: `plugins` from github.com/open-policy-agent/opa/v1/plugins

When creating the discoveryPlugin using discovery.New(), I wanted to pass a `map[string]plugins.Factory` object with two plugins.Factory structs . Here I initialize this map.

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
wisinghe: Added some assertions on the payload sent to S3

go.mod
```diff
@@ -73,49 +74,104 @@ require (
...
 	github.com/Microsoft/go-winio v0.6.2 // indirect
+	github.com/OneOfOne/xxhash v1.2.8 // indirect
 	github.com/agnivade/levenshtein v1.2.1 // indirect
 	github.com/armon/go-metrics v0.4.1 // indirect
+	github.com/aws/aws-sdk-go-v2 v1.41.0 // indirect
+	github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream v1.7.4 // indirect
+	github.com/aws/aws-sdk-go-v2/config v1.32.6 // indirect
+	github.com/aws/aws-sdk-go-v2/credentials v1.19.6 // indirect
+	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.18.16 // indirect
+	github.com/aws/aws-sdk-go-v2/feature/s3/manager v1.20.17 // indirect
```
szuecs: If we merge it we can also add some AWS specific filters, because here we import many cloud provider dependencies.
I wonder how big the binary will be, can you build the binary and share the MB, please ?

wisinghe: The binary is actually 156.9MB in this case. While master branch built one is 80.2.

What would be a recommended way to add AWS specific filters?

filters/openpolicyagent/openpolicyagent.go
torwunder: this comment was about the name of the variable. `plugins` or `pluginFactories` seems to be more descriptive compared to a generic `discoveryOpts` that you pass to `discovery.Factories`

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
```diff
@@ -801,3 +1023,15 @@ func isHeadersAbsent(t *testing.T, unwantedHeaders http.Header, headers http.Hea
...
+func assertDecisionLogJSON(t *testing.T, body []byte) {
+	t.Helper()
+
+	var decisionLog map[string]interface{}
+	err := json.Unmarshal(body, &decisionLog)
+	assert.NoError(t, err, "Expected valid JSON in decision log")
+
+	assert.NotNil(t, decisionLog["decision_id"])
+	assert.NotNil(t, decisionLog["input"])
+	assert.NotNil(t, decisionLog["result"])
```
torwunder, line 1036: given that you already know if the decision is allowed or denied you could verify if the decision is allowed and if the path is what you used for the testcase

filters/openpolicyagent/openpolicyagent.go
wisinghe: Misunderstood the comment before :)

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
wisinghe: Checking test instance related data in s3 calls cause race conditions. Removed this for now

go.mod
szuecs: Oh, that's huge.
The question is if we want to have build tags in order to reduce the binary size.
It's not really something you should care about, but we should care.

So from my side it's not a blocker, but I may create some build tag to also have a non OPA build for people who care about binary size. 

cc @MustafaSaber

szuecs:
@wisinghe there is a conflict

go.mod
```diff
@@ -19,25 +19,26 @@ require (
...
 	github.com/golang-jwt/jwt/v4 v4.5.2
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
-	github.com/hashicorp/memberlist v0.5.4
-	github.com/instana/go-sensor v1.72.0
+	github.com/hashicorp/memberlist v0.5.3
+	github.com/instana/go-sensor v1.71.2
 	github.com/lightstep/lightstep-tracer-go v0.26.0
-	github.com/miekg/dns v1.1.72
+	github.com/miekg/dns v1.1.68
```
szuecs, line 29: please check that we do not downgrade, there are also other libs here that were downgraded

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -18,6 +18,9 @@ import (
 	"text/template"
 	"time"
 
+	dl "github.com/open-policy-agent/eopa/pkg/plugins/decision_logs"
+	"github.com/open-policy-agent/opa/v1/util"
```
szuecs, line 22: can you group the import 5 lines below is a section for opa

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
```diff
@@ -607,6 +608,227 @@ func TestAuthorizeRequestFilter(t *testing.T) {
...
+							"persist": false,
+							"resource": "bundles/test",
+							"service": "test"
+						  }
+				}}
+				}`,
+				}),
+			)
+
+			config := []byte(fmt.Sprintf(`{
```
szuecs, line 731: You can use `fmt.Appendf` instead `[]byte(fmt.Sprintf)`

wisinghe:
Closing as the changes were merged with https://github.com/zalando/skipper/pull/3846

PR #3794, opened 2025-12-17
optimize: routesrv eskip route creation

szuecs:
optimize: routesrv eskip route creation

We store lastRoutes to compare with new routes to reduce processing routes if there was no change detected. We do this because we observe a high memory objects creation and high cpu usage on the processing (Editor and formatAndSet)

### current

memory profile
image

### after change

memory profile
image

## busy cluster

memory profile
image

cpu profile
image

## after change

memory profile
image

cpu profile
image

### DIFF
--- a/routesrv/polling.go
+++ b/routesrv/polling.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"strings"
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

PR #3800, opened 2025-12-19
fix flakytest: often we get an off by +/-1 so define an epsilon band

szuecs:
fix flakytest: often we get an off by +/-1 so define an epsilon and make sure we are within that bounds

refactor: close filter immediately, because it'S only used to test that the filter args are fine


vlktna:
👍

PR #3807, opened 2025-12-30
build: provide diff from `make fmt`

jsoref:
It's helpful show contributors what changes they need to make...

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -41,7 +41,10 @@ jobs:
           check-latest: true
       - run: go version
       - run: make deps
-      - run: make check-fmt
+      - run: |
+          : Check format
+          make fmt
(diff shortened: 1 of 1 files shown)


szuecs:
I think it would make sense to change the makefile target instead of the pipeline

jsoref:
I originally did that and then thought that it might be used by others and thus be an unwanted change there. 

I'm happy with either approach.

jsoref:
Oh, the other reason we can't really change `check-fmt` is that it's reasonable to assume that it won't rewrite files that people are working on.

jsoref:
For the curious, this is what these changes look like: https://github.com/jsoref/skipper/actions/runs/20621396207/job/59223906377

szuecs:
:+1:

greeshma1196:
👍

PR #3811, opened 2026-01-02
test: reduce proxy access logs

szuecs:
test: reduce proxy access logs

These logs were not helpful in the past year to understand the flakytest problem so let's drop them


szuecs:
:+1:

a4180p:
👍

Issue #3799, opened 2025-12-19
valkey swarm


szuecs:
Redis was a good choice back then we implemented the clusterRatelimit and Swarmer interface.
Today there is valkey which is the better choice for opensource projects, because redis got bloated by enterprise features.
Also go-redis was not the best choice of redis clients it seems. go-redis did many breaking changes during the years and just rewrote their client connection pool, that has too many code smells to tell about. We found a memory leak in their connection pool.

With valkey we have the possibility to change to valkey-go which looks much better.

szuecs:
https://github.com/zalando/skipper/pull/3819 implemented base block to have ring sharding for valkey
https://github.com/zalando/skipper/pull/3866 implemented ratelimit features to be able to use valkey

Missing: documentation

PR #3819, opened 2026-01-06
Feature: net package valkey ring client implementation

szuecs:
Feature: net package valkey ring client implementation

This is the basic block to have another Swarmer implementation to implement for example `cluster*Ratelimit` filters.
In order to allow migration from redis to valkey, this is the first step out of two.

ref https://github.com/zalando/skipper/issues/3799
valkey command sheet: https://valkey.io/commands

### DIFF
--- a/net/valkey.go
+++ b/net/valkey.go
@@ -94,7 +94,7 @@ func createValkeyClient(addr string, opt *ValkeyOptions) (valkey.Client, error) 
 	)
 
 	if opt.EnableOTel {
-		valkeyotel.NewClient(clientOptions, opt.OTelOptions...)
+		cli, err = valkeyotel.NewClient(clientOptions, opt.OTelOptions...)
 	} else {
 		cli, err = valkey.NewClient(clientOptions)
 	}
@@ -294,7 +294,7 @@ func (vr *valkeyRing) RunScript(ctx context.Context, script *valkey.Lua, keys []
 	return script.Exec(ctx, shard, keys, args)
 }
 
-// ValkeyRingClient is a wrapper aroung valkey.Client that does access valkey shard by
+// ValkeyRingClient is a wrapper around valkey.Client that does access valkey shard by
 // computing a ring hash. It logs to the logging.Logger interface,
 // that you can pass. It adds metrics and operations are traced with
 // opentracing. You can set timeouts and the defaults are set to be ok
@@ -464,6 +464,9 @@ func (vrc *ValkeyRingClient) SetWithExpire(ctx context.Context, key string, valu
 		if err := res.Error(); err != nil {
 			return err
 		}
+	}
+	if len(results) == 0 {
+		return fmt.Errorf("failed to SetWithExpire, no result")
 	}
 	return results[len(results)-1].Error()
 }


net/valkey.go
```diff
@@ -0,0 +1,557 @@
...
+	}
+	return cli, err
+}
+
+type valkeyRing struct {
+	opt *ValkeyOptions
+
+	// maps int to client for sharding, trades memory for concurrent access
+	// most operations only have to use this lock-free structure
+	shards       [ringSize]valkey.Client
```
szuecs, line 113: this data structure is the key for the lock free access

net/valkey.go
```diff
@@ -0,0 +1,557 @@
...
+
+	for _, cl := range vr.clientMap {
+		clients = append(clients, cl)
+	}
+
+	for i := range ringSize {
+		if i%shardSize == 0 {
+			cur++
+		}
+		vr.shards[i] = clients[cur]
```
szuecs, line 160: here we set the values of the lock free data structure, if we get called by SetAddr() , for example scliang-out valkey instances

net/valkey.go
```diff
@@ -0,0 +1,557 @@
...
+		curAddr = append(curAddr, k)
+	}
+	vr.updateShards(curAddr)
+
+	return nil
+}
+
+// shardForKey does the lookup for valkey most operations to find the valkey ring shard
+func (vr *valkeyRing) shardForKey(key string) valkey.Client {
+	return vr.shards[xxhash.Sum64String(key)%ringSize]
```
szuecs, line 219: this is the O(1) lock free lookup of the valkey client connected to the right shard

net/valkey.go
```diff
@@ -0,0 +1,557 @@
...
+}
+
+// PingAll pings all known shards
+func (vr *valkeyRing) PingAll(ctx context.Context) map[string]valkey.ValkeyResult {
```
a4180p, line 223: This method is not used anywhere. IIUC in the Redis implementation, it removes unreachable shards from the ring. Was it omitted for simplification?

szuecs: Yes it's only the base code such that we can build it (autoscaling and ratelimit) on top.

greeshma1196:
lgtm, nit: can we create a sub-issue highlighting all that is left to complete the implementation?

greeshma1196:
👍🏽

a4180p:
👍

net/valkey.go
```diff
@@ -0,0 +1,557 @@
...
+		// BlockingPipeline:    0,
+
+	}
+	var (
+		cli valkey.Client
+		err error
+	)
+
+	if opt.EnableOTel {
+		valkeyotel.NewClient(clientOptions, opt.OTelOptions...)
```
MustafaSaber: This should return cli and err, no?

net/valkey.go
```diff
@@ -0,0 +1,557 @@
...
+	shard := vr.shardForKey(key)
+	return shard.Do(ctx, shard.B().Zrangebyscore().Key(key).Min(min).Max(max).Withscores().Limit(offset, count).Build())
+}
+
+func (vr *valkeyRing) RunScript(ctx context.Context, script *valkey.Lua, keys []string, args ...string) valkey.ValkeyResult {
+	shard := vr.shardForKey(strings.Join(keys, ""))
+	return script.Exec(ctx, shard, keys, args)
+}
+
+// ValkeyRingClient is a wrapper aroung valkey.Client that does access valkey shard by
```
MustafaSaber: aroung -> around

net/valkey.go
```diff
@@ -0,0 +1,557 @@
...
+}
+
+func (vrc *ValkeyRingClient) SetWithExpire(ctx context.Context, key string, value string, expire time.Duration) error {
+	results := vrc.ring.SetWithExpire(ctx, key, value, expire)
+	for _, res := range results {
+		if err := res.Error(); err != nil {
+			return err
+		}
+	}
+	return results[len(results)-1].Error()
```
MustafaSaber, line 471: This can reference invalid index in case we return empty result (I don't know if it's a valid response)

net/valkey.go
szuecs: haha, we should also use the client :D

Thank you!

net/valkey.go
szuecs: if `len(results) == 0`, better to guard it!

szuecs: fixed

szuecs:
:+1:

MustafaSaber (approved):

PR #3820, opened 2026-01-09
fix: log poisoning via xff header

szuecs:
fix: log poisoning via xff header

In this change we remove ports from access log XFF headers and only log IPs on purpose.
It makes more sense to be able to group by these values without having to strip port part, so we do not expect that anyone depends on this behavior.


szuecs:
:+1:

PR #3825, opened 2026-01-13
context: span tag `shadow: true` should only be set for shadow traffic

madumalt:
Shadow Traffic usually means the duplicated traffic whose response is discarded. The loopbackInternal method get called for teeLoopback filter and loopbackIfStatus filters. At the moment in both the cases we tag the span with `shadow: true`. We wanna do this only for the teeLoopback case and this change accomplishes that.

### DIFF
--- a/proxy/tracing_test.go
+++ b/proxy/tracing_test.go
@@ -411,6 +411,138 @@ func TestTracingProxySpanWithRetry(t *testing.T) {
 	}
 }
 
+func TestTracingProxySpanWithTeeLoopbackFilter(t *testing.T) {
+	// shadow backend server for shadow traffic, if shunt backend is used proxy span will be nil
+	sb := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte("Hello! I'm shadow backend!"))
+	}))
+	defer sb.Close()
+
+	doc := fmt.Sprintf(`
+main: Path("/main") -> teeLoopback("shadow-main") -> inlineContent("<h1>Hello! I'm main :)</h1>") -> <shunt>;
+shadow: Path("/main") && Tee("shadow-main") && Weight(2) -> "%s";
+`, sb.URL)
+
+	tracer := tracingtest.NewTracer()
+	params := Params{
+		OpenTracing: &OpenTracingParams{
+			Tracer: tracer,
+		},
+		Flags: FlagsNone,
+	}
+
+	t.Setenv("HOSTNAME", "shadow-tag.tracing.test")
+
+	tp, err := newTestProxyWithParams(doc, params)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tp.close()
+
+	ps := httptest.NewServer(tp.proxy)
+	defer ps.Close()
+
+	req, err := http.NewRequest("GET", ps.URL+"/main", nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	req.Header.Set("X-Flow-Id", "test-flow-id")
+
+	rsp, err := ps.Client().Do(req)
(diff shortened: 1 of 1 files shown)


szuecs:
no test to reproduce the issue?

proxy/context.go
```diff
@@ -317,10 +317,12 @@ func (c *context) loopbackInternal(discardResponse bool) {
...
 		}
-	}
-	if c.proxySpan != nil {
-		c.proxy.tracing.setTag(c.proxySpan, "shadow", "true")
-		c.proxySpan.Finish()
+
+		// span tag shadow: true should only be set if the response is discarded
+		if c.proxySpan != nil {
+			c.proxy.tracing.setTag(c.proxySpan, "shadow", "true")
+			c.proxySpan.Finish()
```
szuecs: not sure if we should move finish of proxySpan, too.
It will be a logical change if we move this, too and I don't see an explanation why it is correct.

madumalt: > not sure if we should move finish of proxySpan, too.

yeah true, only the shadow tag should be set when discardResponse is true. Moved the proxySpan.Finish back to keep the same existing behaviour.

madumalt:
> no test to reproduce the issue?

I will try to add a test case for this tmrw or so.

madumalt:
> > no test to reproduce the issue?
> 
> I will try to add a test case for this tmrw or so.

Hi @szuecs I added two test cases. `TestTracingProxySpanWithLoopbackIfStatusFilter` fails with the existing implementation but fixed with this PR.

proxy/tracing_test.go
```diff
@@ -411,6 +411,146 @@ func TestTracingProxySpanWithRetry(t *testing.T) {
...
+	tp, err := newTestProxyWithParams(doc, params)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tp.close()
+
+	ps := httptest.NewServer(tp.proxy)
+	defer ps.Close()
+
+	t.Run("teeLoopback sets shadow: true tag in the proxy span", func(t *testing.T) {
```
szuecs: can you please add the whole above creation into the sub-test here?
The reason is that you will not need tracer.Reset(), nor you will depend on a shared state if you add more sub-tests.

szuecs: Sometimes it's needed to do this but then I would create a table driven test that parameterize some things that every test has a separated data set, or you will get flakytests.

madumalt: I removed the t.Run, it is not necessary as you said. I started with having both the test cases in one function but later created two test functions and forgot to take the sub-test (t.Run) part out.

szuecs:
can you check and run the test with `-count=10` and `-count=100` if it does not take >10m. 
```
--- FAIL: TestLoopbackWithClientIP (0.03s)
    teeloopback_test.go:209: failed to receive the right number of requests on the backend; split: 1; shadow: 2
FAIL
```

madumalt:
> can you check and run the test with `-count=10` and `-count=100` if it does not take >10m.
> 
> ```
> --- FAIL: TestLoopbackWithClientIP (0.03s)
>     teeloopback_test.go:209: failed to receive the right number of requests on the backend; split: 1; shadow: 2
> FAIL
> ```

❯ go test -run TestLoopbackWithClientIP -count=100 ./filters/tee
ok  	github.com/zalando/skipper/filters/tee	3.613s

❯ go test -run TestLoopbackWithClientIP -count=10 ./filters/tee
ok  	github.com/zalando/skipper/filters/tee	0.796s

❯ go test -count=100 ./filters/tee
ok  	github.com/zalando/skipper/filters/tee	41.199s

seems like a flakytest, passes when I run it locally.

madumalt:
👍

szuecs:
:+1:

szuecs:
Thank you @madumalt !

PR #3827, opened 2026-01-13
Feature/otel baggage predicate

szuecs:
Feature: OTelBaggage() predicate matches the key of an item

### DIFF
--- a/predicates/otel/otel.go
+++ b/predicates/otel/otel.go
@@ -1,0 +1,44 @@
+package otel
+
+import (
+	"net/http"
+
+	"github.com/zalando/skipper/predicates"
+	"github.com/zalando/skipper/routing"
+	"go.opentelemetry.io/otel/baggage"
+)
+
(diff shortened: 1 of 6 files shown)


szuecs:
:+1:

greeshma1196:
👍

predicates/otel/otel.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Create a predicate instance that always evaluates to baggage
+func (*baggageSpec) Create(args []interface{}) (routing.Predicate, error) {
+	if len(args) != 1 {
+		return nil, predicates.ErrInvalidPredicateParameters
+	}
+
+	var key string
+	if sarg, ok := args[0].(string); ok {
```
MustafaSaber, line 31: I understand this is a starting point, but I think it's better to have multiple values or regex from the ground as it always proved useful

PR #2096, opened 2022-10-12
Allow to selectively disable filters

AlexanderYastrebov:
Signed-off-by: Alexander Yastrebov <alexander.yastrebov@zalando.de>

PR #3830, opened 2026-01-16
security: disable lua scripting by default

szuecs:
There is a weakness that can be abused by a default skipper installation such that you can read arbitrary files as skipper process. It depends on the custom installation and environment if this is actually exploitable by untrusted people. Since 2022 we provide a detailed Lua [config guide](https://opensource.zalando.com/skipper/reference/scripts/#enable-and-disable-lua-sources) such that operators can choose how to use Lua even in less trusted environments. For example you can use -lua-sources=file and only operators that can provide a file accessible to the skipper process are able to reference lua sources and execute provided scripts.

Thanks defang bo providing us a detailed report how to exploit this vulnerability that is available by default in skipper versions <v0.23


AlexanderYastrebov:
Safety first.

AlexanderYastrebov:
👍

szuecs:
@MustafaSaber It feels more complicated that's why I chose to just let everything as it is and use our common strategy to enable/disable features

MustafaSaber (approved):

AlexanderYastrebov:
There is a long history of us trying to make it solid:
* https://github.com/zalando/skipper/pull/2264
* https://github.com/zalando/skipper/pull/2112
* https://github.com/zalando/skipper/pull/2096

but the truth is that [sandboxing Lua](http://lua-users.org/wiki/SandBoxes) is tricky due to its nature so off by default is the right default choice.

PR #3835, opened 2026-01-23
feat(routing): add Print method to RouteLookup for route inspection

MustafaSaber:
Add Print method to RouteLookup struct that returns a formatted string representation of valid routes from the captured routing table. This enables debugging and inspection of the current routing state. Since library users can't access routing table info, this can give nive view only access and to compare against ekip to validate routes.

### DIFF
--- a/routing/routing.go
+++ b/routing/routing.go
@@ -414,6 +414,11 @@ type RouteLookup struct {
 	rt *routeTable
 }
 
+// Print returns valid routes from the captured routing Table.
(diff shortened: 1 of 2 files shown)


greeshma1196:
👍🏽

szuecs:
I don't think that this is needed for the purpose.
However it's also fine to have, at least I don't see an issue merging besides DCO and that I would like to phrase again that we should use less the assert library as it's a hort of bug.

PR #3842, opened 2026-01-26
security: disable kubernetes external name by default

szuecs:
security: disable kubernetes external name by default

This PR breaks end-users and library-users that depend on Kubernetes Ingress referencing a Kubernetes service type ExternalName.

If you are end-user and depend on the former behavior, you can pass `-enable-kubernetes-external-names` to your deployment.
If you are library-user and depend on the former behavior, you can pass the following option to restore behavior:
```go
skipper.Run(skipper.Options{
    EnableExternalNames: true,
})
```

ref: https://github.com/zalando/skipper/security/advisories/GHSA-mxxc-p822-2hx9


a4180p:
👍

szuecs:
:+1:

szuecs:
:+1:

PR #3845, opened 2026-01-27
feat: integrate preprocessors into validation mode

MustafaSaber:
While skipper runs in validation mode, it doesn't apply prepocessors which can result in a valid route being blocked as it have invalid argument before applying preprocessor.

```
r1: SourceFromLast("application") -> <shunt>;
// After preprocessing
r1: * -> SourceFromLast("1.1.1.1") -> <shunt>;
```

This can be test locally via

```bash
~ ./bin/skipper -edit-route=@"application"@"1.1.1.1"@ -validation-webhook-enabled=true -validation-webhook-address=:9085 -validation-webhook-cert-file="validation.crt" -validation-webhook-key-file="validation.key" -enable-advanced-validation=true -application-log-level=debug -kubernetes

~ curl -i -k -X POST -H "Content-Type: application/json" -d @./dataclients/kubernetes/admission/testdata/rg/rg-with-valid-eskip-predicates.json https://127.0.0.1:9085/routegroups
```

PR #3847, opened 2026-01-28
refactor: enhance route validation to handle multi-route preprocessors

MustafaSaber:
This fixes validation issues with preprocessors like OAuth Grant that generate additional routes (e.g., callback routes) which must also be validated.

follow up on https://github.com/zalando/skipper/pull/3845


MustafaSaber:
👍

PR #3851, opened 2026-01-30
fix: implement ConfigDiscoveryHook for PrometheusOverride to enable OPA status metrics

x418:
## What & Why?

Implement ConfigDiscoveryHook.OnConfigDiscovery() for PrometheusOverride to enable Prometheus status metrics when using discovery bundles.

https://github.com/zalando/skipper/pull/3631 added support to also expose native OPA metrics via Skipper's metrics endpoint. This PR extends it by also ensuring that we also enable it when using discovery bundles

_TODO: Extract some reusable methods if this is the way we want to go_


mjungsbluth:
@x418 is this still relevant? Looks like a good approach to capture more metrics that already exist.

PR #3856, opened 2026-02-04
Add routeId to filterContext

xavivars:
This MR exposes the route ID to `filterContext`.

Over the past few years, we've seen many instances where it could have been handy to know the route that was triggered for a specific request: we have predicates that match based on A/B testing, or other input (headers/cookies), and there may be multiple routes that end up in the same origin. 

Being able to track (and potentially log, or apply filters based on) the route ID would be extremely helpful for us.

Slack conversation: https://gophers.slack.com/archives/C82Q5JNH5/p1755799907005649

### DIFF
--- a/filters/filters.go
+++ b/filters/filters.go
@@ -136,7 +136,7 @@ type FilterContext interface {
 
 	Logger() FilterContextLogger
 
-	// Returns the route ID of the currently matched route or an empty string
+	// RouteId Returns the route ID of the currently matched route or an empty string
 	// if no route has been matched yet.
 	RouteId() string
 }


filters/filters.go
```diff
@@ -135,6 +135,10 @@ type FilterContext interface {
 	LoopbackWithResponse()
 
 	Logger() FilterContextLogger
+
+	// Returns the route ID of the currently matched route or an empty string
```
szuecs: Godoc starts with the name of the function.
RouteId...

filters/filters.go
```diff
@@ -135,6 +135,10 @@ type FilterContext interface {
...
+	// if no route has been matched yet.
+	RouteId() string
```
szuecs, line 141: I wonder if we should use ID (used in Go) or stay with Id (used in skipper)

xavivars: I don't have a strong preference...

xavivars: I'll leave it as `Route.Id` for now, as the field in the route is also `Id`. But happy to change it if that's the preference.

PR #981, opened 2019-03-05
Feature/redis based cluster ratelimit

szuecs:
Obsolets https://github.com/zalando/skipper/pull/961 and fix #942 and partially #904.

This PR creates the possibility to choose redis based cluster ratelimits, instead of swim based. The main reason is that swim based cluster ratelimits were not consistent enough, but we would like to work on the consistency in the future.

Redis based cluster ratelimits works with redis ring which is using consistent hashing to find the right shard for a key. The key in this case is the ratelimit group + client data (in case of client ratelimit), for example: `clusterRatelimit("groupA", 100, "10s")` or `clusterClientRatelimit("groupB", 10, "1m")` as documented in https://opensource.zalando.com/skipper/reference/filters/#clusterratelimit and https://opensource.zalando.com/skipper/reference/filters/#clusterclientratelimit.

TODO:
- [x] test redis delete (rolling update) should not cause issues for the redis client. I expect that it will find the redis shard as soon as it is ready again. During this time period rate limit won't happen and all traffic is allowed to pass through.
- [x] redo the poor mans retry `ring.Ping()` check in `newClusterRateLimiterRedis()` with backoff
- [x] review logging
- [x] add and change documentation
- [x] check if there are enough tests

### tests without rolling updates
Tests to show results from laptop to dev cluster in AWS. The goal was to have consistent results with +/- x% where x is as small as possible providing good enough latencies. Latencies with 1000req/s are most likely, caused by the test setup (my local machine). 
```
% for j in 50 50 250 500 1000; do echo "# test $j req/s $(date)"; for i in {1..9}; do echo "GET h
ttps://c3.kiac-test.teapot.zalan.do/" | vegeta attack -rate=$j -duration=10s -http2=false | tee results.bin | vegeta report | grep -E 'Status|Latencies'; done; date; sleep 10; done
# test 50 req/s Tue Mar  5 17:34:55 CET 2019
Latencies     [mean, 50, 95, 99, max]  19.105414ms, 18.511786ms, 20.595503ms, 71.037245ms, 173.501323ms
Status Codes  [code:count]             200:401  429:99
Latencies     [mean, 50, 95, 99, max]  18.233201ms, 18.482938ms, 20.25782ms, 50.624704ms, 96.866689ms
Status Codes  [code:count]             200:403  429:97
Latencies     [mean, 50, 95, 99, max]  18.043883ms, 18.375156ms, 21.042849ms, 29.532214ms, 98.689485ms
Status Codes  [code:count]             429:100  200:400
[...]

PR #3866, opened 2026-02-09
feature: valkey based ratelimit

szuecs:
feature: valkey based cluster ratelimits

The difference between redis and valkey config is based on the fact that go-redis and go-valkey are fundamentally different.
I tested that a local skipper build works with clusterRatelimit() on linux with valkey running
```
% docker run --network host -it valkey/valkey:9-alpine3.23 valkey-server 
```

### DIFF
--- a/ratelimit/cluster_test.go
+++ b/ratelimit/cluster_test.go
@@ -82,6 +82,23 @@ func Test_newClusterRateLimiter(t *testing.T) {
 				TimeWindow: 3 * time.Second,
 			},
 			swarm:      nil,
+			redisRing:  redisRing,
+			valkeyRing: valkeyRing,
+			group:      "mygroup",
+			want: &clusterLimitValkey{
+				group:      "mygroup",
+				maxHits:    10,
+				window:     3 * time.Second,
+				ringClient: valkeyRing,
+			},
+		},
+		{
+			name: "no swarmer, a redis and valkey ring",
+			settings: Settings{
+				MaxHits:    10,
+				TimeWindow: 3 * time.Second,
+			},
+			swarm:      nil,
 			valkeyRing: valkeyRing,
 			group:      "mygroup",
 			want: &clusterLimitValkey{


ratelimit/valkey.go
```diff
@@ -0,0 +1,312 @@
...
+	if span != nil {
+		span.SetTag("allowed", allow)
+	}
+
+	c.measureQuery(valkeyAllowMetricsFormat, valkeyAllowMetricsFormatWithGroup, &failed, now)
+
+	if allow {
+		c.metrics.IncCounter(valkeyMetricsPrefix + "allows")
+	} else {
+		c.metrics.IncCounter(valkeyMetricsPrefix + "forbids")
```
szuecs: @MustafaSaber @ponimas should we change the key "forbids" to "deny" for valkey?

a4180p: `deny` sounds ok, but I would leave the old naming scheme, used with redis.
Also, if we choose `deny` - we should also change `allows` to `allow` for consistency.

ratelimit/cluster_test.go
```diff
@@ -7,18 +7,28 @@ import (
...
 	)
-	defer myring.Close()
+	defer redisRing.Close()
+
+	valkeyRing, err := net.NewValkeyRingClient(&net.ValkeyOptions{
+		Addrs: []string{valkeyAddr},
+	})
+	if err != nil {
+		t.Fatalf("Failed to create valkey ring client: %v", err)
+	}
```
a4180p, line 31:
```suggestion
	}
	defer valkeyRing.Close()
```

ratelimit/registry.go
```diff
@@ -21,22 +21,39 @@ const (
...
+			return reg
+		}
+
+	}
+	return NewRedisSwarmRegistry(swarm, ro, settings...)
+}
+
+// NewRedisSwarmRegistry initializes a registry with an optional swarm and
+// the provided default settings. If swarm is nil, clusterRatelimits
+// will be replaced by voidRatelimit, which is a noop limiter implementation.
```
a4180p: I don't think this comment is true since https://github.com/zalando/skipper/pull/981
IMO, we should split the swim based ratelimit initializer into sesparate function, because rn it's a bit confusing - why does redis registry have swarm and valkey no?

Also the term swarm itself imo is aplicable only to the swim based ratelimiting, but that's another discussion :)

routesrv/valkeyhandler.go
```diff
@@ -0,0 +1,64 @@
...
+	address, err := vh.AddrUpdater()
+	if err != nil {
+		log.Errorf("Failed to update address for valkey handler %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.Write(address)
+}
+
+func getValkeyAddresses(opts *skipper.Options, kdc *kubernetes.Client, m metrics.Metrics) func() ([]byte, error) {
```
a4180p, line 43: Do I get the idea right that we're relying on Kubernetes liveness probes for valkey health checks, which is why we don't ping shards ourselves?
Doesn't that make the valkey-based rate limiter usable only in k8s?

ratelimit/valkey.go
szuecs: ok then I let it like this.

ratelimit/registry.go
szuecs: yeah there was the idea to have a Swarmer interface that can be used for every implementation but if you first specify the interface and then see what the implementations look like, then your abstractions are broken.
We should delete the "swim" implementation. We were never being able to successfully set this up such that the rate limit has results in acceptable bounds.

routesrv/valkeyhandler.go
szuecs: yeah maybe we should do the valkey.PingAll() to  get updates of "reachable" shards.

szuecs: However this is `routesrv` which has anyways only the kubernetes dataclient in use.

ratelimit/registry.go
szuecs: I refactored the code

routesrv/valkeyhandler.go
szuecs: Apropos liveness vs readiness  probe is an interesting detail. We rely on the readinessProbe, which we also do for all endpointslices (pods)

a4180p:
👍

ratelimit/cluster.go
```diff
@@ -9,19 +9,25 @@ const (
...
+// group that can span one or multiple routes.
+func newClusterRateLimiter(s Settings, sw Swarmer, redisRing *net.RedisRingClient, valkeyRing *net.ValkeyRingClient, group string) limiter {
 	if sw != nil {
 		if l := newClusterRateLimiterSwim(s, sw, group); l != nil {
 			return l
 		}
 	}
-	if ring != nil {
-		if l := newClusterRateLimiterRedis(s, ring, group); l != nil {
+	if valkeyRing != nil {
```
MustafaSaber, line 24: minor: maybe we should log somewhere that if user configure have both Rings then valkey will always be favoured

ratelimit/cluster_test.go
```diff
@@ -32,49 +43,66 @@ func Test_newClusterRateLimiter(t *testing.T) {
...
+			swarm:     fake,
+			redisRing: nil,
+			group:     "mygroup",
+			want:      newClusterRateLimiterSwim(settings, fake, "mygroup"),
 		}} {
 		t.Run(tt.name, func(t *testing.T) {
 			defer tt.want.Close()
 
-			got := newClusterRateLimiter(tt.settings, tt.swarm, tt.ring, tt.group)
+			got := newClusterRateLimiter(tt.settings, tt.swarm, tt.redisRing, tt.valkeyRing, tt.group)
```
MustafaSaber, line 122: would add another testcase with providing both rings to document behaviour also

routesrv/testdata/valkey-endpointslice-single.kube
```diff
@@ -0,0 +1 @@
+enable-kubernetes-endpointslices: true
```
MustafaSaber: minor: new line

routesrv/routesrv.go
```diff
@@ -116,6 +116,21 @@ func New(opts skipper.Options) (*RouteServer, error) {
...
+	if opts.KubernetesValkeyServiceNamespace != "" && opts.KubernetesValkeyServiceName != "" {
+		log.Infof("Use endpoints %s/%s to fetch updated valkey shards", opts.KubernetesValkeyServiceNamespace, opts.KubernetesValkeyServiceName)
+		vh = &ValkeyHandler{}
+		_, err := dataclient.LoadAll()
+		if err != nil {
+			return nil, err
+		}
+		vh.AddrUpdater = getValkeyAddresses(&opts, dataclient, m)
+		mux.Handle("/swarm/valkey/shards", vh)
+		println("HERE")
```
MustafaSaber: leftover from debugging

routesrv/routesrv.go
```diff
@@ -116,6 +116,21 @@ func New(opts skipper.Options) (*RouteServer, error) {
...
+	}
+	println("HERE 2")
```
MustafaSaber: Also leftover

routesrv/valkeyhandler.go
```diff
@@ -0,0 +1,64 @@
...
+	return func() ([]byte, error) {
+		a := kdc.GetEndpointAddresses("", opts.KubernetesValkeyServiceNamespace, opts.KubernetesValkeyServiceName)
+		log.Infof("Valkey updater called and found %d valkey endpoints: %v", len(a), a) //TODO(sszuecs): Debugf
```
MustafaSaber: Maybe move to Debugf now?

skipper.go
```diff
@@ -1980,7 +2051,7 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 			ratelimitfilters.NewBackendRatelimit(),
 		)
 
-		if redisOptions != nil {
+		if redisOptions != nil || valkeyOptions != nil {
```
MustafaSaber, line 2054: here if both are supplied we prefer redisOptions, unlike everywhere else where we prefer valkey

valkey_test.go
```diff
@@ -0,0 +1,586 @@
...
+
+	successRate := va.Success()
+	t.Logf("Success [0..1]: %0.2f", successRate)
+	t.Logf("Want: %0.2f", float64(sec*1)/float64(sec*rate))
+
+	epsilon := 0.2
+	// sec * 1 because 1 is the number of requests allowed per second via clusterRatelimit("foo", 1, "1s")
+	expected := float64(sec*1) / float64(sec*rate)
+	// https://github.com/stretchr/testify/issues/1839
+	// if !assert.InEpsilon(t, expected, successRate, epsilon, fmt.Sprintf("Test should have a success rate between %0.2f < %0.2f < %0.2f", expected-epsilon, successRate, expected+epsilon)) {
```
MustafaSaber, line 188: maybe remove this

ratelimit/cluster.go
szuecs: Skipper already logs which implementation is used:

Example
```
% ./bin/skipper -inline-routes='route1: * -> clusterRatelimit("foo", 5, "1m") -> <shunt>' -enable-ratelimits -enable-swarm
[APP]INFO[0000] enable swarm: true                           
[APP]INFO[0000] Start swim based swarm                       

% ./bin/skipper -inline-routes='route1: * -> clusterRatelimit("foo", 5, "1m") -> <shunt>' -enable-ratelimits -enable-swarm -swarm-valkey-urls localhost:6379  feature/valkey-ratelimit
[APP]INFO[0000] enable swarm: true                           
[APP]INFO[0000] Valkey based swarm with 1 shards          

% ./bin/skipper -inline-routes='route1: * -> clusterRatelimit("foo", 5, "1m") -> <shunt>' -enable-ratelimits -enable-swarm -swarm-redis-urls localhost:6379   feature/valkey-ratelimit
[APP]INFO[0000] enable swarm: true                           
[APP]INFO[0000] Redis based swarm with 1 shards            
```

routesrv/routesrv.go
szuecs: oh man good point I dropped even more println everywhere

valkey_test.go
szuecs: It's a copy of redis test and we have there the reference, so I would let the reference. Maybe one day they fix this and we can have a more nice assert use. At least that was the idea.

skipper.go
szuecs: But this is only because of the Lua script that is passed to valkey/redis. In this case there is no preference, code segment is and does not pass any options:
```
		if redisOptions != nil || valkeyOptions != nil {
			o.CustomFilters = append(o.CustomFilters, ratelimitfilters.NewClusterLeakyBucketRatelimit(ratelimitRegistry))
		}
```

routesrv/testdata/valkey-endpointslice-single.kube
szuecs: done

ratelimit/cluster_test.go
szuecs: done

szuecs:
:+1:

PR #3868, opened 2026-02-10
[filter/inlineContentIfStatus] Clear content headers

ponimas:
We should remove Content-Length and Content-Encoding headers  to prevent mismatches between the original backend response and the inline content.


szuecs:
@ponimas can you check the docs/reference/filters.md if this needs also a change?

a4180p:
👍

Issue #931, opened 2019-01-23
Kubernetes: update Ingress status field


ch-dimitri:
When using a `LoadBalancer` Service to provision the frontend to Skipper, it should update the `status` field of Ingress objects assigned to that ingress-class.  ingress-nginx does [this](https://github.com/kubernetes/ingress-nginx/blob/master/internal/ingress/status/status.go#L349-L372).

Skipper should do something similar for cases that don't use the [AWS Ingress Controller|https://github.com/zalando-incubator/kube-ingress-aws-controller], but instead rely upon the native cloud provider.

Example config:
```
kind: Service
apiVersion: v1
metadata:
  name: skipper-ingress
  namespace: kube-system
  labels:
    app.kubernetes.io/name: skipper-ingress
  annotations:
   service.beta.kubernetes.io/aws-load-balancer-type: nlb                                                                                                                                                                                                                                                                                                                 
    service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
spec:
  externalTrafficPolicy: Local
  type: LoadBalancer
  selector:
    application: skipper-ingress
  ports:
  - name: http
    port: 80
    targetPort: http
  - name: https
    port: 443
    targetPort: https
```

szuecs:
Sounds like a good enhancement. Would you like to create a pull request for that?

ch-dimitri:
I'd love to, but have to go through a legal approval process first.

ch-dimitri:
Sure, @szuecs!  Do you want to assign this to me?

PR #3877, opened 2026-02-19
kubernetes: update ingress status from service

lambertpandian:
Add support for -kubernetes-ingress-status-from-service so Skipper can source ingress status addresses from a configured Service and patch ingress /status when needed.

Why: Skipper does not currently provide this capability, and this adds a way to publish Service-derived ingress addresses through ingress status updates.

Refs #931

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -203,7 +203,7 @@ type Config struct {
 	KubernetesBackendTrafficAlgorithm                    kubernetes.BackendTrafficAlgorithm `yaml:"-"`
 	KubernetesDefaultLoadBalancerAlgorithm               string                             `yaml:"kubernetes-default-lb-algorithm"`
 	KubernetesForceService                               bool                               `yaml:"kubernetes-force-service"`
-	KubernetesIngressStatusFromService                   string                             `yaml:"kubernetes-ingress-status-from-service"`
+	KubernetesStatusFromService                          string                             `yaml:"kubernetes-ingress-status-from-service"`
 
 	// Default filters
 	DefaultFiltersDir string `yaml:"default-filters-dir"`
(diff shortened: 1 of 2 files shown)


szuecs:
Can you please --signoff your commits (required for DCO check)?

You need also to add documentation and tell about the RBAC requirement, because before we did not have write access, bt now we need to write the subresource ingress/status.

lambertpandian:
@szuecs 
I have added the --signoff & updated the doc. Thanks.

szuecs:
:+1:

config/config.go
```diff
@@ -203,6 +203,7 @@ type Config struct {
 	KubernetesBackendTrafficAlgorithm                    kubernetes.BackendTrafficAlgorithm `yaml:"-"`
 	KubernetesDefaultLoadBalancerAlgorithm               string                             `yaml:"kubernetes-default-lb-algorithm"`
 	KubernetesForceService                               bool                               `yaml:"kubernetes-force-service"`
+	KubernetesIngressStatusFromService                   string                             `yaml:"kubernetes-ingress-status-from-service"`
```
MustafaSaber: Since we also have RouteGroup, I would recommend dropping the Ingress part from the variable and elsewhere, so we don't have to rename the flag in the follow up PR.

Either we include the RouteGroup implementation in the PR or we will follow up with another PR after this one.

lambertpandian: Done 👍 I’ve removed “Ingress” from the variable.

szuecs:
:+1:

szuecs:
@lambertpandian in case `check-race` fails, please fetch origin/master, rebase master and push -f to fix it.
There was a data race in a dependency that we fixed an until release we apply a temporary fix, that was merged today.

greeshma1196:
👍🏽

config/config.go
```diff
@@ -203,6 +203,7 @@ type Config struct {
 	KubernetesBackendTrafficAlgorithm                    kubernetes.BackendTrafficAlgorithm `yaml:"-"`
 	KubernetesDefaultLoadBalancerAlgorithm               string                             `yaml:"kubernetes-default-lb-algorithm"`
 	KubernetesForceService                               bool                               `yaml:"kubernetes-force-service"`
+	KubernetesStatusFromService                          string                             `yaml:"kubernetes-ingress-status-from-service"`
```
szuecs, line 206: we forgot to change the config to not have ingress in the yaml config

config/config.go
```diff
@@ -548,6 +549,7 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.KubernetesBackendTrafficAlgorithmString, "kubernetes-backend-traffic-algorithm", kubernetes.TrafficPredicateAlgorithm.String(), "sets the algorithm to be used for traffic splitting between backends: traffic-predicate or traffic-segment-predicate")
 	flag.StringVar(&cfg.KubernetesDefaultLoadBalancerAlgorithm, "kubernetes-default-lb-algorithm", kubernetes.DefaultLoadBalancerAlgorithm, "sets the default algorithm to be used for load balancing between backend endpoints, available options: roundRobin, consistentHash, random, powerOfRandomNChoices")
 	flag.BoolVar(&cfg.KubernetesForceService, "kubernetes-force-service", false, "overrides default Skipper functionality and routes traffic using Kubernetes Services instead of Endpoints")
+	flag.StringVar(&cfg.KubernetesStatusFromService, "kubernetes-ingress-status-from-service", "", "when set to <namespace>/<name>, updates ingress status.loadBalancer.ingress from the referenced service")
```
szuecs, line 552: we forgot to remove the ingress name in the flag.

PR #3878, opened 2026-02-21
feature: add auto for X-Forwarded-Proto and X-FOrwarded-Port

### LINKED ISSUE #3871: Improve X-Forwarded-Proto to handle insecure-address config option
X-Forwarded-Proto and X-Forwarded-Port require a value to be set on the command line.
These values could change if your using address and insecure-address.

It would be ideal, if you set X-Forwarded-Proto without a value (=http/=https) it would set http or https depending on if it came in the address port or the insecure-address port, if insecure-address is not default, likely error, cause a value would be required in that case.

The same goes for X-Forwarded-Port, if not defined, it should be set to the port the connection comes in on, but it address or insecure-address

patrickdk77:
closes #3871

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -502,8 +502,8 @@ func NewConfig() *Config {
 		"X-Forwarded-Host sets X-Forwarded-Host value to the request host\n"+
 		"X-Forwarded-Method sets X-Forwarded-Method value to the request method\n"+
 		"X-Forwarded-Uri sets X-Forwarded-Uri value to the requestURI\n"+
-		"X-Forwarded-Port=<port> sets X-Forwarded-Port value\n"+
-		"X-Forwarded-Proto=<http|https> sets X-Forwarded-Proto value")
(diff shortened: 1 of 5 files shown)


szuecs:
@patrickdk77 can you --sign-off your commits to fix DCO check?

szuecs:
👍

szuecs:
@patrickdk77 please fetch origin/master and rebase on master and push -f.
There is a dependency issue that create a data race inside the dependency and I made a fix such that we can merge open PRs.

patrickdk77:
pushing the rebase, but locally here I see no change to the make check errors.

szuecs:
> pushing the rebase, but locally here I see no change to the make check errors.

thanks!
Yes it should not be your problem. It was an OPA dependency that was updated some says ago, but had a race condition, that we could fix in OPA (was merged) and we could unblock the current `pr / check-race` errors that found the race condition.

szuecs:
:+1:

PR #3884, opened 2026-02-25
fix: config and flag for kubernetes status from service

szuecs:
fix: config and flag for kubernetes status from service

We (@MustafaSaber , @lambertpandian and me) agreed to drop `ingress` from the config and flag and skipper.Option. This fixes docs, config and flag according to our agreement

ref: https://github.com/zalando/skipper/releases/tag/v0.24.29
ref: https://github.com/zalando/skipper/pull/3877


szuecs:
:+1:

lambertpandian (approved):
Thanks for the PR.
Looks good to me.

PR #3888, opened 2026-02-28
fix: race condition while updating shards storage

szuecs:
fix: race condition while updating shards storage

Writing the array, while reading created sometimes a panic with data race in tests.
Use atomic to swap the fields and load to get get the value safely.

```
% go test -race . -run TestConcurrentKubernetesClusterStateAccessValkey -count 20
ok      github.com/zalando/skipper      170.118s
```

### DIFF
--- a/net/valkey.go
+++ b/net/valkey.go
@@ -6,6 +6,7 @@ import (
 	"math"
 	"strings"
 	"sync"
+	"sync/atomic"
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

MustafaSaber (approved):

PR #3897, opened 2026-03-04
feature: expose invalid routes via support listener API

### LINKED ISSUE #3488: List invalid routes

In multitenant ingress setting users may define invalid routes and operators may have to figure out where invalid routes are coming from. Currently it is possible via monitoring logs for a [specific error message](https://github.com/zalando/skipper/blob/d05b0409b27a696e981aefa49a9a41bfc77ca4ad/routing/datasource.go#L514).

Skipper [serves valid routes](https://github.com/zalando/skipper/blob/d05b0409b27a696e981aefa49a9a41bfc77ca4ad/routing/routing.go#L279-L338) via support listener API at /routes.
It also have [a list of invalid routes](https://github.com/zalando/skipper/blob/d05b0409b27a696e981aefa49a9a41bfc77ca4ad/routing/datasource.go#L513-L514) so we may add a support API say at /invalid-routes to show them as well.
Along with a new metric for number of invalid routes this could be used for alerting and debugging invalid configurations.

Monitor logs for a specific error message.

The problem is partially addressed by a webhook which prevents users from creating routes with e.g. invalid eskip syntax but it does not (currently) prevent users from creating non-existing or disabled filters.

This is a good first issue for someone :)

bakabaka91:
Fixes #3488

### Context

Operators in multitenant ingress setups currently have no programmatic way to discover which routes failed validation — the only option is grepping logs for `"failed to process route"` messages. The `routeTable` already tracks `invalidRoutes []*eskip.Route` alongside `validRoutes`, but `Routing.ServeHTTP` unconditionally serves only the valid set. Invalid routes are silently dropped from the routing table with no API visibility into what broke or why.

### Approach

This extends the existing `/routes` handler with a `?invalid=true` query parameter rather than adding a separate endpoint. The query param approach follows @szuecs's guidance in the issue — invalid routes are a filtered view of the same data, not a distinct resource, and this avoids new mux registrations or a parallel `ServeHTTP` implementation.

The harder design decision was how to expose error messages (per @AlexanderYastrebov's suggestion). The validation errors are computed in `processRouteDefs` and `newMatcher` but were previously only sent to metrics and discarded. Rather than adding an error field to `eskip.Route` (a core type used across the entire codebase for parsing, serialization, and matching), the errors are stored in a separate `invalidRouteErrors map[string]string` on `routeTable` and merged into the JSON response via a custom `MarshalJSON` on `InvalidRouteResponse`.

### Changes

- `processRouteDefs` returns a third value — error messages keyed by route ID — collected from filter, predicate, backend, and matcher validation failures
- `routeTable` gains an `invalidRouteErrors` field, populated from both `processRouteDefs` and `newMatcher` error paths in `receiveRouteMatcher`
- `Routing.ServeHTTP` switches between `validRoutes` and `invalidRoutes` based on the `invalid` query param; all existing behavior (pagination, content negotiation, `X-Count`/`X-Timestamp` headers) works identically for both sets
- `InvalidRouteResponse` wraps a route with its error string, using custom JSON marshaling to merge the error into the route's existing JSON structure (necessary because `eskip.Route` has its own `MarshalJSON`)
- Documentation added to the "Routing table information" section of `operation.md`

### Testing

Five new tests cover the feature surface: JSON response with error field validation, eskip text format output, HEAD request with correct count, pagination with offset/limit, and backward compatibility (no param, `?invalid=false`, and `?invalid=` all return valid routes). All eleven `TestRoutingHandler*` tests pass, including the six pre-existing ones — no regressions.

### Notes for Reviewer

- The `InvalidRouteResponse.MarshalJSON` round-trips through `json.Marshal` → `map[string]json.RawMessage` → inject error → `json.Marshal`. This is slightly less efficient than direct byte manipulation but avoids fragile assumptions about `eskip.Route`'s JSON output format. The cost is negligible since this endpoint is only hit by operators debugging, not in the request path.
- The issue mentions adding "a new metric for number of invalid routes" — individual route metrics already exist via `SetInvalidRoute` and the gauge tracking added in #3550/#3610, so this PR focuses solely on the API surface.
- Scope excludes routesrv (`cmd/routesrv`), which has its own route handler and would need separate treatment.

### DIFF
--- a/routing/routing_test.go
+++ b/routing/routing_test.go
@@ -1082,7 +1082,10 @@ func TestRoutingHandlerInvalidRoutesJSON(t *testing.T) {
 	// With ?invalid=true: should return only invalid routes with error
 	req, _ = http.NewRequest("GET", server.URL+"?invalid=true", nil)
 	req.Header.Set("accept", "application/json")
-	resp, _ = http.DefaultClient.Do(req)
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		t.Fatalf("failed to request invalid routes: %v", err)
+	}
 	defer resp.Body.Close()
 
 	if resp.Header.Get(routing.RoutesCountName) != "1" {
@@ -1130,7 +1133,10 @@ func TestRoutingHandlerInvalidRoutesText(t *testing.T) {
 	defer server.Close()
 
 	req, _ := http.NewRequest("GET", server.URL+"?invalid=true", nil)
-	resp, _ := http.DefaultClient.Do(req)
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		t.Fatalf("failed to request invalid routes: %v", err)
+	}
 	defer resp.Body.Close()
 
 	if got, want := resp.Header.Get("content-type"), "text/plain"; got != want {
@@ -1169,7 +1175,10 @@ func TestRoutingHandlerInvalidRoutesHEAD(t *testing.T) {
 	defer server.Close()
 
 	req, _ := http.NewRequest("HEAD", server.URL+"?invalid=true", nil)
-	resp, _ := http.DefaultClient.Do(req)
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		t.Fatalf("failed to request invalid routes: %v", err)
+	}
 	defer resp.Body.Close()
 
 	if resp.Header.Get(routing.RoutesCountName) != "1" {


bakabaka91:
@szuecs Thanks for the approval\! Could you please trigger the CI pipeline? It's blocked for external contributors.

bakabaka91:
@szuecs The CI `tests` job failed on `make vet` — three instances of using `resp` before checking for errors in the test file (`routing/routing_test.go`). I have a fix ready (just adding proper error checks before using `resp`).

The `check-race` failure is a pre-existing flaky test (`TestAdmissionControl/medium_error_rate_and_bigger_than_threshold_will_block_some_traffic` in `filters/shedder`) — not related to this PR.

Would it be okay to push the vet fix, or would that require re-approval?

szuecs:
@bakabaka91 tests fail:
```
Error: routing/routing_test.go:1086:8: using resp before checking for errors
Error: routing/routing_test.go:1134:8: using resp before checking for errors
Error: routing/routing_test.go:1173:8: using resp before checking for errors
```

bakabaka91:
@szuecs Pushed the fix for the `go vet` errors — added proper error checks before using `resp` in the three test functions (`TestRoutingHandlerInvalidRoutesJSON`, `TestRoutingHandlerInvalidRoutesText`, `TestRoutingHandlerInvalidRoutesHEAD`).

It looks like the push dismissed your earlier approval. Could you re-approve and trigger the pipeline when you get a chance? The only change from the previous version is this minor test fix.

szuecs:
@bakabaka91 can you please first sign-off your commit? and force push again?
Thanks!

bakabaka91:
@szuecs Done — rebased and signed off both commits with the correct identity. Ready for re-approval and pipeline trigger whenever you get a chance.

szuecs:
:+1:

greeshma1196:
👍🏽

MustafaSaber:
Thanks for your contribution @bakabaka91 !

PR #3900, opened 2026-03-06
OPA: add HTTP send tracing tests for authorization checks, fix nil pointer error

mjungsbluth:
When a Rego policy was using `http.send` and also tracing was enabled, the internally used RoundTrip could be nil, raising an error during evaluation. 

This PR adds a test to replicate the behavior and fixes the issue.

### DIFF
--- a/filters/openpolicyagent/tracing.go
+++ b/filters/openpolicyagent/tracing.go
@@ -49,6 +49,11 @@ func WithTracingOptManager(manager *plugins.Manager) func(*transport) {
 func (*tracingFactory) NewTransport(tr http.RoundTripper, opts opatracing.Options) http.RoundTripper {
 	log := &logging.DefaultLog{}
 
+	// If the underlying transport is nil, use the default transport
+	if tr == nil {
+		tr = http.DefaultTransport
+	}
+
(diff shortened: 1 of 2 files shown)


filters/openpolicyagent/tracing.go
```diff
@@ -49,6 +49,11 @@ func WithTracingOptManager(manager *plugins.Manager) func(*transport) {
 func (*tracingFactory) NewTransport(tr http.RoundTripper, opts opatracing.Options) http.RoundTripper {
 	log := &logging.DefaultLog{}
 
+	// If the underlying transport is nil, use the default transport
+	if tr == nil {
```
szuecs, line 53: Do you  this is  a  missconfiguration, or does it happen on normal configuration?
If miss configuration I would rather fail fast, like a fatal log with a message what was missing

mjungsbluth: This is sadly the normal behavior for this call path (when the http call is triggered from within a policy), for the the other path (for example comms with S3) we have a non-nil value.

szuecs:
:+1:

mjungsbluth:
👍

PR #3903, opened 2026-03-06
Fix: OTel event to attribute

szuecs:
fix: stream_Headers log should be a tag if we switch log to tags
 fix: streamBody.byte log should be a tag if we switch log to tags
 refactor: stream line tracing.go
 refactor: Id to ID


szuecs:
:+1:

greeshma1196:
👍🏽

PR #3911, opened 2026-03-09
Downgrade opa version to last stable

wisinghe:
Downgrade opa version to last stable.

Later need to update to the new version with this fix https://github.com/open-policy-agent/opa/pull/8341


mikkeloscar:
:+1:

wisinghe:
👍

PR #3915, opened 2026-03-09
fix: panic on concurrent access of statebag

szuecs:
httptrace seems to be called concurrently so we can not use unprotected statebag to pass around the time.Time. Add a new member to the context


szuecs:
:+1:

ayruslore:
👍

szuecs:
:+1:

PR #3920, opened 2026-03-14
fix: panic on auto propagator

szuecs:
fix: panic on auto propagator
panic: duplicate registration: "skipper-debug" [recovered, repanicked]

test: add test coverage to otel package


szuecs:
:+1:

PR #3922, opened 2026-03-15
feat: add jwtValidationKeys filter for JWT validation with direct JWKS URL

### LINKED ISSUE #3921: Add JWT verification filter with direct JWKS URL and claims validation
## Summary

The current `jwtValidation` filter discovers JWKS keys exclusively via `.well-known/openid-configuration` and only validates the `sub` claim. This makes it impossible to verify JWTs from services that don't expose a standard OIDC discovery endpoint or that require `iss`/`aud` validation.

A concrete use case is **Google Chat bot webhooks**, where requests are signed with a JWT that must be verified against a non-OIDC JWKS endpoint with specific claim requirements:

- JWKS URL: `https://www.googleapis.com/service_accounts/v1/jwk/chat@system.gserviceaccount.com`
- `iss` must equal `chat@system.gserviceaccount.com`
- `aud` must equal the Google Cloud project number

## Proposal

A new filter (e.g. `jwtDirect`) that accepts a JWKS URL directly and validates arbitrary claim key-value pairs:

```
jwtDirect(
    "https://www.googleapis.com/service_accounts/v1/jwk/chat@system.gserviceaccount.com",
    "iss", "chat@system.gserviceaccount.com",
    "aud", "123456789"
)
```

### Behavior

- First argument: JWKS URL (fetched and cached, reusing the existing `keyfunc` library)
- Remaining arguments: key-value pairs for claim validation (all must match)
- Returns 401 on missing/invalid token or failed claim validation
- Stores validated claims in StateBag under `oidcclaimscachekey` for downstream filters (e.g. `forwardToken`, `forwardTokenField`)

### Why not extend `jwtValidation`?

Changing the argument semantics of the existing filter would be a breaking change. A separate filter keeps the existing interface stable and makes the intent explicit.

## Use cases

- Google Chat bot webhook verification
- Firebase/GCP service account JWT verification
- Any service that publishes a JWKS endpoint without full OIDC discovery
- Scenarios requiring `iss`/`aud` validation at the proxy layer

ivan-digital:
## Summary

- Add new `jwtValidationKeys` filter that verifies JWT Bearer tokens using a JWKS URL directly, without requiring OIDC discovery via `.well-known/openid-configuration`
- Reuses existing `jwtValidationFilter` — the new spec only provides an alternative entry point that skips OIDC discovery
- Claims validation delegated to `oidcClaimsQuery` as per existing convention
- Registered alongside `jwtValidation` in skipper.go

## Motivation

The existing `jwtValidation` filter only supports JWKS discovery via `.well-known/openid-configuration`. Services like Google Chat bots sign webhook requests with JWTs but publish their public keys at non-standard JWKS endpoints without OIDC discovery support, making it impossible to verify these tokens with the current filter.

## Usage

```
jwtValidationKeys("https://www.googleapis.com/service_accounts/v1/jwk/chat@system.gserviceaccount.com")
-> oidcClaimsQuery("/:@_:iss==\"chat@system.gserviceaccount.com\"")
-> oidcClaimsQuery("/:@_:aud==\"123456789\"")
```

Closes #3921

## Test plan

- [x] Spec validation (missing args, too many args, non-string args)
- [x] Valid token, expired token, missing sub claim
- [x] Missing/empty/malformed Bearer tokens
- [x] Algorithm none rejected
- [x] Existing jwtValidation tests still pass

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -1686,6 +1686,7 @@ jwtValidation("https://login.microsoftonline.com/{tenantId}/v2.0")
 The filter works like [jwtValidation](#jwtvalidation) but takes a JWKS URL directly instead of
 discovering it via `/.well-known/openid-configuration`. This is useful for services that publish
 JWKS keys at non-standard endpoints, such as Google Chat service accounts.
+Unlike `jwtValidation`, the `sub` claim is not required — tokens without `sub` are accepted.
 
 The filter stores token claims into the state bag where they can be used by [oidcClaimsQuery](#oidcclaimsquery), [forwardToken](#forwardtoken) or [forwardTokenField](#forwardtokenfield) filters.
 


filters/auth/jwt_direct.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil, err
+	}
+
+	return &jwtDirectFilter{
+		jwksURL: jwksURL,
+		claims:  claims,
+	}, nil
+}
+
+func (f *jwtDirectFilter) Request(ctx filters.FilterContext) {
```
AlexanderYastrebov: Hi. Existing `jwtValidation` filter only validates token and delegates claims check to `oidcClaimsQuery`.
I think it makes sense to do the same here, moreover Request logic seems to be exactly the same and therefore could be reused - you can simply return an instance of jwtValidationFilter from the spec.
As for the name I'd suggest `jwtValidationKeys` as "direct" is a bit obscure.
Please also add docs to filters.md nearby jwtValidation. Thanks!

(PS: I am not actively involved in Skipper development anymore but I recently saw your great work on ASR/TTS so decided to chime in 👋)

filters/builtin/builtin.go
```diff
@@ -232,6 +232,7 @@ func Filters() []filters.Spec {
 		accesslog.NewEnableAccessLog(),
 		auth.NewForwardToken(),
 		auth.NewForwardTokenField(),
+		auth.NewJwtDirect(),
```
AlexanderYastrebov: Register it nearby existing jwtValidation filter instead of builtin.

filters/auth/jwt_direct.go
ivan-digital: Thanks for the review and the valuable simplifications! Addressed all points: renamed to `jwtValidationKeys`, reusing `jwtValidationFilter` from the spec, moved registration next to `jwtValidation` in skipper.go, and added docs to filters.md. Claims validation delegated to `oidcClaimsQuery` as suggested. Thanks for the attention :)

filters/builtin/builtin.go
ivan-digital: Done, moved to skipper.go next to `jwtValidation`.

filters/auth/jwt_direct.go
szuecs: @AlexanderYastrebov it's always a great contribution if you review code, much appreciated :)

szuecs:
code looks fine I would move it to the same file as the implementation that you borrow from or rename to filter name

szuecs:
:+1:

szuecs:
:+1:

PR #3934, opened 2026-03-25
feat: Add print statements from Open Policy Agent to spans

### LINKED ISSUE #3304: OPA: Enable support for print() from Reno
OPA allows to `print()` information from within the policy. This can be helpful to for example output conditions that led to a token being invalid and can cut down debugging times considerably. 

Since we already have support for sending open tracing Spans for the OPA filters, it would be helpful to include the `print()` output that happened as part of an evaluation in the Span's events via the `LogKV` method.

The envoy plugin logs the print output which looses context to the decision made. 

This requires an upgrade to OPA v0.70 as the corresponding envoy plugin was changed to allow configuring a print hook. There is a [test](https://github.com/open-policy-agent/opa-envoy-plugin/blob/bf7dca6d8626cae3bf39834aabc518ff0337adbe/envoyauth/evaluation_test.go#L197) that demonstrates using such a hook. 

Yes

mjungsbluth:
## Summary

Closes #3304

OPA policies can use `print()` for diagnostic output, but until now this output was only logged to OPA's internal logger — completely disconnected from request tracing. This PR captures `print()` output during policy evaluation and emits it as events on the active OpenTracing span, enabling contextual, per-request debugging of policy logic.

The feature is opt-in via a new flag:

```
--enable-open-policy-agent-print-tracing
```

When enabled, each `print()` call in a Rego policy produces a span log event with the following fields:
- `event`: `"print"`
- `opa.print.location`: source location of the `print()` call (e.g. `main.rego:10`)
- `message`: the printed string

## Implementation notes

Two changes are required to make this work end-to-end:

1. **Manager level** (`plugins.EnablePrintStatements`): the OPA compiler silently drops `print()` calls unless explicitly enabled at manager creation time. This flag is now passed conditionally based on `--enable-open-policy-agent-print-tracing`.

2. **Eval level** (`rego.EvalPrintHook`): a custom `spanPrintHook` is registered per evaluation that routes `print()` output to the active OpenTracing span via `LogKV`. It overrides the default opa-envoy-plugin hook (which logs to OPA's internal logger) since `envoyauth.Eval` appends caller-supplied options after its defaults, and the last `EvalPrintHook` wins.

Both changes are no-ops when the flag is disabled, so there is no overhead on existing deployments.

## Test plan

- [x] New test `TestPrintTracing` in `filters/openpolicyagent/openpolicyagent_test.go` verifies:
  - With `--enable-open-policy-agent-print-tracing`: `print()` output appears as span log events with the correct fields
  - Without the flag (default): span logs are empty — no overhead, no leakage

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -340,6 +340,7 @@ type Config struct {
 	OpenPolicyAgentControlLoopMaxJitter                time.Duration `yaml:"open-policy-agent-control-loop-max-jitter"`
 	EnableOpenPolicyAgentDataPreProcessingOptimization bool          `yaml:"enable-open-policy-agent-data-preprocessing-optimization"`
 	EnableOpenPolicyAgentPreloading                    bool          `yaml:"enable-open-policy-agent-preloading"`
+	EnableOpenPolicyAgentPrintTracing                  bool          `yaml:"enable-open-policy-agent-print-tracing"`
 	OpenPolicyAgentConfigTemplate                      string        `yaml:"open-policy-agent-config-template"`
 	OpenPolicyAgentEnvoyMetadata                       string        `yaml:"open-policy-agent-envoy-metadata"`
 	OpenPolicyAgentCleanerInterval                     time.Duration `yaml:"open-policy-agent-cleaner-interval"`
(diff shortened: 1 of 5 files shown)


mjungsbluth:
👍

szuecs:
docs are missing, @mjungsbluth / @mefarazath please provide a follow up PR to add docs/

PR #3935, opened 2026-03-25
fix: valkey updater calculated wrong difference for update required detection

szuecs:
fix: valkey updater calculated wrong difference for update required detection

Problem showed up after a while when valkey instances were cycled and in the unlucky ordering of valkey IPs we do not detect that we need to update the instance list

### DIFF
--- a/net/valkey.go
+++ b/net/valkey.go
@@ -351,7 +351,6 @@ func NewValkeyRingClient(opt *ValkeyOptions) (*ValkeyRingClient, error) {
 		}
 		if err != nil {
 			opt.Log.Errorf("Failed start valkey client: %v", err)
-			return nil, fmt.Errorf("failed start valkey client: %v", err)
 		} else {
 			opt.Addrs = address
 		}


net/valkey.go
```diff
@@ -413,10 +414,10 @@ func (vrc *ValkeyRingClient) startUpdater(ctx context.Context) {
 
 		addrs, err := vrc.options.AddrUpdater()
 		if err != nil {
-			vrc.log.Errorf("Failed to run valkey updater: %v", err)
+			vrc.log.Errorf("Failed to update valkey ring: %v", err)
 			continue
 		}
-		if !init && len(difference(addrs, old)) != 0 {
+		if !init && (len(difference(addrs, old)) != 0 || len(difference(old, addrs)) != 0) {
```
greeshma1196: If we can sort the addresses, we can directly use slices.Equal to do the check.

szuecs: done

greeshma1196:
👍🏽

szuecs:
:+1:

net/valkey_test.go
```diff
@@ -15,6 +15,46 @@ import (
...
+			want: []string{},
+		},
+		{
+			name: "test a != b",
+			a:    []string{"1s"},
+			b:    []string{"2s"},
+			want: []string{"1s"},
+		}} {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := difference(tt.a, tt.b); len(got) != len(tt.want) {
```
MustafaSaber, line 50: Is the function still in use?

szuecs: Yes in some other part which needs set operations like it is.

PR #3940, opened 2026-03-27
Extend predicate OTelBaggage to match with value

greeshma1196:
## Context

Currently the predicate OTelBaggage only matches based on key. 

This needs to be extended to match based on value if the value is not empty.

## Tasks

- [x] Update `baggagePredicate` struct to include optional value field
- [x] Modify `Create()` to accept 1 or 2 arguments (key only or key+value)
- [x] Update `Match()` to check value when provided, key-only when not
- [x] Add test cases for value matching scenarios
- [x] Run tests to verify working implementation


greeshma1196:
👍🏽

PR #3965, opened 2026-04-15
docs(filters): note static() second-parameter file vs. directory pitfall

### LINKED ISSUE #2233: Doc: static() has some interesting redirect behavior in case you reference a file "index.html"
Interesting behavior:
```
[/tmp/foo]% cat index.html
HELO index

% ./bin/skipper -inline-routes='r: * -> static("/foo", "/tmp/foo/index.html") -> <shunt>;'

% curl http://localhost:9090/foo -v
*   Trying ::1...
> GET /foo HTTP/1.1
> Host: localhost:9090
> User-Agent: curl/7.49.0
> Accept: */*
>
< HTTP/1.1 301 Moved Permanently
< Location: ..//
< Server: Skipper
< Date: Tue, 14 Feb 2023 19:59:01 GMT
< Transfer-Encoding: chunked
<
```

https://pkg.go.dev/net/http@go1.20.1#FileServer says 
> As a special case, the returned file server redirects any request ending in "/index.html" to the same path, without the final "index.html". 

I think this should be documented as edge case in our filter docs

mvanhorn:
Issue #2233 documented a confusing redirect: setting `static("/foo", "/tmp/foo/index.html")` (with the second parameter pointing directly at a file named `index.html`) returns `301 Moved Permanently` with `Location: ..//`. The cause is Go's `http.FileServer` "redirect any request ending in /index.html to the same path without the final index.html" behavior, inherited transitively by `static()`.

The reporter asked to document this as an edge case in the filter docs. The existing `Notes` list near the `static` filter does mention the symmetric case (`GET /foo/index.html` redirects to `/foo/`) but doesn't call out that pointing the second parameter at a file - rather than its parent directory - hits the same redirect. This PR adds one bullet covering that case, with the fix (use the parent directory) inline.

No code changes; doc-only.

Closes #2233.

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -1127,6 +1127,7 @@ Notes:
 * redirects to the directory when a file `index.html` exists and it is requested, i.e. `GET /foo/index.html` redirects to `/foo/` which serves then the `/foo/index.html`
 * serves the content of the `index.html` when a directory is requested
 * does a simple directory listing of files / directories when no `index.html` is present
+* the second parameter must be a directory; pointing it at a single file path that ends in `/index.html` (e.g. `static("/foo", "/tmp/foo/index.html")`) inherits Go's [`http.FileServer`](https://pkg.go.dev/net/http#FileServer) behavior of redirecting any request ending in `/index.html` to the same path without the final `index.html`, which can produce surprising `301 Moved Permanently` responses with a `Location: ..//` header. Pass the parent directory instead (e.g. `static("/foo", "/tmp/foo")`).
 
 ### inlineContent
 


szuecs (approved):

szuecs:
@mvanhorn can you please signoff you commit ?
```
git commit --amend --sign-off
```
should do the trick

MustafaSaber:
👍

szuecs:
:+1:

mvanhorn:
Thanks for the merge, @szuecs!

PR #3969, opened 2026-04-16
fix: proxy filter panics did not run applyFiltersOnError

### LINKED ISSUE #3966: filters that were processed on request should be called in response if there was a filter panic
If you have a filter chain:

```
a() -> b() 
```

a() is a scheduler filter like fifo/lifo that implements the proxy internal `errorHandlerFilter interface`:
```
func (f *fifoFilter) HandleErrorResponse() bool {
	return true
}
```

If b() panics we do not execute a() even if we should, because a() was processed correctly.
This can lead to weir behavior that you get panics but then you are also getting errors caused by filter a() for example.

szuecs:
fix https://github.com/zalando/skipper/issues/3966

Example: In the proxy a route with filter chain a() -> b() exists.

b() panics and a() implements errorHandlerFilter interface and return true from HandleErrorResponse().  In this case we should run the filter a().Response(), because we need to cleanup things as expected.

For example scheduler filters like fifo() or lifo() need to decrement the active concurrent requests counter or we will stop requests in the queue, wich would be miss leading.  If we panic in b() and everyone will look at behavior of a(), because we are stuck there caused by not running the cleanup.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -906,7 +906,7 @@ func WithParams(p Params) *Proxy {
 }
 
 // applies filters to a request
-func (p *Proxy) applyFiltersToRequest(f []*routing.RouteFilter, ctx *context) []*routing.RouteFilter {
+func (p *Proxy) applyFiltersToRequest(f []*routing.RouteFilter, ctx *context, counter *int) []*routing.RouteFilter {
 	if len(f) == 0 {
 		return f
 	}
@@ -927,6 +927,7 @@ func (p *Proxy) applyFiltersToRequest(f []*routing.RouteFilter, ctx *context) []
 		ctx.request = ctx.request.WithContext(labelCtx)
 
 		fi.Request(ctx)
+		(*counter)++
 
 		ctx.request = ctx.request.WithContext(parentCtx)
 		pprof.SetGoroutineLabels(parentCtx)
@@ -1246,6 +1247,7 @@ func stack() []byte {
 
 func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 	var requestElapsed, responseElapsed time.Duration
+	processedFiltersCounter := 0
 
 	requestStopWatch, responseStopWatch := newStopWatch(), newStopWatch()
 	requestStopWatch.Start()
@@ -1258,6 +1260,10 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 
 			perr := &proxyError{
 				err: fmt.Errorf("panic caused by: %v", r),
+			}
+
+			if processedFiltersCounter != 0 {
+				p.applyFiltersOnError(ctx, ctx.route.Filters[:processedFiltersCounter])
 			}
 			p.makeErrorResponse(ctx, perr)
 			err = perr
@@ -1302,7 +1308,7 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 	ctx.applyRoute(route, params, p.flags.PreserveHost())
 
 	requestStopWatch.Stop()
-	processedFilters := p.applyFiltersToRequest(ctx.route.Filters, ctx)
+	processedFilters := p.applyFiltersToRequest(ctx.route.Filters, ctx, &processedFiltersCounter)
 	requestStopWatch.Start()
 
 	// not every of these branches could end up in a response to the client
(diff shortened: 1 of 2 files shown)


proxy/proxy_test.go
```diff
@@ -1028,27 +1031,58 @@ func TestFilterPanic(t *testing.T) {
...
 	defer s.Close()
 
+	metrics := &metricstest.MockMetrics{}
+	defer metrics.Close()
+	d := 50 * time.Millisecond
+	reg := scheduler.RegistryWith(scheduler.Options{
+		Metrics:                metrics,
+		EnableRouteFIFOMetrics: true,
+		MetricsUpdateTimeout:   d,
+	})
```
madumalt, line 1041:
```suggestion
	reg := scheduler.RegistryWith(scheduler.Options{
		Metrics:                metrics,
		EnableRouteFIFOMetrics: true,
		MetricsUpdateTimeout:   50 * time.Millisecond,
	})
```

madumalt: why define a var if we can directly assign it. On the other hand its more readable this way for me.

proxy/proxy.go
```diff
@@ -906,7 +906,7 @@ func WithParams(p Params) *Proxy {
 }
 
 // applies filters to a request
-func (p *Proxy) applyFiltersToRequest(f []*routing.RouteFilter, ctx *context) []*routing.RouteFilter {
+func (p *Proxy) applyFiltersToRequest(f []*routing.RouteFilter, ctx *context, counter *int) []*routing.RouteFilter {
```
madumalt, line 909: shouldn't the var name be `index` instead of `counter` ?

proxy/proxy.go
```diff
@@ -1246,6 +1247,7 @@ func stack() []byte {
 
 func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 	var requestElapsed, responseElapsed time.Duration
+	processedFiltersCounter := 0
```
madumalt, line 1250: similar to the above comment, shouldn't this be something like `lastProcessedFilterIndex` ?

madumalt (commented):
some minor comments, suggestions to change some variable names. The logic seems sound to me!

proxy/proxy.go
szuecs: No because we count number of filters executed in the request path

proxy/proxy.go
szuecs: No, same here and "last" would be misleading

proxy/proxy_test.go
```diff
@@ -1059,6 +1093,16 @@ func TestFilterPanic(t *testing.T) {
 	if err = testLog.WaitFor(msg, 100*time.Millisecond); err != nil {
 		t.Errorf("expected '%s' in logs", msg)
 	}
+
+	// wait for updateMetrics to happen
+	time.Sleep(50*time.Millisecond - time.Since(now))
```
szuecs: Here we should use `d`

proxy/proxy_test.go
madumalt: answer: https://github.com/zalando/skipper/pull/3969#discussion_r3098975802

proxy/proxy.go
madumalt: I got this whole idea that is more like an index is with this,

https://github.com/zalando/skipper/pull/3969/changes#diff-1d74870266949f1927a500e5f0002617239d3e3cd9bcceee6be014738e5e0f2bR1266

madumalt: I approved the PR, if the suggestion doen't make sense, just ignore it :)

madumalt (approved):

szuecs:
:+1:

proxy/proxy.go
```diff
@@ -1259,6 +1261,10 @@ func (p *Proxy) do(ctx *context, parentSpan ot.Span) (err error) {
 			perr := &proxyError{
 				err: fmt.Errorf("panic caused by: %v", r),
 			}
+
+			if processedFiltersCounter != 0 {
+				p.applyFiltersOnError(ctx, ctx.route.Filters[:processedFiltersCounter])
```
MustafaSaber, line 1266: What if we have another panic in this call?

szuecs: then we likely panic

MustafaSaber (approved):

PR #3984, opened 2026-04-24
bugfix: routesrv return full routingTable with zone

MustafaSaber:
In routeSRV, when we serve zone aware routes after `zone: r.PathValue("zone")` we only update uncompressed data, while http.Client serves compressed data by default. Which made all requests land in `gzip` branch later.

And `gzip` branch is serving `zdata` which was never updated with the new routingTable.

This commit include a reproducer testcase and the fix is in the next commit.


szuecs (approved):

greeshma1196:
👍🏽

PR #3988, opened 2026-04-29
fix: fatal at startup, if we can not create a valkey connection

szuecs:
fix: fatal at startup, if we can not create a valkey connection, we should assume something is wrong, because we have this dependency


greeshma1196:
👍🏽

szuecs:
:+1:

PR #3989, opened 2026-04-29
feat: add a flag to require dataclients to exist

MustafaSaber:
Add an opt in flag to fatal if some dataclients are not configured. If the flag is set it will ensure that skipper doesn't start without this set of dataclients

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -129,24 +129,25 @@ type Config struct {
 	OpenTelemetry *otel.Options `yaml:"open-telemetry"`
 
 	// route sources:
-	EtcdUrls           string               `yaml:"etcd-urls"`
-	EtcdPrefix         string               `yaml:"etcd-prefix"`
-	EtcdTimeout        time.Duration        `yaml:"etcd-timeout"`
-	EtcdInsecure       bool                 `yaml:"etcd-insecure"`
-	EtcdOAuthToken     string               `yaml:"etcd-oauth-token"`
-	EtcdUsername       string               `yaml:"etcd-username"`
-	EtcdPassword       string               `yaml:"etcd-password"`
-	RoutesFile         string               `yaml:"routes-file"`
-	RoutesURLs         *listFlag            `yaml:"routes-urls"`
-	InlineRoutes       string               `yaml:"inline-routes"`
-	ForwardBackendURL  string               `yaml:"forward-backend-url"`
-	AppendFilters      *defaultFiltersFlags `yaml:"default-filters-append"`
-	PrependFilters     *defaultFiltersFlags `yaml:"default-filters-prepend"`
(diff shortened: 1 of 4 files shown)


skipper.go
```diff
@@ -1245,6 +1252,42 @@ func createDataClients(o Options, cr *certregistry.CertRegistry) ([]routing.Data
...
+		"kubernetes":   {},
+		"etcd":         {},
+		"routesfile":   {},
+		"routesurls":   {},
+		"inlineroutes": {},
+		"plugins":      {},
+	}
+
+	for _, name := range o.RequiredDataClients {
+		if _, ok := validDataClients[name]; !ok {
```
MustafaSaber, line 1266: API validation

skipper.go
```diff
@@ -1245,6 +1252,42 @@ func createDataClients(o Options, cr *certregistry.CertRegistry) ([]routing.Data
...
+			return fmt.Errorf("invalid dataclient name %q; accepted names: %s", name, strings.Join(slices.Sorted(maps.Keys(validDataClients)), ", "))
+		}
+		var active bool
+		switch name {
+		case "kubernetes":
+			active = o.Kubernetes
```
MustafaSaber, line 1272: doesn't check if the dataclient is initiated as it's called after `createDataClients` which will fail if the client failed to be created.

skipper.go
```diff
@@ -1245,6 +1252,42 @@ func createDataClients(o Options, cr *certregistry.CertRegistry) ([]routing.Data
 	return clients, nil
 }
 
+func checkRequiredDataClients(o Options) error {
+	validDataClients := map[string]struct{}{
```
szuecs, line 1256: this you can only do to support "internal" dataclients, but not dataclients that are created by skipper library users

szuecs:
see https://github.com/zalando/skipper/pull/3990 how to support custom dataclients, too

config/config.go
```diff
@@ -385,6 +386,7 @@ func NewConfig() *Config {
 	cfg.EditRoute = routeChangerConfig{}
 	cfg.KubernetesEastWestRangeDomains = commaListFlag()
 	cfg.RoutesURLs = commaListFlag()
+	cfg.RequiredDataClients = commaListFlag("kubernetes", "etcd", "routesfile", "routesurls", "inlineroutes", "plugins")
```
greeshma1196, line 389: I think by default this should be empty.

MustafaSaber: I will close this PR, but no

MustafaSaber: The arg is the allowed set, if you provide the flag it should have one value of those, or don't just provide the flag if no opt in required

MustafaSaber:
close in favor of https://github.com/zalando/skipper/pull/3990

PR #3990, opened 2026-04-29
feature: ensure-dataclient

szuecs:
The new flag -ensure-dataclient="inline,kubernetes,.." can be used to ensure that specific dataclients were created. This makes sense to use if you have templated configuration files and you want to make sure a given dataclient will be started.

Good:
```
% ./bin/skipper -inline-routes='r: * -> status(201) -> <shunt>' -address :9002 -ensure-dataclients "inline"
[APP]INFO[0000] enable swarm: false
[APP]INFO[0000] Replacing tee filter specification
[APP]INFO[0000] Replacing teenf filter specification
[APP]INFO[0000] Replacing teeResponse filter specification
[APP]INFO[0000] route settings, reset, route: r: * -> status(201) -> <shunt>
[APP]INFO[0000] route settings received, id: 1
[APP]INFO[0000] route settings applied, id: 1
[APP]INFO[0000] Dataclients are updated once, first load complete
[APP]INFO[0000] Listen on :9002
```

Bad:
```
% ./bin/skipper -inline-routes='r: * -> status(201) -> <shunt>' -address :9002 -ensure-dataclients "kubernetes"
[APP]FATA[0000] Failed to start skipper: failed to find dataclient that was ensured: "kubernetes"
zsh: exit 1
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -147,6 +147,7 @@ type Config struct {
 	CloneRoute         routeChangerConfig   `yaml:"clone-route"`
 	SourcePollTimeout  int64                `yaml:"source-poll-timeout"`
 	WaitFirstRouteLoad bool                 `yaml:"wait-first-route-load"`
+	EnsureDataClients  *listFlag            `yaml:"ensure-dataclients"`
 
 	// Forwarded headers
 	ForwardedHeadersList            *listFlag            `yaml:"forwarded-headers"`
@@ -394,6 +395,7 @@ func NewConfig() *Config {
 	cfg.ProxyAllowListCIDRs = commaListFlag()
 	cfg.ProxyDenyListCIDRs = commaListFlag()
 	cfg.ProxySkipListCIDRs = commaListFlag()
+	cfg.EnsureDataClients = commaListFlag()
 
 	flag := flag.NewFlagSet("", flag.ExitOnError)
 	flag.StringVar(&cfg.ConfigFile, "config-file", "", "if provided the flags will be loaded/overwritten by the values on the file (yaml)")
@@ -510,6 +512,7 @@ func NewConfig() *Config {
 	flag.Var(&cfg.EditRoute, "edit-route", "match and edit filters and predicates of all routes")
 	flag.Var(&cfg.CloneRoute, "clone-route", "clone all matching routes and replace filters and predicates of all matched routes")
 	flag.BoolVar(&cfg.WaitFirstRouteLoad, "wait-first-route-load", false, "prevent starting the listener before the first batch of routes were loaded")
+	flag.Var(cfg.EnsureDataClients, "ensure-dataclients", `comma separated list of routing.NamedDataClient names: "etcd", "eskipfile-watch", "eskipfile-remote", "eskipfile", "inline", "kubernetes"`)
 
 	// Forwarded headers
 	flag.Var(cfg.ForwardedHeadersList, "forwarded-headers", "comma separated list of headers to add to the incoming request before routing\n"+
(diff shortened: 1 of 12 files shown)


szuecs:
:+1:

routing/routing.go
```diff
@@ -55,6 +55,14 @@ type DataClient interface {
 	LoadUpdate() ([]*eskip.Route, []string, error)
 }
 
+// NamedDataClient is like DataClient but has a name, which can be
+// used to EnsureDataClient or -ensure-dataclients.
+type NamedDataClient interface {
+	LoadAll() ([]*eskip.Route, error)
```
MustafaSaber: I think we should embed `DataClient` interface here

szuecs: fixed

skipper.go
```diff
@@ -1242,6 +1244,20 @@ func createDataClients(o Options, cr *certregistry.CertRegistry) ([]routing.Data
 		clients = append(clients, kubernetesClient)
 	}
 
+	for _, s := range o.EnsureDataClient {
```
MustafaSaber: I think this should be outside createDataClients() and after adding the CustomDataClients

skipper.go
MustafaSaber, line 1763: I think we should have another function and call it here

MustafaSaber, line 1760: And we can also check here that if we also have ensureDataClient this should fail before even calling the function

skipper.go
szuecs: fixed, thanks yeah!

szuecs:
:+1:

config/config.go
```diff
@@ -510,6 +512,7 @@ func NewConfig() *Config {
 	flag.Var(&cfg.EditRoute, "edit-route", "match and edit filters and predicates of all routes")
 	flag.Var(&cfg.CloneRoute, "clone-route", "clone all matching routes and replace filters and predicates of all matched routes")
 	flag.BoolVar(&cfg.WaitFirstRouteLoad, "wait-first-route-load", false, "prevent starting the listener before the first batch of routes were loaded")
+	flag.Var(cfg.EnsureDataClients, "ensure-dataclients", `comma separated list of routing.NamedDataClient names: "etcd", "eskipfile-watch", "eskipfile-remote", "eskipfile", "inline", "kubernetes"`)
```
MustafaSaber, line 515: We don't do any validation to enforce those values for the flag, ideally we should fail if we don't have a known value. If we add the values for the commandListFlag it will do the validation but I don't know how it will work with customDataClients

skipper.go
```diff
@@ -1751,6 +1753,20 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	// append custom data clients
 	dataClients = append(dataClients, o.CustomDataClients...)
 
+	for _, s := range o.EnsureDataClient {
```
MustafaSaber: This is really a minor but I prefer to encapsulate this in a function

skipper_test.go
```diff
@@ -509,124 +509,361 @@ func createRoutesFile(route string) (string, error) {
...
-	<-rt.FirstLoad()
-	tracer := tracingtest.NewTracer()
-	pr := proxy.WithParams(proxy.Params{
-		Routing:     rt,
-		OpenTracing: &proxy.OpenTracingParams{Tracer: tracer},
+
+	t.Run("ensure all dataclients work as expected", func(t *testing.T) {
+		// routesfile
+		routesFileStatus := 201
+		routeStringFmt := `r%d: Path("/routes-file") -> disableAccessLog() -> status(%d) -> inlineContent("Got it") -> <shunt>;`
```
MustafaSaber, line 638: If this doesn't change in all test cases let's move it outside t.Run()

config/config.go
szuecs: It's on purpose, that we do not validate, because we are a library and can't predict users name.
Also if you have a wrong name you will see this fast, you don't need to see the help message to see the fatal log  ;)

skipper_test.go
szuecs: no, it is fine to have it here. We move only things outside that are not quickly starting (docker container for example), rest should be as stateless as possible.

skipper.go
szuecs: I don't. It's good to see this long thing here close to the dataclient creations.

szuecs: However I did an extract method to have it anyways

MustafaSaber (approved):

szuecs:
:+1:

PR #3991, opened 2026-04-30
feature: add cache() HTTP response caching filter

larry-dalmeida:
# Summary

Adds a new `cache()` filter providing proxy-layer HTTP response caching backed by a shared in-memory LRU store.

## Two operating modes

- `cache()` - RFC mode: upstream `Cache-Control` is fully authoritative (freshness, `no-store`, `private`, `Vary`, `s-maxage`, heuristic TTL, conditional revalidation)
- `cache("5m", "15s", "30s"[, "60s"[, "X-Header"]])` - force mode: operator sets TTL, error TTL, and SWR window directly; upstream freshness directives are ignored        

## Behaviours                                                                                                                                                     
                              
- Stale-while-revalidate: serves stale immediately, refreshes in background
- Cold-miss coalescing: concurrent requests for the same uncached key share a single upstream fetch
- Unsafe method invalidation: `POST`/`PUT`/`DELETE`/`PATCH` success invalidates the cached entry and any same-origin `Location`/`Content-Location` URIs
- `HEAD 200` freshens stored headers without replacing the cached body
- Stale-if-error (RFC 5861): serves stale during upstream 5xx within a configurable window                                                                                                                                                      
- Per-header key namespacing via optional `keyHeaders` argument

**RFC 9111 compliance:** `s-maxage` correctly implies proxy-revalidate (§5.2.2.10), `Vary` header support, heuristic freshness (§4.2.2), corrected initial age (§4.2.3).

### Revalidation worker sizing

Single-worker + 256-slot queue design is intentional and sufficient for our target workload driving the need for this implementation.

Screenshot 2026-05-21 at 07 35 05

At ~1.36ms average upstream latency (CURRENT_CACHING_SOLUTION_TO_BE_REPLACED, production, last 7 days),  worker can process ~735 revalidations/sec ceiling, well within capacity. Against a ~267 req/s baseline, the queue will not back up under normal or burst conditions. The `reval_dropped` metric provides visibility if this assumption breaks in a different deployment context.

<details>
<summary>Math</summary>
1000ms / 1.36ms ≈ 735 revals/sec
735 revals/sec capacity vs ~267 req/s demand - that's ~2.75x headroom. 
Even if 100% of requests hit a stale entry simultaneously, the worker clears the queue faster than it fills.
256 slots / 735 revals/sec ≈ 0.35 seconds of burst absorption at full capacity.
</details>

### Demo

Covers MISS→HIT→STALE→HIT, error caching, coalescing, `Age` header, and `no-cache`/`no-store` bypass against a live mockapi backend. Run against a local Skipper on `:9090`.

<details>
<summary>demo-route.eskip + demo.sh</summary>

```eskip
demoBanners:
  Path("/api/v1/banners")
  -> cache("5s", "3s", "5s")
  -> "http://localhost:8888";

demoBanners404:
  Path("/api/v1/banners/0")
  -> cache("5s", "3s", "5s")
  -> "http://localhost:8888";
```

```bash
#!/usr/bin/env bash
set -euo pipefail

SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
SKIPPER_DIR="${SCRIPT_DIR}/skipper"
MOCK_DIR="${SCRIPT_DIR}/mock-upstream"

BASE="http://localhost:9090"
MOCK_ADDR="http://localhost:8888"
N=0  # global request counter
MOCK_PID=""
SKIPPER_PID=""
SKIPPER_LOG="${SCRIPT_DIR}/skipper-demo.log"

# Extract X-Cache-Status from a response; never fails the script.
xcache() {
  grep -i "^x-cache-status:" | tr -d '\r' | awk '{print $2}'
}

req() {
  local path="$1"
  local raw status
  raw=$(curl -si --max-time 10 "${BASE}${path}" 2>/dev/null) || true
  status=$(printf '%s' "$raw" | xcache) || true
  N=$((N + 1))
  printf "[%d] %s\n" "$N" "${status:-NO_HEADER}"
}

cleanup() {
  echo ""
  echo "Stopping mock upstream and Skipper..."
  [[ -n "$MOCK_PID" ]] && kill "$MOCK_PID" 2>/dev/null || true
  [[ -n "$SKIPPER_PID" ]] && kill "$SKIPPER_PID" 2>/dev/null || true
  echo "Skipper logs saved to: ${SKIPPER_LOG}"
}
trap cleanup EXIT

echo "============================================"
echo " Cache Demo: MISS -> HIT -> STALE -> HIT"
echo "============================================"
echo ""

# Build and start the mock upstream.
echo "Building mock upstream..."
cd "${MOCK_DIR}"
go build -o mock-upstream . 2>/dev/null
./mock-upstream -addr :8888 > /tmp/mock-upstream.log 2>&1 &
MOCK_PID=$!
cd "${SCRIPT_DIR}"

# Build and start Skipper.
echo "Building Skipper..."
cd "${SKIPPER_DIR}"
go build -o skipper ./cmd/skipper 2>/dev/null
./skipper -routes-file ../demo-route.eskip -address :9090 > "${SKIPPER_LOG}" 2>&1 &
SKIPPER_PID=$!
cd "${SCRIPT_DIR}"

# Wait for both to be ready.
# Use /__health for Skipper (avoids warming the cache) and a direct hit for mock.
echo "Starting services..."
for i in $(seq 1 20); do
  if curl -s --max-time 1 "${MOCK_ADDR}/api/v1/banners" > /dev/null 2>&1 && \
     curl -s --max-time 1 "${BASE}/__health" > /dev/null 2>&1; then
    break
  fi
  sleep 0.5
done

echo "Services ready."
echo ""

# -- Section 1: MISS -> HIT -> STALE -> revalidation HIT ---------------------
# TTL=5s, SWR=5s, hard expiry=10s
echo "--- Section 1: Happy path (TTL=5s, SWR=5s) ---"
echo ""

echo "Request 1 -- cold cache, expect MISS"
req "/api/v1/banners"                          # T+0: MISS, CreatedAt=T+0

echo "Request 2 -- within TTL, expect HIT"
req "/api/v1/banners"                          # T+0: HIT

echo "Sleeping 6s (past TTL=5s, inside SWR window)..."
sleep 6                               # now T+6

echo "Request 3 -- past TTL, inside SWR window, expect STALE"
req "/api/v1/banners"                          # T+6: STALE, revalidation fires, new CreatedAt=T+6

echo "Sleeping 3s (background revalidation should complete)..."
sleep 3                               # now T+9; revalidated entry is 3s old (< TTL=5s)

echo "Request 4 -- revalidated entry, expect HIT"
req "/api/v1/banners"                          # T+9: HIT (3s old, inside TTL=5s)

echo ""

# -- Section 2: Error caching (404, no SWR) -----------------------------------
# errorTTL=3s, hard expiry=3s (SWR=0 for errors)
echo "--- Section 2: Error caching (errorTTL=3s, no SWR) ---"
echo ""

echo "Request 5 -- cold cache for /api/v1/banners/0, expect MISS (404 stored)"
req "/api/v1/banners/0"                        # T+9: MISS, 404 stored

echo "Request 6 -- within errorTTL, expect HIT (cached 404)"
req "/api/v1/banners/0"                        # T+9: HIT

echo "Sleeping 4s (past errorTTL=3s, no SWR for errors)..."
sleep 4                               # now T+13

echo "Request 7 -- errorTTL expired, expect MISS"
req "/api/v1/banners/0"                        # T+13: MISS

echo ""

# -- Section 3: Thundering herd / coalescing ----------------------------------
# The /api/v1/banners entry was last revalidated during Section 1 (background
# revalidation from the STALE request). With TTL=5s and SWR=5s the hard expiry
# is CreatedAt + 10s. We are at T+13 here; sleep 8s to ensure we are safely
# past even a late revalidation (hard expiry window is at most 10s from T+9).
echo "--- Section 3: Coalescing (5 parallel cold-miss requests) ---"
echo ""

echo "Sleeping 8s for /api/v1/banners to hard-expire..."
sleep 8                               # ensures we are past hard expiry

echo "Firing 5 parallel requests against cold cache -- all should show MISS"
echo "(Typically 1-2 upstream fetches occur; singleflight coalesces the rest)"
echo ""

tmpdir=$(mktemp -d)
pids=()
for i in 1 2 3 4 5; do
  (
    raw=$(curl -si --max-time 10 "${BASE}/api/v1/banners" 2>/dev/null) || true
    xcstat=$(printf '%s' "$raw" | xcache) || true
    printf '%s' "${xcstat:-NO_HEADER}" > "${tmpdir}/result_${i}"
  ) &
  pids+=($!)
done
wait "${pids[@]}"

for i in 1 2 3 4 5; do
  N=$((N + 1))
  printf "[%d] %s\n" "$N" "$(cat "${tmpdir}/result_${i}")"
done
rm -rf "$tmpdir"

echo ""
echo "Request $((N + 1)) -- after coalescing, expect HIT (entry stored by coalesced fetch)"
req "/api/v1/banners"

echo ""

# -- Section 4: Age header -----------------------------------------------------
echo "--- Section 4: Age header ---"
echo ""

echo "Sleeping 11s to hard-expire /api/v1/banners (TTL=5s + SWR=5s = 10s)..."
sleep 11

echo "Request $((N + 1)) -- cold cache, expect MISS"
req "/api/v1/banners"

echo "Sleeping 2s..."
sleep 2

echo "Request $((N + 1)) -- within TTL, expect HIT with Age >= 2"
raw=$(curl -si --max-time 10 "${BASE}/api/v1/banners" 2>/dev/null) || true
age_val=$(printf '%s' "$raw" | grep -i "^age:" | tr -d '\r' | awk '{print $2}') || true
N=$((N + 1))
printf "[%d] Age: %s (expected >= 2)\n" "$N" "${age_val:-MISSING}"
echo ""

# -- Section 5: Request no-cache bypass ----------------------------------------
echo "--- Section 5: Request no-cache bypass ---"
echo ""

echo "Cache should be warm from Section 4."
echo "Request $((N + 1)) -- Cache-Control: no-cache, expect MISS even on warm cache"
raw=$(curl -si --max-time 10 -H "Cache-Control: no-cache" "${BASE}/api/v1/banners" 2>/dev/null) || true
xcstat=$(printf '%s' "$raw" | xcache) || true
N=$((N + 1))
printf "[%d] %s (expected MISS)\n" "$N" "${xcstat:-NO_HEADER}"
echo ""

# -- Section 6: Request no-store -----------------------------------------------
echo "--- Section 6: Request no-store ---"
echo ""

echo "Sleeping 11s to hard-expire cache..."
sleep 11

echo "Request $((N + 1)) -- no-store, expect MISS (and nothing stored)"
raw=$(curl -si --max-time 10 -H "Cache-Control: no-store" "${BASE}/api/v1/banners" 2>/dev/null) || true
xcstat=$(printf '%s' "$raw" | xcache) || true
N=$((N + 1))
printf "[%d] %s (expected MISS)\n" "$N" "${xcstat:-NO_HEADER}"

echo "Request $((N + 1)) -- normal GET after no-store, expect MISS (entry stored now)"
req "/api/v1/banners"
echo ""

# -- Section 7: min-fresh bypass -----------------------------------------------
# Cache is warm from Section 6 Request 18 (stored just above).
# TTL=5s so freshness decays over the next 5s.
echo "--- Section 7: min-fresh (RFC 9111 §5.2.1) ---"
echo ""

echo "Cache is warm from Section 6. Request $((N + 1)) -- min-fresh=1 with plenty of TTL, expect HIT"
raw=$(curl -si --max-time 10 -H "Cache-Control: min-fresh=1" "${BASE}/api/v1/banners" 2>/dev/null) || true
xcstat=$(printf '%s' "$raw" | xcache) || true
N=$((N + 1))
printf "[%d] %s (expected HIT)\n" "$N" "${xcstat:-NO_HEADER}"

echo "Sleeping 4s (TTL=5s, ~1s freshness remaining)..."
sleep 4

echo "Request $((N + 1)) -- min-fresh=2 with ~1s left, expect MISS (2s required, ~1s left)"
raw=$(curl -si --max-time 10 -H "Cache-Control: min-fresh=2" "${BASE}/api/v1/banners" 2>/dev/null) || true
xcstat=$(printf '%s' "$raw" | xcache) || true
N=$((N + 1))
printf "[%d] %s (expected MISS)\n" "$N" "${xcstat:-NO_HEADER}"
echo ""

echo "============================================"
echo " Demo complete."
echo "============================================"

```

</details>

Caching Layer for Contentful - Frame 4

### DIFF
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -3681,3 +3681,96 @@ pr: Path("/products/:productId")
 ```
 consistentHashBalanceFactor(3)
 ```
+
+## Caching
+
+### cache
+
+Caches HTTP responses in a shared in-memory LRU store. Operates in two modes
+selected by the number of arguments.
+
+**RFC mode** (zero arguments): upstream `Cache-Control` is fully authoritative.
+Freshness, `no-store`, `private`, `Vary`, `s-maxage`, heuristic TTL
+(RFC 9111 §4.2.2), and conditional revalidation (`ETag`/`Last-Modified`) are
+all honoured.
+
+**Force mode** (3–5 arguments): operator TTL is authoritative; upstream
+`Cache-Control` freshness directives are ignored. Semantic blockers (`no-store`,
+`private`) are respected in RFC mode but ignored in force mode.
+
+Stale-while-revalidate (SWR): a stale entry within the SWR window is served
+immediately while a background fetch refreshes the entry. Concurrent cold-miss
+requests for the same key are coalesced into a single upstream fetch.
+
+Unsafe methods (`POST`, `PUT`, `DELETE`, `PATCH`) invalidate the cached entry on
+success. `HEAD 200` freshens stored headers without replacing the body.
+
+Parameters (force mode):
+
+* `ttl` – freshness duration for 200 responses (string, e.g. `"5m"`)
+* `errorTTL` – freshness duration for 404/5xx responses (string)
+* `swrWindow` – stale-while-revalidate window after TTL expires (string)
+* `staleIfError` – optional; how long a stale entry may be served on upstream
+  5xx (string, RFC 5861)
+* `keyHeaders` – optional; comma-separated request header names folded into the
+  cache key for per-header isolation (string, e.g. `"Authorization,X-Tenant-ID"`)
+
+Examples:
+
+RFC mode — upstream `Cache-Control` is authoritative:
+
+```
+-> cache() -> "https://cdn.example.org"
+```
+
+Force mode — 5 minute TTL, 15 second error TTL, 30 second SWR window:
+
+```
+-> cache("5m", "15s", "30s") -> "https://cdn.example.org"
+```
+
+Force mode with stale-if-error fallback:
+
+```
+-> cache("5m", "15s", "30s", "60s") -> "https://cdn.example.org"
+```
+
+Force mode with per-tenant cache key isolation:
+
+```
+-> cache("5m", "15s", "30s", "0s", "X-Tenant-ID") -> "https://cdn.example.org"
+```
+
+**Cache key**
+
+The key is derived from route ID + HTTP method + host + path + query string.
+Route ID is included so entries from different routes never collide when sharing
+the same storage instance. Additional request headers can be folded in via
+`keyHeaders`. Without `keyHeaders`, all requests to the same path share one
+cache entry regardless of caller identity.
+
+**Safety**
+
+The filter stores whatever the upstream returns and serves it to any request
+matching the same key. It has no awareness of other filters in the chain.
+
+* **Only use on routes where the response is identical for all callers** —
+  public CMS content, unauthenticated APIs, static assets.
+* **Do not use on routes that return user-specific data or set user-scoped
+  cookies** — those responses will be served to other users.
+* **`Authorization` is not in the key by default.** Requests from different
+  users hitting the same path share a cache entry unless `Authorization` is
+  added to `keyHeaders` or stripped before the filter.
+* **`Cache-Control: private` is ignored in force mode.** Audit the upstream
+  response before enabling force mode on any authenticated route.
+
+!!! note
+    The LRU store is shared across all `cache()` filter instances created from
+    the same spec. The storage budget is divided evenly across 256 internal
+    shards; a single entry larger than one shard's budget is dropped with a
+    warning log.
+
+!!! note
+    `s-maxage` implies `proxy-revalidate` per RFC 9111 §5.2.2.10: stale entries
+    stored under `s-maxage` are never served without revalidation, regardless of
+    the SWR window.
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -3743,7 +3743,8 @@ Force mode with per-tenant cache key isolation:
 
 **Cache key**
 
-The key is derived from route ID + HTTP method + host + path + query string.
+The key is derived from route ID + HTTP method + host + path + query string,
+hashed with SHA-256 for uniform shard distribution.
 Route ID is included so entries from different routes never collide when sharing
 the same storage instance. Additional request headers can be folded in via
 `keyHeaders`. Without `keyHeaders`, all requests to the same path share one
@@ -3758,17 +3759,24 @@ matching the same key. It has no awareness of other filters in the chain.
   public CMS content, unauthenticated APIs, static assets.
 * **Do not use on routes that return user-specific data or set user-scoped
   cookies** — those responses will be served to other users.
-* **`Authorization` is not in the key by default.** Requests from different
-  users hitting the same path share a cache entry unless `Authorization` is
-  added to `keyHeaders` or stripped before the filter.
+* **`Authorization` is not in the key by default.** Responses are stored and
+  served without regard to caller identity. To isolate per-user responses, add
+  `Authorization` to `keyHeaders`; without it, a response stored for one user
+  will be served to all others on the same path.
 * **`Cache-Control: private` is ignored in force mode.** Audit the upstream
   response before enabling force mode on any authenticated route.
 
 !!! note
-    The LRU store is shared across all `cache()` filter instances created from
-    the same spec. The storage budget is divided evenly across 256 internal
-    shards; a single entry larger than one shard's budget is dropped with a
-    warning log.
+    The LRU store is shared across all `cache()` filter instances. The storage
+    budget is divided evenly across 256 internal shards; a single entry larger
+    than one shard's budget is dropped with a warning log.
+
+!!! note
+    Three metrics track LRU behaviour: `lru_eviction` (counter, incremented each
+    time an entry is evicted due to memory pressure), `lru_bytes` (gauge,
+    updated on every eviction to reflect current storage usage in bytes), and
+    `lru_oversized` (counter, incremented when an entry is too large to fit in
+    any shard and is silently dropped).
 
 !!! note
     `s-maxage` implies `proxy-revalidate` per RFC 9111 §5.2.2.10: stale entries
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -3694,6 +3694,13 @@ Freshness, `no-store`, `private`, `Vary`, `s-maxage`, heuristic TTL
 (RFC 9111 §4.2.2), and conditional revalidation (`ETag`/`Last-Modified`) are
 all honoured.
 
+Request `Cache-Control` directives partially honoured: `no-store`, `no-cache`
+(including `max-age=0`), `only-if-cached`, `max-stale`, and `min-fresh`.
+Client `max-age` with non-zero values is not enforced — use `no-cache` to force
+revalidation. Per RFC 9111 §5.2.1, request directives are advisory (`MAY`) for
+shared caches; ignoring client `max-age` prevents cache-bypass abuse on shared
+routes.
+
 **Force mode** (3–5 arguments): operator TTL is authoritative; upstream
 `Cache-Control` freshness directives are ignored. Semantic blockers (`no-store`,
 `private`) are respected in RFC mode but ignored in force mode.
--- a/docs/reference/filters.md
+++ b/docs/reference/filters.md
@@ -3691,13 +3691,13 @@ selected by the number of arguments.
 
 **RFC mode** (zero arguments): upstream `Cache-Control` is fully authoritative.
 Freshness, `no-store`, `private`, `Vary`, `s-maxage`, heuristic TTL
-(RFC 9111 §4.2.2), and conditional revalidation (`ETag`/`Last-Modified`) are
+([RFC 9111 §4.2.2](https://www.rfc-editor.org/rfc/rfc9111#section-4.2.2)), and conditional revalidation (`ETag`/`Last-Modified`) are
 all honoured.
 
 Request `Cache-Control` directives partially honoured: `no-store`, `no-cache`
 (including `max-age=0`), `only-if-cached`, `max-stale`, and `min-fresh`.
 Client `max-age` with non-zero values is not enforced — use `no-cache` to force
-revalidation. Per RFC 9111 §5.2.1, request directives are advisory (`MAY`) for
+revalidation. Per [RFC 9111 §5.2.1](https://www.rfc-editor.org/rfc/rfc9111#section-5.2.1), request directives are advisory (`MAY`) for
 shared caches; ignoring client `max-age` prevents cache-bypass abuse on shared
 routes.
 
@@ -3718,7 +3718,7 @@ Parameters (force mode):
 * `errorTTL` – freshness duration for 404/5xx responses (string)
 * `swrWindow` – stale-while-revalidate window after TTL expires (string)
 * `staleIfError` – optional; how long a stale entry may be served on upstream
-  5xx (string, RFC 5861)
+  5xx (string, [RFC 5861](https://www.rfc-editor.org/rfc/rfc5861))
 * `keyHeaders` – optional; comma-separated request header names folded into the
   cache key for per-header isolation (string, e.g. `"Authorization,X-Tenant-ID"`)
 
@@ -3786,6 +3786,6 @@ matching the same key. It has no awareness of other filters in the chain.
     any shard and is silently dropped).
 
 !!! note
-    `s-maxage` implies `proxy-revalidate` per RFC 9111 §5.2.2.10: stale entries
+    `s-maxage` implies `proxy-revalidate` per [RFC 9111 §5.2.2.10](https://www.rfc-editor.org/rfc/rfc9111#section-5.2.2.10): stale entries
     stored under `s-maxage` are never served without revalidation, regardless of
     the SWR window.
--- a/filters/cache/cache_control.go
+++ b/filters/cache/cache_control.go
@@ -1,29 +1,34 @@
 package cache
 
 import (
+	"math"
 	"net/http"
 	"strconv"
 	"strings"
-	"time"
 )
 
 type cacheDirectives struct {
-	NoStore        bool
-	NoCache        bool
-	Private        bool
-	MustRevalidate bool
-	Public         bool
-	SMaxAge        *time.Duration // RFC 9111 §5.2.2.10; nil means not present
+	NoStore         bool
+	NoCache         bool
+	Private         bool
+	MustRevalidate  bool
+	ProxyRevalidate bool
+	Public          bool
+	MaxAge          int64 // -1 = not present; 0 means max-age=0
+	SMaxAge         int64 // -1 = not present
 }
 
 type requestCacheDirectives struct {
 	NoStore      bool
-	NoCache      bool // includes max-age=0
+	NoCache      bool  // includes max-age=0
 	OnlyIfCached bool
+	MaxStale     int64 // -1 = not present; >= 0 = max staleness seconds client accepts
+	MinFresh     int64 // -1 = not present; >= 0 = min remaining freshness seconds required
 }
 
+// parseRequestCacheControl parses Cache-Control request directives from h.
 func parseRequestCacheControl(h http.Header) requestCacheDirectives {
-	var d requestCacheDirectives
+	d := requestCacheDirectives{MaxStale: -1, MinFresh: -1}
 	for _, line := range h.Values("Cache-Control") {
 		for token := range strings.SplitSeq(line, ",") {
 			parts := strings.SplitN(strings.TrimSpace(token), "=", 2)
@@ -39,6 +44,20 @@ func parseRequestCacheControl(h http.Header) requestCacheDirectives {
 				}
 			case "only-if-cached":
 				d.OnlyIfCached = true
+			case "max-stale":
+				if len(parts) == 2 {
+					if v, err := strconv.ParseInt(strings.TrimSpace(parts[1]), 10, 64); err == nil && v >= 0 {
+						d.MaxStale = v
+					}
+				} else {
+					d.MaxStale = math.MaxInt32 // ~68 years; safe upper bound for *time.Second conversion
+				}
+			case "min-fresh":
+				if len(parts) == 2 {
+					if v, err := strconv.ParseInt(strings.TrimSpace(parts[1]), 10, 64); err == nil && v >= 0 {
+						d.MinFresh = v
+					}
+				}
 			}
 		}
 	}
@@ -47,9 +66,9 @@ func parseRequestCacheControl(h http.Header) requestCacheDirectives {
 
 // parseCacheControl parses Cache-Control response directives from h.
 // Uses Header.Values to handle multiple header lines; matches names
-// case-insensitively per RFC 7234 §5.2.
+// case-insensitively per RFC 9111 §5.2.
 func parseCacheControl(h http.Header) cacheDirectives {
-	var d cacheDirectives
+	d := cacheDirectives{MaxAge: -1, SMaxAge: -1}
 	for _, line := range h.Values("Cache-Control") {
 		for token := range strings.SplitSeq(line, ",") {
 			parts := strings.SplitN(strings.TrimSpace(token), "=", 2)
@@ -63,13 +82,20 @@ func parseCacheControl(h http.Header) cacheDirectives {
 				d.Private = true
 			case "must-revalidate":
 				d.MustRevalidate = true
+			case "proxy-revalidate":
+				d.ProxyRevalidate = true
 			case "public":
 				d.Public = true
+			case "max-age":
+				if len(parts) == 2 {
+					if v, err := strconv.ParseInt(strings.TrimSpace(parts[1]), 10, 64); err == nil {
+						d.MaxAge = v
+					}
+				}
 			case "s-maxage":
 				if len(parts) == 2 {
-					if secs, err := strconv.ParseInt(strings.TrimSpace(parts[1]), 10, 64); err == nil && secs >= 0 {
-						v := time.Duration(secs) * time.Second
-						d.SMaxAge = &v
+					if v, err := strconv.ParseInt(strings.TrimSpace(parts[1]), 10, 64); err == nil {
+						d.SMaxAge = v
 					}
 				}
 			}
--- a/filters/cache/cache_control.go
+++ b/filters/cache/cache_control.go
@@ -1,0 +1,79 @@
+package cache
+
+import (
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type cacheDirectives struct {
+	NoStore        bool
+	NoCache        bool
+	Private        bool
+	MustRevalidate bool
+	Public         bool
+	SMaxAge        *time.Duration // RFC 9111 §5.2.2.10; nil means not present
+}
+
+type requestCacheDirectives struct {
+	NoStore      bool
+	NoCache      bool // includes max-age=0
+	OnlyIfCached bool
+}
+
+func parseRequestCacheControl(h http.Header) requestCacheDirectives {
+	var d requestCacheDirectives
+	for _, line := range h.Values("Cache-Control") {
+		for token := range strings.SplitSeq(line, ",") {
+			parts := strings.SplitN(strings.TrimSpace(token), "=", 2)
+			directive := strings.ToLower(strings.TrimSpace(parts[0]))
+			switch directive {
+			case "no-store":
+				d.NoStore = true
+			case "no-cache":
+				d.NoCache = true
+			case "max-age":
+				if len(parts) == 2 && strings.TrimSpace(parts[1]) == "0" {
+					d.NoCache = true
+				}
+			case "only-if-cached":
+				d.OnlyIfCached = true
+			}
+		}
+	}
+	return d
+}
+
+// parseCacheControl parses Cache-Control response directives from h.
+// Uses Header.Values to handle multiple header lines; matches names
+// case-insensitively per RFC 7234 §5.2.
+func parseCacheControl(h http.Header) cacheDirectives {
+	var d cacheDirectives
+	for _, line := range h.Values("Cache-Control") {
+		for token := range strings.SplitSeq(line, ",") {
+			parts := strings.SplitN(strings.TrimSpace(token), "=", 2)
+			directive := strings.ToLower(strings.TrimSpace(parts[0]))
+			switch directive {
+			case "no-store":
+				d.NoStore = true
+			case "no-cache":
+				d.NoCache = true
+			case "private":
+				d.Private = true
+			case "must-revalidate":
+				d.MustRevalidate = true
+			case "public":
+				d.Public = true
+			case "s-maxage":
+				if len(parts) == 2 {
+					if secs, err := strconv.ParseInt(strings.TrimSpace(parts[1]), 10, 64); err == nil && secs >= 0 {
+						v := time.Duration(secs) * time.Second
+						d.SMaxAge = &v
+					}
+				}
+			}
+		}
+	}
+	return d
+}
--- a/filters/cache/cache_control.go
+++ b/filters/cache/cache_control.go
@@ -20,7 +20,7 @@ type cacheDirectives struct {
 
 type requestCacheDirectives struct {
 	NoStore      bool
-	NoCache      bool  // includes max-age=0
+	NoCache      bool // includes max-age=0
 	OnlyIfCached bool
 	MaxStale     int64 // -1 = not present; >= 0 = max staleness seconds client accepts
 	MinFresh     int64 // -1 = not present; >= 0 = min remaining freshness seconds required
--- a/filters/cache/cache_control.go
+++ b/filters/cache/cache_control.go
@@ -8,54 +8,54 @@ import (
 )
 
 type cacheDirectives struct {
-	NoStore         bool
-	NoCache         bool
-	Private         bool
-	MustRevalidate  bool
-	ProxyRevalidate bool
-	Public          bool
-	MaxAge          int64 // -1 = not present; 0 means max-age=0
-	SMaxAge         int64 // -1 = not present
+	noStore         bool
+	noCache         bool
+	private         bool
+	mustRevalidate  bool
+	proxyRevalidate bool
+	public          bool
+	maxAge          int64 // -1 = not present; 0 means max-age=0
+	sMaxAge         int64 // -1 = not present
 }
 
 type requestCacheDirectives struct {
-	NoStore      bool
-	NoCache      bool // includes max-age=0
-	OnlyIfCached bool
-	MaxStale     int64 // -1 = not present; >= 0 = max staleness seconds client accepts
-	MinFresh     int64 // -1 = not present; >= 0 = min remaining freshness seconds required
+	noStore      bool
+	noCache      bool // includes max-age=0
+	onlyIfCached bool
+	maxStale     int64 // -1 = not present; >= 0 = max staleness seconds client accepts
+	minFresh     int64 // -1 = not present; >= 0 = min remaining freshness seconds required
 }
 
 // parseRequestCacheControl parses Cache-Control request directives from h.
 func parseRequestCacheControl(h http.Header) requestCacheDirectives {
-	d := requestCacheDirectives{MaxStale: -1, MinFresh: -1}
+	d := requestCacheDirectives{maxStale: -1, minFresh: -1}
 	for _, line := range h.Values("Cache-Control") {
 		for token := range strings.SplitSeq(line, ",") {
 			parts := strings.SplitN(strings.TrimSpace(token), "=", 2)
 			directive := strings.ToLower(strings.TrimSpace(parts[0]))
 			switch directive {
 			case "no-store":
-				d.NoStore = true
+				d.noStore = true
 			case "no-cache":
-				d.NoCache = true
+				d.noCache = true
 			case "max-age":
 				if len(parts) == 2 && strings.TrimSpace(parts[1]) == "0" {
-					d.NoCache = true
+					d.noCache = true
 				}
 			case "only-if-cached":
-				d.OnlyIfCached = true
+				d.onlyIfCached = true
 			case "max-stale":
 				if len(parts) == 2 {
 					if v, err := strconv.ParseInt(strings.TrimSpace(parts[1]), 10, 64); err == nil && v >= 0 {
-						d.MaxStale = v
+						d.maxStale = v
 					}
 				} else {
-					d.MaxStale = math.MaxInt32 // ~68 years; safe upper bound for *time.Second conversion
+					d.maxStale = math.MaxInt32 // ~68 years; safe upper bound for *time.Second conversion
 				}
 			case "min-fresh":
 				if len(parts) == 2 {
 					if v, err := strconv.ParseInt(strings.TrimSpace(parts[1]), 10, 64); err == nil && v >= 0 {
-						d.MinFresh = v
+						d.minFresh = v
 					}
 				}
 			}
@@ -68,34 +68,34 @@ func parseRequestCacheControl(h http.Header) requestCacheDirectives {
 // Uses Header.Values to handle multiple header lines; matches names
 // case-insensitively per RFC 9111 §5.2.
 func parseCacheControl(h http.Header) cacheDirectives {
-	d := cacheDirectives{MaxAge: -1, SMaxAge: -1}
+	d := cacheDirectives{maxAge: -1, sMaxAge: -1}
 	for _, line := range h.Values("Cache-Control") {
 		for token := range strings.SplitSeq(line, ",") {
 			parts := strings.SplitN(strings.TrimSpace(token), "=", 2)
 			directive := strings.ToLower(strings.TrimSpace(parts[0]))
 			switch directive {
 			case "no-store":
-				d.NoStore = true
+				d.noStore = true
 			case "no-cache":
-				d.NoCache = true
+				d.noCache = true
 			case "private":
-				d.Private = true
+				d.private = true
 			case "must-revalidate":
-				d.MustRevalidate = true
+				d.mustRevalidate = true
 			case "proxy-revalidate":
-				d.ProxyRevalidate = true
+				d.proxyRevalidate = true
 			case "public":
-				d.Public = true
+				d.public = true
 			case "max-age":
 				if len(parts) == 2 {
 					if v, err := strconv.ParseInt(strings.TrimSpace(parts[1]), 10, 64); err == nil {
-						d.MaxAge = v
+						d.maxAge = v
 					}
 				}
 			case "s-maxage":
 				if len(parts) == 2 {
 					if v, err := strconv.ParseInt(strings.TrimSpace(parts[1]), 10, 64); err == nil {
-						d.SMaxAge = v
+						d.sMaxAge = v
 					}
 				}
 			}
--- a/filters/cache/cache_control_test.go
+++ b/filters/cache/cache_control_test.go
@@ -46,16 +46,17 @@ func TestParseCacheControl(t *testing.T) {
 		header http.Header
 		want   cacheDirectives
 	}{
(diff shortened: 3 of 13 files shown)


filters/cache/filter.go
```diff
@@ -0,0 +1,231 @@
+package cache
```
szuecs, line 1: please create a `docs.go` with only `package cache` and have some package doc:
"Package cache implements a caching family of filters."

This will fix the ST1000

filters/cache/lru_storage.go
```diff
@@ -0,0 +1,69 @@
+package cache
+
+import (
+	"bytes"
+	"context"
+	"encoding/gob"
```
szuecs: are you sure you want to depend on `gob`?
IIRC it depends on Go version so if you update skipper built by a different Go version you could have the problem of a panic or at least failed Marshal/Unmarshal.

Maybe I am wrong, but I remember Arpad said this once to me.

larry-dalmeida: I updated just to be safe. JSON is marginally slower (acceptable imo) and larger but has compatibility guarantees.

larry-dalmeida:
Will update the demo screenshot shortly with the latest run

docs/reference/filters.md
```diff
@@ -3681,3 +3681,73 @@ pr: Path("/products/:productId")
...
+
+Force mode with stale-if-error fallback:
+
+```
+-> cache("5m", "15s", "30s", "60s") -> "https://cdn.example.org"
+```
+
+Force mode with per-tenant cache key isolation:
+
+```
```
larry-dalmeida, line 3747: Recieved feedback, will update docs to clarify:

1.  that logic to construct cache key is not transparent.
2. no explicit guidance on safety of usage - can I cache customer info? is it aware of other filters (auth)? could it cache that by accident?

larry-dalmeida: added

docs/reference/filters.md
```diff
@@ -3681,3 +3681,96 @@ pr: Path("/products/:productId")
...
+The filter stores whatever the upstream returns and serves it to any request
+matching the same key. It has no awareness of other filters in the chain.
+
+* **Only use on routes where the response is identical for all callers** —
+  public CMS content, unauthenticated APIs, static assets.
+* **Do not use on routes that return user-specific data or set user-scoped
+  cookies** — those responses will be served to other users.
+* **`Authorization` is not in the key by default.** Requests from different
+  users hitting the same path share a cache entry unless `Authorization` is
+  added to `keyHeaders` or stripped before the filter.
```
a4180p: If `Authorization` is ignored by default, how can removal of the header affect cache key?

docs/reference/filters.md
```diff
@@ -3681,3 +3681,96 @@ pr: Path("/products/:productId")
...
+* **`Cache-Control: private` is ignored in force mode.** Audit the upstream
+  response before enabling force mode on any authenticated route.
+
+!!! note
+    The LRU store is shared across all `cache()` filter instances created from
+    the same spec. The storage budget is divided evenly across 256 internal
```
a4180p: what does "same spec" means in this context?

filters/cache/filter.go
```diff
@@ -0,0 +1,971 @@
...
+		method := ctx.Request().Method
+		if (method == http.MethodGet || method == http.MethodHead) && evaluateConditionals(ctx.Request(), entry) {
+			notModified := &http.Response{
+				StatusCode: http.StatusNotModified,
+				Header:     rsp.Header.Clone(),
+				Body:       http.NoBody,
+				Request:    ctx.Request(), // link response to originating request per net/http convention
+			}
+			ctx.Serve(notModified)
+			go f.revalidate(key, ctx.Request())
```
a4180p: `ctx.Request()` will return mutated request if there are any filters changing the request earlier in the chain. Is that expected behaviour?

docs/reference/filters.md
```diff
@@ -3681,3 +3681,96 @@ pr: Path("/products/:productId")
...
+
+Force mode with per-tenant cache key isolation:
+
+```
+-> cache("5m", "15s", "30s", "0s", "X-Tenant-ID") -> "https://cdn.example.org"
+```
+
+**Cache key**
+
+The key is derived from route ID + HTTP method + host + path + query string.
```
szuecs: IIRC you use sha256sum to hash data to get a key. I think it would be great to tell here, because if you later write about Authorization header in the cache key, everyone with a security lens to the docs would question what we do.

docs/reference/filters.md
szuecs: It's likely developer internal knowledge: spec is the thing that creates a filter instance by `spec.CreateFilter()`.
These are user docs, so people won't understand this. Maybe tell only the first part:
```
The LRU store is shared across all `cache()` filter instances.
The storage budget...
```

docs/reference/filters.md
```diff
@@ -3681,3 +3681,96 @@ pr: Path("/products/:productId")
...
+* **`Authorization` is not in the key by default.** Requests from different
+  users hitting the same path share a cache entry unless `Authorization` is
+  added to `keyHeaders` or stripped before the filter.
+* **`Cache-Control: private` is ignored in force mode.** Audit the upstream
+  response before enabling force mode on any authenticated route.
+
+!!! note
+    The LRU store is shared across all `cache()` filter instances created from
+    the same spec. The storage budget is divided evenly across 256 internal
+    shards; a single entry larger than one shard's budget is dropped with a
```
szuecs: likely we want to expose a metric, too

filters/cache/cache_control.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"math"
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+type cacheDirectives struct {
+	NoStore         bool
```
szuecs: I wonder why we need these exported on a private struct.

filters/cache/lru.go
```diff
@@ -0,0 +1,162 @@
...
+type lruShard struct {
+	maxBytes     int64
+	currentBytes int64
+	ll           *list.List
+	cache        map[string]*list.Element
+	onEvict      func() // called once per evicted item; nil = no-op
+
+	// mu is a standard Mutex (not RWMutex) because LRU reads (Get)
+	// modify linked list. Contention is mitigated by sharding.
+	mu sync.Mutex
```
szuecs: please put the mutex member on top of what it protects with empty lines as separator for unprotected things

filters/cache/lru.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	return s
+}
+
+// getShard routes a key to its shard via FNV-1a. Keys are expected to be
+// pre-hashed (e.g. SHA-256) by the caller.
+func (s *ShardedByteLRU) getShard(key string) *lruShard {
+	h := fnv.New32a()
+	h.Write([]byte(key))
+	return s.shards[h.Sum32()%shardCount]
```
szuecs: Please use this:
```
s.shards[xxhash.Sum64String(key)%shardCount]

File: filters/cache/filter.go [hunk 12]
```diff
@@ -0,0 +1,971 @@
...
+		listenAddr: listenAddr,
+		client:     sknet.NewClient(sknet.Options{}),
+		storage:    NewLRUStorage(maxBytes, func() { metrics.Default.IncCounter("lru_eviction") }),
+	}
+}
+
+type cacheSpec struct {
+	maxBytes   int64
+	listenAddr string
+	client     *sknet.Client
```
szuecs: sknet -> skpnet (we use it in other places)

filters/cache/filter.go
```diff
@@ -0,0 +1,971 @@
...
+	log "github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/metrics"
+	sknet "github.com/zalando/skipper/net"
+	"golang.org/x/sync/singleflight"
+)
+
+const (
+	Name        = "cache"
+	filterName  = Name
```
szuecs: please add the "Name" in filters/filter.go and use this as reference

filters/cache/filter.go
```diff
@@ -0,0 +1,971 @@
...
+	listenAddr string
+	client     *sknet.Client
+	storage    Storage // shared across all filter instances
+}
+
+func (s *cacheSpec) Name() string { return filterName }
+
+func (s *cacheSpec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) != 0 && (len(args) < 3 || len(args) > 5) {
+		return nil, fmt.Errorf("cache: expected 0 or 3-5 args (ttl, errorTTL, swrWindow[, staleIfError[, keyHeaders]]), got %d", len(args))
```
szuecs: all errors in CreateFilter that are caused by invalid parameters have to be `ErrInvalidFilterParameters`
You can do: `fmt.Errorf("...: %w", ErrInvalidFilterParameters)`

filters/cache/filter.go
```diff
@@ -0,0 +1,971 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("cache: arg 1 (errorTTL): %w", err)
+		}
+
+		swr, err = toDuration(args[2])
+		if err != nil {
+			return nil, fmt.Errorf("cache: arg 2 (swrWindow): %w", err)
+		}
+
+		if len(args) >= 4 {
```
szuecs, line 114: you can close the other if before this one for more readable code

filters/cache/filter.go
```diff
@@ -0,0 +1,971 @@
...
+			return nil, fmt.Errorf("cache: arg 4 (keyHeaders): expected string, got %T", args[4])
+		}
+		for _, h := range strings.Split(raw, ",") {
+			if name := strings.TrimSpace(h); name != "" {
+				keyHeaders = append(keyHeaders, http.CanonicalHeaderKey(name))
+			}
+		}
+	}
+
+	m := metrics.Default
```
szuecs: you don't need a local var for this :)

filters/cache/filter.go
```diff
@@ -0,0 +1,971 @@
...
+	cf := &cacheFilter{
+		storage:      s.storage,
+		listenAddr:   s.listenAddr,
+		ttl:          ttl,
+		errorTTL:     errorTTL,
+		swrWindow:    swr,
+		staleIfError: staleIfError,
+		rfcMode:      rfcMode,
+		metrics:      m,
```
szuecs: `metrics: metrics.Default,`

filters/cache/filter.go
```diff
@@ -0,0 +1,971 @@
...
+
+type cacheFilter struct {
+	storage      Storage
+	listenAddr   string
+	ttl          time.Duration
+	errorTTL     time.Duration
+	swrWindow    time.Duration
+	staleIfError time.Duration
+	keyHeaders   []string // request headers folded into the base cache key
+	// rfcMode true: upstream Cache-Control is authoritative (cache()).
```
szuecs, line 185: please add a newline above this one

filters/cache/filter.go
```diff
@@ -0,0 +1,971 @@
...
+	// false: operator ttl/errorTTL/swrWindow are authoritative (force mode).
+	rfcMode bool
+	sf      singleflight.Group // cold-miss coalescing
```
szuecs: Maybe `coldSF` ?

filters/cache/filter.go
```diff
@@ -0,0 +1,971 @@
...
+//	-> cache("5m", "15s", "30s") -> "https://example.org"
+//
+// Combining force mode with stale-if-error:
+//
+//	-> cache("5m", "15s", "30s", "60s") -> "https://example.org"
+func NewCacheFilter(maxBytes int64, listenAddr string) filters.Spec {
+	return &cacheSpec{
+		maxBytes:   maxBytes,
+		listenAddr: listenAddr,
+		client:     sknet.NewClient(sknet.Options{}),
```
szuecs: we need to pass some options, likely best is to put `skpnet.Options` into the parameter list of NewCacheFilter() and initialize in skipper.go by some of the options we already got from skipper.Options (no need to add more config, but we can select define some hard coded values and re-use existing config where it makes sense (CloseIdleTimeout we can re-use and passing the tracer makes also sense)

filters/cache/filter.go
```diff
@@ -0,0 +1,971 @@
...
+// Response() can populate the cache.
+func (f *cacheFilter) Request(ctx filters.FilterContext) {
+	baseKey := cacheKey(ctx.RouteId(), ctx.Request(), f.keyHeaders)
+	key := baseKey
+	if sentinel, _ := f.storage.Get(ctx.Request().Context(), "vary:"+baseKey); sentinel != nil && len(sentinel.VaryHeaders) > 0 {
+		key = varyKey(baseKey, ctx.Request(), sentinel.VaryHeaders)
+	}
+	ctx.StateBag()[stateBagKey] = key
+	ctx.StateBag()["cache:base-key"] = baseKey
+	ctx.StateBag()["cache:request-time"] = time.Now()
```
szuecs: all stateBag Keys should use a const

filters/cache/filter.go
```diff
@@ -0,0 +1,971 @@
...
+
+	// Unsafe methods skip cache lookup; Response() will invalidate the entry.
+	if isUnsafeMethod(ctx.Request().Method) {
```
szuecs, line 212: should this not be the first line of the func?

filters/cache/filter.go
```diff
@@ -0,0 +1,971 @@
...
+		return
+	}
+
+	reqDir := parseRequestCacheControl(ctx.Request().Header)
+	if reqDir.OnlyIfCached {
+		entry, err := f.storage.Get(ctx.Request().Context(), key)
+		if err != nil || entry == nil || entry.IsStale(time.Now()) {
+			ctx.Serve(&http.Response{
+				StatusCode: http.StatusGatewayTimeout,
+				Header:     http.Header{"X-Cache-Status": {"MISS"}},
```
szuecs: The key and all values of this header should be `const`

filters/cache/filter.go
```diff
@@ -0,0 +1,971 @@
...
+				Body:       http.NoBody,
+			})
+			return
+		}
+	} else if reqDir.NoCache || reqDir.NoStore {
+		if reqDir.NoStore {
+			ctx.StateBag()["cache:no-store"] = true
```
szuecs: also a const

filters/cache/filter.go
szuecs: There are 2 problems here:
1) It's unsafe to pass a reference to Request this in a goroutine, which could lead to panic, see also https://github.com/golang/go/issues/53808#issuecomment-1183496047
2) We spawn a goroutine per request, this will lead to 2x goroutines also code path via line 239 2x

In general: why do we need to revalidate per request? 
For me this sounds more like a single background goroutine per filter, that walks the cache items to revalidate.

larry-dalmeida:
@szuecs @a4180p thanks for the comments 🧡 Was delayed due to conflicting priorities. I am addressing the comments and expect to be complete with most if not all by today.

docs/reference/filters.md
larry-dalmeida: Good catch. To clarify: Authorization is excluded from cache key by default. 
Two requests that differ only in their Authorization header share same cache entry. 
Removing header entirely has no effect on key. Rewritten paragraph now makes default (excluded) explicit and frames `keyHeaders` as an opt-in for per client isolation.

docs/reference/filters.md
larry-dalmeida: Fixed - removed "same spec" and reworded as suggested.

filters/cache/filter.go
larry-dalmeida: Both points are resolved in latest changes:
  
@a4180p: Yes, using the mutated request is intentional - cache key is built from the already-filtered request (mutations by earlier filters like auth header stripping are reflected in the key), so the revalidation request must match that same key. Added a code comment to make this explicit.

@szuecs : Both issues are fixed. The goroutine-per-request problem is solved by a single background goroutine per filter instance that drains a bounded channel (revalJobs, capacity 256). The request is cloned in the calling goroutine before enqueueing, so there is no data race on original `*http.Request`. Concurrent revalidations for the same key are further coalesced via singleflight (revalSF) in the worker, so at most one upstream fetch is in flight per key at any time.

docs/reference/filters.md
larry-dalmeida: thanks! Good call out, fixed.

docs/reference/filters.md
larry-dalmeida: Done - `lru_oversized` (when entry is too large), `lru_eviction` counter (incremented on every eviction) and `lru_bytes` gauge (current bytes in use across all shards, updated on every eviction) are exposed via `metrics.Default`. Also documented in the note.

filters/cache/cache_control.go
larry-dalmeida: Fixed - unexported all fields on `cacheDirectives` and `requestCacheDirectives`. No external references exist as the types are package-internal only.

filters/cache/lru.go
larry-dalmeida: Fixed - `mu` now sits directly above the fields it protects. `maxBytes` and `onEvict` are immutable after construction and appear above with an empty line as separator.

filters/cache/lru.go
larry-dalmeida: Fixed - thanks! that is much better (faster, readable, better distro across shards)

filters/cache/filter.go
larry-dalmeida: Fixed

filters/cache/filter.go
larry-dalmeida: Fixed - wrapped with `fmt.Errorf`

filters/cache/filter.go
larry-dalmeida: Fixed

filters/cache/filter.go
larry-dalmeida: Fixed - keyHeaders parsing `(args[4])` is now in a separate top-level if `len(args) == 5` block, closing the `if len(args) >= 3` block earlier and removing one level of nesting.

filters/cache/filter.go
larry-dalmeida: Fixed

filters/cache/filter.go
larry-dalmeida: Fixed - same change: m removed, metrics.Default used directly.

filters/cache/filter.go
larry-dalmeida: Fixed

filters/cache/filter.go
larry-dalmeida: Fixed - NewCacheFilter now accepts skpnet.Options as a third parameter. In skipper.go the filter is registered after tracer resolution (moved out of filterRegistry() into Run()), passing CloseIdleConnsPeriod → IdleConnTimeout, IdleConnectionsPerHost → MaxIdleConnsPerHost, resolved tracer, and OpenTracingClientTraceByTag.

filters/cache/filter.go
larry-dalmeida: Fixed - all state bag keys are defined as named constants and used consistently throughout.

filters/cache/filter.go
larry-dalmeida: Fixed. `isUnsafeMethod` check now comes after the key is computed and stored in the state bag, but before the cache lookup. The key needs to be set for all requests as Response() reads it to perform invalidation. 

Moving the check before key computation would require recomputing it in Response(), duplicating the SHA-256 hash. Intent of skipping the cache lookup for write requests is satisfied; the key storage is the necessary setup, not the lookup itself.

filters/cache/filter.go
larry-dalmeida: Fixed - cacheStatusHeader and cacheStatusMiss are named constants. Used here and consistently throughout the file.

filters/cache/filter.go
larry-dalmeida: Fixed - stateBagNoStore is a named constant. used here and in Response().

larry-dalmeida:
FYI - Will squash the commits once changes are approved.

filters/cache/filter.go
larry-dalmeida: Done

a4180p:
>  FYI - Will squash the commits once changes are approved.

That's not required, we use squash merge.
PR description will be used as commit message

larry-dalmeida:
@a4180p @szuecs done with changes but I see [ClusterFuzzLite PR fuzzing](https://github.com/zalando/skipper/actions/runs/26211044879/job/77121770158?pr=3991) failing due to flakiness, ~I'm trying to identify root cause, likely related to previous package update and I think it has nothing to with my PR. Let me know if you've seen this before.~ Ah I see I need to rebase.

szuecs:
> @a4180p @szuecs done with changes but I see [ClusterFuzzLite PR fuzzing](https://github.com/zalando/skipper/actions/runs/26211044879/job/77121770158?pr=3991) failing due to flakiness, ~I'm trying to identify root cause, likely related to previous package update and I think it has nothing to with my PR. Let me know if you've seen this before.~ Ah I see I need to rebase.

Yeah the fuzzing is not your problem and the flaky tests also not. Focus on the tests in the area you changed something.

docs/reference/filters.md
```diff
@@ -3681,3 +3681,104 @@ pr: Path("/products/:productId")
...
+## Caching
+
+### cache
+
+Caches HTTP responses in a shared in-memory LRU store. Operates in two modes
+selected by the number of arguments.
+
+**RFC mode** (zero arguments): upstream `Cache-Control` is fully authoritative.
+Freshness, `no-store`, `private`, `Vary`, `s-maxage`, heuristic TTL
+(RFC 9111 §4.2.2), and conditional revalidation (`ETag`/`Last-Modified`) are
```
szuecs: Can you add a link to the RFC here. please?

docs/reference/filters.md
```diff
@@ -3681,3 +3681,104 @@ pr: Path("/products/:productId")
...
+Unsafe methods (`POST`, `PUT`, `DELETE`, `PATCH`) invalidate the cached entry on
+success. `HEAD 200` freshens stored headers without replacing the body.
+
+Parameters (force mode):
+
+* `ttl` – freshness duration for 200 responses (string, e.g. `"5m"`)
+* `errorTTL` – freshness duration for 404/5xx responses (string)
+* `swrWindow` – stale-while-revalidate window after TTL expires (string)
+* `staleIfError` – optional; how long a stale entry may be served on upstream
+  5xx (string, RFC 5861)
```
szuecs: TODO: add rfc link

docs/reference/filters.md
```diff
@@ -3681,3 +3681,104 @@ pr: Path("/products/:productId")
...
+
+!!! note
+    Three metrics track LRU behaviour: `lru_eviction` (counter, incremented each
+    time an entry is evicted due to memory pressure), `lru_bytes` (gauge,
+    updated on every eviction to reflect current storage usage in bytes), and
+    `lru_oversized` (counter, incremented when an entry is too large to fit in
+    any shard and is silently dropped).
+
+!!! note
+    `s-maxage` implies `proxy-revalidate` per RFC 9111 §5.2.2.10: stale entries
```
szuecs: TODO: add link to RFC Paragraph

filters/cache/cache_control.go
```diff
@@ -0,0 +1,105 @@
...
+				if len(parts) == 2 && strings.TrimSpace(parts[1]) == "0" {
+					d.noCache = true
+				}
+			case "only-if-cached":
+				d.onlyIfCached = true
+			case "max-stale":
+				if len(parts) == 2 {
+					if v, err := strconv.ParseInt(strings.TrimSpace(parts[1]), 10, 64); err == nil && v >= 0 {
+						d.maxStale = v
+					}
```
szuecs, line 51: should we have here also an "else" part?
The client can trigger the else branch, by setting `max-stale=3.4` and then `d.maxStale` has the zero value, which is `0`. Maybe it's ok, maybe not.

filters/cache/cache_control.go
```diff
@@ -0,0 +1,105 @@
...
+				} else {
+					d.maxStale = math.MaxInt32 // ~68 years; safe upper bound for *time.Second conversion
+				}
+			case "min-fresh":
+				if len(parts) == 2 {
+					if v, err := strconv.ParseInt(strings.TrimSpace(parts[1]), 10, 64); err == nil && v >= 0 {
```
szuecs, line 57: here we have in both if cases no else branch, maybe it's fine.

filters/cache/cache_control.go
```diff
@@ -0,0 +1,105 @@
...
+				d.private = true
+			case "must-revalidate":
+				d.mustRevalidate = true
+			case "proxy-revalidate":
+				d.proxyRevalidate = true
+			case "public":
+				d.public = true
+			case "max-age":
+				if len(parts) == 2 {
+					if v, err := strconv.ParseInt(strings.TrimSpace(parts[1]), 10, 64); err == nil {
```
szuecs, line 91: Do we need "else"?

filters/cache/cache_control.go
```diff
@@ -0,0 +1,105 @@
...
+						d.maxAge = v
+					}
+				}
+			case "s-maxage":
+				if len(parts) == 2 {
+					if v, err := strconv.ParseInt(strings.TrimSpace(parts[1]), 10, 64); err == nil {
```
szuecs, line 97: Do we need "else"?

filters/cache/cache_control_test.go
```diff
@@ -0,0 +1,68 @@
...
+func TestParseRequestCacheControl(t *testing.T) {
+	cases := []struct {
+		header       string
+		noStore      bool
+		noCache      bool
+		onlyIfCached bool
+	}{
+		{"no-store", true, false, false},
+		{"no-cache", false, true, false},
+		{"max-age=0", false, true, false},
```
szuecs: what if we have `max-age=0,max-age=5` or `max-age=3.4` ?

filters/cache/cache_control_test.go
```diff
@@ -0,0 +1,68 @@
...
+		{"no-cache", http.Header{"Cache-Control": {"no-cache"}}, cacheDirectives{noCache: true, maxAge: -1, sMaxAge: -1}},
+		{"private", http.Header{"Cache-Control": {"private"}}, cacheDirectives{private: true, maxAge: -1, sMaxAge: -1}},
+		{"must-revalidate", http.Header{"Cache-Control": {"must-revalidate"}}, cacheDirectives{mustRevalidate: true, maxAge: -1, sMaxAge: -1}},
+		{"comma-separated", http.Header{"Cache-Control": {"no-store, must-revalidate"}}, cacheDirectives{noStore: true, mustRevalidate: true, maxAge: -1, sMaxAge: -1}},
+		{"multiple lines", http.Header{"Cache-Control": {"no-cache", "must-revalidate"}}, cacheDirectives{noCache: true, mustRevalidate: true, maxAge: -1, sMaxAge: -1}},
+		{"case-insensitive", http.Header{"Cache-Control": {"NO-STORE"}}, cacheDirectives{noStore: true, maxAge: -1, sMaxAge: -1}},
+		{"value suffix stripped", http.Header{"Cache-Control": {`no-cache="x-private"`}}, cacheDirectives{noCache: true, maxAge: -1, sMaxAge: -1}},
+		{"empty", http.Header{}, cacheDirectives{maxAge: -1, sMaxAge: -1}},
+		{"max-age=3600", http.Header{"Cache-Control": {"max-age=3600"}}, cacheDirectives{maxAge: 3600, sMaxAge: -1}},
```
szuecs, line 71: another test case with `max-age=3.4` and having 2 different `max-age` would be interesting.

filters/cache/filter.go
```diff
@@ -0,0 +1,1028 @@
...
+		}
+	}
+
+	var keyHeaders []string
+	if len(args) == 5 {
+		raw, ok := args[4].(string)
+		if !ok {
+			return nil, fmt.Errorf("cache: arg 4 (keyHeaders): expected string, got %T: %w", args[4], filters.ErrInvalidFilterParameters)
+		}
+		for _, h := range strings.Split(raw, ",") {
```
szuecs:
````
for  h := range strings.SplitSeq(raw, ",") {
```

filters/cache/filter.go
szuecs: As long as the implementation of `f.storage.Get()` (see line 204) is not a call to valkey or something like this and only an in-memory lookup, I tend to agree. However if we do or plan to do a network roundtrip in storage.Get(), we should calculate the SHA-256 two times as this is seriously fast compared to network call and also if we have to hold a lock to access an in-memory cache it will be cheaper to calculate SHA256.

filters/cache/filter.go
```diff
@@ -0,0 +1,1028 @@
...
+	// fresh entry. Strip the header so it never reaches the upstream.
+	if ctx.Request().Header.Get(revalidateHeader) != "" {
+		ctx.Request().Header.Del(revalidateHeader)
+		return
+	}
+
+	reqDir := parseRequestCacheControl(ctx.Request().Header)
+	if reqDir.onlyIfCached {
+		entry, err := f.storage.Get(ctx.Request().Context(), key)
+		if err != nil || entry == nil || entry.IsStale(time.Now()) {
```
szuecs, line 228: What happens if context of the request was canceled?
The client will not get the response below, because it already hang up the connection.

filters/cache/filter.go
```diff
@@ -0,0 +1,1028 @@
...
+			return
+		}
+	} else if reqDir.noCache || reqDir.noStore {
+		if reqDir.noStore {
+			ctx.StateBag()[stateBagNoStore] = true
+		}
+		return
+	}
+
+	entry, err := f.storage.Get(ctx.Request().Context(), key)
```
szuecs: If `reqDir.onlyIfCached ` is true we had already the `entry` and we would fetch it again.

filters/cache/filter.go
```diff
@@ -0,0 +1,1028 @@
...
+		rsp := &http.Response{
+			StatusCode: entry.StatusCode,
+			Header:     entry.Header.Clone(),
+			Body:       io.NopCloser(bytes.NewReader(entry.Payload)),
+		}
+		rsp.Header.Set(cacheStatusHeader, cacheStatusMiss)
+		ctx.Metrics().IncCounter("miss")
+		ctx.Serve(headBodyOmitted(ctx.Request().Method, rsp))
+	case <-ctx.Request().Context().Done():
+		// Client disconnected. Remaining waiters still receive their result.
```
szuecs, line 407: 👍

filters/cache/filter.go
```diff
@@ -0,0 +1,1028 @@
...
+	"Trailer":             true,
+	"Transfer-Encoding":   true,
+	"Upgrade":             true,
+}
+
+// stripHopByHop removes hop-by-hop headers from h in-place.
+// Also removes any header named in the Connection value per RFC 9110 §7.6.1.
+func stripHopByHop(h http.Header) {
+	for _, name := range h.Values("Connection") {
+		for _, field := range strings.Split(name, ",") {
```
szuecs:
```
for field := range strings.SplitSeq(name, ",") {
```

filters/cache/cache_control.go
larry-dalmeida: No `else` needed. [RFC 9111 §4.2.1](https://www.rfc-editor.org/rfc/rfc9111.html#name-calculating-freshness-lifet) says treat non-integer freshness values as stale - `max-stale=3.4` falling through to zero (`max-stale=0`, accept no staleness) is correct and RFC-aligned. Will add test coverage to document behaviour explicitly.

filters/cache/cache_control.go
larry-dalmeida: No `else` needed - parse failure leaving `-1` (absent) is fine.

filters/cache/cache_control.go
larry-dalmeida: No `else` needed - [RFC 9111 §4.2.1](https://www.rfc-editor.org/rfc/rfc9111.html#name-calculating-freshness-lifet) says treat non-integer `max-age` as stale. sentinel `-1` (absent) achieves same result downstream.

filters/cache/cache_control.go
larry-dalmeida: Same as above - no `else` needed.

filters/cache/cache_control_test.go
larry-dalmeida: Will add both cases.

filters/cache/filter.go
larry-dalmeida: Storage is in-memory only — no network roundtrip. Will add comment to revisit if remote backend is introduced.

filters/cache/filter.go
larry-dalmeida: Good catch - will add `ctx.Request().Context().Err() != nil` guard before `storage.Get()` at line 241.

filters/cache/filter.go
larry-dalmeida: Good catch - will restructure so unconditional `storage.Get()` is in `else` branch, skipping second lookup when `onlyIfCached` hit is fresh.

filters/cache/filter.go
larry-dalmeida: Note: Guard placed in else branch only (normal miss path). `onlyIfCached` skipped intentionally: path is direct lookup + serve with no `coalesce()` call, so cost of one in-memory lookup on cancelled 
context is negligible vs added branching complexity to correctly serve 504. `noCache/noStore` paths return before touching storage - no guard needed there. Let me know if that sounds good to you.

filters/cache/cache_control_test.go
larry-dalmeida: Added both cases. Note that request max-age with non-zero values is silently discarded - only `=0` is handled (maps to noCache). [RFC 9111 §5.2.1](https://www.rfc-editor.org/rfc/rfc9111.html#name-max-age) marks request directives as advisory (MAY), so not a violation, but intentional behaviour for shared caches as honouring arbitrary client max-age on shared routes is a cache-bypass attack vector. 

Updated docs to list which request directives are honoured and note the max-age limitation explicitly.

filters/cache/cache_control_test.go
larry-dalmeida: Added :)

larry-dalmeida:
Recommended changes are complete. None of the pipeline failures are related to my changes. I am addressing the pipeline flakiness in a separate PR.

filters/cache/filter.go
szuecs: The guard in l241 is good.
The thing is that on client cancel you can not deliver anything (no correct 504 for example) anyways, so simple is better.

filters/cache/filter.go
szuecs: Yeah that's fine

filters/cache/storage.go
```diff
@@ -0,0 +1,40 @@
...
+	StaleIfError time.Duration
+}
+
+// IsStale returns true when the entry is past its TTL but still within the SWR window.
+func (e *Entry) IsStale(now time.Time) bool {
+	return now.After(e.CreatedAt.Add(e.TTL)) && now.Before(e.CreatedAt.Add(e.TTL+e.StaleWhileRevalidate))
+}
+
+type Storage interface {
+	// Should return (nil, nil) if the key is not found
```
szuecs: can you do proper godoc comments please for interface and listed functions?

filters/cache/storage.go
```diff
@@ -0,0 +1,63 @@
...
+
+// IsStale reports whether the entry is past its TTL but still within the
+// stale-while-revalidate window relative to now.
+func (e *Entry) IsStale(now time.Time) bool {
+	return now.After(e.CreatedAt.Add(e.TTL)) && now.Before(e.CreatedAt.Add(e.TTL+e.StaleWhileRevalidate))
+}
+
+// Storage is the backing store abstraction for cached entries. Implementations
+// must be safe for concurrent use.
+type Storage interface {
```
larry-dalmeida:
```suggestion
// Storage is the backing store abstraction for cached entries.
// Implementations must be safe for concurrent use.
type Storage interface {
```

larry-dalmeida:
👍

szuecs:
:+1:

PR #3992, opened 2026-04-30
routesrv: fix zone aware routing table filtering

MustafaSaber:
There are three conditions where we are silently dropping routes:
1. If route endpoints isn't >= minEndpointsByZone
2. If route does not have any endpoints
3. If route does not have any endpoints for a zone

We should add all routes, and if routes endpoints match the condition then we filter the endpoints to only get `zone` endpoints only.

First commit is a reproducer test case (pipeline should fail)


dataclients/kubernetes/testdata/ingressV1/zone-aware-traffic/mixed-zone-threshold.yaml
```diff
@@ -0,0 +1,179 @@
+apiVersion: apps/v1
+kind: Deployment
```
szuecs: Do we read anywhere the deployment?
I think we do not need this

dataclients/kubernetes/testdata/ingressV1/zone-aware-traffic/mixed-zone-threshold.yaml
```diff
@@ -0,0 +1,179 @@
...
+    - 10.3.1.5
+    zone: eu-central-1b
+ports:
+  - name: web
+    port: 8080
+    protocol: TCP
+apiVersion: v1
+---
+apiVersion: apps/v1
+kind: Deployment
```
szuecs: also this one we do not need

dataclients/kubernetes/testdata/ingressV1/zone-aware-traffic/mixed-zone-threshold.yaml
```diff
@@ -0,0 +1,179 @@
...
+    kubernetes.io/service-name: svc-below-threshold
+  name: svc-below-threshold-eps
+  namespace: default
+endpoints:
+  - addresses:
+    - 10.4.0.3
+    zone: eu-central-1a
+  - addresses:
+    - 10.4.0.4
+    zone: eu-central-1a
```
szuecs: both addresses are in zone A and we test for "below threshold", but this would be the same if we would have zone aware configuration but we want to test the fallback.

dataclients/kubernetes/testdata/routegroups/zone-aware-traffic/mixed-zone-threshold.yaml
```diff
@@ -0,0 +1,175 @@
...
+    kubernetes.io/service-name: svc-below-threshold
+  name: svc-below-threshold-eps
+  namespace: default
+endpoints:
+  - addresses:
+    - 10.4.0.3
+    zone: eu-central-1a
+  - addresses:
+    - 10.4.0.4
+    zone: eu-central-1a
```
szuecs: same problem we are running in zone A and have only pods in zone A and so we do not test that we also add pods from other zones...

routesrv/testdata/zone-aware-traffic/mixed-zone-threshold.yaml
```diff
@@ -0,0 +1,127 @@
...
+    kubernetes.io/service-name: svc-below-threshold
+  name: svc-below-threshold-eps
+  namespace: default
+endpoints:
+  - addresses:
+    - 10.4.0.3
+    zone: eu-central-1a
+  - addresses:
+    - 10.4.0.4
+    zone: eu-central-1a
```
szuecs: same problem here with the test data

dataclients/kubernetes/testdata/routegroups/zone-aware-traffic/mixed-zone-threshold.yaml
szuecs: fixed

dataclients/kubernetes/testdata/ingressV1/zone-aware-traffic/mixed-zone-threshold.yaml
szuecs: fixed

routesrv/testdata/zone-aware-traffic/mixed-zone-threshold.yaml
szuecs: fixed

dataclients/kubernetes/testdata/ingressV1/zone-aware-traffic/mixed-zone-threshold.yaml
szuecs: cleaned up

dataclients/kubernetes/testdata/ingressV1/zone-aware-traffic/mixed-zone-threshold.yaml
szuecs: cleaned up

szuecs:
👍

routesrv/polling.go
MustafaSaber: I think it's better to start with default then override it, other than the else branch

```go
for zone, eps := range byZone {
  toAddRoute := r
  if len(eps) >= minEndpointsByZone {
    rCopy := *r
    rCopy.LBEndpoints = eps
    toAddRoute =  &rCopy
  }
  zoneAwareRoutes[zone] = append(zoneAwareRoutes[zone], toAddRoute)
}
```

routesrv/routesrv_test.go
```diff
@@ -113,6 +113,15 @@ func getRoutes(rs *routesrv.RouteServer) *httptest.ResponseRecorder {
 	return w
 }
 
+func getRoutesGzip(rs *routesrv.RouteServer) *httptest.ResponseRecorder {
+	w := httptest.NewRecorder()
```
MustafaSaber: You can use `return getRoutesWithRequestHeadersSetting(rs, map[string]string{"Accept-Encoding": "gzip"})`

routesrv/polling.go
szuecs: you create more garbage like you suggest and you create a side effect, so IMO the code right now is better in all cases.

MustafaSaber: Can you explain why 😄?

routesrv/routesrv_test.go
szuecs: I think it looks much better like we do here. It's much clearer than what you suggest.

routesrv/polling.go
szuecs: setting "route" is a side effect you create and you need to copy another pointer for every iteration, so you create additional memory usage.
So "garbage" meant memory that you have to GC in the next run again.

routesrv/routesrv_test.go
greeshma1196: It looks better this way, but Mustafa is right, its essentially duplicated code.

MustafaSaber: We don't need to remove the new function but it can be a wrapper to the old one

routesrv/polling.go
MustafaSaber: Since it's a matter of taste we can ignore this, but I don't think this will be a problem for GC or cause performance degradation

routesrv/routesrv_test.go
greeshma1196: Addressed

routesrv/testdata/zone-aware-traffic/mixed-zone-threshold-all-endpoints.eskip
```diff
@@ -0,0 +1,4 @@
+kube_default__ingress_above_threshold__above_example_org____svc_above_threshold: Host("^(above[.]example[.]org[.]?(:[0-9]+)?)$")
```
szuecs, line 1: The filename "-all-routes" should be changed because it is miss leading.
For example: "-all-endpoints" or "-all" name suffix would be fine. It has nothing to do that we have less or more routes, it's only the LB members that change.

routesrv/routesrv_test.go
```diff
@@ -1401,3 +1413,55 @@ func TestZoneAwareFallbackToAllRoutesWhenBelowThreshold(t *testing.T) {
 	assert.Equal(t, fullCount, plainResp.Header().Get(routing.RoutesCountName))
 	assert.Equal(t, fullCount, gzipResp.Header().Get(routing.RoutesCountName))
 }
+
+func TestXCountShouldBeSameForZoneAwareAndZoneUnawareRoutes(t *testing.T) {
+	defer tl.Reset()
+
+	// Serve full routing table
```
szuecs: there is nothing "full" here, maybe "all endpoints" or "not zone-aware"

routesrv/routesrv_test.go
```diff
@@ -1401,3 +1413,55 @@ func TestZoneAwareFallbackToAllRoutesWhenBelowThreshold(t *testing.T) {
...
+	defer rs.StopUpdates()
+	require.NoError(t, tl.WaitFor(routesrv.LogRoutesInitialized, waitTimeout))
+
+	expectedAllRoutes := parseEskipFixture(t, "testdata/zone-aware-traffic/mixed-zone-threshold-all-routes.eskip")
+
+	plainRespAllRoutes := getRoutes(rs)
+	require.Equal(t, http.StatusOK, plainRespAllRoutes.Code)
+	gotPlain, err := eskip.Parse(plainRespAllRoutes.Body.String())
+	require.NoError(t, err)
+	assert.Equal(t, expectedAllRoutes, gotPlain, "expected all routes from /routes endpoint")
```
szuecs: `expected all routes from /routes endpoint` -> `expected routes from /routes endpoint`

szuecs: there is no "all" routes, same is true for all occurrences below and above here.

routesrv/testdata/zone-aware-traffic/mixed-zone-threshold-all-endpoints.eskip
greeshma1196: I agree, "all routes" is misleading, will update.

szuecs:
:+1:

routesrv/polling.go
```diff
@@ -130,7 +130,11 @@ func (p *poller) poll(wg *sync.WaitGroup) {
 func filterRoutesByZone(routes []*eskip.Route) map[string][]*eskip.Route {
 	zoneAwareRoutes := make(map[string][]*eskip.Route)
 	for _, r := range routes {
-		byZone := make(map[string][]*eskip.LBEndpoint)
+		byZone := map[string][]*eskip.LBEndpoint{
```
MustafaSaber: This is not right, skipper is not always deployed in `eu-central` specially for OSS. I think it can be handled later without this pre-intitialization

greeshma1196: Removed the hardcoded values.

greeshma1196:
👍🏽

MustafaSaber:
👍

greeshma1196:
👍🏽

PR #3995, opened 2026-05-05
test: add flowstate test 

szuecs:
test: add flowstate test that show what happens if secret was deleted. We can create and extract states if we once created and Encrypter.


szuecs:
:+1:

ayruslore (approved):

PR #3998, opened 2026-05-07
fix: create an access log instance instead of pkg global

szuecs:
fix: create an access log instance instead of pkg global

Library users that would use logging.LogAccess will need to use logging.NewLogAccess() and use inst.LogAccess()
I think this won't be widely in use, because skipper.Run() will do it automatically.

Migration for skipper as library users in case you initialize the logger yourself (very unlikely the case as we automatically do this in a not exported func), replace:
```
logging.Init(logging.Options{
...
skipper.Options{
...
}
```
by
```
lg := logging.NewAccessLogger(logging.Options{
...

skipper.Options{
...
    AccessLogger: accessLogger,
...
```


a4180p:
👍

logging/access.go
```diff
@@ -144,8 +141,8 @@ func hash(val string) uint64 {
 
 // LogAccess logs an access event in Apache combined log format (with a minor customization with the duration).
 // Additional allows to provide extra data that may be also logged, depending on the specific log format.
-func LogAccess(entry *AccessEntry, additional map[string]interface{}) {
-	if accessLog == nil || entry == nil {
+func (alog *AccessLogger) LogAccess(entry *AccessEntry, additional map[string]any) {
+	if entry == nil {
```
MustafaSaber, line 145: shouldn't we also check `if alog != nil`?

MustafaSaber:
Also we need to update version for the breaking changes

logging/access.go
szuecs: We do this in the proxy. In general it's an anti pattern to check "this == nil".
It's a coding error if it happens and it should not be hidden if you have a coding error.

szuecs:
> Also we need to update version for the breaking changes

done

a4180p:
👍

szuecs:
:+1:

PR #4001, opened 2026-05-08
Async decision logging for OPA

wisinghe:
## Async decision logging for OPA

Adds an opt-in async decision logging mode (--enable-open-policy-agent-async-decision-logging) that offloads decision log calls to a background goroutine, preventing slow log plugins (e.g. eopa_dl writing to S3) from adding latency to HTTP responses.

When enabled, each OPA instance maintains a bounded channel. logDecision does a non-blocking send into it; if the buffer is full the event is dropped with a warning rather than blocking the request goroutine. A dedicated runDecisionLogger goroutine drains the channel and performs the actual log call. On shutdown, Close drains the channel to completion before stopping the manager, respecting the context deadline.

The channel buffer size is derived from decision_logs.reporting.buffer_size_limit_events in the OPA config template, keeping it consistent with OPA's own buffer configuration. If not set, it defaults to 1000.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -345,7 +345,7 @@ type Config struct {
 	EnableOpenPolicyAgentDataPreProcessingOptimization bool          `yaml:"enable-open-policy-agent-data-preprocessing-optimization"`
 	EnableOpenPolicyAgentPreloading                    bool          `yaml:"enable-open-policy-agent-preloading"`
 	EnableOpenPolicyAgentAsyncDecisionLogging          bool          `yaml:"enable-open-policy-agent-async-decision-logging"`
-	OpenPolicyAgentDecisionLogBufferSize               int           `yaml:"open-policy-agent-decision-log-buffer-size"`
+	OpenPolicyAgentDecisionLogQueueSize                int           `yaml:"open-policy-agent-decision-log-queue-size"`
 	EnableOpenPolicyAgentPrintTracing                  bool          `yaml:"enable-open-policy-agent-print-tracing"`
 	OpenPolicyAgentConfigTemplate                      string        `yaml:"open-policy-agent-config-template"`
 	OpenPolicyAgentEnvoyMetadata                       string        `yaml:"open-policy-agent-envoy-metadata"`
@@ -615,7 +615,7 @@ func NewConfig() *Config {
 	flag.BoolVar(&cfg.EnableOpenPolicyAgentCustomControlLoop, "enable-open-policy-agent-custom-control-loop", false, "when enabled skipper will use a custom control loop to orchestrate certain opa behaviour (like the download of new bundles) instead of relying on periodic plugin triggers")
 	flag.BoolVar(&cfg.EnableOpenPolicyAgentPreloading, "enable-open-policy-agent-preloading", false, "EXPERIMENTAL: when enabled, OPA instances will be pre-loaded during route processing instead of during filter creation, making filter creation non-blocking")
 	flag.BoolVar(&cfg.EnableOpenPolicyAgentAsyncDecisionLogging, "enable-open-policy-agent-async-decision-logging", false, "when enabled, decision log events are offloaded to a background goroutine so that slow decision log plugins (e.g. eopa_dl with S3 output) do not delay HTTP responses")
-	flag.IntVar(&cfg.OpenPolicyAgentDecisionLogBufferSize, "open-policy-agent-decision-log-buffer-size", openpolicyagent.DefaultDecisionLogChannelSize, "number of decision log events that can be queued before new ones are dropped; only applies when async decision logging is enabled")
+	flag.IntVar(&cfg.OpenPolicyAgentDecisionLogQueueSize, "open-policy-agent-decision-log-queue-size", openpolicyagent.DefaultDecisionLogQueueSize, "number of decision log events that can be queued before new ones are dropped; only applies when async decision logging is enabled")
 	flag.DurationVar(&cfg.OpenPolicyAgentControlLoopInterval, "open-policy-agent-control-loop-interval", openpolicyagent.DefaultControlLoopInterval, "Interval between the execution of the control loop. Only applies if the custom control loop is enabled")
 	flag.DurationVar(&cfg.OpenPolicyAgentControlLoopMaxJitter, "open-policy-agent-control-loop-max-jitter", openpolicyagent.DefaultControlLoopMaxJitter, "Maximum jitter to add to the control loop interval. Only applies if the custom control loop is enabled")
 	flag.BoolVar(&cfg.EnableOpenPolicyAgentDataPreProcessingOptimization, "enable-open-policy-agent-data-preprocessing-optimization", false, "As a latency optimization, open policy agent will read values from in-memory storage as pre converted ASTs, removing conversion overhead at evaluation time. Currently experimental and if successful will be enabled by default")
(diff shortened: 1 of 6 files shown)


filters/openpolicyagent/openpolicyagent.go
```diff
@@ -607,6 +607,21 @@ func (registry *OpenPolicyAgentRegistry) newOpenPolicyAgentInstance(bundleName s
...
+	input  interface{}
+	result *envoyauth.EvalResult
+	err    error
+	sync   chan struct{}
+}
+
+// defaultDecisionLogBufferSize is the number of pending decision log tasks that
+// can be queued before new ones are dropped. Each entry is a small struct; the
+// real memory cost is the decision payload held by *envoyauth.EvalResult.
+const defaultDecisionLogBufferSize = 1000
```
torwunder: ideally this is configurable

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -607,6 +607,21 @@ func (registry *OpenPolicyAgentRegistry) newOpenPolicyAgentInstance(bundleName s
 	return engine, nil
 }
 
+// decisionLogTask holds everything needed to call logDecision off the hot path.
+// If sync is non-nil it is a flush sentinel: the worker closes it and skips logDecision.
+type decisionLogTask struct {
+	ctx    context.Context
```
torwunder: is there something from the context we need in the decisionLogger?

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -607,6 +607,21 @@ func (registry *OpenPolicyAgentRegistry) newOpenPolicyAgentInstance(bundleName s
...
+	input  interface{}
+	result *envoyauth.EvalResult
+	err    error
+	sync   chan struct{}
```
torwunder: what is this used for?

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -979,10 +1002,31 @@ func (opa *OpenPolicyAgentInstance) Close(ctx context.Context) {
 	opa.closingOnce.Do(func() {
 		opa.Logger().Info("Closing OPA instance...")
 		opa.closing = true
+		close(opa.decisionLogChan)
+		select {
+		case <-opa.decisionLogDone:
+		case <-ctx.Done():
```
torwunder: why do we wait for this context?

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -765,8 +783,13 @@ func (registry *OpenPolicyAgentRegistry) new(store storage.Store, bundleName str
 
 		idGenerator: uniqueIDGenerator,
 		logger:      logging.Get().WithFields(map[string]any{"bundle-name": bundleName}),
+
+		decisionLogChan: make(chan decisionLogTask, defaultDecisionLogBufferSize),
+		decisionLogDone: make(chan struct{}),
```
torwunder: decisionsProcessed?

filters/openpolicyagent/openpolicyagent.go
wisinghe: It is handed over to the function in the opa library https://github.com/open-policy-agent/opa/blob/main/v1/plugins/logs/plugin.go#L716

filters/openpolicyagent/openpolicyagent.go
wisinghe: It's actually not used. I will refactor. 

The explanation from claude:

>  Here's the intent behind it: it was designed to be a flush/barrier mechanism. The idea is:                                                                                         
>                                                                                                                                                                                      
>   1. You send a sentinel task (with sync set to a channel, and no real log data) into decisionLogChan.                                                                               
>   2. Because the channel is FIFO, this sentinel sits behind all previously enqueued real tasks.
> The sync field is defined and handled in the worker, but waitDecisionLogDrained is never implemented and sync is never set anywhere — it's dead code. The comment references a function that doesn't exist.

filters/openpolicyagent/openpolicyagent.go
wisinghe: I believe it makes sure Close does not hang forever if the `decisionLogDone` takes longer.

> The select gives the logger a chance to finish its work, while ctx.Done() provides an escape hatch so a misbehaving or slow logger cannot prevent the instance (and ultimately the process) from terminating within its grace period.

filters/openpolicyagent/openpolicyagent.go
torwunder: yes but as the as this is async do we really need to drag the context there? 
not sure if we would want the decision logging to be cancelled if the context is cancelled here

wisinghe: I think now I understand your point. It looks like it was passed over not for any specific data but for the purpose of in case a plugin needs to be aware of the context getting cancelled which in our case we don't need.

As we build the struct with context.WithoutCancel(ctx) [here](https://github.com/zalando/skipper/pull/4001/changes#diff-291cdf60ccfdc27702f75d92c2fdbd4d3bede9dce027b736c92fb65e9a1da792R60), decision logging will not be cancelled with the initial context. But it is not needed for the struct. Will remove and use a context.WithoutCancel at the point we call the opa.logDecision

config/config.go
```diff
@@ -344,6 +344,8 @@ type Config struct {
 	OpenPolicyAgentControlLoopMaxJitter                time.Duration `yaml:"open-policy-agent-control-loop-max-jitter"`
 	EnableOpenPolicyAgentDataPreProcessingOptimization bool          `yaml:"enable-open-policy-agent-data-preprocessing-optimization"`
 	EnableOpenPolicyAgentPreloading                    bool          `yaml:"enable-open-policy-agent-preloading"`
+	EnableOpenPolicyAgentAsyncDecisionLogging          bool          `yaml:"enable-open-policy-agent-async-decision-logging"`
+	OpenPolicyAgentDecisionLogQueueSize                int           `yaml:"open-policy-agent-decision-log-queue-size"`
```
mjungsbluth: The alternative is to use the existing opaconfig as we do for other cases (from the decision_log section).

filters/openpolicyagent/evaluation.go
```diff
@@ -53,9 +53,24 @@ func (opa *OpenPolicyAgentInstance) Eval(ctx context.Context, req *ext_authz_v3.
 			return
 		}
 
-		err := opa.logDecision(ctx, input, result, err)
-		if err != nil {
-			opa.Logger().WithFields(map[string]interface{}{"err": err}).Error("Unable to log decision to control plane.")
+		if opa.decisionLogChan != nil {
```
mjungsbluth: Can we not hide this in the logDecision() method?

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -791,6 +819,21 @@ func allPluginsReady(allPluginsStatus map[string]*plugins.Status, pluginNames ..
...
+func decisionLogBufferSize(rawConfig json.RawMessage) int {
+	const defaultBufferSize = 1000
+	if rawConfig == nil {
+		return defaultBufferSize
+	}
+	var cfg logs.Config
+	if err := json.Unmarshal(rawConfig, &cfg); err != nil {
+		return defaultBufferSize
+	}
+	if cfg.Reporting.BufferSizeLimitEvents != nil && *cfg.Reporting.BufferSizeLimitEvents > 0 {
```
mjungsbluth, line 831: What happens if `cfg.Reporting` is nil?

wisinghe: It's defined as **Reporting       ReportingConfig      `json:"reporting"`** 
Since it's a value type, it cannot be nil.
Calude explanation: 

> When the opaconfig does not define decision_logs.reporting, cfg.Reporting will just be the zero value of ReportingConfig with all fields at their zero values.

mjungsbluth:
👍

filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
```diff
@@ -829,6 +829,258 @@ func TestAuthorizeRequestFilterWithS3DecisionLogPlugin(t *testing.T) {
...
+
+	r := eskip.MustParse(fmt.Sprintf(`* -> opaAuthorizeRequest("test") -> %q`, clientServer.URL))
+	proxy := proxytest.New(fr, r...)
+	defer proxy.Close()
+
+	req, err := http.NewRequest(http.MethodGet, proxy.URL+"/allow", nil)
+	require.NoError(t, err)
+
+	start := time.Now()
+	rsp, err := proxy.Client().Do(req)
```
MustafaSaber, line 928: I think we can make use of vegeta here (we already use vegeta client) and latency mean should be less that s3delay

MustafaSaber (approved):

PR #4007, opened 2026-05-12
Add `decision_log.dropped` counter for async decision logging

x418:
## What & Why?

Follow-up to #4001, which introduced a bounded async channel in front of `logDecision` so that slow log plugins (e.g. `eopa_dl` writing to S3) don't add latency to HTTP responses. When the channel is full, events are silently dropped.

Currently, those drops are only visible as warn log lines. This PR surfaces the same event as a Prometheus counter so we can:

- **Alert** when drops exceed a threshold (should we decide on one based on operational experience)
- **Tune `buffer_size_limit_events`** based on observed drop rate.

The counter is emitted via skipper's standard `metrics.Default`:

```
skipper_custom_total{key="decision_log.dropped.<bundle>"}
```

to match the existing OPA counter convention (`decision.allow.<bundle>`, `decision.deny.<bundle>`, `decision.err.<bundle>`). 

_Given naming is hard, please feel free to suggest a better name for the metric._

### DIFF
--- a/filters/openpolicyagent/evaluation.go
+++ b/filters/openpolicyagent/evaluation.go
@@ -18,6 +18,8 @@ import (
 	"github.com/open-policy-agent/opa/v1/topdown/print"
 	"github.com/opentracing/opentracing-go"
 	pbstruct "google.golang.org/protobuf/types/known/structpb"
+
+	"github.com/zalando/skipper/metrics"
 )
 
 func (opa *OpenPolicyAgentInstance) Eval(ctx context.Context, req *ext_authz_v3.CheckRequest) (*envoyauth.EvalResult, error) {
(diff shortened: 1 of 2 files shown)


wisinghe:
👍

szuecs:
@x418 please fix DCO (--sign-off commits)

x418:
> @x418 please fix DCO (--sign-off commits)

Done

szuecs:
> > @x418 please fix DCO (--sign-off commits)
> 
> Done

the checker does not say so.
Did you do: `git commit --signoff --amend ` ?

x418:
> > > @x418 please fix DCO (--sign-off commits)
> > 
> > 
> > Done
> 
> the checker does not say so. Did you do: `git commit --signoff --amend ` ?

My bad. I use a different email for GitHub and that caused GH to declare that my commit wasn't signed. Fixed now

szuecs:
:+1:

MustafaSaber (approved):

PR #4011, opened 2026-05-15
proxy: always extract baggage header

MustafaSaber:
The OTelBaggage predicate reads baggage via baggage.FromContext, but skipper never extracted the incoming baggage HTTP header into the request context. Both the OT bridge and OT tracers discard baggage extracted from the HTTP header when no valid trace context is present in the same request.

Fix by always calling propagation.Baggage{}.Extract after span setup in ServeHTTP, which always populates the context from the baggage header independently of whether a tracecontext header is present. The W3C propagator also handles percent-decoding (%3D -> =), so OTelBaggage matches the logical value regardless of how the upstream encoded it.

### DIFF
--- a/proxy/breaker_with_teeloopback_test.go
+++ b/proxy/breaker_with_teeloopback_test.go
@@ -31,7 +31,7 @@ func TestBreakerLeak(t *testing.T) {
 		CircuitBreakers:      circuit.NewRegistry(settings...),
 	}
 
-	tp, err := newCustomTestProxy(nil, fr, doc, params, nil)
+	tp, err := newTestProxyWithFiltersAndParams(fr, doc, params, nil)
 	if err != nil {
 		t.Error(err)
 		return
--- a/proxy/breaker_with_teeloopback_test.go
+++ b/proxy/breaker_with_teeloopback_test.go
@@ -31,7 +31,7 @@ func TestBreakerLeak(t *testing.T) {
 		CircuitBreakers:      circuit.NewRegistry(settings...),
 	}
 
-	tp, err := newTestProxyWithFiltersAndParams(fr, doc, params, nil)
+	tp, err := newCustomTestProxy(nil, fr, doc, params, nil)
 	if err != nil {
 		t.Error(err)
 		return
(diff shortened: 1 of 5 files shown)


proxy/breaker_with_teeloopback_test.go
```diff
@@ -31,7 +31,7 @@ func TestBreakerLeak(t *testing.T) {
 		CircuitBreakers:      circuit.NewRegistry(settings...),
 	}
 
-	tp, err := newTestProxyWithFiltersAndParams(fr, doc, params, nil)
+	tp, err := newCustomTestProxy(nil, fr, doc, params, nil)
```
szuecs: why? see below

proxy/proxy_test.go
```diff
@@ -208,7 +208,7 @@ func (srw *syncResponseWriter) Len() int {
 	return srw.body.Len()
 }
 
-func newTestProxyWithFiltersAndParams(fr filters.Registry, doc string, params Params, preprocs []routing.PreProcessor) (*testProxy, error) {
+func newCustomTestProxy(predicates []routing.PredicateSpec, fr filters.Registry, doc string, params Params, preprocs []routing.PreProcessor) (*testProxy, error) {
```
szuecs, line 211: please do not replace but add a new function and call the new one from the old one with "nil" as first parameter.

```
func newTestProxyWithFiltersAndParams(fr filters.Registry, doc string, params Params, preprocs []routing.PreProcessor) (*testProxy, error) {
   return newCustomTestProxy(nil, fr, doc, params, preprocs)
}
```

proxy/tracing_test.go
```diff
@@ -814,6 +816,62 @@ func TestSetTagWithEmptySpan(t *testing.T) {
...
+			Tracer: tracer,
+		},
+	})
+	require.NoError(t, err)
+	defer tp.close()
+
+	ps := httptest.NewServer(tp.proxy)
+	defer ps.Close()
+
+	for _, tc := range []struct {
```
szuecs, line 824: everything before this should be in the body of the test.
This all takes some µs to run so we should not have here shared state so we can not run this later in parallel.

szuecs:
:+1:

a4180p:
👍

PR #4016, opened 2026-05-18
shortcheck: fix some testcases

MustafaSaber:
Use os.MkdirTemp with a short path to avoid hitting the Unix socket path length limit (104 chars on macOS) when t.TempDir() produces a long path.


greeshma1196:
👍🏽

MustafaSaber:
👍

PR #4017, opened 2026-05-19
ratelimiter/leakybucket: Adds Valkey support

madumalt:
This introduces Valkey support for LeakyBucket Ratelimiter. When the Valkey support was introduced somehow we have missed this.


ratelimit/leakybucket_redis.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+const (
+	leakyBucketRedisKeyPrefix = "lkb."
+	leakyBucketMetricPrefix   = "leakybucket.redis."
+	leakyBucketMetricLatency  = leakyBucketMetricPrefix + "latency"
+	leakyBucketSpanName       = "redis_leakybucket"
+)
+
+func newClusterLeakyBucket(ringClient *net.RedisRingClient, capacity int, emission time.Duration, now func() time.Time) *ClusterLeakyBucket {
```
szuecs: let's rename it to `newClusterLeakyBucketRedis`

ratelimit/leakybucket_redis.go
```diff
@@ -0,0 +1,99 @@
...
+	"time"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
+
+	"github.com/zalando/skipper/metrics"
+	"github.com/zalando/skipper/net"
+)
+
+type ClusterLeakyBucket struct {
```
szuecs, line 16: Rename the type to `ClusterLeakyBucketRedis`, it's anyways internal enough. We can increase minor via VERSION file and tell in the release notes that if you use `ratelimit.ClusterLeakyBucket` then you need to replace it with ...

ratelimit/leakybucket_redis.go
```diff
@@ -0,0 +1,99 @@
...
+func (b *ClusterLeakyBucket) add(ctx context.Context, label string, increment int, now time.Time) (added bool, retry time.Duration, err error) {
+	r, err := b.ringClient.RunScript(ctx, b.script,
+		[]string{b.getBucketId(label)},
+		b.capacity,
+		b.emission.Microseconds(),
+		increment,
+		now.UnixMicro(),
+	)
+
+	if err == nil {
```
szuecs: I know you just copied it but I would prefer to have the basic Go way:
- error on the right
- ok path on the left
style of code.

so:
```
if err != nil {
   return
}
x := r.(int64)
if x >= 0 {
	added, retry = true, 0
} else {
	added, retry = false, -time.Duration(x)*time.Microsecond
}
return
```

ratelimit/leakybucket_valkey.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (b *ClusterLeakyBucketValkey) add(ctx context.Context, label string, increment int, now time.Time) (added bool, retry time.Duration, err error) {
+	msg, err := b.ringClient.RunScript(ctx, b.script,
+		[]string{b.getBucketId(label)},
+		strconv.FormatInt(int64(b.capacity), 10),
+		strconv.FormatInt(b.emission.Microseconds(), 10),
+		strconv.Itoa(increment),
+		strconv.FormatInt(now.UnixMicro(), 10),
+	)
+	if err == nil {
```
szuecs: same as in the redis case

ratelimit/leakybucket_redis.go
madumalt: (done)

ratelimit/leakybucket_redis.go
madumalt: (done)

ratelimit/leakybucket_valkey.go
madumalt: (done)

szuecs:
:+1:

madumalt:
👍

PR #4018, opened 2026-05-19
Make testcontainer based tests runnable with Podman/Colima container runtimes on MacOs

madumalt:
Updates readme.md instructing disable ryuk when running tests with Podman/Colima container runtimes and updates the test helpers from container IPs to host plus mapped ports so Podman/Colima on macOS can reach containers. Linux developers are not affected because Docker on Linux already supports this access pattern and tests continue to work the same way.


madumalt:
The fix make other tests fail, fixing them require quite heavy change of skipper. Couldn't find a simpler way of doing it. Given that gain is not that much there is no point in pursing any longer in this one.

PR #4022, opened 2026-05-22
Fix pipeline flakiness

larry-dalmeida:
# Description

Fixes two pre-existing flaky CI failures unrelated to the cache filter changes.

**`net`: `TestValkeyClient/With_metrics_and_AddrUpdater`**

The test used `time.Sleep(2*UpdateInterval - elapsed)` before asserting the shard count. On a loaded CI runner, `RingAvailable()` (a real network ping) takes >100ms, collapsing the sleep below one ticker period - both tick 1 and tick 2 fire before the assertion, leaving the metric at 1 shard instead of 2.

Fix: extract into a standalone `TestValkeyClientMetricsUpdater` with its own `addressUpdater` instance (eliminates shared-`n` coupling across sub-tests) and replace the sleep with a polling loop with a 10-interval deadline.

**`proxy`: `TestBackendTimeoutWithSlowBodyWriterShadow`**

The 80ms client timeout used for the final health-check is too tight when 500 goroutines are running under `-race` (2–5× slower). `TestBackendTimeoutWithSlowBodyShadow`, the passing sibling, uses 800ms. Widened to 500ms.

### DIFF
--- a/net/valkey_test.go
+++ b/net/valkey_test.go
@@ -446,8 +446,6 @@ func TestValkeyClient(t *testing.T) {
 					t.Fatalf("Failed to create ring client: %v", err)
 				}
 			}
-			startedUpdater := time.Now()
-
 			defer func() {
 				if !cli.closed {
 					t.Error("Failed to close valkey ring client")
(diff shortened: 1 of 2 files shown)


szuecs (approved):

a4180p:
👍

PR #4023, opened 2026-05-22
Add decompressRequest filter for handling compressed request body

Abdelghafour01:
### Add `decompressRequest` filter

Adds a new filter `decompressRequest()` that decompresses the request body on the request path, mirroring the existing `decompress()` filter that operates on response bodies.

#### Motivation

Skipper currently supports decompressing response bodies via the `decompress()` filter, but there is no equivalent for request bodies. Backends that don't natively understand compressed request payloads (e.g. clients sending `Content-Encoding: gzip` on `POST`/`PUT`) can't be transparently fronted by Skipper today. This filter closes that gap.

#### Changes

- New filter decompressrequest.go implementing `decompressRequest()`.
  - Reads the request's `Content-Encoding` header and streams the body through the appropriate decoder.
  - Supported encodings: `gzip`, `deflate`, `br`, `zstd` (reuses `newDecodedBody`, `getEncodings`, `encodingsSupported` from the existing `decompress` filter — no duplication of decoder logic or pools).
  - On success: removes `Content-Encoding` and `Content-Length` headers and sets `ContentLength = -1` so the request is sent to the backend as a streamed, decompressed body.
  - On unsupported encoding: leaves the request unmodified and sets `filter::decompress::not-possible` in the state bag.
  - On init failure: closes the body, replaces it with `http.NoBody`, sets both `filter::decompress::not-possible` and `filter::decompress::error` in the state bag, and logs the error. Same state-bag keys as `decompress()` for consistency.
- Registered the new spec in builtin.go and added the `DecompressRequestName` constant in filters.go.
- Tests in decompressrequest_test.go covering:
  - successful gzip decompression and header/`ContentLength` cleanup
  - request without `Content-Encoding` is passed through unchanged
  - unsupported encoding sets `DecompressionNotPossible` and preserves headers
  - invalid gzip payload sets both `DecompressionNotPossible` and `DecompressionError`
  - filter name
- Documentation entry added under filters.md.

#### Note

- The state-bag keys (`DecompressionNotPossible`, `DecompressionError`) are intentionally shared with `decompress()`; if you'd prefer request-specific keys to disambiguate, happy to split them.

### DIFF
--- a/filters/builtin/decompressrequest_test.go
+++ b/filters/builtin/decompressrequest_test.go
@@ -1,0 +1,131 @@
+package builtin
+
+import (
+	"bytes"
+	"compress/gzip"
+	"io"
+	"net/http"
+	"testing"
+
+	"github.com/zalando/skipper/filters/filtertest"
+)
+
+func gzipped(t *testing.T, payload string) *bytes.Buffer {
+	t.Helper()
+	var buf bytes.Buffer
+	gz := gzip.NewWriter(&buf)
+	if _, err := gz.Write([]byte(payload)); err != nil {
+		t.Fatalf("failed to write gzip data: %v", err)
+	}
+	if err := gz.Close(); err != nil {
+		t.Fatalf("failed to close gzip writer: %v", err)
+	}
+	return &buf
+}
+
+func newDecompressRequestContext(t *testing.T, body io.Reader, encoding string) *filtertest.Context {
+	t.Helper()
+	req, err := http.NewRequest("POST", "http://example.com", body)
+	if err != nil {
+		t.Fatalf("failed to create request: %v", err)
+	}
+	if encoding != "" {
+		req.Header.Set("Content-Encoding", encoding)
+	}
+	return &filtertest.Context{
+		FRequest:  req,
+		FStateBag: make(map[string]interface{}),
+	}
+}
+
+func TestDecompressRequestGzip(t *testing.T) {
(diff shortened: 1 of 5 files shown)


filters/builtin/decompressrequest_test.go
```diff
@@ -0,0 +1,131 @@
...
+		t.Fatalf("failed to create filter: %v", err)
+	}
+	f.Request(ctx)
+
+	if _, ok := ctx.StateBag()[DecompressionNotPossible]; !ok {
+		t.Error("expected DecompressionNotPossible to be set in state bag")
+	}
+	if got := ctx.Request().Header.Get("Content-Encoding"); got != "unsupported" {
+		t.Errorf("expected Content-Encoding to be preserved when unsupported, got: %q", got)
+	}
```
szuecs, line 106: Can you please also check the body contains "x"?

filters/builtin/decompressrequest_test.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func TestDecompressRequestInvalidPayload(t *testing.T) {
```
szuecs, line 109: If we detect this we should respond by a 400, right?
If so I would say we should also test it.

szuecs:
@Abdelghafour01 thanks for the pr, looks great and there are only a bit of ideas what we can further improve by testing.
I didn't read any rfc regarding this but I would expect that there will be an rfc that tells what to do in case of header says gzip but body is plain. I am not sure if Go already detecting this case but if not please check and reference the rfc if you can find it.

PR #4024, opened 2026-05-22
Handle errors from SetAddrs in Valkey ring updater

a4180p:
We were ignoring errors returned from `ring.SetAddr`, so [if the creation of the new valkey client failed](https://github.com/zalando/skipper/blob/8fb2ebe40f5f7aafe9db010c85c5efc8bc283ac3/net/valkey.go#L222), the `clientMap` stays in the old (or incomlete) state, but in the [updater](https://github.com/zalando/skipper/blob/8fb2ebe40f5f7aafe9db010c85c5efc8bc283ac3/net/valkey.go#L483) it was considered successful and we never tried to recreate the client or update clientMap.

### DIFF
--- a/net/valkey.go
+++ b/net/valkey.go
@@ -434,9 +434,13 @@ func (vrc *ValkeyRingClient) startUpdater(ctx context.Context) {
 
 	init := true
 	if len(old) != 0 {
-		vrc.SetAddrs(ctx, old)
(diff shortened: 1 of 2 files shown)


madumalt (approved):

szuecs:
:+1:

PR #4026, opened 2026-05-22
Feature: mtls support in httpclient - net.Client 

szuecs:
Example skipper/net.Client with mTLS and cert rotation
```
rootCAs, _ := x509.SystemCertPool()
caCertPEM, _ := os.ReadFile("path/to/your/custom-ca.crt")
rootCAs.AppendCertsFromPEM(caCertPEM)

net.NewClient(net.Options{
	CertFile:            certFile,
	KeyFile:             keyFile,
	CertRefreshInterval: time.Minute,
	RootCAs:             rootCAs,
})
```

### DIFF
--- a/net/httpclient.go
+++ b/net/httpclient.go
@@ -45,7 +45,7 @@ type clientTraceTimeT int
 
 const clientTraceTime clientTraceTimeT = 1
 
-type certReloader struct {
+type CertReloader struct {
 	cert        atomic.Pointer[tls.Certificate]
 	sp          *secrets.SecretPaths
 	certFile    string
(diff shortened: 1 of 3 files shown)


net/httpclient.go
```diff
@@ -74,11 +181,32 @@ func NewClient(o Options) *Client {
...
 		sp := secrets.NewSecretPaths(o.BearerTokenRefreshInterval)
 		if err := sp.Add(o.BearerTokenFile); err != nil {
-			o.Log.Errorf("failed to read secret: %v", err)
+			o.Log.Errorf("Failed to read secret: %v", err)
 		}
 		sr = secrets.NewStaticDelegateSecret(sp, o.BearerTokenFile)
 	}
 
+	var cr *CertReloader
+	if o.CertFile != "" && o.KeyFile != "" && o.Transport == nil {
```
a4180p, line 190: Should we also check for o.RootCAs here?

szuecs: If the default is nil then it will use the system available CAs, `tls.Config.RootCAs`:
```
	// RootCAs defines the set of root certificate authorities
	// that clients use when verifying server certificates.
	// If RootCAs is nil, TLS uses the host's root CA set.
	RootCAs *x509.CertPool
```

a4180p:
👍

szuecs:
:+1:

PR #4034, opened 2026-05-26
fix: eliminate race in TestCacheFilter_ColdMissCoalescing

larry-dalmeida:
## Problem

`TestCacheFilter_ColdMissCoalescing` was intermittently failing under `-race` with:

```
--- FAIL: TestCacheFilter_ColdMissCoalescing (0.00s)
    filter_test.go:401: expected 1 upstream fetch, got 2
```

The test asserts that 50 concurrent cold-miss requests coalesce into exactly one upstream fetch via `singleflight.DoChan`.

The previous synchronization used a `sync.WaitGroup` (`wgIn`) where each goroutine called `wgIn.Done()` **before** calling `f.Request()`. The fetch stub blocked on `wgIn.Wait()`, intending to hold the singleflight open until all goroutines had joined it.

This had a [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to-time-of-use) gap: a goroutine could call `wgIn.Done()`, get preempted, and resume while the fetch stub had already unblocked from `wgIn.Wait()` but had not yet called `storage.Set()`. At that point the goroutine would call `storage.Get()` → `nil` and enter `DoChan()`. If the leader's `fetchFn` had returned in the interim, `DoChan()` would start a new singleflight round and fire a second upstream fetch. Under the race detector's overhead this window was wide enough to trigger reliably.

## Fix

Replaced `wgIn` with a `missCounting` storage wrapper that embeds a `sync.WaitGroup`. Each goroutine calls `wg.Done()` inside `storage.Get()` at the exact moment it observes a cache miss - the last possible point before `coalesce()` → `DoChan()`. The fetch stub calls `wg.Wait()`, which durably blocks `fetchFn` until all N goroutines have confirmed a miss and are either inside `DoChan`'s waitlist or in the scheduler-opaque gap immediately before it. Only then does `fetchFn` call `storage.Set()` and return, closing the singleflight round.

The main goroutine's spin loop was replaced with a `fetchStarted` channel closed by the fetch stub on entry, eliminating all busy-waits.

### Before

Goroutine Missed Window-2026-05-26-141411

### After

Goroutine Missed Window-2026-05-26-141426

No production code was changed.

### DIFF
--- a/filters/cache/filter_test.go
+++ b/filters/cache/filter_test.go
@@ -365,8 +365,11 @@ func (mc *missCounting) Get(ctx context.Context, key string) (*Entry, error) {
 func TestCacheFilter_ColdMissCoalescing(t *testing.T) {
 	// Hold the upstream fetch until all N goroutines have performed their
 	// storage.Get and received a miss. At that point every goroutine is either
-	// already inside DoChan's wait list or in the tiny, scheduler-opaque gap
-	// between the nil check and the DoChan call — guaranteeing exactly 1 fetch.
+	// already inside DoChan's wait list or in the scheduler-opaque gap between
+	// the nil check and the DoChan call. Any goroutine that resumes after the
+	// singleflight completes will hit the double-check inside fetchFn (line 382
+	// of filter.go) and return without a second upstream call, keeping fetchCount
+	// at 1.
 	const N = 50
 	f := newTestFilter(t, time.Minute, 15*time.Second, time.Minute)
 
--- a/filters/cache/filter_test.go
+++ b/filters/cache/filter_test.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"io"
 	"net/http"
-	"runtime"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


szuecs:
`noleak`  found a leaked goroutine, likely something not closed as expected.

larry-dalmeida:
@szuecs 

Maybe I am missing something in the logs, but it looks like it's related to `secrets`?
I'm  both failures aren't related to my changes.

filters/cache package passed cleanly:
Screenshot 2026-05-28 at 05 51 49

noleak failure is in the secrets package - a background goroutine in Encrypter.runCipherRefresher running scrypt.Key that isn't shut down before the test exits:

Screenshot 2026-05-28 at 05 52 45

 This PR doesn't touch secrets at all but will check if there is a quick fix for the noleak.

larry-dalmeida:
Fix: https://github.com/zalando/skipper/pull/4036

szuecs (approved):

szuecs:
:+1:

larry-dalmeida:
👍

PR #4035, opened 2026-05-26
introduce timeout for decicion log task

wisinghe:
## Problem

When asyncDecisionLogging is enabled and a new policy bundle is deployed, eopa_dl.Reconfigure() calls Stop() on the running Benthos stream followed by Start() on a new one. If `runDecisionLogger` is inside stream.Consume() on the old stream at that moment, the Benthos producer closure tries to write the decision event to an internal channel whose reader (inproc input) has been destroyed by Stop().
  
Because `doLogDecision` was called with context.Background() (without deadline) the blocked send never returns. `runDecisionLogger` is permanently stuck on that one call and stops draining `decisionLogChan`.

With the goroutine stuck, once the channel fills to capacity (buffer_size_limit_events), every subsequent `logDecision` call hits the non-blocking default branch and drops the event, logging "Decision log dropped: async buffer full." The warning continues indefinitely with no self-recovery. The only fix prior to this change was a pod restart.
 
## Fix
Set timeout for the context passed for decision logging. The value for the timeout is decided using the configured timeout for the eopa_dl output type

### DIFF
--- a/filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
+++ b/filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/zalando/skipper/eskip"
 	"github.com/zalando/skipper/filters"
 	"github.com/zalando/skipper/filters/builtin"
+	nethttptest "github.com/zalando/skipper/net/httptest"
 	"github.com/zalando/skipper/proxy/proxytest"
 	"github.com/zalando/skipper/tracing/tracingtest"
 
@@ -923,20 +924,13 @@ func TestAuthorizeRequestFilterWithS3DecisionLogPlugin_SlowS3BlocksClientRespons
 	proxy := proxytest.New(fr, r...)
 	defer proxy.Close()
 
-	req, err := http.NewRequest(http.MethodGet, proxy.URL+"/allow", nil)
-	require.NoError(t, err)
+	va := nethttptest.NewVegetaAttacker(proxy.URL+"/allow", 10, time.Second, 2*s3Delay)
+	va.Attack(io.Discard, 3*time.Second, t.Name())
 
-	start := time.Now()
-	rsp, err := proxy.Client().Do(req)
-	elapsed := time.Since(start)
-
-	require.NoError(t, err)
-	defer rsp.Body.Close()
-	assert.Equal(t, http.StatusOK, rsp.StatusCode)
-
-	assert.Less(t, elapsed, s3Delay,
-		"client round-trip (%v) should not be delayed by the S3 upload (%v); "+
-			"logDecision must not run on the request goroutine", elapsed, s3Delay)
+	assert.Equal(t, 1.0, va.Success(), "all requests should succeed")
+	assert.Less(t, va.Metrics().Latencies.Mean, s3Delay,
+		"mean client latency (%v) should not be delayed by the S3 upload (%v); "+
+			"logDecision must not run on the request goroutine", va.Metrics().Latencies.Mean, s3Delay)
 }
 
 // TestFullDecisionLogBufferBlocksClientResponse is the regression test for the buffer full scenario when
--- a/filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
+++ b/filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
@@ -7,10 +7,12 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
 	opasdktest "github.com/open-policy-agent/opa/v1/sdk/test"
+	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/zalando/skipper/eskip"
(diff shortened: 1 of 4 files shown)


filters/openpolicyagent/opaauthorizerequest/opaauthorizerequest_test.go
```diff
@@ -921,20 +924,13 @@ func TestAuthorizeRequestFilterWithS3DecisionLogPlugin_SlowS3BlocksClientRespons
...
+	va := nethttptest.NewVegetaAttacker(proxy.URL+"/allow", 10, time.Second, 2*s3Delay)
+	va.Attack(io.Discard, 3*time.Second, t.Name())
 
-	assert.Less(t, elapsed, s3Delay,
-		"client round-trip (%v) should not be delayed by the S3 upload (%v); "+
-			"logDecision must not run on the request goroutine", elapsed, s3Delay)
+	assert.Equal(t, 1.0, va.Success(), "all requests should succeed")
+	assert.Less(t, va.Metrics().Latencies.Mean, s3Delay,
+		"mean client latency (%v) should not be delayed by the S3 upload (%v); "+
+			"logDecision must not run on the request goroutine", va.Metrics().Latencies.Mean, s3Delay)
```
wisinghe, line 933: Addressing the comment on the previous PR https://github.com/zalando/skipper/pull/4001#discussion_r3226592972

wisinghe:
@x418 could you also have a look :)

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -64,6 +64,7 @@ const (
 	defaultShutdownGracePeriod      = 30 * time.Second
 	DefaultOpaStartupTimeout        = 30 * time.Second
 	DefaultBackgroundTaskBufferSize = 100
+	decisionLogTaskTimeoutMin       = 5 * time.Second
```
mjungsbluth: hardcoded? might be fine but still a bit odd.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -1033,13 +1034,34 @@ func (opa *OpenPolicyAgentInstance) Close(ctx context.Context) {
 	})
 }
 
+// decisionLogTaskTimeout returns the per-task timeout for runDecisionLogger.
+// It is bufSize/200 seconds (time to fill the buffer with 200rps),
+// clamped to [decisionLogTaskTimeoutMin, instanceStartupTimeout].
+func decisionLogTaskTimeout(bufSize int, instanceStartupTimeout time.Duration) time.Duration {
+	t := time.Duration(bufSize/200) * time.Second
```
mjungsbluth: Where does the 200 rps come from? There is also an rps setting on the decision log plugin

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -1033,13 +1034,34 @@ func (opa *OpenPolicyAgentInstance) Close(ctx context.Context) {
...
-func (opa *OpenPolicyAgentInstance) runDecisionLogger() {
+func (opa *OpenPolicyAgentInstance) runDecisionLogger(timeout time.Duration) {
 	defer close(opa.decisionsProcessed)
 	for task := range opa.decisionLogChan {
-		if err := opa.doLogDecision(context.Background(), task.input, task.result, task.err); err != nil {
+		// A context with timeout bounds the time runDecisionLogger spends on a single
+		// doLogDecision call. Without this, a Benthos stream.Consume() call on a
+		// stream that was stopped during eopa_dl.Reconfigure() blocks forever because
+		// the inproc reader is gone and context.Background() provides no escape hatch.
+		ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
mjungsbluth: This looses all context from the actual decision, i.e. the spans will create a new trace and not be correlated with the actual decision. 

You might want to use the context that is used further up the call stack, i.e. via the task

wisinghe: using a separate context was a decision taken with the previous PR to completely decouple decision logging process from the request. The decision logging ideally doesn't have to inherit the deadline or cancellation of the request.

WDYT about keeping the new ctx but copying the needed data into the new context from the old?

filters/openpolicyagent/openpolicyagent.go
wisinghe: This is a minimum value and the actual effective timeout depends on the event buffer limit which is configurable. But I will try to change the timeout calculation logic in a way that the hardcoded magic numbers are avoided

filters/openpolicyagent/openpolicyagent.go
wisinghe: 200 rps is actually random. ZPS has ~100 rps per opa in high throughput cases. It would be nice if we can calculate the timeout without having to configure it. 

The logic behind is, if the decision flushing is slow or stuck, the buffer will start filling and then the decisions will start to drop until the timeout. If the timeout is reached before the buffer is full, no decisions will be dropped.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -819,6 +835,43 @@ func allPluginsReady(allPluginsStatus map[string]*plugins.Status, pluginNames ..
...
+	if err := json.Unmarshal(rawConfig, &cfg); err != nil || cfg.Output == nil {
+		return 0
+	}
+	// Output may be a single object or an array.
+	var outputs []json.RawMessage
+	if err := json.Unmarshal(cfg.Output, &outputs); err != nil {
+		// Not an array — try as a single object.
+		outputs = []json.RawMessage{cfg.Output}
+	}
+	type outputEntry struct {
```
szuecs: type definitions please outside of functions

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -819,6 +835,43 @@ func allPluginsReady(allPluginsStatus map[string]*plugins.Status, pluginNames ..
...
 }
 
+// decisionLogS3OutputTimeout parses the eopa_dl decision_logs config and returns
+// the timeout from the first S3 output entry, or 0 if not present or unparseable.
+func decisionLogS3OutputTimeout(rawConfig json.RawMessage) time.Duration {
+	if rawConfig == nil {
+		return 0
+	}
+	var cfg struct {
+		Output json.RawMessage `json:"output"`
```
szuecs: this is also a type definition, please outside of the function

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -853,23 +859,17 @@ func decisionLogS3OutputTimeout(rawConfig json.RawMessage) time.Duration {
...
+		var entry decisionLogOutputEntry
+		if err := json.Unmarshal(raw, &entry); err != nil || entry.Timeout == "" {
 			continue
 		}
-		if entry.Type == "s3" && entry.Timeout != "" {
-			d, err := time.ParseDuration(entry.Timeout)
-			if err == nil {
-				return d
-			}
+		if d, err := time.ParseDuration(entry.Timeout); err == nil && d > max {
```
szuecs: I think you do not need the `nil` check, because `var max time.Duration` set `max=0` and on error `d=0`, so `d > max` is never true on error.
It's some kind of API contract that nil values do not change, so it's safe to drop the nil check in Go in this kind of case.

szuecs:
I think other then my last comment, it's good to go.

szuecs:
The current pipeline error will likely be fixed by https://github.com/zalando/skipper/pull/4036 , so after merge of the linked one please rebase on master and force push.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -1036,12 +1090,23 @@ func (opa *OpenPolicyAgentInstance) Close(ctx context.Context) {
 // runDecisionLogger drains decisionLogChan in the background so that logDecision
 // is never called on the request goroutine. The goroutine exits when the channel
 // is closed (during Close) and signals completion via decisionsProcessed.
+// A per-task timeout bounds stuck calls (e.g. a Benthos stream.Consume() call on a
```
mjungsbluth, line 1093: Does that effectively mean losing the decisions?

mjungsbluth:
👍

filters/openpolicyagent/openpolicyagent.go
wisinghe: Yes.. from what I understand the plugin Reconfigure() can cause losing the decisions which were in the memory buffer of the eopa_dl plugin.

wisinghe:
👍

PR #4036, opened 2026-05-28
fix: wait for cipher refresher goroutine to exit on Close

larry-dalmeida:
# Related Issue

Fixes a goroutine leak found while working on #4034.

## Problem

`Registry.Close()` called `v.Close()` on each encrypter but returned immediately, leaving the `runCipherRefresher` goroutine running until it observed the closed channel on its next select iteration. Under `noleak` this caused a test failure in the `secrets` package.

## Fix

Initialize `closedHook` inside `runCipherRefresher` (only when a goroutine is actually started) and drain it in `Registry.Close()` so shutdown blocks until the goroutine has exited. The `nil` guard on the drain ensures encrypters created without a refresh interval (no goroutine started) are unaffected.

No behavior change in production - `Close()` is only called at shutdown.

### DIFF
--- a/secrets/encrypter.go
+++ b/secrets/encrypter.go
@@ -157,16 +157,14 @@ func (e *Encrypter) runCipherRefresher(refreshInterval time.Duration) error {
 	if err != nil {
 		return fmt.Errorf("failed to refresh ciphers: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

a4180p:
👍

PR #4037, opened 2026-05-28
feature: proxy mTLS support

szuecs:
cleanroom implementation based on net.Client and net.CertReloader

Added test cases by ai and validated the code that it is testing as expected.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -251,9 +251,11 @@ type Config struct {
 	EnableAdvancedValidation  bool   `yaml:"enable-advanced-validation"`
 
 	// TLS client certs
-	ClientKeyFile  string            `yaml:"client-tls-key"`
-	ClientCertFile string            `yaml:"client-tls-cert"`
-	Certificates   []tls.Certificate `yaml:"-"`
+	ClientKeyFile             string            `yaml:"client-tls-key"`
+	ClientCertFile            string            `yaml:"client-tls-cert"`
+	ClientCertRefreshInterval time.Duration     `yaml:"client-tls-cert-refresh-interval"`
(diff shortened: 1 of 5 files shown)


a4180p:
👍

szuecs:
:+1:

config/config.go
```diff
@@ -629,6 +631,9 @@ func NewConfig() *Config {
 	// TLS client certs
 	flag.StringVar(&cfg.ClientKeyFile, "client-tls-key", "", "TLS Key file for backend connections, multiple keys may be given comma separated - the order must match the certs")
 	flag.StringVar(&cfg.ClientCertFile, "client-tls-cert", "", "TLS certificate files for backend connections, multiple keys may be given comma separated - the order must match the keys")
+	flag.DurationVar(&cfg.ClientCertRefreshInterval, "client-tls-cert-refresh-interval", 0, "How often to reload client TLS certificate and key files for backend connections. Defaults to 5 minutes if certificate files are set.")
```
MustafaSaber, line 634: Why not set the default duration here?

proxy/proxy.go
```diff
@@ -845,6 +864,26 @@ func WithParams(p Params) *Proxy {
...
+		interval := p.ClientCertRefreshInterval
+		if interval == 0 {
+			interval = 5 * time.Minute
+		}
+		var err error
+		cr, err = snet.NewCertReloader(p.ClientCertFile, p.ClientKeyFile, interval, log)
+		if err != nil {
+			log.Errorf("Failed to initialize cert reloader in proxy: %v", err)
+			os.Exit(2)
+		} else {
```
MustafaSaber, line 879: I don't think `else` makes sense, since we already exist in the if

PR #4043, opened 2026-06-01
redis/valkey: get endpoints for any namespace/name

MustafaSaber:
This will allow multiple skipper instance to get redis/valkey endpoints from routeSRV and stop using k8s dataclient


MustafaSaber:
Not needed

PR #4044, opened 2026-06-01
Feature: mtls filters

szuecs:
feature: mtlsSanCIDR(), mtlsSanDNS(), mtlsSanIP(), mtlsSanURI() filters
feature: mtlsIssuerDN() filter
feature: mtlsCN() filter
feature: mtlsAuthn() filter
doc: document mtls authnz filters
doc: mTLS operations guide
test: benchmarks

ref: closing https://github.com/zalando/skipper/issues/3295


docs/reference/filters.md
```diff
@@ -550,6 +550,69 @@ Example:
...
+This authz filter checks the SAN value of the provided certificate. You have
+to use `mtlsAuthn()` to verify validity.
+
+Parameters are one or more:
+
+* DNS hostnames (string)
+* CIDR (string)
+* IPv4 (string)
+* IPv6 (string)
+* URI (string)
```
MustafaSaber, line 657: We don't want to support S/MIME email?

filters/tls/mtls.go
```diff
@@ -0,0 +1,367 @@
...
+					if _, ok := mf.allowedURIs.Load(u.String()); ok {
+						allowed = true
+						auditCertData.WriteString("SAN URI: ")
+						auditCertData.WriteString(u.String())
+					}
+				}
+			}
+
+		case mtlsCN:
+			if _, ok := mf.allowedCN.Load(cert.Issuer.CommonName); ok {
```
shyamz-22: should we rename the filter to mtlsIssuerCN to be explicit given there is also subject commonName?

filters/tls/mtls.go
```diff
@@ -0,0 +1,367 @@
...
+	var auditCertData auditBuffer
+	if mf.enableAuditLog {
+		auditCertData = &strings.Builder{}
+	} else {
+		auditCertData = voidAuditBuffer{}
+	}
+
+	peerCerts := req.TLS.PeerCertificates
+	allowed := false
+	for _, cert := range peerCerts {
```
vlktna: why you verify all certs? `The first element is the leaf certificate`, so the connection should be verified only by the first cert


```
	// PeerCertificates are the parsed certificates sent by the peer, in the
	// order in which they were sent. The first element is the leaf certificate
	// that the connection is verified against.
	//
	// On the client side, it can't be empty. On the server side, it can be
	// empty if Config.ClientAuth is not RequireAnyClientCert or
	// RequireAndVerifyClientCert.
	//
	// PeerCertificates and its contents should not be modified.
	PeerCertificates []*x509.Certificate
```

filters/tls/mtls.go
```diff
@@ -0,0 +1,367 @@
...
+				break
+			}
+			allowed = true
+			auditCertData.WriteString("authn: ")
+			auditCertData.WriteString(cert.Subject.String())
+		}
+	}
+
+	if !allowed {
+		reject(ctx, http.StatusForbidden, "", mf.typ.rejectReason(), req.Host)
```
shyamz-22, line 434: if peerCerts is empty should we send a 401 instead of 403 given nothing was presented?

filters/tls/mtls.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+	if mf.enableAuditLog {
+		logrus.Infof("access allowed by %s: %s", mf.typ.String(), auditCertData.String())
```
shyamz-22, line 438: In the above line, when access is rejected , it will come here and write access allowed which will not match the decision took on line 360.

filters/tls/mtls.go
```diff
@@ -0,0 +1,367 @@
...
+	return &mtlsSpec{
+		typ: mtlsCN,
+	}
+}
+
+// NewMtlsAuthn returns a filter spec for the mtlsAuthn filter. It
+// takes no argument and loads the system CA certificates used to
+// verify the client certificate chain.
+func NewMtlsAuthn() filters.Spec {
+	pool, _ := x509.SystemCertPool()
```
shyamz-22: the error is ignored here, is it fine?

filters/tls/mtls.go
```diff
@@ -0,0 +1,367 @@
...
+
+	default:
+		if len(args) == 0 {
+			return nil, filters.ErrInvalidFilterParameters
+		}
+	}
+
+	switch ms.typ {
+	case mtlsIssuerDN:
+		for _, arg := range args {
```
MustafaSaber, line 218: Shouldn't we fail if we have more than one DN? Or we support multiples?

filters/tls/mtls.go
```diff
@@ -0,0 +1,367 @@
...
+			s, ok := arg.(string)
+			if !ok || s == "" {
+				return nil, filters.ErrInvalidFilterParameters
+			}
+			mf.allowedDN.Store(s, struct{}{})
+		}
+	case mtlsCN:
+		for _, arg := range args {
```
MustafaSaber, line 228: Same as for DN

docs/reference/filters.md
```diff
@@ -550,6 +550,69 @@ Example:
...
 
+### mtlsAuthn
+
+This filter validates the client certificate provided by verifying
+with the configured system CA certificates.
+
+Example:
+
+```
+* -> mtlsAuthn() -> "http://10.2.5.21:8080";
```
MustafaSaber, line 572: Is this reliable to use alone? If not then maybe we embedded it in all other filters instead of having it as a standalone.

filters/tls/mtls.go
MustafaSaber: Because leaf cert can include any data, I think this is mostly for mtlsAuthn filter which will make sure this is a trusted certificate; if that's the case the loop should be only for mtlsAuthn also, and then other filters decide that it have the right data

docs/reference/filters.md
szuecs: Yes it's fine if you want to check: it's a valid user but you do not care about what exactly it is.

filters/tls/mtls.go
szuecs: why should we fail?
We can support multiple certificates or CAs.
For example allow company X and company Y to call us via mTLS with their certificates, would be a valid filter config, no need to separate routes.

docs/reference/filters.md
szuecs: No I don't think we want that :)

filters/tls/mtls.go
szuecs: oh, no it should be only ignored in tests

docs/reference/filters.md
MustafaSaber: But all other filters require this one, no? I men we should require it some how

filters/tls/mtls.go
MustafaSaber: I agree, just docs doesn't show that 😄

filters/tls/mtls.go
szuecs: I wanted to pass a cert pool from skipper.go and then we can make it configurable and error handling can be done outside of the filter spec creation.

szuecs: fixed by commit a31f45cc5bc4402a1d2f7d8774830d865e3f15d2

filters/tls/mtls.go
szuecs: Maybe we should better switch to Subject.CommonName ....
I think the intention was that the CN of the cert and not the CN of the CA that signed the cert.

szuecs: fixed by 1cb6f476da4dbcbb3b8e6226f51cd9b1fa46c6fe

filters/tls/mtls.go
szuecs: yes makes sense, fixed by 2b37804c2dd0b97e67eae65211f8f12ee6899ecf

filters/tls/mtls.go
szuecs: yes makes sense, fixed by 5b1b092e06cba3f26884895dd79feafe4ea33106

filters/tls/mtls.go
szuecs: yes makes sense, it was fixed within 2b37804c2dd0b97e67eae65211f8f12ee6899ecf

filters/tls/mtls.go
szuecs: Ok maybe I should explain.
The DN issuer and CN subject are only one in each cert, so if we allow multiple values in the filter it means "one of these are fine", similar to tokeninfoAnyScope.
So if we have 5 services A,..E and they have A,...E as CN of their cert with DN "Applications,Kubernetes,Zalando", then we could allow A and C to call us by `mtlsAuthn() -> mtlsCN("A", "C")`

szuecs:
Given that auth should be fast, I will also provide 4 more filters for `mtlsSAN` to have them special, for example if you use SPIFFE/SPIRE, you do not want to check DNS or IP in SAN.

filters/tls/mtls.go
MustafaSaber: > Because leaf cert can include any data, I think this is mostly for mtlsAuthn filter which will make sure this is a trusted certificate; if that's the case the loop should be only for mtlsAuthn also, and then other filters decide that it have the right data

It's already done by https://pkg.go.dev/crypto/x509#Certificate.Verify

szuecs:
e634e6e93768541f65321369dd3f60255f382c41 has the 4 new filters

szuecs:
Local test without CA loaded, so empty pool it can not validate certs:
```
% ./bin/skipper -inline-routes='r: * -> mtlsAuthn() -> status(201) -> <shunt>' -address :9002                                                                     feature/mTLS-filters
[APP]INFO[0000] Expose metrics in format: "codahale"
[APP]INFO[0000] enable swarm: false
[APP]INFO[0000] Replacing tee filter specification
[APP]INFO[0000] Replacing teenf filter specification
[APP]INFO[0000] Replacing teeResponse filter specification
[APP]INFO[0000] route settings, reset, route: r: * -> mtlsAuthn() -> status(201) -> <shunt>
[APP]INFO[0000] route settings received, id: 1
[APP]INFO[0000] route settings applied, id: 1
[APP]INFO[0000] Dataclients are updated once, first load complete
[APP]INFO[0000] Listen on :9002
::1 - - [03/Jun/2026:22:10:26 +0200] "GET /foo HTTP/1.1" 401 0 "-" "curl/7.49.0" 0 localhost:9002 - -
```

client call sent with no cert:
```
% curl http://localhost:9002/foo -v
*   Trying ::1...
* Connected to localhost (::1) port 9002 (#0)
> GET /foo HTTP/1.1
> Host: localhost:9002
> User-Agent: curl/7.49.0
> Accept: */*
>
< HTTP/1.1 401 Unauthorized
< Server: Skipper
< Www-Authenticate: localhost:9002
< Date: Wed, 03 Jun 2026 20:10:26 GMT
< Transfer-Encoding: chunked
<
```

We can see good default behaviour and no panic, so integration seems also fine

filters/tls/mtls.go
szuecs: The docs say "varargs"

docs/reference/filters.md
szuecs: It's a split authn/authz filters, very powerful in composite and simple in implementation.
It's similar to other authn/authz filters like: 
```
oauthGrant() -> grantClaimsQuery("/path:@_:sub%\"userid\"")
```

szuecs: grantClaimsQuery also does not authn

filters/tls/mtls.go
```diff
@@ -0,0 +1,511 @@
...
+	case mtlsSAN:
+		var ipStrs []string
+		for _, arg := range args {
+			s, ok := arg.(string)
+			if !ok {
+				return nil, filters.ErrInvalidFilterParameters
+			}
+			if !isValidSAN(s) {
+				return nil, filters.ErrInvalidFilterParameters
+			}
```
a4180p: all checks from this function (with exception of the isValidHostname) are performed below

filters/tls/mtls.go
```diff
@@ -0,0 +1,511 @@
...
+			// IP addresses and CIDRs are collected for the IPSet.
+			if _, err := netip.ParsePrefix(s); err == nil {
+				ipStrs = append(ipStrs, s)
+			} else if _, err := netip.ParseAddr(s); err == nil {
+				ipStrs = append(ipStrs, s)
+			} else if u, err := url.Parse(s); err == nil && u.Scheme != "" {
+				mf.allowedURIs.Store(s, struct{}{})
+			} else {
+				mf.allowedHostnames.Store(strings.ToLower(s), struct{}{})
+			}
```
a4180p:
```suggestion
			} else if isValidHostname(s) {
				mf.allowedHostnames.Store(strings.ToLower(s), struct{}{})
			} else {
				return nil, filters.ErrInvalidFilterParameters
			}
```

filters/tls/mtls.go
a4180p:
```suggestion
```

config/config.go
```diff
@@ -861,6 +867,24 @@ func (c *Config) ParseArgs(progname string, args []string) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to load system cert pool: %v", err)
+		} else {
+			c.MtlsAuthnCA = pool
+		}
+	} else {
+		c.MtlsAuthnCA = x509.NewCertPool()
+	}
+	if c.MtlsAuthnCaFile != "" {
+		pem, err := os.ReadFile(c.MtlsAuthnCaFile)
```
a4180p: docstring says `multiple files may be given comma separated`, but we read a sinle one

config/config.go
```diff
@@ -861,6 +867,24 @@ func (c *Config) ParseArgs(progname string, args []string) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to read %q: %v", c.MtlsAuthnCaFile, err)
+		}
+		c.MtlsAuthnCA.AppendCertsFromPEM(pem)
```
a4180p: return value is ignored, I think we should log if it was unsuccessfull

filters/tls/mtls.go
szuecs: please no suggestions, the feature is broken with DCO

filters/tls/mtls.go
```diff
@@ -0,0 +1,511 @@
...
+		for _, arg := range args {
+			s, ok := arg.(string)
+			if !ok || s == "" {
+				return nil, filters.ErrInvalidFilterParameters
+			}
+			mf.allowedDN.Store(s, struct{}{})
+		}
+	case mtlsCN:
+		for _, arg := range args {
+			if s, ok := arg.(string); !ok {
```
a4180p, line 229: is an empty string allowed?

filters/tls/mtls.go
```diff
@@ -0,0 +1,511 @@
...
+	allowedCN sync.Map // string -> struct{}
+
+	// mtlsIssuerDN allow-list (RFC 2253 DN strings)
+	allowedDN sync.Map // string -> struct{}
+
+	// mtlsSAN allow-lists: hostnames, URIs, and IP/CIDR ranges are stored
+	// separately so the hot path can use a single IPSet.Contains call for IPs
+	// instead of re-parsing every pattern on each request.
+	allowedHostnames sync.Map // string -> struct{} (lowercased)
+	allowedURIs      sync.Map // string -> struct{} (exact match)
```
a4180p: why do we use sync.Map here?
they are not shared between filter instances and writes are performed only in CreateFilter.

filters/tls/mtls.go
a4180p: > please no suggestions, the feature is broken with DCO

if you don't commit them with button everything is good :)

docs/reference/filters.md
```diff
@@ -550,6 +550,131 @@ Example:
...
+to use `mtlsAuthn()` to verify validity.
+
+Parameters are one or more:
+
+* URI (string)
+
+Example:
+
+```
+* -> mtlsAuthn() ->  mtlsSAN("spiffe://my-service.example/app1") -> "http://10.2.5.21:8080";
```
MustafaSaber: wrong filter example

docs/reference/filters.md
```diff
@@ -550,6 +550,131 @@ Example:
...
+
+* CN (string)
+
+Example:
+
+```
+* -> mtlsAuthn() ->  mtlsCN("My CA") -> "http://10.2.5.21:8080";
+```
+
+### mtlsSAN
```
MustafaSaber: I guess this should be removed and use other filters

filters/tls/mtls.go
MustafaSaber: currently they are shared between mtlsSan & the smaller filters mtlsSan*

a4180p: @MustafaSaber  no, they are not. 
Create filter returns new datastructure every time

MustafaSaber: yeah you are right

filters/tls/mtls.go
szuecs: fixed

config/config.go
szuecs: Yeah good point I added a for loop to read them all

config/config.go
szuecs: I return an error

filters/tls/mtls.go
szuecs: Yeah in some cases there are empty SAN, not sure if someone wants to check for it but we also do not need to decide this.

filters/tls/mtls.go
szuecs: Hmm yeah good catch

docs/reference/filters.md
szuecs: It depends, if someone wants to allow a mix of it, but yeah we could also argue to use a filter chain instead.
I agree will delete it later.

filters/tls/mtls.go
szuecs: fixed

szuecs:
@MustafaSaber I can't reply to your message directly so doing it here:
it's fixed

docs/reference/filters.md
szuecs: done

docs/reference/filters.md
szuecs: cleaned up

szuecs:
@MustafaSaber fixed your comments

a4180p:
👍

MustafaSaber (approved):

filters/tls/mtls.go
```diff
@@ -0,0 +1,428 @@
...
+					auditCertData.WriteString("SAN IP: ")
+					auditCertData.WriteString(ip.String())
+				}
+			}
+		}
+
+	case mtlsSanDNS:
+		// Check hostname SANs against the allowlist.
+		for _, dns := range cert.DNSNames {
+			if _, ok := mf.allowedHostnames[strings.ToLower(dns)]; ok {
```
shyamz-22, line 392: in the allowedHostnames, if I have *.example.com, and the certificate has foo.example.com as SAN, in this case it won't match. Should we state explicitly wildcard matching is not allowed?

filters/tls/mtls.go
```diff
@@ -0,0 +1,428 @@
...
+			if !ok || s == "" {
+				return nil, filters.ErrInvalidFilterParameters
+			}
+			mf.allowedDN[s] = struct{}{}
+		}
+
+	case mtlsCN:
+		mf.allowedCN = make(map[string]struct{})
+		for _, arg := range args {
+			if s, ok := arg.(string); !ok {
```
shyamz-22, line 229: should empty common name string be checked as invalid filter parameters?

filters/tls/mtls.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+
+	switch ms.typ {
+	case mtlsAuthn:
+		if len(args) != 0 {
+			return nil, filters.ErrInvalidFilterParameters
+		}
+
+		mf.verifyOpt = x509.VerifyOptions{
+			Roots:     ms.caPool,
```
shyamz-22: This does not have the intermediates set, which means a client cert signed by an intermediate CA will fail unless that intermediate is also in Roots. The test also does not cover this case.

filters/tls/mtls.go
```diff
@@ -0,0 +1,428 @@
...
+	)
+
+	rsp := &http.Response{
+		StatusCode: status,
+		Header:     make(map[string][]string),
+	}
+
+	if hostname != "" {
+		// https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.2
+		rsp.Header.Add("WWW-Authenticate", hostname)
```
shyamz-22, line 338: Is this a valid scheme or challenge for mTLS? Also can req.Host be used without validating/sanitizing? How does a client understand this response?

filters/tls/mtls.go
szuecs: No, because empty string is allowed in CN as far as I know.

filters/tls/mtls.go
szuecs: Good point maybe it's a feature request. I can create a follow up issue for this.

filters/tls/mtls.go
szuecs: 1) You can pass whatever you want to the filter spec creation. If you add intermediates it's up to you, or do I miss something?
2) I am not sure if we should add intermediates explicitly. An operator can add intermediates as CA, too. Do you have an idea if this is needed? What use case would this solve?

filters/tls/mtls.go
szuecs: In case of https it seems to fit. Basically we are telling what hostname did not authenticate. 
In the old days with basic auth and vhosts in Apache it was also helpful to understand that a different route caught this request and you were having authnz problem caused by routing problems.
Because a header is not automatically shown in an HTML page, we do not risk XSS or something like that, so I would say it's fine to use req.Host.

filters/tls/mtls.go
shyamz-22: just wanted to make sure, we allow mtlsCN("") as valid filter correct?

szuecs: added a test case, which shows it's fine

filters/tls/mtls.go
```diff
@@ -0,0 +1,445 @@
...
+		typ: mtlsCN,
+	}
+}
+
+// NewMtlsAuthn returns a filter spec for the mtlsAuthn filter. You
+// need to pass the x509.CertPool that it will use to validate client
+// certificates for authentication. The second parameter can be nil,
+// in case you have no intermediate or can not load intermediates
+// upfront.
+func NewMtlsAuthn(caPool, intermediates *x509.CertPool) filters.Spec {
```
shyamz-22, line 116: I think I was not clear  in my previous comments. Loading intermediates in NewMtlsAuthn is not necessary  Operator configures trusted roots. Client sends leaf + intermediates. Verifier builds chain using both. So the filter already has access to intermediates at request time. which you are doing in Line 421 - 428. Unless we want to use it as a fallback. What I was suggesting is what you have now fixed in 421-429.

filters/tls/mtls.go
```diff
@@ -0,0 +1,445 @@
...
+		if _, ok := mf.allowedCN[leafCert.Subject.CommonName]; ok {
+			allowed = true
+			auditCertData.WriteString("CN: ")
+			auditCertData.WriteString(leafCert.Subject.CommonName)
+		}
+
+	case mtlsAuthn:
+		verifyOpt := mf.verifyOpt
+		// peerCerts -> chain: [Leaf, Intermediate]
+		if len(peerCerts) > 1 && verifyOpt.Intermediates == nil {
```
shyamz-22, line 419: If we remove the loading of intermediates we can remove this `verifyOpt.Intermediates == nil` check. We populate intermediates if peerCerts > 1

filters/tls/mtls.go
```diff
@@ -0,0 +1,445 @@
...
+			intermediates := x509.NewCertPool()
+			for _, cert := range peerCerts[1:] {
+				intermediates.AddCert(cert)
+			}
+			verifyOpt.Intermediates = intermediates
+			verifyOpt.MaxConstraintComparisions = defaultMaxIntermediateChain
```
shyamz-22: If the intention is to limit intermediates to 5, can this be done with Certificate constraints. Usually how I understand the constraints is in CA certificate it will say allowed dns domain is cluster1.zalan.do, and if leaf certificate SAN has cluster2.zalan.do then verification will fail, in this case if the SAN list is more than 5 it will fail for 6th or combination of multiple constraints like [ExcludedDNSDomains, ExcludedIPRanges](https://github.com/golang/go/blob/beed654b97d1fe530df2e7de764079ba4dd5dcc2/src/crypto/x509/x509.go#L860-L868). I am not sure about the naming of the constants and purpose of the option here.

skipper.go
```diff
@@ -2297,6 +2314,15 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 		o.CustomFilters = append(o.CustomFilters, lua)
 	}
 
+	if o.MtlsAuthnCA == nil {
```
shyamz-22, line 2317: will this ever be nil, in config.go  L870 -L879, if  --mtls-authn-append-ca is not set then MtlsAuthnCA is set to non empty cert pool, and so if neither flag is set --mtls-authn-ca  or --mtls-authn-append-ca mtlsAuthn() does not use system roots and rejects otherwise valid client certs. I think we should make clear either one of these must be set for mTLSAuthn() to work successfully. May be you can correct if I understand the code wrong. But its a bit confusing
```
	if c.MtlsAuthnAppendCA {
		pool, err := x509.SystemCertPool()
		....
	} else {
		c.MtlsAuthnCA = x509.NewCertPool()
	}
```

filters/tls/mtls.go
shyamz-22: I addressed it [here](https://github.com/zalando/skipper/pull/4044#discussion_r3429242143), I made the comment in the wrong place. Hope this clarifies my intent better now.

skipper.go
szuecs: config.go is only used in case of skipper as binary, library users have to do it themselves.
So yes default for library users is "nil".

By default it loads system cert pool and if you present a client cert that is valid from your system cert pool you can perfectly use the filters without an issue and no global configuration needed. Less config the better.

I think it's the simplest way and the most "expected" for operations.

filters/tls/mtls.go
szuecs: It's only a maximum per connection such that we have no attack vector that a client passes arbitrary long list of chains to us.
This here configures the verify options such that a client can pass an intermediate chain up to 5. Other clients can present a completely different set of intermediates. So you can have 10000 different intermediates for 10000 different connections to skipper, no problem.

filters/tls/mtls.go
szuecs: Yes, but if we pass intermediates, we can skip loading client presented intermediates. I only allow passing these fixed intermediates as "library" user right now, but maybe we want to also create a config for this.
A tiny `verifyOpt.Intermediates == nil` should not be the issue.

filters/tls/mtls.go
szuecs: https://github.com/zalando/skipper/issues/4073

filters/tls/mtls.go
szuecs: Yes, but you could also pass fixed intermediates and it would be faster to not create the verifyOptions on every request as we do now in lines 421-428

skipper.go
shyamz-22: I am confused by "By default it loads system cert pool ", but we do  `c.MtlsAuthnCA = x509.NewCertPool()` , which is not the system cert pool but a empty pool or am I missing something?

filters/tls/mtls.go
shyamz-22: got it.

filters/tls/mtls.go
shyamz-22: got it.

filters/tls/mtls.go
shyamz-22: MaxConstraintComparisons are for NameConstraints and not checking the intermediate chain length right? In that case shouldn't it check the peerCert length? [Name Constraint is an Extension and is for CA issued certificates](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.10)

```
	// MaxConstraintComparisions is the maximum number of comparisons to
	// perform when checking a given certificate's name constraints. If
	// zero, a sensible default is used. This limit prevents pathological
	// certificates from consuming excessive amounts of CPU time when
	// validating. It does not apply to the platform verifier.
	MaxConstraintComparisions int
```

filters/tls/mtls.go
szuecs: see benchmark results in the commit message

filters/tls/mtls.go
szuecs: I dropped the config as Go authors wrote that they use "sensible default if set to 0"

shyamz-22:
👍🏽

skipper.go
szuecs: Clarified in a call.
We use `x509.SystemCertPool()` by default to initialize, if there is no user chosen config.

szuecs:
:+1:

PR #4056, opened 2026-06-05
feature: routesrv allow arbitrary filter chains to secure call path

szuecs:
feature: routesrv allow arbitrary filter chains to secure the call path to routesrv as you want

### DIFF
--- a/routesrv/filter.go
+++ b/routesrv/filter.go
@@ -1,0 +1,130 @@
+package routesrv
+
+import (
+	"io"
+	"net/http"
+
+	ot "github.com/opentracing/opentracing-go"
+	log "github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/metrics"
+)
+
+// filterContext is a minimal filters.FilterContext implementation for use as
+// HTTP middleware. It supports the Request() path only: filters can read the
+// request, set state-bag values, and short-circuit via Serve(). Methods that
+// are meaningful only inside the proxy's full routing engine (Split, Loopback,
+// backend access, etc.) are not implemented and will panic if called.
+type filterContext struct {
+	w        http.ResponseWriter
+	req      *http.Request
+	rsp      *http.Response
+	metrics  metrics.Metrics
+	served   bool
+	stateBag map[string]interface{}
+}
+
+func (fc *filterContext) ResponseWriter() http.ResponseWriter   { return fc.w }
+func (fc *filterContext) Request() *http.Request                { return fc.req }
+func (fc *filterContext) Response() *http.Response              { return fc.rsp }
+func (fc *filterContext) OriginalRequest() *http.Request        { return fc.req }
+func (fc *filterContext) OriginalResponse() *http.Response      { return fc.rsp }
+func (fc *filterContext) Served() bool                          { return fc.served }
+func (fc *filterContext) MarkServed()                           { fc.served = true }
+func (fc *filterContext) StateBag() map[string]interface{}      { return fc.stateBag }
+func (fc *filterContext) BackendUrl() string                    { return "" }
+func (fc *filterContext) OutgoingHost() string                  { return "" }
+func (fc *filterContext) SetOutgoingHost(string)                {}
+func (fc *filterContext) PathParam(string) string               { return "" }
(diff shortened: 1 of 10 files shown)


routesrv/filter.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	for k, vs := range rsp.Header {
+		for _, v := range vs {
+			fc.w.Header().Add(k, v)
+		}
+	}
+	status := rsp.StatusCode
+	if status == 0 {
+		status = http.StatusOK
+	}
```
madumalt:
```suggestion
	if rsp.StatusCode == 0 {
		rsp.StatusCode = http.StatusOK
	}
```

madumalt: IIUC you want to set the response statusCode to OK if it is empty right?

routesrv/routesrv.go
```diff
@@ -282,10 +357,18 @@ func run(rs *RouteServer, opts skipper.Options, sigs chan os.Signal) error {
 	rs.StartUpdates()
 
 	go rs.startSupportListener()
-	if err = rs.server.ListenAndServe(); err != http.ErrServerClosed {
-		go shutdown(0)
+	if tlsConfig := rs.server.TLSConfig; tlsConfig != nil {
+		if err = rs.server.ListenAndServeTLS("", ""); err != http.ErrServerClosed {
+			go shutdown(0)
+		} else {
+			err = nil
```
madumalt, line 364: IIUC we don't care about the error here unless it is a http.ErrServerClosed. Even if it is the case let's add a info log here saying that there was an error, since it was not ErrServerClosed, we are ignoring it. WDYT?

routesrv/routesrv.go
```diff
@@ -282,10 +357,18 @@ func run(rs *RouteServer, opts skipper.Options, sigs chan os.Signal) error {
...
+		}
 	} else {
-		err = nil
+		if err = rs.server.ListenAndServe(); err != http.ErrServerClosed {
+			go shutdown(0)
+		} else {
+			err = nil
```
madumalt, line 370: same as https://github.com/zalando/skipper/pull/4056/changes#r3458326129

routesrv/routesrv.go
szuecs: Hmm, behaviour did not change and I don't know for sure why we ignored the error before.
I bet because of the `rs.wg.Wait()`, but maybe it would be better to check `err` and if non nil `return err` else we would wait on the waitgroup.

I will change the behaviour.

szuecs: Ah no I understand the behaviour now.
If the error is not `http.ErrServerClosed` we init the shutdown procedure, because the other means that it is already closed and we have nothing to do.
If one of the 2 Listeners is already closed we don't care and that's why we set `err=nil`.

So in total the code is good

madumalt: alright then, shall we add your explanation as a comment there? It will help in the future when someone looking at the code.

madumalt (approved):

szuecs:
retrigger

szuecs:
:+1:

routesrv/routesrv.go
```diff
@@ -169,6 +198,52 @@ func New(opts skipper.Options) (*RouteServer, error) {
...
+	return filterChain, nil
+}
+
+func addCustomerFilters(opts *skipper.Options) {
+	if opts.MtlsAuthnCA != nil {
+		opts.CustomFilters = append(opts.CustomFilters, tlsfilters.NewMtlsAuthn(opts.MtlsAuthnCA, opts.MtlsAuthnInterMediateCA))
+	}
+	if opts.OAuthTokeninfoURL != "" {
+		tio := auth.TokeninfoOptions{
+			URL:                         opts.OAuthTokeninfoURL,
```
MustafaSaber, line 230: I don't think this will work, routeSRV uses regular skipper image

MustafaSaber: I mean the fix won't be in this PR, but we need to do that change first in kubernetes-on-aws, change the image and the healthchecks and readinessProbe

szuecs: yeah of course we need to change our deployment.
However the health check is not part of the filter chain, because kubelet executes healtcheck and has no access to $corp tokens or client TLS certs to access health checks.

PR #4067, opened 2026-06-14
fix: host metrics should be measured and stored correctly in case of HostAny() predicate

szuecs:
fix: host metrics should be measured and stored correctly in case of HostAny() predicate

Before this fix we measured a route like the following in the bucket `"_unknownhost_"`:

```
r: HostAny("www.example.com") -> <shunt>
```

### DIFF
--- a/metrics/metricstest/metricsmock.go
+++ b/metrics/metricstest/metricsmock.go
@@ -5,6 +5,7 @@ package metricstest
 import (
 	"fmt"
 	"net/http"
+	"strconv"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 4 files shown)


szuecs:
:+1:

MustafaSaber:
Tests still fail in the pipeline

szuecs:
> Tests still fail in the pipeline

not the GH Actions, so maybe a bug in CdP?

szuecs:
I can reproduce it locally, so will fix it

szuecs:
:+1:

madumalt (approved):

PR #4078, opened 2026-06-19
PR test suite: split into long running packages and rest

szuecs:
PR test suite: split into long running packages and rest

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -29,7 +29,7 @@ jobs:
           check-latest: true
       - run: go version
       - run: make deps
-      - run: make check-race_split1
+      - run: make check-race_rest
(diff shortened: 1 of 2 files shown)


a4180p:
👍

.github/workflows/pr.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
       # Run the rule checker using the fetched rules
       - run: semgrep ci -f semgrep-go
 
-  check-race:
+  check-race_rest:
```
greeshma1196, line 21: This can be done in another PR; I think we should be a little bit more intentional with the naming other than `rest`, but I am not able to come up with a good one now.  😅

greeshma1196:
👍🏽

PR #4087, opened 2026-06-27
routing: measure data client LoadAll/LoadUpdate duration

### LINKED ISSUE #4086: Expose per-DataClient load latency metrics (LoadAll, LoadUpdate)

Skipper currently exposes route update latency, but it does not expose per-DataClient execution duration for LoadAll and LoadUpdate calls.
Without this visibility, it is difficult to identify which datasource is slow when route updates lag. This can lead to stale routing behavior, such as continuing to route to outdated backends or delayed pickup of newly added routes.

It would be useful to have built-in Prometheus duration metrics for DataClient operations across all DataClients, including the Kubernetes datasource implemented in kube.go.

A possible shape:

* `skipper_custom_duration_seconds{key="data_clients.load_latency"}` histogram metric
* Labels:
    * client (for example kubernetes, etcd, eskipfile, inline)
    * operation (load_all, load_update)
    * result (success, error)

This would provide out-of-the-box observability for datasource performance without requiring custom wrappers.

Custom wrapping of each DataClient is possible, but it is inconsistent, easy to miss for built-in clients, and less useful than a standardized core metric.

There is a related metric for route update latency at routing.go:427.
That metric only measures the duration after the DataClient has returned its routing table. It excludes the DataClient load duration.
The current request is specifically for per-DataClient timing, not to replace that metric. A global metric that encapsulate both of those operations (loading + updating) would also be useful, but might prove harder to achieve with the current code structure.

No

AshrafAhmed9:
Fixes #4086

Adds per-data-client timing of LoadAll and LoadUpdate in receiveFromClient
using the existing Metrics.MeasureSince. Keyed by client name and operation:
- `routes.<client>.load_all`
- `routes.<client>.load_update`

The client name comes from NamedDataClient.Name() where available, falling
back to the concrete type otherwise.

As agreed with @fvigneault in #4086, this encodes the name and
operation in the metric key (rather than a new Prometheus histogram with
explicit labels) and drops the result/error dimension.

### DIFF
--- a/routing/datasource.go
+++ b/routing/datasource.go
@@ -59,6 +59,16 @@ func (d *incomingData) log(l logging.Logger, suppress bool) {
 	}
 }
 
+// dataClientName returns a name for the data client suitable for use in metric
+// keys, using NamedDataClient.Name() when available.
+func dataClientName(c DataClient) string {
+	name := "unknown"
+	if nc, ok := c.(NamedDataClient); ok {
+		name = nc.Name()
+	}
+	return name
+}
+
 // continuously receives route definitions from a data client on the output channel.
 // The function does not return unless quit is closed. When started, it request for the
 // whole current set of routes, and continues polling for the subsequent updates. When a
@@ -67,6 +77,7 @@ func (d *incomingData) log(l logging.Logger, suppress bool) {
 // nondeterministic way, but this may change in the future.
 func receiveFromClient(c DataClient, o Options, out chan<- *incomingData, quit <-chan struct{}) {
 	initial := true
+	clientName := dataClientName(c)
 	var ticker *time.Ticker
 	if o.PollTimeout != 0 {
 		ticker = time.NewTicker(o.PollTimeout)
@@ -74,6 +85,8 @@ func receiveFromClient(c DataClient, o Options, out chan<- *incomingData, quit <
 		ticker = time.NewTicker(time.Millisecond)
 	}
 	defer ticker.Stop()
+	now := time.Now()
+
 	for {
 		var (
 			routes     []*eskip.Route
@@ -83,8 +96,10 @@ func receiveFromClient(c DataClient, o Options, out chan<- *incomingData, quit <
 
 		if initial {
 			routes, err = c.LoadAll()
+			o.Metrics.MeasureSince("routes.load_all."+clientName, now)
 		} else {
 			routes, deletedIDs, err = c.LoadUpdate()
+			o.Metrics.MeasureSince("routes.load_update."+clientName, now)
 		}
 
 		switch {
@@ -93,6 +108,7 @@ func receiveFromClient(c DataClient, o Options, out chan<- *incomingData, quit <
 		case err != nil:
 			o.Log.Error("error while receiving update;", err)
 			initial = true
+			now = time.Now()
 			continue
 		case initial || len(routes) > 0 || len(deletedIDs) > 0:
 			var incoming *incomingData
@@ -111,7 +127,8 @@ func receiveFromClient(c DataClient, o Options, out chan<- *incomingData, quit <
 		}
 
 		select {
-		case <-ticker.C:
+		case t := <-ticker.C:
+			now = t
 		case <-quit:
 			return
 		}
--- a/routing/datasource_test.go
+++ b/routing/datasource_test.go
@@ -774,3 +774,57 @@ func (s slowCreateSpec) CreateFilter(args []interface{}) (filters.Filter, error)
 
 func (s slowCreateFilter) Request(ctx filters.FilterContext) {}
 func (s slowCreateFilter) Response(filters.FilterContext)    {}
+
+func TestDataClientLoadMetrics(t *testing.T) {
+	m := &metricstest.MockMetrics{Now: time.Now()}
+
+	dc, err := testdataclient.NewDoc(`r0: * -> <shunt>`)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	r := routing.New(routing.Options{
+		DataClients:     []routing.DataClient{dc},
+		Metrics:         m,
+		SignalFirstLoad: true,
+	})
+	defer func() {
+		r.Close()
+		dc.Close()
+	}()
+
+	<-r.FirstLoad()
+
+	// LoadAll ran during the first load.
+	waitForMeasure(t, m, "routes.load_all.unknown")
+
+	// Trigger an update to exercise the LoadUpdate path.
+	if err := dc.UpdateDoc(`r1: * -> <shunt>`, nil); err != nil {
+		t.Fatal(err)
+	}
+
+	waitForMeasure(t, m, "routes.load_update.unknown")
+}
+
+func waitForMeasure(t *testing.T, m *metricstest.MockMetrics, key string) {
+	t.Helper()
+	timeout := time.After(time.Second)
+	ticker := time.NewTicker(time.Millisecond)
+	defer ticker.Stop()
+
+	for {
+		found := false
+		m.WithMeasures(func(measures map[string][]time.Duration) {
+			_, found = measures[key]
+		})
+		if found {
+			return
+		}
+
+		select {
+		case <-timeout:
+			t.Fatalf("timed out waiting for measure %q", key)
+		case <-ticker.C:
+		}
+	}
+}
--- a/routing/routing.go
+++ b/routing/routing.go
@@ -268,6 +268,10 @@ func New(o Options) *Routing {
 		o.Log = &logging.DefaultLog{}
 	}
 
+	if o.Metrics == nil {
+		o.Metrics = metrics.Void
+	}
+
 	uniqueClients := make(map[DataClient]struct{}, len(o.DataClients))
 	for i, c := range o.DataClients {
 		if _, ok := uniqueClients[c]; ok {


routing/datasource.go
```diff
@@ -81,11 +91,19 @@ func receiveFromClient(c DataClient, o Options, out chan<- *incomingData, quit <
 			err        error
 		)
 
+		start := time.Now()
 		if initial {
 			routes, err = c.LoadAll()
 		} else {
 			routes, deletedIDs, err = c.LoadUpdate()
 		}
+		if o.Metrics != nil {
```
szuecs: I would say that metrics should never be nil here , so we don't need to check

szuecs (dismissed):

szuecs:
Rest looks good to me

AshrafAhmed9:
Thanks @szuecs! Made metrics default to metrics.Void in New() (matching how Log is defaulted) so it's never nil, and removed the check. Signed off the commit for DCO too.

routing/datasource_test.go
```diff
@@ -774,3 +774,57 @@ func (s slowCreateSpec) CreateFilter(args []interface{}) (filters.Filter, error)
...
+
+	// LoadAll ran during the first load.
+	waitForMeasure(t, m, fmt.Sprintf("routes.%T.load_all", dc))
+
+	// Trigger an update to exercise the LoadUpdate path.
+	if err := dc.UpdateDoc(`r1: * -> <shunt>`, nil); err != nil {
+		t.Fatal(err)
+	}
+
+	waitForMeasure(t, m, fmt.Sprintf("routes.%T.load_update", dc))
```
AlexanderYastrebov: nit: maybe hardcode the constant to see what %T actually produces.

routing/datasource.go
```diff
@@ -59,6 +59,15 @@ func (d *incomingData) log(l logging.Logger, suppress bool) {
 	}
 }
 
+// dataClientName returns a name for the data client suitable for use in metric
+// keys, using NamedDataClient.Name() when available.
+func dataClientName(c DataClient) string {
+	if nc, ok := c.(NamedDataClient); ok {
+		return nc.Name()
+	}
+	return fmt.Sprintf("%T", c)
```
AlexanderYastrebov: What will values looks like? I think there is a constraint charset for metric name that %T might not fit.

routing/datasource.go
```diff
@@ -81,11 +91,17 @@ func receiveFromClient(c DataClient, o Options, out chan<- *incomingData, quit <
...
 		if initial {
 			routes, err = c.LoadAll()
 		} else {
 			routes, deletedIDs, err = c.LoadUpdate()
 		}
+		operation := "load_update"
+		if initial {
+			operation = "load_all"
+		}
+		o.Metrics.MeasureSince("routes."+clientName+"."+operation, start)
```
AlexanderYastrebov: Not sure, is it better to have routes.name.operation or routes.operation.name, i.e. have variable part at the end (for the cases where metric mane is matched via asterisk)

routing/datasource.go
```diff
@@ -59,6 +59,15 @@ func (d *incomingData) log(l logging.Logger, suppress bool) {
 	}
 }
 
+// dataClientName returns a name for the data client suitable for use in metric
+// keys, using NamedDataClient.Name() when available.
+func dataClientName(c DataClient) string {
+	if nc, ok := c.(NamedDataClient); ok {
```
AlexanderYastrebov, line 66: What existing named clients are there? 
How this would look like to k8s, inline routes and url clients (i.e. most common)?

routing/datasource.go
```diff
@@ -81,11 +91,17 @@ func receiveFromClient(c DataClient, o Options, out chan<- *incomingData, quit <
 			err        error
 		)
 
+		start := time.Now()
```
AlexanderYastrebov: nit: its fine like that but there is already a ticker.C that returns current time.

AlexanderYastrebov (commented):
Hi, looks ok, but please check metric name format.

routing/datasource_test.go
AshrafAhmed9: Done — hardcoded the expected names.

routing/datasource.go
AshrafAhmed9: Built-in clients all implement NamedDataClient: k8s → kubernetes, etcd → etcd, eskipfile → eskipfile (plus eskipfile-remote, eskipfile-watch), inline/string routes → inline. So for the common cases the names are clean lowercase. %T only triggers for clients that don't implement Name() — custom/third-party or the test client.

routing/datasource.go
AshrafAhmed9: Right — for named clients the values are clean (list above). %T only applies to unnamed clients and produces e.g. *testdataclient.Client. The key is used as a Prometheus label value (unconstrained) and a CodaHale dotted metric name, so the * would be ugly in CodaHale. Happy to sanitize the fallback or fall back to a fixed "unknown" — which do you prefer?

routing/datasource.go
AshrafAhmed9: Good point — with the client part last, routes.load_all.* matches cleanly in hierarchical backends. I'll switch to `routes.load_all.<client> / routes.load_update.<client>` unless you or @szuecs prefer otherwise.

routing/datasource.go
AshrafAhmed9: ticker.C fires at the poll-interval boundary (it paces the loop), so it wouldn't capture the load call's own duration — kept time.Now() right before the call.

routing/datasource.go
szuecs: @AlexanderYastrebov we had tge need for -ensure-dataclient (if I remember correctly the flag) to have named dataclients

routing/datasource.go
```diff
@@ -81,11 +91,17 @@ func receiveFromClient(c DataClient, o Options, out chan<- *incomingData, quit <
...
 		if initial {
 			routes, err = c.LoadAll()
 		} else {
 			routes, deletedIDs, err = c.LoadUpdate()
 		}
+		operation := "load_update"
+		if initial {
+			operation = "load_all"
```
szuecs: move this into the if block before, no need to have another if block

routing/datasource.go
szuecs: I think @AlexanderYastrebov meant that you could create `now time.Now()`out of the for loop for the first iteration and set it by the ticker driving the loop for the next iteration.
It should work.

AshrafAhmed9: Done — moved now out of the loop and set it from the ticker. Also reset it on the error fast-retry path so the load_all measurement after a failed update stays accurate.

routing/datasource.go
AshrafAhmed9: Done — folded the operation into the if/else and flipped the order to `routes.load_all.<client> `/ `routes.load_update.<client>.`

routing/datasource.go
szuecs: Make Prometheus nice, codahale is planned to be deleted 2026 and was a while deprecated already. Community was asked and nobody uses codahale, so do not focus on codahale!

szuecs:
:+1:

routing/datasource.go
AshrafAhmed9: Thanks — for Prometheus the name is a label value (any string is valid), so it's already fine there even for the fallback. Leaving as is unless you'd prefer it stripped down.

szuecs: All good from my side, no worries. I ask the team to review.

MustafaSaber: I would fallback to unknown, and some time later we can depreciate/remove old dataclient that doesn't  implement Name function


Make NamedDataClient the regular Dataclient.

This will break interface but in the long term it's cleaner and we don't want to maintain to interfaces just because of the extra function and then no need for this function at all, wdyt?

AshrafAhmed9: Good idea — switched the fallback to unknown. On merging NamedDataClient into DataClient: agree it'd be cleaner long-term, but since it's a breaking change touching all client implementations, I'd suggest a separate follow-up rather than expanding this PR. Happy to open an issue to track it — wdyt?

MustafaSaber: Please do 😄!

AshrafAhmed9: Opened #4097 to track this. 🙏

szuecs: > I would fallback to unknown, and some time later we can depreciate/remove old dataclient that doesn't implement Name function
> 
> Make NamedDataClient the regular Dataclient.
> 
> This will break interface but in the long term it's cleaner and we don't want to maintain to interfaces just because of the extra function and then no need for this function at all, wdyt?

No this we do not do @MustafaSaber . We do not break library users and the maintenance of an interface like this is "nothing" compared to the issue we create.

@AshrafAhmed9 please do the following:
You test if the dataclient is a NamedDataClient and if not you set dcName = "unknown", maybe in code it's better to set default to "unknown" and override by `.Name()`

szuecs:
:+1:

routing/datasource.go
szuecs: Current PR is good to go, we do not need a follow up. I closed the issue as "do not want"

AshrafAhmed9: Done — defaulted dcName to "unknown" and override it with .Name(). Thanks, and no problem on dropping the follow-up. 🙏

szuecs:
:+1:

MustafaSaber (approved):

routing/datasource.go
MustafaSaber: We discussed this internally, we don't think the change is that important to break dataclient interface for library users. Maybe if we have more usecases in the future we can reconsider but for now it's not worth it.

AshrafAhmed9:
Thanks for the reviews @szuecs @MustafaSaber @AlexanderYastrebov! 🙏

szuecs:
Thanks for the contribution @AshrafAhmed9

Issue #4098, opened 2026-06-30
feature: decodeBase64 filter


szuecs:
**Is your feature request related to a problem? Please describe.**

I want to rewrite a header value part that is base64 encoded.
The filter should be generic and composable, example decodes the second (index 1), of the Authorization header value:
`decodeBase64("Authorization", 1)`

Example use case:
You have a client side tool that can add a basic auth header for authorization but your server is only able to authnz bearer tokens. In this case you have a base64 encoded string that will be again base64 encoded by the http client.

**Describe the solution you would like**

The solution should be generic and allow arbitrary parts of a header to be decoded. The second argument is optional and if it's not set it decodes the whole value.

`decodeBase64("Authorization", 1)`

PR #4100, opened 2026-06-30
feat: implement decodeBase64 filter for header value decoding

szuecs:
closing https://github.com/zalando/skipper/issues/4098

### DIFF
--- a/filters/builtin/decode_base64.go
+++ b/filters/builtin/decode_base64.go
@@ -1,0 +1,157 @@
+package builtin
+
+import (
+	"encoding/base64"
+	"fmt"
+	"strings"
+
+	"github.com/zalando/skipper/filters"
+)
+
+type decodeBase64Type int
+
+const (
+	decodeRequestHeaderBase64 decodeBase64Type = iota
+	decodeResponseHeaderBase64
+)
+
+// decodeBase64 filter for decoding base64-encoded parts of headers
+type decodeBase64 struct {
+	typ        decodeBase64Type
+	headerName string
+	partIndex  *int // nil means decode entire value, otherwise decode the part at this index
+}
+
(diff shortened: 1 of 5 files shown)


filters/builtin/decode_base64.go
```diff
@@ -0,0 +1,157 @@
...
+	headerName, ok := args[0].(string)
+	if !ok {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	var partIndex *int
+	if len(args) == 2 {
+		// Second parameter is optional and should be a number representing the part index
+		switch v := args[1].(type) {
+		case float64:
```
MustafaSaber, line 73: Why float64?

filters/builtin/decode_base64.go
```diff
@@ -0,0 +1,157 @@
...
+		// Decode the entire value
+		decoded, err := base64.StdEncoding.DecodeString(value)
+		if err != nil {
+			return "", fmt.Errorf("failed to decode base64: %w", err)
+		}
+		return string(decoded), nil
+	}
+
+	// Split by spaces and decode the specified part
+	parts := strings.Fields(value)
```
MustafaSaber, line 102: I wonder how this will work across different headers, this assumes all headers use spaces as separators which is not always the case. Maybe we can add another param for separators & use FieldsFunc instead

filters/builtin/decode_base64.go
szuecs: In skipper and Go it is the default

filters/builtin/decode_base64.go
szuecs: Yeah if we need this we can add another optional parameter to have a separator. The default "space" can stay.

MustafaSaber (approved):

szuecs:
:+1:

PR #4108, opened 2026-07-02
feat: add opt-in support for Prometheus native histograms

### LINKED ISSUE #3898: Optimize prometheus integration
During a session with Frederic from Polarsignals we identified a potential opimization while serving metrics via Prometheus:
- [x] do not use compression to serve metrics
- [ ] check "native" histograms in Prometheus

AshrafAhmed9:
Adds opt-in support for Prometheus native histograms, as discussed in #3898.

- New options: -enable-prometheus-native-histograms and -prometheus-native-histogram-bucket-factor (default 1.1, exported as metrics.DefaultNativeHistogramFactor)
- Dual emission: classic bucketed histograms are kept, so this is fully backward compatible with existing dashboards/alerts; when the flag is off there is no behavior change
- When a native bucket factor is set and `Buckets` is empty, client_golang drops classic buckets instead of applying defaults — the helper falls back to `prometheus.DefBuckets` to preserve them
- `NativeHistogramMaxBucketNumber=160` and `NativeHistogramMinResetDuration=1h`, per the [Prometheus docs](https://prometheus.io/docs/specs/native_histograms/) and OTEL/KEP recommendation, are set to bound memory; happy to make these configurable if preferred
- Native histograms are only transmitted via the protobuf exposition format, so text `/metrics` output is unchanged; docs updated accordingly

Fixes #3898

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -33,47 +33,47 @@ type Config struct {
 	Flags      *flag.FlagSet
 
 	// generic:
-	Address                          string         `yaml:"address"`
-	InsecureAddress                  string         `yaml:"insecure-address"`
-	EnableTCPQueue                   bool           `yaml:"enable-tcp-queue"`
-	ExpectedBytesPerRequest          int            `yaml:"expected-bytes-per-request"`
-	MaxTCPListenerConcurrency        int            `yaml:"max-tcp-listener-concurrency"`
-	MaxTCPListenerQueue              int            `yaml:"max-tcp-listener-queue"`
-	EnableCopyStreamPoolExperimental bool           `yaml:"enable-copy-stream-pool"`
-	IgnoreTrailingSlash              bool           `yaml:"ignore-trailing-slash"`
-	Insecure                         bool           `yaml:"insecure"`
-	ProxyPreserveHost                bool           `yaml:"proxy-preserve-host"`
-	DevMode                          bool           `yaml:"dev-mode"`
-	SupportListener                  string         `yaml:"support-listener"`
-	DebugListener                    string         `yaml:"debug-listener"`
-	CertPathTLS                      string         `yaml:"tls-cert"`
-	KeyPathTLS                       string         `yaml:"tls-key"`
-	StatusChecks                     *listFlag      `yaml:"status-checks"`
-	PrintVersion                     bool           `yaml:"version"`
-	MaxLoopbacks                     int            `yaml:"max-loopbacks"`
-	DefaultHTTPStatus                int            `yaml:"default-http-status"`
-	PluginDir                        string         `yaml:"plugindir"`
-	LoadBalancerHealthCheckInterval  time.Duration  `yaml:"lb-healthcheck-interval"`
-	ReverseSourcePredicate           bool           `yaml:"reverse-source-predicate"`
-	RemoveHopHeaders                 bool           `yaml:"remove-hop-headers"`
-	RfcPatchPath                     bool           `yaml:"rfc-patch-path"`
-	MaxAuditBody                     int            `yaml:"max-audit-body"`
-	MaxMatcherBufferSize             uint64         `yaml:"max-matcher-buffer-size"`
-	EnableBreakers                   bool           `yaml:"enable-breakers"`
-	Breakers                         breakerFlags   `yaml:"breaker"`
-	EnableRatelimiters               bool           `yaml:"enable-ratelimits"`
-	Ratelimits                       ratelimitFlags `yaml:"ratelimits"`
-	EnableRouteFIFOMetrics           bool           `yaml:"enable-route-fifo-metrics"`
-	EnableRouteLIFOMetrics           bool           `yaml:"enable-route-lifo-metrics"`
-	MetricsFlavour                   *listFlag      `yaml:"metrics-flavour"`
-	DisableMetricsCompression        bool           `yaml:"disable-metrics-compression"`
-	EnableNativeHistograms           bool           `yaml:"enable-native-histograms"`
-	NativeHistogramBucketFactor      float64        `yaml:"native-histogram-bucket-factor"`
-	FilterPlugins                    *pluginFlag    `yaml:"filter-plugin"`
-	PredicatePlugins                 *pluginFlag    `yaml:"predicate-plugin"`
-	DataclientPlugins                *pluginFlag    `yaml:"dataclient-plugin"`
-	MultiPlugins                     *pluginFlag    `yaml:"multi-plugin"`
-	CompressEncodings                *listFlag      `yaml:"compress-encodings"`
+	Address                               string         `yaml:"address"`
+	InsecureAddress                       string         `yaml:"insecure-address"`
+	EnableTCPQueue                        bool           `yaml:"enable-tcp-queue"`
+	ExpectedBytesPerRequest               int            `yaml:"expected-bytes-per-request"`
+	MaxTCPListenerConcurrency             int            `yaml:"max-tcp-listener-concurrency"`
+	MaxTCPListenerQueue                   int            `yaml:"max-tcp-listener-queue"`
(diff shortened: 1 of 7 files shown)


metrics/prometheus.go
```diff
@@ -340,6 +343,22 @@ func (p *Prometheus) sinceS(start time.Time) float64 {
...
+// configured, keeping the classic buckets for backward compatibility.
+func (p *Prometheus) histogramOpts(o prometheus.HistogramOpts) prometheus.HistogramOpts {
+	if p.opts.EnableNativeHistograms {
+		// with a native histogram bucket factor set, empty buckets would
+		// disable the classic histogram instead of using the defaults
+		if len(o.Buckets) == 0 {
+			o.Buckets = prometheus.DefBuckets
+		}
+		o.NativeHistogramBucketFactor = p.opts.NativeHistogramBucketFactor
+		o.NativeHistogramMaxBucketNumber = 100
```
AlexanderYastrebov: In this KEP https://www.kubernetes.dev/resources/keps/5808/ they propose 1.1 for the factor and 160 for the max bucket number based on OTEL [recommendation](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#base2-exponential-bucket-histogram-aggregation).

AshrafAhmed9: Good point, thanks! Updated to NativeHistogramMaxBucketNumber = 160 per the KEP/OTEL recommendation — the bucket factor already defaults to 1.1 as proposed there.

AlexanderYastrebov: Prometheus docs also mention 160 for the NativeHistogramMaxBucketNumber  and 1h for the NativeHistogramMinResetDuration [here](https://prometheus.io/docs/specs/native_histograms/#limiting-the-bucket-count) 👍

config/config.go
```diff
@@ -67,6 +67,8 @@ type Config struct {
 	EnableRouteLIFOMetrics           bool           `yaml:"enable-route-lifo-metrics"`
 	MetricsFlavour                   *listFlag      `yaml:"metrics-flavour"`
 	DisableMetricsCompression        bool           `yaml:"disable-metrics-compression"`
+	EnableNativeHistograms           bool           `yaml:"enable-native-histograms"`
+	NativeHistogramBucketFactor      float64        `yaml:"native-histogram-bucket-factor"`
```
AlexanderYastrebov: These are Prometheus-specific so I am wondering whether they should include `prometheus` in the name. @szuecs WDYT?

metrics/prometheus.go
AlexanderYastrebov: Please update PR description accordingly.

config/config.go
szuecs: Yeah makes sense I would propose
- EnablePrometheusNativeHistograms
- PrometheusNativeHistogramBucketFactor

szuecs: Flags should also contain `prometheus`

AshrafAhmed9: Renamed to EnablePrometheusNativeHistograms / PrometheusNativeHistogramBucketFactor, flags to -enable-prometheus-native-histograms / -prometheus-native-histogram-bucket-factor, yaml keys and docs updated accordingly.

docs/operation/operation.md
```diff
@@ -284,6 +284,21 @@ option to enable it:
 It will return [Prometheus](https://prometheus.io/) metrics on the
 common metrics endpoint :9911/metrics.
 
+### Prometheus native histograms
```
szuecs: Can you move all the additions here to just before the `### OpenTelemetry (OTel)`, otherwise we would break here the rest of the docs.

szuecs:
@AshrafAhmed9 one last comment to move the block in the docs, rest is fine and can be merged as is.

AshrafAhmed9:
Done — moved the native histograms block to just before the OpenTelemetry section.

szuecs (approved):

metrics/prometheus.go
```diff
@@ -82,6 +82,9 @@ type Prometheus struct {
 // NewPrometheus returns a new Prometheus metric backend.
 func NewPrometheus(opts Options) *Prometheus {
 	opts = applyCompatibilityDefaults(opts)
+	if opts.EnablePrometheusNativeHistograms && opts.PrometheusNativeHistogramBucketFactor <= 1 {
+		opts.PrometheusNativeHistogramBucketFactor = 1.1
```
MustafaSaber: I would make a variable with the default and use it here and add a comment for why we chose this default

```golang
// KEP or prometheus link
DefaultNativeHistogramFactor = 1.1
...
if opts.EnablePrometheusNativeHistograms && opts.PrometheusNativeHistogramBucketFactor <= 1 {
 opts.PrometheusNativeHistogramBucketFactor = DefaultNativeHistogramFactor      
}
```

This will use the default for library users

config/config.go
```diff
@@ -448,6 +450,8 @@ func NewConfig() *Config {
 	flag.BoolVar(&cfg.EnableRouteLIFOMetrics, "enable-route-lifo-metrics", false, "enable metrics for the individual route LIFO queues")
 	flag.Var(cfg.MetricsFlavour, "metrics-flavour", "Metrics flavour is used to change the exposed metrics format. Supported metric formats: 'codahale', 'prometheus' and 'otel', you can select multiple or all of them by using one option with ',' separated values")
 	flag.BoolVar(&cfg.DisableMetricsCompression, "disable-metrics-compression", false, "disable metrics compression on /metrics handler endpoint.")
+	flag.BoolVar(&cfg.EnablePrometheusNativeHistograms, "enable-prometheus-native-histograms", false, "enables prometheus native histograms in addition to the classic bucketed histograms")
+	flag.Float64Var(&cfg.PrometheusNativeHistogramBucketFactor, "prometheus-native-histogram-bucket-factor", 0, "resolution of prometheus native histograms, must be greater than 1, defaults to 1.1 when native histograms are enabled")
```
MustafaSaber: Would make the default value matches the default in the description so also

```suggestion
	flag.Float64Var(&cfg.PrometheusNativeHistogramBucketFactor, "prometheus-native-histogram-bucket-factor", prometheus.DefaultNativeHistogramFactor, "resolution of prometheus native histograms, must be greater than 1, defaults to 1.1 when native histograms are enabled")
```

This will use the default from the flag

metrics/prometheus.go
```diff
@@ -340,6 +343,22 @@ func (p *Prometheus) sinceS(start time.Time) float64 {
...
+// configured, keeping the classic buckets for backward compatibility.
+func (p *Prometheus) histogramOpts(o prometheus.HistogramOpts) prometheus.HistogramOpts {
+	if p.opts.EnablePrometheusNativeHistograms {
+		// with a native histogram bucket factor set, empty buckets would
+		// disable the classic histogram instead of using the defaults
+		if len(o.Buckets) == 0 {
+			o.Buckets = prometheus.DefBuckets
+		}
+		o.NativeHistogramBucketFactor = p.opts.PrometheusNativeHistogramBucketFactor
+		o.NativeHistogramMaxBucketNumber = 160
```
MustafaSaber, line 365: Another comment with a link for why we choose this one would also be good, same for the time if we have it

MustafaSaber (commented):
Added some minor comments but in general the PR is good and we can merge, but I think it's better to address them to document some decisions that were taken on the fly

config/config.go
AshrafAhmed9: Addressed all three: extracted metrics.DefaultNativeHistogramFactor = 1.1 with a link to the KEP, used it as the flag default as suggested, and added a comment documenting the 160-bucket / 1h reset choice with links to the Prometheus native histograms spec and the KEP.

MustafaSaber (approved):

config/config.go
MustafaSaber: Thank you!

szuecs:
:+1:

szuecs:
Thanks for the great work!

Issue #4103, opened 2026-06-30
RouteSRV: keep zone information while serializing to eskip


greeshma1196:
Currently RouteSRV drops zone information when it serializes it to eskip and sets the endpoints like so `<roundRobin, "http://10.2.0.1:8080", "http://10.2.0.2:8080">`. 
In order for the user to plot latency by zone for skipper-ingress to backend in which RouteSRV is the dataclient, the zone information needs to preserved. One idea is to set the zone as a query parameter for the endpoint `<roundRobin, "http://10.2.0.1:8080?zone=eu-central-1a", "http://10.2.0.2:8080?zone=eu-central-1b">`.

How to measure latency by zone?
Similar to how [MeasureBackend](https://github.com/zalando/skipper/blob/3d2c5a7488aeab25465d94005a445780723c27be/metrics/metrics.go#L43) is implemented, the idea is to do the same but append zone information as well. This implies that the zone data needs to be preserved all the way to the proxy, so that when skipper has to select the endpoint for load balancing it will also pass the zone of where the endpoint is in.

MustafaSaber:
One idea would be by exposing an extra endpoint for endpoints zone, that can return a map of zones & endpoints and we can then match/fill data later in skipper.

szuecs:
I think I like the idea of adding the proposed query better, than an extra API endpoint, because if we need to transfer more metadata about an endpoint we have this in the normal route flow creation.

@AlexanderYastrebov do you have an opinion on that?

MustafaSaber:
This is already done since the PR is merged

PR #4111, opened 2026-07-03
zone aware routing: latency by zone + rps by zone 

greeshma1196:
This PR introduces two changes:
1. When I operate skipper-ingress controller across multiple availability zones, I want backend request latency labeled by the destination endpoint's zone, So that I can plot per-zone latency and RPS by zone

RPS by zone:
image

Latency by zone: 
image

2. When I run skipper-ingress that consume routes from RouteSRV, I want each backend endpoint's availability zone to survive the eskip wire format, So that per-zone backend latency can be plotted when RouteSRV is used, the same as in the in-process kubernetes dataclient mode - ref: https://github.com/zalando/skipper/issues/4103

Result: In the test cluster, there were 2 instances of skipper-ingress one set to Zone-A, the other to Zone-B

route-dump of skipper-ingress in zone A
```
kube_rg__skipper_loadtest_target__loadtest_target__all__0_0: Host(/^(skipper-loadtest[.]pg9-test[.]zalan[.]do[.]?(:[0-9]+)?)$/) && PathSubtree("/")
  -> disableAccessLog(2, 3, 404, 429)
  -> fifo(2000, 20, "1s")
  -> disableAccessLog()
  -> oauthTokeninfoValidate("{optOutAnnotations: [iam.zalando.org/public], optOutHosts: [\"^.*[.]ingress[.]cluster[.]local\"], unauthorizedResponse: \"Authentication required, see https://cloud.docs.zalando.net/howtos/authenticate-requests/#default-authentication\\n\"}")
  -> stateBagToTag("auth-user", "client.uid")
  -> <powerOfRandomNChoices, "http://[2a05:d024:90:1103:6d8f::4]:9090?zone=eu-central-1a", "http://[2a05:d024:90:1103:931f::b]:9090?zone=eu-central-1a", "http://[2a05:d024:90:1103:931f::c]:9090?zone=eu-central-1a">;
```

route-dump of skipper-ingress in zone B
```
kube_rg__skipper_loadtest_target__loadtest_target__all__0_0: Host(/^(skipper-loadtest[.]pg9-test[.]zalan[.]do[.]?(:[0-9]+)?)$/) && PathSubtree("/")
  -> disableAccessLog(2, 3, 404, 429)
  -> fifo(2000, 20, "1s")
  -> disableAccessLog()
  -> oauthTokeninfoValidate("{optOutAnnotations: [iam.zalando.org/public], optOutHosts: [\"^.*[.]ingress[.]cluster[.]local\"], unauthorizedResponse: \"Authentication required, see https://cloud.docs.zalando.net/howtos/authenticate-requests/#default-authentication\\n\"}")
  -> stateBagToTag("auth-user", "client.uid")
  -> <powerOfRandomNChoices, "http://[2a05:d024:90:1104:99f6::20]:9090?zone=eu-central-1b", "http://[2a05:d024:90:1104:99f6::24]:9090?zone=eu-central-1b", "http://[2a05:d024:90:1104:99f6::c]:9090?zone=eu-central-1b">;
```


eskip/eskip.go
```diff
@@ -391,6 +389,36 @@ func (ep LBEndpoint) String() string {
 	return ep.Address
 }
 
+// StringWithZone returns the endpoint address with the availability zone encoded
+// as a "zone" query parameter.
+func (ep LBEndpoint) StringWithZone() string {
+	if ep.Zone == "" {
+		return ep.Address
+	}
+	return ep.Address + "?zone=" + url.QueryEscape(ep.Zone)
```
szuecs: `"zone"` should be a const and used all over the place

szuecs:
Rest is good to go from my side.

szuecs:
:+1:

szuecs:
If you want you can merge and create a new PR to refactor the `const`

eskip/eskip.go
```diff
@@ -391,6 +389,36 @@ func (ep LBEndpoint) String() string {
...
+}
+
+// newLBEndpoint builds an LBEndpoint from a endpoint string, extracting the
+// availability zone from the "zone" query parameter if present. Only the "zone"
+// key is stripped; the remaining address is kept clean. Address must not carry
+// the zone param so that it can be used as it was originally defined
+func newLBEndpoint(s string) *LBEndpoint {
+	u, err := url.Parse(s)
+	if err != nil {
+		return &LBEndpoint{Address: s}
```
MustafaSaber, line 409: This seems wrong, if we have error parsing we return the endpoint without any problem. Is this expected?

CONTEXT.md
```diff
@@ -261,6 +261,13 @@ Central route server precomputes per-zone route sets; each proxy fetches `/route
 **DataClient Mode**:
 Each proxy instance filters endpoints to its own zone at the dataclient layer during endpoint fetch.
 
+**Backend zone**:
+The availability zone of the selected destination endpoint, sourced from the
+Kubernetes EndpointSlice `zone` field (e.g. `eu-central-1c`). It is the zone the
+traffic is sent *to*. Recorded as the `zone` metric label whenever present, else
+empty.
+_Avoid_: AZ, endpoint zone, destination zone
```
MustafaSaber, line 269: What does this avoid mean?

MustafaSaber: ah this is context.md, but also I don't understand how it will avoid them

eskip/eskip.go
greeshma1196: It will likely not hit the error condition here because the upstream that sends the endpoint list does the necessary validation while constructing it; refer https://github.com/zalando/skipper/blob/82d5831eb30c7a42a2d329871bb3d247969ffb44/dataclients/kubernetes/clusterstate.go#L82  https://github.com/zalando/skipper/blob/82d5831eb30c7a42a2d329871bb3d247969ffb44/dataclients/kubernetes/clusterstate.go#L174
Moreover this is exactly what was done earlier, refer https://github.com/zalando/skipper/blob/82d5831eb30c7a42a2d329871bb3d247969ffb44/eskip/eskip.go#L365 If we are not able to parse the string we send the Address as is.

CONTEXT.md
greeshma1196: The idea is to stick to using only "Backend Zone" to refer to the zone of backend endpoint and "avoid" using the other terms. It works when you use the [domain modelling skill](https://github.com/mattpocock/skills/tree/main/skills/engineering/domain-modeling)

I can remove it if you think its not needed

szuecs (approved):

CONTEXT.md
szuecs: As far as I understand the CONTEXT.md is the guideline for the AI to use the words that we want, not the ones that could be "similar".

eskip/eskip.go
szuecs: I think we should not have extra error handling here, so it seems the current state of the change is legit.
Basically nothing will break that was working before but we could also think of what we want, if we want to have it different.
How we would expect the `Address` to be?
Do we support the full URL or URI or only scheme+ip+port?
I would tend to let it like this.

CONTEXT.md
MustafaSaber: > I can remove it if you think its not needed

It's ok I'm just understanding 😄

eskip/eskip.go
MustafaSaber: The new problem introduced is the quey string, which `Parse` can return an error for. In this case I think we should fallback to the default behavior and return the string without the zone

> If we are not able to parse the string we send the Address as is.

My concern is about `10.0.0.1/?=nokey` in this case

At line 415 we will return
```golang
return LBEndpoint{Address: "10.0.0.1/?=nokey"}
```
But we are the one providing this URL so it should be ok

MustafaSaber (approved):

eskip/eskip.go
szuecs: But the same would be true in https://github.com/zalando/skipper/blob/82d5831eb30c7a42a2d329871bb3d247969ffb44/eskip/eskip.go#L365

greeshma1196: One more link for reference, this is where the zone query parameter is added https://github.com/zalando/skipper/blob/2d2524f3c59afeb21c385d19f940f76eb95a1f7d/eskip/eskip.go#L395 so we ensure the endpoint is constructed correctly.

PR #4109, opened 2026-07-02
go dependency: update  oras.land/oras-go/v2 to patched version

szuecs:
fixes several CVEs in the transitive dependency

PR #4112, opened 2026-07-03
Revert "go dependency: update  oras.land/oras-go/v2 to patched version"

szuecs:
Reverts zalando/skipper#4109

Problem is that opa instance times out at startup


szuecs:
:+1:

MustafaSaber (approved):

PR #4122, opened 2026-07-07
Fix ForwardBackend filters initialization in routegroup

a4180p:
Align filter cleanup behavior between ingress and routegroup processing in kubernetes dataclient.
Filters are cleaned up during the processing of ingress and left as is during the processing of routegroups.

IIUC the logic, default filters will not be cleared anyway, because they are added later in polling.go. 
This is the risk we should accept.


szuecs (approved):

a4180p:
👍

PR #4124, opened 2026-07-08
feat: add weightedRoundRobin load balancer algorithm

AshrafAhmed9:
Adds a `weightedRoundRobin` load balancer algorithm, as discussed in #557 with @szuecs and @MustafaSaber.

## How it works

- The endpoint registry now tracks a dynamic per-endpoint weight: the ratio of successful round trips within the last stats reset period, computed in the same loop that derives the passive health check drop probability. `Weight()` is added to the `routing.Metrics` interface (allowed under the weak-compatibility policy for the `routing` package).
- The new algorithm uses [smooth weighted round robin](https://github.com/nginx/nginx/commit/52327e0627f49dbda1e8db695e63a4b0af4448b1) over these weights, keyed by host so it works with the passive-health-check-filtered endpoint subset. With equal weights it behaves exactly like `roundRobin`.
- No eskip or Kubernetes plumbing was needed — the weight travels through the existing `LBEndpoint.Metrics` channel, per @MustafaSaber's suggestion.

## Design decisions to note

- **Weights require passive health check to be enabled**: the weight is updated by the PHC stats loop, so without `-passive-health-check` the algorithm behaves like `roundRobin` (documented in backends.md and the package docs). Happy to add a startup warning for this combination if preferred.
- **Weight floor**: weights are bounded below by `1 - maxHealthCheckDropProbability`, so failing endpoints keep receiving a small share of traffic to detect recovery — mirroring the PHC clamping semantics.
- The RouteGroup CRD enum (both `routegroups_crd.yaml` and the helm chart copy) is updated so the algorithm is accepted in RouteGroups.

## Testing

- Deterministic distribution test (smooth WRR is exactly proportional): weights 0.2/0.8/1.0 over 1000 rounds yield exact proportional shares.
- Equal-weights test asserting identical behavior to `roundRobin` (exact `rounds/N` per endpoint).
- Weight computation asserted in the endpoint registry stats test (success ratio, floor, and reset back to 1.0).
- Benchmark: ~447ns/op at 10 endpoints (vs ~5ns for `random`) — the mutex + map cost of smooth WRR, negligible against a proxied request.

Ref #557

### DIFF
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -272,6 +272,7 @@ func (p *powerOfRandomNChoices) getScore(e routing.LBEndpoint) int64 {
 
 type weightedRoundRobin struct {
 	mu             sync.Mutex
+	rnd            *rand.Rand
 	currentWeights map[string]float64
 }
 
@@ -280,6 +281,7 @@ type weightedRoundRobin struct {
 // registry within the last stats reset period.
 func newWeightedRoundRobin(endpoints []string) routing.LBAlgorithm {
 	return &weightedRoundRobin{
+		rnd:            rand.New(rand.NewPCG(uint64(time.Now().UnixNano()), 0)), // #nosec
 		currentWeights: make(map[string]float64, len(endpoints)),
 	}
 }
@@ -289,8 +291,11 @@ func newWeightedRoundRobin(endpoints []string) routing.LBAlgorithm {
 // Each endpoint accumulates its weight per round and the endpoint with the
 // highest accumulated weight is selected and reduced by the total weight of
 // the round. With equal weights it behaves like the roundrobin algorithm.
+// Iteration starts at a random offset so that ties are not biased towards
+// the first endpoint of the list.
 func (w *weightedRoundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
-	if len(ctx.LBEndpoints) == 1 {
+	ne := len(ctx.LBEndpoints)
+	if ne == 1 {
 		return ctx.LBEndpoints[0]
 	}
 
@@ -300,7 +305,9 @@ func (w *weightedRoundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 	total := 0.0
 	best := 0
 	bestWeight := math.Inf(-1)
-	for i := range ctx.LBEndpoints {
+	offset := w.rnd.IntN(ne) // #nosec
+	for k := 0; k < ne; k++ {
+		i := (offset + k) % ne
 		e := &ctx.LBEndpoints[i]
 		weight := e.Metrics.Weight()
 		total += weight
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -278,9 +278,9 @@ type weightedRoundRobin struct {
 // newWeightedRoundRobin creates a smooth weighted roundrobin algorithm
 // with weights based on the failed round trips observed by the endpoint
 // registry within the last stats reset period.
-func newWeightedRoundRobin([]string) routing.LBAlgorithm {
+func newWeightedRoundRobin(endpoints []string) routing.LBAlgorithm {
 	return &weightedRoundRobin{
-		currentWeights: make(map[string]float64),
+		currentWeights: make(map[string]float64, len(endpoints)),
 	}
 }
 
@@ -313,19 +313,6 @@ func (w *weightedRoundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
 		}
 	}
 	w.currentWeights[ctx.LBEndpoints[best].Host] -= total
-
-	// forget endpoints that are no longer part of the route
-	if len(w.currentWeights) > 2*len(ctx.Route.LBEndpoints) {
-		known := make(map[string]struct{}, len(ctx.Route.LBEndpoints))
-		for i := range ctx.Route.LBEndpoints {
-			known[ctx.Route.LBEndpoints[i].Host] = struct{}{}
-		}
-		for host := range w.currentWeights {
-			if _, ok := known[host]; !ok {
-				delete(w.currentWeights, host)
-			}
-		}
-	}
 
 	return ctx.LBEndpoints[best]
 }
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -3,6 +3,7 @@ package loadbalancer
 import (
 	"errors"
 	"fmt"
+	"math"
 	"math/rand/v2"
 	"sort"
 	"sync"
@@ -34,6 +35,10 @@ const (
 
 	// PowerOfRandomNChoices selects N random endpoints and picks the one with least outstanding requests from them.
 	PowerOfRandomNChoices
+
+	// WeightedRoundRobin distributes requests proportionally to the dynamic
+	// endpoint weights derived from observed failed round trips.
+	WeightedRoundRobin
 )
 
 const powerOfRandomNChoicesDefaultN = 2
@@ -48,6 +53,7 @@ var (
 		Random:                newRandom,
 		ConsistentHash:        newConsistentHash,
 		PowerOfRandomNChoices: newPowerOfRandomNChoices,
+		WeightedRoundRobin:    newWeightedRoundRobin,
 	}
 	defaultAlgorithm = newRoundRobin
 )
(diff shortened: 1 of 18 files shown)


MustafaSaber:
Thanks for the PR! One request would be, can you add a benchmark for other algorithms with & without `updateStats` always firing? (maybe in another PR) My expectation is that, it shouldn't add noticeable overhead

If we see no degradation we can always fire the goroutine and drop the PHC condition in this case we decouple this PHC. The problem is that we not only depend on PHC being enabled but also the maxHealthCheckDropPropability coming from PHC, so if we have some application using powerOfRandomNChoices & another one using weightedRoundRobin if we change maxHealthCheckDropPropability it will affect both (I'm thinking if should be the case)

Also a testcase to simulate a failing backend would be nice to have as I only see happy scenario tests

loadbalancer/algorithm.go
```diff
@@ -264,6 +270,53 @@ func (p *powerOfRandomNChoices) getScore(e routing.LBEndpoint) int64 {
...
+func (w *weightedRoundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
+	if len(ctx.LBEndpoints) == 1 {
+		return ctx.LBEndpoints[0]
+	}
+
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	total := 0.0
+	best := 0
```
szuecs, line 306: it's better to select randomly a default from the list, see also the round robin algorithm

loadbalancer/algorithm.go
```diff
@@ -264,6 +270,53 @@ func (p *powerOfRandomNChoices) getScore(e routing.LBEndpoint) int64 {
...
+// algorithm, see https://github.com/nginx/nginx/commit/52327e0627f49dbda1e8db695e63a4b0af4448b1.
+// Each endpoint accumulates its weight per round and the endpoint with the
+// highest accumulated weight is selected and reduced by the total weight of
+// the round. With equal weights it behaves like the roundrobin algorithm.
+func (w *weightedRoundRobin) Apply(ctx *routing.LBContext) routing.LBEndpoint {
+	if len(ctx.LBEndpoints) == 1 {
+		return ctx.LBEndpoints[0]
+	}
+
+	w.mu.Lock()
```
szuecs, line 302: I have no good feeling about this lock.
The problem here is that the iterations depend on number of LBEndpoints.
Think about: What if we have 1000 of them and 10k RPS in one instance

Maybe the better choice is to use a sync.Map

loadbalancer/algorithm_test.go
```diff
@@ -551,3 +556,127 @@ func BenchmarkRandomAlgorithm(b *testing.B) {
...
+	}
+
+	assert.Len(t, h, N)
+	for host, count := range h {
+		assert.Equal(t, rounds/N, count, "host %s", host)
+	}
+}
+
+func BenchmarkWeightedRoundRobinAlgorithm(b *testing.B) {
+	const N = 10
```
szuecs: can you create a loop to benchmark number of endpoints like this:
````
[]int{10,100,1000,10000}
````

File: loadbalancer/algorithm.go (hunk 1 above)
Comment (AshrafAhmed9) [reply loadbalancer/algorithm.go:306]: Done — Apply now starts iterating from a random offset, so ties in the accumulated weights no longer bias toward the first endpoint.

File: loadbalancer/algorithm.go (hunk 2 above)
Comment (AshrafAhmed9) [reply loadbalancer/algorithm.go:302]: Added the benchmarks you asked for (M5, single goroutine):


10      173 ns/op
100   1,602 ns/op
1000 16,569 ns/op
10000 171,775 ns/op

On sync.Map — I don't think it helps here, and may hurt. Smooth WRR needs the whole pick to be atomic as one transaction: read every endpoint's accumulated weight, add the current weights, select the max, subtract the total from the winner. That invariant (accumulators sum to ~0 across a cycle) only holds if concurrent Apply calls are serialized — sync.Map gives per-key atomicity, not per-Apply, so two requests could interleave mid-accumulation and skew the distribution. It also still does one write per endpoint per request, so it wouldn't shrink the O(n) work the lock holds.

So the real lever is the algorithm, not the map: keeping smooth WRR means keeping the serialized accumulator (~16.6µs under lock at 1000 endpoints, so one route nears saturation around ~60k RPS on a core). If that ceiling is too low for large pools, I'd switch selection to stateless weighted-random — no shared accumulator, lock only around the RNG — which trades the exact smooth distribution for O(n) lock-free reads. Happy to implement that if you prefer; otherwise I'll document the complexity and suggest powerOfRandomNChoices for very large pools.

File: loadbalancer/algorithm_test.go (hunk 3 above)
Comment (AshrafAhmed9) [reply loadbalancer/algorithm_test.go]: Done — BenchmarkWeightedRoundRobinAlgorithm now loops over {10, 100, 1000, 10000} (numbers above).

**Comment by AshrafAhmed9** [comment]
Added a failing-backend test: one endpoint takes 80% failed round trips through the registry's passive health check, and it asserts the healthy endpoints get >2× its traffic while it still receives some (to detect recovery).

Good points on the coupling. The updateStats benchmark (with/without the goroutine always firing) I'll do as the follow-up PR you suggested. On maxHealthCheckDropProbability bleeding across algorithms — agreed that's not ideal; if we decouple the stats loop from PHC there, a dedicated weight floor makes sense. Happy to fold that into the same follow-up once the direction's settled.

File: loadbalancer/algorithm.go (hunk 2 above)
Comment (szuecs) [reply loadbalancer/algorithm.go:302]: My fear is the lick contention over time.
If you 60k rps in one route in one instance at 1000 or 10000 endpoints it is likely a problem.
I don't have time right now to do the calculations but maybe you can try to use ./skptesting and run it for some minutes to a list of backends. I think I would use ai to build a minimal C server that responds with 200 status code and then you can run 10k processes on your machine for the large scale test.
You can add this test and c code to ./skptesting .
It would help a lot to test more of these algorithms.

This year we had one incident with a similar case that did an optimization and all benchmarks showed it's good until production which broke after a couple of minutes.

Comment (AshrafAhmed9) [reply loadbalancer/algorithm.go:302]: Built the harness: `skptesting/okserver.c` (minimal C server, one process per endpoint, or a port range per process for low process limits) + `skptesting/benchmark-lb-algorithms.sh`. Committed so it can be reused for the other algorithms.

Results at **1000 endpoints**, 256 connections, 1M requests each, zero errors (M5 laptop, load generator on the same machine):

    algorithm              req/s    p50     p99
    roundRobin             50,438   4.1ms   17.6ms
    powerOfRandomNChoices  50,454   3.8ms   19.3ms
    weightedRoundRobin     31,038   8.3ms   13.5ms

The lock costs ~38% of peak throughput vs the O(1) algorithms, but still sustains ~31k RPS in one instance — 3× your 10k scenario — with stable latency over the whole run.

Testbed notes: at **10000 endpoints** the single machine hits ephemeral-port exhaustion before any algorithm limit (roundRobin already ~46% 502s), so a true 10k test needs Linux with a wide `ip_local_port_range` — the script is ready for that. I also used `-close-idle-conns-period=60s`; the 3s from the other benchmarks turns large-pool runs into connection-churn benchmarks.

My take: fine for realistic pools and your 10k RPS case; for thousands of endpoints at six-figure RPS I'd document recommending `powerOfRandomNChoices`. If you'd rather remove the ceiling entirely, the stateless weighted-random option still stands.

Comment (szuecs) [reply loadbalancer/algorithm.go:302]: Thanks!

**Review by szuecs — APPROVED** [review]

File: loadbalancer/algorithm.go (hunk 2 above)
Comment (AshrafAhmed9) [reply loadbalancer/algorithm.go:302]: Thanks for the approval! One note on CI: `check-race_skipper` only covers the root package, which hasn't changed since it last passed — the commits since then only touch `loadbalancer/`, `skptesting/`, and `.gitignore`. Root is also where the known-flaky `TestMTLS_ProxyToBackend` (#4096) lives. Could you re-run the failed job?

File: loadbalancer/algorithm_test.go [hunk 4]
```diff
@@ -551,3 +558,215 @@ func BenchmarkRandomAlgorithm(b *testing.B) {
...
+	endpointRegistry.Do([]*routing.Route{r})
+
+	failing := rt[0].LBEndpoints[0].Metrics
+
+	// Continuously feed the registry so that every stats reset period is
+	// populated: the first endpoint sees 80% failed round trips, the other
+	// endpoints only successful ones. This keeps the reduced weight stable
+	// regardless of when the background stats goroutine fires, avoiding a
+	// dependency on a single stats window.
+	done := make(chan struct{})
```
MustafaSaber: The test case is ok, but I think a realistic one was going to be better, make 3 test servers, one of them is failing forever, run vegeta attack to make load on this route then do the assertions after some time

loadbalancer/algorithm_test.go
```diff
@@ -551,3 +558,215 @@ func BenchmarkRandomAlgorithm(b *testing.B) {
...
+	endpointRegistry := routing.NewEndpointRegistry(routing.RegistryOptions{
+		PassiveHealthCheckEnabled:     true,
+		StatsResetPeriod:              100 * time.Millisecond,
+		MinRequests:                   10,
+		MinHealthCheckDropProbability: 0.05,
+		MaxHealthCheckDropProbability: 0.95,
+	})
+	defer endpointRegistry.Close()
+
+	p := NewAlgorithmProvider()
```
MustafaSaber: Please use better naming for variables, I understand this is a testcase but when I read I have to keep going up and down to maintain the knowledge

skptesting/benchmark-lb-algorithms.sh
```diff
@@ -0,0 +1,147 @@
...
+# The backends are okserver processes (see okserver.c), each responding
+# with a fixed 200 on its own port. One skipper instance load balances a
+# single route across all of them, and the load generator drives it for
+# the configured duration, once per algorithm.
+#
+# usage: benchmark-lb-algorithms.sh [backends] [duration] [connections] [algorithms] [skipper-binary]
+#
+# example: benchmark-lb-algorithms.sh 1000 2m 128 "roundRobin weightedRoundRobin"
+#
+# The load is generated with wrk (https://github.com/wg/wrk), or hey
```
MustafaSaber, line 16: For future PRs, we usually use vegeta

MustafaSaber (commented):
Thanks for the follow ups! Small comments then we can merge after. I care more about the naming convention in the test cases

loadbalancer/algorithm_test.go
AshrafAhmed9: Done — replaced it with a realistic test in the proxy package (`TestWeightedRoundRobinWithFailingBackend`): two healthy backends plus one that always exceeds `backendTimeout("20ms")`, load driven by vegeta for 5s, asserting the failing backend gets a reduced share while still receiving some traffic to detect recovery. PHC filtering stays disabled in the proxy params so the distribution shift is attributable to the weights alone, not endpoint dropping.

loadbalancer/algorithm_test.go
AshrafAhmed9: Fair point — renamed throughout (endpointAddresses, selectionCounts, route, registry, ...) and extracted the shared setup into helpers so the test bodies read top-to-bottom.

skptesting/benchmark-lb-algorithms.sh
AshrafAhmed9: Good to know, thanks — used vegeta for the new proxy test; I'll default to it going forward.

proxy/healthy_endpoints_test.go
```diff
@@ -344,3 +345,56 @@ func TestPHCMaxUnhealthyEndpointsRatioParam(t *testing.T) {
 		assert.InDelta(t, 0.0, float64(counters["passive-health-check.requests.passed"]), 0.3*float64(nRequests)) // allow 30% error
 	})
 }
+
+func TestWeightedRoundRobinWithFailingBackend(t *testing.T) {
+	if testing.Short() {
```
MustafaSaber, line 350: 👍

MustafaSaber (approved):
Thanks for the great contribution!

szuecs:
:+1:

MustafaSaber:
@AshrafAhmed9 can you create a follow up on https://github.com/szuecs/routegroup-client to update routegroup CRD with the new algorithm?

AshrafAhmed9:
Done: szuecs/routegroup-client#66

PR #4138, opened 2026-07-17
Fix bounded consistent hash with filtered endpoints

### LINKED ISSUE #4139: Fix bounded consistent hash with filtered endpoints
Fix `boundedLoadSearch` so it advances to the next hash ring entry when the current endpoint has been filtered out. Filtered endpoints can happen before load balancing, for example through fade-in.

In bounded-load mode, the algorithm starts from the normal consistent-hash target. If that endpoint is overloaded, it walks forward in the hash ring looking for another endpoint below the load threshold.

Before this change, if the walk reached a filtered endpoint, the loop continued without advancing the ring index. This could repeatedly check the same filtered endpoint and eventually return it.

Example case:
```
Route endpoints: [A, B, C]
Candidate endpoints: [A, C]    // B was filtered out

Hash ring walk:  A -> B -> C

- A is selected by consistent hash but is overloaded.
- The algorithm moves to B.
- B is not in the candidate list, so it should be skipped.
- Previously the ring index was not advanced when skipping B.

With this change, the algorithm advances to C.
```

## Tests
```
go test ./loadbalancer -run TestConsistentHashBoundedLoadSearchSkipsFilteredEndpoints
go test ./loadbalancer
```

Suggested label: bug

PR: https://github.com/zalando/skipper/pull/4138

shrikant23codes:
Fix `boundedLoadSearch` so it advances to the next hash ring entry when the current endpoint has been filtered out. Filtered endpoints can happen before load balancing, for example through fade-in.

In bounded-load mode, the algorithm starts from the normal consistent-hash target. If that endpoint is overloaded, it walks forward in the hash ring looking for another endpoint below the load threshold.

Before this change, if the walk reached a filtered endpoint, the loop continued without advancing the ring index. This could repeatedly check the same filtered endpoint and eventually return it.

Example case:
```
Route endpoints: [A, B, C]
Candidate endpoints: [A, C]    // B was filtered out

Hash ring walk:  A -> B -> C

- A is selected by consistent hash but is overloaded.
- The algorithm moves to B.
- B is not in the candidate list, so it should be skipped.
- Previously the ring index was not advanced when skipping B.

With this change, the algorithm advances to C.
```

## Tests
```
go test ./loadbalancer -run TestConsistentHashBoundedLoadSearchSkipsFilteredEndpoints
go test ./loadbalancer
```

Suggested label: bug
Fixes: https://github.com/zalando/skipper/issues/4139

### DIFF
--- a/loadbalancer/algorithm.go
+++ b/loadbalancer/algorithm.go
@@ -188,6 +188,7 @@ func (ch *consistentHash) boundedLoadSearch(key string, balanceFactor float64, c
 	for i := 0; i < ch.Len(); i++ {
 		endpointIndex := ch.hashRing[ringIndex].index
 		if skipEndpoint(ctx, endpointIndex) {
+			ringIndex = (ringIndex + 1) % ch.Len()
 			continue
 		}
 		load := ctx.Route.LBEndpoints[endpointIndex].Metrics.InflightRequests()
--- a/loadbalancer/algorithm_test.go
+++ b/loadbalancer/algorithm_test.go
@@ -701,3 +701,81 @@ func BenchmarkWeightedRoundRobinAlgorithm(b *testing.B) {
 		})
 	}
 }
+
+func TestConsistentHashBoundedLoadSearchSkipsFilteredEndpoints(t *testing.T) {
+	endpoints := []string{
+		"http://127.0.0.1:8080",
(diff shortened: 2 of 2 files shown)


MustafaSaber (approved):
Thanks, nice catch!

szuecs (approved):

PR #3882, opened 2026-02-23
tracing: to reduce costs on the next tracing provider, add a config option migrate span logs/events to span tags/attributes

szuecs:
tracing: to reduce costs on the next tracing provider, add a config option migrate span logs/events to span tags/attributes

In some tracing providers this has the advantage that you can create timeseries of the instrumented http clients: example connect() times, ..

PR #4148, opened 2026-07-24
tracing: use tags to reduce cost

MustafaSaber:
This is a follow up on https://github.com/zalando/skipper/pull/3882 where we didn't use the flag in some hotspots and is causing producing huge amount of span events specially filters tracing


szuecs (approved):

MustafaSaber:
👍

PR #4159, opened 2026-07-28
return failClosed for failed request in redis and valkey

vlktna:
For the cases of client canceled requests (i.e. client canceled context) the span should not be marked as valkey.error: failedToDetermineAllow.


MustafaSaber:
Can you please fix DCO?

szuecs (approved):

vlktna:
👍

PR #4166, opened 2026-08-03
fix flaky tests

vlktna:
Fix three failing tests:

- `TestBackendTimeoutWithSlowBodyWriterShadow` remove an incorrect log assertion. The timeout happens while draining the shadow response body after headers were received, so the loopback request error is not always logged
- `TestMtlsRoutesrv` fix the inverted readiness check. The test previously stopped waiting on the first connection error instead of retrying until /health returned 204
- `TestAdvisoryDenyOnPresencePolicyNotBypassed` update the test Rego policy to deny truncated request bodies. Oversized JSON cannot be fully parsed, so the policy must handle input.truncated_body explicitly

### DIFF
--- a/filters/openpolicyagent/ghsa-8qqm-fp2q-v734_test.go
+++ b/filters/openpolicyagent/ghsa-8qqm-fp2q-v734_test.go
@@ -293,6 +293,10 @@ import rego.v1
 default allow := true
 
 allow := false if {
+    input.truncated_body
+}
+
+allow := false if {
     input.parsed_body.action == "delete"
 }
 `,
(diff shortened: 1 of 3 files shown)


MustafaSaber (commented):
Let's focus on this PR on fixing the reproducer  and make other 2 on another PR if it's needed, seems like majority of failures are coming from the reproducer

vlktna:
@MustafaSaber 
[OPA truncated body](https://github.com/zalando/skipper/pull/4168)
[shadow timeout assertion](https://github.com/zalando/skipper/pull/4169)
[routesrv mTLS readiness](https://github.com/zalando/skipper/pull/4170)

didn't change this PR to save check logs for proxy race fix https://github.com/zalando/skipper/pull/4167

PR #4168, opened 2026-08-03
test(opa): deny requests with truncated bodies

vlktna:
TestAdvisoryDenyOnPresencePolicyNotBypassed update the test Rego policy to deny truncated request bodies. Oversized JSON cannot be fully parsed, so the policy must handle input.truncated_body explicitly


MustafaSaber (approved):

vlktna:
👍

PR #4169, opened 2026-08-03
test(proxy): remove timing-dependent shadow log assertion

vlktna:
TestBackendTimeoutWithSlowBodyWriterShadow remove an incorrect log assertion. The timeout happens while draining the shadow response body after headers were received, so the loopback request error is not always logged

### DIFF
--- a/proxy/fifo_backendtimeout_with_teeloopback_test.go
+++ b/proxy/fifo_backendtimeout_with_teeloopback_test.go
@@ -101,12 +101,6 @@ func TestBackendTimeoutWithSlowBodyWriterShadow(t *testing.T) {
 		return
 	}
 
-	if err := proxyLog.WaitFor("failed to execute loopback request: dialing failed false: context deadline exceeded", time.Second); err != nil {
-		t.Fatalf("Failed to get expected error: %v", err)
-	} else {
-		t.Log(`Found "failed to execute loopback request" error log`)
-	}
-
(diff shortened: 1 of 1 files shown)


szuecs:
I don't think that's the correct fix for the flakytest. We should make sure we have the required log found, not sure how but this sounds more like a bug

vlktna:
@szuecs as I understand this test mainly checks the timeout while reading a slow response body `context: error while discarding remainder response body`

the `failed to execute loopback request` log appears only when a request times out before receiving response headers. With 500 requests this sometimes happens because of scheduling, but the test does not guarantee it. More requests increase the chance, but the log can still be missing

we could keep the body-timeout check here and add a separate test where the backend delays the response headers. The new test could reliably check the `failed to execute loopback request` log

szuecs:
> @szuecs as I understand this test mainly checks the timeout while reading a slow response body `context: error while discarding remainder response body`
> 
> the `failed to execute loopback request` log appears only when a request times out before receiving response headers. With 500 requests this sometimes happens because of scheduling, but the test does not guarantee it. More requests increase the chance, but the log can still be missing
> 
> we could keep the body-timeout check here and add a separate test where the backend delays the response headers. The new test could reliably check the `failed to execute loopback request` log

Yes, then let's do this.

vlktna:
👍

szuecs (approved):

Issue #4145, opened 2026-07-21
HTTP/2 Support


molikuner:
**Is your feature request related to a problem? Please describe.**
We're currently running skipper as our API Gateway for external and internal traffic. There are upcoming internal requirements to support gRPC traffic.

**Describe the solution you would like**
Skipper should be able to receive and make unencrypted HTTP/2 requests and responses (including Trailers), such that gRPC traffic can flow through it.

**Describe alternatives you've considered (optional)**
 * Route gRPC traffic outside of Skipper (and optionally replace Skipper entirely)
 * Do not support gRPC traffic

**Additional context (optional)**
At the moment we do not have any strict requirements to inspect the gRPC traffic. For now we would simply like to route traffic based on [Header](https://opensource.zalando.com/skipper/reference/predicates/#header) and potentially [Path](https://opensource.zalando.com/skipper/reference/predicates/#path) predicates to different backends.

**Would you like to work on it?**
Yes, but likely we would need some support to answer some open questions (see below).

**Related Issues**
#1253
#906
#4085

**Identified issues so far**
 * Skipper silently drops Trailers from responses
 * Skipper doesn't accept h2c requests
 * Skipper doesn't create h2c & h2 requests

**Open questions**
 * Would you be open to accepting contributions in this direction?
 * Should there be a dedicated configuration option to enable HTTP/2 support (incoming and/or outgoing)?
 * Should there be a dedicated configuration option to enable Trailers forwarding?
 * #4085 mentions `NewClientConn`. Would contributions to enable HTTP/2 only be accepted if they're migrating to `NewClientConn`, or would it be okay to continue to use the existing `Transport` infrastructure for now?
 * Should the outgoing connections start to using the `AppProtocol`, e.g. values like `kubernetes.io/h2c`? How would this relate to the `zalando.org/skipper-backend-protocol` annotation?

szuecs:
In general it would be a nice contribution. It's really part of the core and likely we need to build it on our own. I waited for the official Go support without x/net.
After my vacation I wanted to focus on this anyways so likely you need to wait a bit.
There are internal needs that also want to have grpc support.
What I'm not sure about is h2c support. I always thought it is a good idea but I think the Go maintainers have a different idea. Therefore I checked SPIFFE/SPIRE and it would enable plain h2. When I am done I will document it so it's easy to build the system, but of course it is not a trivial setup.

Your issue will have to be done by multiple PRs: net.Client, proxy, server handler and kubernetes.

[...]

Issue #1253, opened 2019-11-29
implement H2C to proxy to h2 over TCP


szuecs:
**Is your feature request related to a problem? Please describe.**

For example in Kubernetes, people want to use server push, etc. features from h2.

- server/client ALPN/Upgrade h2/h2c: https://github.com/nekolunar/http2
- very short unmaintained: https://github.com/owlwalks/unload/blob/master/main.go
- h2i console https://godoc.org/golang.org/x/net/http2/h2i
- h2c handler https://godoc.org/golang.org/x/net/http2/h2c#ex-NewHandler
- h2 configure server https://godoc.org/golang.org/x/net/http2#ConfigureServer

**Describe the solution you would like**

The proxy should be able to connect with H2C.

**Describe alternatives you've considered (optional)**

Create mTLS infrastructure, which can be a huge change without providing too much value

**Would you like to work on it?**
Yes

szuecs:
https://github.com/zalando/skipper/pull/4171 was merged and released https://github.com/zalando/skipper/releases/tag/v0.27.57

Issue #3844, opened 2026-01-27
Separate connection pool to backends


szuecs:
During an investigation of a high traffic low latency system we found that access to connection pool in the Go runtime added more than 20ms latency (lock contention in http transport).
We should consider that we can split this by application or route or endpoint.
Likely by route is the simplest but by application the best. 
Maybe by endpoint could be a good compromise.

szuecs:
# Plan: Separate Connection Pool per Endpoint

## Context

Under high-traffic, low-latency conditions, the single shared `*http.Transport` in Skipper causes >20ms latency spikes due to lock contention in Go's runtime connection pool. All routes and all backend endpoints share one transport, so concurrent requests contend on the same mutex inside `http.Transport`.

**Goal:** Split the connection pool so each unique backend endpoint (`scheme://host:port`) gets its own `*http.Transport`, eliminating cross-endpoint lock contention.

**Chosen granularity: by endpoint** — aligns with the existing `EndpointRegistry` (already keyed by `host:port`), avoids transport explosion from routes sharing the same backend, and is simpler than the "by application" option which has no existing concept in routing.

---

## Critical Files

| File | Role |
|---|---|
| `proxy/proxy.go` | Core change: `Proxy` struct, `WithParams()`, `getRoundTripper()`, `Close()` |
| `proxy/transport_registry.go` | New file: `transportRegistry` type |
| `proxy/transport_registry_test.go` | New file: unit tests + benchmark |
| `routing/endpointregistry.go` | Phase 2: add `onRemove` callback for proactive cleanup |
| `proxy/idle_test.go` | Extend with per-endpoint isolation test |

---

## Implementation Steps

### Step 1 — New type `transportRegistry` in `proxy/transport_registry.go`

```go
type transportRegistry struct {
    mu      sync.RWMutex
    base    *http.Transport
    wrap    func(http.RoundTripper) http.RoundTripper
    entries map[string]http.RoundTripper  // key: "scheme://host:port"
}
```

- `base`: the template `*http.Transport` (never sends requests, just cloned for settings)
- `wrap`: `CustomHttpRoundTripperWrap` from `Params`, applied once at creation
- `entries`: per-endpoint map, protected by `sync.RWMutex`

**`getOrCreate(key string) http.RoundTripper`** — hot path:
1. `RLock` → lookup → return on hit (zero write contention on fast path)
2. On miss: write lock + double-checked locking → clone transport → apply wrap → store → return

[...]

PR #4171, opened 2026-08-04
feature: enable h2c for server handler, proxy client and net client

szuecs:
ref: https://github.com/zalando/skipper/issues/4145
ref: https://github.com/zalando/skipper/issues/1253

Test Direct Prior Knowledge:
```
% nghttp -v http://localhost:9001/path
[  0.000] Connected
[  0.000] send SETTINGS frame <length=12, flags=0x00, stream_id=0>
          (niv=2)
          [SETTINGS_MAX_CONCURRENT_STREAMS(0x03):100]
          [SETTINGS_INITIAL_WINDOW_SIZE(0x04):65535]
[  0.001] send PRIORITY frame <length=5, flags=0x00, stream_id=3>
          (dep_stream_id=0, weight=201, exclusive=0)
[  0.001] send PRIORITY frame <length=5, flags=0x00, stream_id=5>
          (dep_stream_id=0, weight=101, exclusive=0)
[  0.001] send PRIORITY frame <length=5, flags=0x00, stream_id=7>
          (dep_stream_id=0, weight=1, exclusive=0)
[  0.001] send PRIORITY frame <length=5, flags=0x00, stream_id=9>
          (dep_stream_id=7, weight=1, exclusive=0)
[  0.001] send PRIORITY frame <length=5, flags=0x00, stream_id=11>
          (dep_stream_id=3, weight=1, exclusive=0)
[  0.001] send HEADERS frame <length=43, flags=0x25, stream_id=13>
          ; END_STREAM | END_HEADERS | PRIORITY
          (padlen=0, dep_stream_id=11, weight=16, exclusive=0)
          ; Open new stream
          :method: GET
          :path: /path
          :scheme: http
          :authority: localhost:9001
          accept: */*
          accept-encoding: gzip, deflate
          user-agent: nghttp2/1.59.0
[  0.001] recv SETTINGS frame <length=30, flags=0x00, stream_id=0>
          (niv=5)
          [SETTINGS_MAX_FRAME_SIZE(0x05):1048576]
          [SETTINGS_MAX_CONCURRENT_STREAMS(0x03):250]
          [SETTINGS_MAX_HEADER_LIST_SIZE(0x06):1048896]
          [SETTINGS_HEADER_TABLE_SIZE(0x01):4096]
          [SETTINGS_INITIAL_WINDOW_SIZE(0x04):1048576]
[  0.001] send SETTINGS frame <length=0, flags=0x01, stream_id=0>
          ; ACK
          (niv=0)
[  0.001] recv SETTINGS frame <length=0, flags=0x01, stream_id=0>
          ; ACK
          (niv=0)
[  0.001] recv WINDOW_UPDATE frame <length=4, flags=0x00, stream_id=0>
          (window_size_increment=983041)
[  0.008] recv (stream_id=13) :status: 201
[  0.008] recv (stream_id=13) date: Tue, 04 Aug 2026 22:19:10 GMT
[  0.008] recv (stream_id=13) server: Skipper
[  0.008] recv HEADERS frame <length=36, flags=0x04, stream_id=13>
          ; END_HEADERS
          (padlen=0)
          ; First response header
[  0.008] recv DATA frame <length=0, flags=0x01, stream_id=13>
          ; END_STREAM
[  0.008] send GOAWAY frame <length=8, flags=0x00, stream_id=0>
          (last_stream_id=0, error_code=NO_ERROR(0x00), opaque_data(0)=[])
```

test with upgrade does not work:
```
% nghttp -uv http://localhost:9001/path
[  0.000] Connected
[  0.000] HTTP Upgrade request
GET /path HTTP/1.1
host: localhost:9001
connection: Upgrade, HTTP2-Settings
upgrade: h2c
http2-settings: AAMAAABkAAQAAP__
accept: */*
user-agent: nghttp2/1.59.0

[  0.002] HTTP Upgrade response
HTTP/1.1 500 Internal Server Error
Content-Length: 22
Server: Skipper
Date: Tue, 04 Aug 2026 22:19:39 GMT

Internal Server Error

[ERROR] HTTP Upgrade failed
Some requests were not processed. total=1, processed=0
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -299,7 +299,6 @@ type Config struct {
 	ExperimentalUpgrade          bool          `yaml:"experimental-upgrade"`
 	ExperimentalUpgradeAudit     bool          `yaml:"experimental-upgrade-audit"`
 	EnableH2cServer              bool          `yaml:"enable-h2c-server"`
-	EnableH2cBackends            bool          `yaml:"enable-h2c-backends"`
 	ReadTimeoutServer            time.Duration `yaml:"read-timeout-server"`
 	ReadHeaderTimeoutServer      time.Duration `yaml:"read-header-timeout-server"`
 	WriteTimeoutServer           time.Duration `yaml:"write-timeout-server"`
@@ -694,8 +693,7 @@ func NewConfig() *Config {
 	flag.DurationVar(&cfg.BackendFlushInterval, "backend-flush-interval", 20*time.Millisecond, "flush interval for upgraded proxy connections")
 	flag.BoolVar(&cfg.ExperimentalUpgrade, "experimental-upgrade", false, "enable experimental feature to handle upgrade protocol requests")
 	flag.BoolVar(&cfg.ExperimentalUpgradeAudit, "experimental-upgrade-audit", false, "enable audit logging of the request line and the messages during the experimental web socket upgrades")
-	flag.BoolVar(&cfg.EnableH2cServer, "enable-h2c-server", false, "enable h2c (HTTP/2 cleartext) on the incoming listener; no effect when TLS is configured")
-	flag.BoolVar(&cfg.EnableH2cBackends, "enable-h2c-backends", false, "enable h2c (HTTP/2 cleartext) for outgoing connections to backends; backends must speak h2c")
+	flag.BoolVar(&cfg.EnableH2cServer, "enable-h2c-server", false, "enable h2c (HTTP/2 cleartext) on the incoming listener")
 	flag.DurationVar(&cfg.ReadTimeoutServer, "read-timeout-server", 5*time.Minute, "set ReadTimeout for http server connections")
 	flag.DurationVar(&cfg.ReadHeaderTimeoutServer, "read-header-timeout-server", 60*time.Second, "set ReadHeaderTimeout for http server connections")
 	flag.DurationVar(&cfg.WriteTimeoutServer, "write-timeout-server", 60*time.Second, "set WriteTimeout for http server connections")
@@ -1158,7 +1156,6 @@ func (c *Config) ToOptions() skipper.Options {
 		ExperimentalUpgrade:          c.ExperimentalUpgrade,
 		ExperimentalUpgradeAudit:     c.ExperimentalUpgradeAudit,
 		EnableH2cServer:              c.EnableH2cServer,
-		EnableH2cBackends:            c.EnableH2cBackends,
 		ReadTimeoutServer:            c.ReadTimeoutServer,
 		ReadHeaderTimeoutServer:      c.ReadHeaderTimeoutServer,
 		WriteTimeoutServer:           c.WriteTimeoutServer,
--- a/config/config.go
+++ b/config/config.go
@@ -298,6 +298,8 @@ type Config struct {
 	BackendFlushInterval         time.Duration `yaml:"backend-flush-interval"`
 	ExperimentalUpgrade          bool          `yaml:"experimental-upgrade"`
 	ExperimentalUpgradeAudit     bool          `yaml:"experimental-upgrade-audit"`
+	EnableH2cServer              bool          `yaml:"enable-h2c-server"`
+	EnableH2cBackends            bool          `yaml:"enable-h2c-backends"`
 	ReadTimeoutServer            time.Duration `yaml:"read-timeout-server"`
 	ReadHeaderTimeoutServer      time.Duration `yaml:"read-header-timeout-server"`
 	WriteTimeoutServer           time.Duration `yaml:"write-timeout-server"`
@@ -692,6 +694,8 @@ func NewConfig() *Config {
 	flag.DurationVar(&cfg.BackendFlushInterval, "backend-flush-interval", 20*time.Millisecond, "flush interval for upgraded proxy connections")
 	flag.BoolVar(&cfg.ExperimentalUpgrade, "experimental-upgrade", false, "enable experimental feature to handle upgrade protocol requests")
 	flag.BoolVar(&cfg.ExperimentalUpgradeAudit, "experimental-upgrade-audit", false, "enable audit logging of the request line and the messages during the experimental web socket upgrades")
+	flag.BoolVar(&cfg.EnableH2cServer, "enable-h2c-server", false, "enable h2c (HTTP/2 cleartext) on the incoming listener; no effect when TLS is configured")
+	flag.BoolVar(&cfg.EnableH2cBackends, "enable-h2c-backends", false, "enable h2c (HTTP/2 cleartext) for outgoing connections to backends; backends must speak h2c")
 	flag.DurationVar(&cfg.ReadTimeoutServer, "read-timeout-server", 5*time.Minute, "set ReadTimeout for http server connections")
 	flag.DurationVar(&cfg.ReadHeaderTimeoutServer, "read-header-timeout-server", 60*time.Second, "set ReadHeaderTimeout for http server connections")
 	flag.DurationVar(&cfg.WriteTimeoutServer, "write-timeout-server", 60*time.Second, "set WriteTimeout for http server connections")
@@ -1153,6 +1157,8 @@ func (c *Config) ToOptions() skipper.Options {
 		BackendFlushInterval:         c.BackendFlushInterval,
 		ExperimentalUpgrade:          c.ExperimentalUpgrade,
 		ExperimentalUpgradeAudit:     c.ExperimentalUpgradeAudit,
+		EnableH2cServer:              c.EnableH2cServer,
+		EnableH2cBackends:            c.EnableH2cBackends,
 		ReadTimeoutServer:            c.ReadTimeoutServer,
 		ReadHeaderTimeoutServer:      c.ReadHeaderTimeoutServer,
 		WriteTimeoutServer:           c.WriteTimeoutServer,
--- a/dataclients/kubernetes/endpointslices.go
+++ b/dataclients/kubernetes/endpointslices.go
@@ -4,7 +4,10 @@ import (
 	"github.com/zalando/skipper/dataclients/kubernetes/definitions"
 )
 
-const endpointSliceServiceNameLabel = "kubernetes.io/service-name"
+const (
+	endpointSliceServiceNameLabel = "kubernetes.io/service-name"
+	appProtocolH2C                = "kubernetes.io/h2c"
+)
 
 // There are [1..N] Kubernetes endpointslices created for a single Kubernetes service.
 // Kubernetes endpointslices of a given service can have duplicates with different states.
@@ -22,8 +25,7 @@ type skipperEndpoint struct {
 	Zone    string
 }
 
-func (eps *skipperEndpointSlice) getPort(protocol, pName string, pValue int) int {
-	var port int
+func (eps *skipperEndpointSlice) getEndpointSlicePort(protocol, pName string, pValue int) *endpointSlicePort {
 	for _, p := range eps.Ports {
 		if protocol != "" && p.Protocol != protocol {
 			continue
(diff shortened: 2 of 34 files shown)


skipper.go
```diff
@@ -1585,6 +1593,15 @@ func listenAndServeQuit(
 		ErrorLog:          newServerErrorLog(),
 	}
 
+	if o.EnableH2cServer && !serveTLS {
```
molikuner: By not allowing `UnencryptedHTTP2` to be set to `true` when `serveTLS` is enabled, it's impossible to both serve H2C and HTTPS traffic using the same skipper instance (as the `srv` instance is shared between both listeners). To my knowledge it's totally fine to set `UnencryptedHTTP2` to `true` for a TLS listener. It will just be ignored.

config/config.go
```diff
@@ -681,6 +683,8 @@ func NewConfig() *Config {
 	flag.DurationVar(&cfg.BackendFlushInterval, "backend-flush-interval", 20*time.Millisecond, "flush interval for upgraded proxy connections")
 	flag.BoolVar(&cfg.ExperimentalUpgrade, "experimental-upgrade", false, "enable experimental feature to handle upgrade protocol requests")
 	flag.BoolVar(&cfg.ExperimentalUpgradeAudit, "experimental-upgrade-audit", false, "enable audit logging of the request line and the messages during the experimental web socket upgrades")
+	flag.BoolVar(&cfg.EnableH2cServer, "enable-h2c-server", false, "enable h2c (HTTP/2 cleartext) on the incoming listener; no effect when TLS is configured")
+	flag.BoolVar(&cfg.EnableH2cBackends, "enable-h2c-backends", false, "enable h2c (HTTP/2 cleartext) for outgoing connections to backends; backends must speak h2c")
```
molikuner: These two new "enable" flags feel a bit odd to me. One forces h2c (outgoing) and the other only allows h2c additionally (incoming). I think it would make sense to at least make this more clear in the description, but ideally even some different name of the option would make it clear. For example one could be `EnableH2cServer` while the other could be `ForceH2cBackends`. wdyt?

proxy/proxy.go
molikuner: Do we need to find a similar solution to DNS changes for the new `h2cTr`?

proxy/proxy.go
```diff
@@ -883,6 +888,31 @@ func WithParams(p Params) *Proxy {
...
 	}
 
+	if p.EnableH2cBackends {
+		if tr.Protocols == nil {
+			tr.Protocols = new(http.Protocols)
+		}
+		tr.Protocols.SetUnencryptedHTTP2(true)
+	}
+
+	h2cTr := &http.Transport{
```
molikuner: Instead of creating a fully new Transport here, did you consider to just clone the "normal" Transport instance? That could prevent configuration drifts, apart from the ones that we want to set explicitly, i.e. the `Protocols`.

skipper_h2c_test.go
```diff
@@ -0,0 +1,154 @@
...
+	serveTLS := tlsCfg != nil
+
+	srv := &http.Server{}
+	if o.EnableH2cServer && !serveTLS {
+		if srv.Protocols == nil {
+			srv.Protocols = new(http.Protocols)
+		}
+		srv.Protocols.SetHTTP1(true)
+		srv.Protocols.SetUnencryptedHTTP2(true)
+	}
```
molikuner: I don't quite get the usefulness of these tests. This is the exact logic we want to test, just hard-coded into the test and not taking anything from the actual code, or not?

molikuner (commented):
Thanks for looking into this already! I left a few comments for things that I noticed. Feel free to ignore, if you don't think they provide value.

skipper.go
szuecs: Yeah it's the first code generation, I have to check better.
My thought was that we split the listeners, the current condition is for sure not great.

config/config.go
szuecs: Yes I think we should completely drop the `-enable-h2c-backends`. It's basically a relict of the steps for generating the code.
We basically only need h2c:// to decide the protocol

proxy/proxy.go
szuecs: I don't know, something to test, likely yes.

proxy/proxy.go
szuecs: My idea was for a while to have a data structure to have N Transports instead of 1. See also https://github.com/zalando/skipper/issues/3844. The problem is with high traffic, low latency the throughput is likely limited by a lock contention on the transport as far as we understand.
The problem with https://github.com/zalando/skipper/issues/3844 is that we can not have a map, because we would need a lock and we will move lock contention from stdlib to the map. A fixed size array could be a solution but then we need to be smart if we want to support a split of transport configurations. We could either use an array per protocol or we use one and split it into sections.

What is your suggestion on the Transport and use of `Protocols`?

My idea is to support h2c:// and http:// and https:// and maybe h2:// (and h3:// whenever it#s ready) so basically being explicit about the protocol.
As far as I understand we can not do (and likely do not want) to support HTTP/1.1 to H2c upgrade, so a connection pool per protocol seems to be not a bad choice.

To omit config drift we can just extract the transport creation into a func and pass the protocols as parameters and then we have no drift.

skipper_h2c_test.go
szuecs: yeah some of claude's unit test generation is not very useful and to call our normal bdd style tests e2e is also funny.

skipper.go
szuecs: So I also read the Go code and you are right and then I also read the skipper code and we already have the split so I just dropped the extra condition as you commented.

config/config.go

proxy/proxy.go

proxy/proxy.go
szuecs: config drift fixed

szuecs: Go's http.Transport actually has `TLSNextProto map[string]func(authority string, c *tls.Conn) RoundTripper` to lookup the RoundTripper from a map, so the idea with an array to split multiple is not so bad to avoid lock contention on the transport.

config/config.go
```diff
@@ -291,6 +291,7 @@ type Config struct {
 	BackendFlushInterval         time.Duration `yaml:"backend-flush-interval"`
 	ExperimentalUpgrade          bool          `yaml:"experimental-upgrade"`
 	ExperimentalUpgradeAudit     bool          `yaml:"experimental-upgrade-audit"`
+	EnableH2cServer              bool          `yaml:"enable-h2c-server"`
```
molikuner, line 301: Minor: I'm not sure how useful this flag is. The server only supports h2c with prior knowledge, so at the moment there are no h2c requests targeting skipper. Or do you think there would be clients relying on the fact that these calls would currently fail and thus it would be not backwards compatible to just start supporting it?

docs/operation/operation.md
```diff
@@ -89,6 +89,29 @@ implementation of DialContext, which is the TCP connection pool used in the
...
+
+Enable h2c on the listener (for clients that speak h2c to skipper):
+
+    -enable-h2c-server
+        enable h2c (HTTP/2 cleartext) on the incoming listener; no effect when TLS is configured
+
+Enable h2c on outgoing backend connections (for backends that speak h2c):
+
+    -enable-h2c-backends
+        enable h2c (HTTP/2 cleartext) for outgoing connections to backends; backends must speak h2c
```
molikuner: This was removed

proxy/proxy.go
```diff
@@ -814,41 +815,12 @@ func WithParams(p Params) *Proxy {
...
-		IdleConnTimeout:       p.CloseIdleConnsPeriod,
-		DisableKeepAlives:     p.DisableHTTPKeepalives,
-		Proxy:                 proxyFromContext,
-	}
+	tr := newTransport(p)
 
 	quit := make(chan struct{})
 	// We need this to reliably fade on DNS change, which is right
 	// now not fixed with IdleConnTimeout in the http.Transport.
 	// https://github.com/golang/go/issues/23427
```
molikuner: This comment should be moved to the lines below where this pro-active disconnect is implemented now.

proxy/proxy.go
molikuner: > The problem with https://github.com/zalando/skipper/issues/3844 is that we can not have a map, because we would need a lock and we will move lock contention from stdlib to the map.

I saw some lock-free map implementations recently. I've not checked how they perform, but that could also be an option to benchmark. 

> What is your suggestion on the Transport and use of Protocols?
> My idea is to support h2c:// and http:// and https:// and maybe h2:// (and h3:// whenever it#s ready) so basically being explicit about the protocol.
As far as I understand we can not do (and likely do not want) to support HTTP/1.1 to H2c upgrade, so a connection pool per protocol seems to be not a bad choice.

Yes, it's definitely something that could work. But I'm not sure how much benefit it provides. For us 99.99% of the traffic us currently `http`. Some might switch to `https`, but even then it would only be distributed to two different Transport instances. This is likely not going to make a huge difference for us, as we don't have much traffic with different protocols happening at the moment.

Would `https://` only support HTTP/1.1 over TLS? This is a bit different from how other tools use the protocol. My expectation would rather be that `https` tries to use the best protocol available, while `h2://` forces HTTP/2. Given that you can also just enable `HTTP1` and `HTTP2` for the same `Transport` and `HTTP2` is preferred if available, this would also follow my expectations. This of course makes it a bit more complex or even impossible to split the Transport instances by protocol.

szuecs: > > The problem with https://github.com/zalando/skipper/issues/3844 is that we can not have a map, because we would need a lock and we will move lock contention from stdlib to the map.
> 
> I saw some lock-free map implementations recently. I've not checked how they perform, but that could also be an option to benchmark. 

Yeah there's sync.Map but I want to have basically an array and do transports[hash(targetHost)] to lookup the transport of a target. It doesn't fit into the current pr, but that's what I want to do in the future but maybe we have to create a flag to set the size of the array.
I think there are also cases that don't need that kind of throughput optimization.

> > What is your suggestion on the Transport and use of Protocols?
> 
> > My idea is to support h2c:// and http:// and https:// and maybe h2:// (and h3:// whenever it#s ready) so basically being explicit about the protocol.
> 
> As far as I understand we can not do (and likely do not want) to support HTTP/1.1 to H2c upgrade, so a connection pool per protocol seems to be not a bad choice.
> 
> Yes, it's definitely something that could work. But I'm not sure how much benefit it provides. For us 99.99% of the traffic us currently `http`. Some might switch to `https`, but even then it would only be distributed to two different Transport instances. This is likely not going to make a huge difference for us, as we don't have much traffic with different protocols happening at the moment.

Yes that's also true in pur case, but we have to support the vision of mixed workloads and large migrations made easy from mixed workloads point of view.

> 
> Would `https://` only support HTTP/1.1 over TLS? This is a bit different from how other tools use the protocol. My expectation would rather be that `https` tries to use the best protocol available, while `h2://` forces HTTP/2. Given that you can also just enable `HTTP1` and `HTTP2` for the same `Transport` and `HTTP2` is preferred if available, this would also follow my expectations. This of course makes it a bit more complex or even impossible to split the Transport instances by protocol.

I am not sure but I think to have control over protocol from the backend owner point of view to set https:// or h2c:// or h2:// is actually a feature.
Keep in mind that HTTP/1.1 in Go is 4x more efficient based on fortinet's benchmarks. Do a general upgrade to h2 might not be what you want.

config/config.go
szuecs: A protocol change could be a huge change, so better to guard.

docs/operation/operation.md
szuecs: thanks dropped

proxy/proxy.go
szuecs: thanks moved

szuecs:
Current state shows that h2c is slower than http/1.1, but I think that's what we get from Go stdlib, so expected.

```
% echo "GET http://localhost:9090/http" | vegeta attack  -rate 100/s -duration 1m | vegeta report -type "hist[0,500µs,1ms,2ms,10ms,100ms]"
Bucket           #     %       Histogram
[0s,     500µs]  0     0.00%   
[500µs,  1ms]    482   8.03%   ######
[1ms,    2ms]    5495  91.58%  ####################################################################
[2ms,    10ms]   23    0.38%   
[10ms,   100ms]  0     0.00%   
[100ms,  +Inf]   0     0.00%   

% echo "GET http://localhost:9090/h2c" | vegeta attack -h2c -rate 100/s -duration 1m | vegeta report -type "hist[0,500µs,1ms,2ms,10ms,100ms]"
Bucket           #     %       Histogram
[0s,     500µs]  0     0.00%   
[500µs,  1ms]    0     0.00%   
[1ms,    2ms]    5519  91.98%  ####################################################################
[2ms,    10ms]   481   8.02%   ######
[10ms,   100ms]  0     0.00%   
[100ms,  +Inf]   0     0.00%   
```

proxy/proxy.go
molikuner: > I am not sure but I think to have control over protocol from the backend owner point of view to set https:// or h2c:// or h2:// is actually a feature.
> Keep in mind that HTTP/1.1 in Go is 4x more efficient based on fortinet's benchmarks. Do a general upgrade to h2 might not be what you want.

Ah, thanks for sharing. I wasn't aware of these efficiency differences. Lets go with manual protocol selection for now. If in future there is a need to have automatic protocol selection it could still be implemented.

szuecs: not quite so slow as I remembered, but https://github.com/zalando/skipper/pull/4171#issuecomment-5240728900 shows it and I see that you saw it :)

dataclients/kubernetes/endpointslices.go
```diff
@@ -32,30 +34,44 @@ func (eps *skipperEndpointSlice) getPort(protocol, pName string, pValue int) int
 		// Optional if only one ServicePort is defined on this service.
 		// Therefore empty name match is fine.
 		if p.Name == pName {
-			port = p.Port
-			break
+			return p
```
a4180p, line 46: nit: imo, it was better before)

a4180p (approved):

dataclients/kubernetes/endpointslices.go
szuecs: You are right, I will check if I can do this easily

szuecs: done

a4180p:
👍

szuecs:
:+1:

PR #4172, opened 2026-08-06
fix: proxy listener on insecure address

szuecs:
fix: if we run tls listener, the insecure listener would never be able to have the haproxy2 proxy protocol support


a4180p:
👍

szuecs:
:+1:

PR #4177, opened 2026-08-10
feat(kubernetes): add flag to preserve Host header for ExternalName ingress backends

### LINKED ISSUE #3343: HTTP header HOST always modified when externalName is used in service backend
**Describe the bug**
I observed strange behavior when externalName type is used in the service backend, the HTTP header HOST of request is **always modified** to the value of externalName.

**To Reproduce**

in namespace NS-1:
ingress
-- Host: foo.com
-- backend: ingressgateway

svc
-- name: ingressgateway
-- externalName: bar.com

When the traffic of foo.com comes, the data flow looks like below:
ingress:foo.com(NS-1) -> svc:ingressgateway(NS-1) -> bar.com

However the request to bar.com, whose HTTP header HOST is **always modified** to **bar.com**, I actually expected the HTTP header HOST is still **foo.com**.

**Expected behavior**
The HTTP header HOST in the request to externalName backend should be kept.

**Observed behavior**
The HTTP header HOST in the request to externalName backend is always modified to the value of externalName.

I checked the documentation, but I can't find any configuration to control this behavior to keep the HOST value or change it to the destination.

pujitha24:
Motivation:
For Ingress backends that are Kubernetes Services of type ExternalName,
skipper always injects a setRequestHeader("Host", "<externalName>")
filter into the generated route. Since this filter runs in the request
filter chain, it unconditionally overwrites the outgoing Host header,
even when the proxy-preserve-host flag is set or a preserveHost()
filter is attached to the route. Users who route internal cluster
traffic through an ExternalName service (e.g. to reference a Service
in another namespace) have no way to keep the original incoming Host
header.

Approach:
Add a new option, -kubernetes-external-name-preserve-host, threaded
through config.Config -> skipper.Options -> kubernetes.Options ->
the ingress data client. When enabled, externalNameRoute() in
dataclients/kubernetes/ingress.go no longer attaches the
setRequestHeader("Host", ...) filter, so the outgoing Host header is
then governed by skipper's normal mechanisms: the global
-proxy-preserve-host flag and/or a preserveHost() route filter. The
new option defaults to false, so existing behavior (Host always
overwritten to the external name) is unchanged unless explicitly
opted in.

Note the new flag alone does not change the resulting Host header:
since -proxy-preserve-host also defaults to false, users must combine
this flag with -proxy-preserve-host or a preserveHost() filter (e.g.
added via the zalando.org/skipper-filter annotation) to actually keep
the original Host. This is intentional per the flag's name: it only
removes the forced override, matching the approach suggested by a
maintainer on the linked report.

Validation:
Ran `go build . ./config/... ./dataclients/kubernetes/...` (passes)
and `go test ./dataclients/kubernetes/...` (passes). Added a new
fixture pair under
dataclients/kubernetes/testdata/ingressV1/ingress-data/
(ing-with-externalname-svc-preserve-host.{yaml,kube,eskip}) exercised
by the existing TestIngressV1Fixtures harness. Confirmed the fixture
is a genuine regression test: temporarily reverting externalNameRoute
to always inject the Host filter makes
TestIngressV1Fixtures/ing-with-externalname-svc-preserve-host fail
with a diff showing the unwanted setRequestHeader filter; restoring
the fix makes it pass again. Did not run a live Kubernetes cluster or
end-to-end reproduction; the repo's CONTRIBUTING.md does not require
one for this kind of change.

Report: https://github.com/zalando/skipper/issues/3343
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

Fixes #3343

### DIFF
--- a/dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-externalname-svc-preserve-host.kube
+++ b/dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-externalname-svc-preserve-host.kube
@@ -1,0 +1,2 @@
+enableExternalNames: true
+externalNamePreserveHost: true
--- a/config/config.go
+++ b/config/config.go
@@ -199,6 +199,7 @@ type Config struct {
 	KubernetesAnnotationFiltersAppend                    []kubernetes.AnnotationFilters     `yaml:"-"`
 	KubernetesEastWestRangePredicates                    []*eskip.Predicate                 `yaml:"-"`
 	EnableKubernetesExternalNames                        bool                               `yaml:"enable-kubernetes-external-names"`
+	KubernetesExternalNamePreserveHost                   bool                               `yaml:"kubernetes-external-name-preserve-host"`
 	KubernetesOnlyAllowedExternalNames                   bool                               `yaml:"kubernetes-only-allowed-external-names"`
 	KubernetesAllowedExternalNames                       regexpListFlag                     `yaml:"kubernetes-allowed-external-names"`
 	KubernetesRedisServiceNamespace                      string                             `yaml:"kubernetes-redis-service-namespace"`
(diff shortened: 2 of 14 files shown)


dataclients/kubernetes/testdata/ingressV1/ingress-data/ing-with-externalname-svc-preserve-host.kube
```diff
@@ -0,0 +1,2 @@
+enableExternalNames: true
+externalNamePreserveHost: true
```
szuecs, line 2: can you add test cases for all kind of combinations of these two configurations, so we are explicitly testing the full matrix?
Thanks!

szuecs (commented):
Please add the 3 other test cases, other than that lgtm

pujitha24:
Added the remaining two combinations (`enableExternalNames: false` with `externalNamePreserveHost` unset and set to `true`) as `ing-with-externalname-svc-disabled` and `ing-with-externalname-svc-disabled-preserve-host` in the same `ingress-data` dir. Both expect zero routes, since `enableExternalNames` gates the whole ExternalName conversion path regardless of `preserveHost` — that matches the existing `not-allowed` fixture's empty-eskip pattern. Combined with the two fixtures already in the PR, that covers all four combinations.

szuecs:
:+1:

MustafaSaber (approved):

Issue #4073, opened 2026-06-17
mTLS filters support globbing or regexp


szuecs:
Right now all `mtls*` filters support only exact match, but for SAN fields like DNS or or URI it makes sense to allow either regexp or at least globbing.

[example](https://github.com/zalando/skipper/issues/1780#issuecomment-1938336694):  `mtlsSanDNS("*.host.example")` or `mtlsSanURI("spiffe://services.example/applications/*")`

another example in https://github.com/zalando/skipper/pull/4044/changes#r3422146715

PR #4178, opened 2026-08-10
feature: support globbing in mtls filters

szuecs:
ref: https://github.com/zalando/skipper/issues/4073


filters/tls/mtls.go
```diff
@@ -184,6 +187,17 @@ func isValidHostname(s string) bool {
...
 }
 
+// matchesDNSWildcard reports whether the lowercased hostname name matches the
+// wildcard pattern "*.suffix". It requires exactly one non-empty label before
+// suffix, matching RFC 6125 single-label wildcard semantics.
+func matchesDNSWildcard(name, suffix string) bool {
+	if !strings.HasSuffix(name, "."+suffix) {
+		return false
+	}
+	label := name[:len(name)-len(suffix)-1]
```
MustafaSaber, line 197: you can use [`strings.CutSuffix`](https://pkg.go.dev/strings#CutSuffix)

filters/tls/mtls.go
```diff
@@ -296,6 +314,12 @@ func (ms *mtlsSpec) CreateFilter(args []any) (filters.Filter, error) {
 			}
 			if u, err := url.Parse(s); err != nil || u.Scheme == "" {
 				return nil, filters.ErrInvalidFilterParameters
+			}
+			if strings.ContainsAny(s, "*?[") {
```
shyamz-22: what happens to the query matching? will spiffe://services.example/application?role=admin and spiffe://services.example/applicationXrole=admin match? In the latter case it will match to /applicationXrole=admin, is this something we want? We can properly distinquish between URI syntax and glob syntax.

filters/tls/mtls.go
```diff
@@ -389,20 +413,48 @@ func (mf *mtlsFilter) Request(ctx filters.FilterContext) {
...
+			if _, ok := mf.allowedHostnames[lower]; ok {
 				allowed = true
 				auditCertData.WriteString("SAN DNS: ")
 				auditCertData.WriteString(dns)
+				break
+			}
+			// Wildcard match: pattern suffix is stored without the leading "*.".
+			// A name matches only when it has exactly one label before the suffix.
+			for _, suffix := range mf.allowedHostnameGlobs {
+				if matchesDNSWildcard(lower, suffix) {
```
shyamz-22, line 426: the allowedHostNameGlobs with *.host.example will match foo.host.com but not foo.bar.host.com  , so is the comment above correct? it does not use path.Match semantics.
allowedHostnameGlobs []string            // lowercased glob patterns (pre-validated, path.Match semantics)

shyamz-22 (commented):
Additionally I miss documentation changes that show glob support, will this be another PR?

szuecs:
> Additionally I miss documentation changes that show glob support, will this be another PR?

?
User docs are changed: `docs/reference/filters.md` ...

filters/tls/mtls.go
szuecs: Not exactly, see CutSuffix and here we have -1 in the index to get the label:
```
func CutSuffix(s, suffix string) (before string, found bool) {
	if !HasSuffix(s, suffix) {
		return s, false
	}
	return s[:len(s)-len(suffix)], true
}
```

filters/tls/mtls.go
szuecs: What would be the globbing config you would have to match or not match?
Do you think we should only glob match the path or not?

szuecs: I think the current way seems fine. Why do you think it might be wrong?

szuecs: I added some testcases to document the behaviour as far as I understood.
Not sure why `spiffe://services.example/applicationXrole=admin` and `spiffe://services.example/application?role=admin` should not match `spiffe://services.example/*`

filters/tls/mtls.go
szuecs: Yes it matches "DNS style", one label, so `*.example` will match `foo.example`, but not `foo.bar.example`
Documentation seems correct to the intended behaviour and implementation seems to work as test coverage shows this behaviour, too.

filters/tls/mtls.go
shyamz-22: ? is a valid glob wildcard, but it is also the URI query separator. This code treats every URI containing a query as a glob. For example, an existing exact pattern spiffe://services.example/application?role=admin would no longer be matched literally because path.Match interprets ? as “any one character.” This may unintentionally broaden authorization and changes existing behavior. Can we restrict globbing to * is what i meant by properly distinquishing between the URI and glob syntax but that might be limiting. Example mtlsSanURI("spiffe://services.example/application?role=admin") will now match spiffe://services.example/applicationXrole=admin also. Do we want this is my question.

shyamz-22:
> > Additionally I miss documentation changes that show glob support, will this be another PR?
> 
> ? User docs are changed: `docs/reference/filters.md` ...

please ignore, for whatever reason the review showed me only mtls filter code and not the documentation, i did not check properly before commenting.

docs/reference/filters.md
```diff
@@ -653,8 +653,9 @@ Example:
 
 ### mtlsSanDNS
 
-This authz filter checks DNS of the SAN value of the provided certificate. You have
-to use `mtlsAuthn()` to verify validity.
+This authz filter checks DNS of the SAN value of the provided
+certificate. It supports exact match and globbing.  You have to use
```
shyamz-22: `It supports exact match and globbing`, but [matchDNSWildCard](https://github.com/zalando/skipper/blob/2f65b9155c3e51a51add43a45011cc5c96ea32f1/filters/tls/mtls.go#L193-L199) does a single-label wildcard semantics with .* meaning `mtlsSanDNS("*.example.com")`  works but `mtlsSanDNS("app-*.example.com")`  will not work as it will be even rejected before iiuc as invalid hostname.

Can we update the doc to be explicit about the type of wild card match we support?

filters/tls/mtls.go
shyamz-22: In [L:99](https://github.com/zalando/skipper/blob/2f65b9155c3e51a51add43a45011cc5c96ea32f1/filters/tls/mtls.go#L99) its said that we support path.Match semantics, however [L:193:199](https://github.com/zalando/skipper/blob/2f65b9155c3e51a51add43a45011cc5c96ea32f1/filters/tls/mtls.go#L193-L199) we do not use that, its more like a confusion following the comments and actual code, does only leading wildcard that matches one DNS label. 

I tried to make the connection in this comment also https://github.com/zalando/skipper/pull/4178/changes#r3764316282

filters/tls/mtls.go
szuecs: Ah ok, yeah of course makes sense

filters/tls/mtls.go
szuecs: Ok I think I fixed it

filters/tls/mtls.go
szuecs: This is fixed

docs/reference/filters.md
szuecs: done

szuecs:
:+1:

shyamz-22:
👍🏽

PR #4183, opened 2026-08-12
fix: skipper startup panic caused by valkey ring not initialized 

szuecs:
fix: valkey startup in case of remote url or kubernetes dataclient. The updater makes now sure that we get non zero shards, such that the ring client is successfully initialized

### DIFF
--- a/dataclients/kubernetes/kubernetestest/api.go
+++ b/dataclients/kubernetes/kubernetestest/api.go
@@ -41,6 +41,10 @@ type api struct {
 	resourceList []byte
 }
 
+type KubeTestAPI interface {
+	UpdateSpec(specs ...io.Reader) error
+}
+
 func NewAPI(o TestAPIOptions, specs ...io.Reader) (*api, error) {
 	a := &api{
 		namespaces: make(map[string]namespace),
@@ -65,6 +69,12 @@ func NewAPI(o TestAPIOptions, specs ...io.Reader) (*api, error) {
 
 	a.resourceList = clrb
 
+	err = a.update(specs...)
+
+	return a, err
+}
+
+func (a *api) update(specs ...io.Reader) error {
 	namespaces := make(map[string]map[string][]interface{})
 	all := make(map[string][]interface{})
 
@@ -106,50 +116,54 @@ func NewAPI(o TestAPIOptions, specs ...io.Reader) (*api, error) {
 			if err := d.Decode(&o); err == io.EOF || err == nil && len(o) == 0 {
 				break
 			} else if err != nil {
-				return nil, err
+				return err
 			}
 
 			kind, ok := o["kind"].(string)
 			if !ok {
-				return nil, errInvalidFixture
+				return errInvalidFixture
 			}
 
 			if kind == "List" {
 				items, ok := o["items"].([]interface{})
 				if !ok {
-					return nil, errInvalidFixture
+					return errInvalidFixture
 				}
 				for _, item := range items {
 					o, ok := item.(map[interface{}]interface{})
 					if !ok {
-						return nil, errInvalidFixture
+						return errInvalidFixture
 					}
 					if err := addObject(o); err != nil {
-						return nil, err
+						return err
 					}
 				}
 			} else {
 				if err := addObject(o); err != nil {
-					return nil, err
+					return err
 				}
 			}
 		}
 	}
 
+	var err error
 	for ns, kinds := range namespaces {
-		var err error
 		a.namespaces[ns], err = initNamespace(kinds)
 		if err != nil {
-			return nil, err
+			return err
 		}
 	}
 
 	a.all, err = initNamespace(all)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	return a, nil
+	return nil
+}
+
+func (a *api) UpdateSpec(specs ...io.Reader) error {
+	return a.update(specs...)
 }
 
 func (a *api) ServeHTTP(w http.ResponseWriter, r *http.Request) {
--- a/dataclients/kubernetes/kubernetestest/api.go
+++ b/dataclients/kubernetes/kubernetestest/api.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"regexp"
 	"strings"
+	"sync"
 
 	"gopkg.in/yaml.v2"
 	yaml2 "sigs.k8s.io/yaml"
@@ -33,6 +34,7 @@ type namespace struct {
 }
 
 type api struct {
+	mu           sync.Mutex
 	failOn       map[string]bool
 	findNot      map[string]bool
 	namespaces   map[string]namespace
@@ -75,6 +77,9 @@ func NewAPI(o TestAPIOptions, specs ...io.Reader) (*api, error) {
 }
 
 func (a *api) update(specs ...io.Reader) error {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
 	namespaces := make(map[string]map[string][]interface{})
 	all := make(map[string][]interface{})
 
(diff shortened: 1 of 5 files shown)


skipper.go
```diff
@@ -2153,8 +2153,9 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 			if kdc != nil {
 				kdc.LoadAll()
 				valkeyOptions.AddrUpdater = getKubernetesAddrUpdater(kdc, true, o.KubernetesValkeyServiceNamespace, o.KubernetesValkeyServiceName, o.KubernetesValkeyServicePort)
+				defer kdc.Close()
```
szuecs, line 2156: This was no done before, but it also does not matter because we never return from run().
We have done the defer kdc.Close some lines below in the other branch, so adding it here, too.
Maybe one day we want to be able to "restart" or "reload" with updated fixed config for some reason and then we should close all things that need to be closed or we have leaks.

dataclients/kubernetes/kubernetestest/api.go
```diff
@@ -106,53 +121,55 @@ func NewAPI(o TestAPIOptions, specs ...io.Reader) (*api, error) {
...
 			} else {
 				if err := addObject(o); err != nil {
-					return nil, err
+					return err
 				}
 			}
 		}
 	}
 
+	var err error
```
madumalt, line 154: we can get rid of this var definition here as we return whenever we see an error.

skipper.go
```diff
@@ -2672,6 +2668,31 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
+	N := 12
+	for i := range N {
+		log.Infof("%d attempt -> %d", i, len(res))
+		res, err = valkeyOptions.AddrUpdater()
+		if err != nil {
+			log.Errorf("Failed to update valkey addresses: %v", err)
+			return err
+		}
+		if len(res) > 0 {
+			log.Infof("Initial valkey address update got %d shards after %d attempts", len(res), i)
```
shyamz-22: why not populate the valkeyOptions.Addrs here, and in the NewValkeyRingClient  and avoid a second fetch during startup?

skipper.go
```diff
@@ -2672,6 +2668,31 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 	return listenAndServeQuit(o.CustomHttpHandlerWrap(proxy), &o, sig, idleConnsCH, mtr, cr)
 }
 
+func initialValkeyAddressUpdate(valkeyOptions *skpnet.ValkeyOptions, dc routing.DataClient) error {
+	var (
+		err error
+		res []string
```
shyamz-22: would addrs be a better name for the variable, give the AddrUpdater returns list of addresses

dataclients/kubernetes/kubernetestest/api.go
```diff
@@ -106,53 +121,55 @@ func NewAPI(o TestAPIOptions, specs ...io.Reader) (*api, error) {
...
+}
 
-	return a, nil
+func (a *api) UpdateSpec(specs ...io.Reader) error {
+	return a.update(specs...)
 }
 
 func (a *api) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
```
madumalt, line 172: just curious why do need to accquire the mutex here? What do we want to protect?

skipper.go
```diff
@@ -1724,7 +1724,7 @@ func getKubernetesAddrUpdater(kdc *kubernetes.Client, loaded bool, ns, name stri
 
 func joinPort(addrs []string, port int) []string {
 	p := strconv.Itoa(port)
-	for i := 0; i < len(addrs); i++ {
+	for i := range addrs {
```
madumalt, line 1727: ❤️

skipper.go
```diff
@@ -2672,6 +2668,31 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
+			return err
+		}
+		if len(res) > 0 {
+			log.Infof("Initial valkey address update got %d shards after %d attempts", len(res), i)
+			return nil
+		}
+		if dc != nil {
+			dc.LoadAll() // enforce state update
+		}
+		time.Sleep(valkeyOptions.UpdateInterval + 1)
```
shyamz-22, line 2691: wouldn't sleep be 1 nanosecond if valkeyOptions.UpdateInterval is 0? Do we always ensure its >0

skipper.go
```diff
@@ -2672,6 +2668,31 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
+		if err != nil {
+			log.Errorf("Failed to update valkey addresses: %v", err)
+			return err
+		}
+		if len(res) > 0 {
+			log.Infof("Initial valkey address update got %d shards after %d attempts", len(res), i)
+			return nil
+		}
+		if dc != nil {
+			dc.LoadAll() // enforce state update
```
shyamz-22, line 2689: I don't understand fully the significance of the code, due to lack of my general knowledge, is it ok to ignore the error LoadAll? What happens if it fails?

dataclients/kubernetes/kubernetestest/api.go
szuecs: No I want to omit the reset when we set err (`a.all` and `a.namespaces[ns]` would be weird to read with `:=` even if it would work correctly)

dataclients/kubernetes/kubernetestest/api.go
szuecs: ServeHTTP and UpdateSpec have a data race on the data structures that lead to panic if you use `-race`, so running tests with the race detector enabled.

skipper.go
szuecs: valkeyOptions.Addrs is static and the function fetches the initial list dynamic from somewhere else.
A second fetch is only done if there is 0 shards available.

skipper.go
szuecs: I changed it to `a`, because local variables should be single letter and `addr` has the meaning of in put for creating a net.Listener so I would avoid being close to this name.

skipper.go
szuecs: in config/config.go the default is 10s, but we do not ensure anything for lib users. It also is not an issue here because we run max 12 times the loop. 
I think it's up to the user to set the config by choice.
The +1 was meant to not race with the update interval, so we wait long than that.

skipper.go
szuecs: We call LoadAll here only to run a clusterstate update: https://github.com/zalando/skipper/blob/master/dataclients/kubernetes/kube.go#L448
If it fails we try again up to 12 times in the hope that things are fixed next time, so err does not matter.

skipper.go
a4180p: would it make sence to make sleep time exponential here?
depending on UpdateInterval these 12 retries can be very fast

skipper.go
```diff
@@ -2672,6 +2668,31 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
...
+		a, err = valkeyOptions.AddrUpdater()
+		if err != nil {
+			log.Errorf("Failed to update valkey addresses: %v", err)
+			return err
+		}
+		if len(a) > 0 {
+			log.Infof("Initial valkey address update got %d shards after %d attempts", len(a), i)
+			return nil
+		}
+		if dc != nil {
```
a4180p, line 2688: is that possible that dataclient will be nil?

szuecs: yes because we pass it only in the kubernetes case, because in the other we do not need this.

skipper.go
szuecs: You do not want to have an UpdateInterval for less than 1s if you are not running in tests.
There is in valkey ring client a background goroutine which would run hot without max retries like we have here, so I would say this here is not the issue. :)

a4180p:
👍

szuecs:
:+1:

Issue #3818, opened 2026-01-06
TestBackendTimeoutWithSlowBodyWriterShadow flakytest


szuecs:
https://github.com/zalando/skipper/actions/runs/20748684995/job/59572170101
```
time="2026-01-06T12:56:17Z" level=error msg="error while proxying after 227.133µs, route groupA with backend lb http://127.0.0.1:40839, status code 502: dialing failed true: failed to do backend roundtrip to 127.0.0.1:40839: dial tcp 127.0.0.1:40839: connect: connection refused, remote host: 127.0.0.1, request: \"GET /a HTTP/1.1\", host: 127.0.0.1:41699, user agent: \"Go-http-client/1.1\""
--- FAIL: TestBackendTimeoutWithSlowBodyWriterShadow (1.22s)
    fifo_backendtimeout_with_teeloopback_test.go:94: metric "fifo.main.active": 1
    fifo_backendtimeout_with_teeloopback_test.go:94: metric "fifo.main.queued": 0
    fifo_backendtimeout_with_teeloopback_test.go:94: metric not found: "fifo.main.error.full"
    fifo_backendtimeout_with_teeloopback_test.go:94: metric not found: "fifo.main.error.other"
    fifo_backendtimeout_with_teeloopback_test.go:94: metric not found: "fifo.main.error.timeout"
    fifo_backendtimeout_with_teeloopback_test.go:94: metric "fifo.shadow.active": 1
    fifo_backendtimeout_with_teeloopback_test.go:94: metric "fifo.shadow.queued": 0
    fifo_backendtimeout_with_teeloopback_test.go:94: metric not found: "fifo.shadow.error.full"
    fifo_backendtimeout_with_teeloopback_test.go:94: metric not found: "fifo.shadow.error.other"
    fifo_backendtimeout_with_teeloopback_test.go:94: metric not found: "fifo.shadow.error.timeout"
    fifo_backendtimeout_with_teeloopback_test.go:99: response code: 200
    fifo_backendtimeout_with_teeloopback_test.go:105: Failed to get expected error: timeout waiting for log entry: failed to execute loopback request: dialing failed false: context deadline exceeded
time="2026-01-06T12:56:25Z" level=error msg="Failed to retry backend request: dialing failed true: failed to do backend roundtrip to [::]:40731: dial tcp [::]:40731: connect: connection refused"
```

szuecs:
```
[...]

PR #4188, opened 2026-08-15
feat: add lifoWithBody and lifoGroupWithBody

### LINKED ISSUE #2579: scheduler filter include streaming part

AshrafAhmed9:
Closes #2579.

lifo drops its queue slot as soon as the backend sends response headers. If the backend then streams a long body the slot is already free, so the route lets in more than MaxConcurrency streams. That's the case in the issue.

fifo got this sorted in #2685 by releasing from the proxy once the body is streamed, lifo never did. So this adds lifoWithBody and lifoGroupWithBody the same way: the filter stashes the release func in the state bag, Response() does nothing, and the proxy calls it after serveResponse(). lifo and lifoGroup are untouched.

They use their own state bag key instead of sharing LIFOKey, otherwise a plain lifo() on the same route pops their release func in the response phase.

Two things I wasn't sure about:

- The preprocessor strips all but the last lifo on a route by exact name, so lifoWithBody isn't deduped. Queues are keyed by route id rather than by filter, so two of them on one route end up sharing a queue and each taking a slot. fifoWithBody has the same gap so I left it consistent, but the lifo docs say only the last one applies and that isn't true for the new filter. Happy to add both names to the dedup if you'd rather have that.
- I turned the fifoWithBody block in ServeHTTP into a loop over the three names instead of copying the if twice. Same behaviour for fifoWithBody. Can split it back out if you prefer it explicit.

For tests, TestLifoWithBody runs the same scenario for all four filters: one request that has its headers but is still streaming, then a second one on a route with MaxConcurrency 1 and no queue. lifo and lifoGroup let it through with 200, the WithBody ones reject it with 502 on the queue timeout. So it fails on the current behaviour rather than only checking the new path. Also added spec cases and docs for both.


szuecs:
> Two things I wasn't sure about:
> 
>     * The preprocessor strips all but the last lifo on a route by exact name, so lifoWithBody isn't deduped. Queues are keyed by route id rather than by filter, so two of them on one route end up sharing a queue and each taking a slot. fifoWithBody has the same gap so I left it consistent, but the lifo docs say only the last one applies and that isn't true for the new filter. Happy to add both names to the dedup if you'd rather have that.

We should just have one scheduler filter and we normally use the last one in the route such that users are able to override if we use prepend `-default-filters-prepend=fifo(..)`. These scheduler filters should be set always at start to cover also rest of the filter chain, that's why prepend makes sense.

> 
>     * I turned the fifoWithBody block in ServeHTTP into a loop over the three names instead of copying the if twice. Same behaviour for fifoWithBody. Can split it back out if you prefer it explicit.

I will check, but I think it should be fine.

> For tests, TestLifoWithBody runs the same scenario for all four filters: one request that has its headers but is still streaming, then a second one on a route with MaxConcurrency 1 and no queue. lifo and lifoGroup let it through with 200, the WithBody ones reject it with 502 on the queue timeout. So it fails on the current behaviour rather than only checking the new path. Also added spec cases and docs for both.

sgtm

szuecs (approved):

szuecs:
@AshrafAhmed9 can you check this test case in ./proxy , please:
```
--- FAIL: TestBackendTimeoutWithSlowBodyWriterShadow (1.51s)
    fifo_backendtimeout_with_teeloopback_test.go:327: Failed to get response: Put "http://127.0.0.1:44895/7": net/http: timeout awaiting response headers
    fifo_backendtimeout_with_teeloopback_test.go:327: Failed to get response: Put "http://127.0.0.1:44895/10": net/http: timeout awaiting response headers
    fifo_backendtimeout_with_teeloopback_test.go:327: Failed to get response: Put "http://127.0.0.1:44895/12": net/http: timeout awaiting response headers
    fifo_backendtimeout_with_teeloopback_test.go:242: backend: failed to read body: unexpected EOF
    fifo_backendtimeout_with_teeloopback_test.go:242: backend: failed to read body: unexpected EOF
    fifo_backendtimeout_with_teeloopback_test.go:242: backend: failed to read body: unexpected EOF
    fifo_backendtimeout_with_teeloopback_test.go:94: metric "fifo.main.active": 261
    fifo_backendtimeout_with_teeloopback_test.go:94: metric "fifo.main.queued": 0
    fifo_backendtimeout_with_teeloopback_test.go:94: metric not found: "fifo.main.error.full"
    fifo_backendtimeout_with_teeloopback_test.go:94: metric not found: "fifo.main.error.other"
    fifo_backendtimeout_with_teeloopback_test.go:94: metric not found: "fifo.main.error.timeout"
    fifo_backendtimeout_with_teeloopback_test.go:94: metric "fifo.shadow.active": 5
    fifo_backendtimeout_with_teeloopback_test.go:94: metric "fifo.shadow.queued": 0
    fifo_backendtimeout_with_teeloopback_test.go:94: metric not found: "fifo.shadow.error.full"
    fifo_backendtimeout_with_teeloopback_test.go:94: metric not found: "fifo.shadow.error.other"
    fifo_backendtimeout_with_teeloopback_test.go:94: metric not found: "fifo.shadow.error.timeout"
    fifo_backendtimeout_with_teeloopback_test.go:96: Failed to get the same amount of responses want 500 got 483
    fifo_backendtimeout_with_teeloopback_test.go:99: response code: 200
    fifo_backendtimeout_with_teeloopback_test.go:107: Found "failed to execute loopback request" error log
    fifo_backendtimeout_with_teeloopback_test.go:111: Failed to get expected error: timeout waiting for log entry: context: error while discarding remainder response body
time="2026-08-17T10:18:49Z" level=error msg="Failed to retry backend request: dialing failed true: failed to do backend roundtrip to [::]:36845: dial tcp [::]:36845: connect: connection refused"
```

I did not check it on my own, but if it fails on your machine we should fix it

AshrafAhmed9:
Checked it — fails on master too, so I don't think it's from this PR.

10 runs each, master with nothing applied:

  with -race:     branch 5/10 failed, master 7/10
  without -race:  branch 10/10, master 9/10

My failures don't match CI's though. On CI the metrics are there
(fifo.main.active: 261) and the count comes up short at 483/500. Locally
none of the fifo.* metrics get recorded at all, the main route returns 200,
and it dies waiting for the loopback timeout log entry. So the timeouts the
test expects never happen on my machine — it doesn't look slow enough to
reproduce what CI hits.

#3818 has been tracking this since January with the same connection refused
pattern.

Happy to take the flake in a separate PR, it leans on wall clock timing in a
few places. Would rather not fold it into this one.

szuecs:
@AshrafAhmed9 ok thanks then please ignore this for this PR.

MustafaSaber (approved):

PR #4197, opened 2026-08-19
fix(oidc flow): Referer header could be deleted, use Cookie instead

szuecs:
fix: Referer header could be deleted by privacy enhancements in the browser, so move the protection to use cookies, which are required for session anyways

follow up on https://github.com/zalando/skipper/pull/4195

thanks @kanywst for the hint


MustafaSaber (approved):

szuecs:
:+1:

PR #4199, opened 2026-08-19
Fix: opa startup not ready

szuecs:
Fix: if opa was not ready skipper would just start, but then fail on requests that hit opa filters. This change makes sure that if we want to start OPA it will block until it's ready to serve

### DIFF
--- a/filters/openpolicyagent/openpolicyagent.go
+++ b/filters/openpolicyagent/openpolicyagent.go
@@ -340,6 +340,7 @@ func NewOpenPolicyAgentRegistry(opts ...func(*OpenPolicyAgentRegistry) error) (*
 	go registry.startCleanerDaemon()
 
 	if registry.enableCustomControlLoop {
+		registry.startAndTriggerOpaPlugins()
 		go registry.startCustomControlLoopDaemon()
 	}
 
@@ -498,22 +499,26 @@ func (registry *OpenPolicyAgentRegistry) startCustomControlLoopDaemon() {
 		case <-registry.quit:
 			return
 		case <-ticker.C:
-			registry.mu.Lock()
-			instances := slices.Collect(maps.Values(registry.instances))
-			registry.mu.Unlock()
-
-			for _, opa := range instances {
-				if opa != nil && !opa.Started() { // Skip retriggering plugins of unstarted instances
-					continue
-				}
-				func() {
-					ctx, cancel := context.WithTimeout(context.Background(), registry.instanceStartupTimeout)
-					defer cancel()
-					opa.triggerPlugins(ctx)
-				}()
-			}
+			registry.startAndTriggerOpaPlugins()
 			ticker.Reset(registry.controlLoopIntervalWithJitter())
 		}
+	}
+}
+
+func (registry *OpenPolicyAgentRegistry) startAndTriggerOpaPlugins() {
+	registry.mu.Lock()
+	instances := slices.Collect(maps.Values(registry.instances))
+	registry.mu.Unlock()
+
+	for _, opa := range instances {
+		if opa != nil && !opa.Started() { // Skip retriggering plugins of unstarted instances
+			continue
+		}
+		func() {
+			ctx, cancel := context.WithTimeout(context.Background(), registry.instanceStartupTimeout)
+			defer cancel()
+			opa.triggerPlugins(ctx)
+		}()
 	}
 }
 
--- a/skipper_opa_test.go
+++ b/skipper_opa_test.go
@@ -1,0 +1,204 @@
+package skipper
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+	"path/filepath"
+	"syscall"
+	"testing"
+	"time"
+
+	opasdktest "github.com/open-policy-agent/opa/v1/sdk/test"
+)
+
+const opaTestBundleName = "test"
+
+func writeOpaConfigFile(t *testing.T, serviceURL string) string {
+	t.Helper()
+	config := fmt.Sprintf(`{
+	"services": {
+		"test": {
+			"url": %q
+		}
+	},
+	"bundles": {
+		"test": {
+			"resource": "/bundles/{{ .bundlename }}"
+		}
+	},
+	"plugins": {
+		"envoy_ext_authz_grpc": {
+			"path": "envoy/authz/allow",
+			"dry-run": false,
+			"skip-request-body-parse": false
+		}
+	}
+}`, serviceURL)
+
+	dir := t.TempDir()
+	path := filepath.Join(dir, "opa-config.json")
+	if err := os.WriteFile(path, []byte(config), 0600); err != nil {
+		t.Fatalf("failed to write OPA config file: %v", err)
+	}
+	return path
+}
+
+// TestOpaStartup verifies that the full skipper proxy starts up correctly
+// with OPA enabled, for both the custom control loop and default loop modes.
+// The custom control loop mode exercises the startup fix: startAndTriggerOpaPlugins()
+// is now called synchronously at registry creation time.
+func TestOpaStartup(t *testing.T) {
+	tests := []struct {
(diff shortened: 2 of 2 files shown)


MustafaSaber:
I think this was done intentionally to avoid slow starts, no?

szuecs:
> I think this was done intentionally to avoid slow starts, no?

Not sure, I just remember that the idea was that we are decoupled from routing updates and that we are.
We should always ensure that we have everything to serve traffic. If OPA is configured why should we start if we can not serve OPA?
Makes no sense to me.

filters/openpolicyagent/openpolicyagent.go
```diff
@@ -340,6 +340,7 @@ func NewOpenPolicyAgentRegistry(opts ...func(*OpenPolicyAgentRegistry) error) (*
 	go registry.startCleanerDaemon()
 
 	if registry.enableCustomControlLoop {
+		registry.startAndTriggerOpaPlugins()
```
Pushpalanka, line 343: We didn't recommend enabling custom control loop due to a CPU usage concern as I remember, when the number of bundles activating within a single Skipper is high. So if we rely on enabling custom control loop it will be a new addition to the mix that will need some benchmark/load testing.

Pushpalanka:
> > I think this was done intentionally to avoid slow starts, no?
> 
> Not sure, I just remember that the idea was that we are decoupled from routing updates and that we are. We should always ensure that we have everything to serve traffic. If OPA is configured why should we start if we can not serve OPA? Makes no sense to me.

We have two different behaviors for Skipper startup with OPA routes and Route updates coming with OPA filter with preloading enabled (default recommendation). 

**Startup with OPA filter**
This is blocking but parallel, which means startup will be blocked until OPA instance becomes ready. It had 30s time out though, so it will not block forever. In the follow up route updates it will again try to recover the OPA instances in the background.  If we need a hard failure here, it might need a modification. I will get back after checking if we covered this in tests. 
PS: This is the test for the scenario, OPA initialization fails at startup and later recovers, https://github.com/zalando/skipper/blob/ed60648be2a8dcae9043d0ffb8512772ffadc52d/filters/openpolicyagent/opaauthorizerequest/openpolicyagent_routeupdate_test.go#L205. It will be the one to check if we change the behavior for start up.

**Route update with OPA filter**
As you mentioned this was decoupled and routes get added even if OPA initialization failed and will keep on giving 503 in the route while attempting to self heal at each route update. 

It was keeping the behavior consistent between startup and route update, so the route gives 503 and keep on trying to self heal. Only difference was Skipper start was blocking and that 30s timeout will be the only indicator there to catch a failure happened. I also understand 503 can be hard to differentiate between if the backend is failing or OPA is failing. On the other hand, we had a concern, if we block Skipper start up at OPA failures and if it was trying to scale for load on other non-OPA routes, we are unnecessarily blocking that scaling as well.

szuecs:
> 
> It was keeping the behavior consistent between startup and route update, so the route gives 503 and keep on trying to self heal. Only difference was Skipper start was blocking and that 30s timeout will be the only indicator there to catch a failure happened. I also understand 503 can be hard to differentiate between if the backend is failing or OPA is failing. On the other hand, we had a concern, if we block Skipper start up at OPA failures and if it was trying to scale for load on other non-OPA routes, we are unnecessarily blocking that scaling as well.

I think the concern is outdated and we should block at startup if we enable OPA in skipper. If we serve 503, but we are ready, then the readiness does not serve the purpose of being ready to serve traffic.

Do you have a suggestion, how we should change as an alternative approach so we have a blocking startup until OPA is ready?
I can also create a separate function that block until OPA responds ready, if the config option is not used in cases.

Pushpalanka:
> I think the concern is outdated and we should block at startup if we enable OPA in skipper. If we serve 503, but we are ready, then the readiness does not serve the purpose of being ready to serve traffic.
> 
> Do you have a suggestion, how we should change as an alternative approach so we have a blocking startup until OPA is ready? I can also create a separate function that block until OPA responds ready, if the config option is not used in cases.

I couldn't still understand why the concern is outdated in the sense why the Skipper startup should as a whole fail when it was only the OPA routes affected. And it is also a possible case multiple OPA routes are working while only one is failing. Isn't it a problem when Skipper tries to scale up and it get blocked due to OPA unable to start with some transient S3 access issue. I think it is then again OPA becoming a blocking neighbor when Skipper is serving non-OPA and working OPA routes as well.

I would suggest we take an observability improvement keeping the current behavior where OPA doesn't disturb other unrelated functionality. The OPA instance health is currently tracked, https://github.com/zalando/skipper/blob/master/filters/openpolicyagent/openpolicyagent.go#L655 if we can expose this in metrics that could indicate and alert on problems. I am not sure exactly how to expose it yet.

szuecs:
> > I think the concern is outdated and we should block at startup if we enable OPA in skipper. If we serve 503, but we are ready, then the readiness does not serve the purpose of being ready to serve traffic.
> > Do you have a suggestion, how we should change as an alternative approach so we have a blocking startup until OPA is ready? I can also create a separate function that block until OPA responds ready, if the config option is not used in cases.
> 
> I couldn't still understand why the concern is outdated in the sense why the Skipper startup should as a whole fail when it was only the OPA routes affected. And it is also a possible case multiple OPA routes are working while only one is failing. Isn't it a problem when Skipper tries to scale up and it get blocked due to OPA unable to start with some transient S3 access issue. I think it is then again OPA becoming a blocking neighbor when Skipper is serving non-OPA and working OPA routes as well.

I think it was a miss understanding. I think @MustafaSaber had the same in his mind, so maybe at some time we made this decision.
If I ever had the opinion to not block until OPA is ready at skipper startup, then it changed.

> 
> I would suggest we take an observability improvement keeping the current behavior where OPA doesn't disturb other unrelated functionality. The OPA instance health is currently tracked, https://github.com/zalando/skipper/blob/master/filters/openpolicyagent/openpolicyagent.go#L655 if we can expose this in metrics that could indicate and alert on problems. I am not sure exactly how to expose it yet.

I think the best improvement is to block at startup and later do updates async of the route processing.
It's simple and in many cases skipper operators will have a dependency on OPA if they enable it, so it's part of the requirement to start a skipper instance. If we do not block we are open to the risk that OPA is forever not working on this instance and failing majority of request processing. I don't think that's a great solution if we do it the current way.

Pushpalanka:
> I think the best improvement is to block at startup and later do updates async of the route processing. It's simple and in many cases skipper operators will have a dependency on OPA if they enable it, so it's part of the requirement to start a skipper instance. If we do not block we are open to the risk that OPA is forever not working on this instance and failing majority of request processing. I don't think that's a great solution if we do it the current way.

Yes, failing fast at startup keeps the design simple and predictable. While blocking the entire Skipper startup because 1 out of 30 or so OPA instances fails might seem unfair, it's a conscious tradeoff to signal failures immediately rather than serving silent 503s. Then we accept that scaling up will be blocked if an OPA dependency is broken, which surfaces the issue where it needs to be fixed. Makes sense.

filters/openpolicyagent/openpolicyagent.go
a4180p: Shouldn't  we have also `if refistry.preloadingEnabled` here?
I think that registry.instances will be still empty here if it is not.

szuecs:
I was discussing today that we can also use `-status-checks` in order to test a route with a trivia OPA filter, such that you make sure OPA is ready but you do not depend on the end user that creates your routes.
I think this would be a good idea to document how to do this.

Basically you would use an inline route `-inline-routes`: 

```
opa_ready: Path("/opa/health") && HostAny("opa.localhost","opa.localhost:9999") -> opaServeResponse("opa-health-bundle") -> <shunt>;
```

and `-status-checks="http://opa.localhost:9999/opa/health"` to make sure you block at startup so long that you have a ready OPA inside skipper.

I create a doc issue for this, instead of following up on this PR.

PR #4200, opened 2026-08-20
apiusagemonitoring: prune filter cache of removed routes

### LINKED ISSUE #3552: add preprocessor to cleanup apimonitoring cache
This https://github.com/zalando/skipper/pull/3551 created unbounded caching.
Right now we think that the cache keys are stable enough to not fail.
However we have already awareness of how to do cleanups via PreProcessor. Check scheduler cleanup as an example.

hdimer:
Fixes #3552.

`apiUsageMonitoringSpec.filterMap`, added in #3551, is only ever written and read, never pruned. Every distinct filter-arg key the process sees stays cached for the lifetime of the instance, so on a long-running skipper whose routes churn the cache grows without bound. Note it is not only route churn: when the kubernetes admission webhook is enabled, `ValidateRoute` also goes through `CreateFilter`, so a parseable `apiUsageMonitoring` config that never becomes a live route still adds a permanent entry.

The spec now implements `routing.PostProcessor`: after each route table rebuild it collects the `*apiUsageMonitoringFilter` instances the new routes reference and drops every cache entry that is not among them. That is the cleanup you pointed at in the issue; the shape is closest to `OpenPolicyAgentRegistry.Do`, which reaps by instance identity rather than by route id like the scheduler does.

A post-processor rather than a pre-processor because filter instances only exist after route creation, so that is the first point where "still in use" is knowable.

No delete delay, unlike the scheduler queues and the OPA instances: those defer because something gets closed, whereas dropping a map entry here does not touch the filter. A route still holding the pointer keeps working, the entry is simply not served from cache again, and the worst case of a mis-scoped route slice is one recompute rather than a closed resource.

The mutex in `Do` is load-bearing, not decoration. Two `Do` calls cannot overlap, but the admission webhook path above reaches `CreateFilter` from request goroutines, so the sweep genuinely races the cache write.

`skipper.go` registers it via an interface assertion, which also encodes "monitoring is disabled": with `-api-usage-monitoring=false` the spec is a `noopSpec` that caches nothing and does not implement `Do`.

Tested: two cached filters and a route table referencing one of them, placed on the second route and behind another filter so the test fails if the scan stops at the first route or the first filter; the referenced entry survives and still hits the cache, the other is dropped and recreated. Plus an empty route table emptying the cache. Ran `gofmt -s`, `go build ./...`, `go vet`, `staticcheck`, and the package and `routing` tests with `-race`.

Written with AI assistance (Claude); I reviewed the change, wrote and ran the tests, and I stand behind it.

### DIFF
--- a/skipper.go
+++ b/skipper.go
@@ -2029,6 +2029,13 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 		OpenTracingClientTraceByTag: o.OpenTracingClientTraceByTag,
 	}
 
+	apiUsageMonitoringFilter := apiusagemonitoring.NewApiUsageMonitoring(
+		o.ApiUsageMonitoringEnable,
+		o.ApiUsageMonitoringRealmKeys,
+		o.ApiUsageMonitoringClientKeys,
+		o.ApiUsageMonitoringRealmsTrackingPattern,
+	)
+
 	admissionControlFilter := shedder.NewAdmissionControl(shedder.Options{
 		Tracer:                      tracer,
 		OpenTracingClientTraceByTag: o.OpenTracingClientTraceByTag,
@@ -2057,12 +2064,7 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 		auth.TokenintrospectionWithOptions(auth.NewSecureOAuthTokenintrospectionAllKV, tio),
 		auth.WebhookWithOptions(who),
 		auth.NewOIDCQueryClaimsFilter(),
-		apiusagemonitoring.NewApiUsageMonitoring(
-			o.ApiUsageMonitoringEnable,
-			o.ApiUsageMonitoringRealmKeys,
-			o.ApiUsageMonitoringClientKeys,
-			o.ApiUsageMonitoringRealmsTrackingPattern,
-		),
+		apiUsageMonitoringFilter,
 		admissionControlFilter,
 	)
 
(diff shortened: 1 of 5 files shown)


szuecs:
Thanks your write up seems to be nailing it :)

szuecs (approved):

szuecs:
@hdimer can you please fetch newest changes from origin/master and rebase your branch on this and do a force push?
The tests break because of Go 1.27 and the old linter version.

skipper.go
```diff
@@ -2514,6 +2516,12 @@ func run(o Options, sig chan os.Signal, idleConnsCH chan struct{}) error {
 		ro.PostProcessors = append(ro.PostProcessors, failClosedRatelimitPostProcessor)
 	}
 
+	// the api usage monitoring spec caches filters and implements
+	// routing.PostProcessor to prune that cache, unless it is disabled
+	if pp, ok := apiUsageMonitoringFilter.(routing.PostProcessor); ok {
```
AlexanderYastrebov: Why conditional if we know it implements the interface?

hdimer: When api-usage-monitoring is disabled (the default), `NewApiUsageMonitoring` returns a `noopSpec`, which implements `filters.Spec` but not `routing.PostProcessor`, so a plain assertion would panic at startup. The comma-ok guard skips registering the cache-pruning post-processor in that case. I can instead gate on `o.ApiUsageMonitoringEnable` explicitly if you find that clearer, though that recouples the two call sites to the flag; the type check keeps this block self-contained. Happy to go whichever way you prefer.

hdimer:
@szuecs rebased onto the latest `master` and force-pushed. Builds and `go test ./filters/apiusagemonitoring/` pass locally on Go 1.27, so the earlier failure should clear.

skipper.go
szuecs: Maybe it would make sense to implement `routing.PostProcessor` with empty body for `noopSpec`

szuecs:
:+1:

skipper.go
hdimer: Good call, done. `noopSpec` now implements `routing.PostProcessor` with a noop `Do` that returns the routes unchanged, so the registration in `skipper.go` is unconditional and the type-assertion guard is gone. Added a small test that the disabled spec satisfies the interface and leaves routes untouched. Force-pushed.

szuecs: Thanks!

szuecs:
:+1:

MustafaSaber (approved):

PR #4201, opened 2026-08-20
feature: add optional response status condition to logBody filter

### LINKED ISSUE #4047: Log body with response status condition
We would like to get more insights on requests failing to be re-routed from Skipper to service due to 5xx error.

We would like skipper to log the body of the request **only when** a 5xx error happens

We tried to use [loopbackifstatus](https://opensource.zalando.com/skipper/reference/filters/#loopbackifstatus) but body is unavailable so [logbody](https://opensource.zalando.com/skipper/reference/filters/#logbody) returned empty

Yes, but no time

somaz94:
logBody can only log unconditionally today, so getting the request body of failing requests means logging every successful one too.

This adds an optional third argument, the response status code from which on to log:

```
* -> logBody("request", 1024, 500) -> "https://www.example.org";
```

For a response body the status is already known when the filter runs, so it is only a check. For a request body it is not known while it streams, so up to limit bytes are buffered and logged once the response arrives. The body itself is not held back and not truncated. Without the third argument nothing changes.

Ran locally: `make fmt`, `make lint` and `go test -race ./filters/diag/` are green. In `make shortcheck` the only failures are the redis/valkey tests that need Docker, and they fail the same way on a clean master here.

One thing I am not sure about: buffering on the request side moves logBody away from "close to no overhead". I kept it in logBody because the limit already caps the buffer, but I am happy to split it into a separate filter if you prefer.

fixes #4047

### DIFF
--- a/filters/diag/logbody.go
+++ b/filters/diag/logbody.go
@@ -24,13 +24,6 @@ type logBody struct {
 }
 
 // NewLogBody creates a filter specification for the 'logBody()' filter.
-//
-// It takes the type of the body to log, "request" or "response", a limit
-// of the number of bytes to log and an optional response status code from
-// which on to log. Without the status code the body is logged in chunks
-// while it streams. With it, a response body is logged the same way once
-// the status matches, while a request body is buffered up to the limit
-// and logged after the response status is known.
 func NewLogBody() filters.Spec { return logBody{} }
 
 // Name returns the logBody filter name.
--- a/filters/diag/logbody.go
+++ b/filters/diag/logbody.go
@@ -18,9 +18,12 @@ type logBody struct {
 	limit    int
 	request  bool
 	response bool
-	// minStatus is the response status code from which on the body is
-	// logged. Zero means no condition, i.e. always log.
-	minStatus int
+	// statusPrefixes selects which responses get logged, following the
+	// same matching rules as the enableAccessLog filter: a value below 10
+	// matches a status class (5 matches 5xx), below 100 a sub-class (50
+	// matches 50x) and any larger value an exact status code. Empty means
+	// no condition, i.e. always log.
+	statusPrefixes []int
 }
 
 // NewLogBody creates a filter specification for the 'logBody()' filter.
@@ -33,12 +36,11 @@ func (logBody) Name() string {
 
 func (logBody) CreateFilter(args []interface{}) (filters.Filter, error) {
 	var (
-		request   = false
-		response  = false
-		minStatus = 0
+		request  = false
+		response = false
 	)
 
-	if len(args) != 2 && len(args) != 3 {
+	if len(args) < 2 {
 		return nil, filters.ErrInvalidFilterParameters
 	}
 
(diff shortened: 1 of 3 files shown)


filters/diag/logbody.go
```diff
@@ -1,20 +1,36 @@
...
 }
 
 // NewLogBody creates a filter specification for the 'logBody()' filter.
+//
+// It takes the type of the body to log, "request" or "response", a limit
+// of the number of bytes to log and an optional response status code from
+// which on to log. Without the status code the body is logged in chunks
+// while it streams. With it, a response body is logged the same way once
+// the status matches, while a request body is buffered up to the limit
+// and logged after the response status is known.
```
szuecs: Please drop this doc here, thanks

szuecs:
@somaz94 looks good, one nit of code docs, but I have one request:

Can you please add a test case that does the request filter with status code config and then do the request but backend does a time.Sleep and the client cancels connection before the sleep is over. I want to see that this does not leak memory.

somaz94:
Added `TestLogBodyRequestStatusConditionClientCancel` in `filters/diag/` and dropped the doc block you flagged.

The test does what you asked: the backend sleeps well past the client timeout, so the client always disconnects before a response status exists. It then asserts three things:

- the buffered body is capped by the filter's `limit` (1 KiB), not by the 256 KiB upload — so a large body cannot inflate what is retained;
- nothing is logged when the client cancels before a status is known, since logging is deferred to the response phase that never runs;
- 50 cancelled requests in a row do not grow the heap.

The workflow runs are sitting at `action_required` and need your approval to start.

szuecs (approved):

filters/diag/logbody.go
```diff
@@ -1,17 +1,26 @@
...
+// defers the decision to log until the response status is known.
+const logBodyRequestKey = "filter." + filters.LogBodyName + ".request"
+
 type logBody struct {
 	limit    int
 	request  bool
 	response bool
+	// minStatus is the response status code from which on the body is
+	// logged. Zero means no condition, i.e. always log.
+	minStatus int
```
MustafaSaber: `>=` doesn't align with what we use in other filters, I would go with the same pattern we use for [`enableAccessLog`](https://opensource.zalando.com/skipper/reference/filters/#enableaccesslog)
 to only log for matched status codes and it can be list of codes.

szuecs: Yeah you are right makes sense to me

somaz94:
Switched to the enableAccessLog prefix form.

The arguments are prefixes rather than full status codes, so `logBody("request", 1024, 5)` covers 5xx and replaces the old `500` threshold, and `logBody("request", 1024, 5, 429)` now works where the threshold could not express a class plus one unrelated code.

One difference from enableAccessLog: I reject prefixes that cannot match any status (0, 6, 99, 600) at filter creation. enableAccessLog accepts those and just never matches — happy to drop the check if you would rather the two behave identically.

Added `TestLogBodyMatchStatus` for the matching rules. The existing 502 test moved from `500` to `5`, since exact matching no longer covers it.

szuecs (approved):

MustafaSaber (approved):
Thanks for your contribution!

PR #4206, opened 2026-08-21
fix(proxy): preserve upstream traceparent when OTel Extract fails

kostz:
## Problem

When skipper uses the OpenTracing→OTel bridge and receives a request whose
`traceparent` header cannot be parsed by the bridge (e.g. the header was
rewritten by instrumentation middleware in the upstream service), `Extract`
fails and skipper starts a new root span with a fresh traceID.

The subsequent `Inject` call then uses the W3C TraceContext propagator which
calls `http.Header.Set("Traceparent", ...)` — unconditionally **replacing** the
incoming `traceparent` with skipper's new, unrelated traceID.

Any downstream service that understands W3C TraceContext will now see a
completely different traceID. Its spans become detached from the upstream trace
and the end-to-end trace chain is broken.

### Why this did not happen with the LightStep tracer

The LightStep OpenTracing tracer's `Inject` writes proprietary headers
(`x-ot-span-context`, etc.) but **never touches `traceparent`**. The original
W3C header therefore passed through unchanged regardless of whether `Extract`
succeeded.

After switching to the OTel bridge, `Inject` delegates to the W3C TraceContext
propagator, which always overwrites the header.

### Failure scenario

```

upstream  ──(traceparent: 00-AAAA...-span1-01)──►  skipper  ──►  downstream
│
Extract fails (AAAA... unrecognised)
skipper creates new root span BBBB...
Inject: traceparent: 00-BBBB...-span2-01
│
downstream sees BBBB... ≠ AAAA...
trace chain broken
```

## Fix

Save the original `traceparent` and `tracestate` values before calling
`Inject`. After the call, if a `traceparent` is present with a **different
traceID**, restore the originals.

```go
origTraceparent := req.Header.Get("Traceparent")
origTracestate  := req.Header.Get("Tracestate")
_ = p.tracing.tracer.Inject(ctx.proxySpan.Context(), ot.HTTPHeaders, carrier)

if origTraceparent != "" && !sameW3CTraceID(origTraceparent, req.Header.Get("Traceparent")) {
    req.Header.Set("Traceparent", origTraceparent)
    if origTracestate != "" {
        req.Header.Set("Tracestate", origTracestate)
    } else {
        req.Header.Del("Tracestate")
    }
}
```

The guard fires only when Extract failed (a new root traceID was injected).
When Extract succeeds, skipper's span is a child of the upstream span and
shares the same traceID — the condition is false and headers are left as-is.

This is safe with the LightStep tracer: Inject never writes traceparent,
so the condition never fires and behaviour is unchanged.

## Tests

TestSameW3CTraceID — unit tests for the helper (same/different traceID, empty input, malformed values).
TestTraceparentPreservedOnExtractFailure — end-to-end test using tracingtest.NewTracer() (whose Extract always returns an error) verifying the original traceparent is forwarded to the backend unchanged.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -1129,7 +1129,21 @@ func (p *Proxy) makeBackendRequest(ctx *context, requestContext stdlibcontext.Co
 	p.setCommonSpanInfo(u, req, ctx.proxySpan)
 
 	carrier := ot.HTTPHeadersCarrier(req.Header)
+	origTraceparent := req.Header.Get("Traceparent")
+	origTracestate := req.Header.Get("Tracestate")
 	_ = p.tracing.tracer.Inject(ctx.proxySpan.Context(), ot.HTTPHeaders, carrier)
+	// If the incoming request had a traceparent that Inject replaced with a different
+	// traceID, the upstream trace context was unrecognized (Extract failed and skipper
+	// started a new root span). Restore the original headers so downstream services
+	// that understand the upstream format keep their trace chain intact.
+	if origTraceparent != "" && !sameW3CTraceID(origTraceparent, req.Header.Get("Traceparent")) {
+		req.Header.Set("Traceparent", origTraceparent)
+		if origTracestate != "" {
+			req.Header.Set("Tracestate", origTracestate)
+		} else {
+			req.Header.Del("Tracestate")
+		}
+	}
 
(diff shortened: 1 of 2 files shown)


szuecs:
The fix seems to be a hack and we need a proper configuration instead.
Please explain why the tracer would not do this automatically.
And yes I don't want to talk to an AI agent.

szuecs:
I think https://github.com/zalando/skipper/pull/4208 will be able to address the issue.

PR #4208, opened 2026-08-21
otel: support configuring the TracerProvider Sampler

madumalt:
Without trace.WithSampler(...), the SDK defaults to ParentBased(AlwaysSample), which respects an upstream trace's sampled flag. Once one hop marks a trace as not sampled, skipper stops sampling its own spans and stops propagating sampled=1 downstream, letting an upstream decision cascade through the rest of the chain.

Add Options.Sampler / Options.SamplerArg (yaml: sampler / samplerArg), mirroring OTEL_TRACES_SAMPLER / OTEL_TRACES_SAMPLER_ARG. Setting sampler: always_on makes skipper always record and always propagate sampled=1 regardless of the incoming trace context. Empty Sampler preserves current behavior (backward compatible).


szuecs (approved):

madumalt:
👍

PR #4209, opened 2026-08-22
feat(auth): add native rfc9421 HTTP Message Signatures filter (#3952)

### LINKED ISSUE #3952: HTTP signature rfc9421
A new filter that implements RFC9421

References 
- https://github.com/yaronf/httpsign
- https://github.com/lestrrat-go/htmsig
- https://github.com/dadrus/httpsig
- https://github.com/common-fate/httpsig
- https://www.rfc-editor.org/rfc/rfc9421
- https://github.com/curl/curl/pull/21239

shardool-patil:
### Description

Implements native support for [RFC 9421 (HTTP Message Signatures)](https://www.rfc-editor.org/rfc/rfc9421) and adds the `rfc9421` egress signing filter.

### Changes
* **RFC 9421 Engine (`rfc/rfc9421/`):**
  * Zero third-party dependencies; implemented using Go standard library crypto packages.
  * Component canonicalization for derived components (`@method`, `@path`, `@query`, `@authority`, `@scheme`, etc.) and HTTP headers.
  * Structured fields formatting for `Signature-Input` and `Signature` headers.
  * Signing support for `hmac-sha256`, `ed25519`, `rsa-pss-sha512`, `rsa-v1_5-sha256`, and `ecdsa-p256-sha256`.
* **Egress Filter (`filters/auth/rfc9421.go`):**
  * Reads private keys dynamically via `x-rfc9421-private-key` header and strips the header before forwarding upstream.
  * Supports PKCS#8, PKCS#1 RSA, EC, Ed25519, and raw HMAC keys.
* **Registry & Docs:** Registered in `filters/builtin/builtin.go` and added filter reference documentation to `docs/reference/filters.md`.

### Verification
* Unit tests for HMAC, Ed25519, RSA-PSS, missing components, and filter parameter parsing.
* Verified `go test` passes across `rfc/rfc9421`, `filters/auth`, and `filters/builtin`.

### Related Issues
* Closes #3952 

@szuecs I have implemented the code as explained just take a look at it when you get time.


szuecs:
@shardool-patil can you add the egress filter also to docs/reference/egress.md ?
One or more sentence(s) and a link to `docs/reference.md#rfc9421` somewhere should be enough.

docs/reference/filters.md
```diff
@@ -2479,6 +2479,40 @@ This filter reads the body in memory. This is needed to generate signature as pe
 This filter resets `read` and `close` implementations of body to default. So when a filter before this filter has some custom implementations of these methods, they would be overwritten.
 
 
+### rfc9421
+
+Signs outgoing HTTP requests according to [RFC 9421 (HTTP Message Signatures)](https://www.rfc-editor.org/rfc/rfc9421). It computes the signature across specified HTTP fields and attaches the `Signature-Input` and `Signature` headers.
+
+The private key must be supplied via the temporary `x-rfc9421-private-key` request header. The filter reads the key into memory and deletes the header before forwarding the request upstream.
```
szuecs: Not a bad idea but I would like to have this integrated via ./secrets module, instead.
The people who create the filter are likely not the ones that run skipper, so they have no idea where to read the key from and should not have access to the key. They could easily get the key:
```
setRequestHeaderFromSecret("x-rfc9421-private-key", "/tmp/secrets/api-secret")
-> logHeader("request")
```
If the operator is serious about this kind of problem then they can disable the filter `setRequestHeaderFromSecret` to ensure that people that can create a route and be able to read logs are not able to read the secret key.

filters/auth/rfc9421.go
```diff
@@ -0,0 +1,170 @@
...
+type rfc9421Filter struct {
+	keyID      string
+	algorithm  string
+	components []string
+	signLabel  string
+}
+
+// NewRFC9421 creates a new filter specification for RFC 9421 HTTP Message Signatures.
+func NewRFC9421() filters.Spec {
+	return &rfc9421Spec{}
```
szuecs: spec should get the reference to the secret key and you will need to add a config flag in config/config.go and pass it to skipper.Options and initiate `NewRFC9421(secretPath) ` in ./skipper.go , there are a bunch of similar cases that are done like this.

filters/auth/rfc9421.go
```diff
@@ -0,0 +1,170 @@
...
+		return
+	}
+
+	privKey, err := parsePrivateKey(keyRaw, f.algorithm)
+	if err != nil {
+		logger.Errorf("rfc9421: failed to parse private key: %v", err)
+		return
+	}
+
+	signer, err := rfc9421.NewSigner(rfc9421.SignerConfig{
```
szuecs: If we have the secret reference this can be created once in `NewRFC9421()` and passed to the filter instance in CreateFilter() without arguments read from the user.
It will be much more fast and less CPU consumption if we do it like this.

filters/builtin/builtin.go
```diff
@@ -236,6 +236,7 @@ func Filters() []filters.Spec {
 		accesslog.NewEnableAccessLog(),
 		auth.NewForwardToken(),
 		auth.NewForwardTokenField(),
+		auth.NewRFC9421(),
```
szuecs: this has to be moved to skipper.go

rfc/rfc9421/crypto.go
```diff
@@ -0,0 +1,116 @@
...
+	"crypto/rsa"
+	"crypto/sha256"
+	"crypto/sha512"
+	"encoding/base64"
+	"fmt"
+)
+
+// signData computes the cryptographic signature over the provided data
+// using the algorithm and private key configured on the Signer.
+func (s *Signer) signData(data []byte) (string, error) {
```
szuecs: all functions are on "*Signer" so I would say we should move this to to `signer.go` instead of having a separate file

rfc/rfc9421/signer.go
```diff
@@ -0,0 +1,118 @@
...
+func (s *Signer) buildSignatureParams(created time.Time) string {
+	var compStrings []string
+	for _, comp := range s.config.Components {
+		clean := strings.TrimSpace(comp)
+		if clean != "" {
+			compStrings = append(compStrings, fmt.Sprintf("%q", clean))
+		}
+	}
+
+	params := fmt.Sprintf("(%s);created=%d", strings.Join(compStrings, " "), created.Unix())
```
szuecs: params is likely better to be a StringBuilder

rfc/rfc9421/signer.go
```diff
@@ -0,0 +1,118 @@
...
+	Components []string
+	SignLabel  string
+	Created    *time.Time
+	Expires    *time.Time
+	Nonce      string
+}
+
+// Signer orchestrates HTTP message signing according to RFC 9421.
+type Signer struct {
+	config SignerConfig
```
szuecs: looks weird to me that we have a SignerConfig. Just move everything to Signer instead.

shardool-patil:
I've updated the PR to address the feedback:

* **Secrets Integration:** Migrated key loading to Skipper's `secrets.SecretsReader` (`sp`) via `auth.NewRFC9421(sp)` in `skipper.go`, removing the temporary header
* **Performance:** Pre-parsed private keys once during `NewSigner()` / `CreateFilter()` at route creation rather than per request
* **Signer Consolidation:** Merged `crypto.go` into `signer.go`, flattened configuration into `Signer`, and switched parameter formatting to `strings.Builder`
* **Filter Registration:** Registered `auth.NewRFC9421(sp)` in `skipper.go` alongside other secrets-backed filters
* **Documentation:** Updated `docs/reference/filters.md` with `secretRef` configuration details and added egress usage to `docs/reference/egress.md`
Take a look at the changes whenever you have time @szuecs

docs/reference/egress.md
```diff
@@ -1,87 +1,90 @@
...
-   * dynamic secrets lookup used by `bearerinjector filter`
-   * encryption and decryption used by [OpenID Connect filters](filters.md#oauthoidcuserinfo)
+Skipper has some features, which are egress specific. Some features, for example
+[dropRequestHeader](filters.md#droprequestheader) or
+[ratelimit](filters.md#ratelimit), might also be used, but are not listed here:
+
+- Circuit breaker filters
+  - [consecutiveBreaker](filters.md#consecutivebreaker)
+  - [rateBreaker](filters.md#ratebreaker)
+  - [disableBreaker](filters.md#disablebreaker)
```
szuecs: I think there is a whitespace missing, there is a need for 3x " " in order to make markdown lists a sub-list

docs/reference/egress.md
```diff
@@ -1,87 +1,90 @@
...
+- [bearerinjector](filters.md#bearerinjector) filter, that injects tokens for an app
+- [rfc9421](filters.md#rfc9421) filter, that signs outgoing HTTP requests according to RFC 9421 HTTP Message Signatures using secrets managed by the secrets module
+- The secrets module that does:
+  - automated secrets rotation read from files used by the `bearerinjector` filter and `rfc9421` filter
```
szuecs: here is also the 3rd whitespace missing

szuecs:
@shardool-patil thanks for the changes! Can you please revert the egress.md changes that are not your additions?

Some of these are breaking markdown and I don't want to see changes that are not part of the feature in this PR.

Rest of the changes lgtm.

Thanks!

shardool-patil:
@szuecs i have updated the requested changes so you can take a look.

szuecs:
Do we need to pass the keyfile in the filter?
I would expect that it's fine if only the operator has to know this by providing a flag and we use always the same key to sign.
I think that it would be beneficial for the user that would use the filter.

szuecs:
There is also a linter error detected:
```
staticcheck -checks "all,-ST1003,-ST1020" ./...
Error: rfc/rfc9421/base.go:1:1: at least one file in a package should have a package comment (ST1000)
Error: rfc/rfc9421/signer.go:1:1: at least one file in a package should have a package comment (ST1000)
```

Can you add a "doc.go" that has a godoc style documentation for the package?
Thanks!

shardool-patil:
> Do we need to pass the keyfile in the filter? I would expect that it's fine if only the operator has to know this by providing a flag and we use always the same key to sign. I think that it would be beneficial for the user that would use the filter.

The initial design took the `keyfile` argument inside the filter to:
1. Follow the pattern used in `bearerinjector("/path/to/secret")`.
2. Support multi-target egress routing where different routes call different external partners that require distinct signing keys.

However, I agree that abstracting the file system path away from route authors provides a cleaner interface and aligns better with operator/developer separation.

We have two ways:
- **Lookup by Key ID:** Use the `keyId` argument directly to resolve the key file from the `-credentials-paths` directory (e.g., `rfc9421("my-key-id", "hmac-sha256", ...)` resolves to `<credentials-path>/my-key-id`).
- **Global Operator Flag:** Introduce a global flag (e.g., `-rfc9421-key-path`) to supply a single signing key across all routes.

@szuecs take a look at this, i have fixed the linter failing ci tests and the codeql one is not from the code that i have worked on.

szuecs:
> > Do we need to pass the keyfile in the filter? I would expect that it's fine if only the operator has to know this by providing a flag and we use always the same key to sign. I think that it would be beneficial for the user that would use the filter.
> 
> The initial design took the `keyfile` argument inside the filter to:
> 
>     1. Follow the pattern used in `bearerinjector("/path/to/secret")`.
> 
>     2. Support multi-target egress routing where different routes call different external partners that require distinct signing keys.
> 
> 
> However, I agree that abstracting the file system path away from route authors provides a cleaner interface and aligns better with operator/developer separation.
> 
> We have two ways:
> 
>     * **Lookup by Key ID:** Use the `keyId` argument directly to resolve the key file from the `-credentials-paths` directory (e.g., `rfc9421("my-key-id", "hmac-sha256", ...)` resolves to `<credentials-path>/my-key-id`).
> 
>     * **Global Operator Flag:** Introduce a global flag (e.g., `-rfc9421-key-path`) to supply a single signing key across all routes.

From what I see  from bearerinjector and also oidc filter vs oauthGrant it's not worth the complexity to have the user reference something that they do not know. Therefore I think the best would be to have the file only in the flags and not in the filter. So also drop keyid from the filter input.

> 
> @szuecs take a look at this, i have fixed the linter failing ci tests and the codeql one is not from the code that i have worked on.

staticcheck says you need a godoc for the package:

```
$ staticcheck -checks "all,-ST1003,-ST1020" ./...
Error: rfc/rfc9421/base.go:1:1: at least one file in a package should have a package comment (ST1000)
Error: rfc/rfc9421/signer.go:1:1: at least one file in a package should have a package comment (ST1000)
make: *** [Makefile:167: staticcheck] Error 1
Error: Process completed with exit code 2.
```

To fix it create  rfc/rfc9421/doc.go with content:

```go
// Package rfc9421 implements [RFC9421](link to rfc) 
package rfc9421
```

shardool-patil:
I was away for a while, now i have updated the code just take a look @szuecs 

- **Package Documentation:** Added `rfc/rfc9421/doc.go` with the package doc linking to RFC 9421, resolving the `staticcheck` ST1000 issue.
- **Simplified Filter Signature:** Removed `keyFile` and `keyID` from the `rfc9421()` filter arguments so route authors only need to configure the signature details (`rfc9421("algorithm", "components"[, "label"])`).
- **Global Operator Flags:** Added `-rfc9421-key-file` and `-rfc9421-key-id` CLI flags in `config/config.go` and wired them through `skipper.Options`.
- **Tests & Verification:** Updated `filters/auth/rfc9421_test.go` to match the new signature and verified that all unit tests and static analysis pass cleanly.

shardool-patil:
As i was making doc.go a newline was not added which was the cause for the failing ci tests, now that has been fixed, so take a look at this @szuecs

filters/auth/rfc9421.go
```diff
@@ -0,0 +1,124 @@
...
+//	[0] algorithm (string): cryptographic signing algorithm (e.g. "hmac-sha256", "ed25519", "rsa-pss-sha512")
+//	[1] components (string): comma-separated HTTP components to cover (e.g. "@method, @path, @authority")
+//	[2] signatureLabel (string, optional): signature label prefix (default: "sig1")
+func (s *rfc9421Spec) CreateFilter(args []interface{}) (filters.Filter, error) {
+	if len(args) < 2 || len(args) > 3 {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	if s.keyFile == "" || s.keyID == "" {
+		return nil, errors.New("rfc9421: key file and key ID must be configured via flags")
```
szuecs: fmt.Errorf("your-msg: %w", filters.ErrInvalidFilterParameters)

filters/auth/rfc9421.go
```diff
@@ -0,0 +1,124 @@
...
+			return nil, fmt.Errorf("rfc9421: failed to read key file %q: %w", s.keyFile, err)
+		}
+	}
+
+	if len(keyBytes) == 0 {
+		return nil, fmt.Errorf("rfc9421: secret %q is empty", s.keyFile)
+	}
+
+	var components []string
+	for _, part := range strings.Split(componentsStr, ",") {
```
szuecs: SplitSeq() instead of Split

szuecs (dismissed):

filters/auth/rfc9421.go
shardool-patil: @szuecs both requested changes have been addressed

szuecs:
👍

shardool-patil:
@szuecs it looks like `check-race_proxy` failed due to a timeout in `TestPHCNoHealthyEndpoints`, how do we proceed next

szuecs:
> @szuecs it looks like `check-race_proxy` failed due to a timeout in `TestPHCNoHealthyEndpoints`, how do we proceed next

no worries I take care, retry helps and the problem is known caused by throttled CI env
It's a statistical test that we have no good solution for throttled env

shardool-patil:
@szuecs danke schon

rfc/rfc9421/base.go
```diff
@@ -0,0 +1,155 @@
...
+
+	var trimmed []string
+	for _, v := range values {
+		trimmed = append(trimmed, strings.TrimSpace(v))
+	}
+	return strings.Join(trimmed, ", "), nil
+}
+
+// CanonicalizeQuery normalizes query parameters if needed.
+func CanonicalizeQuery(rawQuery string) (string, error) {
```
a4180p: this function isn't used anywhere

filters/auth/rfc9421.go
```diff
@@ -0,0 +1,123 @@
...
+
+	log "github.com/sirupsen/logrus"
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/rfc/rfc9421"
+	"github.com/zalando/skipper/secrets"
+)
+
+const (
+	// RFC9421FilterName is the filter name used in eskip routes.
+	RFC9421FilterName = "rfc9421"
```
a4180p: please, add filter name to [this list ](https://github.com/zalando/skipper/blob/284767a948d392a59096a47b9599f29bd103bd6d/filters/filters.go#L221)and use it in Name function

docs/reference/filters.md
```diff
@@ -2548,6 +2479,37 @@ This filter reads the body in memory. This is needed to generate signature as pe
...
+
+Signs outgoing HTTP requests according to [RFC 9421 (HTTP Message Signatures)](https://www.rfc-editor.org/rfc/rfc9421). It retrieves the private key or secret material referenced by `secretRef` from the secrets reader (configured via `-credentials-paths`), computes the signature across specified HTTP components, and attaches the `Signature-Input` and `Signature` headers.
+
+Parameters:
+
+* `secretRef` (string): Path or identifier of the secret in the configured secrets reader.
+* `Key ID` (string): Identifier for the signing key (e.g., `"key-1"`).
+* `Algorithm` (string): Cryptographic signing algorithm. Supported: `"hmac-sha256"`, `"rsa-pss-sha512"`, `"rsa-v1_5-sha256"`, `"ecdsa-p256-sha256"`, `"ed25519"`.
+* `Components` (string): Comma-separated list of HTTP fields and derived components to cover in the signature (e.g., `"@method, @path, @authority, content-type"`).
+* `Signature Label` (string, optional): Label prefix for the headers. Default: `"sig1"`.
```
a4180p, line 2490: it doesn't match the implementation, it doesn't allow more than 3 args 
https://github.com/zalando/skipper/pull/4209/changes#diff-f7c98055c7fcefb99dc6c7613dccb74eb999ecf612e25b5597a953f358d4571fR50

filters/auth/rfc9421.go
```diff
@@ -0,0 +1,123 @@
...
+		customLabel, ok := args[2].(string)
+		if !ok || strings.TrimSpace(customLabel) == "" {
+			return nil, filters.ErrInvalidFilterParameters
+		}
+		label = strings.TrimSpace(customLabel)
+	}
+
+	var keyBytes []byte
+	if s.secretsReader != nil {
+		if kb, ok := s.secretsReader.GetSecret(s.keyFile); ok && len(kb) > 0 {
```
a4180p: it will be read only on CreateFiler, the value will remain the same if key is rotated.

filters/auth/rfc9421.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+	}
+
+	if len(keyBytes) == 0 {
+		var err error
+		keyBytes, err = os.ReadFile(s.keyFile)
+		if err != nil {
+			return nil, fmt.Errorf("rfc9421: failed to read key file %q: %w", s.keyFile, err)
+		}
+	}
```
a4180p: I think we should always use secretsReader.

rfc/rfc9421/signer.go
```diff
@@ -0,0 +1,218 @@
...
+
+	block, _ := pem.Decode(keyData)
+	data := keyData
+	if block != nil {
+		data = block.Bytes
+	}
+
+	if pk, err := x509.ParsePKCS8PrivateKey(data); err == nil {
+		return pk, nil
+	}
```
a4180p: all this must be a part of switch below.
it is possible to pass a mismatching pair now.

rfc/rfc9421/base.go
```diff
@@ -0,0 +1,155 @@
...
+		return "", fmt.Errorf("%w: URL is nil for @request-target", ErrMissingComponent)
+	}
+	path := req.URL.Path
+	if path == "" {
+		path = "/"
+	}
+	if req.URL.RawQuery != "" {
+		path = path + "?" + req.URL.RawQuery
+	}
+	method := strings.ToLower(req.Method)
```
a4180p: method is not a part of the request target according to the examples here
https://www.rfc-editor.org/info/rfc9421/#section-2.2.5

rfc/rfc9421/base.go
```diff
@@ -0,0 +1,155 @@
...
+		if method == "" {
+			method = http.MethodGet
+		}
+		return strings.ToUpper(method), nil
+
+	case "@target-uri":
+		if req.URL == nil {
+			return "", fmt.Errorf("%w: URL is nil for @target-uri", ErrMissingComponent)
+		}
+		return req.URL.String(), nil
```
a4180p: it is possible that URL.String() will omit empty fields https://pkg.go.dev/net/url#URL.String

https://www.rfc-editor.org/info/rfc9421/#section-2.2.2  always requires full URL

rfc/rfc9421/signer.go
```diff
@@ -0,0 +1,218 @@
...
+		hashed := sha256.Sum256(data)
+		return rsa.SignPKCS1v15(rand.Reader, key, crypto.SHA256, hashed[:])
+
+	case AlgEcdsaP256:
+		key, ok := s.parsedKey.(*ecdsa.PrivateKey)
+		if !ok {
+			return nil, errors.New("invalid ECDSA key")
+		}
+		hashed := sha256.Sum256(data)
+		return ecdsa.SignASN1(rand.Reader, key, hashed[:])
```
a4180p: i could be wrong, but according to my (very limited) understanding of the signing process described [here](https://www.rfc-editor.org/info/rfc9421/#name-ecdsa-using-curve-p-256-dss) isn't the same as [SignASN1](https://cs.opensource.google/go/go/+/refs/tags/go1.27.1:src/crypto/ecdsa/ecdsa.go;l=391)

a4180p: Also, test for this algo is missing.

rfc/rfc9421/signer.go
```diff
@@ -0,0 +1,218 @@
...
+	AlgEd25519      = "ed25519"
+)
+
+// Signer orchestrates HTTP message signing according to RFC 9421.
+type Signer struct {
+	KeyID          string
+	Algorithm      string
+	Components     []string
+	SignatureLabel string
+	PrivateKey     []byte
```
a4180p: why do we need this field, its parsed value exist in the same datastructure below?

shardool-patil:
@a4180p thanks for thorough review, all points are addressed, you can take a look

rfc/rfc9421/signer.go
shardool-patil: All the suggestions have been handled

rfc/rfc9421/base.go
szuecs: Interesting that staticcheck Unnn didn't found it

docs/reference/filters.md
a4180p: @shardool-patil this is still valid

shardool-patil: @a4180p i had edited the code to include only 3 parameters in previous commit itself, i am confused why it still is not showing outdated, can you take a look at the change as in the code i can see it has been updated

filters/auth/rfc9421.go
```diff
@@ -0,0 +1,152 @@
...
+		return
+	}
+
+	signer, err := f.getSigner(keyBytes)
+	if err != nil {
+		log.Errorf("rfc9421: failed to get signer: %v", err)
+		return
+	}
+
+	req := ctx.Request()
```
a4180p: The thing is that [this request is later modified](https://github.com/zalando/skipper/blob/master/proxy/proxy.go#L666-L716), so scheme and host can change.

rfc/rfc9421/base.go
```diff
@@ -0,0 +1,173 @@
...
+		return strings.ToLower(scheme), nil
+
+	case "@request-target":
+		return s.getRequestTarget(req)
+
+	case "@path":
+		if req.URL == nil || req.URL.Path == "" {
+			return "/", nil
+		}
+		return req.URL.Path, nil
```
a4180p: URL.Path is decoded - https://cs.opensource.google/go/go/+/refs/tags/go1.27.1:src/net/url/url.go;l=268
RFC requires path to be not decoded - https://www.rfc-editor.org/info/rfc9421/#name-path

rfc/rfc9421/base.go
```diff
@@ -0,0 +1,173 @@
...
+
+	case "@authority":
+		host := req.Host
+		if host == "" && req.URL != nil {
+			host = req.URL.Host
+		}
+		if host == "" {
+			return "", fmt.Errorf("%w: host is empty for @authority", ErrMissingComponent)
+		}
+		return strings.ToLower(host), nil
```
a4180p, line 66: https://www.rfc-editor.org/info/rfc9421/#name-authority says that default port is ommited.
so, it host have :443 for https or :80 for http - it should be removed.

filters/auth/rfc9421.go
szuecs: I would let it like this in the filter and change the documentation that it only works for `@host` if option `-proxy-preserve-host=true` is set.

shardool-patil: I am currently working on the suggestions so which approach should i go with @szuecs @a4180p

szuecs: I talked with @a4180p and we agreed that I write this comment. In case there is someone that needs to rewrite the `@host` we can do some more unclean way of having a filter (basically move the code to the proxy, but I do not expect that we need to do this so it's better that the code is as it is.

szuecs:
:+1:

MustafaSaber:
I didn't go through the code changes yet, but I think having the filter name as the rfc is not a good approach maybe we can have a better name and just document in docs that it's the implementation of the RFC

config/config.go
```diff
@@ -657,6 +654,8 @@ func NewConfig() *Config {
 	flag.Int64Var(&cfg.OpenPolicyAgentMaxRequestBodySize, "open-policy-agent-max-request-body-size", openpolicyagent.DefaultMaxRequestBodySize, "Maximum number of bytes from a http request body that are passed as input to the policy")
 	flag.Int64Var(&cfg.OpenPolicyAgentRequestBodyBufferSize, "open-policy-agent-request-body-buffer-size", openpolicyagent.DefaultRequestBodyBufferSize, "Read buffer size for the request body")
 	flag.Int64Var(&cfg.OpenPolicyAgentMaxMemoryBodyParsing, "open-policy-agent-max-memory-body-parsing", openpolicyagent.DefaultMaxMemoryBodyParsing, "Total number of bytes used to parse http request bodies across all requests. Once the limit is met, requests will be rejected.")
+	flag.StringVar(&cfg.RFC9421KeyFile, "rfc9421-key-file", "", "path to the secret key file for RFC 9421 HTTP Message Signatures")
```
MustafaSaber: I expect to see those in use for the example in egress.md

filters/auth/rfc9421.go
```diff
@@ -0,0 +1,152 @@
...
+
+	componentsStr, ok := args[1].(string)
+	if !ok || strings.TrimSpace(componentsStr) == "" {
+		return nil, filters.ErrInvalidFilterParameters
+	}
+
+	label := "sig1"
+	if len(args) == 3 {
+		customLabel, ok := args[2].(string)
+		if !ok || strings.TrimSpace(customLabel) == "" {
```
MustafaSaber: please TrimSpace from customLabel once, you also don't need the ok in this case since value will be empty if cast fail anyway

```go
        customLabel, _ := args[2].(string)
        customLabel := strings.TrimSpace(customLabel)
		if customLabel == "" {
        ...
        }
        lable = customLabel
```

shardool-patil:
> I didn't go through the code changes yet, but I think having the filter name as the rfc is not a good approach maybe we can have a better name and just document in docs that it's the implementation of the RFC

i havent thougt about a particular name for the feature, if anyone has suggestions and everyone agrees we can change the name to that
@szuecs @a4180p

filters/auth/rfc9421.go
```diff
@@ -0,0 +1,152 @@
...
+		algorithm:     algorithm,
+		components:    components,
+		label:         label,
+		secretsReader: s.secretsReader,
+	}, nil
+}
+
+// Request signs the outgoing request with support for dynamic key rotation.
+func (f *rfc9421Filter) Request(ctx filters.FilterContext) {
+	if f.secretsReader == nil {
```
MustafaSaber: This seems to be out of place, it's required config if it should fail it should fail at filter creation

filters/auth/rfc9421_test.go
```diff
@@ -0,0 +1,329 @@
...
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/zalando/skipper/filters"
+	"github.com/zalando/skipper/filters/filtertest"
+	"github.com/zalando/skipper/rfc/rfc9421"
+)
+
+type mockSecretsReader struct {
```
MustafaSaber: There is already testSecretReader, did you check if it covers what you need?

shardool-patil:
> I didn't go through the code changes yet, but I think having the filter name as the rfc is not a good approach maybe we can have a better name and just document in docs that it's the implementation of the RFC

Looking at existing filters like `jwtValidation` and `awsSigv4`, how about **`httpSignature`** (or **`httpMessageSignature`**), it clearly describes what the filter does while referencing the standard (RFC 9421: HTTP Message Signatures), 
what do you guys think?
once the name is finalized i can push the code, i have already handled the suggestions, waiting for confirmation

filters/auth/rfc9421_test.go
shardool-patil: I wasn't aware of `testSecretsReader` earlier, but I checked it out and it completely covers what's needed (including testing key rotation). I've switched over to using it, removed `mockSecretsReader` and also handled the other suggestions, so take a look at them

szuecs:
> httpMessageSignature

Sounds good to me

shardool-patil:
i have updated the code according to @MustafaSaber suggestions and also updated the filter name to `httpMessageSignature`, so take a look

filters/auth/rfc9421.go
```diff
@@ -0,0 +1,150 @@
...
+// NewHTTPMessageSignature creates a new filter specification for RFC 9421 HTTP Message Signatures.
+func NewHTTPMessageSignature(keyFile, keyID string, secretsReader secrets.SecretsReader) filters.Spec {
+	return &httpMessageSignatureSpec{
+		keyFile:       keyFile,
+		keyID:         keyID,
+		secretsReader: secretsReader,
+	}
+}
+
+// NewRFC9421 is an alias for NewHTTPMessageSignature for backwards compatibility.
```
MustafaSaber: Why backward compatibility? This function is not merged yet and not any use at all, let's clean the old name and also clean it from functions and flags

rfc/rfc9421/doc.go
```diff
@@ -0,0 +1,4 @@
+/*
+Package rfc9421 provides RFC 9421 HTTP Message Signatures support.
+*/
+package rfc9421
```
MustafaSaber, line 4: Would also rename the package and everything

szuecs: @MustafaSaber the RFC has more than signature, I think the rfc as package is fine.

filters/auth/rfc9421.go
shardool-patil: @MustafaSaber i initially added compatibility aliases thinking ahead about future scope, but since the feature hasn't been merged or released yet, there are no existing consumers to break. 
I've cleaned the old names and also updated the flags and config options.

rfc/rfc9421/base.go
```diff
@@ -0,0 +1,173 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+// ErrMissingComponent is returned when a required component is missing from the HTTP request.
+var ErrMissingComponent = errors.New("rfc9421: missing required component")
```
MustafaSaber: Let's start with private access

MustafaSaber (commented):
Thanks! I don't have other comments, but there are some valid comments from @a4180p please check so we can merge

shardool-patil:
> Thanks! I don't have other comments, but there are some valid comments from @a4180p please check so we can merge

Done, updated to private access also checked internal refs and also handled comments by @a4180p

rfc/rfc9421/base.go
szuecs: done

szuecs (approved):

docs/reference/filters.md
```diff
@@ -2548,6 +2479,42 @@ This filter reads the body in memory. This is needed to generate signature as pe
...
 
 
+### httpMessageSignature
+
+Signs outgoing HTTP requests according to [RFC 9421 (HTTP Message Signatures)](https://www.rfc-editor.org/rfc/rfc9421). It computes the signature across specified HTTP components and attaches the `Signature-Input` and `Signature` headers. The signing key and key ID are configured globally via command line flags, and key material is loaded dynamically via the configured secrets reader (set via `-credentials-paths`) to support key rotation.
+
+Parameters:
+
+* `Algorithm` (string): Cryptographic signing algorithm. Supported: `"hmac-sha256"`, `"rsa-pss-sha512"`, `"rsa-v1_5-sha256"`, `"ecdsa-p256-sha256"`, `"ed25519"`.
+* `Components` (string): Comma-separated list of HTTP fields and derived components to cover in the signature (e.g., `"@method, @path, @authority, content-type"`).
```
a4180p, line 2489: @shardool-patil please, make it clear in the docs that this implementation isn't full and do not support all derived components.
As far as I see @query-param is missing https://www.rfc-editor.org/info/rfc9421/#name-derived-components 
There is also no support of signature component parameters https://www.rfc-editor.org/info/rfc9421/#section-6.5.2

nit: if you use llm to assist your coding, you can ask it if your code matches the rfc ;)

shardool-patil: Done, you can take a look.

a4180p:
👍

MustafaSaber (approved):

szuecs:
Thanks @shardool-patil for your dedication on the topic

PR #4217, opened 2026-08-24
fix: deduplicate fifoWithBody and lifoWithBody in the route preprocessor

AshrafAhmed9:
`Registry.Do` keys the queue by route id, so `fifo()` and `fifoWithBody()` on the
same route share one queue, same for `lifo()` and `lifoWithBody()`. The
preprocessor drops all but the last instance to avoid that, but it only counted
`filters.FifoName` and `filters.LifoName`, so the WithBody variants were never
removed. Both then call `Wait()` on the same queue and one request takes two
slots.

A single request to an idle proxy rejects itself:

```bash
./bin/skipper -inline-routes='r: * -> fifoWithBody(1, 0, "1s") -> fifoWithBody(1, 0, "1s") -> "http://127.0.0.1:9001"'

curl -s -o /dev/null -w '%{http_code}\n' localhost:9090
503
```

`lifoWithBody` twice blocks for the timeout and gives 502, and `fifo()` followed
by `fifoWithBody()` gives 503 too. With `fifo()` twice it's 200, since that pair
already gets deduplicated.

Counting per queue family instead of per filter name. `lifoGroup()` stays out of
it, that one is keyed by group name.

## How to test

```
go test ./scheduler/... ./filters/scheduler/...
```

Four cases added to `TestRegistryPreProcessor`. On master:

```
--- FAIL: TestRegistryPreProcessor/two_fifoWithBody
--- FAIL: TestRegistryPreProcessor/two_lifoWithBody
--- FAIL: TestRegistryPreProcessor/fifo_and_fifoWithBody_share_a_queue
--- FAIL: TestRegistryPreProcessor/lifo_and_lifoWithBody_share_a_queue
```

ref #4188

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -542,15 +542,28 @@ func (r *Registry) PreProcessor() routing.PreProcessor {
 
 type registryPreProcessor struct{}
 
+// isFifoFilter and isLifoFilter report whether the filter shares the per route
+// queue that Registry.Do assigns. fifo() and fifoWithBody() use the same queue,
+// as do lifo() and lifoWithBody(), so instances have to be counted and removed
+// per family rather than per filter name. lifoGroup() is not included, it is
+// keyed by group name instead of route id.
+func isFifoFilter(name string) bool {
+	return name == filters.FifoName || name == filters.FifoWithBodyName
+}
+
+func isLifoFilter(name string) bool {
+	return name == filters.LifoName || name == filters.LifoWithBodyName
+}
+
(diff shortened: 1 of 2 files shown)


AshrafAhmed9:
@szuecs could you add the `bugfix` (or `minor`) label? COMPR/pr-labels needs one
and I can't set it from a fork.

Found this while looking at what else was left around the WithBody filters after
#4188 — the preprocessor predates them and only knew about `fifo` and `lifo`, so
the variants were never deduplicated and two of them on a route share the queue.

szuecs (approved):

a4180p:
👍

PR #4218, opened 2026-08-24
fix: linter SA1019 error that shows Go stdlib deprecations

szuecs:
We have net package error.Temporary in use, which is the last we need to get rid of


MustafaSaber (approved):

szuecs:
:+1:

config/config.go
```diff
@@ -1250,6 +1248,11 @@ func (c *Config) ToOptions() skipper.Options {
 		options.EditRoute = append(options.EditRoute, eskipEdit)
 	}
 
+	if c.MetricsListener != "" && c.SupportListener == "" {
+		log.Warn("Deprecated use of -metrics-listener found, please use -support-listener instead")
```
AlexanderYastrebov, line 1252: I remember I did https://github.com/zalando/skipper/pull/1807 to log deprecated flags. Maybe it could be improved to detect deprecated flags from the `*Deprecated*` marker in the flag description...

szuecs: interesting, but it does not support library users, but maybe we should somehow support this, too.
But on the other hand if I don't remember than maybe it does not make sense to have it generic.

PR #4223, opened 2026-08-25
Avoid double parsing in routegroups loading

a4180p:
We parse routegroup filters and predicates twice - during validation and later in converter.
The idea is to cache parsing result during validation and reuse in conversion phase.

Benchstat on the data taken from one of the real clusters
```sh
                          │ /tmp/dump-old.txt │         /tmp/dump-new.txt          │
                          │      sec/op       │   sec/op     vs base               │
RouteGroupsLoadAllDump-10         344.5m ± 2%   153.0m ± 2%  -55.60% (p=0.002 n=6)

                          │ /tmp/dump-old.txt │          /tmp/dump-new.txt          │
                          │       B/op        │     B/op      vs base               │
RouteGroupsLoadAllDump-10        206.6Mi ± 0%   137.0Mi ± 0%  -33.71% (p=0.002 n=6)

                          │ /tmp/dump-old.txt │         /tmp/dump-new.txt          │
                          │     allocs/op     │  allocs/op   vs base               │
RouteGroupsLoadAllDump-10         4.669M ± 0%   1.478M ± 0%  -68.33% (p=0.002 n=6)
```

### DIFF
--- a/dataclients/kubernetes/definitions/routegroups.go
+++ b/dataclients/kubernetes/definitions/routegroups.go
@@ -63,6 +63,55 @@ type RouteGroupSpec struct {
 	// TLS specifies the list of Kubernetes TLS secrets to
 	// be used to terminate the TLS connection
 	TLS []*RouteTLSSpec `json:"tls,omitempty"`
+
+	// parsedFilters and parsedPredicates cache the result of parsing a filter or
+	// predicate definition string, keyed by that string. Because the same strings
+	// repeat across the routes of a RouteGroup, validation and conversion parse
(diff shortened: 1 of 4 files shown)


szuecs:
Captured on a cluster that scaled out routesrv to max caused by cpu:

image

the opportunity is quite low (less than 5%).
If we can omit eskipbytes.formatAndSet() it would be more interesting, because it uses 31%.

Maybe the GC can be better with your change

image

szuecs:
:+1:

a4180p:
👍

szuecs (approved):

PR #4224, opened 2026-08-25
feat(auth): add oauthOidcAnyClaimsDropRegexp filter

rraraujo:
Ref #4225

Add a named variant of `oauthOidcAnyClaims` that removes matching string values from one configured JSON array claim before serializing the session cookie.

## Motivation

Large OIDC claims (e.g. Azure AD group lists with hundreds of entries) produce oversized `skipperOauthOidc*` cookies that exceed browser limits. This filter prunes irrelevant values at callback time, before cookie compression and encryption.

## Changes

- New filter `oauthOidcAnyClaimsDropRegexp` with 12 string arguments (the standard 10 + claim name + Go regexp)
- Pruning runs after distributed claims resolution, covering both direct ID-token claims and Azure AD paginated groups
- Only JSON array claims are pruned; missing or non-array claims are a no-op
- Claim key is preserved (empty array) when all values match, so authorization checks remain valid
- Full test coverage: constructor, CreateFilter validation, pruning helper, cookie-name differentiation, and end-to-end OIDC flow
- Filter reference documentation and fuzz dictionary entry

## Testing

```
GOWORK=off go test ./filters/auth/  # all pass
make fmt                             # clean
GOWORK=off go vet ./...              # clean
```

`make lint` could not be run; see #4227.

### DIFF
--- a/filters/auth/oidc.go
+++ b/filters/auth/oidc.go
@@ -13,6 +13,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -86,6 +87,8 @@ const (
 	paramUpstrHeaders
 	paramSubdomainsToRemove
 	paramCookieName
+	paramDropClaimName
+	paramDropClaimValueRegexp
 )
 
 type OidcOptions struct {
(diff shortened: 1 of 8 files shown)


szuecs:
Can you please post the make lint error message into a separate issue?
Thanks 🙏🏽

filters/auth/oidc.go
```diff
@@ -107,28 +110,31 @@ const oidcSecretRefPrefix = "secretRef:"
...
 	tokenOidcSpec struct {
-		typ             roleCheckType
-		SecretsFile     string
-		secretsRegistry secrets.EncrypterCreator
-		options         OidcOptions
+		typ               roleCheckType
+		SecretsFile       string
+		secretsRegistry   secrets.EncrypterCreator
+		options           OidcOptions
+		dropRegexpVariant bool
```
szuecs: This should be a different typ.

szuecs (changes requested):

rraraujo:
Switched to a dedicated `checkOIDCAnyClaimsDropRegexp` roleCheckType and removed the bool. Authorization behaviour is unchanged: the new typ shares the `validateAnyClaims` path and the same callback handling as `checkOIDCAnyClaims`.

Lint error filed as #4227.

filters/auth/oidc.go
rraraujo: Done in c99b56d4. Added `checkOIDCAnyClaimsDropRegexp` to roleCheckType and removed the bool.

szuecs (approved):

MustafaSaber (approved):

PR #4240, opened 2026-09-02
Override request header value when policy provides one

Pushpalanka:
Overrides the client sent request header when OPA policy is also setting the same. Previous behavior was appending the client sent value.


szuecs (approved):

Pushpalanka:
👍

PR #4239, opened 2026-09-02
cache: use mockMetrics instead of the no-op implementation

MustafaSaber:
see subject

follow up on https://github.com/zalando/skipper/pull/4033

PR #4241, opened 2026-09-02
cache: do the cleanup inside synctest

MustafaSaber:
We did the init inside synctest but made the cleanup in the creation function with `t.Cleanup` which was trying to close routines at the end of the test function (e.g. after synctest) and try to close channels created inside synctest which already finished which cause the panic

see https://github.com/zalando/skipper/pull/4239


szuecs (approved):

greeshma1196:
👍🏽

PR #4246, opened 2026-09-04
Read the HeaderValueOption if passed and overwrite or append accordingly

Pushpalanka:
Current behavior is kept as default and no change to the existing behavior when option is not passed.

Keep the behavior consistent with Envoy external authorization header behavior, as taken to attention with https://github.com/zalando/skipper/security/advisories/GHSA-wvv5-jv5r-xq52


szuecs:
@Pushpalanka see `make fmt`

Pushpalanka:
> * pipeline

Thanks a lot for the hint.

szuecs (approved):

Pushpalanka:
👍

PR #4247, opened 2026-09-04
otel: Test behaviour with different ordering of otel propagators

madumalt:
TestPropagatorExtractMalformedTraceparent documents a trace-disconnection bug in skipper's OTel propagator chain.

### Trigger:
  A request arrives with a traceparent whose 128-bit trace-id is a 64-bit
  ottrace id left-padded with 16 zero nibbles. This is spec-valid (SDK
  only rejects all-zero ids), so tracecontext accepts it.

### Root cause:
  With the current chain [tracecontext, ottrace, b3multi, baggage], the
  last propagator to write a valid SpanContext wins. b3multi runs last
  and overwrites the ottrace/W3C context with an unrelated b3 trace-id
  injected by an intermediate Finagle hop.

### Symptom:
  Because skipper's sampler is always_on, spans are still exported - but
  under the b3 trace-id, so the trace appears disconnected in the
  observability UI.

### Test scope:
  Three configs are compared — (1) current prod (reproduces bug),
  (2) reorder-only workaround keeping b3multi,
  (3) target config dropping b3multi, matching the pre-OTel LightStep
  (ottrace-only) baseline.

### DIFF
--- a/otel/otel_test.go
+++ b/otel/otel_test.go
@@ -2,8 +2,12 @@ package otel
 
 import (
 	"context"
+	"net/http"
 	"os"
 	"testing"
+
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/trace"
 )
 
 func TestOtel(t *testing.T) {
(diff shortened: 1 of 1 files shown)


szuecs (approved):

madumalt:
👍

PR #4258, opened 2026-09-10
fix: modRequestHeader and modResponseHeader drop all but the first header value

kevin9327:
## Symptom

`modResponseHeader` deletes cookies. Rewriting the cookie domain of a backend that sets more than one cookie:

```
r: * -> modResponseHeader("Set-Cookie", "Domain=internal\\.example\\.org", "Domain=example.org") -> "http://backend"
```

Backend response:

```
Set-Cookie: a=1; Domain=internal.example.org
Set-Cookie: b=2; Domain=internal.example.org
```

What the client gets through the proxy:

```
client received Set-Cookie: ["a=1; Domain=example.org"]
```

The second cookie is gone. Expected:

```
client received Set-Cookie: ["a=1; Domain=example.org" "b=2; Domain=example.org"]
```

`modRequestHeader` does the same to the request, so a route rewriting a repeated request header forwards only the first value.

The documentation says the filter replaces "all matched regex expressions in the given header" and says nothing about the other values of that header being discarded.

## Cause

`filters/builtin/mod_header.go`:

```go
resp.Header.Set(f.headerName, f.rx.ReplaceAllString(resp.Header.Get(f.headerName), f.replacement))
```

`Header.Get` returns the first value only, and `Header.Set` replaces the whole entry with the single value it is given, so everything after the first value is dropped whether or not the expression matched it.

## Fix

Apply the expression to every value of the header, in place. The `Host` branch of `modRequestHeader` is untouched.

## Reproduction

`go1.27.1 windows/amd64`. Before, with `filters/builtin/mod_header.go` at master and only the test of this PR applied:

```
$ go test ./filters/builtin -run TestModHeaderMultipleValues -v
=== RUN   TestModHeaderMultipleValues/request
    mod_header_test.go:249: failed to modify all request header values, got: ["public.example.org"], want: ["public.example.org" "public.example.com"]
=== RUN   TestModHeaderMultipleValues/response
    mod_header_test.go:269: failed to modify all response header values, got: ["a=1; Domain=example.org"], want: ["a=1; Domain=example.org" "b=2; Domain=example.org"]
--- FAIL: TestModHeaderMultipleValues (0.00s)
    --- FAIL: TestModHeaderMultipleValues/request (0.00s)
    --- FAIL: TestModHeaderMultipleValues/response (0.00s)
```

The `Set-Cookie` output at the top of this description comes from a throwaway `proxytest` route with the same filter, run once against master and once against this branch, so the symptom is the proxy's and not just the filter's.

After:

```
$ go test ./filters/builtin -run 'TestModHeaderMultipleValues|TestModRequestHeader|TestModResponseHeader|TestModRequestHostHeader|TestModifyHost|TestCreateModHost' -count=3
ok  	github.com/zalando/skipper/filters/builtin

$ go test ./filters/builtin
ok  	github.com/zalando/skipper/filters/builtin	2.165s
```

All existing `TestModRequestHostHeader`, `TestModRequestHeader` and `TestModResponseHeader` cases pass before and after; they only ever use single-valued headers, which is why this went unnoticed.

## Notes

- `gofmt -l -s` and `go vet ./filters/builtin` are clean.
- `-race` needs cgo and there is no C toolchain on this machine, so CI has to cover that.

### DIFF
--- a/filters/builtin/mod_header.go
+++ b/filters/builtin/mod_header.go
@@ -65,11 +65,14 @@ func (f *modRequestHeader) Request(ctx filters.FilterContext) {
 		return
 	}
 
-	if _, ok := req.Header[http.CanonicalHeaderKey(f.headerName)]; !ok {
+	values, ok := req.Header[http.CanonicalHeaderKey(f.headerName)]
+	if !ok {
 		return
 	}
 
-	req.Header.Set(f.headerName, f.rx.ReplaceAllString(req.Header.Get(f.headerName), f.replacement))
+	for i, value := range values {
+		values[i] = f.rx.ReplaceAllString(value, f.replacement)
+	}
 }
 
 func (*modRequestHeader) Response(filters.FilterContext) {}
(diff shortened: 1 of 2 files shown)


szuecs (approved):

a4180p:
👍

PR #4262, opened 2026-09-14
Add RoutePredicates and RouteFilters to FilterContext

xavivars:
Expose read-only lazy-loaded copies of the predicate and filter definitions of the matched route via FilterContext. Each call returns a deep copy so mutations cannot affect the shared routing table. If no filters calls them, there's no copying happening

RoutePredicate and RouteFilter are defined as value types in the filters package to avoid an import cycle (filters -> eskip is not possible due to a transitive dependency back through filters/accesslog).


szuecs:
> Expose read-only lazy-loaded copies of the predicate and filter definitions of the matched route via FilterContext. Each call returns a deep copy so mutations cannot affect the shared routing table. If no filters calls them, there's no copying happening
> 
> RoutePredicate and RouteFilter are defined as value types in the filters package to avoid an import cycle (filters -> eskip is not possible due to a transitive dependency back through filters/accesslog).

I understand why we need this technically but what is reason of having the code?

xavivars:
There are certain things we need to know from the predicates in filters that having this woudl be much much easier.

For example, we have some predicates driving experiments (this is an extremely simplified version of the routes)

```
Path('/foo') && Host('example.com') -> 
    -> https://backend-default.com;

Path('/foo') && Host('example.com') && Experiment('an_experiment_identifier_usually_quite_long', false) -> 
    logExperiment('an_experiment_identifier_usually_quite_long')
    -> https://backend-default.com;

Path('/foo') && Host('example.com') && Experiment('an_experiment_identifier_usually_quite_long', true) -> 
    logExperiment('an_experiment_identifier_usually_quite_long')
    -> https://backend-experiment.com;
```

If we had access, we could have logExperiment being just `logExperiment()`

We have many other similar cases, with filters to modify headers based on PathPrefix's parameter, for example, and we have to duplicate the input right now. 

And with hundreds of thousands of routes, it makes a difference on route-file-size.

Does that answer your questoin?

MustafaSaber:
> If we had access, we could have logExperiment being just `logExperiment()`

I think you can use stateBag to achieve same behavior

szuecs:
stateBag is not accessible in predicates, because predicates don't know if they will be in the best matching route.
You can use preProcessors to rewrite the input to a filter based on the input of the predicate.
Basically you allow foo() and your predicate Bar("x") in tge route. The PreProcessor has acces to it and can rewrite the route.

xavivars:
Yes, but preprocessing (iterating over) half a million routes "just in case" is definitely not a cheap operation 😅

szuecs:
If you want to offload you can invest into routesrv. We run PreProcessors in routesrv and offload the data plane.

xavivars:
Yeah, we discussed that few months ago in slack, but that wouldn't change significantly the problem: downloading half a million routes from routesv that are likely to change every few minutes creates cpu spikes, and is no different from what we were trying to avoid: redundant filters that the only thing they do is "repeat" the information that's already in the predicated.

Because that's what we really need in some scenarios: to know "which route" (which predicates) was the one that matched the current request, and it seems much easier to expose that information than to add a bunch of companion filters.

What is the problem you see with the proposed approach?

szuecs:
> Yeah, we discussed that few months ago in slack, but that wouldn't change significantly the problem: downloading half a million routes from routesv that are likely to change every few minutes creates cpu spikes, and is no different from what we were trying to avoid: redundant filters that the only thing they do is "repeat" the information that's already in the predicated.

I think it's not easy to think about the impact if you don't have the data.
We once added a cache sync.Map to lookup `token->parsedValidatedMap` for 1 minute and it was reducing the total cpu time by 50%.
If we would have imagined the impact we would have done it two years earlier. 

> 
> Because that's what we really need in some scenarios: to know "which route" (which predicates) was the one that matched the current request, and it seems much easier to expose that information than to add a bunch of companion filters.

Maybe you need just one filter if it's always similar.
Do you need to know the predicates or the route.ID?
Do you need it to investigate or for an automated decision handling the request processing?

> 
> What is the problem you see with the proposed approach? 

Architectural separation is the problem.

xavivars:
Oh, we have measured the impact of such route-refresh pattern, and that's why we have now multiple different route files, grouping routes that are likely to change together into the same dataclient/file, so if none of the routes of a given subset changes, we get a 304 straight away. Having it on a single file (or on a single routesrv) was creating 1.5minutes long CPU spikes.

In general, for logging and investigating things, route.Id is more than enough (and that one was added a few months back). 

But knowing the predicates would allow us to avoid iterating over all routes every time they get loaded to do something like:

```
// Inspect each route for InABTest predicates, if found adds an setExperiment filter to add log information
func (spec experimentPreprocessor) Do(routes []*eskip.Route) []*eskip.Route {
	inABTestPredicateName := vppredicates.NewInABTest().Name()
	filterName := vpfilters.NewExperimentFilter().Name()
	for _, route := range routes {
		if route.Predicates != nil {
			experimentAndVariations := []string{}
			for _, predicate := range route.Predicates {
				if predicate.Name == inABTestPredicateName {
					experimentAndVariations = append(experimentAndVariations, predicate.Args[0].(string)+":"+predicate.Args[1].(string))
				}
			}
			if len(experimentAndVariations) > 0 {
				args := make([]interface{}, 1)
				args[0] = strings.Join(experimentAndVariations, "|")
				filter := &eskip.Filter{Name: filterName, Args: args}
				route.Filters = append(route.Filters, filter)
			}
		}
	}
	return routes
}
```

PR #4263, opened 2026-09-14
doc: missing letsencrypt user docs

szuecs:
docs: fixed old versions in readme
docs: added some important features to the readme


docs/operation/operation.md
```diff
@@ -2067,3 +2067,45 @@ flowchart TD
 
 - `cache_status`: `"hit"`, `"miss"`, or `"stale"`
 - `cache_ttl_remaining_ms`: remaining freshness in milliseconds (only set on hits)
+
+## Letsencrypt - autocert
+
+Letsencrypt let's you create and automate TLS certificates.
+Skipper can be used to automate TLS certificate on the first request.
+
+Example configuration flags:
```
MustafaSaber, line 2076: Can you provide an example how you use it after enabling the flags? Some example of the use case and a setup (a graph maybe) would be good not just the flags.

szuecs: Do you mean systemd unit and directory or what do you think?
What kind of graph would be helpful?
Basically you need to add the flags and it works. :)

MustafaSaber: We discussed in SUP, we can add later a tutorial after testing this on some server as locally it can't be tested.

MustafaSaber:
👍

a4180p:
👍

PR #4269, opened 2026-09-17
test(proxylistener): fix flaky TestProxyListenerWithProxyClient

### LINKED ISSUE #4070: FlakyTest TestProxyListenerWithProxyClient
```
2026/06/15 21:46:11 Unsolicited response received on idle HTTP channel starting with "HTTP/1.1 400 Bad Request\r\nContent-Type: text/plain; charset=utf-8\r\nConnection: close\r\n\r\n400 Bad Request"; err=<nil>
--- FAIL: TestProxyListenerWithProxyClient (0.01s)
    --- FAIL: TestProxyListenerWithProxyClient/test_skip_list (0.00s)
        proxylistener_test.go:210: Start shutdown
        proxylistener_test.go:236: Failed to get response: Post "http://[::]:42805/foo": readLoopPeekFailLocked: %!w(<nil>)
FAIL
```

ref: https://github.com/golang/go/issues/31259#issuecomment-925708748
ref: https://caddy.community/t/got-502-with-an-error-message-readlooppeekfaillocked-nil/13616/6

shardool-patil:
### Root Cause
In `TestProxyListenerWithProxyClient/test_skip_list`, the listener is configured with `SkipListCIDRs`, which intentionally skips parsing the PROXY protocol header. When `createProxyClient` dials in, it immediately sends a binary PROXY v2 header directly into `http.Server`.
Because the incoming bytes are not a valid HTTP request line, `http.Server` rejects them with `HTTP/1.1 400 Bad Request` and `Connection: close` before terminating the TCP connection. When using Go's standard `http.Client.Do` on a connection that is abruptly closed while client-side write/read synchronization is in flight, `net/http` hits a known transport race condition (ref: golang/go#31259), returning `readLoopPeekFailLocked` and failing the test.

### Changes
* Replaced `http.Client.Do` in `test_skip_list` with a direct TCP exchange (`executeSkipListRequest`) using `http.ReadResponse`, completely bypassing `http.Transport`'s background `readLoop` race on early connection resets.
* Added `io.Copy(io.Discard, rsp.Body)` before closing response bodies to properly drain remaining socket buffers.
* Added `defer client.CloseIdleConnections()` to prevent lingering idle pooled connections from racing with server shutdown.

### Related Issues
* Closes #4070


proxylistener/proxylistener_test.go
```diff
@@ -105,6 +106,50 @@ func createBogusProxyClient(proxyAddr, destAddr string, destPort int, version by
...
+			Port: 8080,
+		},
+	}
+
+	if _, err := header.WriteTo(conn); err != nil {
+		return nil, err
+	}
+
+	reqStr := fmt.Sprintf("GET /foo HTTP/1.1\r\nHost: %s\r\nConnection: close\r\n\r\n", host)
+	if _, err := io.WriteString(conn, reqStr); err != nil {
```
szuecs: You also use `Fprintf(w io.Writer, format string, a ...any)`, so:
```go
if _, err := fmt.Fprintf(conn, "GET /foo HTTP/1.1\r\nHost: %s\r\nConnection: close\r\n\r\n", host); err != nil {
```

szuecs:
:+1:

proxylistener/proxylistener_test.go
shardool-patil: Done, ill make sure to remember to use similar syntax in future as well

szuecs (approved):

MustafaSaber (approved):

PR #4271, opened 2026-09-18
doc: clarify ingress annotations

szuecs:
thanks to @euriconicacio highlighting the problem with `spec.defaultBackends` vs `spec.rules` created routes.


MustafaSaber (approved):

MustafaSaber:
`spec.defaultBackend` is routegroup thingy but those annotations work in ingress, I don't see how they are related.

szuecs:
> `spec.defaultBackend` is routegroup thingy but those annotations work in ingress, I don't see how they are related.

no RouteGroup uses a completely different idea for "defaultBackend".

see
```
% kubectl explain ing.spec.defaultBackend
GROUP:      networking.k8s.io
KIND:       Ingress
VERSION:    v1

FIELD: defaultBackend <IngressBackend>


DESCRIPTION:
    defaultBackend is the backend that should handle requests that don't match
    any rule. If Rules are not specified, DefaultBackend must be specified. If
    DefaultBackend is not set, the handling of requests that do not match any of
    the rules will be up to the Ingress controller.
    IngressBackend describes all endpoints for a given service and port.
    
FIELDS:
  resource      <TypedLocalObjectReference>
    resource is an ObjectRef to another Kubernetes resource in the namespace of
    the Ingress object. If resource is specified, a service.Name and
    service.Port must not be specified. This is a mutually exclusive setting
    with "Service".

  service       <IngressServiceBackend>
    service references a service as a backend. This is a mutually exclusive
    setting with "Resource".
```

szuecs:
:+1:

PR #4281, opened 2026-09-21
skipper: add component where startup error occurred

vlktna:
add component where startup error occurred


skipper.go
```diff
@@ -1872,11 +1872,18 @@ func getRemoteURLShardAddrUpdater(address string) func() ([]string, error) {
 	}
 }
 
+func wrapError(err error, component string) error {
+	if err == nil {
+		return nil
+	}
```
a4180p, line 1878: i don't think we need this check. 
this function is always called inside of `if err`

szuecs:
I don't like it.

